Issue #2281, opened 2022-08-22
Test Fixtures: Tags, Status


bryanculver:

glennmatthews:
Implemented in #2300 for 1.5.

PR #2300, opened 2022-08-24
Fixtures infra, add and use fixtures for Tag and Status

glennmatthews:
# Closes: #2281 
# What's Changed

- Add `invoke write-fixture` and `invoke load-fixture` tasks
   - Document `load-fixture` in the development getting-started guide
   - Create a new `docs/development/testing.md` file, move some of the content from getting-started that didn't belong there into this file, and add details about test fixture design and usage as well as some other test best practices.
- Create fixtures for `tag` and `status` models
   - Status fixture includes all default statuses, plus a couple of custom statuses that can be used for all content-types
   - Tag fixture includes a number of custom tags, most of which apply to all content-types for convenience in testing
- Add status fixture to (hopefully) all test cases that should use it
   - I identified some of these initially through grepping and code inspection
   - I identified a number of others by temporarily disabling the default creation of Statuses during migrations, running the tests, and seeing what blew up due to the default statuses not existing
   - There may be others that I've missed due to the standard statuses existing by default but this is at least a step in the right direction
- _Doing the above identified an issue in that `CircuitTermination` is flagged by `extras_features` as a model that supports statuses but it doesn't actually have a `status` field; I've removed that erroneous flag as a fix._
- Add tag fixture to many of the test cases that rely on tags
   - Again, grepping was the main approach here to identify candidates.
   - I replaced the use of our custom `create_tags()` function with the fixture wherever I found it, and added a DeprecationWarning to `create_tags` when called warning that it will go away and you should use fixtures instead.

# TODO
- [ ] Add natural-key support to extras.Tag model so that the fixture doesn't have to hard-code PKs
- [x] Unit, Integration Tests


nautobot/circuits/models.py
```diff
@@ -282,7 +282,6 @@ def to_csv(self):
     "graphql",
     "locations",
     "relationships",
-    "statuses",
```
bryanculver, line 285: Should we get a backlog item to add this as intended?

nautobot/extras/tests/test_models.py
```diff
@@ -121,8 +123,8 @@ def setUp(self):
         self.platform = Platform.objects.create(name="Platform")
         self.tenantgroup = TenantGroup.objects.create(name="Tenant Group")
         self.tenant = Tenant.objects.create(name="Tenant", group=self.tenantgroup)
-        self.tag = Tag.objects.create(name="Tag", slug="tag")
-        self.tag2 = Tag.objects.create(name="Tag2", slug="tag2")
+        self.tag = Tag.objects.get_for_model(Device)[0]
+        self.tag2 = Tag.objects.get_for_model(Device)[1]
```
bryanculver, line 127:
```suggestion
        device_tags = Tag.object.get_for_model(Device)
        self.tag = device_tags[0]
        self.tag2 = device_tags[1]
```

Curious of the performance improvement...

nautobot/utilities/testing/mixins.py
```diff
@@ -189,6 +190,13 @@ def assertInstanceEqual(self, instance, data, exclude=None, api=False):
...
         Create and return a Tag instance for each name given.
-        """
 
+        DEPRECATED: use fixtures instead if at all possible.
+        """
+        warnings.warn(
+            "create_tags() is deprecated and will be removed in a future Nautobot release. "
+            "Use fixtures such as nautobot/extras/fixtures.tags.json (provided in Nautobot 1.5 and later) instead.",
+            DeprecationWarning,
+            stacklevel=2,
```
bryanculver, line 200: Should we leave the if Python version check here since we haven't confirmed dropping 3.7 for 1.5?

bryanculver (approved):
Couple `nitpicks` but LGTM. Tests pass. Unit testing might be _slightly_ faster by the last CI run?

nautobot/circuits/models.py
glennmatthews: As in add a `status` field to CircuitTermination model? I don't see a need to do so at this time.

nautobot/utilities/testing/mixins.py
glennmatthews: The python version check is needed for `logging.warning(...,stacklevel=...)` - `warnings.warn(..., stacklevel=...)` works fine in 3.7.

nautobot/extras/tests/test_models.py
glennmatthews: How about:

```suggestion
        self.tag, self.tag2 = Tag.objects.get_for_model(Device)[:2]
```

HanlinMiao (approved):
Amazing

Issue #2273, opened 2022-08-22
Add docs linting rule to disallow relative directory links


glennmatthews:
### Proposed Changes

Per https://github.com/mkdocs/mkdocs/issues/1854, `mkdocs` only validates file-relative links (`../installation/index.md`) and by design does *not* validate directory-relative links (`../installation/`). I'm working on a PR to fix our existing documentation links to follow the validated standard. But it would be very useful to add a docs linting rule to detect and disallow directory-relative links so that they don't creep back in inadvertently.

### Justification

Let's use tooling to enforce that our documentation follows best practices and avoid broken links in our docs.

smk4664:
I am finishing up a markdownlint Custom Rule to flag these, would you mind assigning this to me?

PR #2274, opened 2022-08-22
Replace directory-relative links with file-relative links throughout docs

glennmatthews:
# Closes: #DNE
# What's Changed

By design `mkdocs` checks the validity of file-relative internal links (`../installation/index.md`, `models/extras/status.md`) but not directory-relative internal links (`../installation/`, `models/extras/status/`). This PR changes every instance I found of the latter with a couple of quick greps to follow the former, and fixes a few broken links that mkdocs reported as a result.

Note that mkdocs does *not* appear to validate that linked anchors actually exist within the linked file (`../installation/index.md#no-such-anchor` does not report an error, so long as `../installation/index.md` is a valid file path), but this at least gets us closer to the right direction.

I've also opened #2273 as a housekeeping issue to add a linting rule to explicitly disallow directory-relative links within the docs.

# TODO

- [x] Explanation of Change(s)
- n/a Attached Screenshots, Payload Example
- [x] Unit, Integration Tests
- [x] Documentation Updates (when adding/changing features)
- n/a Example Plugin Updates (when adding/changing features)
- [x] Outline Remaining Work, Constraints from Design

PR #2303, opened 2022-08-25
Markdownlint file relative

smk4664:
# Closes: #2273 
# What's Changed

Adds a new markdownlint custom rule to ensure relative links are using the file-relative instead of directory-relative path.


nautobot/docs/additional-features/napalm.md
```diff
@@ -3,7 +3,7 @@
 Nautobot supports integration with the [NAPALM automation](https://github.com/napalm-automation/napalm/) library. NAPALM allows Nautobot to serve a proxy for operational data, fetching live data from network devices and returning it to a requester via its REST API. Note that Nautobot does not store any NAPALM data locally.
 
 !!! note
-    To enable this integration, the NAPALM library must be installed. See [installation steps](../../installation/nautobot/#napalm) for more information.
+    To enable this integration, the NAPALM library must be installed. See [installation steps](../installation/nautobot.md#napalm) for more information.
```
gsnider2195, line 6:
```suggestion
    To enable this integration, the NAPALM library must be installed. See [installation steps](../installation/nautobot.md#configuring-napalm) for more information.
```

This link was broken. Might as well fix in this pr

scripts/use-relative-md-links.js
```diff
@@ -0,0 +1,54 @@
...
+
+// All external urls should have a scheme, i.e. http://, https:// or mailto:
+const externalUrl = /^[a-z0-9+.-]+:/;
+// linking within the same file is just (#<anchor-name>)
+const anchor = /^#/;
+const markdownFile = /.md/;
+const svg = /.svg/;
+
+module.exports = {
+	names: ['NAUTOBOTMD001', 'nautobotmd.filerelativelinks'],
```
gsnider2195, line 14: Can you convert the tabs to spaces in this file?

gsnider2195 (changes requested):
Just a couple minor nitpicks. Looks good!

glennmatthews:
This obsoletes #2274, correct?

.github/workflows/ci_integration.yml
```diff
@@ -45,6 +45,7 @@ jobs:
           config_file: ".markdownlint.yml"
           files: "nautobot examples *.md"
           ignore_files: "nautobot/project-static"
+          rules: "scripts/use-relative-md-links.js"
```
glennmatthews, line 48: Naive question: this is just specifying *additional* rules right? All of the baseline markdownlint rules will still be enforced?

nautobot/docs/development/index.md
```diff
@@ -42,7 +42,7 @@ The following provides more detail on these.
 * The GitHub **Release Milestone** will track items for the _current_ release.
 
 !!! note
-    Release windows and dates will be updated per [Release Management](release-management).
+    Release windows and dates will be updated per [Release Management](#release-management).
```
glennmatthews, line 45: This is literally linking to the parent section of this note. :-) I'm actually wondering if we'd do better to just remove this note, as it's really not clear to me what it's even saying.

scripts/use-relative-md-links.js
```diff
@@ -0,0 +1,54 @@
...
+					.filter(function filterChild(child) {
+						return child.type === 'link_open';
+					})
+					.forEach(function forToken(link) {
+						if (link.attrs[0] && link.attrs[0][0] === 'href') {
+							const href = link.attrs[0][1];
+
+							let isMarkdownFile = markdownFile.test(href.toLowerCase());
+							let isExternalUrl = externalUrl.test(href.toLowerCase());
+							// There is one relative link in the docs that points to an SVG file.
```
glennmatthews, line 33: We have relative references to other file types too, such as the various jpg (ew) and png images in `docs/media/`. I'm guessing those aren't checked by this rule since they're generally embedded media rather than hyperlinks, but I wonder if it would be cleaner or more compatible in general if instead of checking specifically for `.md` and `.svg` if we just checked generally for a file extension, something like `/\.[a-z]+$/`?

glennmatthews (commented):
Great stuff!

scripts/use-relative-md-links.js
smk4664: Sure, the example from markdownlint was in tabs, so I assumed tabs was a requirement.

smk4664:
> This obsoletes #2274, correct?

It is based on the work done in #2274, so if you want to combine, that would be fine.

scripts/use-relative-md-links.js
smk4664: Correct, it only looks for markdown links, it does not check images. The only reason that the .svg was flagged is that it is a link to an image file.

nautobot/docs/development/index.md
smk4664: I know... and unfortunately this has been broken since we forked from Netbox so I have no idea what this really should point to. I will just remove the note.

.github/workflows/ci_integration.yml
smk4664: Correct, https://github.com/nosborn/github-action-markdown-cli#inputs

```
files - what to process (files, directories, globs)
config_file (optional) - configuration file (JSON or YAML)
ignore_files (optional) - files to ignore/exclude (file, directory, glob)
ignore_path (optional) - path to file with ignore pattern(s)
rules (optional) - custom rule files (file, directory, glob, package)
```

scripts/use-relative-md-links.js
smk4664: I think that may not catch things we believe it should then.  I think a better option is to omit checking this rule on the svg line instead. I will see if this is possible.

glennmatthews (approved):

gsnider2195 (approved):

Issue #2237, opened 2022-08-17
object_edit and bulk_edit ignore permissions for the Note field.


smk4664:
### Environment
* Python version: 3.8
* Nautobot version: 1.4.0

### Steps to Reproduce
1. Create a user with permission to the Site model, but don't grant them extras.Note
2. Create a Site Object using the Add/Edit form
3. Notice that the Notes field is a valid option

### Expected Behavior
The Note field should note be available if the user does not have permissions to add_note.

### Observed Behavior
User is able to create a note, (but then can't view the notes on the object).

PR #2310, opened 2022-08-25
Restrict Note field to perms when editing object

smk4664:
# Closes: #2237 
# What's Changed

When editing or bulk editing an object, the forms will now enforce the user permissions. This also corrects two bugs with "note" permissions.


glennmatthews (commented):
Changes look solid. Would you be able and willing to take a stab at adding a test case or two?

glennmatthews (approved):

Issue #2292, opened 2022-08-23
Enable pylint for local linting and CI


glennmatthews:
### Proposed Changes

Add `pylint` to our standard linting and CI suite.

### Justification

Improved code quality and consistency.

glennmatthews:
Assigning myself as I have some draft diffs for this already.

itdependsnetworks:
Just noting the comment I made recently on a similar topic, my goal was more about exposing for plugins, but obviously the same applies https://github.com/nautobot/nautobot/issues/2257#issuecomment-1220684260

glennmatthews:
After #2314 and #2339, the following Pylint "error"-severity checkers are still globally disabled:

- `E1101 no-member`
- `E1102 not-callable`

I'd love to get to the point where we have all "error"-severity checkers enabled, but these two in particular seem to produce a lot of false positives, so enabling them will require adding a large number of one-off `pylint: disable` comments and explanations.

The following remaining lower-severity checkers are also currently disabled - we should evaluate these and decide which ones are desirable to enable and address.

### Warning

- `W0223 abstract-method`
- `W0237 arguments-renamed`
- `W0201 attribute-defined-outside-init`
- `W0703 broad-except`
- `W0122 exec-used`
  - I think this only warns on our CLI runner code, we could probably enable it globally and just disable it locally there
- `W0511 fixme`
  - Currently configured not to warn on "TODO" comments, but we do have a few "FIXME" and "XXX" comments that need revisiting
- `W0603 global-statement`
- `W0602 global-variable-not-assigned`
- `W1113 keyword-arg-before-vararg`
- `W1202 logging-format-interpolation`
- `W1203 logging-fstring-interpolation`
- `W0104 pointless-statement`
- `W0105 pointless-string-statement`
- `W0212 protected-access`
- `W0707 raise-missing-from`
   - Would be good practice to enable and address
- `W0127 self-assigning-variable`
- `W0222 signature-differs`
   - Definitely good to enable and address
- `W0231 super-init-not-called`
   - I think this only warns in a couple of places where we're intentionally not calling super. Spot disable?
- `W1514 unspecified-encoding`
- `W0613 unused-argument`
- `W0611 unused-import`
  - Should already be generally caught by flake8
- `W0614 unused-wildcard-import`
  - Likely to complain about a bunch of our NestedSerializer imports
- `W0401 wildcard-import`
  - Likely to complain about a bunch of our NestedSerializer imports

### Convention

[...]

PR #2314, opened 2022-08-25
Pylint part 1

### LINKED ISSUE #2198: Enable Example Plugin by Default in Developer Environment
### Proposed Changes

Enable the example plugin by default in the developer `nautobot_config.py`.

### Justification

Almost always it's the first thing enabled when starting from a clean branch or clone.

glennmatthews:
# Progress towards: #2292 
# What's Changed

- Add pylint and pylint-django as dev dependencies
- Add `invoke pylint` task and `nautobot-server pylint` command provided by the example plugin. (The latter is needed because of our custom startup code, meaning that `DJANGO_SETTINGS_MODULE=nautobot.core.settings pylint nautobot/` doesn't work on its own)
- Add example plugin to PLUGINS in the development config (closes #2198) so that the above will work
- Fix and/or spot-disable a bunch of relatively low-hanging pylint warnings throughout our code base. It *may* be easier to review these commit-by-commit rather than looking at the entire diffset at once.
  - https://github.com/nautobot/nautobot/commit/cabb17701734468b26e479191c791a561ee2a3dc bad-super-call,dangerous-default-value,expression-not-assigned,no-self-argument,trailing-comma-tuple
  - https://github.com/nautobot/nautobot/commit/70947aa0ae21bd363ef30530c0f1ef747d603de2 unbalanced-tuple-unpacking, undefined-loop-variable, unused-variable, used-before-assignment
  - https://github.com/nautobot/nautobot/commit/64298f44c01d1899de781e51197647237b97bc33 unnecessary-comprehension, unnecessary-pass, use-dict-literal, use-list-literal, useless-object-inheritance, useless-super-delegation, using-constant-test
  - https://github.com/nautobot/nautobot/commit/4a3a49fc773a6e04d3c5a89ce53ff361c91adcf1 abstract-class-instantiated, assignment-from-no-return, assignment-from-none, duplicate-key, function-redefined, implicit-str-concat
  - https://github.com/nautobot/nautobot/commit/c98694dbc570362d97bcae6bf86d360e421d93dc arguments-differ, deprecated-decorator, deprecated-method, deprecated-module, inconsistent-return-statements, invalid-envvar-default, invalid-overridden-method, invalid-str-returned
  - https://github.com/nautobot/nautobot/commit/d997a251b093b23a73eee012dc4427eb305db684 redefined-argument-from-local, redefined-builtin, redefined-outer-name, redundant-unittest-assert, reimported
- Disable all remaining current active pylint warnings against our code base so that we have a passing baseline.
  - I've got a local branch that re-enables and addresses a bunch more of these, but the diffs get increasingly large and hard to follow, and probably more controversial as well. :-) . I'll open up follow-on PR(s) with those additional changes as time and interest permits, hence why this is a "part 1".
- Make a first stab at adding pylint to CI - as it takes about 10 minutes to run, I've set it as a parallel step to the unit-tests, rather than parallel to the other, much faster linter steps.

# TODO

- n/a Unit, Integration Tests
- [x] git pre-commit hook updates


nautobot/core/runner/settings.py
```diff
@@ -8,26 +8,20 @@
...
-except NameError:  # Python3
-
-    def execfile(afile, globalz=None, localz=None):
-        with open(afile, "r") as fh:
-            exec(fh.read(), globalz, localz)
-
-
 import errno
-import imp
+import imp  # pylint: disable=deprecated-module
```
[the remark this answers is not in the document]
glennmatthews: #2318

nautobot/dcim/fields.py
```diff
@@ -70,7 +70,7 @@ def python_type(self):
         return EUI
 
     @property
-    def validators(self):
+    def validators(self):  # pylint: disable=invalid-overridden-method
```
gsnider2195, line 73: Should we change this to a `@cached_property` decorator like the parent class?

glennmatthews: Yeah, this looks like less-than-ideal code. I'll open a child PR with what I think should be the correct code. Good spot!

glennmatthews: #2332.

bryanculver (approved):

Issue #2307, opened 2022-08-25
Expose dynamic groups containing devices on the devices themselves in GraphQL


Kircheneer:
### As ...

Austin - Network Automation Engineer

### I want ...

a `dynamic_groups` field to be available in GraphQL for devices like

```
query ($device_id: ID!) {
  device(id: $device_id) {
    dynamic_groups
  }
}
```

### So that ...

I can query via GraphQL which dynamic groups a given device belongs to.

### I know this is done when...

the `dynamic_groups` field is available on the GraphQL model for devices.

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [X] Misc (including Data Sources)
- [X] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

None

### External Dependencies

None

Kircheneer:
I will try and tackle this. Should look something like [this](https://github.com/nautobot/nautobot/blob/develop/nautobot/dcim/graphql/types.py#L124).

PR #2315, opened 2022-08-26
Implement a 'dynamic_groups' field on the DeviceType for GraphQL.

Kircheneer:
# Closes: #2307

# What's Changed

- Implement a `dynamic_groups` field on the GraphQL `DeviceType` that shows to which dynamic groups the device belongs to
- Statically create `DynamicGroupType` because otherwise I would get the following error

```
Exception Value: 	

Could not import 'nautobot.core.graphql.schema_init.schema' for Graphene setting 'SCHEMA'. ImportError: Module "nautobot.extras.graphql.types" does not define a "DynamicGroupType" attribute/class.
```

Example screenshot:
Screenshot 2022-08-26 at 09 28 30

# TODO
- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/graphql/types.py
+++ b/nautobot/dcim/graphql/types.py
@@ -57,7 +57,7 @@ class DeviceType(gql_optimizer.OptimizedDjangoObjectType):
     dynamic_groups = graphene.List("nautobot.extras.graphql.types.DynamicGroupType")
 
     def resolve_dynamic_groups(self, args):
-        return DynamicGroup.objects.get_for_object(Device.objects.get(pk=self.pk))
+        return DynamicGroup.objects.get_for_object(self)
 
 
 class RackType(gql_optimizer.OptimizedDjangoObjectType):
@@ -71,7 +71,7 @@ class RackType(gql_optimizer.OptimizedDjangoObjectType):
     dynamic_groups = graphene.List("nautobot.extras.graphql.types.DynamicGroupType")
 
     def resolve_dynamic_groups(self, args):
-        return DynamicGroup.objects.get_for_object(Rack.objects.get(pk=self.pk))
+        return DynamicGroup.objects.get_for_object(self)
 
 
 class CableType(gql_optimizer.OptimizedDjangoObjectType):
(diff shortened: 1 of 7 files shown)


bryanculver (commented):
😎 

Can we see if we can get all of the existing DynamicGroup supporting models (there aren't many at this time) to support this?

As well can we get at least a test case for Device to ensure we get the results we expect?

Kircheneer:
Expanded the availability of the key to all relevant models and added a test. Let me know if you want more or more thorough testing.

nautobot/dcim/graphql/types.py
```diff
@@ -53,6 +54,11 @@ class Meta:
         filterset_class = DeviceFilterSet
         exclude = ["_name"]
 
+    dynamic_groups = graphene.List("nautobot.extras.graphql.types.DynamicGroupType")
+
+    def resolve_dynamic_groups(self, args):
+        return DynamicGroup.objects.get_for_object(Device.objects.get(pk=self.pk))
```
bryanculver, line 60: Until we refactor this as a Mixin or base, any reason why to not change to in all places?

```suggestion
        return DynamicGroup.objects.get_for_object(self.Meta.model.objects.get(pk=self.pk))
```

Just to minimize copy-paste issues.

glennmatthews: Is this level of indirection needed or does `DynamicGroup.objects.get_for_object(self)` work too?

bryanculver: > Is this level of indirection needed or does `DynamicGroup.objects.get_for_object(self)` work too?

Is `self` a `Device` object? or a shadowed one for GraphQL? I'm not familiar how the GraphQL types are built in relation to the normal models.

Kircheneer: It is not in fact needed, it works with `DynamicGroup.objects.get_for_object(self)`.

glennmatthews (approved):

bryanculver (approved):

PR #2136, opened 2022-08-03
Add slug field to CustomField model, add versioned custom-fields REST API, etc.

glennmatthews:
Closes: #1962

# What's Changed
Well, this turned into a whole ball of yarn to unravel, didn't it? A simple request in #1962 to add a `slug` field to the `CustomField` model, and use that field in the UI view URLS for custom-fields, ballooned into all of the following:

- Add a `slug` field to the `CustomField` model class.
    - Change UI URL patterns for CustomField to use `slug` rather than `name`, fixing the symptom described in #1962.
    - Add a data migration to set the `slug` for existing CustomFields.
    - As groundwork towards #824, change the edit form for CustomField to prompt for the `slug` (but see also below...) and `label` rather than the `name`.
         - `label` is still not a mandatory field on the database model, but the UI now treats it as required, and if left unset when creating a new CustomField or migrating from 1.3, it will default to copying the `name` value into `label`.
    - Add a versioned REST API serializer for `/api/extras/custom-fields/` so that in version 1.4 and later the `slug` and `label` are mandatory and the `name` is omitted.
    - As groundwork towards #824, add versioning to `CustomFieldModelSerializer` and related classes so that in version 1.4 and later the REST API represents data as `["custom_fields"][<slug>]` instead of `["custom_fields"][<name>]` - this still maps on the back end to the internal storage in `._custom_field_data[<name>]`, as changing that would be a breaking change to the ORM, but since we have a versioned REST API we can at least start the cut-over there.
- Added `TODO 2.0` comments to just about everywhere I could find that will need to be touched when we do #824 (spoiler: there are a lot of places!)
- Move the documentation files for `CustomField` and `ComputedField` from `docs/additional-features/` to `docs/models/extras/` so that they're correctly linked from the UI when editing these models.
- I observed that using dash-based slugs for `ComputedField`, `CustomField`, and `Relationship` makes for a variety of issues in GraphQL, only some of which are addressed by our usage of `str_to_var_name` in the GraphQL code. A better approach would be to require these models to use exclusively underscore-based slugs as those will not require variable name remapping. Addressing this involved:
    - Add a `slugify_dashes_to_underscores` utility function and use this for ORM-level autogeneration of slugs for these three models so that new instances will default to "safe" slugs when not explicitly specified otherwise.
    - Add an option in UI edit form templates and `forms.js` to have the UI slug fields autopopulate with underscore-based slugs instead of dash-based slugs when desired, and use this option for these three models.
[...]

PR #2317, opened 2022-08-26
Add support for grouping of Custom Fields

HanlinMiao:
All entries expanded by default
Screen Shot 2022-08-26 at 11 00 00 AM
Collapse and Expand
Screen Shot 2022-08-26 at 11 00 18 AM
Collapse All
Screen Shot 2022-08-26 at 11 01 10 AM
Expand All
Screen Shot 2022-08-26 at 11 01 33 AM


nautobot/core/templates/inc/custom_fields/panel_data.html
```diff
@@ -3,37 +3,58 @@
...
+                {% if grouping != "" %}
+                    <tr>
+                        <!-- #TODO
+                        Here we are using forloop.counter0 to guarantee the unqiueness of `data-target` attribute.
+                        This feature on the joblist page is using {{ grouping | slugify }} it should be changed as it breaks `accordion-toggle`
+                        when two distinct groups have the same slugified name. e.g. Group 2, group-2 -->
+                        <th>
+                            <button type="button" class="btn-link accordion-toggle mdi mdi-chevron-down"
+                            name="grouping.{{ grouping }}" data-toggle="collapse"
+                            data-target=".collapseme-{{ count }}">
```
glennmatthews, line 20: nit:

```suggestion
                            <button type="button" class="btn-link accordion-toggle mdi mdi-chevron-down"
                                    name="grouping.{{ grouping }}" data-toggle="collapse"
                                    data-target=".collapseme-{{ count }}">
```

nautobot/core/templates/inc/custom_fields/panel_data.html
```diff
@@ -3,37 +3,58 @@
...
+        <table id="accordion" class="table table-hover panel-body attr-table">
+            {% for grouping, fields in custom_fields.items %}
+            {% with forloop.counter0 as count %}
+                {% if grouping != "" %}
+                    <tr>
+                        <!-- #TODO
+                        Here we are using forloop.counter0 to guarantee the unqiueness of `data-target` attribute.
+                        This feature on the joblist page is using {{ grouping | slugify }} it should be changed as it breaks `accordion-toggle`
+                        when two distinct groups have the same slugified name. e.g. Group 2, group-2 -->
+                        <th>
```
glennmatthews, line 17: Is it the use of `<th>` rather than `<td>` that's making the groupings appear so large in the rendered panel? If so, consider something like `<td><strong>` instead as I think the current display is jarring.

nautobot/core/templates/inc/custom_fields/panel_data.html
```diff
@@ -3,37 +3,58 @@
...
+                            <button type="button" class="btn-link accordion-toggle mdi mdi-chevron-down"
+                            name="grouping.{{ grouping }}" data-toggle="collapse"
+                            data-target=".collapseme-{{ count }}">
+                                {{ grouping }}
+                            </button>
+                        </th>
+                        <td></td>
```
glennmatthews, line 24: Perhaps would be better to use a `colspan="2"` on the first th/td so that it just extends to the full width of the table?

nautobot/core/templates/inc/custom_fields/panel_data.html
```diff
@@ -3,37 +3,58 @@
...
+                    </tr>
+                {% endif %}
+                    {% for field, value in fields %}
+                        <tr class="collapseme-{{ count }} collapse in" data-parent="#accordion">
+                            <td><span title="{{ field.description }}">{{ field }}</span></td>
+                            <td>
+                                {% if field.type == 'boolean' %}
+                                    {{ value | render_boolean }}
+                                {% elif field.type == 'url' and value %}
+                                    <a href="{{ value }}">{{ value|truncatechars:70 }}</a>
```
glennmatthews, line 34: I know this is present in the base template too but it does seem odd to me that we are truncating the displayed URL here.

nautobot/core/templates/inc/custom_fields/panel_data.html
```diff
@@ -3,37 +3,58 @@
...
+                                    {{ value|join:", " }}
+                                {% elif field.type == 'json' and value is not None %}
+                                    <p>
+                                        <button class="btn btn-xs btn-primary" type="button" data-toggle="collapse"
+                                                data-target="#cf_{{ field.name | slugify }}" aria-expanded="false"
+                                                aria-controls="cf_{{ field.name | slugify }}">
+                                            Show/Hide
+                                        </button>
+                                    </p>
+                                    <pre class="collapse" id="cf_{{ field.name | slugify }}"><code>{{ value|render_json }}</code></pre>
```
glennmatthews, line 45: Probably can/should change these to `cf_{{ field.slug }}` now that  we have #2136 in place.

nautobot/extras/forms/forms.py
```diff
@@ -361,6 +365,7 @@ class Meta:
             "filter_logic",
             "advanced_ui",
             "content_types",
+            "grouping",
```
glennmatthews, line 368: I'd suggest moving this up, perhaps even to directly below `label`?

nautobot/extras/models/customfields.py
```diff
@@ -161,6 +161,64 @@ def get_custom_fields(self, advanced_ui=None):
...
+        for field in fields:
+            data = (field, self.cf.get(field.name))
+            # Data made before migration will have a grouping of None if not specified
+            # Data made after migration will have a grouping of "" if not specified.
+            # We put them in the same grouping
+            if field.grouping in ["", None]:
+                field.grouping = ""
+            if field.grouping not in record:
+                record[field.grouping] = []
+            record[field.grouping].append(data)
```
glennmatthews, line 214: Perhaps:

```suggestion
            record.setdefault(field.grouping, []).append(data)
```

nautobot/extras/models/customfields.py
```diff
@@ -161,6 +161,64 @@ def get_custom_fields(self, advanced_ui=None):
...
+        """
+        record = {}
+        fields = CustomField.objects.get_for_model(self)
+        if advanced_ui is not None:
+            fields = fields.filter(advanced_ui=advanced_ui)
+
+        for field in fields:
+            data = (field, self.cf.get(field.name))
+            # Data made before migration will have a grouping of None if not specified
+            # Data made after migration will have a grouping of "" if not specified.
```
glennmatthews, line 208: I think we should be able to avoid this discrepancy.

nautobot/extras/models/customfields.py
```diff
@@ -161,6 +161,64 @@ def get_custom_fields(self, advanced_ui=None):
...
+            if field.grouping in ["", None]:
+                field.grouping = ""
+            if field.grouping not in record:
+                record[field.grouping] = []
+            record[field.grouping].append(data)
+        record = OrderedDict(sorted(record.items()))
+        # Check if custom fields with no grouping specified exist.
+        if record.get(""):
+            # Move it to the front of the OrderedDict (render it first)
+            record.move_to_end("", last=False)
```
glennmatthews, line 219: Shouldn't `""` sort as first automatically?

nautobot/extras/models/customfields.py
```diff
@@ -251,6 +309,12 @@ class CustomField(BaseModel, ChangeLoggedModel, NotesMixin):
         limit_choices_to=FeatureQuery("custom_fields"),
         help_text="The object(s) to which this field applies.",
     )
+    grouping = models.CharField(
+        max_length=50,
```
glennmatthews, line 313: 50 is pretty limiting. JOB_MAX_GROUPING_LENGTH is 255, which may be more than needed, but storage isn't that expensive.

nautobot/extras/models/customfields.py
```diff
@@ -251,6 +309,12 @@ class CustomField(BaseModel, ChangeLoggedModel, NotesMixin):
...
+        null=True,
```
glennmatthews, line 314: I think removing `null=True` will ensure that it defaults to "" on migration, and it is good practice not to set null=True on CharFields anyway.

nautobot/extras/templates/extras/customfield.html
```diff
@@ -101,6 +101,14 @@
                             </ul>
                         </td>
                     </tr>
+                    <tr>
+                        <td>Grouping</td>
```
glennmatthews, line 105: I think it'd make more sense to have this located right after `Label` in the display.

nautobot/extras/templates/extras/customfield.html
```diff
@@ -101,6 +101,14 @@
...
+                        <td>
+                            <ul>
+                                {{ object.grouping }}
```
glennmatthews, line 108:
```suggestion
                                {{ object.grouping | placeholder }}
```

nautobot/extras/templates/extras/customfield_edit.html
```diff
@@ -21,6 +21,7 @@
         <div class="panel-heading"><strong>Assignment</strong></div>
         <div class="panel-body">
             {% render_field form.content_types %}
+            {% render_field form.grouping %}
```
glennmatthews, line 24: I think this would make more sense to move up to follow `form.label`.

nautobot/extras/templates/extras/inc/job_table.html
```diff
@@ -17,18 +17,22 @@
 {% load helpers %}
     <tbody>
         {% for row in table.page.object_list|default:table.rows %}
+            {% with forloop.counter0 as count %}
             {% ifchanged row.record.grouping %}
             <tr>
+                <!-- #TODO
+                Here we are using {{ row.record.grouping|slugify }} it should be changed as it breaks `accordion-toggle`
+                when two distinct groups have the same slugified name. e.g. Group 2, group-2 -->
```
glennmatthews, line 25: remove?

nautobot/core/templates/inc/custom_fields/panel_data.html
HanlinMiao: The .`according-toggle` in `base.css` is setting the font-size to 24 px, I have to override the class attribute in the template itself.
Screen Shot 2022-08-26 at 12 48 09 PM

nautobot/core/templates/inc/custom_fields/panel_data.html
HanlinMiao: Screen Shot 2022-08-26 at 1 00 42 PM
I guess this is the reason lol

nautobot/core/templates/generic/object_retrieve.html
```diff
@@ -208,4 +208,34 @@ <h1>{% block title %}{{ object }}{% endblock %}</h1>
     }
 
 </script>
+<script>
+    console.log("hello")
```
bryanculver, line 212:
```suggestion
```

nautobot/extras/models/customfields.py
```diff
@@ -161,6 +161,53 @@ def get_custom_fields(self, advanced_ui=None):
...
+        Return a dictonary of custom fields grouped by the same grouping in the form
+        {
+            <grouping_1>: [cf1, cf2...],
+            ...
+            <grouping_5>: [cf8, cf9...],
+            ...
+        }
+        which have advanced_ui set to True
+        """
+        return self.get_custom_field_groupings(advanced_ui=True)
```
bryanculver, line 188: Can we get some notes in here as to why these methods need to exist where calling `get_custom_field_groups` with the properly defined `advanced_ui` kwarg is insufficient?

I assume this is due in part to the way they are being called in the template?

nautobot/extras/forms/forms.py
```diff
@@ -333,6 +333,11 @@ class ConfigContextSchemaFilterForm(BootstrapMixin, forms.Form):
 
 class CustomFieldForm(BootstrapMixin, forms.ModelForm):
     label = forms.CharField(required=True, max_length=50, help_text="Name of the field as displayed to users.")
+    grouping = forms.CharField(
+        max_length=255,
+        required=False,
+        help_text="Human-readable grouping that this custom field belongs to",
+    )
```
bryanculver, line 340: Is this redundant with the inclusion of the `grouping` field in the Meta section below? Won't the field max_length, blank, and help_text on the model's field provide these values?

bryanculver (changes requested):

nautobot/extras/forms/forms.py
HanlinMiao: Yeah you are right. Confirmed by removing this code block and form worked as expected.

nautobot/core/templates/inc/custom_fields/panel_data.html
```diff
@@ -1,39 +1,60 @@
 {% load helpers %}
 {% if custom_fields %}
+<style>
+    .accordion-toggle {
+        font-size: 15px;
```
glennmatthews, line 5: Looks like the rest of the elements in this table are all defaulting to 14px, including the panel header, so let's make that consistent.

nautobot/extras/models/customfields.py
```diff
@@ -161,6 +163,55 @@ def get_custom_fields(self, advanced_ui=None):
...
+
+    def get_custom_field_groupings(self, advanced_ui=None):
+        """
+        Return a dictonary of custom fields grouped by the same grouping in the form
+        {
+            <grouping_1>: [cf1, cf2...],
+            ...
+            <grouping_5>: [cf8, cf9...],
+            ...
```
glennmatthews, line 201: If I'm reading the code right, this is actually:

```suggestion
            <grouping_1>: [(cf1, <value for cf1>), (cf2, <value for cf2>), ...],
            ...
```

nautobot/extras/models/customfields.py
```diff
@@ -161,6 +163,55 @@ def get_custom_fields(self, advanced_ui=None):
...
+        """
+        record = {}
+        fields = CustomField.objects.get_for_model(self)
+        if advanced_ui is not None:
+            fields = fields.filter(advanced_ui=advanced_ui)
+
+        for field in fields:
+            data = (field, self.cf.get(field.name))
+            record.setdefault(field.grouping, []).append(data)
+        record = OrderedDict(sorted(record.items()))
```
glennmatthews, line 212: Given dicts are order-preserving in modern Python, does this work just as well if you use `dict(sorted(record.items()))` instead?

nautobot/core/templates/generic/object_retrieve.html
```diff
@@ -208,4 +208,33 @@ <h1>{% block title %}{{ object }}{% endblock %}</h1>
     }
 
 </script>
+<script>
```
glennmatthews, line 211: maybe not needed to close one `script` and start a new `script` -- can this all be one `script` block?

glennmatthews (approved):

bryanculver (approved):

glennmatthews:
See also #3265

Issue #1574, opened 2022-03-30
Epic: Natural keys/uniqueness constraints


lampwins:

Kircheneer:
Does this epic entail implementation of `get_by_natural_key` for all core models?

jathanism:
> Does this epic entail implementation of `get_by_natural_key` for all core models?

In short, yes.

Also this is worth considering for the API: https://github.com/wq/django-natural-keys

bryanculver:
Closing as only #3975 remains which is nested in #3752

PR #2320, opened 2022-08-26
Remove PKs from tag fixture now that we have natural keys for them

glennmatthews:
# Relates-to #1574 
# What's Changed

~Add `slug` as a natural-key for `Tag`. Though it occurs to me we might need to revisit this depending on what we decide in #2312.~

This is a PR atop #2281 and #2360 since the latter allows me to remove the UUID PKs from the Tag fixture file.


glennmatthews:
@lampwins Should I change this (and/or the Status natural-keys support as well) to use `name` instead of `slug` as the key?

lampwins:
@glennmatthews because `TagBase.name` is unique, I would say yes because `name` is already a true natural key.

Incidentally, in double checking that in the taggit code, their handling of the `slug` field is a prime example of one of the reasons I truly believe `slugs` are almost never natural keys. I'll post my thoughts on #2312 :)

glennmatthews:
It might make sense to target this to `develop` instead, actually, since natural-keys for Status were added in 1.4.1. Thoughts?

bryanculver (approved):

Issue #449, opened 2021-05-14
Git repository sync should warn if files appear to be out of place


glennmatthews:
### Environment
* Python version: any
* Nautobot version: 1.0.1

### Proposed Functionality

If a Git repository is mis-structured in an easily detectable way, it would be great if the Git sync process would detect such cases and log appropriate warning/error messages, for example if:

- a Job class is defined in the project root rather than in `/jobs/`
- YAML or JSON files identifiable as config context data are located outside of `/config_contexts/`
- a directory exists under `/export_templates/` that doesn't match a known `app_label` or `modelname` (e.g. `/export_templates/device/` instead of `/export_templates/dcim/device/`)
- etc.

### Use Case

Improved ease of troubleshooting, especially when first integrating a Git repository with Nautobot. Currently the logging during a sync is mostly success-path-focused and doesn't make it obvious why data might not be detected or imported.

### Database Changes

None

### External Dependencies

None

PR #2323, opened 2022-08-26
Improve handling of various Git repository bad-data scenarios

glennmatthews:
# Closes: #449
# What's Changed

This came out of an extended troubleshooting session with a user on public Slack.

- More consistent logging in general if failures occur during various Git repository data sync operations
- Warn the user if directories that should exist under `/config_contexts/` are found in the repository root
- Warn the user if directories that should exist under `/export_templates/` are found in the repository root
- Fail gracefully if a config-context file is missing the `_metadata`->`name` key, instead of throwing a DB error when trying to create a record with null `name`
- Change the log message when encountering an unknown content-type in `/export_templates` from a `failure` to a `warning` - I'm open to persuasion that the current behavior is preferable; my thinking in making this change is that potentially a repo can contain export-templates for plugin models, or for models that are newly added in a later Nautobot release, and the presence of these templates shouldn't necessarily fail the sync when run from an older Nautobot version or in the absence of a given plugin
- Enhance `test_pull_git_repository_and_refresh_data_with_bad_data` test case to check for the specific expected log messages in addition to just checking that the sync failed in this case.

# TODO
- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/datasources/git.py
+++ b/nautobot/extras/datasources/git.py
@@ -371,10 +371,7 @@ def update_git_config_contexts(repository_record, job_result):
         try:
             with open(os.path.join(config_context_path, file_name), "r") as fd:
                 # The data file can be either JSON or YAML; since YAML is a superset of JSON, we can load it regardless
-                try:
-                    context_data = yaml.safe_load(fd)
-                except Exception as exc:
-                    raise RuntimeError(f"Error in loading config context data from `{file_name}`: {exc}")
+                context_data = yaml.safe_load(fd)
 
             # A file can contain one config context dict or a list thereof
             if isinstance(context_data, dict):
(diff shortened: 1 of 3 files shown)


nautobot/extras/datasources/git.py
```diff
@@ -509,8 +516,13 @@ def import_config_context(context_data, repository_record, job_result, logger):
     context_record = None
     # TODO: check context_data against a schema of some sort?
 
+    if "_metadata" not in context_data:
+        raise RuntimeError("data is missing the required `_metadata` key.")
```
itdependsnetworks, line 520: Perhaps something like: “The `data` dictionary is missing the required `_metadata` key.” 

Thought process being is data is a generic term and can be quickly misunderstood. Something similar for all of the messages

glennmatthews: Good thought. `data` isn't actually a key anywhere though. The full message seen by the user would be a la "Error in loading config context data from `foo.yml`: data is missing the required `_metadata` key", do you think that gives enough context?

itdependsnetworks: yes, that is helpful, thanks

nautobot/extras/datasources/git.py
```diff
@@ -411,6 +406,15 @@ def update_git_config_contexts(repository_record, job_result):
         "tenants",
         "tags",
     ):
+        if os.path.isdir(os.path.join(repository_record.filesystem_path, filter_type)):
+            job_result.log(
+                f'Found "{filter_type}" directory in the repository root. If this is meant to contain config contexts, '
+                "it should be moved into a `config_contexts/` subdirectory.",
+                level_choice=LogLevelChoices.LOG_WARNING,
```
bryanculver, line 413: Would this be better suited to be an info instead of a warning? Thought being that the folder named "sites" isn't harmful to not be nested (maybe used for some other purpose) but may not have the desired affect.

glennmatthews: That's why it's a warning instead of a failure :-) IMHO info isn't a "sit up and take notice" log level, and I feel like it should be.

bryanculver (approved):
Net improvement 👍

PR #2332, opened 2022-08-29
Fix code and logic around MACAddressCharField

glennmatthews:
# Closes: #n/a
# What's Changed

@gsnider2195 pointed out in #2314 that pylint was warning about `MACAddressCharField.validators` because we were overriding a `@cached_property` with a `@property`. I dug into this and came up with the following set of changes/fixes:

- Instead of overriding the `validators` cached_property to always return an empty list, add logic in `MACAddressCharField.__init__` to just strip out the problematic validator from the (no longer overridden) `validators` list.
- Additionally, I realized on looking at this code, that after some of the early MySQL work back in 2021, we actually have two separate model field classes - `MACAddressCharField` (what `Interface` and `VMInterface` use now, compatible with both MySQL and PostgreSQL) but also still `MACAddressField` (Postgres-specific, should not be used). This led me to the following fixes:
  - Add `@class_deprecated_in_favor_of` decorator to `MACAddressField`. (This required swapping the order of `MACAddressField` and `MACAddressCharField` in `fields.py`, which makes the diff a bit confusing as they have a lot of duplicated code between them.)
  - Removed Graphene code for mapping the deprecated `MACAddressField` to a string. As far as I can tell `MACAddressCharField` already works fine as-is in GraphQL so we don't need an equivalent function for it.
  - Fixed filter logic in `utilities.filters` to correctly assign a `MultiValueMACAddressFilter` to instances of `MACAddressCharField` (previously it was incorrectly assigning it to instances of the *form* (not *model*) field of `dcim.forms.MACAddressField`, which would be a no-op since no models have a form field as part of the model fields.)
  - Fixed test-only data model in `utilities.tests.test_filters` to use a `MACAddressCharField` instead of a `MACAddressField`, since that matches what we actually use in our real models.

# TODO

- n/a Unit, Integration Tests

### DIFF
--- a/nautobot/utilities/filters.py
+++ b/nautobot/utilities/filters.py
@@ -16,6 +16,7 @@ from django_filters.utils import get_model_field, resolve_field
 from mptt.models import MPTTModel
 from tree_queries.models import TreeNode
 
+from nautobot.dcim.fields import MACAddressCharField
 from nautobot.dcim.forms import MACAddressField
 from nautobot.extras.models import Tag
 from nautobot.utilities.constants import (
@@ -540,7 +541,7 @@ class BaseFilterSet(django_filters.FilterSet):
             models.TimeField: {"filter_class": MultiValueTimeFilter},
             models.URLField: {"filter_class": MultiValueCharFilter},
             models.UUIDField: {"filter_class": MultiValueUUIDFilter},
-            MACAddressField: {"filter_class": MultiValueMACAddressFilter},
+            MACAddressCharField: {"filter_class": MultiValueMACAddressFilter},
             TaggableManager: {"filter_class": TagFilter},
         }
     )
--- a/nautobot/dcim/apps.py
+++ b/nautobot/dcim/apps.py
@@ -1,5 +1,3 @@
-import graphene
-
 from nautobot.core.apps import NautobotConfig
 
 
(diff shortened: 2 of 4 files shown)


nautobot/utilities/filters.py
```diff
@@ -540,7 +541,7 @@ class BaseFilterSet(django_filters.FilterSet):
             models.TimeField: {"filter_class": MultiValueTimeFilter},
             models.URLField: {"filter_class": MultiValueCharFilter},
             models.UUIDField: {"filter_class": MultiValueUUIDFilter},
-            MACAddressField: {"filter_class": MultiValueMACAddressFilter},
+            MACAddressCharField: {"filter_class": MultiValueMACAddressFilter},
```
bryanculver, line 544: Should we not leave the `MACAddressField` filter_class assignment here on the off-chance anyone is using that field type? This seems to just be assigning default filter types. We can leave in `#TODO 2.0` to rip out later.

HanlinMiao (approved):

nautobot/utilities/filters.py
glennmatthews: No, this was outright wrong. The `MACAddressField` here was `dcim.forms.MACAddressField` (*form* field class), not `dcim.fields.MACAddressField` (*model* field class), so it wasn't actually doing anything useful since this dict is for mapping *model* fields to filter classes. It was basically a meaningless entry that never would get used.

I could *add* an additional entry for dcim.fields.MACAddressField, but I don't think that's necessary personally.

bryanculver (approved):

PR #2339, opened 2022-08-30
Pylint part 2

glennmatthews:
# Towards: #2292 
# What's Changed

Trying to keep these PRs a bit more manageable. :-) This one enables a few more of the highest-severity ("Error") pylint checkers and addresses them as described:

- Enable E0203 `access-member-before-definition` and fix the 1 case it flagged (a minor if legitimate bug with a missing default value in `UpdateObjectViewTestCase`)
- Enable E1136 `unsubscriptable-object`, no reports for this warning (probably got fixed along the way in the time since I initially set up the pylint configs)
- Enable E1120 `no-value-for-parameter`; it reported a few warnings in test code due to Pylint not innately understanding how `unittest.mock.patch` works; fixed this by adding `signature-mutators` to the Pylint configuration.
- Enable E1133 `not-an-iterable`; 5 warnings reported, all of which appear to be false positives where we have an optional parameter defaulting to None and pylint not recognizing that we're guarding with a `if parameter is not None` before trying to iterate over that parameter. I added type hints to these in hopes that it would get pylint to take the hint, but it appears not, so I added explicit `pylint: disable` for each of these as well.
- Enhance the pylint invoke task to permit targeted execution like `invoke pylint --target nautobot.utilities.testing`, `invoke pylint --target examples/example_plugin --recursive`.

# TODO
- n/a Unit, Integration Tests


bryanculver (approved):
Curious of the value of keeping the type hints in here if ultimately didn't solve the lint issue.

Maybe we should do `__iter__` checks instead of `is not None`?

Since Python doesn't do strict type enforcement at runtime maybe pylint is being overly pessimistic and wanting to see something like

`if hasattr(self.relationships, '__iter__'):`

glennmatthews:
> Curious of the value of keeping the type hints in here if ultimately didn't solve the lint issue.

I mean in the long run I would **love** to turn on `mypy` for type checking as well... :-) 

> 
> Maybe we should do `__iter__` checks instead of `is not None`?
> 
> Since Python doesn't do strict type enforcement at runtime maybe pylint is being overly pessimistic and wanting to see something like
> 
> `if hasattr(self.relationships, '__iter__'):`

I'll give it a try.

bryanculver:
> > Curious of the value of keeping the type hints in here if ultimately didn't solve the lint issue.
> 
> I mean in the long run I would **love** to turn on `mypy` for type checking as well... :-)

yes please

glennmatthews:
I tried `if hasattr(..., '__iter__')` and even `if isinstance(..., (list, tuple))` and neither made Pylint happy as a guard around some of these `not-an-iterable` cases. Explicitly casting to list (`for item in list(variable)`) did but I am not a big fan of that approach. I was able to address a couple of the cases by tweaking the actual logic flow a bit though.

Issue #1998, opened 2022-07-05
Filter form UI revamp


lampwins:
### As ...

Nelly - Network Engineer

### I want ...

An extensible and easy-to-use filter form UI in all object list views

### So that ...

I have a better experience when trying to filter results which allows me to use all possible filtering options

### Requirements
- Not display all permutations of fields and lookup expressions simultaneously (by default)
- Provide the user an ability to select any field and lookup expression combination supported by the FitlerSet to filter the table by
- Provide the user with an ability to edit, view current filter arguments
- Provide the user the ability to specify an arbitrary number of filters

### Out of scope
- Filtering arbitrary by nested lookup

### Optional - Feature groups this request pertains to.

- UI/UX

### Database Changes

None

### External Dependencies

None?

See #1608

glennmatthews:
Is "saving" a set of filters, or setting them as per-user defaults for a particular object-list, in scope?

lampwins:
> Is "saving" a set of filters, or setting them as per-user defaults for a particular object-list, in scope?

No

bryanculver:
Let's see if we can also get an answer to #2162 before this.

bryanculver:
Closed with #2348.

Great work @timizuoebideri1!

PR #2348, opened 2022-08-31
FilterForm UI Revamp

timizuoebideri1:
# Closes: #1998 
# What's Changed

1. Applied filters
2. Filter form button
Screenshot 2022-10-28 at 5 54 02 PM

1. Add new filter field
2. Remove a filter field
3. Filter fields

Screenshot 2022-10-28 at 5 53 24 PM
Screenshot 2022-10-28 at 5 53 31 PM

# TODO

- [x] Unit, Integration Tests


nautobot/utilities/utils.py
```diff
@@ -672,3 +695,164 @@ def pretty_str(self, node=None, depth=0):
...
+        if "choices" in field.extra:  # Field choices
+            # Problem might arise here if plugin developer do not declare field inheriting from nautobot.utilities.choices.ChoiceSet
+            data = {
+                "type": "static-choices",
+                "choices": field.extra["choices"].CHOICES,
+                "allow_multiple": True,
+            }
+        elif hasattr(field, "queryset"):  # Dynamically populated choices
+            if isinstance(field, StatusFilter):
+                related_model = Status
```
glennmatthews, line 803: Why do we need this special case?

nautobot/utilities/utils.py
```diff
@@ -672,3 +695,164 @@ def pretty_str(self, node=None, depth=0):
...
+                data["value_field"] = value_field
+
+            # Add initial choices if initial_choice is not none
+            # This can be used to populate the selected options for the select field
+            if initial_choice is not None:
+                search_by = field.extra.get("to_field_name") or "id"
+                values = initial_choice if isinstance(initial_choice, (list, tuple)) else [initial_choice]
+                data["choices"] = compile_model_choices(related_model, search_by, values)
+
+    elif isinstance(field, (RelatedMembershipBooleanFilter,)):  # Yes / No choice
```
glennmatthews, line 830: Should we check for the base class `django_filters.BooleanFilter` instead?

nautobot/utilities/utils.py
timizuoebideri1: `StatusFilter` do not have the attribute `queryset`.

glennmatthews: Should it, perhaps?

lampwins:
This looks great! I would like to maintain a simple text search bar for the `q` filter which does not require the user to click the filter dialog button to add this specific filter. Let me know if we need a separate story to track this. cc: @bryanculver

briddo (commented):
Looks great to me so far!

HanlinMiao:
Great Work looks awesome!
I was playing around with it. When I specify a filter but not a value, it says that I specified an invalid filter. Is it intentional or should we change it?
See example below:
Screen Shot 2022-09-02 at 9 45 48 PM
Screen Shot 2022-09-02 at 9 46 08 PM

timizuoebideri1:
> Great Work looks awesome! I was playing around with it. When I specify a filter but not a value, it says that I specified an invalid filter. Is it intentional or should we change it (just ignore a filter with no value specified?) See example below: Screen Shot 2022-09-02 at 9 45 48 PM Screen Shot 2022-09-02 at 9 46 08 PM

Filters with error would be ignored only if `STRICT_FILTERING` is `False`

glennmatthews:
> > Great Work looks awesome! I was playing around with it. When I specify a filter but not a value, it says that I specified an invalid filter. Is it intentional or should we change it (just ignore a filter with no value specified?)
> 
> Filters with error would be ignored only if `STRICT_FILTERING` is `False`

I'd be in favor of silently ignoring filters that have no user-specified value, rather than requiring the user to manually remove these filters to get the view to work.

christhant:
Updated Filter UI

image

christhant:
update filter UI without border panel

image

HanlinMiao (commented):
Screen Shot 2022-09-16 at 12 08 45 PM
The DeviceType Filters in DeviceListView seem to be broken. The choices in the dropdown are not populating, can you confirm that on your end? @timizuoebideri1

timizuoebideri1:
> The DeviceType Filters in DeviceListView seem to be broken. The choices in the dropdown are not populating, can you confirm that on your end? @timizuoebideri1

@HanlinMiao  works for me

nautobot/utilities/utils.py
```diff
@@ -689,3 +720,220 @@ def pretty_str(self, node=None, depth=0):
...
+        label = f"({search.group()})"
+
+    if FILTER_LOOKUP_MAP.get(verbose_name):
+        lookup_label = FILTER_LOOKUP_MAP[verbose_name]
+        verbose_name = "not-" + lookup_label if label.startswith("(n") else lookup_label
+
+    return verbose_name + label
+
+
+def get_all_lookup_exper_for_field(model, field_name):
```
jathanism, line 750: This looks like a typo and should be renamed:

```suggestion
def get_all_lookup_expr_for_field(model, field_name):
```

nautobot/utilities/utils.py
```diff
@@ -689,3 +720,220 @@ def pretty_str(self, node=None, depth=0):
...
+        field_name (str): Field name e.g slug__iew
+        verbose_name (str): The verbose name for the lookup exper which is suffixed to the field name e.g iew -> iendswith
+
+    Examples:
+        >>> build_lookup_label("slug__iew", "iendswith")
+        >>> "ends-with(iew)"
+    """
+
+    label = ""
+    search = re.search(r"(?<=__)\w+", field_name)
```
jathanism, line 739: Please move this `re` pattern to a module-level constant and use `re.compile` so that it is runtime-optimized, and then you can use the pattern to call `.search` here in the business logic.

e.g.

```python
# Defined at top of module..
SEARCH_RE = re.compile(r"(?<=__)\w+")

# later in code...
def build_lookup_table(field_name, verbose_name): 
    # Call it like this:
    search - SEARCH_RE.search(field_name)
```

nautobot/core/api/views.py
```diff
@@ -36,6 +37,11 @@
 from nautobot.core.api import BulkOperationSerializer
 from nautobot.core.api.exceptions import SerializerNotFound
 from nautobot.utilities.api import get_serializer_for_model
+from nautobot.utilities.utils import (
+    get_all_lookup_exper_for_field,
```
jathanism, line 41:
```suggestion
    get_all_lookup_expr_for_field,
```

nautobot/core/api/views.py
```diff
@@ -702,3 +708,88 @@ def execute_graphql_request(self, request, data, query, variables, operation_nam
...
+        if "contenttype" not in request.GET or "field_name" not in request.GET:
+            return Response(
+                "contenttype and field_name are required parameters",
+                status=400,
+            )
+        contenttype = request.GET.get("contenttype")
+        field_name = request.GET.get("field_name")
+        app_label, model_name = contenttype.split(".")
+        model = ContentType.objects.get(app_label=app_label, model=model_name).model_class()
+        data = get_all_lookup_exper_for_field(model, field_name)
```
jathanism, line 754:
```suggestion
        data = get_all_lookup_expr_for_field(model, field_name)
```

nautobot/utilities/tests/test_utils.py
```diff
@@ -1,17 +1,26 @@
...
 from nautobot.core.settings_funcs import is_truthy
+from nautobot.utilities.forms import DatePicker, DateTimePicker, DynamicModelMultipleChoiceField
 from nautobot.utilities.utils import (
+    build_lookup_label,
+    get_values_display_names,
+    convert_querydict_to_factory_formset_acceptable_querydict,
     deepmerge,
     dict_to_filter_params,
-    get_form_for_model,
+    get_all_lookup_exper_for_field,
```
jathanism, line 14:
```suggestion
    get_all_lookup_expr_for_field,
```

nautobot/utilities/tests/test_utils.py
```diff
@@ -346,3 +355,173 @@ def test_slugify_dashes_to_underscores(self):
...
+
+        with self.subTest("Test field with has_ prefix"):
+            lookup_expr = get_all_lookup_exper_for_field(Site, "has_vlans")
+            self.assertEqual(
+                lookup_expr,
+                [{"id": "has_vlans", "name": "exact"}],
+            )
+
+        with self.subTest("Test unknown field"):
+            lookup_expr = get_all_lookup_exper_for_field(Site, "unknown_field")
```
jathanism, line 403:
```suggestion
    def test_get_all_lookup_expr_for_field(self):
        with self.subTest():
            lookup_expr = get_all_lookup_expr_for_field(Site, "status")
            self.assertEqual(
                lookup_expr,
                [{"id": "status", "name": "exact"}, {"id": "status__n", "name": "not-exact(n)"}],
            )

        with self.subTest("Test field with has_ prefix"):
            lookup_expr = get_all_lookup_expr_for_field(Site, "has_vlans")
            self.assertEqual(
                lookup_expr,
                [{"id": "has_vlans", "name": "exact"}],
            )

        with self.subTest("Test unknown field"):
            lookup_expr = get_all_lookup_expr_for_field(Site, "unknown_field")
```

tbotnz:
this looks epic! when are you targeting release?

jathanism:
> this looks epic! when are you targeting release?

1.5 sir

nautobot/utilities/utils.py
```diff
@@ -689,3 +704,220 @@ def pretty_str(self, node=None, depth=0):
...
+        ...     'form-0-lookup_field': ['status'],
+        ...     'form-0-lookup_type': ['status'],
+        ...     'form-0-value': ['active', 'decommissioning'],
+        ...     'form-1-lookup_field': ['name'],
+        ...     'form-1-lookup_type': ['name__ic'],
+        ...     'form-1-value': ['site']
+        ... }
+    """
+    query_dict = QueryDict(mutable=True)
+    filterset_class_fields = filterset_class.base_filters.keys()
```
glennmatthews, line 864: Just checking - does this feature work correctly for dynamically generated filters (filtering on custom fields, relationships)?

nautobot/utilities/utils.py
```diff
@@ -689,3 +704,220 @@ def pretty_str(self, node=None, depth=0):
...
+
+def is_single_choice_field(filterset_class, field_name):
+    # Some filter parameters do not accept multiple values, e.g DateTime fields, boolean fields etc.
+    field = filterset_class.base_filters.get(field_name)
+    if field is None:
+        pass
+        # TODO (timizuo): handle filed not exist
+    if isinstance(field, (DateFilter, DateTimeFilter, TimeFilter, BooleanFilter)):
+        return True
+    return False
```
glennmatthews, line 899:
```suggestion
    return isinstance(field, (DateFilter, DateTimeFilter, TimeFilter, BooleanFilter))
```

nautobot/utilities/utils.py
```diff
@@ -689,3 +704,220 @@ def pretty_str(self, node=None, depth=0):
...
+
+    # Some FilterSet field only accept single choice not multiple choices
+    # e.g datetime field, bool fields etc.
+    filter_params = {
+        field: filter_params.get(field)
+        if is_single_choice_field(filterset_class, field)
+        else filter_params.getlist(field)
+        for field in filter_params.keys()
+        if filter_params.get(field)  # Discard fields without values
+    }
```
glennmatthews, line 922: I'd prefer a `for` loop here as well, personally.

nautobot/extras/models/groups.py
```diff
@@ -211,11 +211,14 @@ def _map_filter_fields(self):
...
-            # not correct for char fields. For boolean fields, we want them to be nullable.
-            if isinstance(modelform_field, (forms.CharField, forms.BooleanField)):
+            # Use filterset_field to generate the correct filterform_field for CharField.
+            # Which is `MultiValueCharField`.
+            if isinstance(modelform_field, forms.CharField):
+                new_modelform_field = filterset_field.field
+                modelform_field = new_modelform_field
+
+            # For boolean fields, we want them to be nullable.
+            if isinstance(modelform_field, forms.BooleanField):
```
jathanism, line 221: I'm not clear on why these were broken out since the code seems to be the same for `CharField` and `BooleanField`?

nautobot/core/settings.py
```diff
@@ -730,3 +730,20 @@ def UI_RACK_VIEW_TRUNCATE_FUNCTION(device_display_name):
     :type: str
     """
     return str(device_display_name).split(".")[0]
+
+
+# Filter lookup Expr verbose name
+FILTERS_VERBOSE_LOOKUPS = {
```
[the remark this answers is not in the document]
timizuoebideri1: Its been used by [django_filters.utils.verbose_lookup_expr()](https://github.com/nautobot/nautobot/pull/2348/files#diff-9daac98565b5e7e2ca25cc63a973b705d837f2be4e212e442952dadf6f62a08dR596).

nautobot/utilities/forms/forms.py
```diff
@@ -233,3 +241,102 @@ def __init__(self, table, *args, **kwargs):
...
+
+        # Update lookup_type and lookup_value fields to match expected field types derived from data
+        # e.g status expects a ChoiceField with APISelectMultiple widget, while name expects a CharField etc.
+        if kwargs.get("data") and kwargs.get("prefix"):
+            data = kwargs["data"]
+            prefix = kwargs["prefix"]
+            lookup_type = data.get(prefix + "-lookup_type")
+            lookup_value = data.getlist(prefix + "-lookup_value")
+
+            if lookup_type and lookup_value and lookup_type in self.filterset_filters:
```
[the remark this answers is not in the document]
timizuoebideri1: Nope.

nautobot/utilities/templates/utilities/templatetags/filter_form.html
```diff
@@ -0,0 +1,59 @@
...
+                            </a>
+                        </div>
+                    </form>
+                {% else %}
+                    <div class="alert alert-warning">
+                        <b>
+                            <i class="mdi mdi-alert"></i>
+                            Please create a FilterSet for this model to use this feature.
+                        </b>
+                    </div>
```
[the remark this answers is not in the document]
timizuoebideri1: Yes manually.

nautobot/core/templates/generic/object_list.html
```diff
@@ -20,14 +35,30 @@
...
+    {% for field, values in filter_params %}
+    <span class="select2 select2-container select2-container--bootstrap display-inline" dir="ltr">
+        <span class="selection">
+        <span class="select2-selection select2-selection--multiple p-5" role="combobox" aria-haspopup="true" aria-expanded="false" tabindex="-1" aria-disabled="false">
+            <b>{{ field }}:</b> <span class="select2-selection__clear remove-filter-param m-0" title="Remove all items" data-field-type="parent" data-field-value={{ field }}>×</span>
+            <ul class="select2-selection__rendered">
+                {% for value in values %}
+                <li class="select2-selection__choice chip" title="{{ value }}"><span class="select2-selection__choice__remove remove-filter-param" data-field-type="child"  data-field-parent={{ field }}  data-field-value={{ value }} role="presentation">×</span>{{ value }}</li>
+                {% endfor %}
+            </ul>
```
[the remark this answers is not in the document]
timizuoebideri1: @christopherthant

nautobot/core/views/generic.py
```diff
@@ -313,7 +319,8 @@ def get(self, request):
             "permissions": permissions,
             "action_buttons": valid_actions,
             "table_config_form": TableConfigForm(table=table),
-            "filter_form": filter_form,
+            "filter_params": display_filter_params,
+            "filter_form": dynamic_filter_form,
```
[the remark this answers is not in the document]
timizuoebideri1: If `{% block content %}` is not overridden, then no breaking change.

nautobot/core/templates/generic/object_list.html
```diff
@@ -20,14 +35,30 @@
...
+<div class="filters-applied">
+    <b>Filters:</b>
+    {% for field, values in filter_params %}
+    <span class="select2 select2-container select2-container--bootstrap display-inline" dir="ltr">
+        <span class="selection">
+        <span class="select2-selection select2-selection--multiple p-5" role="combobox" aria-haspopup="true" aria-expanded="false" tabindex="-1" aria-disabled="false">
+            <b>{{ field }}:</b> <span class="select2-selection__clear remove-filter-param m-0" title="Remove all items" data-field-type="parent" data-field-value={{ field }}>×</span>
+            <ul class="select2-selection__rendered">
+                {% for value in values %}
+                <li class="select2-selection__choice chip" title="{{ value }}"><span class="select2-selection__choice__remove remove-filter-param" data-field-type="child"  data-field-parent={{ field }}  data-field-value={{ value }} role="presentation">×</span>{{ value }}</li>
```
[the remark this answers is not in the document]
timizuoebideri1: @christopherthant

nautobot/utilities/forms/forms.py
```diff
@@ -233,3 +241,102 @@ def __init__(self, table, *args, **kwargs):
...
+        self.fields["lookup_type"].widget.attrs["class"] = "nautobot-select2-api lookup_type-select"
+
+        lookup_value_css = self.fields["lookup_value"].widget.attrs.get("class") or ""
+        self.fields["lookup_value"].widget.attrs["class"] = " ".join(
+            [lookup_value_css, "lookup_value-input form-control"]
+        )
+
+    @staticmethod
+    def capitalize(field):
+        return field  # TODO(Culver): Fix, custom field is not
```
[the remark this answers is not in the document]
timizuoebideri1: @bryanculver

nautobot/utilities/tests/test_utils.py
```diff
@@ -346,3 +355,173 @@ def test_slugify_dashes_to_underscores(self):
...
+        for field in single_choice_fields:
+            self.assertTrue(is_single_choice_field(filterset_class, field))
+
+        multi_choice_fields = ("status", "tenant", "tag")
+        for field in multi_choice_fields:
+            self.assertFalse(is_single_choice_field(filterset_class, field))
+
+    def test_build_lookup_label(self):
+        with self.subTest():
+            label = build_lookup_label("slug__iew", "iendswith")
```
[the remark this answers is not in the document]
timizuoebideri1: That's the verbose name for `iew`, which is needed by `django_filters.utils.verbose_lookup_expr(_verbose_name)` to get the user-friendly version -> `ends-with`.

nautobot/utilities/tests/test_forms.py
```diff
@@ -615,3 +653,232 @@ def test_api_select_add_query_param_with_utf8(self):
...
+        #     self.assertEqual(form.capitalize("test_one"), "Test one")
+
+        with self.subTest("Assert get_lookup_field_choices"):
+            self.assertEqual(
+                form.get_lookup_field_choices(),
+                [
+                    ("color", "Color"),
+                    ("content_types", "Content type(s)"),
+                    ("created", "Created"),
+                    ("id", "Id"),
```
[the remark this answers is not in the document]
timizuoebideri1: `Id` is the field's label. Except the field label is updated to be `ID`.

nautobot/utilities/tests/test_utils.py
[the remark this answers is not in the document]
glennmatthews: And there's no straightforward way for us to go from `slug__iew` --> `iew` --> `iendswith` to avoid needing to pass both `slug__iew` and `iendswith` as distinct parameters?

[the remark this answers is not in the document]
timizuoebideri1: Conflict may arise. An example I can think of is:
`status__n` -> `n` -> `exact`
`tenantgroup__n` -> `n` -> `in`

How do i figure out which is `in` and which is `exact`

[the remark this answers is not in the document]
timizuoebideri1: Currently the `verbose_name` is generated from `filterset_class().filters["status"].lookup_expr`

changes/1998.added
```diff
@@ -0,0 +1 @@
+Added DynamicFilterForm to list views.
```
glennmatthews, line 1: Let's add a section to the Release Overview in `docs/release-notes/version-1.5.md` calling this out - it's a flagship feature of this release.

nautobot/core/api/urls.py
```diff
@@ -36,4 +38,7 @@
     path("graphql/", GraphQLDRFAPIView.as_view(), name="graphql-api"),
     # Plugins
     path("plugins/", include((plugin_api_patterns, "plugins-api"))),
+    # Lookup Expr
+    path("lookup-choices/", GetFilterSetFieldLookupExpressionChoicesAPI.as_view(), name="lookup_choices"),
+    path("lookup-value-dom-element/", GetFilterSetFieldDOMElementAPI.as_view(), name="lookup_value_dom_element"),
```
glennmatthews, line 43: Perhaps something like:

```suggestion
    path("filterset-fields/lookup-choices/", GetFilterSetFieldLookupExpressionChoicesAPI.as_view(), name="lookup_choices"),
    path("filterset-fields/lookup-value-dom-element/", GetFilterSetFieldDOMElementAPI.as_view(), name="lookup_value_dom_element"),
```

nautobot/core/api/views.py
```diff
@@ -702,3 +709,133 @@ def execute_graphql_request(self, request, data, query, variables, operation_nam
...
+                "count": len(data),
+                "next": None,
+                "previous": None,
+                "results": data,
+            }
+        )
+
+
+class GetFilterSetFieldDOMElementAPI(NautobotAPIVersionMixin, APIView):
+    """API View that gets the dom element of a FilterSet field."""
```
glennmatthews, line 796:
```suggestion
    """API View that gets the DOM element representation of a FilterSet field."""
```

nautobot/core/templates/generic/object_list.html
```diff
@@ -20,14 +35,30 @@
...
+    {% endfor %}
+</div>
+<hr>
+{% endif %}
+
 <div class="row">
     <div class="col-md-12">
-        {% if filter_form %}
-            <div class="col-md-3 pull-right right-side-panel noprint">
-                {% include 'inc/search_panel.html' %}
```
glennmatthews, line 27: From a quick grep I see that `inc/search_panel.html` is still used in:

```
nautobot/dcim/templates/dcim/connections_list.html:            {% include 'inc/search_panel.html' %}
nautobot/dcim/templates/dcim/rack_elevation_list.html:            {% include 'inc/search_panel.html' %}
nautobot/extras/templates/extras/plugins_list.html:                    {% include 'inc/search_panel.html' %}
```

Should each of those also be updated to use the new  filtering UI as well? If so, I'm fine with opening that up as a follow-up issue, just want to make sure it gets looked at.

nautobot/core/templates/generic/object_list.html
```diff
@@ -83,8 +114,67 @@ <h1>{% block title %}{{ content_type.model_class|meta:"verbose_name_plural"|bett
...
+        formCssClass: 'dynamic-filterform',
+        added: jsify_form
+    });
+
+    // By default on lookup_value field names are form-\d-lookup_value, thats why
+    // on page load we change all `lookup_value` name to its relevant `lookup_type` value
+    $(".dynamic-filterform").each(function(){
+        lookup_type_value = $(this).find(".lookup_type-select").val()
+        lookup_value = $(this).find(".lookup_value-input")
+        lookup_value.attr("name", lookup_type_value)
```
glennmatthews, line 139: Semicolons at line end here and below, perhaps?

nautobot/core/templates/generic/object_list.html
```diff
@@ -83,8 +114,67 @@ <h1>{% block title %}{{ content_type.model_class|meta:"verbose_name_plural"|bett
...
+    // Remove applied filters
+    $(".remove-filter-param").on("click", function(){
+        let query_params = location.search
+        let type = $(this).attr("data-field-type")
+        let field_value = $(this).attr("data-field-value")
+        let query_string = location.search.substr(1).split("&")
+
+        if (type === "parent")
+            query_string = query_string.filter(item => item.search(field_value) < 0)
+        else {
```
glennmatthews, line 151: Perhaps:

```suggestion
        if (type === "parent") {
            query_string = query_string.filter(item => item.search(field_value) < 0);
        } else {
```

nautobot/core/tests/test_api.py
```diff
@@ -216,3 +216,62 @@ def test_header_and_query_version(self):
...
+    def test_get_lookup_field_data_invalid_query_params(self):
+        url = reverse("lookup_choices")
+        response = self.client.get(url + "?content_type=invalid.contenttypes&field_name=name", **self.header)
+
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.data, "content_type not found")
+
+    def test_get_lookup_choices(self):
+        url = reverse("lookup_choices")
+        response = self.client.get(url + "?content_type=dcim.site&field_name=status", **self.header)
```
glennmatthews, line 238: Do we need a test for invalid `field_name`, e.g. `?content_type=dcim.site&field_name=ice_cream_flavor`?

nautobot/core/tests/test_api.py
```diff
@@ -216,3 +216,62 @@ def test_header_and_query_version(self):
...
+    def test_get_lookup_field_data_invalid_query_params(self):
+        url = reverse("lookup_value_dom_element")
+        response = self.client.get(url + "?content_type=invalid.contenttypes&field_name=name", **self.header)
+
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.data, "content_type not found")
+
+    def test_get_lookup_value_dom_element(self):
+        url = reverse("lookup_value_dom_element")
+        response = self.client.get(url + "?content_type=dcim.site&field_name=name", **self.header)
```
glennmatthews, line 269: Same question here - test for valid content_type but invalid field_name?

nautobot/project-static/js/forms.js
```diff
@@ -402,26 +438,104 @@ function jsify_form(context) {
...
-    // Scroll up an offset equal to the first nav element if a hash is present
-    // Cannot use '#navbar' because it is not always visible, like in small windows
-    function headerOffsetScroll() {
-        if (window.location.hash) {
-            // Short wait needed to allow the page to scroll to the element
-            setTimeout(function() {
-                window.scrollBy(0, -$('nav').height())
-            }, 10);
+function initializeDynamicFilterForm(context){
+    this_context = $(context)
```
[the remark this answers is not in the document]
glennmatthews: Was closed, but doesn't look like it was addressed?

nautobot/utilities/forms/forms.py
```diff
@@ -233,3 +241,102 @@ def __init__(self, table, *args, **kwargs):
...
+
+        # cls.model is set at `dynamic_formset_factory()`
+        self.model = model or getattr(self, "model", None)
+
+        # Raise exception if `cls.model` not set and `model` not passed
+        if self.model is None:
+            raise AttributeError("'DynamicFilterForm' object requires `model` attribute")
+
+        filterset_class = get_filterset_for_model(self.model)
+        if filterset_class is not None:
```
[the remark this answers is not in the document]
glennmatthews: What should we do in the case where it *is* None? Is this expected under normal circumstances, or should it result in an exception or logging an error or something?

nautobot/utilities/tests/test_forms.py
```diff
@@ -615,3 +653,230 @@ def test_api_select_add_query_param_with_utf8(self):
...
+                    ("time_zone", "Time zone"),
+                    ("vlan_groups", "Vlan groups (slug or ID)"),
+                    ("vlans", "Vlans"),
+                ],
+            )
+
+        # TODO(Culver): Assertion doesn't like instantiated filters
+        # with self.subTest("Assert form generates the correct base_filters"):
+        #     self.assertEqual(form.filterset_filters, StatusFilterSet().filters)
+        #     self.assertEqual(site_form.filterset_filters, SiteFilterSet().filters)
```
glennmatthews, line 745: @bryanculver?

nautobot/utilities/tests/test_forms.py
```diff
@@ -615,3 +653,230 @@ def test_api_select_add_query_param_with_utf8(self):
...
+            )
+
+            self.assertEqual(
+                form.fields["lookup_type"].widget.attrs,
+                {
+                    "class": "nautobot-select2-api lookup_type-select",
+                    "placeholder": None,
+                    "data-query-param-field_name": '["$lookup_field"]',
+                    "data-contenttype": "extras.status",
+                    "data-url": "/api/lookup-choices/",
```
glennmatthews, line 774: Should probably use `reverse()` here and below rather than hard-coding this URL.

nautobot/utilities/tests/test_forms.py
```diff
@@ -615,3 +653,230 @@ def test_api_select_add_query_param_with_utf8(self):
...
+                },
+            )
+
+            self.assertEqual(
+                form.fields["lookup_value"].widget.attrs,
+                {"class": "form-control lookup_value-input form-control", "placeholder": "Value"},
+            )
+
+    def test_dynamic_filter_form_with_data_and_prefix(self):
+        """Assert that lookup value implements the right field(CharField, ChoicField etc) and widget"""
```
glennmatthews, line 784:
```suggestion
        """Assert that lookup value implements the right field (CharField, ChoiceField etc.) and widget."""
```

nautobot/utilities/tests/test_forms.py
```diff
@@ -615,3 +653,230 @@ def test_api_select_add_query_param_with_utf8(self):
...
+                {
+                    "class": "nautobot-select2-api lookup_type-select",
+                    "placeholder": None,
+                    "data-query-param-field_name": '["$lookup_field"]',
+                    "data-contenttype": "dcim.site",
+                    "data-url": "/api/lookup-choices/",
+                },
+            )
+            self.assertIsInstance(form.fields["lookup_value"], forms.ChoiceField)
+            self.assertTrue(isinstance(form.fields["lookup_value"].widget, APISelectMultiple))
```
glennmatthews, line 822:
```suggestion
            self.assertIsInstance(form.fields["lookup_value"].widget, APISelectMultiple)
```

nautobot/utilities/tests/test_forms.py
```diff
@@ -615,3 +653,230 @@ def test_api_select_add_query_param_with_utf8(self):
...
+                    "data-url": "/api/lookup-choices/",
+                    "placeholder": None,
+                },
+            )
+            self.assertIsInstance(form.fields["lookup_value"], forms.ChoiceField)
+            self.assertEqual(
+                form.fields["lookup_value"].widget.attrs,
+                {"class": "form-control nautobot-select2-static lookup_value-input form-control"},
+            )
+            self.assertTrue(isinstance(form.fields["lookup_value"].widget, StaticSelect2))
```
glennmatthews, line 853:
```suggestion
            self.assertIsInstance(form.fields["lookup_value"].widget, StaticSelect2)
```

nautobot/utilities/tests/test_forms.py
```diff
@@ -615,3 +653,230 @@ def test_api_select_add_query_param_with_utf8(self):
...
+                form.fields["lookup_type"].widget.attrs,
+                {
+                    "class": "nautobot-select2-api lookup_type-select",
+                    "data-contenttype": "dcim.site",
+                    "data-query-param-field_name": '["$lookup_field"]',
+                    "data-url": "/api/lookup-choices/",
+                    "placeholder": None,
+                },
+            )
+            self.assertTrue(isinstance(form.fields["lookup_value"].widget, DatePicker))
```
glennmatthews, line 868:
```suggestion
            self.assertIsInstance(form.fields["lookup_value"].widget, DatePicker)
```

nautobot/utilities/utils.py
```diff
@@ -689,3 +705,201 @@ def pretty_str(self, node=None, depth=0):
...
+        DateTimePicker,
+        DynamicModelMultipleChoiceField,
+        StaticSelect2,
+        StaticSelect2Multiple,
+        TimePicker,
+    )
+
+    # TODO(timizuo): Raise Error if filterset not found
+    filterset = get_filterset_for_model(model)()
+    # TODO(T=timizuo): Raise Error if field not found
```
glennmatthews, line 775: TODO?

nautobot/utilities/utils.py
```diff
@@ -689,3 +705,201 @@ def pretty_str(self, node=None, depth=0):
...
+            if lookup_type in filterset_class_fields:
+                lookup_field = re.sub(r"__\w+", "", lookup_type)
+                lookup_value = request_querydict.getlist(lookup_type)
+
+                query_dict.setlistdefault(lookup_field_placeholder % num, [lookup_field])
+                query_dict.setlistdefault(lookup_type_placeholder % num, [lookup_type])
+                query_dict.setlistdefault(lookup_value_placeholder % num, lookup_value)
+                num += 1
+
+    query_dict.setdefault("form-TOTAL_FORMS", num if num > 2 else 3)
```
glennmatthews, line 869: Perhaps?

```suggestion
    query_dict.setdefault("form-TOTAL_FORMS", max(num, 3))
```

nautobot/utilities/utils.py
```diff
@@ -689,3 +705,201 @@ def pretty_str(self, node=None, depth=0):
...
+    return query_dict
+
+
+def is_single_choice_field(filterset_class, field_name):
+    # Some filter parameters do not accept multiple values, e.g DateTime fields, boolean fields etc.
+    field = filterset_class().filters.get(field_name)
+    if field is None:
+        pass
+        # TODO (timizuo): handle filed not exist
```
glennmatthews, line 878: TODO

nautobot/project-static/css/base.css
```diff
@@ -388,6 +388,21 @@ ul.nav-tabs li a {
...
+.select2-selection--multiple.p-5 {
+    padding: 5px;
+}
+.select2-container--bootstrap .select2-selection--multiple .select2-selection__clear.remove-filter-param.m-0 {
+    margin: 0px
+}
+.select2-container--bootstrap .select2-selection--multiple .select2-selection__choice.chip {
+    margin-left: 0px;
+    margin-right: 3px;
+}
```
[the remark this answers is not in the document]
timizuoebideri1: @christhant

nautobot/core/templates/generic/object_list.html
timizuoebideri1, line 139:
```suggestion
        lookup_type_value = $(this).find(".lookup_type-select").val();
        lookup_value = $(this).find(".lookup_value-input");
        lookup_value.attr("name", lookup_type_value);
```

nautobot/core/templates/generic/object_list.html
```diff
@@ -83,8 +114,67 @@ <h1>{% block title %}{{ content_type.model_class|meta:"verbose_name_plural"|bett
...
+        lookup_value = $(this).find(".lookup_value-input")
+        lookup_value.attr("name", lookup_type_value)
+    })
+
+    // Remove applied filters
+    $(".remove-filter-param").on("click", function(){
+        let query_params = location.search
+        let type = $(this).attr("data-field-type")
+        let field_value = $(this).attr("data-field-value")
+        let query_string = location.search.substr(1).split("&")
```
timizuoebideri1, line 147:
```suggestion
        let query_params = location.search;
        let type = $(this).attr("data-field-type");
        let field_value = $(this).attr("data-field-value");
        let query_string = location.search.substr(1).split("&");
```

nautobot/core/templates/generic/object_list.html
```diff
@@ -83,8 +114,67 @@ <h1>{% block title %}{{ content_type.model_class|meta:"verbose_name_plural"|bett
...
+
+        if (type === "parent")
+            query_string = query_string.filter(item => item.search(field_value) < 0)
+        else {
+            let parent = $(this).attr("data-field-parent")
```
timizuoebideri1, line 152:
```suggestion
            let parent = $(this).attr("data-field-parent");
```

nautobot/core/templates/generic/object_list.html
```diff
@@ -83,8 +114,67 @@ <h1>{% block title %}{{ content_type.model_class|meta:"verbose_name_plural"|bett
...
+            query_string = query_string.filter(item => item.search(parent + "=" + field_value) < 0)
+        }
+        location.replace("?" + query_string.join("&"))
+    })
```
timizuoebideri1, line 156:
```suggestion
        location.replace("?" + query_string.join("&"));
    });
```

nautobot/core/templates/generic/object_list.html
```diff
@@ -83,8 +114,67 @@ <h1>{% block title %}{{ content_type.model_class|meta:"verbose_name_plural"|bett
...
+
+    // Clear new row values upon creation
+    $(".dynamic-filterform-add .add-row").click(function(){
+        let new_fields_parent_element = $(".dynamic-filterform").last()
+        let lookup_field_classes = [".lookup_field-select", ".lookup_type-select", ".lookup_value-input"]
+        lookup_field_classes.forEach(field_class => {
+            let element = new_fields_parent_element.find(field_class)
+            element.val(null).trigger('change')
+        })
```
timizuoebideri1, line 174:
```suggestion
        let lookup_field_classes = [".lookup_field-select", ".lookup_type-select", ".lookup_value-input"];
        lookup_field_classes.forEach(field_class => {
            let element = new_fields_parent_element.find(field_class);
            element.val(null).trigger('change')
        })
```

nautobot/project-static/js/forms.js
```diff
@@ -402,26 +440,105 @@ function jsify_form(context) {
...
```
timizuoebideri1, line 463:
```suggestion
    this_context = $(context);
```

nautobot/project-static/js/forms.js
```diff
@@ -402,26 +440,105 @@ function jsify_form(context) {
...
+        if ($(this)[0] == lookup_field_element[0]) {
+            lookup_type_element.val(null).trigger('change');
+        }
+        lookup_value_element.val(null).trigger('change')
+
+    })
+}
+
+function initializeSortableList(context){
+    this_context = $(context)
```
timizuoebideri1, line 507:
```suggestion
    this_context = $(context);
```

nautobot/project-static/js/forms.js
```diff
@@ -402,26 +440,105 @@ function jsify_form(context) {
...
+        });
+    });
+    this_context.find('#select-all-options').bind('click', function() {
+        var select_id = '#' + $(this).attr('data-target');
+        $(select_id + ' option').prop('selected',true);
+    });
+}
+
+function initializeImagePreview(context){
+    this_context = $(context)
```
timizuoebideri1, line 538:
```suggestion
    this_context = $(context);
```

nautobot/project-static/js/forms.js
```diff
@@ -456,37 +573,89 @@ function jsify_form(context) {
...
-        var countSelectedOptions = $(select_id + ' option:selected').length;
-        $(select_id + ' option:selected').each(function () {
-            var newPos = $(select_id + ' option').index(this) + countSelectedOptions;
-            if (newPos < countOptions) {
-                $(select_id + ' option').eq(newPos).after("<option value='" + $(this).val() + "' selected='selected'>" + $(this).text() + "</option>");
-                $(this).remove();
-            }
-        });
+function initializeSelectAllForm(context){
+    this_context = $(context)
```
timizuoebideri1, line 579:
```suggestion
    this_context = $(context);
```

nautobot/utilities/tests/test_utils.py
```diff
@@ -346,3 +354,153 @@ def test_slugify_dashes_to_underscores(self):
...
+        with self.subTest("Test unknown field"):
+            lookup_expr = get_all_lookup_expr_for_field(Site, "unknown_field")
+            self.assertEqual(lookup_expr, [])
+
+    def test_get_filterset_parameter_form_field(self):
+        with self.subTest("Test get CharFields"):
+            site_fields = ["comments", "name", "contact_email", "physical_address", "shipping_address"]
+            for field_name in site_fields:
+                form_field = get_filterset_parameter_form_field(Site, field_name)
+                self.assertTrue(isinstance(form_field, forms.CharField))
```
timizuoebideri1, line 410:
```suggestion
                self.assertIsInstance(form_field, forms.CharField)
```

nautobot/utilities/tests/test_forms.py
bryanculver: Need to jump back in to revalidate what I was addressing here. I had changed the implementation somewhat because it had originally used `base_filters` instead of the instantiated filters which end up being different and less complete.

nautobot/utilities/utils.py
```diff
@@ -672,3 +704,282 @@ def pretty_str(self, node=None, depth=0):
...
+    return verbose_name + label
+
+
+def get_all_lookup_exper_for_field(model, field_name):
+    """
+    Return all lookup expressions for `field_name` in `model` filterset
+    """
+    filterset = get_filterset_for_model(model).base_filters
+    if field_name.startswith("has_"):
+        return [{"id": field_name, "name": "exact"}]
```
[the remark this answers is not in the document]
timizuoebideri1: Yes, but line 745 shows that it causes a conflict. This logic yields more lookup expr than anticipated if this use case is not addressed here.

nautobot/extras/models/groups.py
```diff
@@ -393,29 +396,8 @@ def get_initial(self):
         This is intended for use to populate the dynamically-generated filter form created by
         `generate_filter_form()`.
         """
-        filter_fields = self.get_filter_fields()
```
jathanism, line 396: Can you please go ahead and remove the `FIXME` comment above here? Your changes obviate it. Thank you!

nautobot/utilities/forms/forms.py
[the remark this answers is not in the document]
glennmatthews: Presumably the issue with custom fields was due to the fact that the field_name starts with `cf_`, so it was returning something like `Cf my custom field` instead of `My custom field`?

nautobot/utilities/templates/utilities/templatetags/filter_form.html
```diff
@@ -0,0 +1,52 @@
+{% load form_helpers %}
```
glennmatthews, line 1: Would it make sense to name this file to match the templatetag, i.e. `dynamic_filter_form_modal.html` instead?

nautobot/core/api/views.py
```diff
@@ -702,3 +710,140 @@ def execute_graphql_request(self, request, data, query, variables, operation_nam
...
+        contenttype = request.GET.get("content_type")
+        app_label, model_name = contenttype.split(".")
+        try:
+            model = ContentType.objects.get(app_label=app_label, model=model_name).model_class()
+        except ContentType.DoesNotExist:
+            return Response(
+                "content_type not found",
+                status=400,
+            )
+        model_form = get_form_for_model(model)
```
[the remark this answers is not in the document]
timizuoebideri1: Which is preferable: logging a warning and returning an input field, or returning a 400 error "Form missing"? Considering that an input field would generally still function

nautobot/core/views/generic.py
```diff
@@ -313,7 +323,8 @@ def get(self, request):
             "permissions": permissions,
             "action_buttons": valid_actions,
             "table_config_form": TableConfigForm(table=table),
-            "filter_form": filter_form,
+            "filter_params": display_filter_params,
+            "filter_form": self.dynamic_filter_form,
```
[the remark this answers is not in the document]
timizuoebideri1: 1. I believe that 'Filterset' naming standard should be mandated,
2. Due to the fundamental differences between FilterForm and "DynamicFilterForm," it would not function.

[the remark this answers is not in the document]
glennmatthews: I agree with you on 1. but the fact remains there may be existing plugins that are not following this convention, and we'd like to avoid breaking them if at all possible.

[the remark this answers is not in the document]
timizuoebideri1: Do you want support for both FilterForm and DynamicFilterForm?
If yes, I think we should create an issue, cause I taught, the objective was to drop support for FilterForm and use DynamicFilterForm instead.

nautobot/core/templates/generic/object_list.html
timizuoebideri1: Yes they need to be updated

nautobot/project-static/js/forms.js
```diff
@@ -453,37 +570,89 @@ function jsify_form(context) {
...
-        var select_id = '#' + $(this).attr('data-target');
-        $(select_id + ' option').prop('selected',true);
+}
+
+function initializeResultPerPageSelection(context){
+    this_context = $(context);
+    this_context.find('select#per_page').change(function() {
+        this.form.submit();
     });
+}
```
bryanculver, line 591: I broke this when refactoring. I'll jump in and re-add it. Sorry!

bryanculver: Fixed.

bryanculver (approved):
Handful of nitpicks but overall I think this is a sweeping improvement.

Some of the feedback is dependent on #2572 getting merged into this and well as some others might be worthwhile to fix before being shipped.

nautobot/core/api/views.py
```diff
@@ -702,3 +709,90 @@ def execute_graphql_request(self, request, data, query, variables, operation_nam
...
+            )
+        contenttype = request.GET.get("content_type")
+        field_name = request.GET.get("field_name")
+        app_label, model_name = contenttype.split(".")
+        try:
+            model = ContentType.objects.get(app_label=app_label, model=model_name).model_class()
+        except ContentType.DoesNotExist:
+            return Response(
+                "content_type not found",
+                status=400,
```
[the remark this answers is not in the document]
timizuoebideri1: When the primary object is not found, I believe a 404 is returned, not when an improper parameter is supplied.

nautobot/utilities/utils.py
```diff
@@ -689,3 +706,213 @@ def pretty_str(self, node=None, depth=0):
...
+    query_dict = QueryDict(mutable=True)
+    filterset_class_fields = filterset_class().filters.keys()
+
+    query_dict.setdefault("form-INITIAL_FORMS", 0)
+    query_dict.setdefault("form-MIN_NUM_FORMS", 0)
+    query_dict.setdefault("form-MAX_NUM_FORMS", 100)
+
+    lookup_field_placeholder = "form-%d-lookup_field"
+    lookup_type_placeholder = "form-%d-lookup_type"
+    lookup_value_placeholder = "form-%d-lookup_value"
```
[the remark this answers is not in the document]
timizuoebideri1: No, `DynamicFilterFormSet`/`formset_factory` requires this format.

nautobot/utilities/utils.py
```diff
@@ -689,3 +706,213 @@ def pretty_str(self, node=None, depth=0):
...
+        form_field_class.widget = StaticSelect2()
+        form_attr = {"choices": BOOLEAN_CHOICES}
+
+        form_field = form_field_class(**form_attr)
+    elif isinstance(field, DateTimeFilter):
+        form_field.widget = DateTimePicker()
+    elif isinstance(field, DateFilter):
+        form_field.widget = DatePicker()
+    elif isinstance(field, TimeFilter):
+        form_field.widget = TimePicker()
```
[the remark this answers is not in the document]
glennmatthews: At the very least let's get an issue opened to track this.

nautobot/core/api/urls.py
```diff
@@ -36,4 +38,13 @@
...
     path("plugins/", include((plugin_api_patterns, "plugins-api"))),
+    # Lookup Expr
+    path(
+        "filterset-fields/lookup-choices/", GetFilterSetFieldLookupExpressionChoicesAPI.as_view(), name="lookup_choices"
+    ),
+    path(
+        "filterset-fields/lookup-value-dom-element/",
+        GetFilterSetFieldDOMElementAPI.as_view(),
+        name="lookup_value_dom_element",
+    ),
```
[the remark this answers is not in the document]
glennmatthews: Yeah, it's awkward since core's view URLs and API URLs haven't been namespaced to date. It probably is the right direction to move for new URL patterns in core, though.

nautobot/core/api/views.py
```diff
@@ -702,3 +709,90 @@ def execute_graphql_request(self, request, data, query, variables, operation_nam
...
+        if "content_type" not in request.GET or "field_name" not in request.GET:
+            return Response(
+                "content_type and field_name are required parameters",
+                status=400,
+            )
+        contenttype = request.GET.get("content_type")
+        field_name = request.GET.get("field_name")
+        app_label, model_name = contenttype.split(".")
+        try:
+            model = ContentType.objects.get(app_label=app_label, model=model_name).model_class()
```
glennmatthews, line 735: IIRC `model_class` can potentially be `None` in the case where it's a leftover content-type from a plugin that's no longer installed. Probably need to handle that.

nautobot/core/api/views.py
```diff
@@ -702,3 +709,90 @@ def execute_graphql_request(self, request, data, query, variables, operation_nam
...
+
+    @extend_schema(exclude=True)
+    def get(self, request):
+        if "content_type" not in request.GET or "field_name" not in request.GET:
+            return Response(
+                "content_type and field_name are required parameters",
+                status=400,
+            )
+        contenttype = request.GET.get("content_type")
+        field_name = request.GET.get("field_name")
```
glennmatthews, line 732: Just thinking aloud here, what do you think about maybe making these part of the URL path instead of query-parameters, since they're required? I.e. `/api/core/filterset-fields/<app_label>.<model>/<field_name>/choices/` and `/api/core/filterset-fields/<app_label>.<model>/<field_name>/dom-element/`?

nautobot/core/templates/generic/object_list.html
```diff
@@ -20,14 +37,54 @@
     {% endif %}
 </div>
 <h1>{% block title %}{{ content_type.model_class|meta:"verbose_name_plural"|bettertitle }}{% endblock %}</h1>
+{% if filter_params %}
+<div class="filters-applied">
+    <b>Filters:</b>
+    {% for field, values in filter_params %}
+    <span class="select2 select2-container select2-container--bootstrap display-inline" dir="ltr">
```
glennmatthews, line 44: Still not sure I'm comfortable with using select2 classes for non-select2 elements. I'd rather have a distinct class for each of these new UI elements, even if it means duplicating some CSS.

nautobot/core/templates/generic/object_list.html
```diff
@@ -20,14 +37,54 @@
...
+                            data-field-parent={{ field }}
+                            data-field-value={{ value }}
+                            role="presentation"
+                    >×</span>{{ value }}
+                </li>
+                {% endfor %}
+            </ul>
+        </span>
+        </span>
+        <span class="dropdown-wrapper" aria-hidden="true"></span>
```
glennmatthews, line 79: What's this span do?

nautobot/core/templates/generic/object_list.html
```diff
@@ -20,14 +37,54 @@
     {% endif %}
 </div>
 <h1>{% block title %}{{ content_type.model_class|meta:"verbose_name_plural"|bettertitle }}{% endblock %}</h1>
+{% if filter_params %}
+<div class="filters-applied">
+    <b>Filters:</b>
+    {% for field, values in filter_params %}
+    <span class="select2 select2-container select2-container--bootstrap display-inline" dir="ltr">
+        <span class="selection">
```
glennmatthews, line 45: What's this span do?

nautobot/core/api/views.py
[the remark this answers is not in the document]
glennmatthews: I'm in favor of 404 here - the request isn't semantically invalid, it's just asking for something that isn't there.

nautobot/core/templates/generic/object_list.html
```diff
@@ -3,12 +3,29 @@
...
+<style>
+    .btn-filter-value {
+        background: #cfcfcf;
+        margin-right: 5px;
+    }
+    .filters-applied>.btn {
+        margin-right: 5px;
+        margin-bottom: 5px;
+    }
+</style>
```
glennmatthews, line 17: Is this still relevant/needed? I don't see `btn-filter-value` or any `btn` inside of `.filters-applied` in the current HTML.

nautobot/docs/release-notes/version-1.5.md
```diff
@@ -30,6 +30,10 @@ As a result, the value of this setting now defaults to `False`, disabling databa
 !!! important
     Users with existing `nautobot_config.py` files generated from earlier versions of Nautobot will still have `CACHEOPS_ENABLED = True` unless they modify or regenerate their configuration. If users no longer desire caching, please be sure to explicitly toggle the value of this setting to `False` and restart your Nautobot services.
 
+#### Redesigned List Filtering UI
```
glennmatthews, line 33: Are regular `FilterForm` classes now completely obsolete, or do they still serve a purpose? If obsolete,

1) Are we planning to remove them from the code base in 1.5 or does that need to be a 2.0 item?
2) There's various developer documentation that needs to be revised, for example:

```
nautobot/docs/development/best-practices.md:| Table filter forms       | `NautobotFilterForm`       | `BootstrapMixin`           |
nautobot/docs/development/generic-views.md:    filterset_form = forms.ExampleModelFilterForm
nautobot/docs/development/style-guide.md:    * Similarly, for filter forms, `nautobot.extras.forms.NautobotFilterForm` combines `nautobot.utilities.forms.BootstrapMixin`, `nautobot.extras.forms.CustomFieldModelFilterFormMixin`, and `nautobot.extras.forms.RelationshipModelFilterFormMixin`, and should be used where appropriate.
nautobot/docs/models/extras/status.md:- `nautobot.extras.forms.StatusModelFilterFormMixin` should be used to add a non-required, multiple-choice `status` filter field to UI filter forms. This multiple-choice field allows for multiple status values to be selected for filtering objects in list views in the web UI.
nautobot/docs/models/extras/status.md:    In prior Nautobot versions these mixins were named `StatusFilterFormMixin` and `StatusBulkEditFormMixin`; the old names are still available as aliases but will be removed in a future major release.
nautobot/docs/models/extras/status.md:- Filter forms must inherit from `nautobot.extras.forms.StatusModelFilterFormMixin`
nautobot/docs/models/extras/status.md:    In prior Nautobot releases the latter two mixins were named `StatusBulkEditFormMixin` and `StatusFilterFormMixin` respectively; the old names are still available as aliases but will be removed in a future major release.
nautobot/docs/plugins/development.md:Plugins can extend any model-based `FilterSet` and `FilterForm` classes that are provided by the Nautobot core.
nautobot/docs/plugins/development.md:* The original `FilterForm` must follow the pattern: `f"{model.__name__}FilterForm"`, e.g. `TenantFilterForm`
nautobot/docs/plugins/development.md:* The `filterform_fields` attribute must be a valid dictionary of Django form fields
nautobot/docs/plugins/development.md:    A plugin is not required to define both `filterset_fields` and `filterform_fields`.
nautobot/docs/plugins/development.md:    filterset_form_class = YourPluginModelFilterForm
nautobot/docs/plugins/development.md:* `filter_form`: The FilterForm object for the associated model
nautobot/docs/plugins/development.md:    'filter_form': <ProviderFilterForm bound=True, valid=Unknown, fields=(region;site;location;q;asn;tag)>,
nautobot/docs/plugins/development.md:    filterset_form_class = YourPluginModelFilterForm
```

nautobot/utilities/forms/forms.py
```diff
@@ -233,3 +244,107 @@ def __init__(self, table, *args, **kwargs):
...
+        required=False,
+        label="Value",
+    )
+
+    def __init__(self, *args, filterset=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        from nautobot.utilities.forms import add_blank_choice  # Avoid circular import
+
+        # cls.model is set at `dynamic_formset_factory()`
+        self.filterset = filterset or getattr(self, "filterset", None)
```
glennmatthews, line 273: Do we need to support both of these patterns or can we standardize on one or the other?

nautobot/utilities/forms/forms.py
```diff
@@ -233,3 +244,107 @@ def __init__(self, table, *args, **kwargs):
...
+
+        # Raise exception if `cls.filterset` not set and `filterset` not passed
+        if self.filterset is None:
+            raise AttributeError("'DynamicFilterForm' object requires `filterset` attribute")
+
+        model = self.filterset._meta.model
+
+        if self.filterset is not None:
+            filterset = self.filterset()
```
glennmatthews, line 282: Should `cls.filterset` / the `filterset=...` parameter be renamed to `filterset_class` for clarity?

nautobot/utilities/forms/forms.py
```diff
@@ -233,3 +244,107 @@ def __init__(self, table, *args, **kwargs):
...
+
+            # Update lookup_type and lookup_value fields to match expected field types derived from data
+            # e.g status expects a ChoiceField with APISelectMultiple widget, while name expects a CharField etc.
+            if "data" in kwargs and "prefix" in kwargs:
+                data = kwargs["data"]
+                prefix = kwargs["prefix"]
+                lookup_type = data.get(prefix + "-lookup_type")
+                lookup_value = data.getlist(prefix + "-lookup_value")
+
+                if lookup_type and lookup_value and lookup_type in self.filterset_filters:
```
glennmatthews, line 298: if `lookup_type` is not in `self.filterset_filters`, is that something we should log a warning or error about?

nautobot/core/api/urls.py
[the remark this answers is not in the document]
glennmatthews: Also, the names for these views should ideally be namespaced as well, something like `core-api:filtersetfield-list-lookupchoices` and `core-api:filtersetfield-retrieve-lookupvaluedomelement`.

nautobot/utilities/templates/utilities/templatetags/dynamic_filter_form_modal.html
```diff
@@ -0,0 +1,52 @@
...
+        <div class="modal-content">
+            <div class="modal-header">
+                <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+                <h4 class="modal-title">Filter Form</h4>
+            </div>
+            <div class="modal-body">
+                {% if form %}
+                    <form id="dynamic-filter-form" class="form-horizontal" method="get" action=".">
+                        <div class="">
+                            <table class="table" id="secrets" style="table-layout: fixed;">
```
glennmatthews, line 15: `table-layout: fixed` makes it so the table doesn't adjust columns to fit their contents, right? Is that why there's all that dead whitespace to the right of the "delete" button column? Not blocking for this PR but it'd be nice to fix that up with some smarter column widths.

nautobot/utilities/templates/utilities/templatetags/dynamic_filter_form_modal.html
```diff
@@ -0,0 +1,52 @@
...
+                                {% endfor %}
+                            </table>
+                        </div>
+
+                        <div class="text-right noprint">
+                            <button type="submit" class="btn btn-info">
+                                <span class="mdi mdi-magnify" aria-hidden="true"></span> Apply
+                            </button>
+                            <a href="." class="btn btn-default">
+                                <span class="mdi mdi-close-thick" aria-hidden="true"></span> Clear
```
glennmatthews, line 43: Just checking - does this close the form (and revert to any previously applied filters) or does it actually clear all filters that may have been applied by a previous dialog?

nautobot/utilities/tests/test_forms.py
```diff
@@ -615,3 +654,245 @@ def test_api_select_add_query_param_with_utf8(self):
...
+                    ("tenant_group", "Tenant Group (slug)"),
+                    ("tenant_group_id", "Tenant Group (ID)"),
+                    ("tenant_id", 'Tenant (ID) (deprecated, use "tenant" filter instead)'),
+                    ("time_zone", "Time zone"),
+                    ("vlan_groups", "Vlan groups (slug or ID)"),
+                    ("vlans", "Vlans"),
+                ],
+            )
+
+        with self.subTest("Assert form generates the correct base_filters"):
```
glennmatthews, line 744: It's not very clear to me what this is testing, especially given the data transformation involved. Would you mind adding some more context?

nautobot/utilities/tests/test_utils.py
```diff
@@ -346,3 +356,169 @@ def test_slugify_dashes_to_underscores(self):
...
+        with self.subTest("Test get CharFields"):
+            site_fields = ["comments", "name", "contact_email", "physical_address", "shipping_address"]
+            for field_name in site_fields:
+                form_field = get_filterset_parameter_form_field(Site, field_name)
+                self.assertIsInstance(form_field, forms.CharField)
+
+            device_fields = ["serial", "name"]
+            for field_name in device_fields:
+                form_field = get_filterset_parameter_form_field(Device, field_name)
+                self.assertTrue(isinstance(form_field, forms.CharField))
```
glennmatthews, line 428: `assertIsInstance` here and below, please.

nautobot/core/api/views.py
timizuoebideri1: `field_name` is obtained in the UI when the field has been selected.

nautobot/docs/release-notes/version-1.5.md
timizuoebideri1: I believe `FilterForm` would be obsolete @bryanculver ?

timizuoebideri1: Cause currently they don't serve any purpose

nautobot/utilities/forms/forms.py
timizuoebideri1: Both must be supported, except we want developers to consistently declare DynamicFilterForm in this manner:
```python
dynamic_filterform = DynamicFilterForm
dynamic_filterform.filterset = ...
dynamic_filterform(...)
```

nautobot/utilities/forms/forms.py
timizuoebideri1: In the UI, a warning would have been issued for this.

nautobot/utilities/templates/utilities/templatetags/dynamic_filter_form_modal.html
timizuoebideri1: It close the form (and revert to any previously applied filters)

nautobot/utilities/templates/utilities/templatetags/dynamic_filter_form_modal.html
timizuoebideri1: The white space to the right of the "delete" button results from the equal width of each column.

nautobot/utilities/utils.py
```diff
@@ -689,3 +706,213 @@ def pretty_str(self, node=None, depth=0):
...
+    final_filter_params = {}
+    for field in filter_params.keys():
+        if filter_params.get(field):
+            # `is_single_choice_field` implements `get_filterset_field`, which throws an exception if a field is not found.
+            # If an exception is thrown, instead of throwing an exception, set `_is_single_choice_field` to 'False'
+            # because the fields that were not discovered are still necessary.
+            try:
+                _is_single_choice_field = is_single_choice_field(filterset_class, field)
+            except FilterSetFieldNotFound:
+                _is_single_choice_field = False
```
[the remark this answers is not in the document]
timizuoebideri1: `is_single_choice_field` implements `get_filterset_field` which raises this error when `filterset_class().filters.get(field_name)` is `None`.

A simple key search in the filters dict

nautobot/core/api/views.py
glennmatthews: Sorry, I'm not sure I understand how that's a response to my idea?

nautobot/utilities/templates/utilities/templatetags/dynamic_filter_form_modal.html
glennmatthews: OK, so maybe the button should be "Cancel" rather than "Clear"?

nautobot/utilities/forms/forms.py
glennmatthews: Hmm. How would the UI trigger this case? If it's not something that should be possible under normal usage, I feel like a server-side log message would be appropriate.

nautobot/core/api/views.py
```diff
@@ -732,16 +732,22 @@ def get(self, request):
         field_name = request.GET.get("field_name")
         app_label, model_name = contenttype.split(".")
         try:
-            model = ContentType.objects.get(app_label=app_label, model=model_name).model_class()
+            model_contenttype = ContentType.objects.get(app_label=app_label, model=model_name)
+            model = model_contenttype.model_class()
+            if model is None:
+                return Response(
+                    f"model for content_type: <{model_contenttype}> not found",
+                    status=400,
```
glennmatthews, line 740: This feels like maybe a 500 error? A valid content-type was requested, but we couldn't find the model class corresponding to it?

nautobot/core/api/views.py
timizuoebideri1: Here #[2348L306](https://github.com/nautobot/nautobot/pull/2348/files#diff-98f1dfde62e7c5d2714e5fc4be0a06544d109330ccfb842b063be81ec3894559R306)
```python
self.fields["lookup_type"].widget.attrs["data-url"] = reverse("core-api:filtersetfield-list-lookupchoices")
```
The endpoint is added to the `lookup_type` field, however, the `field_name` is currently unknown. The `field name` is obtained only after the user has picked the field in the UI.

timizuoebideri1: hence, using this would be virtually impossible. `reverse("...", kwargs={"model": .., "app_label":.., "field_name":...})`

nautobot/utilities/forms/forms.py
timizuoebideri1: Before using the `DynamicFilterSet`, `filterset = self.filterset(filter params, self.queryset)` in the `ObjectListView` handled this invalid field.

However, I believe it's fantastic to also log a warning.

nautobot/docs/release-notes/version-1.5.md
bryanculver: I don't know where they will be displayed anymore. The intention of this is to expand the options of filtering as well as provide more room on the table.

I guess there's no reason we can't leave the `FilterForm`s in for 1.5 but 🤷

Thoughts @lampwins

nautobot/core/templates/generic/object_list.html
christhant: @glennmatthews I can create separate classes clone from select2.

nautobot/core/templates/generic/object_list.html
christhant: @glennmatthews those spans are used conjunction with select2 class. it was cloned from select2

nautobot/utilities/utils.py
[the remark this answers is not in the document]
glennmatthews: Opened #2634.

nautobot/core/templates/generic/object_list.html
```diff
@@ -20,14 +55,47 @@
     {% endif %}
 </div>
 <h1>{% block title %}{{ content_type.model_class|meta:"verbose_name_plural"|bettertitle }}{% endblock %}</h1>
+{% if filter_params %}
+<div class="filters-applied">
+    <b>Filters:</b>
+    {% for field, values in filter_params %}
+    <span class="filter-container display-inline" dir="ltr">
```
glennmatthews, line 62: Why do we need two separate classes here? Can we remove the `display-inline` class and just apply all necessary CSS to `filter-container`?

nautobot/core/templates/generic/object_list.html
```diff
@@ -20,14 +55,47 @@
...
+        <span class="selection">
+        <span
+                class="filter-selection filter-selection--multiple p-5">
```
glennmatthews, line 65: Why do we need three separate classes here? Can we remove `filter-selection--multiple` and `p-5` and just apply all styling to the `filter-selection` class?

nautobot/core/templates/generic/object_list.html
```diff
@@ -20,14 +55,47 @@
...
+            <b>{{ field }}:</b>
+            <span
+                    class="filter-selection__clear remove-filter-param m-0"
```
glennmatthews, line 68: Why do we need three separate classes here? Can we remove `filter-selection__clear` and `m-0` and just apply all CSS styling to the `remove-filter-param` class?

nautobot/core/templates/generic/object_list.html
glennmatthews: I don't see any styling for this class - can this span be removed?

glennmatthews (approved):
Code changes look good. Looking at the test failures now.

nautobot/extras/tests/test_api.py
```diff
@@ -3242,7 +3242,7 @@ def test_all_relevant_content_types_assigned_to_tags_with_empty_content_types(se
         tag = Tag.objects.get(slug=self.create_data[0]["slug"])
         self.assertEqual(
             tag.content_types.count(),
-            TaggableClassesQuery().as_queryset.count(),
+            TaggableClassesQuery().as_queryset().count(),
```
glennmatthews, line 3245: Looks like there are a couple of additional cases of `as_queryset` at lines 3268 and 3271 that need to be updated.

glennmatthews:
Can you update the TODO items in the PR description with where we're at now - are there any remaining open items?

nautobot/utilities/templates/utilities/templatetags/filter_form_modal.html
```diff
@@ -0,0 +1,87 @@
...
+                                    {% endfor %}
+                                </table>
+                            </div>
+
+                            <div class="text-right noprint">
+                                <button type="submit" class="btn btn-primary">
+                                    <span class="mdi mdi-magnify" aria-hidden="true"></span> Apply
+                                </button>
+                                <a href="." class="btn btn-default">
+                                    <span class="mdi mdi-close-thick" aria-hidden="true"></span> Cancel
```
jathanism, line 77: This "default" tab button is labeled "Clear" and the dynamic tab button is labeled "Cancel", but they both in fact clear the search criteria. This was confusing for me.
```suggestion
                                    <span class="mdi mdi-close-thick" aria-hidden="true"></span> Clear
```

jathanism (changes requested):
Another issue I ran into is this, and not sure how we can solve it, but I wanted to point it out:

1. Click "Filter", then "Advanced" and input an advanced filter such as "name regex = ^ams01".
2. Click "Apply"
3. Observe that the filter `name__re=^ams01` is displayed in list view
4. Click "Filter" again, then select any "Default" filter such as "has interfaces = Yes"
5. Click "Apply"
6. Observe that the previous filter was replaced and that only `has_interfaces=True` is displayed in list view.

This appears to only be the case if you implement "default" filters. If you select the same set of filter criteria using the "advanced" tab, they always persist.

bryanculver (approved):
@timizuoebideri1 Let's go ahead and merge this. So many of the essential functions are here and we aren't going to make any significant breaking changes from here on out. The remainder can be tweaks and patches going forward.

bryanculver:
> Another issue I ran into is this, and not sure how we can solve it, but I wanted to point it out:
> 
> 1. Click "Filter", then "Advanced" and input an advanced filter such as "name regex = ^ams01".
> 2. Click "Apply"
> 3. Observe that the filter `name__re=^ams01` is displayed in list view
> 4. Click "Filter" again, then select any "Default" filter such as "has interfaces = Yes"
> 5. Click "Apply"
> 6. Observe that the previous filter was replaced and that only `has_interfaces=True` is displayed in list view.
> 
> This appears to only be the case if you mix "advanced' and "default" filters using the "default" tab. If you select the same set of filter criteria using only the "advanced" tab, they always persist.

Let's acknowledge this as a known bug and fix forward.

jathanism:
> Let's acknowledge this as a known bug and fix forward.

https://github.com/nautobot/nautobot/issues/2732

jathanism (approved):

PR #2349, opened 2022-08-31
Fix #2342, and updates for Poetry 1.2

### LINKED ISSUE #2342: Docker containers should not include a poetry cache
### Proposed Changes

Run `poetry cache clear -n --all pypi` in the `final` step of the Dockerfile

### Justification

Reduce file size of docker container by removing cached wheel files that are no longer needed

glennmatthews:
# Closes: #2342 
# What's Changed

- Poetry 1.2.0 is released now
   - It has an underdocumented behavior change that `poetry install --extras ...` followed by `poetry install` (without any `--extras`) will result in Poetry removing the previously installed extras.
- Change our install of Nautobot itself in the `dev` image to account for this change
- `poetry install --no-dev` is deprecated in favor of `poetry install --only main`, so I've made that change.
- Add `poetry cache clear --all PyPI` to the `final-dev` image build, fixing #2342 (with a workaround for a couple of Poetry bugs)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -137,7 +137,7 @@ RUN apt-get update && \
 # https://python-poetry.org/docs/master/#installing-with-the-official-installer
 # This also makes it so that Poetry will *not* be included in the "final" image since it's not installed to /usr/local/
 RUN curl -sSL https://install.python-poetry.org -o /tmp/install-poetry.py && \
-    python /tmp/install-poetry.py && \
+    python /tmp/install-poetry.py --version 1.2.0 && \
     rm -f /tmp/install-poetry.py
 
 # Add poetry install location to the $PATH
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -266,6 +266,7 @@ RUN pip install --no-deps --no-cache-dir /source/dist/*.whl && \
 # Remove Poetry's cache of PyPI files to keep image size down
 # poetry cache clear --no-interactive actually doesn't clear anything - see
 # https://github.com/python-poetry/poetry/issues/4998
+# hadolint ignore=DL4006
 RUN yes | poetry cache clear --all PyPI
 
 ################################ Stage: final (production-ready image)
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -162,7 +162,7 @@ WORKDIR /source
 # pyuwsgi wheel doesn't support ssl so we build it from source
 # https://github.com/nautobot/nautobot/issues/193
 RUN pip install --no-cache-dir --no-binary=pyuwsgi pyuwsgi==${PYUWSGI_VER} && \
-    poetry install --no-root --no-dev --no-ansi --extras all
+    poetry install --no-root --only main --no-ansi --extras all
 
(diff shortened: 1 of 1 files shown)


glennmatthews:
Weird, hadolint passes locally. Wonder if I have an old version or something. Will fix when I get time.

glennmatthews:
Actually it's the other way around - CI is using hadolint 2.7 but a locally built dev image has hadolint 2.10. Fails regardless when I run it locally against the actual branch in question 🤦

bryanculver:
> * Poetry 1.2.0 is released now
>   
>   * It has an underdocumented behavior change that `poetry install --extras ...` followed by `poetry install` (without any `--extras`) will result in Poetry removing the previously installed extras.
> * Change our install of Nautobot itself in the `dev` image to account for this change

This seems to be a necessary fix as when we ship a new image we'll probably bring in Poetry 1.2 and plugin's development/testing images won't include the necessary dev dependencies at runtime: https://github.com/nautobot/nautobot-plugin-golden-config/blob/develop/development/Dockerfile#L42

gsnider2195:
Does the poetry installer at https://install.python-poetry.org/ support constraining the version with `>=1.2.0` and if so should we add that constraint as part of this change? I know the installer says it will install the latest version so this may not be an issue?

glennmatthews:
It does always install the latest version by default. We could (and probably should) constrain it with the `--version` argument so that it will use a specific version until we intentionally move to a newer one though.

bryanculver:
> It does always install the latest version by default. We could (and probably should) constrain it with the `--version` argument so that it will use a specific version until we intentionally move to a newer one though.

This is how most plugins are installing Poetry: https://github.com/networktocode/gh-action-setup-poetry-environment/blob/main/action.yml

glennmatthews:
> This is how most plugins are installing Poetry: https://github.com/networktocode/gh-action-setup-poetry-environment/blob/main/action.yml

Sure, though that's a virtualenv workflow, not a Docker workflow.

I've added an explicit `--version 1.2.0` to the `install-poetry` command so that we can migrate to newer versions on our own schedule rather than immediately upon release.

gsnider2195 (approved):

docker/Dockerfile
```diff
@@ -137,7 +137,7 @@ RUN apt-get update && \
 # https://python-poetry.org/docs/master/#installing-with-the-official-installer
 # This also makes it so that Poetry will *not* be included in the "final" image since it's not installed to /usr/local/
 RUN curl -sSL https://install.python-poetry.org -o /tmp/install-poetry.py && \
-    python /tmp/install-poetry.py && \
+    python /tmp/install-poetry.py --version 1.2.0 && \
```
bryanculver, line 140: Just double checking on the comment above. The installer _should_ install it in a place other than `/usr/local` but should we enforce this through providing `--path` or `$POETRY_HOME`?

bryanculver (approved):

docker/Dockerfile
glennmatthews: --path is for specifying where to find the file to install from, if I'm reading the script right. We could explicitly set POETRY_HOME I suppose but I feel like this is a pretty low risk.

PR #2356, opened 2022-09-01
Resolved #2350 - Add ability to age out a redis lock

### LINKED ISSUE #2350: Deadlock on get next prefix API
### Environment
* Nautobot version (Docker tag too if applicable):
* Python version: 3.8
* Database platform, version: 1.3.8 & 1.4.1
* Middleware(s):

### Steps to Reproduce
1. Perform API call for get next prefix
2. abruptly stop Nautobot service before call can complete
3. Restart server

### Expected Behavior
Lock ages out

### Observed Behavior
Places Nautobot in a deadlock state for get-next-prefix

### Proposed Fix
https://github.com/nautobot/nautobot/blob/v1.4.1/nautobot/ipam/api/views.py#L143
Set a configurable timeout that defaults to 300 seconds. A user should never be able to hit 5 minutes in an API call which would safely add the safe measure. And after 5 minutes if the lock is not gracefully released it will be aged out.

whitej6:
# Closes: #2350 
# What's Changed

Add ability to age out a redis lock

# TODO

- [ ] Unit, Integration Tests


nautobot/core/settings.py
```diff
@@ -76,7 +76,7 @@
 HTTP_PROXIES = None
 JOBS_ROOT = os.getenv("NAUTOBOT_JOBS_ROOT", os.path.join(NAUTOBOT_ROOT, "jobs").rstrip("/"))
 MAINTENANCE_MODE = False
-
+AVAILABLE_PREFIX_TIMEOUT = int(os.getenv("NAUTOBOT_AVAILABLE_PREFIX_TIMEOUT", 600))
```
glennmatthews, line 79: pylint will complain that we're defaulting an environment variable to an int rather than a str.

```suggestion
AVAILABLE_PREFIX_TIMEOUT = int(os.getenv("NAUTOBOT_AVAILABLE_PREFIX_TIMEOUT", "600"))
```

nautobot/docs/configuration/optional-settings.md
```diff
@@ -53,6 +53,16 @@ Please see the [official Django documentation on `ADMINS`](https://docs.djangopr
 
 ---
 
+## AVAILABLE_PREFIX_TIMEOUT
+
+Defaul: `600`
```
glennmatthews, line 58:
```suggestion
Default: `600`
```

nautobot/docs/configuration/optional-settings.md
```diff
@@ -53,6 +53,16 @@ Please see the [official Django documentation on `ADMINS`](https://docs.djangopr
...
+
+Environment Variable: `NAUTOBOT_AVAILABLE_PREFIX_TIMEOUT`
+
+Prevents a deadlock scenario where a Redis lock is achieved upon performing API to `/api/ipam/prefixes/{id}/available-prefixes/`. Default is set to 600 seconds (10 minutes) to be longer than any theorhetical API call time. This is to prevent a deadlock scenario where the server did not gracefully exit the with block when acquiring the Redis lock.
```
glennmatthews, line 62:
```suggestion
Prevents a deadlock scenario where a Redis lock is achieved upon performing API to `/api/ipam/prefixes/{id}/available-prefixes/`. Default is set to 600 seconds (10 minutes) to be longer than any theoretical API call time. This is to prevent a deadlock scenario where the server did not gracefully exit the with block when acquiring the Redis lock.
```

nautobot/ipam/api/views.py
```diff
@@ -140,7 +140,9 @@ def available_prefixes(self, request, pk=None):
         prefix = get_object_or_404(self.queryset, pk=pk)
         if request.method == "POST":
 
-            with cache.lock("available-prefixes", blocking_timeout=5):
+            with cache.lock(
+                "available-prefixes", blocking_timeout=5, timeout=get_settings_or_config("AVAILABLE_PREFIX_TIMEOUT")
+            ):
```
glennmatthews, line 145: Is there a similar issue with the "available-ips" feature?

whitej6: Yes there is, didn't think about that. I will update the var name and should along with comments.

nautobot/core/settings.py
```diff
@@ -76,7 +76,7 @@
 HTTP_PROXIES = None
 JOBS_ROOT = os.getenv("NAUTOBOT_JOBS_ROOT", os.path.join(NAUTOBOT_ROOT, "jobs").rstrip("/"))
 MAINTENANCE_MODE = False
-
+AVAILABLE_IP_PREFIX_TIMEOUT = int(os.getenv("NAUTOBOT_AVAILABLE_IP_PREFIX_TIMEOUT", "600"))
```
glennmatthews, line 79: 1. Maybe something more generic like `REDIS_LOCK_TIMEOUT`, so that this can be used if we determine the need for other similar locks in the future?
2. Location in this file seems pretty arbitrary. Maybe move it down to near the caching/Celery configs instead? (We don't really have any other general Redis configs in this file that I see).

nautobot/docs/configuration/optional-settings.md
```diff
@@ -53,6 +53,16 @@ Please see the [official Django documentation on `ADMINS`](https://docs.djangopr
 
 ---
 
+## AVAILABLE_IP_PREFIX_TIMEOUT
+
+Default: `600`
+
+Environment Variable: `NAUTOBOT_AVAILABLE_IP_PREFIX_TIMEOUT`
+
+Prevents a deadlock scenario where a Redis lock is achieved upon performing API to `/api/ipam/prefixes/{id}/available-prefixes/`. Default is set to 600 seconds (10 minutes) to be longer than any theoretical API call time. This is to prevent a deadlock scenario where the server did not gracefully exit the with block when acquiring the Redis lock.
```
glennmatthews, line 62: Perhaps:

```suggestion
Maximum duration of a Redis lock created when calling `/api/ipam/prefixes/{id}/available-prefixes/` or `/api/ipam/prefixes/{id}/available-ips/` to avoid inadvertently allocating the same prefix or IP to multiple simultaneous callers. Default is set to 600 seconds (10 minutes) to be longer than any theoretical API call time. This is to prevent a deadlock scenario where the server did not gracefully exit the `with` block when acquiring the Redis lock.
```

nautobot/ipam/api/views.py
```diff
@@ -140,7 +140,9 @@ def available_prefixes(self, request, pk=None):
         prefix = get_object_or_404(self.queryset, pk=pk)
         if request.method == "POST":
 
-            with cache.lock("available-prefixes", blocking_timeout=5):
+            with cache.lock(
+                "available-prefixes", blocking_timeout=5, timeout=get_settings_or_config("AVAILABLE_IP_PREFIX_TIMEOUT")
```
glennmatthews, line 144: Only need to use `get_settings_or_config` for properties that are alternatively configurable via the Constance (admin-config) UI. As presently implemented, just `settings.AVAILABLE_IP_PREFIX_TIMEOUT` would be more correct.

nautobot/core/settings.py
whitej6: Both makes sense.

nautobot/docs/configuration/optional-settings.md
whitej6: Done

nautobot/ipam/api/views.py
whitej6: Done

nautobot/core/settings.py
whitej6: Done

glennmatthews (approved):
Solid!

bryanculver (approved):
:shipit:

Applied to all existing cache.locks. Would possibly be a great example for a linter to check that calls to lock are called with a timeout.

PR #2357, opened 2022-09-02
Reduce/Eliminate Noisy Test Output [Ready for Review]

HanlinMiao:
The noisy output is due to job result logging jobs that are intended to fail. Solved by:
```
#Disable any logging message up to the severity of error.
logging.disable(logging.ERROR)
......(failure code)
#reenable logging
logging.disable(logging.NOTSET)
```
Celery.ping issue is a known issue https://github.com/celery/celery/issues/3642 and solved by setting the `perform_ping_check` argument to `False`
Screen Shot 2022-09-05 at 7 37 35 PM
🧑‍🍳😚


bryanculver:
🤔 Should all runs of `create_job_result_and_run_job` perform the logging error disable? Should that be an optional argument to the function?

HanlinMiao:
I dont think all runs of `create_job_result_and_run_job` should have logging error disabled. Just the ones that are intended to fail and throw an exception. Also, `create_job_result_and_run_job` runs the job and [job_result.log()](https://github.com/nautobot/nautobot/blob/a3c27aa74a54269b29ea0073c557c0aaa3114360/nautobot/extras/jobs.py#L1096) here is the reason there are errors logged in the unittest. If we set the argument of `log()` `logger = None` [here](https://github.com/nautobot/nautobot/blob/a3c27aa74a54269b29ea0073c557c0aaa3114360/nautobot/extras/models/jobs.py#L762), i think the noise during testing will stop as well. But I am not sure how to do that in the unittest. So I just disabled logging altogether.

gsnider2195 (approved):
Looks good!

Issue #2235, opened 2022-08-17
Location Names Don't Provide Enough Context in Dropdown


bryanculver:
### Discussed in https://github.com/nautobot/nautobot/discussions/2233

<div type='discussions-op-text'>

<sup>Originally posted by **Phill93** August 17, 2022</sup>
Hi,

can someone give me a tip how to best create my location structure in 1.4.0?

I have several buildings that are designated with numbers (e.g. 101,102...) in these the rooms are again designated with numbers (e.g. 001, 101, 201).

I have now created the following location structure in my Nautobot instance: Building -> Floor -> Room (102 -> G -> 101)

If I now have the room 101 in several buildings, Nautobot lists me the 101 several times in the location field without differentiation.

<img width="905" alt="Bildschirmfoto 2022-08-17 um 17 40 31" src="https://user-images.githubusercontent.com/1454797/185182721-a7192cfa-f74b-4952-b1aa-ebdc9c3954f5.png">

Greetings

Daniel

Translated with www.DeepL.com/Translator (free version)

</div>

christhant:
After looking into the content setup and code,  
1. it appears that location types are not allowed to setup with the same name. i.e. all the location type names must be unique.
2. location type dropdown is using Python generic DynamicModelChoiceField. It probably needs to be created with custom model that can take different set of parameters and display hierarchical data structure in the dropdown. 


![Image](https://user-images.githubusercontent.com/110844839/187217204-80d58bbb-1aad-4848-86b9-4a2e64459ca0.png)

bryanculver:
> After looking into the content setup and code,
> 
> 1. it appears that location types are not allowed to setup with the same name. i.e. all the location type names must be unique.
> 2. location type dropdown is using Python generic DynamicModelChoiceField. It probably needs to be created with custom model that can take different set of parameters and display hierarchical data structure in the dropdown.

Take a look at the API response for the Locations endpoint. The Select2 widget should grab a `display_name` field if present in the response. Clicking this dropdown calls said API:

<img width="1495" alt="image" src="https://user-images.githubusercontent.com/31187/187224763-45a78ab4-0f66-4a53-88f0-471a9042ce7c.png">
<img width="1498" alt="image" src="https://user-images.githubusercontent.com/31187/187224879-87001300-d747-4191-889b-bc5b26aab151.png">

PR #2359, opened 2022-09-02
nautobot.2235. location name update

christhant:
# Closes: #2235 
# What's Changed

Changes
- Locations data model : included "display" property. prefix with parent name allows location name identity if there are the same name exists under different parents.
image

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/models/locations.py
+++ b/nautobot/dcim/models/locations.py
@@ -199,6 +199,20 @@ class Location(TreeNode, StatusModel, PrimaryModel):
         """The site that this Location belongs to, if any, or that its root ancestor belongs to, if any."""
         return self.site or self.ancestors().first().site
 
+    @property
+    def display(self):
+        # prefix with parent name allows location name identity if there are the same name exists under different parents.
+        if self.name:
+            if self.parent:
+                if self.parent.parent:
+                    return f"[{self.parent.parent}->{self.parent}] {self.name}"
+                else:
+                    return f"[{self.parent}] {self.name}"
+            else:
+                return self.name
+        else:
+            return ""
+
     def validate_unique(self, exclude=None):
         # Check for a duplicate name on a Location with no parent.
         # This is necessary because Django does not consider two NULL fields to be equal.


nautobot/dcim/models/locations.py
```diff
@@ -199,6 +199,20 @@ def base_site(self):
...
+        if self.name:
+            if self.parent:
+                if self.parent.parent:
+                    return f"[{self.parent.parent}->{self.parent}] {self.name}"
+                else:
+                    return f"[{self.parent}] {self.name}"
+            else:
+                return self.name
+        else:
+            return ""
```
glennmatthews, line 214: I can expect that there will be cases where 2 levels of parentage is considered insufficient to uniquely identify a location. Maybe we can do something like:

```python
def display(self):
    display_str = self.name
    parent = self.parent
    for i in range(10):  # "10 levels should be more than enough for anybody"
        if parent is None:
            break
        display_str = f"{parent.name} → {display_str}"
        parent = parent.parent
    return display_str
```

glennmatthews:
Impact 1 should be addressable by changing the `TREE_LINK` template in `nautobot/dcim/tables/template_code.py`:

```diff
 TREE_LINK = """
 {% load helpers %}
 {% for i in record.tree_depth|as_range %}
     <i class="mdi mdi-circle-small"></i>
 {% endfor %}
-{{ record|hyperlinked_object }}
+<a href="{{ record.get_absolute_url }}">{{ record.name }}</a>
"""
```

glennmatthews:
We should probably add similar display string logic to the `LocationType` class.

nautobot/dcim/models/locations.py
bryanculver: I'm worried about the recursion impact if we go too far with too long of a list. I would guess Django will have to make an individual SQL call to get the next parent. Considering we are looking to ditch cacheops it may make repeated calls for the same more-root parents.

christhant:
> Impact 1 should be addressable by changing the `TREE_LINK` template in `nautobot/dcim/tables/template_code.py`:
> 
> ```diff
>  TREE_LINK = """
>  {% load helpers %}
>  {% for i in record.tree_depth|as_range %}
>      <i class="mdi mdi-circle-small"></i>
>  {% endfor %}
> -{{ record|hyperlinked_object }}
> +<a href="{{ record.get_absolute_url }}">{{ record.name }}</a>
> """
> ```



> Impact 1 should be addressable by changing the `TREE_LINK` template in `nautobot/dcim/tables/template_code.py`:
> 
> ```diff
>  TREE_LINK = """
>  {% load helpers %}
>  {% for i in record.tree_depth|as_range %}
>      <i class="mdi mdi-circle-small"></i>
>  {% endfor %}
> -{{ record|hyperlinked_object }}
> +<a href="{{ record.get_absolute_url }}">{{ record.name }}</a>
> """
> ```

thank you! I will update the code accordingly.

christhant:
> We should probably add similar display string logic to the `LocationType` class.

should I work on LocationType update in separate story?

nautobot/dcim/models/locations.py
christhant: @glennmatthews the result looks great. didn't see too much performance issue. perhaps. I am on local?
image

nautobot/project-static/js/forms.js
```diff
@@ -226,10 +226,7 @@ function jsify_form(context) {
                         // Annotate hierarchical depth for MPTT objects
                         record.text = '--'.repeat(record._depth) + ' ' + record.text;
                     }
-                    if (record.tree_depth) {
-                        // Annotate hierarchical depth for django-tree-queries objects
-                        record.text = '--'.repeat(record.tree_depth) + ' ' + record.text;
-                    }
```
bryanculver, line 232: @christhant With this removal it'll make the LocationType dropdown selection harder to parse

Screen Shot 2022-09-06 at 10 25 18

Therefor I think we should implement a similar `display` property on LocationType as well.

nautobot/dcim/models/locations.py
glennmatthews: Let's check and make sure that both the drop-down and rendered field are OK display-wise when we have a highly nested hierarchy and long names (perhaps `Campus` -> `Building` -> `Floor` -> `Room` as location-types, with locations `Raleigh-Durham Primary Campus` -> `1234 West Really-Long-Street-Name Avenue` -> `Thirteenth Floor` -> `Room 1301`, just as an example)

christhant:
UI preview of dropdown with long text string
image

christhant:
location hierarchy update at the device view.

image

bryanculver (approved):

nautobot/dcim/tables/template_code.py
```diff
@@ -92,7 +92,7 @@
 {% for i in record.tree_depth|as_range %}
     <i class="mdi mdi-circle-small"></i>
 {% endfor %}
-{{ record|hyperlinked_object }}
+<a href="{{ record.get_absolute_url }}">{{ record.name }}</a>
```
glennmatthews, line 95: Not blocking for this PR but we should think about enhancing `hyperlinked_object` to let you specify the display text, something like `{{ record|hyperlinked_object("name") }}` would be pretty convenient.

glennmatthews (approved):

PR #2366, opened 2022-09-04
Remove profile from mkdocs docker

itdependsnetworks:
# Closes: DNE
# What's Changed

There is a Docker profile assigned to the default Docker build. Additionally, there does not seem to be a way to enable the profile via invoke, which makes it difficult for the common Docker development workflow. 

# TODO
N/A

### DIFF
--- a/development/docker-compose.yml
+++ b/development/docker-compose.yml
@@ -78,8 +78,6 @@ services:
         - "15900:5900"
     shm_size: 2g
   mkdocs:
-    profiles:
-      - mkdocs
     image: "networktocode/nautobot-dev-py${PYTHON_VER}:local"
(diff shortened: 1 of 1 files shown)


bryanculver:
This was done intentionally to not require the mkdocs container consuming resources unnecessarily.

Docs were updated when this was added to specify how to start this container: https://github.com/nautobot/nautobot/commit/7375c032a59c0cefdc2107d3b4680b8f54b4aedc#diff-ce9ed0c205466672ab07f05ace4645a56dd0886770d5464314676c16cad64379R611-R615

itdependsnetworks:
If looking to save resources, does it make sense to open up one to limit when selenium runs? Has always been a pet peeve of mine but figured resource savings was not a primary driver (amongst other reasons)

Issue #873, opened 2021-09-01
Make a relationship required


lampwins:
### Environment
* Python version: 3.9
* Nautobot version: 1.1.2

### Proposed Functionality
As Austin the Network Automation Engineer, I want to specify that a relationship is required so that my users are forced to provide a value for the relationship when creating or modifying data.

### Use Case
In many use cases where I am extending the data model with relationships, I need to enforce the integrity of my data by ensuring that a relationship is actually formed between objects. For instance, I might have a relationship defined between racks and vlans to designate the management vlan for that rack. I want to enforce that a rack _always_ has a management vlan and thus I want to ensure that users are required to always provide a value for this relationship when creating or modifying a rack.

This is straightforward for 1-1 and 1-many relationships and implies the constraint should be directional. Many-many is more interesting and my thought is that marking a many-many relationship as required would force the user to provide at least one value for the relationship. That is debatable I suppose though.

### Database Changes
Relationships will need the ability to mark a relation as required.

### External Dependencies
None

### TODO

- Require (being defined as "Not NULL or empty"):
    - Neither side required
    - Source side required
    - Destination side required

### Out of Scope

- Bi-directional requirement
- Non-Deletion Enforcement

glennmatthews:
Points to clarify:

- How to handle "many" relationship side in general
- How to handle existing data when creating a required relationship (or changing a previously non-required relationship to be required henceforth)

jedelman8:
@lampwins can you provide a little more feedback with what you were thinking?

lampwins:
> How to handle "many" relationship side in general

I invasion being able to mark either side (or both) as required. This helps with the common case in which one side is hidden. This means I could mark the many side as not required. In the case where I do mark the many side as required, I would infer that mean I must specify at least one object for the relationship.

> How to handle existing data when creating a required relationship (or changing a previously non-required relationship to be required henceforth)

This should be handled the same way custom fields are today. The "required" constraint is only enforced on object creation/edit. There is a UI tip on objects that have a null value to say that a required value is missing.

jathanism:
The use-cases and implementation for 1-to-1/many is pretty straightforward.

It would be good to have more concrete use-case examples for required many-to-many relationships before we try to tackle that.

bryanculver:
> The use-cases and implementation for 1-to-1/many is pretty straightforward.
> 
> It would be good to have more concrete use-case examples for required many-to-many relationships before we try to tackle that.

[...]

PR #2223, opened 2022-08-16
Augment `get_route_for_model()` to support REST API routes.

jathanism:
# Closes: #DNE
# What's Changed
- This adds an `api=True` argument to `nautobot.utilities.utils.get_route_for_model()`
- Replaced almost all bespoke implementations of manual route construction for plugins or API routes with calls to `get_route_for_model()`
- Expanded the Development Best Practices guide with a section on "Getting URL Routes"

# TODO

- [ ] Explanation of Change(s)
- [ ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example Plugin Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #2371, opened 2022-09-06
Force a relationship to be required

briddo:
# Closes: #873 
# What's Changed
This is very much a work-in-progress at the moment. 

So far, this code will check if a relationship on an object is required and raise a `forms.ValidationError` if the required side hasn't provided the relationship data. It also adds the same functionality to the `create` and `update` methods in `nautobot.extras.api.relationships.RelationshipModelSerializerMixin`.

It will prevent an object creation or edit if there is a required relationship defined for the object to a model which doesn't have any existing objects:

- In the case of the UI, this presents as form validation errors. 
- In the case of the API, it will add `serializers.ValidationError`s with messages detailing which endpoints to post required objects to.

# TODO
- [ ] API
- [ ] Put HTML5 validation back in to the forms when backend validation has been tested
- [ ] Explanation of Changes
- [ ] Unit, Integration Tests
- [ ] Documentation Updates
- [ ] Outline Remaining Work
- [ ] Clean up the code structure/design


nautobot/extras/forms/mixins.py
```diff
@@ -370,6 +370,12 @@ def _append_relationships(self):
                 field_name = f"cr_{relationship.slug}__{peer_side}"
                 self.fields[field_name] = relationship.to_form_field(side=side)
```
briddo, line 372: This is currently commented out because I wanted to test the server-side validation before providing HTML validation.

nautobot/extras/forms/mixins.py
```diff
@@ -370,6 +370,12 @@ def _append_relationships(self):
...
 
+                # if relationship.required in [
+                #     RelationshipSidesRequiredChoices.BOTH_SIDES_REQUIRED,
+                #     RelationshipSidesRequiredChoices.DESTINATION_SIDE_REQUIRED,
+                # ]:
+                #     self.fields[field_name].required = True
+
```
briddo, line 378: This is currently commented out because I wanted to test the server-side validation before providing HTML validation.

nautobot/core/views/generic.py
```diff
@@ -47,6 +47,34 @@
...
+    required_relationships = Relationship.objects.get_required_for_model(obj)
+    relationships_errors = []
+    for side, relations in required_relationships.items():
+        for relation in relations:
+
+            # Skip referencing itself
+            if ContentType.objects.get_for_model(obj) == getattr(relation, f"{relation.required}_type"):
+                continue
+
+            model_class = getattr(relation, f"{side}_type").model_class()
```
glennmatthews, line 60: Defensive-coding-wise, need to handle the case where `model_class` is None (e.g., relationship to a plugin model for a plugin that's not installed at present).

nautobot/core/views/generic.py
```diff
@@ -47,6 +47,34 @@
...
+
+            if not model_class.objects.count():
+                model_meta = model_class._meta
+                required = model_meta.verbose_name
+                try:
+                    add_url = reverse(f"{model_meta.app_label}:{model_meta.model_name}_add")
```
glennmatthews, line 66: Should be able to use `nautobot.utilities.utils.get_route_for_model` here (possibly dependent on #2223?)

nautobot/core/views/generic.py
```diff
@@ -389,6 +417,16 @@ def get(self, request, *args, **kwargs):
...
         restrict_form_fields(form, request.user)
 
+        relationships_errors = get_relationships_errors(request, obj)
+        if relationships_errors:
+            for msg in relationships_errors:
+                messages.add_message(
+                    request,
+                    messages.ERROR,
+                    mark_safe(msg),
+                )
```
glennmatthews, line 427: Might as well do the `messages.add_message` calls inside of `get_relationships_errors()` so that you don't have to repeat this code three times?

nautobot/extras/choices.py
```diff
@@ -304,6 +304,19 @@ class RelationshipTypeChoices(ChoiceSet):
...
+class RelationshipSidesRequiredChoices(ChoiceSet):
+
+    NEITHER_SIDE_REQUIRED = ""
+    SOURCE_SIDE_REQUIRED = "source"
+    DESTINATION_SIDE_REQUIRED = "destination"
+
+    CHOICES = (
+        (NEITHER_SIDE_REQUIRED, "No"),
+        (SOURCE_SIDE_REQUIRED, "Source side required"),
+        (DESTINATION_SIDE_REQUIRED, "Destination side required"),
```
glennmatthews, line 316: Would something like this be accurate?

```suggestion
        (SOURCE_SIDE_REQUIRED, "Source objects MUST implement this relationship"),
        (DESTINATION_SIDE_REQUIRED, "Destination objects MUST implement this relationship"),
```

Just trying to think of a way to make it clearer what "required" actually means here.

nautobot/extras/forms/mixins.py
```diff
@@ -394,6 +400,33 @@ def clean(self):
...
+                if relation.symmetric:
+                    if not self.cleaned_data.get(f"cr_{relation.slug}__peer", []):
+                        model_name = relation.source_type.model_class()._meta.verbose_name
+                        # Only add the error once for this field
+                        if len(self.errors.get(f"cr_{relation.slug}__peer", [])) == 0:
+                            self.add_error(
+                                f"cr_{relation.slug}__peer", forms.ValidationError(f"You must select a {model_name}")
+                            )
+                    continue
```
glennmatthews, line 416: I didn't think we were going to implement/permit symmetric required relations?

nautobot/extras/forms/mixins.py
```diff
@@ -394,6 +400,33 @@ def clean(self):
...
+                # Skip self-referencing
+                if ContentType.objects.get_for_model(self.instance) == getattr(relation, f"{relation.required}_type"):
+                    continue
+
+                model_name = getattr(relation, f"{side}_type").model_class()._meta.verbose_name
+                cr_field_name = f"cr_{relation.slug}__{relation.required}"
+
+                if not self.cleaned_data[cr_field_name]:
+                    self.add_error(
+                        f"cr_{relation.slug}__{side}", forms.ValidationError(f"You must select a {model_name}")
```
glennmatthews, line 427: Need to think about how to make this message accurate for both `*-to-1` relations and `*-to-many` relations.

nautobot/extras/forms/mixins.py
briddo: Ah yes https://github.com/nautobot/nautobot/issues/873#issuecomment-1217088337

nautobot/core/views/generic.py
briddo: Do we know when #2223 is likely to be merged in?

nautobot/extras/forms/mixins.py
```diff
@@ -370,6 +369,12 @@ def _append_relationships(self):
                 field_name = f"cr_{relationship.slug}__{peer_side}"
                 self.fields[field_name] = relationship.to_form_field(side=side)
 
+                # if relationship.required in [
+                #     RelationshipSidesRequiredChoices.BOTH_SIDES_REQUIRED,
+                #     RelationshipSidesRequiredChoices.DESTINATION_SIDE_REQUIRED,
+                # ]:
+                #     self.fields[field_name].required = True
+
```
bryanculver, line 377: Intended?

briddo: Yeah it's commented out as intending to put the HTML5 validation back in when backend is sorted..

changes/873.added
```diff
@@ -0,0 +1 @@
+Relationships were given the ability to be required.
```
glennmatthews, line 1: Perhaps something like:

```suggestion
Made it possible to require Relationships to be included when editing the related models.
```

nautobot/core/views/generic.py
```diff
@@ -384,6 +385,10 @@ def get(self, request, *args, **kwargs):
         form = self.model_form(instance=obj, initial=initial_data)
         restrict_form_fields(form, request.user)
 
+        relationships_errors = get_relationships_errors(request, obj)
+        if len(relationships_errors) > 0:
+            return redirect(get_route_for_model(obj, "list"))
```
glennmatthews, line 390: Seems wrong to me - why should we ever get errors on a GET (i.e., loading the "edit" page initially)? And wouldn't this prevent us from ever editing the object as a result?

nautobot/core/views/generic.py
```diff
@@ -403,6 +408,10 @@ def post(self, request, *args, **kwargs):
         form = self.model_form(data=request.POST, files=request.FILES, instance=obj)
         restrict_form_fields(form, request.user)
 
+        relationships_errors = get_relationships_errors(request, obj)
+        if len(relationships_errors) > 0:
+            return redirect(get_route_for_model(obj, "list"))
```
glennmatthews, line 413: This also seems undesirable to me - if there are missing required relationships, we should remain on the "edit" page and tell the user exactly what's missing, just as in the case of any other form validation error, not send them back to the list view.

Can we instead make it so that `form.is_valid()` below includes the validation of relationships and reports them as form errors if any?

nautobot/core/views/generic.py
```diff
@@ -948,6 +957,10 @@ def post(self, request, **kwargs):
         logger = logging.getLogger("nautobot.views.BulkEditView")
         model = self.queryset.model
 
+        relationships_errors = get_relationships_errors(request, model)
+        if len(relationships_errors) > 0:
+            return redirect(get_route_for_model(model, "list"))
```
glennmatthews, line 962: Same comment here about staying on the bulk-edit view instead of sending them back to the list view, reporting errors on the form if applicable.

nautobot/core/views/generic.py
```diff
@@ -948,6 +957,10 @@ def post(self, request, **kwargs):
         logger = logging.getLogger("nautobot.views.BulkEditView")
         model = self.queryset.model
 
+        relationships_errors = get_relationships_errors(request, model)
```
glennmatthews, line 960: Is it correct that here we're passing the model class as the second parameter, while in the single-edit case we're passing `obj` (model instance) instead? Does `get_relationships_errors` handle both cases?

nautobot/extras/api/relationships.py
```diff
@@ -259,6 +256,49 @@ def to_internal_value(self, data):
         return {self.field_name: output_data}
 
 
+def api_relationships_errors(instance, relationships_data):
```
glennmatthews, line 259: Docstring please? It's not obvious to me without a side-by-side comparison what the difference is between `api_relationships_errors` and `get_relationships_errors`.

nautobot/extras/api/relationships.py
```diff
@@ -259,6 +256,49 @@ def to_internal_value(self, data):
...
+
+            if relation.skip_required(instance, side):
+                continue
+
+            model_name = getattr(relation, f"{side}_type").model_class()._meta.verbose_name
+
+            try:
+                submitted_data = relationships_data.get(relation, "").get(side, "")
+            except AttributeError:
+                submitted_data = []
```
glennmatthews, line 273: Perhaps?

```suggestion
            submitted_data = relationships_data.get(relation, {}).get(side, "")
```

nautobot/extras/api/relationships.py
```diff
@@ -259,6 +256,49 @@ def to_internal_value(self, data):
...
+
+            if len(submitted_data) == 0:
+                if relation.type in [
+                    RelationshipTypeChoices.TYPE_ONE_TO_MANY,
+                    RelationshipTypeChoices.TYPE_MANY_TO_MANY,
+                ]:
```
glennmatthews, line 279: Might be better to check `relation.has_many(side)` or `relation.has_many(RelationshipSideChoices.OPPOSITE[side])` as appropriate?

nautobot/extras/api/relationships.py
```diff
@@ -259,6 +256,49 @@ def to_internal_value(self, data):
...
+                    num_required_verbose = "at least one"
+                else:
+                    num_required_verbose = "a"
+                relationships_errors.append(
+                    {
+                        relation.slug: f"You must specify {num_required_verbose} {model_name} id or slug "
+                        f"in the request body's relationships[“{relation.slug}”][”{side}”][”objects”] list data"
```
glennmatthews, line 286:
```suggestion
                        f'in the request body's relationships["{relation.slug}"]["{side}"]["objects"] list data'
```

nautobot/extras/api/relationships.py
```diff
@@ -267,13 +307,17 @@ class RelationshipModelSerializerMixin(ValidatedModelSerializer):
     def create(self, validated_data):
         relationships = validated_data.pop("relationships", {})
         instance = super().create(validated_data)
+        validate_relationships(self.context["request"], instance, relationships)
```
glennmatthews, line 310: In the case of an error here, does the instance still get created since you're calling `super().create(validated_data)` before the `validate_relationships()` check? Or does raising an exception cause it to not get created after all?

nautobot/extras/choices.py
```diff
@@ -304,6 +304,19 @@ class RelationshipTypeChoices(ChoiceSet):
     )
 
 
+class RelationshipSidesRequiredChoices(ChoiceSet):
```
glennmatthews, line 307:
```suggestion
class RelationshipRequiredSideChoices(ChoiceSet):
```

nautobot/extras/forms/mixins.py
```diff
@@ -394,6 +401,37 @@ def clean(self):
...
+                # Not enforcing required symmetric relationships
+                if relation.symmetric:
+                    continue
+
+                # Skip self-referencing
+                if ContentType.objects.get_for_model(self.instance) == getattr(
+                    relation, f"{relation.required_side}_type"
+                ):
+                    continue
```
glennmatthews, line 417: should we be using `skip_required()` here instead?

nautobot/extras/forms/mixins.py
```diff
@@ -394,6 +401,37 @@ def clean(self):
...
+
+                model_name = getattr(relation, f"{side}_type").model_class()._meta.verbose_name
+                cr_field_name = f"cr_{relation.slug}__{relation.required_side}"
+
+                if not self.cleaned_data[cr_field_name]:
+                    if relation.type in [
+                        RelationshipTypeChoices.TYPE_ONE_TO_MANY,
+                        RelationshipTypeChoices.TYPE_MANY_TO_MANY,
+                    ]:
```
glennmatthews, line 426: Another case where I  think checking `relation.has_many()` would be more correct.

nautobot/extras/models/relationships.py
```diff
@@ -233,6 +235,17 @@ def get_for_model(self, model):
...
 
+    def get_required_for_model(self, model):
+        """
+        Return all required Relationships assigned to the given model.
+        """
+        required_types = [req_type for req_type in RelationshipSidesRequiredChoices.values() if req_type]
+        source, destination = self.get_for_model(model)
+        return {
+            "source": destination.filter(required_side__in=required_types),
+            "destination": source.filter(required_side__in=required_types),
```
glennmatthews, line 246: This seems incorrect/confusing to me:

1) it inverts the meaning of `source` and `destination` as compared to `get_for_model`?
2) why `required_side__in=required_types` in both cases, not just `required_side="source"` on one side and `required_side="destination"` on the other?

nautobot/extras/models/relationships.py
```diff
@@ -249,7 +262,14 @@ class Relationship(BaseModel, ChangeLoggedModel, NotesMixin):
         default=RelationshipTypeChoices.TYPE_MANY_TO_MANY,
         help_text="Cardinality of this relationship",
     )
-
+    required_side = models.CharField(
+        max_length=12,
+        choices=RelationshipSidesRequiredChoices,
+        default=RelationshipSidesRequiredChoices.NEITHER_SIDE_REQUIRED,
+        help_text="Force this relationship to be required. This does not effect symmetrical relationships.",
```
glennmatthews, line 269:
```suggestion
        help_text="Objects on the specified side MUST implement this relationship. Not permitted for symmetric relationships.",
```

nautobot/extras/models/relationships.py
```diff
@@ -736,3 +774,65 @@ def _validate_relationship_filter_restriction(self):
...
+            required_model_class = getattr(relation, f"{side}_type").model_class()
+
+            if required_model_class.objects.count() == 0:
+                model_meta = required_model_class._meta
+                required = model_meta.verbose_name
+
+                if relation.type in [
+                    RelationshipTypeChoices.TYPE_ONE_TO_MANY,
+                    RelationshipTypeChoices.TYPE_MANY_TO_MANY,
+                ]:
```
glennmatthews, line 797: relation.has_many()

nautobot/extras/models/relationships.py
```diff
@@ -736,3 +774,65 @@ def _validate_relationship_filter_restriction(self):
...
+    relationships_errors = []
+    for side, relations in required_relationships.items():
+        for relation in relations:
+
+            if relation.skip_required(obj, side):
+                continue
+
+            required_model_class = getattr(relation, f"{side}_type").model_class()
+
+            if required_model_class.objects.count() == 0:
```
glennmatthews, line 790:
```suggestion
            if not required_model_class.objects.exists()
```

nautobot/extras/models/relationships.py
```diff
@@ -736,3 +774,65 @@ def _validate_relationship_filter_restriction(self):
...
+                        }
+                    )
+
+    if len(relationships_errors) > 0 and output_for == "ui":
+        for msg in relationships_errors:
+            messages.add_message(
+                request,
+                messages.ERROR,
+                mark_safe(msg),
+            )
```
glennmatthews, line 836: Why not do this as part of the other `if output_for == "ui"` block up above?

nautobot/extras/models/relationships.py
```diff
@@ -736,3 +774,65 @@ def _validate_relationship_filter_restriction(self):
                 raise ValidationError(
                     {side_name: (f"{side} violates {self.relationship} {side_name}_filter restriction")}
                 )
+
+
+def get_relationships_errors(request, obj, output_for="ui"):
```
glennmatthews, line 779: So this doesn't actually check for relationship errors - it only checks for the case where none of the required objects exist and tells the user to create them first? I wonder if something like `check_for_missing_required_related_objects` might be a clearer name.

I also wonder if it'd work any better for this to be a class method or instance method on the `RelationshipModel` class instead of a standalone function?

nautobot/extras/models/relationships.py
```diff
@@ -736,3 +774,65 @@ def _validate_relationship_filter_restriction(self):
...
+                    num_required_verbose = "at least one"
+                else:
+                    num_required_verbose = "a"
+
+                if output_for == "ui":
+                    try:
+                        add_url = reverse(get_route_for_model(required_model_class, "add"))
+                        add_message = f"<a href='{add_url}'>Click here</a> to create a {model_meta.verbose_name}."
+                    except NoReverseMatch:
+                        add_message = f"You need to create a {required} first"
```
glennmatthews, line 807:
```suggestion
                        add_message = f"You need to create {num_required_verbose} {required} first"
```

nautobot/extras/models/relationships.py
```diff
@@ -736,3 +774,65 @@ def _validate_relationship_filter_restriction(self):
...
+                    RelationshipTypeChoices.TYPE_MANY_TO_MANY,
+                ]:
+                    num_required_verbose = "at least one"
+                else:
+                    num_required_verbose = "a"
+
+                if output_for == "ui":
+                    try:
+                        add_url = reverse(get_route_for_model(required_model_class, "add"))
+                        add_message = f"<a href='{add_url}'>Click here</a> to create a {model_meta.verbose_name}."
```
glennmatthews, line 805: Does this link work in the UI? I would have thought we'd need a `format_html()` or `mark_safe()` somewhere along the line.

nautobot/extras/tests/test_api.py
```diff
@@ -2466,6 +2472,30 @@ def setUpTestData(cls):
...
                 destination_type=device_type,
             ),
+            Relationship(
+                name="VLANs require at least one device (destination objects)",
+                slug="vlans-devices",
+                type="many-to-many",
+                source_type=vlan_type,
+                destination_type=device_type,
+                required_side="destination",
+            ),
```
glennmatthews, line 2482: I find this feature really confusing, not sure if I'm misunderstanding it, or if this is wrong logic - the description of `DESTINATION_SIDE_REQUIRED` is `Destination objects MUST implement this relationship`, which I would interpret in this case as meaning that each Device in this relationship must specify at least one VLAN - but that's the opposite of what you're naming this relationship, above. Is the name wrong, is the help text wrong, or am I misunderstanding this feature?

nautobot/extras/tests/test_api.py
```diff
@@ -2466,6 +2472,30 @@ def setUpTestData(cls):
...
+            Relationship(
+                name="Platform requires at least one device (destination objects)",
```
glennmatthews, line 2484: Same question here.

nautobot/extras/tests/test_api.py
```diff
@@ -2466,6 +2472,30 @@ def setUpTestData(cls):
...
+                slug="platform-devices",
+                type="one-to-many",
+                source_type=platform_type,
+                destination_type=device_type,
+                required_side="destination",
+            ),
+            Relationship(
+                name="Tenant requires one platform (one source object)",
```
glennmatthews, line 2492: And here.

nautobot/extras/models/relationships.py
```diff
@@ -736,3 +774,65 @@ def _validate_relationship_filter_restriction(self):
...
+                    except NoReverseMatch:
+                        add_message = f"You need to create a {required} first"
+
+                    relationships_errors.append(
+                        f"{obj._meta.verbose_name_plural.capitalize()} require {num_required_verbose} {required}, "
```
glennmatthews, line 810: `capitalize()` will change `VLANs` to `Vlans`, `IP addresses` to `Ip addresses`, etc., which is undesirable. Can we reword this to avoid the need to call `capitalize()`?

nautobot/extras/tests/test_api.py
```diff
@@ -2631,6 +2661,192 @@ def test_populate_relationship_associations_on_site_create(self):
...
+        # Try creating a vlan when no device exists
+        response = self.client.post(
+            reverse(get_route_for_model(VLAN, "list", api=True)),
+            data=create_data,
+            format="json",
+            **self.header,
+        )
+        expected_error_json = [
+            {
+                "cr_vlans-devices__destination": "Vlans require at least one device, but no devices exist yet. "
```
glennmatthews, line 2689: Why `cr_vlans-devices__destination` here but `vlans-devices` in the next expected error? Seems inconsistent to me.

nautobot/extras/tests/test_api.py
```diff
@@ -2631,6 +2661,192 @@ def test_populate_relationship_associations_on_site_create(self):
...
+                }
+            }
+        }
+        response = self.client.post(
+            reverse(get_route_for_model(VLAN, "list", api=True)),
+            data=create_data,
+            format="json",
+            **self.header,
+        )
+        self.assertHttpStatus(response, 201)
```
glennmatthews, line 2731: Perhaps also check that the relationship associations are created as expected?

nautobot/extras/tests/test_api.py
```diff
@@ -2631,6 +2661,192 @@ def test_populate_relationship_associations_on_site_create(self):
...
+        # Try creating a tenant when all required data is present
+        create_data["relationships"] = {"platform-tenant": {"source": {"objects": [{"pk": str(platform.pk)}]}}}
+        response = self.client.post(
+            reverse(get_route_for_model(Tenant, "list", api=True)),
+            data=create_data,
+            format="json",
+            **self.header,
+        )
+        self.assertHttpStatus(response, 201)
+
```
glennmatthews, line 2850: Also need some tests for the `update` case, including IMHO at least the following:

- object is created without the relationship, then the relationship is marked as required, object is updated without the relationship (fails)
- object is created without the relationship, then the relationship is marked as required, object is updated to include the relationship (succeeds)
- object is created with the required relationship, object is updated without specifying any relationship data (succeeds, relationship remains in place)
- object is created with the required relationship, object is updated to remove the relationship data (fails)

nautobot/core/views/generic.py
briddo: My thinking was that, as the requirement is for editing or creating objects, the user should not be allowed be see the edit page without required relationship objects being present in the database. Is there a better/more UX experience-friendly way?

nautobot/core/views/generic.py
briddo: Yeah, seems to!

nautobot/extras/forms/mixins.py
briddo: Well spotted!

nautobot/extras/models/relationships.py
briddo: Line 830:
```
    if len(relationships_errors) > 0 and output_for == "ui":
        for msg in relationships_errors:
            messages.add_message(
                request,
                messages.ERROR,
                mark_safe(msg),
            )
```

nautobot/extras/tests/test_api.py
briddo: I think I confused myself by inverting things as you commented here: https://github.com/nautobot/nautobot/pull/2371#discussion_r981244943

glennmatthews:
Closing in favor of #2612.

Issue #2023, opened 2022-07-08
App ecosystem CI testing against core


lampwins:
As Patti, the Plugin Developer, I want an automated mechanism whereby my plugin's tests can be run against new builds of Nautobot core when they are made available so that I have continuous feedback on integration issues with the core as they arise.

We need to design and publish a methodology for an ecosystem-wide CI pipeline.

Should builds be kicked off:
- For each commit in core?
- For each major branch/merge, build in core (develop/next/main/etc)?
- For each release candidate in core?
- Something else?

Should plugins be able to register build from other dependencies like the SSoT base?

### Outcome

- Proposed design for triggering plugin CI as well as best-practices for plugins to test against desired Nautobot versions
- Proposal for frequency of CI triggering
- Should we consider matrix options?

PR #2377, opened 2022-09-06
Add upstream monitoring reusable workflow

bryanculver:
# Closes: #2023
# What's Changed
### Overview
The high-level goal for this project is to design and publish a methodology for an ecosystem-wide CI pipeline. The desired outcome is:

- Proposed design for triggering plugin CI as well as best-practices for plugins to test against desired Nautobot versions
- Proposal for frequency of CI triggering
- Should we consider matrix options?
- As well as providing a minimal amount of churn or maintenance to plugins as new versions of Nautobot get published.

### Proposal

With the considerations above, our proposal is to have plugins routinely run CI on a schedule to validate their compatibility with different versions of Nautobot. To make this easily manageable for plugins a setup as follows in being recommended:

Nautobot Core should continue to provide helpful Docker image labels:
- `next`: This image will be the most recent, passing build from the `next` branch on GitHub. This will be pre-published, pre-release builds of the next major or minor release of Nautobot.
- `develop`: This image will be the most recent, passing build from the `develop` branch on GitHub. This will be the pre-published patch release builds of the current minor release of Nautobot.

Nautobot Core should provide a Github Action reusable workflow that can be included with minimal involvement from plugin maintainers. This workflow would:
- Provide a core-maintained list of testing matrix of Nautobot versions so as to produce consistent results across all adopting plugins and eliminating churn plugin-side.
- Execute in the context of a plugin's existing workflows. Scalable for any number of plugins, public or private, to benefit from the same Github Action workflow steps
- Be updatable in the future to report statistics for plugin compatibility
- Require a minimal amount of expectations from the running workflow for file structure and test run syntax
- Allow plugins flexibility in how frequent their upstream monitoring run (private could be pricey to run nightly and desire to run weekly)
- Be able to be incorporated into cookie cutters without any post-baking configuring

The testing strategy would be as follows:
- Take a "vanilla" Nautobot Docker image with default dev dependencies
- Install plugin from current repo (permits pre-release versions of plugins)
- Install any additional dev dependencies a plugin may need for their tests to complete
- Run `poetry run invoke unittest`

### Prototype

A prototype exists here: https://github.com/nautobot/nautobot-plugin-netutils/pull/2. This plugin was baked very recently from a cookie so adheres closest to our best practices.

**`.github/workflows/upstream_testing.yml`** (in plugin specific repo)

An example of what plugins would need to add to their repository to pull a shared reusable workflow on a schedule and what minimal information would need to be provided to the workflow to execute.

**`.github/workflows/plugin_upstream_testing_base.yml`** (here)

This is the reusable workflow all plugins would share and would ideally be tagged and stored in either the Nautobot core repo (`nautobot/nautobot`) or a bespoke tools repo.

**`development/Dockerfile`** (in plugin specific repo)

This was modified to get around Poetry side-effects in a more stable way:
- Ignores the Poetry lockfile which can cause forced upgrades/downgrades in the container
- Still uses Poetry to install plugin-specific dev dependencies that may be required for testing but is not available in the base image
- Allows the Nautobot version in the container to be "frozen" and does not require a PyPi tagged version of Nautobot to be reinstalled (pre-release version of Nautobot can be tested)
- Still can allow for custom apt/build steps be added as necessary for plugins with more advanced setups

#### Assumptions
- A plugin provides a Dockerfile in `development/Dockerfile` which builds an image for the purpose of testing that honors the requested Nautobot version, sometimes one that is not available in PyPI
- A plugin uses `invoke` to execute commands and has been reasonably set up to orchestrate a Docker compose or other setup to run the above image (unless testing locally)
- `invoke unittest` is executable and either runs locally without the need of Docker containers or spins up the necessary services to execute the test

### Caveats
- Tests default branch of plugin which may be pre-release and might not reflect most recent released version of plugin on a regular basis
- Not intended for multiple release train plugins


jathanism:
I feel like the name of `upstream` isn't quite descriptive enough. Would something like `upstream-testing` be a tad better?

.github/workflows/plugin_upstream_base.yml
```diff
@@ -0,0 +1,49 @@
...
+
+env:
+  PYTHON_VERSION: "3.7"
+
+jobs:
+  unittest:
+    strategy:
+      fail-fast: false
+      matrix:
+        nautobot-version: ["next", "develop", "stable", "1.4", "1.3"]
```
jathanism, line 22: This is my only concern here which mirrors that of @glennmatthews in the other example repo (https://github.com/nautobot/nautobot-plugin-netutils/pull/2): I don't see that much value in constantly testing against `stable`, which won't change often, and old versions like `1.3` that won't be changing at all.

I think the ability to also be able to run this workflow from a plugin's repo manually against a given target version of Nautobot could be very appealing, however.

If we were to automate the runs, I think weekly or even bi-weekly would be more than sufficient. And I also think that running against a matrix of targeted versions on releases might also be a good idea?

bryanculver:
Please review https://github.com/nautobot/nautobot-plugin-golden-config/pull/344 for working example (still draft pending this making it to `main`).

See also: https://github.com/nautobot/nautobot-plugin-golden-config/actions/runs/3114981878

glennmatthews:
In https://github.com/nautobot/nautobot-plugin-golden-config/actions/runs/3114981878/jobs/5051417897, why is it that we build the container in the `build` step, then re-build it again in the `run tests` step? I noticed that `build` is creating `docker.io/nautobot-plugin-golden-config/nautobot:develop-py3.7` (note `-plugin-`) while `run tests` is creating and running with `nautobot-golden-config/nautobot:develop-py3.7` (no `-plugin-`).

.github/workflows/plugin_upstream_testing_base.yml
```diff
@@ -0,0 +1,49 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        nautobot-version: ["next", "develop"]
+    runs-on: "ubuntu-20.04"
+    steps:
+      - name: "Check out repository code"
+        uses: "actions/checkout@v2"
+      - name: "Setup environment"
+        uses: "networktocode/gh-action-setup-poetry-environment@v2"
```
glennmatthews, line 28: Possibly `@v3` now? #2467

glennmatthews (commented):
Should we add something to the plugin development docs about how to set up a project to make use of this?

bryanculver:
> In [nautobot/nautobot-plugin-golden-config/actions/runs/3114981878/jobs/5051417897](https://github.com/nautobot/nautobot-plugin-golden-config/actions/runs/3114981878/jobs/5051417897), why is it that we build the container in the `build` step, then re-build it again in the `run tests` step? I noticed that `build` is creating `docker.io/nautobot-plugin-golden-config/nautobot:develop-py3.7` (note `-plugin-`) while `run tests` is creating and running with `nautobot-golden-config/nautobot:develop-py3.7` (no `-plugin-`).

See: https://github.com/nautobot/nautobot-plugin-golden-config/pull/344#discussion_r979056951

HanlinMiao (approved):
LGTM!

gsnider2195 (approved):

Issue #2201, opened 2022-08-12
Jobs with `has_sensitive_variables=True` can be scheduled and arguments stored in the database


gsnider2195:
### Environment
* Python version: 3.7.13
* Nautobot version: 1.3.10

The API and UI are relying on `approval_required=True` and `has_sensitive_variables=True` to be mutually exclusive in order to prevent these jobs from being scheduled in the future and potentially sensitive variables being stored in the database. However migration `nautobot/extras/migrations/0036_scheduledjob_crontab.py` to add the `has_sensitive_variables` field sets the default value to true, bypassing model validation and allowing jobs to have both of these variables set to true.

Job run validation in the UI and API is only failed if `has_sensitive_variables=True` AND the job is not scheduled for immediate execution. However a job can be set for immediate execution and have `approval_required=True`, requiring the job to be scheduled for later execution while awaiting approval.

### Steps to Reproduce
1. Start with nautobot v1.3.9
2. Create a job with `approval_required=True` and `enabled=True`
3. Upgrade to nautobot v1.3.10
4. Try to run the job immediately

### Expected Behavior
The job will not be scheduled because `has_sensitive_variables=True`

### Observed Behavior
The job is scheduled because `approval_required=True`

PR #2380, opened 2022-09-07
Prevent jobs from running if has sensitive variables and requires approval

timizuoebideri1:
# Closes: #2201 
# What's Changed

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/tables.py
+++ b/nautobot/extras/tables.py
@@ -71,7 +71,7 @@ GITREPOSITORY_BUTTONS = """
 """
 
 JOB_BUTTONS = """
-<a href="{% url 'extras:job_run' slug=record.slug %}" class="btn btn-primary btn-xs" title="Run/Schedule" {% if not perms.extras.run_job or not record.enabled or not record.installed %}disabled="disabled"{% endif %}><i class="mdi mdi-play" aria-hidden="true"></i></a>
+<a href="{% url 'extras:job_run' slug=record.slug %}" class="btn btn-primary btn-xs" title="Run/Schedule" {% if not perms.extras.run_job or not record.enabled or not record.installed or record.has_sensitive_variables and record.approval_required %}disabled="disabled"{% endif %}><i class="mdi mdi-play" aria-hidden="true"></i></a>
 """
 
 OBJECTCHANGE_OBJECT = """
--- a/nautobot/extras/tables.py
+++ b/nautobot/extras/tables.py
@@ -71,7 +71,7 @@ GITREPOSITORY_BUTTONS = """
 """
 
 JOB_BUTTONS = """
-<a href="{% url 'extras:job_run' slug=record.slug %}" class="btn btn-primary btn-xs" title="Run/Schedule" {% if not perms.extras.run_job or not record.enabled or not record.installed or record.has_sensitive_variables and record.approval_required %}disabled="disabled"{% endif %}><i class="mdi mdi-play" aria-hidden="true"></i></a>
+<a href="{% url 'extras:job_run' slug=record.slug %}" class="btn btn-primary btn-xs" title="Run/Schedule" {% if not perms.extras.run_job or not record.runnable %}disabled="disabled"{% endif %}><i class="mdi mdi-play" aria-hidden="true"></i></a>
 """
 
 OBJECTCHANGE_OBJECT = """
--- a/nautobot/extras/templates/extras/job.html
+++ b/nautobot/extras/templates/extras/job.html
@@ -38,7 +38,8 @@
                     {% if job_model.has_sensitive_variables or job_model.approval_required %}
                         <div class="alert alert-danger">
                             <i class="mdi mdi-alert"></i>
-                            This job may have sensitive variables and cannot be marked as requiring approval.
+                            This job is flagged as possibly having sensitive variables but is also flagged as requiring approval.
+                            One of these two flags must be removed before this job can be run.
                         </div>
                     {% endif %}
                     {% if not job_model.installed or job_model.job_class is None %}
@@ -103,7 +104,7 @@
                         </div>
                         <div class="pull-right">
                             <button type="submit" name="_run" id="id__run" class="btn btn-primary"
-                                {% if not perms.extras.run_job or not job_model.runnable or job_model.has_sensitive_variables or job_model.approval_required %}
+                                {% if not perms.extras.run_job or not job_model.runnable %}
                                     disabled="disabled"
                                 {% endif %}
                             >
--- a/nautobot/extras/templates/extras/job.html
+++ b/nautobot/extras/templates/extras/job.html
@@ -35,6 +35,12 @@
                             You do not have permission to run jobs.
                         </div>
                     {% endif %}
+                    {% if job_model.has_sensitive_variables or job_model.approval_required %}
+                        <div class="alert alert-danger">
+                            <i class="mdi mdi-alert"></i>
+                            This job may have sensitive variables and cannot be marked as requiring approval.
+                        </div>
+                    {% endif %}
                     {% if not job_model.installed or job_model.job_class is None %}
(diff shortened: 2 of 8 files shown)


nautobot/extras/tests/test_views.py
```diff
@@ -1800,6 +1801,38 @@ def test_run_job_with_sensitive_variables_for_future(self, _):
...
+        }
+        for run_url in self.run_urls:
+            # Assert warning message shows in get
+            response = self.client.get(run_url)
+            content = extract_page_body(response.content.decode(response.charset))
+            self.assertIn("This job may have sensitive variables and cannot be marked as requiring approval.", content)
+
+            # Assert run button is disabled
+            search = re.search(r'<button.+ id="id__run" class="btn btn-primary"\s+disabled="disabled"', content)
+            self.assertTrue(search is not None)
```
glennmatthews, line 1824:
```suggestion
            self.assertInHtml('<button id="id__run" class="btn btn-primary" disabled="disabled">', content)
```

nautobot/extras/templates/extras/job.html
```diff
@@ -35,6 +35,12 @@ <h1>{{ job_model }}</h1>
                             You do not have permission to run jobs.
                         </div>
                     {% endif %}
+                    {% if job_model.has_sensitive_variables or job_model.approval_required %}
+                        <div class="alert alert-danger">
+                            <i class="mdi mdi-alert"></i>
+                            This job may have sensitive variables and cannot be marked as requiring approval.
```
glennmatthews, line 41:
```suggestion
                            This job is flagged as possibly having sensitive variables but is also flagged as requiring approval.
                            One of these two flags must be removed before this job can be run.
```

nautobot/extras/templates/extras/job.html
```diff
@@ -97,7 +103,7 @@ <h1>{{ job_model }}</h1>
                         </div>
                         <div class="pull-right">
                             <button type="submit" name="_run" id="id__run" class="btn btn-primary"
-                                {% if not perms.extras.run_job or not job_model.runnable %}
+                                {% if not perms.extras.run_job or not job_model.runnable or job_model.has_sensitive_variables or job_model.approval_required %}
```
glennmatthews, line 106: This seems wrong - this will prevent any `has_sensitive_variables` *or* `approval_required` job from being run?

glennmatthews: Can we perhaps add the `has_sensitive_variables and approval_required` check into the `job_model.runnable` logic directly?

nautobot/extras/views.py
```diff
@@ -1086,6 +1086,11 @@ def post(self, request, class_path=None, slug=None):
             messages.error(request, "Unable to run or schedule job: Job is not enabled to be run.")
         elif job_model.has_sensitive_variables and request.POST["_schedule_type"] != JobExecutionType.TYPE_IMMEDIATELY:
             messages.error(request, "Unable to schedule job: Job may have sensitive input variables.")
+        elif job_model.has_sensitive_variables and job_model.approval_required:
+            messages.error(
+                request,
+                "Unable to run or schedule job: A job that may have sensitive variables cannot be marked as requiring approval.",
```
glennmatthews, line 1092:
```suggestion
                "Unable to run or schedule job: "
                "This job is flagged as possibly having sensitive variables but is also flagged as requiring approval."
                "One of these two flags must be removed before this job can be scheduled or run.",
```

nautobot/extras/views.py
```diff
@@ -1086,6 +1086,11 @@ def post(self, request, class_path=None, slug=None):
             messages.error(request, "Unable to run or schedule job: Job is not enabled to be run.")
         elif job_model.has_sensitive_variables and request.POST["_schedule_type"] != JobExecutionType.TYPE_IMMEDIATELY:
             messages.error(request, "Unable to schedule job: Job may have sensitive input variables.")
+        elif job_model.has_sensitive_variables and job_model.approval_required:
```
glennmatthews, line 1089: Do we need to add similar logic in the REST API view?

glennmatthews (changes requested):

nautobot/extras/tests/test_views.py
timizuoebideri1: Would not work because of how the Html element is formatted.
```jinja
<button type="submit" name="_run" id="id__run" class="btn btn-primary"
      {% if not perms.extras.run_job or not job_model.runnable %}
          disabled="disabled"
      {% endif %}
  >
```
Looks more like this
```html
<button type="submit" name="_run" id="id__run" class="btn btn-primary"
          disabled="disabled"
  >
```

nautobot/extras/tests/test_views.py
```diff
@@ -1800,6 +1801,45 @@ def test_run_job_with_sensitive_variables_for_future(self, _):
...
+            response = self.client.get(run_url)
+            content = extract_page_body(response.content.decode(response.charset))
+            self.assertIn(
+                "This job is flagged as possibly having sensitive variables but is also flagged as requiring approval.",
+                content,
+            )
+
+            # Assert run button is disabled
+            search = re.search(
+                r'<button type="submit" name="_run" id="id__run" class="btn btn-primary"\s+disabled="disabled"', content
```
bryanculver, line 1827:
```suggestion
                r'<button[^>]+id="id__run"[^>]+disabled="disabled"', content
```
Should be a but more durable for any CSS changes.

nautobot/extras/tests/test_views.py
gsnider2195: AssertInHTML is supposed to ignore whitespace. Did it not work for this?

nautobot/extras/tables.py
```diff
@@ -71,7 +71,7 @@
 """
 
 JOB_BUTTONS = """
-<a href="{% url 'extras:job_run' slug=record.slug %}" class="btn btn-primary btn-xs" title="Run/Schedule" {% if not perms.extras.run_job or not record.enabled or not record.installed %}disabled="disabled"{% endif %}><i class="mdi mdi-play" aria-hidden="true"></i></a>
+<a href="{% url 'extras:job_run' slug=record.slug %}" class="btn btn-primary btn-xs" title="Run/Schedule" {% if not perms.extras.run_job or not record.enabled or not record.installed or record.has_sensitive_variables and record.approval_required %}disabled="disabled"{% endif %}><i class="mdi mdi-play" aria-hidden="true"></i></a>
```
glennmatthews, line 74:
```suggestion
<a href="{% url 'extras:job_run' slug=record.slug %}" class="btn btn-primary btn-xs" title="Run/Schedule" {% if not perms.extras.run_job or not record.runnable %}disabled="disabled"{% endif %}><i class="mdi mdi-play" aria-hidden="true"></i></a>
```

nautobot/extras/tests/test_views.py
glennmatthews: Might need to explicitly include the `name="_run"` in the `assertInHtml` fragment?

nautobot/extras/tests/test_views.py
```diff
@@ -1800,6 +1801,43 @@ def test_run_job_with_sensitive_variables_for_future(self, _):
...
+            response = self.client.get(run_url)
+            content = extract_page_body(response.content.decode(response.charset))
+            self.assertIn(
+                "This job is flagged as possibly having sensitive variables but is also flagged as requiring approval.",
+                content,
+            )
+
+            # Assert run button is disabled
+            search = re.search(r'<button[^>]+id="id__run"[^>]+disabled="disabled"', content)
+            self.assertTrue(search)
```
glennmatthews, line 1827: I'd still like to have this use `assertInHTML` rather than `assertTrue` on a `re.search` if at all possible.

glennmatthews (approved):
Please add changelog fragment before merging; otherwise this looks good to me!

PR #2381, opened 2022-09-07
Added MultiValueCharField

HanlinMiao:
Added MultiValueCharField to accomodate the usage of MultiValueCharFilter that is not backed by choices.
# DEPENDENT ON
https://github.com/nautobot/nautobot/pull/2384
# TODO

- [x] Unit Tests
- [ ] Documentation
- [x] Make an `MultiValueCharField` for each `MultiValueCharFilter`


nautobot/utilities/forms/fields.py
```diff
@@ -676,6 +676,25 @@ def has_changed(self, initial, data):
...
+        bound_field = BoundField(form, self, field_name)
+        value = bound_field.value()
+        widget = bound_field.field.widget
+        # Save the selected choices in the widget even after the filterform is submitted
+        if value is not None:
+            widget.choices = [(v, v) for v in value]
+
+        return bound_field
+
+
```
bryanculver, line 697: Looks like this chain needs to be updated:

https://github.com/nautobot/nautobot/blob/develop/nautobot/utilities/filters.py#L527

https://github.com/nautobot/nautobot/blob/b027b6414c0be1222177793a8691793365354afa/nautobot/utilities/filters.py#L73-L74

https://github.com/nautobot/nautobot/blob/b027b6414c0be1222177793a8691793365354afa/nautobot/utilities/filters.py#L36-L60

Which should fix the magic generation issue, eliminating the need to specify the field.

And is the choices only helping on page reload pre-populate the inputted text? Is the necessary?

HanlinMiao: Yeah the `widget.choices = [(v, v) for v in value] `are there after the filter is applied to keep the user input.

bryanculver (approved):
This should get merged and incorporated into #2348

glennmatthews (commented):
Would it be desirable to add an integration test or two?

bryanculver (changes requested):
> Would it be desirable to add an integration test or two?

Valid point @glennmatthews. I had more considered this being a component in the UI-side testing for #2348, however this does introduce a new Filter too which should be tested for the features it's providing.

HanlinMiao:
Seems like `MultiValueCharFilter` is already being tested [here](https://github.com/nautobot/nautobot/blob/40e8db640611381b2d16e0feb75aefb4896daddb/nautobot/utilities/tests/test_filters.py#L512). Should we just test the `MultiValueCharField` by simulating how a user would use it in an browser(name, enter, name enter etc.) and see if it renders the correct result?

bryanculver:
Merged via #2413

PR #2382, opened 2022-09-07
Remove cruft from Docker dev image; use no-binary pyuwsgi

glennmatthews:
# Closes: #DNE
# What's Changed

- Poetry 1.2 adds a `installer.no-binary` config option, which we can use to force a local build of pyuwsgi via Poetry, avoiding the need for a separate PYUWSGI_VER build argument and separate `pip install --no-binary pyuwsgi` command.
- Remove some more unneeded cruft from the `dev` and `final-dev` Docker images:
  - Unlike pip, Poetry doesn't have a `--no-cache-dir` option, and because it uses pip under the hood, various Poetry steps will populate both the pip and Poetry caches on disk. Rather than manually clean up these caches at the end of the build (as was incompletely implemented in #2349), use Docker buildkit [cache-mount](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md#run---mounttypecache) feature to essentially treat these directories as temporary filesystems that don't get propagated forward into the final image.
  - There is some interaction between Poetry 1.2.0, pip, and the `requests` library that results in certificate files being created in `/tmp` for each package Poetry installs. https://github.com/python-poetry/poetry/issues/6425 was closed, but as a workaround, it's simple enough to manually `rm -rf /tmp/tmp*` after each Poetry command we run.

As a result, the `dev`/`final-dev` image `/tmp` directory is once again empty, and the `/root/.cache/` directory is nearly empty as well, reducing the final image size by about 120 MB as compared to the 1.4.2 image:

```
root@bdbf0d213905:/source# ls /tmp
root@bdbf0d213905:/source# du -sh /root/.cache/
12K	/root/.cache/
```

```
docker image ls
REPOSITORY                          TAG                              IMAGE ID       CREATED          SIZE
networktocode/nautobot-dev-py3.7    local                            f6076c14a9d6   22 seconds ago   1.37GB
networktocode/nautobot-dev          1.4.2                            74d5c0da8dc7   46 hours ago     1.49GB
```

Also confirmed that `pyuwsgi` still supports SSL:

```
root@f815d36c2c05:/source# uwsgi --help | grep https
    --https-socket                          bind to the specified UNIX/TCP socket using HTTPS protocol
    --https-socket-modifier1                force the specified modifier1 when using HTTPS protocol
    --https-socket-modifier2                force the specified modifier2 when using HTTPS protocol
    --https                                 add an https router/server on the specified address with specified certificate and key
    --https2                                add an https/spdy router/server using keyval options
    --https-export-cert                     export uwsgi variable HTTPS_CC containing the raw client certificate
    --https-session-context                 set the session id context to the specified value
    --http-to-https                         add an http router/server on the specified address and redirect all of the requests to https
```

# TODO

- [x] CI updates

### DIFF
--- a/.github/workflows/ci_integration.yml
+++ b/.github/workflows/ci_integration.yml
@@ -296,8 +296,6 @@ jobs:
         uses: "actions/checkout@v2"
       - run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
         id: "gitbranch"
-      - run: echo "##[set-output name=version;]$(egrep -m1 'file.*pyuwsgi-' poetry.lock | cut -f2 -d-)"
-        id: "pyuwsgi"
       - name: "Set up QEMU"
         uses: "docker/setup-qemu-action@v1"
       - name: "Set up Docker Buildx"
@@ -338,7 +336,6 @@ jobs:
           context: "."
           build-args: |
             PYTHON_VER=${{ matrix.python-version }}
-            PYUWSGI_VER=${{ steps.pyuwsgi.outputs.version }}
             POETRY_PARALLEL=true
       - name: "Docker Dev Metadata"
         id: "dockerdevmeta"
@@ -370,5 +367,4 @@ jobs:
           context: "."
           build-args: |
             PYTHON_VER=${{ matrix.python-version }}
-            PYUWSGI_VER=${{ steps.pyuwsgi.outputs.version }}
             POETRY_PARALLEL=true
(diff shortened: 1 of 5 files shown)


gsnider2195 (approved):
Installed locally on py3.7 and verified nautobot is working correctly and the files were removed as intended.

```
networktocode/nautobot-dev-py3.7    local      0afc854d4343   17 minutes ago   1.29GB
networktocode/nautobot-dev-py3.9    local      6f284dded9ca   4 days ago       1.51GB
```

bryanculver (approved):
Verified plugins still build correctly as well (since they use `final-dev` as their base).

PR #2389, opened 2022-09-08
Removes source lookup from extras.jobs.Job

abates:
Since source code is no longer displayed in Nautobot Jobs (see nautobot/nautobot/issues#1462), `inspect.getsource` is no longer needed. Additionally, `inspect.getsource` will throw an exception in some cases, such as dynamically created classes.

# What's Changed

This change removes source code lookups for extras.jobs.Job. Since the job source is no longer displayed in the job view, this lookup is not needed.

# TODO

- [ ] ~~Attached Screenshots, Payload Example~~
- [X] Unit, Integration Tests
- [ ] ~~Documentation Updates (when adding/changing features)~~
- [ ] ~~Example Plugin Updates (when adding/changing features)~~
- [ ] ~~Outline Remaining Work, Constraints from Design~~

### DIFF
--- a/nautobot/extras/jobs.py
+++ b/nautobot/extras/jobs.py
@@ -111,9 +111,6 @@ class BaseJob:
         self.failed = False
         self._job_result = None
 
-        # Grab some info about the job
-        self.source = inspect.getsource(self.__class__)
-
         # Compile test methods and initialize results skeleton
         self.test_methods = []
 


bryanculver:
Should we do any defensive coding pre 2.0 to just set this to `None` temporarily?

abates:
> Should we do any defensive coding pre 2.0 to just set this to `None` temporarily?

Based on my discussion with @lampwins it seemed like there wasn't a plan to have the source code be visible. If that isn't the case and the source _may_ be displayed in the future, then I think it would be better to proactively wrap it up in a `try/catch` block to prevent the issue that I stumbled upon today.

lampwins:
My vote is just to rip it out, but if @bryanculver wants to do `self.source = None` that is fine too.

bryanculver:
> My vote is just to rip it out, but if @bryanculver wants to do `self.source = None` that is fine too.
> 
> To my knowledge, `self.source` as never a part of the publicly documented API.

Let's rip it out then.

bryanculver (approved):

glennmatthews (approved):

Issue #2299, opened 2022-08-24
Dynamic Group: AttributeError on detail view of parent group containing child group filtered on string value


jifox:
### Environment
* Python version: 3.9
* Nautobot version: df6620fec8ac (v1.4.1)

### Steps to Reproduce
1. Added dynamic group (dcim | device)
2. Add another dynamic group to use as Child Group
3. Add Child Group to 1st dynamc group --> Error on Update

### Expected Behavior
Display 0 devices if empty or 2 if selected

### Observed Behavior

**Error during template rendering**
In template /usr/local/lib/python3.9/site-packages/nautobot/core/templates/inc/table.html, error at line 20

**'str' object has no attribute 'get_absolute_url'**

```
10	                    {% else %}
11	                        <th {{ column.attrs.th.as_html }}>{{ column.header }}</th>
12	                    {% endif %}
13	                {% endfor %}
14	            </tr>
15	        </thead>
16	    {% endif %}
17	    <tbody>
18	        {% for row in table.page.object_list|default:table.rows %}
19	            <tr {{ row.attrs.as_html }}>
20	                {% for column, cell in row.items %}
21	                    <td {{ column.attrs.td.as_html }}>{{ cell }}</td>
22	                {% endfor %}
23	            </tr>
24	        {% empty %}
25	            {% if table.empty_text %}
26	                <tr>
27	                    <td colspan="{{ table.columns|length }}" class="text-center text-muted">&mdash; {{ table.empty_text }} &mdash;</td>
28	                </tr>
29	            {% endif %}
30	        {% endfor %}
```

**Traceback:**

```
Environment:


Request Method: GET
Request URL: http://localhost:8080/extras/dynamic-groups/test-switches/

Django Version: 3.2.15
Python Version: 3.9.13
Installed Applications:
['django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.humanize',
 'cacheops',
 'corsheaders',
 'django_filters',
 'django_jinja',
 'django_tables2',
 'django_prometheus',
 'mptt',
 'social_django',
 'taggit',
 'timezone_field',
 'nautobot.core.apps.NautobotConstanceConfig',
 'nautobot.core',
 'django.contrib.admin',
 'django_celery_beat',
 'rest_framework',
 'db_file_storage',
 'nautobot.circuits',
 'nautobot.dcim',
[...]

PR #2391, opened 2022-09-09
Fix `AttributeError` on Dynamic Group detail view parent group containing child group filtered on single value.

jathanism:
# Closes: #2299 
# What's Changed

This removes the `filter` column and `render_filter()` method from `DynamicGroupMembershipTable`. The implementation is flawed and of questionable use in its current form and should be revisited in a future release.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/tables.py
+++ b/nautobot/extras/tables.py
@@ -324,47 +324,11 @@ class DynamicGroupMembershipTable(DynamicGroupTable):
             "operator",
             "name",
             "weight",
-            # "filter",  # FIXME(jathan): Revisit rendering this in a future release.
             "members",
             "description",
             "actions",
         )
         exclude = ("content_type",)
-
-    def render_filter(self, value, record):
-        """Turns the filter dict into a prettified list of HTML links."""
-        # Display an empty filter as None
-        if not value:
-            return None
-
-        # Use the filterset for the record to construct links to the objects used in the filter.
-        fs = record.group.filterset_class(record.filter, record.group.get_queryset())
-        fs.is_valid()  # Required or we don't get the inner form's `cleaned_data`
-
-        # Iterate over each key in the filter and extract the value from the inner form's
-        # cleaned_data`, calling `get_absolute_url()` on each to create links.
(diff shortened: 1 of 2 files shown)


bryanculver:
Is there a good use case for keeping this column? The more we discussed it it seems like this column provides little value, especially hyperlinked as it is.

See: https://github.com/nautobot/nautobot/issues/2299#issuecomment-1232097519

jathanism:
> Is there a good use case for keeping this column? The more we discussed it it seems like this column provides little value, especially hyperlinked as it is.
> 
> See: [#2299 (comment)](https://github.com/nautobot/nautobot/issues/2299#issuecomment-1232097519)

I may be biased since I implemented it, but I like the visibility of what each child is providing to the group filter.

bryanculver:
> > Is there a good use case for keeping this column? The more we discussed it it seems like this column provides little value, especially hyperlinked as it is.
> > See: [#2299 (comment)](https://github.com/nautobot/nautobot/issues/2299#issuecomment-1232097519)
> 
> I may be biased since I implemented it, but I like the visibility of what each child is providing to the group filter.

The problem is is that it's not entirely accurate because the group of groups while not having a filter directly does eventually resolve a filter that's being applied. It's not the nested group's filter individually that's being applied but the collective filter from all nested groups, which may be included as a NOT instead of an OR or AND. The filter that is displayed doesn't show that.

jathanism:
As discussed in our stand-up, let's leave this fix in and I'll go ahead and disable the column for now.

nautobot/extras/tables.py
```diff
@@ -346,9 +346,22 @@ def render_filter(self, value, record):
...
+            filters = fs.form.cleaned_data[field_name]
+            links = []
+
+            for item in filters:
+                if isinstance(item, (str, int)):  # single-value char/int filter
+                    multi = False
+                    links.append(repr(item))
+                else:
+                    multi = True
+                    links.append(format_html('<a href="{}">{}</a>', item.get_absolute_url(), item))
```
glennmatthews, line 358: Won't this still fail for multi-value char/int filters?

bryanculver (approved):

Issue #1619, opened 2022-04-07
Bulk Operation Schema Views are Incorrect


bryanculver:
### Environment
* Python version: 3.6+
* Nautobot version: 1.2+ (confirmed in `next` as well)

### Steps to Reproduce
1. View https://demo.nautobot.com/api/docs/#operations-ipam-ipam_ip-addresses_bulk_partial_update
2.
3.

### Expected Behavior

Schema for bulk update should be a list of the models.

### Observed Behavior

Schema shows as a single object for the bulk update methods.

glennmatthews:
This appears to still be an issue after #1544.

PR #2398, opened 2022-09-09
Fixes to OpenAPI schema for bulk API operations

glennmatthews:
# Closes: #1619 
# What's Changed

Because Nautobot's bulk REST API operations (PUT, PATCH, DELETE) are bespoke implementations that are not part of the DRF standard operations set, the OpenAPI schema that drf-spectacular was generating for these operations was incomplete/incorrect in a number of ways, including:

- Request schema for bulk PUT/PATCH was showing as a single object representation, instead of what should be a list of object dictionaries with mandatory `id` field
- Request schema for bulk DELETE was showing no request body, instead of what should be a list of `{id: <uuid>}` dictionaries.
- Response schema for bulk PUT/PATCH was showing as a single object representation, instead of what should be a list

This PR includes the following fixes:

- Upgrade to latest drf-spectacular release 0.23.1
- Enable drf-spectacular `COMPONENT_SPLIT_PATCH` setting, allowing for distinct request schemas for PATCH versus POST/PUT
- Enable drf-spectacular `COMPONENT_SPLIT_REQUEST` setting, allowing for distinct schemas for request vs response
- The folllowing additions/changes to our custom `NautobotAutoSchema` subclass of drf-spectacular's `AutoSchema`:
  - ~Override ~internal `_get_request_body()`~ `get_operation()` method to allow for a request body for bulk-DELETE. drf-spectacular's base implementation here is hard-coded to assume that DELETE requests never have a request body, so we needed to override this to allow our exception to this case.
    - [x] ~TODO: investigate whether it would be possible/cleaner to inject this logic at a higher level without overriding an internal method, even if it means we need to write more lines of code~
  - Extend our overridden `get_request_serializer()` method implementation to handle bulk actions and PATCH actions specifically.
    - Extend our custom `get_writable_class()` method to be capable of generating `BulkWritable*Serializer` dynamic classes as well as the existing `Writable*Serializer` dynamic classes.
  - Override `get_response_serializers()` to recognize that bulk PUT/PATCH return a list of objects rather than a single object
    - Because bulk-PUT/PATCH return a bare list of objects (`[{...}, {...}, ...]`) instead of a standard REST API paginated-style list (`{"results": [{...}, {...}, ...], ...}`) and this list is not filterable like a GET endpoint, I also had to override `get_filter_backends()` and the internal `_get_paginator()` method to have both make an exception for bulk operations. (`_get_paginator` is called in a number of places fairly deep inside the guts of the AutoSchema class, so I don't think it'll be possible to avoid overriding it even though it's a private method)
      - I've added a couple of `2.0 TODO` comments for us to revisit this aspect of the REST API design to be more self-consistent.
  - Override `resolve_serializer()` to recognize that in our custom bulk-PATCH requests, the `id` field is always required (by default drf-spectactular assumes that there are no required fields in any PATCH request)
- Explicitly decorate the `BulkDestroyModelMixin.bulk_destroy` view to represent its request body as a `BulkOperationSerializer`
  - Change `BulkOperationSerializer` to always use a UUID as the `id` field (rather than a generic string), as this is more accurate for 99% of our models.
  - For the `Group` model, which as a Django built-in still uses integer PKs, add `BulkOperationIntegerIDSerializer` and use it where needed.
- Because we now have `BulkWritable*Serializer` as well as `Writable*Serializer`, we need to explicitly name the `Status` choices for *all* Status-containing model in `settings.py` to prevent drf-spectacular from complaining about collisions between enums.

Example bulk PUT/PATCH/DELETE schemas, showing list inputs and outputs and correct marking of required fields:

Screen Shot 2022-09-09 at 10 52 54 AM

Screen Shot 2022-09-09 at 10 53 19 AM

Screen Shot 2022-09-09 at 10 53 34 AM

Bulk PUT/PATCH/DELETE for Groups, with special-case integer PK:

Screen Shot 2022-09-09 at 10 54 01 AM

Screen Shot 2022-09-09 at 10 54 10 AM

Screen Shot 2022-09-09 at 10 54 19 AM

# TODO

- n/a Unit, Integration Tests

### DIFF
--- a/nautobot/core/api/schema.py
+++ b/nautobot/core/api/schema.py
@@ -234,14 +234,14 @@ class NautobotAutoSchema(AutoSchema):
             ref_name = ref_name[: -len("Serializer")]
         return ref_name
 
-    def resolve_serializer(self, serializer, direction):
+    def resolve_serializer(self, serializer, direction, bypass_extensions=False):
         """
         Re-add required `id` field on bulk_partial_update action.
 
         drf-spectacular clears the `required` list for any partial serializers in its `_map_basic_serializer()`,
         but Nautobot bulk partial updates require the `id` field to be specified for each object to update.
         """
-        component = super().resolve_serializer(serializer, direction)
+        component = super().resolve_serializer(serializer, direction, bypass_extensions)
         if self.is_bulk_action and self.is_partial_action and direction == "request":
             component.schema["required"] = ["id"]
         return component
--- a/nautobot/core/api/schema.py
+++ b/nautobot/core/api/schema.py
@@ -4,6 +4,7 @@ import re
 from drf_spectacular.contrib.django_filters import DjangoFilterExtension
 from drf_spectacular.extensions import OpenApiSerializerFieldExtension
 from drf_spectacular.openapi import AutoSchema
+from drf_spectacular.plumbing import build_media_type_object, is_serializer
 from rest_framework import serializers
 from rest_framework.relations import ManyRelatedField
 
@@ -36,6 +37,69 @@ class NautobotAutoSchema(AutoSchema):
         }
     )
 
+    @property
+    def is_bulk_action(self):
+        """Custom property for convenience."""
+        return hasattr(self.view, "action") and self.view.action in self.custom_actions
+
+    @property
+    def is_partial_action(self):
+        """Custom property for convenience."""
+        return hasattr(self.view, "action") and self.view.action in ["partial_update", "bulk_partial_update"]
+
+    def _get_paginator(self):
+        """Nautobot's custom bulk operations, even though they return a list of records, are NOT paginated."""
+        if self.is_bulk_action:
+            return None
+        return super()._get_paginator()
+
+    def _get_request_body(self, direction="request"):
+        """Get the dict of information that defines the request portion of the schema for a given operation.
+
+        We have to override this method because drf-spectacular itself asserts that DELETE operations
+        never have a request body, which is not the case for our bulk_delete action.
+        """
+        request_body = super()._get_request_body(direction=direction)
+        if not request_body and self.is_bulk_action and self.method == "DELETE":
+            # The below is copied from drf-spectacular's `_get_request_body()` method,
+            # minus the "only unsafe methods can have a body" assertion
+            request_serializer = self.get_request_serializer()
+
+            if isinstance(request_serializer, dict):
+                content = []
+                request_body_required = True
+                for media_type, serializer in request_serializer.items():
+                    schema, partial_request_body_required = self._get_request_for_media_type(serializer, direction)
+                    examples = self._get_examples(serializer, direction, media_type)
+                    if schema is None:
+                        continue
+                    content.append((media_type, schema, examples))
+                    request_body_required &= partial_request_body_required
+            else:
+                schema, request_body_required = self._get_request_for_media_type(request_serializer, direction)
+                if schema is None:
+                    return None
+                content = [
+                    (media_type, schema, self._get_examples(request_serializer, direction, media_type))
+                    for media_type in self.map_parsers()
+                ]
+
+            request_body = {
+                "content": {
+                    media_type: build_media_type_object(schema, examples) for media_type, schema, examples in content
+                }
+            }
+            if request_body_required:
+                request_body["required"] = request_body_required
+
+        return request_body
+
+    def get_filter_backends(self):
+        """Nautobot's custom bulk operations, even though they return a list of records, are NOT filterable."""
+        if self.is_bulk_action:
+            return []
+        return super().get_filter_backends()
+
     def get_operation_id(self):
         """Extend the base method to handle Nautobot's REST API bulk operations.
 
@@ -45,7 +109,7 @@ class NautobotAutoSchema(AutoSchema):
 
         With this extension, the bulk endpoints automatically get a different operation-id from the non-bulk endpoints.
         """
-        if hasattr(self.view, "action") and self.view.action in self.custom_actions:
+        if self.is_bulk_action:
             # Same basic sequence of calls as AutoSchema.get_operation_id,
             # except we use "self.view.action" instead of "self.method_mapping[self.method]" to get the action verb
             tokenized_path = self._tokenize_path()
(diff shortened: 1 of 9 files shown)


bryanculver (approved):
Spot checked a few endpoints between generated schemas and actual function and looks good.

nautobot/core/api/schema.py
```diff
@@ -145,6 +234,18 @@ def get_serializer_ref_name(self, serializer):
...
+        """
+        Re-add required `id` field on bulk_partial_update action.
+
+        drf-spectacular clears the `required` list for any partial serializers in its `_map_basic_serializer()`,
+        but Nautobot bulk partial updates require the `id` field to be specified for each object to update.
+        """
+        component = super().resolve_serializer(serializer, direction)
+        if self.is_bulk_action and self.is_partial_action and direction == "request":
+            component.schema["required"] = ["id"]
+        return component
```
jathanism, line 247: I hate that we have to do all of this and if there's any reasonable expectation that we can undo this in v2, can we please add "v2 TODO" for these?

nautobot/core/api/schema.py
```diff
@@ -107,25 +174,47 @@ def get_writable_class(self, serializer):
...
+                            field.required = True
+                        new_fields[name] = field
+                    return new_fields
 
-        writable_class = self.writable_serializers[type(serializer)]
+                properties["get_fields"] = get_fields
+
+            self.writable_serializers[writable_name] = type(writable_name, (type(serializer),), properties)
+
+        writable_class = self.writable_serializers[writable_name]
```
jathanism, line 217: I hate that we have to do all of this and if there's any reasonable expectation that we can undo this in v2, can we please add "v2 TODO" for these?

nautobot/core/api/schema.py
```diff
@@ -36,6 +37,49 @@ class NautobotAutoSchema(AutoSchema):
...
+
+            operation["requestBody"] = {
+                "content": {
+                    media_type: build_media_type_object(build_array_type(component.ref))
+                    for media_type in self.map_parsers()
+                },
+                "required": True,
+            }
+
+        return operation
```
jathanism, line 81: Another ripe opportunity for a "v2 TODO" I hope? Such as if we can align around normalized responses? Or is the lack of filtering/pagination for these bulk responses always going to be an issue here? (Such as how GET will always be paginated, but all the others won't?)

jathanism (commented):
All in all this is pretty concise. My only concerns are around whether or not we can move away from this in v2?

nautobot/core/api/schema.py
glennmatthews: I think bulk-PATCH will always require `id` as a field, as how else to identify the specific instances being patched? Unless we want to get rid of these bulk operations entirely in v2.

nautobot/core/api/schema.py
glennmatthews: This part is specifically for bulk-DELETE; I think we could possibly push this upstream to get drf-spectacular to remove its code that hard-codes all DELETE request bodies to null.

nautobot/core/api/schema.py
```diff
@@ -36,6 +37,49 @@ class NautobotAutoSchema(AutoSchema):
...
+        """Nautobot's custom bulk operations, even though they return a list of records, are NOT paginated."""
+        if self.is_bulk_action:
+            return None
+        return super()._get_paginator()
+
+    def get_filter_backends(self):
+        """Nautobot's custom bulk operations, even though they return a list of records, are NOT filterable."""
+        if self.is_bulk_action:
+            return []
+        return super().get_filter_backends()
```
glennmatthews, line 60: Regarding v2 - I don't think it makes sense for bulk-PATCH and bulk-PUT to support filter parameters or to provide true paginated responses (how would you retrieve "page 2" of a PATCH response?). I'd like to push an issue up to drf-spectacular to make `get_paginator` into a non-private method though.

nautobot/core/api/schema.py
glennmatthews: I'm definitely open to ideas on how we could make the bulk and non-bulk serializers more similar. I don't see an obvious path to doing so though.

nautobot/core/api/schema.py
jathanism: I suppose what I meant is if there's a way we can get away with it without requiring so much high touch. But perhaps now that this is solved it's less of a concern.

nautobot/core/api/schema.py
jathanism: > This part is specifically for bulk-DELETE; I think we could possibly push this upstream to get drf-spectacular to remove its code that hard-codes all DELETE request bodies to null.

I like this idea.

jathanism (approved):

glennmatthews:
Looks like drf-spectacular 0.24 was just released - I'd like to evaluate it and see if it adds anything worth including in this PR.

nautobot/core/api/schema.py
glennmatthews: Opened https://github.com/tfranzel/drf-spectacular/issues/809.

nautobot/core/settings.py
```diff
@@ -221,11 +221,24 @@
         "PowerPortTypeChoices": "nautobot.dcim.choices.PowerPortTypeChoices",
         "RackTypeChoices": "nautobot.dcim.choices.RackTypeChoices",
         "RelationshipTypeChoices": "nautobot.extras.choices.RelationshipTypeChoices",
-        # Because Interface and VMInterface, and Site and Location, have the same default statuses, we get the error:
+        # Each of these StatusModels has bulk and non-bulk serializers, with the same status options, so we'd get error:
```
itdependsnetworks, line 224: get `an` error?

nautobot/core/settings.py
```diff
@@ -221,11 +221,24 @@
...
         #   enum naming encountered a non-optimally resolvable collision for fields named "status"
```
glennmatthews, line 225:
```suggestion
        # Each of these StatusModels has bulk and non-bulk serializers, with the same status options,
        # which confounds drf-spectacular's automatic naming of enums, resulting in the below warning:
        #   enum naming encountered a non-optimally resolvable collision for fields named "status"
        # By explicitly naming the enums ourselves we avoid this warning.
```

Issue #2309, opened 2022-08-25
404 error on scheduled-jobs <id> page


FloLaco:
### Environment
* Python version: 3.8.13
* Nautobot version: 1.3.5

### Steps to Reproduce
1. Add a git repo for importing jobs
2. Create a schedule jobs depending on this repo
3. Modify the name/slug of the git repo on nautobot (and branch) (I'm not the person who edit the name, so I'm supposing the other one did it because of the view of `jobs results` page which show us another name)
4. All futur scheduled jobs are in failed state on job results page
5. On `extras/jobs/scheduled-jobs/` page, when we click on the schedule name, I get the 404 error

### Expected Behavior
Should see the datail of my schedule job

### Observed Behavior
I cannot access to detail. I get a 404 error page.
```
Page not found (404)
Request Method:	GET
Request URL:	https://network-cmdb.carrefour.com/extras/jobs/scheduled-jobs/daa35077-a263-403a-9384-b542b51ebbfc/
Raised by:	nautobot.extras.views.ScheduledJobView
```

```
Using the URLconf defined in nautobot.core.urls, Django tried these URL patterns, in this order:

[name='home']
search/ [name='search']
login/ [name='login']
logout/ [name='logout']
circuits/
dcim/
extras/ changelog/ [name='objectchange_list']
extras/ changelog/<uuid:pk>/ [name='objectchange']
extras/ computed-fields/ [name='computedfield_list']
extras/ computed-fields/add/ [name='computedfield_add']
extras/ computed-fields/delete/ [name='computedfield_bulk_delete']
extras/ computed-fields/<slug:slug>/ [name='computedfield']
extras/ computed-fields/<slug:slug>/edit/ [name='computedfield_edit']
extras/ computed-fields/<slug:slug>/delete/ [name='computedfield_delete']
extras/ computed-fields/<slug:slug>/changelog/ [name='computedfield_changelog']
extras/ config-contexts/ [name='configcontext_list']
extras/ config-contexts/add/ [name='configcontext_add']
extras/ config-contexts/edit/ [name='configcontext_bulk_edit']
extras/ config-contexts/delete/ [name='configcontext_bulk_delete']
extras/ config-contexts/<uuid:pk>/ [name='configcontext']
extras/ config-contexts/<uuid:pk>/edit/ [name='configcontext_edit']
extras/ config-contexts/<uuid:pk>/delete/ [name='configcontext_delete']
[...]

PR #2410, opened 2022-09-12
Change job_class no longer found behavior on ScheduledJobView

bryanculver:
# Closes: #2309
# What's Changed

- No longer return 404 on ScheduledJobView when Job class is not found
- Show alert on ScheduledJobView
- Job kwargs lose labels when code is gone, hide kwargs.

Screen Shot 2022-09-12 at 16 16 13

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/templates/extras/scheduledjob.html
+++ b/nautobot/extras/templates/extras/scheduledjob.html
@@ -13,7 +13,7 @@
 {% endblock buttons %}
 
 {% block content %}
-{% if job_class_found == False %}
+{% if not job_class_found %}
 <div class="alert alert-danger">
     <i class="mdi mdi-alert"></i>
     This job source for this scheduled job is no longer installed.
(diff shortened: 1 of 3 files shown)


nautobot/extras/views.py
```diff
@@ -1354,22 +1354,17 @@ class ScheduledJobApprovalQueueListView(generic.ObjectListView):
...
+        job_class = get_job(instance.job_class)
         labels = {}
-        for name, var in job_class._get_vars().items():
-            field = var.as_field()
-            if field.label:
-                labels[name] = var
-            else:
-                labels[name] = pretty_name(name)
-        return {"labels": labels}
+        if job_class:
```
glennmatthews, line 1360:
```suggestion
        if job_class is not None:
```

nautobot/extras/templates/extras/scheduledjob.html
```diff
@@ -12,6 +12,17 @@
     {% endif %}
 {% endblock buttons %}
 
+{% block content %}
+{% if job_class_found == False %}
```
glennmatthews, line 16:
```suggestion
{% if not job_class_found %}
```

glennmatthews (approved):

PR #2411, opened 2022-09-13
Swagger API Docs render & use appropriate Token header

### LINKED ISSUE #2154: Swagger documentation lists incorrect authorization header
### Environment
* Python version: 3.10
* Nautobot version: 1.39

### Steps to Reproduce
1. Open the swagger API interface 
2. Attempt to use the included curl commands to execute API calls

### Expected Behavior
The listed authorzation header value should be 
Authorization: Token <VALUE>

### Observed Behavior
THe header value is instead
Authorzation <VALUE>

image

whitej6:
# Closes: #2154 
# What's Changed

- Vendored the swagger.js template to account for Token keyword in API
- Added additional value to request data for template
- Pointed swagger html to template to use the new vendored js template

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/core/api/views.py
+++ b/nautobot/core/api/views.py
@@ -491,6 +491,9 @@ class NautobotSpectacularSwaggerView(APIVersioningGetSchemaURLMixin, Spectacular
         response = super().get(request, *args, **kwargs)
         response.data["swagger_settings"] = response.data["settings"]
         del response.data["settings"]
+
+        # Add additional data so drf-spectacular will use the Token keyword in authorization header.
+        response.data["schema_auth_names"] = ["tokenAuth"]
         return response
 
 
--- a/nautobot/core/templates/swagger_ui.js
+++ b/nautobot/core/templates/swagger_ui.js
@@ -84,7 +84,11 @@ const injectAuthCredentials = (request) => {
       return;
     // This conditional resolves issue #2154 and is the only portion that is updated for this patch.
     } else if (authDef.schema.type === "apiKey" && authDef.schema.in === "header" && authDef.schema.description === 'Token-based authentication with required prefix "Token"') {
+      if (authDef.value.startsWith("Token ")) {
+        request.headers["Authorization"] = authDef.value;
+      } else {
         request.headers["Authorization"] = "Token " + authDef.value;
+      }
         return;
     // End of block for resolving issue #2154.
     } else if (authDef.schema.type === "apiKey" && authDef.schema.in === "header") {
--- a/nautobot/core/templates/swagger_ui.js
+++ b/nautobot/core/templates/swagger_ui.js
@@ -82,9 +82,11 @@ const injectAuthCredentials = (request) => {
     } else if (authDef.schema.type === "http" && authDef.schema.scheme === "basic") {
       request.headers["Authorization"] = "Basic " + btoa(authDef.value.username + ":" + authDef.value.password);
       return;
+    // This conditional resolves issue #2154 and is the only portion that is updated for this patch.
     } else if (authDef.schema.type === "apiKey" && authDef.schema.in === "header" && authDef.schema.description === 'Token-based authentication with required prefix "Token"') {
         request.headers["Authorization"] = "Token " + authDef.value;
         return;
+    // End of block for resolving issue #2154.
     } else if (authDef.schema.type === "apiKey" && authDef.schema.in === "header") {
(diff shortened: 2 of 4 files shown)


nautobot/core/api/views.py
```diff
@@ -491,6 +491,9 @@ def get(self, request, *args, **kwargs):
         response = super().get(request, *args, **kwargs)
         response.data["swagger_settings"] = response.data["settings"]
         del response.data["settings"]
+
+        # Add additional data so drf-spectacular will use the Token keyword in authorization header.
+        response.data["schema_auth_names"] = ["tokenAuth"]
```
glennmatthews, line 496: It would be more correct to populate this into drf-spectacular (likely by implementing an [OpenApiAuthenticationExtension class](https://drf-spectacular.readthedocs.io/en/latest/customization.html#specify-authentication-with-openapiauthenticationextension)) rather than hard-coding it to this view.

(From a quick read though, this *should* be working automatically already, given that [drf-spectacular's built in `TokenScheme` says `match_subclasses = True`](https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/authentication.py#L35) and our `TokenAuthentication` class is a subclass of `rest_framework.authentication.TokenAuthentication`... not sure what I'm missing here)

nautobot/core/templates/swagger_ui.js
```diff
@@ -0,0 +1,121 @@
...
+    }
+    if (authDef.schema.type === "http" && authDef.schema.scheme === "bearer") {
+      request.headers["Authorization"] = "Bearer " + authDef.value;
+      return;
+    } else if (authDef.schema.type === "http" && authDef.schema.scheme === "basic") {
+      request.headers["Authorization"] = "Basic " + btoa(authDef.value.username + ":" + authDef.value.password);
+      return;
+    } else if (authDef.schema.type === "apiKey" && authDef.schema.in === "header" && authDef.schema.description === 'Token-based authentication with required prefix "Token"') {
+        request.headers["Authorization"] = "Token " + authDef.value;
+        return;
```
glennmatthews, line 87: Is this the only actual change to the file? Would you mind adding some comments before/after this block to make this easy to locate/identify in the future in case we need to update this file from upstream?

whitej6: Will do

nautobot/core/api/views.py
whitej6: _Maybe_..... IF a deployment just has publicly installable plugins, I see no issue. BUT if someone installs a private package they _may_ not want to expose the schema.

I will dig more into this.

nautobot/core/templates/swagger_ui.js
```diff
@@ -0,0 +1,123 @@
...
+    }
+    if (authDef.schema.type === "http" && authDef.schema.scheme === "bearer") {
+      request.headers["Authorization"] = "Bearer " + authDef.value;
+      return;
+    } else if (authDef.schema.type === "http" && authDef.schema.scheme === "basic") {
+      request.headers["Authorization"] = "Basic " + btoa(authDef.value.username + ":" + authDef.value.password);
+      return;
+    // This conditional resolves issue #2154 and is the only portion that is updated for this patch.
+    } else if (authDef.schema.type === "apiKey" && authDef.schema.in === "header" && authDef.schema.description === 'Token-based authentication with required prefix "Token"') {
+        request.headers["Authorization"] = "Token " + authDef.value;
```
bryanculver, line 87: Can you check `authDef.value` to make sure users haven't already provided `Token ` in their input for backwards compatibility?

And can we strip the cookies from the request so passing a token doesn't silently pass (requests appear to work today with the incorrect value because our API accepts both the session and authorization header).

bryanculver (changes requested):
Changes requested.

Also, we use [sidecar](https://github.com/tfranzel/drf-spectacular-sidecar) as far as I'm aware which is used for `collectstatic` so will this actually work for production deploys?

whitej6:
Unless something changed we already overload and use a custom html template for one specific var that is passed to this template. The JS file isn't served via static from upstream either.

bryanculver:
> Unless something changed we already overload and use a custom html template for one specific var that is passed to this template. The JS file isn't served via static from upstream either.

👍 

Sounds good. Just checking. So yeah if we can gracefully allow `Token ` to also be in the input for this who are doing that already I’m good with this.

bryanculver (approved):

Issue #2417, opened 2022-09-13
Speed Up Docker Build, Bumping Up Cache Limits


bryanculver:
### Proposed Changes
Publish a `nautobot-dependencies` container to base builds off of to improve CI.

### Justification
We are at the limit of GitHub Actions cache (10GB, 7 days) and due to our matrix of builds we rarely have all layers available in cache and eviction rules in GitHub Actions cache can't know which layers to smartly evict or now.

PR #2420, opened 2022-09-14
Publish Dependencies Stage as Image, Speed Up Final Builds

bryanculver:
# Closes: #2417
# What's Changed

- New reusable workflow `.github/workflows/build_dependency_image_conditional.yml` added to determine if recent commit changed key files
    - Tested:
        - No files changed: https://github.com/nautobot/nautobot/actions/runs/3057370602/jobs/4932454078
        - Files changed: https://github.com/nautobot/nautobot/actions/runs/3057383781/jobs/4932480535 
- New reusable workflow `.github/workflows/build_dependency_image.yml` added to abstract dependency image build steps. Permits manual execution to force rebuilds.
    - Publishes an image with docker build target `dependences` to `ghcr.io/nautobot/nautobot-dependencies`
    - We don't cache these layers since the intention is for us to pull these built-image sealed
    - Dependency rebuilds will likely never have cached layers as we are evicting frequently
- Integration, Release, Pre-release workflows updated to call `.github/workflows/build_dependency_image_conditional.yml` before final Docker builds
    - Additionally `DEPENDENCIES_BASE_BRANCH` provided to final build environment
- Dockerfile updated to have layers dependent on `dependencies` stage now reference the pre-built image
    - Because `COPY --from` depends on named stage instead of a specific image and tag, we have to alias it as a stage, `dependencies-as-stage`
- Invoke tasks will now build a "local" dependencies base. This permits consistent previous behavior of running `invoke build` to fully build/rebuild an image, especially when updating dependencies locally
    - Optionally, `invoke build` takes a `--dependencies_base=` argument to reference a specific base to speed up builds. No intuitive way to derive this as it depends on the base of current branch
    - We need to use `buildx` but it has a default platform of `linux/amd64` which could cause the final output of `invoke build` to change due to chaining commands (`docker compose` did not have a platform set)
        - Alternatively, we could change this to default behavior to not provide the argument if not specified (see: https://github.com/nautobot/nautobot/blob/develop/tasks.py#L186-L194)
    - Due to the above `build_dependencies` will check for a "default" platform but fall back to `buildx`'s existing behavior
    - `buildx` was updated to take `DEPENDENCIES_BASE_BRANCH` Dockerfile argument

# TODO

- [ ] Unit, Integration Tests


.github/workflows/build_dependcy_image_conditional.yml
```diff
@@ -0,0 +1,44 @@
+---
```
jathanism, line 1: ^ This filename has a typo.

`build_dependcy_image_conditional.yml` -> `build_dependency_image_conditional.yml`

jathanism (changes requested):

.github/workflows/build_dependcy_image_conditional.yml
bryanculver: 🤦

jathanism (approved):
YOLO

.github/workflows/build_dependency_image.yml
```diff
@@ -0,0 +1,62 @@
+---
+# This resuable workflow always builds & tags a Docker image for the Nautobot dependencies stage
```
glennmatthews, line 2:
```suggestion
# This reusable workflow always builds & tags a Docker image for the Nautobot dependencies stage
```

.github/workflows/build_dependency_image.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: "docker/setup-qemu-action@v1"
+      - name: "Set up Docker Buildx"
+        uses: "docker/setup-buildx-action@v1"
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: "Docker Dependenices Metadata"
```
glennmatthews, line 36:
```suggestion
      - name: "Docker Dependencies Metadata"
```

.github/workflows/build_dependency_image.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: "docker/metadata-action@v3"
+        with:
+          images: "ghcr.io/nautobot/nautobot-dependencies"
+          flavor: |
+            latest=false
+          tags: |
+            type=raw,value=${{ steps.gitbranch.outputs.branch }}-py${{ matrix.python-version }}
+            type=raw,value=${{ steps.gitbranch.outputs.branch }}-{{sha}}-{{date 'X'}}-py${{ matrix.python-version }}
+            type=raw,value=${{ steps.gitbranch.outputs.branch }},enable=${{ matrix.python-version == 3.7 }}
+            type=raw,value=${{ steps.gitbranch.outputs.branch }}-{{sha}}-{{date 'X'}},enable=${{ matrix.python-version == 3.7 }}
```
glennmatthews, line 47: Do we particularly need the `sha` tags for this image? Or the non-versioned tag for python == 3.7?

.github/workflows/ci_integration.yml
```diff
@@ -294,6 +295,8 @@ jobs:
     steps:
       - name: "Check out repository code"
         uses: "actions/checkout@v2"
+      - name: "Check & Build Dependencies Container Image"
+        uses: "./.github/workflows/build_dependency_image_conditional.yml"
```
glennmatthews, line 299: Is this path right? Relative to the repository base directory, rather than to the directory the current workflow is defined in?

.github/workflows/prerelease.yml
```diff
@@ -59,6 +60,8 @@ jobs:
     steps:
       - name: "Check out repository code"
         uses: "actions/checkout@v2"
+      - name: "Check & Build Dependencies Container Image"
+        uses: "./.github/workflows/build_dependency_image_conditional.yml"
```
glennmatthews, line 64: Should `prerelease` and `release` perhaps be unconditional and always build the dependencies image?

changes/2417.changed
```diff
@@ -0,0 +1 @@
+Moved to building a published intermediate dependencies images to speed up container build times.
```
glennmatthews, line 1:
```suggestion
Reworked Docker builds in CI to publish an intermediate "dependencies-only" image to speed up rebuild times.
```

glennmatthews (commented):
I'll admit I don't fully understand this one yet.

.github/workflows/prerelease.yml
bryanculver: Thanks for the callout, jogged my thinking that while we tag off of `main`, what ends up being the "reference" in CI for releases is not the branch but tag, see: https://github.com/nautobot/nautobot/actions/runs/2994196181/jobs/4803333403#step:10:123

I'll have to poke around how to solve this.

.github/workflows/ci_integration.yml
bryanculver: Yes, I've tested this here: https://github.com/nautobot/nautobot-plugin-netutils/pull/2/files#diff-d61e1018c78f1c276c1f4ca893c840c6708bcbe7a364c359540eb764802d5e35R12

.github/workflows/build_dependency_image.yml
bryanculver: Done

.github/workflows/prerelease.yml
bryanculver: Tweaked in latest to publish `nautobot-dependencies:vX.Y.Z-py3.N` on every pre-/release.

glennmatthews (approved):
:shipit:

Issue #950, opened 2021-09-28
Concurrency issues(²?)


u1735067:
### Environment
* Python version: 3.9.7
* Nautobot version: 1.1.3-py39
* Redis: redis:6.2.5-alpine
* PSQL: postgres:13.3-alpine

Docker-compose extract:
```yml
---
version: '3.4'

services:
  nautobot: &nautobot
    image: networktocode/nautobot:1.1.3-py3.9
    restart: unless-stopped
    depends_on:
      - postgres
      - redis
    volumes:
      - ./volumes/media:/opt/nautobot/media:z,rw
      - ./volumes/git:/opt/nautobot/git:z,rw
      - ./volumes/jobs:/opt/nautobot/jobs:z,rw
    environment:
      NAUTOBOT_DB_HOST: postgres
      NAUTOBOT_DB_USER: nautobot
      NAUTOBOT_DB_PASSWORD: nautobot
      NAUTOBOT_DB_NAME: nautobot
      NAUTOBOT_REDIS_HOST: redis
      NAUTOBOT_SECRET_KEY: "*****"
      NAUTOBOT_MAX_PAGE_SIZE: "5000"
      NAUTOBOT_CHANGELOG_RETENTION: "366"
      NAUTOBOT_METRICS_ENABLED: "false"
      ### NAUTOBOT_CACHEOPS_ENABLED: "false"
      ### NAUTOBOT_UWSGI_PROCESSES: 1
    healthcheck:
      interval: 30s
      timeout: 15s
      start_period: 5s
      retries: 3
      test: ["CMD-SHELL", "curl --fail http://localhost:8080/health/ || exit 1"]

  celery_worker:
    <<: *nautobot
    entrypoint: "nautobot-server celery worker -l INFO"
    networks:
      - default
    labels: []
    depends_on:
      - nautobot
    healthcheck:
      interval: 30s
      timeout: 15s
      start_period: 5s
      retries: 3
      test: ["CMD", "nautobot-server", "health_check"]

  # postgres - https://hub.docker.com/_/postgres
  postgres:
    image: postgres:13.3-alpine
    restart: unless-stopped
    volumes:
      - ./volumes/pgsql:/var/lib/postgresql/data
    environment:
      TZ: Europe/Paris
      POSTGRES_USER: nautobot
      POSTGRES_PASSWORD: nautobot
      POSTGRES_DB: nautobot

  # redis - https://hub.docker.com/_/redis
  redis:
[...]

PR #2424, opened 2022-09-15
Fix database concurrency issues with uWSGI pre-forking.

jathanism:
# Closes: #950

# What's Changed

- Implements `@uwsgidecorators.postfork` callback to explicitly close database and cache connections when a uWSGI worker is forked from the main thread.
- Explicitly set `WEBSERVER_WARMUP = False` within `nautobot.core.settings` to work with this `postfork` callback. This is an internal setting that is not intended to be user-serviceable.
- Credit to @u1735067 for the detailed troubleshooting and the patience of a saint in identifying and testing the fix for this.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/950.fixed
+++ b/changes/950.fixed
@@ -1,1 +1,1 @@
-Fix database concurrency issues with uWSGI pre-forking.
+Fixed database concurrency issues with uWSGI pre-forking.
--- a/changes/950.fixed
+++ b/changes/950.fixed
@@ -1,0 +1,1 @@
+Fix database concurrency issues with uWSGI pre-forking.
--- a/nautobot/core/wsgi.py
+++ b/nautobot/core/wsgi.py
@@ -1,4 +1,8 @@
+import logging
+
+from django.core import cache
 from django.core.wsgi import get_wsgi_application
+from django.db import connections
 
 import nautobot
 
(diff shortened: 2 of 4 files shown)


changes/950.fixed
```diff
@@ -0,0 +1 @@
+Fix database concurrency issues with uWSGI pre-forking.
```
gsnider2195, line 1:
```suggestion
Fixed database concurrency issues with uWSGI pre-forking.
```

nautobot/core/wsgi.py
```diff
@@ -10,4 +14,23 @@
...
+# will first call `get_internal_wsgi_application` which does not have `uwsgi` module loaded
+# already. Therefore, `settings.WEBSERVER_WARMUP` to `False` for this code to be loaded.
+try:
+    import uwsgidecorators
+
+    @uwsgidecorators.postfork
+    def fix_uwsgi():
+        import uwsgi
+
+        logging.getLogger("nautobot").info(
```
glennmatthews, line 27: Any reason not to use `logging.getLogger(__name__)` (i.e. "nautobot.core.wsgi") instead?

glennmatthews (approved):
Awesome stuff! If you want, you could possibly add u1735067 as a co-author on the commit?

bryanculver (approved):
Will disabling warmup cause any startup behavior changes (slower?) that we should leave a note about in the changelog?

jathanism:
> Will disabling warmup cause any startup behavior changes (slower?) that we should leave a note about in the changelog?

Slightly slower startup, but it's barely noticeable. I would say it would be more of a concern if Nautobot were an I/O intensive app on bootstrap, but it's not.

gsnider2195 (approved):
Great stuff @jathanism and @u1735067

PR #2425, opened 2022-09-15
Added error handling for `display()` for LocationType and Location.

HanlinMiao:
# What's Changed

Screen Shot 2022-09-15 at 1 13 03 AM

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/models/locations.py
+++ b/nautobot/dcim/models/locations.py
@@ -87,7 +87,17 @@ class LocationType(TreeNode, OrganizationalModel):
 
     @property
     def display(self):
-        # prefix with parent name allows location type name identity if there are the same name exists under different parents.
+        """
+        Prefix with parent name allows location type name identity if there are the same name exists under different parents.
+        self.ancestors() return all the preceding nodes from top down. So if we are looking at node C and its node structure is the following:
+            A
+           /
+          B
+         /
+        C
+        This method will return \"A → B → C\".
+        If self.ancestor() throw an ObjectDoesNotExist Exception during bulk operations, we handle it by only returning the name of the current node.
+        """
         display_str = ""
         try:
             for ancestor in self.ancestors():
@@ -96,7 +106,7 @@ class LocationType(TreeNode, OrganizationalModel):
             display_str += self.name
             return display_str
         except ObjectDoesNotExist:
-            return display_str
+            return self.name
 
 
 @extras_features(
@@ -214,7 +224,17 @@ class Location(TreeNode, StatusModel, PrimaryModel):
 
     @property
     def display(self):
-        # prefix with parent name allows location type name identity if there are the same name exists under different parents.
+        """
+        Prefix with parent name allows location name identity if there are the same name exists under different parents.
+        self.ancestors() return all the preceding nodes from top down. So if we are looking at node C and its node structure is the following:
+            A
+           /
+          B
+         /
+        C
+        This method will return \"A → B → C\".
+        If self.ancestor() throw an ObjectDoesNotExist Exception during bulk operations, we handle it by only returning the name of the current node.
+        """
         display_str = ""
         try:
             for ancestor in self.ancestors():
@@ -223,7 +243,7 @@ class Location(TreeNode, StatusModel, PrimaryModel):
             display_str += self.name
             return display_str
         except ObjectDoesNotExist:
-            return display_str
+            return self.name
 
     def validate_unique(self, exclude=None):
         # Check for a duplicate name on a Location with no parent.
--- a/nautobot/dcim/models/locations.py
+++ b/nautobot/dcim/models/locations.py
@@ -88,15 +88,16 @@ class LocationType(TreeNode, OrganizationalModel):
     @property
     def display(self):
         """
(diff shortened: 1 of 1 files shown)


nautobot/dcim/models/locations.py
```diff
@@ -88,14 +88,15 @@ def clean(self):
...
-        return display_str
+        display_str = ""
+        try:
+            for ancestor in self.ancestors():
+                display_str += ancestor.name
+                display_str += " → "
+            display_str += self.name
+            return display_str
+        except ObjectDoesNotExist:
+            return display_str
```
glennmatthews, line 99: This reads as backwards to me - if we hit an ObjectDoesNotExist on one of the ancestors, we may return a partially-constructed `display_str` that includes some ancestors but not `self.name` (since that's the last thing we add)?

nautobot/dcim/models/locations.py
```diff
@@ -87,15 +87,26 @@ def clean(self):
...
+        """
+        display_str = ""
+        try:
+            for ancestor in self.ancestors():
+                display_str += ancestor.name
+                display_str += " → "
+            display_str += self.name
+            return display_str
+        except ObjectDoesNotExist:
+            return self.name
```
bryanculver, line 109: I still think this is incorrect because it loses potential ancestors that did exist.

What about (Psudocode-ish):

``` py
display_str = ""
member = self
while member is not None:
  try:
    display_str = f"{member.name} → {display_str}"
    member = member.parent
  except ObjectDoesNotExist:
    member = None

return display_str[:len(display_str)-3]
```

smk4664: Or perhaps move the 
Self.name and return display_str to the finally block:
'''
try:
   for ancestor in ancestors():
       display_str += ancestor.name + "->"

except ...

finally:
    return display_str += self.name
'''

nautobot/dcim/models/locations.py
```diff
@@ -213,15 +224,25 @@ def base_site(self):
...
+        """
+        Prefix with parent name allows location name identity if there are the same name exists under different parents.
+        self.ancestors() return all the preceding nodes from top down. So if we are looking at node C and its node structure is the following:
+            A
+           /
+          B
+         /
+        C
+        This method will return \"A → B → C\".
+        If self.ancestor() throws an ObjectDoesNotExist Exception during bulk operations, we handle it by only returning the names of the current node and the existing ancestors."""
```
jathanism, line 236:
```suggestion
        This method will return "A → B → C".
        If `self.ancestors()` throws an `ObjectDoesNotExist` exception during bulk operations, we handle it by only returning the names of the current node and the existing ancestors."""
```

jathanism (changes requested):

nautobot/dcim/models/locations.py
```diff
@@ -87,15 +87,26 @@ def clean(self):
...
+        Prefix with parent name allows location type name identity if there are the same name exists under different parents.
+        self.ancestors() return all the preceding nodes from top down. So if we are looking at node C and its node structure is the following:
+            A
+           /
+          B
+         /
+        C
+        This method will return "A → B → C".
+        If `self.ancestors()` throws an `ObjectDoesNotExist` exception during bulk operations, we handle it by only returning the names of the current node and the existing ancestors.
+        """
```
glennmatthews, line 100: Please wrap to 120 character line length, perhaps:

```suggestion
        """
        Include the parent type names as well in order to provide UI clarity.

        `self.ancestors()` returns all the preceding nodes from the top down.
        So if we are looking at node C and its node structure is the following:
            A
           /
          B
         /
        C
        This method will return "A → B → C".

        Note that `self.ancestors()` may throw an `ObjectDoesNotExist` during bulk-delete operations.
        """
```

nautobot/dcim/models/locations.py
```diff
@@ -213,15 +224,25 @@ def base_site(self):
...
+        Prefix with parent name allows location name identity if there are the same name exists under different parents.
+        self.ancestors() return all the preceding nodes from top down. So if we are looking at node C and its node structure is the following:
+            A
+           /
+          B
+         /
+        C
+        This method will return "A → B → C".
+        If `self.ancestors()` throws an `ObjectDoesNotExist` exception during bulk operations, we handle it by only returning the names of the current node and the existing ancestors."""
```
glennmatthews, line 236: Similarly, perhaps:

```suggestion
        """
        Location name is unique per parent but not globally unique, so include parent information as context.

        `self.ancestors()` returns all the preceding nodes from the top down.
        So if we are looking at node C and its node structure is the following:
            A
           /
          B
         /
        C
        This method will return "A → B → C".
        
        Note that `self.ancestors()` may throw an `ObjectDoesNotExist` during bulk-delete operations.
        """
```

glennmatthews:
I see you marked my docstring comments as resolved, but I don't see any changes to the docstrings?

jathanism (approved):

HanlinMiao:
Shoot sorry I think I did commit your suggestions on GitHub and then my force push might have overriden the commit.

glennmatthews (approved):

Issue #2375, opened 2022-09-06
Must Supply `color` and `type` Filter on Cable List View


bryanculver:
### Environment
* Nautobot version (Docker tag too if applicable): 1.4.2
* Python version: 3.7
* Database platform, version: Postgres
* Middleware(s): None
* Demo: 

### Steps to Reproduce
1. Visit Cable List View (https://demo.nautobot.com/dcim/cables/)
2. Click "Apply" on FilterForm (apply no values to any filter)

### Expected Behavior
List all cables (unfiltered since no values applied).

### Observed Behavior
"Invalid filters were specified:" error. Color and type filters must be specified (none not an option).

<img width="1495" alt="image" src="https://user-images.githubusercontent.com/31187/188693421-8952d622-a769-46fc-add4-1579748ce35a.png">

Seems similar to #2178

PR #2429, opened 2022-09-15
fixed invalid filter error message in CableListView when color and type field is empty

HanlinMiao:
# Closes: #2375 
# What's Changed

Changed color and type filters in CableFilterSet from `MultipleChoiceFilter` to `ChoiceFilter.` `MultipleChoiceFilter` is expecting an empty list therefore it does not take `None` as a valid choice.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/filters.py
+++ b/nautobot/dcim/filters.py
@@ -1540,8 +1540,8 @@ class VirtualChassisFilterSet(NautobotFilterSet):
 
 class CableFilterSet(NautobotFilterSet, StatusModelFilterSetMixin):
     q = SearchFilter(filter_predicates={"label": "icontains"})
-    type = django_filters.MultipleChoiceFilter(choices=CableTypeChoices)
-    color = django_filters.MultipleChoiceFilter(choices=ColorChoices)
+    type = django_filters.ChoiceFilter(choices=CableTypeChoices)
+    color = django_filters.ChoiceFilter(choices=ColorChoices)
     device_id = MultiValueUUIDFilter(method="filter_device", label="Device (ID)")
     device = MultiValueCharFilter(method="filter_device", field_name="device__name", label="Device (name)")
     rack_id = MultiValueUUIDFilter(method="filter_device", field_name="device__rack_id", label="Rack (ID)")
--- a/nautobot/dcim/forms.py
+++ b/nautobot/dcim/forms.py
@@ -3893,7 +3893,7 @@ class CableBulkEditForm(TagsBulkEditFormMixin, StatusModelBulkEditFormMixin, Nau
             raise forms.ValidationError({"length_unit": "Must specify a unit when setting length"})
 
 
-class CableFilterForm(BootstrapMixin, StatusModelFilterFormMixin, forms.Form):
+class CableFilterForm(BootstrapMixin, StatusModelFilterFormMixin):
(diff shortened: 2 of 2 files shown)


nautobot/dcim/forms.py
```diff
@@ -3893,7 +3893,7 @@ def clean(self):
             raise forms.ValidationError({"length_unit": "Must specify a unit when setting length"})
 
 
-class CableFilterForm(BootstrapMixin, StatusModelFilterFormMixin, forms.Form):
+class CableFilterForm(BootstrapMixin, StatusModelFilterFormMixin):
```
HanlinMiao, line 3896: `StatusModelFilterFormMixin` already inherits `forms.Form` so I dont think this is necessary.

glennmatthews: Disagree a bit on this point - because of the way forms.Form works, the mixins do have to inherit from it, but I consider that an implementation detail, rather than something we should assume/rely upon in our usage of these mixins.

changes/2375.fixed
```diff
@@ -0,0 +1 @@
+Fixed must Supply `color` and `type` Filter on Cable List View.
```
glennmatthews, line 1:
```suggestion
Fixed error in Cable list view filtering when `color` and `type` filters are not set.
```

HanlinMiao:
I was playing around with the filterform, apparently region is giving out an invalid filter error message as well. Can someone confirm that on their end? I pushed up the fix for that one as well.
Screen Shot 2022-09-15 at 10 01 24 PM

nautobot/dcim/filters.py
```diff
@@ -1541,13 +1540,17 @@ class Meta:
...
     device_id = MultiValueUUIDFilter(method="filter_device", label="Device (ID)")
     device = MultiValueCharFilter(method="filter_device", field_name="device__name", label="Device (name)")
     rack_id = MultiValueUUIDFilter(method="filter_device", field_name="device__rack_id", label="Rack (ID)")
     rack = MultiValueCharFilter(method="filter_device", field_name="device__rack__name", label="Rack (name)")
     site_id = MultiValueUUIDFilter(method="filter_device", field_name="device__site_id", label="Site (ID)")
     site = MultiValueCharFilter(method="filter_device", field_name="device__site__slug", label="Site (name)")
+    region_id = MultiValueUUIDFilter(method="filter_device", field_name="device__site__region_id", label="Region (ID)")
+    region = MultiValueCharFilter(
+        method="filter_device", field_name="device__site__region__slug", label="Region (name)"
+    )
```
glennmatthews, line 1553: Can we instead just add `region` as a `NaturalKeyOrPKMultipleChoiceFilter`?

HanlinMiao: Im not sure it is possible to use `NaturalKeyOrPKMultipleChoiceFilter`. As you can see from other filters in this filterset, we have to use a custom method `filter_device` to filter the regions for both`termintion_a` and `termination_b`.  In `NaturalKeyOrPKMultipleChoiceFilter` I can only pass in one or the other (`_termination_a_device__site__region`, `_termination_b_device__site__region`) for its `field_name` argument. Unless Im missing something obvious🤔

bryanculver (approved):

Issue #2324, opened 2022-08-27
Reference to Deleted Jobs


itdependsnetworks:
### Environment
* Nautobot version (Docker tag too if applicable): 1.3.10
* Python version: 3.7
* Database platform, version: postgres
* Middleware(s):

### Steps to Reproduce
1. Run a job
2. Delete the job (not the job run instance, but actual job)
3. Go to front page
4. Click on job run instance, and observe a 404

### Expected Behavior

A 2XX page instead of a 4/5XX page

### Observed Behavior

<img width="832" alt="image" src="https://user-images.githubusercontent.com/9260483/187045280-3d81fa20-4932-4272-b29e-50077e58f525.png">

glennmatthews:
Appears to be a cacheops issue. The `JobResult.job_model` is automatically set to `None` after deleting the Job, but cacheops is making Django think that it is still set. Running `cacheops.invalidate_all()` makes the error go away.

PR #2430, opened 2022-09-15
Fixed errors encountered when a job model is deleted while a job is running

gsnider2195:
# Closes: #2324
# What's Changed

- Fixed JobResult page 500 error described in #2324 
- Added exception handling to allow the job result to be saved when a job is deleted while the job is running so the job result isn't stuck in a "running" state

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/changes/2324.fixed
+++ b/changes/2324.fixed
@@ -1,0 +1,1 @@
+Fixed errors encountered when a job model is deleted while a job is running.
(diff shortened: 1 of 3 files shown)


glennmatthews:
It's not clear to me how this changeset addresses #2324 - if I understand the issue right, the problem is not when deleting a job while the job is running, it's in trying to view the JobResult of a Job that no longer exists?

gsnider2195:
That's correct it's a 500 error encountered when running cacheops and viewing the JobResult when a job has been deleted. The 1 line change to change the queryset in the JobResult view `queryset = JobResult.objects.prefetch_related("job_model", "obj_type", "user")` fixed #2324. The other change was another  annoying side effect I noticed during testing and fixed.

glennmatthews (approved):

PR #2438, opened 2022-09-17
Resolves #2437 - Catch CSRF Failure when DEBUG=False

### LINKED ISSUE #2437: Catch CSRF Failure when DEBUG=False
### As ...

Patti - Platform Admin

### I want ...

the 403 CSRF failure to not render the default Django failure html when DEBUG is set to False

### So that ...

When errors occur in Nautobot a user remains in a view that has the same look and feel as nautobot and not a generic view without styling.

- When form is submitted without CSRF token AND DEBUG is set to False a stylized view is returned with the error.
- When form is submitted and fails CSRF validation AND DEBUG is set to False a stylized view is returned with the error.
- When form is submitted without CSRF token AND DEBUG is set to True the default Django view is returned.
- When form is submitted and fails CSRF validation AND DEBUG is set to True the default Django view is returned.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation

### Database Changes

None

### External Dependencies

None

whitej6:
# Closes: #2437 
# What's Changed

- Changed settings.py to use new view for CSRF failure
- Added template for failure
- Added view to render template when DEBUG=False OR return upstream Django view when DEBUG=True
image

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/2437.added
+++ b/changes/2437.added
@@ -1,0 +1,1 @@
+Catch 403 CSRF errors and return stylized nautobot view.
--- a/changes/2437.added
+++ b/changes/2437.added
@@ -1,1 +1,1 @@
-Catch 403 CSRF errors and return stylized nautobot view.
+Added Nautobot-themed error page for handling 403 CSRF errors.
--- a/nautobot/core/templates/403_csrf_failure.html
+++ b/nautobot/core/templates/403_csrf_failure.html
@@ -1,0 +1,9 @@
+{% extends '40x.html' %}
+
+{% block title %}Invalid Form Submission{% endblock %}
+
+{% block icon %}<i class="mdi mdi-alert"></i>{% endblock %}
+
+{% block message %}
+    CSRF failure has occurred.
+{% endblock %}
--- a/nautobot/core/templates/403_csrf_failure.html
+++ b/nautobot/core/templates/403_csrf_failure.html
@@ -5,5 +5,19 @@
 {% block icon %}<i class="mdi mdi-alert"></i>{% endblock %}
 
 {% block message %}
-    CSRF failure has occurred.
+    CSRF failure has occurred.<br/>
+    <p>
+        There was a problem with your request. Please contact an administrator.
+    </p>
+    <p>
+        The reason for failure is provided below:
+    </p>
+<pre><strong>{{ reason }}</strong><br/>
+
+Python version: {{ python_version }}
+Nautobot version: {{ nautobot_version }}
+</pre>
+    <p>
+        If further assistance is required, please join the <b>#nautobot</b> channel on <a href="https://slack.networktocode.com/">Network to Code's Slack community</a> and post your question.
+    </p>
 {% endblock %}
--- a/nautobot/core/views/__init__.py
+++ b/nautobot/core/views/__init__.py
@@ -207,7 +207,7 @@ def csrf_failure(request, reason="", template_name="403_csrf_failure.html"):
     if settings.DEBUG:
         return _csrf_failure(request, reason=reason)
     t = loader.get_template(template_name)
-    context = {"settings": settings}
+    context = {"settings": settings, "reason": reason}
     return HttpResponseForbidden(t.render(context), content_type="text/html")
 
 
--- a/nautobot/core/views/__init__.py
+++ b/nautobot/core/views/__init__.py
@@ -4,13 +4,14 @@ import sys
 
 from django.conf import settings
 from django.contrib.auth.mixins import AccessMixin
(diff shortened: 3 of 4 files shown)


whitej6:
One question, would we want this to be high level that a failure occurred OR add the failure reason to the template?

bryanculver:
Could we instead direct the user back one page in history instead of the homepage?

whitej6:
This doesn't redirect to the homepage, it just returns a rendered error for the same URL which is what we do for catching 403/404/500 currently. 

In all likelihood this is a scenario that should not arise but was discovered when a tool for a customer first would make an API call but the URI variable was not instantiated so the POST went to `/` instead of `/api/dcim/foo/bar/`. The returned HTML had me confused on which item in the stack was the culprit since I had assumed CSRF 403 failures were caught just like typical 403. This was made more confusing due to a Django app that was installed which injected a lot of JS which is not part of Nautobot. The word Nautobot I don't _think_ was in the HTML response either since it was the generic 403 CSRF response.

Only time I have seen this in the wild has been when doing view development and creating custom HTML forms without adding the token to the form.

nautobot/core/views/__init__.py
```diff
@@ -196,6 +197,20 @@ def server_error(request, template_name=ERROR_500_TEMPLATE_NAME):
...
+
+    If Nautobot is set to DEBUG the default view for csrf_failure.
+    `403_csrf_failure.html` template name is used over `403_csrf.html` to account for
+    additional context that is required to render the inherited templates.
+    """
+    if settings.DEBUG:
+        return _csrf_failure(request, reason=reason)
+    t = loader.get_template(template_name)
+    context = {"settings": settings}
+    return HttpResponseForbidden(t.render(context), content_type="text/html")
```
jathanism, line 211: `csrf_failure` accepts a `template_name` argument:

```suggestion
    return _csrf_failure(request, reason, template_name)
```

nautobot/core/views/__init__.py
```diff
@@ -4,13 +4,14 @@
 
 from django.conf import settings
 from django.contrib.auth.mixins import AccessMixin
-from django.http import HttpResponseServerError, JsonResponse
+from django.http import HttpResponseServerError, JsonResponse, HttpResponseForbidden
```
jathanism, line 7:
```suggestion
from django.http import HttpResponseServerError, JsonResponse
```

nautobot/core/templates/403_csrf_failure.html
```diff
@@ -0,0 +1,9 @@
+{% extends '40x.html' %}
+
+{% block title %}Invalid Form Submission{% endblock %}
+
+{% block icon %}<i class="mdi mdi-alert"></i>{% endblock %}
+
+{% block message %}
+    CSRF failure has occurred.
```
jathanism, line 8: Is there not a message that comes from the exception that is raised that is possibly more descriptive here?

Also could we not overload  the `more` context variable used in the [default template](https://github.com/django/django/blob/main/django/views/csrf.py)? vs. introducing another template to manage?

jathanism (changes requested):

nautobot/core/views/__init__.py
whitej6: But does not have our context, which we need settings for the template to render. Initially I just added the template which should have worked if it had access to the context. The upstream view does not allow for outside context.

nautobot/core/views/__init__.py
whitej6: See above response.

nautobot/core/templates/403_csrf_failure.html
whitej6: See above responses and my first comment. 

Happy to add reason to the output.

jathanism: I think that would be a nice middle ground. to say CSRF failed and at least attempt to explain why unless we provide a generic reason like "Refresh the page and try again." or something to that effect.

whitej6: Added in reason and generic boilerplate from the 500.html template

changes/2437.added
```diff
@@ -0,0 +1 @@
+Catch 403 CSRF errors and return stylized nautobot view.
```
glennmatthews, line 1:
```suggestion
Added Nautobot-themed error page for handling 403 CSRF errors.
```

glennmatthews (approved):

jathanism (approved):

Issue #2447, opened 2022-09-19
Dynamic group membership should be rendered as a separate view


lampwins:
### Environment
* Nautobot version (Docker tag too if applicable): 1.4.2
* Python version: 3.9
* Database platform, version: all
* Middleware(s): N/A

### Steps to Reproduce
1. Create a large number of dynamic groups (in this case ~1400 via a perfectly valid use case with golden config settings)
2. Attempt to load the device detail view or any of it's direct tabs
3. Have a bad day

### Expected Behavior
The view should load and I am able to view device details.

### Observed Behavior
Because membership in dynamic groups is evaluated at the time of rendering the detail view, the page times out.

PR #2448, opened 2022-09-19
Move Dynamic Group Model Detail Tab to its Own View.

bryanculver:
# Closes: #2447
# What's Changed
- Created `ObjectDynamicGroupsView` to generically list dynamic group memberships (reverse) of the object in question
    - This uses the same table view as in the Dynamic Groups list view
    - Created `DeviceDynamicGroupsView` to override base template (don't know if this is needed but kept for consistency)
- Add test case to ensure it loads the results
- Add `[content-type]/[slug/pk]/dynamic-groups/` routes to all supporting models
- Added `get_dynamic_groups_url` method to `DynamicGroupMixin` to provide quick access to the dynamic groups url from instance
- Removed old dynamic group tab code

image

# TODO

- [x] Unit, Integration Tests


nautobot/extras/models/mixins.py
```diff
@@ -28,6 +28,23 @@ def dynamic_groups(self):
...
+        # Iterate the pk-like fields and try to get a URL, or return None.
+        fields = ["pk", "slug"]
+        for field in fields:
+            if not hasattr(self, field):
+                continue
+
+            try:
+                return reverse(route, kwargs={field: getattr(self, field)})
+            except NoReverseMatch:
+                continue
```
jathanism, line 44: There's that pattern again. We should probably factor this one into a helper at some point.

nautobot/ipam/urls.py
```diff
@@ -187,6 +187,12 @@
         name="prefix_changelog",
         kwargs={"model": Prefix},
     ),
+    path(
+        "prefixes/<uuid:pk>/dynamic-groups/",
+        ObjectDynamicGroupsView.as_view(),
+        name="prefix_dynamicgroups",
+        kwargs={"model": Prefix},
+    ),
```
jathanism, line 195: Let's make sure we remember to factor this into viewset pattern to be automagic when an object has `dynamic_groups` in its `extras_features`.

nautobot/core/templates/generic/object_retrieve.html
```diff
@@ -85,12 +85,12 @@ <h1>{% block title %}{{ object }}{% endblock %}</h1>
...
             {% endif %}
         {% endif %}
-        {% if perms.extras.view_dynamicgroup and object.dynamic_groups %}
-            <li role="presentation"{% if request.GET.tab == 'dynamic-groups' %} class="active"{% endif %}>
-                <a href="{{ object.get_absolute_url }}#dynamic-groups" onclick="switch_tab(this.href)" aria-controls="advanced" role="tab" data-toggle="tab">
-                    Dynamic Groups
-                </a>
-            </li>
+        {% if perms.extras.view_dynamicgroup and object.get_dynamic_groups_url %}
+            {% if active_tab != 'dynamic-groups' and object.get_dynamic_groups_url or active_tab == 'dynamic-groups' %}
```
jathanism, line 89: Is this logic accurate? It reads funky to me.

nautobot/core/templates/inc/dynamic_groups_panel.html
```diff
@@ -1,15 +0,0 @@
-{% load helpers %}
```
glennmatthews, line 1: Might need to leave this file, just in case we have plugin templates that are expecting to use it?

nautobot/extras/views.py
```diff
@@ -701,6 +701,56 @@ class DynamicGroupBulkDeleteView(generic.BulkDeleteView):
     table = tables.DynamicGroupTable
 
 
+class ObjectDynamicGroupsView(View):
```
glennmatthews, line 704: Should this be `(generic.ObjectView)` instead?

What permissions are enforced by this view? I'd think you'd at least need `view_dynamicgroup` and `view_<object>` permission - are those checked?

nautobot/core/templates/generic/object_retrieve.html
bryanculver: 🤷‍♂️ Mimics what we are doing for changelog and notes tabs around it.

jathanism (approved):

PR #2458, opened 2022-09-21
Added performance testing

HanlinMiao:
# Closes: #2227
# What's Changed

1. Added `GENERATE_PERFORMANCE_REPORT` and `TEST_ALWAYS_GENERATE_BASELINE_REPORT` to `settings.py` and `nautobot_config.py`
2. `GENERATE_PERFORMANCE_REPORT` and `TEST_ALWAYS_GENERATE_BASELINE_REPORT` always default to `False` in settings.py to avoid breaking plugin testing. However, we override `GENERATE_PERFORMANCE_REPORT` to `True` in `nautobot_config.py` as we always need performance-testing for core development purposes.
3. Added `TEST_PERFORMANCE_BASELINE_FILE` = "`nautobot/core/tests/performance_baselines.yml`" in settings.py as baselines to run performance tests against for now. **(Need to get a consistent baseline before making performance tests gating)**
4. Made `django-slowtests` a dev dependency
5. Added `invoke performance-test` to tasks.py and added try..except.. block to catch errors in configuration settings (i.e. trying to run performance-tests without setting `GENERATE_PERFORMANCE_REPORT` to True.
6. Added `performance-test` CI steps to run in python version 3.7 and 3.10 and postgres in parallel with integration test. (not gating)
7. Added `NautobotPerformanceTestRunner` to run performance-specific unittests.

# Caveats

1. Performance baselines are derived from unittests on my local instance.
2. CI test timings are very inconsistent across python versions and runs and I have to loosen my failure criteria so the performance tests do not fail because of some small variations between runs.
3. The criteria right now is that if the tests took 1.5 time longer than the corresponding baselines and 3 seconds longer, show the test to the user.

# TODO

- [x] Method for tracking performance baselines
- [x] Writing Appropriate Performance Tests
- [x] Documentations


bryanculver:
Leaving this here as it was discussed in another channel:

- Can we get a decorator/tag/label to add to tests to mark them as specifically performance targeted tests?
- Step added to Github CI workflow to run this on PR with default Python version & Postgres

nautobot/core/tests/__init__.py
```diff
@@ -0,0 +1,9026 @@
+PERFORMANCE_BASELINES = {
```
glennmatthews, line 1: Would it make more sense to store this as a separate YAML file rather than inline as code?

.github/workflows/ci_pullrequest.yml
```diff
@@ -275,9 +317,7 @@ jobs:
       - name: "Run coverage"
         run: "poetry run invoke unittest-coverage"
     needs:
-      - "pylint"
-      - "tests-postgres"
-      - "tests-mysql"
+      - "performance-test"
```
glennmatthews, line 320: I'd prefer to run integration and performance tests in parallel so that we're not adding another serial step to our already lengthy CI process.

nautobot/core/settings.py
```diff
@@ -116,6 +116,8 @@
 # Test runner that is aware of our use of "integration" tags and only runs
 # integration tests if explicitly passed in with `nautobot-server test --tag integration`.
 TEST_RUNNER = "nautobot.core.tests.runner.NautobotTestRunner"
+NUM_SLOW_TESTS = None
```
glennmatthews, line 119: Is this a `django-slowtests` setting? (If so, can you add some comments?) Or can we rename it to something less generic?

nautobot/core/settings.py
```diff
@@ -116,6 +116,8 @@
...
+ALWAYS_GENERATE_SLOW_REPORT = False
```
glennmatthews, line 120: Do we need this setting at all? If so, can we add a `TEST_` prefix to it to make its scope clearer?

nautobot/core/tests/runner.py
```diff
@@ -30,3 +40,87 @@ def __init__(self, **kwargs):
...
+    def generate_report(self, test_results, result):
+        """
+        Generate Performance Report consists of unittests that are significantly slower than baseline.
+        """
+        test_result_count = len(test_results)
+
+        # Add `--generate_report` to the end of `invoke` commands to generate a report.json file consist of the performance tests result
+        if self.report_path:
+            data = {
+                "tests": [{"name": func_name, "execution_time": float(timing)} for func_name, timing in test_results],
```
glennmatthews, line 53: Is there any intrinsic ordering to `test_results`? If not, can we sort it in some fashion? Thinking about keeping the diffs readable as and when we need to regenerate this file.

(edit) Ah I see, currently it's sorted by execution time? I think sorting by name might be better?

nautobot/core/tests/runner.py
```diff
@@ -30,3 +40,87 @@ def __init__(self, **kwargs):
...
+        # Print the results in the CLI.
+        else:
+            if test_result_count:
+                print(f"\n{test_result_count} abnormally slower tests:")
+                for func_name, timing in test_results:
+                    time = float(timing)
+                    baseline = float(self.baselines[func_name])
+                    print(f"{time:.4f}s {func_name} is significantly slower than the baseline {baseline:.4f}s")
+                # Make the test fail if there are tests significantly slower
+                assert test_result_count == 0, "Performance Tests failed due to significantly slower tests"
```
glennmatthews, line 69: This assertion will always fail since it's inside the `if test_result_count` block - intentional?

nautobot/core/tests/runner.py
```diff
@@ -30,3 +40,87 @@ def __init__(self, **kwargs):
...
+        Generate Performance Report consists of unittests that are significantly slower than baseline.
+        """
+        test_result_count = len(test_results)
+
+        # Add `--generate_report` to the end of `invoke` commands to generate a report.json file consist of the performance tests result
+        if self.report_path:
+            data = {
+                "tests": [{"name": func_name, "execution_time": float(timing)} for func_name, timing in test_results],
+                "nb_tests": result.testsRun,
+                "nb_failed": len(result.errors + result.failures),
```
glennmatthews, line 55: I don't see a great need to abbreviate `nb` - perhaps?

```suggestion
                "test_count": result.testsRun,
                "failed_count": len(result.errors + result.failures),
```

nautobot/core/tests/runner.py
```diff
@@ -30,3 +40,87 @@ def __init__(self, **kwargs):
...
+            if not test_results:
+                print("\nNo tests signficantly slower than baseline. Success!")
+
+    def get_baselines(self):
+        """Load the performance_baselines.yml file for result comparison."""
+        baselines = {}
+        input_file = "nautobot/core/tests/performance_baselines.yml"
+
+        with open(input_file) as f:
+            data = yaml.load(f, Loader=SafeLoader)
```
glennmatthews, line 80: Is this preferable to `yaml.safe_load(f)`?

nautobot/core/tests/runner.py
```diff
@@ -30,3 +40,87 @@ def __init__(self, **kwargs):
...
+        return baselines
+
+    def suite_result(self, suite, result):
+        """Compile the performance test results"""
+        return_value = super(DiscoverSlowestTestsRunner, self).suite_result(suite, result)
+        self.baselines = self.get_baselines()
+
+        # You can set `ALWAYS_GENERATE_SLOW_REPORT` to True or add `--report` to `invoke` commands to generate report.
+        # e.g. `invoke unittest --report`
+        should_generate_report = getattr(settings, "ALWAYS_GENERATE_SLOW_REPORT", True) or self.should_generate_report
```
glennmatthews, line 92:
```suggestion
        should_generate_report = getattr(settings, "ALWAYS_GENERATE_SLOW_REPORT", self.should_generate_report)
```

nautobot/core/tests/runner.py
```diff
@@ -30,3 +40,87 @@ def __init__(self, **kwargs):
...
+        if not should_generate_report:
+            self.remove_timing_tmp_files()
```
glennmatthews, line 94: Where's this method defined?

nautobot/core/tests/runner.py
```diff
@@ -30,3 +40,87 @@ def __init__(self, **kwargs):
...
+                if self.report_path:
+                    test_results.append(entry)
+                else:
+                    # If the test completed under 1.5 times the baseline or the difference between the result and the baseline is less than 5 seconds,
+                    # dont show the test to the user.
+
+                    # baseline duration in milliseconds
+                    baseline_ms = self.baselines.get(entry[0], 0) * 1000
+                    # Arbitrary criteria to not make performance test fail easily
+                    if result_time_ms <= baseline_ms * 1.5 or result_time_ms - baseline_ms <= 5000:
```
glennmatthews, line 120: Not sure about that 5000 check - so if we go from 0.2 seconds to 5.0 seconds (4800 delta) for a given test, that's OK?

nautobot/core/tests/runner.py
```diff
@@ -13,6 +17,12 @@ class NautobotTestRunner(DiscoverRunner):
     inheriting from that class do not need to be explicitly tagged.
 
     Only integration tests that DO NOT inherit from `SeleniumTestCase` will need to be explicitly tagged.
+
+    `ModelViewSetTestCase` is tagged with `performance` to test the time it will take to retrieve, list, create, bulk_create,
```
glennmatthews, line 21:
```suggestion
    `ModelViewTestCase` is tagged with `performance` to test the time it will take to retrieve, list, create, bulk_create,
```

pyproject.toml
```diff
@@ -134,6 +134,7 @@ social-auth-core = {version = "~4.3.0", optional = true, extras = ["openidconnec
 social-auth-app-django = "~5.0.0"
 # Rendering of SVG images (for rack elevations, etc.)
 svgwrite = "~1.4.2"
+django-slowtests = "^1.1.1"
```
glennmatthews, line 137: Please move this up into alphabetical ordering and add a comment.

nautobot/core/tests/runner.py
HanlinMiao: Yes it is intentional. This will make the performance tests fail if any tests failed to meet the criteria derived from its baseline.

nautobot/core/tests/runner.py
HanlinMiao: If `ALWAYS_GENERATE_SLOW_REPORT` exists and is set to `False`, then wouldn't `self.should_generate_report` matter in the suggestion? That eliminates the overridability of the statement I think. Let me know if I'm wrong.

nautobot/core/tests/runner.py
HanlinMiao: This method is defined in `DiscoverSlowestTestsRunner` class inherited by `NautobotTestRunner`.
See https://github.com/realpython/django-slow-tests/blob/2c5673ca5b6eb093a908e4e28143d889a1893113/django_slowtests/testrunner.py#L219

nautobot/core/tests/runner.py
HanlinMiao: Yeah Im not too sure about what kind of criteria for the performance test is considered appropriate. Test times are still pretty fickle across python versions in the CI. so I loosened the restrictions just to make all the tests pass.

pyproject.toml
```diff
@@ -70,6 +70,8 @@ django-prometheus = "~2.2.0"
 django-redis = "~5.2.0"
 # RQ (Redis Queueing) for background handling of webhooks, jobs, etc.
 django-rq = "~2.5.1"
+# DiscoverSlowestTestsRunner for running CI performance tests and benchmarking.
+django-slowtests = "^1.1.1"
```
jathanism, line 74: Does this need to be a core dependency or can we get away with it being a dev dependency?

glennmatthews: Great point! Same potential issue here as #2536.

nautobot/core/tests/runner.py
glennmatthews: You're right. :-)

pyproject.toml
```diff
@@ -155,6 +155,8 @@ black = "~22.8.0"
 coverage = "~6.4.1"
 # Tool for debugging Django
 django-debug-toolbar = "~3.7.0"
+# DiscoverSlowestTestsRunner for running CI performance tests and benchmarking.
+django-slowtests = "^1.1.1"
```
glennmatthews, line 159: In general I'm OK with this being a dev dependency, **but** we need to think about the case of plugins using our test runner to run their own tests. (similar to #2536). Plugins shouldn't have to know about Nautobot's dev-only dependencies in order to run their own unit tests; it should be possible to run `nautobot-server test <plugin>` (or `coverage run --module nautobot.core.cli test <plugin>`, equivalently) with only Nautobot's core dependencies plus the plugin's own dev dependencies installed.

I'm thinking for #2536 I'll probably address this with a settings flag like `TEST_USE_FACTORIES` and appropriate conditional logic around imports and calls to ensure that the factory logic is only called when opted into (i.e. always by core, conditionally by plugins if and only if they decide to opt into and support this feature).

nautobot/core/tests/runner.py
```diff
@@ -69,3 +79,88 @@ def teardown_databases(self, old_config, **kwargs):
...
+                # Make the test fail if there are tests significantly slower
+                assert test_result_count == 0, "Performance Tests failed due to significantly slower tests"
+
+            if not test_results:
+                print("\nNo tests signficantly slower than baseline. Success!")
+
+    def get_baselines(self):
+        """Load the performance_baselines.yml file for result comparison."""
+        baselines = {}
+        input_file = "nautobot/core/tests/performance_baselines.yml"
```
glennmatthews, line 116: This should probably be configurable, as in the future plugins may want to generate and check their own performance baselines using this tooling.

pyproject.toml
HanlinMiao: Just to clarify, this conditional logic will leave `factory-boy` and `django-slowtests` as dev dependencies and only when the plugin decide to opt in to have those test features using a settings flag, they will become available as a plugin own dev dependencies? 
Can we make them as plugin dev dependencies to begin with or is it not worth it?

glennmatthews: If the plugin wants to use these test features, they will have to add those packages as their own dev dependencies. Plugins have no direct visibility into Nautobot's own dev dependencies as those are a Poetry feature, not a Python package/wheel feature.

We could, I suppose, add these as optional primary dependencies under an extra (so you could do `pip install nautobot[test-infra]`) rather than as dev dependencies, which would at least make it easier for plugins to get a consistent version for these libraries... the more I think about it the more I like that idea. But even if we go that route, plugins still need to be able to continue to run their existing tests as-is; these add-ons still need to be opt-in.

What I'm trying to avoid is putting plugins in a position of "Your tests are going to blow up and refuse to run under Nautobot 1.5 unless/until you update your project."

nautobot/core/settings.py
```diff
@@ -117,6 +117,15 @@
...
 TEST_RUNNER = "nautobot.core.tests.runner.NautobotTestRunner"
 
+#
+# django-slowtests
+#
+
+# django-slowtests setting and it is set so that only adding `--report` at the end of the `invoke` command
+# will output the performance test results. e.g. `invoke unittest --report`.
+TEST_ALWAYS_GENERATE_SLOW_REPORT = False
+TEST_PERFORMANCE_BASELINE_FILE = "nautobot/core/tests/performance_baselines.yml"
```
glennmatthews, line 127: These should be documented somewhere - normally these'd belong in `docs/configuration/optional-settings.md`, but since these are dev/test specific maybe `docs/development/testing.md` would be more appropriate.

pyproject.toml
glennmatthews: Thinking about it more, the `extras` approach probably isn't feasible, as IIRC Poetry doesn't let you have different extras dependencies for the same package under different sections, i.e. a plugin *could not* do:

```toml
[tool.poetry.dependencies]
nautobot = "*"

[tool.poetry.dev-dependencies]
nautobot = {"version": "*", extras=["test-infra"]}
```

See #2570 for the approach I ended up going with for factory-boy.

nautobot/core/settings.py
jathanism: I would prefer now that we have sections for "Administratively Configurable Settings" and "Django Configuration Settings", that we just add these as optional settings, and then have clear references to them in the testing docs.

This is s similar pattern to how we have it in our installation docs, where the settings are explained in the context they are needed, but documented centrally in the configuration required/optional settings section.

HanlinMiao:
I vaguely remeber the team agreeing on removing performance-tests from CI until we can track baseline performance consistently. Is that still good?

nautobot/core/tests/runner.py
```diff
@@ -52,3 +51,118 @@ def teardown_databases(self, old_config, **kwargs):
...
+
+# Use django_slowtests only when GENERATE_PERFORMANCE_REPORT flag is set to true
+if settings.GENERATE_PERFORMANCE_REPORT:
+    import json
+    import yaml
+
+    try:
+        from django_slowtests.testrunner import DiscoverSlowestTestsRunner
+
+        print("Using NautobotPerformanceTestRunner to run tests ...")
```
bryanculver, line 64: Should this be more related to `DiscoverSlowestTestsRunner`?

tasks.py
```diff
@@ -589,6 +589,8 @@ def check_schema(context, api_version=None):
         "verbose": "Enable verbose test output.",
         "append": "Append coverage data to .coverage, otherwise it starts clean each time.",
         "skip_docs_build": "Skip (re)build of documentation before running the test.",
+        "report": "Generate Performance Testing report in the terminal. Has to set GENERATE_PERFORMANCE_REPORT=True in settings.py",
+        "generate_report": "Generate a new performance testing report to report.json. Has to set GENERATE_PERFORMANCE_REPORT=True in settings.py",
```
bryanculver, line 593: "Report" to me is more synonymous with coverage report than performance. And the report json would typically be referred to as snapshot.

So maybe something more like:
```suggestion
        "performance_report": "Generate Performance Testing report in the terminal. Has to set GENERATE_PERFORMANCE_REPORT=True in settings.py",
        "performance_snapshot": "Generate a new performance testing report to report.json. Has to set GENERATE_PERFORMANCE_REPORT=True in settings.py",
```

tasks.py
```diff
@@ -603,6 +605,8 @@ def unittest(
     verbose=False,
     append=False,
     skip_docs_build=False,
+    report=False,
+    generate_report=False,
```
bryanculver, line 609: Same as above.

tasks.py
```diff
@@ -682,6 +701,62 @@ def integration_test(
...
+    label="nautobot",
+    failfast=False,
+    buffer=True,
+    tag=None,
+    exclude_tag=None,
+    verbose=False,
+    append=False,
+    skip_docs_build=False,
+    report=True,
+    generate_report=False,
```
bryanculver, line 737: This seems silly. I like `invoke performance_test` but allowing to run it without the performance tests?

bryanculver (commented):
I think some of the naming conventions could be tidied up but love the approach 👍. Keep going!

nautobot/core/tests/runner.py
```diff
@@ -52,3 +51,133 @@ def teardown_databases(self, old_config, **kwargs):
...
+    except ImportError:
+        print("Unable to use NautobotPerformanceTestRunner to run tests. Is the 'django_slowtests' package installed?")
+
+    class NautobotPerformanceTestRunner(NautobotTestRunner, DiscoverSlowestTestsRunner):
```
jathanism, line 67: Is it necessary to define this entire class conditionally? Don't we still need to overload the `TEST_RUNNER` setting (or pass `--testrunner` to `nautobot-server test`) to activate it? Unless I'm misunderstanding?

HanlinMiao: I confuse myself sometimes when Im implementing something but you are absolutely right. The try and except block should cover it.

nautobot/docs/development/testing.md
```diff
@@ -61,15 +61,38 @@ Nautobot's custom [test runner](https://docs.djangoproject.com/en/3.2/topics/tes
...
+Performance baseline for test_bulk_create_objects (nautobot.circuits.tests.test_api.ProviderTest) is not available. Test took 0.2092s to run
+Performance baseline for test_bulk_delete_objects (nautobot.circuits.tests.test_api.CircuitTerminationTest) is not available. Test took 0.1168s to run
+Performance baseline for test_bulk_delete_objects (nautobot.circuits.tests.test_api.CircuitTest) is not available. Test took 0.2762s to run
+Performance baseline for test_bulk_delete_objects (nautobot.circuits.tests.test_api.CircuitTypeTest) is not available. Test took 0.0663s to run
+Performance baseline for test_bulk_delete_objects (nautobot.circuits.tests.test_api.ProviderNetworkTest) is not available. Test took 0.0875s to run
+...
+```
+
+!!! info
+    If performance baselines for tests are available and the time it take to run tests are siginificantly slower than baselines:
```
jathanism, line 92: Good step forward, and I see that you moved the all of the `TEST_GENERATE_PERFORMANCE_REPORT` docs here, but what seems to be missing are any references to passing `--testrunner` or setting the `TEST_RUNNER` in your config?

Perhaps that would be best in the `development/testing.md` file somewhere?

glennmatthews (approved):

changes/2227.added
```diff
@@ -1 +1,2 @@
 Added generating performance report options to 'invoke unittest'.
+Added `invoke performance-test` to task.py.
```
jathanism, line 2:
```suggestion
Added `invoke performance-test` to `tasks.py`.
```

jathanism (changes requested):
Minor last minute docs nits. Can you also replace all instances of "unittests" in the docs with "unit tests" when it's not being explicitly referenced as a command (e.g. `invoke unittest`, et al. can stay), since that's the human-readable way to express that! Thank you!

jathanism (approved):

Issue #1292, opened 2022-01-21
Offer simple ORM interface for scheduling a job


lampwins:
### Environment
* Python version: 3.8.11
* Nautobot version: 1.2.4

### Proposed Functionality
As P.D. the plugin developer, I want a simple interface to schedule a job execution via the ORM so that I can easily perform this action within the business logic of my plugin.

### Use Case
Today it takes ~30 lines of code to create a `ScheduledJob` instance for a nautobot job. Example: https://github.com/nautobot/nautobot/blob/0e491e19d137b41a685c4f97953d2dbe31295862/nautobot/extras/views.py#L820-L857

In the spirit of `JobResult.enqueue_job()`, it should not be this hard to schedule a job for execution. We can likely consolidate the `JobResult.enqueue_job()` method into this too.

### Database Changes
None

### External Dependencies
None

tbotnz:
@lampwins/whoever resolves this issue, would you also be able to consider a pattern where JobResult.enqueue_job() is triggered by django signals.
The current arg (user) on JobResult.enqueue_job() makes this pattern (signals triggering jobs) problematic to implement

lampwins:
Converting this to a spike the goal of defining what the interface for a combined enqueue/scheduling method would look like.

gsnider2195:
> @lampwins/whoever resolves this issue, would you also be able to consider a pattern where JobResult.enqueue_job() is triggered by django signals. The current arg (user) on JobResult.enqueue_job() makes this pattern (signals triggering jobs) problematic to implement

That should be covered by the work for #1631

bryanculver:
Design complete. See #2474. To be implemented via #2535

PR #2474, opened 2022-09-22
Spike: Offer simple ORM interface for scheduling a job

gsnider2195:
# Closes: #1292
# Spike Proposed Design

The minimum arguments required for scheduling a job are `request`, `data`, `commit`, `schedule_type`, `start_time`, `schedule_name`, and `task_queue`. We can derive a default value for `schedule_name` from the other arguments, `start_time` can be optional if using the "immediately" schedule type and `task_queue` can be optional if you want to use the default queue. 

This design uses a method on the `Job` model called `schedule_job` for the public API and private methods called `_run_job` and `_schedule_job` to either run the job immediately with `JobResult.enqueue_job` or create a `ScheduledJob` object. This _should_ be able to replace `JobResult.enqueue_job` for immediate jobs by using the "immediately" schedule type. The spec for the `schedule_job` method is:

```py
def schedule_job(self, request, data, commit, schedule_type, start_time=None, schedule_name=None, task_queue=None):
```

Some of the validation in `schedule_job` is duplicated in the api and ui run job views and could probably be collapsed as part of this design.

Example immediate job run
```py
from nautobot.extras.context_managers import web_request_context
from nautobot.extras.choices import JobExecutionType

j=Job.objects.get(job_class_name='ExampleLoggingJob')
data = {"interval": 1}
commit = True
user = User.objects.first()
with web_request_context(user) as request:
    j.schedule_job(request, data, commit, "immediately")
```

Example scheduled future job
```py
from datetime import timedelta
from django.utils import timezone
from nautobot.extras.context_managers import web_request_context
from nautobot.extras.choices import JobExecutionType

j=Job.objects.get(job_class_name='ExampleLoggingJob')
data = {"interval": 1}
commit = True
user = User.objects.first()
with web_request_context(user) as request:
    start_time = timezone.now() + timedelta(seconds=20)
    j.schedule_job(request, data, commit, JobExecutionType.TYPE_FUTURE, start_time=start_time)
```

# TODO

This is a proof of concept design. In order to implement the following would need to be completed:

- Documentation
- Testing

### DIFF
--- a/nautobot/extras/models/jobs.py
+++ b/nautobot/extras/models/jobs.py
@@ -42,10 +42,13 @@ from nautobot.extras.utils import (
     FeatureQuery,
     extras_features,
     get_job_content_type,
+    get_worker_count,
     jobs_in_directory,
 )
 from nautobot.utilities.fields import JSONArrayField
+from nautobot.utilities.exceptions import CeleryWorkerNotRunningException
 from nautobot.utilities.logging import sanitize
+from nautobot.utilities.utils import copy_safe_request
 
 from .customfields import CustomFieldModel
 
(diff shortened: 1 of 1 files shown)


bryanculver:
@lampwins Do you think the interface above is an acceptable improvement against the current way to schedule a job? Would you like to have the `web_request_context` abstracted away into the `.schedule_job()` function, requiring a user to be passed in always?

lampwins:
This looks good to me.

> Would you like to have the `web_request_context` abstracted away into the `.schedule_job()` function, requiring a user to be passed in always?

I interpret the requirement to be a `request` object, meaning if you already have one, you don't need `web_request_context`, so the current implementation is fine because it offers the required flexibility.

Issue #2283, opened 2022-08-22
Test Fixtures: Aggregate, Prefix, IP address, VRF, VLAN


bryanculver:

gsnider2195:
@glennmatthews I'm going to take a look at IP address and prefix

glennmatthews:
Implemented through #2479, #2514, and #2579.

Issue #2226, opened 2022-08-17
Test Fixtures: Setup Realistic Data for Test Environment


bryanculver:

itdependsnetworks:
between factory boy and data generation, I would say this is complete.

PR #2479, opened 2022-09-23
Factory-boy fixtures: Tenant, TenantGroup, RIR, Aggregate

glennmatthews:
# Partial progress on: #2283 (addresses RIR and Aggregate but not other IPAM models)
# What's Changed

- Add `factory-boy` as a dev dependency
- Define factories for the Tenant, TenantGroup, RIR, and Aggregate models
  - (I added Tenant and TenantGroup because Aggregates can have a Tenant, and I wanted to make fully populated Aggregates from the factory. This ended up having a lot broader impact than I anticipated; mea culpa!)
- Update NautobotTestRunner to automatically populate these four models before running all tests, as a global fixture. The intent is that other factories will be added here as we make progress on #2226, and concurrently we can reduce the amount of bespoke data that each test is generating for itself.
- Having globally populated Tenants and TenantGroups ended up impacting pretty much every `test_filters.py` that has a tenant-related filter, because all of those tests were assuming that the only Tenants/TenantGroups in the DB were the ones that they themselves had created. I've had to modify the test cases accordingly to instead associate to a number of the fixture Tenants, and for the filter tests to correctly select the relevant Tenants to filter against. I think this is a good direction for us to go in, in general, as it cuts down on the number of hard-coded magic numbers in our filter tests, but it does make for large diffs. Sorry again!
- Similarly, having a network of globally populated objects at all breaks the default assumption of our API/view "delete" test cases, which were assuming that any arbitrary instance of the model under test could be freely deleted. This is not true whenever we have foreign keys with `on_delete=models.PROTECT` - for example we can no longer arbitrarily pick any RIR(s) to delete, because many of the fixture RIRs have associated Aggregates. To address this, I added the new test methods `get_deletable_object` (for single delete) and `get_deletable_object_pks` (for bulk delete) to both API and View tests; these methods default to the previous behavior of picking more or less arbitrary instances (so will not break plugins and the like) but can be overridden if an individual test needs to specially select (or create) instances that are valid for deletion.
- Added `get_random_tags_for_model` helper function; this doesn't actually achieve much in the present implementation (because Tags are still, for now, created on a per-testcase basis via a JSON fixture, not generated by a factory during the global test setup, there are never any Tags to choose from during global setup), but I've confirmed that it works well when manually running the factories, and will be useful when we migrate Tags to a factory as well
- Added natural-key support to the RIR model (this was done during my initial work using JSON fixtures; it's no longer directly relevant to the factory-boy alternate approach, but it certainly doesn't hurt anything)
- Fixed a couple of issues in the RIR and Aggregate UI views that I noticed during manual testing (a couple of table columns missing `linkify=True`, a toggle button on the RIR view that does nothing and was apparently leftover from *old* NetBox code)

image

image

image

image

In my local testing, the full unit-test suite appears to be about 30 seconds slower than `next` baseline (1700 seconds versus 1650 seconds) but that may fall within the range of normal variation. I was hoping to notice substantial speed gains but that does not appear to be the immediate result.

# TODO

- [x] Unit, Integration Tests


nautobot/ipam/templates/ipam/rir_list.html
```diff
@@ -1,23 +0,0 @@
-{% extends 'generic/object_list.html' %}
```
bryanculver, line 1: What's this deletion for?

glennmatthews: Removing unused/unneeded UI cruft - all this file did was add an "IPv4 Stats"/"IPv6 Stats" toggle button that never (in Nautobot, at least) ever actually worked - try it out at https://demo.nautobot.com/ipam/rirs/ for yourself :)

bryanculver:
> In my local testing, the full unit-test suite appears to be about 30 seconds slower than `next` baseline (1700 seconds versus 1650 seconds) but that may fall within the range of normal variation. I was hoping to notice substantial speed gains but that does not appear to be the immediate result.

I suspect some of the `get_deletable_object` calls aren't helping here but I think necessary to not massively complicate the refactor here.

nautobot/ipam/templates/ipam/rir_list.html
bryanculver: 🤦 Good catch!

HanlinMiao (approved):
LGTM! I wonder if the create_batch() method can be used to write new performance tests where we measure the time to create, list and delete a large number of model instances.

PR #2493, opened 2022-09-26
Resolves #2075 - Fix dark flash

### LINKED ISSUE #2075: Darkmode is not consistent, page flashes in default mode during page load
### Environment
* Python version: 3.9
* Nautobot version: v1.4.0a2

### Steps to Reproduce
1. Enable dark mode
2. Go to any other page then the current one

### Expected Behavior
Dark mode is permanent active

### Observed Behavior
During load, I can see the light mode pop up for less then a second before it switches to dark mode.
This breaks the user experience and also hurts in the eye.

matt852:
# Closes: #2075
# What's Changed

Fixes white screen flash in dark mode.  This is caused as the CSS theme is loaded client-side via JavaScript.  However the page loads the default white theme before the client-side theme selection occurs, causing a brief white screen flash.

This works locally when testing across multiple browsers, and I've done the best I can to verify it is working better.  There is still an occasional, brief white screen flash on initially loading Nautobot, but the flash when switching pages seems to have gone away in my testing.

# TODO

- [ X ] Explanation of Change(s)
- [ X ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests


HanlinMiao (approved):
Looks Awesome!

nautobot/project-static/css/hide_body.css
```diff
@@ -0,0 +1,5 @@
+body {
```
glennmatthews, line 1: Maybe add this to `base.css` instead of creating a separate stylesheet? Also please add a comment explaining why/how this is used.

nautobot/project-static/css/dark.css
```diff
@@ -25,3 +25,72 @@ html[data-theme="dark"] .opblock-summary-method {
...
+}
+
+/* Breadcrumbs */
+html[data-theme="dark"] .breadcrumb>li+li:before {
+  color: rgb(142, 142, 142);
+}
+
+/* Code */
+html[data-theme="dark"] code {
+  color: rgb(0, 255, 0);
```
glennmatthews, line 95: I would have preferred that this be a separate PR from the bug fix one, but won't ask you to carve it out separately at this point. Can you provide some side-by-side screenshots showing what's changed by these / why they're needed?

changes/2075.fixed
```diff
@@ -0,0 +1 @@
+Fixed white screen flash in dark mode. Updated dark mode theme to be less...dark.
```
glennmatthews, line 1:
```suggestion
Fixed white screen flash in dark mode.
```

You can add a separate `2493.changed` file for `Updated dark mode theme to be less...dark.`

nautobot/project-static/css/hide_body.css
matt852: I'll add a comment for sure.
However this CSS code cannot go into `base.css`.  It's purpose is to hide the contents of `<body>` until after the Javascript event `DOMContentLoaded` is triggered, which occurs after the page is fully loaded.  Then as part of the JS, `<body>` is displayed.  This is what helps with the white screen flashing issue.

changes/2075.fixed
matt852: Done

nautobot/project-static/css/hide_body.css
glennmatthews: OK, thanks for the explanation!

nautobot/core/templates/admin/base.html
```diff
@@ -8,6 +8,9 @@
 <html lang="en">
 <head>
     <title>{% block title %}Admin Home{% endblock %} - Nautobot</title>
+    <link rel="stylesheet"
+          href="{% static 'css/hide_body.css' %}"
```
glennmatthews, line 12: Possibly?

```suggestion
          href="{% static 'css/hide_body.css' %}?v{{ settings.VERSION }}"
```

Same comment elsewhere this file is referenced. IIRC this is a mechanism to ensure that when upgrading to a new version of Nautobot the browser is hinted to re-download this file rather than relying on cache, in case it changed in the upgrade.

nautobot/project-static/css/hide_body.css
```diff
@@ -0,0 +1,14 @@
+/* This CSS code cannot go into base.css.
+It is loaded immediately in the <head> section before anything else.
```
glennmatthews, line 2: Still a bit curious about this. `base.css` is also loaded in the `<head>` of the file. Are we just relying on the order of declarations in `<head>` to ensure that this file gets loaded early enough that we avoid the white flash? Would it be more reliable to embed this CSS directly in the page template itself rather than linking it as a separate file for the browser to load?

matt852: No specific reason it is in a separate file, other than that's how the other CSS files are loaded (via separate files).
I do not know if it would be more reliable, but I am ok with embedding it directly if desired.

Yes the ordering is important, as we want to hide the `<body>` contents right away, before anything else starts to load.  Then we render `<body>` after everything is loaded, which helps with the brief switchover from white/base CSS (server side) to when JS loads dark mode CSS (client side).

bryanculver (commented):
@matt852

What if instead...:

```diff
diff --git a/nautobot/core/templates/base.html b/nautobot/core/templates/base.html
index f6eff06ec..2ce67368e 100644
--- a/nautobot/core/templates/base.html
+++ b/nautobot/core/templates/base.html
@@ -3,7 +3,7 @@
 {% load plugins %}

 <!DOCTYPE html>
-<html lang="en">
+<html lang="en"{% if context.request.COOKIES|get_item:theme == 'dark' %} data-theme="dark"{% endif %}>
 <head>
     <title>{% block title %}Home{% endblock %} - {{ settings.BRANDING_TITLE }}</title>
     {% include 'inc/media.html' %}
diff --git a/nautobot/project-static/js/theme.js b/nautobot/project-static/js/theme.js
index 2b39f5c2a..5d24aeea5 100644
--- a/nautobot/project-static/js/theme.js
+++ b/nautobot/project-static/js/theme.js
@@ -80,12 +80,14 @@ function detectThemeSettings() {
 function setDarkTheme() {
     htmlEl.dataset.theme = "dark";
     darkElement.disabled = undefined;
+    document.cookie = "theme=dark; path=/";
     /* Highlight dark theme image to show it's active */
 }

 function setLightTheme() {
     htmlEl.dataset.theme = "light";
     darkElement.disabled = "disabled";
+    document.cookie = "theme=light; path=/";
 }
```

matt852:
@bryanculver Updated with your suggestions.  I cleaned up the "show/hide <body>" toggle code.

I left in the client-side enable/disable of the CSS loading as well for 2 scenarios:
1. Nautobot defaults to Light theme when first loading the page, even when the user system is set to dark.  Leaving client-side enable/disable of the CSS prevents this, and while there is an initial, 1-time "flash" from light to dark, it's only on the initial load.
2. Deleting cookies and reopening the page has the same effect as above.

I haven't seen any other issues in testing with either loading the correct theme, or the white flashing issue (after first-time loading of home page).

bryanculver (approved):

Issue #2330, opened 2022-08-29
Change Log tab missing in device component detail views


glennmatthews:
### Environment
* Nautobot version (Docker tag too if applicable): 1.4.2b1
* Python version:
* Database platform, version: MySQL
* Middleware(s): n/a

### Steps to Reproduce
1. Create a device with components (interfaces, power ports, console ports, etc - all components appear to be affected)
2. Navigate to the device detail view, then to the appropriate device component list view (interfaces, power ports, etc.), then click on an individual component (interface, power port, etc.) in the table, e.g. http://localhost:8080/dcim/console-ports/8b5d0e98-8ed9-4d93-8451-68202fc8253a/

![image](https://user-images.githubusercontent.com/5603551/187236121-4259316a-3410-495c-bb61-82a8ed6f44be.png)

### Expected Behavior

"Change Log" tab to be included in the page.

### Observed Behavior

"Change Log" tab is missing, however the changelog view does exist as you can manually enter the URL such as http://localhost:8080/dcim/console-ports/8b5d0e98-8ed9-4d93-8451-68202fc8253a/changelog/ in order to view it.

This is probably because the various device component models do not actually inherit from the base `ChangeLoggedModel` mixin class at present, and so are lacking an `instance.get_changelog_url` method.

glennmatthews:
May be a regression resulting from #2049. See also #1424.

lampwins:
Let's bite the bullet and turn these into proper change-logged models. For the migrations to do that, let's set the timestamps of the component's device timestamp. In this context, it is less confusing than using the timestamp of when the migration is run.

Issue #1424, opened 2022-02-25
Update all objectchange-capable models to inherit from ChangeLoggedModel


glennmatthews:
### Proposed Changes

There are a number of models (`CircuitTerminationModel`, `ComponentModel`, `ComponentTemplateModel`, `VMInterface`, possibly others) that implement a `to_objectchange` method but do not inherit from `ChangeLoggedModel`. It would be good to make them into ChangeLoggedModel subclasses (or more generally, they may be able to convert into `OrganizationalModel` or even `PrimaryModel` subclasses.)

### Justification

1. Reduction of code duplication - these models could then call `super().to_objectchange()` instead of implementing their own construction of `ObjectChange` instances
2. Consistent functionality - these models would then have `created` and `last_updated` fields like most other models do. If they become `PrimaryModel` subclasses, they would also add support for Tags which may also be desirable.

bryanculver:
In `next`:

- `CircuitTermination` is now a `PrimaryModel`
- `ComponentModel` is now a `PrimaryModel`
- `ComponentTemplateModel` inherits from `ChangeLoggedModel`
- `CustomFieldChoice` inherits from `ChangeLoggedModel`
- `VMInterface` inherits from `PrimaryModel`

No other models implement their own `to_objectchange`. All above call super when they overload.

PR #2503, opened 2022-09-27
Fixed missing Change Log tab on device component detail views

jathanism:
# Closes: #2330 
# What's Changed

- This updates `ComponentModel` to inherit from `PrimaryModel` therefore making all device components change-logged models now.
- This updates `ComponentTemplateModel` to also inherit from `ChangedLoggedModel` therefore making all device components change-logged models now.
- Data migration explicitly nullifies the values of `created`/`last_updated` fields on these objects.
- Updated the `inc/created_updated.html` to not show `Created` field if it is null.
- Updated docs for `Devices` and `Device Types` to include version-added admonition for `created`/`last_updated` to indicate that they will be null.

# TODO

- [ ] Unit, Integration Tests


glennmatthews:
Toward #1424, should we go ahead and do the same for `ComponentTemplateModel` as well?

glennmatthews:
Do we need a data migration (https://github.com/nautobot/nautobot/issues/2330#issuecomment-1232084079) to set appropriate initial values for these new fields?

jathanism:
> Do we need a data migration ([#2330 (comment)](https://github.com/nautobot/nautobot/issues/2330#issuecomment-1232084079)) to set appropriate initial values for these new fields?

The date fields? No, they are auto fields (`auto_now_add=True`).

glennmatthews:
> > Do we need a data migration ([#2330 (comment)](https://github.com/nautobot/nautobot/issues/2330#issuecomment-1232084079)) to set appropriate initial values for these new fields?
> 
> The date fields? No, they are auto fields (`auto_now_add=True`).

See the linked comment - @lampwins had requested that we set the created/last_updated timestamps of migrated components (I'm not sure about component templates) to match the parent device, instead of the arbitrary time of the migration being run. I don't personally feel strongly about this, but wanted to make sure it didn't get lost in the shuffle.

jathanism:
> > > Do we need a data migration ([#2330 (comment)](https://github.com/nautobot/nautobot/issues/2330#issuecomment-1232084079)) to set appropriate initial values for these new fields?
> > 
> > 
> > The date fields? No, they are auto fields (`auto_now_add=True`).
> 
> See the linked comment - @lampwins had requested that we set the created/last_updated timestamps of migrated components (I'm not sure about component templates) to match the parent device, instead of the arbitrary time of the migration being run. I don't personally feel strongly about this, but wanted to make sure it didn't get lost in the shuffle.

This is a lot of work for not much benefit in my opinion.

nautobot/dcim/models/device_component_templates.py
```diff
@@ -64,8 +64,10 @@ def instantiate(self, device):
         """
         raise NotImplementedError()
 
-    def to_objectchange(self, action):
-        # Annotate the parent DeviceType
+    def to_objectchange(self, action, *, related_object=None, object_data_extra=None, object_data_exclude=None):
+        """
+        Return a new ObjectChange with the `related_object` pinned to the `device_type` by default.
+        """
```
glennmatthews, line 70: Should this be changed to call `super().to_objectchange()` now instead of constructing an `ObjectChange` directly?

nautobot/dcim/models/device_components.py
```diff
@@ -81,7 +76,10 @@ def __str__(self):
             return f"{self.name} ({self.label})"
         return self.name
 
-    def to_objectchange(self, action):
+    def to_objectchange(self, action, *, related_object=None, object_data_extra=None, object_data_exclude=None):
```
glennmatthews, line 79: Same question/suggestion here.

glennmatthews (approved):

Issue #270, opened 2021-04-09
Implement user guide for working with custom fields


glennmatthews:
### Environment
* Python version:
* Nautobot version: 1.0.0b3

### Proposed Functionality

Add a "user guide" to the documentation specifically focusing on use cases, applications, tips, and caveats in using the Custom Fields feature.

### Use Case

Custom fields are a surprisingly complex and nuanced feature. The current documentation is fairly bare-bones in this regard, and doesn't describe:

- worked examples of regular expression validation
- explanation of how creating or changing a custom field does (or does not!) apply to existing data records, especially as involves "required" custom fields
- examples of filtering ORM queries, REST API queries, GraphQL queries based on custom field values

### Database Changes

None

### External Dependencies

None

brobare:
Related to custom fields and documentation, I was wondering if it might be possible to add "help" context bubbles to custom fields in the UI. 

A lot of the other areas in the UI have the "help" icon that points at, I believe, markdown files. An example: https://demo.nautobot.com/dcim/sites/ams01/edit/

I was thinking it would be nice to be able to point at the location of a file when defining the custom field. Then, whenever that custom field is rendered, the "help" icon would be there and clicking on it would pull up the markdown. 

Thoughts? Can open a new issue if that would be a better path.

PR #2504, opened 2022-09-27
Add Custom Field user guide to documentation

gsnider2195:
# Closes: #270
# What's Changed

- Add Custom Field user guide to documentation 
- Enable mkdocs-material details extension for collapsible sections using `???` syntax and update existing docs that used html for collapsible sections

localhost_8001_en_stable_user-guides_custom-fields_ (5)

# TODO

- n/a Unit, Integration Tests


HanlinMiao (commented):
Very solid and comprehensive! I learned a lot from just reading this PR.
One thing I would say is to add some screenshots to show the valid format for default values for different type of fields. Especially `Selection`, `Multiple Selection`, `JSON`, since the formats for those are not immediately obvious IMO.

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,344 @@
...
+### GraphQL
+
+#### Retrieve Data for a Custom Field in GraphQL
+
+Individual custom fields can be retrieved in GraphQL queries by using the `cf_<fieldname>` field name format:
+
+```graphql
+{
+  devices {
+    cf_text_custom_field
```
itdependsnetworks, line 179: Could we use something a bit more real life over `text_custom_field`? It kinda makes you think that you standard is actually `cf_<fieldname>_custom_field`,

mkdocs.yml
```diff
@@ -133,6 +133,7 @@ nav:
           - The Search Bar: "user-guides/getting-started/search-bar.md"
           - Git as a Data Source: "user-guides/git-data-source.md"
           - GraphQL: "user-guides/graphql.md"
+          - Custom Fields: "user-guides/custom-fields.md"
```
glennmatthews, line 136: I'll register my usual complaint here that I think our separate `User Guides` section is not a great pattern. We already have `models/extras/customfield.md`, what's the purpose/benefit to having a new, separate, second page of documentation for custom fields, instead of adding all of this great content to the existing page?

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,344 @@
+# Custom Fields
+
+Custom fields are a method of adding new fields to existing Nautobot models. For more general information on how custom fields work, refer to the [custom fields model documentation](../models/extras/customfield.md).
```
glennmatthews, line 3: If we are going to go with two separate custom-field documentation pages, then `models/extras/customfield.md` should also have a link back to this new file to aid with discovery.

glennmatthews, line 3: Perhaps:
```suggestion
Custom fields are a convenient method of adding new fields to existing Nautobot models without needing to modify the underlying source code or database scheme. For more general information on how custom fields work, refer to the [custom fields model documentation](../models/extras/customfield.md).
```

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,344 @@
...
+## When to use Custom Fields
+
+Custom fields are commonly used for fields that need different values across individual objects. For example, a custom field on devices to reference an internal ticket number that inventories each device in Nautobot. If you need a solution for marking multiple objects with a common flag, [tags](../models/extras/tag.md) may be a better fit.
+
+## Create a Custom Field
+
+Navigate to the custom fields page by clicking on **Extensibility -> Custom Fields** in the Nautobot menu. Click on **Add** to create a new custom field.
+
+!!! note
+    Custom fields are initialized on associated objects when a content type is added to the custom field, including when the custom field is created. The initial value will be set to the `default` value of the custom field.
```
glennmatthews, line 14: From a developer perspective I understand this note but I don't think it's super clear to an end user. Perhaps:

1) Move this down to the `Default` or "Content Types" sub-sections below?
2) Rephrase to something like "When a custom field is created or associated to a new content type (model), all affected existing objects will be updated to include the custom field, receiving its `default` value. This update runs as a background task via Celery (insert link here), so it may take a few seconds or more before all objects reflect the new custom field, depending on the size of your database."

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,344 @@
...
+#### Grouping
+
+The optional grouping field allows you to group custom fields into collapsible menus.
+
+![Custom Field Grouping](./images/custom-fields/custom_field_detail_grouped.png)
+
+#### Slug
+
+The slug is used to create the URL endpoint for the custom field and is also used as the key in the custom field dictionary. This is automatically created from the label if not supplied. The default value should be sufficient for most deployments.
+
```
glennmatthews, line 33: Perhaps add an admonition about using underscores rather than dashes or whitespace if defining a custom slug, similar to the tip that we have in `customfield.md`?

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,344 @@
...
+#### Type
+
+The type of data that the custom field will store. Valid choices are documented in the custom field [model documentation](../models/extras/customfield.md#creating-custom-fields).
+
+#### Weight
+
+Weight determines how custom fields are sorted in forms and object detail views. Higher-weight fields will be ordered lower on the page.
```
glennmatthews, line 40:
```suggestion
Weight determines how custom fields are sorted in forms and object detail views. Higher-weight fields will be ordered lower on the page; if multiple fields have the same weight, they will be listed alphabetically.
```

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,344 @@
...
+#### Required
+
+Check the required box if this field cannot be null on the associated objects.
+
+!!! warning
+    If an associated object does not have a valid value assigned to a required custom field, that field must be updated with a valid value before the object can be saved. Try to supply a valid default value when creating required custom fields. Since automatic provisioning is only performed when a custom field's content types change, if an existing custom field is changed from optional to required the associated objects will have to be updated manually.
+
+#### Default
+
+The default value for the custom field. This form field only accepts JSON data so if you want to set the field default to a string of `foo` you must supply the JSON string `"foo"`. Boolean field valid values are `true` and `false` (all lowercase). Date fields are strings in the format `YYYY-MM-DD`.
```
glennmatthews, line 57: Should this be?

```suggestion
The default value for the custom field. This form field only accepts JSON data so if you want to set the field default to a string of `foo` you must supply the JSON string `"foo"`. Boolean field valid values are `true` and `false` (all lowercase). Date fields are strings in the format `"YYYY-MM-DD"`.
```

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,344 @@
...
+
+#### Filter Logic
+
++/- 1.4.0
+    Custom field [extended filtering](../rest-api/filtering.md#lookup-expressions) introduced extended lookup expression filters for `exact` and `icontains`, duplicating the functionality of both the `Strict` and `Loose` settings.
```
glennmatthews, line 62: This should really be called out in the appropriate section of `customfield.md` as well.

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,344 @@
...
+
+The filter logic setting applies to filtering on custom fields in the UI and API. For example, when filtering in the API to find a device with the custom field `cf1` set to `"abc"` you would query `/api/dcim/devices/?cf_cf1=abc`. If the filter logic setting is set to `Loose` this would match on `"ABC"` and `"abcdef"`. If the filter logic setting is set to `Strict` only devices with the custom field set to exactly "abc" (case sensitive) would be returned. If the filter logic setting is set to `disabled`, no filters will be available for this custom field, including extended lookup filters. The `Loose` and `Strict` settings only change the behavior of the default filter (`cf_customfieldname`) on `text`, `url` and `json` custom fields.
+
+#### Move to Advanced Tab
+
+When selected, the custom field will appear in the advanced tab of the object detail view instead of the default tab.
```
glennmatthews, line 68:
```suggestion
When selected, the custom field will appear in the "Advanced" tab of the object detail view instead of the default tab.
```

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,344 @@
...
+
+The choices to be presented for `Selection` and `Multiple selection` custom field types. These are displayed in the order of the weight values supplied with the lowest weight on top. If regex validation is being used, these choices must match the regular expression.
+
+![Custom Field Choices](./images/custom-fields/custom_field_choices.png)
+
+![Custom Field Select](./images/custom-fields/custom_field_select.png)
+
+## Editing Custom Fields
+
+Since automatic provisioning is only performed when a custom field's content types change, some changes made to existing custom fields are not reflected on the associated objects automatically. Some examples of cases where this might cause unexpected behavior are changes to the `required`, `default` and validation fields. If a custom field is created with `required=False` and then later changed to `required=True`, all of the associated objects will fail validation the next time they're saved unless the custom field is updated with a valid value.
```
glennmatthews, line 120:
```suggestion
Since automatic provisioning is only performed when a custom field's content types change, some changes made to existing custom fields are not reflected on the associated objects automatically. Some examples of cases where this might cause unexpected behavior are changes to the `required`, `default` and validation fields.

If a custom field is created with `required=False` and then later changed to `required=True`, all of the associated objects that do not yet have a value for this field will fail validation the next time they're saved unless updated with a valid value. Similarly, changes to the validation fields for a custom field will not automatically result in changes to affected objects, but may require changes the next time those objects are edited in order to bring them into compliance with the updated validation rules.

If a custom field's `default` value is changed, newly created objects will use the new default, but existing objects that were set to the previous default value will *not* automatically be updated.
```

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,344 @@
...
+
+## Deleting Custom Fields
+
+Custom fields are removed from associated objects when a content type is removed from the custom field, including when the custom field is deleted.
```
glennmatthews, line 124: Perhaps add note about this happening as a background task?

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,344 @@
...
+
+## Retrieving Custom Field Data
+
+Custom fields augment an existing model so retrieving custom field values is different from native fields. All custom field data is stored as a dictionary in the model field named `_custom_field_data` but there is a property named `cf` to make accessing this field easier. Example:
```
glennmatthews, line 128: Probably worth explaining the use of field `slug`s (actually `name` at present 🤐) as keys to this dictionary.

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,344 @@
...
+```json
+{
+  "id": "ffd8df99-6d1a-41c3-b19f-b8357eefc481",
+  ...
+  "custom_fields": {
+    "boolean_custom_field": true,
+    "date_custom_field": "1970-01-01",
+    "grouped_custom_field_1": "",
+    "grouped_custom_field_2": "",
+    "grouped_custom_field_3": "",
```
glennmatthews, line 161: Might leave these out, they're not adding much value to the example?

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,344 @@
...
+        "name": "phx-leaf1-1-1",
+        "id": "8bd9ed2b-3774-4806-9d17-c9f21f2c73e4",        
+        "_custom_field_data": {
+          "date_custom_field": null,
+          "text_custom_field": "This is the custom field value",
+          "boolean_custom_field": true,
+          "integer_custom_field": 12345,
+          "grouped_custom_field_1": null,
+          "grouped_custom_field_2": null,
+          "grouped_custom_field_3": null
```
glennmatthews, line 246: Same suggestion to remove here.

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,344 @@
...
+
+```json
+{
+  "data": {
+    "devices": [
+      {
+        "name": "phx-leaf1-1-1",
+        "id": "8bd9ed2b-3774-4806-9d17-c9f21f2c73e4",        
+        "_custom_field_data": {
+          "date_custom_field": null,
```
glennmatthews, line 240: Just checking - is `null` the expected value here?

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,344 @@
...
+### Modify Custom Field Data in Nautobot Shell
+
+Custom field data behaves like a python dictionary in the Nautobot Shell. When modifying custom fields through the Nautobot Shell, make sure to use the `.validated_save()` method to save the object to ensure that custom field validation is performed.  Example:
+
+```py
+>>> device.cf["text_custom_field"]
+'Some other value'
+>>> d.cf["text_custom_field"] = "Changed"
+>>> d.validated_save()
+>>> device.cf["text_custom_field"]
```
glennmatthews, line 307: Inconsistent use of `device` versus `d` here?

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,344 @@
...
+
+#### Weight
+
+Weight determines how custom fields are sorted in forms and object detail views. Higher-weight fields will be ordered lower on the page.
+
+#### Description
+
+The description of a custom field is shown as a mouseover tooltip in object detail views and as help text under form fields.
+
+![Custom Field Edit](./images/custom-fields/custom_field_edit.png)
```
glennmatthews, line 46: Would be good to use more realistic field labels here as well.

glennmatthews (commented):
Good content!

mkdocs.yml
gsnider2195: Agreed that this is not much of a user guide but more in depth documentation.

nautobot/docs/user-guides/custom-fields.md
gsnider2195: It's valid but I'll change it to avoid confusion

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,323 @@
...
+#### Slug
+
+The slug is used to create the URL endpoint for the custom field and is also used as the key in the underlying custom field data dictionary. This is automatically created from the label if not supplied. The default value should be sufficient for most deployments.
+
+!!! tip
+    Because custom field data is included in the database, in the REST API and in GraphQL, we strongly recommend that when defining a custom field, you provide a `slug` that contains underscores rather than dashes (`my_field_slug`, not `my-field-slug`), as some features may not work optimally if dashes are included in the slug. Similarly, the provided `name` should also contain only alphanumeric characters and underscores, as it is currently treated in some cases like a slug.
+
+#### Type
+
+The type of data that the custom field will store. Valid choices are documented in the custom field [model documentation](../models/extras/customfield.md#creating-custom-fields).
```
jathanism, line 36:
```suggestion
The type of data that the custom field will store. Valid choices are documented in the [custom field model documentation](../models/extras/customfield.md#creating-custom-fields).
```

mkdocs.yml
```diff
@@ -133,6 +134,7 @@ nav:
...
```
jathanism, line 137: Can we please alpha sort the User Guides while you're in here touching this? I think it will help navigation.

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,323 @@
...
+The description of a custom field is shown as a mouseover tooltip in object detail views and as help text under form fields.
+
+![Custom Field Edit](./images/custom-fields/custom_field_edit.png)
+
+#### Required
+
+Check the required box if this field cannot be null on the associated objects.
+
+!!! warning
+    If an associated object does not have a valid value assigned to a required custom field, that field must be updated with a valid value before the object can be saved. Try to supply a valid default value when creating required custom fields. Since automatic provisioning is only performed when a custom field's content types change, if an existing custom field is changed from optional to required the associated objects will have to be updated manually.
```
jathanism, line 53:
```suggestion
    If an associated object does not have a valid value assigned to a required custom field, that field must be updated with a valid value before the object can be saved. Try to supply a valid default value when creating required custom fields. Since automatic provisioning is only performed when the assigned content types for a custom field change, if an existing custom field is changed from optional to required the associated objects will need to be updated manually.
```

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,323 @@
...
+
+#### Default
+
+The default value for the custom field. This form field only accepts JSON data so if you want to set the field default to a string of `foo` you must supply the JSON string `"foo"`. Boolean field valid values are `true` and `false` (all lowercase). Date fields are strings in the format `"YYYY-MM-DD"`. Select and multiselect field default must match one of the field's choices.
```
jathanism, line 57:
```suggestion
The default value for the custom field. This form field only accepts JSON data so if you want to set the field default to a string of `foo` you must supply the JSON string `"foo"`. Boolean field valid values are `true` and `false` (all lowercase). Date fields are strings in the format `"YYYY-MM-DD"`. Select and multi-select field default must match one of the field's choices.
```

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,323 @@
...
+
+!!! note
+    The default value for a select or multiselect field must match one of the existing choices. If the desired default value is not in the list of choices, the choices must be updated and saved before the default can be changed. As a result of this behavior, default values cannot be set on select and multiselect fields when a custom field is created.
```
jathanism, line 60:
```suggestion
    The default value for a select or multi-select field must match one of the existing choices. If the desired default value is not in the list of choices, the choices must be updated and saved before the default can be changed. As a result of this behavior, default values cannot be set on select and multi-select fields when a custom field is created.
```

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,323 @@
...
+
+#### Move to Advanced Tab
+
+When selected, the custom field will appear in the "Advanced" tab of the object detail view instead of the default tab.
+
+### Assignment
+
+#### Content Types
+
+The list of content types to add this custom field to. Only models that subclass the `nautobot.extras.models.customfields.CustomFieldModel` model class can be selected.
```
jathanism, line 77:
```suggestion
The list of content types to add this custom field to. Only models that inherit from the `nautobot.extras.models.customfields.CustomFieldModel` model class can be selected.
```

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,323 @@
...
+> ^\d+$
+
+Must be exactly 8 alphanumeric characters
+> ^[0-9a-zA-Z]{8}$
+
+Must be between 8 and 10 alphanumeric characters and underscore
+> ^\w+{8,10}$
+
+Must contain anything that is not whitespace
+> \S
```
jathanism, line 114: IMO change these example regex to code gates with `no-highlight` vs. block quote.

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,323 @@
...
+## Retrieving Custom Field Data
+
+Custom fields augment an existing model so retrieving custom field values is different from native fields. All custom field data is stored as a dictionary in the model field named `_custom_field_data` but there is a property named `cf` to make accessing this field easier. Example:
+
+### Retrieve Custom Field Data in Nautobot Shell
+
+```py
+# retrieve all custom field data
+>>> device.cf
+{'eol_date': '1970-01-01', 'support_group': 'Network Operations (555-4357)', 'dmz_device': True, 'cmdb_id': 12345}
```
jathanism, line 145: Just to improve readability so the output doesn't scroll to the right.

```suggestion
{'eol_date': '1970-01-01', 
 'support_group': 'Network Operations (555-4357)',
 'dmz_device': True,
 'cmdb_id': 12345}
```

nautobot/docs/user-guides/custom-fields.md
```diff
@@ -0,0 +1,323 @@
...
+    }
+    ```
+
+### GraphQL
+
+#### Retrieve Data for a Custom Field in GraphQL
+
+Individual custom fields can be retrieved in GraphQL queries by using the `cf_<fieldname>` field name format:
+
+```graphql
```
jathanism, line 179: Unfortunately it doesn't look like the `graphql` syntax is highlighting. Is there another plugin we need to install/enable?

gsnider2195: It doesn't look like graphql is supported in the code highlighting library that material for mkdocs uses, [`pygments`](https://pygments.org/languages/).  #2571

jathanism (approved):
adam handshake michelangelo

PR #2514, opened 2022-09-28
Add RouteTarget and VRF factories. Add TenancyFilterTestCaseMixin and refactor

glennmatthews:
# Towards: #2283 
# What's Changed

- Added factories for RouteTarget, VRF, Role, VLANGroup, and VLAN models and add these factories to the test runner.
- Added `OrganizationalModelFactory` and `PrimaryModelFactory` base classes - currently these are mostly empty but I've left TODO comments regarding potential areas for extension of these. I moved the Tag mapping logic out of the individual factories and into `PrimaryModelFactory` so that it doesn't need to be repeated everywhere.
  - Also added `BaseModelFactory` as a common parent of these two classes, and overrode the `_create` API (which despite the leading underscore is a documented extension point) to call our `.validated_save()` method when creating model instances to ensure they are valid.
- Added `get_random_instances` factory helper function, and used it for Tag mapping as well as `import_target`/`export_target` mapping on the VRF factory.
- As a result of the factory updates, I discovered a bug in the filtering test updates for `tenant_group` filters that I made in #2479, specifically that the tests weren't accounting for the fact that `tenant_group` is a TreeNodeMultipleChoice filter and so matches not only against the specific tenant group(s) provided, but *also any of their descendants*. As a result many of the `test_tenant_group` tests began failing after I modified the factories (and hence the specific set of random data being generated) because they weren't accounting for the possibility of sub-groups.
   - Testing for `tenant` and `tenant_group` filters is basically boilerplate, and I got tired of making the same change in a dozen different places, so I went ahead and created a `TenancyFilterTestCaseMixin` class and made use of it throughout our tests. This results in a decent amount of test code reduction, yay!

image

image

# TODO

- [x] Unit, Integration Tests


nautobot/ipam/factory.py
```diff
@@ -111,3 +99,85 @@ def prefix(self, n):
...
+    - "<4-byte ASN>:<2-byte integer>"
+    """
+    branch = randgen.randint(0, 2)
+    if branch == 0:
+        # 16-bit ASNs 64496–64511 are reserved for documentation and sample code
+        return f"{randgen.randint(64496, 64511)}:{randgen.randint(0, 2**32 - 1)}"
+    if branch == 1:
+        # The official stance of factory-boy maintainers is that evaluate() is a private method, but there doesn't
+        # seem to be any convenient alternative way to do this at present.
+        return f"{factory.Faker('ipv4_private').evaluate(None, None, {'locale': None})}:{randgen.randint(0, 2**16 - 1)}"
```
gsnider2195, line 120: For this it would probably be fine to just go straight to Faker
```py
from faker import Faker
>>> Faker().ipv4_private()
'10.100.193.83'
```

glennmatthews: I need to do some more exploration here - my concern is not having reproducible behavior, if I do this I may need to also be careful to initialize the Faker random seed directly as well as the FactoryBoy random seed?

bryanculver (commented):
This is looking pretty good. I'd like to understand a little bit more the need for two separate random instance getters.

glennmatthews:
> This is looking pretty good. I'd like to understand a little bit more the need for two separate random instance getters.

One is for getting a single random instance (e.g., ForeignKey), the other is for getting 0-to-n random instances (e.g., ManyToManyField). The latter is "new" in this PR but only inasmuch as it's a generalization of the `get_random_tags_for_model` function that was introduced in #2479. (Which I just realized is still left in the code at the moment, even though it's no longer used - will get that fixed in the next commit.)

nautobot/ipam/factory.py
gsnider2195: It looks like it works.

```py
>>> factory.random.reseed_random("Nautobot")
>>> factory.Faker("ipv4_private").evaluate(None, None, {"locale": None})
'172.19.147.184'
>>> factory.Faker("ipv4_private").evaluate(None, None, {"locale": None})
'10.87.130.179'
>>> 
>>> factory.random.reseed_random("Nautobot")
>>> Faker().ipv4_private()
'172.19.147.184'
>>> factory.Faker("ipv4_private").evaluate(None, None, {"locale": None})
'10.87.130.179'
```

glennmatthews: Ah yeah, thanks for testing. Looks like `factory.random.reseed_random()` actually reseeds both libraries :)

nautobot/ipam/factory.py
```diff
@@ -111,3 +99,85 @@ def prefix(self, n):
...
+        exclude = (
+            "has_description",
+            "has_rd",
+            "has_tenant",
+        )
+
+    # note that name is *not* globally unique
+    name = factory.Faker("color_name")
+
+    # RD needs to be globally unique, but the random route-distinguisher generation space is large enough that
```
gsnider2195, line 162: Are these are deterministically randomized so they're the same every time tests are run? So we shouldn't need to be concerned about this unless we change the seed or number of objects generated?

glennmatthews: Or we make a change to the factories that results in a different path through the RNG. But yes, if no changes in any of thosse areas, we're supposed to get the same instances every time.

nautobot/ipam/factory.py
```diff
@@ -111,3 +107,199 @@ def prefix(self, n):
...
+            "has_site",
+        )
+
+    # TODO: name is not globally unique, but (site, name) tuple must be.
+    # The likelihood of collision with random names is pretty low, but non-zero.
+    # We might want to consider *intentionally* using non-globally-unique names for testing purposes?
+    name = factory.Faker("word", part_of_speech="noun")
+
+    has_description = factory.Faker("pybool")
+    description = factory.Maybe("has_description", factory.Faker("sentence"), "")
```
gsnider2195, line 220: Should we use `Faker("text")` instead of `sentence` so that we can set max characters to match the model's `max_length`?

nautobot/ipam/factory.py
```diff
@@ -111,3 +107,199 @@ def prefix(self, n):
...
+    - "<2-byte ASN>:<4-byte integer>"
+    - "<IPv4 address>:<2-byte integer>"
+    - "<4-byte ASN>:<2-byte integer>"
+    """
+    branch = randgen.randint(0, 2)
+    if branch == 0:
+        # 16-bit ASNs 64496–64511 are reserved for documentation and sample code
+        return f"{randgen.randint(64496, 64511)}:{randgen.randint(0, 2**32 - 1)}"
+    if branch == 1:
+        return f"{faker.Faker().ipv4_private()}:{randgen.randint(0, 2**16 - 1)}"
```
gsnider2195, line 126: Any reason to use `randgen.randint` over `Faker("pyint")`?

nautobot/utilities/testing/views.py
```diff
@@ -574,42 +574,29 @@ def test_list_objects_anonymous(self):
...
-            if hasattr(self.model, "name"):
-                self.assertIn(instance1.name, content, msg=content)
-                self.assertNotIn(instance2.name, content, msg=content)
-            try:
-                self.assertIn(self._get_url("view", instance=instance1), content, msg=content)
-                self.assertNotIn(self._get_url("view", instance=instance2), content, msg=content)
-            except NoReverseMatch:
-                pass
+            # We should have exactly one table row, and that row is "even" (I guess counting from zero?)
+            self.assertIn('<tr class="even">', content, msg=content)
```
gsnider2195, line 583: After removing the above tests does it make sense to count the rows here to make sure we did only get one row back?

gsnider2195 (approved):

nautobot/utilities/factory.py
```diff
@@ -13,14 +36,62 @@ class Meta:
...
-    return factory.LazyFunction(
-        lambda: factory.random.randgen.choice(model.objects.all()) if model.objects.count() else None
+    branch = factory.random.randgen.randint(0, 2)
+    queryset = _get_queryset_from_model_or_queryset_or_lambda(model_or_queryset_or_lambda)
+    count = queryset.count()
+    if maximum is None:
+        maximum = count
+    if branch == 0 or count == 0 or maximum == 0:
+        return []
+    if branch == 1 or count == 1 or maximum == 1:
```
jathanism, line 90: IMO `any()` is more readable in times like this.

```suggestion
    if not any([branch, count, maximum]):
        return []
    if any([branch == 1, count == 1, maximum == 1]):
```

nautobot/utilities/factory.py
```diff
@@ -1,10 +1,33 @@
+from django.db.models import Model
+
 import factory
 import factory.random
 
 
-def random_instance(model):
+def _get_queryset_from_model_or_queryset_or_lambda(model_or_queryset_or_lambda):
```
jathanism, line 7: Docstring pls

jathanism (changes requested):

nautobot/utilities/testing/views.py
glennmatthews: I'm actually going to mostly revert this set of changes as it's still too fragile (fails for Job table, for example, since it has a custom table HTML template).

nautobot/ipam/factory.py
glennmatthews: It's a bit shorter, compared to `faker.Faker().pyint(0, 2**16 - 1)`? :)

Not really - this was the best approach I came up with before you pointed out that I can use faker.Faker() directly as needed. I'll revisit this.

nautobot/ipam/factory.py
glennmatthews: Good thought, will do.

glennmatthews:
One change I want to draw attention to is the change in the various `test_filter` cases from:

```python
self.assertEqual(self.filterset(params, self.queryset).qs.count(), HARDCODED_NUMBER)
```

to:

```python
self.assertQuerysetEqual(self.filterset(params, self.queryset), self.queryset.filter(...))
```

First, it accounts for the fact that due to factories we don't necessarily have a specific number of filtered instances known in advance. But secondly, using `assertQuerysetEqual` instead of just `assertEqual(count, count)` makes for a stronger test IMHO.

In some cases, the queryset ordering may not necessarily be guaranteed, such as in the VRF case where we might have two VRFs with the same `name` and a null `rd` value, the relative ordering of which would not be guaranteed. In these cases (only) we need to pass `ordered=False` to the `assertQuerysetEqual()` call.

nautobot/utilities/factory.py
```diff
@@ -13,14 +46,62 @@ class Meta:
...
+            # This needs to be done this way because .get_for_model() evaluates a ContentType queryset,
+            # and we need to defer evaluation of that queryset as well.
+            status = random_instance(lambda: Status.objects.get_for_model(Object), allow_null=False)
+    """
+
+    def get_random_instance():
+        queryset = _get_queryset_from_model_or_queryset_or_lambda(model_or_queryset_or_lambda)
+
+        if not allow_null and not queryset.exists():
+            raise RuntimeError(f"No objects in queryset for {model_or_queryset_or_lambda}! {queryset.explain()}")
```
gsnider2195, line 68: Ran into an issue with `queryset.explain()` when using a django tree query queryset:

```
>>> Location.objects.all().explain()
  File "/usr/local/lib/python3.10/site-packages/django/db/backends/utils.py", line 84, in _execute
psycopg2.errors.SyntaxError: syntax error at or near "EXPLAIN"
LINE 26:     EXPLAIN SELECT (__tree.tree_depth) AS "tree_depth", (__t...
```

You get this when using `random_instance(Location, allow_null=False)` when no Location instances exist.

glennmatthews: Interesting! Sounds like maybe a bug worth reporting to django-tree-queries? Since this code is supposed to raise an exception *anyway*, do you think it's a problem that in this case we raise a `SyntaxError` instead of a `RuntimeError`?

nautobot/dcim/tests/test_filters.py
```diff
@@ -892,30 +866,42 @@ def test_has_prefixes(self):
             self.assertEqual(self.filterset(params, self.queryset).qs.count(), 1)
 
     def test_vlan_groups(self):
-        vlan_groups = VLANGroup.objects.all()[:2]
+        vlan_groups = list(VLANGroup.objects.filter(site__isnull=False))[:2]
         params = {"vlan_groups": [vlan_groups[0].pk, vlan_groups[1].slug]}
-        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 2)
+        self.assertQuerysetEqual(
+            self.filterset(params, self.queryset).qs, self.queryset.filter(vlan_groups__in=vlan_groups).distinct()
```
gsnider2195, line 872: Why was distinct required here?

glennmatthews: In general it seems that querysets using a reverse lookup, e.g. `queryset.filter(<related_name>__in=...)` can return duplicate entries. Not sure if every single case where I added `distinct()` is needed but at least some of them definitely were.

nautobot/utilities/factory.py
gsnider2195: It looks like this may be specific to `LocationQuerySet` because `Region.objects.all().explain()` works. I'll get an issue opened to look into `LocationQuerySet`

glennmatthews: Region is an MPTT model, not a django-tree-queries model, so that tracks.

nautobot/utilities/testing/views.py
```diff
@@ -580,6 +580,8 @@ def test_list_objects_filtered(self):
             content = extract_page_body(response.content.decode(response.charset))
             # TODO: it'd make test failures more readable if we strip the page headers/footers from the content
             if hasattr(self.model, "name"):
+                # TODO: this is fragile as we move toward more autogenerated test fixtures,
+                #       as "instance.name" may appear coincidentally in other page text if we're not careful.
```
jathanism, line 584: Since I'm currently in the midst of invententorying `TODO` comments I'm a bit sensitive to this right now. Would the preceding comment about stripping headers/footers help improve this test and make it be less fragile? Is there something immediately actionable otherwise?

nautobot/utilities/testing/views.py
```diff
@@ -623,7 +618,10 @@ def test_list_objects_unknown_filter_no_strict_filtering(self):
             content = extract_page_body(response.content.decode(response.charset))
             # TODO: it'd make test failures more readable if we strip the page headers/footers from the content
             self.assertNotIn("Unknown filter field", content, msg=content)
+            self.assertNotIn(f"No {self.model._meta.verbose_name_plural} found", content, msg=content)
             if hasattr(self.model, "name"):
+                # TODO: this is fragile as we move toward more autogenerated test fixtures,
+                #       as "instance.name" may appear coincidentally in other page text if we're not careful.
```
jathanism, line 624: Same question here as the previous `TODO` comment...

jathanism (approved):
I only have the `TODO` questions, but nothing blocking. This is looking pretty solid. Another @glennmatthews hit.

nautobot/utilities/testing/views.py
glennmatthews: Nothing immediately actionable - in an earlier version of this PR I changed things to look for the table rows instead of the instance names, but that broke for the Job table/list view since it renders the table rows differently, so I reverted things back to this baseline. Stripping headers/footers would probably help but it's still a risk - the main mitigation is setting up factories to generate names that are not likely to appear by coincidence in the page text, such as by using `.capitalize()` or `.upper()` on random-word names so that they don't accidentally match in the middle of some other word.

Do you want me to remove the `TODO` portion of the comment and just leave it as an informational note?

glennmatthews: Opened https://github.com/nautobot/nautobot/issues/2580 as a housekeeping item. I'll add it to the comment here as a reference.

nautobot/utilities/factory.py
bryanculver: Issue opened as #2577, which required fix in `django-tree-queries` which has been done in `0.12`: https://github.com/matthiask/django-tree-queries/blob/main/CHANGELOG.rst#012-2022-11-30

Issue #2509, opened 2022-09-28
Notes and ChangeLog tabs using wrong model templates


Kircheneer:
### Environment
* Nautobot version (Docker tag too if applicable): 1.4.1
* Python version: 3.10.4
* Database platform, version: Docker image `postgres:13-alpine`
* Middleware(s): -

### Steps to Reproduce
1. Create a model in a plugin
2. Call the basic retrieve template for it `$(MODELNAME)_retrieve` (as I assume is best practice due to `NautobotUIViewSet` using this)
3. Create an instance of that model
4. Navigate to the Notes or Changelog tab of that model

### Expected Behavior

The tab displays correctly including the rendering of the other tabs.

### Observed Behavior

Only the tab content renders, not the other tabs.

![image](https://user-images.githubusercontent.com/12400533/192784849-54fdf82c-ec57-4f45-b7a1-5f6f569bfe59.png)

#### Notes

A potential fix for the Notes issue could be [here](https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/views.py#L1700):

```python
-       # Default to using "<app>/<model>.html" as the template, if it exists. Otherwise,
+       # Default to using "<app>/<model>_retrieve.html" as the template, if it exists. Otherwise,
        # fall back to using base.html.
        if self.base_template is None:
-           self.base_template = f"{model._meta.app_label}/{model._meta.model_name}.html"
+           self.base_template = f"{model._meta.app_label}/{model._meta.model_name}_retrieve.html"
            # TODO: This can be removed once an object view has been established for every model.
            try:
                template.loader.get_template(self.base_template)
            except template.TemplateDoesNotExist:
                self.base_template = "base.html"
```

#### Changelog

A potential fix for the Changelog issue could be [here](https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/views.py#L1620):

```python
-       # Default to using "<app>/<model>.html" as the template, if it exists. Otherwise,
+       # Default to using "<app>/<model>_retrieve.html" as the template, if it exists. Otherwise,
        # fall back to using base.html.
        if self.base_template is None:
-           self.base_template = f"{model._meta.app_label}/{model._meta.model_name}.html"
+           self.base_template = f"{model._meta.app_label}/{model._meta.model_name}_retrieve.html"
            # TODO: This can be removed once an object view has been established for every model.
            try:
                template.loader.get_template(self.base_template)
            except template.TemplateDoesNotExist:
                self.base_template = "base.html"

[...]

PR #2522, opened 2022-09-28
Fixed template lookup logic in ObjectNotesView, ObjectDynamicGroupsVi…

HanlinMiao:
…ew and ObjectChangeLogView.

# Closes: #2509 
# What's Changed

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/2509.fixed
+++ b/changes/2509.fixed
@@ -1,0 +1,1 @@
+Fixed template lookup logic in ObjectNotesView, ObjectDynamicGroupsView and ObjectChangeLogView.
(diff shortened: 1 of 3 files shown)


glennmatthews (commented):
Change looks reasonable, but would it make sense to create a common helper function or base class that these three views could share so as to not need code in triplicate?

bryanculver (approved):
@HanlinMiao Just to confirm, is the issue that plugins are trying to use the new template naming convention even though it was most heavily used in `NautobotUIViewSet`?

HanlinMiao:
> @HanlinMiao Just to confirm, is the issue that plugins are trying to use the new template naming convention even though it was most heavily used in `NautobotUIViewSet`?

Yeah, the new plugins are trying to follow the new naming conventions but we have to incoporate both template lookup logic so as not to break some of the existing plugins.

bryanculver:
> 

👍 I know we added backwards compatibility templates our side to address this fact but good to know.

Issue #2555, opened 2022-10-03
collapsing of Job groupings is broken in next


glennmatthews:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.0a0

### Steps to Reproduce
1. Go to https://next.demo.nautobot.com/extras/jobs/
2. Click the "Data Quality" heading

### Expected Behavior

All five jobs under "Data Quality" to be hidden

### Observed Behavior

Only the first job is hidden, the other four remain visible.

This appears to be a regression due to changes to the job_table.html template in #2317.

jathanism:
I confirmed that reverting `job_table.html` to the previous version from commit d80a3cf94928f988a16a5d99d7fdc4ade025782f fixes this issue. 

It's not clear to me why this template needed to change for the custom field grouping feature. If it's not, we should just be able to revert the changes to this file to fix this bug.

PR #2559, opened 2022-10-04
Fixed broken accordion for Job list view.

jathanism:
# Closes: #2555 
# What's Changed

- This reverts `extras/inc/ job_table.html` to the previous version from commit d80a3cf94928f988a16a5d99d7fdc4ade025782f.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/templates/extras/inc/job_table.html
+++ b/nautobot/extras/templates/extras/inc/job_table.html
@@ -17,30 +17,28 @@
 {% load helpers %}
     <tbody>
         {% for row in table.page.object_list|default:table.rows %}
-            {% with forloop.counter0 as count %}
-                {% ifchanged row.record.grouping %}
-                <tr>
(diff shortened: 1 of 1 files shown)


glennmatthews (approved):
This does reintroduce the hypothetical issue that #2317 was attempting to fix, i.e. that this may not work correctly if we have multiple different groupings that slugify down to the same string, but IMHO that's an unlikely corner case that I'm OK with not mitigating at this time.

jathanism:
> This does reintroduce the hypothetical issue that #2317 was attempting to fix, i.e. that this may not work correctly if we have multiple different groupings that slugify down to the same string, but IMHO that's an unlikely corner case that I'm OK with not mitigating at this time.

Right yeah, that's the part I wasn't clear on, so thanks for clarifying. I think the only thing would be to combine the slug with the index. But I didn't really spend time experimenting.

bryanculver (approved):

Issue #2536, opened 2022-09-29
FactoryBoy implementation doesn't account for plugin testing using Nautobot test runner


glennmatthews:
### As ...

P.D. - Plugin Developer

### I want ...

To be able to continue to use `nautobot-server test` to execute my plugin unit tests. This is impacted by work done in #2479 (as part of #2283 <- #2226 <- #1999) because that unconditionally enables global fixture data as part of NautobotTestRunner initialization. My tests may not be designed to account for the presence of that data in the database; furthermore, the global fixture data depends on `factory_boy` library, which is a dev-only dependency of Nautobot and would need to be added as an explicit dev dependency of my plugin as well.

### So that ...

My plugin testing (local and CI) doesn't break when run against Nautobot 1.5. and doesn't require substantial updates to run successfully against Nautobot 1.5.

### I know this is done when...

- My existing plugin unit test automation (using `nautobot-server test` or `coverage run --module nautobot.core.cli test`) continues to execute successfully when updating my development environment to Nautobot 1.5.
- I do not need to update my plugin's development dependencies to have my unit test automation run successfully against Nautobot 1.5.
- As and when I update my plugin tests to account for the possibility of test fixture data, I can *opt in* to the test fixture/factory data that Nautobot core can provide.

### Optional - Feature groups this request pertains to.

- [X] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [X] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

None

### External Dependencies

None

glennmatthews:
Will be fixed in `next` by way of #2514.

PR #2570, opened 2022-10-04
Make use of test factories opt-in so as to not break plugin tests

glennmatthews:
# Closes: #2536 
# What's Changed

- Add `TEST_USE_FACTORIES` settings variable. Default it to `False` in `settings.py` and override to `True` in `nautobot/core/tests/nautobot_config.py`
- Add `TEST_FACTORY_SEED` settings variable.
- Extract factory calls to a `nautobot-server populate_database` command. Call this command from the test runner only if `TEST_USE_FACTORIES == True`.
- Document the new subcommand and new optional settings.

Additionally:

- Make one test case slightly more resilient to counter a failure I saw when testing with different random seeds
- Move `nautobot/core/management/validate_models.py` to `nautobot/core/management/commands/validate_models.py` so that it actually shows up as a nautobot-server subcommand.

# TODO
- [x] Unit, Integration Tests


nautobot/docs/configuration/optional-settings.md
```diff
@@ -960,6 +960,34 @@ If set to `False`, unknown/unrecognized filter parameters will be discarded and
...
+## TEST_USE_FACTORIES
+
++++ 1.5.0
+
+Default: `False`
+
+If set to `True`, the Nautobot test runner will call `nautobot-server populate_database --flush --seed ...` before executing any test cases, pre-populating the test database with various pseudo-random instances of many models.
+
+!!! warning
+    This functionality requires the installation of the [`factory-boy`](https://pypi.org/project/factory-boy/) Python package, which is present in any Nautobot development environment, but is _not_ an inherent dependency of the Nautobot package when installed otherwise.
```
bryanculver, line 982: This should be available to any plugin that uses the `nautobot-dev` containers for their development environment, correct? Just if they want to use local testing and it's not in their dev dependency.

bryanculver (approved):

nautobot/core/management/commands/populate_database.py
```diff
@@ -0,0 +1,69 @@
+from django.core.management import call_command
```
jathanism, line 1: IMO this should be named `populate_test_database` just so it's crystal clear that it is for testing.

nautobot/docs/administration/nautobot-server.md
```diff
@@ -272,6 +272,40 @@ $ nautobot-server nbshell
...
+
+`nautobot-server populate_database [--flush] --seed SEED`
+
+Populate the database with various data as a baseline for testing (automated or manual).
+
+`--flush`  
+Flush any existing data from the database before generating new data.
+
+!!! danger
+    Running this command with the `--flush` argument will clear all existing data in your database. You have been warned.
```
jathanism, line 287: There is no mercy in this dojo!!!!

nautobot/core/settings.py
```diff
@@ -116,6 +116,11 @@
 # Test runner that is aware of our use of "integration" tags and only runs
 # integration tests if explicitly passed in with `nautobot-server test --tag integration`.
 TEST_RUNNER = "nautobot.core.tests.runner.NautobotTestRunner"
+# Disable test data factories by default so as not to cause issues for plugins.
+# The nautobot_config.py that Nautobot core uses for its own tests will override this to True.
+TEST_USE_FACTORIES = False
+# Pseudo-random number generator seed, for reproducibility of test results.
+TEST_FACTORY_SEED = os.getenv("NAUTOBOT_TEST_FACTORY_SEED", None)
```
jathanism, line 123: We might consider making these able to be provided by environment variables unless there's a specific reason you think we shouldn't!

nautobot/docs/configuration/optional-settings.md
jathanism: Ya, but if this were run on a `pip`-installed instance, it wouldn't be there.

nautobot/core/management/commands/populate_database.py
glennmatthews: I'm 50/50 on this. That might also mislead people into thinking that it always populates a test database, rather than populating whatever database is currently live?

glennmatthews: Maybe something like "generate_test_data"?

nautobot/docs/administration/nautobot-server.md
glennmatthews: I should probably add something similar to https://github.com/django/django/blob/main/django/core/management/commands/flush.py#L59, i.e. a confirmation prompt if someone runs this command interactively?

nautobot/docs/configuration/optional-settings.md
bryanculver: Correct. But this is strictly a development focused feature so it would affect users who have `nautobot-server test <plugin_module>` running locally instead of a container. Again, not zero but should be considered if they only exclusively develop by running Nautobot & tests in a container.

glennmatthews: Even then, I'd expect that the plugin *should* declare a wildcard development dependency on `factory-boy`, just like they do for `black`, `pylint`, etc. even though those are also provided by the dev container.

nautobot/core/management/commands/populate_database.py
```diff
@@ -36,6 +43,20 @@ def handle(self, *args, **options):
...
         if options["flush"]:
+            if options["interactive"]:
+                confirm = input(
+                    f"""\
+You have requested a flush of the database before generating new data.
+This will IRREVERSIBLY DESTROY all data in the "{connections[DEFAULT_DB_ALIAS].settings_dict['NAME']}" database,
+including all user accounts, and return each table to an empty state.
+Are you SURE you want to do this?
+
+Type 'yes' to continue, or 'no' to cancel:"""
```
jathanism, line 54: I know this is super pedantic but the UX on having the space after the prompt helps!

```suggestion
Type 'yes' to continue, or 'no' to cancel: """
```

Issue #2460, opened 2022-09-21
New filter form search box


lampwins:
### As ...

Nelly - Network Engineer

### I want ...

An always present search box in list views

### So that ...

I can always quickly type in a string value for a search that utilizes the `q` filter, without having to click the advanced filtering form and add an explicit filter.

### I know this is done when...

A search text field is always present in list views, next to the filter form button.

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

None

### External Dependencies

None

lampwins:
A part of #1998

bryanculver:
@lampwins Just to confirm that this search box would be combined with the filters applied via the dynamic filter form.

lampwins:
@bryanculver correct.

PR #2572, opened 2022-10-04
Add search box filter form to list view

timizuoebideri1:
# Closes: #2460 
# What's Changed

- Add a search box to the list view UI, which allows users to utilize the q filter, without having to click the advanced filtering form and add an explicit filter.
- Remove `q` field from DynamicFilterForm

Screenshot 2022-10-05 at 9 16 02 PM

# TODO

- [x] Unit, Integration Tests


changes/2460.added
```diff
@@ -0,0 +1 @@
+Add search box filter form to list view.
```
jathanism, line 1:
```suggestion
Added search box filter form to generic list views.
```

jathanism (commented):
This is pretty great, my only concern may be with where it is placed. Currently we have a search box on the top of every page to the right of the nav bar, and this appears to duplicate it, so we need to be cautious about the UX so that it isn't confusing.

timizuoebideri1:
> This is pretty great, my only concern may be with where it is placed. Currently we have a search box on the top of every page to the right of the nav bar, and this appears to duplicate it, so we need to be cautious about the UX so that it isn't confusing.

@jathanism What do you suggest

jathanism:
> > This is pretty great, my only concern may be with where it is placed. Currently we have a search box on the top of every page to the right of the nav bar, and this appears to duplicate it, so we need to be cautious about the UX so that it isn't confusing.
> 
> @jathanism What do you suggest

I think previously we had a grayed out text in the search box that said something like `Search for `{verbose_name_plural}` but that no longer seems to be the case. 

Perhaps the global search box in the nav bar could say "Global Search" (it currently only says "Search"), and the list view search could say "Search for devices" (using the verbose name plural) value, so that it's clear they are different?

bryanculver:
> > > This is pretty great, my only concern may be with where it is placed. Currently we have a search box on the top of every page to the right of the nav bar, and this appears to duplicate it, so we need to be cautious about the UX so that it isn't confusing.
> > 
> > @jathanism What do you suggest
> 
> I think previously we had a grayed out text in the search box that said something like `Search for `{verbose_name_plural}` but that no longer seems to be the case. 
> 
> Perhaps the global search box in the nav bar could say "Global Search" (it currently only says "Search"), and the list view search could say "Search for devices" (using the verbose name plural) value, so that it's clear they are different?

Can we keep the `Search {verbose_name_plural}` for this input field and update the input in the menubar to say `Search {BRANDING_TITLE}`?

bryanculver:
> > > This is pretty great, my only concern may be with where it is placed. Currently we have a search box on the top of every page to the right of the nav bar, and this appears to duplicate it, so we need to be cautious about the UX so that it isn't confusing.
> > 
> > 
> > @jathanism What do you suggest
> 
> ~I think previously we had a grayed out text in the search box that said something like `Search for `{verbose_name_plural}` but that no longer seems to be the case.~ For some reason in our object detail view template we have this search box that has "Search {verbose_name_plural}" that is on the same row as the breadcrumbs, but not in the list view. We should 100% align these.
> 
> Example:
> 
> ams01-dist-01_-_Nautobot
> 
> This is coded here:
> 
> https://github.com/nautobot/nautobot/blob/f99b59954bc14338804388f3a9ff8e3b0456243b/nautobot/core/templates/generic/object_retrieve.html#L30
> 
> Perhaps the global search box in the nav bar could say "Global Search" (it currently only says "Search"), and the list view search could say "Search `{verbose_name_plural}`", so that it's clear they are different?

@timizuoebideri1 how difficult would it be to put the breadcrumb bar on the list view (don't know how difficult the breadcrumbs themselves would be so could be done later) and putting the input in that bar like @jathanism's screenshot?

jathanism:
> > > > This is pretty great, my only concern may be with where it is placed. Currently we have a search box on the top of every page to the right of the nav bar, and this appears to duplicate it, so we need to be cautious about the UX so that it isn't confusing.
> > > 
> > > 
> > > @jathanism What do you suggest
> > 
> > 
> > ~I think previously we had a grayed out text in the search box that said something like `Search for `{verbose_name_plural}` but that no longer seems to be the case.~ For some reason in our object detail view template we have this search box that has "Search {verbose_name_plural}" that is on the same row as the breadcrumbs, but not in the list view. We should 100% align these.
> > Example:
> > ams01-dist-01_-_Nautobot
> > This is coded here:
> > https://github.com/nautobot/nautobot/blob/f99b59954bc14338804388f3a9ff8e3b0456243b/nautobot/core/templates/generic/object_retrieve.html#L30
> > 
> > Perhaps the global search box in the nav bar could say "Global Search" (it currently only says "Search"), and the list view search could say "Search `{verbose_name_plural}`", so that it's clear they are different?
> 
> @timizuoebideri1 how difficult would it be to put the breadcrumb bar on the list view (don't know how difficult the breadcrumbs themselves would be so could be done later) and putting the input in that bar like @jathanism's screenshot?

Seems to me the correct solution would be to always show the breadcrumb bar similarly to how it's done in the browsable API. Example: Cursor_and_Device_–_Django_REST_framework

(It's worth nothing that the plural name would be nice to see on the browsable API as well and make the breadcrumbs look the same, too, but that's a separate issue.)

jathanism (approved):

nautobot/core/templates/generic/object_list.html
```diff
@@ -17,6 +17,38 @@
...
+                </a></li>
+                {% endif %}
+                {% block extra_breadcrumbs %}{% endblock extra_breadcrumbs %}
+            {% endblock breadcrumbs %}
+            </ol>
+        </div>
+        <div class="col-sm-4 col-md-3">
+            <form action="#" method="get" id="search-form">
+                <div class="input-group">
+                    {{ search_form.q }}
```
glennmatthews, line 38: Do we really need a `search_form` at all, given we're really only just using this widget? Seems like it might be simpler to hard-code it into the template, similar to what we're already doing in object_retrieve.html, a la:

```
<input type="text" name="q" class="form-control" placeholder="Search {{ content_type.model_class|meta:'verbose_name_plural' }}">
```

nautobot/core/templates/generic/object_list.html
```diff
@@ -163,7 +195,38 @@ <h1>{% block title %}{{ content_type.model_class|meta:"verbose_name_plural"|bett
...
-        location.replace("?" + $(this).serialize())
+
+        let query_params = $(this).serialize()
+
+        // check if `?q=` or `&q=` in query_params
+        if(location.search && /(\?|\&)q/.test(location.search)){
+            // Append `q` value to `query_params`
+            query_params += "&q=" + /(?<=(\?|\&)q=)\w+/.exec(location.search)[0]
+        }
+        location.replace("?" + query_params)
```
glennmatthews, line 206: This is confusing to me, at least based on the comments... if q is in the query_params, we append it to the query_params?

nautobot/core/templates/generic/object_list.html
```diff
@@ -163,7 +195,38 @@ <h1>{% block title %}{{ content_type.model_class|meta:"verbose_name_plural"|bett
...
+        let search_value = $(this).find("input[name='q']").val();
+
+        if(query_params){
+            // check if `?q=` or `&q=` in query_params
+            if(/(\?|\&)q/.test(query_params)){
+                // change value of q in the query string
+                query_params = query_params.replace(/(?<=(\?|\&)q=)\w+/, search_value);
+            }
+            else {
+                query_params += "&q=" + search_value;
```
glennmatthews, line 223: Couldn't this be, for more self-consistency:

```suggestion
                query_params += "&" + $(this).serialize();
```

nautobot/core/templates/generic/object_list.html
```diff
@@ -163,7 +195,38 @@ <h1>{% block title %}{{ content_type.model_class|meta:"verbose_name_plural"|bett
         // By removing all name attr with 'form-' prefix form input/select elements
         // only filterable fields would be picked up by `$(this).serialize()`
         $(this).find(`input[name*="form-"], select[name*="form-"]`).removeAttr("name")
-        location.replace("?" + $(this).serialize())
+
+        let query_params = $(this).serialize()
+
+        // check if `?q=` or `&q=` in query_params
```
timizuoebideri1, line 201:
```suggestion
        // check if `?q=` or `&q=` in URL query params
```

nautobot/core/templates/generic/object_list.html
timizuoebideri1: updated the comment: check if `?q=` or `&q=` in  url query params

nautobot/core/templates/generic/object_list.html
```diff
@@ -163,7 +195,38 @@ <h1>{% block title %}{{ content_type.model_class|meta:"verbose_name_plural"|bett
...
+    $("#search-form").on("submit", function(e){
+        e.preventDefault();
+
+        let query_params = location.search;
+        let search_value = $(this).find("input[name='q']").val();
+
+        if(query_params){
+            // check if `?q=` or `&q=` in query_params
+            if(/(\?|\&)q/.test(query_params)){
+                // change value of q in the query string
```
timizuoebideri1, line 219:
```suggestion
                // change the value of `q` in the `query_params`
```

nautobot/core/templates/generic/object_list.html
timizuoebideri1: With this method, I will have to manually handle the input value for "q" if the query string contains it.

glennmatthews: Fair point. Aren't we already doing something like that for all the other query parameters in the parent branch though?

timizuoebideri1: No

bryanculver (approved):

Issue #2573, opened 2022-10-05
Invalid Link at VLAN Group


johannwagner:
### Environment
* Nautobot version (Docker tag too if applicable): v1.4.4
* Python version: Demo-Instance
* Database platform, version: Demo-Instance
* Middleware(s): Demo-Instance

### Steps to Reproduce
1. Create a VLAN Group
2. Click on link on detail page to see all vlans

### Expected Behavior
I expect to see all VLANs assigned to this VLAN group.

### Observed Behavior
I get an error message, that this is not a valid search request.

### Screenshots

![Screenshot from 2022-10-05 10-32-33](https://user-images.githubusercontent.com/12380026/194017188-83586be9-6826-4977-b411-f9b4badd599a.png)
![Screenshot from 2022-10-05 10-34-42](https://user-images.githubusercontent.com/12380026/194017289-c5eec360-397c-4bf3-92b6-3430ff681d3d.png)

johannwagner:
I would be happy to provide a PR. 
Question: Is `group` as a search parameter correct and the filter page is borked or is `group_id` as a search parameter correct and the VLAN Group page is borked?

glennmatthews:
Thanks for the report, and for the offer of a fix!

Both filters are supported, but `group` filters on group `slug`, and `group_id` filters on group `id`. Looks like we have a simple mismatch here (using an `id` with `group`); I have no strong preference as to which direction you go in fixing it (either `?group_id=<uuid>` or `?group=<slug>` would be fine).

PR #2576, opened 2022-10-05
Fix broken link in vlan group template

johannwagner:
# Closes: #2573 
# What's Changed
- Changing a link from an id to a slug to match filter form


johannwagner:
This actually does introduce a different problem. The implemented solution does filter correctly but does not show the filter in the sidebar. I will provide a different implementation.

johannwagner:
This now shows the selected VLAN group in the filter form on the right side.

changes/2573.fixed
```diff
@@ -0,0 +1 @@
+Fixing link to VLAN list to match filter form.
```
glennmatthews, line 1: Perhaps?

```suggestion
Fixed link from VLAN group detail view to filtered VLAN list.
```

glennmatthews (approved):
Thank you!

PR #2540, opened 2022-09-30
factory_boy test fixture: Region, Site, Location, LocationType

HanlinMiao:
# Closes: #2282
# What's Changed

Added factory boy fixture for `Site`, `Region`, `Location`, and `LocationType`.
Changed unittest to use those fixtures as much as possible.
In some `test_filter` unittests `assertEqual` statement, I used` queryset.filter()` to obtain the number of instances qualified.
In some I have to hardcode the number of instances.
There are some complex test cases that involve creating new `Sites`, `Region`, `Location`, and `LocationType` with distinct attributes and I decided to leave those test cases alone for now.

# TODO

- [x] Region, Site, Location, LocationType Factories
- [x] Address broken tests

PR #2579, opened 2022-10-05
Add prefix and ipaddress test fixtures

gsnider2195:
# Closes: #2283
# What's Changed

- Add factoryboy factories for `Prefix` and `IPAddress` models
- Add `Prefix` as a subfactory of `Aggregate` to create prefixes inside the aggregate's IP space
- Add `IPAddress` as a subfactory of non-container prefixes to create IP addresses inside the prefix's IP space
- Add `Prefix` as a subfactory of container prefixes
- Force `AggregateFactory` to create at least 5 aggregates with tenant_group and 5 with a `tenant` and no `tenant_group` to have adequate test data
- Update tests where possible to remove static test objects
  - Some of these were left alone because they would impact performance or be too complex to use randomized test data (ex: creating all available IP addresses in a prefix with a billion IPs would be slow)
- Move queryset filter tests from `nautobot.ipam.tests.test_filters.py` to `nautobot.ipam.tests.test_querysets.py`
- Update generic test util `nautobot.utilities.testing.utils.post_data()` to properly nullify fields in API tests
- Add generic test util `nautobot.utilities.testing.utils.get_deletable_objects()` to allow `DeleteObjectViewTestCase` and `BulkDeleteObjectsViewTestCase` to retrieve objects with no protected relationships that would prevent deletion
- Increase the `PAGINATE_COUNT` setting on `nautobot.utilities.testing.TestCase` to account for the increased number of test objects in the database
- Add a new helper method to `RestrictedQuerySet` called `distinct_values_list`
- Add a new assertion method to `NautobotTestCaseMixin` called `AssertQuerysetEqualAndNotEmpty`

# TODO

- [x] Unit, Integration Tests


nautobot/ipam/factory.py
```diff
@@ -54,6 +72,45 @@ class Meta:
...
+
+        # Raise exception for invalid cidr length (>128 for ipv6, >32 for ipv4)
+        if prefix_cidr > 128 or self.family == 4 and prefix_cidr > 32:
+            raise ValueError(f"Unable to create {prefix_count} prefixes in aggregate {self.cidr_str}")
+
+        # Create child prefixes, preserving tenant and is_ipv6 from aggregate
+        for count, subnet in enumerate(self.prefix.subnet(prefix_cidr)):
+            if count == max_count:
+                break
+            method(prefix=str(subnet.cidr), tenant=self.tenant, is_ipv6=is_ipv6, **kwargs)
```
glennmatthews, line 112: Since Aggregate.tenant is optional, maybe allow prefixes of a non-tenanted Aggregate to have random tenants of their own? Something like:

```suggestion
            if self.tenant is not None:
                method(prefix=str(subnet.cidr), tenant=self.tenant, is_ipv6=is_ipv6, **kwargs)
            else:
                method(prefix=str(subnet.cidr), is_ipv6=is_ipv6, **kwargs)
```

nautobot/ipam/factory.py
```diff
@@ -304,3 +361,184 @@ def _adjust_kwargs(cls, **kwargs):
...
+    """
+    Create random prefixes and 50% of the time generate child prefixes and ip addresses within the prefix IP space.
+    Containers will create child prefixes while prefixes that are not containers will create ip addresses in the
+    prefix's address space. Defaults to creating 0-4 children. Set children__max_count to an integer when calling
+    the factory creation methods (`create()`, `create_batch()`, etc) to override the maximum number of children
+    generated.
+
+    Examples:
+    PrefixFactory.create_batch(20) # create 20 prefixes, approximately half will generate 0-4 children
+    AggregateFactory.create_batch(children__max_count=0) # create 20 prefixes with no children
```
glennmatthews, line 386:
```suggestion
    PrefixFactory.create_batch(children__max_count=0) # create 20 prefixes with no children
```

nautobot/ipam/factory.py
```diff
@@ -304,3 +361,184 @@ def _adjust_kwargs(cls, **kwargs):
...
+    class Params:
+        has_description = factory.Faker("pybool")
+        has_location = factory.Faker("pybool")
+        has_role = factory.Faker("pybool")
+        has_site = factory.Faker("pybool")
+        has_tenant = factory.Faker("pybool")
+        has_vlan = factory.Faker("pybool")
+        has_vrf = factory.Faker("pybool")
+        is_container = factory.Faker("pybool")
+        is_ipv6 = factory.Faker("pybool")
```
glennmatthews, line 401: Did we identify any tradeoff between the `Meta.exclude` approach versus the `Params` approach for these flags? We should settle on one and be consistent about it.

nautobot/ipam/factory.py
```diff
@@ -304,3 +361,184 @@ def _adjust_kwargs(cls, **kwargs):
...
+        child_count = faker.Faker().pyint(min_value=0, max_value=min(max_count, self.prefix.size))
+        if child_count == 0:
+            return
+
+        if factory == IPAddressFactory:
+            # Create child ip addresses, preserving vrf, tenant and is_ipv6 from parent
+            for count, address in enumerate(self.prefix.iter_hosts()):
+                if count == child_count:
+                    break
+                method(address=str(address), vrf=self.vrf, tenant=self.tenant, is_ipv6=is_ipv6, **kwargs)
```
glennmatthews, line 479: Similar thought here about non-tenanted Prefixes getting to have child IPs that have a random tenant?

nautobot/ipam/factory.py
```diff
@@ -304,3 +361,184 @@ def _adjust_kwargs(cls, **kwargs):
...
+        max_count = int(kwargs.pop("max_count", 4))
+        child_count = faker.Faker().pyint(min_value=0, max_value=min(max_count, self.prefix.size))
+        if child_count == 0:
+            return
+
+        if factory == IPAddressFactory:
+            # Create child ip addresses, preserving vrf, tenant and is_ipv6 from parent
+            for count, address in enumerate(self.prefix.iter_hosts()):
+                if count == child_count:
+                    break
```
glennmatthews, line 478: This will always give us a contiguous block of allocated IPs, right? Wonder if we could add some coin-flip logic to randomly skip ahead, something like:
```suggestion
            count = 0
            for address in self.prefix.iter_hosts():
                if not faker.Faker().pybool():
                    continue
                count += 1
                if count == child_count:
                    break
```

nautobot/ipam/factory.py
```diff
@@ -304,3 +361,184 @@ def _adjust_kwargs(cls, **kwargs):
...
+            # Create child prefixes, preserving tenant, site, location, vrf and is_ipv6 from parent
+            for count, address in enumerate(self.prefix.subnet(child_cidr)):
+                if count == child_count:
+                    break
+                method(
+                    prefix=str(address.cidr),
+                    tenant=self.tenant,
+                    site=self.site,
+                    location=self.location,
+                    is_container=False,
```
glennmatthews, line 497: No containers of containers?

nautobot/ipam/factory.py
```diff
@@ -304,3 +361,184 @@ def _adjust_kwargs(cls, **kwargs):
...
+    nat_inside = factory.SubFactory(
+        "nautobot.ipam.factory.IPAddressFactory",
+        nat_inside=None,
+        is_ipv6=factory.SelfAttribute("..is_ipv6"),
+    )
+    role = factory.Maybe("has_role", factory.LazyAttribute(lambda obj: obj.role_choice[0]), "")
+    status = factory.Maybe(
+        "is_ipv6",
+        random_instance(lambda: Status.objects.get_for_model(IPAddress), allow_null=False),
+        random_instance(lambda: Status.objects.get_for_model(IPAddress).exclude(name="SLAAC"), allow_null=False),
```
glennmatthews, line 537: :+1: Good catch.

nautobot/ipam/tests/test_filters.py
```diff
@@ -222,215 +239,150 @@ class PrefixTestCase(FilterTestCases.FilterTestCase, FilterTestCases.TenancyFilt
...
         for value in test_values:
             params = {"q": value}
-            self.assertEqual(self.filterset(params, self.queryset).qs.count(), 1)
+            count = self.queryset.string_search(value).count()
+            self.assertEqual(self.filterset(params, self.queryset).qs.count(), count)
 
     def test_family(self):
         params = {"family": "6"}
-        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 5)
+        ipv6_prefixes = [prefix for prefix in Prefix.objects.all() if prefix.family == 6]
```
glennmatthews, line 256: Can we do `Prefix.objects.ip_family(6)`?

nautobot/ipam/tests/test_filters.py
```diff
@@ -222,215 +239,150 @@ class PrefixTestCase(FilterTestCases.FilterTestCase, FilterTestCases.TenancyFilt
...
+        self.assertEqual(self.filterset(params, self.queryset).qs.count(), len(ipv6_prefixes))
 
     def test_is_pool(self):
         params = {"is_pool": "true"}
-        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 2)
+        self.assertEqual(self.filterset(params, self.queryset).qs.count(), Prefix.objects.filter(is_pool=True).count())
```
glennmatthews, line 261: Can we change these to `assertQuerysetEqual`?

nautobot/ipam/tests/test_views.py
```diff
@@ -32,10 +32,7 @@ class VRFTestCase(ViewTestCases.PrimaryObjectViewTestCase):
     @classmethod
     def setUpTestData(cls):
 
-        tenants = (
-            Tenant.objects.create(name="Tenant A", slug="tenant-a"),
-            Tenant.objects.create(name="Tenant B", slug="tenant-b"),
-        )
+        tenants = Tenant.objects.all()[:2]
```
glennmatthews, line 35: :+1: Not sure how I missed these in the initial tenant factory PR - thanks!

nautobot/ipam/tests/test_views.py
```diff
@@ -266,10 +225,10 @@ def test_empty_queryset(self):
         but the same behavior was observed in other filters, such as IPv4/IPv6.
         """
         prefixes = self._get_queryset().all()
-        self.assertEqual(prefixes.count(), 3)
+        self.assertNotEqual(prefixes.count(), 0)
 
         url = self._get_url("list")
-        response = self.client.get(f"{url}?status=deprecated")
+        response = self.client.get(f"{url}?status=nonexistentstatus")
```
glennmatthews, line 231: This will trip over STRICT_FILTERING won't it? Maybe better to create a one-off Status that is valid for Prefix but doesn't have any assigned to it.

glennmatthews (commented):
Great progress!

nautobot/ipam/factory.py
```diff
@@ -20,12 +30,26 @@ class Meta:
...
+    methods (`create()`, `create_batch()`, etc) to override the maximum number of child prefixes generated. Set
+    child_prefixes__children__max_count to an integer when calling the factory creation methods (`create()`, `batch_create()`, etc)
+    to override the maximum number of grandchildren generated.
+
+    Examples:
+    AggregateFactory.create_batch(20) # create 20 aggregates, approximately half will generate 0-4 child prefixes
+                                      # which will create child prefixes and ip addresses
+    AggregateFactory.create_batch(child_prefixes__max_count=0) # create 20 aggregates with no child prefixes
+    AggregateFactory.create_batch(child_prefixes__children__max_count=0) # create 20 aggregates, approximately
+                                      # half will generate 0-4 child prefixes that will not create any children
```
jathanism, line 50: This reads a little funky. Could we instead have each call on a line of its own with the text preceding it instead of being inline as a comment?

Something like this:


```suggestion

    Create 20 Aggregates, approximately half will generate 0-4 child prefixes which will create child prefixes and ip addresses:
    
        >>> AggregateFactory.create_batch(20)
    
    Create 20 Aggregates with no child prefixes:
    
        >>> AggregateFactory.create_batch(child_prefixes__max_count=0)
    
    Create 20 Aggregates, approximately half will generate 0-4 child prefixes that will not create any children:
    
        >>> AggregateFactory.create_batch(child_prefixes__children__max_count=0) 
```

nautobot/ipam/factory.py
```diff
@@ -44,19 +68,48 @@ class Meta:
...
-    # TODO custom field data?
-
     @factory.post_generation
-    def tags(self, create, extracted, **kwargs):
-        if create:
-            if extracted:
-                self.tags.set(extracted)
-            else:
-                self.tags.set(get_random_tags_for_model(self._meta.model))
+    def child_prefixes(self, create, extracted, **kwargs):
```
jathanism, line 76: Can you please document the arguments?

nautobot/ipam/factory.py
```diff
@@ -111,3 +164,381 @@ def prefix(self, n):
...
+    @factory.post_generation
+    def import_targets(self, create, extracted, **kwargs):
+        if create:
+            if extracted:
+                self.import_targets.set(extracted)
+            else:
+                self.import_targets.set(get_random_instances(RouteTarget))
+
+    @factory.post_generation
+    def export_targets(self, create, extracted, **kwargs):
```
jathanism, line 242: If these are user-facing methods, can you please add docstrings for them?

nautobot/ipam/factory.py
```diff
@@ -111,3 +164,381 @@ def prefix(self, n):
...
+    """
+    Create random prefixes and 50% of the time generate child prefixes and ip addresses within the prefix IP space.
+    Containers will create child prefixes while prefixes that are not containers will create ip addresses in the
+    prefix's address space. Defaults to creating 0-4 children. Set children__max_count to an integer when calling
+    the factory creation methods (`create()`, `create_batch()`, etc) to override the maximum number of children
+    generated.
+
+    Examples:
+    PrefixFactory.create_batch(20) # create 20 prefixes, approximately half will generate 0-4 children
+    PrefixFactory.create_batch(children__max_count=0) # create 20 prefixes with no children
```
jathanism, line 386: Similar feedback for the example formatting as for `AggregateFactory`, pls.

nautobot/ipam/factory.py
```diff
@@ -111,3 +164,381 @@ def prefix(self, n):
...
+        None,
+    )
+    vrf = factory.Maybe(
+        "has_vrf",
+        factory.SubFactory(VRFGetOrCreateFactory, tenant=factory.SelfAttribute("..tenant")),
+        None,
+    )
+
+    @factory.post_generation
+    def children(self, create, extracted, **kwargs):
```
jathanism, line 445: arg docs pls

nautobot/ipam/factory.py
```diff
@@ -111,3 +164,381 @@ def prefix(self, n):
...
+                    site=self.site,
+                    location=self.location,
+                    is_container=False,
+                    is_ipv6=is_ipv6,
+                    vrf=self.vrf,
+                    **kwargs,
+                )
+
+
+class IPAddressFactory(PrimaryModelFactory):
```
jathanism, line 504: Seems like we'll want a docstring for this factory as well in keeping with `AggregateFactory` and `PrefixFactory`?

nautobot/ipam/tests/test_filters.py
```diff
@@ -49,86 +50,79 @@
...
-        vrfs[1].export_targets.add(route_targets[1])
-        vrfs[2].import_targets.add(route_targets[2])
-        vrfs[2].export_targets.add(route_targets[2])
+    tenancy_related_name = "vrfs"
+
+    # Note: all assertQuerySetEqual() calls here must use ordered=False,
+    # because order_by=["name", "rd"] but name is not globally unique and rd can be null,
+    # so relative ordering of VRFs with identical name and null rd is not guaranteed.
+
+    def test_gary(self):
```
jathanism, line 62: 🤨

nautobot/ipam/tests/test_filters.py
```diff
@@ -49,86 +50,79 @@
...
-        vrfs[0].export_targets.add(route_targets[0])
-        vrfs[1].import_targets.add(route_targets[1])
-        vrfs[1].export_targets.add(route_targets[1])
-        vrfs[2].import_targets.add(route_targets[2])
-        vrfs[2].export_targets.add(route_targets[2])
+    tenancy_related_name = "vrfs"
+
+    # Note: all assertQuerySetEqual() calls here must use ordered=False,
+    # because order_by=["name", "rd"] but name is not globally unique and rd can be null,
+    # so relative ordering of VRFs with identical name and null rd is not guaranteed.
```
jathanism, line 60: This seems like it might be better served as a part of the docstring for this test-case so that it doesn't get overlooked.

nautobot/ipam/tests/test_filters.py
```diff
@@ -277,304 +222,180 @@ def test_prefix(self):
...
-        params = {"vlan_vid": vlans[0].vid}
-        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 2)
 
-    def test_role(self):
-        roles = Role.objects.all()[:2]
-        params = {"role_id": [roles[0].pk, roles[1].pk]}
-        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 4)
-        params = {"role": [roles[0].slug, roles[1].slug]}
-        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 4)
+# def test_site(self):
```
jathanism, line 362: Are these deprecated or will these be restored?

nautobot/ipam/tests/test_ordering.py
```diff
@@ -1,314 +1,34 @@
...
-                status=self.status_active,
-                vrf=vrfb,
-                prefix=netaddr.IPNetwork("172.17.4.0/24"),
-            ),
+        prefixes = sorted(
+            Prefix.objects.filter(vrf__isnull=True),
+            key=attrgetter("prefix.network.packed", "prefix_length"),
+        ) + sorted(
+            Prefix.objects.filter(vrf__isnull=False),
+            key=attrgetter("vrf.name", "prefix.network.packed", "prefix_length"),
```
jathanism, line 30: \**chef's kiss*\*

nautobot/ipam/factory.py
gsnider2195: I don't see any functional difference. Neither of them are available from a `post_generation` decorated function and both of them are available using the `factory.SelfAttribute` class from within `Params` and factory attributes.

nautobot/ipam/factory.py
gsnider2195: With our current numbers (average 0.25 child containers created per parent prefix) the subfactories are safe from infinite loops. It actually works pretty well as-is if I remove this line but we'll need some extra logic if we want this to be flexible enough to be able to increase the `children__max_count` and not end up with an infinite loop.

nautobot/ipam/tests/test_views.py
gsnider2195: Confirmed that it does break on "Invalid filters were specified"

nautobot/ipam/tests/test_filters.py
gsnider2195: This file wasn't ready for review yet

nautobot/ipam/tests/test_filters.py
```diff
@@ -222,215 +226,184 @@ class PrefixTestCase(FilterTestCases.FilterTestCase, FilterTestCases.TenancyFilt
...
+        params = {"site_id": [test_sites[0].pk, test_sites[1].pk]}
+        self.assertQuerysetEqual(
+            self.filterset(params, self.queryset).qs,
+            self.queryset.filter(site__in=[test_sites[0], test_sites[1]]),
+        )
+        params = {"site": [test_sites[0].slug, test_sites[1].slug]}
+        self.assertQuerysetEqual(
+            self.filterset(params, self.queryset).qs,
+            self.queryset.filter(site__in=[test_sites[0], test_sites[1]]),
+        )
```
glennmatthews, line 386: These will change/simplify once we have #2540, I presume?

nautobot/ipam/tests/test_filters.py
```diff
@@ -579,70 +555,88 @@ def setUpTestData(cls):
...
+            ipv6_hextets[0],  # 2001
+            f"{ipv6_hextets[0]}:",  # 2001:
+            f"{ipv6_hextets[0]}::",  # 2001::
+            "no match",
+        ]
+
+        for term in search_terms:
+            with self.subTest(term):
+                params = {"q": term}
+                self.assertQuerysetEqual(self.filterset(params, self.queryset).qs, self.queryset.string_search(term))
```
glennmatthews, line 582: Wish we could make this one a bit more robust - as written it'll pass if `string_search()` returns no matches for every term, right?

I wonder if we might want to create a custom test method like `assertQuerysetEqualAndNotEmpty()` to use in all of the filter tests (not just those in this PR)?

It's less of a concern for tests where we're using standard queryset filters (`vrf__isnull=False` and the like) as it'd be pretty weird for those to not work as expected, but for all of these ones involving custom filters (`string_search`, `net_*`) I'm concerned.

glennmatthews (commented):
This is looking really promising!

nautobot/ipam/tests/test_filters.py
gsnider2195: Yes. PrefixFactory is already setup to grab a random site when they're available.

gsnider2195: (Site and Location)

nautobot/ipam/tests/test_filters.py
gsnider2195: I was thinking the same when adding this line to the test but ended up going this route anyway because there are already tests to test that `string_search` is working properly in [test_querysets.py](https://github.com/nautobot/nautobot/blob/0333de62a936d15761e8725747d7d4045a31d7df/nautobot/ipam/tests/test_querysets.py#L76). Since this is testing that the filterset is working properly, I was satisfied that the combination of these tests and the next test that's making sure a blank query returns all objects would be sufficient. I wrote this test using an existing object's properties to ensure that even with randomized data the filter should always return at least 1 object but an `assertQuerysetEqualAndNotEmpty()` would help to enforce that.

nautobot/ipam/tests/test_api.py
```diff
@@ -192,114 +192,106 @@ class PrefixTest(APIViewTestCases.APIViewTestCase):
...
-            is_pool=True,
-            status=self.status_active,
-        )
+        # Find prefix with no child prefixes and large enough to create 4 child prefixes
+        prefix = None
+        for instance in Prefix.objects.filter(vrf__isnull=False):
+            if instance.get_child_prefixes().count() == 0 and instance.prefix.size > 2:
+                prefix = instance
+                break
+        self.assertNotEqual(None, prefix, "Suitable prefix fixture not found")
```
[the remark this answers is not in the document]
gsnider2195: Thanks for the suggestion. I didn't like using an assertion to fail the test setup.

nautobot/ipam/tests/test_filters.py
glennmatthews: I filed #2633 to track the `assertQuerysetEqualAndNotEmpty` idea.

nautobot/utilities/testing/mixins.py
```diff
@@ -182,6 +182,17 @@ def assertInstanceEqual(self, instance, data, exclude=None, api=False):
...
 
+    def assertQuerysetEqualAndNotEmpty(self, qs, values, *args, **kwargs):
+        """Wrapper for assertQuerysetEqual with additional logic to assert input queryset and values are not empty"""
+
+        if isinstance(qs, QuerySet):
+            qs_count = qs.count()
+        else:
+            qs_count = len(qs)
+        self.assertNotEqual(qs_count, 0, "Queryset cannot be empty")
+        self.assertNotEqual(len(values), 0, "Values cannot be empty")
```
glennmatthews, line 193: Should `values` also have the `isinstance(QuerySet)` check for efficiency?

glennmatthews (approved):

PR #2593, opened 2022-10-07
Replace tag/status fixtures with factories

glennmatthews:
# Closes: #DNE
# What's Changed

- Introduce TagFactory and StatusFactory
- Delete static test fixtures for Tag and Status
- Remove documentation and tooling for working with static test fixtures since we've changed direction.
- Update VLANFactory.name and VRFFactory.name to reduce likelihood of the names appearing by coincidence in page HTML and causing test cases failures as a consequence.
- Revert changes to `clear_status_choices` that were made in #2514 and instead just remove the tests for this function.

# TODO
- [x] Unit, Integration Tests


nautobot/extras/factory.py
```diff
@@ -0,0 +1,61 @@
...
+
+# Status isn't technically an OrganizationalModel, but it has all of its features **except** dynamic-groups
+class StatusFactory(OrganizationalModelFactory):
+    class Meta:
+        model = Status
+        exclude = ("has_description",)
+
+    name = factory.LazyFunction(
+        lambda: "".join(word.title() for word in faker.Faker().words(nb=2, part_of_speech="adjective", unique=True))
+    )
```
jathanism, line 22: This pattern for joining words is cleaner than the one using inner `Params` class IMO. Compare to what I did for `PlatformFactory`:

```python
    class Params:
        first = UniqueFaker("word")
        last = UniqueFaker("word")

    # Slug isn't defined here since it will always inherit from name.
    name = factory.LazyAttribute(lambda o: f"{o.first} {o.last}".title())
```

nautobot/extras/factory.py
```diff
@@ -0,0 +1,61 @@
...
+
+from nautobot.core.factory import OrganizationalModelFactory
+from nautobot.extras.models import Status, Tag
+from nautobot.extras.utils import FeatureQuery, TaggableClassesQuery
+from nautobot.utilities.choices import ColorChoices
+from nautobot.utilities.factory import get_random_instances
+
+
+# Status isn't technically an OrganizationalModel, but it has all of its features **except** dynamic-groups
+class StatusFactory(OrganizationalModelFactory):
```
jathanism, line 15: IMO make this comment a docstring instead so that it carries over in IDEs and stuff.

nautobot/extras/factory.py
```diff
@@ -0,0 +1,61 @@
...
+            else:
+                self.content_types.set(
+                    get_random_instances(
+                        lambda: ContentType.objects.filter(FeatureQuery("statuses").get_query()), minimum=1
+                    )
+                )
+
+
+# Tag isn't technically an OrganizationalModel, but it has all of its features **except** dynamic-groups
+class TagFactory(OrganizationalModelFactory):
```
jathanism, line 43: IMO make this comment a docstring instead so that it carries over in IDEs and stuff.

nautobot/ipam/tests/test_filters.py
```diff
@@ -837,12 +834,14 @@ def setUpTestData(cls):
         )
 
     def test_name(self):
-        params = {"name": VLAN.objects.all().values_list("name", flat=True)[:2]}
-        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 2)
+        names = list(VLAN.objects.all().values_list("name", flat=True))[:2]
```
jathanism, line 837: I'm realizing in cases like this it's gonna be a tad more optimal to slice the queryset first, and cast to a list second, assuming we have more than a handful of VLAN objects, that is.

For example:

```suggestion
        names = list(VLAN.objects.all().values_list("name", flat=True)[:2])
```

Perhaps since we're starting to repeat this pattern a lot as we refactor to use the fixture factories, we might consider factoring it into a test case helper method? Just thinking aloud. Please don't change it right now.

nautobot/utilities/factory.py
```diff
@@ -72,7 +72,7 @@ def get_random_instance():
     return factory.LazyFunction(get_random_instance)
 
 
-def get_random_instances(model_or_queryset_or_lambda, maximum=None):
+def get_random_instances(model_or_queryset_or_lambda, minimum=0, maximum=None):
```
jathanism, line 75: Please don't forget to revise the docstring with the new signature! :)

jathanism (commented):
👀 
image

nautobot/extras/factory.py
glennmatthews: Sorry, I'm having a bit of trouble parsing this. Are you saying you'd prefer for me to use the `Params` approach, or are you saying this is fine as-is?

nautobot/ipam/tests/test_filters.py
glennmatthews: Good thought. Since we've already got this pattern in a bunch of tests after #2479 and #2514, maybe we should get a housekeeping item open to clean up all of these?

nautobot/ipam/factory.py
```diff
@@ -158,7 +158,7 @@ class Meta:
         )
 
     # note that name is *not* globally unique
-    name = factory.Faker("color_name")
+    name = factory.LazyFunction(lambda: f"{faker.Faker().color_name()}{faker.Faker().pyint()}")
```
gsnider2195, line 161: Would this make sense to be a [LazyAttributeSequence](https://factoryboy.readthedocs.io/en/stable/reference.html#lazyattributesequence)?

nautobot/ipam/tests/test_filters.py
jathanism: Solid plan.

nautobot/extras/factory.py
```diff
@@ -0,0 +1,63 @@
...
+    """Status isn't technically an OrganizationalModel, but it has all of its features **except** dynamic-groups."""
+
+    class Meta:
+        model = Status
+        exclude = ("has_description",)
+
+    name = factory.LazyFunction(
+        lambda: "".join(word.title() for word in faker.Faker().words(nb=2, part_of_speech="adjective", unique=True))
+    )
+    slug = factory.LazyAttribute(lambda tag: slugify(tag.name))
```
gsnider2195, line 24:
```suggestion
    slug = factory.LazyAttribute(lambda status: slugify(status.name))
```

nautobot/extras/factory.py
jathanism: Oh sorry, I was saying I find this approach to be better/cleaner than what I was attempting. Sorry about the confusion.

nautobot/ipam/factory.py
glennmatthews: Could do, then the names would be (random color + sequential number) instead of (random color + random number). I don't feel strongly about one versus the other, I just needed to add something because color names alone weren't proving to be unambiguous enough.

nautobot/ipam/tests/test_filters.py
```diff
@@ -837,12 +834,14 @@ def setUpTestData(cls):
...
+        params = {"name": names}
+        self.assertQuerysetEqual(self.filterset(params, self.queryset).qs, self.queryset.filter(name__in=names))
 
     def test_vid(self):
-        params = {"vid": VLAN.objects.all().values_list("vid", flat=True)[:3]}
-        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 3)
+        vids = list(VLAN.objects.all().values_list("vid", flat=True))[:3]
```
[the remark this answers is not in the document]
glennmatthews: Without casting to a list, `values_list` returns a QuerySet. This is fine for the `self.filterset` call, but on MySQL the `self.queryset.filter()` call chokes if given a sliced QuerySet (rather than a list), e.g.:

```python
>>> VLAN.objects.filter(vid__in=VLAN.objects.all().values_list("vid", flat=True)[:3])
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/mysql/base.py", line 73, in execute
    return self.cursor.execute(query, args)
  File "/usr/local/lib/python3.7/site-packages/MySQLdb/cursors.py", line 206, in execute
    res = self._query(query)
  File "/usr/local/lib/python3.7/site-packages/MySQLdb/cursors.py", line 319, in _query
    db.query(q)
  File "/usr/local/lib/python3.7/site-packages/MySQLdb/connections.py", line 254, in query
    _mysql.connection.query(self, query)
MySQLdb.NotSupportedError: (1235, "This version of MySQL doesn't yet support 'LIMIT & IN/ALL/ANY/SOME subquery'")
```

It's fine if you cast it to a list first though:

```python
>>> VLAN.objects.filter(vid__in=list(VLAN.objects.all().values_list("vid", flat=True)[:3]))
<RestrictedQuerySet [<VLAN: Layer (274)>, <VLAN: Detail (359)>, <VLAN: Wind (635)>]>
```

nautobot/ipam/factory.py
gsnider2195: My thought was this line would be easier to read as:
```py
name = factory.LazyAttributeSequence(lambda n: f"{faker.Faker().color_name()}{n}")
```

gsnider2195 (approved):

nautobot/ipam/factory.py
glennmatthews: Easier to read but guarantees that names are globally unique, which isn't actually a requirement of the model.

Issue #2595, opened 2022-10-08
UI ViewSet Default Unauthenticated Behavior Differs From Nautobot Patterns


whitej6:
### Environment
* Nautobot version (Docker tag too if applicable): 1.4.5
* Python version: 3.8
* Database platform, version: NA
* Middleware(s): NA

### Steps to Reproduce
1. As an unauthenticated user browse to a view that is registered via NautobotUIViewSetRouter
  * Can be verified [here](https://demo.nautobot.com/circuits/circuits/) as unauthenticated

### Expected Behavior
Redirect to login page with `next` query param for:
* list view
* detail view
* create view
* edit view

404 Response:
* notes view
* changelog view

### Observed Behavior
Page loads
* list view is rendered without any instances populated
* detail view returns 404
* create returns rendered form but no nested objects
* edit returns 404

404 Response:
* notes view
* changelog view

PR #2598, opened 2022-10-09
fixed NautobotUIViewSet Circuits App unauthenticated redirect error

HanlinMiao:
# Closes: #2595 
# What's Changed

1. Override default `dispatch()` method to `return self.handle_no_permission` (redirect to login_url) when the user is not logged in.
2. Improved unittest to test the redirect in `test_get_object_anonymous()` and `test_list_object_anonymous()`.
3. Changed the slug of the first secret_group to conform to slug conventions.

# TODO

- [ ] Unit, Integration Tests

https://user-images.githubusercontent.com/46973263/194735381-ea1ce82b-1954-4281-a9a5-db357e638d87.mp4


glennmatthews (commented):
Seems like we should have some test automation that would have caught this?

nautobot/core/views/mixins.py
```diff
@@ -120,13 +120,20 @@ def has_permission(self):
...
         Used to determine whether the user has permissions to a view and object-level permissions.
         Using AccessMixin handle_no_permission() to deal with Object-Level permissions and API-Level permissions in one pass.
         """
-        if not self.has_permission():
-            self.handle_no_permission()
+        # initialize_request also instantiates self.action which is needed for permission checks.
+        self.initialize_request(request, *args, **kwargs)
+        has_permission = self.check_permissions(request)
+
+        if not has_permission:
```
jathanism, line 133: Was this not possible with the previous method override for `check_permission()` but just fixing it to `return` the output of `self.handle_no_permission()` instead of just calling it?

I'm asking, because I'd like to see us avoid having to overload `dispatch()` if we can!

HanlinMiao: That was my initial approach but [this](https://github.com/encode/django-rest-framework/blob/56946fac8f29aa44ce84391f138d63c4c8a2a285/rest_framework/views.py#L415) is where drf calls self.check_permissions(). Unless we override initial() and do return self.check_permissions, self.handle_no_permission() (which is a redirect) does not do anything.

HanlinMiao: Lemme know if im making sense whatsoever

jathanism: Ah. Got it. Check permissions isn't meant to return anything.

jathanism (approved):

nautobot/core/views/mixins.py
```diff
@@ -101,7 +101,7 @@ def get_required_permission(self):
             )
         return self.get_permissions_for_model(queryset.model, permissions)
 
-    def has_permission(self):
+    def check_permissions(self, request):
```
glennmatthews, line 104: Is this renamed function now being used in the same sense as DRF? Looking at the views.py you linked, it looks like the base `check_permissions` is taking action (calling `self.permission_denied()`), not returning a boolean. What's the reason for renaming this?

jathanism: There doesn't even appear to be a `has_permission` method on any DRF views. So the correct path would be merging the functionality of the prior iteration into `check_permissions`, to have it call `self.permission_denied()` vs. returning a boolean.

jathanism: Reference:

```python
    def check_permissions(self, request):
        """
        Check if the request should be permitted.
        Raises an appropriate exception if the request is not permitted.
        """
        for permission in self.get_permissions():
            if not permission.has_permission(request, self):
                self.permission_denied(
                    request,
                    message=getattr(permission, 'message', None),
                    code=getattr(permission, 'code', None)
                )
```

glennmatthews: Hm. `self.permission_denied()` [raises an exception](https://github.com/encode/django-rest-framework/blob/master/rest_framework/views.py#L169) in DRF, which I'm not sure we have the correct handling in place for at this time?

jathanism: It should be handled by default as it raises the `PermissionDenied` exception which results in a 403.

HanlinMiao: like this?
```
def check_permissions(self, request):
        """
        Check whether the user has the permissions needed to perform certain actions.
        """
        user = self.request.user
        permission_required = self.get_required_permission()
        # Check that the user has been granted the required permission(s).
        if user.has_perms(permission_required):

            # Update the view's QuerySet to filter only the permitted objects
            for permission in permission_required:
                action = resolve_permission(permission)[1]
                queryset = queryset.restrict(user, action)
        else:
            self.permission_denied(
                request,
                message=getattr(permission_required, "message", None),
                code=getattr(permission_required, "code", None),
            )
```

HanlinMiao: and in dispatch:
    ```
def dispatch(self, request, *args, **kwargs):
        """
        Override the default dispatch() method to check permissions first.
        Used to determine whether the user has permissions to a view and object-level permissions.
        Using AccessMixin handle_no_permission() to deal with Object-Level permissions and API-Level permissions in one pass.
        """
        # initialize_request also instantiates self.action which is needed for permission checks.
        self.initialize_request(request, *args, **kwargs)
        try:
            self.check_permissions(request)
        except:
            return self.handle_no_permission()

        return super().dispatch(request, *args, **kwargs)

```

Comment (jathanism) [reply nautobot/core/views/mixins.py:103]: In the former, the `has_perms` stuff looks accurate. In the `dispatch` method, the only thing I don't like ks the naked `except`.

I am realizing we have a little confusion here, which should be clarified and documented for posterity.

1. My comment about `self.permission_denied()` raising `PermissionDenied` exception is the **DRF version of this exception** which when raised will return a JSON 403 DRF `Response` object. See: https://www.django-rest-framework.org/api-guide/exceptions/#permissiondenied

2. The method `self.handle_no_permission()` *optionally* raises Django's version of `PermissionDenied` if `raise_exception = True` is set on the view class (if it inherits from `AccessMixin`), otherwise it will d a login redirect. See: https://docs.djangoproject.com/en/4.1/ref/exceptions/#permissiondenied

Comment (HanlinMiao) [reply nautobot/core/views/mixins.py:103]: I tried to catch the DRF version of 403 `PermissionDenied` exception by `self.permission_denied()`. However, before that exception is even raised I ran into an `AttributeError `which is similar to what is explained [here](https://github.com/encode/django-rest-framework/issues/7210). The request.authenticator is a part of `self.permission_denied()` and is only accessible in API requests apparently. 
I think we can either stick with my current approach just to catch all exceptions and return `self.handle_no_permission()` (which I dont like it either) or we can revert to my previous approach: instead of raising an DRF exceptions we can return True or False from `self.check_permissions()`. What do you think? 🤔

Comment (jathanism) [reply nautobot/core/views/mixins.py:103]: Ahh, thanks for that context. Honestly since it seems like we'll be expecting this exception every time, we could just catch both `AttributeError` and `PermissionDenied` to split the difference and I'd be happy. Can you also please provide a comment linking to https://github.com/encode/django-rest-framework/issues/7210?

**Review by jathanism — CHANGES_REQUESTED** [review]
See latest comment. Let's make that change and IMO this is good-to-go.

File: nautobot/core/views/mixins.py (hunk 2 above)
Comment (HanlinMiao) [reply nautobot/core/views/mixins.py:103]: Okay so apparently I can convert a WSGI request to an API request by doing `api_request = self.initialize_request(request, *args, **kwargs)` and pass `api_request` to self.check_permissions() instead. Now DRF's `exceptions.NotAuthenticated, exceptions.PermissionDenied` can be raised properly and handled by `self.handle_no_permission` in the except block. Check out this [commit](https://github.com/nautobot/nautobot/pull/2598/commits/4f98e27b0436271a817a5257958a5500d041e27b) to see the change I made. I also added some comments.

File: nautobot/utilities/testing/views.py [hunk 3]
```diff
@@ -125,6 +125,10 @@ def test_get_object_anonymous(self):
             self.client.logout()
             response = self.client.get(self._get_queryset().first().get_absolute_url())
             self.assertHttpStatus(response, 200)
+            response_body = response.content.decode(response.charset)
+            self.assertIn(
+                "/login/?next=" + self._get_queryset().first().get_absolute_url(), response_body, msg=response_body
+            )
```
glennmatthews, line 131: Should we have a corresponding `assertNotIn` added to `test_get_object_without_permission()`?

nautobot/core/views/mixins.py
```diff
@@ -122,13 +122,16 @@ def dispatch(self, request, *args, **kwargs):
         Used to determine whether the user has permissions to a view and object-level permissions.
         Using AccessMixin handle_no_permission() to deal with Object-Level permissions and API-Level permissions in one pass.
         """
-        # initialize_request also instantiates self.action which is needed for permission checks.
-        self.initialize_request(request, *args, **kwargs)
+        # self.initialize_request() convert a WSGI request and return an API request object which can be passed into self.check_permissions()
+        # If the user is not authenticated or does not have the permission to perform certain actions,
+        # DRF NotAuthenticated or PermissionDenied exceptions can be properly raised and handled by self.handle_no_permission() in the UI.
+        # initialize_request() also instantiates self.action which is needed for permission checks.
+        api_request = self.initialize_request(request, *args, **kwargs)
```
jathanism, line 132: 👨‍🍳 💋

jathanism (approved):

nautobot/utilities/testing/views.py
```diff
@@ -141,7 +141,12 @@ def test_get_object_without_permission(self):
 
             # Try GET without permission
             with disable_warnings("django.request"):
-                self.assertHttpStatus(self.client.get(instance.get_absolute_url()), [403, 404])
+                response = self.client.get(instance.get_absolute_url())
+                self.assertHttpStatus(response, [403, 404])
+                response_body = response.content.decode(response.charset)
+                self.assertNotIn(
+                    "/login/?next=" + self._get_queryset().first().get_absolute_url(), response_body, msg=response_body
```
glennmatthews, line 148: Why `self._get_queryset().first().get_absolute_url()` rather than `instance.get_absolute_url()`?

Or for that matter, why not just `self.assertNotIn("/login/", response_body, msg=response_body)`?

glennmatthews (approved):

Issue #2477, opened 2022-09-23
Unable to stand up nautobot in a HA mode (one write instance & one read-only instance)


gneville-ot:
### Environment
* Nautobot version (Docker tag too if applicable): 1.3.6
* Python version: 3.10
* Database platform, version: postgres
* Middleware(s):

### Steps to Reproduce
1. Standup nautobot using docker-compose but change the compose file so the database port is exposed
2. Connect to the postgres database and create a read-only user and set transactions to read-only
```
CREATE USER nautobot_read WITH PASSWORD 'readonly';
GRANT CONNECT ON DATABASE nautobot TO nautobot_read;
GRANT USAGE ON SCHEMA public TO nautobot_read;
GRANT SELECT ON ALL TABLES IN SCHEMA public TO nautobot_read;
GRANT SELECT ON ALL SEQUENCES IN SCHEMA public TO nautobot_read;
ALTER USER nautobot_read set default_transaction_read_only = on;
```
4. Copy the docker-compose directory to another folder and change the compose file to use a different port for exposing the web interface and remove the db container and volume
5. Change the local.env to update the DB settings so they use a DB host of the local IP on the machine and the database port that was exposed in step 1 and the DB account/password for the read user setup in Step 2.

### Expected Behavior
I'm able to spin up a write instance and a separate read-only instance.

The idea being that nautobot can be hosted in two different sites, one site with write the other for read, the redis cluster is synced cross-site and the database has read-only replicas in the other site.

### Observed Behavior
The read-only instance complains it doesn't have the permissions to perform INSERT/UPDATE statements in different tables.

```
nauto_read-celery_beat-1      | Traceback (most recent call last):
nauto_read-celery_beat-1      |   File "/usr/local/lib/python3.10/site-packages/django/db/backends/utils.py", line 84, in _execute
nauto_read-celery_beat-1      |     return self.cursor.execute(sql, params)
nauto_read-celery_beat-1      | psycopg2.errors.ReadOnlySqlTransaction: cannot execute UPDATE in a read-only transaction
nauto_read-celery_beat-1      | 
nauto_read-celery_beat-1      | 
nauto_read-celery_beat-1      | The above exception was the direct cause of the following exception:
nauto_read-celery_beat-1      | 
nauto_read-celery_beat-1      | Traceback (most recent call last):
nauto_read-celery_beat-1      |   File "/usr/local/lib/python3.10/site-packages/django_celery_beat/schedulers.py", line 320, in update_from_dict
[...]

Issue #2578, opened 2022-10-05
Many settings as defined in settings.py do not respect the documented environment variables


glennmatthews:
### Environment
* Nautobot version (Docker tag too if applicable): 1.4.5
* Python version: any
* Database platform, version: any
* Middleware(s): n/a

### Steps to Reproduce
1. Use a minimal, stripped-down `nautobot_config.py`, rather than one containing all optional settings like the `nautobot/core/templates/nautobot_config.py.j2` template does.
2. Attempt to influence various optional settings by setting environment variables for the Nautobot process, e.g. `NAUTOBOT_DEBUG=True`

### Expected Behavior

Environment variables should be respected as documented.

### Observed Behavior

Many environment variables are not respected by the defaults defined in `settings.py`, only the overrides for these settings in `nautobot_config.py.j2` respect environment variables. Related: #2574. The below list may not be comprehensive:

- CACHEOPS_DEFAULTS / NAUTOBOT_CACHEOPS_DEFAULTS
- CACHEOPS_ENABLED / NAUTOBOT_CACHEOPS_ENABLED
- CACHEOPS_REDIS / NAUTOBOT_CACHEOPS_REDIS
- CORS_ALLOW_ALL_ORIGINS / NAUTOBOT_CORS_ALLOW_ALL_ORIGINS
- DEBUG / NAUTOBOT_DEBUG
- DISABLE_PREFIX_LIST_HIERARCHY / NAUTOBOT_DISABLE_PREFIX_LIST_HIERARCHY (making this inherit from a variable by default would possibly break our Constance config though)
- ENFORCE_GLOBAL_UNIQUE / NAUTOBOT_ENFORCE_GLOBAL_UNIQUE
- MAINTENANCE_MODE / NAUTOBOT_MAINTENANCE_MODE
- METRICS_ENABLED / NAUTOBOT_METRICS_ENABLED
- SESSION_COOKIE_AGE / NAUTOBOT_SESSION_COOKIE_AGE
- SESSION_FILE_PATH / NAUTOBOT_SESSION_FILE_PATH
- STRICT_FILTERING / NAUTOBOT_STRICT_FILTERING
- TIME_ZONE / NAUTOBOT_TIME_ZONE, as well as all of the related time/date format variables

Solution should *probably* be to just strip down `nautobot_config.py.j2` and move the environment variable checks directly into `settings.py`.

This also relates to #1150.

jathanism:
[...]

PR #2600, opened 2022-10-10
[WIP] Exposes `SESSION_ENGINE` as ENVVAR Setting, Always Run `collectstatic`

gneville-ot:
# Closes: #2477
# What's Changed

- Changes the Docker Entrypoint so it runs `nautobot-server collectstatic` if `NAUTOBOT_DOCKER_SKIP_INIT=True`, this is needed so static content can be served when standing up in maintenance mode.
- Allows for easier overriding of the SESSION_ENGINE default with ENVVAR `NAUTOBOT_SESSION_ENGINE`, this is changed when standing up in maintenance mode.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/core/templates/nautobot_config.py.j2
+++ b/nautobot/core/templates/nautobot_config.py.j2
@@ -224,10 +224,6 @@ PLUGINS = []
 REMOTE_AUTH_AUTO_CREATE_USER = False
 REMOTE_AUTH_HEADER = "HTTP_REMOTE_USER"
 
-# Set the SESSION_ENGINE
-# By default, Nautobot will use 'django.contrib.sessions.backends.db' however if you need a maintenance mode / read-only instance then this can be overriden
-SESSION_ENGINE = os.getenv("NAUTOBOT_SESSION_ENGINE", "django.contrib.sessions.backends.db")
-
 # The length of time (in seconds) for which a user will remain logged into the web UI before being prompted to
 # re-authenticate. (Default: 1209600 [14 days])
 SESSION_COOKIE_AGE = int(os.getenv("NAUTOBOT_SESSION_COOKIE_AGE", "1209600"))  # 2 weeks, in seconds
--- a/nautobot/core/templates/nautobot_config.py.j2
+++ b/nautobot/core/templates/nautobot_config.py.j2
@@ -224,6 +224,10 @@ PLUGINS = []
 REMOTE_AUTH_AUTO_CREATE_USER = False
 REMOTE_AUTH_HEADER = "HTTP_REMOTE_USER"
 
+# Set the SESSION_ENGINE
+# By default, Nautobot will use 'django.contrib.sessions.backends.db' however if you need a maintenance mode / read-only instance then this can be overriden
+SESSION_ENGINE = os.getenv("NAUTOBOT_SESSION_ENGINE", "django.contrib.sessions.backends.db")
+
 # The length of time (in seconds) for which a user will remain logged into the web UI before being prompted to
 # re-authenticate. (Default: 1209600 [14 days])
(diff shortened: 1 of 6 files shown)


jathanism:
Thanks for this submission! Sorry we didn't get to it for the 1.4.~6~7 release.

nautobot/core/templates/nautobot_config.py.j2
```diff
@@ -224,6 +224,10 @@ PLUGINS = []
 REMOTE_AUTH_AUTO_CREATE_USER = False
 REMOTE_AUTH_HEADER = "HTTP_REMOTE_USER"
 
+# Set the SESSION_ENGINE
+# By default, Nautobot will use 'django.contrib.sessions.backends.db' however if you need a maintenance mode / read-only instance then this can be overriden
+SESSION_ENGINE = os.getenv("NAUTOBOT_SESSION_ENGINE", "django.contrib.sessions.backends.db")
+
```
glennmatthews, line 230: Can this be implemented in `nautobot/core/settings.py` instead? (See also #2578)

glennmatthews (commented):
Thank you!

gneville-ot:
Hi @jathanism, @glennmatthews ,

I've moved the setting in to `nautobot/core/settings.py` and tested and still works.

Thanks

jathanism:
@nniehoff Can you please give this a look!

nniehoff (changes requested):
This makes sense in this case, however we definitely have use cases we need to account for where we don't want to run collectstatic either.  We need to account for these.

gneville-ot:
Hi @nniehoff,

As a regular user of nautobot I won't know all of these use cases. 
As I see it now, `collectstatic` always runs if you're spinning up with docker and doesn't if you set `NAUTOBOT_DOCKER_SKIP_INIT=True`. Is there any reason not to run `collectstatic` when setting  `NAUTOBOT_DOCKER_SKIP_INIT=True`?

I won't be able to progress this PR without feedback from you.

nniehoff:
@gneville-ot that's exactly what I am saying there are use cases out there where when `NAUTOBOT_DOCKER_SKIP_INIT=True` we truly do mean skip all initialization to include `collectstatic`.  Mainly these are around where we use shared file systems for static such as S3.  Nearly all of the HA deployments we do are active/active deployments of Nautobot with active/active HA postgres behind them.  I'm sorry for the trailblazing you are having to do with this issue.  For this PR I'd like to see some other options maybe if `NAUTOBOT_DOCKER_SKIP_INIT=True`  we have a number of other variables to address all of the components of `post_upgrade` which can be found [here](https://github.com/nautobot/nautobot/blob/develop/nautobot/core/management/commands/post_upgrade.py#L76-L114)

glennmatthews (changes requested):
Nick's concerns need to be addressed. I agree that we probably need to keep `NAUTOBOT_DOCKER_SKIP_INIT` as meaning "skip all init tasks" and instead add a different (set of) variables that get passed through to the `nautobot-server post_upgrade` command so that we can for example run `post_upgrade --no-migrate --no-clearsessions --no-remove-stale-contenttypes --no-trace-paths --no-invalidate-all` if it's desirable to only run `collectstatic`... looking at that complexity, maybe we could consider adding a `--only` option to the `post_upgrade` command (`nautobot-server post_upgrade --only collectstatic`)?

bryanculver:
Pending feedback: https://github.com/nautobot/nautobot/issues/2477#issuecomment-1329703574

bryanculver:
Closing as this largely needs to be broken out into its own two issues that this doesn't fully address:
- We have a larger need around what maintenance mode does and does not support, see #3183
- Based on @glennmatthews' comment above we could have a few more knobs for the Docker entryfile to tune what commands get run on startup instead of an all-or-nothing post_upgrade. The use case here is just one of several but there is no simple way to "override" the entryfile without drift management, see #3733

PR #2602, opened 2022-10-11
Extend `ObjectChange` `change_context_detail` to 400 chars

chadell:
# Closes: https://github.com/nautobot/nautobot-plugin-circuit-maintenance/issues/218
# What's Changed

- The `change_context_detail` added in version 1.4.1 was limited to 100 chars
- Some applications have very long names (>100), as the one referenced in the issue: plugins-nautobot_circuit_maintenance-handle_notifications-handler-handlecircuitmaintenancenotifications
- The fix increases `change_context_detail` from 100 to 400 chars

# TODO

- [ ] Unit, Integration Tests


chadell:
Adding @gsnider2195 as the original code contributor

jvanderaa:
Confirms to fix https://github.com/nautobot/nautobot-plugin-circuit-maintenance/issues/218

glennmatthews:
Can we also make this more robust in general so that we don't hit this same issue in the future if (heaven forbid) we somehow end up with a plugin with a 200-character viewname? It'd be good to ensure that we automatically truncate the change_context_detail to fit, e.g. here at least:

https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/signals.py#L83

changes/2602.changed
```diff
@@ -0,0 +1 @@
+`ObjectChange.change_context_detail` extended from 100 to 200 chars
```
glennmatthews, line 1:
```suggestion
Increased size of `ObjectChange.change_context_detail` field from 100 to 200 chars.
```

glennmatthews (commented):
Come to think of it, we have a similar potential issue with `object_repr` - it would be good to add a length check and automatic truncation [here](https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/models/change_logging.py#L40) as well.

glennmatthews:
Ah - this is actually coming from https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/jobs.py#L1243. Since Jobs can have a slug [up to 320 characters in the worst case](https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/constants.py#L39), we definitely need some truncation logic and/or to increase the limit to more than 200 chars.

chadell:
> https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/jobs.py#L1243

I agree on the idea. I would cover the `object_repr` too.
I believe that the truncation is the most conservative approach, as it will cover our backs whatever we find in the future.
If there are no concerns, I would apply a 200 to both attributes, with truncation.

jvanderaa:
@chadell can we move to a 320 size on the context to match that of a slug?

chadell:
We could move to 400 (to account for the application name). 
But, I will also apply truncation to solve this once and for all.

bryanculver:
@chadell Moved to draft to indicate you're still updating the PR to make it visually distinct.

Please select "Ready for review" once the PR has been updated.

glennmatthews (commented):
Looks functionally solid to me. Can we define and use some constants in `nautobot.extras.constants` for these two values so that we don't have to have hard-coded `200` and `400` in various places?

glennmatthews (approved):

PR #2612, opened 2022-10-13
Make a Relationship Required

briddo:
# Closes: #873 
# What's Changed
So far, this code will check if a relationship on an object is required and raise a `forms.ValidationError` if the required side hasn't provided the relationship data. It also adds the same functionality to the `create` and `update` methods in `nautobot.extras.api.relationships.RelationshipModelSerializerMixin`.

It will prevent an object creation or edit if there is a required relationship defined for the object to a model which doesn't have any existing objects:

- In the case of the UI, this presents as form validation errors. 
- In the case of the API, it will add `serializers.ValidationError`s with messages detailing which endpoints to post required objects to.

# TODO
- [x] Modify existing UI tests (currently in PR [#2371)](https://github.com/nautobot/nautobot/pull/2371) to work with this refactor
- [x] The following additional tests:
    * object is created without the relationship, then the relationship is marked as required, object is updated without the relationship (fails)
    * object is created without the relationship, then the relationship is marked as required, object is updated to include the relationship (succeeds)
    * object is created with the required relationship, object is updated without specifying any relationship data (succeeds, relationship remains in place)
    * object is created with the required relationship, object is updated to remove the relationship data (fails)
- [x] Docs


nautobot/extras/forms/mixins.py
```diff
@@ -394,6 +394,15 @@ def clean(self):
...
           a different destination object.
         """
+
+        required_relationships_errors = Relationship.required_related_objects_errors(
+            self.instance, "ui", self.cleaned_data
+        )
+        for error_dict in required_relationships_errors:
+            self.add_error(None, error_dict)
+            for key in error_dict.keys():
+                self.add_error(None, error_dict[key])
```
glennmatthews, line 404: Can we add errors to the specific form field(s) corresponding to the missing relationship(s)? Something like `self.add_error(key, error_dict[key])`?

nautobot/extras/models/relationships.py
```diff
@@ -249,6 +261,14 @@ class Relationship(BaseModel, ChangeLoggedModel, NotesMixin):
         default=RelationshipTypeChoices.TYPE_MANY_TO_MANY,
         help_text="Cardinality of this relationship",
     )
+    required_side = models.CharField(
+        max_length=12,
+        choices=RelationshipRequiredSideChoices,
+        default=RelationshipRequiredSideChoices.NEITHER_SIDE_REQUIRED,
+        help_text="Objects on the specified side MUST implement this relationship. Not permitted for symmetric relationships.",
+        verbose_name="Required Side",
+        blank=True,
```
glennmatthews, line 270: I don't think we should allow this field to be blank (what would that mean?), especially given we have an appropriate default value.

nautobot/extras/migrations/0051_relationship_required_side.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ("extras", "0050_add_job_task_queues"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="relationship",
+            name="required_side",
+            field=models.CharField(blank=True, default="", max_length=12),
```
glennmatthews, line 16: This looks like it doesn't match with the model field definition below (notably the default value) - might need to regenerate this migration

nautobot/extras/models/relationships.py
```diff
@@ -233,6 +235,16 @@ def get_for_model(self, model):
...
 
+    def get_required_for_model(self, model):
+        """
+        Return a queryset with all required Relationships on the given model.
+        """
+        content_type = ContentType.objects.get_for_model(model._meta.concrete_model)
+        return self.get_queryset().filter(
+            (Q(source_type=content_type) | Q(destination_type=content_type))
+            & Q(required_side__in=["source", "destination"])
+        )
```
glennmatthews, line 246: This "feels" wrong to me. As implemented, won't it return cases where this model is on the *non*-required side of a required relationship as well?

nautobot/extras/models/relationships.py
```diff
@@ -249,6 +261,14 @@ class Relationship(BaseModel, ChangeLoggedModel, NotesMixin):
         default=RelationshipTypeChoices.TYPE_MANY_TO_MANY,
         help_text="Cardinality of this relationship",
     )
+    required_side = models.CharField(
+        max_length=12,
+        choices=RelationshipRequiredSideChoices,
+        default=RelationshipRequiredSideChoices.NEITHER_SIDE_REQUIRED,
+        help_text="Objects on the specified side MUST implement this relationship. Not permitted for symmetric relationships.",
```
glennmatthews, line 268: Probably need `clean()` logic to raise an error if (symmetric && required)

nautobot/extras/models/relationships.py
```diff
@@ -533,6 +553,115 @@ def clean(self):
                     " are present in the database, delete all associations first before modifying the destination type."
                 )
 
+    def skip_required(self, referenced_instance_or_class, side):
```
glennmatthews, line 556: Docstring please.

nautobot/extras/models/relationships.py
```diff
@@ -533,6 +553,115 @@ def clean(self):
...
+
+        # Not enforcing required symmetric relationships
+        if self.symmetric:
+            return True
+
+        # Skip self-referencing
+        if ContentType.objects.get_for_model(referenced_instance_or_class) == getattr(self, f"{side}_type"):
+            return True
```
glennmatthews, line 564: Maybe something else to include in the help_text and enforce in `clean()`?

nautobot/extras/models/relationships.py
```diff
@@ -533,6 +553,115 @@ def clean(self):
...
+                        pass
+
+                error_message = mark_safe(
+                    f"{name_plural[0].upper()}{name_plural[1:]} require "
+                    f"{num_required_verbose} {required_model_meta.verbose_name}, but no "
+                    f"{required_model_meta.verbose_name_plural} exist yet. {hint}"
+                )
+                field_errors[field_key].append(error_message)
+
+            if isinstance(initial_data, dict):
```
glennmatthews, line 637: When is it not a dict? What should we do in that case?

nautobot/extras/models/relationships.py
```diff
@@ -533,6 +553,115 @@ def clean(self):
...
+
+                supplied_data = []
+
+                if output_for == "ui":
+                    supplied_data = initial_data.get(field_key, [])
+                elif output_for == "api":
+                    supplied_data = initial_data.get(relation, {}).get(opposite_side, {})
```
glennmatthews, line 644: Shouldn't the lookup here be using `field_key` instead of `relation`?

nautobot/extras/tests/test_api.py
```diff
@@ -2627,6 +2657,191 @@ def test_populate_relationship_associations_on_site_create(self):
...
+
+        # Try creating a platform when no device exists
+        response = self.post_to_endpoint(Platform, create_data)
+        expected_error_json = [
+            {
+                "platform-devices": [
+                    "Platforms require at least one device, but no devices exist yet. "
+                    "Create a device by posting to /api/dcim/devices/",
+                    'You must specify at least one device [<device.id>, ...] in relationships["platform-devices"]'
+                    '["destination"]["objects"].',
```
glennmatthews, line 2761: This seems a bit confusing to me - I could read this as saying that you need to specify `relationships["platform-devices"]["destination"]['objects"]` when posting to `/api/dcim/devices/`?

nautobot/extras/models/relationships.py
briddo: I was thinking "" should be the default (neither side required)

nautobot/extras/models/relationships.py
briddo: Ah yes!

nautobot/extras/models/relationships.py
briddo: It's `None` by default in kwargs so that this method can be used to check required relationships errors without the need to supply data.

nautobot/extras/models/relationships.py
```diff
@@ -533,6 +558,144 @@ def clean(self):
...
+                # required relationship associations already exist, and continue (ignore validation) if so
+                if (
+                    instance_or_class.present_in_database is True
+                    and initial_data.get(relation, {}).get(opposite_side, {}) == {}
+                    and not relationships_key_specified
+                ):
+                    existing_associations_count = 0
+                    for _, objects in instance_or_class.get_relationships()[relation.required_on].items():
+                        if len(objects) > 0:
+                            existing_associations_count += len(objects)
```
glennmatthews, line 632: I think I'm missing something here - why does this need to be a `for` loop? Shouldn't we just do `instance_or_class.get_relationships()[relation.required_on][relation].count()`? Or for that matter, even something like `RelationshipAssociation.objects.filter(**{f"{side}_id": instance.pk})` directly to be more efficient?

nautobot/extras/models/relationships.py
```diff
@@ -533,6 +558,144 @@ def clean(self):
...
+            if relation.has_many(opposite_side):
+                num_required_verbose = "at least one"
+            else:
+                num_required_verbose = "a"
+
+            if output_for == "api":
+                # If this is a model instance and the relationships json data key is missing, check to see if
+                # required relationship associations already exist, and continue (ignore validation) if so
+                if (
+                    instance_or_class.present_in_database is True
```
glennmatthews, line 625: `present_in_database` is only an attribute of a model instance, not a model class. Is `instance_or_class` a misnomer, is this parameter always going to be an instance?

nautobot/extras/tests/test_relationships.py
```diff
@@ -954,3 +980,372 @@ def test_relationship_table_render(self):
...
+
+                if interact_with == "api":
+
+                    """
+                    - Relationship is marked as being not required
+                    - Object is created without the required relationship data (succeeds)
+                    - Relationship is marked as being required
+                    - Object is updated without the required relationship data (fails)
+                    - Object is updated with the required relationship data (succeeds)
+                    """
```
glennmatthews, line 1256: Should this test apply to the UI as well?

changes/873.added
```diff
@@ -0,0 +1 @@
+Made it possible to require Relationships to be included when editing the related models.
```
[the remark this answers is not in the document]
briddo: Yeah I was getting to that..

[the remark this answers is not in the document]
briddo: Yeah I haven't got to doing docs yet. They'll be in there eventually.

nautobot/extras/models/relationships.py
```diff
@@ -533,6 +558,144 @@ def clean(self):
                     " are present in the database, delete all associations first before modifying the destination type."
                 )
 
+    def skip_required(self, referenced_instance_or_class, side):
+        """
+        This takes an instance or class and a side and checks if it should
+        be skipped or not when validating required relationships.
+        It will skip when any of the following conditions are True:
+         - a relationship is marked as symmetric
+         - a relationship references itself
```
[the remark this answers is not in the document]
briddo: Ah this was a check I'd put there ages ago which is the same thing as testing for a symmetrical relationship. So not needed now!

nautobot/extras/models/relationships.py
```diff
@@ -533,6 +558,144 @@ def clean(self):
...
+        # Handle the case where required_model_class is None (e.g., relationship to a plugin
+        # model for a plugin that's not installed at present):
+        if required_model_class is None:
+            return True
+
+        return False
+
+    @classmethod
+    def required_related_objects_errors(
+        cls, instance_or_class, output_for="ui", initial_data=None, relationships_key_specified=False
```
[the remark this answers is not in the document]
briddo: I went for the latter (put it in `RelationshipModel`)

bryanculver:
Let's focus on the documentation and any remaining test for the non-bulk use case. Opening issue to address bulk enforcement.

nautobot/docs/release-notes/version-1.5.md
```diff
@@ -10,6 +10,17 @@ If you are a user migrating from NetBox to Nautobot, please refer to the ["Migra
 
 ### Added
 
+#### Required Relationships ([#873](https://github.com/nautobot/nautobot/issues/873))
```
glennmatthews, line 13: Historically we've alphabetized the list of headers in the Release Overview section, in which case this would belong after `Nestable LocationTypes`.

glennmatthews (approved):

bryanculver (approved):

Issue #1892, opened 2022-06-15
Device Cluster/HA Modeling


lampwins:
### As ...

Nelly - Network Engineer

### I want ...

To model various network device clustering and high availability topologies & configuration

### So that ...

I can easily track cluster peers, clustering type, and other high availability constraints for network devices like firewalls and router pairs.

### I know this is done when...

- I can create device clusters depicting Active/Passive with 2 or more nodes
- I can create device clusters depicting Active/Active with 2 or more nodes
- The concept of the cluster is its own model which supports extensibility features like custom fields

### Optional - Feature groups this request pertains to.

- DCIM

### Database Changes

- A new model for clusters (will need to consider naming here with virtualization.cluster)

### External Dependencies

None

### Specific User Stories

> As a user I would like to store/retrieve in Nautobot information to be able to deploy a Palo Alto firewall cluster. I will need to know what devices are in the cluster, the cluster's redundancy strategy, the order of which devices will be used (priority/weight).

> As a user I would like to be able to constrain a cluster type to a specific set of device types.

> As a user I would like to be prevented from changing attributes of the cluster type that would cause the current data to be invalid (example removing support for a particular device type while an existing cluster included one).

### TODO:

- [x] #2275
- To be defined based upon outcome of spike
- Name: RedundancyGroup
- Failover strategy on RedundancyGroup model (ENUM, "Active/Active", "Active/Passive")
- FK relationship from Device to RedundancyGroup
- Device holds priority attribute in RedundancyGroup

lampwins:
Also of note, virtual chassis are meant to model multiple devices that share a common control plane. What I am after here are cases where the control plane is separate, but also cases where an engineer simply wants to designate that Device A is the "primary" and Device B is the "backup" even though the two devices may not even know about each other in the network design.

itdependsnetworks:
I think we need to document where each would be handled, it is quickly getting confusing, at least to this Nautobot enthusiast.  Feel free to add to the list, but just off the top of my head. 

Example | virtual chassis | Device Cluster
-------- | --------------- | --------------
Cisco Stackwise |  <ul><li>[ ] </li></ul> |  <ul><li>[ ] </li></ul>
Cisco VSS |  <ul><li>[ ] </li></ul> |  <ul><li>[ ] </li></ul>
Nexus VPC |  <ul><li>[ ] </li></ul> |  <ul><li>[ ] </li></ul>
[...]

PR #2615, opened 2022-10-14
Device Redundancy Groups

bryanculver:
# Closes: #1892
# What's Changed

Starting `RedundancyGroup` implementation.

So far what's implemented:
- `DeviceRedundancyGroup` model, supporting custom fields, tags, statuses, `SecretGroups`
    - `failover_strategy` is a choice, currently identified as `Active/Active` or `Active/Passive`
- `Device` has been updated:
    - A FK to `DeviceRedundancyGroup` as `device_redundancy_group`
    - A PositiveSmallInt for optional priority as `device_redundancy_group_priority`
- `ConfigContext` has been updated:
    - A `ManyToManyField` to `dcim.DeviceRedundancyGroup` as `device_redundancy_groups` for the purposes of applying a ConfigContext based upon a `Device`s `DeviceRedundancyGroup` membership
    - Necessary form, queryset, filter updates to make use of this change
    - Bug fixed on `ConfigContextFilterSet`: `schema` not a valid filter but exists on form, see https://demo.nautobot.com/extras/config-contexts/
    - `schema` filter form field on `ConfigContextFilterForm` wasn't compatible with proper filter field introduced above

Other changes due to changes in Factory fuzzing exposing test failures
- `Device`s form test (`DeviceTestCase`) was flaky
    - A `DeviceType` from factory may not be found that has a `Manufacturer` with an associated `Platform`s that is full depth and `1`U height
        - We just needed a `DeviceType` of non-zero U height
        - Tests updated if needed to be offset by a derived U height
- `BaseNetworkQuerySet` and `IPAddressQuerySet`'s `parse_network_string` function overhauled (used when performing incomplete searches for network addresses)
    - Searching for `b2a` (a valid abbreviated first hextet for an IPv6 address) would not return results. This was due to `RE_HEXTET` requiring a unabbreviated hextet for searching
        - `RE_HEXTET` updated to permit abbreviated forms of hextet
        - This change exacerbated the bug below
    - Searching for `10` would always result in only IPv4 searches, even though it also is a valid IPv6 beginning hextet. This is an ambiguous search string so should return results for both.
        - `_is_ambiguous_network_string` introduced to help determine if inputted string could be either a valid IPv4 beginning octet or a valid IPv6 beginning hextet
        - `string_search` updated to search for both IPv4 and IPv6 addresses if an ambiguous search string is given
    - To keep the output of `parse_network_string` consistent (both derived IP version and always a single `netaddr.IPNetwork`) and DRY the following changes were made:
        - `_safe_parse_network_string` introduced to forcibly try to parse inputted search string as either a IPv4 or IPv6 address or fragment
            - This calls internal `parse_ipv4` and `parse_ipv6` as `parse_network_string` did before
            - This keeps `parse_network_string`'s original error handling for thrown `netaddr.core.AddrFormatError` and ensures always return a valid `netaddr.IPNetwork` object
        - `_check_and_prep_ipv6` introduced to perform much of the "probably IPv6 address" logic originally done in `parse_network_string`
            - Will also prepare the string for `parse_ipv6` by appending the necessary `:`s
        -  `parse_network_string` simplified significantly by leveraging the discrete functions above
    - No tests were harmed in this rewrite however new ones were added to test the discrete functions added above
 
- `RouteTarget`s View test case (`RouteTargetTestCase`) was exposed to have a flaky `csv_data` structure
    - Sometimes `Tenant`s name can include a comma so structure was updated to always escape this field

- `PrefixFactory` may randomly decided to create a child of `2.2.2.2/32`. This is because it checks the size of the prefix and attempts to create children if it's of a large enough size.
     - Weird behavior of `prefix.size`:
         - `Prefix(2.2.2.2/31).size` correctly returns `2` which can have two children in it, `2.2.2.2/32` and `2.2.2.3/32`
         - `Prefix(2.2.2.2/32).size` correctly returns `1` but cannot have a child
     - Factory updated to not create children if `prefix.size` is `1`
- `VirtualChassis` Filter test case (`VirtualChassisTestCase`) had too narrow of a region lookup
    - `Region.objects.filter(sites__isnull=False, children__isnull=True, parent__isnull=True)` may not have returned the required 3 Regions. The tests below were improved in earlier commits to not require no parents and no children so updated to `cls.regions = Region.objects.filter(sites__isnull=False)[:3]`
- Similar failure for `DynamicFilterLookupExpressionTest` Filter test as `VirtualCassis` filter test above.
- `RackGroup` Model test (`test_change_rackgroup_site`) may have randomly chosen the same site to update the rackgroups existing membership to
    - `site_b` updated to be the last object returned from an exclusion of the existing membership site
- Similar failure for `Prefix` Model test as `RackGroup` above.
- `Location` View test case (`LocationTestCase`) was exposed to have a flaky `csv_data` structure
    - Sometimes names can include a comma so structure was updated to always escape these fields
- `RIR` API View test case (`RIRTest`) may have had no deleteable objects to test
    - `Aggregate` has in inbound protected relationship to `RIR` which would prevent deletion. Instead of increasing random pool size, an ensured deletable object was created
- `test_slug_not_modified` was updated to ensure no collision on new slug source value as well as changing lookup expressions from `__contains` to `__exact` in the case where auto-generated names have nested details (ex: Device names: `Device A` and `Device A - PSU`)

# TODO

- [x] Unit, Integration Tests


nautobot/dcim/api/views.py
```diff
@@ -820,6 +821,18 @@ class PowerFeedViewSet(PathEndpointMixin, StatusViewSetMixin, NautobotModelViewS
     filterset_class = filters.PowerFeedFilterSet
 
 
+#
+# Redundancy Groups
+#
+
+
+class RedundancyGroupViewSet(StatusViewSetMixin, NautobotModelViewSet):
+    # v2 TODO(jathan): Replace prefetch_related with select_related
```
bryanculver, line 830: Copy🍝. Should this be replaced now instead of V2 for a new ViewSet?

nautobot/dcim/choices.py
```diff
@@ -1194,3 +1194,35 @@ class PowerFeedPhaseChoices(ChoiceSet):
...
+#
+
+
+class RedundancyGroupStatusChoices(ChoiceSet):
+
+    STATUS_PLANNED = "planned"
+    STATUS_STAGING = "staging"
+    STATUS_ACTIVE = "active"
+    STATUS_DECOMMISSIONING = "decommissioning"
+    STATUS_RETIRED = "retired"
```
bryanculver, line 1210: Different statuses?

nautobot/dcim/filters.py
```diff
@@ -979,6 +981,12 @@ class DeviceFilterSet(
         field_name="virtual_chassis",
         label="Is a virtual chassis member",
     )
+    redundancy_group = NaturalKeyOrPKMultipleChoiceFilter(
+        field_name="redundancy_group",
+        to_field_name="name",
+        queryset=RedundancyGroup.objects.all(),
+        label="Power outlets (name or ID)",
```
bryanculver, line 988: Copy🍝

nautobot/dcim/forms.py
```diff
@@ -4446,3 +4451,32 @@ class PowerFeedFilterForm(NautobotFilterForm, StatusModelFilterFormMixin):
     amperage = forms.IntegerField(required=False)
     max_utilization = forms.IntegerField(required=False)
     tag = TagFilterField(model)
+
+
+class RedundancyGroupForm(NautobotModelForm):
```
bryanculver, line 4456: I think this is missing handling for config contexts.

nautobot/dcim/models/devices.py
```diff
@@ -544,6 +545,20 @@ class Device(PrimaryModel, ConfigContextModel, StatusModel):
...
+    redundancy_group = models.ForeignKey(
+        to="RedundancyGroup",
+        on_delete=models.SET_NULL,
+        related_name="members",
+        blank=True,
+        null=True,
+    )
+    redundancy_group_priority = models.PositiveSmallIntegerField(
+        blank=True,
+        null=True,
```
bryanculver, line 557: Allow null if `redundancy_group` set? Data validation for identical priorities?

nautobot/dcim/models/devices.py
```diff
@@ -998,3 +1013,69 @@ def to_csv(self):
...
+    class Meta:
+        ordering = ("name",)
+
+    def __str__(self):
+        return self.name
+
+    def get_absolute_url(self):
+        return reverse("dcim:redundancygroup", args=[self.pk])
+
+    def to_csv(self):
```
bryanculver, line 1074: Untested

nautobot/dcim/tables/devices.py
```diff
@@ -946,3 +948,24 @@ class Meta(BaseTable.Meta):
...
         fields = ("pk", "name", "domain", "master", "member_count", "tags")
         default_columns = ("pk", "name", "domain", "master", "member_count")
+
+
+#
+# Redundancy Group
+#
+
+
+class RedundancyGroupTable(BaseTable):
```
bryanculver, line 958: Validate

nautobot/dcim/templates/dcim/redundancygroup_retrieve.html
```diff
@@ -0,0 +1,22 @@
+{% extends 'generic/object_retrieve.html' %}
```
bryanculver, line 1: Missing left-hand side sections.

nautobot/dcim/views.py
```diff
@@ -3206,3 +3209,25 @@ class PowerFeedBulkDeleteView(generic.BulkDeleteView):
     queryset = PowerFeed.objects.prefetch_related("power_panel", "rack")
     filterset = filters.PowerFeedFilterSet
     table = tables.PowerFeedTable
+
+
+class RedundancyGroupUIViewSet(NautobotUIViewSet):
```
bryanculver, line 3214: 😎 @HanlinMiao

nautobot/dcim/views.py
```diff
@@ -3206,3 +3209,25 @@ class PowerFeedBulkDeleteView(generic.BulkDeleteView):
...
+    model = RedundancyGroup
+    # bulk_create_form_class = forms.*CSVForm
+    # bulk_update_form_class = forms.*BulkEditForm
```
bryanculver, line 3217: TODO

lampwins:
> Built-in validation for no two `Device`s with same priority?

I'd say no to this validation. There are valid real-world use cases where two devices would have the same priority. For instance, I might have a cluster of 4 firewalls with two `10` and two `20`.

nautobot/dcim/models/devices.py
```diff
@@ -544,6 +545,20 @@ class Device(PrimaryModel, ConfigContextModel, StatusModel):
         blank=True,
         null=True,
     )
+    redundancy_group = models.ForeignKey(
```
mzbroch, line 548: I think we should allow for the device to be a member of multiple redundancy groups/domains.

bryanculver: @lampwins?

Is this a known need at this time?

nautobot/core/forms.py
```diff
@@ -20,6 +20,7 @@
             ("devicetype", "Device types"),
             ("device", "Devices"),
             ("virtualchassis", "Virtual Chassis"),
+            ("redundancygroup", "Redundancy Groups"),
```
glennmatthews, line 23: Hm. Refresh my memory as to what this data structure does? I note that we don't have `location` in here, nor do we have any of the `extras` models.

nautobot/dcim/api/urls.py
```diff
@@ -68,6 +68,9 @@
 router.register("power-panels", views.PowerPanelViewSet)
 router.register("power-feeds", views.PowerFeedViewSet)
 
+# Redundancy Group
+router.register("redunancy-groups", views.RedundancyGroupViewSet)
```
glennmatthews, line 72:
```suggestion
router.register("redundancy-groups", views.RedundancyGroupViewSet)
```

nautobot/dcim/api/serializers.py
```diff
@@ -1120,6 +1122,20 @@ class Meta:
         ]
 
 
+class RedundancyGroupSerializer(NautobotModelSerializer, TaggedObjectSerializer, StatusModelSerializerMixin):
+    url = serializers.HyperlinkedIdentityField(view_name="dcim-api:redundancygroup-detail")
+    failover_strategy = ChoiceField(choices=RedundancyGroupFailoverStrategyChoices)
```
glennmatthews, line 1127: Should the serializer also have a nested serializer for the associated devices?

nautobot/dcim/api/serializers.py
```diff
@@ -1120,6 +1122,20 @@ class Meta:
...
+
+    class Meta:
+        model = RedundancyGroup
+        fields = [
+            "url",
+            "name",
+            "failover_strategy",
```
glennmatthews, line 1134: `secrets_group`?

nautobot/dcim/homepage.py
```diff
@@ -87,6 +87,14 @@ def _connected_powerports_count(request):
                 description="Represents a set of devices which share a common control plane",
                 weight=400,
             ),
+            HomePageItem(
+                name="Redundancy Group",
+                link="dcim:redundancygroup_list",
+                model=models.RedundancyGroup,
+                permissions=["dcim.view_redundancygroup"],
+                description="Represents a set of devices which operate in a failover/HA group",
+                weight=400,
```
glennmatthews, line 96: weight should be unique, possibly 450? Or else 500 and increment all the items below accordingly.

nautobot/dcim/models/devices.py
```diff
@@ -998,3 +1013,69 @@ def to_csv(self):
...
+
+    comments = models.TextField(blank=True)
+
+    secrets_group = models.ForeignKey(
+        to="extras.SecretsGroup",
+        on_delete=models.SET_NULL,
+        default=None,
+        blank=True,
+        null=True,
+    )
```
glennmatthews, line 1048: What's the significance/usage of this FK? I don't recall secrets coming up in our design discussions for this model, and I'm not seeing anywhere in the code that's actually *using* this at present. YAGNI?

nautobot/dcim/navigation.py
```diff
@@ -307,6 +307,22 @@
                             ),
                         ),
                     ),
+                    NavMenuItem(
+                        link="dcim:redundancygroup_list",
+                        name="Redundancy Group",
```
glennmatthews, line 312:
```suggestion
                        name="Redundancy Groups",
```

nautobot/dcim/views.py
HanlinMiao: lets gooooo!

nautobot/dcim/models/devices.py
```diff
@@ -11,7 +11,7 @@
 from django.urls import reverse
 from django.utils.safestring import mark_safe
 
-from nautobot.dcim.choices import DeviceFaceChoices, SubdeviceRoleChoices
+from nautobot.dcim.choices import DeviceFaceChoices, SubdeviceRoleChoices, RedundancyGroupFailoverStrategyChoices
```
HanlinMiao, line 14:
```suggestion
from nautobot.dcim.choices import DeviceFaceChoices, RedundancyGroupFailoverStrategyChoices, SubdeviceRoleChoices
```

nautobot/dcim/urls.py
```diff
@@ -33,6 +34,10 @@
 )
 
 app_name = "dcim"
+
+router = NautobotUIViewSetRouter()
```
HanlinMiao, line 38: 😍

nautobot/dcim/models/devices.py
```diff
@@ -998,3 +1013,69 @@ def to_csv(self):
             self.domain,
             self.master.name if self.master else None,
         )
+
+
+@extras_features(
+    "custom_fields",
```
jathanism, line 1019: Seeing as this is a `ConfigContextModel` does `config_context_owners` need to be included here in `extras_features`?

nautobot/dcim/models/devices.py
jathanism, line 1048: I'm realizing we don't have a `SecretsGroupModelMixin` (akin to `StatusModel[Mixin])` to supply this by default. We might want to backlog a TODO for this so that this pattern doesn't need to proliferate by having to manually include this field on models and model forms.

nautobot/dcim/models/devices.py
```diff
@@ -998,3 +1013,69 @@ def to_csv(self):
...
+    "relationships",
+    "statuses",
+    "webhooks",
+)
+class RedundancyGroup(PrimaryModel, ConfigContextModel, StatusModel):
+    """
+    A RedundancyGroup represents a logical grouping of physical hardware for the purposes of indicating redundancy pairs of high-availability.
+    """
+
+    name = models.CharField(max_length=100, unique=True)
```
jathanism, line 1034: Should we also include a `slug` and `description` for RDs? What's the thought process here between `description` and/or `comments` or both?

nautobot/dcim/models/devices.py
bryanculver: Stubbed out here for the migration (so I didn't have to rebuild) but yes I need to implement the rest of the UIs to show it.

The reason I added it is I thought there might be a shared key for HA between the devices.

Thoughts @lampwins?

nautobot/dcim/views.py
```diff
@@ -3206,3 +3209,25 @@ class PowerFeedBulkDeleteView(generic.BulkDeleteView):
...
+    filterset_class = filters.RedundancyGroupFilterSet
+    filterset_form_class = forms.RedundancyGroupFilterForm
+    form_class = forms.RedundancyGroupForm
+    queryset = RedundancyGroup.objects.prefetch_related("status").annotate(
```
jathanism, line 3221:
```suggestion
    queryset = RedundancyGroup.objects.select_related("status", "members").annotate(
```

nautobot/dcim/views.py
```diff
@@ -3206,3 +3209,25 @@ class PowerFeedBulkDeleteView(generic.BulkDeleteView):
...
+        member_count=count_related(Device, "redundancy_group")
+    )
+    serializer_class = serializers.RedundancyGroupSerializer
+    table_class = tables.RedundancyGroupTable
+    lookup_field = "pk"
+
+    def get_extra_context(self, request, instance):
+        members = Device.objects.restrict(request.user).filter(redundancy_group=instance)
```
jathanism, line 3229: Please include call to `context = super.get_extra_context()` here and add `members` to that.

jathanism (changes requested):

nautobot/dcim/models/devices.py
lampwins: Yeah since we have the functionally already there to support this let's move forward with it, unless it presents significant effort.

nautobot/dcim/models/devices.py
bryanculver: Validate/clean method to throw error/unset `redundancy_group_priority` if `redundancy_group` isn't set?

nautobot/dcim/models/devices.py
bryanculver: `slug` is currently in contention? Name is unique so should we use `natural_key` instead?

nautobot/dcim/views.py
bryanculver: 🦤 👍

nautobot/dcim/api/serializers.py
mzbroch: @bryanculver @glennmatthews Will we only have *Device* redundancyGroups in the future? Can Interface / VirtualChassis in the future form its RedundancyGroup?

Shouldn't we rename this model to `DeviceRedundancyGroup` ?

nautobot/core/forms.py
bryanculver: image

bryanculver: Potentially we shouldn't include it if it's not in the "All Objects" search.

nautobot/dcim/models/devices.py
bryanculver: `Device` doesn't either? 🤷

nautobot/dcim/models/devices.py
mzbroch: Un-resolving - Sorry I missed the answer written in invisible font during moonlight. :-)

nautobot/dcim/api/serializers.py
lampwins: Hi @mzbroch one of the major goals here is to not boil the ocean and solve the problem in an agile sense.

The implementation in this PR is about modeling device chassis-level redundancy, nothing more. In doing so, it is trying to provide a data model specific to the needs of that use case.

#2621 was proposed recently as a related solution to hext-hop redundancy protocol modeling. It is similar, yes, but in many ways distinct from this use case. For instance, the ability for interfaces to participate in multiple hext-hop protocols, or the need to associate IP addresses.

My problem with a generic grouping feature for these use cases is that it is just that, generic. As an end user, I would hate to have to create a bunch of custom fields and relationships just to solve the redundancy use cases at the 80% level. We should be providing better data models & UX for those needs.

Let me be clear though, I _do_ see a need for a "static grouping" feature, just not to solve this particular use case.

So I do see us having several models related to redundancy, but targeted at their respective, specific networking use cases.

nautobot/dcim/tests/test_api.py
```diff
@@ -1454,6 +1457,46 @@ def test_patching_primary_ip4_success(self):
         )
         self.assertHttpStatus(response, status.HTTP_200_OK)
 
+    def test_patching_device_redundancy_group(self):
+        """
+        Validate we can set primary_ip4 on a device using a PATCH.
```
bryanculver, line 1462:
```suggestion
        Validate we can set device redundancy group on a device using a PATCH.
```

nautobot/dcim/models/__init__.py
```diff
@@ -23,7 +23,7 @@
     PowerPort,
     RearPort,
 )
-from .devices import Device, DeviceRole, DeviceType, Manufacturer, Platform, VirtualChassis
+from .devices import Device, DeviceRole, DeviceType, Manufacturer, Platform, DeviceRedundancyGroup, VirtualChassis
```
HanlinMiao, line 26: Do we need to alpha sort this perhaps?

nautobot/dcim/models/__init__.py
```diff
@@ -65,6 +65,7 @@
     "RackRole",
     "RearPort",
     "RearPortTemplate",
+    "DeviceRedundancyGroup",
```
HanlinMiao, line 68: Alpha sort?

nautobot/dcim/tests/test_filters.py
```diff
@@ -2647,6 +2661,21 @@ def test_search(self):
...
+        with self.subTest():
+            params = {"device_redundancy_group": [device_redundancy_groups[0].name]}
+            self.assertEqual(self.filterset(params, self.queryset).qs.count(), 2)
+        with self.subTest():
+            params = {
+                "device_redundancy_group": [device_redundancy_groups[0].pk],
+                "device_redundancy_group_priority": [1],
+            }
+            self.assertEqual(self.filterset(params, self.queryset).qs.count(), 1)
+
```
HanlinMiao, line 2678: Are we using `device_redundancy_groups[1]` here at all? It also looks like we only have one `DeviceRedundacyGroup` instance created.

HanlinMiao (commented):
Looks great! Just a couple of alpha sorts in some imports and choicesets.

nautobot/dcim/forms.py
```diff
@@ -2138,6 +2143,8 @@ class DeviceBulkEditForm(
     platform = DynamicModelChoiceField(queryset=Platform.objects.all(), required=False)
     serial = forms.CharField(max_length=255, required=False, label="Serial Number")
     secrets_group = DynamicModelChoiceField(queryset=SecretsGroup.objects.all(), required=False)
+    device_redundancy_group = DynamicModelChoiceField(queryset=DeviceRedundancyGroup.objects.all(), required=False)
+    device_redundancy_group_priority = forms.IntegerField(min_value=1, required=False)
```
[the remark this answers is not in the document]
bryanculver: `DeviceBulkEditForm` doesn't have a `Meta.fields`

nautobot/dcim/api/views.py
```diff
@@ -820,6 +821,17 @@ class PowerFeedViewSet(PathEndpointMixin, StatusViewSetMixin, NautobotModelViewS
     filterset_class = filters.PowerFeedFilterSet
 
 
+#
+# Device Redundancy Groups
+#
+
+
+class DeviceRedundancyGroupViewSet(StatusViewSetMixin, NautobotModelViewSet):
+    queryset = DeviceRedundancyGroup.objects.select_related("status").prefetch_related("members")
```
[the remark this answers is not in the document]
bryanculver: We use `prefetch_related` in many other places. `select_related` is preferred if it's not inverse related field.

nautobot/dcim/models/devices.py
```diff
@@ -786,6 +802,13 @@ def clean(self):
                     }
                 )
 
+        if self.device_redundancy_group_priority is not None and self.device_redundancy_group is None:
+            raise ValidationError(
+                {
+                    "device_redundancy_group_priority": "Must assign a redundancy group when defining a redundancy group priority."
+                }
+            )
```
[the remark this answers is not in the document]
bryanculver: `None`/`null` priority is an acceptable case.

nautobot/dcim/tables/devices.py
```diff
@@ -188,6 +190,12 @@ class DeviceTable(StatusTableMixin, BaseTable):
     virtual_chassis = tables.LinkColumn(viewname="dcim:virtualchassis", args=[Accessor("virtual_chassis__pk")])
     vc_position = tables.Column(verbose_name="VC Position")
     vc_priority = tables.Column(verbose_name="VC Priority")
+    device_redundancy_group = tables.LinkColumn(
+        viewname="dcim:deviceredundancygroup", args=[Accessor("device_redundancy_group__pk")]
+    )
+    device_redundancy_group_priority = tables.TemplateColumn(
+        template_code="""{% if record.device_redundancy_group %}<span class="badge badge-default">{{ record.device_redundancy_group_priority|default:'None' }}</span>{% else %}—{% endif %}"""
```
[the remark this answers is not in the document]
bryanculver: No. Because we want to show "None" priority only if it's in a redundancy group, otherwise we should show "-" because it's not relevant.

nautobot/dcim/tables/devices.py
```diff
@@ -946,3 +956,23 @@ class Meta(BaseTable.Meta):
...
+#
+# Device Redundancy Group
+#
+
+
+class DeviceRedundancyGroupTable(BaseTable):
+    pk = ToggleColumn()
+    name = tables.Column(linkify=True)
+    member_count = tables.TemplateColumn(
+        template_code="""<a href="{{ record.get_absolute_url }}">{{ value }}</a>""",
```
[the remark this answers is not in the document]
bryanculver: @lampwins? Since the members table of the redundancy group is a devices table I fell this is better UX

nautobot/dcim/tables/devices.py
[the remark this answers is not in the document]
bryanculver: image

nautobot/dcim/tables/devices.py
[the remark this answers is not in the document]
lampwins: Yeah this is more in line with what is in place for virtual chassis today.

nautobot/dcim/tables/devices.py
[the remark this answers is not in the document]
glennmatthews: Misreading on my part. Thanks for the clarification.

nautobot/ipam/tests/test_views.py
```diff
@@ -72,8 +72,8 @@ def setUpTestData(cls):
 
         cls.csv_data = (
             "name,tenant,description",
-            f"65000:1004,{tenants[0].name},Foo",
-            f"65000:1005,{tenants[1].name},Bar",
+            f'65000:1004,"{tenants[0].name}",Foo',
+            f'65000:1005,"{tenants[1].name}",Bar',
```
bryanculver, line 76: Sometimes `Tenant` names will include commas from factory.

nautobot/ipam/querysets.py
```diff
@@ -38,7 +38,7 @@ class BaseNetworkQuerySet(RestrictedQuerySet):
     RE_COLON = re.compile(".*::$")
 
     # Match string from "0000" to "ffff" with no trailing ":"
```
bryanculver, line 40:
```suggestion
    # Match string from "0" to "ffff" with no trailing ":"
    # Allows for abbreviated and non-abbreviated hextet forms
```

nautobot/extras/querysets.py
```diff
@@ -24,6 +24,9 @@ def get_for_object(self, obj):
         cluster = getattr(obj, "cluster", None)
         cluster_group = getattr(cluster, "group", None)
 
+        # Virtualization cluster for VirtualMachine
```
bryanculver, line 27:
```suggestion
```

nautobot/dcim/tests/test_models.py
```diff
@@ -696,6 +698,11 @@ def setUp(self):
         self.location_2 = Location.objects.create(
             name="Leaf", status=self.device_status, location_type=self.location_type_2, parent=self.location_1
         )
+        self.device_redundancy_group = DeviceRedundancyGroup.objects.create(
+            name="Redundancy Group 1",
+            status=Status.objects.get_for_model(DeviceRedundancyGroup).get(slug="active"),
+            failover_strategy=DeviceRedundancyGroupFailoverStrategyChoices.FAILOVER_ACTIVE_ACTIVE,
+        )
```
bryanculver, line 705:
```suggestion
        self.device_redundancy_group = DeviceRedundancyGroup.objects.first()
```

Since we have factory now.

nautobot/dcim/tests/test_filters.py
```diff
@@ -4753,4 +4776,71 @@ def test_cable(self):
...
+            failover_strategy=DeviceRedundancyGroupFailoverStrategyChoices.FAILOVER_ACTIVE_ACTIVE,
+            status=status_planned,
+        )
+
+        DeviceRedundancyGroup.objects.create(
+            name="Device Redundancy Group 3",
+            failover_strategy=DeviceRedundancyGroupFailoverStrategyChoices.FAILOVER_ACTIVE_PASSIVE,
+            status=status_planned,
+            secrets_group=secrets_groups[1],
+        )
```
glennmatthews, line 4812: Can we use the groups created by the factory?

glennmatthews (approved):

nautobot/dcim/api/views.py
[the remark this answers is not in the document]
bryanculver: I took it on @jathanism's suggestion to add it in now where applicable but should this be changed back and adding a `#TODO`?

nautobot/dcim/tests/test_filters.py
bryanculver: We could but in fairness it'll probably be the same amount of DB calls as this because I need to create groups of particular conditions for searching: 
- Name
- Slug
- Associated Secrets Group

I figure I'll be searching as much as I am creating but I'll collapse it.

nautobot/dcim/filters.py
```diff
@@ -1039,6 +1041,11 @@ class DeviceFilterSet(
         field_name="virtual_chassis",
         label="Is a virtual chassis member",
     )
+    device_redundancy_group = NaturalKeyOrPKMultipleChoiceFilter(
```
gsnider2195, line 1044: Do we want a `has_device_redundancy_group` filter?

bryanculver: Does that hold up the merge?

gsnider2195: no

nautobot/utilities/testing/views.py
```diff
@@ -491,7 +495,22 @@ def get_deletable_object(self):
...
+            """
+            Get a list of PKs corresponding to objects that can be safely bulk-deleted.
+
+            For some models this may just be any random objects, but when we have FKs with `on_delete=models.PROTECT`
+            (as is often the case) we need to find or create an instance that doesn't have such entanglements.
+            """
+            instances = get_deletable_objects(self.model, self._get_queryset()).values_list("pk", flat=True)[:3]
+            if len(instances) < 3:
+                self.fail(f"Couldn't find 3 deletable objects, only found {len(instances)}!")
+            return instances
```
glennmatthews, line 513: Is this the right place to be adding this? I'd thought that the `nautobot/utilities/testing/api.py` was the one we were having issues with. Do the `views.py` tests even need/use a `get_deletable_object_pks` function?

bryanculver: I'm sorry I thought that was where you wanted it added. I'll go ahead and fix.

nautobot/tenancy/tests/test_views.py
```diff
@@ -59,3 +59,11 @@ def setUpTestData(cls):
         }
         cls.slug_source = "name"
         cls.slug_test_object = "Tenant 8"
+
+    def get_deletable_object_pks(self):
```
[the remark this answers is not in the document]
bryanculver: A tenant is created at the beginning of `setUpTestData` which should be deletable always.

nautobot/dcim/forms.py
```diff
@@ -4448,3 +4472,67 @@ class PowerFeedFilterForm(NautobotFilterForm, StatusModelFilterFormMixin):
...
+            "failover_strategy",
+            "secrets_group",
+        ]
+
+
+class DeviceRedundancyGroupCSVForm(StatusModelCSVFormMixin, CustomFieldModelCSVForm):
+    failover_strategy = CSVChoiceField(
+        choices=DeviceRedundancyGroupFailoverStrategyChoices, required=False, help_text="Failover Strategy"
+    )
+
```
jathanism, line 4528:
```suggestion
```

nautobot/dcim/models/devices.py
```diff
@@ -998,3 +1025,74 @@ def to_csv(self):
...
+
+    @property
+    def members_sorted(self):
+        return self.members.order_by("device_redundancy_group_priority")
+
+    def __str__(self):
+        return self.name
+
+    def natural_key(self):
+        return (self.name,)
```
jathanism, line 1086: It looks like you forgot to include adding `get_by_natural_key()` method to the queryset/manager for this model, which means you'll need to subclass `RestrictedQuerySet` and then overload `objects =` on this model. 😶

jathanism (changes requested):
Small requests here. Loving the changes to `parse_network_string`!

nautobot/utilities/testing/views.py
```diff
@@ -383,14 +383,18 @@ def test_slug_not_modified(self):
...
             if self.slug_source is not None:
+                new_slug_source_value = "kwyjibo"
+
                 obj = self.model.objects.get(**{self.slug_source: self.slug_test_object})
                 expected_slug = self.slugify_function(getattr(obj, self.slug_source))
                 # Update slug source field str
-                filter_ = self.slug_source + "__contains"
-                self.model.objects.filter(**{filter_: self.slug_test_object}).update(**{self.slug_source: "Test"})
+                filter_ = self.slug_source + "__exact"
+                self.model.objects.filter(**{filter_: self.slug_test_object}).update(
```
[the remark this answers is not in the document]
bryanculver: Did you test that suggestion?

nautobot/dcim/models/devices.py
bryanculver: Addressed this @jathanism. Please review for any other show stoppers.

jathanism (approved):

PR #2617, opened 2022-10-15
Add a dynamic filter form to a list view that lacks this feature.

timizuoebideri1:
# Closes: #2348 
# What's Changed

Add dynamic filter form feature to Rack Elevation, Console Connections, Power Connections, Interface Connections

Screenshot 2022-10-17 at 8 50 34 PM
Screenshot 2022-10-17 at 8 51 05 PM

# TODO

- [ ] Unit, Integration Tests


changes/2617.added
```diff
@@ -0,0 +1 @@
+Add dynamic filter form to list views that lacks this feature.
```
glennmatthews, line 1:
```suggestion
Added dynamic filter form support to specialized list views.
```

glennmatthews (commented):
Code changes look good. Would you mind providing a few screenshots of these updated pages with filters in effect?

nautobot/core/views/generic.py
```diff
@@ -255,6 +255,16 @@ def get(self, request):
                 for field_name, values in filter_params.items()
             ]
 
+            if request.GET:
```
glennmatthews, line 258: I see that the `if self.filterset is not None` guard was removed - is that intended/desired?

nautobot/core/views/generic.py
```diff
@@ -255,6 +255,16 @@ def get(self, request):
...
+                factory_formset_params = convert_querydict_to_factory_formset_acceptable_querydict(
+                    request.GET, self.filterset
+                )
+                self.dynamic_filter_form = DynamicFilterFormSet(
```
glennmatthews, line 262: Why does this need to be stored on `self`, instead of just being a local variable?

nautobot/core/templates/generic/object_list.html
```diff
@@ -3,35 +3,19 @@
...
-        margin-right: 5px;
-        margin-bottom: 5px;
-    }
-</style>
-{% endblock extra_styles %}
-
 {% block header %}
     <div class="row noprint">
-        {% with list_url=content_type.model_class|validated_viewname:"list" %}
         <div class="col-sm-8 col-md-9">
```
glennmatthews, line 8: Nit, but in the case where we don't have a search_form, there's no point in leaving a blank space for it - this should be simply `col` in that case so it spans the full width.

nautobot/core/views/generic.py
timizuoebideri1: Yes, it was moved to the position where `if self.filterset` was first stated. No need to have two `if self.filterset` statements.

glennmatthews (approved):

glennmatthews:
Are we waiting on anything before merging this?

PR #2618, opened 2022-10-16
SSO for Local Development of Nautobot

whitej6:
# What's Changed
- Added the ability to ship a prebuilt keycloak container for local sso development
- Provides ability to smoke test social auth core integration
- Precursor to multiple sso

# TODO


development/docker-compose.keycloak.yml
```diff
@@ -0,0 +1,25 @@
+---
+version: "3.4"
+services:
+  keycloak:
+    image: "quay.io/keycloak/keycloak:15.0.2"
```
glennmatthews, line 5: Do we need to be this specific on the tag, or would something like `keycloak:15` be sufficient?

glennmatthews (commented):
I love where this is going! Probably would be good to have some documentation for using this setup in the developer docs?

development/docker-compose.keycloak.yml
whitej6: Can take a peak at their versioning. Just copied a known working example for CI

gsnider2195:
Would we gain anything by running the container in [dev mode](https://www.keycloak.org/server/containers#_trying_keycloak_in_development_mode)?

gsnider2195:
Is it normal for the SSO credential to get cached? Ex

- Login to Nautobot with SSO
- Logout of Nautobot
- Click login again, click login with SSO
- Automatically logged in with the same SSO user that you logged in with before

whitej6:
The caching of creds is actually a cached session with keycloak not cached with nautobot. The timeout/auth settings can be updated from keycloak if you would like to chance this behavior.

whitej6:
@gsnider2195 dev mode _may_ give more logging but I am not familiar enough to say for sure. Please review the updated docs I just added.

development/docker-compose.keycloak.yml
```diff
@@ -0,0 +1,25 @@
+---
+version: "3.4"
+services:
+  keycloak:
+    image: "quay.io/keycloak/keycloak:15"
```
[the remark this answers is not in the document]
whitej6: Feel free to revert this if you want to merge before I'm back from vacation

whitej6, line 5:
```suggestion
    image: "quay.io/keycloak/keycloak:15.1.1"
```

whitej6:
@glennmatthews @gsnider2195 @bryanculver Outside of adding the documentation, I have resolved all outstanding items. If we are good with how it's implemented I will start the SSO dev docs. If you could 👍 this comment to let me know it would be appreciated.

gsnider2195:
If you forget to run `invoke provision-sso` and try to bring up the keycloak container docker ends up creating a `development/nautobot-realms.json/` directory which breaks subsequent calls to `invoke provision-sso`. Can we use a small bash or python script to generate that json file when the container starts up instead of using a separate invoke command? I think this would be a better developer experience

gsnider2195:
Still waiting to convert the `invoke provision_sso` logic into something that happens in the keycloak container at runtime.

bryanculver:
Please also address merge conflicts.

nautobot/docs/development/sso.md
```diff
@@ -0,0 +1,51 @@
...
+```yaml
+---
+nautobot:
+  compose_files:
+    - "docker-compose.yml"
+    - "docker-compose.postgres.yml"
+    - "docker-compose.dev.yml"
+    - "docker-compose.keycloak.yml"
+```
+
```
gsnider2195, line 38: These are identical now and can be condensed into a single example

bryanculver:
@nautobot/core having this in place will help for #3142

gsnider2195 (approved):

Issue #2406, opened 2022-09-12
Copy button - not working for ip address


david-kn:
### Environment
* Nautobot version (Docker tag too if applicable): 1.4.1
* Python version: 3.7.13
(tested also on demo nautobot)

### Steps to Reproduce
1. Open any device object with IP address assigned in Nautobot GUI
2. Click on a "copy" button next to `Primary IP` address when hovering with mouse 

### Expected Behavior
The IP value should have been copied in clipboard.

### Observed Behavior
Nothing is copied (if a clipboard contained something before, it stays there without any change).

Probably the issues is due to fact that the IP address HTML element is missing the ID `ipv4` that is supposed to be used for element selection (see image) because the "copy" button works for device name though (in title) where element ID is properly defined.

btw. I didn't want to create two seperated issues for this topic / for another (un)expected behavior but are you aware that when the copy buton is used for a device name it copies tha device name + it adds a whitespace in front of and behind the name itself? The clipboard is NOT just the name itself (e.g. `ams01-dist-01`)

I would be nice to fix this and also extend in the future with already existing request: https://github.com/nautobot/nautobot/issues/1519 (just saying :-) )
![image](https://user-images.githubusercontent.com/14314517/189653292-f1d6fd1a-b266-46f7-9a5d-accd541a89c6.png)

### TODO
- Revert:
- https://github.com/nautobot/nautobot/pull/1833/files#diff-f0afa17f447ab2c1d36a0f264153145754fa67028968d3d9ff6ea103198b03beL161-R143
- https://github.com/nautobot/nautobot/pull/1833/files#diff-f0afa17f447ab2c1d36a0f264153145754fa67028968d3d9ff6ea103198b03beL182-R164

glennmatthews:
Thanks for the report! Looks like this was broken by #1833.

PR #2616, opened 2022-10-14
Nestable LocationTypes

glennmatthews:
# Closes: #2608 
# What's Changed

- Added `LocationType.nestable` flag. When set, we allow for Locations of this type to be parents/children of other Locations of the same type. This lets us model hierarchies like:

```
* Main Campus ("Building Group" location type)
    * West Campus (nested Building Group)
        * Building A ("Building" location type)
        * Building B (Building)
    * East Campus (nested Building Group)
        * Building C (Building)
        * Building D (Building)
    * South Campus (nested Building Group)
        * Western South Campus (nested Building Group)
            * Building G (Building)
* Satellite Campus (Building Group)
    * Building Z (Building)
```

- To support this in the UI, I had to convert the `LocationFilterSet.child_location_type` filter to be a method filter so as to include nestable locations in the filter results. Hopefully I've followed the correct best practices here.
- Added the option for the `hyperlinked_object` template filter to take an optional `field` argument, i.e `{{ location|hyperlinked_object:"name"}}`, and fixed a few places where using the full `display` of a LocationType or Location had become unwieldy after #2235.

image

image

image

# TODO

- [x] Explanation of Change(s)
- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [x] Attached Screenshots, Payload Example
- [x] Unit, Integration Tests
- [x] Documentation Updates (when adding/changing features)
- n/a Example Plugin Updates (when adding/changing features)
- [x] Outline Remaining Work, Constraints from Design

PR #2626, opened 2022-10-17
fix: add html element id for hyperlinks

david-kn:
# Closes: #2406 
# What's Changed
Hi, a simple PR for adding an ID element for IP address object (IPv4 and IPv6) within html page. This makes the js clipboard tool actually work for this type of data. A unit test added as well.

No element ID before:
image

HTML tag with ID after update:
image

I just tried to do a simple straightforward fix. Any suggestions that would need to be implemented? 

# TODO
- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/utilities/templatetags/helpers.py
+++ b/nautobot/utilities/templatetags/helpers.py
@@ -57,10 +57,13 @@ def hyperlinked_object(value):
     if value is None:
         return placeholder(value)
     display = value.display if hasattr(value, "display") else str(value)
+    element_id = f" id=ipv{value.family}" if hasattr(value, "family") else ""
     if hasattr(value, "get_absolute_url"):
         if hasattr(value, "description") and value.description:
-            return format_html('<a href="{}" title="{}">{}</a>', value.get_absolute_url(), value.description, display)
-        return format_html('<a href="{}">{}</a>', value.get_absolute_url(), display)
+            return format_html(
+                '<a href="{}" title="{}"{}>{}</a>', value.get_absolute_url(), value.description, element_id, display
+            )
+        return format_html('<a href="{}"{}>{}</a>', value.get_absolute_url(), element_id, display)
     return format_html("{}", display)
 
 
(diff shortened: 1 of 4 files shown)


nautobot/utilities/templatetags/helpers.py
```diff
@@ -57,10 +57,13 @@ def hyperlinked_object(value):
     if value is None:
         return placeholder(value)
     display = value.display if hasattr(value, "display") else str(value)
+    element_id = f" id=ipv{value.family}" if hasattr(value, "family") else ""
```
bryanculver, line 60: HTML tags should be in quotes (so `id="ipv4"`)

bryanculver: Instead of introspecting properties of `value`, what about allowing custom HTML properties via optional args (example: https://github.com/django/django/blob/3.2/django/template/defaultfilters.py#L675-L683)

@glennmatthews?

bryanculver (changes requested):

nautobot/utilities/templatetags/helpers.py
david-kn: > HTML tags should be in quotes (so `id="ipv4"`)

I think that also unquoted attributes are valid W3C syntax for "normal" HTML but you're right: it's recommended to use quotes and to be consistent with you code, it's definitely a better choice. Thanks, Bryan.

> I'm already adding a `field` optional arg to this filter as part of #2616 to solve an issue with `value.display` not always being the way we want to render the value - Django doesn't have a straightforward way to define a filter that takes multiple arguments, that I'm aware of, so maybe this problem needs to be solved in a different way - wonder how difficult it would be to implement an `add_element_id` filter so we could do something like `{{ object|hyperlinked_object|add_element_id:"ipv4" }}` instead?

A neat idea to use another specific filter that would be generic for adding element id. I wanted to come up with some quick fix that would fit its purpose, but this makes more sense and has wider potential at the end of the day.

glennmatthews (approved):

Issue #2086, opened 2022-07-22
JobResult logs can be exported as CSV


chadell:
### As ...

Ozzie - Operator

### I want ...

to have an easy option to export the JobResult logs into a CSV file

### So that ...

I could take this output into other systems

### I know this is done when...

- In the JobResult page there is a button to "Export to CSV" the generates and downloads the CSV file

### Optional - Feature groups this request pertains to.

- [x] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [X] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

Nothing

glennmatthews:
What data in particular are you looking to include in the export? Is just job/timestamp/user/status sufficient or would you be looking to somehow include the logs as well?

chadell:
yes, I would like the logs:  job/user/timestamp/grouping/level/object/message

glennmatthews:
OK, so this is more about exporting the logs for a single JobResult than it is for exporting the overall JobResults table, correct?

chadell:
Yes, will try to refine the description to explain it better

PR #2643, opened 2022-10-19
Jobresult csv export

qduk:
# Closes: #2086
# What's Changed

This PR adds an export button on the JobResult page. It exports only the columns shown in the JobEntryLog table to csv. 
 
Screenshot 2022-10-18 195351

# TODO

- [X] Unit, Integration Tests

### DIFF
--- a/changes/2086.added
+++ b/changes/2086.added
@@ -1,1 +1,1 @@
-Added JobResult export feature
+Added JobResult export feature.
--- a/changes/2086.added
+++ b/changes/2086.added
@@ -1,0 +1,1 @@
+Added JobResult export feature
--- a/nautobot/extras/models/jobs.py
+++ b/nautobot/extras/models/jobs.py
@@ -475,7 +475,7 @@ class JobLogEntry(BaseModel):
     log_object = models.CharField(max_length=JOB_LOG_MAX_LOG_OBJECT_LENGTH, null=True, blank=True)
     absolute_url = models.CharField(max_length=JOB_LOG_MAX_ABSOLUTE_URL_LENGTH, null=True, blank=True)
 
-    csv_headers = ["Time", "Grouping", "Level", "Object", "Message"]
+    csv_headers = ["created", "grouping", "log_level", "log_object", "message"]
 
     def __str__(self):
         return self.message
--- a/nautobot/extras/models/jobs.py
+++ b/nautobot/extras/models/jobs.py
@@ -475,6 +475,8 @@ class JobLogEntry(BaseModel):
     log_object = models.CharField(max_length=JOB_LOG_MAX_LOG_OBJECT_LENGTH, null=True, blank=True)
     absolute_url = models.CharField(max_length=JOB_LOG_MAX_ABSOLUTE_URL_LENGTH, null=True, blank=True)
 
+    csv_headers = ["Time", "Grouping", "Level", "Object", "Message"]
+
     def __str__(self):
         return self.message
 
@@ -482,6 +484,10 @@ class JobLogEntry(BaseModel):
         ordering = ["created"]
         get_latest_by = "created"
         verbose_name_plural = "job log entries"
+
+    def to_csv(self):
+        """Indicates model fields to return as csv."""
+        return (self.created, self.grouping, self.log_level, self.log_object, self.message)
 
 
 #
--- a/nautobot/extras/models/jobs.py
+++ b/nautobot/extras/models/jobs.py
@@ -487,7 +487,7 @@ class JobLogEntry(BaseModel):
 
     def to_csv(self):
         """Indicates model fields to return as csv."""
-        return (self.created, self.grouping, self.log_level, self.log_object, self.message)
+        return (str(self.created), self.grouping, self.log_level, self.log_object, self.message)
 
 
 #
--- a/nautobot/extras/tests/test_models.py
+++ b/nautobot/extras/tests/test_models.py
@@ -1260,9 +1260,9 @@ class JobLogEntryTest(TestCase):
         self.assertEqual(log_object.log_level, log.log_level)
         self.assertEqual(log_object.grouping, log.grouping)
 
-    def test_to_csv(self):
+    def test_to_csv_no_log_object(self):
         """Check that `to_csv` returns the correct data from the JobLogEntry model."""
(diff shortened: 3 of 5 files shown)


nautobot/extras/views.py
```diff
@@ -1487,6 +1491,48 @@ class JobResultView(generic.ObjectView):
...
+
+        changelog_url = None
+
+        if isinstance(instance, ChangeLoggedModel):
+            changelog_url = instance.get_changelog_url()
+
+        if "export" in request.GET:
+            response = HttpResponse(self.queryset_to_csv(instance), content_type="text/csv")
+            underscore_filename = slugify(instance.job_model).replace("-", "_")
+            filename = f"{underscore_filename}_logs.csv"
```
glennmatthews, line 1520: nit: maybe add something like `instance.completed` to the filename as well, to make it easier to distinguish between multiple result exports for the same job?

nautobot/extras/views.py
```diff
@@ -1487,6 +1491,48 @@ class JobResultView(generic.ObjectView):
...
+        instance = get_object_or_404(self.queryset, **kwargs)
+
+        changelog_url = None
+
+        if isinstance(instance, ChangeLoggedModel):
+            changelog_url = instance.get_changelog_url()
+
+        if "export" in request.GET:
+            response = HttpResponse(self.queryset_to_csv(instance), content_type="text/csv")
+            underscore_filename = slugify(instance.job_model).replace("-", "_")
```
glennmatthews, line 1519:
```suggestion
            underscore_filename = instance.job_model.slug.replace("-", "_")
```

nautobot/extras/views.py
```diff
@@ -34,6 +36,7 @@
     get_table_for_model,
     prepare_cloned_fields,
     pretty_print_query,
+    csv_format,
```
glennmatthews, line 39: nit: can you move this up within the list here, as they're alphabetized?

nautobot/extras/tests/test_models.py
```diff
@@ -1257,6 +1260,23 @@ def test_log_entry_creation(self):
         self.assertEqual(log_object.log_level, log.log_level)
         self.assertEqual(log_object.grouping, log.grouping)
 
+    def test_to_csv(self):
+        """Check that `to_csv` returns the correct data from the JobLogEntry model."""
+        expected_data = (datetime.datetime(2020, 1, 26, 15, 37, 36), "run", "success", "", "Django Test")
```
glennmatthews, line 1265: Might be good to have a test for a non-empty `log_object`/`absolute_url` as well?

glennmatthews, line 1265: Shouldn't the CSV data contain the string representation of the timestamp, rather than a `datetime` object?

nautobot/extras/models/jobs.py
```diff
@@ -475,6 +475,8 @@ class JobLogEntry(BaseModel):
     log_object = models.CharField(max_length=JOB_LOG_MAX_LOG_OBJECT_LENGTH, null=True, blank=True)
     absolute_url = models.CharField(max_length=JOB_LOG_MAX_ABSOLUTE_URL_LENGTH, null=True, blank=True)
 
+    csv_headers = ["Time", "Grouping", "Level", "Object", "Message"]
```
glennmatthews, line 478: I believe that in general, the csv_headers are lowercase, and they should ideally match the model fields so as to make hypothetical future csv-import for these logs easier:

```suggestion
    csv_headers = ["created", "grouping", "log_level", "log_object", "message"]
```

changes/2086.added
```diff
@@ -0,0 +1 @@
+Added JobResult export feature
```
glennmatthews, line 1: nit:

```suggestion
Added JobResult export feature.
```

nautobot/extras/views.py
```diff
@@ -1487,6 +1491,48 @@ class JobResultView(generic.ObjectView):
...
+        Generic GET handler for accessing an object by PK or slug
+        """
+        instance = get_object_or_404(self.queryset, **kwargs)
+
+        changelog_url = None
+
+        if isinstance(instance, ChangeLoggedModel):
+            changelog_url = instance.get_changelog_url()
+
+        if "export" in request.GET:
```
glennmatthews, line 1517: Could probably move this block up a few lines since in the export case we don't need or use `changelog_url`?

glennmatthews (commented):
Thanks for the PR!

nautobot/extras/views.py
qduk: I did some formatting of the `instance.completed`. Let me know your thoughts!

nautobot/extras/views.py
qduk: Good point. Done!

nautobot/extras/tests/test_models.py
qduk: Added additional test.

nautobot/extras/views.py
```diff
@@ -1487,6 +1490,49 @@ class JobResultView(generic.ObjectView):
...
+            request,
+            self.get_template_name(),
+            {
+                "object": instance,
+                "verbose_name": self.queryset.model._meta.verbose_name,
+                "verbose_name_plural": self.queryset.model._meta.verbose_name_plural,
+                "changelog_url": changelog_url,  # TODO: Remove in 2.0. This information can be retrieved from the object itself now.
+                **self.get_extra_context(request, instance),
+            },
+        )
```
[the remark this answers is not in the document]
qduk: good call. Thanks Glenn!

nautobot/extras/views.py
```diff
@@ -1487,6 +1489,34 @@ class JobResultView(generic.ObjectView):
     queryset = JobResult.objects.prefetch_related("job_model", "obj_type", "user")
     template_name = "extras/jobresult.html"
 
+    def queryset_to_csv(self, instance):
```
glennmatthews, line 1492: nit: I know this is patterned after `generic.ObjectListView.queryset_to_csv()` but since here this takes an instance rather than a queryset, the name is misleading here. Would something like `instance_to_csv()` be better? Or alternatively should this just be rolled inline under `get()` rather than existing as a separate function at all?

qduk: I'm a proponent of the first option! I think that's clearer. PR updated to reflect that. Thanks Glenn. I appreciate the review.

Issue #1660, opened 2022-04-19
Job (model) changelog are displayed without page header


u1735067:
### Environment
* Python version: 3.10
* Nautobot version: 1.3.0

### Steps to Reproduce
1. Install an instance and install jobs
2. Navigate to the Job model page
3. Click "Changelog"

### Expected Behavior
Job model header and navbar is still displayed

### Observed Behavior
Bare changelog is displayed

bryanculver:
Thanks for the report @u1735067. Confirmed.

u1735067:
I can't test right now, but a fix might be:

* https://github.com/nautobot/nautobot/blob/v1.3.5/nautobot/extras/views.py : 
```py
from nautobot.extras.views import ObjectChangeLogView

class JobChangeLogView(ObjectChangeLogView):
    base_template = "base.html"
```

* https://github.com/nautobot/nautobot/blob/v1.3.5/nautobot/extras/urls.py#L378:
```py
    path(
        "jobs/<slug:slug>/changelog/",
        views.JobChangeLogView.as_view(),
        name="job_changelog",
        kwargs={"model": Job},
    )
```

To force the generic template, because the `job.html` template is very specific.

PR #2654, opened 2022-10-20
Fix Job ObjectChange and Notes not displaying page header

timizuoebideri1:
# Closes: #1660 
# What's Changed

Screenshot 2022-10-20 at 7 39 28 AM
Screenshot 2022-10-20 at 7 39 41 AM

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/views.py
+++ b/nautobot/extras/views.py
@@ -1609,7 +1609,6 @@ class ObjectChangeLogView(View):
         RequestConfig(request, paginate).configure(objectchanges_table)
 
         self.base_template = get_base_template(self.base_template, model)
-        print("\n\n ====>", self.base_template)
 
         return render(
             request,
--- a/nautobot/extras/views.py
+++ b/nautobot/extras/views.py
@@ -1609,6 +1609,7 @@ class ObjectChangeLogView(View):
         RequestConfig(request, paginate).configure(objectchanges_table)
 
         self.base_template = get_base_template(self.base_template, model)
+        print("\n\n ====>", self.base_template)
 
         return render(
             request,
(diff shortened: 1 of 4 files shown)


nautobot/extras/views.py
```diff
@@ -2091,3 +2091,18 @@ class WebhookDeleteView(generic.ObjectDeleteView):
...
+# NOTE: Due to inheritance, JobObjectChangeLogView and JobObjectNotesView can only be
+# constructed below # ObjectChangeLogView and ObjectNotesView.
+
+
+class JobObjectChangeLogView(ObjectChangeLogView):
+    base_template = "extras/job_detail.html"
+
+
+class JobObjectNotesView(ObjectNotesView):
+    base_template = "extras/job_detail.html"
```
glennmatthews, line 2108: As an alternative, could we maybe rename `job_detail.html` to `job_retrieve.html`? I think then these views would work without subclassing being needed?

timizuoebideri1: Nope it did not work,

timizuoebideri1:
```python
def get_base_template(base_template, model):
    """
    Returns the name of the base template, if the base_template is not None
    Otherwise, default to using "<app>/<model>.html" as the base template, if it exists.
    Otherwise, check if "<app>/<model>_retrieve.html" used in `NautobotUIViewSet` exists.
    If both templates do not exist, fall back to "base.html".
    """
    if base_template is None:
        base_template = f"{model._meta.app_label}/{model._meta.model_name}.html"
        # TODO: This can be removed once an object view has been established for every model.
        try:
            get_template(base_template)
        except TemplateDoesNotExist:
            base_template = f"{model._meta.app_label}/{model._meta.model_name}_retrieve.html"
            try:
                get_template(base_template)
            except TemplateDoesNotExist:
                base_template = "base.html"
    return base_template
```

timizuoebideri1: It first determines whether `extras/jobs.html` exists, which it does. 
I'm not sure if addressing this would result in a breaking change.

glennmatthews: Got it, thanks for the analysis.

glennmatthews (approved):

Issue #2513, opened 2022-09-28
DynamicGroup Unable To Filter On Region, Location


whitej6:
### Environment
* Nautobot version (Docker tag too if applicable): v1.4.X
* Python version: 3.8
* Database platform, version: PG 13 & 14
* Middleware(s):

### Steps to Reproduce
1. Create set of devices in a site that is nested in a Region
2. Create DynamicGroup with dcim.Device content-type and only filter being the created Region

Replicable in demo.nautobot.com

### Expected Behavior
Ability use DynamicGroups to filter based on most child Region AND any level of parent Region

### Observed Behavior
When Region filter is applied DynamicGroup.members returns zero Device objects.

bryanculver:
Confirmed this also is an issue for Locations as well.

I suspect this is an issue with the logic for handling `TreeNodeMultipleChoiceFilter`. It's resulting in a `__in` lookup expression but `TreeNodeMultipleChoiceFilter` has an overloaded `filter` method which isn't being evaluated for Dynamic Groups.

jathanism:
Related to #2487

jathanism:
The bug is here in `NaturalKeyOrPKMultipleChoiceFilter.get_filter_predicate()`:

```python
        # If it's not a pk and not a list/qs generate then it's a slug and the filter predicate
        # needs to be nested (e.g. `{"site__slug": "ams01"}`) so that it can be useable in `Q`
        # objects.
        #
        # FIXME(jathan): It feels weird to have a list/qs make its way to `get_filter_predicate()`
        # which if you inspect the source for `django_filters.MultipleChoiceFilter.filter()` should
        # be handling those to generate singular filter predicates and concatentating them.
        if not is_pk and not isinstance(v, (list, models.QuerySet)):  # <-- HERE
            name = f"{self.field_name}__{self.field.to_field_name}"
        # Otherwise just trust the field_name
        else:
            logger.debug("UUID or list/qs detected: Filtering using field name")
            name = self.field_name
```

If we set Region to "Netherlands", it SHOULD result in a filter predicate of:

```
{'site__region__slug__in': ['netherlands']}

```

But instead it is resolving to:

```
{'site__region__in': ['netherlands']}
```

Causing the filter to return an empty queryset.

jathanism:
> Confirmed this also is an issue for Locations as well.
> 
[...]

PR #2665, opened 2022-10-20
Fixed an issue with `DynamicGroup` not properly filtering by `Region` or `Location`.

jathanism:
# Closes: #2513 
# What's Changed

- `NaturalKeyOrPKMultipleChoiceFilter.get_filter_predicate()` has been revised to only return a filter for a single value, and for slugs it wil always be a nested lookup.
- `TreeNodeMultipleChoiceFilter.filter()` was revised to always return a flattened list of objects in the event that there may sometimes be a nested a list of `QuerySets`
 - Implemented a new `flatten_iterable` utility to smartly flatten nested lists while skipping strings that is used in `TreeNodeMultipleChoiceFilter.filter()` to flatten nested lists/`QuerySets` before returned.
 - `DynamicGroup.members` will now return group queryset only if it has child groups, otherwise just uses its own filter directly.
 - `DynamicGroup.count` now calls `members.count()` for consistency.

`Region`
image

`Location`
image

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/utilities/filters.py
+++ b/nautobot/utilities/filters.py
@@ -1,7 +1,6 @@
 from collections import OrderedDict
 from copy import deepcopy
 import logging
-import itertools
 import uuid
 
 from django import forms
@@ -26,6 +25,7 @@ from nautobot.utilities.constants import (
     FILTER_NUMERIC_BASED_LOOKUP_MAP,
 )
 from nautobot.utilities.forms.fields import MultiMatchModelMultipleChoiceField
+from nautobot.utilities.utils import flatten_iterable
 
 from taggit.managers import TaggableManager
 
@@ -487,28 +487,18 @@ class TreeNodeMultipleChoiceFilter(NaturalKeyOrPKMultipleChoiceFilter):
     """
 
     def filter(self, qs, value):
-        new_value = []
         if value:
-            for node in value:
+            if any(isinstance(node, TreeNode) for node in value):
                 # django-tree-queries
-                if isinstance(node, TreeNode):
-                    method = "descendants"
+                value = [node.descendants(include_self=True) if not isinstance(node, str) else node for node in value]
+            elif any(isinstance(node, MPTTModel) for node in value):
                 # django-mptt
-                elif isinstance(node, MPTTModel):
-                    method = "get_descendants"
-                else:
-                    method = None
-
-                if method is not None:
-                    descendants = getattr(node, method)(include_self=True)
-                else:
-                    descendants = node
-
-                new_value.append(descendants)
+                value = [
+                    node.get_descendants(include_self=True) if not isinstance(node, str) else node for node in value
+                ]
 
         # This new_value is going to be a list of querysets that needs to be flattened.
-        if new_value:
-            value = list(itertools.chain.from_iterable(new_value))
+        value = list(flatten_iterable(value))
         return super().filter(qs, value)
 
 
--- a/nautobot/utilities/filters.py
+++ b/nautobot/utilities/filters.py
@@ -24,7 +24,6 @@ from nautobot.utilities.constants import (
     FILTER_CHAR_BASED_LOOKUP_MAP,
     FILTER_NEGATION_LOOKUP_MAP,
     FILTER_NUMERIC_BASED_LOOKUP_MAP,
-    FILTER_TREENODE_NEGATION_LOOKUP_MAP,
 )
 from nautobot.utilities.forms.fields import MultiMatchModelMultipleChoiceField
 
@@ -564,10 +563,6 @@ class BaseFilterSet(django_filters.FilterSet):
         ):
             lookup_map = FILTER_NUMERIC_BASED_LOOKUP_MAP
 
-        elif isinstance(existing_filter, (TreeNodeMultipleChoiceFilter,)):
-            # TreeNodeMultipleChoiceFilter only support negation but must maintain the `in` lookup expression
-            lookup_map = FILTER_TREENODE_NEGATION_LOOKUP_MAP
-
         # These filter types support only negation
         elif isinstance(
             existing_filter,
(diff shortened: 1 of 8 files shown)


nautobot/utilities/filters.py
```diff
@@ -492,20 +487,23 @@ class TreeNodeMultipleChoiceFilter(NaturalKeyOrPKMultipleChoiceFilter):
...
-                value = [
-                    node.get_descendants(include_self=True) if not isinstance(node, str) else node for node in value
-                ]
+            for node in value:
+                if isinstance(node, TreeNode):
+                    method = "descendants"
+                elif isinstance(node, MPTTModel):
+                    method = "get_descendants"
+                else:
+                    continue
```
glennmatthews, line 499: Should we at least log an error in here or something?

jathanism: I'm realizing I didn't refactor this correctly. It's not meant to be an error case.

nautobot/utilities/tests/test_filters.py
```diff
@@ -671,9 +671,9 @@ def test_tag_filter(self):
 
     def test_tree_node_multiple_choice_filter(self):
         self.assertIsInstance(self.filters["treeforeignkeyfield"], TreeNodeMultipleChoiceFilter)
-        self.assertEqual(self.filters["treeforeignkeyfield"].lookup_expr, "in")
+        self.assertEqual(self.filters["treeforeignkeyfield"].lookup_expr, "exact")
         self.assertEqual(self.filters["treeforeignkeyfield"].exclude, False)
-        self.assertEqual(self.filters["treeforeignkeyfield__n"].lookup_expr, "in")
+        self.assertEqual(self.filters["treeforeignkeyfield__n"].lookup_expr, "exact")
         self.assertEqual(self.filters["treeforeignkeyfield__n"].exclude, True)
```
glennmatthews, line 677: Interestingly, in NetBox, `treeforeignkeyfield` is expected to be `exact` but `treeforeignkeyfield__n` is expected to be `in`. What's the functional impact of this change?

nautobot/utilities/filters.py
```diff
@@ -560,23 +553,22 @@ def _get_filter_lookup_dict(existing_filter):
...
-            lookup_map = FILTER_TREENODE_NEGATION_LOOKUP_MAP
-
         # These filter types support only negation
         elif isinstance(
             existing_filter,
             (
                 django_filters.ModelChoiceFilter,
                 django_filters.ModelMultipleChoiceFilter,
                 TagFilter,
+                TreeNodeMultipleChoiceFilter,
```
glennmatthews, line 563: What's the functional impact of this change?

nautobot/utilities/tests/test_filters.py
jathanism: I fixed it.

nautobot/utilities/filters.py
jathanism: It's that these filters aren't special and `in` was never the correct lookup expression here. Multiple choice filters are handled like a list of `exact` filters.

See my comment here: https://github.com/nautobot/nautobot/issues/2513#issuecomment-1287282263

nautobot/utilities/filters.py
```diff
@@ -445,16 +445,10 @@ def get_filter_predicate(self, v):
...
-        # which if you inspect the source for `django_filters.MultipleChoiceFilter.filter()` should
-        # be handling those to generate singular filter predicates and concatentating them.
-        if not is_pk and not isinstance(v, (list, models.QuerySet)):
+        # If it's not a pk, then it's a slug and the filter predicate needs to be nested (e.g.
+        # `{"site__slug": "ams01"}`) so that it can be usable in `Q` objects.
+        if not is_pk:
             name = f"{self.field_name}__{self.field.to_field_name}"
-        # Otherwise just trust the field_name
         else:
             logger.debug("UUID or list/qs detected: Filtering using field name")
```
gsnider2195, line 453:
```suggestion
            logger.debug("UUID detected: Filtering using field name")
```

nautobot/utilities/filters.py
glennmatthews: I still don't get why the `TreeNodeMultipleChoiceFilterTest.test_filter_*_exclude` tests are passing regardless of this change from `in` to `exact` for e.g. `region__n` filters, but I'll take the fact that they *are* still passing as evidence that you're correct here. :-)

glennmatthews (approved):

nautobot/utilities/filters.py
gsnider2195: `TreeNodeMultipleChoiceFilter.filter()` was returning a list of querysets so [`super().filter()`](https://github.com/carltongibson/django-filter/blob/2c81768188cd4ce65d5fd20919ea4a8b2b5f214b/django_filters/filters.py#L243) was building

`Q(location__in=(qs1)) | Q(location__in=(qs2) ...`

Now it's returning a list of instances so it needs to build

`Q(location=(location1)) | Q(location=(location2)) ...`

gsnider2195 (approved):

nautobot/utilities/filters.py
jathanism: It's more that the lookup expression matters less so long as the behavior of the filter is aligned with the filter predicate when processing a multi-choice filter vs. a single filter.

That's really the only fundamental difference between `ChoiceFilter` and `MultipleChoiceFilter`, which is the `filter()` method iterates over a list of values.

gsnider2195: That's what I said! It was broken in 2 places so it would work while still being broken 😆

jathanism: See also: https://django-filter.readthedocs.io/en/stable/ref/filters.html#baseinfilter

Issue #2682, opened 2022-10-25
Hide Nautobot version in footer for unauthenticated users


Jonher937:
### As ...

Isabella - Infrastructure Owner

### I want ...

the Nautobot version to be hidden on the navigating pages unless the user is logged in

### So that ...

we comply with company policies

### I know this is done when...

- The version is not displayed on the front-page unless the user is authenticated

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

- None

### External Dependencies

- None

Kircheneer:
Duplicate of #1704? Also see this #1610

Issue #1610, opened 2022-04-04
Revisit `/api/status` API endpoint and contents


glennmatthews:
### Proposed Changes

The `/api/status/` REST API endpoint requires no authentication and returns a large amount of information about the internals of Nautobot's installation, for example:

```json
{
  "django-version": "3.2.12",
  "installed-apps": {
    "cacheops": "5.0.1",
    "constance": "2.8.0",
    "constance.backends.database": null,
    "corsheaders": null,
    "db_file_storage": null,
    "debug_toolbar": "3.2.4",
    "django.contrib.admin": null,
    "django.contrib.auth": null,
    "django.contrib.contenttypes": null,
    "django.contrib.humanize": null,
    "django.contrib.messages": null,
    "django.contrib.sessions": null,
    "django.contrib.staticfiles": null,
    "django_ajax_tables": null,
    "django_celery_beat": "2.2.1..",
    "django_extensions": "3.1.5",
    "django_filters": "21.1",
    "django_jinja": "3.2.12.final.0",
    "django_prometheus": "2.2.0",
    "django_rq": "2.5.1",
    "django_tables2": "2.4.1",
    "drf_spectacular": "0.22.0",
    "drf_spectacular_sidecar": "2022.3.28",
    "graphene_django": "2.15.0",
    "health_check": null,
    "health_check.storage": null,
    "mptt": "0.13.4",
    "nautobot.circuits": null,
    "nautobot.core": null,
    "nautobot.dcim": null,
    "nautobot.extras": null,
    "nautobot.ipam": null,
    "nautobot.tenancy": null,
    "nautobot.users": null,
    "nautobot.utilities": null,
    "nautobot.virtualization": null,
    "rest_framework": "3.13.1",
    "social_django": "5.0.0",
    "taggit": "1.5.1",
    "timezone_field": "4.1.2"
  },
  "nautobot-version": "1.3.0b1",
  "plugins": {},
  "python-version": "3.7.9",
  "rq-workers-running": 1,
  "celery-workers-running": 1
}
```

We should think about whether it's appropriate to expose all of this information on an unauthenticated endpoint, or whether it would be better (particularly from a security standpoint) to remove some or all of this information or move it to an authenticated endpoint.

### Justification

Exposing the versions of Python, Django, Nautobot, and every active Django app provides a wealth of information to a potential attacker.

glennmatthews:
[...]

PR #2683, opened 2022-10-25
Hide version in footer for unauthenticated users

Jonher937:
Closes: #2682

# What's Changed
- Nautbot version number is hidden from the footer for unauthenticated users

# TODO

> Note the version number being present next to the hostname in the `before`, but not in the `after`.

## Before
before

## After
after

### DIFF
--- a/nautobot/core/tests/test_views.py
+++ b/nautobot/core/tests/test_views.py
@@ -86,7 +86,10 @@ class HomeViewTestCase(TestCase):
 
         footer_hostname_version_pattern = re.compile('<p class="text-muted">\\s+\\S+\\s+\\(v[\\d.]+\\)\\s+<\\/p>')
         self.assertRegex(response_content, footer_hostname_version_pattern)
+
         self.client.logout()
+        response = self.client.get(url)
+        response_content = response.content.decode(response.charset).replace("\n", "")
         self.assertNotRegex(response_content, footer_hostname_version_pattern)
 
 
--- a/nautobot/core/tests/test_views.py
+++ b/nautobot/core/tests/test_views.py
@@ -79,6 +79,16 @@ class HomeViewTestCase(TestCase):
         self.assertIsNotNone(nav_search_bar_result)
         self.assertIsNotNone(body_search_bar_result)
 
+    def test_footer_version_visible_authenticated_users_only(self):
+        url = reverse("home")
+        response = self.client.get(url)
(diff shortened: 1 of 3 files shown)


glennmatthews (commented):
This looks reasonable enough, although as @Kircheneer pointed out, there's still the unauthenticated `/api/status/` endpoint to consider from a security standpoint (#1610).

Would you be willing to add a unit test for this rendering logic, probably in `nautobot.core.tests.test_views.HomeViewTestCase`?

Jonher937:
> This looks reasonable enough, although as @Kircheneer pointed out, there's still the unauthenticated `/api/status/` endpoint to consider from a security standpoint (#1610).
> 
> Would you be willing to add a unit test for this rendering logic, probably in `nautobot.core.tests.test_views.HomeViewTestCase`?

That was not my experience, see my comment in the issue.

Sure, the devcontainer seems broken for VSCode (requires "db" but db is not defined, amongst other things). I've pushed a commit with a test, let's hope I nailed it, but noone usually does on the first try :sweat_smile:

glennmatthews (approved):
Looks good to me. We might want to revisit the test case at some point to make it more robust but it's definitely good enough for now. Thank you!

Jonher937:
Sorry for the re-reviews. The CI only runs partial jobs since I'm a new contributor.

nautobot/core/tests/test_views.py
```diff
@@ -79,6 +79,16 @@ def test_search_bar_visible_if_hide_restricted_ui_False(self):
...
 
+    def test_footer_version_visible_authenticated_users_only(self):
+        url = reverse("home")
+        response = self.client.get(url)
+        response_content = response.content.decode(response.charset).replace("\n", "")
+
+        footer_hostname_version_pattern = re.compile('<p class="text-muted">\\s+\\S+\\s+\\(v[\\d.]+\\)\\s+<\\/p>')
+        self.assertRegex(response_content, footer_hostname_version_pattern)
+        self.client.logout()
+        self.assertNotRegex(response_content, footer_hostname_version_pattern)
```
glennmatthews, line 90:
```suggestion
        self.client.logout()

        response = self.client.get(url)
        response_content = response.content.decode(response.charset).replace("\n", "")
        self.assertNotRegex(response_content, footer_hostname_version_pattern)
```

PR #2686, opened 2022-10-26
Add test helper method to `FilterTestCases` to find values suitable for testing multiple choice filters.

gsnider2195:
# Closes: #DNE
# What's Changed

Add a helper method to `FilterTestCase` to find unique values for testing multiple choice filters. This will help with rewriting tests to use factory fixtures.

```py
        def get_filterset_test_values(self, field_name, queryset=None):
            """Returns a list of distinct values from the requested queryset field to use in filterset tests.

            Returns a list for use in testing multiple choice filters. The size of the returned list is random
            but will contain at minimum 2 unique values. The list of values will match at least 2 instances when
            passed to the queryset's filter(field_name__in=[]) method but will fail to match at least one instance.

            Args:
                field_name: The name of the field to retrieve test values from.
                queryset: The queryset to retrieve test values. Defaults to `self.queryset`.

            Returns:
                list: A list of unique values derived from the queryset.

            Raises:
                ValueError: Raised if unable to find a combination of 2 or more unique values
                    to filter the queryset to a subset of the total instances.
            """
```

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/utilities/testing/filters.py
+++ b/nautobot/utilities/testing/filters.py
@@ -1,3 +1,6 @@
+import random
+
+from django.db.models import Count
 from django.test import tag
 from nautobot.utilities.testing.views import TestCase
 
@@ -6,8 +9,48 @@ from nautobot.tenancy.models import Tenant, TenantGroup
 
 @tag("unit")
 class FilterTestCases:
-    class FilterTestCase(TestCase):
+    class BaseFilterTestCase(TestCase):
         """Base class for testing of FilterSets."""
+
+        def get_filterset_test_values(self, field_name, queryset=None):
+            """Returns a list of distinct values from the requested queryset field to use in filterset tests.
+
+            Returns a list for use in testing multiple choice filters. The size of the returned list is random
+            but will contain at minimum 2 unique values. The list of values will match at least 2 instances when
+            passed to the queryset's filter(field_name__in=[]) method but will fail to match at least one instance.
+
+            Args:
+                field_name: The name of the field to retrieve test values from.
+                queryset: The queryset to retrieve test values. Defaults to `self.queryset`.
+
+            Returns:
+                list: A list of unique values derived from the queryset.
+
+            Raises:
+                ValueError: Raised if unable to find a combination of 2 or more unique values
+                    to filter the queryset to a subset of the total instances.
+            """
+            test_values = []
+            if queryset is None:
+                queryset = self.queryset
+            qs_count = queryset.count()
+            values_with_count = queryset.values(field_name).annotate(count=Count(field_name)).order_by("count")
+            for value in values_with_count:
+                # randomly break out of loop after 2 values have been selected
+                if len(test_values) > 1 and random.choice([True, False]):
+                    break
+                if value["count"] < qs_count:
+                    qs_count -= value["count"]
+                    test_values.append(value[field_name])
+
+            if len(test_values) < 2:
+                raise ValueError(
+                    f"Cannot find valid test data for {queryset.model._meta.object_name} field {field_name}"
+                )
+            return test_values
+
+    class FilterTestCase(BaseFilterTestCase):
+        """Add common tests for all FilterSets."""
 
         queryset = None
         filterset = None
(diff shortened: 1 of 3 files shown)


nautobot/utilities/tests/test_filters.py
```diff
@@ -991,6 +992,48 @@ def test_device_comments_multiple_value_charfield_iregex_negation(self):
...
+    def test_empty_queryset(self):
+        with self.assertRaisesMessage(ValueError, self.exception_message):
+            self.get_filterset_test_values("description")
+
+    def test_object_return_count(self):
+        for n in range(1, 11):
+            Site.objects.create(name=f"Site{n}", status=self.status_active, description=f"description {n}")
+        test_values = self.get_filterset_test_values("description", Site.objects.all())
+        self.assertNotEqual(test_values, 0)
+        self.assertNotEqual(test_values, Site.objects.count())
```
glennmatthews, line 1017: `test_values` is a list, right? These assertions seem incorrect to me if we're comparing a list against an int.

nautobot/utilities/tests/test_filters.py
```diff
@@ -991,6 +992,48 @@ def test_device_comments_multiple_value_charfield_iregex_negation(self):
...
+
+    @classmethod
+    def setUpTestData(cls):
+        statuses = Status.objects.get_for_model(Site)
+        cls.status_active = statuses.get(slug="active")
+
+        Site.objects.all().delete()
+
+    def test_empty_queryset(self):
+        with self.assertRaisesMessage(ValueError, self.exception_message):
```
glennmatthews, line 1009: I didn't know about `assertRaisesMessage`... there are quite a few tests I've written where this would be useful!

nautobot/utilities/tests/test_filters.py
```diff
@@ -991,6 +992,48 @@ def test_device_comments_multiple_value_charfield_iregex_negation(self):
...
+
+    queryset = Site.objects.all()
+    exception_message = "Cannot find valid test data for Site field description"
+
+    @classmethod
+    def setUpTestData(cls):
+        statuses = Status.objects.get_for_model(Site)
+        cls.status_active = statuses.get(slug="active")
+
+        Site.objects.all().delete()
```
glennmatthews, line 1006: Do we have any models in or planned for our factories that have a PROTECT foreign key to Site, which would make this problematic? I wonder if instead of deleting all pre-existing Sites, we could instead use a filtered Site queryset for the below tests that only contains the explicitly tested Site(s)?

glennmatthews (commented):
I like this a lot!

nautobot/utilities/tests/test_filters.py
gsnider2195: Sure is! Thanks for catching that

nautobot/utilities/tests/test_filters.py
gsnider2195: We should. There are a lot of protected relationships to Site. Thanks for the heads up, I'll change this.

nautobot/utilities/testing/filters.py
```diff
@@ -6,9 +9,49 @@
...
+            test_values = []
+            if queryset is None:
+                queryset = self.queryset
+            qs_count = queryset.count()
+            values_with_count = queryset.values(field_name).annotate(count=Count(field_name)).order_by("count")
+            for value in values_with_count:
+                # randomly break out of loop after 2 values have been selected
+                if len(test_values) > 1 and random.choice([True, False]):
+                    break
+                if value["count"] < qs_count:
```
HanlinMiao, line 42: Correct me if I'm wrong. Would `value["count"]` ever be greater than `qs_count`? And we do not append the `value[field_name]` when `qs_count == value["count"]`. Does that ignore the last eligibe value in `values_with_count`?

gsnider2195: > Correct me if I'm wrong. Would value["count"] ever be greater than qs_count? 

`value["count"]` would never be greater than `qs_count` because the former is a subset of the latter. The max it could ever be is `qs_count`.


> And we do not append the value[field_name] when qs_count == value["count"]. Does that ignore the last eligibe value in values_with_count?

That's correct and it's by design. The condition `if value["count"] >= qs_count` will only ever match on the final object in the loop and we always want to test a filter on a subset of objects in the queryset. When a filter fails it returns the original queryset. If our test expects a filter to return all of the objects in the queryset we don't know the difference between success and failure.

glennmatthews (approved):

Issue #2253, opened 2022-08-18
Nautobot silently discards interface's tagged vlans if mode is wrong


retnuh:
### Environment
* Python version:   N/A
* Nautobot version:  1.3.8

### Steps to Reproduce
1.  Add a vlan to an interfaces tagged vlans, but don't set the mode to `MODE_TAGGED`
2.  call `save` or `validated_save` on the interface
3.  vlan is no longer in tagged_vlans

### Expected Behavior

I think this should be a validation error, rather than a silent discard - or at least _add_ a validation error, so it will be caught in `validated_save`

Here is the code:

```
    def clean(self):
        # Remove untagged VLAN assignment for non-802.1Q interfaces
        if not self.mode and self.untagged_vlan is not None:
            raise ValidationError({"untagged_vlan": "Mode must be set when specifying untagged_vlan"})

    def save(self, *args, **kwargs):

        # Only "tagged" interfaces may have tagged VLANs assigned. ("tagged all" implies all VLANs are assigned.)
        if self.present_in_database and self.mode != InterfaceModeChoices.MODE_TAGGED:
            self.tagged_vlans.clear()

        return super().save(*args, **kwargs)
```

### Observed Behavior

A validation error or other error so that when doing things programmatically, these things don't happen silently.

glennmatthews:
Sounds related to #895

PR #2688, opened 2022-10-27
Raise an exception if the interface contains a tagged vlan but the mode is not set to tagged

timizuoebideri1:
# Closes: #2253 
# What's Changed

# TODO

- [ ] Unit, Integration Tests


glennmatthews (commented):
Do we need to also cover the case where an interface is initially tagged mode and has tagged_vlans, then the user changes the mode to untagged but doesn't clear the vlans? More generally, how does this impact the UI flow when creating or editing interfaces?

changes/2688.fixed
```diff
@@ -0,0 +1 @@
+Fix exception no raised when the interface contains a tagged vlan but mode is not set to tagged
```
gsnider2195, line 1:
```suggestion
Fixed exception not raised when the interface contains a tagged vlan but mode is not set to tagged.
```

nautobot/dcim/api/serializers.py
```diff
@@ -1025,8 +1025,14 @@ def validate(self, data):
                 )
 
         # Validate many-to-many VLAN assignments
+        tagged_vlans = data.get("tagged_vlans", [])
+        if tagged_vlans and data.get("mode") != InterfaceModeChoices.MODE_TAGGED:
+            raise serializers.ValidationError(
+                {"tagged_vlans": f"Mode must be set to {InterfaceModeChoices.MODE_TAGGED} when specifying tagged_vlan"}
```
gsnider2195, line 1031:
```suggestion
                {"tagged_vlans": f"Mode must be set to {InterfaceModeChoices.MODE_TAGGED} when specifying tagged_vlans"}
```

nautobot/dcim/api/serializers.py
```diff
@@ -1025,8 +1025,14 @@ def validate(self, data):
                 )
 
         # Validate many-to-many VLAN assignments
+        tagged_vlans = data.get("tagged_vlans", [])
+        if tagged_vlans and data.get("mode") != InterfaceModeChoices.MODE_TAGGED:
```
gsnider2195, line 1029: This doesn't work on a `PATCH` to add `tagged_vlans` when `mode` is not specified in the request:

```json
{
  "id": "03d0892a-4b1e-4956-bade-792a3ead7643",
  "display": "Ethernet0",
  "url": "http://localhost:8080/api/dcim/interfaces/03d0892a-4b1e-4956-bade-792a3ead7643/",
  "device": {
    "display": "rtr1",
    "id": "9a3cf918-b8da-40eb-9883-cb51b63ef40e",
    "url": "http://localhost:8080/api/dcim/devices/9a3cf918-b8da-40eb-9883-cb51b63ef40e/",
    "name": "rtr1"
  },
  "name": "Ethernet0",
  "label": "",
  "type": {
    "value": "1000base-t",
    "label": "1000BASE-T (1GE)"
  },
  "enabled": true,
  "parent_interface": null,
  "bridge": null,
  "lag": null,
  "mtu": null,
  "mac_address": null,
  "mgmt_only": false,
  "description": "",
  "mode": {
    "value": "tagged",
    "label": "Tagged"
  },
  "untagged_vlan": null,
  "tagged_vlans": [],
  "cable": null,
  "cable_peer": null,
  "cable_peer_type": null,
  "connected_endpoint": null,
  "connected_endpoint_type": null,
  "connected_endpoint_reachable": null,
  "count_ipaddresses": 0,
  "created": "2022-10-27",
  "last_updated": "2022-10-27T13:29:45.009498Z",
  "tags": [],
  "notes_url": "http://localhost:8080/api/dcim/interfaces/03d0892a-4b1e-4956-bade-792a3ead7643/notes/",
  "custom_fields": {}
}
```


```
curl -X 'PATCH' \
  'http://localhost:8080/api/dcim/interfaces/03d0892a-4b1e-4956-bade-792a3ead7643/' \
  -H 'X-CSRFTOKEN: MaDmbYcv6DjjmlxvfSQockE4q2Svb5pACqya1DWcoOFw7oDtLM49jx8FYdlhEBVS' \
  -d '{
  "tagged_vlans": [
    "ae4a4270-9e68-4040-8000-9da70767f039"
  ]
}'

Response:
{
  "tagged_vlans": [
    "Mode must be set to tagged when specifying tagged_vlan"
  ]
}
```

nautobot/virtualization/api/serializers.py
```diff
@@ -217,6 +217,12 @@ def validate(self, data):
                 )
 
         # Validate many-to-many VLAN assignments
+        tagged_vlans = data.get("tagged_vlans", [])
+        if tagged_vlans and data.get("mode") != InterfaceModeChoices.MODE_TAGGED:
+            raise serializers.ValidationError(
+                {"tagged_vlans": f"Mode must be set to {InterfaceModeChoices.MODE_TAGGED} when specifying tagged_vlan"}
```
gsnider2195, line 223:
```suggestion
                {"tagged_vlans": f"Mode must be set to {InterfaceModeChoices.MODE_TAGGED} when specifying tagged_vlans"}
```

nautobot/dcim/tests/test_models.py
```diff
@@ -1121,3 +1122,30 @@ def test_power_panel_validation(self):
...
+        )
+        interface = Interface.objects.create(
+            name="Int1",
+            type=InterfaceTypeChoices.TYPE_VIRTUAL,
+            device=device,
+        )
+        with self.assertRaises(ValidationError) as err:
+            interface.tagged_vlans.add(vlan1)
+        self.assertEqual(
+            err.exception.message_dict["tagged_vlans"][0], "Mode must be set to tagged when specifying tagged_vlan"
```
gsnider2195, line 1150:
```suggestion
            err.exception.message_dict["tagged_vlans"][0], "Mode must be set to tagged when specifying tagged_vlans"
```

nautobot/dcim/tests/test_api.py
```diff
@@ -1847,6 +1847,20 @@ def test_interface_not_belonging_to_common_device_or_vc_not_allowed(self):
...
+            "device": self.devices[0].pk,
+            "name": "Tagged Interface",
+            "type": "1000base-t",
+            "mode": InterfaceModeChoices.MODE_ACCESS,
+            "tagged_vlans": [self.vlans[0].pk, self.vlans[1].pk],
+            "untagged_vlan": self.vlans[2].pk,
+        }
+        response = self.client.post(self._get_list_url(), data=payload, format="json", **self.header)
+        self.assertHttpStatus(response, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.data["tagged_vlans"][0], "Mode must be set to tagged when specifying tagged_vlan")
```
gsnider2195, line 1862:
```suggestion
        self.assertEqual(response.data["tagged_vlans"][0], "Mode must be set to tagged when specifying tagged_vlans")
```

nautobot/virtualization/tests/test_models.py
```diff
@@ -69,3 +70,20 @@ def test_vm_duplicate_name_per_cluster(self):
...
+
+        clustertype = ClusterType.objects.create(name="Test Cluster Type 1", slug="test-cluster-type-1")
+        cluster = Cluster.objects.create(name="Test Cluster 1", type=clustertype)
+        virtualmachine = VirtualMachine.objects.create(cluster=cluster, name="Test VM 1")
+
+        interface = VMInterface.objects.create(virtual_machine=virtualmachine, name="Interface 1")
+        with self.assertRaises(ValidationError) as err:
+            interface.tagged_vlans.add(vlan)
+        self.assertEqual(
+            err.exception.message_dict["tagged_vlans"][0], "Mode must be set to tagged when specifying tagged_vlan"
```
gsnider2195, line 88:
```suggestion
            err.exception.message_dict["tagged_vlans"][0], "Mode must be set to tagged when specifying tagged_vlans"
```

nautobot/virtualization/tests/test_api.py
```diff
@@ -381,6 +381,19 @@ def test_untagged_vlan_requires_mode(self):
...
+        vlan = VLAN.objects.first()
+        virtualmachine = VirtualMachine.objects.first()
+        payload = {
+            "virtual_machine": virtualmachine.pk,
+            "name": "Tagged Interface",
+            "tagged_vlans": [vlan.pk],
+        }
+        response = self.client.post(self._get_list_url(), data=payload, format="json", **self.header)
+        self.assertHttpStatus(response, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.data["tagged_vlans"][0], "Mode must be set to tagged when specifying tagged_vlan")
```
gsnider2195, line 395:
```suggestion
        self.assertEqual(response.data["tagged_vlans"][0], "Mode must be set to tagged when specifying tagged_vlans")
```

glennmatthews:
> > Do we need to also cover the case where an interface is initially tagged mode and has tagged_vlans, then the user changes the mode to untagged but doesn't clear the vlans? More generally, how does this impact the UI flow when creating or editing interfaces?
> 
> The [`BaseInterface.save()` method](https://github.com/nautobot/nautobot/blob/d53914ca9bfe42a6d6a4be2ae5da70dc15464a15/nautobot/dcim/models/device_components.py#L532-L534) handles clearing `tagged_vlans` if the mode isn't tagged.

Isn't that the original problem described by #2253 though? It silently clears and discards `tagged_vlans` instead of warning the user about the invalid data.

nautobot/dcim/tests/test_api.py
```diff
@@ -1847,6 +1847,41 @@ def test_interface_not_belonging_to_common_device_or_vc_not_allowed(self):
...
+                "name": "Tagged Interface",
+                "type": "1000base-t",
+                "status": "active",
+                "mode": InterfaceModeChoices.MODE_ACCESS,
+                "tagged_vlans": [self.vlans[0].pk, self.vlans[1].pk],
+                "untagged_vlan": self.vlans[2].pk,
+            }
+            response = self.client.post(self._get_list_url(), data=payload, format="json", **self.header)
+            self.assertHttpStatus(response, status.HTTP_400_BAD_REQUEST)
+            print(response.data)
```
glennmatthews, line 1864:
```suggestion
```

glennmatthews (commented):
Based on the description of #2253 I would have expected some change to `Interface.clean()` and/or `Interface.save()`. At the very least it would seem like the logic to remove tagged vlans when changing interface mode should no longer be needed?

gsnider2195:
> Based on the description of #2253 I would have expected some change to `Interface.clean()` and/or `Interface.save()`. At the very least it would seem like the logic to remove tagged vlans when changing interface mode should no longer be needed?

I think this needs to go in the form and serializer validation since `tagged_vlans` is a `ManyToManyField`.

timizuoebideri1:
> > Based on the description of #2253 I would have expected some change to `Interface.clean()` and/or `Interface.save()`. At the very least it would seem like the logic to remove tagged vlans when changing interface mode should no longer be needed?
> 
> I think this needs to go in the form and serializer validation since `tagged_vlans` is a `ManyToManyField`.

Because the UI clears values in the `tagged_vlans` field when the `mode` is changed, it would be redundant in form clean.

nautobot/dcim/api/serializers.py
```diff
@@ -1025,6 +1025,16 @@ def validate(self, data):
                 )
 
         # Validate many-to-many VLAN assignments
+        mode = data.get("mode", getattr(self.instance, "mode", None))
+        has_tagged_vlans = data.get("tagged_vlans", False)
+        if not has_tagged_vlans and self.instance and self.instance.tagged_vlans.exists():
```
gsnider2195, line 1030: Should this only raise an exception if trying to set `tagged_vlans` but not setting the mode to `tagged`? Should changing the mode from `tagged` to `access` and not specifying any tagged vlans also generate an exception? That's how it works in the UI.

timizuoebideri1: Only when mode is set to `tagged` is a tagged VLAN necessary; why should this generate an exception if mode is changed to 'access' and tagged VLANs are not added?

nautobot/dcim/tests/test_models.py
```diff
@@ -1121,3 +1122,30 @@ def test_power_panel_validation(self):
...
+        interface = Interface.objects.create(
+            name="Int1",
+            type=InterfaceTypeChoices.TYPE_VIRTUAL,
+            device=device,
+        )
+        with self.assertRaises(ValidationError) as err:
+            interface.tagged_vlans.add(vlan1)
+        self.assertEqual(
+            err.exception.message_dict["tagged_vlans"][0], "Mode must be set to tagged when specifying tagged_vlans"
+        )
```
glennmatthews, line 1151: I still feel like there should be a test here where we start with an interface with mode `tagged` and `tagged_vlans` assigned, then change it to `access` without clearing `tagged_vlans`, which should result in an error (and not the current behavior of silently clearing the `tagged_vlans`?

timizuoebideri1: Perhaps adding to the documentation that changing the interface mode from tagged to something else clears all tagged VLANs.

If that procedure could be automated, I don't believe this is required.
```json
{
   "mode": "access",
   "tagged_vlans": null,
   "untagged_vlan": "...",
}
```

nautobot/dcim/api/serializers.py
gsnider2195: Sorry my message was confusing. I think we should remove this part

```py
self.instance and self.instance.tagged_vlans.exists()
```

With this behavior you have to specify `mode: tagged` on every API call to an interface with `tagged_vlans` right?

gsnider2195: What happens if an interface has `tagged_vlans` and is `mode: tagged` and you want to change it to `mode: access`?

glennmatthews: The request in #2253, as I understand it, would be that you need to explicitly clear the tagged_vlans as part of the mode change.

nautobot/dcim/models/device_components.py
```diff
@@ -519,6 +519,9 @@ def clean(self):
         if not self.mode and self.untagged_vlan is not None:
             raise ValidationError({"untagged_vlan": "Mode must be set when specifying untagged_vlan"})
 
+        if self.tagged_vlans.exists() and self.mode != InterfaceModeChoices.MODE_TAGGED:
+            raise ValidationError({"tagged_vlans": f"Clear tagged_vlans to set mode to {self.mode}"})
```
glennmatthews, line 523: Does this make the `self.tagged_vlans.clear()` in `save()` below now redundant? Should it be removed?

nautobot/virtualization/tests/test_models.py
```diff
@@ -69,3 +71,31 @@ def test_vm_duplicate_name_per_cluster(self):
...
+
+    def test_tagged_vlan_raise_error_if_mode_not_set_to_tagged(self):
+        interface = VMInterface.objects.create(virtual_machine=self.virtualmachine, name="Interface 1")
+        with self.assertRaises(ValidationError) as err:
+            interface.tagged_vlans.add(self.vlan)
+        self.assertEqual(
+            err.exception.message_dict["tagged_vlans"][0], "Mode must be set to tagged when specifying tagged_vlans"
+        )
+
+    def test_tagged_vlan_raise_error_if_mode_is_changed_withoout_cleating_tagged_vlans(self):
```
glennmatthews, line 92:
```suggestion
    def test_tagged_vlan_raise_error_if_mode_is_changed_without_clearing_tagged_vlans(self):
```

nautobot/dcim/tests/test_models.py
```diff
@@ -1168,6 +1170,49 @@ def test_power_panel_validation(self):
...
+            type=InterfaceTypeChoices.TYPE_VIRTUAL,
+            device=self.device,
+        )
+        with self.assertRaises(ValidationError) as err:
+            interface.tagged_vlans.add(self.vlan)
+        self.assertEqual(
+            err.exception.message_dict["tagged_vlans"][0], "Mode must be set to tagged when specifying tagged_vlans"
+        )
+
+    def test_tagged_vlan_raise_error_if_mode_is_changed_withoout_cleating_tagged_vlans(self):
```
glennmatthews, line 1201:
```suggestion
    def test_tagged_vlan_raise_error_if_mode_is_changed_without_clearing_tagged_vlans(self):
```

glennmatthews (approved):

nautobot/dcim/api/serializers.py
gsnider2195, line 1030: L1030 is doing a boolean check against `self.instance` but L1028 is not. Checking to see if this be a problem when creating a new interface

gsnider2195: Serializer `instance` is initialized to `None` so this is good.

nautobot/dcim/models/device_components.py
```diff
@@ -519,6 +519,9 @@ def clean(self):
         if not self.mode and self.untagged_vlan is not None:
             raise ValidationError({"untagged_vlan": "Mode must be set when specifying untagged_vlan"})
 
+        if self.tagged_vlans.exists() and self.mode != InterfaceModeChoices.MODE_TAGGED:
```
gsnider2195, line 522: If validation logic for `untagged_vlan` and `tagged_vlans` is going to exist in `BaseInterface` then those fields should be moved to `BaseInterface` or the validation logic moved to the models where those fields exist? New issue perhaps?

gsnider2195 (approved):

nautobot/dcim/models/device_components.py
gsnider2195: Would probably have to move to a mixin since they're related fields. It's probably not worth worrying about today.

phoenixswiss:
Hi

After implementing this fix, I'm unable to change an interface from mode 'tagged' with assigned vlans to mode 'access' in one api call (e.g. via pynautobot). I have to remove all the vlans from the interface in the first api call, save it and then change the mode to untagged and save it again.

The same issue occurs in the nautobot gui. 
On an interface configured with mode "tagged" and vlans assigned to aninterface, it's not possible to remove all vlans form the "Tagged VLANs" field and change the mode to "access" in one step. After doing so an pressing the update button, the interface edit page loads again without an invisible error message. The mode is set to "access" and the untagged vlan is also set to the expected values. If you change the mode back to "tagged" the additional "Tagged VLANs" field with the error message "Clear tagged_vlans to set mode to access" appears now. At this stage you can't modify the interface further. You have to press cancel on the interface edit page and open the interface again. Next all tagged vlans have to be removed from the interface and change must be first saved via the update button. Next you can change the interface mode to untagged and save it again. Thats not very elegant from a user perspective and doubles the amount of api calls if you have to change an interface mode from "tagged" to "untagged".

Is this an expected behaviour?

Thanks

Issue #2693, opened 2022-10-27
Dynamic Groups with nested object filters like `Location` don't work in child groups


jathanism:
### Environment
* Nautobot version (Docker tag too if applicable): 1.4.7
* Python version: 3.10
* Database platform, version: n/a
* Middleware(s): n/a

### Steps to Reproduce

This is a follow-up to https://github.com/nautobot/nautobot/pull/2665, which fixed the ability for Dynamic Groups to filter properly on nested object types like `Location` and `Region` by augmenting `TreeNodeMultipleChoiceFilter.filter()` to iterate the tree descendants (e.g. `Location A --> Location B`).

#### Create Locations

1. Create a `LocationType` called `My Region` and content-type `dcim | device`
2. Create a `LocationType` called `My Site` with parent set to `LocationType("MyRegion")` and content-type `dcim | device`

![image](https://user-images.githubusercontent.com/138052/198370544-4d605315-f380-4c89-a5a9-b7875f5dfbb4.png)

3. Create a new `Location` with `LocationType` of `My Region` called `Location A` with site `AMS01` (which is part of the demo dataset)
4. Create a new `Location` with `LocationType` of `My Region --> MySite` called `Location B` with parent `Location A` 

![image](https://user-images.githubusercontent.com/138052/198370969-9743eba9-d1ba-4a29-83fe-a00cad8f6bec.png)

#### Create Devices
1. Create a new `Device` named `new-device1` with site `AMS01` and location set to `Location A` 
2. Create a new `Device` named `new-device2` with site `AMS01` and location set to `Location B` 

![image](https://user-images.githubusercontent.com/138052/198371095-3d56afe3-cd44-4697-99e1-294ac4c3269f.png)

#### Create Dynamic Groups
1. Create a `DynamicGroup` with a filter for a `location = "Location A"` named `devices-location-a`

![image](https://user-images.githubusercontent.com/138052/198371211-a5b59e1a-0145-4942-8b65-abfba195e29a.png)

2. Create a `DynamicGroup` with an empty filter as a parent group named `location-parent`
3. Edit `location-parent` and set the first child group to `devices-location-a` with operator `Include (OR)`.

![image](https://user-images.githubusercontent.com/138052/198371323-53faa1ae-8cdc-41d8-af4c-c5f461cfe880.png)

### Expected Behavior

- Navigate to Dynamic Groups list view
- Observe that `devices-location-a` has 2 members
- Observe that `location-parent` also has 2 members

[...]

PR #2695, opened 2022-10-27
Fixed filtering by tree node objects in nested Dynamic Groups.

jathanism:
# Closes: #2693 
# What's Changed

- `TreeNodeMultipleChoiceFilter.filter()` refactored to call a new `generate_filter()` method that returns a `Q` objectthat includes filter predicates for all descendants.
- `DynamicGroup.generate_query_for_filter()` has been extended to look for `generate_query()` methods on filters and call that to include the generated `Q` object in group filters.

Proof that it works for Locations:

image

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/models/groups.py
+++ b/nautobot/extras/models/groups.py
@@ -495,13 +495,36 @@ class DynamicGroup(OrganizationalModel):
         """
         query = models.Q()
 
+        field_name = filter_field.field_name
+
+        # Attempt to account for `ModelChoiceFilter` where `to_field_name` MAY be set.
+        to_field_name = getattr(filter_field.field, "to_field_name", None)
+        if to_field_name is not None:
+            field_name = f"{field_name}__{to_field_name}"
+
+        lookup = f"{field_name}__{filter_field.lookup_expr}"
+
         # Explicitly call generate_query_{filter_method} for a method filter.
         if filter_field.method is not None and hasattr(filter_field.parent, "generate_query_" + filter_field.method):
             filter_method = getattr(filter_field.parent, "generate_query_" + filter_field.method)
-            query |= models.Q(filter_method(value))
+            query |= filter_method(value)
 
-        # In this case we want all values in a set union (boolean OR) because we want ANY of the
-        # filter values to match.
+        # Explicitly call `filter_field.generate_query` for a reversible filter.
+        elif hasattr(filter_field, "generate_query"):
+            filter_method = getattr(filter_field, "generate_query")
+
+            # Is this a list of strings? Well let's resolve it to related model objects so we can
+            # pass it to `generate_query` to get a correct Q object back out.
+            if value and isinstance(value, list) and isinstance(value[0], str):
+                related_model = self.model._meta.get_field(filter_field.field_name).related_model
+                lookup_kwargs = {f"{to_field_name}__in": value}
+                gq_value = related_model.objects.filter(**lookup_kwargs)
+            else:
+                gq_value = value
+            query |= filter_method(gq_value)
+
+        # For vanilla multiple-choice filters, we want all values in a set union (boolean OR)
+        # because we want ANY of the filter values to match.
         elif isinstance(filter_field, django_filters.MultipleChoiceFilter):
             for v in value:
                 query |= models.Q(**filter_field.get_filter_predicate(v))
@@ -511,14 +534,6 @@ class DynamicGroup(OrganizationalModel):
         # multiple-choice. This is safe for singular filters except `ModelChoiceFilter`, because they
         # do not support `to_field_name`.
         else:
-            field_name = filter_field.field_name
-
-            # Attempt to account for `ModelChoiceFilter` where `to_field_name` MAY be set.
-            to_field_name = getattr(filter_field.field, "to_field_name", None)
-            if to_field_name is not None:
-                field_name = f"{field_name}__{to_field_name}"
-
-            lookup = f"{field_name}__{filter_field.lookup_expr}"
             query |= models.Q(**{lookup: value})
 
         return query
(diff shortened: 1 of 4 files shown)


nautobot/extras/models/groups.py
```diff
@@ -495,13 +495,36 @@ def generate_query_for_filter(self, filter_field, value):
...
+
+        # Explicitly call `filter_field.generate_query` for a reversible filter.
+        elif hasattr(filter_field, "generate_query"):
+            filter_method = getattr(filter_field, "generate_query")
+
+            # Is this a list of strings? Well let's resolve it to related model objects so we can
+            # pass it to `generate_query` to get a correct Q object back out.
+            if value and isinstance(value, list) and isinstance(value[0], str):
+                related_model = self.model._meta.get_field(filter_field.field_name).related_model
+                lookup_kwargs = {f"{to_field_name}__in": value}
```
jathanism, line 520: This assertion is fragile. A list of strings might not necessarily coincide with a filter that has `to_field_name` set. This will be shaken out in testing.

jathanism:
So, there’s a chicken and egg problem here I’ve realized with the “forward” filtering using filtersets, and the “reverse” filtering using Qobjects and generate_query methodology.

It only really becomes apparent with the tree node filtering, so it may be less of a problem in the common case, but … I’m just gonna submit this PR for review without any new tests just now to show my hand for the moment while I contemplate this.

nautobot/extras/models/groups.py
```diff
@@ -495,13 +495,36 @@ def generate_query_for_filter(self, filter_field, value):
...
+        if to_field_name is not None:
+            field_name = f"{field_name}__{to_field_name}"
+
+        lookup = f"{field_name}__{filter_field.lookup_expr}"
+
         # Explicitly call generate_query_{filter_method} for a method filter.
         if filter_field.method is not None and hasattr(filter_field.parent, "generate_query_" + filter_field.method):
             filter_method = getattr(filter_field.parent, "generate_query_" + filter_field.method)
-            query |= models.Q(filter_method(value))
+            query |= filter_method(value)
```
jathanism, line 510: I realized that this was wrapping a `Q` object in a `Q` object, which while not broken, was redundant.

nautobot/extras/models/groups.py
jathanism: I'm realizing that this also may have implications for `SearchFilter`, since it also implements `generate_query` but its signature is different. Good times.

nautobot/extras/models/groups.py
```diff
@@ -495,13 +495,36 @@ def generate_query_for_filter(self, filter_field, value):
...
+
+        # Explicitly call `filter_field.generate_query` for a reversible filter.
+        elif hasattr(filter_field, "generate_query"):
+            filter_method = getattr(filter_field, "generate_query")
```
glennmatthews, line 514:
```suggestion
            filter_method = filter_field.generate_query
```

nautobot/extras/models/groups.py
```diff
@@ -495,13 +495,36 @@ def generate_query_for_filter(self, filter_field, value):
...
+
+            # Is this a list of strings? Well let's resolve it to related model objects so we can
+            # pass it to `generate_query` to get a correct Q object back out.
```
glennmatthews, line 517: Could we maybe put a concrete example in the comment here? I'll admit that `related_model = self.model._meta.get_field(filter_field.field_name).related_model` makes my head spin.

nautobot/extras/models/groups.py
```diff
@@ -511,14 +534,6 @@ def generate_query_for_filter(self, filter_field, value):
...
         # do not support `to_field_name`.
         else:
-            field_name = filter_field.field_name
-
-            # Attempt to account for `ModelChoiceFilter` where `to_field_name` MAY be set.
-            to_field_name = getattr(filter_field.field, "to_field_name", None)
-            if to_field_name is not None:
-                field_name = f"{field_name}__{to_field_name}"
-
-            lookup = f"{field_name}__{filter_field.lookup_expr}"
```
glennmatthews, line 521: Why's this being removed?

jathanism: I moved it.

nautobot/extras/models/groups.py
jathanism:
```python
                model_field = self.model._meta.get_field(filter_field.field_name)
                related_model = model_field.related_model
```
😉

nautobot/extras/models/groups.py
```diff
@@ -494,14 +494,37 @@ def generate_query_for_filter(self, filter_field, value):
...
-            query |= models.Q(filter_method(value))
+            query |= filter_method(value)
+
+        # Explicitly call `filter_field.generate_query` for a reversible filter.
+        elif hasattr(filter_field, "generate_query"):
+            # Is this a list of strings? Well let's resolve it to related model objects so we can
+            # pass it to `generate_query` to get a correct Q object back out. When values are being
+            # reconstructed from saved filters, lists of slugs are common e.g. (`{"site": ["ams01",
+            # "ams02"]}`, the value being a list of site slugs (`["ams01", "ams02"]`).
+            if value and isinstance(value, list) and isinstance(value[0], str):
```
bryanculver, line 517: Any chance that `value` may be a list but of length 0?

bryanculver (approved):
Validated manually. LGTM. Existing filters appear to work fine as well.

As usual, these types of meta-level programming stuff are voodoo.

Would be nice to add a test though.

nautobot/extras/models/groups.py
jathanism: Probably, but if that were the case an empty list would evaluate to `False` and wouldn't be in scope.

Issue #2419, opened 2022-09-14
null status on device interface created through template


Thetacz:
### Environment
* Nautobot version (Docker tag too if applicable): 1.4.1
* Python version: 3.10
* Database platform, version:
* Middleware(s):

#1753 introduced status field for interfaces. We have interface templates on some device types. If we create new device (tested via UI and csv import) with type containing interface template, the newly created device has interface without status set.
Creating device interface directly without status is not possible, the field is required.

### Steps to Reproduce
1. Create device type
2. Add interface to the device type (no field 'status' there)
3. Create device with that device type

### Expected Behavior
Newly created interface on the new device has status filled

### Observed Behavior
Newly created interface on the new device has empty status

### TODO
- Add status as an option to interface templates on device templates
- For migration, we should default existing templates to "Active" status, if it doesn't exist create it
- For backwards compatibility, no specifying a status should default to "Active" status, throw error if it does not exist

glennmatthews:
Thanks for the report! A "simple" fix might be to just set status Active for all new interfaces created from templates; but I wonder if a more desirable solution might be to allow individual interface templates to actually specify their default creation status.

martin-networker:
@glennmatthews Device type/template usually contains only physical interfaces because virtual interfaces vary device by device. With physical interfaces I doubt that any customer will have same sort of ports active from implementation. So as per my opinion the solution: set status Active for all new interfaces created from templates is great.

Thetacz:
Thanks for the fast reply! The extra status on device templates is not much usable for us, but others might like it. I'll leave that part completely on you :) 
just IMHO the fix should contain DB migration to add missing statuses, as #1753 added 'active' statuses to all interfaces when introducing the filed.

glennmatthews:
Good points, thank you both! I suspect that rather than a new DB migration, we'll want to add a management command similar to `nautobot-server fix_custom_fields`, `nautobot-server trace_paths`, etc. that can be run as a one-off (or as part of the `nautobot-server post_upgrade` command) to fix up any interfaces that may be missing statuses.

PR #2698, opened 2022-10-28
Fix interface status missing when creating a device with a device role associated to a InterfaceTemplate

timizuoebideri1:
# Closes: #2419 
# What's Changed

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/models/device_component_templates.py
+++ b/nautobot/dcim/models/device_component_templates.py
@@ -10,12 +10,13 @@ from nautobot.dcim.choices import (
     PowerOutletTypeChoices,
     PowerOutletFeedLegChoices,
     InterfaceTypeChoices,
+    InterfaceStatusChoices,
     PortTypeChoices,
 )
 
 from nautobot.core.models import BaseModel
 from nautobot.dcim.constants import REARPORT_POSITIONS_MAX, REARPORT_POSITIONS_MIN
-from nautobot.extras.models import ChangeLoggedModel, CustomField, CustomFieldModel, RelationshipModel
+from nautobot.extras.models import ChangeLoggedModel, CustomField, CustomFieldModel, RelationshipModel, Status
 from nautobot.extras.utils import extras_features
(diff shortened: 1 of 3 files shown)


nautobot/dcim/models/device_component_templates.py
```diff
@@ -258,11 +259,16 @@ class Meta:
         unique_together = ("device_type", "name")
 
     def instantiate(self, device):
+        try:
+            status = Status.objects.get_for_model(Interface).get(slug=InterfaceStatusChoices.STATUS_ACTIVE)
+        except Status.DoesNotExist:
+            status = Status.objects.get_for_model(Interface).first()
```
glennmatthews, line 265: I know we're moving away from `cacheops` - I wonder if we should cache this in memory somewhere? When instantiating a DeviceType with a couple hundred InterfaceTemplates, is this going to hit the database each time?

nautobot/dcim/tests/test_models.py
```diff
@@ -147,6 +147,34 @@ def test_instantiate_model(self):
         self.assertEqual(Interface.objects.get(pk=interfaces[1].pk).cf["field_3"], "value_3")
 
 
+class InterfaceTemplateTestCase(TestCase):
```
gsnider2195, line 150: Can you add a test for when the "active" status doesn't exist?

nautobot/dcim/models/device_component_templates.py
timizuoebideri1: Do you recommend that I use an alternative memory cache, such as `django.core.cache.backends.locmem.LocMemCache`?
Or i add a TODO for this

glennmatthews: I wasn't thinking anything so fancy - I was thinking more of an approach like we have for `Prefix.STATUS_CONTAINER` where we just look it up and store it on the class for future re-use.

glennmatthews (approved):

jathanism (approved):

Issue #2494, opened 2022-09-26
Custom field regex validation doesn't work for URL field types


gsnider2195:
### Environment
* Nautobot version (Docker tag too if applicable): v1.5.0a0
* Python version: 3.10.7
* Database platform, version: Postgres 13
* Middleware(s):

### Steps to Reproduce
1. Create a URL custom field with regex validation (ex ^https)
2. Update custom field on an associated object with a non-matching value (ex http://www.abc.com)

### Expected Behavior
Validation error

### Observed Behavior
Custom field saves successfully

Issue #2489, opened 2022-09-26
Custom field regex validation always anchors to beginning of string


gsnider2195:
### Environment
* Nautobot version (Docker tag too if applicable): v1.5.0a0
* Python version: 3.10.7
* Database platform, version: Postgres 13
* Middleware(s):

### Steps to Reproduce
1. Create a custom field with type Text and regex validation of string `foo`
2. Update custom field on an object to `barfoo`

### Expected Behavior
`barfoo` matches `foo` (regex is not anchored on beginning of line)

### Observed Behavior
`barfoo` does not match `foo` because Nautobot is using `re.match` instead of `re.search` [here](https://github.com/nautobot/nautobot/blob/633a95d3befaa4440a37bd36f377de9878623000/nautobot/extras/models/customfields.py#L535) and [here](https://github.com/nautobot/nautobot/blob/633a95d3befaa4440a37bd36f377de9878623000/nautobot/extras/models/customfields.py#L628)

PR #2699, opened 2022-10-28
fix custom field regex validation bugs #2489 and #2494

gsnider2195:
# Closes: #2489, Closes: #2494
# What's Changed

- Add regex validation to URL custom fields
- Change regex validation from `re.match` to `re.search` to properly match the [`RegexValidator`](https://docs.djangoproject.com/en/3.2/ref/validators/#regexvalidator) used in the form and to allow the regex to match anywhere on the string instead of anchored at the beginning.
- Add more testing for regex validation on the `CustomField` and `CustomFieldChoice` models.

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/changes/2489.fixed
+++ b/changes/2489.fixed
@@ -1,0 +1,1 @@
+Fixed custom field regex validation always anchored to beginning of string.
--- a/changes/2494.fixed
+++ b/changes/2494.fixed
@@ -1,0 +1,1 @@
+Fixed custom field regex validation not being enforced on URL custom fields.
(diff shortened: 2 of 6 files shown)


changes/2489.fixed
```diff
@@ -0,0 +1 @@
+Fixed custom field regex validation always anchored to beginning of string.
```
glennmatthews, line 1: This is arguably a behavior change rather than a bug fix. It's going to make existing regex rules less strict?

gsnider2195: I would argue that it was broken since that behavior wasn't documented anywhere and the form validator uses `re.search`. I guess it could have been worked around by starting the regex with `.*` and this change wouldn't break that workaround.

gsnider2195: Happy to discuss though

glennmatthews: I didn't know that about the form field validator. In that case I'm fine with treating this as a bug fix.

glennmatthews (approved):

Issue #2256, opened 2022-08-18
Site latitude or longitude doesn't work with float


joewesch:
### Environment
* Python version: 3.7.13
* Nautobot version: 1.4.0

### Steps to Reproduce
```
>>> from nautobot.dcim.models import Site
>>> site = Site.objects.first()
>>> site.longitude = 33.33
>>> site.longitude
33.33
>>> site.validated_save()
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/nautobot/core/models/__init__.py", line 51, in validated_save
    self.full_clean()
  File "/usr/local/lib/python3.7/site-packages/django/db/models/base.py", line 1251, in full_clean
    raise ValidationError(errors)
django.core.exceptions.ValidationError: {'longitude': ['Ensure that there are no more than 6 decimal places.']}
>>>
```

### Expected Behavior
I expected the longitude to be accepted since it is less than 6 decimal places

### Observed Behavior
It threw a Validation error that there were more than 6 decimal places.

### Analysis
It seems that since the field is a Decimal field it only accepts strings:
```
>>> site.longitude = "33.33"
>>> site.longitude
'33.33'
>>> site.validated_save()
>>> site.longitude
Decimal('33.33')
>>>
```

### TODO

- In validate/clean, help user by converting Floats inputted into a Decimal. No need to throw error for loss of precision, 6th decimal place precision gets us down to 10cm.

jathanism:
While this is a user experience surprise, the root cause here is actually because a `float` and `Decimal` are not as precise:

```
In [3]: import decimal

In [4]: decimal.Decimal(33.33)
Out[4]: Decimal('33.3299999999999982946974341757595539093017578125')

In [5]: decimal.Decimal("33.33")
Out[5]: Decimal('33.33')
```

The former is why this error about decimal places occurs.

PR #2706, opened 2022-11-01
Fix Site latitude or longitude float error

timizuoebideri1:
# Closes: #2256 
# What's Changed

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/2706.fixed
+++ b/changes/2706.fixed
@@ -1,0 +1,1 @@
+Fixed site latitude or longitude field raise error if value specified is a float.
--- a/changes/2706.fixed
+++ b/changes/2706.fixed
@@ -1,1 +1,1 @@
-Fixed site latitude or longitude field raise error if value specified is a float.
+Fixed a ValidationError that could occur when specifying Site latitude/longitude as a float.
--- a/nautobot/dcim/models/sites.py
+++ b/nautobot/dcim/models/sites.py
@@ -215,3 +215,10 @@ class Site(PrimaryModel, StatusModel):
             self.contact_email,
             self.comments,
         )
+
+    def clean_fields(self, exclude=None):
+        if self.longitude is not None:
+            self.longitude = str(self.longitude)
+        if self.latitude is not None:
+            self.latitude = str(self.latitude)
+        super().clean_fields(exclude)
--- a/nautobot/dcim/models/sites.py
+++ b/nautobot/dcim/models/sites.py
@@ -218,8 +218,11 @@ class Site(PrimaryModel, StatusModel):
 
     def clean_fields(self, exclude=None):
         """Explicitly convert latitude/longitude to strings to avoid floating-point precision errors."""
+
         if self.longitude is not None:
-            self.longitude = str(self.longitude)
+            decimal_places = self._meta.get_field("longitude").decimal_places
+            self.longitude = f"{self.longitude:.{decimal_places}f}"
         if self.latitude is not None:
-            self.latitude = str(self.latitude)
+            decimal_places = self._meta.get_field("latitude").decimal_places
+            self.latitude = f"{self.latitude:.{decimal_places}f}"
         super().clean_fields(exclude)
--- a/nautobot/dcim/models/sites.py
+++ b/nautobot/dcim/models/sites.py
@@ -217,6 +217,7 @@ class Site(PrimaryModel, StatusModel):
         )
 
     def clean_fields(self, exclude=None):
+        """Explicitly convert latitude/longitude to strings to avoid floating-point precision errors."""
         if self.longitude is not None:
             self.longitude = str(self.longitude)
         if self.latitude is not None:
--- a/nautobot/dcim/tests/test_models.py
+++ b/nautobot/dcim/tests/test_models.py
@@ -1136,5 +1136,5 @@ class SiteTestCase(TestCase):
         )
         site.validated_save()
 
-        self.assertTrue(site.longitude, 55.22)
-        self.assertTrue(site.latitude, 55.22)
+        self.assertEqual(site.longitude, 55.22)
+        self.assertEqual(site.latitude, 55.22)
--- a/nautobot/dcim/tests/test_models.py
+++ b/nautobot/dcim/tests/test_models.py
@@ -1121,3 +1121,19 @@ class PowerPanelTestCase(TestCase):
             'Rack group "Rack Group 1" belongs to a location ("Location 2") that does not contain "Location 1"',
             str(cm.exception),
         )
+
+
+class SiteTestCase(TestCase):
+    def test_latitude_or_longitude(self):
+        """Test latitude and longitude is parsed to sting"""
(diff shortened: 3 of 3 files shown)


changes/2706.fixed
```diff
@@ -0,0 +1 @@
+Fixed site latitude or longitude field raise error if value specified is a float.
```
glennmatthews, line 1:
```suggestion
Fixed a ValidationError that could occur when specifying Site latitude/longitude as a float.
```

nautobot/dcim/models/sites.py
```diff
@@ -215,3 +215,10 @@ def to_csv(self):
             self.contact_email,
             self.comments,
         )
+
+    def clean_fields(self, exclude=None):
+        if self.longitude is not None:
+            self.longitude = str(self.longitude)
```
glennmatthews, line 221: Is this sufficient or do we need to explicitly specify the target precision when converting to a string? What happens if you try to set `longitude = 0.123456789" for example?

nautobot/dcim/models/sites.py
```diff
@@ -215,3 +215,10 @@ def to_csv(self):
             self.contact_email,
             self.comments,
         )
+
+    def clean_fields(self, exclude=None):
```
glennmatthews, line 219:
```suggestion
    def clean_fields(self, exclude=None):
        """Explicitly convert latitude/longitude to strings to avoid floating-point precision errors."""
```

nautobot/dcim/tests/test_models.py
```diff
@@ -1121,3 +1121,19 @@ def test_power_panel_validation(self):
...
+        site = Site(
+            name="Site A",
+            slug="site-a",
+            status=active_status,
+            longitude=55.22,
+            latitude=55.22,
+        )
+        site.validated_save()
+
+        self.assertTrue(site.present_in_database)
```
glennmatthews, line 1139: Doesn't seem like it's necessary/relevant? Unless maybe you also want to check the values that `site.longitude` and `site.latitude` return after cleaning and saving?

nautobot/dcim/models/sites.py
HanlinMiao: Yeah I think there is a 8 digit limit on both of these fields.

timizuoebideri1: Raising an error whenever the maximum number of digits is reached makes perfect sense.

nautobot/dcim/tests/test_models.py
```diff
@@ -1121,3 +1121,20 @@ def test_power_panel_validation(self):
...
+            name="Site A",
+            slug="site-a",
+            status=active_status,
+            longitude=55.22,
+            latitude=55.22,
+        )
+        site.validated_save()
+
+        self.assertTrue(site.longitude, 55.22)
+        self.assertTrue(site.latitude, 55.22)
```
glennmatthews, line 1140:
```suggestion
        self.assertEqual(site.longitude, 55.22)
        self.assertEqual(site.latitude, 55.22)
```

timizuoebideri1: 😅

nautobot/dcim/models/sites.py
jathanism: This is strange to me because it's a `DecimalField` which means the internal value is a `decimal.Decimal` object.

There appears to be some incongruence here in these fields as defined on the model, could this be the culprit?

```python
    latitude = models.DecimalField(
        max_digits=8,
        decimal_places=6,
        blank=True,
        null=True,
        help_text="GPS coordinate (latitude)",
    )
    longitude = models.DecimalField(
        max_digits=9,
        decimal_places=6,
        blank=True,
        null=True,
        help_text="GPS coordinate (longitude)",
    )
```

It should be `8`, but our field has it defined as `9`. https://en.wikipedia.org/wiki/Decimal_degrees#Precision

nautobot/dcim/tests/test_models.py
```diff
@@ -1121,3 +1121,20 @@ def test_power_panel_validation(self):
             'Rack group "Rack Group 1" belongs to a location ("Location 2") that does not contain "Location 1"',
             str(cm.exception),
         )
+
+
+class SiteTestCase(TestCase):
+    def test_latitude_or_longitude(self):
+        """Test latitude and longitude is parsed to sting"""
```
jathanism, line 1128:
```suggestion
        """Test latitude and longitude is parsed to string."""
```

Good to have a test case here, however, its his correct? The test values here are floats, and since this is a `Decimal` field, the more decimal places, the less precision there will be unless using `Decimal` objects, yes?

nautobot/dcim/models/sites.py
glennmatthews: I'd assumed `max_digits` was total digits before and after the decimal, i.e. `123.456789` with 6 decimal places but 9 digits?

jathanism: Ah yes, that vibes. 3 + 6 since that falls under "individual humans" in the table I linked.

nautobot/dcim/tests/test_models.py
timizuoebideri1: Yes, but for a test case this should do right?

nautobot/dcim/models/sites.py
jathanism: The real root cause is that a `float` can in fact be cast to a `Decimal`, except that the precision is lost:

```python
In [16]: decimal.Decimal(55.22)
Out[16]: Decimal('55.219999999999998863131622783839702606201171875')

In [17]: decimal.Decimal("55.22")
Out[17]: Decimal('55.22')
```

So when you try to pass a `float` through a `DecimalValidator` (the validator used underneath by `DecimalField`, this happens:

```python
In [20]: from django.core.validators import DecimalValidator

In [21]: validator = DecimalValidator(max_digits=9, decimal_places=6)

In [22]: validator(decimal.Decimal(55.22))
---------------------------------------------------------------------------
ValidationError                           Traceback (most recent call last)
Cell In [22], line 1
----> 1 validator(decimal.Decimal(55.22))

File ~/Library/Caches/pypoetry/virtualenvs/nautobot-b2ttWva6-py3.10/lib/python3.10/site-packages/django/core/validators.py:476, in DecimalValidator.__call__(self, value)
    473 whole_digits = digits - decimals
    475 if self.max_digits is not None and digits > self.max_digits:
--> 476     raise ValidationError(
    477         self.messages['max_digits'],
    478         code='max_digits',
    479         params={'max': self.max_digits, 'value': value},
    480     )
    481 if self.decimal_places is not None and decimals > self.decimal_places:
    482     raise ValidationError(
    483         self.messages['max_decimal_places'],
    484         code='max_decimal_places',
    485         params={'max': self.decimal_places, 'value': value},
    486     )

ValidationError: ['Ensure that there are no more than 9 digits in total.']
```

glennmatthews (approved):

Issue #2671, opened 2022-10-21
nautobot-dev (final-dev) docker image doesn't start


joewesch:
### Environment
* Nautobot version (Docker tag too if applicable): 1.4.2 -> latest
* Python version: 3.7 -> 3.10
* Database platform, version: n/a
* Middleware(s): n/a

### Steps to Reproduce
1. docker run --rm -it ghcr.io/nautobot/nautobot-dev:1.4.7-py3.10

### Expected Behavior
Nautobot Dev should start up

### Observed Behavior
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/nautobot/extras/plugins/utils.py", line 64, in load_plugin
    plugin = importlib.import_module(plugin_name)
  File "/usr/local/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'example_plugin'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/nautobot/core/runner/runner.py", line 116, in settings_callback
    initializer(
  File "/usr/local/lib/python3.10/site-packages/nautobot/core/cli.py", line 176, in _configure_settings
    load_plugins(settings)
  File "/usr/local/lib/python3.10/site-packages/nautobot/extras/plugins/utils.py", line 54, in load_plugins
    load_plugin(plugin_name, settings)
  File "/usr/local/lib/python3.10/site-packages/nautobot/extras/plugins/utils.py", line 67, in load_plugin
    raise PluginNotFound(
nautobot.extras.plugins.exceptions.PluginNotFound: Unable to import plugin example_plugin: Module not found. Check that the plugin module has been installed within the correct Python environment.
```

glennmatthews:
Thanks for the report! Hypothesis:

- `example_plugin` is copied in under `/source/examples/example_plugin` and installed in the `dependencies-dev-python` target in editable mode
- both `dev` and `final-dev` targets use `development/nautobot_config.py` as their Nautobot config file, which includes the example plugin
- `dev` (what we use for Nautobot core development) works just fine this way.
[...]

PR #2715, opened 2022-11-02
Use same nautobot_config.py in final and final-dev images

glennmatthews:
# Closes: #2671 
# What's Changed

- Previously the `final-dev` images (published as `nautobot-dev`) were using the `development/nautobot_config.py` file as their default configuration. This caused failures in some downstream dev environments (such as `pynautobot`) that didn't provide their own `nautobot_config.py` overrides, because that config file enables the `example_plugin` in Nautobot, which isn't actually included in the `final-dev` image build. I've changed the Docker build for `final-dev` so that it now defaults to the same templated file (from `nautobot-server init`) as the `final` image, which should unblock these downstream dev environments. (TODO below - I want to actually try this image with the `pynautobot` dev environment)
- I found it repeatedly confusing that our local docker-compose dev environment defaults to building and using `dev` images with the label/tag `networktocode/nautobot-dev-py${PYTHON_VER}:local`, since these images shouldn't ever be published to dockerhub, and the label is confusing against the `networktocode/nautobot-dev:${VER}-py${PYTHON_VER}` images which actually **are** published. I changed the dev image label to instead be `local/nautobot-dev:local-py${PYTHON_VER}` to make its local purpose more clear.
- Similarly, I updated the `docker-compose.build.yml` that can be used for running with locally built `final` images to label these images as `local/nautobot-final:local-py${PYTHON_VER}` instead of `networktocode/nautobot-py${PYTHON_VER}:local`, which was also confusing against the actual published images on dockerhub at `networktocode/nautobot:${VER}-py${PYTHON_VER}`.
- I also renamed `development/docker-compose.build.yml` to `development/docker-compose.final.yml` and updated the docs accordingly.
- We didn't have a docker-compose file for building/testing with `final-dev` targets, so I've added `development/docker-compose.final-dev.yml`.

# TODO

- [ ] Test updated `final-dev` Docker image with `pynautobot` dev environment.
- n/a Unit, Integration Tests

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -250,9 +250,6 @@ RUN npm install --global markdownlint-cli@0.31.1
 # /source from build-nautobot will include docs from build-docs as well
 COPY --from=build-nautobot /source /source
 
-# TODO Use nautobot init to generate the same config for all use cases
-COPY development/nautobot_config.py /opt/nautobot/nautobot_config.py
-
 # Run Nautobot development server by default
 EXPOSE 8080
 CMD ["nautobot-server", "runserver", "0.0.0.0:8080", "--insecure"]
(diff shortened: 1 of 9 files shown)


jathanism (approved):

docker/Dockerfile
```diff
@@ -308,7 +311,7 @@ USER nautobot
 
 WORKDIR /opt/nautobot
 
-RUN nautobot-server init
+COPY --from=final-dev /opt/nautobot/nautobot_config.py /opt/nautobot/nautobot_config.py
```
Yannis100, line 314: Got an issue `cannot create /opt/nautobot/nautobot_config.py: Permission denied`
Running `cat /opt/nautobot/nautobot_config.append.py >> /opt/nautobot/nautobot_config.py` while building homemade Dockerfile using Nautobot docker image as base

I think adding `--chown=nautobot` should solve the issue

Issue #2740, opened 2022-11-04
NautobotUIViewSet should include routes for change log and notes


lampwins:
### As ...

P.D. - Plugin Developer

### I want ...

The NautobotUIViewSet to automatically include routes for change log and notes

### So that ...

I do not have manually account for these routes in the traditional pattern when using the viewset

### I know this is done when...

- Routes for notes and changelog are automatically accounted for when I use the viewset and a corresponding model
- The viewset router can handle cases where the model does not support change logging or notes appropriately

### Optional - Feature groups this request pertains to.

- Plugins (and other Extensibility)

### Database Changes

None

### External Dependencies

None

smk4664:
I would like to take this, but on the condition I can call in backup if I get too far over my head.

Issue #1813, opened 2022-05-25
Create/Update Example Plugin to use NautobotUIViewSet


lampwins:
The examples/example_plugin plugin should use NautobotUIViewSet since that's the approach we're recommending for all new plugin development, and this is meant to serve as an example of good plugin development practices.

bryanculver:
Closed via #2752

PR #2752, opened 2022-11-05
Create ObjectChangeLog and Notes Viewset

smk4664:
# Closes: #2740 #1813 
# What's Changed
Add ObjectNotesView and ObjectChangeLogView to NautobotUIViewSet and NautobotUIViewSetRouter.

I discovered while trying to update the Example Plugin to remove the notes/changelog url that it was not using the NautbotUIViewSet. I went ahead and made those changes as well.

While the issue called out "The viewset router can handle cases where the model does not support change logging or notes appropriately" I noticed that the Plugin Development guide calls out how to remove urls (by inheriting from the mixins instead of NautobotUIViewSet), in addition get_notes_url and get_changelog_url will not return a URL, so there will be no links to the notes tab or changelog tab.

## Remaining Work - Possibly

I noticed that Example Plugin uses Dynamic Groups, which wasn't originally in the scope of work. Should this be a separate issue or should I include it in these changes.

# TODO

- [x] Unit, Integration Tests


bryanculver:
👏

nautobot/core/views/mixins.py
```diff
@@ -903,3 +911,72 @@ def perform_bulk_update(self, request, **kwargs):  # pylint: disable=arguments-d
...
+        # v2 TODO(jathan): Replace prefetch_related with select_related
+        objectchanges = (
+            ObjectChange.objects.restrict(request.user, "view")
+            .prefetch_related("user", "changed_object_type")
+            .filter(
+                Q(changed_object_type=content_type, changed_object_id=obj.pk)
+                | Q(related_object_type=content_type, related_object_id=obj.pk)
+            )
+        )
+        objectchanges_table = ObjectChangeTable(data=objectchanges, orderable=False)
```
smk4664, line 935: Part of me wants to move this into the `renders.NautobotHTMLRenderer.construct_table` function, but this table deviates a little from that. What do y'all think? Would it be worth the effort?

nautobot/core/views/mixins.py
```diff
@@ -903,3 +911,72 @@ def perform_bulk_update(self, request, **kwargs):  # pylint: disable=arguments-d
...
+    def notes(self, request, *args, **kwargs):
+        obj = self.get_object()
+        content_type = ContentType.objects.get_for_model(obj)
+        notes_form = NoteForm(
+            initial={
+                "assigned_object_type": content_type,
+                "assigned_object_id": obj.pk,
+            }
+        )
+        notes_table = NoteTable(obj.notes)
```
smk4664, line 967: See other comment about `construct_table`.

HanlinMiao:
Nice work!

nautobot/core/views/mixins.py
HanlinMiao: IMHO, I think we should try to move this into the `construct_table` function.
Just stealing from @jathanism: renderer should do all the heavy lifting and keep the mixins as lightweight as possible.

smk4664: Updated.

nautobot/core/views/mixins.py
smk4664: Updated

nautobot/core/views/mixins.py
```diff
@@ -903,3 +914,33 @@ def perform_bulk_update(self, request, **kwargs):  # pylint: disable=arguments-d
...
+    """
+    UI mixin to list a model's changelog queryset
+    """
+
+    base_template = None
+
+    def changelog(self, request, *args, **kwargs):
+        data = {
+            "base_template": self.base_template,
+            "template": "extras/object_changelog.html",
```
HanlinMiao, line 929: Not blocking and def a nitpick, but can we use the generic `get_template_name()` here instead of hardcoding the template name?

nautobot/core/views/mixins.py
```diff
@@ -903,3 +914,33 @@ def perform_bulk_update(self, request, **kwargs):  # pylint: disable=arguments-d
...
+    """
+    UI Mixin for an Object's Notes.
+    """
+
+    base_template = None
+
+    def notes(self, request, *args, **kwargs):
+        data = {
+            "base_template": self.base_template,
+            "template": "extras/object_notes.html",
```
HanlinMiao, line 944: Not blocking and def a nitpick, but can we use the generic `get_template_name()` here instead of hardcoding the template name?

HanlinMiao (commented):
Awesome work thank you! just some nitpicks

examples/example_plugin/example_plugin/views.py
```diff
@@ -67,98 +69,38 @@ def post(self, request):
...
+):
+    bulk_update_form_class = forms.AnotherExampleModelBulkEditForm
+    filterset_class = filters.AnotherExampleModelFilterSet
+    filterset_form_class = forms.AnotherExampleModelFilterForm
+    form_class = forms.AnotherExampleModelForm
     queryset = AnotherExampleModel.objects.all()
+    serializer_class = serializers.AnotherExampleModelSerializer
+    table_class = tables.AnotherExampleModelTable
+    lookup_field = "pk"
+    action_buttons = ["add", "export"]
```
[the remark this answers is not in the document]
smk4664: Sure! I just copied the attributes from the docs (I think).

examples/example_plugin/example_plugin/templates/example_plugin/anotherexamplemodel_retrieve.html
```diff
@@ -0,0 +1,88 @@
+<!-- this is a view demonstrating the capabilities of generic/object_detail.thml -->
+{% extends 'generic/object_detail.html' %}
```
glennmatthews, line 2:
```suggestion
<!-- this is a view demonstrating the capabilities of generic/object_retrieve.html -->
{% extends 'generic/object_retrieve.html' %}
```

examples/example_plugin/example_plugin/templates/example_plugin/examplemodel_retrieve.html
```diff
@@ -0,0 +1,87 @@
+<!-- this is a view demonstrating the capabilities of generic/object_detail.thml -->
+{% extends 'generic/object_detail.html' %}
```
glennmatthews, line 2:
```suggestion
<!-- this is a view demonstrating the capabilities of generic/object_retrieve.html -->
{% extends 'generic/object_retrieve.html' %}
```

nautobot/core/views/mixins.py
```diff
@@ -369,6 +378,8 @@ class BulkCreateForm(BootstrapMixin, Form):
                 csv_file = CSVFileField(from_form=self.bulk_create_form_class)
 
             form_class = BulkCreateForm
+        elif self.action == "notes":
+            form_class = NoteForm
         else:
             form_class = getattr(self, f"{self.action}_form_class", None)
```
glennmatthews, line 384: Nit, but maybe we could default `notes_form_class = NoteForm` on the class and avoid the need to add an `elif self.action == "notes"`? Would allow for the (admittedly unlikely) case where a view wants to provide a custom notes form too.

nautobot/core/views/mixins.py
bryanculver: Because we are essentially making `changelog` and `notes` an "action" for NautobotUIViewSets I disagree: https://github.com/nautobot/nautobot/blob/develop/nautobot/core/views/mixins.py#L316

`get_template_name` should know about these templates and we should align them with the template patterns we've set forth with NautobotUIViewSets.

nautobot/core/views/mixins.py
smk4664: On the `NautobotUIViewSet` class?

smk4664:
Retargetting Develop.

nautobot/docs/plugins/development.md
```diff
@@ -1124,6 +1124,9 @@ urlpatterns = [
 urlpatterns += router.urls
 ```
 
++++ 1.5.0
+    Changelog and Notes urls are now provided in the NautobotUIViewSet and NautobotUIViewSetRouter.
```
HanlinMiao, line 1128:
```suggestion
    Changelog and Notes views and urls are now provided in the NautobotUIViewSet and NautobotUIViewSetRouter.
```

HanlinMiao (approved):
This is good to go IMO

nautobot/docs/plugins/development.md
smk4664: The reason for this comment was to point out how you don't have to add the URLs for Notes and Changelog, then I think this was updated to also add in the NautobotUIViewSet, so now it makes more since to add what you commented.

nautobot/core/templates/generic/object_notes.html
```diff
@@ -0,0 +1,33 @@
...
+                    <div class="panel-body">
+                    {% csrf_token %}
+                    {% render_form form %}
+                    </div>
+                </div>
+                </div>
+                <div class="col-md-6 col-md-offset-3 text-right" style="margin-bottom: 20px;">
+                <button type="submit" class="btn btn-primary">Create</button>
+                <a href="{{ object.get_absolute_url }}" class="btn btn-default">Cancel</a>
+                </div>
```
glennmatthews, line 26: Readability nitpick:

```suggestion
                <div class="col-md-6 col-md-offset-3">
                    <h3>Add a new note</h3>
                    <div class="panel panel-default">
                        <div class="panel-heading">
                            <strong>Note</strong>
                        </div>
                        <div class="panel-body">
                            {% csrf_token %}
                            {% render_form form %}
                        </div>
                    </div>
                </div>
                <div class="col-md-6 col-md-offset-3 text-right" style="margin-bottom: 20px;">
                    <button type="submit" class="btn btn-primary">Create</button>
                    <a href="{{ object.get_absolute_url }}" class="btn btn-default">Cancel</a>
                </div>
```

nautobot/core/views/viewsets.py
```diff
@@ -9,8 +9,11 @@ class NautobotUIViewSet(
     mixins.ObjectBulkDestroyViewMixin,
     mixins.ObjectBulkCreateViewMixin,
     mixins.ObjectBulkUpdateViewMixin,
+    mixins.ObjectChangeLogViewMixin,
+    mixins.ObjectNotesViewMixin,
```
glennmatthews, line 13: Does this pose any issue for existing plugins that are using the 1.4 version of `NautobotUIViewSet` and separately implementing the changelog and notes views as was previously needed? Want to make sure we're not doing something backwards-incompatible.

nautobot/core/views/routers.py
```diff
@@ -79,4 +79,22 @@ class NautobotUIViewSetRouter(SimpleRouter):
...
+        ),
+        Route(
+            url=r"^{prefix}/{lookup}/notes/$",
+            mapping={
+                "get": "notes",
+            },
+            name="{basename}_notes",
+            detail=True,
+            initkwargs={"suffix": "Notes"},
+        ),
```
glennmatthews, line 99: Does this pose any issue for existing plugins that are using the 1.4 version of `NautobotUIViewSetRouter` and separately implementing the changelog and notes URL patterns as was previously needed? Want to make sure we're not doing something backwards-incompatible.

nautobot/docs/plugins/development.md
```diff
@@ -1124,6 +1124,9 @@ urlpatterns = [
 urlpatterns += router.urls
 ```
 
++++ 1.5.0
+    Changelog and Notes views and urls are now provided in the NautobotUIViewSet and NautobotUIViewSetRouter.
```
glennmatthews, line 1128:
```suggestion
+++ 1.5.1
    Changelog and Notes views and URLs are now provided in the NautobotUIViewSet and NautobotUIViewSetRouter.
```

nautobot/core/views/viewsets.py
smk4664: I just tested with Providers (by adding back in the path to urls.py, and it doesn't break anything. I am not sure which view it is loading... But nothing broke.

nautobot/core/views/routers.py
smk4664: I just tested with Providers (by adding back in the path to urls.py, and it doesn't break anything. I am not sure which view it is loading... But nothing broke.

nautobot/core/views/renderers.py
```diff
@@ -233,6 +261,20 @@ def get_context(self, data, accepted_media_type, renderer_context):
...
+                        "active_tab": "changelog",
+                    }
+                )
+            elif view.action == "notes":
+                context.update(
+                    {
+                        "base_template": get_base_template(data.get("base_template"), model),
+                        "active_tab": "notes",
+                    }
+                )
```
jathanism, line 277: Since these are pretty much identical, couldn't we get away with this?
```suggestion
            elif view.action in ["changelog", "notes"]:
                context.update(
                    {
                        "base_template": get_base_template(data.get("base_template"), model),
                        "active_tab": view.action,
                    }
                )
```

nautobot/core/views/mixins.py
```diff
@@ -71,6 +75,7 @@ class NautobotViewSetMixin(GenericViewSet, AccessMixin, GetReturnURLMixin, FormV
     # serializer_class has to be specified to eliminate the need to override retrieve() in the RetrieveModelMixin for now.
     serializer_class = None
     table_class = None
+    notes_form_class = NoteForm
```
jathanism, line 78: Why is this defined here? I don't see it being used anywhere else?

smk4664: get_form_class uses it. Glenn asked me to move it here.

jathanism: Ah okay, I didn't catch that before. Thanks!

```python
            form_class = getattr(self, f"{self.action}_form_class", None)
```

glennmatthews (approved):

Issue #2744, opened 2022-11-04
Bulk Create/Edit of Required Relationships Doesn't Peform Enforcement


bryanculver:
### Environment
* Nautobot version (Docker tag too if applicable): `next`, `1.5`
* Python version: 3.7+
* Database platform, version: All
* Middleware(s): None

### Steps to Reproduce
1. Mark a relationship as required
2. Perform a bulk API/UI operation on required fields (adding/removing)

### Expected Behavior

The required relationships should be enforced.

### Observed Behavior

They are not.

_This is a follow-up from #2612, #873._

PR #2753, opened 2022-11-05
Bulk Create/Edit of Required Relationships

briddo:
# Closes: #2744 
# What's Changed
UI bulk edit and API bulk edit/add now enforces required relationships.

# TODO
- [x] UI bulk edit check for existence of required objects
- [x] UI bulk edit check each object's required relationships data from form submission
- [ ] UI bulk edit tests
- [x] API bulk edit
- [ ] API bulk edit tests
- [x] API bulk create
- [ ] API bulk create tests
- [ ] Docs

### DIFF
--- a/nautobot/extras/api/relationships.py
+++ b/nautobot/extras/api/relationships.py
@@ -1,7 +1,7 @@
 import logging
 
 from django.contrib.contenttypes.models import ContentType
-from django.core.exceptions import ValidationError as CoreValidationError
+from django.core.exceptions import ValidationError as DjangoValidationError
 from django.db.models import Q
 from drf_spectacular.utils import extend_schema_field
 from rest_framework.exceptions import PermissionDenied
--- a/nautobot/extras/api/relationships.py
+++ b/nautobot/extras/api/relationships.py
@@ -1,6 +1,7 @@
 import logging
 
 from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import ValidationError as CoreValidationError
 from django.db.models import Q
 from drf_spectacular.utils import extend_schema_field
 from rest_framework.exceptions import PermissionDenied
@@ -271,10 +272,12 @@ class RelationshipModelSerializerMixin(ValidatedModelSerializer):
         )
         if required_relationships_errors:
             raise ValidationError(required_relationships_errors)
-
         instance = super().create(validated_data)
         if relationships_data:
-            self._save_relationships(instance, relationships_data)
+            try:
+                self._save_relationships(instance, relationships_data)
+            except CoreValidationError as error:
+                raise ValidationError(str(error))
         return instance
 
     def update(self, instance, validated_data):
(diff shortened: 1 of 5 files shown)


bryanculver:
Ready for implementation review.

nautobot/extras/models/relationships.py
```diff
@@ -303,14 +301,19 @@ def required_related_objects_errors(
...
+                missing_data = False
 
                 if output_for == "ui":
                     supplied_data = initial_data.get(field_key, [])
+                    if not supplied_data:
+                        missing_data = True
+
                 elif output_for == "api":
                     supplied_data = initial_data.get(relation, {}).get(opposite_side, {})
+                    if len(supplied_data) == 0:
```
jathanism, line 313: Any reason not to align the patterns used her to check there's any `supplied_data`?

```suggestion
                    if  not supplied_data:
```

nautobot/extras/models/relationships.py
```diff
@@ -257,10 +257,8 @@ def required_related_objects_errors(
                     and initial_data.get(relation, {}).get(opposite_side, {}) == {}
                     and not relationships_key_specified
                 ):
-                    if (
-                        RelationshipAssociation.objects.filter(**{f"{relation.required_on}_id": instance.pk}).count()
-                        > 0
-                    ):
+                    filter_kwargs = {"relationship": relation, f"{relation.required_on}_id": instance.pk}
+                    if RelationshipAssociation.objects.filter(**filter_kwargs).count() > 0:
```
jathanism, line 261: This should be an `exists()` query instead:

```suggestion
                    if RelationshipAssociation.objects.filter(**filter_kwargs).exists():
```

nautobot/extras/api/relationships.py
```diff
@@ -1,6 +1,7 @@
 import logging
 
 from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import ValidationError as CoreValidationError
```
jathanism, line 4:
```suggestion
from django.core.exceptions import ValidationError as DjangoValidationError
```

nautobot/extras/api/relationships.py
```diff
@@ -271,10 +272,12 @@ def create(self, validated_data):
         )
         if required_relationships_errors:
             raise ValidationError(required_relationships_errors)
-
         instance = super().create(validated_data)
         if relationships_data:
-            self._save_relationships(instance, relationships_data)
+            try:
+                self._save_relationships(instance, relationships_data)
+            except CoreValidationError as error:
```
jathanism, line 279:
```suggestion
            except DjangoValidationError as error:
```

jathanism (changes requested):

nautobot/extras/models/relationships.py
briddo: Ah! I think I was told off by a linter or even flake8 before because I wasn't being explicit enough, but maybe it wasn't setup correctly. So every time I had to check an int was 0, I just used `len(i) == 0` instead of just `not i`. Good to know  I can use the implicit check again now!

Issue #2732, opened 2022-11-03
Dynamic filter form field value persistence is surprising.


jathanism:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.x
* Python version: *
* Database platform, version: n/a
* Middleware(s): n/a

### Steps to Reproduce

> This was done using the demo dataset.

0. Go to device list view
1. Click "Filter", then "Advanced" and input an advanced filter such as "name regex = ^ams01".
2. Click "Apply"
3. Observe that the filter `name__re=^ams01` is displayed in list view
4. Click "Filter" again, then select any "Default" filter such as "has interfaces = Yes"
5. Click "Apply"
6. Observe that the previous filter was replaced and that only `has_interfaces=True` is displayed in list view.

### Expected Behavior

The filter values between both tabs should persist.

### Observed Behavior

This appears to only be the case if you mix "advanced' and "default" filters using the "default" tab. If you select the same set of filter criteria using only the "advanced" tab, they always persist.

gsnider2195:
Potential fix in django if a JS fix isn't feasible:

```diff
diff --git a/nautobot/extras/forms/base.py b/nautobot/extras/forms/base.py
index d4e8febe3..60da60697 100644
--- a/nautobot/extras/forms/base.py
+++ b/nautobot/extras/forms/base.py
@@ -1,3 +1,4 @@
+from django import forms
 from nautobot.utilities.forms import BootstrapMixin
 from .mixins import (
     CustomFieldModelBulkEditFormMixin,
@@ -38,6 +39,20 @@ class NautobotFilterForm(BootstrapMixin, CustomFieldModelFilterFormMixin, Relati
     needed.
     """
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        for name, value in self.data.items():
+            if name == "q" or name not in self.fields:
+                if isinstance(value, (list, tuple)):
+                    widget = forms.widgets.MultipleHiddenInput
+                else:
+                    widget = forms.widgets.HiddenInput
+                self.fields[name] = forms.CharField(
+                    required=False,
+                    initial=value,
+                    widget=widget,
+                )
+
 
 class NautobotBulkEditForm(
     BootstrapMixin, CustomFieldModelBulkEditFormMixin, RelationshipModelBulkEditFormMixin, NoteModelBulkEditFormMixin
```

gsnider2195:
[...]

PR #2757, opened 2022-11-07
Fix filter fields on default filter replaces filters on dynamic filter on submit

timizuoebideri1:
# Closes: #2732 
# What's Changed

Merge advanced and default form values and delete duplicates before applying the filter.

NOTE:
If a select field on either the Default filter form or the Dynamic filter form is changed, repeat that change in the other if the fields have the same 'name' attribute. This would avoid the following event from occurring:

1. Apply the filter after changing the default filter status fields to `active` and `planned.`
4. Check the advanced filter to see if the applicable status is also applied there.
5. Remove `planned` from the advance status filter field, then apply the filter.
6. Despite removing `planned` from the advanced filter, both the default and advance status fields have both `active` and `planned`.

# TODO

- [ ] Unit, Integration Tests


nautobot/core/templates/generic/object_list.html
```diff
@@ -220,5 +226,41 @@ <h1>{% block title %}{{ title }}{% endblock %}</h1>
...
+        let default_filters_field_dom = $(`#default-filter form select[name=${field_name}]`);
+        let advanced_filters_field_dom = $(`#advanced-filter #filterform-table tbody tr td select[name=${field_name}]`);
+
+        // Only apply logic if fields with same name attr are on both advanced and default filter form
+        if(default_filters_field_dom.length && advanced_filters_field_dom.length){
+            let default_filters_field_ids = default_filters_field_dom.select2('data').map(data => data["id"]);
+            let advanced_filters_field_ids = advanced_filters_field_dom.select2('data').map(data => data["id"]);
+
+            // Only change field value if both fields do not have equal values
+            if (JSON.stringify(advanced_filters_field_ids) !== JSON.stringify(default_filters_field_ids)){
```
HanlinMiao, line 254: nice!

gsnider2195 (approved):

Issue #2755, opened 2022-11-07
Dockerfile permission denied config file


Yannis100:
### Environment
* Nautobot version (Docker tag too if applicable): 1.4.8 -> 1.4.9
* Python version: 3.10.6
* Database platform, version: postgres 14 alpine
* Middleware(s):

Change in #2715 #2671 
### Steps to Reproduce
1. Own Dockerfile based on official Nautobot docker image
2. Try to overwrite config file with `cat /opt/nautobot/nautobot_config.append.py >> /opt/nautobot/nautobot_config.py`
3. `cannot create /opt/nautobot/nautobot_config.py: Permission denied`

### Expected Behavior
Able to change the configuration file while building the image

### Observed Behavior
Permission denied error

I think changing the line https://github.com/nautobot/nautobot/blob/9f7b414408cef9931d675abfbc5d394e65a6e420/docker/Dockerfile#L314 :
`COPY --from=final-dev /opt/nautobot/nautobot_config.py /opt/nautobot/nautobot_config.py` to 
`COPY --chown=nautobot --from=final-dev /opt/nautobot/nautobot_config.py /opt/nautobot/nautobot_config.py`
should solve the issue

bryanculver:
Thanks for the report @Yannis100!

Can you provide a sanitized example of your Dockerfile?

Going to use it as an example for some future housekeeping tests.

glennmatthews:
Confirmed:

```
nautobot@692f1c8a9a0d:~$ ls -l
total 32
-rw-r--r-- 1 nautobot nautobot  1980 Nov  4 21:36 nautobot.crt
-rw------- 1 nautobot nautobot  3276 Nov  4 21:36 nautobot.key
-rw-r--r-- 1 root     root     18818 Nov  4 21:36 nautobot_config.py
-rw-r--r-- 1 nautobot nautobot  2198 Nov  4 21:27 uwsgi.ini
```

Yannis100:
Thanks for the quick fix !

@bryanculver 
My Dockerfile is basically a mix between the LDAP and Plugins ones in docker-compose project https://github.com/nautobot/nautobot-docker-compose/blob/main/Dockerfile-LDAP & https://github.com/nautobot/nautobot-docker-compose/blob/main/plugin_example/Dockerfile-Plugins

I just didn't want to copy a full config file as it might change between versions, so I use
````
COPY --chown=nautobot config/nautobot_config.append.py /opt/nautobot/
[...]

PR #2758, opened 2022-11-07
Fix incorrect permissions on /opt/nautobot/nautobot_config.py in final images

glennmatthews:
# Closes: #2755 
# What's Changed

- Fix ownership of `/opt/nautobot/nautobot_config.py` in `final` images, a bug introduced by #2715.

Before:

```bash
nautobot@692f1c8a9a0d:~$ ls -l
total 32
-rw-r--r-- 1 nautobot nautobot  1980 Nov  4 21:36 nautobot.crt
-rw------- 1 nautobot nautobot  3276 Nov  4 21:36 nautobot.key
-rw-r--r-- 1 root     root     18818 Nov  4 21:36 nautobot_config.py
-rw-r--r-- 1 nautobot nautobot  2198 Nov  4 21:27 uwsgi.ini
```

After:

```bash
nautobot@89d310d03b6c:~$ ls -l
total 32
-rw-r--r-- 1 nautobot nautobot  1980 Nov  7 15:03 nautobot.crt
-rw------- 1 nautobot nautobot  3272 Nov  7 15:03 nautobot.key
-rw-r--r-- 1 nautobot nautobot 18818 Nov  7 15:03 nautobot_config.py
-rw-r--r-- 1 nautobot nautobot  2198 Oct  4 13:08 uwsgi.ini
```

- Fix a less severe issue also introduced by #2715 in which plugins running with the default `nautobot_config.py` using the `final-dev` image would have verbose console logging enabled by default during the tests, which is generally not desirable.

# TODO

- n/a Unit, Integration Tests

### DIFF
--- a/nautobot/core/settings.py
+++ b/nautobot/core/settings.py
@@ -1,6 +1,7 @@
 import os
 import platform
 import re
+import sys
 
 from django.contrib.messages import constants as messages
 import django.forms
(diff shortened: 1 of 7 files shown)


nautobot/core/settings.py
```diff
@@ -283,40 +284,49 @@
 DEBUG = is_truthy(os.getenv("NAUTOBOT_DEBUG", "False"))
 INTERNAL_IPS = ("127.0.0.1", "::1")
 FORCE_SCRIPT_NAME = None
+
+TESTING = len(sys.argv) > 1 and sys.argv[1] == "test"
```
HanlinMiao, line 288: Is the second `sys.argv` always guaranteed be `test`? Can the user chain commands with `nautobot-server...`?
perhaps?
```suggestion
TESTING = len(sys.argv) > 1 and "test" in sys.argv
```

tasks.py
```diff
@@ -652,7 +652,8 @@ def unittest(
         build_and_check_docs(context)
 
     append_arg = " --append" if append else ""
-    command = f"coverage run{append_arg} --module nautobot.core.cli --config=nautobot/core/tests/nautobot_config.py test {label}"
+    command = f"coverage run{append_arg} --module nautobot.core.cli test {label}"
+    command += " --config=nautobot/core/tests/nautobot_config.py"
```
HanlinMiao, line 656: Is this change strictly for code formatting?

nautobot/core/settings.py
glennmatthews: It's not guaranteed, for example `nautobot-server --config=my_nautobot_config.py test` would be a valid command that wouldn't trip this check. But this is a pretty standard approach and matches most extant plugins.

I considered the `test in sys.argv` approach but would rather err in the direction of false negatives (not TESTING when we actually are running a test) rather than false positives (setting TESTING when we're actually in a live environment).

tasks.py
glennmatthews: Also re-ordering the parameters so that the above `TESTING` definition would correctly detect this as a test.

gsnider2195 (approved):

HanlinMiao (approved):

PR #2759, opened 2022-11-07
Add prometheus metrics for health check results.

Kircheneer:
# Closes: DNE
# What's Changed
Exposes metrics for the health check results under `/metrics` as follows:

```
# HELP health_check_database_info Multiprocess metric
# TYPE health_check_database_info gauge
health_check_database_info 1.0
# HELP health_check_cache_ops_redis_info Multiprocess metric
# TYPE health_check_cache_ops_redis_info gauge
health_check_cache_ops_redis_info 0.0
# HELP health_check_redis_backend_info Multiprocess metric
# TYPE health_check_redis_backend_info gauge
health_check_redis_backend_info 1.0
```

I have defaulted to enabling the metrics endpoint in the development configuration. I am happy to change this back if you want.

I have explicitly added
```
# Custom prometheus metrics
prometheus-client = "~0.14.1"
```
to `pyproject.toml`. This is a transitive dependency through `django-prometheus`, so it is not technically a new dependency, 

# TODO
- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/health_checks.py
+++ b/nautobot/extras/health_checks.py
@@ -1,5 +1,5 @@
 """Nautobot custom health checks."""
-from abc import ABC
+from typing import Optional
 from urllib.parse import urlparse
 
 from django.conf import settings
@@ -15,7 +15,7 @@ from redis.sentinel import Sentinel
 from .models import HealthCheckTestModel
 
 
-class NautobotHealthCheckBackend(BaseHealthCheckBackend, ABC):
+class NautobotHealthCheckBackend(BaseHealthCheckBackend):
     """Automatically set metric based on the outcome of the check.
 
     Due to the multiprocess nature of Nautobot we have multiple processes generating health check metrics. Most of these
@@ -27,15 +27,18 @@ class NautobotHealthCheckBackend(BaseHealthCheckBackend, ABC):
     MULTIPROCESS_MODE = "max"
 
     states = {"unknown": -1, "down": 0, "up": 1}
-    metric: Gauge  # Set this in subclasses
+    metric: Optional[Gauge] = None  # Set this in subclasses
 
     def __init__(self):
         super().__init__()
-        # Initialize the metric as -1
-        self.metric.set(self.states["unknown"])
+        # Initialize the metric as -1 if present
+        if self.metric:
+            self.metric.set(self.states["unknown"])
 
     def run_check(self):
         super().run_check()
+        if not self.metric:
+            return
         if self.errors:
             self.metric.set(self.states["down"])
         else:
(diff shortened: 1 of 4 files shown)


Kircheneer:
For some reason the cache ops health check always seems to fail, but that health check is not included in the (default) `/health` endpoint anyway, possibly its broken?

bryanculver:
> For some reason the cache ops health check always seems to fail, but that health check is not included in the (default) `/health` endpoint anyway, possibly its broken?

Well we are ripping out Cacheops in 2.0 and it's already been disabled by default for new installs in 1.5.

What is the intended consumer of the `/metrics` endpoint?

Kircheneer:
> What is the intended consumer of the /metrics endpoint?

Prometheus/Telegraf for telemetry purposes.

nautobot/extras/health_checks.py
```diff
@@ -1,21 +1,56 @@
...
+from prometheus_client import Gauge
 from redis import exceptions, from_url
 from redis.client import Redis
 from redis.sentinel import Sentinel
 
 from .models import HealthCheckTestModel
 
 
-class DatabaseBackend(BaseHealthCheckBackend):
+class NautobotHealthCheckBackend(BaseHealthCheckBackend, ABC):
```
bryanculver, line 18: We don't seem to define any abstract methods. Is there a reason for inheriting from `ABC` anyway?

nautobot/extras/health_checks.py
```diff
@@ -67,6 +102,12 @@ def check_redis(self, redis_url=None, **kwargs):
 class CacheopsRedisBackend(RedisHealthCheck):
     """Health check for Cacheops Redis."""
 
+    metric = Gauge(
+        "health_check_cache_ops_redis_info",
+        "State of cache ops redis backend",
+        multiprocess_mode=NautobotHealthCheckBackend.MULTIPROCESS_MODE,
+    )
+
```
bryanculver, line 110: Per your comment this doesn't seem to be working and CacheOps is going away soon. YAGNI.

bryanculver (changes requested):

nautobot/extras/health_checks.py
Kircheneer: My intent was to mark this ABC as there is the `metric` class attribute you need to set in subclasses. I can either make it an `@abstractproperty` or just omit the entire ABC deal - what are your thoughts?

Kircheneer: Scratch that, I made the metric optional to deal with cacheops and that resolves this.

bryanculver (approved):

Issue #1519, opened 2022-03-18
Copy button on hover for more fields


Thetacz:
### As ...

Ozzie - Operator

### I want ...

to be able to copy data with single click when viewing device detail

### So that ...

- I always copy the whole part and nothing extra
- the whole copy process is easier in general

### I know this is done when...

- I am able to click copy button when hovering over device attribute

Our original internal request is only for device type and asset number, but it does not make sense to limit to only these 2 attributes. It would be nice for this to be a general rule - whichever attribute I hover over, I can get the copy button there.

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

None

### External Dependencies

None

Thetacz:
This is probably as easy as adding `<span class="hover_copy">` over each attribute, but probably would be nice to do this somehow automatically.

christhant:
Fields that need to display copy button when hover.
1. https://demo.nautobot.com/ipam/prefixes/2e8dd8c7-4ec2-4415-b64d-ce34cf5b2b09/?tab=main
- IPAM=> pre fix => ip address title needs to display copy button
- IP Address columns
2. https://demo.nautobot.com/ipam/vlans/52a457f4-e5d7-4efb-845d-abdde337e560/?tab=main
- IPAM=> vlan => prefix column needs to display copy button
3. https://demo.nautobot.com/dcim/devices/e71d056c-1b24-40f7-b4ec-276dfa3ebbad/?tab=main
asset tag needs to display copy button
4. https://demo.nautobot.com/circuits/circuits/8f361c7c-b477-48cc-9624-38e217f7e967?tab=main
circuit Id title need to display copy button
circuit table => Circuit ID need to display copy button

PR #2762, opened 2022-11-07
Extend IP Address Copy Link

christhant:
# Closes: #1519 
# What's Changed
1. Extended IP address link for Interface with copy button when hover
image

2.  Extended IP address link for IPAM prefix with copy button when hover
image

image

image

3. Extended IP address prefix link for vlan with copy button when hover
image

4. Extended Circuit ID link when copy button when hover
image

Explanation of Change(s)
---------------------------
core/templates/generic/object_list.html
- include ClipboardJS to able to use copying data to clipboard from the data list views.
ipam/tables.py
- extending the existing template codes (PREFIX_LINK, IPADDRESS_LINK, IPADDRESS_ASSIGN_LINK) with (PREFIX_COPY_LINK, IPADDRESS_COPY_LINK, IPADDRESS_ASSIGN_COPY_LINK) to allow displaying copy icon when mouse hover
-project-static/css/base.css
instead of create a new class, I used html object based css definition
when copy icon is placed between H1 tag, its size becomes larger in accordance to H1. therefore, when copy icon span class="btn-inline" class is placed inside H1, this change will bring down copy icon size to medium.

# TODO

- [ ] Unit, Integration Tests


changes/1519.fixed
```diff
@@ -0,0 +1 @@
+Extending the model table columns thant need to display copy button when hover
```
HanlinMiao, line 1:
```suggestion
Extending the model table columns that need to display copy button when hovered over.
```

HanlinMiao (changes requested):
I pulled down the branch and played around with the copy button. It does not seem to be working?

nautobot/ipam/tables.py
```diff
@@ -62,10 +62,30 @@
 {% endif %}
 """
 
+IPADDRESS_COPY_LINK = """
```
jathanism, line 65: Since you're in here, this seems like a good opportunity to move away from `template_code` and instead use `template_name` so that these template snippets can be stored as template files instead of hard-coded template strings.

Perhaps it's out of scope for this, but inevitably is something we need to address for the UI overhaul.

changes/1519.fixed
christhant: @HanlinMiao lol. thank you! I fixed it now.

nautobot/ipam/tables.py
christhant: Thanks @jathanism. @bryanculver  can you please let me know how I can create template_name?

bryanculver (approved):

Issue #2789, opened 2022-11-09
CSS footer margin is gone and swagger UI authorization box is broken


gsnider2195:
### Environment
* Nautobot version (Docker tag too if applicable): v1.5.1b1
* Python version: 3.10.7
* Database platform, version: Postgres 13
* Middleware(s):

### Steps to Reproduce
Go to test.nautobot.com, scroll to the footer and then navigate to the API 

![image](https://user-images.githubusercontent.com/75227981/200877779-41df0791-10c6-4e55-81fd-6be7a6f3f455.png)
![image](https://user-images.githubusercontent.com/75227981/200877839-3171dd45-7ebe-4098-9ceb-3e7acca0cc0b.png)
![image](https://user-images.githubusercontent.com/75227981/200877916-bbac42cb-c1bf-4a6e-8e0c-25b447c76ab2.png)

It looks like this was a regression from #2553. #2747 resolves the footer issue without impacting the margin or the swagger API. Reverting #2553 appears to resolve this issue.

glennmatthews:
In general if we're making changes to the header and footer, we need to check for impact to:

1) Ordinary Nautobot UI pages
2) Swagger UI
3) GraphiQL

gsnider2195:
> In general if we're making changes to the header and footer, we need to check for impact to:
> 
> 1. Ordinary Nautobot UI pages
> 2. Swagger UI
> 3. GraphiQL

Agreed. After reverting I made sure this looks ok in large and small viewports on the home page, list view with a lot of objects, swagger ui and graphql. We should also check light mode and dark mode.

PR #2747, opened 2022-11-04
fix scrollbar being covered by nav bar

gsnider2195:
# Closes: #2266
# What's Changed

- Fix the scrollbar being covered by the navbar
- Fix some problems with the dark mode theme when applying the scrollbar fix

- Fix the footer detaching from the bottom of the page when the viewport width gets too small:
before:
image
after:
image

# TODO

- [x] Explanation of Change(s)
- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [x] Attached Screenshots, Payload Example
- n/a Unit, Integration Tests
- n/a Documentation Updates (when adding/changing features)
- n/a Example Plugin Updates (when adding/changing features)
- [x] Outline Remaining Work, Constraints from Design

Issue #2552, opened 2022-10-02
Footer On Resized Window Not Displaying Correctly


whitej6:
### Environment
* Nautobot version (Docker tag too if applicable): 1.4.X (includes demo)
* Python version: py3.8
* Database platform, version: Postgres
* Middleware(s):

### Steps to Reproduce
1. Resize window so links on the right begin to touch the the timestamp
2. Scroll to bottom of screen
3.

### Expected Behavior
Footer expands size with right hand elements begin to stack and new size inherits footer CSS 

### Observed Behavior
Footer does expand in size, right hand elements being to stack BUT overflow into parent HTML element.

<img width="1269" alt="image" src="https://user-images.githubusercontent.com/26559188/193465455-e404a985-938c-4af4-92fb-b872bb83202d.png">

PR #2790, opened 2022-11-09
Fix web UI footer margin and swagger UI authorization box size

gsnider2195:
# Closes: #2789 
# What's Changed

- Fix swagger UI authorization box vertical size being too large
- Fix footer margin that was no longer working
- Fix tiny white line underneath the footer on dark mode when the viewport is wide

Tested against home page, list view with over 50 objects, graphql ui and swagger api on wide and narrow viewports in light mode and dark mode.

image

image

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/project-static/css/dark.css
+++ b/nautobot/project-static/css/dark.css
@@ -36,6 +36,7 @@ html[data-theme="dark"] .container-fluid.wrapper {
 
 /* Lighter than the background */
 html[data-theme="dark"] .navbar,                          /* Top menu navbar */
+html[data-theme="dark"] .footer,                          /* Bottom footer bar */
 html[data-theme="dark"] .footer .row {                    /* Bottom footer bar */
   background-color: rgb(209, 216, 221) !important;
 }
--- a/changes/2789.fixed
+++ b/changes/2789.fixed
@@ -1,0 +1,1 @@
+Fixed web UI footer margin and swagger UI authorization box size.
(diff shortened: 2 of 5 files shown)


nautobot/project-static/css/dark.css
```diff
@@ -36,6 +36,7 @@ html[data-theme="dark"] .container-fluid.wrapper {
 
 /* Lighter than the background */
 html[data-theme="dark"] .navbar,                          /* Top menu navbar */
+html[data-theme="dark"] .footer,                          /* Bottom footer bar */
 html[data-theme="dark"] .footer .row {                    /* Bottom footer bar */
```
glennmatthews, line 40: Is it correct to have both of these?

glennmatthews:
Have you confirmed that #2552 is still resolved?

nautobot/project-static/css/dark.css
gsnider2195: If we don't set the footer row explicitly it will inherit the background color property from this section in `base.css` instead of from the parent element:
```css
.footer .row {
    background-color: #f5f5f5;
}
```

The footer row overlaps the entire footer in every case except on the graphql ui when the viewport is wide enough so that the footer doesn't overflow. In that case if the footer color isn't set you'd see a 1 or 2 pixel white line at the very bottom of the page. I'm not sure why the graphql ui is different.

gsnider2195: Actually the base.css change was something I changed in #2747. Let me see if I can condense this.

gsnider2195: The base.css change was the fix for #2552. So both are needed and I just updated the light mode theme to fix the tiny white line in the graphql that was so subtle I didn't notice it until I zoomed it to 800%

glennmatthews (approved):

Issue #204, opened 2021-03-26
Refactor all imports to use module namespaces


jathanism:
### Proposed Changes

Currently in the code base there are many examples of large imports into the global namespace. For example, here is a sample of forms-related imports from `nautobot/extras/forms.py`:

```python
from django import forms

from nautobot.utilities.forms import (
    add_blank_choice,
    APISelectMultiple,
    BootstrapMixin,
    BulkEditForm,
    BulkEditNullBooleanSelect,
    ColorSelect,
    CSVModelChoiceField,
    CSVModelForm,
    CSVMultipleContentTypeField,
    DateTimePicker,
    DynamicModelChoiceField,
    DynamicModelMultipleChoiceField,
    JSONField,
    MultipleContentTypeField,
    SlugField,
    StaticSelect2,
    StaticSelect2Multiple,
    BOOLEAN_WITH_BLANK_CHOICES,
)
```

I'm proposing we move to this pattern:

```
 from django import forms as djforms  # Django forms module
 from nautobot.utilities import forms  # Our utilities.forms module
    
 # In code..
class SomeForm(forms.CustomFieldModelForm, djforms.ModelForm):
    name = djforms.CharField(required=False)
    slug = forms.CSVContentTypeField()

<!-- Provide justification for the proposed change(s). -->
### Justification

The current pattern is not only is unwieldy, but complicates merges, rebases, and refactors and in some cases can result in circular imports.

This way it's clear what is coming from Django `forms` and what is coming from internal `forms` modules, and any time something new is needed or used from `nautobot.utilities.forms` it doesn't need to be explicitly imported.

This also applies to more than just forms, but I was just selecting forms as a singular example of imports for a single module.

gsnider2195:
[refactor_imports.zip](https://github.com/nautobot/nautobot/files/10061540/refactor_imports.zip)

itdependsnetworks:
I have added this to an internal tracker of best practices. In reality, if we can get it into pylint-nautobot, then we do it, otherwise just a soft recommendation that we shouldn't track as a one off.

PR #2794, opened 2022-11-09
update module imports and add module import section to best practices

gsnider2195:
# Closes: #204
# What's Changed

This PR is only updating a single module and adding documentation for a style guide for module imports in Nautobot. See the comments on the code changes below for conversation on the thought process for this standard. Also includes some typo fixes in docs.

localhost_8001_projects_core_en_stable_development_style-guide_ (3)

# TODO

- [x] Explanation of Change(s) - and add comments to file diffs


nautobot/utilities/filters.py
```diff
@@ -1,35 +1,24 @@
-from collections import OrderedDict
-from copy import deepcopy
 import logging
 import uuid
+from collections import OrderedDict
+from copy import deepcopy
 
-from django import forms
+import django_filters
```
bryanculver, line 6: Is there a reason this was moved up above its other imports below?

nautobot/docs/development/best-practices.md
```diff
@@ -400,3 +400,102 @@ filterset.qs.filter(query).count()  # 339
...
 +++ 1.4.0
     Using `NautobotUIViewSet` for [plugin development](../plugins/development.md#nautobotuiviewset) is strongly recommended.
+
+## Style Guidelines for Importing Python Packages
+
+To prevent circular dependency errors and improve code readability, the following standards should be followed when importing from other python modules.
+
+### PEP8 Style Guide
+
+Nautobot follows the [PEP8 style guide's](https://peps.python.org/pep-0008/#imports) standard for importing modules. Libraries should be imported in these groups: standard library, third party libraries, then `nautobot` packages and finally try/except imports. The groups should be separated by a single blank line. Within these groups, top level packages should be imported first, then sub-package imports. Import lines should be sorted alphanumerically as well as lists of names imported from packages.
```
glennmatthews, line 410: Can we recommend `isort` as the tool for checking/enforcing this style? Obviously we can't enable it project-wide yet or as a CI step, but it's at least useful for local development and incremental improvement. Would be nice to even provide an appropriate config in `pyproject.toml`, something like:

```toml
[tool.isort]
profile = "black"
line_length = 120
force_sort_within_sections = true  # personal preference, see other comments
```

nautobot/utilities/filters.py
```diff
@@ -1,35 +1,24 @@
...
-)
-from nautobot.utilities.forms.fields import MultiMatchModelMultipleChoiceField, MultiValueCharField
-from nautobot.utilities.utils import flatten_iterable
-
-
 from taggit.managers import TaggableManager
+from tree_queries.models import TreeNode
 
+from nautobot.dcim import fields as dcim_fields
+from nautobot.dcim import forms as dcim_forms
```
glennmatthews, line 19: Should these two be merged?

```suggestion
from nautobot.dcim import fields as dcim_fields, forms as dcim_forms
```

HanlinMiao (commented):
Are we changing the entire `utilities` folder or just `utilities/forms`?

nautobot/utilities/filters.py
gsnider2195: I think that when aliasing module names it's easier to read if they're on separate lines. Black, isort and flake8 don't complain about it (with the isort config above)

nautobot/docs/development/best-practices.md
```diff
@@ -400,3 +400,102 @@ filterset.qs.filter(query).count()  # 339
...
+    # inter-app import
+    from nautobot.dcim import models as dcim_models
+
+    # intra-app import
+    from nautobot.extras import constants
+    ```
+
+#### Resolving Name Conflicts
+
+When using external libraries you may need to import multiple different modules with the same name. In this case, the namespace from the external package should be aliased. A common source of namespace conflicts in Nautobot are django packages. For django, packages should be renamed `dj<module>`. For other external libraries, use `<package>_<module>`.
```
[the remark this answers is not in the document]
gsnider2195: I don't know if avoiding aliasing completely is the right approach. Typing out the full module path is going to make the code more difficult to read and aliasing is a good middle ground of making the object name shorter while still giving us enough information about where the object came from. (`dcim_fields.JSONPathField` tells me which app and usually which file I can find that definition in without scrolling back and forth to the top of the file)

I'm not attached to making a special case for django imports. That was @jathanism's idea 😀

nautobot/docs/development/best-practices.md
gsnider2195: `isort` is already a dev dependency for a lot of our dependencies and it's a direct dependency for `pylint` (so it's already in the poetry.lock file). I think making it part of the style guide and adding the dev dependency + config is a good idea. `force_sort_within_sections` isn't my favorite but I don't have a strong opinion on it. Speaking of the style guide, I'm thinking all of these categories should go there and not in best practices.

nautobot/docs/development/best-practices.md
```diff
@@ -400,3 +400,102 @@ filterset.qs.filter(query).count()  # 339
...
+    import django_filters
+    from django.db.models import CharField, DecimalField, TextField
+
+    from nautobot.dcim import models as dcim_models
+    from nautobot.extras import models
+    ```
+
+### Absolute Imports
+
+Always use absolute imports instead of relative imports.
```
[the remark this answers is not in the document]
gsnider2195: Is there any limitation on absolute imports where we would have to use relative imports in a `__init__.py`? The only reason I didn't call out an exception for `__init__.py` files was just for consistency.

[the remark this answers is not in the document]
glennmatthews: I'd thought there was but it looks like not. So you can disregard this. :)

nautobot/utilities/filters.py
glennmatthews: Fair enough. :-)

glennmatthews (approved):
Let's reach a decision as a team about `djforms` vs `django_forms` and the `force_sort_within_sections` toggle. Otherwise looks great to me.

gsnider2195:
As discussed we will enable `force_sort_within_sections` in `isort`

nautobot/docs/development/best-practices.md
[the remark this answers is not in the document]
jathanism: Make it `django_{foo}`.

Issue #2257, opened 2022-08-19
Codify Standards


itdependsnetworks:
### As ...

P.D. - Plugin Developer

### I want ...

the ability to be inline with the current standards and best practices, without requiring a substantial time to be able to learning that system.

### So that ...

I know that plugin is as rock solid as it can be

### I know this is done when...

- I have an easy method of enforcement of the best practices
    - In an ideal (but unrealistic) world I would have a "formatter" similar to black
    - In a realistic world, I have the ability to quickly create tests that let me know which standards I have met or not

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

N/A

### External Dependencies

- Potentially linters

itdependsnetworks:
As always, I will separate a potential solution from the definition of done. I think it makes sense to use a custom Pylint checker.

itdependsnetworks:
Examples include:

* Ensuring that navigation permissions are in place
* Filters
    * ensure using best practice base class
    * Ensure that you are using the right "type", e.g.  NameSlugSearchFilterSet vs ???
    * I think you are not supposed to set fields when the attribute exists already?
    * Ensure classes are using proper name
    * Ensure there is a q filter
* Forms
    * Ensure csvform uses the csv_headers attribute on the model
    * Ensure classes are using proper name, e.g. <model_name>Form / <model_name>FilterForm
    * Ensure there is a q filter
* Models
    * Ensure classes are using proper name
    * Ensure we don't override defaults on model fields
    * Ensure to_csv and csv_headers are set
    * Ensure extra features are set and/or don't override each other (i think there is something with status here? )
* Ensure using views

bryanculver:
Assigning myself to Move these over to pylint-nautobot

PR #2801, opened 2022-11-10
[WIP] Best Practice Linter

Kircheneer:
# Closes: #2257
# What's Changed

*Note: This was originally not using pylint, I have force pushed over the old, manual approach.*

A couple of custom pylint rules have been added:
- Check if classes that inherit from certain base classes (e.g. `django_filters.BaseFilterSet`) also inherit from the respective Nautobot version (i.e. `nautobot.utilities.filters.BaseFilterSet`)
- Check that `instance.pk` is not used as an if condition

Current example output:
```
************* Module nautobot.tenancy.forms
nautobot/tenancy/forms.py:115:0: W5001: Uses correct base classes. (incorrect-base-class)
nautobot/tenancy/forms.py:129:0: W5001: Uses correct base classes. (incorrect-base-class)
************* Module nautobot.tenancy.filters
nautobot/tenancy/filters/__init__.py:78:0: W5001: Uses correct base classes. (incorrect-base-class)
```

# TODO
- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/utilities/pylint.py
+++ b/nautobot/utilities/pylint.py
@@ -1,0 +1,97 @@
+import inspect
+
+from astroid import ClassDef, Assign, Const, Attribute
+from django.db.models import Model
+from django.forms import Form
+from django_filters import FilterSet
+from pylint.checkers import BaseChecker
+
+from nautobot.core.models import BaseModel
+from nautobot.utilities.filters import BaseFilterSet
+from nautobot.utilities.forms import BootstrapMixin
+
+
+def to_path(obj):
+    """Given an object, return its fully qualified import path."""
+    return f"{inspect.getmodule(obj).__name__}.{obj.__name__}"
+
+
+def is_abstract(node):
+    """Given a node, returns whether it is an abstract base model."""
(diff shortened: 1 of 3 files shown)


itdependsnetworks:
flipping epic man!!

itdependsnetworks:
I think would be great to partner with you on converting this to a pylint package. I think that once we setup a pattern, it shouldn't be too bad to build out. This would make is simple for plugins to implement as well.

Kircheneer:
@itdependsnetworks one thing I'm wondering is whether I can use inspection like this in pylint, from what I can tell they lint based on the AST/raw source code rather then the already interpreted objects: https://pylint.pycqa.org/en/latest/development_guide/how_tos/custom_checkers.html

We need to have interpreted objects as for example we have to inspect the whole class hierarchy for these checks to make sense.

Do you have any experience there?

Kircheneer:
New draft utilizing pylint this time. Check [pylint CI output](https://github.com/nautobot/nautobot/actions/runs/4606499834/jobs/8139982799?pr=2801) for details on what this currently reports.

PR #2816, opened 2022-11-12
Bug changing interface mode

timizuoebideri1:
# What's Changed

Remove `tagged_vlan` validation from the model `clean` method to `Form` and `Serializer` because we can't know if the `tagged_valn` has been cleared or not. As a result, regardless of whether the `tagged_valn` has been cleared or not via form or API, it raises an exception.

Add validation to Form and serializer

# TODO

- [ ] Unit, Integration Tests


nautobot/dcim/forms.py
```diff
@@ -183,17 +183,21 @@ def clean(self):
 
         parent_field = "device" if "device" in self.cleaned_data else "virtual_machine"
         tagged_vlans = self.cleaned_data["tagged_vlans"]
+        mode = self.cleaned_data["mode"]
+
+        if mode != InterfaceModeChoices.MODE_TAGGED and tagged_vlans:
+            raise forms.ValidationError({"tagged_vlans": f"Clear tagged_vlans to set mode to {self.mode}"})
```
timizuoebideri1, line 189: Since `tagged_vlans` is auto cleared via Javascript whenever the mode is changed in UI, this is essentially redundant.

nautobot/dcim/forms.py
```diff
@@ -183,17 +183,21 @@ def clean(self):
...
 
         # Untagged interfaces cannot be assigned tagged VLANs
-        if self.cleaned_data["mode"] == InterfaceModeChoices.MODE_ACCESS and tagged_vlans:
+        if mode == InterfaceModeChoices.MODE_ACCESS and tagged_vlans:
             raise forms.ValidationError({"mode": "An access interface cannot have tagged VLANs assigned."})
```
gsnider2195, line 193: Will this code path ever run? It looks like the statement above will catch this same condition.

gsnider2195 (approved):

Issue #1063, opened 2021-11-08
Add a generic Role model


dgarros:
### Environment
* Python version: 3.77
* Nautobot version: 1.1.4

### Proposed Functionality
Add a generic Role model, similar to Status to avoid creating multiple models that are describing pretty much the same things.
Currently in Core, we are leveraging roles in multiple places : 
- Dynamic based on database model : `ipam:Role`, `dcim:DeviceRole`, `dcim: RackRole`
- Static : `ipam ? IPAddressRoleChoices`
> I wouldn't be surprised if we are using roles in more places

### Use Case
- Improve the user experience by allowing users to manage roles in a single place.
- Reduce the effort to add role to more models, as part of core or as part of a plugin 
- Allow static role to become dynamic.

### Database Changes
Yes
At a minimum we need to add a new role model, most likely as part of Extras that will be very similar to Status. 
At some point, it would be good to migrate all existing roles to this new model but it would most likely be a breaking change

### External Dependencies
No

jathanism:
Yeah we talked about this back pre-1.0 for things like "profiles" e.g. "interface profiles" but just using roles generically. I'm a fan.

glennmatthews:
This would probably be difficult to implement in a non-breaking / backwards compatible way. Need to consider impact to the REST API, plugins potentially consuming these data models, etc.

Setting that aside, it would be worth looking at whether we even need separate data models for "status" and "role" or whether these are just two different groupings/uses of the same generic model type behind the scenes? (In theory even "tag" could be subsumed into this as well...)

Currently we're leaning towards targeting this as a 2.x feature - not to rule out implementing it before then if these sorts of concerns can be addressed in a backwards-compatible way.

dgarros:
Understood, retrofitting the existing models wouldn't be backward compatible

It might be possible to introduce the new Role model first, to allow plugins to leverage it and retrofit the existing models later as part of a major release.

jvanderaa:
This is probably something that could be implemented across the boards, including supplanting Location Type. @bryanculver

Kircheneer:
I can also see use in this as a feature for plugin developers, if all I need is an extendable list of possible types otherwise I would have to define a custom model with lots of extra overhead, if we had this and it was extendable to plugin models that could save me as a plugin developer some time. One example from a currently open-sourced plugin would be `nautobot_bgp_models.models.PeeringRole`, this is a very simple model without custom logic that just provides an extendable role field.

[...]

PR #2798, opened 2022-11-10
Mixin classes renaming and filter code reorganization

glennmatthews:
# Closes: #2779
# What's Changed

Groundwork for #2723 and 2.0.

- `nautobot.dcim.filter_mixins` becomes `nautobot.dcim.filters.mixins` submodule
- Create `nautobot.extras.filters.mixins` and `nautobot.extras.filters.customfields` submodules
- Create `nautobot.tenancy.filters.mixins` submodule
- Rename a bunch of mixin classes for consistency and clarity, with backwards-compatibility aliases as appropriate:

| Former Name                    | New Name                                     |
| ------------------------------ | -------------------------------------------- |
| `nautobot.dcim.filters.CableTerminationFilterSet`    | `nautobot.dcim.filters.mixins.CableTerminationModelFilterSetMixin`        |
| `nautobot.dcim.api.serializers.CableTerminationSerializer`   | `nautobot.dcim.api.serializers.CableTerminationModelSerializerMixin`       |
| `nautobot.dcim.api.serializers.ConnectedEndpointSerializer`  | `nautobot.dcim.api.serializers.PathEndpointModelSerializerMixin`           |
| `nautobot.dcim.filters.ConnectionFilterSet`          | `nautobot.dcim.filters.mixins.ConnectionFilterSetMixin`                   |
| `nautobot.extras.filters.CreatedUpdatedFilterSet`      | `nautobot.extras.filters.mixins.CreatedUpdatedModelFilterSetMixin`          |
| `nautobot.extras.filters.CustomFieldModelFilterSet`    | `nautobot.extras.filters.mixins.CustomFieldModelFilterSetMixin`             |
| `nautobot.extras.api.customfields.CustomFieldModelSerializer`   | `nautobot.extras.api.customfields.CustomFieldModelSerializerMixin`            |
| `nautobot.dcim.filters.DeviceComponentFilterSet`     | `nautobot.dcim.filters.mixins.DeviceComponentModelFilterSetMixin`         |
| `nautobot.dcim.filters.DeviceTypeComponentFilterSet` | `nautobot.dcim.filters.mixins.DeviceComponentTemplateModelFilterSetMixin` |
| `nautobot.extras.filters.LocalContextFilterSet`        | `nautobot.extras.filters.mixins.LocalContextModelFilterSetMixin`            |
| `nautobot.dcim.filters.PathEndpointFilterSet`        | `nautobot.dcim.filters.mixins.PathEndpointModelFilterSetMixin`            |
[...]

Issue #2722, opened 2022-11-03
Apps Refactor: Collapse Extras App to Core


bryanculver:
**Moving Non-Model Features** -3
- [ ] Subdivide different components of extras
- [ ] Lift-and-shift the code from extras to core
- [ ] Correct import paths from other apps to the now core import path

**Moving Model Features**
_Needs to be broken out to it's own story_
- [x] #2826
- Migration import paths change (looking at data migrations)
- Model changes incur impact to URLs, APIs, etc
-

glennmatthews:
After #3132, the merged core+utilities app has the following remaining imports from extras:

```
> git grep 'import.*extras\|extras.*import' nautobot/core
nautobot/core/api/metadata.py:from nautobot.extras.api.fields import StatusSerializerField
nautobot/core/api/urls.py:from nautobot.extras.plugins.urls import plugin_api_patterns
nautobot/core/apps/__init__.py:from nautobot.extras.plugins.utils import import_object
   import_object should become a core utility or be removed.
nautobot/core/apps/__init__.py:from nautobot.extras.registry import registry
   extras.registry should become a core utility 
nautobot/core/apps/__init__.py:        from nautobot.extras.models import mixins
   This is for the TaggableManager monkey_mix, which is a 2.0 TODO comment to rework already.
nautobot/core/celery/schedulers.py:from nautobot.extras.models import ScheduledJob, ScheduledJobs
   May be addressed by work like #2991?
nautobot/core/cli.py:from nautobot.extras.plugins.utils import load_plugins
nautobot/core/factory.py:from nautobot.extras.models import Tag
   Tags should be moved to core
nautobot/core/filters.py:        from nautobot.extras import models as extras_models  # avoid circular import
   Tags should be moved to core
nautobot/core/forms/fields.py:        from nautobot.extras import utils as extras_utils
   FeatureQuery should be moved to core
nautobot/core/graphql/__init__.py:from nautobot.extras.models import GraphQLQuery
   GraphQLQuery model should be moved to core
nautobot/core/graphql/generators.py:from nautobot.extras.choices import RelationshipSideChoices
nautobot/core/graphql/generators.py:from nautobot.extras.models import RelationshipAssociation
nautobot/core/graphql/schema.py:from nautobot.extras.registry import registry
nautobot/core/graphql/schema.py:from nautobot.extras.models import ComputedField, CustomField, Relationship
[...]

PR #2822, opened 2022-11-14
Role model definition

timizuoebideri1:
# Closes: #1063 
# What's Changed

Role model and further related classes that are pertinent to Role models were created.

This PR only focuses on how role models would look like; If accepted, a different PR would be made that incorporates the role into the necessary models.

# TODO

- [ ] Unit, Integration Tests

PR Milestones:
- [x] Create Role model
- [x] API view for role model, this PR would include: `filterform`, `serializers`, `ViewSet` and URLs -> [PR](https://github.com/nautobot/nautobot/pull/2884)
- [x] UI view for role model - CRUD operations -> [PR](https://github.com/nautobot/nautobot/pull/2885)
- [x] Add data migrations for roles i.e In IPAddress, there are several default roles that should be introduced automatically. and obtain existing roles in all apps -> [PR](https://github.com/nautobot/nautobot/pull/2911)
- [x] Add/update role fields to these models: Racks, Device, VirtualMachine, prefix, vlan, ipaddress
- [x] Remove outdated code and Models that was rendered obsolete by the implementation of a role model.
- [x] Add testcases
- [ ] Add documentation


nautobot/extras/utils.py
```diff
@@ -172,6 +172,44 @@ def get_choices(self):
...
+class RoleModelsQuery(FeaturedQueryMixin):
+    """
+    Helper class to get ContentType models that implements role.
+    """
+
+    def list_subclasses(self):
+        """
+        Return a list of classes that implements roles e.g roles = ...
+        """
+        return [_class for _class in apps.get_models() if hasattr(_class, "roles")]
```
glennmatthews, line 210: Probably good to check that `roles` is actually a ReverseRelation to the `Role` model?

nautobot/extras/utils.py
```diff
@@ -172,6 +172,44 @@ def get_choices(self):
...
+        for model in self.list_subclasses():
+            query |= Q(app_label=model._meta.app_label, model=model.__name__.lower())
+
+        return query
+
+    def as_queryset(self):
+        return ContentType.objects.filter(self()).order_by("app_label", "model")
+
+    def get_choices(self):
+        return [(f"{ct.app_label}.{ct.model}", ct.pk) for ct in self.as_queryset()]
```
glennmatthews, line 197: Is the plan to refactor `TaggableClassesQuery` to inherit from `FeaturedQueryMixin`? Seems like there's a lot of duplication otherwise.

glennmatthews (commented):
Looks like a solid start!

nautobot/extras/utils.py
timizuoebideri1: Yup

nautobot/extras/utils.py
gsnider2195: Would `issubclass(_class, RoleModel)` work here?

nautobot/extras/models/roles.py
```diff
@@ -0,0 +1,93 @@
...
+    #  all contain similar fields
+
+    content_types = models.ManyToManyField(
+        to=ContentType,
+        related_name="roles",
+        verbose_name="Content type(s)",
+        limit_choices_to=RoleModelsQuery(),
+        help_text="The content type(s) to which this role applies.",
+    )
+    name = models.CharField(max_length=50, unique=True)
```
gsnider2195, line 35: Should the uniqueness constraint be on name and content_types? I could see the use case for having multiple roles with the same name but different content types and different descriptions.

nautobot/extras/models/roles.py
```diff
@@ -0,0 +1,93 @@
...
+    content_types = models.ManyToManyField(
+        to=ContentType,
+        related_name="statuses",
+        verbose_name="Content type(s)",
+        limit_choices_to=RoleModelsQuery(),
+        help_text="The content type(s) to which this status applies.",
+    )
+    name = models.CharField(max_length=50, unique=True)
+    color = ColorField(default=ColorChoices.COLOR_GREY)
+    slug = AutoSlugField(populate_from="name", max_length=50)
```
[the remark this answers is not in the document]
timizuoebideri1: IMO `http://localhost:8080/a-slug-used` is more url friendly than `http://localhost:8080/a%20slug%20used`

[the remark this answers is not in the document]
bryanculver: I agree with both but I think we can leave this as-is and once we address #2834 (part of #1574) we can probably use the natural key in the URL instead of with spaces.

nautobot/extras/fields.py
```diff
@@ -0,0 +1,96 @@
...
+
+from django.contrib.contenttypes.models import ContentType
+from django.db import models
+from django.utils.encoding import force_str
+from django.utils.hashable import make_hashable
+
+from nautobot.utilities.forms import DynamicModelChoiceField
+
+
+class LimitedChoiceField(models.ForeignKey):
```
glennmatthews, line 11: Can we give this a less generic name, perhaps something like `ForeignKeyLimitedByContentTypes` or `ContentTypeLimitedForeignKey`?

nautobot/extras/fields.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+    Abstract model database field that automatically limits custom choices.
+
+    The limit_choices_to for the field are automatically derived from:
+
+        - the content-type to which the field is attached (e.g. `dcim.device`)
```
glennmatthews, line 17: I don't have exact wording in mind but think it would be good to elaborate on this statement a bit for clarity. This is a bit complex/confusing of a feature.

nautobot/extras/fields.py
```diff
@@ -0,0 +1,96 @@
...
+            field_method = getattr(self, field.name)
+            return getattr(field_method, "color")
+
+        # Install `.get_FOO_color()` onto the model using our own version.
+        if f"get_{self.name}_color" not in cls.__dict__:
+            setattr(
+                cls,
+                f"get_{self.name}_color",
+                partialmethod(_get_FIELD_color, field=self),
+            )
```
glennmatthews, line 85: While I understand this is common between `status` and `role`, it feels out of scope for this generic class to me.

nautobot/extras/models/base_properties.py
```diff
@@ -0,0 +1,78 @@
...
+        Return all `self.model` instances assigned to the given model.
+        """
+        content_type = ContentType.objects.get_for_model(model._meta.concrete_model)
+        return self.filter(content_types=content_type)
+
+    def get_by_natural_key(self, name):
+        return self.get(name=name)
+
+
+class BasePropertiesModel(BaseModel, ChangeLoggedModel, CustomFieldModel, RelationshipModel, NotesMixin):
```
glennmatthews, line 28: Definitely need a less generic name for this class.

glennmatthews, line 28: Any reason this can't be an `OrganizationalModel` subclass?

nautobot/extras/models/base_properties.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+
+    # Todo(timizuo): Tag should inherit from this model; but
+    #  cant because of field conflicts: name and slug field.
+    content_types = models.ManyToManyField(
+        to=ContentType,
+        help_text="The content type(s) to which this model applies.",
+    )
+    name = models.CharField(max_length=50, unique=True)
+    slug = AutoSlugField(populate_from="name", max_length=50)
```
glennmatthews, line 41: Consider increasing the limit on these - while `Status` is currently limited to 50 characters,  `DeviceRole`, `RackRole`,  and `ipam.Role` all allow up to 100 characters for name/slug.

nautobot/extras/models/roles.py
```diff
@@ -0,0 +1,49 @@
...
+    """Model database field that automatically limits role choices
+    depending on the model implementing it.
+    """
+
+    def set_defaults(self, **kwargs):
+        kwargs.setdefault("to", Role)
+        return super().set_defaults(**kwargs)
+
+
+class RoleModel(models.Model):
```
glennmatthews, line 38: We should consider possibly `RoleModelMixin`, in the spirit of #2798?

nautobot/extras/models/roles.py
```diff
@@ -0,0 +1,49 @@
...
+    "webhooks",
+)
+class Role(BasePropertiesModel):
+    content_types = models.ManyToManyField(
+        to=ContentType,
+        related_name="roles",
+        verbose_name="Content type(s)",
+        limit_choices_to=RoleModelsQuery(),
+        help_text="The content type(s) to which this role applies.",
+    )
```
glennmatthews, line 25: Looking at the models this might eventually replace, `ipam.Role` also has a `weight` field, so we might consider including that on this model?

nautobot/extras/models/statuses.py
```diff
@@ -98,75 +55,9 @@ class StatusField(models.ForeignKey):
...
-        """Return a prepped formfield for use in model forms."""
-        defaults = {
-            "form_class": DynamicModelChoiceField,
-            "queryset": Status.objects.all(),
-            # label_lower e.g. "dcim.device"
-            "query_params": {"content_types": self.model._meta.label_lower},
-        }
-        defaults.update(**kwargs)
-        return super().formfield(**defaults)
+        return kwargs
```
glennmatthews, line 60: missing `super()` call here?

nautobot/extras/models/base_properties.py
timizuoebideri1: status do not inherit from `DynamicGroupMixin` guess its same for role.

nautobot/extras/models/roles.py
timizuoebideri1: Currently this would cause partial import because of  `SetFieldColorAndDisplayMixin`

nautobot/extras/models/roles.py
timizuoebideri1: Make sense, but the problem is i cant use `unique_together` on a `ManyToMany` field, i could possibly index `name` and add a custom validation to the `save` method. What do you think @gsnider2195

nautobot/extras/models/base_properties.py
timizuoebideri1: Any ideas?

glennmatthews: Is there a reason it *shouldn't* inherit from DynamicGroupMixin?

glennmatthews: `NameColorContentTypesModel`? :-)

nautobot/extras/models/roles.py
glennmatthews: Not sure I understand your response. I was just asking about a possible name change for this class.

nautobot/extras/fields.py
```diff
@@ -0,0 +1,47 @@
...
+class ForeignKeyLimitedByContentTypes(models.ForeignKey):
+    """
+    An abstract model field that automatically restricts ForeignKey options based on content_types.
+
+    For instance, if the model "Role" contains two records: role_1 and role_2, role_1's content_types
+    are set to "dcim.site" and "dcim.device" while the role_2's content_types are set to
+    "circuit.circuit" and "dcim.site."
+
+    If Device Model contains a field role, then role_1 is the only Role that is available,
+    while role_1 & role_2 are the only Roles that are available for Status.
```
glennmatthews, line 16:
```suggestion
    Then, for the field `role` on the Device model, role_1 is the only Role that is available,
    while role_1 & role_2 are both available for the Site model.
```

nautobot/extras/models/roles.py
timizuoebideri1: Oooh 🙊

nautobot/extras/models/base_properties.py
jathanism: No there isn't a good reason. Status seems like it should/could certainly benefit from supporting dynamic groups, but we need to be cautious about the chicken and the egg problem as it relates to multiple inheritance here, in case there is one.

nautobot/extras/models/base_properties.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+
+    # Todo(timizuo): Tag should inherit from this model; but
+    #  cant because of field conflicts: name and slug field.
+    content_types = models.ManyToManyField(
+        to=ContentType,
+        help_text="The content type(s) to which this model applies.",
+    )
+    name = models.CharField(max_length=100, unique=True)
+    slug = AutoSlugField(populate_from="name", max_length=100)
```
jathanism, line 41: Since this is "new world" for v2, and we're trying to elminate slugs, let's go ahead and remove this field?

jathanism, line 41: Since this is "new world" for v2, and we're trying to eliminate slugs, let's go ahead and remove this field?

nautobot/extras/models/base_properties.py
```diff
@@ -0,0 +1,78 @@
...
+    def __str__(self):
+        return self.name
+
+    def natural_key(self):
+        return (self.name,)
+
+    # TODO(timizuo): When view url has been implemented for role; visit this
+    def get_absolute_url(self):
+        ct = f"{self._meta.app_label}:{self._meta.model_name}"
+        return reverse(ct, args=[self.slug])
```
jathanism, line 66: Similarly here, since we're getting rid of `slug`, this should just point to `pk` instead? There's no easy solution here, and for now we can just leave this until the later refactoring of removing/replacing slugs with natural keys is complete?

jathanism (changes requested):
The changes I'm requesting require an immediate decision on the path forward for slugs/URLs with slugs vs. pk/natural_key.

nautobot/extras/models/base_properties.py
bryanculver: We don't need to have the natural key methods as a part of base problem completely solved at this point but let's leave a TODO here for us to review as we get to that point. Name should be sufficient for the time being.

nautobot/extras/fields.py
```diff
@@ -0,0 +1,47 @@
+from django.contrib.contenttypes.models import ContentType
+from django.db import models
+
+from nautobot.utilities.forms import DynamicModelChoiceField
+
+
+class ForeignKeyLimitedByContentTypes(models.ForeignKey):
```
bryanculver, line 7: Trying to curtail the sprawl of additional items in `extras` (seeing as we can't accomplish #2722) in 2.0, can this live in `core`?

nautobot/extras/fields.py
```diff
@@ -0,0 +1,47 @@
+from django.contrib.contenttypes.models import ContentType
+from django.db import models
```
bryanculver, line 2: Since we only use one base here is seems?
```suggestion
from django.db.models import ForeignKey
```

nautobot/extras/models/name_color_content_types.py
```diff
@@ -0,0 +1,88 @@
...
+    NotesMixin,
+    DynamicGroupMixin,
+):
+    """
+    This abstract base properties model contains fields and functionality that are
+    shared amongst models that requires these fields: name, color, content_types and description.
+    """
+
+    # Todo(timizuo): Tag should inherit from this model; but
+    #  cant because of field conflicts: name and slug field.
```
bryanculver, line 43: `TODO` what do you propose here?

nautobot/extras/models/name_color_content_types.py
```diff
@@ -0,0 +1,88 @@
...
+
+# Inheriting from OrganizationalModel here causes partial import error
+class NameColorContentTypesModel(
+    BaseModel,
+    ChangeLoggedModel,
+    CustomFieldModel,
+    RelationshipModel,
+    NotesMixin,
+    DynamicGroupMixin,
+):
```
bryanculver, line 36: I'm going to guess this is going to cause import nightmares to be able to make this abstract base class live in `core`?

nautobot/extras/models/roles.py
```diff
@@ -0,0 +1,51 @@
...
+@extras_features(
+    "custom_fields",
+    "custom_links",
+    "custom_validators",
+    "export_templates",
+    "graphql",
+    "relationships",
+    "webhooks",
+)
+class Role(NameColorContentTypesModel):
```
bryanculver, line 19: Same as the abstract base class, is there any possibility we could have a "freeze" for all new models being added to extras, since we won't be able to do #2722 in 2.0?

nautobot/extras/models/name_color_content_types.py
timizuoebideri1: I'm not intending on fixing this in this PR or any other PR relating to this issue

nautobot/core/models/name_color_content_types.py
```diff
@@ -0,0 +1,88 @@
...
+        Return all `self.model` instances assigned to the given model.
+        """
+        content_type = ContentType.objects.get_for_model(model._meta.concrete_model)
+        return self.filter(content_types=content_type)
+
+    def get_by_natural_key(self, name):
+        return self.get(name=name)
+
+
+# Inheriting from OrganizationalModel here causes partial import error
```
glennmatthews, line 28: Can we put a TODO here?

Alternately, could we make this a mixin instead of a base class, and thus have Role defined as `class Role(NameColorContentTypesMixin, OrganizationalModel):`?

nautobot/core/models/name_color_content_types.py
```diff
@@ -0,0 +1,88 @@
...
+    RelationshipModel,
+    NotesMixin,
+    DynamicGroupMixin,
+):
+    """
+    This abstract base properties model contains fields and functionality that are
+    shared amongst models that requires these fields: name, color, content_types and description.
+    """
+
+    # Todo(timizuo): Tag should inherit from this model; but
```
glennmatthews, line 42:
```suggestion
    # TODO(timizuo): Tag should inherit from this model; but
```

nautobot/core/models/name_color_content_types.py
```diff
@@ -0,0 +1,88 @@
...
+    def natural_key(self):
+        return (self.name,)
+
+    # TODO(timizuo): When view url has been implemented for role; visit this
+    def get_absolute_url(self):
+        ct = f"{self._meta.app_label}:{self._meta.model_name}"
+        # TODO(timizuo): Replace self.slug with natural key or pk
+        return reverse(ct, args=[self.slug])
+
+    def get_label(self):
```
glennmatthews, line 78: Is there a reason this is called `get_label`? I don't see what a list of content-types has to do with labels?

nautobot/extras/migrations/0054_role_and_alter_status.py
```diff
@@ -0,0 +1,63 @@
...
+    dependencies = [
+        ("contenttypes", "0002_remove_content_type_name"),
+        ("extras", "0053_relationship_required_on"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="status",
+            name="name",
+            field=models.CharField(max_length=100, unique=True),
```
glennmatthews, line 23: Why isn't there also a migration on `status.slug`, which is also changing max_length?

nautobot/extras/models/name_color_content_types.py
bryanculver: What do you propose we do for `Tag` then?

nautobot/extras/models/roles.py
bryanculver: Is it possible for this to live in Core?

nautobot/extras/models/mixins.py
```diff
@@ -143,3 +146,71 @@ def formfield(self, form_class=DynamicModelMultipleChoiceField, **kwargs):
         kwargs.setdefault("required", False)
         kwargs.setdefault("query_params", {"content_types": self.model._meta.label_lower})
         return self._no_monkey.formfield(self, form_class=form_class, **kwargs)
+
+
+class SetFieldColorAndDisplayMixin:
```
bryanculver, line 151: Can this Mixin live in Core?

nautobot/extras/models/mixins.py
```diff
@@ -143,3 +146,71 @@ def formfield(self, form_class=DynamicModelMultipleChoiceField, **kwargs):
...
+            field_method = getattr(self, field.name)
+            return getattr(field_method, "color")
+
+        # Install `.get_FOO_color()` onto the model using our own version.
+        if f"get_{self.name}_color" not in cls.__dict__:
+            setattr(
+                cls,
+                f"get_{self.name}_color",
+                partialmethod(_get_FIELD_color, field=self),
+            )
```
glennmatthews, line 216: I wonder whether this is a pattern we actually want to perpetuate. I only see one instance of `.get_status_color()` in the current code, and I don't really see how it's in any way preferable to just calling `.status.color` instead. IIRC this is a leftover from old NetBox code where statuses were a hard-coded list of choices instead of a model with an actual `.color` attribute.

`get_status_display()` is used a lot more widely in the current code, but again, it should be trivially replaceable with `.status.name` throughout as well.

nautobot/extras/models/roles.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class RoleModelMixin(models.Model):
+    """
+    Abstract base class for any model which may have roles.
+    """
+
+    role = RoleField(
+        on_delete=models.PROTECT,
+        related_name="%(app_label)s_%(class)s_related",  # e.g. dcim_device_related
```
glennmatthews, line 47: Bearing in mind #2960, we should make sure we don't have to change this later to comply with #2960 - not sure if that means this should be just `related_name="%(class)ss"` or what exactly, but let's try to get it right.

nautobot/extras/utils.py
```diff
@@ -73,38 +73,44 @@ def image_upload(instance, filename):
...
-    """
+class FeaturedQueryMixin:
+    """Mixin class that gets a list of featured models."""
 
     def list_subclasses(self):
-        """
-        Return a list of classes that implement the to_objectchange method
-        """
-        return [_class for _class in apps.get_models() if hasattr(_class, "to_objectchange")]
+        """Return a list of classes that has implements this `name`."""
```
glennmatthews, line 81:
```suggestion
```

nautobot/extras/utils.py
```diff
@@ -155,21 +161,26 @@ def list_subclasses(self):
...
+        """
+        Return a list of classes that implements roles e.g roles = ...
+        """
+        # Avoid circular imports
+        from nautobot.extras.models.roles import RoleField
+
+        model_classes = []
+        for model_class in apps.get_models():
+            role_field = getattr(model_class, "roles", None)
+            if role_field and isinstance(role_field, RoleField):
```
glennmatthews, line 181: Perhaps?

```suggestion
            if hasattr(model_class, "roles") and isinstance(model_class.roles, RoleField):
```

glennmatthews (changes requested):

nautobot/core/models/name_color_content_types.py
timizuoebideri1: Importing `OrganizationalModel` in status / role causes partial import error

nautobot/extras/migrations/0054_role_and_alter_status.py
timizuoebideri1: migration file 0014_auto_slug shows slugs' `max_length` has been 100.

nautobot/extras/models/roles.py
timizuoebideri1: i think it would be more of `related_name="%(app_label)s"` => Role.devices
@gsnider2195

nautobot/extras/models/name_color_content_types.py
timizuoebideri1: replace `TagBase` with `NameColorContentTypesModel`

nautobot/dcim/api/serializers.py
```diff
@@ -782,7 +750,7 @@ class Meta:
             "url",
             "name",
             "device_type",
-            "device_role",
+            "role",
```
glennmatthews, line 753: This change needs to be documented in the v2 migration docs. Take a look at the format I've used for similar changes in https://github.com/nautobot/nautobot/pull/3012/files#diff-d5e7d13f177913ec08c4f0735b2744d57d1b223ee1eb55fbd577505cb16fea23 and see what you think.

nautobot/dcim/factory.py
```diff
@@ -121,20 +119,6 @@ class Params:
     comments = factory.Maybe("has_comments", factory.Faker("paragraph"), "")
 
 
-class DeviceRoleFactory(OrganizationalModelFactory):
```
glennmatthews, line 124: Need to create/enable a new `nautobot.extras.factory.RoleFactory` equivalent.

nautobot/dcim/filters/__init__.py
```diff
@@ -122,7 +120,6 @@
     "RackFilterSet",
     "RackGroupFilterSet",
     "RackReservationFilterSet",
-    "RackRoleFilterSet",
```
glennmatthews, line 125: also need to remove "DeviceRoleFilterSet" from this list?

nautobot/docs/additional-features/jobs.md
```diff
@@ -753,7 +753,7 @@ These variables are presented as a web form to be completed by the user. Once su
 ```python
 from django.utils.text import slugify
 
-from nautobot.dcim.models import Device, DeviceRole, DeviceType, Manufacturer, Site
+from nautobot.dcim.models import Device, DeviceType, Manufacturer, Site
 from nautobot.extras.models import Status
```
glennmatthews, line 757:
```suggestion
from nautobot.extras.models import Role, Status
```

nautobot/docs/additional-features/jobs.md
```diff
@@ -753,7 +753,7 @@ These variables are presented as a web form to be completed by the user. Once su
...
 from nautobot.extras.jobs import *
```
glennmatthews, line 758: Also need to replace `DeviceRole` with `Role` later in this example Job.

nautobot/extras/api/serializers.py
```diff
@@ -36,6 +35,7 @@
     ObjectChangeActionChoices,
 )
 from nautobot.extras.datasources import get_datasource_content_choices
+from ...core.models.mixins import LimitQuerysetChoicesSerializerMixin
```
glennmatthews, line 38:
```suggestion
from nautobot.core.models.mixins import LimitQuerysetChoicesSerializerMixin
```

nautobot/extras/api/serializers.py
```diff
@@ -112,6 +113,7 @@
     NestedTagSerializer,
     NestedWebhookSerializer,
     NestedJobHookSerializer,
+    NestedRoleSerializer,
```
glennmatthews, line 116: Can you move these two to their appropriate alphabetized position in the list please?

nautobot/extras/filters/__init__.py
```diff
@@ -216,12 +218,12 @@ class ConfigContextFilterSet(BaseFilterSet):
...
+        queryset=Role.objects.all(),
         label="Role",
     )
     role = django_filters.ModelMultipleChoiceFilter(
         field_name="roles__slug",
-        queryset=DeviceRole.objects.all(),
+        queryset=Role.objects.all(),
         to_field_name="slug",
         label="Role (slug)",
     )
```
glennmatthews, line 229: Possibly collapse these to a single `NaturalKeyOrPKMultipleChoiceFilter` while you're touching this? Or is that out of scope?

nautobot/extras/forms/mixins.py
```diff
@@ -687,6 +691,43 @@ def _append_relationships_side(self, relationships, initial_side):
...
+    Mixin to add non-required `role` multiple-choice field to filter forms.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["role"] = DynamicModelMultipleChoiceField(
+            required=False,
+            queryset=Role.objects.all(),
+            query_params={"content_types": self.model._meta.label_lower},
+            to_field_name="name",  # TODO(timizuo): Might be slug but i dont think so
```
glennmatthews, line 718: At the very least it appears to be inconsistent with `StatusModelFilterFormMixin` -- intentional?

nautobot/extras/models/models.py
```diff
@@ -102,7 +102,12 @@ class ConfigContext(BaseModel, ChangeLoggedModel, ConfigContextSchemaValidationM
     regions = models.ManyToManyField(to="dcim.Region", related_name="+", blank=True)
     sites = models.ManyToManyField(to="dcim.Site", related_name="+", blank=True)
     locations = models.ManyToManyField(to="dcim.Location", related_name="+", blank=True)
-    roles = models.ManyToManyField(to="dcim.DeviceRole", related_name="+", blank=True)
+    roles = models.ManyToManyField(
+        to="extras.Role",
+        related_name="+",
+        blank=True,
+        limit_choices_to=RoleModelsQuery(),
```
glennmatthews, line 109: Maybe I'm missing something - I don't see why `RoleModelsQuery` applies here. That returns a list of ContentTypes, not a list of Roles?

nautobot/extras/querysets.py
```diff
@@ -14,8 +14,8 @@ def get_for_object(self, obj):
         Return all applicable ConfigContexts for a given object. Only active ConfigContexts will be included.
         """
 
-        # `device_role` for Device; `role` for VirtualMachine
-        role = getattr(obj, "device_role", None) or obj.role
+        # `role` for Device; `role` for VirtualMachine
```
glennmatthews, line 17: probably can just remove this comment :)

nautobot/extras/tests/integration/__init__.py
```diff
@@ -1,12 +1,12 @@
...
+from nautobot.dcim.models import Device, DeviceType, Manufacturer, Site
 
 
 def create_test_device():
     test_uuid = str(uuid.uuid4())
-    device_role = DeviceRole.objects.create(
+    device_role = Role.objects.create(
         name=f"Test Role {test_uuid}",
         slug=f"test-role-{test_uuid}",
     )
```
glennmatthews, line 12: probably need to set `device_role.content_types` too?

nautobot/ipam/factory.py
```diff
@@ -362,7 +346,6 @@ class Meta:
     )
 
     has_role = factory.Faker("pybool")
-    role = factory.Maybe("has_role", random_instance(Role), None)
```
glennmatthews, line 365: this needs to change to use Roles from the extras factory, instead of being removed

nautobot/ipam/factory.py
```diff
@@ -439,7 +422,6 @@ class Params:
     location = factory.Maybe(
         "has_location", random_instance(lambda: Location.objects.get_for_model(Prefix), allow_null=False), None
     )
-    role = factory.Maybe("has_role", random_instance(Role), None)
```
glennmatthews, line 442: this needs to change to use Roles from the extras factory, instead of being removed

nautobot/ipam/tables.py
```diff
@@ -44,25 +42,6 @@
...
-{% load helpers %}
-{% for i in record.parents|as_range %}
-    <i class="mdi mdi-circle-small"></i>
-{% endfor %}
-<span class="hover_copy"><a href="{% if record.present_in_database %}{% url 'ipam:prefix' pk=record.pk %}{% else %}{% url 'ipam:prefix_add' %}?prefix={{ record }}{% if object.vrf %}&vrf={{ object.vrf.pk }}{% endif %}{% if object.site %}&site={{ object.site.pk }}{% endif %}{% if object.tenant %}&tenant_group={{ object.tenant.group.pk }}&tenant={{ object.tenant.pk }}{% endif %}{% endif %}" id="copy_{{record.id}}">{{ record.prefix }}</a><button type="button" class="btn btn-inline btn-default hover_copy_button" data-clipboard-target="#copy_{{record.id}}">
-                                <span class="mdi mdi-content-copy"></span>
-                            </button>
-                        </span>
-"""
-
```
glennmatthews, line 57: Why is this removed?

nautobot/utilities/templatetags/helpers.py
```diff
@@ -679,3 +679,16 @@ def custom_branding_or_static(branding_asset, static_asset):
     if settings.BRANDING_FILEPATHS.get(branding_asset):
         return f"{ settings.MEDIA_URL }{ settings.BRANDING_FILEPATHS.get(branding_asset) }"
     return StaticNode.handle_simple(static_asset)
+
+
+@library.filter()
+@register.filter()
+def render_related_name_and_color(model, field_name):
```
glennmatthews, line 686: Given that this doesn't actually need model at all (only obj) what if we changed it to just take the obj directly as input? 

```
{{ object.status | render_name_and_color }}
{{ object.role | render_name_and_color }}
```

Actually, what if we took this even a step further, and just made this an enhancement to the existing `hyperlinked_object` filter (i.e. handle the case where the object has a `color` field)?

glennmatthews (commented):
Porting over comments from #2911.

nautobot/utilities/templatetags/helpers.py
timizuoebideri1: Changing the '`hyperlinked_object`' to add button styling for `objs` with `color` field has an effect on the breadcrumb. That is, breadcrumbs of roles and status, or anything with a `color` field, appears as a button.

timizuoebideri1:
Also, I've moved most of the `classes` and `functions` relevant to `roles` to `core`; those that haven't been moved generated partial import issues, therefore they haven't been transferred to `core`.

nautobot/utilities/templatetags/helpers.py
glennmatthews: Ah, makes sense. Thanks.

glennmatthews:
> Also, I've moved most of the `classes` and `functions` relevant to `roles` to `core`; those that haven't been moved generated partial import issues, therefore they haven't been transferred to `core`.

At a quick glance, I'm not sure having this feature partly in core and partly in extras is an improvement. Makes it just that much less intuitive to find things. (nested serializer in core, but non-nested serializer in extras? filterset and forms in core, but views in extras?)

bryanculver:
> > Also, I've moved most of the `classes` and `functions` relevant to `roles` to `core`; those that haven't been moved generated partial import issues, therefore they haven't been transferred to `core`.
> 
> At a quick glance, I'm not sure having this feature partly in core and partly in extras is an improvement. Makes it just that much less intuitive to find things. (nested serializer in core, but non-nested serializer in extras? filterset and forms in core, but views in extras?)

It was my request to not add additional models into extras if we could avoid it. I'd like to see us try to keep it in core and see how non-ideal the imports need to be to be able to do that. At least until we can do #2722 maybe we can draw the line at no new ones.

nautobot/dcim/tests/test_views.py
```diff
@@ -111,12 +110,13 @@ class RegionTestCase(ViewTestCases.OrganizationalObjectViewTestCase):
     def setUpTestData(cls):
 
         # Create three Regions
+        parent_region = Region.objects.filter(children__isnull=True, parent__isnull=True).first()
```
glennmatthews, line 113: Why `children__isnull=True`?

nautobot/extras/factory.py
```diff
@@ -5,12 +5,37 @@
...
+    class Meta:
+        model = Role
+        exclude = ("has_description",)
+
+    name = factory.LazyFunction(
+        lambda: "".join(word.title() for word in faker.Faker().words(nb=2, part_of_speech="adjective", unique=True))
+    )
+    slug = factory.LazyAttribute(lambda role: slugify(role.name))
+    color = factory.Iterator(ColorChoices.CHOICES, getter=lambda choice: choice[0])
+
```
glennmatthews, line 26: Need `has_weight` and `weight` attributes too, I believe?

nautobot/extras/migrations/0054_role_and_alter_status.py
glennmatthews: Huh, that seems like a bug in our `AutoSlugField` class then, since Status declares:

```python
    slug = AutoSlugField(populate_from="name", max_length=50)
```

Out of scope for this PR though.

nautobot/extras/models/models.py
```diff
@@ -102,7 +101,9 @@ class ConfigContext(BaseModel, ChangeLoggedModel, ConfigContextSchemaValidationM
     regions = models.ManyToManyField(to="dcim.Region", related_name="+", blank=True)
     sites = models.ManyToManyField(to="dcim.Site", related_name="+", blank=True)
     locations = models.ManyToManyField(to="dcim.Location", related_name="+", blank=True)
-    roles = models.ManyToManyField(to="dcim.DeviceRole", related_name="+", blank=True)
+    # TODO(timizuo): Find a way to limit role choices to Device; as of now using
+    #  limit_choices_to=Role.objects.get_for_model(Device), causes a partial import error
```
glennmatthews, line 105: Actually I think we'd need to limit to Device *and/or* VirtualMachine related Roles.

nautobot/extras/tests/test_filters.py
```diff
@@ -1516,3 +1514,52 @@ def test_search(self):
...
+        rack_roles = self.queryset.filter(content_types=self.rack_ct)
+        params = {"content_types": ["dcim.rack"]}
+        self.assertQuerysetEqualAndNotEmpty(self.filterset(params, self.queryset).qs, rack_roles)
+
+    def test_color(self):
+        """Test the color search field."""
+        params = {"color": [ColorChoices.COLOR_AMBER]}
+        expected_count = Role.objects.filter(color=ColorChoices.COLOR_AMBER).count()
+        self.assertEqual(self.filterset(params, self.queryset).qs.count(), expected_count)
+
```
glennmatthews, line 1553: Also test filtering by `weight`?

nautobot/extras/tests/test_views.py
```diff
@@ -2285,3 +2287,36 @@ def setUpTestData(cls):
...
+            "content_types": [content_type.pk],
+        }
+
+        cls.csv_data = (
+            "name,slug,color,content_types",
+            "test_role1,test-role1,ffffff,dcim.device",
+            'test_role2,test-role2,ffffff,"dcim.device,dcim.rack"',
+            'test_role3,test-role3,ffffff,"dcim.device,ipam.prefix"',
+            'test_role4,test-role4,ffffff,"ipam.ipaddress,ipam.vlan"',
+        )
```
glennmatthews, line 2315: Testing `weight` as well?

nautobot/ipam/models.py
```diff
@@ -386,46 +385,6 @@ def get_utilization(self):
...
-)
-class Role(OrganizationalModel):
-    """
-    A Role represents the functional role of a Prefix or VLAN; for example, "Customer," "Infrastructure," or
-    "Management."
-    """
-
-    name = models.CharField(max_length=100, unique=True)
-    slug = AutoSlugField(populate_from="name")
-    weight = models.PositiveSmallIntegerField(default=1000)
```
glennmatthews, line 403: Does changing from a `ipam.Role.weight` default of `1000` to a `extras.Role.weight` default of `None` have any impacts to the Prefix and VLAN models? I don't think there are a lot of places we actually **use** `weight` in the core code, but want to be sure.

nautobot/extras/models/roles.py
```diff
@@ -0,0 +1,61 @@
...
+)
+class Role(NameColorContentTypesModel):
+    content_types = models.ManyToManyField(
+        to=ContentType,
+        related_name="roles",
+        verbose_name="Content type(s)",
+        limit_choices_to=RoleModelsQuery(),
+        help_text="The content type(s) to which this role applies.",
+    )
+    weight = models.PositiveSmallIntegerField(null=True, blank=True)
```
glennmatthews, line 26: `ipam.Role` had `ordering` of `(weight, name)` rather than just `name`. Wonder whether it'd be desirable to do the same for the new Role model?

nautobot/utilities/templatetags/helpers.py
```diff
@@ -679,3 +679,15 @@ def custom_branding_or_static(branding_asset, static_asset):
...
     return StaticNode.handle_simple(static_asset)
+
+
+@library.filter()
+@register.filter()
+def render_related_name_and_color(obj):
+    """Render the display view of an object."""
+    if obj:
+        content = (
+            f'<span class="label" style="color: {fgcolor(obj.color)}; background-color: #{obj.color}">{obj.name}</span>'
```
glennmatthews, line 690: Would it make sense for this to also hyperlink to the detail view for `obj`?

glennmatthews:
> > > Also, I've moved most of the `classes` and `functions` relevant to `roles` to `core`; those that haven't been moved generated partial import issues, therefore they haven't been transferred to `core`.
> > 
> > 
> > At a quick glance, I'm not sure having this feature partly in core and partly in extras is an improvement. Makes it just that much less intuitive to find things. (nested serializer in core, but non-nested serializer in extras? filterset and forms in core, but views in extras?)
> 
> It was my request to not add additional models into extras if we could avoid it. I'd like to see us try to keep it in core and see how non-ideal the imports need to be to be able to do that. At least until we can do #2722 maybe we can draw the line at no new ones.

Understood - but in my opinion this partial implementation in core and partial in extras is worse than just adding it in extras for the time being.

nautobot/ipam/models.py
timizuoebideri1: On data that currently exists No I dont think there is any impact, because all records would be transferred. But when creating a new Role with its '`content type`' set to '`ipam.Vlan`'/ '`ipam.Prefix`', That default of 1000 would not be automatically added, I suppose that would be the impact

glennmatthews: Understood. I was asking more specifically what impact that has on the operation of the application - i.e. is there anywhere that's expecting all Prefix/VLAN roles to have a non-null `weight` that would have issues as a result of this change?

nautobot/dcim/forms.py
```diff
@@ -647,7 +621,6 @@ class RackForm(LocatableModelFormMixin, NautobotModelForm, TenancyForm):
         required=False,
         query_params={"site_id": "$site"},
     )
-    role = DynamicModelChoiceField(queryset=RackRole.objects.all(), required=False)
```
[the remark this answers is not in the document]
timizuoebideri1: It is not necessary because `role` form field widget is configured in `ForeignKeyLimitedByContentTypes` by overriding the `formfield` method.

nautobot/ipam/models.py
timizuoebideri1: I dont think so/ I cant find any

nautobot/utilities/utils.py
```diff
@@ -824,7 +824,7 @@ def get_filterset_parameter_form_field(model, parameter):
             "to_field_name": field.extra.get("to_field_name", "id"),
         }
         # Status and Tag api requires content_type, to limit result to only related content_types
-        if related_model in [Status, Tag]:
+        if related_model in [Status, Tag, Role]:
```
HanlinMiao, line 827:
```suggestion
        if related_model in [Role, Status, Tag]:
```
Nitpick sorry

nautobot/ipam/factory.py
```diff
@@ -362,7 +347,6 @@ class Meta:
     )
 
     has_role = factory.Faker("pybool")
-    role = factory.Maybe("has_role", random_instance(Role), None)
```
HanlinMiao, line 365: Do we need the `has_role` attribute above if we are not using it?

nautobot/docs/release-notes/version-2.0.md
```diff
@@ -10,6 +10,10 @@ If you are a user migrating from Nautobot v1.X, please refer to the ["Upgrading
 
 ### Added
 
+#### Generic Role Model ([#1063](https://github.com/nautobot/nautobot/issues/1063))
+
+DeviceRole, RackRole, IPAM Role, and IPAddressRoleChoices have all been merged into a single generic Role model. A role can now be created and associated to one or more of the content-types that previously implemented role as a field. These model content-types include dcim.device, dcim.rack, virtualization.virtualmachine, ipam.ipaddress, ipam.prefix, and ipam.vlan.
```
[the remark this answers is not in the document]
timizuoebideri1: `content-types` looks better

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -6,6 +6,7 @@
 
 | Model         | Renamed Field | New Name     |
 |---------------|---------------|--------------|
+| Device        | `device_role` | `role`       |
```
glennmatthews, line 9: This should also be documented under `Renamed Serializer Fields` I think?

nautobot/docs/models/extras/role.md
```diff
@@ -0,0 +1,81 @@
...
+
+### `RoleField` model field
+
+The `RoleField` field type subclass of a `django.db.models.ForeignKey`
+
+This model field also emits its own form field to eliminate the requirement for a form field to be explicitly added to model forms.
+
+### `RoleFilter` filter field
+
+Any filter that is intended to have a `role` field must inherit from `nautobot.extras.filters.RoleModelFilterSetMixin`. This will add a `nautobot.extras.filters.RoleFilter` to the filter, which allows filtering by the `name` of the role.
```
glennmatthews, line 34: Name only? Not name or PK?

nautobot/docs/models/extras/role.md
```diff
@@ -0,0 +1,81 @@
...
+
+When created, a `Role` can be associated to one or more model content-types using a many-to-many relationship. The relationship to each model is referenced across all user interfaces using the `{app_label}.{model}` naming convention (e.g. `dcim.device`).
+
+Roles may be managed by navigating to **Organization > Roles** in the navigation menu.
+
+### Importing Objects with a `role` Field
+
+When using CSV import to reference a `role` field on an object, the `Role.name` field is used.
+
+## Role Internals
```
glennmatthews, line 17: This is another example where we really need to separate user-facing documentation from developer-facing documentation. This whole section is very overwhelming to an end-user given that it's so much longer than the part that's actually relevant to them above. Made worse by the fact that this content is embedded in the user docs in three different places.

Can we perhaps move this content into a separate documentation file under  `nautobot/docs/development/` and just add a cross-reference here?

nautobot/docs/models/extras/role.md
```diff
@@ -0,0 +1,81 @@
...
+!!! warning
+    The section below is largely intended for developers who may need to create
+    data models of their own that implement a `role` field. Proceed at your
+    own risk!
+
+Any model that is intended to have a `role` field must inherit from either of these two mixins: `nautobot.extras.models.roles.RoleModelMixin` or `nautobot.extras.models.roles.RoleRequiredRoleModelMixin`. The `RoleModelMixin` adds a nullable `role` field, while the `RoleRequiredRoleModelMixin` adds a required `role` field. This abstract model will add an `nautobot.extras.models.roles.RoleField` to the model. The abstract base will automatically assign a `related_name` for the reverse relationship back to the inheriting model's name (e.g. `dcim_device_related`).
+
+### `RoleField` model field
+
+The `RoleField` field type subclass of a `django.db.models.ForeignKey`
```
glennmatthews, line 28:
```suggestion
The `RoleField` field type is a subclass of `django.db.models.ForeignKey`.
```

nautobot/docs/models/extras/role.md
```diff
@@ -0,0 +1,81 @@
...
+
+This model field also emits its own form field to eliminate the requirement for a form field to be explicitly added to model forms.
+
+### `RoleFilter` filter field
+
+Any filter that is intended to have a `role` field must inherit from `nautobot.extras.filters.RoleModelFilterSetMixin`. This will add a `nautobot.extras.filters.RoleFilter` to the filter, which allows filtering by the `name` of the role.
+
+### Form fields
+
+Any model form that is intended to have a `role` field must inherit from one of two mixins, depending on the use-case:
```
glennmatthews, line 38: This is misleadingly phrased. Not all model forms need these mixins (a basic create/edit form does not, for one). Should probably rephrase to make it clear that filter forms (only) need to use `RoleModelFilterFormMixin` and bulk-edit forms (only) need to use `RoleModelBulkEditFormMixin`.

nautobot/docs/models/extras/role.md
```diff
@@ -0,0 +1,81 @@
...
+- `nautobot.extras.api.serializers.RoleModelSerializerMixin` should be used if `role` field is non-required.
+- `nautobot.extras.api.serializers.RoleRequiredRoleModelSerializerMixin`should be used if `role` field is required.
+
+This adds an `nautobot.extras.api.fields.RoleSerializerField` to the serializer.
+
+### Table field
+
+If you wish for a table to include a `role` column, your table must inherit from `nautobot.extras.tables.RoleTableMixin`. This includes a `ColorColumn` with the header `role` on the table.
+
+## Role object integrations
```
glennmatthews, line 56: Is this just a repeat/summary of the previous section? Seems redundant to me.

nautobot/docs/release-notes/version-2.0.md
```diff
@@ -10,6 +10,10 @@ If you are a user migrating from Nautobot v1.X, please refer to the ["Upgrading
...
+
 #### Added Site Fields to Location ([#2954](https://github.com/nautobot/nautobot/issues/2954))
```
glennmatthews, line 18: Need an item under `### Changed` as well to indicate that the IPAddress.role field has been changed in its behavior.

Also should probably revise the role-related content in `nautobot/docs/models/ipam/ipaddress.md`.

nautobot/core/models/mixins.py
```diff
@@ -0,0 +1,14 @@
+class LimitQuerysetChoicesSerializerMixin:
```
glennmatthews, line 1: Is `core/models/mixins.py` the appropriate place for this? Seems like it's specific to the REST API, should it be in `core/api/mixins.py` instead?

nautobot/extras/factory.py
```diff
@@ -5,12 +5,42 @@
...
+            "has_weight",
+        )
+
+    name = factory.LazyFunction(
+        lambda: "".join(word.title() for word in faker.Faker().words(nb=2, part_of_speech="adjective", unique=True))
+    )
+    slug = factory.LazyAttribute(lambda role: slugify(role.name))
+    color = factory.Iterator(ColorChoices.CHOICES, getter=lambda choice: choice[0])
+    has_weight = factory.Faker("pybool")
+    weight = factory.Maybe("has_weight", factory.Faker("pyint"), None)
```
glennmatthews, line 30: Interesting that `pyint` by default only generates numbers between 0 and 9999. I was worried that it would potentially generate negative numbers that would be invalid for `weight` as a `PositiveSmallIntegerField`.

nautobot/extras/filters/__init__.py
```diff
@@ -297,6 +290,13 @@ class ConfigContextFilterSet(BaseFilterSet):
         to_field_name="slug",
         label="Tag (slug)",
     )
+    # TODO(timizuo): Limit Role choices by Device/VM but can't
+    #  import device cause of partial import error
+    roles = NaturalKeyOrPKMultipleChoiceFilter(
```
glennmatthews, line 295: Should probably be `role` (as in the removed code above) as the other filters are single rather than plural as well.

nautobot/extras/filters/__init__.py
```diff
@@ -214,17 +218,6 @@ class ConfigContextFilterSet(BaseFilterSet):
         queryset=Location.objects.all(),
         label="Location (slug)",
     )
-    role_id = django_filters.ModelMultipleChoiceFilter(
-        field_name="roles",
-        queryset=DeviceRole.objects.all(),
-        label="Role",
-    )
```
glennmatthews, line 221: Need to document this in the upgrading guide as a removed filter field.

nautobot/extras/filters/mixins.py
```diff
@@ -24,12 +25,26 @@
...
+
+class ConfigContextRoleFilter(NaturalKeyOrPKMultipleChoiceFilter):
+    """Limit role choices to the available role choices for Device and VM"""
+
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault("to_field_name", "slug")
+        kwargs.setdefault("queryset", Role.objects.get_for_models([Device, VirtualMachine]))
+        kwargs.setdefault("label", "Role (slug or ID)")
+
+        super().__init__(*args, **kwargs)
```
glennmatthews, line 47: Is this class unused?

nautobot/extras/migrations/0057_alter_role_options.py
```diff
@@ -0,0 +1,17 @@
...
+
+    dependencies = [
+        ("extras", "0056_alter_configcontext_roles"),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name="role",
+            options={"ordering": ("weight", "name")},
+        ),
```
glennmatthews, line 16: Does this need to be a separate migration, or can it be merged up into `0054_role_and_alter_status.py`?

nautobot/extras/migrations/0056_alter_configcontext_roles.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ("extras", "0055_collect_roles_from_related_apps_roles"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="configcontext",
+            name="roles",
+            field=models.ManyToManyField(blank=True, related_name="_extras_configcontext_roles_+", to="extras.Role"),
```
glennmatthews, line 16: Where do existing ConfigContexts that reference `DeviceRole` instances get migrated to instead reference the new `Role` instances?

nautobot/extras/querysets.py
```diff
@@ -123,7 +122,7 @@ def _get_config_context_filters(self):
         )
 
         if self.model._meta.model_name == "device":
-            base_query.add((Q(roles=OuterRef("device_role")) | Q(roles=None)), Q.AND)
+            base_query.add((Q(roles=OuterRef("role")) | Q(roles=None)), Q.AND)
```
glennmatthews, line 125: Should be able to pull this line out of the `if device... elif virtualmachine` blocks since it's now the same for both.

nautobot/extras/tests/test_filters.py
```diff
@@ -1519,3 +1518,62 @@ def test_search(self):
...
+            Role.objects.create(name="Role 1", color=ColorChoices.COLOR_GREY, weight=100),
+            Role.objects.create(name="Role 2", color=ColorChoices.COLOR_GREY, weight=20),
+            Role.objects.create(name="Role 3", color=ColorChoices.COLOR_AMBER, weight=10),
+            Role.objects.create(name="Role 4", color=ColorChoices.COLOR_AMBER, weight=100),
+            Role.objects.create(name="Role 5", color=ColorChoices.COLOR_AMBER, weight=100),
+        ]
+        for i in range(2):
+            roles[i].content_types.add(cls.device_ct)
+        for i in range(2, 4):
+            roles[i].content_types.add(cls.rack_ct)
```
glennmatthews, line 1541: Can we instead use the roles created by the factory?

nautobot/ipam/factory.py
```diff
@@ -361,9 +346,6 @@ class Meta:
         factory.Maybe("has_location", random_instance(Location, allow_null=False), None),
     )
 
-    has_role = factory.Faker("pybool")
```
glennmatthews, line 364: By removing `has_role` you've changed the behavior here - instead of 50% of VLANs having a `role` and 50% having none, now all of them have a `role`. Reduces the diversity of our test data; I'm against this change. Same comment on the Prefix factory below.

nautobot/ipam/views.py
```diff
@@ -316,92 +315,6 @@ class AggregateBulkDeleteView(generic.BulkDeleteView):
...
-
-        vlan_table = tables.VLANTable(vlans)
-        vlan_table.columns.hide("role")
-
-        RequestConfig(request, paginate).configure(vlan_table)
-
-        return {
-            "prefix_table": prefix_table,
-            "vlan_table": vlan_table,
-        }
```
glennmatthews, line 382: Some net loss of UI functionality here - the new generic Role detail view doesn't show prefixes and vlans that are using this role. Might be nice to keep that functionality and extend it to include devices/racks/ipaddresses as well?

glennmatthews (changes requested):

nautobot/extras/migrations/0057_alter_role_options.py
timizuoebideri1: I tried that but, `makemigration` kept on creating this migration file regardless.

nautobot/docs/development/role-internals.md
```diff
@@ -0,0 +1,37 @@
...
+
+## `RoleFilter` filter field
+
+Any filter that is intended to have a `role` field must inherit from `nautobot.extras.filters.RoleModelFilterSetMixin`. This will add a `nautobot.extras.filters.RoleFilter` to the filter, which allows filtering by the `name` or `id` of the role.
+
+## Form fields
+
+Any filter form that is intended to have a `role` field must inherit from `nautobot.extras.forms.RoleModelFilterFormMixin`. This mixin adds a non-required, multiple-choice `role` filter field to the filter form.
+
+Any bulk edit form that is intended to have a `role` field must inherit from `nautobot.extras.forms.RoleModelBulkEditFormMixin`. This mixin adds a non-required `role` field to the form, and constrains the eligible role choices to the object type being edited.
```
glennmatthews, line 24: Also need to mention `RoleModelCSVFormMixin` I think?

nautobot/extras/forms/mixins.py
```diff
@@ -687,6 +691,43 @@ def _append_relationships_side(self, relationships, initial_side):
...
+        )
+        self.order_fields(self.field_order)  # Reorder fields again
+
+
+class RoleModelCSVFormMixin(CSVModelForm):
+    """Mixin to add a required `role` choice field to CSV import forms."""
+
+    role = CSVModelChoiceField(
+        queryset=Role.objects.all(), to_field_name="name", required=False, help_text="Assigned role"
+    )
```
glennmatthews, line 728: Do we need to provide required-role/non-required-role variants of this mixin?

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -43,20 +44,33 @@
...
 
+### Replaced Models
+
+The `ipam.Role`, `dcim.RackRole`, and `dcim.DeviceRole` models have been removed and replaced by a single `extras.Role` model. This means that any references to the removed models in the code now use the `extras.Role` model instead.
+
+| Removed Model     | Replaced With  |
+|-------------------|----------------|
+| `dcim.DeviceRole` | `extras.Role`  |
+| `dcim.RackRole`   | `extras.Role`  |
+| `extras.Role`     | `extras.Role`  |
```
glennmatthews, line 55:
```suggestion
```

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -43,20 +44,33 @@
...
 ### API Behavior Changes
 
-| Model       | Field          | Changes                                                                                                 |
-|-------------|----------------|---------------------------------------------------------------------------------------------------------|
-| RackGroup   | `rack_count`   | Now only counts Racks directly belonging to this RackGroup, not those belonging to its descendants.     |
-| Region      | `site_count`   | Now only counts Sites directly belonging to this Region, not those belonging to its descendants.        |
-| TenantGroup | `tenant_count` | Now only counts Tenants directly belonging to this TenantGroup, not those belonging to its descendants. |
+| Model       | Field          | Changes                                                                                                  |
+|-------------|----------------|----------------------------------------------------------------------------------------------------------|
+| Ipaddress   | `role`         | `/ipam/ip-addresses/` endpoint now uses role nested serializer for the role field, rather than a string. |
```
glennmatthews, line 64:
```suggestion
| IPAddress   | `role`         | `/ipam/ip-addresses/` endpoint now uses role nested serializer for the role field, rather than a string. |
```

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -68,6 +82,16 @@
...
 
+### Replaced Endpoints
+
+These endpoints `/ipam/roles/`, `/dcim/rack-roles/` and `/dcim/device-roles/` are no longer available. Instead,  use the `/extras/roles/` endpoint to retrieve and manipulate `role` data.
+
+| Removed Endpoints      | Replaced With      |
+|------------------------|--------------------|
+| `/dcim/device-roles/`  | `_/extras/roles/`  |
+| `/dcim/rack-roles/`    | `_/extras/roles/`  |
+| `/ipam/roles/`         | `_/extras/roles/`  |
```
glennmatthews, line 93:
```suggestion
| `/dcim/device-roles/`  | `/extras/roles/`  |
| `/dcim/rack-roles/`    | `/extras/roles/`  |
| `/ipam/roles/`         | `/extras/roles/`  |
```

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -152,71 +176,72 @@ Below is a table documenting [corrected filter field changes](../release-notes/v
...
+| Device             | `region_id`          | instead of `/dcim/devices/?region_id=<uuid>`, use `region=<uuid>`                     |
+|                    | `site_id`            | instead of `/dcim/devices/?site_id=<uuid>`, use `site=<uuid>`                         |
+|                    | `manufacturer_id`    | instead of `/dcim/devices/?manufacturer_id=<uuid>`, use `manufacturer=<uuid>`         |
+|                    | `model`              | instead of `/dcim/devices/?model=<uuid>`, use `device_type=<uuid>`                    |
+|                    | `role_id`            | instead of `/dcim/devices/?role_id=<uuid>`, use `role=<uuid>`                         |
+|                    | `platform_id`        | instead of `/dcim/devices/?platform_id=<uuid>`, use `platform=<uuid>`                 |
+|                    | `secrets_group_id`   | instead of `/dcim/devices/?secrets_group_id=<uuid>`, use `secrets_group=<uuid>`       |
+|                    | `pass_through_ports` | instead of `/dcim/devices/?pass_through_ports=<bool>`, use `has_front                 |rear_ports`      |
```
glennmatthews, line 197: Needs fixing

nautobot/docs/models/extras/role.md
```diff
@@ -0,0 +1,17 @@
...
+
+When created, a `Role` can be associated to one or more model content-types using a many-to-many relationship. The relationship to each model is referenced across all user interfaces using the `{app_label}.{model}` naming convention (e.g. `dcim.device`).
+
+Roles may be managed by navigating to **Organization > Roles** in the navigation menu.
+
+### Importing Objects with a `role` Field
+
+When using CSV import to reference a `role` field on an object, the `Role.name` field is used.
+
+Visit [role-internals](../../development/role-internals.md) to learn more about working with `role` as a developer.`
```
glennmatthews, line 17:
```suggestion
Visit [role-internals](../../development/role-internals.md) to learn more about working with `role` as a developer.
```

nautobot/docs/release-notes/version-2.0.md
```diff
@@ -60,6 +64,12 @@ This has been addressed in v2.x as follows:
 
 Check out the specific changes documented in the table at [UI and REST API Filter Changes](../installation/upgrading-from-nautobot-v1.md#corrected-filter-fields)
 
+#### Generic Role Model ([#1063](https://github.com/nautobot/nautobot/issues/1063))
+
+The `DeviceRole`, `RackRole`, `ipam.Role`, and `IPAddressRoleChoices` have all been removed and replaced with a `extras.Role` model, This means that all references to any of the replaced models and choices now points to this generic role model.
+
+In Addition, The `role` field of the `IPAddress` model has also been changed from a choice field to a foreign key related field to the `extras.Role` model.
```
glennmatthews, line 71:
```suggestion
In addition, the `role` field of the `IPAddress` model has also been changed from a choice field to a foreign key related field to the `extras.Role` model.
```

nautobot/extras/templates/extras/role_retrieve.html
```diff
@@ -0,0 +1,96 @@
...
+                </td>
+            </tr>
+        </table>
+    </div>
+{% endblock content_left_page %}
+
+{% block content_full_width_page %}
+    <div class="panel panel-default">
+         <div class="panel-heading">
+             <strong>IPAddresses</strong>
```
glennmatthews, line 57:
```suggestion
             <strong>IP Addresses</strong>
```

nautobot/extras/templates/extras/role_retrieve.html
```diff
@@ -0,0 +1,96 @@
...
+         </div>
+         {% include 'inc/table.html' with table=ipaddress_table %}
+         {% if perms.ipam.add_ipaddress %}
+             <div class="panel-footer text-right noprint">
+                 <a href="{% url 'ipam:ipaddress_add' %}?role={{ object.pk }}" class="btn btn-xs btn-primary">
+                     <span class="mdi mdi-plus-thick" aria-hidden="true"></span> Add IPAddress
```
glennmatthews, line 63:
```suggestion
                     <span class="mdi mdi-plus-thick" aria-hidden="true"></span> Add IP Address
```

nautobot/extras/templates/extras/role_retrieve.html
```diff
@@ -0,0 +1,96 @@
...
+     </div>
+     <div class="panel panel-default">
+         <div class="panel-heading">
+             <strong>Prefixes</strong>
+         </div>
+         {% include 'inc/table.html' with table=prefix_table %}
+         {% if perms.ipam.add_prefix %}
+             <div class="panel-footer text-right noprint">
+                 <a href="{% url 'ipam:prefix_add' %}?role={{ object.pk }}" class="btn btn-xs btn-primary">
+                     <span class="mdi mdi-plus-thick" aria-hidden="true"></span> Add prefix
```
glennmatthews, line 76:
```suggestion
                     <span class="mdi mdi-plus-thick" aria-hidden="true"></span> Add Prefix
```

nautobot/extras/templates/extras/role_retrieve.html
```diff
@@ -0,0 +1,96 @@
...
+     </div>
+     <div class="panel panel-default">
+         <div class="panel-heading">
+             <strong>VLANs</strong>
+         </div>
+         {% include 'inc/table.html' with table=vlan_table %}
+         {% if perms.ipam.add_vlan %}
+             <div class="panel-footer text-right noprint">
+                 <a href="{% url 'ipam:vlan_add' %}?role={{ object.pk }}" class="btn btn-xs btn-primary">
+                     <span class="mdi mdi-plus-thick" aria-hidden="true"></span> Add vlan
```
glennmatthews, line 89:
```suggestion
                     <span class="mdi mdi-plus-thick" aria-hidden="true"></span> Add VLAN
```

nautobot/extras/templates/extras/role_retrieve.html
```diff
@@ -0,0 +1,96 @@
...
+                 </a>
+             </div>
+         {% endif %}
+     </div>
+     {% include 'inc/paginator.html' with paginator=vlan_table.paginator page=vlan_table.page %}
```
glennmatthews, line 94: Seems off to me that we include paginator.html for the VLAN table but not for the IPAddress and Prefix tables?

Also, any reason we don't also list Devices, VirtualMachines, and Racks using this role?

nautobot/docs/installation/upgrading-from-nautobot-v1.md
timizuoebideri1: Not sure I see anything that needs to be fixed here.

glennmatthews:
```suggestion
|                    | `pass_through_ports` | instead of `/dcim/devices/?pass_through_ports=<bool>`, use `has_front|rear_ports`      |
```

glennmatthews (approved):
Fix the integration tests and I think this is otherwise ready to go!

PR #2846, opened 2022-11-16
Eliminate region_id and site_id filters with NauturalKeyOrPKMultipleC…

HanlinMiao:
Eliminated `site_id` and `region_id` in dcim filters and made changes accordingly in forms and unittests

> Tactical PR to merge into `next` via #2849


glennmatthews (approved):
Looks reasonable. Thanks!

PR #2877, opened 2022-11-17
nautobot.apps module, plugin developer docs -> app developer docs

glennmatthews:
# Closes: #2723
# What's Changed

- Add `nautobot.apps` submodule and make a first pass at identifying which classes and functions should be exposed via this module for app (plugin) developers to make use of.
   - Relevant modules were identified by reviewing several of our more feature-intensive published apps, as well as the features identified in the plugin developer documentation as suitable to consume.
   - *Reminder that we can certainly add to this over time, but can't remove from it as that would be a breaking change - please keep me honest here and feel free to suggest things that should be OMITTED from this API!*
- Update example plugins to use the new `nautobot.apps` imports where appropriate.
- Move "plugin" developer documentation to be its own top-level docs heading distinct from "core" developer documentation.
   - Do a loose search-and-replace of the term "plugin" with the term "app" in the main ~plugin~ app developer documentation file. I've left some cases as "plugin" because they refer to things that I'm not changing in this PR.
- Make use of `mkdocstrings` and `mkdocs-gen-files` to autogenerate code documentation for the `nautobot.apps` module (alone, at present - can extend in the future to include the rest of nautobot as desired and as time permits)
    - Fix mkdocs warnings that this uncovered with various docstrings.
- One CSS change to the docs rendering, disallowing `code` in table columns from wrapping mid-word.

Before:

image

After:

image

# TODO

- n/a Unit, Integration Tests
- [x] Documentation Updates (when adding/changing features) - see https://docs-landing-page--2877.org.readthedocs.build/projects/core/en/2877/plugins/development/ for rendered docs


nautobot/utilities/management/commands/makemigrations.py
```diff
@@ -1,10 +1,8 @@
 # noinspection PyUnresolvedReferences
-from django.core.management.commands.makemigrations import Command
+from django.core.management.commands.makemigrations import Command  # noqa: F401
 from django.db import models
 
 from . import custom_deconstruct
 
-__all__ = ("Command",)
-
```
bryanculver, line 8: Why this? I figure we don't intend on anyone importing this file so is this just unnecessary details? Should we publish `__all__ = []` so any IDE doesn't suggest to import `Command` from here instead of the source?

bryanculver (approved):

nautobot/utilities/management/commands/makemigrations.py
glennmatthews: Having this was making mkdocstrings unhappy. I don't remember the exact error message, but given this was only added here to make flake8 happy (dd1d02363fd6fbdc0e42bb77f9b72c9d1bc85a7a) it seemed safe enough enough to remove it and add the `noqa` comment instead.

glennmatthews:
Question: for the app imports that I renamed on import (`PluginTemplateExtension` -> `TemplateExtension`, `PluginBanner` -> `Banner`, `PluginConfig` -> `NautobotAppConfig`, `PluginFilterExtension` -> `FilterExtension`), should I go ahead and rename them in the source and add a `class_deprecated_in_favor_of` for the old names?

bryanculver:
> Question: for the app imports that I renamed on import (`PluginTemplateExtension` -> `TemplateExtension`, `PluginBanner` -> `Banner`, `PluginConfig` -> `NautobotAppConfig`, `PluginFilterExtension` -> `FilterExtension`), should I go ahead and rename them in the source and add a `class_deprecated_in_favor_of` for the old names?

Because this will be more subtle than "Change `nautobot.core.X import Y` to `nautobot.apps import Y` it would be a good idea to rename those.

I would say it's not necessary overall to emit a warning that apps should import from apps verses their home because that will just get clunky.

nautobot/docs/plugins/development.md
```diff
@@ -130,100 +133,100 @@ class AnimalSoundsConfig(PluginConfig):
...
 
 ```python
 PLUGINS = ["animal_sounds"]
 ```
 
-### Verify that the Plugin is Installed
+### Verify that the App is Installed
 
-In the Nautobot UI, navigate to **Plugins -> Installed Plugins**. The newly installed plugin should appear in the displayed table if everything is configured correctly. You can also click on the plugin's name in this table to view more detailed information about this plugin based on its PluginConfig and other contents.
+In the Nautobot UI, navigate to **Plugins -> Installed Plugins**. The newly installed app should appear in the displayed table if everything is configured correctly. You can also click on the app's name in this table to view more detailed information about this app based on its NautobotAppConfig and other contents.
```
gsnider2195, line 223: Nitpick

```suggestion
After restarting the Nautobot server, the newly installed app should appear in **Plugins -> Installed Plugins** if everything is configured correctly. You can also click on the app's name in this table to view more detailed information about this app based on its NautobotAppConfig and other contents.
```

nautobot/apps/filters.py
```diff
@@ -0,0 +1,33 @@
...
+    RelationshipModelFilterSetMixin,
+    StatusModelFilterSetMixin,
+)
+from nautobot.extras.plugins import FilterExtension
+from nautobot.tenancy.filters import TenancyModelFilterSetMixin
+from nautobot.utilities.filters import (
+    BaseFilterSet,
+    MultiValueCharFilter,
+    NaturalKeyOrPKMultipleChoiceFilter,
+    SearchFilter,
```
gsnider2195, line 16:
```suggestion
    RelatedMembershipBooleanFilter,
    SearchFilter,
```

nautobot/apps/filters.py
```diff
@@ -0,0 +1,33 @@
...
+
+__all__ = (
+    "BaseFilterSet",
+    "CreatedUpdatedModelFilterSetMixin",
+    "CustomFieldModelFilterSetMixin",
+    "FilterExtension",
+    "MultiValueCharFilter",
+    "NaturalKeyOrPKMultipleChoiceFilter",
+    "NautobotFilterSet",
+    "RelationshipModelFilterSetMixin",
```
gsnider2195, line 28:
```suggestion
    "RelatedMembershipBooleanFilter",
    "RelationshipModelFilterSetMixin",
```

nautobot/apps/forms.py
```diff
@@ -0,0 +1,52 @@
+"""Forms and fields for apps to use."""
+
+from nautobot.extras.forms import (
+    CustomFieldModelBulkEditFormMixin,
+    CustomFieldModelCSVForm,
+    CustomFieldModelFormMixin,
+    NautobotBulkEditForm,
+    NautobotModelForm,
```
gsnider2195, line 8: Should `NautobotFilterForm` be exposed in `nautobot.apps`?

glennmatthews: My thinking was that it's "quasi-obsolete" now that we have dynamic filter UI, but maybe I'm looking too far into the future?

nautobot/apps/testing.py
```diff
@@ -0,0 +1,12 @@
+"""Utilities for apps to implement test automation."""
+
+from nautobot.utilities.testing.api import APITestCase, APIViewTestCases
```
gsnider2195, line 3: Should these be added?

```py
from nautobot.utilities.testing import CeleryTestCase, run_job_for_testing, TransactionTestCase
```

nautobot/apps/forms.py
gsnider2195: Probably better to leave it out and add it if we find that we still need to use it. Don't want to create unnecessary breaking changes.

gsnider2195 (approved):

Issue #2837, opened 2022-11-15
Invalid filter menu option for prefixes


matt852:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.0 (demo.nautobot.com) and on v1.5.1b1
* Python version:  unknown
* Database platform, version: unknown
* Middleware(s): unknown

### Steps to Reproduce
1. Go to IPAM --> Prefixes
2. Select the new Dynamic Filter box (NOT the "Max Length" filter box)
3. In the modal that pops up, on the Default tab, select ANY number for "Mask Length"
4. Click Apply

### Expected Behavior

The filter should work as expected.

### Observed Behavior

This error pops up for the handful of numbers I've tested this filter with:

```
Invalid filters were specified:
mask_length
Enter a number.
```

According to the UI, the filter appears to be applied but doesn't appear to be working.

![Screen Shot 2022-11-15 at 10 53 24 AM](https://user-images.githubusercontent.com/536467/201966128-17e21a12-4e14-4473-a94b-a31f63b6dd86.png)

PR #2879, opened 2022-11-18
Fix to ensure django_filters.NumberFilter is treated as a single choice field

briddo:
# Closes: #2837 
# What's Changed

`django_filters.NumberFilter` is now treated a single choice field which fixes this filter globally. Before, we would get the following with any uses of `django_filters.NumberFilter`:

```
Invalid filters were specified:
Enter a number.
```

# TODO
- [ ] Unit, Integration Tests

### DIFF
--- a/changes/2837.fixed
+++ b/changes/2837.fixed
@@ -1,1 +1,1 @@
-Changed nautobot.utilities.utils.is_single_choice_field to check if field is an instance MultipleChoiceFilter.
+Fixed incorrect logic in `nautobot.utilities.utils.is_single_choice_field` that was causing valid filters to report as invalid.
--- a/changes/2837.fixed
+++ b/changes/2837.fixed
@@ -1,1 +1,1 @@
-Added NumberFilter to nautobot.utilities.utils.is_single_choice_field check list, fixing the filter globally.
+Changed nautobot.utilities.utils.is_single_choice_field to check if field is an instance MultipleChoiceFilter.
--- a/changes/2837.fixed
+++ b/changes/2837.fixed
@@ -1,0 +1,1 @@
+Added NumberFilter to nautobot.utilities.utils.is_single_choice_field check list, fixing the filter globally.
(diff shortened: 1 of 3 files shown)


nautobot/utilities/utils.py
```diff
@@ -933,7 +933,7 @@ def convert_querydict_to_factory_formset_acceptable_querydict(request_querydict,
 def is_single_choice_field(filterset_class, field_name):
     # Some filter parameters do not accept multiple values, e.g DateTime fields, boolean fields, q field etc.
     field = get_filterset_field(filterset_class, field_name)
-    return isinstance(field, (DateFilter, DateTimeFilter, TimeFilter, BooleanFilter)) or field_name == "q"
+    return isinstance(field, (DateFilter, DateTimeFilter, NumberFilter, TimeFilter, BooleanFilter)) or field_name == "q"
```
glennmatthews, line 936: Actually, this might break `MultiValueNumberFilter` since it's a subclass of `NumberFilter`. Maybe we should invert this to be simply `not isinstance(field, django_filters.MultipleChoiceFilter)`?

glennmatthews: Probably incorrect as currently implemented for `MultiValueDateFilter`, `MultiValueDateTimeFilter`, `MultiValueTimeFilter` in fact...

changes/2837.fixed
```diff
@@ -0,0 +1 @@
+Changed nautobot.utilities.utils.is_single_choice_field to check if field is an instance MultipleChoiceFilter.
```
glennmatthews, line 1: Perhaps:

```suggestion
Fixed incorrect logic in `nautobot.utilities.utils.is_single_choice_field` that was causing valid filters to report as invalid.
```

nautobot/utilities/utils.py
```diff
@@ -931,9 +932,9 @@ def convert_querydict_to_factory_formset_acceptable_querydict(request_querydict,
 
 
 def is_single_choice_field(filterset_class, field_name):
-    # Some filter parameters do not accept multiple values, e.g DateTime fields, boolean fields, q field etc.
+    # Some filter parameters do not accept multiple values, e.g DateTime, Boolean, Int fields and the q field, etc.
     field = get_filterset_field(filterset_class, field_name)
-    return isinstance(field, (DateFilter, DateTimeFilter, TimeFilter, BooleanFilter)) or field_name == "q"
+    return not isinstance(field, django_filters.MultipleChoiceFilter) or field_name == "q"
```
glennmatthews, line 937: May not need the special case for `field_name == "q"` any more, since SearchFilter is not a MultipleChoiceFilter?

glennmatthews (approved):

Issue #2875, opened 2022-11-17
NaturalKeyOrPKMultipleChoiceFilter should support filtering using a combination of fields (true natural key)


HanlinMiao:
### As ...

Dan - Data Owner

### I want ...

To be able to filter data based on the model's true Natural Keys and the current implmentation of `NaturalKeyOrPKMultipleChoiceFilter` does not support that.

A scenario:

1. Create some powerpanels with the same name and different sites
2. Try to filter in SiteListView with the filter `Power Panel(name or ID)`
3. The `value` column only allows you to enter one value.
4. All the sites with the power panel of that name is returned.
5. If all the sites have a power panel with the same name, this filter is essentially useless.

<img width="1348" alt="Screen Shot 2022-11-17 at 12 55 25 PM" src="https://user-images.githubusercontent.com/46973263/202521382-85d5ac74-3e9d-40c7-9209-be09782c967e.png">

The real issue is:
Name for `PowerPanel` model is not a true NaturalKey, (name, site) is.
`NaturalKeyOrPKMultipleChoiceFilter` should support filtering using true uniqueness constraint (e.g. a combination of fields) and not just `to_field_name` which allows only one field to be specified as Natural Key. (in some cases, one model field is not enough to identify the natural keys)

### So that ...

I can accurately filter my data and get unique data back.

### I know this is done when...

`NaturalKeyOrMultipleChoiceFilter` supports filtering based on True Natural Keys. (In some cases a combination of fields)

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [X] Documentation
- [ ] Other (not directly a platform feature)

glennmatthews:
Probably this should support input of the `natural_key_slug` property (added in #3423) as a filter value.

glennmatthews:
I have some local code that was working pretty well towards being able to use `composite_key` as a filter string, but given #4346, we'll have to come up with an alternate approach.

PR #2882, opened 2022-11-18
Replace NaturalKeyOrPKMultipleChoiceFilter with ModelMultipleChoiceFilter where appropriate

HanlinMiao:
# What's Changed

# TODO

- [ ] Unit, Integration Tests


nautobot/dcim/tests/test_filters.py
```diff
@@ -3550,22 +3550,18 @@ def test_description(self):
...
 
     def test_lag(self):
         lag_interfaces = Interface.objects.filter(type=InterfaceTypeChoices.TYPE_LAG)[:2]
-        with self.subTest():
-            params = {"lag_id": [lag_interfaces[0].pk, lag_interfaces[1].pk]}
-            self.assertEqual(self.filterset(params, self.queryset).qs.count(), 2)
-        with self.subTest():
-            params = {"lag": [lag_interfaces[0].pk, lag_interfaces[1].name]}
-            self.assertEqual(self.filterset(params, self.queryset).qs.count(), 2)
+        params = {"lag_id": [lag_interfaces[0].pk, lag_interfaces[1].pk]}
```
glennmatthews, line 3563: Same comment here, seems like it should be `lag` that we keep.

nautobot/dcim/filters.py
```diff
@@ -1471,15 +1486,12 @@ class InventoryItemFilterSet(BaseFilterSet, DeviceComponentFilterSet):
         to_field_name="name",
         label="Device (name)",
     )
+    # TODO: change this filter to NaturalKeyOrPKMultipleChoiceFilter once
+    # https://github.com/nautobot/nautobot/issues/2875 is resolved.
     parent_id = django_filters.ModelMultipleChoiceFilter(
         queryset=InventoryItem.objects.all(),
         label="Parent inventory item (ID)",
     )
-    parent = NaturalKeyOrPKMultipleChoiceFilter(
```
glennmatthews, line 1478: Shouldn't we remove `parent_id` and keep `parent`?

nautobot/dcim/filters.py
```diff
@@ -1262,26 +1275,25 @@ class InterfaceFilterSet(
...
+        label="Bridge interface (ID)",
     )
+    # TODO: change this filter to NaturalKeyOrPKMultipleChoiceFilter once
+    # https://github.com/nautobot/nautobot/issues/2875 is resolved.
     lag_id = django_filters.ModelMultipleChoiceFilter(
         field_name="lag",
         queryset=Interface.objects.filter(type=InterfaceTypeChoices.TYPE_LAG),
         label="LAG interface (ID)",
     )
-    lag = NaturalKeyOrPKMultipleChoiceFilter(
```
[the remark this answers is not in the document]
HanlinMiao: I dont want to blanketly say that it was implemented incorrectly, but it is definitely not filtering on the model's true Natural Keys. So `NaturalKeyOrPKMultipleChoiceFilter` is kind of a misnomer in that sense but it is functionally correct. 
e.g. `device's` `name` attribute is not a sole unique identifier of the `device` but it is used as if it was a Natural Key in a lot of the filters; however, the filter is filtering on the name of the devices and returning the correct results.

nautobot/dcim/filters.py
```diff
@@ -184,10 +184,11 @@ class SiteFilterSet(NautobotFilterSet, TenancyFilterSet, StatusModelFilterSetMix
...
     )
-    devices = NaturalKeyOrPKMultipleChoiceFilter(
+    # TODO: change this filter to NaturalKeyOrPKMultipleChoiceFilter once
+    # https://github.com/nautobot/nautobot/issues/2875 is resolved.
+    devices = django_filters.ModelMultipleChoiceFilter(
         queryset=Device.objects.all(),
-        to_field_name="name",
-        label="Devices (name or ID)",
+        label="Devices (ID)",
     )
```
glennmatthews, line 192: A lot of these are arguably a loss of functionality, since `device.name`, `rack.name`, etc. are "locally unique" fields and it's probably pretty useful to be able to filter on those even if they're not "true" natural keys. I get that this change is more correct from a code perspective but I'm still a bit unhappy about it.

HanlinMiao: Yeah I understand, let me ask the team for their opinion on this.

bryanculver: Let's keep the semi-incorrect behavior in for now with the `TODO`. We're going to be cleaning up natural keys in 2.0 along with the fact that 2804 not going in until 2.0 and there would be a strong desire to fix #2875 in 2.0 anyway.

jathanism: Thrived with what @bryanculver said. This isn't going to be finalized so for now just keeping it working while we keep moving along towards the fix in #2875 is the right path forward.

PR #2911, opened 2022-11-25
Role model data migrations, UI and API related changes

timizuoebideri1:
# Closes: #1063 
# What's Changed

# TODO
- Replace Prefix, Device, VM, IPAddress, Racks role field with new Role field and create data migration.
- These models' UI and API-related codes have been updated to function with our new Role model.

- [ ] Unit, Integration Tests


nautobot/extras/migrations/0055_collect_roles_from_related_apps_roles.py
```diff
@@ -0,0 +1,79 @@
+# Generated by Django 3.2.16 on 2022-11-19 23:15
+
+from collections import namedtuple
+from django.db import migrations, models
+
+from nautobot.ipam.choices import IPAddressRoleChoices
```
timizuoebideri1, line 6: NOTE: This would be removed when removing obsolete code; either hardcode '`IPAddressRoleChoices`' choices here or do not remove '`IPAddressRoleChoices`'. Because its been used in data migration below,

nautobot/dcim/migrations/0020_alter_device_and_rack_role_add_new_role.py
```diff
@@ -0,0 +1,62 @@
...
+        migrations.AlterField(
+            model_name="device",
+            name="device_role",
+            field=models.ForeignKey(
+                on_delete=django.db.models.deletion.PROTECT,
+                related_name="devices",
+                to="dcim.devicerole",
+                null=True,
+            ),
+        ),
```
glennmatthews, line 28: What's changing here? I notice no corresponding AlterField on the Rack `role` field below.

nautobot/dcim/models/devices.py
```diff
@@ -578,7 +577,8 @@ class Device(PrimaryModel, ConfigContextModel, StatusModel):
 
     csv_headers = [
         "name",
-        "device_role",
+        # TODO(timizuo): Device Role Reassign
+        # "device_role",
```
glennmatthews, line 581: This just becomes `"role"` right? Same below?

nautobot/dcim/models/racks.py
```diff
@@ -456,7 +448,9 @@ def get_rack_units(
             # Retrieve all devices installed within the rack
             queryset = (
                 # v2 TODO(jathan): Replace prefetch_related with select_related
-                Device.objects.prefetch_related("device_type", "device_type__manufacturer", "device_role")
+                # TODO(timizuo): Device Role Reassign
+                # Device.objects.prefetch_related("device_type", "device_type__manufacturer", "device_role")
+                Device.objects.prefetch_related("device_type", "device_type__manufacturer")
```
glennmatthews, line 453: Perhaps:

```suggestion
                Device.objects.prefetch_related("device_type", "device_type__manufacturer", "role")
```

nautobot/extras/migrations/0055_collect_roles_from_related_apps_roles.py
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    for role_model_or_choices in related_role_models_or_choices:
+        try:
+            app_name, model_class = role_model_or_choices.model_or_choices.split(".")
+            related_role_model_class = apps.get_model(app_name, model_class)
+            existing_roles = Role.objects.values_list("name", flat=True)
+            roles_to_create = related_role_model_class.objects.filter(~models.Q(name__in=existing_roles))
+        except AttributeError:
+            # An AttributeError would occur when trying to perform a split on a choice e.g. IPAddressRoleChoices
```
glennmatthews, line 47: This feels inelegant to me. Instead of trying to handle the IPAddressRoleChoices in the same `for` loop as the actual Role models, would it be cleaner to handle it separately (perhaps with a common helper function to provide shared logic between the two cases)?

nautobot/extras/migrations/0055_collect_roles_from_related_apps_roles.py
```diff
@@ -0,0 +1,90 @@
...
+        RelatedRoleModelOrChoices("ipam.Role", ["ipam.rack", "ipam.vlan"]),
+        RelatedRoleModelOrChoices(IPAddressRoleChoices, ["ipam.ipaddress"]),
+    )
+
+    color_map = {
+        "default": "9e9e9e",  # Grey
+        "primary": "2196f3",  # Blue
+        "warning": "ffc107",  # Amber
+        "success": "4caf50",  # Green
+    }
```
glennmatthews, line 38: Should be able to use values from `nautobot.utilities.choices.ColorChoices` here.

nautobot/extras/migrations/0055_collect_roles_from_related_apps_roles.py
```diff
@@ -0,0 +1,90 @@
...
+    """Populate Role models using records from other related role models or choices from different apps."""
+
+    Role = apps.get_model("extras", "Role")
+    ContentType = apps.get_model("contenttypes", "ContentType")
+
+    related_role_models_or_choices = (
+        RelatedRoleModelOrChoices("dcim.DeviceRole", ["dcim.device", "virtualization.virtualmachine"]),
+        RelatedRoleModelOrChoices("dcim.RackRole", ["dcim.rack"]),
+        RelatedRoleModelOrChoices("ipam.Role", ["ipam.rack", "ipam.vlan"]),
+        RelatedRoleModelOrChoices("ipam.Role", ["ipam.rack", "ipam.vlan"]),
```
glennmatthews, line 29: Why twice?

nautobot/extras/migrations/0055_collect_roles_from_related_apps_roles.py
```diff
@@ -0,0 +1,90 @@
...
+            Role(
+                name=data.name,
+                slug=data.slug,
+                description=data.description,
+                color=data.color if hasattr(data, "color") else color_map["default"],
+                weight=data.weight if hasattr(data, "weight") else None,
+            )
+            for data in roles_to_create
+        ]
+        role_instances = Role.objects.bulk_create(roles, batch_size=1000)
```
glennmatthews, line 65: Won't this fail if we have any overlap between the role records, e.g. a `DeviceRole` with slug `router` and an ipam `Role` with slug `router` too?

nautobot/extras/migrations/0055_collect_roles_from_related_apps_roles.py
```diff
@@ -0,0 +1,90 @@
...
+
+        filter_ct_by = models.Q()
+        for app_and_model in role_model_or_choices.implemented_by:
+            app_label, model_name = app_and_model.split(".")
+            filter_ct_by |= models.Q(app_label=app_label, model=model_name)
+
+        content_types = ContentType.objects.filter(filter_ct_by)
+        for role in role_instances:
+            role.content_types.set(content_types)
```
glennmatthews, line 74: Similarly, if we have overlapping DeviceRole/RackRole/Role records, won't the last content-type encountered overwrite the previous content_types? I'd think we'd want an `add` here rather than a `set`?

nautobot/extras/utils.py
```diff
@@ -399,3 +399,49 @@ def refresh_job_model_from_job_class(job_model_class, job_source, job_class, *,
...
+        model(Model): Model with role field to alter
+        role_model(Model): Role Model (optional)
+        is_choice_field(bool): True if Model's role field is a choice field not an FK field else False
+        role_choiceset(ChoiceSet): Role ChoiceSet (optional)
+        legacy_role(str): Models role field legacy name; This is the current role field name
+        new_role(str): Models role field new name; This is the name role field should be updated to
+    """
+
+    legacy_role_name = legacy_role if is_choice_field else legacy_role + "__name"
+    queryset = model.objects.filter(**{legacy_role + "__isnull": False}).values("pk", legacy_role_name)
```
glennmatthews, line 425: Is there actually a case with null `legacy_role` that we need to handle? Or is this just overly cautious?

nautobot/ipam/migrations/0010_migrate_ipam_role_data.py
```diff
@@ -0,0 +1,55 @@
...
+            role_model = apps.get_model("ipam", value["role_model"])
+        else:
+            role_choiceset = value["role_choiceset"]
+        migrate_role_data(
+            model=model,
+            role_model=role_model,
+            role_choiceset=role_choiceset,
+            legacy_role="new_role",
+            new_role="legacy_role",
+        )
```
glennmatthews, line 44: On these two also, I personally feel it'd be cleaner to just handle IPAddress roles differently from the ipam.Role model rather than these somewhat confusing data structures.

nautobot/extras/migrations/0055_collect_roles_from_related_apps_roles.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def populate_roles_from_related_app_roles(apps, schema_editor):
+    """Populate Role models using records from other related role models or choices from different apps."""
+
+    Role = apps.get_model("extras", "Role")
+    ContentType = apps.get_model("contenttypes", "ContentType")
+
+    related_role_models_or_choices = (
+        RelatedRoleModelOrChoices("dcim.DeviceRole", ["dcim.device", "virtualization.virtualmachine"]),
```
glennmatthews, line 26: The migration of DeviceRole should refer to the `vm_role` flag on each record to determine whether that particular Role should be assigned to `virtualization.virtualmachine`.

nautobot/dcim/migrations/0020_alter_device_and_rack_role_add_new_role.py
timizuoebideri1: I had to set `device_role` to allow `null` cause when reversing migration i get this error "legacy_role cannot have null values"

nautobot/extras/utils.py
timizuoebideri1: Yr apart from `Device` other model role fields allow null

nautobot/utilities/templatetags/helpers.py
```diff
@@ -681,3 +681,16 @@ def custom_branding_or_static(branding_asset, static_asset):
...
+
+@library.filter()
+@register.filter()
+def render_obj_display(model, field_name):
+    """Render the display view of an object."""
+    obj = getattr(model, field_name)
+    if obj:
+        color = getattr(obj, "color", None)
+        display = getattr(model, f"get_{field_name}_display")
+        content = f'<span class="label" style="color: {fgcolor(color)}; background-color: #{color}">{display()}</span>'
```
glennmatthews, line 694: See my [comment](https://github.com/nautobot/nautobot/pull/2822#discussion_r1048841424) on #2822 - we shouldn't need to use `get_{field_name}_display`, should just be able to do `getattr(obj, "name")`

nautobot/utilities/templatetags/helpers.py
```diff
@@ -681,3 +681,16 @@ def custom_branding_or_static(branding_asset, static_asset):
...
+        return mark_safe(content)
```
glennmatthews, line 695: Please test this against Status/Role objects that have a `name` containing HTML-unsafe characters like `<` and `>` - I suspect we have an issue here. Probably should be using `format_html()` here.

nautobot/circuits/templates/circuits/circuit_retrieve.html
```diff
@@ -14,7 +14,7 @@
                 <tr>
                     <td>Status</td>
                     <td>
-                        <span class="label" style="color: {{ object.status.color|fgcolor }}; background-color: #{{ object.status.color }}">{{ object.get_status_display }}</span>
+                        {{ object | render_obj_display:"status" }}
```
glennmatthews, line 17: Instead of `render_obj_display` perhaps something like `render_related_name_and_color` would be clearer?

nautobot/dcim/api/serializers.py
```diff
@@ -705,8 +707,6 @@ class Meta:
 
 class DeviceRoleSerializer(NautobotModelSerializer):
     url = serializers.HyperlinkedIdentityField(view_name="dcim-api:devicerole-detail")
-    device_count = serializers.IntegerField(read_only=True)
-    virtualmachine_count = serializers.IntegerField(read_only=True)
```
glennmatthews, line 709: Is this a removal of existing functionality? Should be documented in the changelog and in the migration docs if so.

...Actually, shouldn't this serializer go away entirely with the removal of `DeviceRole` as a distinct model?

nautobot/dcim/api/serializers.py
```diff
@@ -769,7 +768,7 @@ class Meta:
             "url",
             "name",
             "device_type",
-            "device_role",
+            "role",
```
glennmatthews, line 771: This change needs to be documented in the v2 migration docs.

nautobot/dcim/filters.py
```diff
@@ -436,10 +437,18 @@ class RackRoleFilterSet(NautobotFilterSet, NameSlugSearchFilterSet):
 
     class Meta:
         model = RackRole
-        fields = ["id", "name", "slug", "color", "description", "racks"]
+        # TODO (timizuo): Racks Role
+        # fields = ["id", "name", "slug", "color", "description", "racks"]
```
glennmatthews, line 441: What needs to be done here?

nautobot/dcim/filters.py
```diff
@@ -878,18 +881,9 @@ class Meta:
 
 
 class DeviceRoleFilterSet(NautobotFilterSet, NameSlugSearchFilterSet):
```
glennmatthews, line 883: Shouldn't this entire class go away if we're removing `DeviceRole` as a model?

nautobot/dcim/forms.py
```diff
@@ -130,6 +130,7 @@
     Site,
     VirtualChassis,
 )
+from ..extras.forms.mixins import RoleModelBulkEditFormMixin, RoleModelCSVFormMixin, RoleModelFilterFormMixin
```
glennmatthews, line 133:
```suggestion
from nautobot.extras.forms.mixins import RoleModelBulkEditFormMixin, RoleModelCSVFormMixin, RoleModelFilterFormMixin
```

nautobot/dcim/tests/test_cablepaths.py
```diff
@@ -44,12 +43,13 @@ def setUpTestData(cls):
 
         manufacturer = Manufacturer.objects.create(name="Generic", slug="generic")
         device_type = DeviceType.objects.create(manufacturer=manufacturer, model="Test Device")
-        device_role = DeviceRole.objects.create(name="Device Role", slug="device-role")
+        # TODO(timizuo): Device Role Reassign
```
glennmatthews, line 46: What work is needed here?

nautobot/dcim/views.py
```diff
@@ -1269,12 +1269,12 @@ def get_extra_context(self, request, instance):
         # v2 TODO(jathan): Replace prefetch_related with select_related
         devices = (
             Device.objects.restrict(request.user, "view")
-            .filter(device_role=instance)
+            .all()
```
glennmatthews, line 1272: seems wrong to me, though I presume this entire view is going away soon anyway?

nautobot/dcim/views.py
```diff
@@ -1269,12 +1269,12 @@ def get_extra_context(self, request, instance):
...
             .prefetch_related("status", "site", "tenant", "rack", "device_type")
         )
 
         device_table = tables.DeviceTable(devices)
-        device_table.columns.hide("device_role")
+        device_table.columns.hide()
```
glennmatthews, line 1277: also seems wrong - will this hide all columns, or is it an error?

nautobot/extras/filters.py
```diff
@@ -1153,6 +1154,31 @@ class Meta:
         ]
 
 
+class RoleFilter(NaturalKeyOrPKMultipleChoiceFilter):
```
glennmatthews, line 1157: Is this going to be reused somewhere else, or why does it need to be a custom class instead of something inline within `RoleModelFilterSetMixin`?

nautobot/extras/models/models.py
```diff
@@ -103,7 +103,8 @@ class ConfigContext(BaseModel, ChangeLoggedModel, ConfigContextSchemaValidationM
     regions = models.ManyToManyField(to="dcim.Region", related_name="+", blank=True)
     sites = models.ManyToManyField(to="dcim.Site", related_name="+", blank=True)
     locations = models.ManyToManyField(to="dcim.Location", related_name="+", blank=True)
-    roles = models.ManyToManyField(to="dcim.DeviceRole", related_name="+", blank=True)
+    # TODO(timizuo): Figure out if roles is still meant to be here.
+    roles = models.ManyToManyField(to="extras.Role", related_name="+", blank=True, null=True)
```
glennmatthews, line 107: Should still be needed, else we're removing functionality as far as assigning a ConfigContext to apply to only specific device role(s).

nautobot/extras/models/roles.py
```diff
@@ -45,6 +45,8 @@ class RoleModelMixin(models.Model):
     role = RoleField(
         on_delete=models.PROTECT,
         related_name="%(app_label)s_%(class)s_related",  # e.g. dcim_device_related
+        blank=True,
+        null=True,
```
glennmatthews, line 49: Hmm, we might need two different mixins - IIRC there are some models for which a role was mandatory, and others for which it was optional?

nautobot/extras/api/serializers.py
```diff
@@ -142,6 +144,16 @@ class NautobotModelSerializer(
...
 
 
+class RoleSerializerField(LimitQuerysetChoicesSerializerMixin, NestedRoleSerializer):
+    """NestedSerializer field for `Role` object fields."""
+
+
+class RoleModelSerializerMixin(BaseModelSerializer):
+    """Mixin to add `role` choice field to model serializers."""
+
+    role = RoleSerializerField(required=False)
```
glennmatthews, line 154: Do we need separate mixins for models for which role is required versus models that permit a null role?

nautobot/dcim/api/serializers.py
timizuoebideri1: Every thing relating to `DeviceRole` and `ipam.Role` would all be removed

nautobot/dcim/tests/test_cablepaths.py
timizuoebideri1: It would be replaced with Role, in the next PR where i write test

nautobot/dcim/views.py
timizuoebideri1: `DeviceRoleView` would be removed, so this change is irrelevant, Had to make this change cause it was causing errors

nautobot/extras/filters.py
timizuoebideri1: on line 1179, I altered the queryset  by applying `.get_for_model(self.model)`, Im not sure there is any-other way to get the `self.model`

nautobot/docs/additional-features/jobs.md
```diff
@@ -753,7 +753,7 @@ These variables are presented as a web form to be completed by the user. Once su
 ```python
 from django.utils.text import slugify
 
-from nautobot.dcim.models import Device, DeviceRole, DeviceType, Manufacturer, Site
+from nautobot.dcim.models import Device, DeviceType, Manufacturer, Site
 from nautobot.extras.models import Status
 from nautobot.extras.jobs import *
```
glennmatthews, line 758: Also need to replace `DeviceRole` with `Role` later in this example Job.

nautobot/extras/api/serializers.py
```diff
@@ -36,6 +35,7 @@
     ObjectChangeActionChoices,
 )
 from nautobot.extras.datasources import get_datasource_content_choices
+from ...core.models.mixins import LimitQuerysetChoicesSerializerMixin
```
glennmatthews, line 38:
```suggestion
from nautobot.core.models.mixins import LimitQuerysetChoicesSerializerMixin
```

nautobot/extras/api/serializers.py
```diff
@@ -113,6 +113,7 @@
     NestedTagSerializer,
     NestedWebhookSerializer,
     NestedJobHookSerializer,
+    NestedRoleSerializer,
```
glennmatthews, line 116: Can you move these two to their appropriate alphabetized position in the list please?

nautobot/extras/filters/__init__.py
```diff
@@ -217,12 +218,12 @@ class ConfigContextFilterSet(BaseFilterSet):
...
+        queryset=Role.objects.all(),
         label="Role",
     )
     role = django_filters.ModelMultipleChoiceFilter(
         field_name="roles__slug",
-        queryset=DeviceRole.objects.all(),
+        queryset=Role.objects.all(),
         to_field_name="slug",
         label="Role (slug)",
     )
```
glennmatthews, line 229: Possibly collapse these to a `NaturalKeyOrPKMultipleChoiceFilter` while you're touching this? Or is that out of scope?

nautobot/extras/forms/mixins.py
```diff
@@ -687,6 +691,43 @@ def _append_relationships_side(self, relationships, initial_side):
...
+    Mixin to add non-required `role` multiple-choice field to filter forms.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["role"] = DynamicModelMultipleChoiceField(
+            required=False,
+            queryset=Role.objects.all(),
+            query_params={"content_types": self.model._meta.label_lower},
+            to_field_name="name",  # TODO(timizuo): Might be slug but i dont think so
```
glennmatthews, line 718: At the very least it appears to be inconsistent with `StatusModelFilterFormMixin` -- intentional?

nautobot/extras/models/models.py
```diff
@@ -102,7 +102,12 @@ class ConfigContext(BaseModel, ChangeLoggedModel, ConfigContextSchemaValidationM
     regions = models.ManyToManyField(to="dcim.Region", related_name="+", blank=True)
     sites = models.ManyToManyField(to="dcim.Site", related_name="+", blank=True)
     locations = models.ManyToManyField(to="dcim.Location", related_name="+", blank=True)
-    roles = models.ManyToManyField(to="dcim.DeviceRole", related_name="+", blank=True)
+    roles = models.ManyToManyField(
+        to="extras.Role",
+        related_name="+",
+        blank=True,
+        limit_choices_to=RoleModelsQuery(),
```
glennmatthews, line 109: Maybe I'm missing something - I don't see why `RoleModelsQuery` applies here. That returns a list of ContentTypes, not a list of Roles?

nautobot/extras/querysets.py
```diff
@@ -14,8 +14,8 @@ def get_for_object(self, obj):
         Return all applicable ConfigContexts for a given object. Only active ConfigContexts will be included.
         """
 
-        # `device_role` for Device; `role` for VirtualMachine
-        role = getattr(obj, "device_role", None) or obj.role
+        # `role` for Device; `role` for VirtualMachine
```
glennmatthews, line 17: probably can just remove this comment :)

nautobot/extras/tests/integration/__init__.py
```diff
@@ -1,12 +1,12 @@
...
+from nautobot.dcim.models import Device, DeviceType, Manufacturer, Site
 
 
 def create_test_device():
     test_uuid = str(uuid.uuid4())
-    device_role = DeviceRole.objects.create(
+    device_role = Role.objects.create(
         name=f"Test Role {test_uuid}",
         slug=f"test-role-{test_uuid}",
     )
```
glennmatthews, line 12: probably need to set `device_role.content_types`?

nautobot/dcim/factory.py
```diff
@@ -121,20 +119,6 @@ class Params:
     comments = factory.Maybe("has_comments", factory.Faker("paragraph"), "")
 
 
-class DeviceRoleFactory(OrganizationalModelFactory):
```
glennmatthews, line 124: Need to create/enable a new `nautobot.extras.factory.RoleFactory` equivalent.

nautobot/ipam/factory.py
```diff
@@ -362,7 +346,6 @@ class Meta:
     )
 
     has_role = factory.Faker("pybool")
-    role = factory.Maybe("has_role", random_instance(Role), None)
```
glennmatthews, line 365: this needs to change to use Roles from the extras factory, instead of being removed.

nautobot/ipam/factory.py
```diff
@@ -439,7 +422,6 @@ class Params:
     location = factory.Maybe(
         "has_location", random_instance(lambda: Location.objects.get_for_model(Prefix), allow_null=False), None
     )
-    role = factory.Maybe("has_role", random_instance(Role), None)
```
glennmatthews, line 442: this needs to change to use Roles from the extras factory, instead of being removed.

nautobot/ipam/tables.py
```diff
@@ -44,25 +42,6 @@
...
-PREFIX_COPY_LINK = """
-{% load helpers %}
-{% for i in record.parents|as_range %}
-    <i class="mdi mdi-circle-small"></i>
-{% endfor %}
-<span class="hover_copy"><a href="{% if record.present_in_database %}{% url 'ipam:prefix' pk=record.pk %}{% else %}{% url 'ipam:prefix_add' %}?prefix={{ record }}{% if object.vrf %}&vrf={{ object.vrf.pk }}{% endif %}{% if object.site %}&site={{ object.site.pk }}{% endif %}{% if object.tenant %}&tenant_group={{ object.tenant.group.pk }}&tenant={{ object.tenant.pk }}{% endif %}{% endif %}" id="copy_{{record.id}}">{{ record.prefix }}</a><button type="button" class="btn btn-inline btn-default hover_copy_button" data-clipboard-target="#copy_{{record.id}}">
-                                <span class="mdi mdi-content-copy"></span>
-                            </button>
-                        </span>
-"""
```
glennmatthews, line 56: Why is this removed?

glennmatthews (changes requested):

PR #2922, opened 2022-11-29
[WIP] Update HTTP_ACCEPT header in test without api_version

chadell:
the `set_api_version` is already taking care of having a null api_version, setting HTTP_ACCEPT header to "application/json". This is not a big deal because it's de the default for API calls, but it seems more easy to follow.

# What's Changed

Explicitly using the `set_api_version` even no api_version is defined.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/utilities/testing/api.py
+++ b/nautobot/utilities/testing/api.py
@@ -49,8 +49,7 @@ class APITestCase(ModelTestCase):
         super().setUpNautobot(client=False)
         self.token = Token.objects.create(user=self.user)
         self.header = {"HTTP_AUTHORIZATION": f"Token {self.token.key}"}
-        if self.api_version:
-            self.set_api_version(self.api_version)
+        self.set_api_version(self.api_version)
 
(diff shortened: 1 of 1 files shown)


chadell:
to investigate why tests fail:
```
FAIL: test_root (nautobot.circuits.tests.test_api.AppTest)
  File "/home/runner/work/nautobot/nautobot/nautobot/circuits/tests/test_api.py", line 15, in test_root
AssertionError: 406 != 200
```

bryanculver (changes requested):
The tests are failing because in the root case the response is not JSON. Please view: http://nautobot.test/api/circuits/?format=api.

`?format=api` is the browsable API response (see: https://www.django-rest-framework.org/topics/browsable-api/#formats)

I agree this change is cleaner but this may need to be rearchitected so the root tests can exist (supplying their own overloaded `Accept` header potentially) or the condition needs to change entirely.

Issue #2924, opened 2022-11-29
Conditional Deprecation Warnings Causing Import/Startup Errors


bryanculver:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.2
* Python version: 3.7+
* Database platform, version: All
* Middleware(s): N/A

### Steps to Reproduce
1. Install plugins into 1.5.2 container using recently deprecated `PluginConfig`
2. Startup

### Expected Behavior
Starts up with deprecation warnings.

### Observed Behavior
Failure to start: https://github.com/nautobot/nautobot-plugin-golden-config/actions/runs/3529187176/jobs/5920005352#step:7:325

PR #2925, opened 2022-11-29
Fix import error

bryanculver:
# Closes: #2924
# What's Changed
- Moves deprecation warning flag check to being environment variable only
- Updates documentation

# TODO

- NA Attached Screenshots, Payload Example
- NA Unit, Integration Tests

### DIFF
--- a/nautobot/utilities/deprecation.py
+++ b/nautobot/utilities/deprecation.py
@@ -5,7 +5,7 @@ import sys
 import traceback
 import warnings
 
-from django.conf import settings
+from nautobot.core.settings import LOG_DEPRECATION_WARNINGS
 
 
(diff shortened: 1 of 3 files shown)


nautobot/utilities/deprecation.py
```diff
@@ -5,7 +5,7 @@
 import traceback
 import warnings
 
-from django.conf import settings
+from nautobot.core.settings import LOG_DEPRECATION_WARNINGS
```
glennmatthews, line 8: We should probably move this out of nautobot.core.settings entirely and just check the environment variable directly instead.

glennmatthews (changes requested):

glennmatthews (approved):

Issue #2936, opened 2022-12-02
POST to `/circuits/circuit-types/<slug>/delete/` returning a 404 in v1.5.3


gsnider2195:
### Environment
* Nautobot version (Docker tag too if applicable): v1.5.3
* Python version: 
* Database platform, version: 
* Middleware(s):

### Steps to Reproduce
1. Create a test circuit type
2. Navigate to the circuit type object detail view
3. Click on delete to try to delete the object
4. Click confirm

### Expected Behavior
Circuit type deletes successfully

### Observed Behavior
Nautobot returns a 404

bryanculver:
This was introduced due to #2473 / #2644 because we "fixed the glitch" of DRF advertising we support multipart and urlencoded forms on the API.

PR #2937, opened 2022-12-02
Restore Delete Function on NautobotUIViewSet

bryanculver:
# Closes: #2936
# What's Changed

- Added `x-www-form-urlencoded` Parser to `NautobotUIViewSet`
    - This is how the web form deletes work
- Added Test Case for `x-www-form-urlencoded` for `DeleteObjectViewTestCase`
    - Previous tests were submitting as `multipart/form-data` which is not how the web UI actually submits the form:
    - image

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/utilities/testing/views.py
+++ b/nautobot/utilities/testing/views.py
@@ -9,6 +9,7 @@ from django.core.files.uploadedfile import SimpleUploadedFile
 from django.test import TestCase as _TestCase, override_settings, tag
 from django.urls import reverse, NoReverseMatch
 from django.utils.text import slugify
+from django.utils.http import urlencode
 
 from nautobot.extras.choices import CustomFieldTypeChoices, RelationshipSideChoices, ObjectChangeActionChoices
 from nautobot.extras.models import ChangeLoggedModel, CustomField, Relationship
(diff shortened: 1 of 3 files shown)


gsnider2195 (approved):

HanlinMiao (approved):

nautobot/utilities/testing/views.py
```diff
@@ -545,6 +546,35 @@ def test_delete_object_with_permission(self):
                 self.assertEqual(len(objectchanges), 1)
                 self.assertEqual(objectchanges[0].action, ObjectChangeActionChoices.ACTION_DELETE)
 
+        @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
+        def test_delete_object_with_permission_and_xwwwformurlencoded(self):
```
HanlinMiao, line 550: Wonder if we should add this to `CircuitTypeTestCase` or any test cases that use `NautobotUIViewSet` Views specifically? Lets wait to see if the test cases for others that use `generic.views` pass.

HanlinMiao: The tests passed so I think we can keep it here

Issue #2663, opened 2022-10-20
Tags on ConfigContext View is inconsistent with DynamicModelMultiChoiceField


nniehoff:
### Environment
* Nautobot version (Docker tag too if applicable): 1.4.7
* Python version: 3.7
* Database platform, version: Postgres
* Middleware(s):

### Steps to Reproduce
1.  Go to https://demo.nautobot.com/extras/config-contexts/add/
2.  Notice the tags field does not use `DynamicModelMultiChoiceField` and the current field choice is confusing

### Expected Behavior

The form should use the `DynamicModelMultiChoiceField` to be consistent with other views.

### Observed Behavior

The Form uses a different choice field and it is not clear how to remove the tags

PR #2961, opened 2022-12-06
Override tags field in ConfigContextForm as DynamicModelMultipleChoiceField

HanlinMiao:
# Closes: #2663 
# What's Changed

So all the `PrimaryModels` have their `tags` field defined as `TaggableManager` from `django-taggit`, whereas ConfigContext is the only one with `tags` defined as `models.ManyToManyField`. I wonder if we should change the tags field to`TaggableManager`. But for now I only override the tags field in the `ConfigContextForm` and it works as expected without a migration involved.

Screen Shot 2022-12-06 at 11 03 18 AM
Screen Shot 2022-12-06 at 11 03 05 AM

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/2663.changed
+++ b/changes/2663.changed
@@ -1,1 +1,1 @@
-Changed `tags` field in ConfigContextForm to `DynamicModelMultipleChoiceField`
+Changed `tags` field in ConfigContextForm to `DynamicModelMultipleChoiceField`.
--- a/changes/2663.changed
+++ b/changes/2663.changed
@@ -1,0 +1,1 @@
+Changed `tags` field in ConfigContextForm to `DynamicModelMultipleChoiceField`
--- a/nautobot/extras/templates/extras/configcontext_edit.html
+++ b/nautobot/extras/templates/extras/configcontext_edit.html
@@ -26,9 +26,9 @@
             {% render_field form.tenant_groups %}
             {% render_field form.tenants %}
             {% render_field form.device_redundancy_groups %}
+            {% render_field form.tags %}
         </div>
     </div>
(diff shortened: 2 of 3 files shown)


glennmatthews:
For context: `ConfigContext.tags` is not *tags applied to a ConfigContext*, it is *tags used to control which Devices/VirtualMachines this ConfigContext applies to*.  So it's a different semantic than the `tags` field on most other models.

nautobot/extras/templates/extras/configcontext_edit.html
```diff
@@ -26,9 +26,9 @@
             {% render_field form.tenant_groups %}
             {% render_field form.tenants %}
             {% render_field form.device_redundancy_groups %}
-            {% render_field form.tags %}
         </div>
     </div>
+    {% include 'inc/extras_features_edit_form_fields.html' %}
     <div class="panel panel-default">
```
glennmatthews, line 32: I think this change is undesirable; again this is not "tags for this config context" it's "this config context applies to devices/vms that have the specified tag(s)". As such, grouping the `tags` field under the `Assignment` panel is correct.

glennmatthews (approved):

changes/2663.changed
```diff
@@ -0,0 +1 @@
+Changed `tags` field in ConfigContextForm to `DynamicModelMultipleChoiceField`
```
glennmatthews, line 1:
```suggestion
Changed `tags` field in ConfigContextForm to `DynamicModelMultipleChoiceField`.
```

jathanism (approved):

Issue #2826, opened 2022-11-14
Spike: Work necessary to migrate models between apps


bryanculver:
Investigate the complexity and feasibility of migrating models between apps and the impact to data and other relationships (including content types and data migrations).

**Outcome:** Tasks identified to migrate a model between apps to be used to determine if we want to continue with #2722 or only extract extras features into core on a case-by-case basis.

bryanculver:
Analysis completed by @glennmatthews in #2977. Going to shelve #2722 for a future major release.

PR #2977, opened 2022-12-09
[WIP] Migrate DynamicGroup model and functionality from extras to core

glennmatthews:
# Closes: #2826 
# What's Changed

- Move the `DynamicGroup` model and associated classes (filters, forms, views, API, etc.) from the `extras` app to the `core` app.
- Create migrations (based on example from https://davit.tech/django-move-model) to move the existing database tables and content-types between apps instead of deleting the existing tables and creating new empty tables.
    - Migrating the content-type isn't covered in the linked blog post but it's necessary to ensure that GFKs such as in ObjectChange don't get broken.
    - I have manually tested migration back-and-forth between this branch and next with a DynamicGroup with a change history; I haven't tested the migrations against an empty table but CI should prove that out one way or the other.
- To get URL patterns to work correctly, I had to split `nautobot.core.urls` into `nautobot.core.urls` (app-level URLs, such as now dynamic-groups) and `nautobot.core.urls_root` (the base urlconf for all of Nautobot), and similarly split `nautobot.core.api.urls` into `nautobot.core.api.urls` and `nautobot.core.api.urls_root`.
- To avoid circular imports between `extras` and `core` after the move, I had to:
    - convert `nautobot.core.api.views` and `nautobot.core.views` into submodules
    - move the setup of the dynamic-group signal handlers into the AppConfig's `ready()` method instead of connecting them at import time
- One lingering issue I haven't been able to solve for is adding any nav menu items in `nautobot.core.navigation`. Some issue with application loading and initialization order - the URL resolution for the nav menu items causes `nautobot.core.admin` to be imported and evaluated early, apparently too early? and so app startup fails.

# TODO
- [ ] Unit, Integration Tests


nautobot/core/api/__init__.py
```diff
@@ -5,11 +5,11 @@
     TimeZoneSerializerField,
 )
 from .routers import OrderedDefaultRouter
+from .nested_serializers import WritableNestedSerializer
 from .serializers import (
     BaseModelSerializer,
     BulkOperationSerializer,
     ValidatedModelSerializer,
-    WritableNestedSerializer,
```
glennmatthews, line 12: Disregard changes to this file - artifact of some early attempts at refactoring that turned out to be unnecessary.

nautobot/core/api/serializers/__init__.py
```diff
@@ -139,37 +139,6 @@ def get_field_names(self, declared_fields, info):
         return fields
 
 
-class ValidatedModelSerializer(BaseModelSerializer):
```
glennmatthews, line 142: Disregard changes to this file (other than converting it into a submodule) - the relocation of `ValidatedModelSerializer` within this file are leftovers from some early refactoring that proved unnecessary.

nautobot/core/navigation.py
```diff
@@ -0,0 +1,34 @@
+from nautobot.core.apps import NavMenuAddButton, NavMenuGroup, NavMenuItem, NavMenuTab
+
+
+# TODO fix initialization order issue with admin site
+_menu_items = (
```
glennmatthews, line 5: As noted in the PR description, setting `menu_items` here causes an order-of-initialization problem with `nautobot.core.admin`. TODO.

nautobot/core/api/urls.py
```diff
@@ -1,57 +1,29 @@
...
-    StatusView,
-    NautobotSpectacularSwaggerView,
-    NautobotSpectacularRedocView,
-)
-from nautobot.extras.plugins.urls import plugin_api_patterns
+from nautobot.core.api.views.dynamic_groups import DynamicGroupViewSet, DynamicGroupMembershipViewSet
+from . import routers, views
 
+router = routers.OrderedDefaultRouter()
+# router.APIRootView = views.CoreRootView
```
glennmatthews, line 7: Uncommenting this line (following a similar pattern for other Nautobot apps' `api.urls` code) results in an error from inside Django. I haven't chased it down.

glennmatthews:
OK, that's a bizarre error from unit test initialization. Any ideas? Will investigate on Monday.

glennmatthews:
Seems to be an issue with the test-only model defined in [`nautobot/core/tests/test_models.py`](https://github.com/nautobot/nautobot/blob/develop/nautobot/core/tests/test_models.py) not getting a table created in the database. Not sure why my changes are influencing anything here; but based on some googling it perhaps shouldn't be working in the first place as written? Could use some extra eyes here.

glennmatthews:
One extra concern not directly addressed by this PR is how this change would be handled by apps (plugins) that have foreign-keys to any of the Extras models, such as to in this case a DynamicGroup. Probably the data migration needs to inspect the migrated model for any and all reverse-relations that it contains, and handle all of those identified foreign-keys at once.

bryanculver:
😢 

At the moment we're going to shelve moving extras for the time being (maybe 3.0) due to the significant impact to migration for 2.0. Let's let the Plugin API be more widely adopted and clean up some other things first before we move extras.

PR #2991, opened 2022-12-13
Implements initial database backend for Celery task results.

jathanism:
# Closes: #2993
# What's Changed
- Added `django-celery-results` as a core dependency
- Added `django_celery_results` to `INSTALLED_APPS` in `nautobot.core.settings`
- Implemented a subclass of `django_celery_results.backends.DatabaseBackend` at `nautobot.core.celery.backends.NautobotDatabaseBackend`
- Changed default value for `CELERY_RESULT_BACKEND` to point to this new backend subclass
- Implemented a subclass of `celery.Task` at `nautobot.core.celery.task.NautobotTask`
- The core `Celery` app changed to use `NautobotTask` as the default task class for all tasks decorated using `@nautobot_task`

As of these updates, the default functionality of Nautobot as it relates to executing Celery tasks and running Jobs and retreiving job results using `JobResult` objects DOES NOT CHANGE.

The `TaskResult` object provided by `django-celery-results` will now be where Celery task tombstones are stored, more tightly integrating our state machine for managing Jobs and JobResults.

These changes merely lay the groundwork for upcoming changes with the following next steps in mind:

- The database backend subclass will be used to implement custom hooks for Celery task state changes
- Effort will be made to merge the functionality of Nautobot's `JobResult` model and the `TaskResult` model from `django-celery-results` to eliminate the state transitions required inside of the `run_job()` generic Celery task
- The task subclass will be used for integrating any work required to be run solely by workers for handling Job exeuction
- We may also need to customize Celery `Request` used for when a worker is running or tracing a task.

# TODO

- [ ] Unit, Integration Tests


nautobot/core/settings.py
```diff
@@ -725,7 +726,9 @@
 CELERY_BROKER_URL = os.getenv("NAUTOBOT_CELERY_BROKER_URL", parse_redis_connection(redis_database=0))
 
 # Celery results backend URL to tell workers where to publish task results
-CELERY_RESULT_BACKEND = os.getenv("NAUTOBOT_CELERY_RESULT_BACKEND", parse_redis_connection(redis_database=0))
+CELERY_RESULT_BACKEND = os.getenv(
+    "NAUTOBOT_CELERY_RESULT_BACKEND", "nautobot.core.celery.backends.NautobotDatabaseBackend"
+)
```
glennmatthews, line 731: Do we still even want to keep this setting configurable by environment variable?

nautobot/core/settings.py
```diff
@@ -391,6 +391,7 @@
     "nautobot.core",
     "django.contrib.admin",  # Must be after `nautobot.core` for template overrides
     "django_celery_beat",  # Must be after `nautobot.core` for template overrides
+    "django_celery_results",  # Must be after `nautobot.core` for template overrides
```
glennmatthews, line 394: Comment seems misleading - does django-celery-results even have any page templates that we are planning to override?

nautobot/core/settings.py
jathanism: Yeah great question. This was provided before due to varying Redis instances. We should just remove it.

jathanism: But we'll also need to update the docs and put it into the release notes/migration guides that this value shall not be overridden anymore due to Jobs overhaul, etc, etc.

bryanculver:
Cleaning up the dependencies PRs in `next` so ~if~ when we get merge conflicts @jathanism for `poetry.lock` I'll take care of them.

gsnider2195 (approved):

PR #2992, opened 2022-12-13
Bump mkdocs-material to 8.5.11

bryanculver:
Collapses #2785 and #2786, and since this is patch bump should go to `develop`.


bryanculver:
Same as #2802. Canceling as duplicate.

PR #2995, opened 2022-12-13
Update lockfile from Poetry 1.3 update.

bryanculver:
#2994 is very busy because Poetry 1.3 just recently rolled out which introduced a new lockfile format: https://github.com/python-poetry/poetry/releases/tag/1.3.0

This should be reviewed to see if it's backward compatible with Poetry 1.2.


jathanism (approved):

glennmatthews (approved):
What's +/- 1500 LOC between friends? :-)

glennmatthews:
Though from the linked PR it does sound like this may not be backwards-compatible with 1.2...? I'm OK with requiring developers to move to 1.3 but we should document this clearly in the docs and release-notes if so.

bryanculver (commented):
https://github.com/nautobot/nautobot/blob/develop/docker/Dockerfile#L134-L141

😢

HanlinMiao (approved):
It seems like 1.2.x did retain the ability to read new format of lock files here https://github.com/python-poetry/poetry/pull/6608. So this bump should be good.

jathanism (approved):

Issue #2999, opened 2022-12-14
WebUI InterfaceBulkCreateForm has not status field


rhinoau:
### Environment
* Nautobot version: 1.5.4
* Python version: 3.8.10
* Database platform, version: postgresql 12+214ubuntu0.1 
* Middleware(s): n/a

### Steps to Reproduce
1. WebUI - select one or more devices from /dcim/devices/
2. Add components -> Interfaces button at bottom, resulting in entry form at /dcim/devices/interfaces/add/
3. Submit completed form

### Expected Behavior
Specified interface to be created on the selected devices.

### Observed Behavior
Error due to status field not being defined:
```
<class 'ValueError'>

'InterfaceBulkCreateForm' has no field named 'status'.

Python version: 3.8.10
Nautobot version: 1.5.4
```

glennmatthews:
Looks to have been an oversight in #1753. Thanks for the report!

PR #1753, opened 2022-05-05
Add status field to Interface

timizuoebideri1:
# Closes: #984 
# What's Changed
Add status field to Interface Model

# Checklist 

- [x] status is a new mandatory field on Interface
- [x] REST API would need to be versioned
- [x] Interface Statuses are: Active, Planned, Maintenance, Failed, and Decommissioned
- [x] Default value for migration, legacy API would be "Active"

### Interface List UI
Screenshot 2022-05-05 at 12 52 15 PM

### Interface Detail UI
Screenshot 2022-05-05 at 12 52 40 PM

### Device Interface List UI
Screenshot 2022-05-05 at 12 53 14 PM

# TODO

- [x] Explanation of Change(s)
- [x] Attached Screenshots, Payload Example
- [x] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example Plugin Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #1505, opened 2022-03-16
Scope Tags to Content Types

timizuoebideri1:
### Fixes: #872

- Scope tags to content types
 - Create PrimaryModelRelatedContentType as an helper class to get ContentType models that are subclasses of PrimaryModel
 
### Features
Create a Tag with a content type
_**NB**: If `content_type` is left blank, Tag would be available to all `content_types`_
Screenshot 2022-03-21 at 12 08 32 PM

Tag constrained to a particular content type cannot be applied to other content types
Tags
Screenshot 2022-03-21 at 12 28 19 PM

(example 1) Site
Screenshot 2022-03-21 at 12 21 57 PM

(example 2) Device
Screenshot 2022-03-21 at 12 24 15 PM

PR #3008, opened 2022-12-15
Add missing fields to component BulkCreateForms and add unit test coverage

glennmatthews:
# Closes: #2999 
# What's Changed

- When bulk-adding Interfaces to a list of Devices, or VMInterfaces to a list of VirtualMachines, the UI form was missing a `status` field; after #1753, this meant that component creation would fail validation. I've added this missing field to both forms.
- Additionally, when bulk-adding Interfaces, the `type` field was incorrectly marked as optional rather than mandatory. I've fixed this.
- All of the DeviceBulkAddComponentForm subclasses were missing a `tags` field (it was removed inadvertently in #1505) so I've added it back in to these forms.
- I also added `mode` as a field on `InterfaceBulkCreateForm` as this field was present on `VMInterfaceBulkCreateForm` but missing here for no apparent reason.

Before:

image

After:

image

Additionally, as far as I can tell, there was **no** existing unit test automation for this set of views/forms; I've added a generic test case to the `DeviceComponentViewTestCase` base class to provide basic success-path coverage for these views.

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/utilities/testing/views.py
+++ b/nautobot/utilities/testing/views.py
@@ -1328,3 +1328,53 @@ class ViewTestCases:
         """
 
         maxDiff = None
+        bulk_add_data = None
+        """Used for bulk-add (distinct from bulk-create) view testing; self.bulk_create_data will be used if unset."""
+
+        @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
+        def test_bulk_add_component(self):
+            """Test bulk-adding this component to devices/virtual-machines."""
+            obj_perm = ObjectPermission(name="Test permission", actions=["add"])
+            obj_perm.save()
+            obj_perm.users.add(self.user)
+            obj_perm.object_types.add(ContentType.objects.get_for_model(self.model))
+
+            initial_count = self._get_queryset().count()
+
+            data = (self.bulk_add_data or self.bulk_create_data).copy()
+
+            # Load the device-bulk-add or virtualmachine-bulk-add form
+            if "device" in data:
+                url = reverse(f"dcim:device_bulk_add_{self.model._meta.model_name}")
+                request = {
+                    "path": url,
+                    "data": post_data({"pk": data["device"]}),
+                }
+            else:
+                url = reverse(f"virtualization:virtualmachine_bulk_add_{self.model._meta.model_name}")
+                request = {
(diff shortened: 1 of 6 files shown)


bryanculver (approved):

nautobot/utilities/testing/views.py
```diff
@@ -1328,3 +1328,53 @@ class DeviceComponentViewTestCase(
...
+            request["data"] = post_data(data)
+            self.assertHttpStatus(self.client.post(**request), 302)
+
+            updated_count = self._get_queryset().count()
+            self.assertEqual(updated_count, initial_count + self.bulk_create_count)
+
+            matching_count = 0
+            for instance in self._get_queryset().all():
+                try:
+                    self.assertInstanceEqual(instance, (self.bulk_add_data or self.bulk_create_data))
```
HanlinMiao, line 1376: could we just do this?
```suggestion
                    self.assertInstanceEqual(instance, data)
```

HanlinMiao (approved):
Can we consolidate `bulk_add_data` and `bulk_create_data` in the future? They seem super similar to me

glennmatthews:
> Can we consolidate `bulk_add_data` and `bulk_create_data` in the future? They seem super similar to me

It's coded so that in the case where they're the same, you can omit `bulk_add_data` and the test will just use `bulk_create_data` in its place.

There are a couple of cases where the set of fields available for devices-bulk-add are different than the set of fields for bulk-create, such as Interface where the devices-bulk-add form doesn't let you specify VLANs (probably because those may differ by device?) but the bulk-create form for a single device does, so for that test, we need to have different form data for the two cases.

nautobot/utilities/testing/views.py
glennmatthews: Not quite. Note that `self.bulk_add_data` or `self.bulk_create_data` have a `device` or `virtual_machine` key (which corresponds to the equivalent field/attribute on the `instance`), while we modified `data` above to match the form's expectations where it instead has a `pk` key and a `_create` key. Good question though!

HanlinMiao: Damn I missed that part. Thank you for the clarification

Issue #2989, opened 2022-12-13
Legacy Filter Display Not Ideal for User Experience


bryanculver:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5
* Python version: 3.7+
* Database platform, version: Postgres
* Middleware(s): N/A

### Steps to Reproduce
1. Visit any model (example Device)
2. Filter using the Default form (example "Model")
3. Click "Apply"

### Expected Behavior
Displayed filters at top of the form display the same as what user can experience in viewing the default form.

### Observed Behavior
UUIDs returned for foreign-key fields.

<img width="330" alt="image" src="https://user-images.githubusercontent.com/31187/207357847-9cc1efb0-766d-44a6-adc2-791126351bdd.png">

Investigate a user-experience improvement for the lifespan of 1.5 to allow a graceful transition to 2.0 (knowing that #1483 is part of 2.0).

PR #3010, opened 2022-12-15
Show Filter Label, `display` for Field Choices for Active Filter Display

bryanculver:
# Closes: #2989
# What's Changed
- Attempt to show a label for a filter field instead of the filter name.
- Attempt to show a user-friendly display of the filter choice.

This:
image

instead of:
image

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/core/views/generic.py
+++ b/nautobot/core/views/generic.py
@@ -257,6 +257,10 @@ class ObjectListView(ObjectPermissionRequiredMixin, View):
 
             def check_for_display(field_name, values):
                 values = values if isinstance(values, (list, tuple)) else [values]
+
+                if not hasattr(filterset_filters, field_name):
+                    return [field_name, values]
+
                 label = filterset_filters[field_name].label
                 if len(values) == 0 or not is_uuid(values[0]):
                     return [label, values]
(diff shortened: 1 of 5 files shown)


nautobot/core/views/generic.py
```diff
@@ -252,9 +255,25 @@ def get(self, request):
...
+
+                label = filterset_filters[field_name].label
+                if len(values) == 0 or not is_uuid(values[0]):
+                    return [label, values]
+                else:
+                    try:
+                        filtered_results = filterset_filters[field_name].queryset.filter(pk__in=values)
+                        new_values = [result.display for result in filtered_results]
+                        return [label, new_values]
+                    except (FieldError, AttributeError):
```
HanlinMiao, line 272: Just curious, which statement could cause a FieldError?

nautobot/core/views/generic.py
```diff
@@ -252,9 +255,25 @@ def get(self, request):
                 )
                 self.queryset = self.queryset.none()
 
+            def check_for_display(field_name, values):
```
HanlinMiao, line 258: Shall we add one for `NautobotUIViewSet's` mixin as well? Maybe this should be a helper method in `utilities` so both places can use it without much redundant code.

HanlinMiao (changes requested):

nautobot/core/views/generic.py
bryanculver: `result.display`, just in case the object doesn't have a `display` property

HanlinMiao: Ahhh i thought that was `AttributeError`

HanlinMiao (approved):
Love the super comprehensive test case!

PR #3011, opened 2022-12-15
[POC] What Changes if We Drop Python 3.7 Today

bryanculver:
# Closes: #DNE
# What's Changed
Commits were done in a consecutive fashion to drop 3.7 support and see what changes. This is a canary check to see which of our dependencies have already posted patch releases that have dropped 3.7 support officially (we are seeing some say they stopped supporting 3.7 but are still publishing them).

- https://github.com/nautobot/nautobot/pull/3011/commits/e735c56ca9045fcb713a0455388991b3c3ee7ead run `poetry update` against our current `pyproject.toml` to eliminate any false positives
  - Seems like Renovate is ignoring some dependencies. Seems to only push for updating our direct dependencies and not (N+1)-order dependencies. Might be good to update these in concert with #2994 
- https://github.com/nautobot/nautobot/pull/3011/commits/102d099895e4c56141a898c03257cbb1f745512a What I changed in our `pyproject.toml` to "drop" Python 3.7 support
- https://github.com/nautobot/nautobot/pull/3011/commits/b0ef58d77dd08684b4a92f4fb52d21d398da21d5 Running `poetry lock --no-update` to see what extraneous entries in our lockfile go away with no longer needing 3.7 sections (mainly dropping pre-3.8 backwards compatibility libraries like typing and `importlib-metadata`
- https://github.com/nautobot/nautobot/pull/3011/commits/57dda53c7ab86e8038a3cf4aa8fe5137dcf49b07 Only `django-health-check` gets a necessary update, which we suspected because of #2994 (we can drop `importlib-metadata` since the only feature we needed exists as a part of stdlib in 3.8+)

Now, there could be many more packages that have gone to a new major/minor release that dropped Python 3.7 support but as a part of policy we don't do non-patch updates unless the needs strikes. If a CVE of a dependency comes in and is only patched in their major/minor release that has dropped 3.7 support then that is a challenge we will have to face.

# TODO

- [NA] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -39,9 +39,6 @@ files = [
     {file = "asgiref-3.5.2.tar.gz", hash = "sha256:4a29362a6acebe09bf1d6640db38c1dc3d9217c68e6f9f6204d72667fc19a424"},
 ]
 
-[package.dependencies]
-typing-extensions = {version = "*", markers = "python_version < \"3.8\""}
-
 [package.extras]
 tests = ["mypy (>=0.800)", "pytest", "pytest-asyncio"]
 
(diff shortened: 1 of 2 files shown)


bryanculver:
Not today satan!

Issue #510, opened 2021-05-30
Django MPTT is no longer maintained - Migrate Existing Models to `django-tree-queries`


lampwins:
### Proposed Changes
The django-mptt project is [no longer maintained](https://github.com/django-mptt/django-mptt#this-project-is-currently-unmaintained) so we need to find a new pattern for our tree use cases (Regions, rack groups, etc). I think we should revisit these use cases related to storage and access patterns to see if we really need the complexity of the MPTT structure, to begin with.

### Justification
The django-mptt project is no longer maintained and as a core dependency, we need to find an alternative path.

### TODO

Migrate
- [ ] RackGroup
- [ ] TenantGroup
- [ ] InventoryItem

_Footnote: `Region` is an MPTT model but should be collapsed into `Location` sometime in this release and either this issue or that one can be the one to "turn the lights off" on MPTT (See: #2517)_

jathanism:
Sounds like we really need to review for what, exactly, we're using MPTT, and how we can centralize the pattern. We already attacked this to some degree w/ the network model overhaul to decouple form Postgres using the MPTT pattern, but we could go deeper and make it more efficient by establishing some kind of base "tree" model that uses an adjacency list (self-referential foreign key) to a `parent`, and then have all the tree methods (`get_ancestors`, `get_children`, `root`, `parent`, etc.).

glennmatthews:
### Requirements

The main requirements, as I understand them are:
1. Reasonably efficient iteration over ancestors (`get_ancestors()`)
    - `nautobot/dcim/templates/dcim/site.html` and others - `for region in object.region.get_ancestors` etc.
2. Reasonably efficient iteration over children and/or descendants (`get_children()`, `get_descendants()`)
    - `nautobot.dcim.signals.handle_rackgroup_site_change` - recursively setting the Site of all child RackGroups when a parent RackGroup changes Site
3. Reasonably efficient tree membership queries (closely related to the previous two requirements)
    - `TreeNodeMultipleChoiceFilter` - "get me all Sites that belong to Region ABC _or any of its descendants_"
    -  `ConfigContextModelQuerySet` - "get me all ConfigContexts that are assigned to Region XYZ _or any of its ancestors_"
4. Reasonably efficient QuerySet annotation such as [`add_related_count`](https://github.com/django-mptt/django-mptt/blob/main/mptt/managers.py#L402)
    - `RegionListView` - "How many Sites belong to each of the following Regions _or any of their descendants_"
[...]

PR #3012, opened 2022-12-15
Remove django-mptt

glennmatthews:
# Closes: #510
# What's Changed

- Remove `django-mptt` dependency.
- Migrate all former MPTT models (`Region`, `RackGroup`, `InventoryItem`, `TenantGroup`) to use `django-tree-queries` instead
    - The serializers for these models used to have a `_depth` field; I've replaced this with `tree_depth` both for clarity and for consistency with the `Location` and `LocationType` serializers, and added a `TreeModelSerializerMixin` class that provides this field.
    - `django-tree-queries` didn't like RackGroup having `ordering = ["site", "name"]` and was throwing database errors; I've changed it to simply `ordering = ("name",)` which I think is perfectly acceptable in any case.
    - To reflect API differences between mptt and tree-queries, I've had to replace calls to `get_ancestors()` and `get_descendants()` with `ancestors()` and `descendants()` correspondingly. Similarly there's no `get_children()` method but we can just use `.children` instead.
- `django-tree-queries` does not have an `add_related_count(..., cumulative=True)` equivalent (https://github.com/matthiask/django-tree-queries/issues/21) so for the time being I've replaced the cases using this annotation with `count_related()` -- the difference here is that `add_related_count(..., cumulative=True)` includes objects associated with the descendants of each object in the count, while `count_related()` only counts the objects that are directly associated to the object. I think this is probably fine, but it *is* a behavior change and therefore worth discussing
- Tests ~are probably broken at this time. It's a WIP.~ should be passing now. :-)
- ~I commented out the MPTT-specific Region filtering logic in `ConfigContextModelQuerySet` for now; that will need to be restored and fixed.~ Fixed

# TODO

- [x] Unit, Integration Tests
- [x] Reach a decision regarding `add_related_count()` removal and plan to proceed
- [x] Fix config-context handling of Regions
- [x] Update `display` of these models for clarity in menu presentation, similar to what was done with Locations and LocationTypes


nautobot/dcim/models/device_components.py
```diff
@@ -88,10 +87,6 @@ def to_objectchange(self, action, **kwargs):
 
         return super().to_objectchange(action, related_object=device, **kwargs)
 
-    @property
-    def parent(self):
-        return getattr(self, "device", None)
-
```
glennmatthews, line 94: Defining `parent` on the `ComponentModel` base class was problematic because it was shadowing the `InventoryItem.parent` DB field. This appears to have been the case for a LONG time - no idea why it didn't cause issues before.

It's really only needed on `CableTermination` subclasses, not on all `ComponentModels`, and fortunately `InventoryItem` isn't a `CableTermination`, so I was able to move it as seen below.

nautobot/dcim/filters/__init__.py
```diff
@@ -1504,13 +1503,13 @@ class InventoryItemFilterSet(BaseFilterSet, DeviceComponentModelFilterSetMixin):
...
-    child_items = NaturalKeyOrPKMultipleChoiceFilter(
+    children = NaturalKeyOrPKMultipleChoiceFilter(
         queryset=InventoryItem.objects.all(),
         to_field_name="name",
         label="Child items (name or ID)",
     )
-    has_child_items = RelatedMembershipBooleanFilter(
-        field_name="child_items",
+    has_children = RelatedMembershipBooleanFilter(
+        field_name="children",
```
glennmatthews, line 1512: Unlike MPTT, django-tree-queries doesn't generally let you change the related-name for the `parent` FK - it has to be `children`.

nautobot/dcim/migrations/0001_initial_part_1.py
```diff
@@ -526,10 +526,10 @@ class Migration(migrations.Migration):
...
                 ("asset_tag", models.CharField(blank=True, max_length=50, null=True, unique=True)),
                 ("discovered", models.BooleanField(default=False)),
-                ("lft", models.PositiveIntegerField(editable=False)),
-                ("rght", models.PositiveIntegerField(editable=False)),
-                ("tree_id", models.PositiveIntegerField(db_index=True, editable=False)),
-                ("level", models.PositiveIntegerField(editable=False)),
+                ("lft", models.PositiveIntegerField(default=0, editable=False)),
+                ("rght", models.PositiveIntegerField(default=0, editable=False)),
+                ("tree_id", models.PositiveIntegerField(default=0, editable=False)),
+                ("level", models.PositiveIntegerField(default=0, editable=False)),
```
glennmatthews, line 532: Addition of defaults here is to make reverse-migration (rolling back the dcim/0021 migration) possible - since MPTT is no longer auto-populating these fields, they needed to have a defined default value else the DB complains about trying to set them to null when doing the reverse migration.

nautobot/dcim/models/device_components.py
```diff
@@ -1019,7 +1037,7 @@ class InventoryItem(MPTTModel, ComponentModel):
     ]
 
     class Meta:
-        ordering = ("device__id", "parent__id", "_name")
+        ordering = ("_name",)
```
glennmatthews, line 1040: django-tree-queries doesn't appear happy when using a compound `ordering` definition, and ordering by IDs (PKs) doesn't make sense for UUID PKs anyway.

nautobot/dcim/templates/dcim/site.html
```diff
@@ -17,11 +17,6 @@
                 <tr>
                     <td>Region</td>
                     <td>
-                        {% if object.region %}
-                            {% for region in object.region.get_ancestors %}
-                                {{ region|hyperlinked_object }} /
-                            {% endfor %}
-                        {% endif %}
```
glennmatthews, line 24: Removed since `region.display` now includes the ancestors of the region, so listing them individually was now redundant.

nautobot/dcim/migrations/0001_initial_part_1.py
bryanculver: While I think it's extremely likely that if you upgrade to 2.0 you won't really be able to roll back, it's good to try.

changes/510.changed
```diff
@@ -0,0 +1,3 @@
+Removed dependency on `django-mptt`, migrating `Region`, `RackGroup`, `TenantGroup`, and `InventoryItem` to use `django-tree-queries` instead. This does change the API for certain tree operations on these models, for example `get_ancestors()` is now `ancestors()` and `get_descendants()` is now `descendants()`.
```
bryanculver, line 1: Should this not partially be `changes/510.removed`?

bryanculver (approved):

changes/510.changed
glennmatthews: Probably so. Good thought.

PR #3019, opened 2022-12-17
Update Select2 Widgets to Know if They Are in a Modal

### LINKED ISSUE #3018: Cannot Search Fields in Advanced Filter Form
### Environment
* Nautobot version (Docker tag too if applicable): 1.5+
* Python version: 3.7+
* Database platform, version: NA
* Middleware(s): Na

### Steps to Reproduce
1. Click the "Filter" button on any list view
2. Click the "Advanced" tab
3. Click a field dropdown
4. Try typing in the search box

### Expected Behavior
I can filter down the fields to filter on.

### Observed Behavior
Typing does nothing.

**Demo:**

https://user-images.githubusercontent.com/31187/208224701-197e044b-39f3-4346-b80e-0cb090b4a277.mov

This is due to a Select2 & Bootstrap modal issue. Certain Select2 fields hidden in modals lose certain functionality (top "Common Problem" of Select2): 
- https://select2.org/troubleshooting/common-problems#select2-does-not-function-properly-when-i-use-it-inside-a-bootst
- https://github.com/select2/select2/issues/600
- See linked issues to ☝️

bryanculver:
# Closes: #3018
# What's Changed
- Add `dropdownParent` argument to Select2 initializers (defaults to null for legacy calls)
- Loop over modals to reinitialize their Select2 widgets to update the dropdownParent property

**Demo:**

https://user-images.githubusercontent.com/31187/208224990-b454b072-c5a3-447d-976a-1e66b8301532.mov

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/3018.fixed
+++ b/changes/3018.fixed
@@ -1,1 +1,1 @@
-Update Select2 widgets to know they are in a modal.
+Fixed rendering of Select2 widgets in modal dialogs.
--- a/changes/3018.fixed
+++ b/changes/3018.fixed
@@ -1,0 +1,1 @@
+Update Select2 widgets to know they are in a modal.
--- a/nautobot/project-static/js/forms.js
+++ b/nautobot/project-static/js/forms.js
@@ -619,7 +619,7 @@ function initializeInputs(context) {
     initializeSelectAllForm(this_context)
     initializeMultiValueChar(this_context)
 
-    $(".modal").each(function() {
+    $(this_context).each(function() {
         this_modal = $(this)
         initializeStaticChoiceSelection(this_modal, this_modal)
         initializeColorPicker(this_modal, this_modal)
--- a/nautobot/project-static/js/forms.js
+++ b/nautobot/project-static/js/forms.js
@@ -619,7 +619,7 @@ function initializeInputs(context) {
     initializeSelectAllForm(this_context)
     initializeMultiValueChar(this_context)
 
-    $(this_context).each(function() {
+    $(this_context).find(".modal").each(function() {
         this_modal = $(this)
         initializeStaticChoiceSelection(this_modal, this_modal)
         initializeColorPicker(this_modal, this_modal)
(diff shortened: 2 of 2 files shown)


changes/3018.fixed
```diff
@@ -0,0 +1 @@
+Update Select2 widgets to know they are in a modal.
```
glennmatthews, line 1:
```suggestion
Fixed rendering of Select2 widgets in modal dialogs.
```

nautobot/project-static/js/forms.js
```diff
@@ -613,6 +618,15 @@ function initializeInputs(context) {
     initializeDynamicFilterForm(this_context)
     initializeSelectAllForm(this_context)
     initializeMultiValueChar(this_context)
+
+    $(".modal").each(function() {
```
glennmatthews, line 622: This gets all `.modal` in the entire document right? Should it search only within `this_context` instead?

nautobot/project-static/js/forms.js
```diff
@@ -50,13 +50,14 @@ function colorPickerClassCopy(data, container) {
 */
 
 // Static choice selection
-function initializeStaticChoiceSelection(context){
+function initializeStaticChoiceSelection(context, dropdownParent=null){
```
glennmatthews, line 53: Instead of manually specifying a `dropdownParent` (and reinitializing menus inside a modal a second time), could we do something more automatic, like (not tested):

```js
this_context.find('.nautobot-select2-static').each(function() {
    $(this).select2({
        ...
        dropdownParent: $(this).closest(".modal")[0],
    });
)}
```

bryanculver: Since there are many more select widget that aren't in modals I thought the approach you were suggesting would require more iterations and checks. Reinitializing seems to be no issue for Select2 and would be a much smaller subset of widgets (since the call inwards would be scoped to the modal's context).

nautobot/project-static/js/forms.js
bryanculver: Possibly/probably? But in fact, this should probably live outside of `initializeInputs` because I don't see the use case where we are dynamically creating and registering new modals in the current UI paradigm.

glennmatthews (approved):

Issue #3013, opened 2022-12-16
Instrumentation of the job feature


Kircheneer:
### As ...

Patti - Platform Admin

### I want ...

metrics that describe the following aspects of the job feature:

- When, why (scheduling, manual, hook receiver, ...?), which exact jobs were executed
- What the result of that execution was
- How long did the execution take

### So that ...

I can use these metrics inside of my telemetry and observability solution.

### I know this is done when...

metrics are exposed under `/metrics` that detail the things listed above.

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [X] Other (not directly a platform feature)

### Database Changes

None

### External Dependencies

None

Kircheneer:
[Flower](https://flower.readthedocs.io/en/latest/prometheus-integration.html#available-metrics) helps with some of these, but to me it feels like it sits one abstraction level to high. It can tell me whether the celery task ran and if it succeeded, but succeeded there means that it terminated, which to Nautobot could be both success and failure. Furthermore it can't tell me for example which exact job was run, just that `run_job` was the celery task that was executed.

Kircheneer:
Possibly relates to #2993

Kircheneer:
#3045 should also provide a pattern for enabling plugin job instrumentation, which in the same vain currently is not possible.

PR #3045, opened 2022-12-23
Add prometheus instrumentation for job results.

Kircheneer:
# Closes: #3013 
# What's Changed

- Add `CELERY_WORKER_PROMETHEUS_PORT` configuration setting
- Add prometheus HTTP server listening on the worker to expose worker metrics
- Add nautobot_job_results_total counter metric that reports on job execution duration and outcome

# Example

This is an example output of the metrics at `http://worker:8080/metrics`:

```
# HELP nautobot_job_results_total Multiprocess metric
# TYPE nautobot_job_results_total histogram
nautobot_job_results_total_sum{module="example_plugin.jobs",name="ExampleLoggingJob",status="completed"} 1.080877
nautobot_job_results_total_sum{module="example_plugin.jobs",name="ExampleLoggingJob",status="errored"} 0.019623
nautobot_job_results_total_bucket{le="0.005",module="example_plugin.jobs",name="ExampleLoggingJob",status="completed"} 0.0
nautobot_job_results_total_bucket{le="0.01",module="example_plugin.jobs",name="ExampleLoggingJob",status="completed"} 0.0
nautobot_job_results_total_bucket{le="0.025",module="example_plugin.jobs",name="ExampleLoggingJob",status="completed"} 0.0
nautobot_job_results_total_bucket{le="0.05",module="example_plugin.jobs",name="ExampleLoggingJob",status="completed"} 0.0
nautobot_job_results_total_bucket{le="0.075",module="example_plugin.jobs",name="ExampleLoggingJob",status="completed"} 0.0
nautobot_job_results_total_bucket{le="0.1",module="example_plugin.jobs",name="ExampleLoggingJob",status="completed"} 0.0
nautobot_job_results_total_bucket{le="0.25",module="example_plugin.jobs",name="ExampleLoggingJob",status="completed"} 0.0
nautobot_job_results_total_bucket{le="0.5",module="example_plugin.jobs",name="ExampleLoggingJob",status="completed"} 0.0
nautobot_job_results_total_bucket{le="0.75",module="example_plugin.jobs",name="ExampleLoggingJob",status="completed"} 0.0
nautobot_job_results_total_bucket{le="1.0",module="example_plugin.jobs",name="ExampleLoggingJob",status="completed"} 0.0
nautobot_job_results_total_bucket{le="2.5",module="example_plugin.jobs",name="ExampleLoggingJob",status="completed"} 1.0
nautobot_job_results_total_bucket{le="5.0",module="example_plugin.jobs",name="ExampleLoggingJob",status="completed"} 1.0
nautobot_job_results_total_bucket{le="7.5",module="example_plugin.jobs",name="ExampleLoggingJob",status="completed"} 1.0
nautobot_job_results_total_bucket{le="10.0",module="example_plugin.jobs",name="ExampleLoggingJob",status="completed"} 1.0
nautobot_job_results_total_bucket{le="+Inf",module="example_plugin.jobs",name="ExampleLoggingJob",status="completed"} 1.0
nautobot_job_results_total_count{module="example_plugin.jobs",name="ExampleLoggingJob",status="completed"} 1.0
nautobot_job_results_total_bucket{le="0.005",module="example_plugin.jobs",name="ExampleLoggingJob",status="errored"} 0.0
nautobot_job_results_total_bucket{le="0.01",module="example_plugin.jobs",name="ExampleLoggingJob",status="errored"} 0.0
nautobot_job_results_total_bucket{le="0.025",module="example_plugin.jobs",name="ExampleLoggingJob",status="errored"} 1.0
nautobot_job_results_total_bucket{le="0.05",module="example_plugin.jobs",name="ExampleLoggingJob",status="errored"} 1.0
nautobot_job_results_total_bucket{le="0.075",module="example_plugin.jobs",name="ExampleLoggingJob",status="errored"} 1.0
nautobot_job_results_total_bucket{le="0.1",module="example_plugin.jobs",name="ExampleLoggingJob",status="errored"} 1.0
nautobot_job_results_total_bucket{le="0.25",module="example_plugin.jobs",name="ExampleLoggingJob",status="errored"} 1.0
nautobot_job_results_total_bucket{le="0.5",module="example_plugin.jobs",name="ExampleLoggingJob",status="errored"} 1.0
nautobot_job_results_total_bucket{le="0.75",module="example_plugin.jobs",name="ExampleLoggingJob",status="errored"} 1.0
nautobot_job_results_total_bucket{le="1.0",module="example_plugin.jobs",name="ExampleLoggingJob",status="errored"} 1.0
nautobot_job_results_total_bucket{le="2.5",module="example_plugin.jobs",name="ExampleLoggingJob",status="errored"} 1.0
nautobot_job_results_total_bucket{le="5.0",module="example_plugin.jobs",name="ExampleLoggingJob",status="errored"} 1.0
nautobot_job_results_total_bucket{le="7.5",module="example_plugin.jobs",name="ExampleLoggingJob",status="errored"} 1.0
nautobot_job_results_total_bucket{le="10.0",module="example_plugin.jobs",name="ExampleLoggingJob",status="errored"} 1.0
nautobot_job_results_total_bucket{le="+Inf",module="example_plugin.jobs",name="ExampleLoggingJob",status="errored"} 1.0
nautobot_job_results_total_count{module="example_plugin.jobs",name="ExampleLoggingJob",status="errored"} 1.0
```

# TODO
- [ ] Unit, Integration Tests
- [x] Example Plugin Updates (when adding/changing features) -> I don't believe this is applicable here


Kircheneer:
Please let me know if you do want tests for this feature. I think it could be done, but it would definitely require some thought.

glennmatthews:
Should these metrics include a term like `duration` in their names? I would naively expect `nautobot_job_results_total_sum` to be a sum of results, rather that what I think I understand here to be a sum of execution times?

development/docker-compose.yml
```diff
@@ -30,6 +30,8 @@ services:
         - "http://localhost:8080/health/"
   celery_worker:
     image: "local/nautobot-dev:local-py${PYTHON_VER}"
+    ports:
+      - "8081:8080"
```
glennmatthews, line 34: How does this work in the case where there are multiple celery workers for a given Nautobot instance? Would you just collect metrics from one arbitrary worker, since I think each worker will have identical metrics based on this implementation?

nautobot/core/celery/__init__.py
```diff
@@ -39,6 +43,28 @@
...
+        return
+
+    logger.info("Setting up prometheus metrics HTTP server for celery worker.")
+
+    # Ensure that the multiprocess coordination directory exists
+    multiprocess_coordination_directory = Path(os.environ["prometheus_multiproc_dir"])
+    multiprocess_coordination_directory.mkdir(parents=True, exist_ok=True)
+    # Delete any files lefts in the multiprocess coordination directory
+    for filepath in multiprocess_coordination_directory.glob("*.db"):
+        filepath.unlink()
```
glennmatthews, line 60: Any resource contention we would need to worry about if there are multiple workers sharing a filesystem, or if the worker shares a filesystem with the core app?

nautobot/extras/models/jobs.py
```diff
@@ -60,6 +61,14 @@
 # objects being created within transaction.atomic().
 JOB_LOGS = "job_logs"
 
+# The JOB_RESULT_METRIC variable is a counter metric that counts executions of jobs,
+# including information beyond what a tool like flower could get by introspecting
+# the celery task queue. This is accomplished by looking one abstraction deeper into
+# the job model of Nautobot.
+JOB_RESULT_METRIC = Histogram(
+    "nautobot_job_duration_seconds", "Results of Nautobot jobs.", ["module", "name", "status"]
```
glennmatthews, line 69: `nautobot_job_duration_seconds` here but `nautobot_job_results_total` in the changelog?

nautobot/extras/models/jobs.py
```diff
@@ -669,7 +678,14 @@ def set_status(self, status):
         """
         self.status = status
         if status in JobResultStatusChoices.TERMINAL_STATE_CHOICES:
+            # Only add metrics if we have a related job model. If we are moving to a terminal state we should always
+            # have a related job model, so this shouldn't be too tight of a restriction.
             self.completed = timezone.now()
+            if self.job_model:
+                duration = self.completed - self.created
+                JOB_RESULT_METRIC.labels(self.job_model.module_name, self.job_model.job_class_name, status).observe(
```
glennmatthews, line 686: Would it be preferable to use `grouping` and `name` (human-readable, configurable) instead of `module_name` and `job_class_name` (hard-coded)? Not sure, so I thought I'd ask.

nautobot/extras/models/jobs.py
Kircheneer: Good catch, I changed the metric name after adding the changelog and forgot to reflect that.

development/docker-compose.yml
Kircheneer: Excellent point. I'm going to write some thoughts about it here before I leave for the holidays.

# Multiple workers - one filesystem

In this case I think we should be good by wrapping a try/except around the `start_http_server` call as we don't really care which exact worker exposes the metrics and they all share the same filesystem directory (by the multiproc mechanism from the Python Prometheus client). This should work, because all of our workers should share the same lifetime - one wouldn't shut down significantly before another would, right?

# Multiple workers - multiple filesystems

In this case each worker will expose the metrics on its own port 8080, which the scraping configuration for a given Prometheus instance should be able to account for AFAIK.

nautobot/core/celery/__init__.py
Kircheneer: That is an excellent point. I included this as part of a snippet in an [issue](https://github.com/korfuri/django-prometheus/issues/339) that details this problem - if it doesn't cause any problems we should absolutely drop these two lines.

nautobot/extras/models/jobs.py
Kircheneer: I will gather some opinions about this and loop back around here.

Kircheneer: Changed.

nautobot/core/celery/__init__.py
Kircheneer: Changed to not delete files and added a comment.

Kircheneer:
> Should these metrics include a term like `duration` in their names? I would naively expect `nautobot_job_results_total_sum` to be a sum of results, rather that what I think I understand here to be a sum of execution times?

The metric is now called `nautobot_job_duration_seconds` - what it actually represents is buckets of durations (think 10ms, 20ms, [...], 1s, 5s) and the corresponding number of job results (with appropriate labeling) that fall into these duration buckets. Does this work for you?

development/docker-compose.yml
Kircheneer: These are my findings here.

**One worker - one machine - one process**: No problem.
**One worker - one machine - multiple processes**: No problem as the server starts on worker startup rather than child process startup.
**Multiple workers - one machine - any amount of processes**: This would currently break with this implementation. Is this a scenario that has any practical relevance at all, do we need it?
**Multiple workers - one per machine - any amount of processes**: No problem - in this case every worker exposes their metrics on the configured port. Depending on which `JobResult` instance is processed where, the metrics would be generated on the corresponding worker. In the Prometheus scraping config we can then scrape all the workers as individual targets, Prometheus itself should take care of the appropriate aggregation.
**Multiple workers - multiple per machine - any amount of processes**: See "Multiple workers - one per machine - any amount of processes".

Kircheneer: I have addressed the cases where multiple workers run on one machine now, as well.

nautobot/core/templates/nautobot_config.py.j2
```diff
@@ -293,6 +293,10 @@ SECRET_KEY = os.getenv("NAUTOBOT_SECRET_KEY", "{{ secret_key }}")
 # CELERY_TASK_SOFT_TIME_LIMIT = int(os.getenv("NAUTOBOT_CELERY_TASK_SOFT_TIME_LIMIT", str(5 * 60)))
 # CELERY_TASK_TIME_LIMIT = int(os.getenv("NAUTOBOT_CELERY_TASK_TIME_LIMIT", str(10 * 60)))
 
+# Port for prometheus metric HTTP server running on the celery worker
+# Set this to 0 to disable it.
```
glennmatthews, line 297: Please duplicate the very helpful comments from `settings.py` here so that they're more visible to a user setting up a new config file.

nautobot/core/templates/nautobot_config.py.j2
```diff
@@ -293,6 +293,10 @@ SECRET_KEY = os.getenv("NAUTOBOT_SECRET_KEY", "{{ secret_key }}")
...
+# CELERY_WORKER_PROMETHEUS_PORT = int(os.getenv("NAUTOBOT_CELERY_WORKER_PROMETHEUS_PORT"), 0)
```
glennmatthews, line 298: This seems wrong by comparison to `settings.py` - shouldn't it be the same logic as above, i.e. a comma-separated list?

changes/3013.added
```diff
@@ -0,0 +1,3 @@
+Added `CELERY_WORKER_PROMETHEUS_PORT` configuration setting
+Added prometheus HTTP server listening on the worker to expose worker metrics
+Added `nautobot_job_results_seconds` counter metric that reports on job execution
```
glennmatthews, line 3:
```suggestion
Added `nautobot_job_duration_seconds` counter metric that reports on job execution
```

glennmatthews (changes requested):

Kircheneer:
Feedback addressed. I forgot to copy the updated metric name/comments over from settings to the template.

glennmatthews:
I think I'm still missing a key point here. Why does this need to be a per-worker metric rather than part of the core Nautobot application's metrics? Is the idea that each worker only exports the histogram for the jobs that ran on that worker, not an aggregate for the whole system?

Kircheneer:
My understanding of this is as follows:

The worker(s) run(s) job code. As such, `JobResult.set_status` is also run in the worker context. Now the worker and the application _might_ share the same filesystem, in which case the main applications metrics endpoint would also expose the worker metrics (I guess we should document this). We can't assume this though, as it might not be the case. Therefore, we cannot (always) directly export these metrics into the web applications metrics endpoint.

Now there is a way to aggregate metrics from multiple places into one, using the [Pushgateway](https://prometheus.io/docs/practices/pushing/). In the docs for this it says:
> Usually, the only valid use case for the Pushgateway is for capturing the outcome of a service-level batch job

Although this is pretty much what we are doing, I still don't think its the right thing to do. It introduces more complexity (as in an additional service). It also introduces a single point of failure for metrics. Through the use of labels, Prometheus can handle the same metrics from multiple workers very well and it is possible to aggregate them after the fact. It might also be interesting to know if a certain worker is performing worse than others, right?

lampwins:
Yeah, it would be pretty typical to have the node exporter on each worker collect and expose these metrics on its own. No need to have the application try to do that for you, and often when you do that you lose interesting dimensions in the data.

nautobot/core/templates/nautobot_config.py.j2
glennmatthews:
```suggestion
# CELERY_WORKER_PROMETHEUS_PORT = []
# if os.getenv("NAUTOBOT_CELERY_WORKER_PROMETHEUS_PORT"):
#     CELERY_WORKER_PROMETHEUS_PORT = [
#         int(value) for value in os.getenv("NAUTOBOT_CELERY_WORKER_PROMETHEUS_PORT").split(",")
#     ]
```

nautobot/docs/configuration/optional-settings.md
```diff
@@ -257,6 +257,24 @@ The global Celery task hard timeout (in seconds). Any background task that excee
 
 ---
 
+## CELERY_WORKER_PROMETHEUS_PORT
+
++++ 1.5.x
+
+Default: `` (disabled)
```
glennmatthews, line 264:
```suggestion
Default: `[]` (disabled)
```

nautobot/docs/configuration/optional-settings.md
```diff
@@ -257,6 +257,24 @@ The global Celery task hard timeout (in seconds). Any background task that excee
...
+
+Environment Variable: `NAUTOBOT_CELERY_WORKER_PROMETHEUS_PORT`
+
+Ports for prometheus metric HTTP server running on the celery worker.
```
glennmatthews, line 268:
```suggestion
Ports for Prometheus metric HTTP server running on the celery worker(s).
```

nautobot/docs/configuration/optional-settings.md
```diff
@@ -257,6 +257,24 @@ The global Celery task hard timeout (in seconds). Any background task that excee
...
+
```
glennmatthews, line 269: Nit: would it be more accurate to change the variables to plural, i.e. `CELERY_WORKER_PROMETHEUS_PORTS` and `NAUTOBOT_CELERY_WORKER_PROMETHEUS_PORTS`?

nautobot/docs/configuration/optional-settings.md
```diff
@@ -257,6 +257,24 @@ The global Celery task hard timeout (in seconds). Any background task that excee
...
+Normally this should be set to a single port, unless you have multiple workers running on a single machine, i.e.
+sharing the same available ports. In that case you need to specify a range of ports greater than or equal to the
+highest amount of workers you are running on a single machine (comma-separated, like "8080,8081,8082"). You can then use
+the `target_limit` parameter to the Prometheus `scrape_config` to ensure you are not getting duplicate metrics in that
+case. Set this to an empty string to disable it.
```
glennmatthews, line 274:
```suggestion
case. Set this to an empty list to disable it.
```

glennmatthews (approved):
Just nits at this point. Thank you!

nautobot/docs/configuration/optional-settings.md
Kircheneer: Wouldn't it be an empty string still in this case, as in `""`? I realize this maps to an empty list through the configuration mechanism, but the actual setting is still a string type, right?

nautobot/docs/configuration/optional-settings.md
Kircheneer: Done.

nautobot/docs/configuration/optional-settings.md
glennmatthews: If I'm reading right, `CELERY_WORKER_PROMETHEUS_PORTS` should be an empty list; to make this happen with the built-in environment variables, you would set `NAUTOBOT_CELERY_WORKER_PROMETHEUS_PORTS` to an empty string?

glennmatthews (approved):

nautobot/docs/configuration/optional-settings.md
Kircheneer: Ahh right, I was mixing up environment variables and the actual settings. Fixed it, thanks!

Kircheneer:
Alright, I think that should be everything now. 👍🏻

nautobot/core/celery/__init__.py
```diff
@@ -39,6 +43,34 @@
...
+    # Don't set up the server if the port is undefined
+    if not settings.CELERY_WORKER_PROMETHEUS_PORTS:
+        return
+
+    logger.info("Setting up prometheus metrics HTTP server for celery worker.")
+
+    # Ensure that the multiprocess coordination directory exists. Note that we explicitly don't clear this directory
+    # out because the worker might share its filesystem with the core app or another worker. The multiprocess
+    # mechanism from prometheus-client takes care of this.
+    multiprocess_coordination_directory = Path(os.environ["prometheus_multiproc_dir"])
```
jathanism, line 58: This is the only reference to this environment variable here. Is this expected to be one set by the Prometheus client? Also curiously is it supposed to be lowercase? Thanks.

jathanism (commented):
Just curious about the environment variable here.

nautobot/core/celery/__init__.py
Kircheneer: Good catch! Here are the [docs](https://github.com/prometheus/client_python/blob/master/README.md#multiprocess-mode-eg-gunicorn) about it. As to your questions:
1. Why is this the only reference? While its the only reference in _this PR_ - there are more references in Nautobot, such as [this](https://docs.nautobot.com/projects/core/en/stable/additional-features/prometheus-metrics/?h=multi+process#multi-processing-notes) or [this](https://github.com/nautobot/nautobot/blob/develop/docker/Dockerfile#L100)
2. Why is it lowercase? Following the existing pattern here. You make a good point that it should probably be upper-case because it is documented like that, but it seems to work either way.

So I see two options, either we fix it _everywhere_ (7 matches in 6 files), or we leave it like this. What's your take?

jathanism: IMO since it is an environment variable, it would be nice to have it uppercased everywhere for consistency. But, seeing as it's already been like this, it's a whatever for me. Could leave it here, and then make it upper-cased in another smaller patch.

jathanism (approved):

Issue #2163, opened 2022-08-09
Eliminate `NullableCharFieldFilter`, and review `CharField` where `null=True`


jathanism:
### Proposed Changes

`NullableCharFieldFilter` (and `nautobot.NullableCharField` itself) needs to go away in v2. It is not even used.

Guidance here:
- Remove `nautobot.utilities.filters.NullableCharFieldFilter` and `nautobot.utilities.fields.NullableCharField`
- Establish style enforcement that any model `CharField` must never have `null=True` and `blank=True` set UNLESS `unique=True`
- Where not explicitly required, any `CharField` with `null=True` should be removed.

### Justification

It is bad data design to have `CharField` with `null=True` because this introduces two ways such a field can be considered "empty" (both an empty string or `None`).

From the Django docs on [`null`](https://docs.djangoproject.com/en/4.0/ref/models/fields/#null):

> Avoid using [`null`](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.Field.null) on string-based fields such as [`CharField`](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.CharField) and [`TextField`](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.TextField). If a string-based field has `null=True`, that means it has two possible values for “no data”: `NULL`, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not `NUL`L. One exception is when a [`CharField`](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.CharField) has both `unique=True` and `blank=True` set. In this situation, `null=True` is required to avoid unique constraint violations when saving multiple objects with blank values.

glennmatthews:
Review of existing cases:

- `BaseInterface.mac_address` has `null=True` but is not required to be unique. Should be made non-nullable.
- `JobLogEntry.log_object` and `JobLogEntry.absolute_url` have `null=True` but probably should be changed.
- `ScheduledJob.queue` has `null=True` and may have logic around  this, but should be changed to use blank instead.
- `WebHook.ca_file_path` has `null=True`, change it.

- `InventoryItem.asset_tag` has `null=True` but also `unique=True`, leave as is.
- `Device.name` has `null=True` and `db_index=True`, and is part of a `unique_together` constraint, leave as is for now.
- `Device.asset_tag` has `null=True` but also `unique=True`, leave as is.
[...]

PR #3046, opened 2022-12-23
Remove 'null=True' from CharFields that don't need it, remove related unused classes

glennmatthews:
# Closes: #2163 
# What's Changed

- As discussed in #2163, remove `null=True` from CharFields that don't actually need to support null values as part of a uniqueness constraint. Affected fields:
  - `Interface.mac_address` and `VMInterface.mac_address`
    - Since these use a `MACAddressCharField` which already translates between the DB (CharField) and Python (EUI instance), I've updated that class so that it still accepts `None` on write (translating it to "" in the DB) and translates DB value of "" back to `None` in Python.
  - `JobLogEntry.log_object`
  - `JobLogEntry.absolute_url`
  - `ScheduledJob.queue`
  - `Webhook.ca_file_path`
- Provide data migrations to correct existing records for all of the above
- Remove `NullableCharField` and `NullableCharFieldFilter` as unused/unneeded classes
- Remove `MACAddressField` (not to be confused with `MACAddressCharField` which is currently used by Interface/VMInterface) as unused/unneeded

# TODO
- [x] Unit, Integration Tests


nautobot/dcim/fields.py
```diff
@@ -56,42 +55,16 @@ def to_python(self, value):
...
         if isinstance(value, str):
             value = value.strip()
-        try:
-            return EUI(value, version=48, dialect=mac_unix_expanded_uppercase)
-        except AddrFormatError:
-            raise ValidationError(f"Invalid MAC address format: {value}")
-
-    def get_prep_value(self, value):
-        if not value:
+        if value == "":
```
HanlinMiao, line 58: Do we still need the `if value is None:` condition on line 54 if we are already checking that in `get_prep_value()` by `if not value: return ""`?
Also what happens when `value` is "  "? If it satisfies the condition on line 58, that would return `None` is that what we want?

gsnider2195:
Just for clarification, the reason we're keeping `null=True` on unique fields is because when checking for uniqueness `null != null` but `"" == ""`?

gsnider2195 (approved):

nautobot/dcim/fields.py
glennmatthews: The diffs make it a bit confusing here but IIRC this is necessary.

`to_python()` is converting the DB value to a Python value, and this check here is needed to ensure that a DB value of `""` maps to a Python value of `None` (so that the Python value is either an EUI object or None, never a string). I could *probably* rearrange things here to combine the checks from line 54 and line 58 into a single check but I though this was perhaps clearer (at least, when not viewed as a diff)?

Conversely, `get_prep_value()` is converting the Python value to a DB value, thus the change on line 67 below to ensure that the DB value is now `""` rather than `null` since the field is no longer nullable.

glennmatthews:
> Just for clarification, the reason we're keeping `null=True` on unique fields is because when checking for uniqueness `null != null` but `"" == ""`?

Correct - for those fields we currently permit multiple records with a value of `null`, but the uniqueness constraint would prevent us from permitting multiple records with `""`.

HanlinMiao (approved):

Issue #3026, opened 2022-12-19
Client-Side UI: Extend React Prototype


bryanculver:
Draw the rest of the owl time:

- [x] Core rendering HTML content provided by a plugin (see detail view plugin content left-side/right-side) `PluginTemplateExtension.[right/left/full_width]_page `
- [x] Plugin providing additional tabs on object detail view `PluginTemplateExtension.detail_tabs`
- [x] Core view rendering new components provided by a plugin (New, should be available via React)
- [x] Plugin providing buttons in list view (TBD)
    - With what we've implemented for the plugin features above this should be straight forward to implement
- [x] Plugin overriding entire core view `Views.override_views` (assuming Django routing is no longer in play)
- [ ] Plugin's home & config views:
   - `://nautobot/plugins/example-plugin/config/`
   - `://nautobot/plugins/example-plugin/`

gsnider2195:
## Core rendering HTML content provided by a plugin (see detail view plugin content left-side/right-side) PluginTemplateExtension.[right/left/full_width]_page 

- Works similarly to Nautobot 1.x
- No changes for plugin developers
- Plugin template content fetched from django by javascript and loaded into the js ui via the [`dangerouslySetInnerHTML`](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml) property

## TODO

- [ ] Load all tabs on page load but don't load content until active tab changes

gsnider2195:
## Plugin providing additional tabs on object detail view PluginTemplateExtension.detail_tabs

- Works similarly to Nautobot 1.x
- No changes for plugin developers
- Plugin template content fetched from django by javascript and loaded into the js ui via the [dangerouslySetInnerHTML](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml) property

## TODO

- [ ] Load all tabs on page load but don't load content until active tab changes

gsnider2195:
## Core view rendering new components provided by a plugin (New, should be available via React)

- TODO

gsnider2195:
## Plugin providing buttons in list view (TBD)

- TODO?

gsnider2195:
## Plugin overriding entire core view Views.override_views (assuming Django routing is no longer in play)

- TODO

gsnider2195:
## Plugin's home & config views

- TODO

gsnider2195:
[...]

PR #3061, opened 2022-12-30
Nautobot dev with react

timizuoebideri1:
# What's Changed

## Routing
The goal of moving routing to React using a JSON-like configuration is to simplify the process of creating and defining routes.

One reason for doing this is that not all routes need to interact with a database or API, and in those cases, it may be unnecessary to create a route in Django. By using a JSON-like configuration in React, these types of routes can be defined more easily and efficiently.

Another benefit of using this approach is that it allows plugin developers to create custom routes in React, rather than having to define them in both Django and React. This avoids the need to maintain two separate sets of routes and simplifies the process of adding new routes.

Finally, defining routes using this method is generally easier and more convenient than using either Django routing or React routing alone. By combining the advantages of both systems, it is possible to create a more flexible and user-friendly routing system that is better suited to a variety of different use cases.

# TODO

- [ ] Unit, Integration Tests


bryanculver:
Another prototype in the pursuit of #3026

PR #3074, opened 2023-01-04
Enforce sandboxed rendering for Jinja2 templates.

jathanism:
# Closes: DNE
# What's Changed

As a preventative measure, this change forces the Jinja2 template engine to use a `SandboxedEnvironment`.

- This addresses any unsafe code execution everywhere `render_jinja2` is called.
- Test added to assert that unsafe object attributes cannot be accessed and will result in a `SecurityError`.
- Documentation referencing direct use of `jinja2.Template` replaced with `render_jinja2`

# TODO

- [ ] Unit, Integration Tests


glennmatthews (approved):

changes/3074.security
```diff
@@ -0,0 +1 @@
+Sandboxed rendering of Jinja2 templates is now enforced by default in keeping with [Jinja2 best practices](https://jinja.palletsprojects.com/en/3.0.x/sandbox/#sandbox). To enable template sandboxing in a Nautobot instance without needing to upgrade, add the following value to your `nautobot_config.py`: `TEMPLATES[1]["OPTIONS"]["environment"] = "jinja2.sandbox.SandboxedEnvironment"` and restart your Nautobot services.
```
HanlinMiao, line 1: Quick question, is it guaranteed that "Jinja" and its related attributes will always be the second entry (`TEMPLATES[1]`) in the `TEMPLATES` configuration list? If not, should we do a more generic version of this value?

HanlinMiao (approved):
It looks good except for the comment I made

changes/3074.security
jathanism: Yes, because we configure "django" and "jinja" out of the box. See: https://github.com/nautobot/nautobot/blob/c3559e30ef8adb1b7d47e84432807d8bf5774ad7/nautobot/core/settings.py#L434-L473

HanlinMiao: I see I see, I thought the user might be able to change the order of things in their own `nautobot_config.py`. Thank you for the clarification

jathanism: They would have to do so explicitly, in which case, I would reckon they would know how to update it themselves. :)

PR #3105, opened 2023-01-10
Avoid accessing the DB when serializing/deserializing NautobotFakeRequest

glennmatthews:
# Closes: #2244 

(May not address all symptoms reported over the life of #2244 but it seems to at least address the latest symptoms reported on that issue.)

# What's Changed

- When deserializing a previously serialized `NautobotFakeRequest`, we were immediately loading the fake request's `user` attribute from the DB. This was proving problematic in some environments as the Celery worker main process would be hitting the DB before it even forked off a worker process to handle the Job request.
   - Solution part 1: when deserializing a `NautobotFakeRequest` from JSON, instead of loading the User from the database immediately, set it as a lazy reference (similar to how real Django Requests normally operate).
   - This wasn't sufficient to fix the early DB access from the Celery worker, because it turns out that after deserializing the object from JSON, Celery itself then tries to `pickle` it (yes, even though we have Celery configured throughout to only use JSON serialization/deserialization!), which was itself automatically resolving the lazy reference.
   - Solution part 2: implement the `pickle` API on `NautobotFakeRequest` and ensure that when pickling it, if we already have the user PK cached on the object, we don't do an unnecessary DB access to look it up again.

This is a different fix than that proposed by @lampwins in https://github.com/nautobot/nautobot/issues/2244#issuecomment-1232110673, which is something we still might want to pursue separately?

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/utilities/tests/test_utils.py
+++ b/nautobot/utilities/tests/test_utils.py
@@ -623,7 +623,7 @@ class NautobotFakeRequestTest(TestCase):
         self.assertEqual(fake_request.user, self.user)
 
     def test_fake_request_json_no_extra_db_access(self):
-        """Verify that serializing and deserializing a NautobotFakeRequest as JSON doesn't necessarily access the DB."""
+        """Verify that serializing and deserializing a NautobotFakeRequest as JSON doesn't unnecessarily access the DB."""
         real_request = self.get_with_user("/")
         fake_request = copy_safe_request(real_request)
         with self.assertNumQueries(0):
@@ -637,7 +637,7 @@ class NautobotFakeRequestTest(TestCase):
             new_fake_request.user
 
     def test_fake_request_pickle_no_extra_db_access(self):
-        """Verify that pickling and unpickling a NautobotFakeRequest doesn't necessarily access the DB."""
+        """Verify that pickling and unpickling a NautobotFakeRequest doesn't unnecessarily access the DB."""
         real_request = self.get_with_user("/")
         fake_request = copy_safe_request(real_request)
         with self.assertNumQueries(0):
--- a/nautobot/utilities/tests/test_utils.py
+++ b/nautobot/utilities/tests/test_utils.py
@@ -1,12 +1,19 @@
+import pickle
+
 from django import forms
 from django.core.exceptions import ValidationError
 from django.db.models import Q
 from django.http import QueryDict
-from django.test import TestCase
+from django.test import RequestFactory
 
 from nautobot.core.settings_funcs import is_truthy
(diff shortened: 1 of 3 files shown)


gsnider2195:
This change looks good to me. I think a unit test would be a good idea here.

gsnider2195:
I was thinking maybe something like this?

```py
from nautobot.core import celery

# Do we need to allow one database query to retrieve and store the user pk?
with self.assertNumQueries(1):
    request = {"user": self.user}
    serialized_request = celery._dumps(request)
    deserialized_request = celery._loads(serialized_request)
```

nautobot/utilities/tests/test_utils.py
```diff
@@ -587,3 +593,59 @@ def test_ensure_content_type_and_field_name_inquery_params(self):
...
+        # After creating the new instance, its `user` is a lazy attribute, which should evaluate on demand:
+        with self.assertNumQueries(1):
+            new_fake_request.user
+            self.assertEqual(new_fake_request.user, self.user)
+        # It should then be cached and not require re-lookup from the DB
+        with self.assertNumQueries(0):
+            new_fake_request.user
+
+    def test_fake_request_pickle_no_extra_db_access(self):
+        """Verify that pickling and unpickling a NautobotFakeRequest doesn't necessarily access the DB."""
```
gsnider2195, line 640:
```suggestion
        """Verify that pickling and unpickling a NautobotFakeRequest doesn't unnecessarily access the DB."""
```

nautobot/utilities/tests/test_utils.py
```diff
@@ -587,3 +593,59 @@ def test_ensure_content_type_and_field_name_inquery_params(self):
...
+        self.assertEqual(real_request.GET, fake_request.GET)
+        self.assertEqual(real_request.user, fake_request.user)
+        self.assertEqual(real_request.path, fake_request.path)
+        for key in fake_request.META.keys():
+            self.assertIn(key, HTTP_REQUEST_META_SAFE_COPY)
+            self.assertEqual(real_request.META[key], fake_request.META[key])
+        self.assertEqual(fake_request.user, self.user)
+
+    def test_fake_request_json_no_extra_db_access(self):
+        """Verify that serializing and deserializing a NautobotFakeRequest as JSON doesn't necessarily access the DB."""
```
gsnider2195, line 626:
```suggestion
        """Verify that serializing and deserializing a NautobotFakeRequest as JSON doesn't unnecessarily access the DB."""
```

gsnider2195 (approved):

Issue #3104, opened 2023-01-10
Unable to remove filter with X button in ui


gsnider2195:
### Environment
* Nautobot version (Docker tag too if applicable): v1.5.7
* Python version: 3.7.15
* Database platform, version: pg13
* Middleware(s):

### Steps to Reproduce
1. go to `/dcim/devices`
2. filter on any site
3. try to click the X on the filter to remove the filter

![image](https://user-images.githubusercontent.com/75227981/211598269-5dd90f18-c835-418d-8f70-4c2d874ac3bf.png)

### Expected Behavior
Filter is removed

### Observed Behavior
Filter remains

PR #3106, opened 2023-01-10
Fix bug preventing filters from being removed from list views

gsnider2195:
# Closes: #3104 
# What's Changed

- Fix bug in `nautobot.core.utilities.check_filter_for_display()` and update `object_list.html` to use the new `dict` format of the view's `filter_params` context variable. 
- Add an integration test to prevent this from happening going forward

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/3104.fixed
+++ b/changes/3104.fixed
@@ -1,0 +1,1 @@
+Fixed bug preventing filters from being removed from list views.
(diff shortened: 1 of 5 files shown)


glennmatthews (commented):
Not obvious to me why the integration tests are failing, but other than that this is looking like a solid fix.

gsnider2195:
> Not obvious to me why the integration tests are failing, but other than that this is looking like a solid fix.

The integration test should be working now. I had to create my own test data instead of relying on the factory data since this is a `TransactionTestCase`.

glennmatthews (approved):

Issue #1721, opened 2022-05-02
Remove django-cacheops dependency/usage


glennmatthews:
### As ...

Nelly - Network Engineer

### I want ...

To have confidence that the Nautobot UI and API are accurately showing me data as it _is_ in the database, not as it _was_ in the past. 

### So that ...

I will not be shown outdated information that misleads me as to what the current state of the database is, such as can be seen in #1627.

### I know this is done when...

- [x] Nautobot no longer uses `django-cacheops`
- [ ] The `invalidate` mgmt command is removed
- [ ] Nautobot overall performance is acceptably similar to where it was when cacheops was still in use
- [ ] Issues such as #1627 are no longer seen

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [X] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

N/A

### External Dependencies

Remove `django-cacheops` dependency; shouldn't add any new dependencies.

## TODO

- [ ] Blocked by #1633
- [ ] Remove any stray `prefetch_related` that could be moved to `select_related` that wasn't already addressed in #3112 
- [ ] Remove Cacheops

bryanculver:
We need to be mindful of how to not regress on #1515.

We leveraged `cacheops` because if it's file-based caching for the problems it tried to solve there.

jathanism:
One option is that we could leave `cacheops` as a dependency for using `@cacheops.file_cache.cached()` but disable it for use in queryset caching

Another would be to switch to another library that supports disk caching. One of the newer ones [`DiskCache` has built-in support for Django caching framework](https://grantjenks.com/docs/diskcache/tutorial.html#djangocache). A decorator pattern would still need to be considered.

bryanculver:
Let's also see if this resolves #1723, #1694

bryanculver:
We will likely need to keep the dependency in the short term due to being leveraged by plugins, but ultimately rip out in the future.

jathanism:
Last comment is that we also need to consider the monkey-mixin pattern, to carry forward into v2 when we ditch it. We could forklift the requisite code and ditch cacheops and of course provide attribution in the `NOTICE` file.

[...]

PR #3112, opened 2023-01-12
Converted eligible `prefetch_related` to `select_related` queries

lampwins:
# What's Changed

I've changed all eligible `prefetch_related()` field queries to `select_related()`

I have not attempted to change any logic related to prefetch handling in the `ModelViewSet` with `brief_prefetch_fields`. We should leave that to 2.0.

# TODO

- [x] Unit, Integration Tests


glennmatthews (commented):
Since this'll (as I understand it) cause all sorts of issues if a deployment still has cacheops enabled, we should probably add a startup check to enforce that cacheops must be disabled as a part of this change.

nautobot/dcim/views.py
```diff
@@ -2684,15 +2621,13 @@ class CableBulkImportView(generic.BulkImportView):
 
 
 class CableBulkEditView(generic.BulkEditView):
-    # v2 TODO(jathan): Replace prefetch_related with select_related
     queryset = Cable.objects.prefetch_related("termination_a", "termination_b")
```
glennmatthews, line 2624: For my understanding - no select_related here because these are generic foreign keys?

lampwins: Yep, the django docs are a little confusing on this point IMO, but it is here https://docs.djangoproject.com/en/dev/ref/models/querysets/#django.db.models.query.QuerySet.prefetch_related

nautobot/core/checks.py
```diff
@@ -40,6 +40,13 @@
     obj=settings,
 )
 
+W006 = Warning(
```
lampwins, line 43: @glennmatthews I went with a warning because there is technically nothing stopping you from doing it, if you accept the risk of invalid results.

This is not a hill I will die on though, if we want to make this an error.

nautobot/core/checks.py
```diff
@@ -40,6 +40,13 @@
...
+    "CACHEOPS_ENABLED is set to True but cachops is no longer supported in v1.5.8. It can still be used but may lead to "
```
glennmatthews, line 44:
```suggestion
    "CACHEOPS_ENABLED is set to True but cacheops is no longer recommended in v1.5. It can still be used but may lead to "
```

changes/3112.changed
```diff
@@ -0,0 +1 @@
+Converted eligible `prefetch_related()` to `select_related()` queries. Users should note a perforamnce gain from this change, but note that cacheops is no longer recomended in v1.5 and this change will likely result in invalide data responses if cacheops remains enabled in your environment. Cacheops will be removed entirely in a future release.
```
glennmatthews, line 1:
```suggestion
Converted eligible `prefetch_related()` to `select_related()` queries. Users should note a performance gain from this change, but note that cacheops is no longer recommended in v1.5 and this change will likely result in invalid data responses if cacheops remains enabled in your environment. Cacheops will be removed entirely in a future release.
```

glennmatthews (approved):

bryanculver:
@lampwins @glennmatthews would you consider this done then: https://github.com/nautobot/nautobot/issues/1721#issuecomment-1315897140

lampwins:
@bryanculver I would

Issue #3068, opened 2023-01-03
Fix circuits, ipam, tenancy, users apps FKs and related names


gsnider2195:
See spike https://github.com/nautobot/nautobot/issues/2960#issuecomment-1349809073 for details

bryanculver:
Closed via #3090 and #3094

PR #3113, opened 2023-01-12
Renamed reverse-relation `user_set` to `users` on the Group model.

HanlinMiao:
# Closes: #3068 
# What's Changed

# TODO

- [ ] Unit, Integration Tests


glennmatthews (commented):
I'm nervous about this one, since it's changing a database field that's built in to Django itself. I don't see anywhere in Django core that's making use of `group.user_set` but I'd worry about any of our dependencies (social-auth for example?) potentially doing so. I'd question whether the benefit here is worth the risk.

HanlinMiao:
> I'm nervous about this one, since it's changing a database field that's built in to Django itself. I don't see anywhere in Django core that's making use of `group.user_set` but I'd worry about any of our dependencies (social-auth for example?) potentially doing so. I'd question whether the benefit here is worth the risk.

I see, should we just close this PR and not touch the `users` app?

HanlinMiao:
See @glennmatthews's comment for closing this PR. Basically we do not want to change any native models and attributes from Django. The risks outweigh the benefits here.

Issue #2721, opened 2022-11-03
Apps Refactor: Collapse Utilities App to Core


bryanculver:
The bulk of the guts of the `utilities` app should no longer exist in v2. The primitives such as filters, forms, fields, template tags, etc., should move to `core`.

Ultimately, we might need to leave it around until v2.2 if we decide we want to maintain any modicum of forward-compatibility, otherwise we may just consider this entire endeavor a breaking change.

- [x] Subdivide different components of utilities
- [x] Lift-and-shift the code from utilities to core
- [x] Correct import paths from other apps to the now core import path
- [x] Create migration docs for Apps to fix imports from now missing module

glennmatthews:
Implemented by #3132 for 2.0.

PR #3132, opened 2023-01-19
Collapse utilities app into core

glennmatthews:
# Closes: #2721 
# What's Changed

- Eliminate the `nautobot.utilities` app, folding it in under `nautobot.core`
  - Many modules moved over one-for-one as new modules, e.g. `nautobot.utilities.forms` --> `nautobot.core.forms`
  - Many other modules merged into the existing core modules by the same name, e.g. `nautobot.utilities.factory` was merged into the existing `nautobot.core.factory` module
  - A few modules were renamed for self-consistency and structural purposes, e.g. `nautobot.utilities.api` -> `nautobot.core.api.utils`
  - A number of top-level `nautobot.utilities.*` modules were moved to `nautobot.core.utils.*` to reduce top-level clutter.
  - **I'm absolutely open to discussion and revisiting of any specific module name/location decisions!!**
- One circular dependency issue that the above introduced was with `nautobot.core.constants.SEARCH_TYPES`, the definition of which required importing models, tables, and filtersets from all other apps, which in turn were importing lots of stuff from `utilities`, which when merged into `core` introduced circular imports.
   - This constant is only used once, in `nautobot.core.views.SearchView`.
   - We don't want anything in `core` to depend on imports from other apps.
   - ~For the moment, I "fixed" this issue by adding a `get_search_types()` API to the `NautobotConfig` class and letting each app provide its own additions to this "constant". This is still less than ideal since the models/tables/filtersets can't be imported at app-level but need to be imported locally within the `get_search_types` function, but it works for the moment. Longer-term I feel like this "should" be discoverable more cleanly.~
   - **Thanks to functions like `reverse()`, `resolve()`, `get_route_for_model` we can actually derive everything the search needs from a simple list of apps and modelnames per app!** I've refactored the code to take advantage of this; as a bonus this now allows apps to easily register their own models for the "global search" as well!
- Update all imports within core to use the new modules
  - Reorder imports where appropriate to reflect the new alphabetical order
  - There is one place where doing this introduced a circular import loop between `nautobot.core.models` and `nautobot.extras.models`. For the moment, keeping the `core.models` import after the `extras.models` import in the `nautobot.dcim.models.cables` file seems to be sufficient to prevent this loop.  I've added a TODO comment in this file; probably this will be cleanly resolvable only by collapsing extras into core as well (#2722).

~I haven't yet updated the 1.x-to-2.x migration guide but that should be done before merging this.~ See below for additional TODOs.

# TODO

- [x] Revise `get_search_types` pattern
- [ ] Update `nautobot.apps` with redirects to any classes/functions/modules that we expect apps to make use of
   - [ ] Backport these additions to the namespace into 1.5.x
- [x] Unit, Integration Tests


nautobot/core/constants.py
```diff
@@ -1,262 +1,50 @@
-from collections import OrderedDict
```
glennmatthews, line 1: Consider moving `nautobot.utilities.config` to `nautobot.core.utils.config` instead of the top-level `nautobot.core.config` so as to reduce the likelihood of confusing it with `nautobot.core.settings`?

nautobot/core/factory.py
```diff
@@ -1,8 +1,9 @@
+from django.db.models import Model
```
glennmatthews, line 1: Consider merging `nautobot.utilities.error_handlers` into `nautobot.core.views.utils` instead of a new top-level `nautobot.core.error_handlers` module, as it's only used in view code.

nautobot/core/utils/utils.py
```diff
@@ -27,8 +27,8 @@
 from django_filters.utils import verbose_lookup_expr
```
glennmatthews, line 27: I don't like having `nautobot.core.utils.utils` - would appreciate suggestions as to a better name for this submodule.

nautobot/core/constants.py
bryanculver: I know we went back and forth in the discussion last week but I thought we landed on

`nautobot.core.[module].utils` instead of `nautobot.core.utils.[module]`

nautobot/core/utils/utils.py
glennmatthews: Possibly `utils.functions`? Or possibly split this into several more narrowly targeted modules, such as `utils.lookup` (`get_route_for_model`, `get_related_class_for_model`, etc.)?

nautobot/core/constants.py
glennmatthews: The thing is a lot of these don't *have* directly corresponding top-level modules... `nautobot.core.config.utils` isn't any less messy than a top-level `nautobot.core.config`, whereas `nautobot.core.utils.config` at least groups a lot of disparate files under the `.utils` namespace.

For cases where there *is* a top-level module already, I've done my best to follow that (`nautobot.core.api.utils`, `nautobot.core.views.utils`). It's a messy problem to be sure. :-)

nautobot/core/utils/utils.py
glennmatthews: > Could we eliminate the folder and just have `utils.py`?

Are you thinking to merge all of these files into a single `utils.py`? To my mind that's not a great solution since it results in that one file being a massive dumping-ground for miscellaneous functionality:

```
> ls -1 nautobot/core/utils
__init__.py
config.py
deprecation.py
git.py
logging.py
ordering.py
paginator.py
permissions.py
query_functions.py
utils.py
```

jathanism:
re this comment:

> For the moment, I "fixed" this issue by adding a get_search_types() API to the NautobotConfig class and letting each app provide its own additions to this "constant". This is still less than ideal since the models/tables/filtersets can't be imported at app-level but need to be imported locally within the get_search_types function, but it works for the moment. Longer-term I feel like this "should" be discoverable more cleanly.

Longer term, we should be using real a search backend and not just iterating over an ever-growing list of unindexed query sets.

nautobot/core/utils/utils.py
bryanculver: ACK

How big is too big to be broken up into smaller files? < 500?

nautobot/core/constants.py
bryanculver: Since relocating is pain, I'd like to follow a convention if possible. Thankfully this ditches the entire utilities app but `core.utils.config` and `core.views.utils` still seems opaque to me.

nautobot/core/forms/search.py
```diff
@@ -1,8 +1,8 @@
+from django.apps import apps
 from django import forms
 
-from nautobot.utilities.forms import (
-    BootstrapMixin,
-)
+from nautobot.core.forms import BootstrapMixin
+
 
 OBJ_TYPE_CHOICES = (
```
bryanculver, line 7: No sense in keeping this since it's no longer referenced anywhere.

jathanism (approved):
IMO this is good enough to merge and iterate upon based on any comments/findings.

bryanculver (approved):
:shipit:

changes/3132.added
```diff
@@ -0,0 +1 @@
+Added the ability for apps to register their models for inclusion in the global Nautobot search.
```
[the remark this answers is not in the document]
glennmatthews: Happened as a serendipitous side effect of my refactoring of the search setup. :-)

nautobot/core/constants.py
glennmatthews: `core.utils` is still a bit of a catch-all, but I think this is a net improvement (new diffs coming soon):

```
> ls -1 nautobot/core/**/utils.py
nautobot/core/api/utils.py
nautobot/core/forms/utils.py
nautobot/core/graphql/utils.py
nautobot/core/models/utils.py
nautobot/core/testing/utils.py
nautobot/core/views/utils.py
```

```
> ls -1 nautobot/core/utils/
color.py
config.py
data.py
deprecation.py
filtering.py
git.py
logging.py
lookup.py
permissions.py
requests.py
```

So far I haven't been able to come up with good "top-level" `nautobot.core` namespaces to put any of the remainder into. (I made a stab at converting `nautobot.core.filters` to a submodule and moving `nautobot.core.utils.filtering` to `nautobot.core.filters.utils` but that introduced some unpleasant circular dependencies between `nautobot.core.models`, `nautobot.core.forms`, and `nautobot.core.filters`...)

nautobot/core/utils/utils.py
glennmatthews: See [comment above](https://github.com/nautobot/nautobot/pull/3132#discussion_r1085963290) for my attempt at getting rid of `nautobot.core.utils.utils` by splitting it out into other modules.

nautobot/core/models/name_color_content_types.py
```diff
@@ -3,18 +3,17 @@
...
 from django.urls import reverse
 
-from nautobot.core.fields import AutoSlugField
+from nautobot.core.choices import ColorChoices
+from nautobot.core.fields import AutoSlugField, ColorField
 from nautobot.core.models import BaseModel
+from nautobot.core.models.querysets import RestrictedQuerySet
 
 # Importing CustomFieldModel, ChangeLoggedModel, RelationshipModel from  nautobot.extras.models
 # caused circular import error
```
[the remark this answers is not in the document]
glennmatthews: Looks like it. The issue is that `nautobot.extras.models.roles` imports this file, so we can't have this file import the entire `nautobot.extras.models` module here without it becoming a circular import.

nautobot/dcim/models/cables.py
```diff
@@ -20,9 +21,13 @@
 )
 from nautobot.extras.models import Status, StatusModel
 from nautobot.extras.utils import extras_features
+
+# TODO: There's an ugly circular-import pattern where if we move this import "up" to above, we get into an import loop
+# from dcim.models.cables to core.models.generics to extras.models.datasources to core.models.generics.
+# Deferring the update to here works for now; fixing so that core.models.generics doesn't depend on extras.models
+# would be the much more invasive but much more "correct" fix.
```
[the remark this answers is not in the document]
glennmatthews: I think so but not sure we'll be able to pull it off without collapsing all/most of extras into core as well. TBD.

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -289,3 +289,41 @@ Below is a table documenting [removed redundant filter field changes](../release
...
+| `nautobot.utilities.query_functions` | `nautobot.core.utils.query_functions` |
+| `nautobot.utilities.querysets`       | `nautobot.core.models.querysets`      |
+| `nautobot.utilities.tables`          | `nautobot.core.tables`                |
+| `nautobot.utilities.tasks`           | `nautobot.core.tasks`                 |
+| `nautobot.utilities.templatetags`    | `nautobot.core.templatetags`          |
+| `nautobot.utilities.testing`         | `nautobot.core.testing`               |
+| `nautobot.utilities.tree_queries`    | `nautobot.core.models.tree_queries`   |
+| `nautobot.utilities.utils`           | `nautobot.core.utils.utils`           |
+| `nautobot.utilities.validators`      | `nautobot.core.models.validators`     |
+| `nautobot.utilities.views`           | `nautobot.core.views.utility_mixins`  |
```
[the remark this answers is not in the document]
glennmatthews: Will give it a shot.

glennmatthews:
> > * Update nautobot.apps with redirects to any classes/functions/modules that we expect apps to make use of
> >   
> >   * Backport these additions to the namespace into 1.5.x
> 
> Is this one we should apply to `develop` first, merge `develop` into `next`, then make the changes? If so I'd say ship this for now and leave it as a PR fix when merging `develop` into `next`.

We need to review the changes being made in this PR (which I've summarized in the migrating docs, https://github.com/nautobot/nautobot/pull/3132/files#diff-d5e7d13f177913ec08c4f0735b2744d57d1b223ee1eb55fbd577505cb16fea23) and decide on which of the changed modules/classes/functions are worth exposing in the `nautobot.apps` namespace in order to reduce the impact to app developers when moving from 1.5 to 2.0. I'm fine with doing that as a subsequent task but it will likely need to be done separately in both `next` and `develop`.

nautobot/core/utils/git.py
```diff
@@ -6,7 +6,7 @@
 
 from git import Repo
 
-logger = logging.getLogger("nautobot.utilities.git")
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
glennmatthews: Opened #3160.

nautobot/core/filters.py
```diff
@@ -15,10 +15,9 @@
 from drf_spectacular.utils import extend_schema_field
 from taggit.managers import TaggableManager
 
-from nautobot.dcim import fields as dcim_fields
-from nautobot.dcim import forms as dcim_forms
-from nautobot.extras import models as extras_models
-from nautobot.utilities import constants, forms, utils
+from nautobot.core import constants, forms
+from nautobot.core.models import fields as core_fields
+from nautobot.core.utils import data as data_utils
```
bryanculver, line 20: Why did the `dcim` app disappear here? Was utilities importing core forms via DCIM?

glennmatthews: Nice spot! Yeah, we were pulling in `MACAddressField` (form field) and `MACAddressCharField` (model field) from DCIM. I ran into some dependency issues around that which I fixed by moving those two classes from DCIM into core.

nautobot/dcim/models/cables.py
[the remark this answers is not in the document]
glennmatthews: Left a comment on #2722 regarding the remaining cases where we're importing `extras` into `core`. TLDR: I think moving `Tag` and its associated code into core would *help* but I'm not sure it's enough on its own.

PR #3047, opened 2022-12-23
Create Region and Site Location Types and Locations during data migration

HanlinMiao:
# Closes: #2955 
# What's Changed

# TODO

- [ ] Explanation of Change(s)
- [ ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [ ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example Plugin Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #3148, opened 2023-01-23
Revert "Create Region and Site Location Types and Locations during data migration"

HanlinMiao:
Reverts nautobot/nautobot#3047


glennmatthews (approved):
Ruh roh!

Issue #3085, opened 2023-01-05
Jobs Overhaul: Merge `TaskResult` and `JobResult` data models


jathanism:
- [x] Augment `JobResult` with `TaskResult` fields
- [x] Replace `JobResultStatusChoices` enum values with those of Celery task states
- [x] Replace `TaskResult` with `JobResult` on `NautobotDatabaseBackend`
- [x] Migrate state transition machinery from `run_job()` task to `NautobotDatabaseBackend` state handlers and `NautobotTask` state handlers
- [x] Fix double JSON encoding/decoding for task args/kwargs (especially `request`) and result/data fields.
- [x] Fix system tasks compatibility with `JobResult` (these weren't previously stored in database)
- [x] Assert that `JobResult.data` is updated correctly.
- [x] Get all the tests running and updated
- [x] Remove `NAUTOBOT_CELERY_RESULT_BACKEND` environment variable
- [x] Remove optional settings docs for `CELERY_RESULT_BACKEND` and `CELERY_RESULT_BACKEND_TRANSPORT_OPTIONS`

bryanculver:
Let's see if we can pull in #3089 as a part of this.

bryanculver:
Let's figure out how much of https://github.com/nautobot/nautobot/issues/275 is quasi-resolved or started because of the changes for this (#3164 👀)

Issue #3065, opened 2023-01-03
Unable to Edit/Delete/Add Objects in v1.5.6


phoenixswiss:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.6
* Python version: 3.8.10
* Database platform, version: postgresql, Ubuntu 12.12-0ubuntu0.20.04.1
* Middleware(s): -

### Steps to Reproduce
1. Create/Edit/Delete an object (e.g. Interface/IP-Address) in Nautobot either via Web GUI, API-Page or pynautobot

### Expected Behavior
The creation / edit / deletion of the object will be made without an error. Worked on 1.4.10 Release until upgrade to 1.5.6

### Observed Behavior
**Web GUI:** An HTTP 400 error (Bad Request (400)) will be shown and no change will be made. The following log will be created in the syslog
```
[...] nautobot-server[2528]: ERROR:django.security.DisallowedHost:Invalid HTTP_HOST header: 'testserver'. You may need to add 'testserver' to ALLOWED_HOSTS.
[...] nautobot-server[2528]: WARNING:django.request:Bad Request: /dcim/interfaces/add/
[...] nautobot-server[2528]: [pid: 2528|app: 0|req: 154/154] x.x.x.x () {76 vars in 1973 bytes}  POST /dcim/interfaces/add/?device=397a12bb-db12-4b87-9283-a3c7cf231c91&return_url=/dcim/devices/397a12bb-db12-4b87-9283-a3c7cf231c91/interfaces/ => generated 143 bytes in 69 msecs (HTTP/1.1 400) 6 headers in 182 bytes (1 switches on core 0)
```
**API-Page:**
_Request_
```
curl -X 'DELETE' \
  'https://***********/api/dcim/interfaces/da7c8120-5da4-42aa-b724-79760c900587/' \
  -H 'accept: */*' \
  -H 'X-CSRFTOKEN: SsqqVYngkfktBhknV2cChj9vDSU91nOsjgp70RSpd0MwU0BbGFakLA5pTNH9oM3u'
```

_Response body_
```
Download
{
  "error": "Invalid HTTP_HOST header: 'testserver'. You may need to add 'testserver' to ALLOWED_HOSTS.",
  "exception": "DisallowedHost",
  "nautobot_version": "1.5.6",
  "python_version": "3.8.10"
}
```

_Response headers_
```
connection: keep-alive 
 content-length: 191 
 content-type: application/json 
[...]

Issue #2726, opened 2022-11-03
Jobs Overhaul: Jobs should no longer wrap in atomic, leave to author


bryanculver:
As opposed to #1991, we should remove all atomic "magic" behavior and leave it to the Job author.

This allows authors to decide where and when they want certain database calls to be atomic. Since more and more jobs are interacting with other systems, wrapping the entire job in `with transaction.atomic:` causes data inconsistencies: Nautobot can roll back its database changes but API calls to other systems that may have created data.

Remove the testing specific workaround and any tests related to the atomic block functionality.

lampwins:
Need good documentation on migration path for this.

gsnider2195:
Do we want to continue to provide the `commit` argument from core or leave that up to the job author to implement as well?

gsnider2195:
Decided to remove `commit` argument

gsnider2195:
What's the plan for `dry-run` and scheduled job approval? If a job requires approval but dry-run is set the job can be run without approval. Does that feature get dropped as part of this issue or does the commit flag need to remain to support this?

bryanculver:
I think `dry-run` needs to be dropped as well because AFAIK that's being done via aborting the transaction (as in commit=False).

Job approval would stay though. This would provide greater weight to the job approval process: ensuring the job has a dry-run option it builds if necessary or preventing accidental non-dry-runs.

lampwins:
Thinking about this... I feel we should maintain the `dry-run` primitive as an interface, but provide no implementation by default. In essence this is the same as ansible's `--check` in that Job authors ("modules" in ansible speak) must implement support for `dry-run` in their job but doing so will enable a standard and consistent UX because this is a common use case and is expected in an automation platform. I see us implementing another `Meta` flag like `provides_dry_run = True` and we would use that in the UI and the approval/schedule flow to allow the relevant functionality.

So the onus is still on the Job author, but we provide the standard interface pattern. The Job author could then rely on a standard `dry_run` input and they can chose what to do with that in their code.

bryanculver:
Closed as was done by not implementing in #3483

Issue #1807, opened 2022-05-24
post_run method fails to add exceptions to job log


dnewood:
### Environment
* Python version: 3.7.13
* Nautobot version: 1.3.2

### Steps to Reproduce
1. Create a job with a `post_run()` method that should raise an Exception. Example:
```python
class TestPostRun(Job):

    class Meta:
        name = "Test Post Run"
        description = "Testing Post Run Stuffz"
        commit_default = True

    def run (self, data, commit):
        self.log_success(message="Yay, Job ran!")
    
    def post_run(self):
        self.log_success(message="Post run has started! Let's break stuff.")
        
        # Purposely cause an exception
        lol_fake_object.add(idk_this_should_break)
        
        self.log_info(message="Testing to see if stuff is broken.")
```
2. Run the job.

### Expected Behavior
Expected to see a log entry within the job showing that the `post_run` group experienced a NameError exception.

### Observed Behavior
The post_run method just "eats" the exception and stops the job silently. You can however see the expected exception in the worker logs:
```
[2022-05-24 22:06:00,915: ERROR/ForkPoolWorker-2] Task nautobot.extras.jobs.run_job[a9df438c-d93c-47ca-8771-a1b7f0af9fa6] raised unexpected: NameError("name 'lol_fake_object' is not defined")
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/celery/app/trace.py", line 451, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/celery/app/trace.py", line 734, in __protected_call__
    return self.run(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/nautobot/extras/jobs.py", line 1185, in run_job
    _run_job()
  File "/usr/local/lib/python3.7/site-packages/nautobot/extras/jobs.py", line 1172, in _run_job
    output = job.post_run()
  File "/opt/nautobot/jobs/CreateFwPolicyv2.py", line 463, in post_run
    lol_fake_object.add(idk_this_should_break)
NameError: name 'lol_fake_object' is not defined
```
However the Job in the GUI makes it seem like everything went fine:
![Screen Shot 2022-05-24 at 6 06 42 PM](https://user-images.githubusercontent.com/67761754/170143376-65261572-1c27-481c-912b-90a38a3ec106.png)

jathanism:
[...]

PR #3164, opened 2023-01-25
Merge `TaskResult` into `JobResult` objects

jathanism:
# Closes: #3085 
# What's Changed

**WARNING: This is ugly. Getting this out there to show a checkpoint.**

JobResult fields renamed:
- `job_id` => `task_id`
- `created` => `date_created`
- `completed` => `date_done`
- `job_kwargs` => `task_kwargs`

JobResult fields changed
- `status` choices changed to mirror that of `TaskResult.status.choices`
- Added all the missing fields found on `TaskResult` to `JobResult`
-`JobResult.task_name` now mirrors `TaskResult.task_name` and is used distinctly
- `obj_type` is now nullable so that system jobs like `nautobot.extras.tasks.provision_field` can be run and stored in DB, but we might want to revisit this to make it conditional around "system jobs", but this was necessary to make non-Job Celery tasks run at all without an `IntegrityError`
- The `JobResultManager` now sets `name` to match `task_name` if `name` isn't set.
- Caught some field names I missed in the Job Result templates.

Other changes
- Celery testing has now been updated to run tasks with `ALWAYS_EAGER=True` so that instead of requiring a single-threaded worker process using the `solo` pool strategy, we can just run them "synchronously" and have them correctly update the state of the `JobResult`.
- `nautobot.core.testing.run_job_for_testing` now stores the Celery result from a job on the `JobResult` object it returns as `JobResult.celery_result`
- This was necessary due to the fact that `nautobot.extras.jobs.run_job` no longer handles exceptions internally and allows them to bubble up. `run_job` is called synchronously within `run_job_for_testing`, so a need to catch exceptions was introduced as a result.
- Now instead of using a custom context manager, `run_job_for_testing`, a dummy `Request` object is created (if one isn't passed) and wrappped using `copy_safe_request`
- Added the `traceback` field to the "Additional Data" tab on `JobResult` detail view template
- Tests for extras API and views had to be revised for new expectations in that there will ALWAYS be `JobResult` objects now and to filter them by the name of the Job model class.
- Revised `nautobot.extras.tests.test_jobs.JobFileUploadTest.test_run_job_fail` to check that an exception was raised by its job run and that the exception matches the expected type.
- Placed `TODO` comments across various places in `run_job()` where I had to restore calls to `job_result.save()` to preserve the functionality of how the job model is mutating `JobResult.data` as it goes. This pattern needs to be refactored for this new paradigm as we work to eliminate `run_job` entirely.
- Excluded `content_type` and `content_encoding` fields from `JobResult`. Had to effectively duplicate the base `JobResultManager.store_result()` manager method instead of overloading it and calling the base in order to accomplish this, but it should be more efficient overall since it won't result in double saving of an object.

Outstanding issues

- [x] Still need to migrate business logic from `run_job` into db backend/task classes
- [x] There's an issue where `date_done` is being nullified that needs to be addressed likely because of some race condition between `run_job` internals and database backend state engine (see previous bullet)
- [x] `TaskStateChoices` and `celery_states` need to be folded into a revamped `JobResultStatusChoices` choice set, so please wait on nitpicking that.
- [x] Most of UI is working except for tracking success/done state refresh of results
- [x] Unit tests omega broken again because of the status enums

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/models/jobs.py
```diff
@@ -530,19 +538,59 @@ class JobResult(BaseModel, CustomFieldModel):
...
+
+    date_created = models.DateTimeField(auto_now_add=True)
+
+    @property
+    def created(self):
+        return self.date_created
+
+    @created.setter
+    def set_created(self):
+        self.date_created = created
```
jathanism, line 550: These will go away so we can make sure to find calling code.

nautobot/extras/models/jobs.py
```diff
@@ -530,19 +538,59 @@ class JobResult(BaseModel, CustomFieldModel):
...
+
+    date_done = models.DateTimeField(null=True, blank=True)
+
+    @property
+    def completed(self):
+        return self.date_done
+
+    @completed.setter
+    def set_completed(self, completed):
+        self.date_done = completed
```
jathanism, line 560: Same here. Will go away.

nautobot/extras/models/jobs.py
```diff
@@ -530,19 +538,59 @@ class JobResult(BaseModel, CustomFieldModel):
...
+    content_type = models.CharField(
+        default="application/x-nautobot-json",
+        max_length=128,
+        verbose_name=('Result Content Type'),
+        help_text='Content type of the result data')
+    content_encoding = models.CharField(
+        default="utf-8",
+        max_length=64,
+        verbose_name=('Result Encoding'),
+        help_text='The encoding used to save the task result data')
```
jathanism, line 586: These are currently required by Celery, but since we're always using JSON this is just a passthrough right now. Need to review the base `DatabaseBackend` from `django-celery-results` and see if we can just always use JSON and eliminate or ignore these fields.

nautobot/extras/models/jobs.py
```diff
@@ -693,10 +752,20 @@ def enqueue_job(cls, func, name, obj_type, user, *args, celery_kwargs=None, sche
...
+        # Explicitly create a TaskResult before the Celery task is called to
+        # test that we can have it there. This was confirmed to work.
+        """
+        from django_celery_results.models import TaskResult
+        from django.utils import timezone
+        t = TaskResult.objects.create(task_id=job_result.task_id)
+        t.date_created = timezone.datetime(2021, 12, 28, 19, 30, 23, 742973, tzinfo=timezone.now().tzinfo)
+        t.save()
+        """
+
```
jathanism, line 768: Remove. This was for early debug.

jathanism:
This is the current issue with encoding, specifically the args/kwargs seem to not being Kombu encoded: 

```
{'_state': <django.db.models.base.ModelState at 0x11843b850>,
 'id': UUID('f9a98b40-77cf-4e50-bf7c-66b0a5175395'),
 '_custom_field_data': {},
 'job_model_id': UUID('64b239ab-71aa-4a59-a99f-002665f642bb'),
 'name': 'plugins/example_plugin.jobs/ExampleLoggingJob',
 'task_name': 'nautobot.extras.jobs.run_job',
 'obj_type_id': 16,
 'date_created': datetime.datetime(2023, 1, 27, 19, 48, 31, 714111, tzinfo=<UTC>),
 'date_done': datetime.datetime(2023, 1, 27, 19, 48, 36, 222092, tzinfo=<UTC>),
 'user_id': UUID('2b12f865-52db-4405-8fc8-1c4d4375c2c5'),
 'status': 'SUCCESS',
 'data': None,
 'periodic_task_name': None,
 'worker': 'celery@jathy-studio.lan',
 'task_args': '"()"',
 'task_kwargs': '"{\'data\': {\'interval\': 4, \'_task_queue\': \'default\'}, \'request\': <nautobot.utilities.utils.NautobotFakeRequest object at 0x1149db280>, \'commit\': True, \'task_queue\': \'default\', \'job_result_pk\': UUID(\'f9a98b40-77cf-4e50-bf7c-66b0a5175395\')}"',
 'content_type': 'application/x-nautobot-json',
 'content_encoding': 'utf-8',
 'result': 'null',
 'traceback': None,
 'meta': '{"children": []}',
 'schedule_id': None,
 'task_id': UUID('7ded996a-e588-4fef-b214-7df7d6e699db'),
 'use_job_logs_db': True}
```

nautobot/core/celery/backends.py
```diff
@@ -1,7 +1,60 @@
...
+        return "application/x-nautobot-json", "utf-8", data
+
+    def decode_content(self, obj, content):
+        """Pass through decoding since we're storing as JSON explicitly."""
+        return content
+
+    def _get_extended_properties(self, request, traceback):
+        """
+        Overload default so that `argsrepr` and `kwargsrepr` aren't used to construct `args` and
+        `kwargs`. Our need to properly (de)serialize `NautobotFakeRequest` facilitates this.
```
glennmatthews, line 24: If we get rid of `NautobotFakeRequest`, will that get rid of the need to override this private method? Not saying that has to be done as part of this PR though.

nautobot/extras/api/views.py
```diff
@@ -709,10 +709,10 @@ def list(self, request, *args, **kwargs):
             r.name: r
             for r in JobResult.objects.filter(
                 obj_type=job_content_type,
-                status__in=JobResultStatusChoices.TERMINAL_STATE_CHOICES,
+                status__in=JobResultStatusChoices.READY_STATES,
```
glennmatthews, line 712: Which states are READY_STATES? At a glance this change seems odd or perhaps even reversed from the old TERMINAL_STATE_CHOICES - are SUCCESS and FAILURE really the READY_STATES?

nautobot/extras/datasources/registry.py
```diff
@@ -46,7 +46,7 @@ def refresh_datasource_content(model_name, record, request, job_result, delete=F
                     job_result.log(
                         f"Error while refreshing {entry.name}: {exc}", level_choice=LogLevelChoices.LOG_FAILURE
                     )
-                    job_result.set_status(JobResultStatusChoices.STATUS_ERRORED)
+                    job_result.set_status(JobResultStatusChoices.STATUS_FAILURE)
```
glennmatthews, line 49: Does this file also need a TODO about removing the `set_status` calls?

nautobot/extras/jobs.py
```diff
@@ -1185,37 +1151,35 @@ def _run_job():
...
-                    raise AbortTransaction()
+                    raise AbortTransaction("Database changes have been reverted automatically.")
 
         except AbortTransaction:
             if not job_model.read_only:
                 job.log_info(message="Database changes have been reverted automatically.")
 
-        except Exception as exc:
-            stacktrace = traceback.format_exc()
-            job.log_failure(message=f"An exception occurred: `{type(exc).__name__}: {exc}`\n```\n{stacktrace}\n```")
```
glennmatthews, line 1204: How do exceptions thrown during job execution get recorded now?

nautobot/extras/jobs.py
```diff
@@ -1225,36 +1189,24 @@ def _run_job():
...
-            message = (
-                f"An exception occurred during job post_run(): `{type(exc).__name__}: {exc}`\n```\n{stacktrace}\n```"
-            )
-            output = message
-            job.log_failure(message=message)
-            job_result.set_status(JobResultStatusChoices.STATUS_ERRORED)
-        finally:
-            if output:
-                job.results["output"] += "\n" + str(output)
+        output = job.post_run()
```
glennmatthews, line 1192: The try/except was added specifically for #1807 - are exceptions captured/logged differently now?

nautobot/extras/managers.py
```diff
@@ -0,0 +1,30 @@
+from celery import states
+from django.utils import timezone
+from django_celery_results.managers import TaskResultManager, transaction_retry
+
+from nautobot.core.models.querysets import RestrictedQuerySet
+
+
+# This subclass is a hack. We'll fix it in post.
```
glennmatthews, line 8: :-) Can we at least make this a TODO comment?

nautobot/extras/managers.py
```diff
@@ -0,0 +1,30 @@
...
+class JobResultManager(RestrictedQuerySet.as_manager().__class__, TaskResultManager):
+    @transaction_retry(max_retries=2)
+    def store_result(self, *args, **kwargs):
+        """
+        Overload default to manage custom behaviors for Nautobot integration.
+        """
+        obj = super().store_result(*args, **kwargs)
+
+        # Make sure `date_done` is allowed to stay null until the task reacheas
+        # a ready state.
```
glennmatthews, line 18: What's the default behavior of django-celery-results here that we have to override??

nautobot/extras/models/jobs.py
```diff
@@ -521,27 +522,57 @@ class JobResult(BaseModel, CustomFieldModel):
     )
 
     name = models.CharField(max_length=255, db_index=True)
+    task_name = models.CharField(max_length=255, null=True, db_index=True)
```
glennmatthews, line 525: Comments or help_text explaining the difference between these two fields?

nautobot/extras/models/jobs.py
```diff
@@ -521,27 +522,57 @@ class JobResult(BaseModel, CustomFieldModel):
...
+    )
+    result = models.TextField(
+        null=True,
+        default=None,
+        editable=False,
+        verbose_name="Result Data",
+        help_text="The data returned by the task. Use content_encoding and content_type fields to read.",
+    )
+    traceback = models.TextField(blank=True, null=True)
+    meta = models.JSONField(null=True, default=None, editable=False)
```
glennmatthews, line 575: Surprising to me that there isn't an abstract base class we can inherit from for this set of fields.

nautobot/extras/jobs.py
jathanism: They get stored on the `JobResult.traceback` field and unless otherwise handled by decorating the task with a `throws=` argument, they result in the job status of `FAILURE`.

Cursor_and_Job_Result__File_Upload_Failure_-_Nautobot

nautobot/extras/jobs.py
jathanism: Yes. We should allow all exceptions to bubble up and they are handled by Celery and then the traceback gets captured.

nautobot/extras/managers.py
jathanism: Default behavior is having this field as a `DateField(auto_now=True)` which just automatically updates the `date_done` field on every state transition. This is different than Celery's default behavior (and the current behavior of Nautobot) to keep it null until there is a state transition to a ready state (e.g. `SUCCESS`, `REVOKED`, `FAILURE`).

So I made it to align with expectations that `date_done` will remain null until there is a ready state, just how it works in Celery core. The default Celery backend does this. The DCR one does not.

nautobot/extras/models/jobs.py
jathanism: That would be lovely. Alas: https://github.com/celery/django-celery-results/blob/47402b8120b7b84e741b1a38d67caf05146c62a9/django_celery_results/models.py#L18

nautobot/extras/api/views.py
jathanism: This is using the same terminology as within Celery core. Ready states are terminal states (success, failure, revoked). There are various other states.

nautobot/core/celery/backends.py
jathanism: Yes, I think so. The other problem are the nullable `CharField` fields we forklifted. If we want "correctness" I would rather those default to `""` vs. `None` but that's actually more invasive than what I had to do here.

nautobot/extras/datasources/registry.py
jathanism: Actually I meant to remove all of those. Anywhere explicit status changes are being made must go away. We want Celery to make these decisions, along with any explicit calls to `JobResult.save()`, moving those to the `NautobotTask` class or the backend or a mix of both, depending.

nautobot/extras/models/jobs.py
jathanism: I still want to see if I can eliminate these fields since they are 100% no longer necessary. This would be done in `JobResultManager.store_result()`.

nautobot/extras/api/views.py
glennmatthews: Ack. Can we add a comment to that effect in the JobResultStatusChoices definition?

nautobot/core/testing/__init__.py
```diff
@@ -86,16 +88,20 @@ def run_job_for_testing(job, data=None, commit=True, username="test-user", reque
...
 
-    @contextmanager
-    def _web_request_context(user):
-        if request:
-            yield request
-        else:
-            with web_request_context(user=user) as ctx_request:
-                yield ctx_request
+    if request is None:
+        factory = RequestFactory()
```
glennmatthews, line 92: We shouldn't use RequestFactory in non-test code. (#3065).

nautobot/core/utils/requests.py
```diff
@@ -47,12 +47,11 @@ def nautobot_serialize(self):
...
+        user = data.pop("user")
+        data.pop("_cached_user", None)
         if "_user_pk" not in data:
             # We have a user but haven't stored its PK yet, so look it up and store it
-            data["_user_pk"] = data.pop("user").pk
-        else:
-            # We already have stored the PK, and we need to AVOID actually resolving the lazy user reference.
-            data.pop("user")
+            data["_user_pk"] = user.pk
         return data
```
glennmatthews, line 55: Wonder if this should be back-ported to 1.5.

nautobot/extras/templates/extras/jobresult.html
```diff
@@ -100,6 +100,14 @@
                         {% include 'extras/inc/json_data.html' with data=result.data format="json" %}
                     </div>
                 </div>
+                <div class="panel panel-default">
+                    <div class="panel-heading">
+                        <strong>Traceback</strong>
+                    </div>
+                    <div class="panel-body">
+                        {% include 'extras/inc/json_data.html' with data=result.traceback format="python" %}
```
glennmatthews, line 108: Wondering if we should do a check for settings.DEBUG here - similar to how we only present full tracebacks for other exceptions (500 errors) if it's True?

nautobot/extras/managers.py
bryanculver: Would it be more desirable to document that `date_done` behaves more like `last_updated` and that + status would let you know if it's date done or not?

Would we want to change this in the future?

nautobot/extras/datasources/registry.py
bryanculver: @jathanism Do you intend to do this in this PR or open a follow-on story?

nautobot/extras/filters/__init__.py
```diff
@@ -655,13 +655,13 @@ class JobResultFilterSet(BaseFilterSet, CustomFieldModelFilterSetMixin):
...
-    created = django_filters.DateTimeFilter()
-    completed = django_filters.DateTimeFilter()
+    date_created = django_filters.DateTimeFilter()
+    date_done = django_filters.DateTimeFilter()
     status = django_filters.MultipleChoiceFilter(choices=JobResultStatusChoices, null_value=None)
 
     class Meta:
         model = JobResult
-        fields = ["id", "created", "completed", "status", "user", "obj_type", "name"]
+        fields = ["id", "date_created", "date_done", "name", "status", "user", "obj_type"]
```
bryanculver, line 664: Let's be sure to add either as a requirement to this PR or a follow-on to create the migration docs for all the model changes here.

Probably would be a good pre-release checklist item is to review every migration in `next` and ensure we have a migration guide time around it.

nautobot/extras/jobs.py
```diff
@@ -1185,27 +1163,24 @@ def _run_job():
...
-                else:
-                    job.logger.info("job completed successfully")
-                    job_result.set_status(JobResultStatusChoices.STATUS_COMPLETED)
+                job.logger.info("job completed successfully")
 
                 if not commit:
-                    raise AbortTransaction()
+                    raise AbortTransaction("Database changes have been reverted automatically.")
 
         except AbortTransaction:
```
bryanculver, line 1171: I think it would help if we completed #2726 shortly after merging this.

nautobot/core/utils/requests.py
jathanism: It might not be necessary, seeing as we don't have the recurring serialize/deserialize issue we have here with the database backend. The `_cached_user` only becomes a part of the payload when it is re-serialized from a deserialized payload a 2nd time.

nautobot/extras/jobs.py
jathanism: Any traceback will result in a failure unless it's defined as a known exception type such as using [`autoretry_for`](https://docs.celeryq.dev/en/stable/userguide/tasks.html?highlight=exceptions#automatic-retry-for-known-exceptions)

Also traceback is currently part of the "Additional Data" tab:

image

Unless you're thinking of showing it in the job log as well?

nautobot/extras/templates/extras/jobresult.html
jathanism: IMO if there is a traceback we should show it. I was thinking that since it's stored on the object, presenting it if it exists is easy. Is there a concern about data exposure or something tangible?

glennmatthews: Yeah, that was my thinking.

nautobot/core/celery/task.py
```diff
@@ -8,10 +8,20 @@
...
-        logger.debug(">>>  KWARGS = %s", kwargs)
-
 
 Task = NautobotTask  # noqa: So that the class path resolves.
+
+
+# TODO(jathan): Remove this once this body of work is done. This is just useful for debugging but it
+# results int a lot of noise and slows things down.
+# from celery import signals
+# @signals.task_prerun.connect
```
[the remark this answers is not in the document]
jathanism: I just uncomment it locally when I need it for now. Leaving it in for testing generates a stupendous amount of noise.

I think documenting how to use it would be valuable.

nautobot/extras/datasources/registry.py
jathanism: Follow-on as a part of the "make `run_job` a thing of the past" work.

nautobot/extras/templates/extras/jobresult.html
jathanism: That's valid. IMO let's leave that as a v2 UI thing. I've tracked it to add to the backlog.

nautobot/extras/jobs.py
jathanism: I'm thinking that'll just be another one for ripping out `run_job`. Leave out the atomic block while we're at it.

PR #3167, opened 2023-01-26
Fix Role Migrations

bryanculver:
# Closes: #DNE
# What's Changed

- [x] `VLAN.legacy_role__name` is not an attribute. Can't use the QuerySet filter to access this. Instead, check if we don't get a string back and instead grab the name of the associated object.
- [x] `ObjectChange` records associated to `ipam.Role` make stale content type clean up sad
- [x] Probably need to do the same cleanup for `DeviceRole`...
- [ ] There exists another bug using the demo dataset and get collisions if an `ipam.Role` has the same slug as the previously choice-only role (possibly if name doesn't match but slug does?)

# TODO

- [?] Unit, Integration Tests


nautobot/dcim/migrations/0025_device_and_rack_roles_data_migrations.py
```diff
@@ -12,6 +12,22 @@ def migrate_data_from_legacy_role_to_new_role(apps, schema):
         model = apps.get_model("dcim", model_name)
         migrate_role_data(model=model, role_model=role_model)
 
+    # TODO: We need to decide how these should properly be done...
+    # - Do we keep them but update ObjectChange to allow changed_object_type to be NULL?
+    #   - What will break in UI/elsewhere that expects to be able to get the changed_object?
+    # - Do we move the ObjectChange records?
+    #   - _technically_ this isn't correct because the records are now associated to more than just Prefix/VLAN roles
```
glennmatthews, line 19: I still believe this is the least-bad option.

nautobot/extras/utils.py
```diff
@@ -435,7 +435,7 @@ def migrate_role_data(
         queryset=queryset,
         role_model=role_model,
         role_choiceset=role_choiceset,
-        legacy_role=legacy_role_name,
+        legacy_role=legacy_role,
```
glennmatthews, line 438: is `legacy_role_name` now unused? The logic above checks for a special case of `not (is_choice_field or is_m2m_field)` - is that a case we need to handle specially in some other way?

nautobot/extras/migrations/0057_collect_roles_from_related_apps_roles.py
```diff
@@ -61,22 +61,46 @@ def bulk_create_roles(apps, roles_to_create, content_types):
...
-            weight=getattr(data, "weight", None),
+        (
+            data,
+            Role(
+                name=data.name,
+                slug=data.slug,
+                description=data.description,
+                color=getattr(data, "color", color_map["default"]),
+                weight=getattr(data, "weight", None),
+            ),
```
glennmatthews, line 85: Should we instead be using `Role.objects.get_or_create()` as a way to handle existing roles?

nautobot/extras/migrations/0057_collect_roles_from_related_apps_roles.py
```diff
@@ -61,22 +61,46 @@ def bulk_create_roles(apps, roles_to_create, content_types):
...
     if roles_instances:
-        Role.objects.bulk_create(roles_instances, batch_size=1000)
+        for old_role, new_role in roles_instances:
+            # TODO: We need to handle new role collisions
+            new_role.save()
+            if old_role_ct:
+                # Move over existing object change records to the new role we created
+                ObjectChange.objects.filter(changed_object_type=old_role_ct, changed_object_id=old_role.pk).update(
+                    changed_object_type=new_role_ct, changed_object_id=new_role.pk
+                )
```
glennmatthews, line 98: Won't this skip over ObjectChanges for roles that happen to already have been created with a different content-type?

nautobot/extras/migrations/0057_collect_roles_from_related_apps_roles.py
```diff
@@ -61,22 +61,46 @@ def bulk_create_roles(apps, roles_to_create, content_types):
...
+
+    # This is for all of the change records for roles which no longer exist
+    if old_role_ct:
+        # TODO: Should we null out the changed_object_id?
```
glennmatthews, line 102: I don't think we do for object deletes in general?

bryanculver: We don't which is why I didn't do it here and a UUID is a UUID and even if it were in the same ContentType the likelihood you'll re-use a UUID is staggeringly small.

nautobot/extras/migrations/0057_collect_roles_from_related_apps_roles.py
bryanculver: Unless my logic is extremely faulty the way I've read through the order of these migrations is that won't ever happen. The migration doesn't appear to collapse Roles into a single object, hence the collision problem that still exists.

I would ultimately like to refactor the way these migrations happen before we ship 2.0.

Currently, we have a dependency on the non-extras apps schema migrations on this data migration and we loosely associate the old role name to the new ones by hoping they have a similar name and/or slug.

I've tested this pretty extensively locally, restoring the DBs and re-running for different scenarios.

nautobot/extras/migrations/0057_collect_roles_from_related_apps_roles.py
bryanculver: Maybe? But there's issues in the data migrations in the soon-to-be-deleted roles and associated objects.

We need some logic around "if slug matches but name differs" or "if name matches but description differs" etc.

glennmatthews (commented):
Just tried running the updated migrations from a `develop` dataset to this branch, and while it no longer errors out, I get the following:

```
  Applying ipam.0010_migrate_ipam_role_data...
16:48:20.419 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.422 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.427 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.429 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.432 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.434 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.437 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.438 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.439 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.443 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.445 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.447 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.450 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.453 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.454 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.456 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.462 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.466 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.473 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.476 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
16:48:20.480 ERROR   nautobot.extras.utils utils.py        get_instances_to_pk_and_new_role() :
  Role with name  not found
Role with name  not found
```

which seems suspect to me?

glennmatthews (approved):
Good enough for the moment. Let's make sure we get a backlog item opened to revisit this though.

Issue #1380, opened 2022-02-16
Add the device name to the header (Rename interfaces)


jifox:
### Environment
* Python version: 3.8
* Nautobot version: 1.2.5

### Proposed Functionality

The header should be "**Renaming 56 Interfaces (DEVICE-NAME)**"

### Use Case

As Nelly the Network Engineer I want to ensure not to rename the interfaces of a different device before applying the changes. In the table no reference to the device name is present

![image](https://user-images.githubusercontent.com/8406301/154272899-86d1661f-261f-4b2d-aff7-ba06e1ce86e7.png)

Alternativly use the "Rename" header and add the device e.g. "Rename interfaces of ROUTER1"

![image](https://user-images.githubusercontent.com/8406301/154273255-ce951c65-c0b6-4510-9138-188a7cd5e5bc.png)

### Database Changes

None

PR #3168, opened 2023-01-26
Add device name to bulk interface rename header

timizuoebideri1:
# Closes: #1380 
# What's Changed

# TODO

- [ ] Unit, Integration Tests


nautobot/dcim/templates/dcim/interface_bulk_rename.html
```diff
@@ -0,0 +1,5 @@
+{% extends 'generic/object_bulk_rename.html' %}
+{% load helpers %}
+{% load form_helpers %}
+
+{% block title %}Renaming {{ selected_objects|length }} {{ device_name }} {{ obj_type_plural|bettertitle }}{% endblock %}
```
glennmatthews, line 5: Is the need for this fix exclusive to interfaces? I'd think all device-components (console ports, power ports, vminterfaces, etc. etc.) would also benefit from this. Should this be an across-the-board enhancement to `BulkRenameView` instead?

nautobot/dcim/tests/test_views.py
```diff
@@ -1971,6 +1972,39 @@ def setUpTestData(cls):
...
+            response = self.client.post(self._get_url("bulk_rename"), data, follow=True)
+            expected_message = """
+            <div class="alert alert-warning alert-dismissable" role="alert">
+                <button type="button" class="close" data-dismiss="alert" aria-label="Close">
+                    <span>&times;</span>
+                </button>
+                No interfaces were selected.
+            </div>
+            """
+            self.assertInHTML(expected_message, response.content.decode(response.charset))
```
glennmatthews, line 2006: Similarly, this check (at least) should probably be part of the base `DeviceComponentViewTestCase`.

nautobot/dcim/views.py
```diff
@@ -136,6 +136,11 @@ def post(self, request):
         )
 
 
+class BaseDeviceComponentsBulkRenameView(generic.BulkRenameView):
+    def get_selected_objects_parents_name(self, selected_objects):
+        return selected_objects.first().device.name
```
glennmatthews, line 141: Can we use `.parent.name` so that both VMInterface and all the DeviceComponent views all can share the same code?

nautobot/utilities/testing/views.py
```diff
@@ -1404,3 +1404,45 @@ def test_bulk_add_component(self):
...
+            with self.subTest("Assert if no objects selected return with error"):
+                data["pk"] = []
+                response = self.client.post(self._get_url("bulk_rename"), data, follow=True)
+                expected_message = f"""
+                <div class="alert alert-warning alert-dismissable" role="alert">
+                    <button type="button" class="close" data-dismiss="alert" aria-label="Close">
+                        <span>&times;</span>
+                    </button>
+                    No {verbose_name_plural} were selected.
+                </div>
```
glennmatthews, line 1446: Not sure it's in our best interest to check for the entire HTML block here, as it seems like that'll make the test pretty fragile as we update the UI. Maybe just check for `No {verbose_name_plural} were selected` instead?

nautobot/core/templates/generic/object_bulk_rename.html
```diff
@@ -3,7 +3,7 @@
 {% load form_helpers %}
 
 {% block content %}
-    <h1>{% block title %}Renaming {{ selected_objects|length }} {{ obj_type_plural|bettertitle }}{% endblock %}</h1>
+    <h1>{% block title %}Renaming {{ selected_objects|length }} {{ parent_name }} {{ obj_type_plural|bettertitle }}{% endblock %}</h1>
```
glennmatthews, line 6: Possibly?

```suggestion
    <h1>{% block title %}Renaming {{ selected_objects|length }} {{ obj_type_plural|bettertitle }} on {{ parent_name }}{% endblock %}</h1>
```

nautobot/core/views/generic.py
```diff
@@ -1149,6 +1149,10 @@ def get_required_permission(self):
     def post(self, request):
         logger = logging.getLogger("nautobot.views.BulkRenameView")
 
+        if not request.POST.getlist("pk"):
```
glennmatthews, line 1152: Something I noticed in looking at a different issue with bulk-edit views is that we silently filter out and discard any objects that are in `pk` list but aren't editable by the current user (`selected_objects = self.queryset.filter(pk__in=form.initial["pk"])`). So there is a case where the provided list is non-empty, but the actual operation will be a no-op because no valid objects were included. Might want to handle that case?

nautobot/utilities/testing/views.py
timizuoebideri1: For that to work illd have to use `assertIn` ratter than `assertInHTML`, cause for some reason `assertInHTML` do not work

nautobot/dcim/views.py
```diff
@@ -136,6 +136,11 @@ def post(self, request):
         )
 
 
+class BaseDeviceComponentsBulkRenameView(generic.BulkRenameView):
+    def get_selected_objects_parents_name(self, selected_objects):
+        return selected_objects.first().parent.name
+
+
```
glennmatthews, line 143: This feels like it *should* be possible to include on the base `BulkRenameView` class at this point - that view is AFAIK only used for device-components + VMInterface. I don't think this needs to be a separate subclass.

nautobot/dcim/views.py
```diff
@@ -136,6 +136,11 @@ def post(self, request):
         )
 
 
+class BaseDeviceComponentsBulkRenameView(generic.BulkRenameView):
+    def get_selected_objects_parents_name(self, selected_objects):
+        return selected_objects.first().parent.name
```
HanlinMiao, line 141: We probably need some error checking and catching here. what if `selected_objects` is `None` or `.parent` is `None`. Give user a message to let them define `selected_objects` in their tests?

HanlinMiao, line 141: Should this be `.device `rather than `.parent `since InventoryItem does not have `parent @property `defined, so probably safer to just do `.device` or implement `parent @property` on InventoryItem?

HanlinMiao (approved):

HanlinMiao (changes requested):
I meant to select changes requested, sorry for the mess!

nautobot/dcim/views.py
glennmatthews: Ugh, good catch. In `next` `InventoryItem.parent` isn't the device, it's the parent InventoryItem if any. There's probably a bug [here](https://github.com/nautobot/nautobot/blob/next/nautobot/core/views/generic.py#L1491) as a result of that. You're right, we'll need to explicitly go to `.device` or `.virtual_machine` instead of relying on `.parent`.

glennmatthews (commented):
Failing pylint but otherwise looks good!

nautobot/virtualization/views.py
```diff
@@ -455,6 +455,11 @@ class VMInterfaceBulkRenameView(generic.BulkRenameView):
     queryset = VMInterface.objects.all()
     form = forms.VMInterfaceBulkRenameForm
 
+    def get_selected_objects_parents_name(self, selected_objects):
+        selected_object = selected_objects.first()
+        if selected_object:
+            return selected_object.virtual_machine.name
+
```
timizuoebideri1, line 462:
```suggestion
        if selected_object:
            return selected_object.virtual_machine.name
        return None
```

HanlinMiao (approved):

Issue #2957, opened 2022-12-05
Site, Region -> Location: Change constraints of objects that require Site or Region to now require Location


bryanculver:
- Involves updating both uniqueness constraints and required fields
- Tests will be updated (and broken!)
- Docs update

HanlinMiao:
Closed by https://github.com/nautobot/nautobot/pull/3195

PR #3174, opened 2023-01-26
[Do Not Review Reference Only] Remove Site and Region FKs from model classes and enforce Location FK not null.

HanlinMiao:
# Closes: #2957 
# What's Changed

# TODO

- [ ] Unit, Integration Tests


changes/2957.removed
```diff
@@ -0,0 +1 @@
+Change constraints of objects (CircuitTermination, Device, PowrePanel, PowerFeed, RackGroup, Rack, Prefix, VLAN, VLANGroup, Cluster) that require Site or Region to now require Location.
```
glennmatthews, line 1:
```suggestion
Changed constraints of objects (CircuitTermination, Device, PowerPanel, PowerFeed, RackGroup, Rack, Prefix, VLAN, VLANGroup, Cluster) that required Site or Region to now require Location.
```

nautobot/circuits/api/serializers.py
```diff
@@ -98,17 +97,15 @@ class Meta:
 
 class CircuitCircuitTerminationSerializer(WritableNestedSerializer, NotesSerializerMixin):
     url = serializers.HyperlinkedIdentityField(view_name="circuits-api:circuittermination-detail")
-    site = NestedSiteSerializer()
     location = NestedLocationSerializer(required=False, allow_null=True)
     provider_network = NestedProviderNetworkSerializer()
-    connected_endpoint = NestedInterfaceSerializer()
+    connected_endpoint = NestedCircuitTerminationSerializer()
```
glennmatthews, line 102: Hmm. Actually neither of these is "correct" - `connected_endpoint` is an alias to the generic foreign key `circuit_termination._path.destination`, which could be any subclass of `PathEndpoint`. It looks to me like this serializer should probably inherit from `nautobot.dcim.api.serializers.PathEndpointModelSerializerMixin` rather than implementing its own `connected_endpoint` logic.

changes/2957.removed
glennmatthews, line 1: `.changed` instead of `.removed` perhaps?

nautobot/circuits/filters.py
```diff
@@ -198,7 +177,6 @@ class Meta:
 class CircuitTerminationFilterSet(
     BaseFilterSet,
     CableTerminationModelFilterSetMixin,
-    LocatableModelFilterSetMixin,
```
glennmatthews, line 201: Why remove this and manually declare a `location` filter? Should be able to just update the mixin class instead.

nautobot/core/tests/test_forms.py
```diff
@@ -692,7 +692,6 @@ def test_dynamic_filter_form(self):
                 site_form._get_lookup_field_choices(),
```
glennmatthews, line 692: This whole test will need to be updated to use something other than `site` as its basis, at some point.

nautobot/dcim/forms.py
```diff
@@ -3654,34 +3654,31 @@ class ConnectCableToRearPortForm(ConnectCableToDeviceForm):
 
 class ConnectCableToCircuitTerminationForm(ConnectCableExcludeIDMixin, NautobotModelForm):
     termination_b_provider = DynamicModelChoiceField(queryset=Provider.objects.all(), label="Provider", required=False)
-    termination_b_region = DynamicModelChoiceField(queryset=Region.objects.all(), label="Region", required=False)
-    termination_b_site = DynamicModelChoiceField(
-        queryset=Site.objects.all(),
-        label="Site",
+    termination_b_location = DynamicModelChoiceField(
```
glennmatthews, line 3657: Hm. Something we missed supporting in 1.4/1.5? If we can't specify a Location when connecting a cable to a circuit-termination that's something we may need a backported fix for.

nautobot/circuits/filters.py
HanlinMiao: Yeah I was gonna do this one step at a time, If I remove the Site and Region filters from the mixin class, the tests are just gonna break all at once.

glennmatthews: Gotcha. Sounds good.

HanlinMiao:
see https://github.com/nautobot/nautobot/pull/3195 for the reopened PR

Issue #3187, opened 2023-01-30
DynamicModelChoiceField doesn't properly select empty option


abates:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.4 (problem exists in earlier versions as well)
* Python version: 3.8.15
* Database platform, version: PostgreSQL 13
* Middleware(s): None

When a form includes the `DynamicModelCHoiceField` that can accept an empty value, the UI does not properly show a selected empty option. Consider the following field from an edit form:

```python
vrf = DynamicModelChoiceField(
  empty_label="Global",
  null_option="Global",
  queryset=VRF.objects.all(),
  required=True,
)
```

A null VRF indicates the "global" VRF. When the edit view's object has `None` for the VRF value, then an empty option with the label "Global" should be shown in the UI. However no option is present in this case. Additionally, the form cannot be submitted without first updating the VRF field since the validations indicate a required field is not set. This diminishes the user experience by requiring a user to update a field through the UI when the underlying data field won't actually change.

### Steps to Reproduce
See detailed description above.
1. Go to `IPAM` -> `IP Addresses` -> `Add`
2. View that the VRF default selected option is `---------`

### Expected Behavior
1. Expected default option to be `Global`

![Screen Shot 2023-01-30 at 7 35 08 AM](https://user-images.githubusercontent.com/358901/215478615-c5fae4db-1412-4e99-af8a-2c28923fdb24.png)

### Observed Behavior
![Screen Shot 2023-01-30 at 7 38 30 AM](https://user-images.githubusercontent.com/358901/215479290-91b14c9d-5ebd-484b-8382-967e9a4b8ac5.png)

abates:
This issue appears to be related to the `select2` jquery widget that is used. Empty values are expected to have a value of `null` but when Django renders the HTML the empty option has a string value of `None`. The fix should be to update the `ModelChoiceIterator` to replace `None` values with `null`. I will submit a PR.

PR #3188, opened 2023-01-30
Fixes DynamicModelChoiceField empty placeholder

abates:
# Closes: #3187
# What's Changed

This fixes the behavior of the UI select box for `DynamicModelChoiceField` and `DynamicModelMultipleChoiceField`. With either of those fields, if an `empty_label`/`null_option` were specified, the resulting HTML/Javascript select box would not correctly select the empty option even if the object being edited had an empty/null value for the field.

This change corrects that behavior by converting the Python `None` value to the Javascript `null` value in the field's options.

Before:
Screen Shot 2023-01-30 at 7 38 30 AM

After:
Screen Shot 2023-01-30 at 7 35 08 AM

# TODO

- [ ] ~Attached Screenshots, Payload Example~
- [X] Unit, Integration Tests
- [ ] ~Example Plugin Updates (when adding/changing features)~


gsnider2195:
This looks good so far. I'm wondering why we use a `null_option` argument instead of leveraging some of the functionality of Django's [`empty_label`](https://docs.djangoproject.com/en/3.2/ref/forms/fields/#django.forms.ModelChoiceField.empty_label) argument? They appear to do the same thing unless I'm missing something.

abates:
> This looks good so far. I'm wondering why we use a `null_option` argument instead of leveraging some of the functionality of Django's [`empty_label`](https://docs.djangoproject.com/en/3.2/ref/forms/fields/#django.forms.ModelChoiceField.empty_label) argument? They appear to do the same thing unless I'm missing something.

When I first started looking into this I wondered the same thing. I don't know the history behind the use of `null_option` and I certainly don't know what the impact would be in changing it.

bryanculver:
The UX portion of this has been addressed in #3315. However the breaking change to no longer need the duplicate `null_option` and `empty_label` should be open as a separate issue and could revive a portion of this PR.

PR #3200, opened 2023-01-31
Followup to #3198

glennmatthews:
I realized after we merged #3198 that there's a reference in the docs that needs updating as well.


HanlinMiao (approved):

bryanculver (approved):
Thank you sir. Shame on me for not assuming this required docs updates.

Issue #1362, opened 2022-02-09
Prefix Type


jvanderaa:
### Environment
* Python version: 3.8
* Nautobot version: 1.2.5

### Proposed Functionality

Proposed to add a `type` field to better handle difference Prefix types, rather than holding them in the status field. Today if one wants to have a status on a prefix of active/standby/maintenance/retired or other and is also a container, then the field cannot be changed to reflect the status of the Prefix. Container and pools are "types" of prefixes, and should be it's own type.

### Use Case

Proposed to add a `type` field to better handle difference Prefix types, rather than holding them in the status field. Today if one wants to have a status on a prefix of active/standby/maintenance/retired or other and is also a container, then the field cannot be changed to reflect the status of the Prefix. Container and pools are "types" of prefixes, and should be it's own type.

### Database Changes

* Addition of prefix_type field
* Change of handling on code of prefix status to prefix types

Likely some more changes than just this.

### External Dependencies

None

Referenced in #1309 for discussion.

### TODO

- [x] Add enum property `type` of values: "Network" (the default), "Container", and "Pool"
    - "Aggregate" is a top-level "Container"
        - This will collapse the Aggregate model into a Prefix of type "Container"
    - Any existing prefix that is marked both "Container" and "Pool" in the current setup, "Pool" takes precedence
- [x] Update usage calculation to use the `type` to influence logic for calculation
    - [ ] Split out into `get_prefix_utilization` and `get_ip_address_utilitization`
- [x] Remove `is_pool` field and `container` status

martin-networker:
The thing that container as a status prevents from setting real status to the prefix is one problem.

Second issue is that Nautobot can't properly determine the utilization of a prefix if it is a container.
For example, if prefix 10.1.1.0/24 is a container and has two child prefixes 10.1.1.0/25 which is utilized on 100% and 10.1.1.128/25 which is utilized on 2%, the parent prefix 10.1.1.0/24 shows utilization of 100% instead of 51%.
This leads to a problem of finding available prefix space.
As searching of free prefix is quite difficult in Nautobot we have a job to do that. Unfortunately this job does not work in all cases because of this false utilization enumeration.

bryanculver:
Closed with #3211

bryanculver:
Closed via #3211. Aggregate collapse moved to #3302

Issue #2580, opened 2022-10-06
Improve row checks in ListObjectsViewTestCase.test_list_objects_filtered


glennmatthews:
### Proposed Changes

`ListObjectsViewTestCase.test_list_objects_filtered` and `ListObjectsViewTestCase.test_list_objects_unknown_filter_no_strict_filtering` generic test cases both have logic like the following to check whether the object list view contains or doesn't contain various instances as expected:

```python
            if hasattr(self.model, "name"):
                self.assertIn(instance1.name, content, msg=content)
                self.assertNotIn(instance2.name, content, msg=content)
```

This is fragile as it's scanning the entire HTML response for what may be a very small substring; as we move to have more and more of our test data defined by pseudo-random factories, including ones like `name = factory.Faker("word", part_of_speech="noun")` it's quite possible that generated test data may include text that coincidentally appears elsewhere in the HTML, causing false positive or false negative test results.

One present mitigation is to use capitalization in object names to reduce the likelihood of false matches (e.g. `Red` or `RED` is less likely to appear as a fragment of another word than `red` is) but more broadly it would be good to revisit this pattern and come up with a more robust way to look for specific rows in the table.

An approach I experimented with was to just look at the presence, absence, or count of `<td class="even">` and `<td class="odd">` substrings in the HTML instead, but that ended up failing for certain test cases such as the Job list view tests, since that view has custom HTML rendering. 

Another possibility to consider would be moving these tests from unit to integration, since in integration tests we can use things like xpath lookups rather than naive substring matching...?

### Justification

Improve test robustness.

PR #3211, opened 2023-02-01
Add Prefix.type field

gsnider2195:
# Closes: #1362
# What's Changed

NOTE: `Aggregate` will be removed in a future PR

- Added `Prefix.type` field with choices:
  - `TYPE_CONTAINER`
  - `TYPE_NETWORK`
  - `TYPE_POOL`
- Removed Prefix status `container`
- Removed `Prefix.is_pool` boolean
- Added migration to migrate prefixes with `status=container` to `Prefix.type=TYPE_CONTAINER` and `is_pool=True` to `Prefix.type=TYPE_POOL`
  - If `status=container` and `is_pool=True`, `Prefix.type` is set to `TYPE_POOL`
- All other prefixes will default to `TYPE_NETWORK`

# TODO

- [x] Unit, Integration Tests


nautobot/ipam/factory.py
```diff
@@ -457,12 +457,11 @@ class Params:
...
-    status = factory.Maybe(
+    status = random_instance(lambda: Status.objects.get_for_model(Prefix), allow_null=False)
+    type = factory.Maybe(
         "is_container",
-        factory.LazyFunction(lambda: Prefix.STATUS_CONTAINER),
-        random_instance(
-            lambda: Status.objects.get_for_model(Prefix).exclude(pk=Prefix.STATUS_CONTAINER.pk), allow_null=False
-        ),
+        PrefixTypeChoices.TYPE_CONTAINER,
+        factory.Faker("random_element", elements=PrefixTypeChoices.values()),
```
glennmatthews, line 464: OK to possibly pick `TYPE_CONTAINER` randomly even if `is_container` is false?

nautobot/ipam/migrations/0015_prefix_type.py
```diff
@@ -0,0 +1,68 @@
...
+
+def revert_prefix_type(apps, schema_editor):
+
+    Prefix = apps.get_model("ipam", "Prefix")
+    Status = apps.get_model("extras", "Status")
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    prefix_ct = ContentType.objects.get_for_model(Prefix)
+
+    # Create container status
+    status_container, _ = Status.objects.get_or_create(slug="container", defaults={"name": "Container"})
```
glennmatthews, line 44: do we need a default color here?

nautobot/ipam/migrations/0015_prefix_type.py
```diff
@@ -0,0 +1,68 @@
...
+        prefix_default_status = Status.objects.filter(content_types=prefix_ct).exclude(slug="container").first()
+
+    # Set Prefix.type to container for prefixes with status of container
+    Prefix.objects.filter(status__slug="container").update(
+        status=prefix_default_status,
+        type=choices.PrefixTypeChoices.TYPE_CONTAINER,
+    )
+
+    # Set Prefix.type to pool for prefixes with `is_pool=True`
+    Prefix.objects.filter(is_pool=True).update(type=choices.PrefixTypeChoices.TYPE_POOL)
```
glennmatthews, line 26: Might be a good idea to include a couple of `print()` calls here to tell the user what this migration is doing?

nautobot/ipam/models.py
```diff
@@ -611,7 +604,7 @@ def get_child_prefixes(self):
         Return all Prefixes within this Prefix and VRF. If this Prefix is a container in the global table, return child
         Prefixes belonging to any VRF.
```
glennmatthews, line 605: This is definitely a weird corner case, looking forward to changing this when we have proper prefix parentage...

nautobot/ipam/models.py
```diff
@@ -679,19 +676,23 @@ def get_first_available_ip(self):
     def get_utilization(self):
         """Get the child prefix size and parent size.
 
-        For Prefixes with a status of "container", get the number child prefixes. For all others, count child IP addresses.
+        For Prefixes with a type of "container", get the number child prefixes. For all others, count child IP addresses.
```
glennmatthews, line 679: IMHO as part of the 2.0 changes we should think about splitting this function into a `get_prefix_utilization` and `get_ip_address_utilitization` function.

nautobot/ipam/tests/test_api.py
```diff
@@ -182,7 +182,7 @@ def test_list_available_prefixes(self):
         """
         Test retrieval of all available prefixes within a parent prefix.
         """
-        prefix = Prefix.objects.ip_family(6).filter(prefix_length__lt=128).exclude(status__slug="container").first()
+        prefix = Prefix.objects.ip_family(6).filter(prefix_length__lt=128).exclude(type="container").first()
```
glennmatthews, line 185: Possibly?

```suggestion
        prefix = Prefix.objects.ip_family(6).filter(prefix_length__lt=128).exclude(type=choices.PrefixTypeChoices.TYPE_CONTAINER).first()
```

nautobot/tenancy/tests/test_filters.py
```diff
@@ -455,19 +455,6 @@ def test_vrfs(self):
             self.queryset.filter(vrfs__in=vrfs).distinct(),
         )
 
-    def test_has_vrfs(self):
```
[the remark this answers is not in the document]
gsnider2195: This and the generic test both failed so I went ahead and removed this one because they're identical. #3213 is open to migrate the rest of the tests covered by the new generic filter tests

nautobot/ipam/models.py
```diff
@@ -645,7 +638,11 @@ def get_available_ips(self):
         available_ips = prefix - child_ips
 
         # IPv6, pool, or IPv4 /31-32 sets are fully usable
-        if self.family == 6 or self.is_pool or (self.family == 4 and self.prefix.prefixlen >= 31):
+        if (
+            self.family == 6
+            or self.type == choices.PrefixTypeChoices.TYPE_POOL
+            or (self.family == 4 and self.prefix.prefixlen >= 31)
+        ):
```
jathanism, line 645: IMO when it comes to multiple `or` matches, I find that using `any()` is more readable:
```suggestion
        if any([
            self.family == 6,
            self.type == choices.PrefixTypeChoices.TYPE_POOL,
            self.family == 4 and self.prefix.prefixlen >= 31,
        ]):
```

nautobot/ipam/models.py
```diff
@@ -679,19 +676,23 @@ def get_first_available_ip(self):
...
             return UtilizationData(numerator=child_prefixes.size, denominator=self.prefix.size)
 
         else:
             prefix_size = self.prefix.size
-            if self.prefix.version == 4 and self.prefix.prefixlen < 31 and not self.is_pool:
+            if (
+                self.prefix.version == 4
+                and self.prefix.prefixlen < 31
+                and self.type != choices.PrefixTypeChoices.TYPE_POOL
+            ):
```
jathanism, line 695: Similarly, when it comes to chained `and` matches, I find `all()` to be more readable:

```suggestion
            if all([
                self.prefix.version == 4,
                self.prefix.prefixlen < 31,
                self.type != choices.PrefixTypeChoices.TYPE_POOL,
            ]):
```

nautobot/core/factory.py
```diff
@@ -2,13 +2,28 @@
...
+
+    def clear(self, locale=None):
+        subfaker = self._get_faker(locale)
+        subfaker.clear()
+
+
 class BaseModelFactory(DjangoModelFactory):
     """Base class for all Nautobot model factories."""
 
+    id = UniqueFaker("uuid4")
```
glennmatthews, line 25: Huh. Making the factory UUIDs deterministic? What was the driver for this? I worry that it may mask some bad test practices if we rely on specific UUIDs being present.

nautobot/core/factory.py
```diff
@@ -147,9 +162,25 @@ def get_random_instances(model_or_queryset_or_lambda, minimum=0, maximum=None):
...
+    50/50 distribution of values. Used in factories when a data set must contain both
+    True and False values.
+    """
+
+    def _nautobot_boolean_iterator_sample(self):
+        iterator = [True, True, True, True, False, False, False, False]
+        factory.random.randgen.shuffle(iterator)
+        return iterator
+
+    def __init__(self, *args, cycle=True, getter=None):
```
glennmatthews, line 178: not blocking for this PR, but it'd be sweet to add support for a [`chance_of_getting_true`](https://faker.readthedocs.io/en/stable/providers/faker.providers.misc.html#faker.providers.misc.Provider.boolean) parameter that could be used to weight the scales even further. :-)

nautobot/core/testing/views.py
```diff
@@ -655,8 +656,8 @@ def test_list_objects_filtered(self):
                 # TODO: https://github.com/nautobot/nautobot/issues/2580
                 #       This is fragile as we move toward more autogenerated test fixtures,
                 #       as "instance.name" may appear coincidentally in other page text if we're not careful.
-                self.assertIn(instance1.name, content, msg=content)
-                self.assertNotIn(instance2.name, content, msg=content)
+                self.assertRegex(content, r">\s*" + re.escape(instance1.name) + r"\s*<", msg=content)
+                self.assertNotRegex(content, r">\s*" + re.escape(instance2.name) + r"\s*<", msg=content)
```
glennmatthews, line 660: Do you think this change is sufficient to resolve #2580 for now (and permit removing the above TODO)?

nautobot/dcim/factory.py
```diff
@@ -8,7 +8,13 @@
 from django.contrib.contenttypes.models import ContentType
 from django.db.models import Q
 
-from nautobot.core.factory import OrganizationalModelFactory, PrimaryModelFactory, UniqueFaker, random_instance
+from nautobot.core.factory import (
+    OrganizationalModelFactory,
+    PrimaryModelFactory,
+    UniqueFaker,
+    NautobotBoolIterator,
+    random_instance,
```
glennmatthews, line 16:
```suggestion
    NautobotBoolIterator,
    OrganizationalModelFactory,
    PrimaryModelFactory,
    UniqueFaker,
    random_instance,
```

nautobot/dcim/tests/test_filters.py
```diff
@@ -1119,10 +1119,9 @@ def test_base_site(self):
 
     def test_subtree(self):
         params = {"subtree": [self.loc1.slug, self.nested_loc.pk]}
-        self.assertQuerysetEqualAndNotEmpty(
-            self.filterset(params, self.queryset).qs,
-            Location.objects.get(name=self.loc1.name).descendants(include_self=True),
-        )
+        expected = Location.objects.get(name=self.loc1.name).descendants(include_self=True)
+        expected |= Location.objects.get(name=self.nested_loc.name).descendants(include_self=True)
```
glennmatthews, line 1123: Interesting. Not sure how this was ever passing as previously implemented. :-)

nautobot/docs/rest-api/overview.md
```diff
@@ -441,16 +445,16 @@ GET /api/ipam/prefixes/7d2d24ac-4737-4fc1-a850-b30366618f3d/
     "tenant": null,
     "vlan": null,
     "status": {
-        "value": "container",
-        "label": "Container"
+        "value": "reserved",
+        "label": "Reserved"
     },
```
glennmatthews, line 450: Looks like I missed updating the status representation here and below in #3225. Mind fixing these to use the new NestedStatusSerializer representation?

nautobot/ipam/factory.py
```diff
@@ -9,10 +9,12 @@
     PrimaryModelFactory,
     UniqueFaker,
     get_random_instances,
+    NautobotBoolIterator,
```
glennmatthews, line 12: mind sorting the imports in this list?

nautobot/ipam/models.py
```diff
@@ -19,7 +19,7 @@
 from nautobot.extras.models import RoleModelMixin, Status, StatusModel
 from nautobot.extras.utils import extras_features
 from nautobot.virtualization.models import VirtualMachine, VMInterface
-from .choices import ServiceProtocolChoices
+from nautobot.ipam import choices
```
glennmatthews, line 22: nit: move this line above `from nautobot.virtualization.models ...`?

nautobot/tenancy/factory.py
```diff
@@ -1,6 +1,12 @@
 import factory
 
-from nautobot.core.factory import OrganizationalModelFactory, PrimaryModelFactory, UniqueFaker, random_instance
+from nautobot.core.factory import (
+    OrganizationalModelFactory,
+    PrimaryModelFactory,
+    UniqueFaker,
+    NautobotBoolIterator,
```
glennmatthews, line 7:
```suggestion
    NautobotBoolIterator,
    OrganizationalModelFactory,
    PrimaryModelFactory,
    UniqueFaker,
```

nautobot/virtualization/factory.py
```diff
@@ -1,6 +1,12 @@
 import factory
 
-from nautobot.core.factory import OrganizationalModelFactory, PrimaryModelFactory, UniqueFaker, random_instance
+from nautobot.core.factory import (
+    OrganizationalModelFactory,
+    PrimaryModelFactory,
+    UniqueFaker,
+    NautobotBoolIterator,
```
glennmatthews, line 7:
```suggestion
    NautobotBoolIterator,
    OrganizationalModelFactory,
    PrimaryModelFactory,
    UniqueFaker,
```

jathanism (approved):
No qualms here.

nautobot/core/factory.py
gsnider2195: The only reason I added this was to make sure that my factory changes didn't break the determinism of the factories (which I did on my first couple passes). It also makes the diffs easier to manage if we ever decide to generate fixtures and use those instead of running the factories every time (as in #3255). I'm fine with removing this but I don't know if I agree with leaving it out to enforce better coding practices.

nautobot/dcim/tests/test_filters.py
gsnider2195: Luck 😆

nautobot/core/factory.py
gsnider2195: Do you think it would be useful to be able to explicitly set the number of `True` and `False` values in the list?

glennmatthews: Yeah, that could be another way to achieve the same result. Good thought. I was thinking something probably dumb like:

```python
iterator = [(i < chance_of_getting_true) for i in range(0, 100)]
```

nautobot/core/testing/views.py
gsnider2195: It's probably good enough. Probably needs to be backported to 1.5 though?

jathanism (approved):
:shipit:

nautobot/core/factory.py
gsnider2195: https://github.com/nautobot/nautobot/pull/3211/commits/91f60cf8e5281d976c870aa3df5e200cf3294b1b

gsnider2195: Now I'm thinking it needs tests

glennmatthews: yes please! :-) No good deed, etc. etc.

nautobot/core/testing/views.py
glennmatthews: Wouldn't hurt

glennmatthews (approved):

PR #3232, opened 2023-02-05
[WIP] Typing POC

Kircheneer:
# Closes: #3243 
# What's Changed

- Add a mypy plugin that runs `nautobot.setup()` as a kind of pre-init hook to enable usage of mypy on Nautobot
- Add mypy, django-stubs and things around
- Do some example type hinting for `nautobot.dcim.models.devices`

Interesting files are
- `mypy_nautobot_plugin.py`
- `pyproject.toml`

Run mypy with `inv mypy` - currently there is no setup for running these locally.

# TODO

Everything! This is merely meant to save the (quite straight-forward) how-to of using django-stubs with Nautobot. I don't plan on getting this to through any time soon.


bryanculver:
@Kircheneer Thanks for the prototype. Mind opening a Feature Request for introducing typing so we can evolve the use case and break out the work needed to be done. Obviously this would be large body of work so we will need to carve it out.

Kircheneer:
> Mind opening a Feature Request
#3243

glennmatthews:
I pulled in latest `next` into this branch and updated to the latest (1.1.1) `mypy` release. I note that when running mypy against nautobot.dcim.models.devices, it crashes out due to https://github.com/typeddjango/django-stubs/issues/626 - not sure if this may be a blocking issue for us to proceed here?

Kircheneer:
> I pulled in latest `next` into this branch and updated to the latest (1.1.1) `mypy` release. I note that when running mypy against nautobot.dcim.models.devices, it crashes out due to [typeddjango/django-stubs#626](https://github.com/typeddjango/django-stubs/issues/626) - not sure if this may be a blocking issue for us to proceed here?

That is correct. I tried working on this but I didn't get very far yet.

Kircheneer:
Possible fix for mypy crash coming here: https://github.com/typeddjango/django-stubs/pull/1436

Kircheneer:
Unfortunately the released fix doesn't work.. I will have to re-investigate this

Issue #3224, opened 2023-02-03
Remove Support for 1.X REST APIs in 2.0


bryanculver:
### Proposed Changes
Update the minimum supported REST API version in Nautobot 2.0 to be API version “2.0”, explicitly dropping support for the 1.x REST API versions, rather than trying to retain fully-backwards-compatible versions of all REST API endpoints.

### Justification

The scope of the breaking data model and filtering changes in 2.0 (https://docs.nautobot.com/projects/core/en/v2.0.0-alpha.1/installation/upgrading-from-nautobot-v1/ as a starting point, with more planned before the 2.0.0 initial release) is such that it will involve a tremendous amount of developer work and code to retain support for the 1.x (“1.2”/“1.3"/“1.4”/“1.5") REST API versions in 2.0.

The current 2.0.0-alpha.1 already has changed the schema for numerous REST API endpoints, as seen in the aforementioned document, and as currently implemented those changes generally apply to the “1.x” API versions as well, which is a breaking change to that API in many cases. If we want to retain an unchanged 1.x REST API for backward compatibility in 2.0, we will need to go back and write versioned serializers and views for all of the changed endpoints, writing shims for any removed/renamed/changed fields, etc. More such work will be required as we make more breaking changes to the “2.0" REST API, and some types of changes (e.g. FilterSet changes) may be difficult or impossible to vary across API versions.

Docs that need to be updated to clarify this change:

- https://docs.nautobot.com/projects/core/en/stable/rest-api/overview/#breaking-changes
- https://docs.nautobot.com/projects/core/en/stable/development/#overview
- https://docs.nautobot.com/projects/core/en/stable/development/#deprecation-policy

Issue #824, opened 2021-08-17
Change custom field "name" to a proper "slug" field


glennmatthews:
### Proposed Changes

Follow-on to discussion originally in #464. Custom fields currently have both a `label` (which is meant to be human-readable) and a `name` (which is used internally as a dictionary key, as well as in the GraphQL schema).

Even though the `name` is used in various programmatic and API features, it is currently treated more like the human-readable `name` present on various other models than like a proper `slug` - in particular, it doesn't currently have any of the format enforcement applied to similarly-used `slug` fields in most other models.

The proposal here is to migrate `name` to `slug` in the next major (breaking) release of Nautobot, both for self-consistency and also for the prevention of bugs such as #464.

Note that some of the (non-breaking) groundwork for this change has already been laid in #735; implementing this change should include a thorough review of the various TODO comments added in that PR.

### Justification

- Provide a proper CustomField slug field that can be used safely/correctly in URL patterns, dictionary keys, GraphQL, REST API, etc.
- Reduce the number of redundant "human-readable" fields (name/label/description) currently present on CustomField model.

### TODO

- Update all references in the codebase which previously used a custom field's `name` property to now use the `slug` property (see https://github.com/nautobot/nautobot/issues/824#issuecomment-1315888981)
- Rename `slug` to `key` with some help text to make it more apparent what this field is to be used for
- Remove the custom field's `name` property

Future follow-on: Should `label` on both custom and computed fields now become the "name" of the field, with a call-out that it is the human-readable version

smk4664:
I know there was some interest in standardizing Slug fields across all of Nautobot.  I am curious why we would choose to depreciate the `name` field when that is the field that the Slug is taken from in most of the other models.  Would it make more sense to depreciate the `label` instead?   

I don't think all of this needs to wait till the next major release to introduce all of these changes.  I believe we can safely accomplish most of the slug implementations ahead of time.

In addition to this, I was working on a Data migration the other day, and found the Django System Check Framework, which I would like to add to the chosen depreciated field so we can start alerting users ahead of time of the new field.

glennmatthews:
We could arguably rename `label` to `name` as part of this same migration, that might be a good thing to do. Ultimately we would want this to match other models in having a `name` and a `slug`; the issue is that right now the field called `name` is being used in places where a `slug` would be more appropriate.

jathanism:
Inventory of changes needing to be made to satisfy this based on `2.0 TODO` comments in the code base, broken down by filename:

[...]

PR #3249, opened 2023-02-06
Remove legacy REST API versions

glennmatthews:
# Closes: #3224 
# What's Changed

- Change default REST API version to 2.0
- Remove declared support for REST API versions 1.2 through 1.5
- Remove REST API serializers specific to versions of the API prior to 1.5/2.0
  - Remove viewset versioning that used these older serializers
  - Remove test cases that were specific to older API versions

# TODO

- [x] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features) - need to document changes to API deprecation policy.


gsnider2195:
Does this doc still need to be updated? https://docs.nautobot.com/projects/core/en/stable/development/#deprecation-policy

glennmatthews:
> Does this doc still need to be updated? https://docs.nautobot.com/projects/core/en/stable/development/#deprecation-policy

Yes, my understanding was that @bryanculver was planning to handle that with a separate PR into `develop`.

bryanculver:
> > Does this doc still need to be updated? https://docs.nautobot.com/projects/core/en/stable/development/#deprecation-policy
> 
> Yes, my understanding was that @bryanculver was planning to handle that with a separate PR into `develop`.

I think we should still document which APIs and serializers are being removed as of this PR which hopefully I can get to before EOD.

nautobot/extras/api/customfields.py
```diff
@@ -42,7 +31,7 @@ def __call__(self, serializer_field):
         # Populate the default value for each CustomField
         value = {}
         for field in fields:
-            key = field.slug if use_slug else field.name
+            key = field.name
```
gsnider2195, line 34: Does this need to wait for #824 before switching it back to slug?

glennmatthews: Yes. I initially changed this as `field.slug` here but that made things work incorrectly - need to use `name` here because that's what is still used in `<model>._custom_field_data[...]`

bryanculver:
Should we also address this: https://github.com/nautobot/nautobot/blob/gfm-issue-3224/nautobot/extras/api/views.py#L606-L609 in this PR?

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -133,6 +133,21 @@ The `ipam.Role`, `dcim.RackRole`, and `dcim.DeviceRole` models have been removed
...
+
+- `/api/dcim/interfaces`
+- `/api/ipam/ip-addresses`
+- `/api/extras/custom-fields`
+- `/api/extras/jobs`
+- `/api/extras/tags`
+- `/api/virtualization/interfaces`
+
+Please ensure you are using the latest representations of request/response representations as seen in the API docs or Swagger.
+
```
bryanculver, line 150: @glennmatthews thoughts?

glennmatthews:
> Should we also address these:
> 
>     * https://github.com/nautobot/nautobot/blob/gfm-issue-3224/nautobot/extras/api/views.py#L606-L609
> 
>     * https://github.com/nautobot/nautobot/blob/gfm-issue-3224/nautobot/extras/api/views.py#L640-L644
> 
> 
> in this PR? Or is that separate because it's not necessarily versioned.

We should definitely consider removing them in 2.0, but since those aren't versioned endpoints, I think they're beyond the scope of this PR.

nautobot/docs/installation/upgrading-from-nautobot-v1.md
glennmatthews: Hmm. I worry that people are going to misread this as saying we've removed those API endpoints altogether.

Requesting 1.x API versions of *any* REST API endpoint will fail after this changeset, so I'm not sure what's gained by calling out these ones specifically?

gsnider2195 (approved):

nautobot/docs/installation/upgrading-from-nautobot-v1.md
bryanculver: > Hmm. I worry that people are going to misread this as saying we've removed those API endpoints altogether.
> 
> Requesting 1.x API versions of _any_ REST API endpoint will fail after this changeset, so I'm not sure what's gained by calling out these ones specifically?

My thought was to draw attention to the endpoints vs the version. I suspect integrators know more which endpoints they hit than the versions of which endpoints they hit.

Maybe rephrasing the list heading to something like `Endpoints with Legacy Versions`?

jathanism (approved):

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -133,6 +133,12 @@ The `ipam.Role`, `dcim.RackRole`, and `dcim.DeviceRole` models have been removed
 |----------------|----------------------|--------------------------------------|
 | `/api/status/` | `rq-workers-running` | Removed as RQ is no longer supported |
 
+### Removed 1.X Version Endpoints and Serializer Representations
+
+Nautobot 2.0 removes support for 1.X versioned REST APIs and their Serializers. Requesting [older API versions](../rest-api/overview.md#versioning) will result in a `400 Bad Request` error.
+
+Please ensure you are using the latest representations of request/response representations as seen in the API docs or Swagger.
+
```
bryanculver, line 141: @glennmatthews updated

bryanculver (approved):

Issue #633, opened 2021-06-30
Jobs terminated by Worker report as "Running" forever


glennmatthews:
### Environment
* Python version:
* Nautobot version: 1.0.3

### Steps to Reproduce
1. Run a Job that allocates more memory than is actually available (e.g. with the worker running in a memory-limited Docker container) so that it gets killed by the system.
2. Nautobot JobResult gets stuck as "Running" and never shows as completed in the user UI
3. An admin can access `admin/background-tasks/` and see that the RQ task associated with the Job was killed, but this information is not captured in the JobResult or otherwise visible to non-admin users.

Similar behavior is likely to be seen if a job runs for longer than the configured maximum timeout and is killed by RQ because of that.

I recognize that this specific symptom may be changed as we move to replace RQ with Celery (#531) but these same sorts of scenarios likely need to be accounted for with the Celery worker as well.

### Expected Behavior

Nautobot needs to be made aware when a worker task fails or aborts and update the JobResult accordingly. For RQ, the [docs](https://python-rq.org/docs/exceptions/) have some possible approaches; I'm sure there are similar options for Celery.

jathanism:
Any hunch as to what is actually happening here? Is the kernel scheduler killing it and then the state isn't being accurately reported back to the handling code?

jathanism:
As RQ is deprecated we won't need to solve for that.

We addressed the timeout scenario w/ the Celery workers by implementing soft/hard time limits. But what about if a Celery worker dies/disappears, where are we expecting the call flow to result in the job getting updated by something else? (Such as the connective tissue between the Celery task state and the `JobResult` object).

glennmatthews:
Thinking a bit more about this, this may be something we want to address once #275 / #374 get implemented - have a periodic task that looks for JobResults that are marked as `Running` but aren't actually being used by any task, perhaps?

lampwins:
See #765

bryanculver:
See #1622 as solving this.

jathanism:
As a follow-up even with the work in #1622, when a hard `time_limit` is reached, Celery does a `SIGKILL` (-9) on the worker process executing the long-running task, leaving the task in a hung state. Internally, a `mark_as_failure` event is fired, but the process is killed before the event can culminate and update the backend correctly.

[...]

PR #3254, opened 2023-02-07
Added a system task for handling Jobs where the hard `time_limit` is hit.

jathanism:
# Closes: #633
# What's Changed
**Note: This is not testable because our unit test suite no longer uses `CeleryTestCase` and eager results don't use workers. Additionally, in the previous iteration they wouldn't be either because `CeleryTestCase` used a `solo` worker and time-limits are only supported on the pre-fork worker we use for production deployments.**

- This introduces a `NautobotRequest` for Celery that subclasses the core `Request` object to extend the `on_timeout` handler to fire off a `handle_time_limit_exceeded` task when the hard time limit is hit.
- The new `handle_time_limit_exceeded` task takes a `task_id` and sets the `JobResult.status` to `FAILURE` and also saves a `TimeLimitExceeded` exception as the `traceback`.

## Example job
This is the job I used to test this.

```python
# bunkjob.py
import time

from celery.exceptions import SoftTimeLimitExceeded

from nautobot.extras.jobs import Job


class BunkJob(Job):

    class Meta:
        soft_time_limit = 5
        time_limit = 10
        description = "I sleep for 30 seconds."

    def run(self, data, commit):
        try:
            time.sleep(30)
            self.log_info("Nap complete.")
        except SoftTimeLimitExceeded:
            self.log_warning("Time's up! About to crash.")
            time.sleep(30)
```

## Screenshots
Proof that the job failed after logging "Time's up! About to crash." That the job fails. 
image

Traceback stored on the `JobResult`

image

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/core/celery/task.py
+++ b/nautobot/core/celery/task.py
@@ -1,12 +1,24 @@
 from celery.utils.log import get_task_logger
 from celery import Task
+from celery.worker.request import Request
 
 
 logger = get_task_logger(__name__)
 
 
+class NautobotRequest(Request):
+    def on_timeout(self, soft, timeout):
+        from nautobot.core.tasks import handle_time_limit_exceeded
+
+        if not soft:
+            handle_time_limit_exceeded.delay(self.id)
+        super().on_timeout(soft, timeout)
+
+
 class NautobotTask(Task):
     """Nautobot extensions to tasks for integrating with Job machinery."""
+
+    Request = NautobotRequest
 
 
 Task = NautobotTask  # noqa: So that the class path resolves.
--- a/nautobot/core/tasks.py
+++ b/nautobot/core/tasks.py
@@ -1,6 +1,6 @@
-import logging
-
 from cacheops.simple import CacheMiss, cache
+from celery.utils.log import get_task_logger
+from celery.exceptions import TimeLimitExceeded
 from django.conf import settings
 from packaging import version
 import requests
@@ -8,8 +8,9 @@ import requests
 from nautobot.core import celery
 from nautobot.core.utils import config
 
+
 # Get an instance of a logger
-logger = logging.getLogger("nautobot.releases")
+logger = get_task_logger(__name__)
 
 
 @celery.nautobot_task
@@ -53,3 +54,20 @@ def get_releases(pre_releases=False):
 
     # Since this is a Celery task, we can't return Version objects as they are not JSON serializable.
     return [(str(version), url) for version, url in releases]
+
+
+@celery.nautobot_task(ignore_result=True)
+def handle_time_limit_exceeded(task_id):
+    """
+    Handler for when a running `Job` hits its hard `time_limit` to ensure that
+    the `JobResult` is marked as a failure and the traceback is saved.
+    """
+    from nautobot.extras.choices import JobResultStatusChoices
+    from nautobot.extras.models import JobResult
+
+    jr = JobResult.objects.get(task_id=task_id)
+    # TODO(jathan): Assert this is the correct accessor for `time_limit`
+    time_limit = jr.job_model.time_limit
+    jr.set_status(JobResultStatusChoices.STATUS_FAILURE)
+    jr.traceback = str(TimeLimitExceeded(time_limit))
+    jr.save()
--- a/changes/633.fixed
+++ b/changes/633.fixed
@@ -1,0 +1,1 @@
+Added a system task for handling Jobs where the hard `time_limit` is hit.


nautobot/core/celery/task.py
```diff
@@ -1,13 +1,25 @@
...
 from celery import Task
+from celery.worker.request import Request
 
 
 logger = get_task_logger(__name__)
 
 
+class NautobotRequest(Request):
+    def on_timeout(self, soft, timeout):
+        from nautobot.core.tasks import handle_time_limit_exceeded
```
glennmatthews, line 11: I can see why this inner import is needed (avoiding a circular import loop between core.celery.task and core.tasks), but would it be better to just define `handle_time_limit_exceeded` here in this file and avoid the import?

nautobot/core/tasks.py
```diff
@@ -1,15 +1,16 @@
...
 from packaging import version
 import requests
 
 from nautobot.core import celery
 from nautobot.core.utils import config
 
+
 # Get an instance of a logger
-logger = logging.getLogger("nautobot.releases")
+logger = get_task_logger(__name__)
```
glennmatthews, line 13: What's the function of `get_task_logger` versus `logging.getLogger`?

nautobot/core/tasks.py
```diff
@@ -53,3 +54,20 @@ def get_releases(pre_releases=False):
...
+    the `JobResult` is marked as a failure and the traceback is saved.
+    """
+    from nautobot.extras.choices import JobResultStatusChoices
+    from nautobot.extras.models import JobResult
+
+    jr = JobResult.objects.get(task_id=task_id)
+    # TODO(jathan): Assert this is the correct accessor for `time_limit`
+    time_limit = jr.job_model.time_limit
+    jr.set_status(JobResultStatusChoices.STATUS_FAILURE)
+    jr.traceback = str(TimeLimitExceeded(time_limit))
```
glennmatthews, line 72: Can we get the actual traceback in some fashion? Would be really helpful to know where in the task/job the time limit was hit, for example.

nautobot/core/tasks.py
jathanism: https://docs.celeryq.dev/en/stable/userguide/tasks.html#logging

nautobot/core/tasks.py
jathanism: Here is what the logs look like:

```
[2023-02-06 16:06:20,128: WARNING/MainProcess] Soft time limit (0.5s) exceeded for nautobot.extras.jobs.run_job[3796a944-7220-45d4-9ad5-604f2f62ecde]
16:06:20.178 WARNING nautobot.jobs.BunkJob jobs.py                                    log() :
  Time's up! About to crash.
[2023-02-06 16:06:20,178: WARNING/ForkPoolWorker-8] Time's up! About to crash.
  Performing nautobot serialization on <nautobot.core.utils.requests.NautobotFakeRequest object at 0x126337eb0> for type nautobot.core.utils.requests.NautobotFakeRequest
[2023-02-06 16:06:20,670: ERROR/MainProcess] Task handler raised error: TimeLimitExceeded(1.0)
  File "/Users/jathan/Library/Caches/pypoetry/virtualenvs/nautobot-b2ttWva6-py3.9/lib/python3.9/site-packages/billiard/pool.py", line 684, in on_hard_timeout
    raise TimeLimitExceeded(job._timeout)
billiard.exceptions.TimeLimitExceeded: TimeLimitExceeded(1.0,)
[2023-02-06 16:06:20,677: ERROR/MainProcess] Hard time limit (1.0s) exceeded for nautobot.extras.jobs.run_job[3796a944-7220-45d4-9ad5-604f2f62ecde]
  Performing nautobot serialization on <nautobot.core.utils.requests.NautobotFakeRequest object at 0x126337eb0> for type nautobot.core.utils.requests.NautobotFakeRequest
[2023-02-06 16:06:20,783: INFO/MainProcess] Task nautobot.core.tasks.handle_time_limit_exceeded[68b530db-a0a9-43db-8d75-a7302aabde5b] received
[2023-02-06 16:06:20,786: ERROR/MainProcess] Process 'ForkPoolWorker-8' pid:79723 exited with 'signal 9 (SIGKILL)'
[2023-02-06 16:06:20,841: INFO/ForkPoolWorker-1] Task nautobot.core.tasks.handle_time_limit_exceeded[68b530db-a0a9-43db-8d75-a7302aabde5b] succeeded in 0.05540520799999982s: None
```

- First there is the `SoftTimeLimitExceeded` which we catch and handle.
- Then there is a `TimeLimitExceeded` exception that is raised but it's caught by the `billiard.pool` code and that's what results in the process on the worker being called
- Therefore the context of the exception is lost as far as I can tell
- The `Request.on_timeout()` handler on the database backend makes a call to the [backend for the task to `mark_as_failure()`](https://github.com/celery/celery/blob/main/celery/worker/request.py#L528-L531) but best I can tell here, this call is never made because the process is terminated before it even gets here. Could merely be a timing issue to the database connection, but it requires more R&D to get to the bottom of it.

So... I figured having the exception that resulted in the termination is better than nothing.

nautobot/core/celery/task.py
jathanism: I am thinking I could probably just import the task here. I was going with already established pattern of having tasks in `nautobot.{{app_label}.tasks`, but yeah technically nothing stopping us from just defining it here either.

gsnider2195:
It looks like this is a result of `run_job` clobbering DCR's default behavior [here](https://github.com/nautobot/nautobot/blob/jathanism-celery-handle_hard_timeout/nautobot/extras/jobs.py#L1186-L1190)

I was able to work around this with the following patch on 40e0ad59b8e7d5d5d212a4c26a02378a2c9bc1e7

```diff
diff --git a/nautobot/extras/jobs.py b/nautobot/extras/jobs.py
index c5d5450e9..838af6a69 100644
--- a/nautobot/extras/jobs.py
+++ b/nautobot/extras/jobs.py
@@ -1183,9 +1183,11 @@ def run_job(data, request, job_result_pk, commit=True, *args, **kwargs):
         # `NautobotTask` class.
         finally:
             try:
+                job_result.refresh_from_db()
                 job_result.save()
             except IntegrityError:
                 # handle job_model deleted while job was running
+                job_result.refresh_from_db()
                 job_result.job_model = None
                 job_result.save()
             if file_ids:
```

jathanism:
Replaced by #3273

PR #3255, opened 2023-02-07
Speed up unittests

gsnider2195:
# Closes: #DNE
# What's Changed

- `invoke unittest --cache-test-fixtures` argument added
  - Can also be set permanently in `invoke.yml`
- `--cache-test-fixtures` argument added to management commands `test` and `generate_test_data`
  - defaults to filename `development/factory_dump.json`

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/3255.added
+++ b/changes/3255.added
@@ -1,0 +1,1 @@
+Added --cache-test-fixtures command line argument to nautobot unit and integration tests.
--- a/changes/3255.added
+++ b/changes/3255.added
@@ -1,1 +1,1 @@
-Added --cache-test-fixtures command line argument to nautobot unit and integration tests.
+Added `--cache-test-fixtures` command line argument to Nautobot unit and integration tests.
--- a/nautobot/core/management/commands/generate_test_data.py
+++ b/nautobot/core/management/commands/generate_test_data.py
@@ -26,11 +26,18 @@ class Command(BaseCommand):
             help="Do NOT prompt the user for input or confirmation of any kind.",
         )
         parser.add_argument(
+            "--cache-test-fixtures",
+            action="store_true",
+            help="Save test database to a json fixture file to re-use on subsequent tests.",
+        )
+        parser.add_argument(
             "--fixture-file",
-            help="Dump database to fixture file after generating factory data and re-use on subsequent test runs.",
+            default="development/factory_dump.json",
+            help="Fixture file to use with --cache-test-fixtures.",
         )
 
-    def handle(self, *args, **options):
+    def _generate_factory_data(self, seed):
+
         try:
             import factory.random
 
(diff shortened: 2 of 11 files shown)


nautobot/core/management/commands/generate_test_data.py
```diff
@@ -80,95 +86,111 @@ def handle(self, *args, **options):
...
+
+            if options["fixture_file"]:
+                call_command(
+                    "dumpdata",
+                    "--natural-foreign",
+                    "--natural-primary",
+                    indent=2,
+                    format="json",
+                    exclude=["contenttypes", "auth.permission", "extras.job", "extras.customfield"],
+                    output=options["fixture_file"],
```
jathanism, line 191: Since this is adding a new flag it would be cool if we could include that inside of `tasks.py` for when it runs the tests as another optional flag.

nautobot/core/management/commands/generate_test_data.py
```diff
@@ -80,95 +86,111 @@ def handle(self, *args, **options):
...
-        # TODO: nautobot.tenancy.tests.test_filters currently calls the following additional factories:
-        # UserFactory.create_batch(10)
-        # RackFactory.create_batch(10)
-        # RackReservationFactory.create_batch(10)
-        # ClusterTypeFactory.create_batch(10)
-        # ClusterGroupFactory.create_batch(10)
-        # ClusterFactory.create_batch(10)
-        # VirtualMachineFactory.create_batch(10)
-        # We need to remove them from there and enable them here instead, but that will require many test updates.
+            self.stdout.write("Creating Roles...")
```
glennmatthews, line 96: Let's think about pulling all this out into a separate function just so we don't have this gigantic `else` block to keep track of.

nautobot/core/tests/runner.py
```diff
@@ -47,6 +47,8 @@ def setup_databases(self, **kwargs):
             command = ["generate_test_data", "--flush", "--no-input"]
             if settings.TEST_FACTORY_SEED is not None:
                 command += ["--seed", settings.TEST_FACTORY_SEED]
+            if settings.TEST_FACTORY_FIXTURE_FILE is not None:
```
glennmatthews, line 50: Do we need to explicitly default this setting to `None` in `nautobot/core/settings.py`? I'm thinking probably so?

glennmatthews (commented):
Awesome!

changes/3255.added
```diff
@@ -0,0 +1 @@
+Added --cache-test-fixtures command line argument to nautobot unit and integration tests.
```
glennmatthews, line 1:
```suggestion
Added `--cache-test-fixtures` command line argument to Nautobot unit and integration tests.
```

glennmatthews (approved):

PR #3264, opened 2023-02-08
Changed `DynamicGroup.objects.get_for_object()` to be a little more efficient

jathanism:
# Closes: #3270 
# What's Changed
- Reverse membership query for an object to groups of which it may be a member will now always only be a single-row query vs an N-row query based on the number of members of a particular group.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/querysets.py
+++ b/nautobot/extras/querysets.py
@@ -175,7 +175,7 @@ class DynamicGroupQuerySet(RestrictedQuerySet):
         my_groups = []
         # TODO(jathan): 3 queries per DynamicGroup instance
         for dynamic_group in eligible_groups.iterator():
-            if obj.pk in dynamic_group.members.values_list("pk", flat=True):
+            if dynamic_group.members.filter(pk=obj.pk).values_list("pk", flat=True).exists():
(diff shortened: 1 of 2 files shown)


bryanculver (approved):

nautobot/extras/querysets.py
```diff
@@ -175,7 +175,7 @@ def get_for_object(self, obj):
         my_groups = []
         # TODO(jathan): 3 queries per DynamicGroup instance
         for dynamic_group in eligible_groups.iterator():
-            if obj.pk in dynamic_group.members.values_list("pk", flat=True):
+            if dynamic_group.members.filter(pk=obj.pk).values_list("pk", flat=True).exists():
```
glennmatthews, line 178: I think I asked you this before but I don't remember the answer. :-) Why do we need the values_list call? Can't we just do `dynamic_group.members.filter(pk=obj.pk).exists()`?

jathanism: I'll triple check but I thought it still tries to retrieve all the fields otherwise. Just a query optimization in other words.

glennmatthews (approved):

Issue #2799, opened 2022-11-10
Default Nautobot API Version


itdependsnetworks:
### As ...

Austin - Network Automation Engineer

### I want ...

Use the latest features that are offered by the API

### So that ...

* I may get maximum benefit from the latest features.  
* Know when there are small breaking changes vs only when there are major breaking changes

### I know this is done when...

Without setting a header, I get the latest API functionality given I am using the latest version of code.

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

N/A

### TODO

- Verify DRF's default behavior when `DEFAULT_VERSION` is `None`, expecting it chooses the latest
- Remove `DEFAULT_VERSION` from `settings.py`
- Make it easier via config and documentation to override "default to latest" behavior

bryanculver:
Allowing a "magic" setting of `latest` as well as a specific version to be provided here: https://github.com/nautobot/nautobot/blob/develop/nautobot/core/settings.py#L195 should suffice. We should by default still fall back to the oldest minor version if not provided.

PR #3272, opened 2023-02-08
Updates to deprecation policy.

bryanculver:
# Closes: #3224 
# What's Changed

- Update LTM Overview section detailing what may be dropped per release
- Update Deprecation Policy section to detail deprecated API versions
- Update REST API Versioning to change when deprecated functionality can be expected to be dropped

# TODO


itdependsnetworks:
We spoke about (and perhaps documented in an issue) about defaulting the API to latest. Even if that is not going to be the change in behaviour, can I get an official ruling?

bryanculver:
> We spoke about (and perhaps documented in an issue) about defaulting the API to latest. Even if that is not going to be the change in behaviour, can I get an official ruling?

https://github.com/nautobot/nautobot/issues/2799#issuecomment-1315850641

nautobot/docs/development/index.md
```diff
@@ -152,20 +153,22 @@ If for any reason the next maintenance release is delayed, we will continue to s
...
-    Breaking changes in a `x.0.0` release may have an effect on deprecated APIs. While we do our best to keep the deprecated APIs intact to simplify migrations to newer releases, breaking changes are inherently breaking and will cause some changes to current and previous API versions. Types of breaking changes include, but are not limited to, removing or collapsing of objects and changing object relationships. This will also mean we will not publish deprecated APIs for models that no longer exist.
+    Deprecated APIs will be removed in a `x.0.0` release. Deprecated APIs are kept intact during the same major release to simplify migrations to newer minor releases, however even subtle breaking changes in an `x.0.0` release can cause breaking changes to previous API versions. Therefore it is preferred to remove these API versions entirely to prevent accidental errors when their exact behavior cannot be kept intact.  Types of breaking API changes include, but are not limited to, renaming of fields or filters, removing or collapsing of data models, and changing object relationships. This will also mean we will not publish deprecated APIs for models that no longer exist.
```
glennmatthews, line 166: Probably should be clear that we are talking specifically about the REST API here, I think?

```suggestion
    Deprecated REST APIs will be removed in a `x.0.0` release. Deprecated REST APIs are kept intact during the same major release to simplify migrations to newer minor releases, however even subtle breaking changes in an `x.0.0` release can cause breaking changes to previous API versions. Therefore it is preferred to remove these API versions entirely to prevent accidental errors when their exact behavior cannot be kept intact.  Types of breaking REST API changes include, but are not limited to, renaming of fields or filters, removing or collapsing of data models, and changing object relationships. This will also mean we will not publish deprecated APIs for models that no longer exist.
```

glennmatthews (approved):

PR #3283, opened 2023-02-10
DynamicGroups filter data migration from Site/Region to Location

HanlinMiao:
# What's Changed

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/migrations/0030_migrate_region_and_site_data_to_locations.py
+++ b/nautobot/dcim/migrations/0030_migrate_region_and_site_data_to_locations.py
@@ -242,10 +242,8 @@ def migrate_site_and_region_data_to_locations(apps, schema_editor):
         DynamicGroup = apps.get_model("extras", "DynamicGroup")
         dynamic_groups = DynamicGroup.objects.all()
         for dg in dynamic_groups:
-            if "location" not in dg.filter.keys():
-                dg.filter["location"] = []
-            dg.filter["location"] += dg.filter.get("region", [])
-            dg.filter.pop("region", None)
+            if "region" in dg.filter:
+                dg.filter.setdefault("location", []).extend(dg.filter.pop("region"))
             if not dg.filter["location"]:  # Remove the location key if the list is empty
                 dg.filter.pop("location")
             dg.save()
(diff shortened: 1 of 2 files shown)


nautobot/dcim/migrations/0030_migrate_region_and_site_data_to_locations.py
```diff
@@ -239,6 +239,17 @@ def migrate_site_and_region_data_to_locations(apps, schema_editor):
...
 
+        DynamicGroup = apps.get_model("extras", "DynamicGroup")
+        dynamic_groups = DynamicGroup.objects.all()
+        for dg in dynamic_groups:
+            if "location" not in dg.filter.keys():
+                dg.filter["location"] = []
+            dg.filter["location"] += dg.filter.get("region", [])
+            dg.filter.pop("region", None)
+            if not dg.filter["location"]:  # Remove the location key if the list is empty
+                dg.filter.pop("location")
```
glennmatthews, line 250: Maybe?

```python
if "region" in dg.filter:
    dg.filter.setdefault("location", []).extend(dg.filter.pop("region"))
```

glennmatthews (approved):
Looks valid to me.

This makes me think that it would be a good feature for us to introduce something like a `nautobot-server audit_dynamic_groups` command which would iterate over all DGs and make sure their filters are valid. Because we're not currently fixing up DGs to account for all of the other filtering changes that we're making in 2.0...

nautobot/dcim/migrations/0030_migrate_region_and_site_data_to_locations.py
```diff
@@ -239,6 +239,15 @@ def migrate_site_and_region_data_to_locations(apps, schema_editor):
 
         # Custom Links and Image Attachements do not have Region ContentType as one of its ContentType options
 
+        DynamicGroup = apps.get_model("extras", "DynamicGroup")
+        dynamic_groups = DynamicGroup.objects.all()
+        for dg in dynamic_groups:
+            if "region" in dg.filter:
+                dg.filter.setdefault("location", []).extend(dg.filter.pop("region"))
+            if not dg.filter["location"]:  # Remove the location key if the list is empty
+                dg.filter.pop("location")
```
glennmatthews, line 248: Should these two lines be removed? Same question below.

HanlinMiao: Right right they should, I was thinking that "location" key might be an empty list, but we are already checking the "region" key to ensure the list wont be empty. Sorry!

Issue #3290, opened 2023-02-10
netutils filters are not available in Django templates, only Jinja2 templates


glennmatthews:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.10

### Steps to Reproduce
1. Embed a netutils filter in a Jinja2 template, such as a CustomField's content: `{{ 1000 | bits_to_name }}`
2. It works
3. Embed a netutils filter in a Django template, such as `site.html`: `{{ 1000 | bits_to_name }}`
4. Template rendering error due to unknown/unrecognized filter

### Expected Behavior

#1083 claimed that these filters are supported in both contexts, but only added testing covering their use in Jinja2 templates. 

### Observed Behavior

It looks like the Django case has been broken all this time.

PR #1083, opened 2021-11-17
Add Netutils and register all jinja filters into django-jinja

dgarros:
### Fixes: #1082

This PR adds Netutils and register all jinja filters into django-jinja (~53 filters). The complete list of filter available is available [here](https://github.com/networktocode/netutils/blob/develop/netutils/utils.py#L4) 
As it it's functional but before putting more time into it I would like to understand the level of interest around this PR

> Right now I'm registering the filter as part of the extras app, not sure if this is the best place to do that in Nautobot

```python
### Nautobot interactive shell (e2010bc85319)
### Python 3.6.15 | Django 3.1.13 | Nautobot 1.2.0a1
### lsmodels() will show available models. Use help(<model>) for more info.
>>>
>>>
>>> from nautobot.utilities.utils import render_jinja2
>>> render_jinja2("{{ ip | ip_to_hex }}", {"ip": "1.1.1.1" })
'01010101'
>>>
```

### TODO:
- Update documentation
- Add unit tests

PR #3296, opened 2023-02-13
Add support for netutils in Django templates and add test coverage

glennmatthews:
# Closes: #3290 
# What's Changed

- The way support for netutils was implemented way back in #1083 worked for Jinja2 templates (computed fields, webhooks, export templates, etc) but was wrong for Django. We didn't catch this at the time because #1083 only added automated test coverage for the Jinja2 case.
- Fix the Django case by explicitly importing the netutils functions as a `netutils` templatetags library, making it possible to do `{% load netutils %}` in a Django template then call the netutils functions as template tags.
- Add unit test coverage.
- Update docs.

Example that I hacked in to `site.html` for manual testing purposes:

```django
        {% load netutils %}
        FYI, according to Netutils, 1234567890 bits per second is {% bits_to_name 1234567890 %}
        or {% bits_to_name 1234567890 nbr_decimal=3 %}.
```

Screenshot:

image

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/utilities/tests/test_templatetags_netutils.py
+++ b/nautobot/utilities/tests/test_templatetags_netutils.py
@@ -1,0 +1,52 @@
+import inspect
+
+from django.template import Context, Engine
+from django.template.exceptions import TemplateSyntaxError
+from django.test import TestCase
+
+from netutils.utils import jinja2_convenience_function
+
+
+class NautobotTemplateTagsNetutilsTest(TestCase):
+    """Test the use of netutils functions as Django template filters."""
+
+    def test_netutils_filters_in_django(self):
+        """Verify that each function is at least available as a filter."""
+        engine = Engine.get_default()
+        context = Context({})
+        for filter_name, filter_func in jinja2_convenience_function().items():
+            with self.subTest(f'Testing filter "{filter_name}"'):
+                template_string = "{% load netutils %}{% " + filter_name
+                signature = inspect.signature(filter_func)
+                i = 1
+                for param_name, param in signature.parameters.items():
+                    template_string += f" {param_name}="
+                    if param.annotation == str:
+                        template_string += f'"{i}"'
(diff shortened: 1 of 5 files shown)


nautobot/utilities/tests/test_templatetags_netutils.py
```diff
@@ -0,0 +1,52 @@
...
+                    self.fail(str(exc))
+                except (AttributeError, KeyError, TypeError, ValueError):
+                    # We didn't pass "valid" params to the function, but at least Django saw it as a template tag.
+                    pass
+                except ConnectionError:
+                    # Yes, some netutils functions such as tcp_ping actually make network calls when invoked. Eeek.
+                    pass
+                except Exception:
+                    # Catch-all - at least it wasn't a TemplateSyntaxError, so good enough for now.
+                    pass
```
jathanism, line 52: This is a step forward, but having all these expected error types and passing them through feels a little icky. Without boiling the ocean is there a way we could know which ones we expect to error or not and test those separately with `assertRaises`?

I don't want to block on this. More food for thought than anything.

jathanism (approved):

nautobot/utilities/tests/test_templatetags_netutils.py
glennmatthews: I agree, but it's a quick-and-dirty way to test all of the functions without needing to write separate tests per function (and update those when new netutils releases add more). I actually originally started with just `except TemplateError: fail; except Exception: pass` but tried to add more specific ones to see if I could use `inspect` to automatically provide "correct" inputs. That turned out to be unlikely, in particular because there's lots of functions that take `str` as input but actually expect/enforce that it's a string of an IP address or similar.

The alternate option would be to just test a few specific known functions (since we're using a big `for` loop anyway to load them, either they should all work or none will), which might be preferable?

Issue #3291, opened 2023-02-12
RoleRequiredRoleModelMixin Generates Migrations with `blank=True, null=True`


whitej6:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0a1
* Python version: 3.8
* Database platform, version: postgres
* Middleware(s):

### Steps to Reproduce
1. Create model with `RoleRequiredRoleModelMixin`
2. Run `makemigrations`
3.

### Expected Behavior
`blank=False, null=False` in the migration

### Observed Behavior
`blank=True, null=True` in the migration

jathanism:
It looks like the nomenclature and assertions here need to be fixed: 

https://github.com/nautobot/nautobot/blob/next/nautobot/extras/models/roles.py#L45-L64

1. `RoleModelMixin` is calling `RoleField()` with no arguments. This should be the "required" version and should explicitly call `RoleField(null=True, blank=True)`.
2. `RoleRequiredModelMixin` is subclassing `RoleModelMixin` and instead should just also inherit from `models.Model` similar to `RoleModelMixin`, and should not be calling `RoleField(null=False, blank=False)`.

It appears the naming is backward here and the inheritance should not be done in this way.

glennmatthews:
Implemented in #3301.

PR #3301, opened 2023-02-13
Fixed inheritance and `RoleField` definition on `Role` model mixins.

jathanism:
# Closes: #3291 
# What's Changed

- Updated `RoleModelMixin` to make `RoleField` null and blank explicitly
- Updated `RoleRequiredRoleMixin` base to inherit from `models.Model` and not subclass `RoleModelMixin`

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/models/roles.py
+++ b/nautobot/extras/models/roles.py
@@ -47,13 +47,13 @@ class RoleModelMixin(models.Model):
     Abstract base class for any model which may have roles.
     """
 
-    role = RoleField()
+    role = RoleField(null=True, blank=True)
 
     class Meta:
         abstract = True
 
 
-class RoleRequiredRoleModelMixin(RoleModelMixin):
+class RoleRequiredRoleModelMixin(models.Model):
     """
     Abstract base class for any model which may have roles with role field required.
     """
--- a/changes/3291.fixed
+++ b/changes/3291.fixed
@@ -1,0 +1,1 @@
+Fixed inheritance and `RoleField` definition on `Role` model mixins.


nautobot/extras/models/roles.py
```diff
@@ -47,13 +47,13 @@ class RoleModelMixin(models.Model):
...
 
-    role = RoleField()
+    role = RoleField(null=True, blank=True)
 
     class Meta:
         abstract = True
 
 
-class RoleRequiredRoleModelMixin(RoleModelMixin):
+class RoleRequiredRoleModelMixin(models.Model):
```
glennmatthews, line 56: Why was this problematic, for my education?

nautobot/extras/models/roles.py
```diff
@@ -47,13 +47,13 @@ class RoleModelMixin(models.Model):
     Abstract base class for any model which may have roles.
     """
 
-    role = RoleField()
+    role = RoleField(null=True, blank=True)
```
timizuoebideri1, line 50:
```suggestion
    role = RoleField()
```
`blank` and `null` are set to be `True` by default, Why setting it here again?

jathanism: I was just being explicit. I think perhaps fixing the inheritance is enough to alleviate the reported issue.

nautobot/extras/models/roles.py
jathanism: Subclassing an abstract model to create another abstract model that is overloading the same field is apparently not doing what it was intended to do according to the bug report. I was able to confirm the somewhat confusing behavior.

timizuoebideri1 (approved):

Issue #3302, opened 2023-02-13
Remove `Aggregate` Model, Move to `Prefix`


bryanculver:
Follow-on from #1362.

## TODO
- [ ] Remove Aggregate model and migrate any existing instances to Prefix containers
  - [ ] `network`/`broadcast`/`prefix_length`/`tenant`/`description` copied over to `Prefix`
  - [ ] `rir` -> Nullable FK from Prefix
  - [ ] `date_added` -> `date_allocated` Optional DateTimeField

bryanculver:
Closed with #3305

PR #3305, opened 2023-02-13
Fold aggregate into prefix

gsnider2195:
# Closes: #3302 
# What's Changed

- Remove Aggregate model and migrate any existing aggregates to Prefix

### Migration mismatch output

```
Running migrations:
  Applying ipam.0021_prefix_add_rir_and_date_allocated... OK
  Applying ipam.0022_aggregate_to_prefix_data_migration...
 Unable to migrate all fields from Aggregate 10.0.0.0/8 to Prefix due to an existing Prefix. Some of the following data may have to be migrated manually: {'description': 'Aggregate Date added 2023-02-15 with green and blue tags', 'type': 'container'}

 Unable to migrate all fields from Aggregate 172.16.0.0/24 to Prefix due to an existing Prefix. Some of the following data may have to be migrated manually: {'description': 'Aggregate Date added 2023-02-15 with green and blue tags', 'type': 'container'}

 Unable to migrate all fields from Aggregate 192.168.0.0/24 to Prefix due to an existing Prefix. Some of the following data may have to be migrated manually: {'description': 'Aggregate Date added 2023-02-15 with green and blue tags', 'type': 'container'}

 Unable to migrate all fields from Aggregate 192.168.100.128/25 to Prefix due to an existing Prefix. Some of the following data may have to be migrated manually: {'description': 'Aggregate with 2 CFs set, 1 will mismatch existing prefix', 'type': 'container', '_custom_field_data': {'aggregate_and_prefix_cf': 'Aggregate and Prefix CF set on the Aggregate'}}
 OK
```

#### Before

image

#### After

image

#### Before

image

#### After

image

#### Before

image

#### After

image

#### Before

image

#### After

image

#### Before

image

#### After

image

#### Before

image

#### After

image

#### Before

image

#### After

image

#### Before

image

#### After

image

# TODO

- [x] Unit, Integration Tests
- [x] Add a ForeignKey from the Aggregate to the migrated Prefix for Apps to use during migration of Aggregate ForeignKeys
- [x] Re-use existing Aggregate PK on newly created Prefixes
- [x] Migrate tags from Aggregate to Prefix
- [x] Migrate notes from Aggregate to Prefix
- [x] Migrate object permissions from Aggregate to Prefix
- [x] Migrate object changes from Aggregate to Prefix
- [x] Migrate custom fields from Aggregate to Prefix
- [x] Migrate computed fields from Aggregate to Prefix
- [x] Migrate custom relationships from Aggregate to Prefix
- [x] Migrate custom links from Aggregate to Prefix


nautobot/ipam/factory.py
```diff
@@ -38,162 +39,6 @@ class Meta:
...
-        # Create child prefixes, preserving tenant and is_ipv6 from aggregate
-        for count, subnet in enumerate(self.prefix.subnet(prefix_cidr)):
-            if count == max_count:
-                break
-            method(prefix=str(subnet.cidr), is_ipv6=is_ipv6, **kwargs)
-
-    @factory.lazy_attribute_sequence
-    def prefix(self, n):
-        """
-        Yes, this is probably over-complicated - but it's realistic!
```
glennmatthews, line 145: RIP my overengineered code. :-)

nautobot/ipam/migrations/0020_delete_aggregate.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("ipam", "0019_aggregate_to_prefix_data_migration"),
+    ]
+
+    operations = [
+        migrations.DeleteModel(
```
glennmatthews, line 13: While I think the app "blast radius" of removing Aggregate is likely much smaller than that of removing Region and Site, the same set of concerns apply - if an app has data models that depend on Aggregate, we need a documented and tested solution for how they can cleanly migrate to using Prefix in its place.

nautobot/ipam/models.py
```diff
@@ -469,18 +314,22 @@ class Prefix(PrimaryModel, StatusModel, RoleModelMixin):
         "vlan",
         "status",
         "role",
+        "rir",
+        "date_allocated",
```
glennmatthews, line 318: Probably need to update `to_csv` to include these fields too?

nautobot/ipam/tests/test_models.py
```diff
@@ -112,33 +112,6 @@ def test_get_db_prep_value_mysql(self):
         self.assertEqual(prepped, manual)
 
 
-class TestAggregate(TestCase):
-    def test_get_utilization(self):
```
glennmatthews, line 116: Should any of this test content port over to TestPrefix.test_get_utilization?

nautobot/ipam/views.py
```diff
@@ -150,7 +148,7 @@ class RouteTargetBulkDeleteView(generic.BulkDeleteView):
 
 
 class RIRListView(generic.ObjectListView):
-    queryset = RIR.objects.annotate(aggregate_count=count_related(Aggregate, "rir"))
+    queryset = RIR.objects.annotate(prefix_count=count_related(Prefix, "rir"))
```
glennmatthews, line 151: Maybe a TODO comment here? Aggregate count is pretty self-explanatory; prefix count is a bit ambiguous as to whether it's only counting top-level prefixes, counting prefixes at any level that have an assigned RIR, or counting all such prefixes and their descendants. We should disambiguate this at some point.

glennmatthews (commented):
Great start!

nautobot/core/testing/filters.py
```diff
@@ -42,7 +44,10 @@ def get_filterset_test_values(self, field_name, queryset=None):
                     break
                 if value[field_name] and value["count"] < qs_count:
                     qs_count -= value["count"]
-                    test_values.append(str(value[field_name]))
+                    if isinstance(value[field_name], (datetime.date, datetime.time)):
+                        test_values.append(localize_input(value[field_name]))
```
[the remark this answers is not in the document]
gsnider2195: It's misleading. What it's effectively doing for date/time/datetimes is converting them to strings that django forms will accept:

```py
    elif isinstance(value, datetime.datetime):
        value = datetime_safe.new_datetime(value)
        format = default or get_format('DATETIME_INPUT_FORMATS')[0]
        return value.strftime(format)
```

nautobot/ipam/tests/test_models.py
gsnider2195: Most of this logic is already covered there but we could add one of the larger networks to make sure we don't overflow ints?

nautobot/core/tests/test_factory.py
```diff
@@ -0,0 +1,25 @@
...
+from nautobot.core import testing
+from nautobot.core.factory import NautobotBoolIterator
+
+
+class TestFactory(factory.DictFactory):
+    attribute1 = NautobotBoolIterator(cycle=False, chance_of_getting_true=80, length=10)
+    attribute2 = NautobotBoolIterator()
+
+
+class FactoryTestCase(testing.TestCase):
```
jathanism, line 14: Docstrings please. Not clear on what these are for?

nautobot/ipam/migrations/0020_delete_aggregate.py
jathanism: As we had discussed yesterday, having a guide that explains using `run_before` in migrations would be the golden ticket.

nautobot/ipam/tests/test_models.py
```diff
@@ -347,6 +321,50 @@ def test_get_utilization(self):
...
+        )
+        self.assertEqual(large_prefix_v6.get_utilization(), (2**118, 2**120))
+
+        # 50% utilization
+        Prefix.objects.bulk_create((Prefix(prefix=netaddr.IPNetwork("ab40::/10")),))
+        self.assertEqual(large_prefix_v6.get_utilization(), (2**119, 2**120))
+
+        # 100% utilization
+        Prefix.objects.bulk_create((Prefix(prefix=netaddr.IPNetwork("ab80::/9")),))
+        self.assertEqual(large_prefix_v6.get_utilization(), (2**120, 2**120))
```
glennmatthews, line 366: It occurs to me that this *may* be problematic when we get around to reporting utilization in the revamped UI, as IIRC Javascript has problems with numbers larger than 2**53.

gsnider2195: We might have to return a string for the value and a percentage for the utilization chart?

nautobot/ipam/views.py
gsnider2195: I'd like to fix it in this PR. I think the only place it's really exposed to the user is on the RIR serializer:

```json
        {
            "id": "b8c65d37-c37c-4db3-81cd-c00610b6f002",
            "display": "AFRINIC",
            "url": "http://localhost:8080/api/ipam/rirs/b8c65d37-c37c-4db3-81cd-c00610b6f002/",
            "name": "AFRINIC",
            "slug": "afrinic",
            "is_private": false,
            "description": "Performance people west need themselves measure ask. Perhaps fine exactly grow everybody seem second.\nThird perhaps glass article day including. Treatment site recent.",
            "prefix_count": 4123,
            "created": "2023-02-15",
            "last_updated": "2023-02-15T16:53:05.790659Z",
            "notes_url": "http://localhost:8080/api/ipam/rirs/b8c65d37-c37c-4db3-81cd-c00610b6f002/notes/",
            "custom_fields": {}
        },
```

What about `related_prefix_count`?

nautobot/core/tests/test_factory.py
```diff
@@ -0,0 +1,28 @@
...
+        attribute1 = NautobotBoolIterator(cycle=False, chance_of_getting_true=80, length=10)
+        attribute2 = NautobotBoolIterator()
+
+    def test_nautobot_bool_iterator(self):
+        batch = FactoryTestCase.TestDictFactory.build_batch(10)
+        attribute1 = [obj["attribute1"] for obj in batch]
+        attribute2 = [obj["attribute2"] for obj in batch]
+        with self.subTest(r"NautobotBoolIterator, 25% chance of True, length 10"):
+            self.assertCountEqual(attribute1, list(itertools.repeat(True, 8)) + list(itertools.repeat(False, 2)))
+        with self.subTest(r"NautobotBoolIterator, 50% chance of True, length 8"):
```
glennmatthews, line 24: `8` is the default `length` of a NautobotBoolIterator? That feels a bit arbitrary and not ideal to have hard-coded in the test - maybe need a constant declared somewhere?

nautobot/core/tests/test_graphql.py
```diff
@@ -1470,39 +1471,65 @@ def test_query_frontport_filter_second_level(self):
...
+                query = "query { devices{ id, frontports(" + filterv + "){ id }}}"
                 result = self.execute_query(query)
                 self.assertIsNone(result.errors)
-                self.assertEqual(len(result.data["devices"][0]["frontports"]), nbr_expected_results)
+                for device in result.data["devices"]:
+                    qs = FrontPort.objects.filter(device_id=device["id"])
+                    expected_count = qs.filter(qs_filter).count()
+                    matched = max(matched, len(device["frontports"]))
+                    self.assertEqual(len(device["frontports"]), expected_count)
+                self.assertNotEqual(matched, 0, msg="At least one object matched GraphQL query")
```
glennmatthews, line 1499: nit: maybe `assertGreater` rather than `assertNotEqual` here and below?

nautobot/core/tests/test_graphql.py
```diff
@@ -1620,43 +1674,81 @@ def test_query_interfaces_filter_second_level(self):
...
+                query = "query { devices{ id, interfaces(" + filterv + "){ id }}}"
                 result = self.execute_query(query)
                 self.assertIsNone(result.errors)
-                self.assertEqual(len(result.data["devices"][0]["interfaces"]), nbr_expected_results)
+                for device in result.data["devices"]:
+                    qs = Interface.objects.filter(device_id=device["id"])
+                    expected_count = qs.filter(qs_filter).count()
+                    matched = max(matched, len(device["interfaces"]))
+                    self.assertEqual(len(device["interfaces"]), expected_count)
+                self.assertNotEqual(matched, 0, msg="At least one object matched GraphQL query")
```
glennmatthews, line 1710: Might consider moving all of this repeated logic into a helper function, something that could be called like:

```python
self.validate_graphql_filters(filters, query_start="query { devices{ id, interfaces(", query_end="){ id }}}")
```

?

nautobot/docs/core-functionality/ipam.md
```diff
@@ -1,9 +1,5 @@
 # IP Address Management
 
-{%
-    include-markdown "../models/ipam/aggregate.md"
-    heading-offset=1
-%}
 {%
```
glennmatthews, line 3: Might consider adding a `--- 2.0.0` comment to this document noting that the Aggregate model was removed? Or is that covered in this page by your notes added to `prefix.md` and `rir.md` do you think?

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -87,6 +88,8 @@
 
 ### Replaced Models
 
+#### Generic Role Model
```
glennmatthews, line 91: Misleading heading - this section covers both Role and Location changes. Might need to split this into two sections and carve up the "Removed Model/Replaced With" table between those sections if so.

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -99,6 +102,22 @@ The `dcim.Region` and `dcim.Site` models have been removed and replaced by `dcim
 | `dcim.Site`       | `dcim.Location`|
 | `ipam.Role`       | `extras.Role`  |
 
+#### Aggregate Migrated to Prefix
+
+The `ipam.Aggregate` model has been removed and all existing aggregates will be migrated to `ipam.Prefix` with `type` set to "Container". The `Aggregate.date_added` field will be migrated to `Prefix.date_allocated` and changed from a Date field to a DateTime field with the time set to `00:00`. `Aggregate.tenant`, `Aggregate.rir` and `Aggregate.description` will be migrated over to the same fields on `Prefix`.
```
glennmatthews, line 107:
```suggestion
The `ipam.Aggregate` model has been removed and all existing Aggregates will be migrated to `ipam.Prefix` records with their `type` set to "Container". The `Aggregate.date_added` field will be migrated to `Prefix.date_allocated` and changed from a Date field to a DateTime field with the time set to `00:00` UTC. `Aggregate.tenant`, `Aggregate.rir` and `Aggregate.description` will be migrated over to the equivalent fields on the new `Prefix`.
```

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -99,6 +102,22 @@ The `dcim.Region` and `dcim.Site` models have been removed and replaced by `dcim
...
+
+If a `Prefix` already exists with the same network and prefix length as a previous `Aggregate`, the `rir` and `date_added` fields will be copied to the existing prefix object. Messages will be output during migration (`nautobot-server migrate` or `nautobot-server post-upgrade`) if the `tenant`, `description` or `type` fields need to be manually migrated.
```
glennmatthews, line 109:
```suggestion
If a `Prefix` already exists with the same network and prefix length as a previous `Aggregate`, the `rir` and `date_added` fields will be copied to the `rir` and `date_allocated` fields on the existing Prefix object. Messages will be output during migration (`nautobot-server migrate` or `nautobot-server post_upgrade`) if the `tenant`, `description` or `type` fields need to be manually migrated.
```

nautobot/docs/models/ipam/rir.md
```diff
@@ -2,6 +2,9 @@
...
 
-Users can create whatever RIRs they like, but each aggregate must be assigned to one RIR. The RIR model includes a boolean flag which indicates whether the RIR allocates only private IP space.
+Users can create whatever RIRs they like and optionally assign prefixes to an RIR. The RIR model includes a boolean flag which indicates whether the RIR allocates only private IP space.
 
-For example, suppose your organization has been allocated 104.131.0.0/16 by ARIN. It also makes use of RFC 1918 addressing internally. You would first create RIRs named "ARIN" and "RFC 1918," then create an aggregate for each of these top-level prefixes, assigning it to its respective RIR.
+For example, suppose your organization has been allocated 104.131.0.0/16 by ARIN. It also makes use of RFC 1918 addressing internally. You would first create RIRs named "ARIN" and "RFC 1918," then create a prefix for each of these top-level networks, assigning it to its respective RIR.
+
++/- 2.0.0
+    The `Aggregate` model and all relationships to `RIR` were migrated to the `Prefix` model.
```
glennmatthews, line 10:
```suggestion
    The `Aggregate` model and its relationships to `RIR` were migrated to the `Prefix` model.
```

nautobot/docs/release-notes/version-2.0.md
```diff
@@ -206,6 +206,12 @@ Existing prefixes with a status of "Container" will be migrated to the "Containe
 
 The "Container" status will be removed and all prefixes will be migrated to the "Active" status if it exists. If the "Active" status was deleted, prefixes will be migrated to the first available prefix status in the database that is not "Container".
 
+#### Aggregate model Migrated to Prefix ([#3302](https://github.com/nautobot/nautobot/issues/3302))
```
glennmatthews, line 209: Nit: I'd prefer for us to keep the subsections under `### Changed` in alphabetical order just so we have some sort of consistent pattern. Looks like several of the existing subsections are already violating this pattern but let's at least not make it worse :-)

nautobot/ipam/factory.py
```diff
@@ -449,6 +296,8 @@ class Params:
         factory.SubFactory(VRFGetOrCreateFactory, tenant=factory.SelfAttribute("..tenant")),
         None,
     )
+    rir = factory.Maybe("has_rir", random_instance(RIR, allow_null=False), None)
```
glennmatthews, line 299: Do we have any easy way to tell whether this is a root-level prefix or not? Would it be more "realistic" (maybe??) if only root-level prefixes get RIRs?

nautobot/ipam/filters.py
```diff
@@ -236,12 +205,20 @@ class PrefixFilterSet(
         field_name="vlan__vid",
         label="VLAN number (1-4095)",
     )
+    rir = NaturalKeyOrPKMultipleChoiceFilter(
```
glennmatthews, line 208: Our upgrading guide currently doesn't list *net-new* filter fields. Wonder if it should? Those are of course less impactful than the renamed/removed/changed filter fields when upgrading, but maybe we should include for completeness?

nautobot/ipam/forms.py
```diff
@@ -360,8 +283,16 @@ class Meta:
             "description",
             "tenant_group",
             "tenant",
+            "rir",
+            "date_allocated",
             "tags",
         ]
+        help_texts = {
+            "rir": "Regional Internet Registry responsible for this prefix",
```
glennmatthews, line 291: Should this help_text be defined directly on the Prefix model instead of here in the form?

nautobot/ipam/forms.py
```diff
@@ -503,6 +438,11 @@ class PrefixFilterForm(
         null_option="Global",
     )
     present_in_vrf_id = DynamicModelChoiceField(queryset=VRF.objects.all(), required=False, label="Present in VRF")
+    type = forms.MultipleChoiceField(
+        required=False,
+        choices=PrefixTypeChoices,
+        widget=StaticSelect2Multiple(),
+    )
```
glennmatthews, line 445: Should we also add an RIR field to the FilterForm? (I don't know whether we're going to get rid of FilterForms entirely in 2.0 at this point?)

nautobot/ipam/forms.py
```diff
@@ -471,7 +407,6 @@ class PrefixFilterForm(
         "role",
         "tenant_group",
         "tenant",
```
glennmatthews, line 409: Do we need to add `type` and/or `rir` to the field_order?

nautobot/ipam/migrations/0020_aggregate_to_prefix_data_migration.py
```diff
@@ -0,0 +1,81 @@
...
+    Prefix = apps.get_model("ipam", "Prefix")
+    Aggregate = apps.get_model("ipam", "Aggregate")
+    Status = apps.get_model("extras", "Status")
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    prefix_ct = ContentType.objects.get_for_model(Prefix)
+
+    try:
+        prefix_default_status = Status.objects.get(content_types=prefix_ct, slug="active")
+    except ObjectDoesNotExist:
+        prefix_default_status = Status.objects.filter(content_types=prefix_ct).first()
```
glennmatthews, line 22: Corner case, but should we explicitly check for the case where there are *no* prefix Statuses?

nautobot/ipam/migrations/0020_aggregate_to_prefix_data_migration.py
```diff
@@ -0,0 +1,81 @@
...
+
+    for instance in Aggregate.objects.all():
+        # convert date to datetime
+        date_allocated = instance.date_added
+        if date_allocated is not None:
+            date_allocated = datetime.combine(date_allocated, time.min)
+            if settings.USE_TZ:
+                date_allocated = make_aware(date_allocated)
```
glennmatthews, line 30: I'll turn your question back around at you  :-) Does this make for 00:00 local time, or 00:00 UTC?

nautobot/ipam/tests/test_api.py
```diff
@@ -114,55 +113,31 @@ def get_deletable_object_pks(self):
...
             },
             {
                 "prefix": "2001:db8:abcd:12::/80",
                 "status": cls.status_active.pk,
+                "rir": rir.pk,
             },
             {
                 "prefix": "192.168.6.0/24",
                 "status": cls.status_active.pk,
+                "rir": rir.pk,
```
glennmatthews, line 140: Let's have one of these with no `rir` (to validate that it's an optional field) and at least one that specifies a `date_allocated`?

glennmatthews: Probably also should add some `type` values in here as well...

nautobot/ipam/tests/test_filters.py
```diff
@@ -168,58 +166,14 @@ def test_is_private(self):
...
-        params = {"rir": [rirs[0].slug, rirs[1].slug]}
-        self.assertQuerysetEqual(self.filterset(params, self.queryset).qs, self.queryset.filter(rir__in=rirs))
-
-
 class PrefixTestCase(FilterTestCases.FilterTestCase, FilterTestCases.TenancyFilterTestCaseMixin):
     queryset = Prefix.objects.all()
     filterset = PrefixFilterSet
     tenancy_related_name = "prefixes"
     generic_filter_tests = (
+        ["date_allocated"],
```
glennmatthews, line 174: Can/should we test the `__gte` and `__lte` filters for this field as well?

nautobot/ipam/templates/ipam/aggregate_list.html
```diff
@@ -1,14 +0,0 @@
...
-{% load humanize %}
-
-{% block sidebar %}
-    <div class="panel panel-default">
-        <div class="panel-heading">
-            <strong><i class="mdi mdi-chart-areaspline"></i> Statistics</strong>
-        </div>
-        <ul class="list-group">
-            <li class="list-group-item">Total IPv4 IPs <span class="badge">{{ ipv4_total|intcomma }}</span></li>
-            <li class="list-group-item">Total IPv6 /64s <span class="badge">{{ ipv6_total|intcomma }}</span></li>
```
glennmatthews, line 11: Should we consider (if even possible before the other IPAM model changes) having something like this added to the Prefix list view?

nautobot/ipam/views.py
glennmatthews: Here at least it is a column in the RIR table in the UI as well. Maybe `assigned_prefix_count`?

nautobot/ipam/views.py
```diff
@@ -193,118 +191,11 @@ class RIRBulkImportView(generic.BulkImportView):
...
-            if aggregate.prefix.version == 6:
-                # Report equivalent /64s for IPv6 to keep things sane
-                ipv6_total += int(aggregate.prefix.size / 2**64)
-            else:
-                ipv4_total += aggregate.prefix.size
-
-        return {
-            "ipv4_total": ipv4_total,
-            "ipv6_total": ipv6_total,
-        }
```
glennmatthews, line 235: Would this make sense to add similar to the Prefix list view?

nautobot/ipam/views.py
```diff
@@ -361,11 +252,6 @@ class PrefixView(generic.ObjectView):
     )
```
glennmatthews, line 252: Probably should add `"rir"` to the `select_related`? Same on `PrefixListView._queryset` above.

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -310,7 +329,6 @@ Their filters are also being replaced by `?location=<uuid/slug>`. For example `/
 
 | Model              | Removed Filter Field  | UI and API endpoints that are no longer supported in v2.X                                     |
 |--------------------|-----------------------|-----------------------------------------------------------------------------------------------|
-| Aggregate          | `tenant_group_id`     |                                                                                               |
```
glennmatthews, line 313: Probably need to document removal of `Tenant` `aggregates` and `has_aggregates` filters here.

glennmatthews (commented):
Sorry for the big pile o' comments - this really is looking very solid :-)

nautobot/core/testing/filters.py
```diff
@@ -42,7 +44,10 @@ def get_filterset_test_values(self, field_name, queryset=None):
                     break
                 if value[field_name] and value["count"] < qs_count:
                     qs_count -= value["count"]
-                    test_values.append(str(value[field_name]))
+                    if isinstance(value[field_name], (datetime.date, datetime.time)):
```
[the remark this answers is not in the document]
gsnider2195: `datetime` is a subclass of `date` so it is covered by this

nautobot/docs/core-functionality/ipam.md
gsnider2195: This file is just embedding all of the other files so I don't think it would belong here. I believe it's covered with the changes in `prefix.md`, `rir.md` and `migrating-from-netbox.md`

nautobot/ipam/forms.py
gsnider2195: It's showing up automatically in the dynamic groups filter form and the advanced filter form. Is there anywhere else that would benefit from adding it to the simple filter form?

glennmatthews: Just the "non-advanced" filter tab, AFAIK.

nautobot/ipam/migrations/0020_aggregate_to_prefix_data_migration.py
gsnider2195: If I don't set a time zone it gets set to midnight UTC:

```
nautobot=# select date_allocated from ipam_prefix;
     date_allocated     
------------------------
 2023-02-21 00:00:00+00
(1 row)
```

If I set `NAUTOBOT_TIME_ZONE="America/Chicago"` it gets set to midnight in Chicago (but still stores it in the DB as UTC)
```
nautobot=# select date_allocated from ipam_prefix;
     date_allocated     
------------------------
 2023-02-21 06:00:00+00
(1 row)
```

That seems acceptable to me?

glennmatthews: It's a bit inconsistent with #3316, but as long as we document it clearly I think it's probably fine.

nautobot/ipam/factory.py
gsnider2195: I updated `PrefixFactory` to set `has_rir=False` on the child prefixes created by `PrefixFactory.children`

nautobot/ipam/tests/test_filters.py
gsnider2195: I see tests for those filter in the generic tests in `BaseFilterSetTest`. Should we write a generic test to test the generated lookup expression filters on all filtersets? (`n`, `ie`, `nie`, `gt`, `gte`, etc)

nautobot/core/tests/test_graphql.py
gsnider2195: We would have to pass in the `query` from line 1702, the `qs` from line 1706 and the key names in lines 1708-1709. I tried combining the logic and it made the tests more confusing to read and only saved a few lines of code. I recommend leaving them as is for now.

nautobot/ipam/views.py
gsnider2195: Maybe another issue to add this back in? The sidebar got ripped out in 1.5 but adding the sidebar back in is probably going to require some redesign since there's no search form anymore. Ex:

### v1.4
image

### v2.0
image

nautobot/ipam/views.py
```diff
@@ -161,19 +159,19 @@ class RIRView(generic.ObjectView):
 
     def get_extra_context(self, request, instance):
 
-        # Aggregates
-        aggregates = Aggregate.objects.restrict(request.user, "view").filter(rir=instance).select_related("tenant")
+        # Prefixes
+        prefixes = Prefix.objects.restrict(request.user, "view").filter(rir=instance).select_related("tenant")
 
-        aggregate_table = tables.AggregateTable(aggregates)
+        prefix_table = tables.PrefixTable(prefixes)
```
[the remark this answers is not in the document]
gsnider2195: What do you think about "Allocated Prefixes"/`allocated_prefix_table` to align with our `allocated_date` on the `Prefix` model?

nautobot/ipam/tests/test_filters.py
HanlinMiao: I think we should and it would be great. But might be out of scope for this issue

nautobot/ipam/tests/test_forms.py
```diff
@@ -46,24 +46,18 @@ def test_address_missing_mask(self):
...
     field_name = "prefix"
     object_name = "prefix"
 
     def setUp(self):
         super().setUp()
-        self.extra_data = {"status": Status.objects.get(slug="active"), "type": "network"}
+        self.extra_data = {
+            "status": Status.objects.get(slug="active"),
+            "type": "network",
+            "rir": models.RIR.objects.first(),
```
HanlinMiao, line 59: Should we add "date_allocated" here as well?

nautobot/ipam/migrations/0022_aggregate_to_prefix_data_migration.py
```diff
@@ -0,0 +1,235 @@
...
+        # migrate all object changes related to this aggregate to the new prefix
+        _migrate_aggregate_object_changes_to_prefix(apps, instance, aggregate_ct, prefix_ct)
+
+        # migrate all tags related to this aggregate to the new prefix
+        _migrate_aggregate_tags_to_prefix(apps, instance, aggregate_ct, prefix_ct)
+
+        # migrate custom fields on this aggregate to the new prefix
+        mismatches.update(_migrate_aggregate_custom_fields_to_prefix(instance))
+
+        # migrate all relationships related to this aggregate to the new prefix
```
HanlinMiao, line 195: This might be a bit more accurate? Same with the function name
```suggestion
        # migrate all relationship associations related to this aggregate to the new prefix
```

HanlinMiao (approved):
Looks good, Non-blocking but adding DataMigrationTest might be helpful :)

nautobot/ipam/views.py
jathanism: Seeing as we're completely overhauling the UI, I don't think we should spend too much time making the existing v1 UI vibe in v2 alpha.

nautobot/core/tests/test_factory.py
```diff
@@ -0,0 +1,34 @@
...
+class FactoryTestCase(testing.TestCase):
+    """Test Nautobot factory base classes."""
+
+    class TestDictFactory(factoryboy.DictFactory):
+        """Dummy factory to run tests that don't require testing database functionality."""
+
+        attribute1 = factory.NautobotBoolIterator(cycle=False, chance_of_getting_true=80, length=10)
+        attribute2 = factory.NautobotBoolIterator()
+
+    def test_nautobot_bool_iterator(self):
```
jathanism, line 19: Docstring pls.

nautobot/core/tests/test_migration.py
```diff
@@ -0,0 +1,40 @@
+from django.apps import apps
```
jathanism, line 1: Did you add this? I recall there being something @HanlinMiao did with this class. IMO this should be in a library called `migrations.py` since it's publishing a base class.

jathanism (commented):
This one is a doozy. I'm taking it in pieces.

jathanism (changes requested):
Meant to request changes, not comment on my last one.

nautobot/core/tests/test_migration.py
gsnider2195: This is pulled over from Hanlin's branch. I think this should go in `nautobot.core.testing.migrations`

HanlinMiao: I just changed the test to be in `nautobot.core.testing.migrations`

jathanism (approved):

Issue #3203, opened 2023-01-31
UI 2.0: Create Build Pipeline for Core (Generates UI on startup/post-upgrade)


bryanculver:
This translates as hiding `npm` behind a management command (e.g. `nautobot-server build`) and then having it as something that can then be called by other commands such as `post_upgrade`. (See similar example from NSoT project embedding this into `setup.py build`: https://github.com/dropbox/nsot/blob/develop/setup.py#L66-L90)

Webpack should be configured to deploy files to the `STATICFILES_DIRS` which defaults to `project-static` in the `js` directory? (See: https://pascalw.me/blog/2020/04/19/webpack-django.html):

> 1. Have Webpack write to a directory in `STATICFILES_DIRS`
> 2. Configure output.publicPath to match Django's `STATIC_URL`
> 3. Don't let Webpack hash filenames (except for chunks, see below)
> 4. Make sure `webpack-dev-server` writes files to disk (so Django can serve them in development)

```[tasklist]
### Tasks
- [x] Create new `build` management command
- [x] Configure web pack to deploy bundles to `project-static/js`
- [x] Assert that `build` is called by `post_upgrade`
```

bryanculver:
Closed with #3306

Issue #3205, opened 2023-01-31
UI 2.0: Common Components Library, Think `nautobot-ui` as a Package


bryanculver:

bryanculver:
Created, can be found here: https://github.com/nautobot/nautobot-ui (still private until we get baseline components merged).

PR #3306, opened 2023-02-14
Added a `nautobot-server build` command

jathanism:
# Closes: #3203 
# What's Changed
- This command does the following from the `nautobot_ui` directory
  - Generates the `plugin_imports.js` file from any installed plugins
  - Runs `npm install`
  - Runs `npm run build`
- This command is now also called by `nautobot-server post_upgrade`
- Updated the `config-overrides.js` to not hash JS or CSS files so that their raw filename can be stored in static files and referenced using `{% static 'js/main.ss %}` without having to overload the `base.html` template for the web UI
- Updated `nautobot/core/templates/base.html` to link directly to the static files for JS and CSS that get pulled in from the UI build environment by `nautobot-server collectstatic` 
- Added a basic README file back to `nautobot_ui`
- Parameterized the API key needed for now to read from `NAUTOBOT_API_TOKEN`
- `plugin_imports.js` now renders with relative paths vs. absolute paths
- Restored the `/plugins/` route and the `InstalledPlugins.jsx` module

# TODO

- [ ] Unit, Integration Tests


nautobot/core/management/commands/build.py
```diff
@@ -0,0 +1,112 @@
...
+
+            # TODO: Assert that a plugin has a UI folder.
+            jsconfig["compilerOptions"]["paths"][f"@{plugin_name}/*"] = [f"{plugin_path}/ui/*"]
+
+        with open(jsconfig_file_path, "w", encoding="utf-8") as generated_config_file:
+            json.dump(jsconfig, generated_config_file, indent=4)
+
+        plugin_imports_final_file_path = os.path.join(NAUTOBOT_UI_DIR, "src", "plugin_imports.js")
+
+        environment = jinja2.sandbox.SandboxedEnvironment(loader=jinja2.FileSystemLoader(NAUTOBOT_UI_DIR))
```
jathanism, line 69: Forgot to add that I made sure this was using sandboxed env.

nautobot_ui/README.md
```diff
@@ -0,0 +1,29 @@
...
+nautobot-server build
+```
+
+Configure Nautobot to pass through authentication over CORS by putting this into your `nautobot_config.py`:
+
+```python
+CORS_ALLOW_ALL_ORIGINS = True
+CORS_ALLOW_CREDENTIALS = True
+SESSION_COOKIE_SAMESITE = None
+```
```
jathanism, line 17: Methinks this is no longer necessary since the web UI has basic (even if somewhat flaky) session authentication.

nautobot_ui/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+Set the `NAUTOBOT_API_TOKEN` prior to starting Node.js:
```
jathanism, line 19:
```suggestion
Set the `NAUTOBOT_API_TOKEN` prior to starting the server:
```

nautobot/core/management/commands/build.py
```diff
@@ -0,0 +1,112 @@
...
+            abs_plugin_path = pathlib.Path(app_config.path).resolve()
+            plugin_path = os.path.relpath(abs_plugin_path, NAUTOBOT_UI_DIR)
+
+            # TODO: Assert that a plugin has a UI folder.
+            jsconfig["compilerOptions"]["paths"][f"@{plugin_name}/*"] = [f"{plugin_path}/ui/*"]
+
+        with open(jsconfig_file_path, "w", encoding="utf-8") as generated_config_file:
+            json.dump(jsconfig, generated_config_file, indent=4)
+
+        plugin_imports_final_file_path = os.path.join(NAUTOBOT_UI_DIR, "src", "plugin_imports.js")
```
glennmatthews, line 67: Should this write directly to STATICFILES_DIR or to a tempfile instead? I'm worried that there may be cases where the src directory is not writable by the user.

nautobot/core/management/commands/build.py
```diff
@@ -0,0 +1,112 @@
...
+
+        environment = jinja2.sandbox.SandboxedEnvironment(loader=jinja2.FileSystemLoader(NAUTOBOT_UI_DIR))
+        template = environment.get_template("plugin_imports.js.j2")
+
+        content = template.render(plugins=settings.PLUGINS)
+
+        with open(plugin_imports_final_file_path, "w", encoding="utf-8") as generated_import_file:
+            generated_import_file.write(content)
+
+        pathlib.Path(router_file_path).touch()
```
glennmatthews, line 77: Similar concern here.

nautobot/core/management/commands/build.py
```diff
@@ -0,0 +1,112 @@
...
+        if options.get("render_plugins"):
+            self.render_plugin_imports()
+
+        # Try to run the build command. Make sure that if it errors, it prints
+        # something like "Hey couldn't find npm, did you install it?"
+        if options.get("npm_install"):
+            self.run_command("npm install", ">>> Installing Nautobot dependencies...")
+
+        if options.get("npm_build"):
+            self.run_command("npm run build", ">>> Compiling Nautobot UI...")
```
glennmatthews, line 112: Should we print a success message at the end of this function?

nautobot_ui/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+```no-highlight
+export NAUTOBOT_API_TOKEN=nnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnn
```
glennmatthews, line 22: Presumably you need to actually create this API token as well... wonder if we should override `createsuperuser` to give the option to provision a token too?

glennmatthews (commented):
Looks like a great step forward!

nautobot/core/management/commands/build.py
jathanism: I was thinking the same thing, but I think this is a good first pass.

We need to think about that. I want to iterate on this to make it something tenable in a production environment.

nautobot/core/management/commands/build.py
jathanism: This touch is more of a development thing I carried over for the moment.

nautobot/core/management/commands/build.py
jathanism:
```
>>> 🥳 UI BUILD COMPLETE 🎉
```

nautobot/core/management/commands/post_upgrade.py
```diff
@@ -11,6 +11,7 @@
 
 - migrate
 - trace_paths
+- build
```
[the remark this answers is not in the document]
jathanism: I dunno, if there were ever another reason to `build`, perhaps?

nautobot_ui/README.md
jathanism: I'm actually thinking we should have a `Token` created by default for any new user. Simplify.

nautobot_ui/README.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+Start Nautobot on port `8888/tcp`:
+
+```no-highlight
+nautobot-server runserver 0.0.0.0:8888 --insecure
```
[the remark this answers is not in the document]
jathanism: We can change it to whatever we want, so yes.

nautobot_ui/public/index.html
```diff
@@ -1,3 +1,6 @@
+{% load static %}
```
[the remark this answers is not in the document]
jathanism: Yes, this is now redundant. It is required if we're going to let Webpack hash the files and update the `index.html` to point to them. With Django managing static files we can now take advantage of `{% static %}` and this file can go away.

nautobot_ui/src/views/InstalledPlugins.jsx
```diff
@@ -0,0 +1,7 @@
+function init_plugins() {
+    return (
+        <></>
```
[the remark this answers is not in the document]
jathanism: Still blank but I distinctly remember it working before. I think we broke it somewhere.

nautobot_ui/src/components/core/Plugins.jsx
```diff
@@ -23,16 +23,20 @@ function get_components() {
     for (const [plugin_name, import_promise] of Object.entries(NautobotPlugins)) {
         import_promise.then((value) => {
             if (value?.default?.view_overrides) {
+                // eslint-disable-next-line
```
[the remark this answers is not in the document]
jathanism: It wants return values from nested functions. The correct fix is what I did below in `nautobot_ui/src/views/generic/ObjectRetrieve.jsx`. Instead of updating a list as you iterate, returning a value from the `map` calls.

```
src/components/core/Plugins.jsx
  Line 26:83:  Array.prototype.map() expects a return value from arrow function  array-callback-return
  Line 27:74:  Array.prototype.map() expects a return value from arrow function  array-callback-return
  Line 34:95:  Array.prototype.map() expects a return value from arrow function  array-callback-return
  Line 36:48:  Array.prototype.map() expects a return value from arrow function  array-callback-return
```

nautobot_ui/src/views/BSCreateViewTemplate.jsx
```diff
@@ -20,8 +20,8 @@ export default function BSCreateViewTemplate({ list_url }) {
   if (!data) return <></>
 
   const post_schema = data.serializer.schema
-  const ui_schema = data.serializer.uiSchema
-  const log = (type) => console.log.bind(console, type)
+  // const ui_schema = data.serializer.uiSchema
+  // const log = (type) => console.log.bind(console, type)
```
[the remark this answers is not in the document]
jathanism: Linting fixes since these weren't used. I wanted to keep them around "for now" since this code is still janky.

nautobot_ui/public/index.html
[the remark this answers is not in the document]
jathanism: Looks like we could just disable the `HtmlWebpackPlugin` perhaps to not even care about pulling in and rendering this file.

nautobot/core/management/commands/build.py
jathanism: Just digging in, all the other JS source is here in `src` and so dropping this file there makes sure that it gets bundled. But there's still the issue of "where should `src` live in a production deployment?".

nautobot/core/management/commands/post_upgrade.py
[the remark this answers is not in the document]
glennmatthews: Well at the very least there's `poetry build` to compile the wheel...

nautobot_ui/README.md
glennmatthews: Hmm, not sure I'd agree with that - especially if we are wanting to change things so that an admin cannot retrieve/view the tokens of other users, how would the newly created token get communicated to the user in a more typical workflow?

nautobot/core/management/commands/build.py
jathanism: `src` is now bundled with the distribution in `nautobot/ui/src` so this file will be updated when the UI is compiled.

nautobot/ui/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "nautobot-ui",
```
glennmatthews, line 2: Looks like `nautobot-ui` is likely to be the name of the library added by #3205... let's brainstorm what we might want to call this instead?

nautobot/ui/package.json
```diff
@@ -1,6 +1,6 @@
...
-  "version": "0.1.0",
+  "version": "2.0.0a1",
```
glennmatthews, line 3: We might want to add a `tasks.py` wrapper for `poetry version ...` that makes sure this gets updated in parallel when updating the version in `pyproject.toml`.

nautobot/core/templates/base.html
```diff
@@ -29,15 +32,13 @@
...
+    <script defer src="{% static 'js/main.js' %}"></script>
+    <link href="{% static 'css/main.css' %}" rel="stylesheet">
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
     <div id="root"></div>
     <div id="legacy-content" style="display:none">
-      {% load static %}
-      {% load helpers %}
-      {% load plugins %}
```
bryanculver, line 40: This (and the line below it) allows us to still render tab views for objects. I know we discussed forcing the tab views to be implemented without being complete views but knowing that this will prevent the block below from rendering content.

nautobot/ui/package.json
bryanculver: Agreed. We can just call it `nautobot` since it's the compiled app and won't be published to NPM (although we should squat the name).

nautobot/ui/jsconfig.json
```diff
@@ -21,7 +21,7 @@
                 "src/components/core/*"
             ],
             "@example_plugin/*": [
-                "/source/examples/example_plugin/example_plugin/ui/*"
+                "../../examples/example_plugin/example_plugin/ui/*"
```
bryanculver, line 24: TODO Template out

nautobot/core/management/commands/build.py
bryanculver: Wonder if there is a Pylint rule for this...

nautobot/core/management/commands/build.py
```diff
@@ -0,0 +1,138 @@
...
+
+        # Generate `app_imports.js`
+        if options["render_apps"]:
+            self.render_app_imports()
+
+        # Run `npm install` and keep it silent by default.
+        if options["npm_install"]:
+            args = "install"
+            if verbosity <= 1:
+                args += " -s --no-progress"
```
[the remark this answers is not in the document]
bryanculver: Also I think we may need to move this. I think we should only run install in a post_upgrade but in all other cases we don't want/need to install. Apps aren't "installed" they are just linked from where they live. They don't packages themselves as NPM packages (as seen by the example).

In the future we will want to allow apps to bring in packages but I think that's out of scope.

bryanculver (changes requested):
I'd like to see the `npm install` not always being run on build. More dangerous IMHO and likely won't be needed. That could be better served as a `post_upgrade` step.

jathanism:
> I'd like to see the `npm install` not always being run on build. More dangerous IMHO and likely won't be needed. That could be better served as a `post_upgrade` step.

There's a flag to disable it `--no-npm-install`, it just runs by default. Are you saying to decouple them?

bryanculver:
> > I'd like to see the `npm install` not always being run on build. More dangerous IMHO and likely won't be needed. That could be better served as a `post_upgrade` step.
> 
> There's a flag to disable it `--no-npm-install`, it just runs by default. Are you saying to decouple them?

Yes.

nautobot/ui/jsconfig.json
bryanculver: This is baking the example plugin by default but that assumes the example plugin is installed and enabled. It shouldn't exist unless it's defined in `PLUGINS`

jathanism:
I inverted the toggle. So now it's `build_ui --npm-install` explicitly or it won't run.

changes/3203.added
```diff
@@ -0,0 +1 @@
+Added a `nautobot-server build` command for building the web UI
```
glennmatthews, line 1:
```suggestion
Added a `nautobot-server build_ui` command for building the web UI.
```

nautobot/core/management/commands/build_ui.py
```diff
@@ -0,0 +1,141 @@
...
+        verbosity = options["verbosity"]
+
+        # NAUTOBOT_DEBUG is also set here so that `console.log` isn't suppressed when we want
+        # verbosity.
+        if verbosity > 1:
+            os.environ["NAUTOBOT_DEBUG"] = "1"
+
+        verbosity_map = {
+            0: "silent",
+            1: "silent",  # Default is "notice" but still too chatty, so "silent".
```
glennmatthews, line 121: What about `warn` or `error`? Still too chatty?

nautobot/core/management/commands/post_upgrade.py
```diff
@@ -85,30 +93,36 @@ def handle(self, *args, **options):
 
         # Run trace_paths
         if options.get("trace_paths"):
-            print("Generating cable paths...")
+            self.stdout.write("Generating cable paths...")
             call_command("trace_paths", no_input=True)
             print()
```
glennmatthews, line 98: Should these prints also be changed to self.stdout.write()?

nautobot/ui/README.md
```diff
@@ -0,0 +1,21 @@
+# Nautobot v2 UI Prototype
+
+## Running it
+
+Install and compile dependencies:
+
+```no-highlight
+nautobot-server build
```
glennmatthews, line 8:
```suggestion
nautobot-server build_ui
```

Do we need to include `--npm-install` on the first invocation of this command?

nautobot/ui/jsconfig.json
glennmatthews: Is this another file that should be gitignored and always regenerated from jsconfig-base.json?

nautobot/ui/public/index.html
```diff
@@ -29,14 +32,13 @@
...
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
     <div id="root"></div>
     <div id="legacy-content" style="display:none">
-      {% load static %}
-      {% load helpers %}
-      {% load plugins %}
       {% block content %}{% endblock content %}
     </div>
     <!--
```
glennmatthews, line 44: Can/should the logo files below be symlinked to their equivalents in `nautobot/project-static`? Not blocking for this PR.

nautobot/core/management/commands/build_ui.py
jathanism: Warn/error are less than notice and yea, still too chatty for the default. I can demonstrate.

nautobot/core/management/commands/post_upgrade.py
jathanism: Yes.

nautobot/ui/README.md
jathanism: Yes. Thanks.

nautobot/ui/public/index.html
jathanism: Yeah kind of. We have a bigger fish to to fry when it comes to having Django serve the static files which includes images, using the manifest static files storage. So we'll cover that then.

Issue #3274, opened 2023-02-09
Circuits export does not respect current filter


Thetacz:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.9
* Python version: 3.10.9
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Filter circuits in UI
2. Export found results
3. Check downloaded csv

### Expected Behavior
Only circuits shown in filtered results are exported.

### Observed Behavior
All circuits regardless of filter are exported.

glennmatthews:
Thanks for the report! Looks like some missing logic in the `ObjectListViewMixin` as compared to the legacy `ObjectListView`.

PR #3307, opened 2023-02-14
Fixed ObjectListViewMixin's filtering when exporting objects in NautobotUIViewSet.

HanlinMiao:
# Closes: #3274
# What's Changed

Moved `get_filter_params()` from `renders.py` to `mixins.py` so that the export feature will respect the filters applied from the request.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/core/views/mixins.py
+++ b/nautobot/core/views/mixins.py
@@ -41,6 +41,7 @@ from nautobot.utilities.forms import (
 from nautobot.core.views.renderers import NautobotHTMLRenderer
 from nautobot.utilities.utils import (
     csv_format,
+    get_filterable_params_from_filter_params,
     prepare_cloned_fields,
 )
 from nautobot.utilities.views import GetReturnURLMixin
@@ -78,6 +79,7 @@ class NautobotViewSetMixin(GenericViewSet, AccessMixin, GetReturnURLMixin, FormV
     serializer_class = None
     table_class = None
     notes_form_class = NoteForm
+    nonefilter_params = None
 
     def get_permissions_for_model(self, model, actions):
         """
@@ -304,6 +306,11 @@ class NautobotViewSetMixin(GenericViewSet, AccessMixin, GetReturnURLMixin, FormV
 
         return obj
 
+    def get_filter_params(self, request):
+        """Helper function - take request.GET and discard any parameters that are not used for queryset filtering."""
+        filter_params = request.GET.copy()
+        return get_filterable_params_from_filter_params(filter_params, self.non_filter_params, self.filterset_class)
+
     def get_queryset(self):
         """
         Get the list of items for this view.
@@ -436,9 +443,28 @@ class ObjectListViewMixin(NautobotViewSetMixin, mixins.ListModelMixin):
         "sort",  # table sorting
     )
 
+    def get_filtered_queryset_for_export(self):
+        """
+        Helper method for NautobotUIViewSet to recognize request's filter_params when exporting.
+        Since the export response cannot be captured from NautobotHTMLRenderer,
+        We have to do it here in the ObjectListViewMixin class.
+        """
+        queryset = self.get_queryset()
+        filter_params = self.get_filter_params(self.request)
+        if self.filterset_class is not None:
+            filterset = self.filterset_class(filter_params, self.queryset)
+            queryset = filterset.qs
+            if not filterset.is_valid():
+                messages.error(
+                    self.request,
+                    mark_safe(f"Invalid filters were specified: {filterset.errors}"),
+                )
+                queryset = queryset.none()
+        return queryset
+
     def check_for_export(self, request, model, content_type):
         # Check for export template rendering
-        queryset = self.get_queryset()
+        queryset = self.get_filtered_queryset_for_export()
         if request.GET.get("export"):
             et = get_object_or_404(
                 ExportTemplate,
@@ -473,7 +499,7 @@ class ObjectListViewMixin(NautobotViewSetMixin, mixins.ListModelMixin):
         """
         Export the queryset of objects as concatenated YAML documents.
         """
-        queryset = self.get_queryset()
+        queryset = self.get_filtered_queryset_for_export()
         yaml_data = [obj.to_yaml() for obj in queryset]
 
         return "---\n".join(yaml_data)
@@ -482,7 +508,7 @@ class ObjectListViewMixin(NautobotViewSetMixin, mixins.ListModelMixin):
         """
         Export the queryset of objects as comma-separated value (CSV), using the model's to_csv() method.
         """
-        queryset = self.get_queryset()
+        queryset = self.get_filtered_queryset_for_export()
         csv_data = []
         custom_fields = []
         # Start with the column headers
--- a/nautobot/core/views/mixins.py
+++ b/nautobot/core/views/mixins.py
@@ -25,6 +25,7 @@ from rest_framework.response import Response
 from rest_framework.viewsets import GenericViewSet
 
 from drf_spectacular.utils import extend_schema
+
 from nautobot.core.api.views import BulkCreateModelMixin, BulkDestroyModelMixin, BulkUpdateModelMixin
 from nautobot.extras.models import CustomField, ExportTemplate
 from nautobot.extras.forms import NoteForm
@@ -69,6 +70,9 @@ class NautobotViewSetMixin(GenericViewSet, AccessMixin, GetReturnURLMixin, FormV
     logger = logging.getLogger(__name__)
     lookup_field = "slug"
     # Attributes that need to be specified: form_class, queryset, serializer_class, table_class for most mixins.
+    # filterset and filter_params will be initialized in filter_queryset() in ObjectListViewMixin
+    filter_params = None
+    filterset = None
     filterset_class = None
     filterset_form_class = None
     form_class = None
@@ -78,7 +82,6 @@ class NautobotViewSetMixin(GenericViewSet, AccessMixin, GetReturnURLMixin, FormV
     serializer_class = None
     table_class = None
     notes_form_class = NoteForm
-    nonefilter_params = None
 
     def get_permissions_for_model(self, model, actions):
         """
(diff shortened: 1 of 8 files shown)


nautobot/utilities/testing/views.py
```diff
@@ -747,6 +749,30 @@ def test_list_objects_with_permission(self):
...
+                # This block of code is for checking the content of the csv output.
+                # But it does not work for some of the models.
+                # csv_headers = getattr(self.model, "csv_headers")
+                # model_field = (
+                #     "name" or "slug" or "username"
+                #     if ("name" or "username" or "slug") in csv_headers
+                #     else csv_headers[0]
+                # )
+                # self.assertIn(str(getattr(instance1, model_field)), response_body)
+                # self.assertIn(str(getattr(instance2, model_field)), response_body)
```
HanlinMiao, line 775: I need some help figuring out a way to check the content of the csv response based on `instance1` attributes for all models.

timizuoebideri1:
```python
reader = csv.DictReader(StringIO(response_body))
data = [dict(row) for row in reader]


# Sample response
# [{"name": "Device 1", ...}, ...]
```
I believe with this format you can check what ever you need to check

timizuoebideri1: Note: `instance1_csv_data` do not account for custom fields, which might be found in `data`

nautobot/core/views/mixins.py
```diff
@@ -436,9 +443,28 @@ class ObjectListViewMixin(NautobotViewSetMixin, mixins.ListModelMixin):
         "sort",  # table sorting
     )
 
+    def get_filtered_queryset_for_export(self):
```
jathanism, line 446: This seems like you're reinventing `filter_queryset()` which is something inherited from `GenericViewSet` that is a base for NautobotViewSetMixin`. 

See: https://github.com/encode/django-rest-framework/blob/master/rest_framework/generics.py#L140-L151

Just hoping to simplify here. Please correct me if I'm wrong!

HanlinMiao: Thanks for pointing this out! Using it directly seems to cause some problems because the `export` query string passed in from the request isn't recognized as a filter for `NautobotFilterBackEnd` So I always get this error.
But it seems to be a good method to override and use.
```
  File "/usr/local/lib/python3.7/site-packages/django/test/utils.py", line 387, in inner
    return func(*args, **kwargs)
  File "/source/nautobot/utilities/testing/views.py", line 803, in test_queryset_to_csv
  File "/source/nautobot/utilities/testing/mixins.py", line 167, in assertHttpStatus
    self.assertIn(response.status_code, expected_status, err_message)
AssertionError: 400 not found in [200] : Expected HTTP status(es) [200]; received 400: {'export': [ErrorDetail(string='Unknown filter field', code=None)]}
```

nautobot/utilities/testing/views.py
```diff
@@ -799,6 +795,30 @@ def test_list_view_plugin_banner(self):
...
+                reader = csv.DictReader(StringIO(response_body))
+                data = [dict(row) for row in reader]
+                csv_headers = getattr(self.model, "csv_headers")
+                for _, entry in enumerate(data):
+                    # x might be None, in that case use an empty sting
+                    instance1_csv_data = [str(x) if x is not None else "" for x in instance1.to_csv()]
+                    for header in csv_headers:
+                        # ignore "", None and etc, we only want "name", "slug" or similar identifiers
+                        if entry[header]:
+                            self.assertIn(entry[header], instance1_csv_data)
```
timizuoebideri1, line 820:
```suggestion
                data = [dict(row) for row in reader][0]
                instance1_csv_data = [str(x) if x is not None else "" for x in instance1.to_csv()]
                # append instance
                instance1_ct_values = list(instance1.get_custom_fields().values())
                instance1_csv_data += instance1_ct_values
                # Since values in `data` are all in str; cast all values in instance1_csv_data to str
                instance1_csv_data = [str(val) for val in instance1_csv_data]
                self.assertEquals(instance1_csv_data, list(data.values()))
```
Since this approach takes into account customs types, I think this might be good to consider. I also don't see the need to obtain the csv headers, unless you want to add a test to determine whether the `instance` csv headers matches the response csv header, which I think would also be a reasonable thing to test for.

nautobot/core/views/mixins.py
```diff
@@ -25,7 +25,6 @@
 from rest_framework.viewsets import GenericViewSet
 
 from drf_spectacular.utils import extend_schema
-
```
glennmatthews, line 28: Personal preference (non-blocking) would be to restore this line, I like keeping Nautobot imports separate from other package imports.

nautobot/core/views/mixins.py
```diff
@@ -78,6 +78,7 @@ class NautobotViewSetMixin(GenericViewSet, AccessMixin, GetReturnURLMixin, FormV
     serializer_class = None
     table_class = None
     notes_form_class = NoteForm
+    nonefilter_params = None
```
glennmatthews, line 81: Unused?

nautobot/core/views/renderers.py
```diff
@@ -163,18 +156,11 @@ def get_context(self, data, accepted_media_type, renderer_context):
             form = data["form"]
         else:
             if view.action == "list":
-                filter_params = self.get_filter_params(view, request)
                 if view.filterset_class is not None:
+                    view.queryset = view.filter_queryset(view.queryset)
+                    filter_params = view.get_filter_params(request)
                     filterset = view.filterset_class(filter_params, view.queryset)
```
glennmatthews, line 162: Seems a bit awkward that we instantiate `filterset` at least twice in this flow (once in `view.filter_queryset` and once here)... wonder if there's any cleaner way we could do this? Maybe setting a `view.filterset` attribute or something?

Similar nit about calling `get_filter_params` twice.

nautobot/extras/models/datasources.py
```diff
@@ -106,6 +106,7 @@ def to_csv(self):
             self.slug,
             self.remote_url,
             self.branch,
+            self.secrets_group.name if self.secrets_group else None,
```
glennmatthews, line 109: Looks like you caught some good bugs here! :+1:

nautobot/utilities/testing/views.py
```diff
@@ -799,6 +795,36 @@ def test_list_view_plugin_banner(self):
...
+                instance1_cf_values = [
+                    "" if value is None else value for value in instance1.get_custom_fields().values()
+                ]
+                instance1_csv_data += instance1_cf_values
+                # Since values in `data` are all in str; cast all values in instance1_csv_data to str
+                instance1_csv_data = [str(val) for val in instance1_csv_data]
+                instance1_cf_headers = ["cf_" + str(cf.slug) for cf in instance1.get_custom_fields().keys()]
+                instance1_csv_headers = list(getattr(self.model, "csv_headers")) + instance1_cf_headers
+                self.assertEquals(instance1_csv_headers, list(data.keys()))
+                self.assertEquals(instance1_csv_data, list(data.values()))
```
glennmatthews, line 826:
```suggestion
                self.assertEqual(instance1_csv_headers, list(data.keys()))
                self.assertEqual(instance1_csv_data, list(data.values()))
```

nautobot/utilities/testing/views.py
```diff
@@ -799,6 +795,36 @@ def test_list_view_plugin_banner(self):
                 f"<div>You are viewing a table of {self.model._meta.verbose_name_plural}</div>", response_body
             )
 
+        @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
+        def test_queryset_to_csv(self):
+            # Built-in CSV export
+            if hasattr(self.model, "csv_headers"):
```
glennmatthews, line 801: Perhaps?

```suggestion
            if not hasattr(self.model, "csv_headers"):
                self.skip(f"{self.model} has no csv_headers attribute?")
            ...
```

nautobot/utilities/testing/views.py
```diff
@@ -799,6 +795,36 @@ def test_list_view_plugin_banner(self):
...
+                instance1 = self._get_queryset().first()
+                # With filtering
+                response = self.client.get(f"{self._get_url('list')}?export&id={instance1.pk}")
+                self.assertHttpStatus(response, 200)
+                self.assertEqual(response.get("Content-Type"), "text/csv")
+                response_body = response.content.decode(response.charset)
+                reader = csv.DictReader(StringIO(response_body))
+                print(response_body)
+                data = [dict(row) for row in reader][0]
+                print(data)
```
glennmatthews, line 814: Possibly remove `print` here and above?

nautobot/utilities/testing/views.py
```diff
@@ -799,6 +795,36 @@ def test_list_view_plugin_banner(self):
...
+                instance1_csv_data = [str(x) if x is not None else "" for x in instance1.to_csv()]
+                # append instance
+                instance1_cf_values = [
+                    "" if value is None else value for value in instance1.get_custom_fields().values()
+                ]
+                instance1_csv_data += instance1_cf_values
+                # Since values in `data` are all in str; cast all values in instance1_csv_data to str
+                instance1_csv_data = [str(val) for val in instance1_csv_data]
+                instance1_cf_headers = ["cf_" + str(cf.slug) for cf in instance1.get_custom_fields().keys()]
+                instance1_csv_headers = list(getattr(self.model, "csv_headers")) + instance1_cf_headers
```
glennmatthews, line 824:
```suggestion
                instance1_csv_headers = list(self.model.csv_headers) + instance1_cf_headers
```

nautobot/utilities/testing/views.py
```diff
@@ -799,6 +795,36 @@ def test_list_view_plugin_banner(self):
...
+                self.assertEqual(response.get("Content-Type"), "text/csv")
+                instance1 = self._get_queryset().first()
+                # With filtering
+                response = self.client.get(f"{self._get_url('list')}?export&id={instance1.pk}")
+                self.assertHttpStatus(response, 200)
+                self.assertEqual(response.get("Content-Type"), "text/csv")
+                response_body = response.content.decode(response.charset)
+                reader = csv.DictReader(StringIO(response_body))
+                print(response_body)
+                data = [dict(row) for row in reader][0]
```
glennmatthews, line 813: This conversion is a bit unintuitive to me - maybe add a comment explaining what it's doing?

nautobot/utilities/testing/views.py
HanlinMiao: Thank you!

glennmatthews (approved):

PR #3315, opened 2023-02-14
Update DynamicModelChoiceField Select2 Widget to Use Specified Empty Label as Placeholder

bryanculver:
# Closes: #3187
# What's Changed

- Update `PrefixForm`'s `vrf` field to provide an empty label
- Update `initializeDynamicChoiceSelection` to use the null option label as the placeholder value instead of always `---------`. Mimics non-JS behavior.

**Before**
Screenshot 2023-02-14 at 14 20 11

**After**
Screenshot 2023-02-14 at 14 18 49

# TODO

- [NA?] Unit, Integration Tests

### DIFF
--- a/nautobot/project-static/js/forms.js
+++ b/nautobot/project-static/js/forms.js
@@ -158,10 +158,6 @@ function initializeDynamicChoiceSelection(context, dropdownParent=null){
     this_context = $(context);
     this_context.find('.nautobot-select2-api').each(function(){
         thisobj = $(this)
-        // console.log(thisobj.attr("name"));
-        // console.log(thisobj.val());
-        // console.log(thisobj.attr("data-null-option"));
-        // console.log("===");
         placeholder = thisobj.attr("data-null-option") || "---------";
         thisobj.select2({
             allowClear: true,
@@ -297,7 +293,6 @@ function initializeDynamicChoiceSelection(context, dropdownParent=null){
                     results = Object.values(results);
 
                     // Handle the null option, but only add it once
-                    console.log(data.previous)
                     if (element.getAttribute('data-null-option') && data.previous === null) {
                         results.unshift({
                             id: 'null',
--- a/nautobot/project-static/js/forms.js
+++ b/nautobot/project-static/js/forms.js
@@ -156,157 +156,166 @@ function initializeColorPicker(context, dropdownParent=null){
 // Dynamic Choice Selection
 function initializeDynamicChoiceSelection(context, dropdownParent=null){
     this_context = $(context);
-    this_context.find('.nautobot-select2-api').select2({
-        allowClear: true,
-        placeholder: "---------",
-        theme: "bootstrap",
-        width: "off",
-        dropdownParent: dropdownParent,
-        ajax: {
-            delay: 500,
+    this_context.find('.nautobot-select2-api').each(function(){
+        thisobj = $(this)
+        // console.log(thisobj.attr("name"));
+        // console.log(thisobj.val());
+        // console.log(thisobj.attr("data-null-option"));
+        // console.log("===");
+        placeholder = thisobj.attr("data-null-option") || "---------";
+        thisobj.select2({
+            allowClear: true,
+            placeholder: placeholder,
+            theme: "bootstrap",
+            width: "off",
+            dropdownParent: dropdownParent,
+            ajax: {
+                delay: 500,
 
-            url: function(params) {
-                var element = this[0];
-                var url = parseURL(element.getAttribute("data-url"));
+                url: function(params) {
+                    var element = this[0];
+                    var url = parseURL(element.getAttribute("data-url"));
 
-                if (url.includes("{{")) {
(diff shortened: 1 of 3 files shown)


nautobot/project-static/js/forms.js
```diff
@@ -156,157 +156,161 @@ function initializeColorPicker(context, dropdownParent=null){
...
-                                }
-
-                                if (ref_field.val() && ref_field.is(":visible")) {
-                                    value = ref_field.val();
-                                } else if (ref_field.attr("required") && ref_field.attr("data-null-option")) {
-                                    value = "null";
-                                } else {
-                                    return true;  // Skip if ref_field has no value
+    this_context.find('.nautobot-select2-api').each(function(){
+        thisobj = $(this)
```
timizuoebideri1, line 160:
```suggestion
        thisobj = $(this);
```

nautobot/project-static/js/forms.js
```diff
@@ -156,157 +156,161 @@ function initializeColorPicker(context, dropdownParent=null){
...
+                    var parameters = {
+                        q: params.term,
+                        limit: 50,
+                        offset: offset,
+                    };
+
+                    // Set api_version
+                    api_version = $(element).attr("data-api-version")
+                    if(api_version)
+                    parameters["api_version"] = api_version
```
timizuoebideri1, line 196:
```suggestion
                    api_version = $(element).attr("data-api-version");
                    if(api_version){
                        parameters["api_version"] = api_version;
                    }
```

nautobot/project-static/js/forms.js
```diff
@@ -156,157 +156,161 @@ function initializeColorPicker(context, dropdownParent=null){
...
+
+                    // Attach any extra query parameters
+                    $.each(element.attributes, function(index, attr){
+                        if (attr.name.includes("data-query-param-")){
+                            var param_name = attr.name.split("data-query-param-")[1];
+
+                            $.each($.parseJSON(attr.value), function(index, value) {
+                                // Referencing the value of another form field
+                                if (value.startsWith('$')) {
+                                    let element_id = $(element).attr("id")
```
timizuoebideri1, line 210:
```suggestion
                                    let element_id = $(element).attr("id");
```

nautobot/project-static/js/forms.js
```diff
@@ -156,157 +156,161 @@ function initializeColorPicker(context, dropdownParent=null){
...
+                                    let ref_field;
+
+                                    if(element_id.includes("id_form-")){
+                                        let id_prefix = element_id.match(/id_form-[0-9]+-/i, "")[0]
```
timizuoebideri1, line 214:
```suggestion
                                        let id_prefix = element_id.match(/id_form-[0-9]+-/i, "")[0];
```

nautobot/project-static/js/forms.js
```diff
@@ -156,157 +156,161 @@ function initializeColorPicker(context, dropdownParent=null){
...
+                        else if( record.site !== undefined && record.site !== null ) {
+                            results[record.site.name] = results[record.site.name] || { text: record.site.name, children: [] };
+                            results[record.site.name].children.push(record);
+                        }
+                        else if ( (record.group !== undefined || record.group == null) && (record.site !== undefined || record.site === null) ) {
+                            results['global'] = results['global'] || { text: 'Global', children: [] };
+                            results['global'].children.push(record);
+                        }
+                        else {
+                            results[idx] = record
```
timizuoebideri1, line 287:
```suggestion
                            results[idx] = record;
```

timizuoebideri1 (approved):

nautobot/project-static/js/forms.js
```diff
@@ -156,157 +156,162 @@ function initializeColorPicker(context, dropdownParent=null){
...
+                    },Object.create(null));
+
+                    results = Object.values(results);
+
+                    // Handle the null option, but only add it once
+                    if (element.getAttribute('data-null-option') && data.previous === null) {
+                        results.unshift({
+                            id: 'null',
+                            text: element.getAttribute('data-null-option')
+                        });
```
glennmatthews, line 301: Does this need logic to handle the case where there's no `data-null-option`?

glennmatthews (approved):

nautobot/project-static/js/forms.js
bryanculver: `element.getAttribute('data-null-option')` should return undefined. I haven't modified any of this code, it just got indented because of the change from
`this_context.find('.nautobot-select2-api').select2({ #...`
to
`this_context.find('.nautobot-select2-api').each(function(){ #...`

glennmatthews: I meant more do we need to do anything to `results` in the case where there is no `data-null-option` to handle the `--------` case?

bryanculver: I think that would address a different UX in always allowing you to select the null option from the list, but that might not be useful for required fields.

PR #3313, opened 2023-02-14
Remove Region and Site models

HanlinMiao:
# Closes: #2958 
# What's Changed

We need to test and document the app migration strategies.

# TODO

- [ ] Explanation of Change(s)
- [ ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [ ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example Plugin Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #3316, opened 2023-02-14
Fix #2076 - change created field to a DateTimeField

### LINKED ISSUE #2076: Change `ChangeLoggedModel.created` from a DateField to a DateTimeField
### As ...

Dan - Data Owner

### I want ...

The `created` field on all change-logged models to include a time stamp as well as a date, rather than just a date with no associated time information.

### So that ...

I can have more granular insight into when data was first entered into Nautobot, especially as part of a global operation spanning multiple time zones.

- Objects automatically store a timestamp as well as a datestamp when first created.
- Object detail pages show the timestamp as well as the datestamp as the "Created ..." text below the page header

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [X] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

- Change `ChangeLoggedModel.created` from a `DateField` to a `DateTimeField`.
- Implement data migration for existing models.
    - Midnight on the same day is sufficient, just needs to be documented
    - May already be addressed in the column/field migration

### External Dependencies

None expected.

glennmatthews:
# Closes: #2076 
# What's Changed

**It probably makes sense to hold this PR until #3305 and #3313 are merged since these migrations *will* conflict**

- Change `ChangeLoggedModel.created` from a `DateField` to a `DateTimeField`.
- Generate the many, many migrations that this causes.
    - I manually verified that existing records `changed` fields are automatically migrated to a DateTime with 00:00:00Z as their time.
- Updated `CreatedUpdatedModelFilterSetMixin` to use DateTimeFilters for `created` and related filters.
- Updated REST API `CreatedUpdatedFilterTest` test case to check created as a datetime filter in the REST API.
- Updated documentation in various places.

### Migrated existing record:

image

### Newly created post-migration record:

image

# TODO

- [x] Unit, Integration Tests


glennmatthews:
Some explanation of fd11b732866ace61a8b06662b008756400c161d2:

- The extras 0004 data migration was not specifying which models to create/associate Statuses for, which meant that as we introduced new Status-supporting models in later migrations (Interface, VMInterface, DeviceRedundancyGroup, Location), even though we had specific data migrations written for the Statuses for those models, this migration would attempt to create/associate their Statuses as well.
- This was problematic of course due to the way migrations use historical models; the new/updated models didn't actually exist as historical models at this point in the migration sequence so it wouldn't be possible to look them up using the migration's `apps`. This was worked around (incorrectly!) by *not* passing the historical `apps` into `create_custom_statuses()` but instead using Django's global `apps`, in other words pulling in the latest versions of the models instead of the correct historical models.
    - Once I made the change for `created` from a DateField to a DateTimeField, this model mismatch caused initial migrations to fail on MySQL (but oddly enough, not on Postgres?) because it was trying to use the latest (DateTimeField) version of the Status model instead of the historical (DateField) version.
- The correct fix here is as follows:
   1. Pass the historical `apps` to `create_custom_statuses` from the migration, not the global `apps`
   2. Explicitly declare which models the extras 0004 migration is responsible for creating Statuses for (so that later-added models are handled by the later data migrations, instead of this one trying to handle them)
   3. Add explicit `dependencies` to the initial migrations for all other apps into the extras 0004 migration so as to ensure that the initial set of models are all defined before trying to create their associated Statuses (without this, this migration was actually running before the initial `circuits` migrations, and so it would error out when trying to create Statuses for the Circuit model)

glennmatthews:
It *might* be good to back-port fd11b732866ace61a8b06662b008756400c161d2 to 1.5 - it's not directly causing failures in that release that I know of, but it is definitely a bug

nautobot/core/tests/test_utils.py
```diff
@@ -11,6 +11,7 @@
 from nautobot.core.api import utils as api_utils
 from nautobot.core.models import fields as core_fields, utils as models_utils
 from nautobot.core.utils import data as data_utils, filtering, lookup, requests
+from nautobot.circuits import models as circuits_models
```
timizuoebideri1, line 14: Alphabetic import order

glennmatthews: You mean `ci` doesn't come after `co`? :-) Will fix.

changes/2076.changed
```diff
@@ -0,0 +1 @@
+Changed the `created` field of all models from a DateField to a DateTimeField for added granularity. Preexisting records will show as created at midnight UTC on their original creation date.
```
gsnider2195, line 1: Nit: Is it actually midnight UTC or is it midnight in the database's TZ?

glennmatthews: Fair question! Looks to be UTC:

```python
>>> settings.TIME_ZONE
'America/Chicago'
>>> Site.objects.first().created
datetime.datetime(2023, 2, 15, 0, 0, tzinfo=<UTC>)
```

image

gsnider2195: Thanks for checking!

timizuoebideri1 (approved):

gsnider2195 (approved):

Issue #766, opened 2021-08-03
Epic: Data Insights


jedelman8:
This is a larger issue to track several issues that will be opened up to increase visibility into the data stored in Nautobot. When they are opened, they will be linked and referenced accordingly:

- [ ] Create dashboards to simplify consumption of data already in Nautobot.  Pick 2 or 3 data models to start with to get feedback on the approach, e.g.  % of particular device types, % of circuit providers, etc. 

```[tasklist]
### Tasks
- [ ] https://github.com/nautobot/nautobot/issues/710
- [ ] https://github.com/nautobot/nautobot/issues/1629
- [ ] https://github.com/nautobot/nautobot/issues/4051
- [ ] https://github.com/nautobot/nautobot/issues/4053
- [ ] https://github.com/nautobot/nautobot/issues/4054
- [ ] https://github.com/nautobot/nautobot/issues/4065
```

glennmatthews:
See #710 for worker/job metrics requirements.

jedelman8:
#710 takes care of the first two items in this Epic.

Kircheneer:
> Add functionality from the Capacity Metrics plugin into core

Is this a definitive plan?

bryanculver:
> > Add functionality from the Capacity Metrics plugin into core
> 
> Is this a definitive plan?

It's the desire at the time but for now, this isn't committed in any significant way.

Kircheneer:
For capacity-metrics -> core my thoughts are the following: 
- RQ functionality: Drop
- Model count stats: Transfer this to core, keeping in mind that in a multiprocess environment only one process needs to calculate these, as they are the same across all processes
- Metric extensibility: Capacity-metrics offers the tools for plugins to provide custom metrics - this needs some more thought

lampwins:
Closing to focus on more concretely defined issues

PR #3324, opened 2023-02-16
Port metric extensibility approach from capacity-metrics.

Kircheneer:
# Relates to: #766 
# What's Changed

This ports the metric extensibility from capacity-metrics to this plugin as a first effort to move capacity-metrics into core. There is an example of what this does in the example plugin, here is what that looks like rendered:

```
# HELP nautobot_example_metric_count Nautobot Example Count Metric
# TYPE nautobot_example_metric_count gauge
nautobot_example_metric_count{name="Test"} 100.0
# HELP nautobot_app_metrics_processing_ms Time in ms to generate the app metrics
# TYPE nautobot_app_metrics_processing_ms gauge
nautobot_app_metrics_processing_ms 5.14007
```

I overwrote the `/metrics` view of `django_prometheus` in order to inject these custom app metrics and then copied the registry approach from the jobs, putting metrics into `$app.metrics.metrics`.

Of the two existing and one planned approaches in capacity-metrics to allow for custom metrics I chose to just implement the planned approach of auto-discovering from via path. I think since all the three ways (explicitly calling a function on the metric, configuring it in the app settings and discovering via path) are equally powerful, its best to go for the approach which mimics other app extensibility features such as jobs here.

# TODO
- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/core/urls.py
+++ b/nautobot/core/urls.py
@@ -3,7 +3,7 @@ from django.conf.urls import include
 from django.urls import path
 from django.views.static import serve
 
-from nautobot.core.views import CustomGraphQLView, HomeView, StaticMediaFailureView, SearchView
+from nautobot.core.views import CustomGraphQLView, HomeView, StaticMediaFailureView, SearchView, custom_app_metric_view
 from nautobot.extras.plugins.urls import (
     plugin_admin_patterns,
     plugin_patterns,
@@ -62,7 +62,7 @@ if settings.DEBUG:
 
 if settings.METRICS_ENABLED:
     urlpatterns += [
-        path("", include("django_prometheus.urls")),
+        path("metrics/", custom_app_metric_view),
     ]
 
 handler404 = "nautobot.core.views.resource_not_found"
--- a/nautobot/core/urls.py
+++ b/nautobot/core/urls.py
@@ -62,7 +62,7 @@ if settings.DEBUG:
 
 if settings.METRICS_ENABLED:
     urlpatterns += [
-        path("metrics/", custom_app_metric_view),
+        path("metrics/", custom_app_metric_view, name="metrics"),
     ]
 
 handler404 = "nautobot.core.views.resource_not_found"
--- a/nautobot/core/urls.py
+++ b/nautobot/core/urls.py
@@ -3,7 +3,7 @@ from django.conf.urls import include
 from django.urls import path
 from django.views.static import serve
 
-from nautobot.core.views import CustomGraphQLView, HomeView, StaticMediaFailureView, SearchView, custom_app_metric_view
+from nautobot.core.views import CustomGraphQLView, HomeView, StaticMediaFailureView, SearchView, nautobot_metrics_view
 from nautobot.extras.plugins.urls import (
     plugin_admin_patterns,
     plugin_patterns,
@@ -62,7 +62,7 @@ if settings.DEBUG:
 
 if settings.METRICS_ENABLED:
     urlpatterns += [
-        path("metrics/", custom_app_metric_view, name="metrics"),
+        path("metrics/", nautobot_metrics_view, name="metrics"),
     ]
 
 handler404 = "nautobot.core.views.resource_not_found"
--- a/nautobot/core/views/__init__.py
+++ b/nautobot/core/views/__init__.py
@@ -1,10 +1,12 @@
 import os
 import platform
 import sys
+import time
 
+import prometheus_client
 from django.conf import settings
 from django.contrib.auth.mixins import AccessMixin
-from django.http import HttpResponseServerError, JsonResponse, HttpResponseForbidden
+from django.http import HttpResponseServerError, JsonResponse, HttpResponseForbidden, HttpResponse
 from django.shortcuts import render
 from django.template import loader, RequestContext, Template
 from django.template.exceptions import TemplateDoesNotExist
@@ -15,6 +17,9 @@ from django.views.csrf import csrf_failure as _csrf_failure
 from django.views.generic import TemplateView, View
 from packaging import version
 from graphene_django.views import GraphQLView
+from prometheus_client import multiprocess
+from prometheus_client.metrics_core import GaugeMetricFamily
+from prometheus_client.registry import Collector
 
 from nautobot.core.constants import SEARCH_MAX_RESULTS, SEARCH_TYPES
 from nautobot.core.forms import SearchForm
@@ -220,3 +225,34 @@ class CustomGraphQLView(GraphQLView):
         data["saved_graphiql_queries"] = GraphQLQuery.objects.all()
         data["form"] = GraphQLQueryForm
         return render(request, self.graphiql_template, data)
+
+
+class NautobotAppMetricsCollector(Collector):
+    """Custom Nautobot metrics collector.
+
+    Metric collector that reads from registry["plugin_metrics"] and yields any metrics registered there."""
+
+    def collect(self):
+        """Collect metrics from plugins."""
+        start = time.time()
+        for metric in registry["plugin_metrics"]:
+            yield from metric()
+        gauge = GaugeMetricFamily("nautobot_app_metrics_processing_ms", "Time in ms to generate the app metrics")
+        duration = time.time() - start
+        gauge.add_metric([], format(duration * 1000, ".5f"))
+        yield gauge
+
+
+def custom_app_metric_view(request):
+    """Exports /metrics.
+
+    This overwrites the default django_prometheus view to inject metrics from Nautobot apps."""
+    if "PROMETHEUS_MULTIPROC_DIR" in os.environ or "prometheus_multiproc_dir" in os.environ:
+        prometheus_registry = prometheus_client.CollectorRegistry()
+        multiprocess.MultiProcessCollector(prometheus_registry)
+    else:
+        prometheus_registry = prometheus_client.REGISTRY
+    nb_app_collector = NautobotAppMetricsCollector()
+    prometheus_registry.register(nb_app_collector)
+    metrics_page = prometheus_client.generate_latest(prometheus_registry)
+    return HttpResponse(metrics_page, content_type=prometheus_client.CONTENT_TYPE_LATEST)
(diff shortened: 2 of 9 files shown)


Kircheneer:
Looking for some feedback before I undertake the docs. Thanks!

nautobot/core/urls.py
```diff
@@ -62,7 +62,7 @@
 
 if settings.METRICS_ENABLED:
     urlpatterns += [
-        path("", include("django_prometheus.urls")),
+        path("metrics/", custom_app_metric_view),
```
glennmatthews, line 65: Does this cause us to *lose* the current django_prometheus metrics, or is it only augmenting those?

nautobot/extras/plugins/__init__.py
```diff
@@ -35,6 +35,7 @@
 registry["plugin_graphql_types"] = []
 registry["plugin_jobs"] = []
 registry["plugin_template_extensions"] = collections.defaultdict(list)
+registry["plugin_metrics"] = []
```
glennmatthews, line 38: Since we're trying to rename `"plugins"` to `"apps"` in the near term, let's use `app_metrics` throughout here even though it's less self-consistent for now.

nautobot/extras/plugins/__init__.py
```diff
@@ -147,6 +149,12 @@ def ready(self):
             register_jobs(jobs)
             self.features["jobs"] = jobs
 
+        # Import metrics (if present)
+        metrics = import_object(f"{self.__module__}.{self.metrics}")
+        if metrics is not None:
+            register_metrics(metrics)
+            self.features["metrics"] = metrics
```
glennmatthews, line 156: Just a note that you'll want to add this to `nautobot/extras/templates/extras/plugin_detail.html`.

nautobot/core/urls.py
Kircheneer: That is correct. If you compare the sources between the `django_prometheus`-provided URL and `custom_app_metric_view` then there are only two lines added which register an _additional_ collector into the view (which is our app extensibility collector).

nautobot/extras/plugins/__init__.py
Kircheneer: Done.

nautobot/extras/plugins/__init__.py
Kircheneer: Done.

Kircheneer:
Hm the failing CI step works on my machine (™️) - any pointers here? _Maybe_ just a restart? 🙏🏻

glennmatthews:
Looks like the ready() call is somehow happening before the database migrations have actually run. Taking a look.

glennmatthews:
Does moving `from example_plugin.models import ExampleModel` inside the `metric_example()` function make any difference?

Kircheneer:
> Does moving `from example_plugin.models import ExampleModel` inside the `metric_example()` function make any difference?

Unfortunately not..

Kircheneer:
I can avoid actually calling the metric callables in `ready()`, but then the info on the plugin detail view won't be as helpful (I couldn't use the actual metric names, but rather just the name of the callable).

Kircheneer:
I added some docs, but I think there will need to be some consolidation with [this](https://docs.nautobot.com/projects/core/en/stable/additional-features/prometheus-metrics/).

bryanculver:
Updates from out-of-band conversations:

@Kircheneer is going to:

- Deploy his branch into a demo environment to ensure the existing metrics aren't reporting differently
- Write a test/validate that output of `/metrics` is unchanged
- Investigate if we can subclass `ExportToDjangoView` to see if we can ensure we stay in-sync with the upstream implementation, just adding our own metrics collector

nautobot/core/views/__init__.py
```diff
@@ -220,3 +225,34 @@ def render_graphiql(self, request, **data):
...
+        start = time.time()
+        for metric in registry["app_metrics"]:
+            yield from metric()
+        gauge = GaugeMetricFamily("nautobot_app_metrics_processing_ms", "Time in ms to generate the app metrics")
+        duration = time.time() - start
+        gauge.add_metric([], format(duration * 1000, ".5f"))
+        yield gauge
+
+
+def custom_app_metric_view(request):
```
bryanculver, line 246: Let's see if we can subclass the existing Django Prometheus view, `ExportToDjangoView`.

nautobot/core/views/__init__.py
```diff
@@ -220,3 +225,34 @@ def render_graphiql(self, request, **data):
...
+    """Exports /metrics.
+
+    This overwrites the default django_prometheus view to inject metrics from Nautobot apps."""
+    if "PROMETHEUS_MULTIPROC_DIR" in os.environ or "prometheus_multiproc_dir" in os.environ:
+        prometheus_registry = prometheus_client.CollectorRegistry()
+        multiprocess.MultiProcessCollector(prometheus_registry)
+    else:
+        prometheus_registry = prometheus_client.REGISTRY
+    nb_app_collector = NautobotAppMetricsCollector()
+    prometheus_registry.register(nb_app_collector)
```
bryanculver, line 256: Let's see if there isn't a potentially better place to register this collector without needing to do it every call time.

nautobot/core/views/__init__.py
Kircheneer: Unfortunately this is not possible. I have added a comment that explains this into the code.

nautobot/core/views/__init__.py
Kircheneer: I could theoretically evaluate persisting the registry over multiple calls to the view, but I don't think this is a good idea in the end for two reasons:
1. This would have us deviate a lot further from the proven `django_prometheus` implementation of this view
2. The collector themselves are implemented as generators. When the metrics are generated, these generators are depleted. Thus, I would have to work around that, which feels hacky.

Kircheneer: I explained some of this in a comment as well.

bryanculver (approved):
After the external validation and feedback from @Kircheneer, I'm comfortable with where this has wound up.

@Kircheneer If you could toss in the automated test that would be 💯:
> Write a test/validate that output of /metrics is unchanged

Kircheneer:
There is now one unit test that tests for this behavior.

nautobot/core/views/__init__.py
```diff
@@ -220,3 +225,40 @@ def render_graphiql(self, request, **data):
...
+
+
+def custom_app_metric_view(request):
+    """Exports /metrics.
+
+    This overwrites the default django_prometheus view to inject metrics from Nautobot apps.
+
+    Note that we cannot use `prometheus_django.ExportToDjangoView`, as that is a simple function, and we need access to
+    the `prometheus_registry` variable that is defined inside of it."""
+    if "PROMETHEUS_MULTIPROC_DIR" in os.environ or "prometheus_multiproc_dir" in os.environ:
```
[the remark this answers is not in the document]
Kircheneer: Since the lower-case spelling didn't originate in this PR, should I open a separate issue that addresses this?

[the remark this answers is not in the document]
bryanculver: I think this is more for us to keep in mind since we are mimicking their implementation. _If_ on the off-chance someone is still using it since it's still supported I think we shouldn't drop it pre-emptively, especially not in a patch release.

nautobot/core/views/__init__.py
Kircheneer: I think you raise a good point here. Just for my technical understanding - when would we be running in a non-multiprocess scenario - on a deployment directly to a VM?

Kircheneer: Win for the tests! They replicated this error you caught: https://github.com/nautobot/nautobot/actions/runs/4366588475/jobs/7636881567

nautobot/core/views/__init__.py
```diff
@@ -220,3 +225,40 @@ def render_graphiql(self, request, **data):
...
+        start = time.time()
+        for metric in registry["app_metrics"]:
+            yield from metric()
+        gauge = GaugeMetricFamily("nautobot_app_metrics_processing_ms", "Time in ms to generate the app metrics")
+        duration = time.time() - start
+        gauge.add_metric([], format(duration * 1000, ".5f"))
+        yield gauge
+
+
+def custom_app_metric_view(request):
```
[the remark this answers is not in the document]
Kircheneer: Good point. I ended up with `nautobot_metrics_view` because I realized this is now not just for metrics extensibility but for all metrics.

nautobot/core/views/__init__.py
Kircheneer: Fixed now.

nautobot/docs/plugins/development.md
```diff
@@ -1370,6 +1370,32 @@ except ImportError:
     pass
 ```
 
+## Prometheus Metrics
+
++++ 1.5.?
+
+It is possible for Nautobot apps to provide their own Prometheus metrics. There are two general ways to achieve this:
+
+1. Use the prometheus_client library directly in your app code. Depending on whether that code runs in the web server or the worker context, the metric will show up in the respective `/metrics` endpoint(s).
```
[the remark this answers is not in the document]
Kircheneer: Done!

Kircheneer:
Alright, I believe that should be it 🚀

nautobot/docs/plugins/development.md
```diff
@@ -1370,6 +1370,32 @@ except ImportError:
...
 ```
 
+## Prometheus Metrics
+
++++ 1.5.?
+
+It is possible for Nautobot apps to provide their own [Prometheus metrics](../additional-features/prometheus-metrics.md). There are two general ways to achieve this:
+
+1. Use the `prometheus_client` library directly in your app code. Depending on whether that code runs in the web server or the worker context, the metric will show up in the respective `/metrics` endpoint(s) (i.e. metrics generated in the worker context show up in the worker's endpoint and those generated in the web application's context show up in the web application's endpoint).
+2. If the metric cannot be generated alongside existing code, apps can implement individual metric generator functions and register them into a list called `metrics` in a file named `metrics.py`at the root of the app. Nautobot will automatically read these and expose them via its `/metrics` endpoint. The following code snippet shows an example metric defined this way:
```
glennmatthews, line 1380:
```suggestion
2. If the metric cannot be generated alongside existing code, apps can implement individual metric generator functions and register them into a list called `metrics` in a file named `metrics.py` at the root of the app. Nautobot will automatically read these and expose them via its `/metrics` endpoint. The following code snippet shows an example metric defined this way:
```

glennmatthews (approved):
Magnifique. :shipit:

Issue #3196, opened 2023-01-31
Filter on custom field (selection) can't be changed once filter is configured


phoenixswiss:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.9
* Python version: 3.8.10
* Database platform, version: postgresql, Ubuntu 12.12-0ubuntu0.20.04.1
* Middleware(s): -

### Steps to Reproduce
1. Create a custom selection field and assign it to the object interface with at least two options (optionA, optionB)
2. Configure/select different options on the newly created custom field on some interfaces
3. On the interfaces page, go to the filter input and select optionA from the drop down box and click Apply
4. Change the filter and select optionB instead of optionA and click Apply again

### Expected Behavior
The filter will be modified and all interfaces with optionB defined will be displayed.

### Observed Behavior
The filter will not be changed and interfaces with optionA will be displayed even optionB is selected in the filter window.

glennmatthews:
Appears to be the case for all custom field filtering, not just selection-type custom fields. Reproduced on demo.nautobot.com:

1. go to https://demo.nautobot.com/dcim/interfaces
2. Click the "Filter" button; in the standard filter form (not "advanced") enter `leaf` for the `Role` custom field and click Apply.
3. Page refreshes as https://demo.nautobot.com/dcim/interfaces/?&enabled=&mgmt_only=&mac_address=&nautobot_bgp_models_interfaces_bgp_routing_instance=&cf_role=leaf&q= (note `cf_role=leaf`)
4. Click the "Filter" button, enter `peer` for the `Role` custom field, and Apply.
5. Page refreshes as https://demo.nautobot.com/dcim/interfaces/?&enabled=&mgmt_only=&mac_address=&nautobot_bgp_models_interfaces_bgp_routing_instance=&cf_role=peer&cf_role=leaf&q=  (note `cf_role=peer&cf_role=leaf` instead of the intended `cf_role=peer`. Since `cf_role` is not a multi-value filter, the last-specified value of the query parameter (`leaf`) wins out.

glennmatthews:
Advanced filtering of custom fields appears to be even more broken:

![image](https://user-images.githubusercontent.com/5603551/215803744-670a0ba4-959e-4197-8739-2f65b0eb7f80.png)

results in the query:

[...]

PR #3326, opened 2023-02-16
Fix Filter on custom field (selection) can't be changed once filter is configured

timizuoebideri1:
# Closes: #3196 
# What's Changed

Because the custom field(selection) on the Default filter form and the DynamicFilterForm had different field types, the DynamicFilterForm field did not update when the choice field on the Default filter form did. The correct field type for DefaultFilter is a choice field, whereas the field type for DynaicFilterForm is a multichar field.

To fix this bug, a revamp of `get_filterset_parameter_form_field` was performed (Function used in generating the Dom representation of the filter set field). 

- Instead of constructing each form field representation of the filterset field, I retrieve it from the model's Form; most form fields that need to be generated already exist in its ModelForm, so instead of reconstructing this ModelForm field, I simply retrieve it. 
- In some cases, such as related name fields and other fields that do not exist as fields in the ModelForm, the field is constructed only in these cases.
- The ModelForm field widget for most fields does not allow multiple selection, which was addressed in this PR by swapping out the default ModelForm field widget with its multiple equilivant.

# TODO

- [ ] Unit, Integration Tests
- [ ] On page load dynamic filter form do not have the correct field type of custom field(selection)
- [ ] `has_` shows unknown as first choice


glennmatthews:
At a quick glance, this *feels* overcomplicated to me. Would it be feasible to instead fix the filter forms so that the two have the same field type?

timizuoebideri1:
C

PR #3332, opened 2023-02-16
Fix missing imports in example secrets.py file in `development.md`

alextremblay:
Fix missing imports in example secrets.py file

This PR makes the example secrets provider plugin useable

Without this change, anyone attempting to implement a secret provider in their own apps using the `secrets.py` file from the documentation will run into errors regarding undefined classes


bryanculver (approved):
@alextremblay 

Please add a `changes/3332.fixed` changelog fragment (per documentation: https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments).

This change mimics what's in the example plugin.

PR #3333, opened 2023-02-17
Update `netutils` to ^1.4.0 from ~1.4.0.

itdependsnetworks:
# Closes: #DNE
# What's Changed

Dependency Version

# TODO

### DIFF
--- a/changes/3333.dependencies
+++ b/changes/3333.dependencies
@@ -1,1 +1,1 @@
-Update `netutils` to ^1.4.0 from ~1.4.0.
+Updated `netutils` constraint from ~1.4.0 to ^1.4.0 to permit semver upgrades.
--- a/changes/3333.dependencies
+++ b/changes/3333.dependencies
@@ -1,0 +1,1 @@
+Update `netutils` to ^1.4.0 from ~1.4.0.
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -118,7 +118,7 @@ napalm = {version = "~3.4.1", optional = true}
 # IP prefix and address handling
 netaddr = "~0.8.0"
 # Library of network related Jinja Filters
```
glennmatthews, line 120: Maybe add a comment as well, something like:

```python
# netutils is well-behaved with respect to semver, so we permit a looser (^) version constraint here.
```

changes/3333.dependencies
```diff
@@ -0,0 +1 @@
+Update `netutils` to ^1.4.0 from ~1.4.0.
```
glennmatthews, line 1:
```suggestion
Updated `netutils` constraint from ~1.4.0 to ^1.4.0 to permit semver upgrades.
```

bryanculver: `s/permit semver/permit minor semver/` to @glennmatthews's suggestion

bryanculver (approved):

PR #3335, opened 2023-02-17
Fix Filter on custom field (selection) can't be changed once filter is configured

timizuoebideri1:
# Closes: #3196 
# What's Changed

- Because the custom field(selection) on the Default filter form and the DynamicFilterForm had different field types, the DynamicFilterForm field did not update when the choice field on the Default filter form did. The correct field type for DefaultFilter is a choice field, whereas the field type for DynaicFilterForm is a multichar field.

- Previously, this updating of fields on DynamicFilterForm and DefualtFilterForm was only performed on select fields. Added JavaScript logic to update on input fields also.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/3335.fixed
+++ b/changes/3335.fixed
@@ -1,1 +1,1 @@
-Fix Filter on custom field (selection) can't be changed once filter is configured.
+Fixed inability to change filtering on custom field (selection) once filter is configured.
--- a/changes/3335.fixed
+++ b/changes/3335.fixed
@@ -1,0 +1,1 @@
+Fix Filter on custom field (selection) can't be changed once filter is configured.
--- a/nautobot/core/tests/integration/test_filters.py
+++ b/nautobot/core/tests/integration/test_filters.py
@@ -22,6 +22,7 @@ class ListViewFilterTestCase(SeleniumTestCase):
         self.user.is_superuser = True
         self.user.save()
 
+        # TODO(timizuo): changing these from name to slug when resolving issue #824
         self.cf_text_field_name = "text_field"
         self.cf_integer_field_name = "integer_field"
         self.cf_select_field_name = "select_field"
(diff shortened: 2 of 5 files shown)


nautobot/core/tests/integration/test_filters.py
```diff
@@ -15,6 +18,20 @@ def setUp(self):
         self.login(self.user.username, self.password)
         factory.RegionFactory.create_batch(15, has_parent=False)
         factory.SiteFactory.create_batch(15, has_tenant=False)
+        # set test user to admin
+        self.user.is_superuser = True
+        self.user.save()
+
+        self.cf_text_field_name = "text_field"
+        self.cf_integer_field_name = "integer_field"
+        self.cf_select_field_name = "select_field"
```
glennmatthews, line 27: Mind adding a # TODO comment here about changing these from `name` to `slug` in #824?

changes/3335.fixed
```diff
@@ -0,0 +1 @@
+Fix Filter on custom field (selection) can't be changed once filter is configured.
```
glennmatthews, line 1:
```suggestion
Fixed inability to change filtering on custom field (selection) once filter is configured.
```

glennmatthews (commented):
Haven't tested it myself but it looks reasonable. :-)

nautobot/utilities/utils.py
```diff
@@ -869,7 +869,9 @@ def get_filterset_parameter_form_field(model, parameter):
     form_field = field.field
 
     # TODO(Culver): We are having to replace some widgets here because multivalue_field_factory that generates these isn't smart enough
-    if isinstance(field, NumberFilter):
+    if isinstance(field, CustomFieldFilterMixin):
+        form_field = field.custom_field.to_form_field()
```
bryanculver, line 873: Sadly long-term, we should merge this logic here (in `get_filterset_parameter_form_field`) with `multivalue_field_factory` and the field+filter+widget association.

bryanculver (approved):
Validated locally as well.

Issue #3080, opened 2023-01-05
Missing REST API listing for content types


Thetacz:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.6
* Python version: 3.10.9
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Create a Job with following variable:
```
content_type = ObjectVar(
    model=ContentType,
)
```
2. Navigate to Job run form in UI

### Expected Behavior
Able to run the Job

### Observed Behavior
![image](https://user-images.githubusercontent.com/26114847/210722543-45002bd4-d61b-4a70-87a1-096777d0dad3.png)

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/django/urls/base.py", line 71, in reverse
    extra, resolver = resolver.namespace_dict[ns]
KeyError: 'contenttypes-api'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.10/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.10/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/nautobot/utilities/views.py", line 116, in dispatch
    return super().dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/nautobot/extras/views.py", line 1107, in get
    return render(
  File "/usr/local/lib/python3.10/site-packages/django/shortcuts.py", line 19, in render
    content = loader.render_to_string(template_name, context, request, using=using)
  File "/usr/local/lib/python3.10/site-packages/django/template/loader.py", line 62, in render_to_string
    return template.render(context, request)
  File "/usr/local/lib/python3.10/site-packages/django/template/backends/django.py", line 61, in render
    return self.template.render(context)
  File "/usr/local/lib/python3.10/site-packages/django/template/base.py", line 170, in render
    return self._render(context)
  File "/usr/local/lib/python3.10/site-packages/django/template/base.py", line 162, in _render
[...]

Issue #3300, opened 2023-02-13
Unable to query `content_types` via GraphQL


joewesch:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.10
* Python version: 3.7.16
* Database platform, version: postgres, 13
* Middleware(s): N/A

### Steps to Reproduce
1. Use the following GraphQL query:
```
query {
  content_types {
    app_label
    model
  }
}
```

### Expected Behavior
I expected the response to provide all app labels and models

### Observed Behavior
```
{
  "errors": [
    {
      "message": "'ContentTypeManager' object has no attribute 'restrict'",
      "locations": [
        {
          "line": 33,
          "column": 3
        }
      ],
      "path": [
        "content_types"
      ]
    }
  ],
  "data": {
    "content_types": null
  }
}
```

The REST endpoint `/api/extras/content-types/` does work, however, as a workaround for now.

lampwins:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Nautobot-Core-Product-64a8783b7a3f4b1225973a0c/issues/nautobot/nautobot/3300?planning-poker) with Zenhub @bryanculver @gsnider2195 @HanlinMiao @jathanism

PR #3341, opened 2023-02-20
Fix get_route_for_model for ContentTypes

timizuoebideri1:
# Closes: #3080 
# What's Changed

Running `get_route_for_model(ConentType, api=True...)` would yield `content-types-api:contenttype-list` rather than the newly created endpoint for contenttypes because `ContentType` is not a core model in nautobot (`extras-api:contenttype-list`)

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/3341.fixed
+++ b/changes/3341.fixed
@@ -1,0 +1,1 @@
+Fix the issue with the ContentType reverse return, i.e get_route_for_model(ContentType, api=True...).
--- a/nautobot/utilities/tests/test_utils.py
+++ b/nautobot/utilities/tests/test_utils.py
@@ -1,6 +1,7 @@
 import pickle
 
 from django import forms
+from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ValidationError
 from django.db.models import Q
 from django.http import QueryDict
@@ -231,6 +232,8 @@ class GetFooForModelTest(TestCase):
         self.assertEqual(get_route_for_model(Device, "list", api=True), "dcim-api:device-list")
         self.assertEqual(get_route_for_model("dcim.site", "detail", api=True), "dcim-api:site-detail")
         self.assertEqual(get_route_for_model(Site, "detail", api=True), "dcim-api:site-detail")
+        self.assertEqual(get_route_for_model(ContentType, "list", api=True), "extras-api:contenttype-list")
+        self.assertEqual(get_route_for_model(ContentType, "detail", api=True), "extras-api:contenttype-detail")
         self.assertEqual(
             get_route_for_model("example_plugin.examplemodel", "list", api=True),
             "plugins-api:example_plugin-api:examplemodel-list",
--- a/nautobot/utilities/tests/test_utils.py
+++ b/nautobot/utilities/tests/test_utils.py
@@ -1,6 +1,7 @@
 import pickle
 
 from django import forms
+from django.contrib.auth.models import Group
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ValidationError
 from django.db.models import Q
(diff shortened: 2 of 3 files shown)


bryanculver (approved):
LGTM. Curious to know why we can't use the Django default for ContentType. Doesn't appear we do anything fancy with our ContentType view. Is their default not read-only?

glennmatthews:
> LGTM. Curious to know why we can't use the Django default for ContentType. Doesn't appear we do anything fancy with our ContentType view. Is their default not read-only?

There is no default REST API provided by the `contenttypes` app AFAIK. We provide one for the ContentType model through our `extras` app.

nautobot/utilities/utils.py
```diff
@@ -105,7 +105,13 @@ def get_route_for_model(model, action, api=False):
         model = get_model_from_name(model)
 
     suffix = "" if not api else "-api"
-    prefix = f"{model._meta.app_label}{suffix}:{model._meta.model_name}"
+    # Because ContentType is not a core model in nautobot, running a 'ContentType. meta.app label'
+    # would return "contenttypes" rather than "extras".
```
glennmatthews, line 109: Perhaps:

```suggestion
    # The `contenttypes` app doesn't provide REST API endpoints,
    # but Nautobot provides one for the ContentType model in our `extras` app.
```

glennmatthews (commented):
1) Do we need a similar fix for the `Group` model and the `users/groups/` REST URL?
2) Would you mind looking into how difficult it would be to fix #3300 in this same PR?

bryanculver:
> > LGTM. Curious to know why we can't use the Django default for ContentType. Doesn't appear we do anything fancy with our ContentType view. Is their default not read-only?
> 
> There is no default REST API provided by the `contenttypes` app AFAIK. We provide one for the ContentType model through our `extras` app.

DOH. Didn't think beyond the `get_route_for_model` being our invention.

jathanism (approved):

timizuoebideri1:
> * Do we need a similar fix for the `Group` model and the `users/groups/` REST URL?
> * Would you mind looking into how difficult it would be to fix #3300 in this same PR?

1. I suppose we'll need a similar fix for 'Group'. 
2. I have little to no idea how GraphQL works, so it may take longer because I need to learn how it works first.

glennmatthews:
> > * Would you mind looking into how difficult it would be to fix [Unable to query `content_types` via GraphQL #3300](https://github.com/nautobot/nautobot/issues/3300) in this same PR?
>     2. I have little to no idea how GraphQL works, so it may take longer because I need to learn how it works first.

No worries, that can be deferred since it's a separate issue. Thanks!

changes/3341.fixed
```diff
@@ -0,0 +1 @@
+Fix the issue with the ContentType reverse return, i.e get_route_for_model(ContentType, api=True...).
```
glennmatthews, line 1: Perhaps:

```suggestion
Fixed missing `get_route_for_model()` logic for the `ContentType` and `Group` models.
```

nautobot/utilities/tests/test_utils.py
```diff
@@ -231,6 +233,10 @@ def test_get_route_for_model(self):
         self.assertEqual(get_route_for_model(Device, "list", api=True), "dcim-api:device-list")
         self.assertEqual(get_route_for_model("dcim.site", "detail", api=True), "dcim-api:site-detail")
         self.assertEqual(get_route_for_model(Site, "detail", api=True), "dcim-api:site-detail")
+        self.assertEqual(get_route_for_model(ContentType, "list", api=True), "extras-api:contenttype-list")
+        self.assertEqual(get_route_for_model(ContentType, "detail", api=True), "extras-api:contenttype-detail")
+        self.assertEqual(get_route_for_model(Group, "list", api=True), "users-api:group-list")
+        self.assertEqual(get_route_for_model(Group, "detail", api=True), "users-api:group-detail")
```
glennmatthews, line 239: Hmm, I suppose for the `api=False` case we don't actually have viable routes for either of these models, do we? I guess it's inline with how `get_route_for_model` currently works that it just tries to guess the route name, but doesn't actually verify that it's valid though. Ah well.

glennmatthews (approved):

Issue #3214, opened 2023-02-02
BaseFilterSet not using multiple choice filters for CharFields with choices


gsnider2195:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0a1
* Python version: 3.10
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. from nbshell inspect the filter for `Interface.mode`:
```py
from nautobot.dcim.filters import InterfaceFilterSet
InterfaceFilterSet.get_filters().get("mode")
# returns <django_filters.filters.ChoiceFilter object at 0x7ff8f8a57e80>
```

### Expected Behavior
This should be a `MultipleChoiceFilter`

### Observed Behavior
The filter is being set to `ChoiceFilter`

bryanculver:
In the same family as #1889 and #1483

bryanculver:
Issues like cause PRs like this: https://github.com/nautobot/nautobot/pull/2754#issuecomment-1311199557 when trying to adhere to standards defined here: https://github.com/nautobot/nautobot/issues/1606

bryanculver:
Closed via #3342

Issue #3213, opened 2023-02-02
Migrate filter tests to generic tests


gsnider2195:
### Proposed Changes

New generic tests were added to the [base filter test case](https://github.com/nautobot/nautobot/blob/bad9bb228da1d78fbc9b284b0a33510e673a9241/nautobot/core/testing/filters.py#L53) (`test_filters_generic` and `test_boolean_filters_generic`). Most `has_*` filter tests should already be covered by `test_boolean_filters_generic` and can be safely removed. Most other basic filter tests can be replaced by `test_filters_generic` but will need to be individually checked for compatibility with the generic test and added to the `generic_filter_tests` list on the test case. 

### Justification

Code clean up and reusing test code

timizuoebideri1:
Test case changes on https://github.com/nautobot/nautobot/pull/3342 can now be moved into `generic_filter_tests` since they are all now all `multi-value filters`

glennmatthews:
#5378 addresses the boolean filter tests only. More to be done.

PR #3342, opened 2023-02-20
Fix BaseFilterSet not using multiple choice filters

timizuoebideri1:
# Closes: #3214 
# What's Changed

- Override `BaseFilterSet.filter_for_lookup` to add logic that checks if the lookup type is exact and if the field has attribute `choices` (as only IntegerField/CharField with choices can have this attribute), and returns MulipleChoiceFilter if the conditions are meet. 
- I also removed this block of code `elif existing filter.extra.get("choices")` because we now expect the choice field to be `MulipleChoiceFilter`, which would require all relevant lookup expressions of a `MulipleChoiceFilter` to be included.

# TODO

- [ ] Unit, Integration Tests


jathanism (approved):

nautobot/extras/tests/test_filters.py
```diff
@@ -1084,7 +1084,7 @@ def setUpTestData(cls):
         ).validated_save()
 
     def test_type(self):
-        params = {"type": "one-to-many"}
+        params = {"type": ["one-to-many"]}
         self.assertEqual(self.filterset(params, self.queryset).qs.count(), 1)
```
bryanculver, line 1088: With all of these string to list changes I feel like this is going to be a breaking change, if not at least for all apps who's tests were expecting to filter on a choice of one.

bryanculver (changes requested):

nautobot/extras/tests/test_filters.py
glennmatthews: Yeah, it's only a breaking change to things that are programmatically using the filterset (UI and API filtering should "just work") but that's definitely worth calling out.

jathanism: A bug fix that is a breaking change! What is the world coming to!

timizuoebideri1: I guess illd have to rebase to next?

nautobot/utilities/filters.py
```diff
@@ -10,6 +10,7 @@
 from django.forms.utils import ErrorDict, ErrorList
 
 import django_filters
+from django_filters import MultipleChoiceFilter
```
glennmatthews, line 13: I'd prefer to skip this additional import and just reference it as `django_filters.MultipleChoiceFilter` where needed, personally.

nautobot/utilities/filters.py
```diff
@@ -720,6 +717,15 @@ def get_filters(cls):
...
         return filters
 
+    @classmethod
+    def filter_for_lookup(cls, field, lookup_type):
+        filter_class, params = super().filter_for_lookup(field, lookup_type)
+
+        if lookup_type == "exact" and getattr(field, "choices", None):
+            return MultipleChoiceFilter, {"choices": field.choices}
+
+        return filter_class, params
```
glennmatthews, line 727: Couldn't this be:

```suggestion
        if lookup_type == "exact" and getattr(field, "choices", None):
            return MultipleChoiceFilter, {"choices": field.choices}
        return super().filter_for_lookup(field, lookup_type)
```

nautobot/extras/tests/test_filters.py
bryanculver: Since this isn't a ground-breaking bug (I'm not seeing hard-stop issues for users) can I request this go into `next` so as to not induce unnecessary churn for apps and their automated tests?

nautobot/utilities/tests/test_filters.py
```diff
@@ -660,35 +662,36 @@ def test_multi_value_time_filter(self):
...
-        self.assertEqual(self.filters["multiplechoicefield__niew"].exclude, True)
-        self.assertEqual(self.filters["multiplechoicefield__re"].lookup_expr, "regex")
-        self.assertEqual(self.filters["multiplechoicefield__re"].exclude, False)
-        self.assertEqual(self.filters["multiplechoicefield__nre"].lookup_expr, "regex")
-        self.assertEqual(self.filters["multiplechoicefield__nre"].exclude, True)
-        self.assertEqual(self.filters["multiplechoicefield__ire"].lookup_expr, "iregex")
-        self.assertEqual(self.filters["multiplechoicefield__ire"].exclude, False)
-        self.assertEqual(self.filters["multiplechoicefield__nire"].lookup_expr, "iregex")
-        self.assertEqual(self.filters["multiplechoicefield__nire"].exclude, True)
+        for field in ["multiplechoicefield", "charchoicefield"]:
```
glennmatthews, line 665: Am I missing something, or is there no test case for the existing `"choicefield"` filter?

nautobot/extras/tests/test_filters.py
glennmatthews: I'd be fine with keeping this in `develop` as long as we add a clear note in the release-notes.

nautobot/utilities/tests/test_filters.py
timizuoebideri1: Nope, or not that I could find

glennmatthews: Would you mind adding one then? :-)

timizuoebideri1, line 665:
```suggestion
        for field in ["multiplechoicefield", "charchoicefield", "choicefield"]:
```

bryanculver (approved):

glennmatthews (commented):
Does this have any impact on the use of these filters in the UI, or how they are presented in the advanced filtering form?

glennmatthews (changes requested):
Would like to clarify the impact of this on the UI and forms before it gets merged.

jathanism:
> Would like to clarify the impact of this on the UI and forms before it gets merged.

With screenshots? :)

timizuoebideri1:
It does not seem to have any impact on UI filtering both on Default and Advance

Screenshot 2023-02-27 at 23 21 39
Screenshot 2023-02-27 at 23 21 45
Screenshot 2023-02-27 at 23 21 49
Screenshot 2023-02-27 at 23 21 57
Screenshot 2023-02-27 at 23 23 07

nautobot/dcim/tests/test_filters.py
```diff
@@ -1087,7 +1087,7 @@ def test_outer_unit(self):
         with self.subTest():
             self.assertEqual(Rack.objects.exclude(outer_unit="").count(), 3)
         with self.subTest():
-            params = {"outer_unit": RackDimensionUnitChoices.UNIT_MILLIMETER}
+            params = {"outer_unit": [RackDimensionUnitChoices.UNIT_MILLIMETER]}
```
glennmatthews, line 1090: Should be able to change most of these affected tests to entries in `generic_filter_test` rather than separate test functions now that these are multi-value filters, though it may take a bit of modifying to the test data to make it diverse enough.

glennmatthews (approved):

nautobot/dcim/tests/test_filters.py
timizuoebideri1: It does not need to be done on this PR right?

glennmatthews: I won't block this PR on waiting for this change but it should be a pretty straightforward one to do. Fine with me if you want to open a new issue or add a comment on #3213 calling out these tests.

Issue #3066, opened 2023-01-03
Fix dcim models FKs and related names


gsnider2195:
See spike https://github.com/nautobot/nautobot/issues/2960#issuecomment-1349809073 for details

gsnider2195:
I am changing `BaseInterface.parent_interface` and related name `child_interfaces` as part of #3069

glennmatthews:
Implemented in #3344 for 2.0.

Issue #1483, opened 2022-03-11
Assert that `FilterSet` filter fields are law and derive forms from them automatically


jathanism:
### Proposed Changes

I propose that we refactor ALL `FilterSet` objects to be authoritative for object filtering as  a superset of all desired object fields or custom filters (e.g. Boolean membership tests). Internally FilterSet objects generate their own `Form` class used for validating that incoming query filtering parameters are valid. When `filterset.is_valid()` is called, this in fact internally calls `filterset.form.is_valid()`. Additionally the validated data for the filtered queryset is provided by the internal `filterset.form.cleaned_data`. 

So: The complete proposal is asserting that filtersets are law, and filter forms are just generated from the filtersets and have their desired fields reduced at render time using something defined on the model meta e.g.`Device.Meta.ui_filter_fields = [...]`. And then we eliminate all custom `FilterForm` classes in exchange for model metadata paired with factory methods to generate filter forms when they are needed, derived from the filter sets. 

There would need to be assertion that any given filter field on a `FilterSet` class must be of the correct type (e.g. multiple-choice) and that the field it emits for the UI filter form is also of the correct type.  **Any custom type coercions of filter fields on `FilterSet` classes would then become the exception and not the norm**. 

Additionally this would eliminate the need for bespoke `FooFilterForm` classes because **we would just use the form already generated by the `FilterSet` internally** and merely reduce down the fields used at render time.

There would likely be some other improvements to gluing model objects to their `FilterSet` class. For example consider how often we see this pattern in `FilterSet` classes:

```python
    def search(self, queryset, name, value):
        if not value.strip():
            return queryset
        return queryset.filter(
            Q(name__icontains=value)
            | Q(slug__icontains=value)
            | Q(description__icontains=value)
            | Q(content_type__app_label__icontains=value)
            | Q(content_type__model__icontains=value)
        )
```

What if we had another model `Meta` attribute called `filter_search_fields`, such as:

```python
class MyModel(PrimaryModel):
    class Meta:
        filter_search_fields = ["name", "slug", "description", "content_type__app_label", "content_type__model"]
```

And then there could be a generic `search` method on `BaseFilterSet`:

```python
    def search(self, queryset, name, value):
        if not value.strip():
            return queryset
        query = Q()
        for search_field in queryset.model._meta.filter_search_fields:
            query |= Q(**{f"{search_field}__icontains": value})

        return query
```

### Justification

[...]

PR #3344, opened 2023-02-21
DCIM ForeignKey related_name fixups and related changes

glennmatthews:
# Closes: #3066 
# What's Changed

See changelog fragments for details. In brief this:
- renames a whole bunch of `related_name` strings for better self-consistency (#3066)
- introduces a new `ForeignKeyWithAutoRelatedName` helper class that we can use on abstract base classes to get `related_name` values like `console_port_templates` instead of `dcim_consoleporttemplate_related`
   - [x] (opened #3351) Issue to update `ConfigContextSchema` to use this helper so we can rename its `dcim_device_related` and `virtualization_virtualmachine_related` related-names.
- Converts `dcim/tests/test_filters` to use Gary's awesome `generic_filter_tests` for all filters that it currently can be applied to.
- [x] ~Currently the DeviceType YAML import test is failing as I haven't updated the test to account for the renamed fields on `FrontPortTemplate` and `PowerOutletTemplate`. This pair of changes breaks compatibility with the devicetype-library repository so I want to carefully consider our approach here.~ (solved, see https://github.com/nautobot/nautobot/pull/3344/commits/f8401ae2e87ecca78e73e8a4bf6ad78f9740662d)
- Fixes a minor bug in the example-plugin banner code that showed up when running tests with `--no-buffer`
- Fixes an intermittent failure in the extras test for the `nautobot-server runjob` command.
- Fixes some incorrect logic in our generic filter code that was treating FloatFields and DecimalFields as if they were integer fields.

# TODO

- [x] Unit, Integration Tests


nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -21,40 +21,74 @@
...
+|                       | `terminations`                          | `circuit_terminations`                         |
+|                       | `type`                                  | `circuit_type`                                 |
+| Cluster               | `group`                                 | `cluster_group`                                |
+|                       | `type`                                  | `cluster_type`                                 |
+| ConfigContextSchema   | `device_set`                            | `dcim_device_related`                          |
+|                       | `virtualmachine_set`                    | `virtualization_virtualmachine_related`        |
+| Device                | `consoleports`                          | `console_ports`                                |
+|                       | `consoleserverports`                    | `console_server_ports`                         |
+|                       | `devicebays`                            | `device_bays`                                  |
+|                       | `device_role`                           | `role`                                         |
```
itdependsnetworks, line 44: I am trying to understand this one?

itdependsnetworks:
Can we document the best practices that are related here? The ones I see are:

* has_* format
* the *_count format
* Assurance that there is not multiple words without an underscore
* Why have vlan_group and device_bay but not device_role?

nautobot/docs/development/best-practices.md
```diff
@@ -210,73 +250,76 @@ class UserFilter(NautobotFilterSet):
...
 
-- Filters **must** declare [`field_name`](https://django-filter.readthedocs.io/en/stable/ref/filters.html#field-name) when they have a different name than the underlying model field they are referencing. Where possible the suffix component of the filter name **must** map directly to the underlying field name.
+    is_virtual_chassis_member = RelatedMembershipBooleanFilter(
+        field_name="virtual_chassis",
+        label="Is a virtual chassis member",
+    )
+```
 
-  For example, `DeviceFilterSet.has_console_ports` could be better named, to assert that the filter name following the `has_` prefix is a one-to-one mapping to the underlying model's related field name (`consoleports`) therefore `field_name` must point to the field name as defined on the model:
+- Whenever possible otherwise, filter names **must** correspond exactly to the underlying model field they are referencing.
```
itdependsnetworks, line 303: Perhaps handle like Boolean filters and declare the must here and the exception below?

nautobot/docs/installation/upgrading-from-nautobot-v1.md
itdependsnetworks: Thanks for that explanation.

nautobot/docs/development/best-practices.md
```diff
@@ -74,7 +68,44 @@ The intended audience for the `validated_save()` convenience method is Job autho
...
+Rack.rack_group
+DeviceType.console_server_port_templates
+Device.ip_addresses
+```
+
+#### Foreign Key `related_name` for Abstract Model Classes
+
++/- 2.0.0
+
+If an abstract model class defines a foreign key to a concrete model class, Django's default `related_name` functionality doesn't provide great options - the best you could normally do for a `related_name` on a `ForeignKey` from an abstract base class, for example, would be `"%(class)ss"` (potentially resulting in related names like `"devices"` or, less optimally, `"ipaddresss"`) or `"%(app_label)s_%(class)s_related"` (resulting in related names like `"dcim_device_related"` or `"ipam_ipaddress_related"`, which while at least consistent, are rather clunky).
```
itdependsnetworks, line 99:
```suggestion
If an abstract model class defines a foreign key to a concrete model class, Django's default `related_name` functionality doesn't provide great options - the best you could normally do for a `related_name` on a `ForeignKey` from an abstract base class, for example, would be `"%(class)s"` (potentially resulting in related names like `"devices"` or, less optimally, `"ipaddresss"`) or `"%(app_label)s_%(class)s_related"` (resulting in related names like `"dcim_device_related"` or `"ipam_ipaddress_related"`, which while at least consistent, are rather clunky).
```

itdependsnetworks:
```suggestion
If an abstract model class defines a foreign key to a concrete model class, Django's default `related_name` functionality doesn't provide great options - the best you could normally do for a `related_name` on a `ForeignKey` from an abstract base class, for example, would be `"%(class)ss"` (potentially resulting in related names like `"devices"` or, less optimally, `"ipaddresss"`) or `"%(app_label)s_%(class)s_related"` (resulting in related names like `"dcim_devices_related"` or `"ipam_ipaddresss_related"`, which while at least consistent, are rather clunky).
```

nautobot/dcim/models/device_component_templates.py
```diff
@@ -195,12 +195,12 @@ class PowerOutletTemplate(ComponentTemplateModel):
...
 
     type = models.CharField(max_length=50, choices=PowerOutletTypeChoices, blank=True)
-    power_port = models.ForeignKey(
+    power_port_template = models.ForeignKey(
         to="dcim.PowerPortTemplate",
         on_delete=models.SET_NULL,
         blank=True,
         null=True,
-        related_name="poweroutlet_templates",
+        related_name="power_outlet_templates",
```
itdependsnetworks, line 203: I am sure this makes sense, but why outlet and not ports?

glennmatthews: `related_name` is what the reverse-relation gets named from the related model back to this one. So here we have:

- (variable name given on `PowerOutletTemplate`) `PowerOutletTemplate.power_port_template`, a ForeignKey from a given `PowerOutletTemplate` to a single `PowerPortTemplate`
- (from the given `related_name`) `PowerPortTemplate.power_outlet_templates`, a [Manager](https://docs.djangoproject.com/en/3.2/topics/db/queries/#backwards-related-objects) back from a given `PowerPortTemplate` to the set of zero or more `PowerOutletTemplate` that reference it

nautobot/docs/development/best-practices.md
glennmatthews: This is correct as is, this feature is just confusing. :-) `%(class)ss` in the first example is actually `%(class)s` followed by a literal `s`, hence why you would get `device`+ `s` or `ipaddress` + `s`. Conversely, in the second example we're just using `%(app_label)s` and `%(class)s` (no extra `s`) so the singular names are correct. (Both of these are patterns that have been used in the code base before, they aren't totally random examples.)

nautobot/docs/development/best-practices.md
glennmatthews: Not sure I understand the suggestion. Can you provide more details?

nautobot/docs/development/best-practices.md
bryanculver: @glennmatthews So what you're saying is `%(class)s` => `Device` and `%(class)ss` => `Devices` where `class` is Device?

nautobot/core/models/fields.py
```diff
@@ -133,6 +134,28 @@ def get_slug_fields(self, model_instance, lookup_value):
             return ""
 
 
+class ForeignKeyWithAutoRelatedName(models.ForeignKey):
+    """
+    Extend base ForeignKey functionality to create a smarter default `related_name`.
+
+    For example, "ip_addresses" instead of "ipaddresss".
+
+    Primarily useful for cases of abstract base classes that define ForeignKeys, such as dcim.ComponentModel.
```
jathanism, line 143:
```suggestion
    Primarily useful for cases of abstract base classes that define ForeignKeys, such as `nautobot.dcim.models.device_components.ComponentModel`.
```

nautobot/core/utils/filtering.py
```diff
@@ -110,6 +110,7 @@ def get_filterset_parameter_form_field(model, parameter):
 
     # TODO(Culver): We are having to replace some widgets here because multivalue_field_factory that generates these isn't smart enough
     if isinstance(field, NumberFilter):
+        # TODO(Glenn): This is wrong for MultiValueDecimalFilter and MultiValueFloatFilter filters!
```
jathanism, line 113: What's the fix? More specific inheritance check?

jathanism (approved):
Me happy

nautobot/docs/development/best-practices.md
lampwins: Yeah this is old-school string parameterization/formatting, where the first `s` is denoting the type of the parameter `class` which is `str` in this case. If `class` were an int you would use `d`.

nautobot/core/utils/filtering.py
glennmatthews: I think this falls under the #1483 umbrella - really our `BaseFilterSet` should generate valid forms on its own, obviating the need for this function to override the form fields on a case-by-case basis. #3214 is also related.

glennmatthews: Easy enough to fix for now though.

nautobot/docs/development/best-practices.md
glennmatthews: > @glennmatthews So what you're saying is `%(class)s` => `Device` and `%(class)ss` => `Devices` where `class` is Device?

To be pedantic, `%(class)s` => `device` and `%(class)ss` => `devices`, yes.

Similarly for "IPAddress" this would result in `ipaddress` and `ipaddresss`.

Issue #2958, opened 2022-12-05
Site, Region -> Location: Remove Site, Region Models


bryanculver:

HanlinMiao:
Closed by #3313

PR #3350, opened 2023-02-22
Add Nautobot App migration docs for Region and Site to Location data migration

HanlinMiao:
# Closes: #2958 
# What's Changed

Added `region-and-site-data-migration-guide.md` in `docs/installation` folder and added references to it in `upgrading-from-nautobot-v1.md`, `version-2.0.md` and `mkdocs.yml`.

# TODO

- [ ] Unit, Integration Tests


nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -92,6 +92,9 @@ The `ipam.Role`, `dcim.RackRole`, and `dcim.DeviceRole` models have been removed
 
 The `dcim.Region` and `dcim.Site` models have been removed and replaced by `dcim.Location` model. This means that any references to the removed models in the code now use the `dcim.Location` model instead with `LocationType` "Site" and "Region".
 
+!!! note
```
glennmatthews, line 95:
```suggestion
!!! important
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
+# Region and Site Related Data Model Migration Guide
+
+`Region` and `Site` models are collapsed into `Location` in Nautobot V2.0. As a result, all the `Region` and `Site` related data models are being migrated to use `Location` as well. Below is a comprehensive guide for Nautobot App developers to migrate their `Region` and `Site` related data models to `Location`.
```
glennmatthews, line 3: I feel like the first and second sentences here are redundant/repetitive but I'm not sure what to suggest as a rewrite.

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+
+We will be using `ExampleModel` as a relatively simple and hands-on example throughout this guide to better your understanding of the migration process.
+
+## Add Location Fields to Site and Region Related Data Models
+
+`ExampleModel` right now has a `site` ForeignKey field but it does not have a `location` ForeignKey field. We need to add one before we run any other migrations in this guide.
```
glennmatthews, line 9:
```suggestion
If the `ExampleModel` currently has a `site` ForeignKey field but it does not have a `location` ForeignKey field, you will need to add the `location` field before any other migrations in this guide.
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+
+!!! note
+    You can skip this step your data models already have both `site`/`region` fields and a `location` field.
```
glennmatthews, line 12:
```suggestion
    You can skip this step only if your data models already have both a `site` (or `region`) field and a `location` field.
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+    name = models.CharField(max_length=20, help_text="The name of this Example.")
+    number = models.IntegerField(default=100, help_text="The number of this Example.")
+    csv_headers = ["name", "number"]
+
+    class Meta:
+        ordering = ["name"]
+...
+```
+
+**DO NOT** delete the `site` ForeignKey field yet.
```
glennmatthews, line 28:
```suggestion
**DO NOT** delete the `site` ForeignKey field yet. As a first step, just add a `ForeignKey` to `dcim.Location` with all other arguments identical to the existing `dcim.Site` `ForeignKey`:
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+# models.py
+
+class ExampleModel(OrganizationalModel):
+    site = models.ForeignKey(to="dcim.Site", on_delete=models.PROTECT, blank=True, null=True)
+    name = models.CharField(max_length=20, help_text="The name of this Example.")
+    number = models.IntegerField(default=100, help_text="The number of this Example.")
+    csv_headers = ["name", "number"]
+
+    class Meta:
+        ordering = ["name"]
```
glennmatthews, line 24: Not sure `number`, `csv_headers`, and the `Meta` class add much to this example - consider removing them?

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+nautobot-server makemigrations [app_name] -n [migration_name]
+nautobot-server makemigrations example_plugin -n add_location_field_to_example_model
+```
+
+Apply the migration file by running:
+
+```shell
+nautobot-server migrate [app_name]
+nautobot-server migrate example_plugin
+```
```
glennmatthews, line 57:
````suggestion
```shell
# nautobot-server makemigrations [app_name] -n [migration_name]
nautobot-server makemigrations example_plugin -n add_location_field_to_example_model
```

Apply the migration file by running:

```shell
# nautobot-server migrate [app_name]
nautobot-server migrate example_plugin
```
````

File: nautobot/docs/plugins/region-and-site-data-migration-guide.md [hunk 8]
```diff
@@ -0,0 +1,285 @@
...
+
+## Create an Empty Migration File to Manually Migrate the Data
```
glennmatthews, line 59: I'm concerned that "manually migrate" will scare people off here. :-) The migration will still be automated, we'll just need to write some code that isn't auto-generated. Perhaps:

```suggestion
## Create an Empty Migration File and Write the Data Migration
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+
+After you make sure that all `Site`/`Region` related models have `location` fields on them, it is time to migrate `Site`/`Region` Data to `Location`.
```
glennmatthews, line 61:
```suggestion
After you make sure that all `Site`/`Region` related models have `location` fields on them, it is time to migrate `Site`/`Region` references in your data to `Location`.
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+
+We have to create an empty migration file and run an manual migration.
```
glennmatthews, line 63:
```suggestion
Django doesn't automatically know how to do this; we have to create an empty migration file and write the migration script ourselves.
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+
+We can do that by creating a migration file first:
+
+```shell
+nautobot-server makemigrations [app_name] -n [migration_file_name] --empty
```
glennmatthews, line 68:
```suggestion
# nautobot-server makemigrations [app_name] -n [migration_file_name] --empty
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+    class Meta:
+        ordering = ["name"]
+...
+```
+
+Make the migration file by running:
+
+```shell
+nautobot-server makemigrations [app_name] -n [migration_name]
+nautobot-server makemigrations example_plugin -n add_location_field_to_example_model
```
glennmatthews, line 49: Let's change `example_plugin` to `example_app` throughout this document?

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+from django.db import migrations
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("example_plugin", "0007_add_location_field_to_example_model"),
+    ]
+    operations = []
+```
+
+First we need to add a mandatory dependency to the migration file, namely `("dcim", "0030_migrate_region_and_site_data_to_locations")`. This dependent migration is very important as it creates the `Site`/`Region` `LocationType` locations for you to migrate your data to.
```
glennmatthews, line 87:
```suggestion
First we need to add a mandatory dependency to a Nautobot 2.0 migration file, namely `("dcim", "0030_migrate_region_and_site_data_to_locations")`. This dependent migration is very important as it creates the `Location` and `LocationType` records corresponding to the existing `Site`/`Region` records, which you will need to reference to migrate your data.
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+
+**Without it, your manual migration might not work!**
+
+```python
+# Generated by Django 3.2.17 on 2023-02-22 15:38
+# 0008_migrate_example_model_data_from_site_to_location
+
+from django.db import migrations
+
+class  Migration(migrations.Migration):
```
glennmatthews, line 97:
```suggestion
class Migration(migrations.Migration):
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("example_plugin", "0007_add_location_field_to_example_model"),
+    ]
+    operations = []
+```
+
+First we need to add a mandatory dependency to the migration file, namely `("dcim", "0030_migrate_region_and_site_data_to_locations")`. This dependent migration is very important as it creates the `Site`/`Region` `LocationType` locations for you to migrate your data to.
+
+**Without it, your manual migration might not work!**
```
glennmatthews, line 89:
```suggestion
**Without it, your data migration might not work!**
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+    dependencies = [
+        # The dcim migration creates the Site Type and Region Type Locations that
+        # your data models are migrating to. It has to be run **before** this migration.
+        ("dcim", "0030_migrate_region_and_site_data_to_locations"),
+        ("example_plugin", "0007_add_location_field_to_example_model"),
+    ]
+    operations = []
+```
+
+Now we can write the function that does the manual migration. We need to query `ExampleModel` instances that have non-null `site` field and point their `location` field to "Site" `LocationType` locations that have the same names and other attributes as their respective `Sites`.
```
glennmatthews, line 107:
```suggestion
Now we can write the function that will perform the data migration. We need to query `ExampleModel` instances that have non-null `site` field and point their `location` field to the new `Location` records that have the same names and other attributes as their respective `Sites`.
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+
+```python
+
+# Generated by Django 3.2.17 on 2023-02-22 15:38
+# 0008_migrate_example_model_data_from_site_to_location
+
+from django.db import migrations
+
+def migrate_example_model_data_to_locations(apps, schema_editor):
+
```
glennmatthews, line 119:
```suggestion
    # Always use the provided `apps` to look up models, rather than importing them directly!
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+    ExampleModel = apps.get("example_plugin", "examplemodel")
+    LocationType = apps.get("dcim", "locationtype")
+    Location = apps.get("dcim", "location")
+    # Get "Site" LocationType
+    site_location_type = LocationType.objects.get(name="Site")
+
+    if ExampleModel.objects.exists():
```
glennmatthews, line 126: I'd be tempted to leave this check out for simplicity - if the queryset below evaluates to empty, the for loop just is a no-op, right?

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+        # Query ExampleModel instances with non-null site field
+        example_models = ExampleModel.objects.filter(site__isnull=False).select_related("site", "location")
```
glennmatthews, line 128: Should we also check for `location__isnull=True` just to make sure we don't overwrite instances that already have a location set?

(edit) I see the admonition to this effect below, but still wonder if we should just *always* include it?

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+        for example_model in example_models:
+            site_name = example_model.site.name
+            # Point the location field to the corresponding "Site" LocationType Location
+            # with the same name.
+            example_model.location = Location.objects.get(location_type=site_location_type, name=site_name)
+            ExampleModel.objects.bulk_update(example_models, ["location"], 1000)
```
glennmatthews, line 134:
```suggestion
        ExampleModel.objects.bulk_update(example_models, ["location"], 1000)
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+
+class  Migration(migrations.Migration):
```
glennmatthews, line 136:
```suggestion
class Migration(migrations.Migration):
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+    ```
+
+Finally, we need to add `migrations.RunPython` to the `operations` attribute in the migration class to execute this function when the migration is applied:
+
+```python
+
+# Generated by Django 3.2.17 on 2023-02-22 15:38
+# 0008_migrate_example_model_data_from_site_to_location
+
+from  django.db  import  migrations
```
glennmatthews, line 163:
```suggestion
from django.db import migrations
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+            code=migrate_example_model_data_to_locations,
+            reverse_code=migrations.operations.special.RunPython.noop,
+        )
+    ]
+
+```
+
+## Remove Site/Region Related Fields from Migrated Data Models
+
+After you data migration is successful, we need to remove the `site`/`region` fields from your data model. You can do that by simply removing the `site`/`region` attributes from your model class:
```
glennmatthews, line 202:
```suggestion
After the data migration is successful, we need to remove the `site`/`region` fields from your data model so that Nautobot will be able to remove the `Site` and `Region` models. You can do that by simply removing the `site`/`region` attributes from your model class:
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+
+```python
+
+# models.py
+class  ExampleModel(OrganizationalModel):
```
glennmatthews, line 207:
```suggestion
class ExampleModel(OrganizationalModel):
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+```
+
+After removing the `site` attribute, make the migration file by running:
+
+```shell
+
+nautobot-server makemigrations [app_name] -n [migration_name]
+nautobot-server makemigrations example_plugin -n remove_site_field_from_example_model
+
+```
```
glennmatthews, line 227:
````suggestion
```shell
# nautobot-server makemigrations [app_name] -n [migration_name]
nautobot-server makemigrations example_plugin -n remove_site_field_from_example_model
```
````

File: nautobot/docs/plugins/region-and-site-data-migration-guide.md [hunk 26]
```diff
@@ -0,0 +1,285 @@
...
+            model_name='examplemodel',
+            name='site',
+        ),
+    ]
+
+```
+
+!!! important
+    Before you apply this migration, you have to add an `run_before` attribute in this migration file to make sure that you remove `site`/`region` fields before `Site` and `Region` models themselves are removed. **Without it, your migration files might be out of order and your app will not start**.
+
```
glennmatthews, line 255: This raises an important question - how does this work if we have these migrations (including the `run_before`) in an app **that we install for the first time AFTER Nautobot has already migrated to 2.0**?

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+            model_name="examplemodel",
+            name="site",
+        ),
+    ]
+```
+
+Apply the migration file by running:
+
+```shell
+nautobot-server migrate [app_name]
```
glennmatthews, line 283:
```suggestion
# nautobot-server migrate [app_name]
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
jathanism, line 89: IMO make this a big WARNING admonition.

nautobot/docs/plugins/region-and-site-data-migration-guide.md
jathanism, line 107: Let's be cautious about using controlled vocabulary. The data migration must be created manually, but it's actually a data migration, so I think it would be better to describe that and then use "data migration" consistently.

```suggestion
Now we can write the function that does the data migration. We need to query `ExampleModel` instances that have non-null `site` field and point their `location` field to "Site" `LocationType` locations that have the same names and other attributes as their respective `Sites`.
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+            ExampleModel.objects.bulk_update(example_models, ["location"], 1000)
+
+class  Migration(migrations.Migration):
+    dependencies = [
+        # The dcim migration creates the Site Type and Region Type Locations that
+        # your data models are migrating to. It has to be run **before** this migration.
+        ("dcim", "0030_migrate_region_and_site_data_to_locations"),
+        ("example_plugin", "0007_add_location_field_to_example_model"),
+    ]
+    operations = []
```
jathanism, line 143: I think as you are building the code, it would be better to only show what is added, and then show the entire file towards the end. It feels like it's becoming like a "spot the difference" puzzle to follow the lines of code that are adding/changed as you go. Also it would be more concise to only show the smaller snippets.

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+        ("dcim", "0030_migrate_region_and_site_data_to_locations"),
+        ("example_plugin", "0007_add_location_field_to_example_model"),
+    ]
+    operations = [
+        migrations.RunPython(
+            # Execute the function
+            code=migrate_example_model_data_to_locations,
+            reverse_code=migrations.operations.special.RunPython.noop,
+        )
+    ]
```
jathanism, line 196: Following the previous suggestion about reducing the code, this is the only thing that changed from the previous code gate, so maybe only highlight this, as I had recommended?

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,285 @@
...
+
+```
```
jathanism, line 198: And then here at the end, show the entire migration file that was created.

nautobot/docs/plugins/region-and-site-data-migration-guide.md
jathanism, line 227: Same example command suggestion here.

nautobot/docs/release-notes/version-2.0.md
```diff
@@ -92,6 +92,8 @@ The `region` and `site` fields are being removed in the `filter` data of `Dynami
 
 Check out the API and UI endpoints changes incurred by the changes stated above in the ["Upgrading from Nautobot v1.X"](../installation/upgrading-from-nautobot-v1.md) guide.
 
+Check out this [migration guide](../plugins/region-and-site-data-migration-guide.md#region-and-site-related-data-model-migration-guide) to migrate your Nautobot Apps and data models from `Site` and `Region` to `Location`.
```
jathanism, line 95:
```suggestion
Check out the [Region and Site Related Data Model Migration Guide](../plugins/region-and-site-data-migration-guide.md#region-and-site-related-data-model-migration-guide) to learn how to migrate your Nautobot Apps and data models from `Site` and `Region` to `Location`.
```

jathanism (changes requested):
Suggestions for consistency, clarity, and conciseness.

nautobot/docs/plugins/region-and-site-data-migration-guide.md
HanlinMiao: Hmmm, I don't see a way for apps that are developed based on Nautobot v1.x to be installed after Nautobot has already migrated to v2.0.
1. Either revert Nautobot to v1.x, install the app, and then migrate to v2.0 together. (I don't even think the reversion is possible with the number of migrations)
2. Install the apps before migrating Nautobot to v2.0. 
3. Develop new apps that are the same as the old ones but v2.0 compatible?

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,246 @@
...
+    site = models.ForeignKey(to="dcim.Site", on_delete=models.PROTECT, blank=True, null=True)
+    location = models.ForeignKey(to="dcim.Location", on_delete=models.PROTECT, blank=True, null=True)
+    name = models.CharField(max_length=20, help_text="The name of this Example.")
+...
+```
+
+Make the migration file by running `nautobot-server makemigrations [app_name] -n [migration_name]`, for example:
+
+```shell
+nautobot-server makemigrations example_plugin -n add_location_field_to_example_model
```
bryanculver, line 38:
```suggestion
nautobot-server makemigrations example_app -n add_location_field_to_example_model
```

bryanculver (approved):
I feel like I know what to do in other apps after reading your guide. LGTM!

smk4664:
Having a reverse migration that is a no-op can be dangerous. I would recommend providing a real reverse migration with Example App instead of the no-op. If Site were a required field, reversing migrations may be impossible with a No-op.

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -92,6 +92,9 @@ The `ipam.Role`, `dcim.RackRole`, and `dcim.DeviceRole` models have been removed
 
 The `dcim.Region` and `dcim.Site` models have been removed and replaced by `dcim.Location` model. This means that any references to the removed models in the code now use the `dcim.Location` model instead with `LocationType` "Site" and "Region".
 
+!!! important
+    If you are an Nautobot App developer, please review the [Region and Site Related Data Model Migration Guide](../plugins/region-and-site-data-migration-guide.md#region-and-site-related-data-model-migration-guide) to learn how to migrate your apps and models from `Site` and `Region` to `Location`.
```
gsnider2195, line 96:
```suggestion
    If you are a Nautobot App developer, please review the [Region and Site Related Data Model Migration Guide](../plugins/region-and-site-data-migration-guide.md#region-and-site-related-data-model-migration-guide) to learn how to migrate your apps and models from `Site` and `Region` to `Location`.
```

bryanculver:
> Having a reverse migration that is a no-op can be dangerous. I would recommend providing a real reverse migration with Example App instead of the no-op. If Site were a required field, reversing migrations may be impossible with a No-op.

It is exceedingly unlikely that we will restore their database in as-good of a state with a reverse from 2.0. If you make any data changes at all post-migration, you will be re-inventing data and relationships that didn't exist in the roll-back.

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,246 @@
...
+
+Apply the migration file by running `nautobot-server migrate [app_name]`, for example:
+
+```shell
+nautobot-server migrate example_app
+```
+
+## Create an Empty Migration File and Write the Data Migration
+
+After you make sure that all `Site`/`Region` related models have `location` fields on them, it is time to migrate `Site`/`Region` references in your data to `Location`.
```
gsnider2195, line 49:
```suggestion
After you make sure that all `Site`/`Region` related models have `Location` fields on them, it is time to migrate `Site`/`Region` references in your data to `Location`.
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,246 @@
...
+
+Django doesn't automatically know how to do this; we have to create an empty migration file and write the migration script ourselves. This is also known as a [data migration](https://docs.djangoproject.com/en/4.1/topics/migrations/#data-migrations).
```
gsnider2195, line 51:
```suggestion
Django doesn't automatically know how to do this; we have to create an empty migration file and write the migration script ourselves. This is also known as a [data migration](https://docs.djangoproject.com/en/3.2/topics/migrations/#data-migrations).
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,246 @@
...
+
+Create a migration file first by running `nautobot-server makemigrations [app_name] -n [migration_file_name] --empty`, for example:
+
+```shell
+nautobot-server makemigrations example_app -n migrate_app_data_from_site_to_location --empty
+```
+
+The empty migration file will look like this with the only dependency being our previous migration that added a `location` ForeignKey field to our `ExampleModel`:
```
gsnider2195, line 59:
```suggestion
The empty migration file will look like this with the only dependency being our previous migration that added a `Location` ForeignKey field to our `ExampleModel`:
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,246 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("example_app", "0007_add_location_field_to_example_model"),
+    ]
+    operations = []
+```
+
+!!! warning
+    First we need to add a mandatory dependency to a Nautobot 2.0 migration file, namely `("dcim", "0030_migrate_region_and_site_data_to_locations")`. This dependent migration is very important as it creates the `Location` and `LocationType` records corresponding to the existing `Site`/`Region` records, which you will need to reference to migrate your data.
+    **Without it, you data migration might not work!**
```
gsnider2195, line 76:
```suggestion
    **Without it, your data migration might not work!**
```

smk4664:
> > Having a reverse migration that is a no-op can be dangerous. I would recommend providing a real reverse migration with Example App instead of the no-op. If Site were a required field, reversing migrations may be impossible with a No-op.
> 
> 
> 
> It is exceedingly unlikely that we will restore their database in as-good of a state with a reverse from 2.0. If you make any data changes at all post-migration, you will be re-inventing data and relationships that didn't exist in the roll-back.

But, if someone tries to uninstall a plugin, and the rollbacks fail due to a missing reverse migration, that would be a concern.

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,246 @@
...
+    ExampleModel = apps.get("example_app", "examplemodel")
+    LocationType = apps.get("dcim", "locationtype")
+    Location = apps.get("dcim", "location")
+    # Get "Site" LocationType
+    site_location_type = LocationType.objects.get(name="Site")
+
+    if  ExampleModel.objects.exists():
+        # Query ExampleModel instances with non-null site field
+        example_models = ExampleModel.objects.filter(site__isnull=False).select_related("site", "location")
+        for  example_model  in  example_models:
```
gsnider2195, line 142:
```suggestion
        for  example_model in example_models:
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,246 @@
...
+            # Execute the function
+            code=migrate_example_model_data_to_locations,
+            reverse_code=migrations.operations.special.RunPython.noop,
+        )
+    ]
+```
+
+## Remove Site/Region Related Fields from Migrated Data Models
+
+After the data migration is successful, we need to remove the `site`/`region` fields from your data model so that Nautobot will be able to remove the `Site` and `Region` models. You can do that by simply removing the `site`/`region` attributes from your model class:
```
gsnider2195, line 167: Maybe call out here that this needs to go in a separate migration file from the first, since the first is migrating data and this is changing the schema?

jathanism (changes requested):
This is looking great and much easier to follow now! Only thing is this document appears to be hard to find seeing as it's not a part of the table of contents in mkdocs?

The URL is `/plugins/region-and-site-data-migration-guide/#region-and-site-related-data-model-migration-guide` but outside of navigating to the 1.x to 2.0 migration guide, the only link to this I could find is in the admonition:

Cursor_and_Upgrading_Nautobot_from_v1_X_to_v2_0_-_Nautobot_Documentation

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,247 @@
...
+            # Execute the function
+            code=migrate_example_model_data_to_locations,
+            reverse_code=migrations.operations.special.RunPython.noop,
+        )
+    ]
+```
+
+## Remove Site/Region Related Fields from Migrated Data Models
+
+After the data migration is successful, we need to remove the `site`/`region` fields from your data model so that Nautobot will be able to remove the `Site` and `Region` models. Note that we need to remove those attributes in a separate migration file from the previous since they are two different operations. You can do that by simply removing the `site`/`region` attributes from your model class:
```
glennmatthews, line 166:
```suggestion
After the data migration is successful, we need to remove the `site`/`region` fields from your data model so that Nautobot will be able to remove the `Site` and `Region` models. Note that we need to remove those attributes in a separate migration file from the previous one, as it's never a good practice to combine data migrations and schema migrations in the same file. You can do this by simply removing the `site`/`region` attributes from your model class:
```

nautobot/docs/plugins/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,247 @@
...
+
+    operations = [
+        migrations.RemoveField(
+            model_name="examplemodel",
+            name="site",
+        ),
+    ]
+```
+
+Apply the migration file by running `nautobot-server migrate [app_name]`, for example:
```
glennmatthews, line 243: Is this applying all three of the migrations that were just authored in a single pass? I don't see this step earlier in the file.


What version of Nautobot itself needs to be running, and which of its migrations must and must not have already run for this migration to be successful? (Presumably the `example_app` migration will fail if `dcim.0034_remove_region_and_site` has already run, for example?)

glennmatthews (changes requested):

HanlinMiao:
> > > > > Having a reverse migration that is a no-op can be dangerous. I would recommend providing a real reverse migration with Example App instead of the no-op. If Site were a required field, reversing migrations may be impossible with a No-op.
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > It is exceedingly unlikely that we will restore their database in as-good of a state with a reverse from 2.0. If you make any data changes at all post-migration, you will be re-inventing data and relationships that didn't exist in the roll-back.
> > 
> > 
> > > 
> > 
> > 
> > > But, if someone tries to uninstall a plugin, and the rollbacks fail due to a missing reverse migration, that would be a concern.
> > 
> > 
> > Very valid point! Definitely need to consider the uninstall case, but that might need to be a whole separate PR after this one?
> 
> Sure! I just want to be careful recommending noop. I have been bitten in the past with migrations I could not reverse.

Opened up an issue to track this. https://github.com/nautobot/nautobot/issues/3374

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,405 @@
+# Region and Site to Location Migration Guide
+
+In Nautobot 2.0.0, all the `Region` and `Site` related data models are being migrated to use `Location`. Below is a comprehensive guide for Nautobot users and Nautobot App developers to migrate their `Region` and `Site` related models and apps to `Location`.
+
+## Migrate ObjectPermission instances in Nautobot from Region and Site to Location
+
+Nautobot admins need to migrate `Site` and `Reigon` Related `ObjectPermission` instances to `Location`. The correct way to do it are documented below with practical examples.
```
glennmatthews, line 7:
```suggestion
Nautobot admins need to migrate `Site` and `Region` Related `ObjectPermission` instances to `Location`. The correct way to do it are documented below with practical examples.
```

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,405 @@
...
+
+### Region Specific ObjectPermission
+
+Add `Location` ContentType to `object_types` field of the `ObjectPermission` if it is not already present.
+
+Since `Location` has all the filters that `Region` had and they retain the same functionalities. We do not need to modify the `constraints` field of any `Region` specific `ObjectPermission` instances.
```
glennmatthews, line 13: Might be good to talk about adding a `location_type` constraint to the permission, e.g. if the old permission was one that allowed for editing Regions but not editing Sites? Same comment on the next section as well.

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,405 @@
...
+
+In Nautobot 2.0.0, all the `Region` and `Site` related data models are being migrated to use `Location`. Below is a comprehensive guide for Nautobot users and Nautobot App developers to migrate their `Region` and `Site` related models and apps to `Location`.
+
+## Migrate ObjectPermission instances in Nautobot from Region and Site to Location
+
+Nautobot admins need to migrate `Site` and `Reigon` Related `ObjectPermission` instances to `Location`. The correct way to do it are documented below with practical examples.
+
+### Region Specific ObjectPermission
+
+Add `Location` ContentType to `object_types` field of the `ObjectPermission` if it is not already present.
```
glennmatthews, line 11: Here and below, is "add" correct, or should the user "replace" the Region `object_types` entry with a Location entry?

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,405 @@
...
+    "region__parent__slug": "north-america",
+    "region__parent__slug__in": ["north-america", "europe"],
+    "region__parent__id": "9mkl123094fc2-0326-40db-b3a3-3afe2207c1ad",
+    "region__parent__id__in": ["9mkl123094fc2-0326-40db-b3a3-3afe2207c1ad", "82132yhks-875njjni121-0091213"],
+}
+```
+
+To modify the data correctly, we need to replace query filter strings to those of model class `Location`:
+    1. Replace all occurrences of "region" with "parent" in the **Key** portion (before ":") of the data **not Value** (after ":").
+    2. Since we are keeping the uuids of the newly created "Site"/"Region" type locations the same from those of the old `Site`/`Region` instances, we do not need to change the uuid values in "..__id" and "..__id__in" Keys.
```
glennmatthews, line 46:
```suggestion
    2. Since Nautobot carries over the UUIDs of the old `Site`/`Region` instances when creating the new "Site"/"Region" type `Location` instances, we do not need to change the UUID values in `...__id` and `...__id__in` Keys.
```

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,405 @@
...
+The old `constraints` field for `Site` specifc `ObjectPermission` instances might look like this:
+
+```json
+{
+    "name": "AMS01",
+    "name__in": ["AMS01", "ATL01", "EDX01"],
+    "slug": "ams01",
+    "slug__in": ["ams01", "atl01", "edx01"],
+    "id": "9121jfjnjas9fc2-0326-40db-b3a3-3af1231ad",
+    "id__in": ["9121jfjnjas9fc2-0326-40db-b3a3-3afe2207c1ad", "mklmjb79012mkjasd-skmk198721-091231"],
```
glennmatthews, line 28: Ok, these definitely aren't real/valid UUIDs. :-) Can you put some "real" UUIDs in here? I got distracted by seeing these nonsense strings.

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,405 @@
...
+    "region__slug": "united-states",
+    "region__slug__in": ["united-states", "greece", "england"],
+    "region__id": "9e0e9fc2-0326-40db-b3a3-3afe2207c1ad",
+    "region__id__in": ["9e0e9fc2-0326-40db-b3a3-3afe2207c1ad", "9e0e9fc2-0326-40db-2222-222222222222"],
+    "region__parent__name": "North America",
+    "region__parent__name__in": ["North America", "Europe"],
+    "region__parent__slug": "north-america",
+    "region__parent__slug__in": ["north-america", "europe"],
+    "region__parent__id": "9mkl123094fc2-0326-40db-b3a3-3afe2207c1ad",
+    "region__parent__id__in": ["9mkl123094fc2-0326-40db-b3a3-3afe2207c1ad", "82132yhks-875njjni121-0091213"],
```
glennmatthews, line 40: Here too, fake UUIDs. :-)

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,405 @@
...
+    "device__site__region__parent__slug__in": ["north-america", "europe", "south-america"],
+    "device__site__region__parent__id": "0kmskqjneq14-945840-2132f6-b964-9ekmwkq12c48ce0",
+    "device__site__region__parent__id__in": ["0kmskqjneq14-945840-2132f6-b964-9ekmwkq12c48ce0", "qwe0kmjn1lohg-12olm-kmlie-8as1231-km2l1321isdan234f5"],
+}
+```
+
+To modify the data correctly, we need to replace query filter strings to those of model class `Location`:
+    1. Replace all occurrences of "site" with "location" in the **Key** portion (before ":") of the data **not Value** (after ":").
+    2. Replace all occurrences of "region" with "parent" in the **Key** portion (before ":") of the data **not Value** (after ":").
+    3. Since we are keeping the uuids of the newly created "Site"/"Region" type locations the same from those of the old `Site`/`Region` instances, we do not need to change the uuid values in "..__id" and "..__id__in" Keys.
```
glennmatthews, line 103:
```suggestion
    3. Since Nautobot carries over the UUIDs of the old `Site`/`Region` instances when creating the new "Site"/"Region" type `Location` instances, we do not need to change the UUID values in `...__id` and `...__id__in` Keys.
```

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,405 @@
...
+    "device__site__slug": "ams01",
+    "device__site__slug__in": ["ams01", "atl01", "etx02"],
+    "device__site__id": "0ab47314-2944-45f6-b964-9e009fc48ce0",
+    "device__site__id__in": ["0ab47314-2944-45f6-b964-9e009fc48ce0", "b09545d4-6e2b-471e-8f07-27f25ca308f5"],
+    "device__site__region__name": "United States",
+    "device__site__region__name__in": ["United States", "United Kingdom", "Greece"],
+    "device__site__region__slug": "united-states",
+    "device__site__region__slug__in": ["united-states", "united-kingdom", "greece"],
+    "device__site__region__id": "0ab47314-123144-2132f6-b964-9ekmwkq12c48ce0",
+    "device__site__region__id__in": ["0ab47314-123144-2132f6-b964-9ekmwkq12c48ce0", "kml21345d4-6e123-471e-8as1231-27f25lmsahca308f5"],
```
glennmatthews, line 90: More fake UUIDs! Where will the deception end?! ;-)

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,405 @@
...
+    "device__location__parent__parent__name": "North America",
+    "device__location__parent__parent__name__in": ["North America", "Europe", "South America"],
+    "device__location__parent__parent__slug": "north-america",
+    "device__location__parent__parent__slug__in": ["north-america", "europe", "south-america"],
+    "device__location__parent__parent__id": "0kmskqjneq14-945840-2132f6-b964-9ekmwkq12c48ce0",
+    "device__location__parent__parent__id__in": ["0kmskqjneq14-945840-2132f6-b964-9ekmwkq12c48ce0", "qwe0kmjn1lohg-12olm-kmlie-8as1231-km2l1321isdan234f5"],
+}
+```
+
+### Other Data Model Specific ObjectPermission e.g. ObjectChange
```
glennmatthews, line 130: Use of ObjectChange as a permissions example here feels a bit contrived to me. If we're looking for one with a content-type reference and/or a generic foreign key, maybe something like ExportTemplate might be a better example?

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,405 @@
...
+{
+    "changed_object_type": "dcim.site",
+    "changed_object_id": "0ab47314-2944-45f6-b964-9e009fc48ce0",
+    "changed_object_id__in": ["0ab47314-2944-45f6-b964-9e009fc48ce0", "b09545d4-6e2b-471e-8f07-27f25ca308f5"],
+}
+```
+
+To modify the data correctly, we need to replace query filter strings to those of model class `Location`:
+    1. Replace all occurrences of "dcim.site" with "dcim.location" in the **Value** portion (after ":").
+    3. Since we are keeping the uuids of the newly created "Site"/"Region" type locations the same from those of the old `Site`/`Region` instances, we do not need to change the uuid values in "changed_object_id" and "changed_object_id__in" Keys.
```
glennmatthews, line 144:
```suggestion
    3. Since Nautobot carries over the UUIDs of the old `Site`/`Region` instances when creating the new "Site"/"Region" type `Location` instances, we do not need to change the UUID values in the `changed_object_id` and `changed_object_id__in` Keys.
```

glennmatthews (approved):
Nits only at this point.

(I just noticed the commit message on your last commit - what happened?)

HanlinMiao:
> Nits only at this point.
> 
> (I just noticed the commit message on your last commit - what happened?)

Oh no, I did `id` in my commit message, and it looks like it did a command substitution :(

nautobot/docs/installation/region-and-site-data-migration-guide.md
HanlinMiao: It seems like `owner_content_type` and `owner_object_id` only involve "extras.gitrepository" and `GitRepository` uuids. I might opt for `Note` model as an example.

glennmatthews:
> > Nits only at this point.
> > (I just noticed the commit message on your last commit - what happened?)
> 
> Oh no, I did `id` in my commit message, and it looks like it did a command substitution :(

Yeah if you're going to use backticks in the commit message, make sure you wrap the message in single-quotes rather than double-quotes. :-)

mkdocs.yml
```diff
@@ -111,6 +111,7 @@ nav:
           - Upgrading Nautobot: "installation/upgrading.md"
           - Upgrading Nautobot from v1.X to v2.0: "installation/upgrading-from-nautobot-v1.md"
           - Migrating from PostgreSQL to MySQL: "installation/migrating-from-postgresql.md"
+          - Migrating from Region and Site to Location: "installation/region-and-site-data-migration-guide.md"
```
glennmatthews, line 114: Perhaps a more intuitive ordering?

```suggestion
          - Migrating from PostgreSQL to MySQL: "installation/migrating-from-postgresql.md"
          - Upgrading Nautobot: "installation/upgrading.md"
          - Upgrading Nautobot from v1.X to v2.0: "installation/upgrading-from-nautobot-v1.md"
          - Migrating from Region and Site to Location: "installation/region-and-site-data-migration-guide.md"
```

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,408 @@
...
+    "region__slug": "united-states",
+    "region__slug__in": ["united-states", "greece", "england"],
+    "region__id": "f77f5706-e5b3-49e0-9749-b8f818319c40",
+    "region__id__in": ["f77f5706-e5b3-49e0-9749-b8f818319c40", "7df99dc2-283a-4130-8125-60b9ca293131"],
+    "region__parent__name": "North America",
+    "region__parent__name__in": ["North America", "Europe"],
+    "region__parent__slug": "north-america",
+    "region__parent__slug__in": ["north-america", "europe"],
+    "region__parent__id": "c1a816df-876f-44d4-8ea0-335898998780",
+    "region__parent__id__in": ["c1a816df-876f-44d4-8ea0-335898998780", "a68b0838-d7fb-416c-b4ba-a3e464e552ba"],
```
glennmatthews, line 40:
```suggestion
    "region__parent__id__in": ["c1a816df-876f-44d4-8ea0-335898998780", "a68b0838-d7fb-416c-b4ba-a3e464e552ba"]
```

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,408 @@
...
+    "parent__slug": "united-states",
+    "parent__slug__in": ["united-states", "greece", "england"],
+    "parent__id":  "f77f5706-e5b3-49e0-9749-b8f818319c40",
+    "parent__id__in": ["f77f5706-e5b3-49e0-9749-b8f818319c40", "7df99dc2-283a-4130-8125-60b9ca293131"],
+    "parent__parent__name": "North America",
+    "parent__parent__name__in": ["North America", "Europe"],
+    "parent__parent__slug": "north-america",
+    "parent__parent__slug__in": ["north-america", "europe"],
+    "parent__parent__id": "c1a816df-876f-44d4-8ea0-335898998780",
+    "parent__parent__id__in": ["c1a816df-876f-44d4-8ea0-335898998780", "a68b0838-d7fb-416c-b4ba-a3e464e552ba"],
```
glennmatthews, line 70:
```suggestion
    "parent__parent__id__in": ["c1a816df-876f-44d4-8ea0-335898998780", "a68b0838-d7fb-416c-b4ba-a3e464e552ba"]
```

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,408 @@
...
+    "device__site__region__slug": "united-states",
+    "device__site__region__slug__in": ["united-states", "united-kingdom", "greece"],
+    "device__site__region__id": "f1a79a3c-d980-40e1-979d-abdb0f83388e",
+    "device__site__region__id__in": ["f1a79a3c-d980-40e1-979d-abdb0f83388e", "6335a61e-503d-463c-99c2-9c87ef8354d9"],
+    "device__site__region__parent__name": "North America",
+    "device__site__region__parent__name__in": ["North America", "Europe", "South America"],
+    "device__site__region__parent__slug": "north-america",
+    "device__site__region__parent__slug__in": ["north-america", "europe", "south-america"],
+    "device__site__region__parent__id": "6695809c-b33b-4f12-b0de-a4969000434d",
+    "device__site__region__parent__id__in": ["6695809c-b33b-4f12-b0de-a4969000434d", "e51d07bb-3fcf-4306-9d87-6b1ff6dd6378"],
```
glennmatthews, line 97:
```suggestion
    "device__site__region__parent__id__in": ["6695809c-b33b-4f12-b0de-a4969000434d", "e51d07bb-3fcf-4306-9d87-6b1ff6dd6378"]
```

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,408 @@
...
+    "device__location__parent__slug": "united-states",
+    "device__location__parent__slug__in": ["united-states", "united-kingdom", "greece"],
+    "device__location__parent__id": "f1a79a3c-d980-40e1-979d-abdb0f83388e",
+    "device__location__parent__id__in": ["f1a79a3c-d980-40e1-979d-abdb0f83388e", "6335a61e-503d-463c-99c2-9c87ef8354d9"],
+    "device__location__parent__parent__name": "North America",
+    "device__location__parent__parent__name__in": ["North America", "Europe", "South America"],
+    "device__location__parent__parent__slug": "north-america",
+    "device__location__parent__parent__slug__in": ["north-america", "europe", "south-america"],
+    "device__location__parent__parent__id": "6695809c-b33b-4f12-b0de-a4969000434d",
+    "device__location__parent__parent__id__in": ["6695809c-b33b-4f12-b0de-a4969000434d", "e51d07bb-3fcf-4306-9d87-6b1ff6dd6378"],
```
glennmatthews, line 129:
```suggestion
    "device__location__parent__parent__id__in": ["6695809c-b33b-4f12-b0de-a4969000434d", "e51d07bb-3fcf-4306-9d87-6b1ff6dd6378"]
```

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,408 @@
...
+
+### Other Data Model Specific ObjectPermission e.g. Note
+
+The old `constraints` field for a `Site`/`Region` related data model's (e.g. `Note`) `ObjectPermission` instance might look like this:
+
+```json
+{
+    "assigned_object_type": "dcim.site",
+    "assigned_object_id": "932d94ee-5571-40a0-903f-4274fcfbed32",
+    "assigned_object_id__in": ["932d94ee-5571-40a0-903f-4274fcfbed32", "e383db9a-dd55-464d-9e56-2f18bc03b32c"],
```
glennmatthews, line 141:
```suggestion
    "assigned_object_id__in": ["932d94ee-5571-40a0-903f-4274fcfbed32", "e383db9a-dd55-464d-9e56-2f18bc03b32c"]
```

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,408 @@
...
+    1. Replace all occurrences of "dcim.site" and/or "dcim.region" with "dcim.location" in the **Value** portion (after ":") of the `assigned_object_type` Key.
+    2. Since Nautobot carries over the UUIDs of the old `Site`/`Region` instances when creating the new "Site"/"Region" type `Location` instances, we **do not** need to change the UUID values in the `assigned_object_id` and `assigned_object_id__in` Keys.
+
+The updated JSON data might look like this:
+
+```json
+{
+    "assigned_object_type": "dcim.location",
+    "assigned_object_id": "932d94ee-5571-40a0-903f-4274fcfbed32",
+    "assigned_object_id__in": ["932d94ee-5571-40a0-903f-4274fcfbed32", "e383db9a-dd55-464d-9e56-2f18bc03b32c"],
```
glennmatthews, line 155:
```suggestion
    "assigned_object_id__in": ["932d94ee-5571-40a0-903f-4274fcfbed32", "e383db9a-dd55-464d-9e56-2f18bc03b32c"]
```

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,408 @@
...
+        migrations.RemoveField(
+            model_name="examplemodel",
+            name="site",
+        ),
+    ]
+```
+
+!!!important
+    Before you apply the aforementioned migration files to your app in the following step, please make sure that Nautobot is on version 1.x and not on version 2.0. More importantly, please make sure that `("dcim", "0030_migrate_region_and_site_data_to_locations")` is applied to your Nautobot instance and that `("dcim", "0034_remove_region_and_site")` is **not** applied.
+
```
glennmatthews, line 401: Would be good to flesh this out a bit further with exact steps, something like:

- Start with a working Nautobot 1.5.x installation with baseline migrations all run
- Stop Nautobot server
- Update installed version of Nautobot to 2.0.0
- Run `nautobot-server migrate dcim 0030_migrate_region_and_site_data_to_locations` (?)
- Updated installed version of *all* Apps to include any necessary migration files as described above
- For each affected app, run `nautobot-server migrate <app_name>` (?)
- Run `nautobot-server migrate` to finish all migrations (?)
- Start Nautobot server

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,408 @@
...
+Replace `Location` ContentType to `object_types` field of the `ObjectPermission` if it is not already present and add `"location_type__name": "Region"` to `constraints` field if the old `ObjectPermission` only allows operations on `Region` instances and not on `Site` instances.
+
+Since `Location` has all the filters that `Region` had and they retain the same functionalities. We **do not** need to modify the `constraints` field of any `Region` specific `ObjectPermission` instances any further.
+
+### Site Specific ObjectPermission
+
+Replace `Location` ContentType to `object_types` field of the `ObjectPermission` if it is not already present and add `location_type__name: "Site"` to `constraints` field if the old `ObjectPermission` only allows operations on `Site` instances and not on `Region` instances.
+
+The old `constraints` field for `Site` specifc `ObjectPermission` instances might look like this:
```
glennmatthews, line 19:
```suggestion
The old `constraints` field for a `Site` specific `ObjectPermission` instance might look like this:
```

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,408 @@
...
+# models.py
+
+class ExampleModel(OrganizationalModel):
+    site = models.ForeignKey(to="dcim.Site", on_delete=models.PROTECT, blank=True, null=True)
+    location = models.ForeignKey(to="dcim.Location", on_delete=models.PROTECT, blank=True, null=True)
+    name = models.CharField(max_length=20, help_text="The name of this Example.")
+...
+```
+
+Make the migration file by running `nautobot-server makemigrations [app_name] -n [migration_name]`, for example:
```
glennmatthews, line 193: What state do the Nautobot migrations need to be in at this point? Still a 1.5.x instance, or should 2.0 be already installed (but not yet migrated) at this point?

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,408 @@
...
+    ]
+```
+
+!!! important
+    Before you apply this migration, you have to add a `run_before` attribute in this migration file to make sure that you remove `site`/`region` fields before `Site` and `Region` models themselves are removed. **Without it, your migration files might be out of order and your app will not start**.
+
+```python
+    # Ensure this migration is run before the migration that removes Region and Site Models
+    run_before = [
+        # TODO we need to change the name after we reorder migration files
```
glennmatthews, line 367: When will this be done?

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -164,6 +164,9 @@ The `ipam.Role`, `dcim.RackRole`, and `dcim.DeviceRole` models have been removed
 
 The `dcim.Region` and `dcim.Site` models have been removed and replaced by `dcim.Location` model. This means that any references to the removed models in the code now use the `dcim.Location` model instead with `LocationType` "Site" and "Region".
 
+!!! important
+    If you are a Nautobot App developer, please review the [Region and Site Related Data Model Migration Guide](../installation/region-and-site-data-migration-guide.md#region-and-site-related-data-model-migration-guide-for-existing-nautobot-app-installations) to learn how to migrate your apps and models from `Site` and `Region` to `Location`.
```
glennmatthews, line 168: Perhaps?

```suggestion
    If you are a Nautobot App developer, or have any Apps installed that include data models that reference `Site` or `Region`, please review the [Region and Site Related Data Model Migration Guide](../installation/region-and-site-data-migration-guide.md#region-and-site-related-data-model-migration-guide-for-existing-nautobot-app-installations) to learn how to migrate your apps and models from `Site` and `Region` to `Location`.
```

glennmatthews (approved):

nautobot/docs/installation/region-and-site-data-migration-guide.md
HanlinMiao: I moved this to before the guide starts

nautobot/docs/installation/region-and-site-data-migration-guide.md
HanlinMiao: After we merge #3313 I think, added that to the TODO comment.

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,438 @@
...
+!!!important  
+    Before you follow the guide, please make sure that these operations are completed:  
+        1. Make sure your working Nautobot is on Version 1.5.x with baseline migrations all run.  
+        2. Stop Nautobot Server.  
+        3. Update installed version of Nautobot to 2.0.0.  
+        4. Run `nautobot-server migrate dcim 0030_migrate_region_and_site_data_to_locations`. (This step will ensure that `("dcim", "0030_migrate_region_and_site_data_to_locations")` is applied to your Nautobot instance and that `("dcim", "0034_remove_region_and_site")` is **not** applied.)  
+    After you complete those operations, follow the guide below for each of your installed apps to:  
+        1. Make all necessary migration files for each app.  
+        2. Run `nautobot-server migrate [app_name]` to apply those migration files to each app.  
+        3. Finally, Start Nautobot Server after **all** the migration files are applied and **all** your affected apps are updated.
```
jathanism, line 174: This is a lot of information stuffed into an admonition.

It would be better as an h3 section with an admonition at the top that is more than just "important". Probably "warning" so that it's orange and stands out. 

It would be good for this admonition to say something like "You must perform these steps." It needs to be very explicit that if you just run the migrations without stopping at `0030` that you absolutely will nuke your data.

Additionally, there should be an extra step after `2. Stop Nautobot Server` that tells the reader to make a database backup. We don't have any documentation on how to backup data. The closest thing we have is a section on "Export data from PostgreSQL" in the "Migrating from PostgreSQL to MySQL" document.

The suggested changes are not complete. Just to give an example.

```suggestion
### Before you Begin 

!!! warning
    You **must** perform these steps before proceeding. Failing to follow them properly could result in data loss. Always backup your database before performing any migration steps.
    
Before you follow the guide, please make sure that these operations are completed:  

1. Make sure your working Nautobot is on Version 1.5.x with baseline migrations all run.  
2. Stop Nautobot Server.  
3. Create a backup of your Nautobot database.
4. Update installed version of Nautobot to 2.0.0.  
5. Run `nautobot-server migrate dcim 0030_migrate_region_and_site_data_to_locations`. (This step will ensure that `("dcim", "0030_migrate_region_and_site_data_to_locations")` is applied to your Nautobot instance and that `("dcim", "0034_remove_region_and_site")` is **not** applied.)  

After you complete those operations, follow the guide below for each of your installed apps to:  

1. Make all necessary migration files for each app.  
2. Run `nautobot-server migrate [app_name]` to apply those migration files to each app.  
3. Finally, Start Nautobot Server after **all** the migration files are applied and **all** your affected apps are updated.  
```

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -0,0 +1,438 @@
...
+    ]
+```
+
+Apply the migration files by running `nautobot-server migrate [app_name]`, for example:
+
+```shell
+nautobot-server migrate example_app
+```
+
+## Region and Site Related Data Model Migration Guide For New Nautobot App installations in an Existing Nautobot 2.0 Environment
```
jathanism, line 438: Is this still a work-in-progress? Why an empty section?

jathanism (changes requested):

nautobot/docs/installation/region-and-site-data-migration-guide.md
HanlinMiao: I need to get an app that has long-running migrations and that uses `Site` and `Region` models in it. So I can try to install it in our alpha environment to know what exactly to do when installing new apps in the 2.0.0 environment.

HanlinMiao: I didn't want to hold up this PR anymore so I thought I could just merge it and add on to it later.

nautobot/docs/installation/region-and-site-data-migration-guide.md
HanlinMiao: Is `nautobot-server dumpdata` something you can use to back up your data?

nautobot/docs/installation/region-and-site-data-migration-guide.md
jathanism: It's fine then, if you wouldn't mind submitting that on the backlog so we don't lose track of it if it isn't already?

nautobot/docs/installation/region-and-site-data-migration-guide.md
```diff
@@ -162,16 +162,20 @@ In Nautobot 2.0.0, all the `Region` and `Site` related data models are being mig
...
-        2. Run `nautobot-server migrate [app_name]` to apply those migration files to each app.  
-        3. Finally, Start Nautobot Server after **all** the migration files are applied and **all** your affected apps are updated.  
+### Before you Begin 
+!!! warning
+    You **must** perform these steps before proceeding. Failing to follow them properly could result in data loss. **Always** backup your database before performing any migration steps.
+    
+Before you follow the guide, please make sure that these operations are completed:  
+1. Make sure your working Nautobot is on Version 1.5.x with baseline migrations all run.  
+2. Stop Nautobot Server.  
+3. Create a backup of your Nautobot database. (Check out the [Export Data Guide](../installation/migrating-from-postgresql.md#export-data-from-postgresql) for this operation)
```
jathanism, line 173: This is great! My only concern here, and I think it should be done in a separate story is this:

The document that this links to is explicitly under the heading “Export data from PostgreSQL”, so I’m thinking what if readers are using MySQL and see this, will they get stuck because it says PG?

We should revise the migration docs to have another section for generic “dump/restore” Nautobot that isn’t necessarily specific to a backend. OR if it is specific to a backend, it has instructions for each (e.g. using `pgdump` vs. `mysqldump` ).

jathanism (approved):

nautobot/docs/installation/region-and-site-data-migration-guide.md
jathanism: Short answer is yes. It is a solution that works across any database backend.

nautobot/docs/installation/region-and-site-data-migration-guide.md
HanlinMiao: https://github.com/nautobot/nautobot/issues/3387

nautobot/docs/installation/region-and-site-data-migration-guide.md
HanlinMiao: https://github.com/nautobot/nautobot/issues/3386

Issue #3242, opened 2023-02-06
UI 2.0: Authentication (session, cookies, etc) to be able to hit both the API and UI


bryanculver:
This should be possible using cookies and session-based authentication. We may need to implement some additional API endpoints to check for session validity, but perhaps not. 

We already have DRF configured for `rest_framework.authentication.SessionAuthentication`, so this shouldn't be too bad.

Some references:

- https://testdriven.io/blog/django-spa-auth/
- https://www.django-rest-framework.org/api-guide/authentication/#sessionauthentication

gsnider2195:
May be able to use an implementation similar to [this](https://blog.logrocket.com/adding-login-authentication-secure-react-apps/)

bryanculver:
Closed via #3356

PR #3356, opened 2023-02-23
Logged in state for nautobot-ui

timizuoebideri1:
# Closes: #3242 
# What's Changed

I simply added an endpoint to retrieve logged in user detail because most of the session base authentication has already been implemented for production (serving react built/complied file). 
Because the menu bar design only shows log In or Log Out button without showing any user detail, this endpoint to retrieve logged in user detail is not currently being used to its full potential, but in the future there will be a username place there instead of just a logout icon.

# TODO

- [ ] Unit, Integration Tests
- [ ] Authenticate using session in react running on node js


bryanculver (changes requested):
@timizuoebideri1 in part of #3242 we should be providing a way for the user to log in with the React UI as well. Maybe that's just redirecting to a Django view with a form but I feel that path would be harder to maintain UI-design-wise. We will also have to consider the SSO sign-in flow as well.

gsnider2195:
Login seems to work for me. We need to create a logout route and see if using the `fetch` api will work for setting the session cookie on login/logout instead of asking the browser to post directly to Django

gsnider2195:
> Login seems to work for me. We need to create a logout route and see if using the `fetch` api will work for setting the session cookie on login/logout instead of asking the browser to post directly to Django

Just to clarify we don't need to retrieve any data from the login view from django, only the return code. I'd like to see if we can have the react UI POST the form data to /login and see if the browser intercepts the cookie and sets it

nautobot/core/api/views.py
```diff
@@ -11,6 +11,8 @@
 from django.db import transaction
 from django.db.models import ProtectedError
 from django.shortcuts import redirect
+from django.utils.decorators import method_decorator
+from django.views.decorators.csrf import ensure_csrf_cookie
```
timizuoebideri1, line 15:
```suggestion
```

nautobot/ui/src/components/common/BSNavBar.jsx
```diff
@@ -1,14 +1,26 @@
...
+import axios from "axios";
 
 
-const fetcher = (url) => fetch(url, { credentials: "include" }).then((res) => res.json());
+const fetcher = (url) => fetch(url, { credentials: "include" }).then((res) => {
+  if(res.status !== 200){
+    throw new Error(res.json());
+  }
+  return res.json();
+});
```
timizuoebideri1, line 15:
```suggestion
const fetcher = (url) => fetch(url, { credentials: "include" }).then((res) => res.json());
```

nautobot/ui/src/components/common/BSNavBar.jsx
```diff
@@ -1,14 +1,21 @@
...
+import axios from "axios";
 
 
 const fetcher = (url) => fetch(url, { credentials: "include" }).then((res) => res.json());
 
 export default function BSNavBar() {
   const { data, error } = useSWR("/api/get-menu/", fetcher)
+  const [ isLoggedIn, setIsLoggedIn] = useState(false)
+  const logout = () => {
+    axios.get("/api/users/tokens/logout/")
```
timizuoebideri1, line 16: not using `useSWR` cause this is an operation that would only be done once, and I don't really need the benefits of  `useSWR` here

gsnider2195: Do we need to use three different methods for fetch? The only reason I used useSWR is because it's recommended by CRA but if Axios is better we could align on that

nautobot/users/api/serializers.py
```diff
@@ -113,3 +114,18 @@ class Meta:
             "actions",
             "constraints",
         )
+
+
+class UserLoginSerializer(serializers.Serializer):
+    username = serializers.CharField(max_length=250)
+    password = serializers.CharField(max_length=250)
```
HanlinMiao, line 121: Nit: Is the `max_length=250` a bit wasteful here?

nautobot/ui/src/components/common/BSNavBar.jsx
```diff
@@ -1,17 +1,33 @@
...
 const fetcher = (url) => fetch(url, { credentials: "include" }).then((res) => res.json());
 
 export default function BSNavBar() {
+  const navigate = useNavigate();
+
   const { data, error } = useSWR("/api/get-menu/", fetcher)
+  const [ isLoggedIn, setIsLoggedIn] = useState(false)
+  useEffect(() => {
+    setIsLoggedIn(localStorage.getItem("nautobot-user") != null)
+  }, [])
```
HanlinMiao, line 19: Can you add some comments for this block of code pretty pls? Just wondering what it does.

nautobot/ui/src/components/common/BSNavBar.jsx
```diff
@@ -1,17 +1,34 @@
...
 
 const fetcher = (url) => fetch(url, { credentials: "include" }).then((res) => res.json());
 
 export default function BSNavBar() {
+  const navigate = useNavigate();
+
   const { data, error } = useSWR("/api/get-menu/", fetcher)
+  const [ isLoggedIn, setIsLoggedIn] = useState(false)
+  useEffect(() => {
+    // Check if `nautobot-user` exist in localStorage; if found set setIsLoggedIn to true else false
```
HanlinMiao, line 18: 🚀

HanlinMiao (approved):
LGTM!

nautobot/users/api/views.py
```diff
@@ -63,6 +69,19 @@ def authentication_classes(self):
...
+        serializer.is_valid(raise_exception=True)
+        user = serializer.validated_data["user"]
+        login(request, user=user)
+        return Response(status=200)
+
+    @action(methods=["GET"], detail=False, permission_classes=[AllowAny])
+    def logout(self, request):
+        logout(request)
+        return Response(status=200)
+
```
bryanculver, line 84: Is there a reason we can't just `POST` to the existing login views? We'll also need to handle the SSO route as well.

timizuoebideri1: That would need a csrf token to be placed in the form, just guessed that's not "API" way of doing things,
But illd try it out

timizuoebideri1: I just attempted it, and I ran into this issue: The status codes 302[OK] and 200[Failure] indicate whether a request was successful or not, but there is no method to manage 302 in "axios."

bryanculver: Confirmed this locally... eww...

bryanculver: Last feedback item: Is there a reason we need to set `permission_classes=[AllowAny]`?

timizuoebideri1: For logout, that is a mistake, for login yes

bryanculver: Is this because it's nested under the tokenviewset?

timizuoebideri1: Yes

bryanculver: OK, can you toss in a `TODO` to potentially move elsewhere? We either create a user login viewset or fix the weird error codes in the default path.

bryanculver (approved):

PR #3357, opened 2023-02-23
Batch update next dependencies within scope, Feb 2023

bryanculver:
Renovate config still needs to be tweaked because #3158 is still too broad. But I wanted to get ahead of some patch updates in `next`.


glennmatthews:
Should we do a bunch of these in `develop` as well?

Issue #826, opened 2021-08-17
Using OrganizationalModel or PrimaryModel shouldn't require also specifying @extras_features(custom_fields, relationships)


glennmatthews:
### Proposed Changes

We should investigate the feasibility and desirability of removing the use of and need for `@extras_features("custom_fields", "relationships")` on models derived from `OrganizationalModel` or `PrimaryModel` (or more precisely, all models using the `CustomFieldModel` and/or `RelationshipModel` mixins).

- Models inheriting from `CustomFieldModel` should automatically support all custom-field-related functionality without requiring an additional decorator
- Models inheriting from `RelationshipModel` should automatically support all relationship-related functionality without requiring an additional decorator

This won't forbid the use of `@extra_features` for custom fields and relationships, but just not depend on them to advertise support for these features. Instead inspect inheritance or properties to determine functionality

### Justification

Code simplification, removing "gotchas".

bryanculver:
Related: #1505

bryanculver:
Closed via #3360

PR #3360, opened 2023-02-24
Alternate approach to updating registry without adding @extra_features to Model

timizuoebideri1:
# Closes: #826 
# What's Changed

Rather than requiring Models to have the decorator @extras_feature added in order to be added to the registry model_features, this prototype concentrates on updating registry without having to add the @extras_feature to model cause it would be smart enough to tell which model needs to be added to a particular feature.

This can be archived by adding some relevant data to lookup_confs
```jsonc
{
    "model": CustomField,
    "lookup_by": "field", 
    "field_names": ["_custom_field_data"],
}
```
1. `model`: model feature to be included in the registry model_features
2. `lookup_by`: This describes how to find the model that satisfies the criteria; options include field or related model
      1. `field`: During the process of iterating through all models, a particular model field name would be looked up, and if found, this is one of the models that would be returned.
      2. `related_model`: While iterating through all models and their fields, we look for instances where a field has a connection to the related_model that has been specified.
3. `field_names`:  The field_names are the field names that will be searched for across all models IF lookup_by is field, and models that have all fields specified will be chosen.
4. `related_model`: If lookup_type is related_model, we would search for this related model in each model field, and if a model field has a relationship with it or has its related_model set to the specified, it would be chosen.
5. `extras`: Just in case we needed additional verifications to focus the search.

I made it my goal to update the registry because I thought that running this code at startup and storing the results there would be the best course of action. Additionally, this logic could be expanded to take into account almost all of the items on extra_features.

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/utils.py
```diff
@@ -207,6 +208,92 @@ def wrapper(model_class):
...
+                extras=lookup_conf.get("extras"),
+            )
+        else:
+            result = lookup_by_related_model(app_models=app_models, related_model=lookup_conf["related_model"])
+        registry_items = {key: list(value) for key, value in result.items()}
+        feature_name = lookup_conf["model"]._meta.model_name
+        registry["model_features"][feature_name] = registry_items
+
+
+def lookup_by_field(app_models, field_names, extras=None):
```
glennmatthews, line 249: Maybe `field_attributes` instead of `extras`?

nautobot/extras/utils.py
```diff
@@ -207,6 +208,92 @@ def wrapper(model_class):
...
+    for lookup_conf in lookup_confs:
+        if lookup_conf["lookup_by"] == "field":
+            result = lookup_by_field(
+                app_models=app_models,
+                field_names=lookup_conf["field_names"],
+                extras=lookup_conf.get("extras"),
+            )
+        else:
+            result = lookup_by_related_model(app_models=app_models, related_model=lookup_conf["related_model"])
+        registry_items = {key: list(value) for key, value in result.items()}
```
glennmatthews, line 244: Can `lookup_by_field` and `lookup_by_related_model` just return the expected data structure directly, so that we don't have to transform `result` to `registry_items`?

nautobot/extras/utils.py
```diff
@@ -207,6 +208,92 @@ def wrapper(model_class):
...
+        feature_name = lookup_conf["model"]._meta.model_name
```
glennmatthews, line 245: It might be better to have the `feature_name` be an explicitly declared value in `lookup_conf`? As won't this currently result in `"customfield"` and `"relationship"` rather than `"custom_fields"` and `"relationships"`?

nautobot/extras/utils.py
```diff
@@ -207,6 +208,92 @@ def wrapper(model_class):
...
 
+def update_registry_with_new_apps():
+    """
+    Updates the registry with new apps.
+
+    This function updates the registry model features for models.
+    """
+    from nautobot.extras.models import CustomField, Relationship, RelationshipAssociation, Role
+
+    lookup_confs = [
```
glennmatthews, line 219: Please add a comment/docstring explaining this data structure in detail. :-)

nautobot/extras/utils.py
```diff
@@ -207,6 +208,92 @@ def wrapper(model_class):
...
+        registry["model_features"][feature_name] = registry_items
+
+
+def lookup_by_field(app_models, field_names, extras=None):
+    """
+    Find all models that have fields with the specified names, and return them grouped by app.
+
+    Args:
+        app_models: A list of model classes to search through.
+        field_names: A list of names of fields to look for.
```
glennmatthews, line 255: If I understand this right, this will add the model to the result if *any* of these field names are valid? Should it instead only do so if *all* of the field names are valid?

nautobot/extras/utils.py
```diff
@@ -207,6 +208,92 @@ def wrapper(model_class):
...
+    """
+    apps_and_models = {}
+    for model_class in app_models:
+        app_label, model_name = model_class._meta.label_lower.split(".")
+        for field_name in field_names:
+            try:
+                field = model_class._meta.get_field(field_name)
+                if extras is None or (
+                    extras and all([getattr(field, item, None) == value for item, value in extras.items()])
+                ):
```
glennmatthews, line 269: Perhaps:

```suggestion
    if extras is None:
        extras = {}
    apps_and_models = {}
    for model_class in app_models:
        app_label, model_name = model_class._meta.label_lower.split(".")
        for field_name in field_names:
            try:
                field = model_class._meta.get_field(field_name)
                if all([getattr(field, item, None) == value for item, value in extras.items()]):
```

nautobot/extras/utils.py
```diff
@@ -207,6 +208,92 @@ def wrapper(model_class):
...
+            "lookup_by": "field",
+            "field_names": ["_custom_field_data"],
+        },
+        {
+            "model": Relationship,
+            "lookup_by": "field",
+            "field_names": ["source_for_associations", "destination_for_associations"],
+            "extras": {"related_model": RelationshipAssociation},
+        },
+        {"model": Role, "lookup_by": "related_model", "related_model": Role},
```
glennmatthews, line 231: Wouldn't this just be equivalent to:

```suggestion
        {
            "model": Role,
            "lookup_by": "field",
            "field_names": ["role"],
            "extras": {"related_model": Role},
        },
```

removing the need for a separate `lookup_by_related_model` function?

glennmatthews (commented):
Cool idea!

nautobot/extras/apps.py
```diff
@@ -9,7 +9,6 @@
 from nautobot.core.apps import NautobotConfig
 from nautobot.core.signals import nautobot_database_ready
 
-
 logger = logging.getLogger("nautobot.extras.apps")
```
timizuoebideri1, line 12:
```suggestion

logger = logging.getLogger("nautobot.extras.apps")
```

nautobot/extras/tests/test_registry.py
```diff
@@ -48,3 +51,22 @@ class DuplicateSecretsProvider(EnvironmentVariableSecretsProvider):
 
         with self.assertRaises(KeyError):
             register_secrets_provider(DuplicateSecretsProvider)
+
+    def test_lookup_by_field(self):
+        """Assert lookup_by_field returns the expected values"""
+
+        # Test for model features with field_attributes
```
timizuoebideri1, line 58:
```suggestion
```

timizuoebideri1:
This PR/Issue does not include refactoring / cleaning up all models with `@extras_features`; however, a new issue could be opened to address this. 
1. Add `roles`, `statuses`, `tags`, and `changelogs` to the '`update registry with new apps`' and '`lookup confs`. 
2. Remove '`RoleModelsQuery`,' '`TaggableClassesQuery`,' and '`ChangeLoggedModelsQuery`,' as all of these classes' functionality can now be obtained from '`FeatureQuery`.' 
3. Remove "custom fields," "relationships," and "statuses" from all '`@extras_features`'.

changes/3360.added
```diff
@@ -0,0 +1 @@
+Added an alternate approach to updating model feature registry without having to decorate a model with @extra_features.
```
glennmatthews, line 1:
```suggestion
Added an alternate approach to updating model feature registry without having to decorate a model with `@extras_features`.
```

nautobot/extras/utils.py
```diff
@@ -207,6 +208,74 @@ def wrapper(model_class):
     return wrapper
 
 
+def update_registry_with_new_apps():
```
glennmatthews, line 211: Perhaps `populate_model_features_registry` or similar might be a clearer name?

nautobot/extras/utils.py
```diff
@@ -207,6 +208,74 @@ def wrapper(model_class):
...
+        registry_items = lookup_by_field(
+            app_models=app_models,
+            field_names=lookup_conf["field_names"],
+            field_attributes=lookup_conf.get("field_attributes"),
+        )
+        feature_name = lookup_conf["feature_name"]
+        registry["model_features"][feature_name] = registry_items
+
+
+def lookup_by_field(app_models, field_names, field_attributes=None):
```
glennmatthews, line 252: Perhaps something like `find_models_with_matching_fields` or similar might be clearer?

nautobot/extras/utils.py
glennmatthews: Still an open question, thanks! :-)

nautobot/extras/utils.py
```diff
@@ -207,6 +208,74 @@ def wrapper(model_class):
...
+    for model_class in app_models:
+        app_label, model_name = model_class._meta.label_lower.split(".")
+        for field_name in field_names:
+            try:
+                field = model_class._meta.get_field(field_name)
+                if all((getattr(field, item, None) == value for item, value in field_attributes.items())):
+                    registry_items.setdefault(app_label, set()).add(model_name)
+            except FieldDoesNotExist:
+                pass
+    registry_items = {key: list(value) for key, value in registry_items.items()}
```
glennmatthews, line 275: Possibly, just for more predictable outcomes:

```suggestion
    registry_items = {key: sorted(value) for key, value in registry_items.items()}
```

glennmatthews (changes requested):
Still like this a lot. Needs documentation, model updates to remove the `custom_fields` and `relationships` extras_features, and verification that everything still works afterwards.

nautobot/extras/utils.py
timizuoebideri1: Oooh Yes, if all field_name are present on the model

timizuoebideri1:
> Still like this a lot. Needs documentation, model updates to remove the `custom_fields` and `relationships` extras_features, and verification that everything still works afterwards.

I manually checked that it functions and produces the same results as `@extras_feature`, but I had intended to refactor in a separate issue (my comment above). Do you believe I should just remove only `relationship` and `custom_field` from all `@extras_fields` in this PR and not create a new issue to address all?

glennmatthews:
> > Still like this a lot. Needs documentation, model updates to remove the `custom_fields` and `relationships` extras_features, and verification that everything still works afterwards.
> 
> I manually checked that it functions and produces the same results as `@extras_feature`, but I had intended to refactor in a separate issue (my comment above). Do you believe I should just remove only `relationship` and `custom_field` from all `@extras_fields` in this PR and not create a new issue to address all?

IMHO, removing just `relationships` and `custom_fields` from `@extras_features` usage is desirable for this PR, with a new issue to address the remaining features.

timizuoebideri1:
> Needs documentation

Any ideas on where this documentation should be added.

nautobot/extras/utils.py
```diff
@@ -207,6 +219,74 @@ def wrapper(model_class):
...
+    lookup_confs = [
+        {
+            "feature_name": "custom_fields",
+            "field_names": ["_custom_field_data"],
+        },
+        {
+            "feature_name": "relationships",
+            "field_names": ["source_for_associations", "destination_for_associations"],
+            "field_attributes": {"related_model": RelationshipAssociation},
+        },
```
jathanism, line 249: I think it would be good to document in the core developer guide how we might go to implement a new "default" feature in the future that doesn't require use of `@extras_features`. 

The comments you provided above helps explain the intent of how one might provide `lookup_confs`, but it's not clear to me why for example `field_attributes` is optional and why it is required for relationships but not custom fields.

I think what is missing is to understand the process that you took to arrive at this elegant solution, and how we can maintain or add more models/features to it in the future.

jathanism (changes requested):

nautobot/docs/development/model-features.md
```diff
@@ -0,0 +1,40 @@
...
+            "field_names": ["bar"],
+            "field_attributes": {"choices": FooBarChoices}
+        },
+       ...
+    ]
+```
+With this only Models which has a field name of `bar` and its field `bar` has the attribute choices=FooBarChoices would be a valid model for the feature `"foo"`
+
+!!! note
+   Only add a feature to lookup_confs if it does not require the use of `@extras_features`.
```
jathanism, line 40:
```suggestion
   Only add a feature to `lookup_confs` if it does not require the use of `@extras_features`.
```

nautobot/docs/development/model-features.md
```diff
@@ -0,0 +1,40 @@
...
+
+A dictionary of particular features (e.g. custom fields) mapped to the Nautobot models which support them, arranged by app.
+
+## Add a new feature to the `model_features` registry
+
+The `populate_model_features_registry()` function updates the registry model features with new apps. This is done by defining a list of dictionaries called `lookup_confs`. Each dictionary in `lookup_confs` contains the following three keys:
+
+- `feature_name`: The name of the feature to be updated in the registry.
+- `field_names`: A list of names of fields that must be present in order for the model to be considered a valid `model_feature`.
+- `field_attributes`: An optional dictionary of attributes to filter the fields by. Only model which fields match all the attributes specified in the dictionary will be considered. This parameter can be useful to narrow down the search for fields that match certain criteria. For example, if `field_attributes` is set to `{"related_model": RelationshipAssociation}`, only fields attr with `related_model` of `RelationshipAssociation` will be considered.
```
jathanism, line 11:
```suggestion
- `field_attributes`: An optional dictionary of attributes used to filter the fields. Only model which fields match all the attributes specified in the dictionary will be considered. This parameter can be useful to narrow down the search for fields that match certain criteria. For example, if `field_attributes` is set to `{"related_model": RelationshipAssociation}`, only fields attr with `related_model` of `RelationshipAssociation` will be considered.
```

jathanism (changes requested):
Also obviously `markdownlint` was not happy.

nautobot/docs/development/model-features.md
```diff
@@ -0,0 +1,40 @@
+# Registry Model Features
```
glennmatthews, line 1: I'm not opposed to splitting this out into a separate document, but note that there was some existing documentation of the `model_features` registry in `nautobot/docs/development/application-registry.md` that can now probably be removed and replaced with a link to this file?

nautobot/docs/development/model-features.md
```diff
@@ -0,0 +1,40 @@
...
+
+A dictionary of particular features (e.g. custom fields) mapped to the Nautobot models which support them, arranged by app.
```
glennmatthews, line 3: Seems like it'd be logical to document `@extras_features` in this same file, unless we're planning to remove it entirely before 2.0.0 ships?

nautobot/docs/development/model-features.md
```diff
@@ -0,0 +1,40 @@
...
+
+The `populate_model_features_registry()` function updates the registry model features with new apps. This is done by defining a list of dictionaries called `lookup_confs`. Each dictionary in `lookup_confs` contains the following three keys:
+
+- `feature_name`: The name of the feature to be updated in the registry.
+- `field_names`: A list of names of fields that must be present in order for the model to be considered a valid `model_feature`.
+- `field_attributes`: An optional dictionary of attributes to filter the fields by. Only model which fields match all the attributes specified in the dictionary will be considered. This parameter can be useful to narrow down the search for fields that match certain criteria. For example, if `field_attributes` is set to `{"related_model": RelationshipAssociation}`, only fields attr with `related_model` of `RelationshipAssociation` will be considered.
+
+To add a new feature to the `lookup_confs` list, follow these steps:
+
+1. Determine the name of the feature to be added.
```
glennmatthews, line 15: Should this be a `slug` rather than a `name`? I don't think there's anything *preventing* you from using a string with whitespace, capital letters, etc., but by convention we're using `snake_case` for all existing features.

nautobot/docs/development/model-features.md
```diff
@@ -0,0 +1,40 @@
...
+        ...,
+        {
+            "feature_name": "foo",
+            "field_names": ["bar"],
+            "field_attributes": {"choices": FooBarChoices}
+        },
+       ...
+    ]
+```
+With this only Models which has a field name of `bar` and its field `bar` has the attribute choices=FooBarChoices would be a valid model for the feature `"foo"`
```
glennmatthews, line 37:
```suggestion
With this only Models which have a field name of `bar`, which in turn has the attribute `choices=FooBarChoices`, would be a valid model for the feature `"foo"`
```

nautobot/docs/plugins/development.md
```diff
@@ -785,7 +785,7 @@ For more advanced usage, you may want to instead inherit from one of Nautobot's
 | [Tags](../models/extras/tag.md) | ❌ | ❌ | ❌ | ✅ |
 
 !!! note
-    When using `OrganizationalModel` or `PrimaryModel`, you also must use the `@extras_features` decorator to specify support for (at a minimum) the `"custom_fields"` and `"relationships"` features.
+    When using `OrganizationalModel` or `PrimaryModel`, you also can use the `@extras_features` decorator to specify support for features e.g `"webhooks""`, `"graphql"`.
```
glennmatthews, line 788: I'd vote for just removing this note, it's no longer critical. Also the edit here is misleading as you can use `@extras_features` on `BaseModel` subclasses too...

nautobot/extras/utils.py
```diff
@@ -207,6 +219,78 @@ def wrapper(model_class):
...
+        registry_items = find_models_with_matching_fields(
+            app_models=app_models,
+            field_names=lookup_conf["field_names"],
+            field_attributes=lookup_conf.get("field_attributes"),
+        )
+        feature_name = lookup_conf["feature_name"]
+        registry["model_features"][feature_name] = registry_items
+
+
+def find_models_with_matching_fields(app_models, field_names, field_attributes=None):
```
glennmatthews, line 267: This probably belongs in `nautobot/core/models/utils.py` alongside the `get_all_concrete_models` function being added by #3368.

More generally, I wonder whether we need to align the approach between these two PRs. This PR is using duck-typing to identify relevant models, while #3368 is mostly using class inheritance to identify its relevant models. Would it make sense to pick one or the other and use it the same for both features?

nautobot/extras/utils.py
```diff
@@ -129,6 +130,17 @@ def get_query(self):
...
+        # The `populate_model_features_registry` function is called in the `FeatureQuery().get_query` method instead of
+        # `ExtrasConfig.ready` because `FeatureQuery().get_query` is called before `ExtrasConfig.ready`.
+        # This is because `FeatureQuery` is a helper class used in `Forms` and `Serializers` that are called during the
+        # initialization of the application, before `ExtrasConfig.ready` is called.
+        # Calling `populate_model_features_registry` in `ExtrasConfig.ready` would lead to an outdated `model_features`
+        # `registry` record being used by `FeatureQuery`.
+        # As the `populate_model_features_registry` can be resource-intensive. The check the conditional check is used to
+        # avoid calling the function multiple times and optimize the performance of the application.
+        if not registry["model_features"].get("relationships"):
+            populate_model_features_registry()
```
glennmatthews, line 143: I gotta say I don't like this. Should we instead be deferring the evaluation of `FeatureQuery()` in some fashion? Why wasn't this an issue before this PR?

glennmatthews (changes requested):

nautobot/docs/development/model-features.md
timizuoebideri1: Sure it can be done

nautobot/docs/development/model-features.md
timizuoebideri1, line 15: Maybe?
```suggestion
1. Determine the name of the feature to be added, This name should be in `snake_case` as per convention
```

nautobot/extras/utils.py
timizuoebideri1: This was not an issue because the models `@extras_features` are evaluated first before serializers and forms `FeatureQuery().get_query()`.

nautobot/docs/development/model-features.md
timizuoebideri1: I don't think there is a plan on removing `@extra_features`? @nautobot/core.

Im not also sure there is any detailed documentation for `@extra_features` currently existing in our docs

nautobot/extras/utils.py
timizuoebideri1: I would support this approach because it takes into consideration models that do not derive from a `ModelMixin`. As far as I know, not all plugin developers would add `status`, `role`, and other features to their models using inheritance, unless we plan to enforce inheritance.

nautobot/extras/utils.py
timizuoebideri1: Any ideas on how I can archive delay the evaluation of `FeatureQuery()`?

nautobot/docs/development/model-features.md
```diff
@@ -0,0 +1,40 @@
...
+        ...,
+        {
+            "feature_name": "foo",
+            "field_names": ["bar"],
+            "field_attributes": {"choices": FooBarChoices}
+        },
+       ...
+    ]
+```
+With this only Models which have a field name of `bar`, which in turn has the attribute `choices=FooBarChoices`, would be a valid model for the feature `"foo"`
```
HanlinMiao, line 37: same comment here

HanlinMiao (commented):
Just some nits, looking solid based on the discussion in standup

jathanism (approved):

bryanculver (approved):
I think this approach is good. It has a smooth path to adoption but I would suspect we would prefer inheritance checks over duck-typing but we don't have a formal style preference on that.

This pattern also seems to allow for `@extra_features` to hang around for now if a model would like to declare support explicitly still.

nautobot/docs/development/model-features.md
glennmatthews: That was my point :) We don't have a good document for that but if it's going to stick around we *should* document it, and this would be the logical file to do so in. Not blocking for this PR though.

nautobot/docs/development/model-features.md
```diff
@@ -0,0 +1,44 @@
...
+            "field_attributes": {"related_model": RelationshipAssociation},
+        },
+       ...
+    ]
+```
+
+With this only Models which have fields names of `source_for_associations` and `destination_for_associations`, which in turn has the attribute `related_model=RelationshipAssociation`, would be a valid model for the feature `relationships`.
+
+!!! note
+   Only add a feature to `lookup_confs` if it does not require the use of `@extras_features`.
```
glennmatthews, line 44: Without documenting how/where to use `extras_features`, this note is a bit obtuse in isolation. I'd think a note like the following might be better?

```suggestion
!!! note
   `populate_model_features_registry()` and `lookup_confs` provide an alternative to the older method of feature flagging models via the `@extras_features` decorator. In general new feature flags should preferentially be implemented via additions to `lookup_confs`, *not* by any new additions to `extras_features`.
```

nautobot/extras/utils.py
glennmatthews: Half-baked idea that came to me just now - what if instead of having a separate `populate_model_features_registry()` that we need to call directly after all model classes are loaded, what if we added logic to `BaseModel` (probably in something like `__init_subclass__()` ??) to where the model automatically adds itself to the appropriate model_features in the registry when it's declared/loaded?

glennmatthews (approved):

nautobot/extras/utils.py
```diff
@@ -129,6 +129,17 @@ def get_query(self):
...
+
+        # The `populate_model_features_registry` function is called in the `FeatureQuery().get_query` method instead of
+        # `ExtrasConfig.ready` because `FeatureQuery().get_query` is called before `ExtrasConfig.ready`.
+        # This is because `FeatureQuery` is a helper class used in `Forms` and `Serializers` that are called during the
+        # initialization of the application, before `ExtrasConfig.ready` is called.
+        # Calling `populate_model_features_registry` in `ExtrasConfig.ready` would lead to an outdated `model_features`
+        # `registry` record being used by `FeatureQuery`.
+        # As the `populate_model_features_registry` can be resource-intensive. The check the conditional check is used to
+        # avoid calling the function multiple times and optimize the performance of the application.
+        if not registry["model_features"].get("relationships"):
```
timizuoebideri1, line 141:
```suggestion
        # TODO(timizuo): Provide a better solution for this; check https://github.com/nautobot/nautobot/pull/3360/files#r1131373797 comment.
        if not registry["model_features"].get("relationships"):
```

Issue #3033, opened 2022-12-20
Job Buttons - Single Click Execution


jvanderaa:
### As ...

Ozzie - Operator

### I want ...

I am looking to be able to make a custom link that will make a POST request, an action button.

### So that ...

This would allow for Nautobot to source a custom activity, such as launch a job on another system.

### I know this is done when...

- There is the ability to do web hook like activity from a button from a custom link.

### Optional - Feature groups this request pertains to.

- [X] Automation
- [X] Circuits
- [X] DCIM
- [X] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

Unsure

### External Dependencies

None

itdependsnetworks:
Action == a job hook essentially?

jeffkala:
POST request without the requirement of "CRUD" operation on the object. A custom link on a Nautobot object that can (webhook) to an external system. The key to this requirement is that clicking the link does this versus a change in state on the Nautobot object itself.

jvanderaa:
Correct, a POST and I think if we are looking at that need to be able to support parameters to be included in the POST URL as well. I think that the Jinja would take care of that natively.

itdependsnetworks:
Love the idea! 

One other consideration is making this a part of template_content, may be able to be a bit more expressive/descriptive there.

glennmatthews:
Based on discussion this morning it sounds like what's really being asked for is the ability to provide a button to trigger either a WebHook or a JobHook (with all of the functionality and configurability that either of those features offers).

bryanculver:
What is the event data for the button though?

JobHooks and WebHooks are firing off of an event (created/edited [what changed]/deleted). Should the payload of this just be a snapshot of the object at present?

jvanderaa:
There are two possibilities here. 

1) An action button similar to custom link, that would just be a link addition with a POST request instead of a GET request. @bryanculver is that something done via a browser then? I think it is.
2) Creating this as a webhook, where there is then significant payload additions.

bryanculver:
>This would allow for Nautobot to source a custom activity, such as launch a job on another system.

A job on another system likely has it's own payload requirements but with WebHooks and JobHooks the sender defines the structure, not the receiver.

[...]

Issue #3141, opened 2023-01-23
Standardize Custom Links and Computed Fields


itdependsnetworks:
### Proposed Changes

Standardize the context provided with both Custom Links and Computed Fields, such as  user, perms, request.

### Justification

Allowing more standard experience

PR #3362, opened 2023-02-24
[WIP]: Job Button model and features

joewesch:
# Closes: #3033
# What's Changed

This PR adds a Job Button model and associated features. I modeled it mostly off Custom Links in terms of UI placement as well as content types to be used with.

Example screenshots:

### Buttons:
Buttons

### Confirmation Modal (if enabled on button):
Confirmation Modal

### Info Message w/ link to Job Result:
Info Message

# TODO

- [ ] Unit, Integration Tests


itdependsnetworks:
Perhaps we should consider providing user, perms, and requests that are in other similar fields? See #3141

joewesch:
> Perhaps we should consider providing user, perms, and requests that are in other similar fields? See #3141

I believe I already included them (copied from Custom Links):
https://github.com/nautobot/nautobot/pull/3362/files#diff-666d45365d47ee0a33a96a8e0279b8b9baac487b3f9ee8dd9ebb05add5e1d090R84-R86

Or are you speaking of adding them elsewhere?

bryanculver:
Do we have to use Extra Features to implement this? That means app-provided models won't get this for free, which they should.

joewesch:
> Do we have to use Extra Features to implement this? That means app-provided models won't get this for free, which they should.

No, I was simply mimicking Custom Links.

nautobot/extras/urls.py
```diff
@@ -482,6 +486,20 @@
...
+        views.ObjectChangeLogView.as_view(),
+        name="jobbutton_changelog",
+        kwargs={"model": JobButton},
+    ),
+    path(
+        "job-button/<uuid:pk>/notes/",
+        views.ObjectNotesView.as_view(),
+        name="jobbutton_notes",
+        kwargs={"model": JobButton},
+    ),
```
bryanculver, line 501: These should be included with a `NautobotUIViewSet`, `NautobotUIViewSetRouter` now: https://github.com/nautobot/nautobot/pull/2752, https://docs.nautobot.com/projects/core/en/stable/plugins/development/#nautobotuiviewsetrouter

nautobot/extras/views.py
```diff
@@ -1567,6 +1570,51 @@ def get(self, request, pk=None):
...
+            func=run_job,
+            name=job_model.class_path,
+            obj_type=get_job_content_type(),
+            user=request.user,
+            data={
+                "obj": post_data["obj"],
+            },
+            request=copy_safe_request(request),
+            commit=True,
+        )
```
bryanculver, line 1612: Can we make the signature for this more like `JobHookReceiver`? The current implementation appears it will allow a user to associate a Job that doesn't take the object on the `obj` key.

https://github.com/nautobot/nautobot/blob/develop/examples/example_plugin/example_plugin/jobs.py#L60-L81

Also, I'm assuming if I read this correctly the only jobs candidates for Job Buttons do not allow other arguments to be entered into them?

joewesch: > Can we make the signature for this more like JobHookReceiver? The current implementation appears it will allow a user to associate a Job that doesn't take the object on the obj key.

I'll look at it and see the difference would be.

> Also, I'm assuming if I read this correctly the only jobs candidates for Job Buttons do not allow other arguments to be entered into them?

Yes, this is [based on the discussion](https://github.com/nautobot/nautobot/issues/3033#issuecomment-1380968777) on the linked issue:

> The job would have to have no input.

itdependsnetworks:
Nope, you have it, thanks!!

joewesch:
> Do we have to use Extra Features to implement this? That means app-provided models won't get this for free, which they should.

@bryanculver I agree that it would be nice if app-provided models could have access to it. Since this is primarily a GUI element, is there any concern of allowing a Job Button to have a FK relationship to a Content Type that doesn't implement the `object_retrieve.html` template?

glennmatthews:
Yeah given that we're trying to reduce reliance on `extras_features` (see #3360) I'd definitely prefer that we not add a new entry to all models for this feature. Maybe instead use the same feature as custom-links, if anything is needed at all?

joewesch:
I have removed the use of `extras_features` as suggested.

joewesch:
@bryanculver I have refactored to use the JobButtonReceiver class as we discussed. I will continue on to documentation and testing. Let me know if there are any other changes you would like to see or discuss.

nautobot/docs/models/extras/jobbutton.md
```diff
@@ -0,0 +1,166 @@
...
+    job_class.log_info(obj=obj, message="Running Site Job.")
+    # Add Site job logic here
+
+
+def _run_device_job(job_class, obj):
+    job_class.log_info(obj=obj, message="Running Device Job.")
+    # Add Device job logic here
+```
+
+## Context Data
```
itdependsnetworks, line 134: I am not seeing the context data in any of the examples above? To me that would have broken the signature, so I am not sure I am fully tracking.

nautobot/docs/models/extras/jobbutton.md
```diff
@@ -0,0 +1,166 @@
...
+* **Name** - A unique name for the Job Button.
+* **Object type(s)** - The type or types of Nautobot object that the button will be associated to.
+* **Text** - The text that will be displayed on the button.
+* **Job** - The [Job Button Receiver](#job-button-receivers) that this button will run.
+* **Weight** - The number used for determining the order the buttons will appear.
+* **Group** - The name of the dropdown group to add this button into (optional).
+* **Button Class** - The button CSS class, which dictates the color.
+* **Confirmation** - Should the button pop up a confirmation dialog before running.
+
+!!! warning
```
itdependsnetworks, line 20: I would think we should default safe for both commit and confirmation meaning

confirmation = True
commit = False

examples/example_plugin/example_plugin/jobs.py
```diff
@@ -85,4 +86,93 @@ def validate_serial(self, serial):
...
+
+
+def _run_site_job(job_class, obj):
+    job_class.log_info(obj=obj, message="Running Site Job.")
+    # Add Site job logic here
+
+
+def _run_device_job(job_class, obj):
+    job_class.log_info(obj=obj, message="Running Device Job.")
+    # Add Device job logic here
```
bryanculver, line 165: This _might_ be better served as a blog post/user guide on "How to combine a `Job` and `JobButtonReciver`" and might be an advanced use case.

nautobot/docs/models/extras/jobbutton.md
bryanculver: I _think_ @joewesch this is so you could do something like `JobButton(text="Check {obj.name}")`, right?

nautobot/docs/models/extras/jobbutton.md
itdependsnetworks: after speaking with @joewesch I think the only actual request is to update the `help_text` of confirmation to be more clear on the implication of turning that off.

nautobot/docs/models/extras/jobbutton.md
itdependsnetworks: Spoke with @joewesch, I had it backwards, the context of user and similar is already in `self`, Joe is kind enough to add an example of that, but no coding change requested once I understood.

examples/example_plugin/example_plugin/jobs.py
joewesch: Fair.

nautobot/extras/choices.py
```diff
@@ -71,11 +71,11 @@ class CustomFieldTypeChoices(ChoiceSet):
 
 
 #
-# CustomLinks
+# Button Classes
 #
 
 
-class CustomLinkButtonClassChoices(ChoiceSet):
+class ButtonClassChoices(ChoiceSet):
```
bryanculver, line 78: Since this is going into `develop` let's not assume no one is using this. Please use the `@class_deprecated_in_favor_of` decorator to seamlessly rename this class.

(suggesstion may not pass Black, Flake8)
```suggestion
@class_deprecated_in_favor_of(ButtonClassChoices)
class CustomLinkButtonClassChoices(ButtonClassChoices):
    pass

class ButtonClassChoices(ChoiceSet):
```

nautobot/extras/utils.py
```diff
@@ -360,6 +363,7 @@ def refresh_job_model_from_job_class(job_model_class, job_source, job_class, *,
             "grouping": job_class.grouping[:JOB_MAX_GROUPING_LENGTH],
             "name": job_class.name[:JOB_MAX_NAME_LENGTH],
             "is_job_hook_receiver": issubclass(job_class, JobHookReceiver),
+            "is_job_button_receiver": issubclass(job_class, JobButtonReceiver),
```
bryanculver, line 366: Since we now have multiple subclasses for jobs, can you throw in a check for `isinstance(job_class, JobHookReceiver) and isinstance(job_class, JobButtonReceiver)` and reject the job?

bryanculver (changes requested):
99% there @joewesch, this is awesome. A few changes requests above, as well as:

- [x] Can we get a few screenshots? Namely the Job Button form and a button's presentation in the UI on the object.
- [x] RE: https://github.com/nautobot/nautobot/pull/3362/files#r1131915019
    - I think having the Context Data and conditional formatting sections around the Job Button section itself would be helpful since they are attributes of the Job Button, not the Job Button Receiver. Hopefully, it'll read better.
- [ ] https://github.com/nautobot/nautobot/pull/3362/files#r1132477306
- [ ] Remove the advanced job pattern example from the example job (see: https://github.com/nautobot/nautobot/pull/3362/files#r1132449092)
- [x] ~Tests!~
    - 🤦 Dunce of the week. I marked them as viewed and when I went back to summarize I missed them. Sorry!

nautobot/extras/utils.py
joewesch: Added in latest commits.

nautobot/extras/choices.py
joewesch: Added in latest commits.

nautobot/docs/models/extras/jobbutton.md
joewesch: Added in latest commits.

nautobot/extras/models/jobs.py
```diff
@@ -859,7 +859,10 @@ class JobButton(BaseModel, ChangeLoggedModel, NotesMixin):
         choices=ButtonClassChoices,
         default=ButtonClassChoices.CLASS_DEFAULT,
     )
-    confirmation = models.BooleanField(help_text="Enable confirmation pop-up box", default=True)
+    confirmation = models.BooleanField(
+        help_text="Enable confirmation pop-up box. <font color='#ff0000'>WARNING: unselecting this option will allow the Job to run (and commit changes) with a single click!</font>",
```
bryanculver, line 863: Can you do
```suggestion
        help_text="Enable confirmation pop-up box. <span class='text-danger'>WARNING: unselecting this option will allow the Job to run (and commit changes) with a single click!</span>",
```
instead?

bryanculver (approved):

nautobot/extras/tests/test_filters.py
```diff
@@ -924,6 +926,55 @@ def test_search(self):
...
+        params = {"name": ["JobButton1"]}
+        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 1)
+
+    def test_job(self):
+        params = {"job": [Job.objects.get(job_class_name="TestJobButtonReceiverSimple").pk]}
+        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 2)
+
+    def test_weight(self):
+        params = {"weight": [50]}
+        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 1)
```
gsnider2195, line 971: Can you update these to test for multiple values in the filter? Also, we've been trying to move away from `count` assertions in filter tests and instead use `assertQuerysetEqualAndNotEmpty`. Example for `test_name`:

```py
    def test_name(self):
        params = {"name": ["JobButton1", "JobButton2"]}
        self.assertQuerysetEqualAndNotEmpty(self.filterset(params, self.queryset).qs, self.queryset.filter(name__in=["JobButton1", "JobButton2"]))
```

nautobot/docs/models/extras/jobbutton.md
```diff
@@ -0,0 +1,133 @@
...
+
+Job Buttons can be created in web UI located in the navbar under Jobs > Job Buttons. Each button can be associated with multiple Nautobot object types (site, device, prefix, etc.) and will be displayed on all of the associated object detail views. The text displayed on the button supports Jinja2 templating which allows for using [context data](#context-data) to dynamically update or [even be hidden under certain conditions](#conditional-rendering).
+
+The buttons appear at the top right corner of an object's individual detail page for each object type they are associated to. They can be either individual buttons or grouped together in a dropdown for better organization. In either case, the order of the buttons are influenced by the `weight` number configured for each button.
```
gsnider2195, line 7: Please update if this is incorrect. I didn't see any docs on how the `weight` works.

```suggestion
The buttons appear at the top right corner of an object's individual detail page for each object type they are associated to. They can be either individual buttons or grouped together in a dropdown for better organization. Buttons will be sorted from left to right based on their `weight` with the lowest `weight` button on the left. Any groups of buttons will be displayed to the right of all ungrouped buttons. Group dropdown buttons will inherit the button class from the button with the lowest `weight` in the group and will be sorted from top to bottom with the lowest `weight` button on top.
```

nautobot/docs/models/extras/jobbutton.md
```diff
@@ -0,0 +1,133 @@
...
+* **Object type(s)** - The type or types of Nautobot object that the button will be associated to.
+* **Text** - The text that will be displayed on the button.
+* **Job** - The [Job Button Receiver](#job-button-receivers) that this button will run.
+* **Weight** - The number used for determining the order the buttons will appear.
+* **Group** - The name of the dropdown group to add this button into (optional).
+* **Button Class** - The button CSS class, which dictates the color.
+* **Confirmation** - Should the button pop up a confirmation dialog before running.
+
+!!! warning
+    As you can see, there is no `commit` option for a Job Button like there is for a normal Job. All Job Buttons will run `commit=True` **implicitly**.
```
gsnider2195, line 21:
```suggestion
    As you can see, there is no `commit` option for a Job Button like there is for a normal Job. All Job Buttons will run with `commit=True` **implicitly**.
```

nautobot/extras/models/jobs.py
```diff
@@ -807,6 +819,61 @@ def log(
...
+        help_text="Buttons with the same group will appear as a dropdown menu. Group dropdown buttons will inherit the button class from the button with the lowest weight in the group.",
+    )
+    button_class = models.CharField(
+        max_length=30,
+        choices=ButtonClassChoices,
+        default=ButtonClassChoices.CLASS_DEFAULT,
+    )
+    confirmation = models.BooleanField(
+        help_text="Enable confirmation pop-up box. <span class='text-danger'>WARNING: unselecting this option will allow the Job to run (and commit changes) with a single click!</span>",
+        default=True,
```
gsnider2195, line 864: Can the default for this field be set to `False` since setting it to `True` is probably not the desired default setting considering how dangerous it can be?

nautobot/docs/models/extras/jobbutton.md
```diff
@@ -0,0 +1,133 @@
+# Job Buttons
+
+Job Buttons are predefined buttons that allow users to run jobs directly from within Nautobot object views. It uses the object where the button was pressed as the only input to the job. These are helpful when you want to start a job that requires minimal or no input without having to use the standard job form. For example, you may have a job that only requires a user to select a device. Instead, they can now go to that device in the web UI and click the associated Job Button instead.
```
gsnider2195, line 3:
```suggestion
+++ 1.5.14

Job Buttons are predefined buttons that allow users to run jobs directly from within Nautobot object views. It uses the object where the button was pressed as the only input to the job. These are helpful when you want to start a job that requires minimal or no input without having to use the standard job form. For example, you may have a job that only requires a user to select a device. Instead, they can now go to that device in the web UI and click the associated Job Button instead.
```

nautobot/docs/models/extras/jobbutton.md
```diff
@@ -0,0 +1,133 @@
...
+
+| Variable | Description |
+|----------|-------------|
+| `obj`      | The Nautobot object being displayed |
+| `debug`    | A boolean indicating whether debugging is enabled |
+| `request`  | The current WSGI request |
+| `user`     | The current user (if authenticated) |
+| `perms`    | The [permissions](https://docs.djangoproject.com/en/stable/topics/auth/default/#permissions) assigned to the user |
+
+All [built-in Jinja2 filters](../../additional-features/template-filters.md) are available and it's also possible to [develop and register a custom Jinja2 filters](../../plugins/development.md#including-jinja2-filters).
```
gsnider2195, line 42:
```suggestion
All [built-in Jinja2 filters](../../additional-features/template-filters.md) are available and it's also possible to [develop and register custom Jinja2 filters](../../plugins/development.md#including-jinja2-filters).
```

nautobot/extras/models/jobs.py
joewesch: I believe you may be confusing `confirmation` with `commit`. The default here says all new Job Buttons will default to having a confirmation dialog pop-up modal rather than run via a single click.

There is no setting to enable or disable `commit` for the Job - it is always enabled.

nautobot/extras/tests/test_filters.py
joewesch: Changed in latest commits

nautobot/docs/models/extras/jobbutton.md
```diff
@@ -0,0 +1,133 @@
...
+{% if perms.extras.run_job %}Provision{% endif %}
+```
+
+The button will only appear if they have the permission to run jobs.
+
+![Job Buttons on Site object](../../media/models/site_jobbuttons.png "Job Buttons on Site object")
+
+## Job Button Receivers
+
+Job Buttons are only able to initiate a specific type of job called a **Job Button Receiver**. These are jobs that subclass the `nautobot.extras.jobs.JobButtonReceiver` class. Job Button Receivers are similar to normal jobs except they are hard coded to accept only `object_pk` and `object_model_name` [variables](../../additional-features/jobs.md#variables). Job Button Receivers are hidden from the jobs listing UI by default but otherwise function similarly to other jobs. The `JobButtonReceiver` class only implements one method called `receive_job_button`.
```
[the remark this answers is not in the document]
joewesch: Added in latest commits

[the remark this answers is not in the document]
joewesch: Added via note in latest commits

nautobot/extras/tests/test_filters.py
```diff
@@ -924,6 +926,64 @@ def test_search(self):
...
+        params = {"job": [Job.objects.get(job_class_name="TestJobButtonReceiverComplex").pk]}
+        self.assertQuerysetEqualAndNotEmpty(
+            self.filterset(params, self.queryset).qs, self.queryset.filter(name__in=["JobButton3"])
+        )
+
+    def test_weight(self):
+        params = {"weight": [50]}
+        self.assertQuerysetEqualAndNotEmpty(
+            self.filterset(params, self.queryset).qs, self.queryset.filter(name__in=["JobButton3"])
+        )
```
gsnider2195, line 977:
```suggestion
                confirmation=True,
                weight=30,
            ),
            JobButton.objects.create(
                name="JobButton2",
                text="JobButton2",
                job=Job.objects.get(job_class_name="TestJobButtonReceiverSimple"),
                confirmation=False,
                weight=40,
            ),
            JobButton.objects.create(
                name="JobButton3",
                text="JobButton3",
                job=Job.objects.get(job_class_name="TestJobButtonReceiverComplex"),
                confirmation=True,
                weight=50,
            ),
        )

        site_ct = ContentType.objects.get_for_model(Site)
        for jb in job_buttons:
            jb.content_types.set([site_ct])

    def test_name(self):
        params = {"name": ["JobButton1", "JobButton2"]}
        self.assertQuerysetEqualAndNotEmpty(
            self.filterset(params, self.queryset).qs, self.queryset.filter(name__in=["JobButton1", "JobButton2"])
        )

    def test_job(self):
        job = Job.objects.get(job_class_name="TestJobButtonReceiverSimple")
        params = {"job": [job.pk]}
        self.assertQuerysetEqualAndNotEmpty(
            self.filterset(params, self.queryset).qs, self.queryset.filter(job__pk=job.pk)
        )

        params = {"job": [job.slug]}
        self.assertQuerysetEqualAndNotEmpty(
            self.filterset(params, self.queryset).qs, self.queryset.filter(job__slug=job.slug)
        )

    def test_weight(self):
        params = {"weight": [30, 50]}
        self.assertQuerysetEqualAndNotEmpty(
            self.filterset(params, self.queryset).qs, self.queryset.filter(weight__in=[30, 50])
        )
```

gsnider2195: Updated `job` and `weight` tests so the right side of the assertion will match even if the test data changes. Also made sure they're testing multiple values

PR #3367, opened 2023-02-24
UI 2.0 remove nginx container

gsnider2195:
# Closes: #DNE
# What's Changed

- Replace nginx reverse proxy with [CRA proxy](https://create-react-app.dev/docs/proxying-api-requests-in-development/) feature
- Replace `node:latest` docker image with `node:slim` (reduced filesize from 1GB to 250MB)

# TODO

- [ ] Unit, Integration Tests


development/docker-compose.dev.yml
```diff
@@ -32,17 +31,9 @@ services:
     user: node
     volumes:
       - ../nautobot/ui/:/opt/node/nautobot
+      - ../examples/:/opt/examples
```
glennmatthews, line 34: Curious why this is needed? Can we not pull the example plugin in from `/source`?

nautobot/ui/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "nautobot",
   "version": "2.0.0-alpha.1",
   "private": true,
+  "proxy": "http://nautobot:8080",
```
glennmatthews, line 5: This feels like a strange place for this, though I'm definitely no expert here. What's the impact of this addition?

gsnider2195: This is what tells node to proxy any unknown routes to the backend nautobot container (so we can remove nginx)

development/docker-compose.dev.yml
gsnider2195: We're using relative paths in our `jsconfig.json` so our path to the example plugin ui looks like `../../examples/ui/`. This is to support the same relative path in the node container.

bryanculver (approved):

glennmatthews (approved):

Issue #1422, opened 2022-02-25
Openapi schema inconsistencies


UnityDG:
### Environment
* Python version: 3.9
* Nautobot version: 1.2.4

Found a few mismatches between the defined schema and the returned data when generating bindings for go with openapi-generator-cli

1.
Sample URL: /api/dcim/interfaces/XXX/
Required data: Untaagged vlans applied to interface
Sample Definition: VmInterface + Interface:
Field: tagged_vlans
Schema type: string[] of uuids
Actually returned: NestedVLAN[]
```
  "tagged_vlans": [
    {
      "id": "89913e46-82b4-5ced-9b54-a459b8777b3a",
      "url": "http://localhost:8080/api/ipam/vlans/XX/",
      "vid": 10,
      "name": "XX",
      "display": "XX"
    },
    {
      "id": "ec495efc-1f11-586e-88b9-427ed1659717",
      "url": "http://localhost:8080/api/ipam/vlans/XX/",
      "vid": 20,
      "name": "XX",
      "display": "XX"
    }
  ],
```

2.
Sample URL: /api/dcim/interfaces/XXX/
Required data: Connection between two interfaces
Sample Definition: Anywhere cable_peer / connected_endpoint is used
Field: cable_peer + connected_endpoint
Schema type: nullable string
Actually returned: NestedInterface?
```
  "cable_peer": {
    "id": "4ae220eb-9161-526e-8ecf-9a338b3c0d0c",
    "url": "http://localhost:8080/api/dcim/interfaces/4ae220eb-9161-526e-8ecf-9a338b3c0d0c/",
    "device": {
      "id": "a1a6fc59-8edb-52fd-8177-f08abc23bf7c",
      "url": "http://localhost:8080/api/dcim/devices/a1a6fc59-8edb-52fd-8177-f08abc23bf7c/",
      "name": "X",
      "display": "X"
    },
    "name": "swp51",
    "cable": "61d5e1e0-66a4-55b7-8209-001bbab4df1d",
    "display": "swp51"
  },
  "cable_peer_type": "dcim.interface",
  "connected_endpoint": {
    "id": "4ae220eb-9161-526e-8ecf-9a338b3c0d0c",
    "url": "http://localhost:8080/api/dcim/interfaces/4ae220eb-9161-526e-8ecf-9a338b3c0d0c/",
    "device": {
      "id": "a1a6fc59-8edb-52fd-8177-f08abc23bf7c",
[...]

PR #3262, opened 2023-02-08
Fix extras FKs and related names

timizuoebideri1:
# Closes: #3067 
# What's Changed

- Fix all related names

# TODO

- [ ] Explanation of Change(s)
- [ ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [ ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example Plugin Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #3368, opened 2023-02-24
Add polymorphic types for serializer GenericForeignKeys

glennmatthews:
# Closes: #1422
# What's Changed

This all builds atop drf-spectacular's `PolymorphicProxySerializer` class, but some details of the changes I made to make it fit our use case:

- Add an `object_type` field to all nested serializers, which simply is a string representation of the content-type of this serializer (e.g. `"dcim.device"`)
    - When using `PolymorphicProxySerializer` there needs to be some field present on each serializer that uniquely identifies which serializer it is. We didn't have that, so I added it.
    - Some of our nested serializers weren't actually inheriting from `WritableNestedSerializer` (our standard base class) so I had to change them over.
- Subclass `PolymorphicProxySerializer` to add a couple of features missing in the base class.
   - [ ] I'd like to consider offering these upstream to drf-spectacular, as they seem generally applicable.
- For each `SerializerMethodField` that we have that maps to one of several distinct nested serializers (due to use of GenericForeignKey), update the schema annotation (`@extend_schema_field`) using our custom `PolymorphicProxySerializer` to identify the possible serializers that may be seen.
   - I added a couple of helpful methods `get_all_concrete_subclasses` and `get_serializers_for_models` to make it easier to dynamically identify the possible serializers for each given field.

Before:

image

After:

image

# TODO

- n/a Unit, Integration Tests
   - [ ] We should consider whether to backport this to 1.5 LTM. It won't patch cleanly due to #3132, but this *should* be low risk to include in there as a bug fix.


nautobot/extras/api/nested_serializers.py
```diff
@@ -150,7 +158,7 @@ class Meta:
         fields = ["id", "url", "name"]
 
 
-class NestedJobLogEntrySerializer(BaseModelSerializer):
+class NestedJobLogEntrySerializer(WritableNestedSerializer):
```
bryanculver, line 161: Should we allow writing/editing Nested Job Log Entries?

nautobot/extras/api/nested_serializers.py
```diff
@@ -167,7 +175,7 @@ class Meta:
         ]
 
 
-class NestedJobResultSerializer(BaseModelSerializer):
+class NestedJobResultSerializer(WritableNestedSerializer):
```
bryanculver, line 178: Should we allow writing/editing Nested Job Result Entries?

nautobot/extras/api/nested_serializers.py
```diff
@@ -134,7 +142,7 @@ class Meta:
         fields = ["id", "url", "name", "image"]
 
 
-class NestedJobSerializer(BaseModelSerializer):
+class NestedJobSerializer(WritableNestedSerializer):
```
bryanculver, line 145: Do Scheduled Job and Job Result need their nested Job Model writable?

nautobot/extras/api/nested_serializers.py
glennmatthews: `WritableNestedSerializer` is not well-named - just using it doesn't automatically make a serializer or a field writable, it just provides common infra (including the new `object_type` parameter I'm adding in this PR) for serializers that *are* declared as writable. By writable it just means that *if* you're updating a foreign object reference on the base serializer, you can identify the new foreign object either by its PK or by any unique set of fields.

The only place we actually use `NestedJobLogEntrySerializer` at present is on the `api/extras/job-results/<uuid>/logs/` endpoint, which is read-only. 

Our own docs specify that nested serializers should *always* use `WritableNestedSerializer` so this is making things align better with our own docs.

glennmatthews: Looking back at this, this change is not *strictly* necessary since `JobLogEntry` is not a full-featured model so it doesn't *have* to have an `object_type` field since it can't be used with Notes or ObjectChanges. But I think it's still a good change.

nautobot/extras/api/nested_serializers.py
glennmatthews: See above comment. The only places we use NestedJobResultSerializer at present are read-only. As with JobLogEntry, `JobResult` doesn't support Notes or ObjectChanges at present so this change ends up not being *strictly* necessary but I think it's still desirable from a consistency standpoint.

nautobot/extras/api/nested_serializers.py
glennmatthews: `Job` *does* support Notes and ObjectChanges, so we need an  `object_type` on NestedJobSerializer.

JobResultSerializer and ScheduledJobSerializer both specify that their associated NestedJobSerializer is `read_only=True`.

nautobot/extras/api/nested_serializers.py
```diff
@@ -222,7 +230,19 @@ class Meta:
...
+    url = serializers.HyperlinkedIdentityField(view_name="extras-api:scheduledjob-detail")
+    name = serializers.CharField(max_length=255, required=False)
+    start_time = serializers.DateTimeField(format=None, required=False)
+
+    class Meta:
+        model = models.ScheduledJob
+        fields = ["url", "name", "start_time", "interval", "crontab"]
+
+
+class NestedScheduledJobCreationSerializer(BaseModelSerializer):
```
glennmatthews, line 243: Just calling attention to this addition in my last commit. When we specify schedule information on a JobInputSerializer, this is *not* a standard use of a nested serializer, as it's defining a new schedule for the job, *not* referencing an existing ScheduledJob instance. `WritableNestedSerializer` expects that any parameters you provide it are meant to be used to locate an existing instance and will throw a ValidationError if it doesn't find one, so for the JobInputSerializer case I had to create this separate class which is *not* a WritableNestedSerializer.

jathanism (changes requested):

nautobot/core/api/fields.py
```diff
@@ -105,6 +105,25 @@ def to_representation(self, obj):
...
+        """Default read_only to True as this should never be a writable field."""
+        super().__init__(*args, read_only=read_only, **kwargs)
+
+    def to_representation(self, _value):
+        """
+        Get the content-type of this serializer's model.
+
+        Implemented this way because `_value` may be None when generating the schema.
+        """
+        model = self.parent.Meta.model
```
HanlinMiao, line 123: Just asking this for my own knowledge, what is `self.parent` here?

nautobot/core/api/serializers.py
```diff
@@ -186,8 +187,22 @@ class WritableNestedSerializer(BaseModelSerializer):
...
+    which of several possible nested serializers are in use for a given attribute.
     """
 
+    object_type = ObjectTypeField()
+
+    def get_field_names(self, declared_fields, info):
+        """Ensure that the "object_type" field is always included in self.fields."""
+        fields = list(super().get_field_names(declared_fields, info))
+        if "object_type" not in fields:
+            fields.append("object_type")
```
HanlinMiao, line 203: Saved one line hehe 
```suggestion
        fields = set(super().get_field_names(declared_fields, info))
        fields.add("object_type")
```

nautobot/dcim/api/serializers.py
```diff
@@ -139,7 +143,16 @@ def get_cable_peer_type(self, obj):
...
         return None
 
-    @extend_schema_field(serializers.DictField(allow_null=True))
+    @extend_schema_field(
+        PolymorphicProxySerializer(
+            component_name="cable_peer",
+            resource_type_field_name="object_type",
+            serializers=lambda: get_serializers_for_models(
+                get_all_concrete_subclasses(CableTermination), prefix="Nested"
+            ),
```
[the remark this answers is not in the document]
glennmatthews: I was thinking that `get_all_concrete_subclasses` should be useful in other future cases as well, such as perhaps the follow-on work after #3360.

nautobot/extras/api/nested_serializers.py
```diff
@@ -222,7 +230,19 @@ class Meta:
...
+    """Unlike WritableNestedSerializer, this allows for creation of a ScheduledJob, not just finding an existing one."""
```
glennmatthews, line 244:
```suggestion
    """
    Nested serializer specifically for use with `JobInputSerializer.schedule`.

    We don't use `WritableNestedSerializer` here because this is not used to look up
    an existing `ScheduledJob`, but instead used to specify parameters for creating one.
    """
```

nautobot/extras/api/nested_serializers.py
glennmatthews: Will do.

HanlinMiao (approved):
Do we need any tests for this feature?

nautobot/extras/api/serializers.py
```diff
@@ -918,7 +949,7 @@ def validate(self, data):
 class JobInputSerializer(serializers.Serializer):
     data = serializers.JSONField(required=False, default=dict)
     commit = serializers.BooleanField(required=False, default=None)
-    scheduled_job = NestedScheduledJobSerializer(required=False)
+    schedule = NestedScheduledJobCreationSerializer(required=False)
```
glennmatthews, line 952: Here and in `extras/api/views.py` and in `extras/tests/test_api.py` I'm reverting this serializer field from `scheduled_job` (change made in #3262) back to `schedule`; as discussed, this isn't actually a reference to a ScheduledJob instance, it's a schedule for the job that may or may not result in the creation of a ScheduledJob instance, so I think `schedule` is more accurate.

nautobot/ipam/api/serializers.py
```diff
@@ -349,7 +352,17 @@ class Meta:
...
 
-    @extend_schema_field(serializers.DictField(allow_null=True))
+    @extend_schema_field(
+        PolymorphicProxySerializer(
+            component_name="assigned_object",
+            resource_type_field_name="object_type",
+            serializers=[
+                NestedInterfaceSerializer,
+                NestedVMInterfaceSerializer,
+            ],
```
glennmatthews, line 362: Could arguably use `get_all_concrete_subclasses(BaseInterface)` here instead.

nautobot/dcim/api/serializers.py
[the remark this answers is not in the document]
glennmatthews: Also there are cases where I'm using `get_serializers_for_models` separarately, such as:

```python
            serializers=lambda: get_serializers_for_models(
                FeatureQuery("config_context_owners").list_subclasses(), prefix="Nested"
            ),
```

Or were you suggesting that I keep both of these as separate functions but add a third wrapper function that combines the two?

glennmatthews:
> Do we need any tests for this feature?

We don't have any current tests for the generated schema (other than running `nautobot-server check-schema` in the CI to make sure that it at least generates without errors). I suppose it'd be possible to write a set of tests that would render the schema to YAML then parse the YAML and look for specific data structures. Mind if I add that as a backlog item?

HanlinMiao:
> > Do we need any tests for this feature?
> 
> We don't have any current tests for the generated schema (other than running `nautobot-server check-schema` in the CI to make sure that it at least generates without errors). I suppose it'd be possible to write a set of tests that would render the schema to YAML then parse the YAML and look for specific data structures. Mind if I add that as a backlog item?

Yeah, I think the test is out of scope for this PR at least.

glennmatthews:
On hold until I can verify that this works correctly with `oapi-codegen`.

glennmatthews:
> > > Do we need any tests for this feature?
> > 
> > 
> > We don't have any current tests for the generated schema (other than running `nautobot-server check-schema` in the CI to make sure that it at least generates without errors). I suppose it'd be possible to write a set of tests that would render the schema to YAML then parse the YAML and look for specific data structures. Mind if I add that as a backlog item?
> 
> Yeah, I think the test is out of scope for this PR at least.

I added a basic test framework and a couple of sample tests in f08809d.

nautobot/dcim/api/serializers.py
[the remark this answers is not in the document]
jathanism: Oh yeah, I was thinking to leave the other two helpers in place, but just having this other one for the common case to wrap them if the decomposition isn't needed. Sorry fit hat wasn't clear.

A third helper to wrap the two helpers for the common case.

nautobot/dcim/tests/test_schema.py
```diff
@@ -0,0 +1,84 @@
+from nautobot.core.models.utils import get_all_concrete_models
+from nautobot.core.testing.schema import OpenAPISchemaTestCases
+from nautobot.dcim.models import CableTermination, PathEndpoint
+
+
+class DCIMOpenAPISchemaTestCase(OpenAPISchemaTestCases.BaseSchemaTestCase):
```
jathanism, line 6: I want to be clear in that I think this is awesome.

What I don't like is how there is so much hard-coding of object names and nested fields relationships. I wish there were a way we could abstract this somehow.

If we were, for example, to manually add schema tests across the board, it would be incredibly tedious.

I know you added these as sample tests and it seems reasonable that this lays the foundation for implementing generic tests based on expectations provided by the way things are marked up using schema decorators?

bryanculver:
What requires this to go to `next` vs `develop`? If I understand the PR correctly it just makes the schema reflect more accurately what you've always had to send to the API but previously was described as a generic JSON dictionary. Is it quasi-breaking in that it will do better serializer enforcement? Or because of:
> Add an `object_type` field to all nested serializers, which simply is a string representation of the content-type of this serializer (e.g. "dcim.device")


will this reject previous calls missing the `object_type` in a request payload?

bryanculver (approved):

bryanculver:
> > What requires this to go to `next` vs `develop`?
> 
> Nothing specific, I just started working on this under `next` initially as I wasn't initially sure of the scope and impact. As I noted in the PR description:
> 
> > _We should consider whether to backport this to 1.5 LTM. It won't patch cleanly due to #3132, but this should be low risk to include in there as a bug fix._
> 
> I'll add that the later-added changes in this PR to the `JobInputSerializer` are specific to `next` and wouldn't apply to `develop` at all. So it's not a clean port over to `develop` for a number of reasons, but it _is_ a bug fix and _could_ be committed back to `develop` for the 1.5 LTM if we think the effort is justified.

Missed the last line in the original description. I think it's fine to leave it to `next` unless there is a significant push to backport it.

nautobot/dcim/tests/test_schema.py
glennmatthews: As discussed, I'll take a quick stab at refining this a bit to reduce the repetition.

glennmatthews: Or spend a couple hours on it, whichever. 😅 Please take a look at the updated tests.

nautobot/core/api/serializers.py
```diff
@@ -249,6 +264,51 @@ def to_internal_value(self, data):
...
+    def serializers(self):
+        """Return the serializers provided for this instance, evaluating them first if needed."""
+        if callable(self._serializers):
+            self._serializers = self._serializers()
+        return self._serializers
+
+    @serializers.setter
+    def serializers(self, value):
+        """Set the serializers for this instance."""
+        self._serializers = value
```
glennmatthews, line 304: I contributed the serializers-as-callable feature upstream as https://github.com/tfranzel/drf-spectacular/pull/953, so once there's a new release we can remove some of this code.

bryanculver: 😎 https://github.com/tfranzel/drf-spectacular/releases/tag/0.26.0

bryanculver: I have removed these methods and updated to `0.26` with 1b10237.

Since `allow_null` is not upstream I figured it would be preferred to be consistent to still use our overloaded class than split hairs between where we used ours vs what could use the default from DRF-Spectacular (example `origin` on `CablePathSerializer`).

glennmatthews: Yup, good stuff. I think the `allow_null` here doesn't quite work across the board, or at least that's what I was seeing when I put a little time last week into trying to port it upstream as well. (IIRC it works for our case where we're using a polymorphic serializer for a specific field on an API endpoint, but it didn't seem to work as expected for the upstream test case that has an entire API endpoint that's polymorphic.) So for the moment keeping our local `allow_null` enhancement here makes sense.

jathanism (approved):

Issue #3337, opened 2023-02-17
Prototype: Network Namespaces, RouteDistinguisher, VRF Models


bryanculver:
Implement the `Namespace`, `RouteDistinguisher`, and `VRF` models (new or changed) from the design described here: https://github.com/nautobot/nautobot/issues/3245#issuecomment-1434806956 and https://github.com/nautobot/nautobot/issues/3245#issuecomment-1433620448.

This is only a prototype so there is no expectation of implementing the data migrations from existing models (current VRF will likely be split up to RD and VRF). A UI-only implementation will be sufficient for getting data entered.

The purpose of this prototype is to take a vanilla install and confirm if the model supports expected use cases. The goal is to ship this in Alpha 2 where we can get real-world validation of these new models.

### To Be Implemented
- [x] `Namespace` model
- [x] `RouteDistinguisher` model
- [x] Updates to `VRF` model
- [x] FK from `Prefix` to `Namespace`
- [x] Optional FK from `Interface` to `VRF`
- [ ] Uniqueness constraints necessary to validate the ultimate goal of #2909 to support #2915 and #1673

### To NOT Be Implemented (we will NOT ship a beta without these)
- Fixing existing test cases
    - Comment them, set them to "expected to fail", leave `TODOs`
- Extensive new test cases
    - We will have minimal-to-missing validations so creating a rich test suite is not necessary at this point
- IPtoInterface M2M
    - (that will come via #3236)
- Extensive data enforcement
    - Ideally based upon the design linked to above an Interface cannot have an IP address from a Prefix in a Namespace it's parent device isn't a part of via VRF->RD
- Overhauls to the natural keys of Prefix, VRF, etc

gsnider2195:
# Use cases

## No namespace, VRF or RD

```mermaid
classDiagram
    class RootPrefix
    RootPrefix : network 10.0.0.0/8
    RootPrefix : namespace null
    class Prefix1
    Prefix1 : network 10.0.0.0/16
    RootPrefix <-- Prefix1
    class Prefix2
    Prefix2 : network 10.128.0.0/16
    RootPrefix <-- Prefix2
    class IPAddress1
    IPAddress1 : address 10.0.0.1
    Prefix1 <-- IPAddress1
    class IPAddress2
    IPAddress2 : address 10.128.0.1
    Prefix2 <-- IPAddress2
    Interface1 --> Device1
    IPAddress1 <-- "primary" Interface1
[...]

Issue #3245, opened 2023-02-06
Spike: Network Namespaces


lampwins:
As the Nautobot PO, I want an understanding of the requirements for namespaces so the team can begin to break down the work into several stories.

AC:
- What are the bounds for a namespace (IPAM only, devices, all of nautobot, etc)?
- What do namespaces do (provide uniqueness, etc)?
- What are the requirements for overlap, nesting, etc?

**Latest Design:** (see: https://github.com/nautobot/nautobot/issues/3245#issuecomment-1434806956)

![Untitled Diagram drawio-2](https://user-images.githubusercontent.com/31187/219696314-335ac810-c833-4e82-b95a-1a539622bbb3.png)

lampwins:
Our stance on tenants is that they are to be used as ownership labels. Namespaces are closer to a non-functional requirement to be able to define bounds for uniqueness constraints.

glennmatthews:
It seems to me that one of the first questions to answer is whether we're trying to design namespaces as an enabler (even if only in the future) of true multitenancy (not to be confused with the current Tenant/TenantGroup model) and resource isolation within Nautobot or whether that's explicitly out of scope. 

Namespaces for multitenancy:
- would need to apply to ALL Nautobot data models (with few/no exceptions)
- would need first-class support in user and permission definitions (i.e. make it really easy to restrict a user to only be able to access objects (of *any* type) that correspond to a specific namespace(s))

Namespaces for resource isolation within a single-tenant Nautobot instance:
- would only need to apply to data models that make sense to support namespacing (perhaps not models such as RIR, DeviceType, Provider, Location...?) though it might prove simpler from an implementation standpoint to make this an all-models feature anyway?
- would not require, but probably would still benefit from, first-class user/permission support, probably?

So maybe it's less of an issue than I was first thinking. Still worth further thought though.

glennmatthews:
Broadly speaking, the purpose of a namespace (as currently scoped) as I see it is to allow for multiple database records to coexist that would otherwise fall afoul of appropriately-defined uniqueness constraints, and conversely to provide a baseline for defining uniqueness constraints that make it feasible to have natural keys for all models and minimize data duplication or ambiguity. The devil is in the details of course! Some thoughts:

- Perhaps host address + namespace should uniquely identify an IPAddress record (presuming requirements such as #2403 are met).
[...]

Issue #2403, opened 2022-09-12
IPAM Overhaul: Assign IP Address to Multiple Interfaces


martin-networker:
```[tasklist]
### TODO
- [x] #3190
- [x] #3236
- [ ] https://github.com/nautobot/nautobot/issues/3411
```

```[tasklist]
### High Priority
- [ ] #3634
- [ ] #3635
- [ ] #3636
- [ ] #3637
- [ ] https://github.com/nautobot/nautobot/issues/4100
- [ ] https://github.com/nautobot/nautobot/issues/4104
- [ ] #3536
```

---
### As ...

Austin - Network Automation Engineer

### I want ...

To be able to assign one IP address to multiple interfaces/devices.

### So that ...

We can assign anycast and other shared IP addresses to all interfaces/devices which has this IP address assigned instead of assigning this IP.

### I know this is done when...

Reason for request:
In our infrastructure we use anycast IP addresses extensively. For example in on DC around two hundred switches share about 40 anycast IP addresses. Another case is for NVE interface switches in VPC cluster has to share one IP so in every VPC cluster two switches share one IP.

Current Nautobot options:
  1) Assign shared IP address to one device:
  	 This will be incorrect information and it will lead to confusions and cause mistakes.
  2) Create a custom field for the interface object that points to the IP address object:
     This will not give us any connection between IP and interfaces. User or script would have to go through all interfaces of all devices to find out where is IP assigned which makes it unusable.
  3) Create a One-To-Many relationship which will connect IP address to multiple interfaces:
     IP address is not seen in Interfaces view (if you open Device and switch to tab interfaces) you must open every interface to find where it is assigned.
	 IP address does not have assignment (Operational Details > Assignment).
	 This means the relationship is not visible from Interfaces view nor from IP address view.

Possible solution:
  Change assignment of IPs from one-to-one to one-to-many. This would be a breaking change and require a database change.
  In GUI it would be great to keep the current functionality and create condition:
    a) if IP is assigned to only one Interface:
	in Operational Details > Assignment it will show the one Device directly (just like it is now)
    b) if IP is assigned to more than one Interface (or maybe it can be based on Role so if IP Role is Anycast/Shared/etc.):
	Operational Details > Assignment will show something like Relationship is showing. Parameter interfaces and value will be link (x interfaces) which would open new window IP Address Associations (similar to Relationship Associations).
  If the condition would cause difficulty the b) option would be sufficient.
[...]

PR #3378, opened 2023-02-28
Introduce IPAM Namespaces

jathanism:
# Closes: #3337, #3429, #3681
# What's Changed

This is the very rough around the edges prototype of IPAM Namespaces. It is not yet complete.

It has drifted somewhat from what was discussed in the design chats and what is scoped on #3337, due to the experimentation in the data design during prototyping.

## The gist

- VRFs can have prefixes assigned to them
- VRFs can have devices assigned to them
- VRFs can be used across devices
- Any number of unique VRFs may be assigned to any number of devices, therefore allowing VRF names to be reused across devices
- An RD is only significant upon assignment of a VRF to a device; therefore the same VRF can be assigned to multiple devices, each with a distinct (including null) RD if so desired 
- Any number of unique Prefixes may be assigned to any number of VRFs, so long as they are in the same namespace

## Changes that drifted related to VRF/RD

- VRF has FK to Namespace instead of deriving namespace through RD
- VRF has m2m to Device, Prefix with through tables for each
- VRF through table for Device is where RD is assigned; validation asserts that RD (if set) namespace matches VRF namespace
- VRF through table for Device; validation asserts that if set, VRF namespace must match Prefix namespace

## Caveats

- This will likely not work in any API or UI views. It's rough.
- This is tested at the ORM only at this time, expecting manual creation of objects to assert functionality.
- There are fields commented out in serializers/views/forms/filters so that database migrations work without complaining 

# TODO

- [ ] Unit, Integration Tests


bryanculver:
@jathanism Can you describe the why we deviated from the original design? As far as I can tell this moves the validation into Python as opposed to schema/DB. Is this long-term desirable? What were the limitation with the original?

> An RD is only significant upon assignment of a VRF to a device; therefore the same VRF can be assigned to multiple devices, each with a distinct (including null) RD if so desired

Is this, could this not be, done behind the scenes from the API interfaces? Where we get schema performant validation but simplified APIs for integrators? Is that the plan and this just a compromise for the prototype?

glennmatthews:
Could you add a class diagram similar to https://github.com/nautobot/nautobot/issues/3337#issuecomment-1448756759 and https://github.com/nautobot/nautobot/issues/3245#issuecomment-1434806956? I'm having a hard time telling from the text description alone how this implementation differs from those and what the implications are.

jathanism:
> Could you add a class diagram similar to [#3337 (comment)](https://github.com/nautobot/nautobot/issues/3337#issuecomment-1448756759) and [#3245 (comment)](https://github.com/nautobot/nautobot/issues/3245#issuecomment-1434806956)? I'm having a hard time telling from the text description alone how this implementation differs from those and what the implications are.



> @jathanism Can you describe the why we deviated from the original design? As far as I can tell this moves the validation into Python as opposed to schema/DB. Is this long-term desirable? What were the limitation with the original?
> 
> > An RD is only significant upon assignment of a VRF to a device; therefore the same VRF can be assigned to multiple devices, each with a distinct (including null) RD if so desired
> 
> Is this, could this not be, done behind the scenes from the API interfaces? Where we get schema performant validation but simplified APIs for integrators? Is that the plan and this just a compromise for the prototype?

Context, descriptions and some Mermaid diagrams have been provided in the issue: https://github.com/nautobot/nautobot/issues/3337#issuecomment-1450855876

jathanism:
> Could you add a class diagram similar to [#3337 (comment)](https://github.com/nautobot/nautobot/issues/3337#issuecomment-1448756759) and [#3245 (comment)](https://github.com/nautobot/nautobot/issues/3245#issuecomment-1434806956)? I'm having a hard time telling from the text description alone how this implementation differs from those and what the implications are.

This is the output from:

```
nautobot-server graph_models -a -g -I Namespace,Prefix,IPAddress,RouteDistinguisher,VRF,VRFDeviceAssignment,VRFPrefixAssignment,Interface,Device` -o nautobot_ipam_namespace.png
```

 The boxes for `Device` and `Interface` were cropped out of this render for conciseness, but you can see the relationship lines pointing to them at the top of the image.

image

HanlinMiao:
https://github.com/nautobot/nautobot/pull/3395
This is the implementation based on the original discussion we had. @jathanism and I collaborated on this.

jathanism:
# Latest changes: 

Got UI/API working with bare minimum effort to keep things flowing.

## Namespace

- Went with default "Global" namespace for foreign keys FOR NOW just to keep things moving
-  Basic API/UI views
image
image


## General

- Had to remove `vrf` from `select_related` for `Prefix`, `IPAddress` in about 7,000 places

## RouteDistinguisher

- Basic API/UI views

image


## VRF

- VRF edit includes namespace/devices, prefixes limited by namespace
- VRF detail includes devices/prefixes
- Filter by namespace now possible

image
image

image


## Prefix

- `Prefix.objects.annotate_tree` now filters on namespace not VRF
- Edit view includes namespace
- List view groups by namepace
- Filter by namespace now possible
- `Prefix.get_duplicates()` changed to a noop (returns `queryset.none()`)
- Available prefixes now filtered by namespace instead of VRF

image

image

image


## IPAddress

- WIP: Not filtering on VRF, but also not updated for Prefix parenting yet

## Generic

- Had to patch `ObjectEditView` to catch `ValidationError` for `PrefixEditView` because of the `unique_together` constraint ValidationError/IntegrityError not being caught the form validation due to `network` and `prefix_length` fields being excluded from validation due to not being included in the form fields.

# Not Done

- Linking RD to VRF when assigned to Devices "somehow"
- Show assigned VRFs in Device detail view
- Assigning a VRF to an Interface
- Showing assigned VRF in Interface detail view

nautobot/dcim/models/device_components.py
```diff
@@ -662,6 +669,11 @@ def to_csv(self):
     def clean(self):
         super().clean()
 
+        # VRF validation
+        if self.vrf and self.vrf not in self.device.vrfs.all():
+            # TODO(jathan): Or maybe we automatically add the VRF to the device?
+            raise ValidationError({"vrf": "VRF must be assigned to same Device."})
+
```
jathanism, line 676: This was a quick hack. UI should only allow selection of VRFs that are already assigned to the device.

nautobot/ipam/filters.py
```diff
@@ -44,6 +46,19 @@
 )
 
 
+class NamespaceFilterSet(NautobotFilterSet):
```
jathanism, line 49: Needs `LocatableModelFilterSetMixin`.

nautobot/ipam/signals.py
```diff
@@ -0,0 +1,44 @@
...
+@receiver(pre_save, sender=VRF.devices.through)
+@receiver(pre_save, sender=VRF.prefixes.through)
+def ipam_object_saved(sender, instance, raw=False, **kwargs):
+    """
+    Forcibly call `full_clean()` when a new IPAM object is manually saved to prevent creation of
+    invalid objects.
+    """
+    if raw:
+        return
+    instance.full_clean()
```
bryanculver, line 19: This is going to create some noise, as it's causing a bunch of tests to blow up everywhere. See: https://github.com/nautobot/nautobot/pull/3378/files#diff-00a7372c809230ca828aa70f90352848b6654faadb78e4762a8f55ee6c59ce4fR584-R585

bryanculver (commented):
Notes:
- I think there was unanimous agreement that there shouldn't be a VRF property on IPAddress so I've updated places that might have caused issues in testing in UI (instead of commenting out), there might be some utility in a helper method to get the parent prefix's vrf
-

nautobot/ipam/models.py
```diff
@@ -48,6 +47,46 @@
...
+    """Return the Global namespace for use in default value for foreign keys."""
+    obj, _ = Namespace.objects.get_or_create(
+        name="Global", defaults={"description": "Default Global namespace. Created by Nautobot."}
+    )
+
+    return obj.pk
+
+
+@extras_features(
+    "custom_fields",
```
bryanculver, line 89: 👀 Silly `git`

bryanculver:
I was reviewing the test changes in the prototype and the changes being made in #2403 and I still have some thoughts around the VRF-Prefix M2M:

- Intuitively we will be adding an optional FK on (VM)Interface to VRF.
- Considering #2403 change, or even the old behavior where an Interface can have many IPs, it would be expected that if I have a VRF set on the Interface, I should only see the IPAddress selections from Prefixes that are a part of the same VRF
    - Hand-wavey: We should eventually as a part of UX make this easier for the user to assign the VRF/Prefixes to each other but call out that we are building it for them (think modal popup "add prefix to VRF" when editing Interface)
- If a Prefix was allowed to be a part of multiple VRFs (via M2M) then is that address assignable because it's through a different VRF or is that not a separate network?
		- `Interface(ETH0)` is a member of `VRF(Orange)`
		- `Interface(ETH1)` is a member of `VRF(Blue)`
		- `Prefix(10.0.0.0/24)` is a member of `VRF(Orange)` and `VRF(Blue)`
		- I want to assign `IPAddress(10.0.0.100)` which has a (concrete relationship to `Prefix(10.0.0.0/24)`)  to `Interface(ETH0)`, this would work but then it's an assignment to `Interface(ETH1)` would be :sus: because 
			- These are effectively the same network (and impose the same VRF) and I would get an address conflict, or at least would have to be a "backup" IP and have additional validation check
			- These are different networks and then I should not be using the same `IPAddress` object and that would require a different `Namespace` and `VRF` entirely


Also, Prefix-VRF was non-M2M before and this never came up before as a limitation, but it was a uniqueness constraint and it would create a unique object because it was implied to be a different network and on the off-chance you did want them to be the same you would have to manage the duplicates manually between these prefixes.

Though there's probably a single sentence that unravels my train of thought here...

nautobot/extras/views.py
```diff
@@ -1828,7 +1828,7 @@ def get_extra_context(self, request, instance):
                 "device_type",
             )
             ipaddress = instance.ip_addresses.select_related(
-                "vrf",
+                # "vrf",
```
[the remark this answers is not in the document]
jathanism: > Notes:
> 
> * I think there was unanimous agreement that there shouldn't be a VRF property on IPAddress so I've updated places that might have caused issues in testing in UI (instead of commenting out), there might be some utility in a helper method to get the parent prefix's vrf

Oh yeah, for sure, I stopped at the places that were causing utter breakage in other views. I didn't actually complete the necessary work for IPAddress.

nautobot/ipam/signals.py
jathanism: That is correct. This is there for now to force model validation in lieu of having formsets for m2m intermediate relationships for the prototype. I don't think it's worth creating the formset code right now if we're just going to rip it out and replace it for the v2 UI.

jathanism: I disabled it for `Prefix` and `VRF`.

jathanism:
> I was reviewing the test changes in the prototype and the changes being made in #2403 and I still have some thoughts around the VRF-Prefix M2M:
> 
> * Intuitively we will be adding an optional FK on (VM)Interface to VRF.
> * Considering [IPAM Overhaul: Assignment IP Address to Multiple Interfaces #2403](https://github.com/nautobot/nautobot/issues/2403) change, or even the old behavior where an Interface can have many IPs, it would be expected that if I have a VRF set on the Interface, I should only see the IPAddress selections from Prefixes that are a part of the same VRF

These are done as of 4019193

>   * Hand-wavey: We should eventually as a part of UX make this easier for the user to assign the VRF/Prefixes to each other but call out that we are building it for them (think modal popup "add prefix to VRF" when editing Interface)
> * If a Prefix was allowed to be a part of multiple VRFs (via M2M) then is that address assignable because it's through a different VRF or is that not a separate network?
>   - `Interface(ETH0)` is a member of `VRF(Orange)`
>   - `Interface(ETH1)` is a member of `VRF(Blue)`
>   - `Prefix(10.0.0.0/24)` is a member of `VRF(Orange)` and `VRF(Blue)`
>   - I want to assign `IPAddress(10.0.0.100)` which has a (concrete relationship to `Prefix(10.0.0.0/24)`)  to `Interface(ETH0)`, this would work but then it's an assignment to `Interface(ETH1)` would be :sus: because
>   - These are effectively the same network (and impose the same VRF) and I would get an address conflict, or at least would have to be a "backup" IP and have additional validation check
>   - These are different networks and then I should not be using the same `IPAddress` object and that would require a different `Namespace` and `VRF` entirely

They are in fact the same networks. And so being able to assign an IP to an interface with `VRF(Orange)` or `VRF(Blue)` shouldn't matter. So currently as of 4019193 this is currently implemented.

> Also, Prefix-VRF was non-M2M before and this never came up before as a limitation, but it was a uniqueness constraint and it would create a unique object because it was implied to be a different network and on the off-chance you did want them to be the same you would have to manage the duplicates manually between these prefixes.

If you do want the Prefixes to be distinct from each other, this would be the use-case for using a `Namespace` to differentiate them.

> Though there's probably a single sentence that unravels my train of thought here...

"Use namespaces."

nautobot/core/tests/test_graphql.py
```diff
@@ -228,6 +228,7 @@ def test_extend_schema_null_field_choices(self):
         self.assertIsInstance(getattr(schema, "resolve_mode"), types.FunctionType)
 
 
+@skip("Something really funky is broken here. Setup fails due to content types not existing...")
```
gsnider2195, line 231: Fix before merging?

nautobot/extras/tests/test_api.py
```diff
@@ -88,6 +90,7 @@ def test_root(self):
 #
 
 
+@skip(reason="Content Types are BROKEN")
```
gsnider2195, line 93: Fix before merging?

gsnider2195: Same question throughout this file

bryanculver (approved):

Issue #3295, opened 2023-02-13
Jobs Issue with Celery Result Extended


mattmiller87:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.9
* Python version: 3.9.7
* Database platform, version: 10.17
* Middleware(s): None

### Steps to Reproduce
1. Enable extended celery results in nautobot_config.py: “CELERY_RESULT_EXTENDED = True”  
2. Run any job and review worker logs if the exception doesn’t appear on the result page

### Expected Behavior
Job Results return

### Observed Behavior
Logs indicate:
```kombu.exceptions.EncodeError: Object of type User is not JSON serializable```

PR #3379, opened 2023-03-01
Fix kombu serialization error on jobs

dsykes16:
# Closes: #3295
# What's Changed

Resolves the `kombu.exceptions.EncodeError` raised when either `CELERY_RESULT_EXTENDED == True` or `enqueue_job` is called from within another `Job` and the existing Job's request is passed to `enqueue_job`.

Fix is implemented by ensuring `_cached_user` is removed from the request is present, which appears to occur when an existing Job's request is passed to `enqueue_job` (i.e. a `Job` that queues up another `Job`) or through some other scenario.

This also appears to fix the linked issue which details a similar bug when `CELERY_RESULT_EXTENDED == True`. The exact code-path leading to that bug was not investigated; however, testing was completed to confirm this resolves the issue.

This fix is a back-port of a fix already implemented in v2.0.0-alpha.1 ([original fix](https://github.com/nautobot/nautobot/commit/1b5d62dd861c0c0a0550d49802eb6ac356c02334#diff-75faa4e2846155ade0f3e2d35048efb5cf59dbee86c9a090c78b42b7350b7442))

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/3295.fixed
+++ b/changes/3295.fixed
@@ -1,0 +1,1 @@
+Fixed kombu serialization error on `User` object that arose when `CELERY_RESULT_EXTENDED == True` or when `enqueue_job` was called from within an existing `Job`.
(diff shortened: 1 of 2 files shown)


jathanism (approved):
Thanks for taking the time to back-port this from the v2 alpha. We had discussed whether or not to do it and didn't, so thanks for the help!

glennmatthews (approved):
Thank you!

Issue #3240, opened 2023-02-06
UI 2.0: How should we serve the UI & API seamlessly?


bryanculver:
- `/api`, `/ui`, etc
- Going to need to keep legacy UI views servable
- Easy upgrading/migrating existing views that can easily adopt new UI

glennmatthews:
Per https://github.com/nautobot/nautobot/pull/3285#discussion_r1103060850, current plan is to group all UI-related API URLs under `/api/ui/`, e.g. `/api/ui/get-menu/`, `/api/ui/filterset-fields/lookup-choices/` etc.

bryanculver:
> Per [#3285 (comment)](https://github.com/nautobot/nautobot/pull/3285#discussion_r1103060850), current plan is to group all UI-related API URLs under `/api/ui/`, e.g. `/api/ui/get-menu/`, `/api/ui/filterset-fields/lookup-choices/` etc.

Can we ensure we don't needlessly spend time publishing/maintaining serializers or swagger docs for our UI? They are not needed and can and should break as they will be tightly coupled with our UI and are not meant for integrating externally.

glennmatthews:
It's absolutely possible to exclude endpoints from the schema, and I agree that we should do so for the API-UI endpoints.

jathanism:
> It's absolutely possible to exclude endpoints from the schema, and I agree that we should do so for the API-UI endpoints.

Further, the UI API endpoints should be considered "private" despite being public, so that we don't have to support them for anything other than what we need until such time we feel it is sufficiently stable to say otherwise. This give us license to say "sorry, please use the primitives, calling the UI API endpoints directly is not supported".

HanlinMiao:
Closed by #3385

PR #3385, opened 2023-03-02
move 'api/' ui related urls under 'api/ui/'

HanlinMiao:
# Closes: #3240 
# What's Changed

1. Created a new `urls.py` file in `nautobot/core/api/ui/` folder and moved current UI-related API endpoints into that file.
2. Changed all occurrences of UI-related `/api/` to `/api/ui/`.
3. Excluded `/api/ui/get-menu/` endpoint from OpenAPI schema
4. Added documentation elaborating on UI-related endpoints support and versioning.

# TODO

- [ ] Unit, Integration Tests


nautobot/core/urls.py
```diff
@@ -29,6 +29,7 @@
     path("virtualization/", include("nautobot.virtualization.urls")),
     # API
     path("api/", include("nautobot.core.api.urls")),
+    path("api/ui/", include("nautobot.core.api.ui.urls")),
```
bryanculver, line 32: Nit: Would this be better added into the URL patterns in `/core/api/urls.py`? UI APIs are a subset of APIs of APIs, which has includes already for plugins and GraphQL.

nautobot/docs/rest-api/ui-related-endpoints.md
```diff
@@ -0,0 +1,8 @@
+# UI Related API Endpoints
+
++++ 2.0.0
+
+Nautobot now retrieves its UI component information from API endpoints.
+
+!!!important
+    Note that these API endpoints are hidden from OpenAPI schema and do not support REST API versioning.
```
jathanism, line 8: Let's also make it very explicit that these endpoints are NOT supported and are for internal use only and therefore are not subject to SemVer and can change at any time to support the UI.

That is of course, until such point that we decide we are ready to support any UI-specific endpoints.

While on this topic: We should be very deliberate in only having UI-specific endpoints as an exception and not the norm. Non-UI REST API endpoints should continue to be utilized wherever possible.

nautobot/ui/urls.py
```diff
@@ -0,0 +1,29 @@
+# UI-related API endpoints
```
jathanism, line 1: I don't think we should put this file here, seeing as `nautobot/ui` is not a Python package nor is it a Django application. Since the views/serializers are defined in `core.api` let's just put these url patterns for the UI endpoints in there as well.

nautobot/ui/urls.py
```diff
@@ -0,0 +1,29 @@
...
+    path(
+        "filterset-fields/lookup-value-dom-element/",
+        GetFilterSetFieldDOMElementAPIView.as_view(),
+        name="filtersetfield-retrieve-lookupvaluedomelement",
+    ),
+]
+ui_api_patterns = [
+    # Lookup Expr
+    path("core/", include((core_api_patterns, "core-api"))),
+    path("get-menu/", GetMenu.as_view(), name="get-menu"),
```
jathanism, line 28: I know it's not in this file, but while you're touching this, can you please rename `GetMenu` to `GetMenuAPIView`?

jathanism (changes requested):

bryanculver (approved):

timizuoebideri1 (approved):

jathanism (approved):

PR #3397, opened 2023-03-07
Develop to next post 1.5.12

bryanculver:
Waiting on #3396 which can be merged into this branch.


nautobot/core/management/commands/generate_test_data.py
```diff
@@ -97,7 +97,7 @@ def _generate_factory_data(self, seed):
         self.stdout.write("Creating LocationTypes...")
         LocationTypeFactory.create_batch(7)  # only 7 unique LocationTypes are hard-coded presently
         self.stdout.write("Creating Locations...")
-        LocationFactory.create_batch(40)  # we need more locations with sites since it can be nested now.
+        LocationFactory.create_batch(65)  # we need more locations with sites since it can be nested now.
```
bryanculver, line 100: Needed more locations that had no parent, no children in the circuit tests.

nautobot/core/filters.py
```diff
@@ -367,7 +367,7 @@ def noop(v):
 
             # Or set it to what was defined by caller
             elif isinstance(lookup_info, dict):
-                lookup_expr = lookup_info["lookup_expr"]
+                lookup_expr = lookup_info.get("lookup_expr")
```
bryanculver, line 370: Pylint

nautobot/core/management/commands/generate_test_data.py
glennmatthews: Can we handle this more elegantly rather than just throwing larger numbers at the test? Should we instead do a separate explicit `LocationFactory.create_batch(..., has_parent=False)` call?

glennmatthews (approved):

nautobot/core/management/commands/generate_test_data.py
bryanculver: > Can we handle this more elegantly rather than just throwing larger numbers at the test? Should we instead do a separate explicit `LocationFactory.create_batch(..., has_parent=False)` call?

I feel like we're walking into a territory where we need to create a bunch of one-off objects for even some generic tests. Probably could revisit this factory strategy.

HanlinMiao (approved):

timizuoebideri1 (approved):

Issue #3236, opened 2023-02-06
Assign IP to Multiple Interfaces: Prototype M2M Change


bryanculver:
Investigate how other aspects need to be changed/improved (potentially create some one-off stories for future follow-on in 2.0)

- Primary IP on Device is a 1-to-1 enforcement, this likely can be derived 
- What additional fields should live on the through table
- Primary IP of interface vs Primary IP of device
    - #1099

This should be the master issue and branch to go into `next`. Other subsequent PRs will target this (as it likely won't be mergable initially, forms and filters will be broken).

Since a spike of this nature will almost certainly be a "draw the rest of the owl" problem, it will build on #3190. It will help size the following story in #2403.

gsnider2195:
## Through table fields

|  field  |  type  |  choices  |  constraints  |
|-------|--------|-----------|---------------|
| `source`[^3] | `BooleanField` | `True`/`False` | |
| `destination`[^3] | `BooleanField` | `True`/`False` | |
| `primary`[^1] | `BooleanField` | `True`/`False` | |
| `secondary`[^2] | `BooleanField` | `True`/`False` | |
| `default`[^1] | `BooleanField` | `True`/`False` | |
| `preferred`[^1] | `BooleanField` | `True`/`False` | |
| `primary_for_device` | `BooleanField` | `True`/`False` | Limited to one per device |



[^1]: [JunOS interface configuration](https://www.juniper.net/documentation/us/en/software/junos/interfaces-fundamentals/topics/topic-map/protocol-family-interface-address-properties.html#id-configuring-default-primary-and-preferred-addresses-and-interfaces)
[^2]: [Cisco IOS configuration](https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/ipaddr/command/ipaddr-cr-book/ipaddr-c1.html#wp1930336422)
[^3]: [Cisco tunnel configuration](https://www.cisco.com/c/en/us/td/docs/security/vpn_modules/6342/vpn_cg/6342site3.html#wp1036092)

lampwins:
One point of clarification on primary IP, it needs to be zero or one of each family type (v4, and v6) per device. I.e. a device may have both a primary v4 and v6 address at the same time.

gsnider2195:
> One point of clarification on primary IP, it needs to be zero or one of each family type (v4, and v6) per device. I.e. a device may have both a primary v4 and v6 address at the same time.

Thanks for the reminder. I probably would have forgotten to account for different address families!

[...]

PR #3399, opened 2023-03-07
Change interface/ip address relationship to many-to-many

gsnider2195:
# Closes: #3236
# What's Changed

# TODO

- [ ] Unit, Integration Tests


nautobot/ipam/models.py
```diff
@@ -966,6 +941,39 @@ def _set_mask_length(self, value):
...
+    interface = models.ForeignKey("dcim.Interface", blank=True, null=True, on_delete=models.CASCADE)
+    vm_interface = models.ForeignKey("virtualization.VMInterface", blank=True, null=True, on_delete=models.CASCADE)
+    source = models.BooleanField(default=False, help_text="Source address")
+    destination = models.BooleanField(default=False, help_text="Destination address")
+    default = models.BooleanField(default=False, help_text="Default address")
+    preferred = models.BooleanField(default=False, help_text="Preferred address")
+    primary = models.BooleanField(default=False, help_text="Primary address")
+    secondary = models.BooleanField(default=False, help_text="Secondary address")
+    standby = models.BooleanField(default=False, help_text="Standby address")
+    primary_for_device = models.BooleanField(default=False, help_text="Primary address for parent device")
```
HanlinMiao, line 955: Just for my knowledge here, why do all the addresses are `BooleanFields`? Do these indicate whether `IPAddressToInterface` assignments have them?

nautobot/ipam/migrations/0022_interface_ipaddress_m2m_data_migration.py
```diff
@@ -0,0 +1,55 @@
...
+    IPAddress = apps.get_model("ipam", "IPAddress")
+    IPAddressToInterface = apps.get_model("ipam", "IPAddressToInterface")
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    Interface = apps.get_model("dcim", "Interface")
+    Device = apps.get_model("dcim", "Device")
+    VMInterface = apps.get_model("virtualization", "VMInterface")
+    VirtualMachine = apps.get_model("virtualization", "VirtualMachine")
+
+    for ip_address in IPAddress.objects.filter(assigned_object_id__isnull=False):
+        related_ct = ContentType.objects.get(id=ip_address.assigned_object_type_id)
```
HanlinMiao, line 15: Can't we use `assigned_object_type` here directly?

nautobot/ipam/models.py
gsnider2195: Here's a configuration example:

```
interface Tunnel0
 bandwidth 180
 ip address 172.17.3.3 255.255.255.0
 no ip directed-broadcast
 tunnel source 10.0.2.4
 tunnel destination 192.24.2.5
 crypto map s1first
!
```

For `Interface` Tunnel0 you would have 3 assocations:

```py
IPAddressToInterface.objects.create(
    ip_address=<IPAddress 172.17.3.3>,
    interface=<Interface Tunnel0>,
)
IPAddressToInterface.objects.create(
    ip_address=<IPAddress 10.0.2.4>,
    interface=<Interface Tunnel0>,
    source=True,
)
IPAddressToInterface.objects.create(
    ip_address=<IPAddress 192.24.2.5>,
    interface=<Interface Tunnel0>,
    destination=True,
)
```

glennmatthews: Maybe it's the field names and/or `help_text` that's misleading then. Should these be things like `is_default`, `is_preferred`, etc.?

nautobot/ipam/models.py
```diff
@@ -966,6 +941,59 @@ def _set_mask_length(self, value):
...
+                    Q(ip_address__in=IPAddress.objects.ip_family(4)) & parent_q
+                ).exclude(id=self.id).update(primary_for_device=False)
+                parent.primary_ip4 = self.ip_address
+                parent.save()
+            if self.ip_address.family == 6 and parent.primary_ip6 != self.ip_address:
+                IPAddressToInterface.objects.filter(
+                    Q(ip_address__in=IPAddress.objects.ip_family(6)) & parent_q
+                ).exclude(id=self.id).update(primary_for_device=False)
+                parent.primary_ip6 = self.ip_address
+                parent.save()
```
glennmatthews, line 988: Two minor optimizations:

1) If the database record was already set as `primary_for_device` (this isn't a change) then there's no need to do this processing and it could be skipped.
2) The `IPAddressToInterface.objects.filter()` could be narrowed down to only select any `primary_for_device=True` instance(s).

nautobot/ipam/models.py
gsnider2195: > Maybe it's the field names and/or `help_text` that's misleading then. Should these be things like `is_default`, `is_preferred`, etc.?

That sounds like a good idea to me

nautobot/ipam/models.py
```diff
@@ -966,6 +941,59 @@ def _set_mask_length(self, value):
...
+
+        if self.interface is not None and self.vm_interface is not None:
+            raise ValidationError(
+                {"interface": "Cannot use a single instance to associate to both an Interface and a VMInterface."}
+            )
+
+        if self.interface is None and self.vm_interface is None:
+            raise ValidationError({"interface": "Must associate to either an Interface or a VMInterface."})
+
+        # if primary_for_device=True, set primary_for_device=False for all other instances on the same device in the same address family
```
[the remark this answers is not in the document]
gsnider2195: I agree and felt the same way while writing this code. I definitely don't think this kind of magic should happen on the through table model but I do think we should check for bad data and raise a validation error

nautobot/ipam/models.py
gsnider2195: I rewrote this to optimize the queryset looking for conflicting instances. Regarding item 1, I didn't know if it was possible to check if the field value has changed in the model `clean` method. I thought this was something we only did in serializer `validate` and form `clean` methods.

glennmatthews: Pattern would be something like:

```python
if self.present_in_database:
    original_data = IPAddressToInterface.objects.get(id=self.id)
    if self.primary_for_device and not original_data.primary_for_device:
        ...
```

nautobot/ipam/filters.py
```diff
@@ -353,29 +353,15 @@ class IPAddressFilterSet(
...
         to_field_name="name",
-        label="VM interface (ID or name)",
-    )
-    vminterface_id = django_filters.ModelMultipleChoiceFilter(
-        field_name="vminterface",
-        queryset=VMInterface.objects.all(),
-        label="VM interface (ID) - Deprecated (use vminterface filter)",
-    )
-    assigned_to_interface = django_filters.BooleanFilter(
-        method="_assigned_to_interface",
```
gsnider2195, line 377: TODO: Should IPAddress have an `assigned_to_interface` filter?

nautobot/ipam/tests/test_forms.py
```diff
@@ -83,51 +83,3 @@ def test_slaac_status_invalid_ipv4(self):
         form = self.form_class(data={self.field_name: "192.168.0.1/32", "status": Status.objects.get(slug="slaac")})
         self.assertFalse(form.is_valid())
         self.assertEqual("Only IPv6 addresses can be assigned SLAAC status", form.errors["status"][0])
-
-    def test_primary_ip_not_altered_if_adding_a_new_IP_to_diff_interface(self):
```
gsnider2195, line 87: TODO: Probably need a similar test in dcim?

gsnider2195:
Ready for review pending test updates from @bryanculver

bryanculver (commented):
Let's make sure if we are pretty sure will need a data migration in DCIM or Virtualization we stub them out now so people aren't having to re-order migrations knowingly in the future.

I don't _think_ we will and I think the one in IPAM should be the only one that needs to exist.

bryanculver (approved):
☝️ 

I have a list of follow-on items noted in the PR description that I think can all be in another PR. This allows people to have the M2M in place and not have a bunch of contentious migrations in-flight. If anyone has other thoughts or thinks any of my suggestions should hold up this PR, I wouldn't argue.

nautobot/dcim/forms.py
```diff
@@ -1712,31 +1712,33 @@ def __init__(self, *args, **kwargs):
...
-                        assigned_object_type=ContentType.objects.get_for_model(Interface),
-                        assigned_object_id__in=interface_ids,
-                    )
-                    .prefetch_related("assigned_object")
-                )
-                if interface_ips:
-                    ip_list = [(ip.id, f"{ip.address} ({ip.assigned_object})") for ip in interface_ips]
+                interface_ip_assignments = IPAddressToInterface.objects.filter(
+                    interface__in=interface_ids
+                ).prefetch_related("ip_address")
```
jathanism, line 1717: This should be `select_related`, and then you can also ditch that `v2 TODO` comment. :) 

```suggestion
                ).select_related("ip_address")
```

jathanism (commented):
Nothing blocking, just a few observations.

nautobot/dcim/forms.py
jathanism: Covered by #3411

PR #3408, opened 2023-03-09
UI docker dev environment tweaks

glennmatthews:
# Closes: #n/a
# What's Changed

- Add some exclusions to `.dockerignore` so that the docker build context doesn't bloat with local `node_modules` and `build` files when switching between local and docker setups.
- Changed `build_ui` handling of npm commands to more verbose by default, but capturing stdout/stderr and only showing them to the user if a failure occurs
- Instead of mounting `nautobot/ui/node_modules` into the container, use a Docker volume for this purpose and share it between `nautobot` and `nodejs` containers.
   - Appears much faster when building from zero
   - Means that `invoke destroy` will blow away this volume, good for troubleshooting or returning to a clean slate
   - Since `nautobot` runs as the `root` user currently, I had to change the `nodejs` to also run as `root` rather than `node` because otherwise file permissions on `node_modules` became a problem.
- To avoid contention between both `nautobot` and `nodejs` containers trying to run `npm install` on startup, I changed it so that `nodejs` waits until `nautobot` is healthy (`build_ui` has completed successfully) before starting and no longer takes responsibility for running `npm install` itself.

# TODO

- n/a Unit, Integration Tests

### DIFF
--- a/nautobot/core/management/commands/build_ui.py
+++ b/nautobot/core/management/commands/build_ui.py
@@ -94,19 +94,24 @@ class Command(BaseCommand):
             message (str): Message to display when command is executed
         """
         self.stdout.write(self.style.WARNING(message))
+        self.stdout.write(f"Running '{command}' in '{cwd}'...")
 
         try:
-            out = subprocess.run(
+            result = subprocess.run(
                 shlex.split(command),
                 check=False,
                 cwd=cwd,
                 env={**os.environ.copy()},
+                encoding="utf-8",
+                capture_output=True,
             )
         except FileNotFoundError as err:
-            raise CommandError(f"`{command}` failed with error: {err}")
+            raise CommandError(f"'{command}' failed with error: {err}")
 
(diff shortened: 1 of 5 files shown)


nautobot/core/management/commands/build_ui.py
```diff
@@ -94,19 +94,24 @@ def run_command(self, command, message, cwd=settings.NAUTOBOT_UI_DIR):
...
                 shlex.split(command),
                 check=False,
                 cwd=cwd,
                 env={**os.environ.copy()},
+                encoding="utf-8",
+                capture_output=True,
             )
         except FileNotFoundError as err:
-            raise CommandError(f"`{command}` failed with error: {err}")
+            raise CommandError(f"'{command}' failed with error: {err}")
```
jathanism, line 109: One thing of note is that this is what happens when `npm` isn't found, such as if it wasn't installed.

jathanism (approved):

Issue #3407, opened 2023-03-09
Non-admin users don't have permission to ContentType API endpoint


joewesch:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.7
* Python version: 3.7.16
* Database platform, version: Docker
* Middleware(s): N/A

### Steps to Reproduce
1. Create a model that uses a ManyToManyField to ContentType
```
class MyModel(PrimaryModel):
    content_types = models.ManyToManyField(to=ContentType)
```
2. Create a form for that Model
```
class MyModelForm(BootstrapMixin, forms.ModelForm):
    content_types = DynamicModelMultipleChoiceField(
        queryset=ContentType.objects.all(),
        label="Object Types",
        widget=APISelectMultiple(
            api_url="/api/extras/content-types/",
        ),
    )
```
3. Create a non-admin user with applicable/all permissions
4. Try and create the object with the non-admin user

### Expected Behavior
The non-admin user should be able to select a content type

### Observed Behavior
After clicking the dropdown for "Object Types" you get the following error:
> The results could not be loaded.

In the Nautobot logs, you see the following:
> Forbidden: /api/extras/content-types/
> "GET /api/extras/content-types/?q=&limit=50&offset=0&brief=true HTTP/1.1" 403 63

### Analysis
I was able to reproduce the issue with the demo user on demo.nautobot.com.
1. Go to the Change Logs at https://demo.nautobot.com/extras/changelog/
2. Click on the "Filter" button
3. Try and select an "Object Type" filter

You can also see this by going to https://demo.nautobot.com/api/extras/content-types/ and getting this error:
```
{
    "detail": "You do not have permission to perform this action."
}
```

PR #3409, opened 2023-03-09
Removed permission checks for ContentTypeAPIViewSet

HanlinMiao:
# Closes: #3407 
# What's Changed

Now ContentType API endpoints are accessible for non-admin users.
Screen Shot 2023-03-09 at 12 08 43 PM
Screen Shot 2023-03-09 at 12 08 35 PM

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/3407.removed
+++ b/changes/3407.removed
@@ -1,0 +1,1 @@
+Removed permission checks for ContentTypeAPIViewSet.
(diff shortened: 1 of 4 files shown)


jathanism (approved):

HanlinMiao:
> Do we have _any_ tests for this viewset? Should we add one for this issue?
Yeah I can add one

HanlinMiao:
> > Do we have _any_ tests for this viewset? Should we add one for this issue?
> > Yeah I can add one

So It turns out that we already have tests for ContentTypeViewSet but with permission override. So I removed the permission override and the tests should still pass.

glennmatthews:
> > > Do we have _any_ tests for this viewset? Should we add one for this issue?
> > > Yeah I can add one
> 
> So It turns out that we already have tests for ContentTypeViewSet but with permission override. So I removed the permission override and the tests should still pass.

Thanks! Go figure, I checked and that set of tests dates back several years...

Issue #2311, opened 2022-08-25
Editing a Location doesn't retrieve parent


itheodoridis:
### Environment
* Python version: 3.10
* Nautobot version: 1.4.1

### Steps to Reproduce
1. Create a location and a child location (e.g. Building, Floor).
2. Edit the child location (e.g. Floor)
3. Observe the parent field.

### Expected Behavior
The parent location should be retrieved and displayed in the parent field box.

### Observed Behavior
The parent location is not retrieved and if one hits the update button regardless a warning is issued to select a parent location and the update is denied.

PR #553, opened 2021-06-09
Correctly handle single values when prepopulating a DynamicModelMultipleChoiceField

glennmatthews:
### Fixes: #512 

When we're pre-populating a form based on GET request query parameters (e.g. `/dcim/devices/add/?tags=c671a001-4c17-4ca1-80fd-fe1609bcadec`), when a single value is provided for a given parameter, the generic request handling has no way to know whether this string value is a value for a single-value field (`DynamicModelChoiceField`) or a list of one value to use for a multi-value field (`DynamicModelMultipleChoiceField`). In the latter case, if we provide a string to initialize a multi-value field, something in Django and/or `django-filter` (I have had trouble tracking down exactly where and why) ends up misinterpreting the string as a list of single characters, and of course single characters are not valid UUIDs, resulting in the error reported in #512.

> Interestingly, this appears to be something of a baseline bug inherited from NetBox, although the symptom is different there due to the use of integer PKs instead of UUIDs - a single-value integer PK passed as a parameter actually works correctly **so long as it is a single-digit PK**, but a request providing a single multi-digit PK appears to silently fail somewhere in the code and the field gets an empty value instead of being pre-populated as intended.

The best solution I've found is to implement the `prepare_value` function for `DynamicModelMultipleChoiceField` to intercept single string values and wrap them into a single-entry list before passing them up the chain. Interestingly, `prepare_value` can also be called with an object instance rather than a string/PK, and in that case we need to preserve the single value as such and *not* wrap it into a list, as doing so causes the field to not correctly pre-populate (although no error is reported). Again, I'm not entirely sure why, but can confirm this behavior.

Issue #512, opened 2021-05-31
Cloning Device or IP with a tag applied results in a Server Error


waydh:
### Environment
* Ubuntu 20.04
* Python version: 3.8.5
* Nautobot version: 1.0.2

### Steps to Reproduce
1. Install NautoBot 1.0.2 per the directions on https://nautobot.readthedocs.io
2. Create a Site, Manufacturer, Device Role, Tag, and then a Device with all applied
3. Click clone on the created device and receive a Server Error
4. Remove the tag from the device (so there are no tags at all) click Clone and it works
5. Issue also occurs with IPs

### Expected Behavior
Clone button works without error allowing the device to be cloned

### Observed Behavior
Server Error like the following - the invalid UUID is not always 'a', it changes alphabet or number

 Server Error

There was a problem with your request. Please contact an administrator.

The complete exception is provided below:

<class 'django.core.exceptions.ValidationError'>

['“a” is not a valid UUID.']

Python version: 3.8.5
Nautobot version: 1.0.2

If further assistance is required, please post to the #nautobot channel on NetworkToCode's Slack.

glennmatthews:
Thanks for the report! This may be related to #499.

glennmatthews:
Looks to be an issue with pre-populating values for any form that uses a `DynamicModelMultipleChoiceField`. Continuing to investigate.

PR #3412, opened 2023-03-10
Fix `LocationForm.parent` auto-fill (again), by revamping `prepare_value` yet again

### LINKED ISSUE #2470: Child location's slug is autogenerated using parent's UUID rather than slug
### Environment
* Nautobot version (Docker tag too if applicable): v1.4.2
* Python version: 3.8
* Database platform, version: PostgreSQL v13

### Steps to Reproduce
1. Create root location type
2. Create child location type of the root location type
3. Create root location
4. Create child location of the root location

### Expected Behavior
Child Location's slug auto generated using the parent location's slug

Screen Shot 2022-09-22 at 10 25 28 AM

### Observed Behavior
Child Location's slug auto generated using the parent location's UUID

Screen Shot 2022-09-22 at 10 28 18 AM

### LINKED ISSUE #3347: Editing a Location does not auto-fill parent field in edit form
### Environment
* Nautobot version (v2.0.0a1)
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Create a hierarchy of LocationTypes
2. Create parent Location
3. Create child Location
4. Edit child Location

### Expected Behavior
Newly opened Location edit-form should auto-fill parent

### Observed Behavior
Newly opened Location edit-form does not auto-fill parent
Screenshot 2023-02-22 at 09 52 14
Screenshot 2023-02-22 at 09 52 30

glennmatthews:
# Closes: #3347 
# What's Changed

It turned out that #2773, in trying to fix #2470, did partially regress #2311, resulting in the filing of #3347.  Along the way, fixing this intersected with code introduced way back in #553 to address #512, resulting in me needing to rework that bit of code as well. Then manual testing to verify that I've fixed #3347/#2311 without reintroducing #2470 or #512. Wheeeee!

- The root of the issue in #512 was that `normalize_querydict` could only guess whether a given query parameter should be recorded as a single value or a list of values based on how many times it appeared in the query, meaning that a list parameter that was only used once in a given query would be incorrectly treated as a single value rather than a list with one entry. This would then cause problems in prepopulating an object form based on this incorrect treatment.
   - Solution then: add a special case in `DynamicModelMultipleChoiceField.prepare_value` to convert a single string value into a list value containing that single string.
       - The problem with this was, because of some Django internals involving rendering the individual `<option>` elements for the form field, that when provided a single *object* value (a model instance), it needed to stay as that single value, and *not* be converted to a list of a single object.
   - Solution now: remove the former code, and instead add an optional `form_class` parameter to `normalize_querydict` and update model views to provide this parameter as appropriate; when present, this parameter is used to check what form field a given query parameter corresponds to, and if the field is a multiple-choice field, correctly convert the single-valued parameter to a single-entry list.
   - This did require moving some code around in extras to support the `JobView` case, where the form class is generated on-the-fly from the selected Job, to make sure the form class was available at the time that the querydict is interpreted. This should have no functional impact.

- The underlying issue with #2311 and #2470 "feels" like a latent Django bug to me but I'm probably missing something. Basically what it boils down to is that with a baseline `ModelChoiceField` that defines a `to_field_name` value:
    - `prepare_value(object_instance)` returns the value of the instance's field corresponding to the `to_field_name` specified (or the instance's PK if no `to_field_name`), which correctly allows the rendered form to pre-select the appropriate `<option value="field_value">` in the HTML.
    - but `prepare_value(object_pk)` simply returns the provided PK,, *regardless of any `to_field_name`*, which doesn't match up with any `<option value="field_value">` in the HTML, resulting in no pre-selection in the rendered form.
    - So it *appears* that the fix here is to subclass the `prepare_value()` method and ensure that, when given a PK, it looks up the actual object instance (if any) and passes that instance through to the superclass for processing, ensuring correct form rendering in this case as well.

Cases manually tested:

- Look for regression of #512 by pressing the `Clone` button on object instances with zero, exactly one, or more than one Tag applied, (in other words, zero, one, and more than one occurrence of `tags=` in the GET query parameters) and verifying that in all three cases the form is correctly prepopulated with the appropriate tag selection:

image

image

image

- Look for regression of #2311/#2470/#3347 by:
   - Editing an existing Location and confirming that the `parent` and all other model choice fields are correctly populated (this case was missed in testing #2773 it appears, resulting in #3347) and that slug autogeneration in the UI works correctly

image

   - Hitting the `Clone` button on an existing location and confirming that the `parent` and other appropriate model choice fields are correctly populated (this case was tested for in #2773) and that slug autogeneration in the UI works correctly

image

   - Hitting the `Add child` button on an existing location and confirming that the `parent` field is correctly populated and that slug autogeneration in the UI works correctly (this is functionally quite similar to the "Clone" case but with far fewer autopopulated selections)

image

**I have not spent time writing unit or integration test automation for the above cases because much of this functionality will likely change drastically in 2.0 rendering such tests irrelevant.**

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/changes/3347.fixed
+++ b/changes/3347.fixed
@@ -1,0 +1,1 @@
+Fixed (again) `Location.parent` not populating correctly in the form when editing an existing Location.
--- a/nautobot/core/views/generic.py
+++ b/nautobot/core/views/generic.py
@@ -414,7 +414,7 @@ class ObjectEditView(GetReturnURLMixin, ObjectPermissionRequiredMixin, View):
     def get(self, request, *args, **kwargs):
         obj = self.alter_obj(self.get_object(kwargs), request, args, kwargs)
 
-        initial_data = normalize_querydict(request.GET)
+        initial_data = normalize_querydict(request.GET, form_class=self.model_form)
         form = self.model_form(instance=obj, initial=initial_data)
         restrict_form_fields(form, request.user)
 
--- a/nautobot/core/views/renderers.py
+++ b/nautobot/core/views/renderers.py
@@ -170,7 +170,7 @@ class NautobotHTMLRenderer(renderers.BrowsableAPIRenderer):
             elif view.action == "destroy":
                 form = form_class(initial=request.GET)
             elif view.action in ["create", "update"]:
-                initial_data = normalize_querydict(request.GET)
+                initial_data = normalize_querydict(request.GET, form_class=form_class)
                 form = form_class(instance=instance, initial=initial_data)
                 restrict_form_fields(form, request.user)
             elif view.action == "bulk_destroy":
--- a/nautobot/extras/jobs.py
+++ b/nautobot/extras/jobs.py
@@ -303,6 +303,15 @@ class BaseJob:
         """
         return self.job_result.data if self.job_result else None
 
+    def as_form_class(self):
+        """
+        Dynamically generate a Django form class corresponding to the variables in this Job.
+
+        In most cases you should use `.as_form()` instead of calling this method directly.
+        """
+        fields = {name: var.as_field() for name, var in self._get_vars().items()}
+        return type("JobForm", (JobForm,), fields)
+
     def as_form(self, data=None, files=None, initial=None, approval_view=False):
         """
         Return a Django form suitable for populating the context data required to run this Job.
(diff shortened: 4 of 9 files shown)


bryanculver (approved):
This is so meta it hurts my brain.

PR #3413, opened 2023-03-10
Update DCIM ChoiceSets from netbox

### LINKED ISSUE #3377: Add new `dcim.choices`
### As ...

Ozzie - Operator

### I want ...

to have more `dcim.choices` for:

- `PortTypeChoices`
- `PowerPortTypeChoices`
- `InterfaceTypeChoices`

and others, such as the ones added to Netbox lately. Idea: An import of the differences in the dcim/choices.py file in [Netbox](https://github.com/netbox-community/netbox/blob/develop/netbox/dcim/choices.py) vs [Nautobot](https://github.com/nautobot/nautobot/blob/develop/nautobot/dcim/choices.py) would make it pretty straight-forward.

### So that ...

I could have more type options to choose.

- The newer type choices are available when creating the related `dcim` objects

- [ ] Automation
- [ ] Circuits
- [X] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

_No response_

### External Dependencies

_No response_

alextremblay:
# Closes: #3377
# What's Changed

Brought over missing entries from netbox for the following `ChoicesSet`s in `nautobot/dcim/choices.py`:
- `RackTypeChoices`
- `ConsolePortTypeChoices`
- `PowerPortTypeChoices`
- `PowerOutletTypeChoices`
- `InterfaceTypeChoices`
- `PortTypeChoices`
- `CableLengthUnitChoices`

# TODO

- [ ] Unit, Integration Tests


nautobot/dcim/choices.py
```diff
@@ -762,7 +865,7 @@ class InterfaceTypeChoices(ChoiceSet):
     TYPE_16GFC_SFP_PLUS = "16gfc-sfpp"
     TYPE_32GFC_SFP28 = "32gfc-sfp28"
     TYPE_64GFC_QSFP_PLUS = "64gfc-qsfpp"
-    TYPE_128GFC_QSFP28 = "128gfc-sfp28"
+    TYPE_128GFC_QSFP28 = "128gfc-qsfp28"
```
bryanculver, line 868: If I'm not mistaken this will cause data issues for those that would have the previous value.

bryanculver (changes requested):

nautobot/dcim/choices.py
alextremblay: right, that's the string that ends up in the database. 

What would be the optimal solution? should I put it back or make a migration?

bryanculver: Put it back for now. A migration right now in `develop` for something like this is overkill. It's not like there is a `TYPE_128GFC_SFP28` and `TYPE_128GFC_QSFP28` where the choice set is having a hard time picking between the two. In practice, the user has no significant impact, other than what they see in the API.

I could be persuaded otherwise @nautobot/core

changes/3377.added
```diff
@@ -0,0 +1 @@
+Added addiotional choices for many data types in `nautobot.dcim`
```
bryanculver, line 1:
```suggestion
Added additional choices for many data types in `nautobot.dcim`
```

nautobot/dcim/choices.py
glennmatthews: FWIW here's the commit in netbox that made this change and included the necessary corresponding data migration: https://github.com/netbox-community/netbox/commit/b7b5a5788f64e8b2a89b8153fd4a2554cb0f647d.

We *might* want to actually make the change and migration in 2.0, as my understanding is that `qsfp` is more correct in the case of this interface type?

nautobot/dcim/choices.py
```diff
@@ -1117,14 +1312,21 @@ class CableStatusChoices(ChoiceSet):
 
 class CableLengthUnitChoices(ChoiceSet):
 
+    # Metric
+    UNIT_KILOMETER = "km"
     UNIT_METER = "m"
     UNIT_CENTIMETER = "cm"
+
+    # Imperial
+    UNIT_MILE = "mi"
```
glennmatthews, line 1321: These need corresponding logic in the `to_meters` utility function to handle these units - see https://github.com/netbox-community/netbox/commit/a6d937b366931670d8cc19b5dd6564f2efee3d85 and https://github.com/netbox-community/netbox/blob/develop/netbox/utilities/utils.py#L276 for reference.

changes/3377.added
```diff
@@ -0,0 +1 @@
+Added additional choices for many data types in `nautobot.dcim`
```
glennmatthews, line 1:
```suggestion
Added additional choices for many data types in `nautobot.dcim`.
```

glennmatthews (changes requested):
Good stuff!

nautobot/dcim/choices.py
alextremblay: good catch! I've added it

nautobot/utilities/utils.py
```diff
@@ -340,6 +340,8 @@ def to_meters(length, unit):
         return length * Decimal("0.3048")
     if unit == CableLengthUnitChoices.UNIT_INCH:
         return length * Decimal("0.3048") * 12
+    if unit == CableLengthUnitChoices.UNIT_MILE:
```
glennmatthews, line 343: Don't we also need to add a case for `CableLengthUnitChoices.UNIT_KILOMETER`?

alextremblay: ah shoot, you're right. I totally missed that.

It's now been added.

glennmatthews (approved):

jmcgill298:
@alextremblay I added `TYPE_OTHER` to a couple of choicesets and put a PR to your branch: https://github.com/alextremblay/nautobot/pull/1

timizuoebideri1 (approved):

alextremblay:
> @alextremblay I added `TYPE_OTHER` to a couple of choicesets and put a PR to your branch: [alextremblay#1](https://github.com/alextremblay/nautobot/pull/1)

Thank you!

bryanculver (approved):

Issue #2900, opened 2022-11-22
Implement `django-natural-keys` as base for `BaseModel`


jathanism:
### Proposed Changes

Base models should inherit from [`django-natural-keys`](https://github.com/wq/django-natural-keys) base `NaturalKeyModel` and use `NaturalKeyModelManager` so that…

- `natural_key()` is automatically provided using the uniqueness constraints
    - Currently for each model, we must define `natural_key()` method 
- `Foo.objects.get_by_natural_key()` is also automatically provided, reducing boilerplate
    - Currently for each model, we must subclass `RestrictedQuerySet` in order to publish a `get_by_natural_key()` method
    - And then set `objects = FooRestrictedQuerySet.as_manager()` on the model
- Each model will also receive a `natural_key_slug` that is *immutable* derived from the natural key fields for that object. This can be used in URL fragments in the  "natural key or PK" pattern we have begun to adopt for filters. 

### Justification

We have decided to align around natural keys. We need less boilerplate, not more. In the common case, this should work automatically and result in objects having deterministic uniqueness/natural key constraints with minimal effort. For the uncommon/edge case, we will need to adapt and revise the pattern.

bryanculver:
https://github.com/nautobot/nautobot/pull/2692

glennmatthews:
Since `django-natural-keys` provides a minimal/naive `natural_key_slug` implementation (for one, it doesn't account for the possibility that any of the key components may contain a `-` character), we probably need to come up with our own design here. Since one of the main uses of natural-key-slugs may be in URL patterns, we need something that's URL-safe - therefore I'm proposing that we consider either [percent-encoding](https://en.wikipedia.org/wiki/Percent-encoding) or [punycode](https://en.wikipedia.org/wiki/Punycode) as a starting point.

bryanculver:
Completed with #3423

PR #3423, opened 2023-03-14
Natural keys and natural_key_slug inspired by django-natural-keys

glennmatthews:
# Closes: #2900
# What's Changed

- ~Add `django-natural-keys` aka `natural_keys` as a library dependency~  No longer a library dependency
  - ~Currently I'm overriding/extending roughly 50% of the code base in `django-natural-keys` to add features and fixes that we need (primarily handling for `Location`'s variadic natural key and the fact that `ContentType` is not a `NaturalKeyModel` but we have models that have natural-keys referencing it). I think a number of my extensions here would make sense to at least try contributing upstream, although it doesn't appear to be a very active project presently.~
  - [x] TODO: review all of my overrides/extensions and see if there's a smaller/simpler set of changes that could achieve the same results; the current changes have been added piecemeal as I encountered various limitations and issues.
- ~Change Nautobot's `BaseModel` to inherit from `NaturalKeyModel` and `RestrictedQuerySet` to inherit from `NaturalKeyQuerySet`.~ Additionally add a new `BaseManager` as derived from `NaturalKeyManager`.
   - Change all model `objects` declarations as needed to ensure that they correctly are based on `BaseManager`.
- Add a custom implementation of `natural_key_slug` based around URL-encoding of strings to produce a "slug-like" string that is lossless and can still be converted back to the original natural key on demand.
- Add a `ModelTestCases.BaseModelTestCase` class in core that includes a couple of natural-key tests, and extended a number of existing `test_models.py` test files to use this base class where doing so would be reasonably simple.
   - [ ] TODO: add basic tests for all models as a way to assert that all models' natural-keys are working correctly. (We get pretty good coverage of this for many models already when using `--cache-test-fixtures` and running tests, as that uses dumpdata/loaddata with natural-keys enabled...)
- Add a custom natural-key implementation for `Location` because it has a self-referential/recursive natural key that needs some special handling to avoid infinite recursion when trying to determine the potential scope of the natural key.
- **For now, *exclude* `Prefix` and `IPAddress` from having a `natural_key` implementation because those models don't actually have a reasonable natural key as presently defined.**

```python
>>> DeviceType.objects.first().natural_key()
['Arista', 'Arista DeviceType 15']

>>> DeviceType.objects.get_by_natural_key("Arista", "Arista DeviceType 15")
<DeviceType: Arista DeviceType 15>

>>> DeviceType.objects.first().natural_key_slug
'Arista&Arista+DeviceType+15'

>>> DeviceType.objects.get(natural_key_slug='Arista&Arista+DeviceType+15')
<DeviceType: Arista DeviceType 15>
```

# TODO

- [x] Unit, Integration Tests
  - [x] Investigate how hard it would be to add URL patterns for natural-key-slugs for all models in any sort of automated way?


nautobot/core/models/tree_queries.py
```diff
@@ -15,10 +14,13 @@ def max_tree_depth(self):
         """
         Get the maximum depth of any tree in this queryset.
         """
-        return self.with_tree_fields().extra(order_by=["-__tree.tree_depth"]).first().tree_depth
+        deepest = self.with_tree_fields().extra(order_by=["-__tree.tree_depth"]).first()
+        if deepest is not None:
+            return deepest.tree_depth
+        return 0
```
jathanism, line 20: I'm curious why this change was necessary?

nautobot/core/models/__init__.py
```diff
@@ -1,11 +1,58 @@
...
+            try:
+                nested_key = extract_nested_key(kwargs, rel_to, name)
+            except AttributeError:
+                # Handle case where rel_to isn't a NaturalKeyModel, e.g. ContentType
+                nested_key = []
+                for key in kwargs.keys():
+                    if key.startswith(f"{name}__"):
+                        nested_key.append(kwargs[key])
+                if all(key is None for key in nested_key):
+                    nested_key = None
```
glennmatthews, line 41: There's almost certainly a more elegant way to do this - possibly by reimplementing `extract_nested_key` directly to handle the non-natural-key-model cases within it.

nautobot/core/models/__init__.py
```diff
@@ -1,11 +1,58 @@
...
+from django.contrib.contenttypes.models import ContentType
 from django.db import models
+from natural_keys import NaturalKeyModel, NaturalKeyModelManager
+from natural_keys.models import extract_nested_key
 
 from nautobot.core.models.querysets import RestrictedQuerySet
 
 
-class BaseModel(models.Model):
+class BaseManager(NaturalKeyModelManager):
```
jathanism, line 12: I'm thinking it's probably best to have this live in `nautobot.core.models.managers` just to preempt any potential import issues.

nautobot/core/models/__init__.py
```diff
@@ -50,3 +97,92 @@ def validated_save(self):
...
+        """
+        vals = []
+        for name_list in [name.split("__") for name in self.get_natural_key_fields()]:
+            val = self
+            for attr_name in name_list:
+                val = getattr(val, attr_name)
+                if val is None:
+                    break
+            vals.append(val)
+        # Strip trailing Nones from vals
```
glennmatthews, line 116: This means an (as-yet-undocumented) best practice is, for any model like Location that has a variadic natural key, the variadic portion *needs* to be the last item in the natural key - i.e. `(name, parent)`, not `(parent, name)` in the Location case.

nautobot/core/models/__init__.py
```diff
@@ -50,3 +97,92 @@ def validated_save(self):
...
+        cleaned_vals = []
+        for val in reversed(vals):
+            if val is None and not cleaned_vals:
+                continue
+            cleaned_vals = [val] + cleaned_vals
+        return cleaned_vals
```
glennmatthews, line 122: There's probably a clever one-liner with `filter()` or similar that could be used to map `vals` to `cleaned_vals`.

nautobot/core/models/__init__.py
```diff
@@ -50,3 +97,92 @@ def validated_save(self):
...
+
+    natural_key_separator = "/"
```
glennmatthews, line 124: As Gary points out, per https://www.rfc-editor.org/rfc/rfc2396 there are several other characters we might consider, such as `~` or `!`

nautobot/core/models/__init__.py
```diff
@@ -50,3 +97,92 @@ def validated_save(self):
...
+
+    @property
+    def natural_key_slug(self):
+        """Less naive implementation than django-natural-keys provides by default, based around URL percent-encoding."""
```
glennmatthews, line 128: Add a note that this also renders `None` values as encoded null characters (`%00`).

nautobot/extras/managers.py
```diff
@@ -1,14 +1,16 @@
...
 
+from nautobot.core.models import BaseManager
 from nautobot.core.models.querysets import RestrictedQuerySet
 
 
 # TODO(jathan): This subclass is a hack. We'll fix it in post. Realistically the work for
 # `django-natural-keys` and establishing managers vs. querysets.as_manager() patterns across the
 # board will be the right palce to do this, as we can then just subclass the manager class.
-class JobResultManager(RestrictedQuerySet.as_manager().__class__, TaskResultManager):
+class JobResultManager(BaseManager.from_queryset(RestrictedQuerySet), TaskResultManager):
```
jathanism, line 13: With this change, the above `TODO` comment can be removed.

nautobot/core/models/__init__.py
```diff
@@ -50,3 +97,92 @@ def validated_save(self):
...
+        return self.natural_key_separator.join(
+            quote_plus(str(value) if value is not None else "\0") for value in self.natural_key()
+        )
+
+    @classmethod
+    def get_natural_key_def(cls):
+        """
+        Extend django-natural-keys implementation to recognize that UUIDFields are not relevant to the natural key.
```
glennmatthews, line 136: It might be better to instead specifically exclude the `id`/`pk` field rather than excluding all AutoField/BigAutoField/UUIDField fields?

nautobot/core/models/__init__.py
```diff
@@ -50,3 +97,92 @@ def validated_save(self):
...
+        natural_key_fields = []
+        for name, rel_to in cls.get_natural_key_info():
+            if not rel_to:
+                natural_key_fields.append(name)
+            else:
+                try:
+                    nested_key = rel_to.get_natural_key_fields()
+                except AttributeError:
+                    if rel_to == ContentType:
+                        nested_key = ["app_label", "model"]
```
glennmatthews, line 184: I don't particularly like this special case but it works. If we reimplement/vendor `django-natural-keys` in its entirety, we might try to get rid of `get_natural_key_fields` somehow to where this wouldn't be needed?

nautobot/core/models/tree_queries.py
glennmatthews, line 20: Not sure why this hasn't been an issue previously, but I definitely hit a case while testing this where `max_tree_depth` was being called with no model instances present, hence this fix.

nautobot/dcim/models/locations.py
```diff
@@ -114,11 +114,29 @@ def clean(self):
 
 
 class LocationQuerySet(TreeQuerySet):
+    @classmethod
+    def as_manager(cls, with_tree_fields=False):
+        """Get an appropriate TreeManager subclass for working with Locations."""
+        LocationManager = TreeManager.from_queryset(cls)
+        LocationManager._built_with_as_manager = True
+        LocationManager._with_tree_fields = with_tree_fields
+        return LocationManager()
```
glennmatthews, line 123: Need to revisit this - I don't think this override of `as_manager` *should* be needed but I was running into issues without it early on.

nautobot/dcim/models/locations.py
```diff
@@ -114,11 +114,29 @@ def clean(self):
...
+
     def get_for_model(self, model):
         """Filter locations to only those that can accept the given model class."""
         content_type = ContentType.objects.get_for_model(model._meta.concrete_model)
         return self.filter(location_type__content_types=content_type)
 
+    def natural_key_kwargs(self, *args):
```
glennmatthews, line 130: Now that I have a custom `natural_key_kwargs` on `RestrictedQuerySet`, it feels wrong to also be overriding it in `LocationQuerySet` - need to revisit this.

nautobot/dcim/models/locations.py
```diff
@@ -251,6 +269,32 @@ class Meta:
     def __str__(self):
         return self.name
 
+    @classmethod
+    def get_natural_key_fields(cls):
+        natural_key = []
+        name = "name"
+        for _ in range(cls.objects.max_tree_depth() + 1):
```
jathanism, line 276: Nevermind, I see now. This is why you're explicitly returning `0` from `max_tree_depth`

nautobot/dcim/models/locations.py
```diff
@@ -251,6 +269,32 @@ class Meta:
...
+    def natural_key(self):
+        """
+        Custom natural key implementation for Location.
+
+        Needed because with the uniqueness constraint of (name, parent) we end up with a theoretically infinite
+        recursive natural key.
+        """
+        try:
+            return [ancestor.name for ancestor in reversed(self.ancestors(include_self=True))]
+        except Location.DoesNotExist:
```
glennmatthews, line 290: I ran into a case (IIRC it was during dumpdata/loaddata) where `self.ancestors()` was returning objects that threw a DoesNotExist exception, hence this fallback logic.

nautobot/dcim/tests/test_models.py
```diff
@@ -686,49 +695,50 @@ def setUp(self):
 
         DeviceBayTemplate(device_type=self.device_type, name="Device Bay 1").save()
 
-    def test_device_creation(self):
-        """
-        Ensure that all Device components are copied automatically from the DeviceType.
-        """
-        d = Device(
+        self.device = Device(
```
glennmatthews, line 698: Diffs here are noisy but basically just boil down to moving a common `Device` instance out of a bunch of individual test functions and into `setUp` so that it exists for the natural-key tests in `BaseModelTestCase`.

nautobot/extras/models/jobs.py
```diff
@@ -594,6 +593,8 @@ class JobResult(BaseModel, CustomFieldModel):
 
     objects = JobResultManager()
 
+    _natural_key = ["task_id"]
```
glennmatthews, line 596: Needed presently because `task_id` is `unique=True` but since it's a `UUIDField` it gets overlooked by django-natural-keys.

nautobot/extras/models/models.py
```diff
@@ -519,6 +519,8 @@ class FileProxy(BaseModel):
     )
     uploaded_at = models.DateTimeField(auto_now_add=True)
 
+    _natural_key = ["name", "uploaded_at"]
```
glennmatthews, line 522: Not actually a unique together constraint presently but "close enough" since we're unlikely to have two files uploaded with the same name at the exact same timestamp.

nautobot/extras/models/tags.py
```diff
@@ -80,9 +80,28 @@ def validate_content_types_removal(self, content_types_id):
...
+
     class Meta:
         index_together = ("content_type", "object_id")
+
+    @classmethod
+    def get_natural_key_fields(cls):
+        return ["tag__name", "content_type__app_label", "content_type__model", "object_id"]
+
+    def natural_key(self):
+        return [self.tag.name, self.content_type.app_label, self.content_type.model, self.object_id]
```
glennmatthews, line 107: Need to revisit this - might not be needed now that I have more generic handling for content-type FKs in the base classes.

nautobot/ipam/models.py
```diff
@@ -870,6 +890,8 @@ class Meta:
         verbose_name = "VLAN group"
         verbose_name_plural = "VLAN groups"
 
+    _natural_key = ["name", "location"]
```
glennmatthews, line 893: Not actually a uniqueness constraint on VLANGroup due to the good old NULL != NULL problem. But close for the moment.

nautobot/users/models.py
```diff
@@ -29,6 +28,14 @@
 #
 
 
+class UserManager(BaseManager, UserManager_):
+    """
+    Natural-key subclass of Django's UserManager.
+
+    Note that this is *NOT* based around RestrictedQuerySet.
```
glennmatthews, line 35: Not sure why this is the case, but changing it caused some RackReservation test cases to fail due to newly appearing User access constraints.

nautobot/extras/models/tags.py
```diff
@@ -80,9 +80,28 @@ def validate_content_types_removal(self, content_types_id):
...
 
 
+class TaggedItemManager(BaseManager.from_queryset(RestrictedQuerySet)):
+    def get_by_natural_key(self, *args):
+        return self.get(
+            tag__name=args[0],
+            content_type__app_label=args[1],
+            content_type__model=args[2],
+            object_id=args[3],
+        )
```
jathanism, line 90: Instead of indexing here with positional args, can we please use named positional arguments?
```suggestion
    def get_by_natural_key(self, tag_name, app_label, model, object_id):
        return self.get(
            tag__name=tag_name,
            content_type__app_label=app_label,
            content_type__model=model,
            object_id=object_id,
        )
```

nautobot/extras/models/jobs.py
jathanism, line 596: This is required because we told it to not include `UUIDField` by default, right? Can you please add a brief comment for that?

jathanism (changes requested):

nautobot/ipam/models.py
jathanism: So what happens if `location` is null in this instance?

nautobot/core/models/__init__.py
jathanism: Good idea actually since we know we may very well have a need for `UUIDField` in the future, such as the current example for `JobResult.task_id`.

nautobot/core/models/__init__.py
jathanism: We need to be cautious about whether these values would be permitted in fields for which uniqueness would be considered, for example `Device.name` or `Interface.name`, etc.

nautobot/core/models/__init__.py
jathanism: All things considered this isn't all that bad.

nautobot/core/models/__init__.py
jathanism: Variadic in that it depends on how deeply nested it is?

glennmatthews: Yeah exactly - it can be `("me")` (if no parent), or `("me", "parent", "grandparent")`, or `("me", "parent", "grandparent", "great-gp", ...)`.

nautobot/ipam/models.py
glennmatthews: if we have two VLANGroups with the same name and a null location, then get_by_natural_key will be unhappy:

```python
>>> VLANGroup.objects.first()
<VLANGroup: AGE>
>>> VLANGroup.objects.create(name="AGE")
<VLANGroup: AGE>
>>> VLANGroup.objects.get_by_natural_key("AGE")
  File "<console>", line 1, in <module>
  File "/source/nautobot/core/models/__init__.py", line 52, in get_by_natural_key
    return self.get(**kwargs)
  File "/usr/local/lib/python3.7/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/cacheops/query.py", line 351, in get
    return qs._no_monkey.get(qs, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/django/db/models/query.py", line 442, in get
    num if not limit or num < limit else 'more than %s' % (limit - 1),
nautobot.ipam.models.VLANGroup.MultipleObjectsReturned: get() returned more than one VLANGroup -- it returned 2!
```

nautobot/core/models/__init__.py
glennmatthews: I'm exploring changing this to `&` in an upcoming commit, with the reasoning as follows:

```python
# For our purposes, NATURAL_KEY_SLUG_SEPARATOR needs to be:
# 1. Safe in a URL path component (so that we can do URLS like "/dcim/devices/<natural_key_slug>/delete/")
#    Per RFC3986 section 2.3 the general "unreserved" characters are ALPHA and DIGIT and the characters -._~
#    Per RFC3986 section 3.3 path components also permit characters :@ and the "sub-delims" characters !$&'()*+,;=
# 2. Not readily confused with characters commonly seen in un-escaped natural key component fields
#    "." is already ruled out as an unreserved character but also would appear in IPv4 IPAddress and Prefix objects
#    ":" similarly would appear in IPv6 IPAddress/Prefix objects
#    "/" would appear in Prefix objects as well as various numbered device component names
```

nautobot/extras/models/jobs.py
glennmatthews: Addressed by instead reworking the logic so that it doesn't automatically exclude UUIDField, only `id`.

glennmatthews:
OK! I managed to get rid of `django-natural-keys` as a dependency without adding too much more to our own code, and I at least personally find the reimplemented code easier to understand than `django-natural-keys`'s original implementation. Hope y'all agree with me on that!

Also I made a stab at adding/updating some developer documentation.

nautobot/core/models/__init__.py
gsnider2195: Maybe?

```suggestion
        while vals and vals[-1] is None:
            vals.pop()
        return vals
```

jathanism: You can't modify a list while iterating, but you could work with a copy of it.

nautobot/docs/development/extending-models.md
```diff
@@ -41,19 +41,17 @@ Add the name of the new field to `csv_headers` and included a CSV-friendly repre
 
 ## 4. Update relevant querysets
 
-<!-- v2 TODO(jathan): Replace prefetch_related with select_related -->
-
-If you're adding a relational field (e.g. `ForeignKey`) and intend to include the data when retrieving a list of objects, be sure to include the field using `prefetch_related()` as appropriate. This will optimize the view and avoid extraneous database queries.
+If you're adding a relational field (e.g. `ForeignKey`) and intend to include the data when retrieving a list of objects, be sure to include the field using `select_related()` and/or `prefetch_related()` as appropriate. This will optimize the view and avoid extraneous database queries.
```
jathanism, line 44: This is great improvement, but without cacheops it's not telling the entire story.

- `select_related` for `ForeignKey` forward lookups
- `prefetch_related` for `ManyToManyField` forward lookups, and reverse lookups of `ForeignKey` relationships.

nautobot/docs/development/natural-keys.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Implementing the Natural Key APIs
+
+In many model cases, Nautobot's default implementation of these APIs will suffice - as long as your model has any of the following, a default natural key will be automatically made available:
+
+- One or more `UniqueConstraint` declarations
+- Any `unique_together` declaration
+- Any field (other than `id`) that is set as `unique=True`.
+
+There are a few special cases that will need special handling.
```
jathanism, line 45:
```suggestion
There are a few special cases that will need special handling as described below.
```

nautobot/docs/development/natural-keys.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Self-Referential Natural Keys
+
+An example of this can be seen with the `Location` model, where a given instance is only uniquely identified by its name **in combination with its parent**, which is another `Location`. Nautobot's default implementation would fall into an infinite recursion when trying to identify the Location's natural key fields, since they would be identified as  `("name", "parent__name", "parent__parent__name", "parent__parent__parent__name", ...)`.
+
+In a case like this, Nautobot is able to support _variadic_ natural keys, where the number and listing of natural keys may vary depending on the data of a given instance. To make this work, you will need to override two natural key related APIs on your model, `natural_key_field_lookups` and `natural_key_args_to_kwargs`, as follows:
```
jathanism, line 51:
```suggestion
In a case like this, Nautobot is able to support _variadic_ natural keys, where the number and listing of natural keys may vary depending on the data of a given instance. To make this work, you will need to override two APIs related to natural keys on your model (`natural_key_field_lookups` and `natural_key_args_to_kwargs`) as follows:
```

nautobot/dcim/models/locations.py
```diff
@@ -251,6 +256,34 @@ class Meta:
...
+        name = "name"
+        for _ in range(cls.objects.max_tree_depth() + 1):
+            lookups.append(name)
+            name = f"parent__{name}"
+        return lookups
+
+    @classmethod
+    def natural_key_args_to_kwargs(cls, args):
+        """Handle the possibility that more recursive "parent" lookups were specified than we initially expected."""
+        args = list(args)
```
jathanism, line 279:
```suggestion
    def natural_key_args_to_kwargs(cls, *args):
        """Handle the possibility that more recursive "parent" lookups were specified than we initially expected."""
```

nautobot/docs/development/natural-keys.md
```diff
@@ -0,0 +1,135 @@
...
+        name = "name"
+        for _ in range(cls.objects.max_tree_depth() + 1):
+            lookups.append(name)
+            name = f"parent__{name}"
+        return lookups
+
+    @classmethod
+    def natural_key_args_to_kwargs(cls, args):
+        """Handle the possibility that more recursive "parent" lookups were specified than we initially expected."""
+        args = list(args)
```
jathanism, line 79:
```suggestion
    def natural_key_args_to_kwargs(cls, *args):
        """Handle the possibility that more recursive "parent" lookups were specified than we initially expected."""
```

nautobot/docs/development/natural-keys.md
```diff
@@ -0,0 +1,135 @@
...
+        while len(args) < len(natural_key_field_lookups):
+            args.append(None)
+        while len(args) > len(natural_key_field_lookups):
+            natural_key_field_lookups.append(f"parent__{natural_key_field_lookups[-1]}")
+        return dict(zip(natural_key_field_lookups, args))
+```
+
+### Natural Keys Referencing a Different Self-Referential Model
+
+Relatedly, if you have a model whose natural keys include a ForeignKey to a model like `Location` with self-referential and variadic natural keys, for the related model to be handled properly, you must always ensure that the related field is the **last** such field in your model's uniqueness constraint or `natural_key_field_names` declaration. Instead of:
```
jathanism, line 90:
```suggestion
Similarly, if you have a model whose natural keys include a `ForeignKey` to a model like `Location` with self-referential and variadic natural keys, for the related model to be handled properly, you must always ensure that the related field is the **last** such field in your model's uniqueness constraint or `natural_key_field_names` declaration.

Instead of this:
```

nautobot/docs/development/natural-keys.md
```diff
@@ -0,0 +1,135 @@
...
+
+```python
+class VLANGroup(BaseModel):
+    class Meta:
+        unique_together = [["location", "name"]]   # wrong, nested location natural key cannot be variadic
+```
+
+do:
```
jathanism, line 98:
```suggestion
Do this:
```

nautobot/docs/release-notes/version-2.0.md
```diff
@@ -18,6 +18,26 @@ DeviceRole, RackRole, IPAM Role, and IPAddressRoleChoices have all been merged i
...
+<DeviceType: Arista DeviceType 15>
+
+>>> DeviceType.objects.first().natural_key_slug
+'Arista&Arista+DeviceType+15'
+
+>>> DeviceType.objects.get(natural_key_slug='Arista&Arista+DeviceType+15')
+<DeviceType: Arista DeviceType 15>
+```
+
+Developers can refer to the [documentation](../development/natural-keys.md) for details on how to support and use this feature.
```
jathanism, line 39:
```suggestion
Developers can refer to the [documentation on natural keys](../development/natural-keys.md) for details on how to support and use this feature.
```

nautobot/docs/development/natural-keys.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+Similarly, the `natural_key_slug` and `get(natural_key_slug=...`) APIs are also symmetric:
+
+```python
+# Note that natural_key_slug is a property, not a method!
+>>> DeviceType.objects.first().natural_key_slug
+'Arista&Arista+DeviceType+15'
+
+>>> DeviceType.objects.get(natural_key_slug='Arista&Arista+DeviceType+15')
```
jathanism, line 30:
```suggestion
>>> DeviceType.objects.get(natural_key_slug="Arista&Arista+DeviceType+15")
```

nautobot/docs/development/natural-keys.md
```diff
@@ -0,0 +1,135 @@
...
+<DeviceType: Arista DeviceType 15>
+```
+
+!!! note
+    The `natural_key_slug` is designed to be suitable for use in URL patterns, such as an object detail endpoint `/app/model/<natural_key_slug>/`, though some caution must be taken when defining URL patterns containing such a slug, as it **can** contain the character `/` as a component.
+
+## Implementing the Natural Key APIs
+
+In many model cases, Nautobot's default implementation of these APIs will suffice - as long as your model has any of the following, a default natural key will be automatically made available:
```
jathanism, line 39:
```suggestion
In many model cases, Nautobot's default implementation of these APIs will suffice. As long as your model has any of the following, a default natural key will be automatically made available:
```

nautobot/extras/models/tags.py
```diff
@@ -86,3 +80,8 @@ class TaggedItem(BaseModel, GenericUUIDTaggedItemBase):
 
     class Meta:
         index_together = ("content_type", "object_id")
+        # TODO: unique_together = [["content_type", "object_id", "tag"]]
```
jathanism, line 83: Not in this PR, but IMO we should do this as a part of v2 seeing as this [exact constraint is implemented in the core `TaggedItem` from `django-taggit`](https://github.com/jazzband/django-taggit/blob/master/taggit/models.py#L191).

jathanism (changes requested):

nautobot/core/models/utils.py
```diff
@@ -170,3 +173,31 @@ def find_models_with_matching_fields(app_models, field_names, field_attributes=N
...
+    """
+    Convert the given list of natural key values to a single URL-path-usable string.
+
+    - Non-URL-safe characters are percent-encoded.
+    - Null (`None`) values are percent-encoded as a literal null character `%00`.
+
+    Reversible by `deconstruct_natural_key_slug()`.
+    """
+    values = [str(value) if value is not None else "\0" for value in values]
+    values = NATURAL_KEY_SLUG_SEPARATOR.join(quote_plus(value, safe="/.:") for value in values)
```
glennmatthews, line 188: Consider **not** flagging `/` as a safe character (requiring it to be encoded/quoted instead) just to make it easier to use natural-key-slugs in URL patterns.

nautobot/docs/development/natural-keys.md
glennmatthews: The reason for the explicit cast to `list()` is because `args` may be an immutable tuple, so without casting it to a list we cannot append to it.

I can change the signature from `(args)` to `(*args)` if you think that would be preferable, but in practice I found that most of the time I already have a list or tuple of args to pass in, rather than distinct individual parameters, so I'd have to explode that list/tuple when calling the function, i.e. instead of (current pattern):

```python
args = ["hello", "world"]
kwargs = model.natural_key_args_to_kwargs(args)
```

I'd have to do:

```python
args = ["hello", "world"]
kwargs = model.natural_key_args_to_kwargs(*args)
```

which just felt clunky to me.

nautobot/docs/development/extending-models.md
glennmatthews:
```suggestion
If you're adding a relational field (e.g. `ForeignKey`) and intend to include the data when retrieving a list of objects, be sure to include the field using `select_related()` (for `ForeignKey` forward lookups) and/or `prefetch_related()` (for `ForeignKey` reverse lookups and `ManyToManyField` lookups) as appropriate. This will optimize the view and avoid extraneous database queries.
```

nautobot/docs/development/natural-keys.md
```diff
@@ -0,0 +1,135 @@
...
+# Note that natural_key_slug is a property, not a method!
+>>> DeviceType.objects.first().natural_key_slug
+'Arista&Arista+DeviceType+15'
+
+>>> DeviceType.objects.get(natural_key_slug='Arista&Arista+DeviceType+15')
+<DeviceType: Arista DeviceType 15>
+```
+
+!!! note
+    The `natural_key_slug` is designed to be suitable for use in URL patterns, such as an object detail endpoint `/app/model/<natural_key_slug>/`, though some caution must be taken when defining URL patterns containing such a slug, as it **can** contain the character `/` as a component.
```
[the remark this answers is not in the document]
glennmatthews: Holding on this comment until we reach a conclusion about inclusion or escaping of `/` characters (https://github.com/nautobot/nautobot/pull/3423/files#r1139300825)

nautobot/extras/models/tags.py
glennmatthews: Opened #3437.

nautobot/docs/development/extending-models.md
jathanism: :shipit:

nautobot/docs/development/natural-keys.md
jathanism: Ahh, I see. I also realize I was incorrect in my thought that the unpacked list would be passed through as such, and it is not. It comes through as a `tuple`.

nautobot/core/models/utils.py
glennmatthews: We decided to quote/encode/escape the `/` character for now.

nautobot/core/models/constants.py
```diff
@@ -0,0 +1,11 @@
...
+
+# For our purposes, NATURAL_KEY_SLUG_SEPARATOR needs to be:
+# 1. Safe in a URL path component (so that we can do URLS like "/dcim/devices/<natural_key_slug>/delete/")
+#    Per RFC3986 section 2.3 the general "unreserved" characters are ALPHA and DIGIT and the characters -._~
+#    Per RFC3986 section 3.3 path components also permit characters :@ and the "sub-delims" characters !$&'()*+,;=
+# 2. Not readily confused with characters commonly seen in un-escaped natural key component fields
+#    "." is already ruled out as an unreserved character but also would appear in IPv4 IPAddress and Prefix objects
+#    ":" similarly would appear in IPv6 IPAddress/Prefix objects
+#    "/" would appear in Prefix objects as well as various numbered device component names
+NATURAL_KEY_SLUG_SEPARATOR = "&"
```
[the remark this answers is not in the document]
glennmatthews: In the current encoding `+` is being used as a replacement for ` ` (space) in the individual natural key components. If we want to use `+` as the separator then we'd have to encode ` ` as `%20` in the individual components.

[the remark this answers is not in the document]
glennmatthews: What's your use case for natural slugs - do you anticipate `&` to be problematic in that use case, or what's driving the question?

jathanism (approved):
:shipit:

nautobot/extras/tests/test_models.py
```diff
@@ -309,7 +310,7 @@ def test_multiple_tags_return_distinct_objects(self):
         device.tags.add(self.tag2)
 
         annotated_queryset = Device.objects.filter(name=device.name).annotate_config_context_data()
-        self.assertEqual(ConfigContext.objects.get_for_object(device).count(), 1)
+        self.assertEqual(ConfigContext.objects.get_for_object(device).count(), 2)  # including one from setUp()
```
HanlinMiao, line 313: Non-blocking: can we use `self.assertQuerysetEqualAndNotEmpty` here?

nautobot/ipam/tests/test_models.py
```diff
@@ -159,7 +160,7 @@ def test_get_db_prep_value_mysql(self):
         self.assertEqual(prepped, manual)
 
 
-class TestPrefix(TestCase):
+class TestPrefix(TestCase):  # TODO change to BaseModelTestCase
```
HanlinMiao, line 163: Sorry but whats the reason here for not changing it to `BaseModelTestCase`? I think we do have a `PrefixFactory`.

HanlinMiao (approved):
LGTM besides some nitpicks. Really impressive work!!🚀

nautobot/ipam/tests/test_models.py
glennmatthews: See line 165/166 below -- this test case currently does a `Prefix.objects.all().delete()` as part of its `setUp()` and then only creates individual Prefix instances in specific test steps -- so the generic test wouldn't be able to use the factory Prefixes, unfortunately.

nautobot/ipam/models.py
```diff
@@ -650,6 +660,16 @@ def get_absolute_url(self):
     def get_duplicates(self):
         return IPAddress.objects.filter(vrf=self.vrf, host=self.host).exclude(pk=self.pk)
 
+    # TODO: The current IPAddress model has no appropriate natural key available yet.
+    #       However, by default all BaseModel subclasses now have a `natural_key` property;
+    #       but for this model, accessing the natural_key will raise an exception.
+    #       The below is a hacky way to "remove" the natural_key property from this model class for the time being.
+    class AttributeRemover:
```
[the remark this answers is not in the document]
glennmatthews: I'd prefer to keep it inline just so that it's clear this is a short-term hack - putting it into `core.utils` makes it more "official" and I'm worried that we might then have to support this antipattern long-term. :-)

nautobot/extras/tests/test_models.py
glennmatthews: I'd be more tempted to remove this honestly. It's not a filter test so there's not really a different queryset to compare against that I can see.

Issue #3312, opened 2023-02-14
Custom fields not created on objects if Django ORM methods are used


phoenixswiss:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.10
* Python version: 3.8.10
* Database platform, version: postgresql, Ubuntu 12.12-0ubuntu0.20.04.1
* Middleware(s): -

### Steps to Reproduce
1. Create a custom field of any type with ("") or without (`null`) as default value and assign it to the object type interface.
2. Create a new job and in that job create a new interface for an existing device using Django ORM methods:

```python
nautobot_ip_interface = Interface(
    device=device,
    name=ip_interface_name,
    type=InterfaceTypeChoices.TYPE_VIRTUAL,
    status=status_active,
    label=ip_interface_name,
    mgmt_only=mgmt_only,
    enabled=True,
    untagged_vlan=untagged_vlan,
    mode="access",
)
nautobot_ip_interface.validated_save()
```

3. List the custom field keys by calling `nautobot_ip_interface.custom_field_data.keys()` or navigate to the Device Change Log site in Nautobot and open the `interface created` entry for the interface, which you have created in step 2. Under Difference note the value `Null` and the new value (JSON) with the property 

```
 "custom_fields": { 
       "your_custom_field": null,
  }
```

4. On the right side of this change log page under `Object Data` the custom field section `your_custom_field` is missing.

### Expected Behavior
All custom fields should be created with default values or null on an object if Django ORM methods are used.

### Observed Behavior
None custom fields are created.

If you add the custom field data manually before saving, the field will be created:

```python
nautobot_ip_interface.custom_field_data["your_custom_field"] = ""
nautobot_ip_interface.validated_save()
```

If pynautobot (REST-API) is used, the custom fields will be created during object creation.

This issue has been discussed here: https://github.com/nautobot/nautobot/discussions/3298#discussioncomment-4960531

bryanculver:
@phoenixswiss Do you have a celery worker running? And if so after you save the object can you run a `refresh_from_db` after waiting a minute after the object save?

phoenixswiss:
@bryanculver Yes, I have celery worker running. Saving the object and waiting 60 seconds and then refreshing the object from db had the same effect.

bryanculver:
> @bryanculver Yes, I have celery worker running. Saving the object and waiting 60 seconds and then refreshing the object from db had the same effect.

Have you tried this outside a job as well?

[...]

PR #3426, opened 2023-03-15
Remove `name` field from CustomField and rename `slug` to `key`

HanlinMiao:
# Closes: #824
# What's Changed

1. Rename `CustomField` `slug` field to `key`
2. Make sure `CustomField` `key` is unique during migration
3. Migrate non-empty `name` value to `label` for `CustomField`
4. Change `CustomField`'s Models' `_custom_field_data.keys()` to use `cf.key` instead of `cf.name`
5. Remove `name` field from `CustomField`

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/migrations/0070_make_cf_name_field_nullable.py
```diff
@@ -0,0 +1,27 @@
...
+        ),
+        migrations.AlterField(
+            model_name='customfield',
+            name='label',
+            field=models.CharField(max_length=50),
+        ),
+        migrations.AlterField(
+            model_name='customfield',
+            name='name',
+            field=models.CharField(blank=True, max_length=50, unique=True),
```
HanlinMiao, line 25: This `blank=True` is to make sure the migration is reversible. aka in data migration tests.

jvanderaa:
Is this keeping label rather than name on Custom Fields? Should it remove label instead as suggested by @smk4664 on #824?

glennmatthews:
The reasoning for keeping `label` is that `name` is ambiguous in its meaning in this context, and IMHO is the root cause of the long-standing issues we've had with this feature. `label` and `slug` are clearer that one is explicitly human-facing and the other is explicitly code-facing.

nautobot/extras/templates/extras/customfield.html
```diff
@@ -34,10 +33,6 @@
                         <td>Weight</td>
                         <td><span>{{ object.weight }}</span></td>
                     </tr>
-                    <tr>
-                        <td>Label</td>
```
glennmatthews, line 38: Looks like `slug` isn't currently on this page? Given its relevance for custom fields it might be good to add even if it's a bit redundant with the `advanced` tab.

nautobot/extras/templates/extras/customfield.html
```diff
@@ -22,9 +22,8 @@
                 </div>
                 <table class="table table-hover panel-body attr-table">
                     <tr>
-                        {% comment %}2.0 TODO: #824 remove name field{% endcomment %}
-                        <td>Name</td>
-                        <td><span>{{ object.name }}</span></td>
+                        <td>Label</td>
+                        <td><span>{{ object.label | placeholder }}</span></td>
```
glennmatthews, line 26: Shouldn't need a `placeholder` here, label should always be set now.

nautobot/extras/tests/integration/test_customfields.py
```diff
@@ -206,30 +206,29 @@ def test_custom_field_advanced_ui(self):
         custom_field.save()
         device_content_type = ContentType.objects.get_for_model(Device)
         custom_field.content_types.set([device_content_type])
-        # 2.0 TODO: #824 replace custom_field.name with custom_field.slug
-        device.cf[custom_field.name] = "This is some testing text"
+        device.cf[custom_field.slug] = "this-is-some-testing-text"
```
glennmatthews, line 209: Why the text change? The old value should still be permissible.

nautobot/extras/tests/test_changelog.py
```diff
@@ -23,17 +23,16 @@ class ChangeLogViewTest(ModelViewTestCase):
...
-        cf = CustomField(type=CustomFieldTypeChoices.TYPE_TEXT, name="my_field", required=False)
+        cf = CustomField(type=CustomFieldTypeChoices.TYPE_TEXT, label="my_field", required=False)
         cf.validated_save()
         cf.content_types.set([ct])
 
         # Create a select custom field on the Location model
         cf_select = CustomField(
             type=CustomFieldTypeChoices.TYPE_SELECT,
-            name="my_field_select",
+            label="my_field_select",
```
glennmatthews, line 35: Nit: let's use human-readable labels for these, i.e. `My Text Field` and `My Select Field`

nautobot/extras/tests/test_customfields.py
```diff
@@ -116,36 +111,34 @@ def test_simple_fields(self):
...
-            cf = CustomField(type=data["field_type"], name="my_field", required=False)
+            cf = CustomField(type=data["field_type"], label="my_field", required=False)
             cf.save()  # not validated_save this time, as we're testing backwards-compatibility
             cf.content_types.set([obj_type])
             # Assert that slug and label were auto-populated correctly
             # 2.0 TODO: slug and label will become mandatory fields to specify.
             cf.refresh_from_db()
-            self.assertEqual(cf.label, cf.name)
-            self.assertEqual(cf.slug, cf.name)
+            self.assertEqual(cf.slug, cf.label)
```
glennmatthews, line 122: Probably should change this test to use a human-readable `label` and assert that `cf.slug` equals `slugify_dashes_to_underscores(cf.label)`?

nautobot/extras/tests/test_customfields.py
```diff
@@ -116,36 +111,34 @@ def test_simple_fields(self):
         obj_type = ContentType.objects.get_for_model(Location)
 
         for data in DATA:
-
             # Create a custom field
             # 2.0 TODO: #824 slug rather than name
```
glennmatthews, line 115: Can remove the various `2.0 TODO` comments in this file.

nautobot/extras/tests/test_customfields.py
```diff
@@ -194,7 +187,7 @@ def test_multi_select_field(self):
         # Create a custom field
         cf = CustomField(
             type=CustomFieldTypeChoices.TYPE_MULTISELECT,
-            name="my_field",
+            label="my_field",
```
glennmatthews, line 190: Human-readable `labels` throughout please.

nautobot/extras/tests/test_dynamicgroups.py
```diff
@@ -887,7 +886,7 @@ def test_filter_custom_fields(self):
         device = self.devices[0]
 
         cf = CustomField.objects.create(
-            name="favorite_food",
+            label="favorite_food",
```
glennmatthews, line 889:
```suggestion
            label="Favorite Food",
```

nautobot/extras/tests/test_filters.py
```diff
@@ -365,15 +364,15 @@ class CustomFieldChoiceFilterSetTestCase(FilterTestCases.FilterTestCase):
 
     generic_filter_tests = (
         ["value"],
-        ["custom_field", "custom_field__name"],
+        ["custom_field", "custom_field__label"],
```
glennmatthews, line 367: Hm, might want to discuss this one. Should it be based around `slug` rather than `label`?

glennmatthews (commented):
Looks like a great start!

nautobot/dcim/tests/test_views.py
```diff
@@ -578,7 +574,6 @@ class DeviceTypeTestCase(
 
     @classmethod
     def setUpTestData(cls):
-
```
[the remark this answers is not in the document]
HanlinMiao: Oh yeah I'm a big stat padder

jvanderaa:
I really feel like this will cause confusion. Would much prefer to keep "The name of the custom field" versus "The label of the custom field", in conversation name would stick to it. Label is often referenced similarly to tag.

jvanderaa:
And most code bases is probably using the `name` key on the custom field rather than `label`. Which to me moving to label would cause more code rework than name.

bryanculver:
> I really feel like this will cause confusion. Would much prefer to keep "The name of the custom field" versus "The label of the custom field", in conversation name would stick to it. Label is often referenced similarly to tag.

> And most code bases is probably using the `name` key on the custom field rather than `label`. Which to me moving to label would cause more code rework than name.

The desire it to _remove_ ambiguity between slug/name/label.

Is `name` what you see in the UI or is it the name of the field seen in the API?

- `slug` is much more clearly something in the API/ORM
- `label` is much more clearly something in the UI

nautobot/extras/api/customfields.py
```diff
@@ -52,18 +52,10 @@ def _get_custom_fields(self):
         return self._custom_fields
 
     def to_representation(self, obj):
-        # 2.0 TODO: #824 use cf.slug as lookup key instead of cf.name
-        return {cf.slug: obj.get(cf.name) for cf in self._get_custom_fields()}
+        return {cf.slug: obj.get(cf.slug) for cf in self._get_custom_fields()}
```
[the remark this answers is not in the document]
HanlinMiao: The conversion is still needed, because `self._get_custom_fields()` returns a `RestrictedQueryset` object.

bryanculver:
> > I really feel like this will cause confusion. Would much prefer to keep "The name of the custom field" versus "The label of the custom field", in conversation name would stick to it. Label is often referenced similarly to tag.
> 
> > And most code bases is probably using the `name` key on the custom field rather than `label`. Which to me moving to label would cause more code rework than name.
> 
> The desire it to _remove_ ambiguity between slug/name/label.
> 
> Is `name` what you see in the UI or is it the name of the field seen in the API?
> 
> * `slug` is much more clearly something in the API/ORM
> * `label` is much more clearly something in the UI

@jvanderaa after MUCH deliberation we have decided to also rename `slug` to `key`.

nautobot/core/testing/views.py
```diff
@@ -113,6 +113,12 @@ def _get_url(self, action, instance=None):
                 return reverse(url_format.format(action), kwargs={"slug": instance.slug})
             except NoReverseMatch:
                 pass
+        # Attempt to resolve using key as the unique identifier if one exists
```
gsnider2195, line 116: Here and the other generic test in `nautobot/core/testing/api.py` it feels like there's a better approach than checking for attrs on the model but I don't know what that approach should be.

nautobot/core/testing/api.py
```diff
@@ -370,7 +370,10 @@ def check_expected_slug(self, obj):
                     expected_slug += self.slugify_function(val)
 
             self.assertNotEqual(expected_slug, "")
-            self.assertEqual(obj.slug, expected_slug)
+            if hasattr(obj, "slug"):
```
gsnider2195, line 373: It feels like there's a better approach than checking for attrs on the model but I don't know what that approach should be.

nautobot/core/views/generic.py
```diff
@@ -1041,11 +1024,10 @@ def post(self, request, **kwargs):
 
                             # Update custom fields
                             for field_name in form_custom_fields:
-                                # 2.0 TODO: #824 when we use slug in obj.cf we can just do obj.cf[field_name[3:]]
                                 if field_name in form.nullable_fields and field_name in nullified_fields:
-                                    obj.cf[form_cf_to_key[field_name]] = None
+                                    obj.cf[field_name[3:]] = None
```
jathanism, line 1028: The previous iteration was a little more readable since it was clear what `form_cf_to_key` was doing and this slicing isn't as obvious. Can you please add a brief comment above this `if/else` stating that `[3:]` is trimming `cf_`?

nautobot/core/views/mixins.py
```diff
@@ -985,7 +982,7 @@ def _process_bulk_update_form(self, form):
             if field not in form_custom_fields + form_relationships + ["pk"] + ["object_note"]
         ]
         nullified_fields = request.POST.getlist("_nullify")
-        form_cf_to_key = {f"cf_{cf.slug}": cf.name for cf in CustomField.objects.get_for_model(model)}
+        form_cf_to_key = {f"cf_{cf.key}": cf.name for cf in CustomField.objects.get_for_model(model)}
```
jathanism, line 985: Why did we decide to stick with the `form_to_cf_key` pattern here, but not the other place? IMO we should align them, and make the appropriate changes below on lines 1013 and 1015 to use slicing.

nautobot/core/testing/api.py
gsnider2195: With slugs going away this probably isn't worth spending time on

nautobot/ipam/tests/test_migrations.py
```diff
@@ -421,7 +421,8 @@ def test_aggregate_to_prefix_migration_object_changes(self):
         "mysql does not support rollbacks",
     )
     def test_aggregate_to_prefix_migration_custom_fields(self):
-        self.assertEqual(self.custom_field.objects.exclude(name="example_plugin_auto_custom_field").count(), 3)
+        # This change is necessary because name attribute is not specified now in example_plugin's signal.py
+        self.assertEqual(self.custom_field.objects.exclude(name__exact="").count(), 3)
```
jathanism, line 425: Not a huge deal but `name=` and `name__exact=` are the same thing.

```suggestion
        self.assertEqual(self.custom_field.objects.exclude(name="").count(), 3)
```

jathanism (changes requested):
Minor nitpicks, but nitpicks all the same.

nautobot/core/views/mixins.py
gsnider2195, line 985: This should have broken some tests?
```suggestion
        form_cf_to_key = {f"cf_{cf.key}": cf.key for cf in CustomField.objects.get_for_model(model)}
```

gsnider2195: Or should it be 
```suggestion
        form_cf_to_key = {f"cf_{cf.key}": cf.label for cf in CustomField.objects.get_for_model(model)}
```

nautobot/dcim/tests/test_views.py
```diff
@@ -1204,7 +1196,7 @@ def setUpTestData(cls):
         status_active = statuses.get(slug="active")
 
         cls.custom_fields = (
-            CustomField.objects.create(type=CustomFieldTypeChoices.TYPE_INTEGER, name="crash-counter", default=0),
+            CustomField.objects.create(type=CustomFieldTypeChoices.TYPE_INTEGER, label="Crash Counter", default=0),
```
gsnider2195, line 1199: Update the Device creates below to change `crash-counter` to `crash_counter`?

```py
                _custom_field_data={"crash-counter": 5},
```

nautobot/extras/api/customfields.py
```diff
@@ -52,18 +52,10 @@ def _get_custom_fields(self):
         return self._custom_fields
 
     def to_representation(self, obj):
-        # 2.0 TODO: #824 use cf.slug as lookup key instead of cf.name
-        return {cf.slug: obj.get(cf.name) for cf in self._get_custom_fields()}
+        return {cf.key: obj.get(cf.key) for cf in self._get_custom_fields()}
```
gsnider2195, line 55: Does this need to be?
```suggestion
        return {cf.key: obj.get(cf.label) for cf in self._get_custom_fields()}
```

nautobot/extras/migrations/0070_make_cf_name_field_nullable.py
gsnider2195: Would it be better to let the reverse migration fail than to leave the database `name` fields empty and silently let your custom field records be in a broken state?

nautobot/extras/models/customfields.py
```diff
@@ -398,29 +392,19 @@ class CustomField(BaseModel, ChangeLoggedModel, NotesMixin):
     ]
 
     class Meta:
-        ordering = ["weight", "name"]
+        ordering = ["weight", "label"]
```
gsnider2195, line 395: Should this be `["weight", "key"]` instead?

HanlinMiao: I feel like it would make more sense to use `label` here since the __str__() method is returning `self.label` but I could go either way do not have a super strong opinion here.

nautobot/extras/models/customfields.py
```diff
@@ -398,29 +392,19 @@ class CustomField(BaseModel, ChangeLoggedModel, NotesMixin):
...
-        # in 2.0 we should make `label` a mandatory field when getting rid of `name`.
-        if self.name and not self.label:
-            self.label = self.name
-
-        # This is to fix up existing ORM usage when caller doesn't specify a slug since it wasn't a field before.
-        if not self.slug:
-            self.slug = slugify_dashes_to_underscores(self.label or self.name)
+        # This is to fix up existing ORM usage when caller doesn't specify a key since it wasn't a field before.
+        if not self.key:
+            self.key = slugify_dashes_to_underscores(self.label)
```
gsnider2195, line 407: Is this/was this required with `key` being an `AutoSlugField`? I commented these lines out and it seemed to work fine.

```py
>>> cf = CustomField(label="this is my custom field")
>>> cf.save()
>>> cf.refresh_from_db()
>>> cf.key
'this_is_my_custom_field'
```

nautobot/extras/tables.py
```diff
@@ -207,21 +207,9 @@ def render(self, record):
 class CustomFieldTable(BaseTable):
     pk = ToggleColumn()
     label = tables.Column(linkify=True)
-    # 2.0 TODO: #824 Remove name column
-    name = tables.TemplateColumn(
+    key = tables.TemplateColumn(
```
gsnider2195, line 210: This doesn't need to be a `TemplateColumn` anymore does it?

nautobot/extras/views.py
```diff
@@ -481,8 +480,7 @@ def construct_custom_field_delete_tasks(self, queryset):
         Helper method to construct a list of celery tasks to execute when bulk deleting custom fields.
         """
         tasks = [
-            # 2.0 TODO: #824 use obj.slug instead of obj.name
-            delete_custom_field_data.si(obj.name, set(obj.content_types.values_list("pk", flat=True)))
+            delete_custom_field_data.si(obj.key, set(obj.content_types.values_list("pk", flat=True)))
```
gsnider2195, line 483: Since this shorthand isn't used anywhere else in the code

```suggestion
            delete_custom_field_data.signature(obj.key, set(obj.content_types.values_list("pk", flat=True)))
```

nautobot/extras/models/customfields.py
gsnider2195: I was thinking it would match the sort order on the object's `_custom_field_data`. Probably not a big deal either way

gsnider2195 (changes requested):
Would like to see if `_fixup_empty_fields` can go away entirely with these changes

nautobot/extras/tables.py
```diff
@@ -207,11 +207,7 @@ def render(self, record):
 class CustomFieldTable(BaseTable):
     pk = ToggleColumn()
     label = tables.Column(linkify=True)
-    key = tables.TemplateColumn(
-        template_code="""
-{{ value }}
-"""
-    )
+    key = tables.Column()
```
gsnider2195, line 210: Does it still work if you remove this line?

nautobot/extras/models/customfields.py
HanlinMiao: I guess this https://github.com/nautobot/nautobot/actions/runs/4443594213/jobs/7801045894?pr=3426 is the sequence of code to execute if you want that error to pop out.

nautobot/extras/api/customfields.py
HanlinMiao: Nah actually the `obj.cf` has `cf.key` as its key as well

changes/824.changed
```diff
@@ -0,0 +1 @@
+Renamed `slug` field to `key` field on CustomField Model Class.
```
glennmatthews, line 1:
```suggestion
Renamed `slug` field to `key` on CustomField model class.
```

nautobot/core/graphql/generators.py
```diff
@@ -92,17 +92,16 @@ def resolve_filter(self, *args, **kwargs):
...
 
 
-# 2.0 TODO: #824 rename `name` to `slug`
-def generate_custom_field_resolver(name, resolver_name):
+def generate_custom_field_resolver(key, resolver_name):
     """Generate function to resolve each custom field within each DjangoObjectType.
 
     Args:
-        name (str): name of the custom field to resolve
+        key (str): name of the custom field to resolve
```
glennmatthews, line 99:
```suggestion
        key (str): unique key of the custom field to resolve
```

nautobot/extras/models/customfields.py
```diff
@@ -315,18 +312,15 @@ class CustomField(BaseModel, ChangeLoggedModel, NotesMixin):
...
-    # 2.0 TODO: #824 remove `name` field as redundant, make `label` mandatory, populate `slug` from `label` field.
-    name = models.CharField(max_length=50, unique=True, help_text="Human-readable unique name of this field.")
     label = models.CharField(
         max_length=50,
-        blank=True,
-        help_text="Name of the field as displayed to users (if not provided, the field's name will be used.)",
+        help_text="Name of the field as displayed to users.",
     )
-    slug = AutoSlugField(
+    key = AutoSlugField(
```
glennmatthews, line 319: Wondering out loud: should we also rename `ComputedField.slug` and `Relationship.slug` for the same reasoning and self-consistency?

nautobot/core/graphql/schema.py
```diff
@@ -228,40 +228,38 @@ def extend_schema_type_custom_field(schema_type, model):
...
     for field in cfs:
-        # 2.0 TODO: #824 replace field.name with field.slug
-        field_name = f"{prefix}{str_to_var_name(field.name)}"
-        if str_to_var_name(field.name) != field.name:
-            # 2.0 TODO: str_to_var_name is lossy, it may cause different fields to map to the same field_name
+        field_key = f"{prefix}{str_to_var_name(field.key)}"
+        if str_to_var_name(field.key) != field.key:
+            # 2.0 TODO: str_to_var_name is lossy, it may cause different fields to map to the same field_key
             # In 2.0 we should simply omit fields whose names/slugs are invalid in GraphQL, instead of mapping them.
             warnings.warn(
```
glennmatthews, line 235: Still in favor of biting the bullet and making this change for 2.0 - just use `{prefix}{field.key}` and if that isn't a "valid" graphql identifier, throw it out instead of converting it. And again, same idea for computed-field and relationship, though I'm fine if we want to do that as a sequel PR.

nautobot/core/testing/api.py
glennmatthews: Just a note that this intersects with #3439 also making changes to this test. Probably good to keep in sync.

nautobot/core/testing/views.py
```diff
@@ -817,7 +825,7 @@ def test_queryset_to_csv(self):
             instance1_csv_data += instance1_cf_values
             # Since values in `data` are all in str; cast all values in instance1_csv_data to str
             instance1_csv_data = [str(val) for val in instance1_csv_data]
-            instance1_cf_headers = ["cf_" + str(cf.slug) for cf in instance1.get_custom_fields().keys()]
+            instance1_cf_headers = ["cf_" + str(cf.key) for cf in instance1.get_custom_fields().keys()]
```
glennmatthews, line 828: We really ought to have a standard helper function `cf -> "cf_{cf.key}"` and use it all over the place... - maybe an instance method or property on the CustomField class?

nautobot/core/views/generic.py
```diff
@@ -186,27 +186,25 @@ def queryset_to_csv(self):
...
 
         # Iterate through the queryset appending each object
         for obj in self.queryset:
             data = obj.to_csv()
 
-            # 2.0 TODO: #824 use custom_field_slug
-            for custom_field_name in custom_field_names:
-                data += (obj.cf.get(custom_field_name, ""),)
+            for custom_field_key in custom_field_keys:
+                data += (obj.cf.get(custom_field_key, ""),)
```
glennmatthews, line 207: Maybe?

```suggestion
                data.append(obj.cf.get(custom_field_key, ""))
```

nautobot/core/views/generic.py
```diff
@@ -379,7 +376,7 @@ def get_required_permission(self):
     def get_object(self, kwargs):
         """Retrieve an object based on `kwargs`."""
         # Look up an existing object by slug, PK, or name, if provided.
-        for field in ("slug", "pk", "name"):
+        for field in ("slug", "pk", "name", "key"):
             if field in kwargs:
                 return get_object_or_404(self.queryset, **{field: kwargs[field]})
```
glennmatthews, line 381: Longer term will be great to replace this with something based on `natural_key_slug`...

nautobot/core/views/generic.py
glennmatthews: Maybe we can add a classmethod on `CustomField` that's the inverse of the `cf -> "cf_{key}"` function?

nautobot/extras/management/commands/fix_custom_fields.py
```diff
@@ -17,17 +17,15 @@ def handle(self, *args, **kwargs):
             self.stdout.write(self.style.SUCCESS(f"Processing ContentType {content_type}"))
             model = content_type.model_class()
             custom_fields_for_content_type = content_type.custom_fields.all()
-            # 2.0 TODO: #824 use cf.slug rather than cf.name
-            custom_field_names_for_content_type = [cf.name for cf in custom_fields_for_content_type]
+            custom_field_names_for_content_type = [cf.key for cf in custom_fields_for_content_type]
```
glennmatthews, line 20: nit - `custom_field_keys_for_content_type`?

nautobot/extras/migrations/0069_move_cf_name_value_to_label.py
```diff
@@ -0,0 +1,28 @@
...
+
+def move_cf_name_value_to_label(apps, schema_editor):
+    CustomField = apps.get_model("extras", "customfield")
+    # Filtering on null or empty labels
+    custom_fields = CustomField.objects.filter(label__isnull=True).union(CustomField.objects.filter(label__exact=""))
+
+    for cf in custom_fields:
+        cf.label = cf.name
+        cf.save()
+
```
glennmatthews, line 14: Aren't we missing somewhere a data migration that will iterate over all CustomFieldModels and migrate `self._custom_field_data[cf.name]` to `self._custom_field_data[cf.key]` across the board?

nautobot/extras/models/customfields.py
glennmatthews: Good catch. That's built-in logic in django-extensions AutoSlugField IIRC, we might need to override it in our AutoSlugField subclass.

nautobot/extras/tests/test_customfields.py
```diff
@@ -509,7 +478,6 @@ def test_get_single_object_without_custom_field_data(self):
         self.assertEqual(response.data["name"], self.locations[0].name)
         # A model directly instantiated via the ORM does NOT automatically receive custom field default values.
         # This is arguably a bug.
```
glennmatthews, line 480: Can add a link to https://github.com/nautobot/nautobot/issues/3312 here specifically :)

nautobot/ipam/tests/test_migrations.py
```diff
@@ -421,7 +421,8 @@ def test_aggregate_to_prefix_migration_object_changes(self):
         "mysql does not support rollbacks",
     )
     def test_aggregate_to_prefix_migration_custom_fields(self):
-        self.assertEqual(self.custom_field.objects.exclude(name="example_plugin_auto_custom_field").count(), 3)
+        # This change is necessary because name attribute is not specified now in example_plugin's signal.py
+        self.assertEqual(self.custom_field.objects.exclude(name="").count(), 3)
```
glennmatthews, line 425: Seems weird to me - `name` is gone as a field right? Should we instead look for specific `label` values instead of excluding empty `name`s?

glennmatthews (changes requested):

nautobot/ipam/tests/test_migrations.py
HanlinMiao: The `extras` migration that deletes the `name` field has not been run yet after `(dcim, 0030)` migration is executed in this test. So CustomField still has the `name` attribute

nautobot/extras/models/customfields.py
HanlinMiao: We were talking about it yesterday, we def should. Gonna make an issue after standup.

nautobot/extras/models/customfields.py
```diff
@@ -481,7 +461,7 @@ def clean(self):
             )
 
     def save(self, *args, **kwargs):
-        # Prior to Nautobot 1.4, `slug` was a non-existent field, but now it's mandatory.
+        # Prior to Nautobot 1.4, `key` was a non-existent field, but now it's mandatory.
         # Protect against get_or_create() or other ORM usage where callers aren't calling clean() before saving.
         # Normally we'd just say "Don't do that!" but we know there are some cases of this in the wild.
         self._fixup_empty_fields()
```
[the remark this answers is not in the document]
HanlinMiao: It still is needed yeah, unfortunately, there is a lot of test data that are not calling `clean()`. https://github.com/nautobot/nautobot/actions/runs/4443594213/jobs/7801045894?pr=3426 This is the sequence of code to execute if you want that error that `_fixup_empty_fields` is preventing pop out.

jvanderaa:
I'm looking at it from ORM and from the API. In the ORM you can probably handle that easily with the get request. will need to look at it from the API, but I'm assuming `name` is what is used often today.

nautobot/extras/models/customfields.py
HanlinMiao: This behavior seems to go away after I restored `_fixup_empty_fields()`. Now this is the output
`DETAIL:  Key (key)=(this_is_my_custom_field) already exists.`

nautobot/extras/tests/test_filters.py
```diff
@@ -365,15 +364,15 @@ class CustomFieldChoiceFilterSetTestCase(FilterTestCases.FilterTestCase):
 
     generic_filter_tests = (
         ["value"],
-        ["custom_field", "custom_field__name"],
+        ["custom_field", "custom_field__key"],
```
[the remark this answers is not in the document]
HanlinMiao: I do not think you can here, this is the code for `CustomFieldChoiceFilterSet`:
```
class CustomFieldChoiceFilterSet(BaseFilterSet):
    q = SearchFilter(filter_predicates={"value": "icontains"})
    custom_field = NaturalKeyOrPKMultipleChoiceFilter(
        queryset=CustomField.objects.all(),
        to_field_name="key",
        label="Field (ID or Key)",
    )
```
The filter field `custom_field` is mapped to field `key` in this case is the `NaturalKey`

nautobot/core/views/generic.py
```diff
@@ -186,27 +187,25 @@ def queryset_to_csv(self):
...
         # Start with the column headers
         headers = self.queryset.model.csv_headers.copy()
 
         # Add custom field headers, if any
         if hasattr(self.queryset.model, "_custom_field_data"):
             for custom_field in CustomField.objects.get_for_model(self.queryset.model):
-                headers.append("cf_" + custom_field.slug)
-                # 2.0 TODO: #824 custom_field.slug
-                custom_field_names.append(custom_field.name)
+                headers.append("cf_" + custom_field.key)
```
glennmatthews, line 198:
```suggestion
                headers.append(custom_field.add_prefix_to_cf_key())
```

nautobot/core/tests/integration/test_filters.py
```diff
@@ -139,9 +139,9 @@ def test_input_field_gets_updated(self):
         """Assert that a filter input/select field on Dynamic Filter Form updates if same field is updated."""
         self.browser.visit(f'{self.live_server_url}{reverse("dcim:location_list")}')
 
-        text_field_name = "cf_" + self.cf_text_field_name
-        integer_field_name = "cf_" + self.cf_integer_field_name
-        select_field_name = "cf_" + self.cf_select_field_name
+        text_field_name = "cf_" + self.custom_fields[0].key
+        integer_field_name = "cf_" + self.custom_fields[1].key
+        select_field_name = "cf_" + self.custom_fields[2].key
```
glennmatthews, line 144: Should these be using `.add_prefix_to_cf_key()`?

nautobot/core/views/generic.py
```diff
@@ -1041,11 +1025,11 @@ def post(self, request, **kwargs):
 
                             # Update custom fields
                             for field_name in form_custom_fields:
-                                # 2.0 TODO: #824 when we use slug in obj.cf we can just do obj.cf[field_name[3:]]
                                 if field_name in form.nullable_fields and field_name in nullified_fields:
-                                    obj.cf[form_cf_to_key[field_name]] = None
+                                    # [3:] is truncating the "cf_" from the field_name
```
glennmatthews, line 1029: Comment no longer needed?

nautobot/core/views/mixins.py
```diff
@@ -1013,9 +1009,10 @@ def _process_bulk_update_form(self, form):
                 # Update custom fields
                 for field_name in form_custom_fields:
                     if field_name in form.nullable_fields and field_name in nullified_fields:
-                        obj.cf[form_cf_to_key[field_name]] = None
+                        # [3:] is truncating the "cf_" from the field_name
```
glennmatthews, line 1012:
```suggestion
```

nautobot/dcim/tests/test_views.py
```diff
@@ -1290,7 +1282,7 @@ def setUpTestData(cls):
             "comments": "A new device",
             "tags": [t.pk for t in Tag.objects.get_for_model(Device)],
             "local_config_context_data": None,
-            "cf_crash-counter": -1,
+            "cf_crash_counter": -1,
             "cr_router-id": None,
```
glennmatthews, line 1286: Why we need similar validation and enforcement for relationship ~slugs~ keys, exhibit 1 of N.

nautobot/extras/migrations/0072_migrate_custom_field_data.py
```diff
@@ -0,0 +1,62 @@
...
+        cf_key = original_cf_key
+        append_counter = 2
+        while cf_key in cf_keys:
+            cf_key_append = f"_{append_counter}"
+            max_key_length = 50 - len(cf_key_append)
+            cf_key = original_cf_key[:max_key_length] + cf_key_append
+            append_counter += 1
+        if cf_key != original_cf_key:
+            print(
+                f'  CustomField Class "{CustomField.key}" key "{original_cf_key}" is not unique, changing to "{cf_key}".'
```
glennmatthews, line 24: Is `{CustomField.key}` the correct thing to be printing here?

nautobot/extras/models/customfields.py
```diff
@@ -434,12 +419,8 @@ def clean(self):
             # Check immutable fields
```
glennmatthews, line 419: Still feel like this is the right time to add logic in `clean()` or `save()` to **enforce** that `key` must be a valid Python/GraphQL identifier (underscores rather than dashes or whitespace, no starting with a digit, etc.)

nautobot/extras/tests/test_filters.py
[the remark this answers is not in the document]
glennmatthews: Ah yeah, we'd have to extend `NaturalKeyOrPKMultipleChoiceFilter` to allow matches on multiple distinct natural keys. Darn.

nautobot/extras/models/customfields.py
```diff
@@ -315,18 +312,16 @@ class CustomField(BaseModel, ChangeLoggedModel, NotesMixin):
...
         help_text="The type of value(s) allowed for this field.",
     )
-    # 2.0 TODO: #824 remove `name` field as redundant, make `label` mandatory, populate `slug` from `label` field.
-    name = models.CharField(max_length=50, unique=True, help_text="Human-readable unique name of this field.")
     label = models.CharField(
         max_length=50,
-        blank=True,
-        help_text="Name of the field as displayed to users (if not provided, the field's name will be used.)",
+        help_text="Name of the field as displayed to users.",
+        blank=False,
```
glennmatthews, line 318: Should we think about making `label` be `unique=True` also? I know it'd be more impact in the data migration, not sure if worth the effort at this time.

nautobot/extras/migrations/0072_migrate_custom_field_data.py
HanlinMiao: Ur right should be `label`.

nautobot/extras/migrations/0072_migrate_custom_field_data.py
```diff
@@ -0,0 +1,62 @@
...
+from django.db import migrations
+from nautobot.core.models.fields import slugify_dashes_to_underscores
+
+
+def generate_unique_custom_field_slug_and_migrate_custom_field_data(apps, schema_editor):
+    CustomField = apps.get_model("extras", "customfield")
+    ContentType = apps.get_model("contenttypes", "ContentType")
+
+    # Make sure that the key is unique during migration
+    cf_keys = []
```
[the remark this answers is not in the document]
HanlinMiao: This is kinda like a record for `cf.key` that we have already encountered in the for loop. So it should be empty at first so that as we go through the `CustomField` instances if we have duplicate `keys` we append indices at the end of the `key`. If we fill it out first, as we go through the instances, every `key` would be changed.

nautobot/extras/models/customfields.py
HanlinMiao: IMHO I think since `label` is user-facing and almost for display purposes only. (plus it can be modified anyway), it is probably not worth the effort to make it unique. But I could be swayed either way.

HanlinMiao: Filter field label for CustomField models does derive from `cf.label`, so we would have duplicate filter field label if we allow duplicate `cf.label`. That again is a display issue.
Screen Shot 2023-03-20 at 3 08 51 PM
The filter tags are constructed by` cf.key` tho
Screen Shot 2023-03-20 at 3 09 07 PM

nautobot/extras/migrations/0072_migrate_custom_field_data.py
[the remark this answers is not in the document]
glennmatthews: I guess maybe I was thinking of something like:

1. key = "Custom Field"
2. key = "custom-field"
3. key = "custom_field_2"

On processing, as currently implemented, 1 becomes `custom_field`, 2 becomes `custom_field_2`, and 3 gets renamed to `custom_field_2_2`, which doesn't seem like the best experience? Admittedly this is a contrived corner case.

nautobot/extras/migrations/0072_migrate_custom_field_data.py
```diff
@@ -12,20 +15,18 @@ def generate_unique_custom_field_slug_and_migrate_custom_field_data(apps, schema
...
             cf_key_append = f"_{append_counter}"
             max_key_length = 50 - len(cf_key_append)
             cf_key = original_cf_key[:max_key_length] + cf_key_append
             append_counter += 1
         if cf_key != original_cf_key:
-            print(
-                f'  CustomField Class "{CustomField.key}" key "{original_cf_key}" is not unique, changing to "{cf_key}".'
-            )
-            custom_field.key = slugify_dashes_to_underscores(cf_key)
+            print(f'  CustomField Class "{CustomField.label}" key "{original_cf_key}" is being changed to "{cf_key}".')
```
glennmatthews: should this be `custom_field.label`?

nautobot/extras/utils.py
```diff
@@ -496,6 +497,19 @@ def remove_prefix_from_cf_key(field_name):
...
+    """
+    Helper method to check if a key field is Python/GraphQL safe.
+    Used in CustomField for now, should be used in ComputedField and Relationship as well.
+    """
+    if key[0].isdigit():
+        raise ValidationError({"key": "The first letter of the CustomField key cannot be a number"})
+    if " " in key:
+        raise ValidationError({"key": 'There cannot be whitespaces " " in CustomField key'})
+    if "-" in key:
+        raise ValidationError({"key": 'There cannot be hypens "-" in CustomField key'})
```
glennmatthews, line 510: Probably should be more stringent still - per https://graphql-rules.com/rules/naming, the key should match the regexp `[_A-Za-z][_0-9A-Za-z]*`

HanlinMiao: The latest implementation outputs the following `ValidationError` when there is a whitespace in the `key`:
```
>>> cf = CustomField.objects.create(label="cf 1", key="123 2")
>>> cf.validated_save()
  File "<console>", line 1, in <module>
  File "/source/nautobot/core/models/__init__.py", line 51, in validated_save
    self.full_clean()
  File "/usr/local/lib/python3.10/site-packages/django/db/models/base.py", line 1251, in full_clean
    raise ValidationError(errors)
django.core.exceptions.ValidationError: {'key': ['Enter a valid “slug” consisting of letters, numbers, underscores or hyphens.'], "CustomField's specific key attribute": ['This key is not Python/GraphQL safe. Please do not start the key with a digit and do not use hypens or whitespaces']}
>>> 
```

It outputs two exception messages. The first one is emitted from Django's `SlugField` and the second one is our custom message. As you can see, these two messages are contradictory regarding hyphen usage. Wonder if is there an easy way to resolve this.

nautobot/extras/migrations/0072_migrate_custom_field_data.py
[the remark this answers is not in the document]
HanlinMiao: I see. The corner case definitely is something to consider. 
But we do have the `print` statement (mapping the old_key to the new_key) so that the user can easily find the `CustomField` with the weird new `key` and modify them by hand. and the corner case might not be as prevalent as we think.

What do you think? I am definitely open to suggestions on how to make this processing algorithm more robust.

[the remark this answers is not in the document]
glennmatthews: Remember that `key` is not user editable for existing custom fields, but this is probably good enough for now.

nautobot/core/testing/views.py
glennmatthews, line 116: Still needed now that we're using UUID:pk for custom-field URLs?

nautobot/core/views/generic.py
glennmatthews: bump :-)

nautobot/core/views/generic.py
```diff
@@ -379,7 +377,7 @@ def get_required_permission(self):
     def get_object(self, kwargs):
         """Retrieve an object based on `kwargs`."""
         # Look up an existing object by slug, PK, or name, if provided.
-        for field in ("slug", "pk", "name"):
+        for field in ("slug", "pk", "name", "key"):
```
glennmatthews, line 380: Here too, since we're using uuid for custom-field URLs, is this change needed?

nautobot/extras/tests/test_customfields.py
```diff
@@ -960,32 +924,31 @@ class CustomFieldImportTest(TestCase):
...
-            CustomField(name="text", type=CustomFieldTypeChoices.TYPE_TEXT),
-            CustomField(name="integer", type=CustomFieldTypeChoices.TYPE_INTEGER),
-            CustomField(name="boolean", type=CustomFieldTypeChoices.TYPE_BOOLEAN),
-            CustomField(name="date", type=CustomFieldTypeChoices.TYPE_DATE),
-            CustomField(name="url", type=CustomFieldTypeChoices.TYPE_URL),
+            CustomField(label="Text", type=CustomFieldTypeChoices.TYPE_TEXT),
+            CustomField(label="Integer", type=CustomFieldTypeChoices.TYPE_INTEGER),
+            CustomField(label="Boolean", type=CustomFieldTypeChoices.TYPE_BOOLEAN),
+            CustomField(label="Date", type=CustomFieldTypeChoices.TYPE_DATE),
+            CustomField(label="Url", type=CustomFieldTypeChoices.TYPE_URL),
```
glennmatthews, line 932: :-)

```suggestion
            CustomField(label="URL", type=CustomFieldTypeChoices.TYPE_URL),
```

nautobot/extras/models/customfields.py
```diff
@@ -430,16 +415,13 @@ def clean_fields(self, exclude=None):
     def clean(self):
         super().clean()
 
+        check_if_key_is_graphql_safe(self.__class__.__name__, self.key)
```
glennmatthews, line 418: Do we have any tests exercising this?

nautobot/extras/models/customfields.py
gsnider2195: I still think this can go away. Can you try setting `separator="_"` on the `AutoSlugField`?

HanlinMiao:
```
cf = CustomField(type=CustomFieldTypeChoices.TYPE_TEXT, label="My Field", required=False)
cf.validated_save()
```
It did fix the trailing index problem, but when I did the above operation in the ORM, it failed without `_fixup_empty_fields()`.

```
django.core.exceptions.ValidationError: {'key': ['This field cannot be blank.'], "CustomField's specific key attribute": ['This key is not Python/GraphQL safe. Please do not start the key with a digit and do not use hyphens or whitespace']}
```

gsnider2195: That's because `blank=False` is set on the field. This is probably the reason the default for `AutoSlugField` is `blank=True`

gsnider2195: Maybe just add a TODO note or open an issue to look into whether the AutoSlugField is being used incorrectly?

gsnider2195 (approved):

nautobot/extras/utils.py
```diff
@@ -487,6 +489,29 @@ def refresh_job_model_from_job_class(job_model_class, job_source, job_class, *,
...
+def check_if_key_is_graphql_safe(model_name, key):
+    """
+    Helper method to check if a key field is Python/GraphQL safe.
+    Used in CustomField for now, should be used in ComputedField and Relationship as well.
+    """
+    graphql_safe_pattern = re.compile("[_A-Za-z][_0-9A-Za-z]*")
+    if not graphql_safe_pattern.fullmatch(key):
+        raise ValidationError(
+            {
+                f"{model_name}'s specific key attribute": "This key is not Python/GraphQL safe. Please do not start the key with a digit and do not use hypens or whitespaces"
```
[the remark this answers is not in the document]
glennmatthews: How does this render in the UI if the user specifies an invalid `key`?

nautobot/extras/tests/integration/test_customfields.py
```diff
@@ -196,7 +196,7 @@ def test_custom_field_advanced_ui(self):
         and checks it appears ONLY there!.
         """
         device = self.device
-        custom_field = CustomField(
+        custom_field = CustomField.objects.create(
```
glennmatthews, line 199: Not sure I'm a fan of this latest batch of changes - by calling `objects.create()` before we do a `save()` or `validated_save()` on the object, it gets written initially to the DB even if it later fails validation. Is this just needed because of removing `_fixup_empty_fields`?

HanlinMiao: Yeah it is needed because we removed _fixup_empty_fields. 
```
cf = CustomField(type=CustomFieldTypeChoices.TYPE_TEXT, label="My Field", required=False)
cf.validated_save()
```
This sequence of code makes the validated_save() fail.

nautobot/extras/utils.py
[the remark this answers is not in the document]
HanlinMiao: Wow it actually throws a value error 
```
ValueError: 'CustomFieldForm' has no field named 'CustomField's specific key attribute'.
```
Screen Shot 2023-03-23 at 4 04 30 PM
I changed it back to `"key":`
Now it is rendering correctly on the form

nautobot/extras/tests/integration/test_customfields.py
glennmatthews: *Why* though?? We have plenty of other models where you can omit a `slug` on create and it will automatically assign one, e.g.:

```python
>>> Location(name="hello", location_type=LocationType.objects.get(name="Root"), status=Status.objects.get(name="Active")).validated_save()
>>> Location.objects.get(name="hello").slug
'hello'
```

So what's missing here that it doesn't automatically work for CustomField?

Happy to help look into this tomorrow.

glennmatthews: Ok, I did a bit of digging, and it looks like the issue is that `AutoSlugField` gets its automatic value on `save()` (actually `pre_save()`), rather than on `clean()`. So if you don't set a `key` value before cleaning, you get:

```
django.core.exceptions.ValidationError: {'key': ['This field cannot be blank.', 'This key is not Python/GraphQL safe. Please do not start the key with a digit and do not use hyphens or whitespace']}
```

- The first error is because the field is defined as `blank=False`, where it should probably be changed to the `AutoSlugField` default of `blank=True`.
- The second error is because our new clean logic rejects a blank value as not matching the regex. This could probably be addressed by changing `check_if_key_is_graphql_safe` to accept blank values or changing our clean function to not even call that function if the key is blank.
    - There's a bit of trickiness here though as this would mean that a CF with `label="123 Main St"` and no specified key would, on save, get an auto-key of `123_main_st` which isn't GraphQL valid.
    - I *think* the approach here should be to augment `slugify_dashes_to_underscores` so that it automatically makes GraphQL valid slugs in such places, perhaps by just prepending `a` or `_` to any generated slug that would otherwise start with an invalid character?

glennmatthews:
FYI it looks like these migrations may not be reversible:

```
root@716a8e3344f5:/source# nautobot-server migrate extras 0073
Operations to perform:
  Target specific migration: 0073_remove_gitrepository_fields, from extras
Running migrations:
  Rendering model states... DONE
  Unapplying extras.0076_remove_name_field_and_make_label_field_non_nullable... OK
  Unapplying extras.0075_migrate_custom_field_data... OK
  Unapplying extras.0074_rename_slug_to_key_for_custom_field...Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/mysql/base.py", line 73, in execute
    return self.cursor.execute(query, args)
  File "/usr/local/lib/python3.7/site-packages/MySQLdb/cursors.py", line 206, in execute
    res = self._query(query)
  File "/usr/local/lib/python3.7/site-packages/MySQLdb/cursors.py", line 319, in _query
    db.query(q)
  File "/usr/local/lib/python3.7/site-packages/MySQLdb/connections.py", line 254, in query
    _mysql.connection.query(self, query)
MySQLdb.IntegrityError: (1062, "Duplicate entry '' for key 'extras_customfield.extras_customfield_name_2fe72707_uniq'")
```

nautobot/extras/tests/integration/test_customfields.py
HanlinMiao: Thank you very much for looking into this! The latest commit has the following behavior:
```
>>> cf = CustomField(label="125 main st", required=False)
>>> cf.key
''
>>> cf.validated_save()
>>> cf.key
'a125_main_st'
```
I am prepending an "a" to the non graphql-safe `keys`. Since `slugify()` seems to remove the "_" I prepend to the `key`.

jathanism (approved):

glennmatthews:
> 
> Is the migrations being reversible a requirement here? I think we have to write a reverse migration in `0075_migrate_custom_field_data` to re-populate the name field again. Furthermore, I ran into this error when I got past that stage:

You're the one who's written migrations tests :-)  Are we OK with not being able to run those any more? 

> 
> ```
> django.core.exceptions.FieldError: Cannot resolve keyword 'key' into field. Choices are: advanced_ui, content_types, created, custom_field_choices, default, description, filter_logic, grouping, id, label, last_updated, name, required, slug, type, validation_maximum, validation_minimum, validation_regex, weight
> ```

Would need the whole traceback to understand where/why this is happening,

HanlinMiao:
> > Is the migrations being reversible a requirement here? I think we have to write a reverse migration in `0075_migrate_custom_field_data` to re-populate the name field again. Furthermore, I ran into this error when I got past that stage:
> 
> You're the one who's written migrations tests :-) Are we OK with not being able to run those any more?
> 
> > ```
> > django.core.exceptions.FieldError: Cannot resolve keyword 'key' into field. Choices are: advanced_ui, content_types, created, custom_field_choices, default, description, filter_logic, grouping, id, label, last_updated, name, required, slug, type, validation_maximum, validation_minimum, validation_regex, weight
> > ```
> 
> Would need the whole traceback to understand where/why this is happening,

```
  File "/usr/local/bin/nautobot-server", line 6, in <module>
    sys.exit(main())
  File "/source/nautobot/core/cli.py", line 54, in main
    run_app(
  File "/source/nautobot/core/runner/runner.py", line 263, in run_app
    management.execute_from_command_line([runner_name, command] + command_args)
  File "/usr/local/lib/python3.10/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.10/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.10/site-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.10/site-packages/django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python3.10/site-packages/django/core/management/base.py", line 89, in wrapped
    res = handle_func(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/django/core/management/commands/migrate.py", line 268, in handle
    emit_post_migrate_signal(
  File "/usr/local/lib/python3.10/site-packages/django/core/management/sql.py", line 42, in emit_post_migrate_signal
    models.signals.post_migrate.send(
  File "/usr/local/lib/python3.10/site-packages/django/dispatch/dispatcher.py", line 180, in send
    return [
  File "/usr/local/lib/python3.10/site-packages/django/dispatch/dispatcher.py", line 181, in <listcomp>
    (receiver, receiver(signal=self, sender=sender, **named))
  File "/source/nautobot/core/apps/__init__.py", line 645, in post_migrate_send_nautobot_database_ready
    nautobot_database_ready.send(sender=app_conf, app_config=app_conf, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/django/dispatch/dispatcher.py", line 180, in send
    return [
  File "/usr/local/lib/python3.10/site-packages/django/dispatch/dispatcher.py", line 181, in <listcomp>
    (receiver, receiver(signal=self, sender=sender, **named))
  File "/source/examples/example_plugin/example_plugin/signals.py", line 30, in nautobot_database_ready_callback
    field, _ = CustomField.objects.update_or_create(
  File "/usr/local/lib/python3.10/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/django/db/models/query.py", line 608, in update_or_create
    obj, created = self.select_for_update().get_or_create(defaults, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/django/db/models/query.py", line 581, in get_or_create
    return self.get(**kwargs), False
  File "/usr/local/lib/python3.10/site-packages/cacheops/query.py", line 351, in get
    return qs._no_monkey.get(qs, *args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/django/db/models/query.py", line 424, in get
    clone = self._chain() if self.query.combinator else self.filter(*args, **kwargs)
  File "/source/nautobot/core/models/querysets.py", line 115, in filter
    return super().filter(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/django/db/models/query.py", line 941, in filter
    return self._filter_or_exclude(False, args, kwargs)
  File "/usr/local/lib/python3.10/site-packages/django/db/models/query.py", line 961, in _filter_or_exclude
    clone._filter_or_exclude_inplace(negate, args, kwargs)
  File "/usr/local/lib/python3.10/site-packages/django/db/models/query.py", line 968, in _filter_or_exclude_inplace
    self._query.add_q(Q(*args, **kwargs))
  File "/usr/local/lib/python3.10/site-packages/django/db/models/sql/query.py", line 1416, in add_q
    clause, _ = self._add_q(q_object, self.used_aliases)
  File "/usr/local/lib/python3.10/site-packages/django/db/models/sql/query.py", line 1435, in _add_q
    child_clause, needed_inner = self.build_filter(
  File "/usr/local/lib/python3.10/site-packages/django/db/models/sql/query.py", line 1309, in build_filter
    lookups, parts, reffed_expression = self.solve_lookup_type(arg)
  File "/usr/local/lib/python3.10/site-packages/django/db/models/sql/query.py", line 1135, in solve_lookup_type
    _, field, _, lookup_parts = self.names_to_path(lookup_splitted, self.get_meta())
  File "/usr/local/lib/python3.10/site-packages/django/db/models/sql/query.py", line 1562, in names_to_path
    raise FieldError("Cannot resolve keyword '%s' into field. "
django.core.exceptions.FieldError: Cannot resolve keyword 'key' into field. Choices are: advanced_ui, content_types, created, custom_field_choices, default, description, filter_logic, grouping, id, label, last_updated, name, required, slug, type, validation_maximum, validation_minimum, validation_regex, weight
```

It is the `example_plugin's` signal to create a `CustomField` at `examples/example_plugin/example_plugin/signals.py`. I specified a `key` here but I think it is expecting a `slug` field when I roll the migrations back :(

```
field, _ = CustomField.objects.update_or_create(
        key="example_plugin_auto_custom_field",  # Note underscores rather than dashes!
        defaults={
            "type": CustomFieldTypeChoices.TYPE_TEXT,
            "label": "Example Plugin Automatically Added Custom Field",
            "default": "Default value",
        },
    )
    field.content_types.set([ContentType.objects.get_for_model(Location)])
```

glennmatthews:
Ooh, that's a gross case, but a good example of a situation that a real app might need to handle. Might need to add some if/else logic there along the lines of `if hasattr(CustomField, "slug"):... else:` to allow the plugin to work correctly with both before/after?

PR #3434, opened 2023-03-16
Change docs to not recommend AbortTransaction.

Kircheneer:
# Closes: DNE
# What's Changed

Change docs to not recommend use of `AbortTransaction` to end jobs. When doing this, the job status is not set properly and the job result always stays in running.

# TODO


nautobot/docs/additional-features/jobs.md
```diff
@@ -472,7 +472,7 @@ Markdown rendering is supported for log messages.
...
 
 !!! note
-    Using `self.log_failure()`, in addition to recording a log message, will flag the overall job as failed, but it will **not** stop the execution of the job, nor will it result in an automatic rollback of any database changes made by the job. To end a job early, you can use a Python `raise` or `return` as appropriate. Raising `nautobot.utilities.exceptions.AbortTransaction` will ensure that any database changes are rolled back as part of the process of ending the job.
+    Using `self.log_failure()`, in addition to recording a log message, will flag the overall job as failed, but it will **not** stop the execution of the job, nor will it result in an automatic rollback of any database changes made by the job. To end a job early, you can use a Python `raise` or `return` as appropriate. Raising any exception (e.g. `ValueError` for malformed input values) will ensure that any database changes are rolled back as part of the process of ending the job. `AbortTransaction` from Nautobot, which was recommended in past versions of the docs, should explicitly _not_ be used, as it is treated specially by Nautobot Jobs.
```
gsnider2195, line 475: Maybe the change below or something similar to explicitly state that `AbortTransaction` should not be used by job authors?

```suggestion
    Using `self.log_failure()`, in addition to recording a log message, will flag the overall job as failed, but it will **not** stop the execution of the job, nor will it result in an automatic rollback of any database changes made by the job. To end a job early, you can use a Python `raise` or `return` as appropriate. Raising any exception (e.g. `ValueError` for malformed input values) will ensure that any database changes are rolled back as part of the process of ending the job. `AbortTransaction` from Nautobot, which was recommended in past versions of the docs, should explicitly _not_ be used, as it is only intended for use by Nautobot's internal job handling.
```

gsnider2195 (approved):

PR #3444, opened 2023-03-17
[WIP] UI 2.0 Cleanup

bryanculver:
# Closes: 😵 
# What's Changed

What started as fixing CSRF snowballed into some other changes.

# TODO

- [ ] Unit, Integration Tests


docker/docker-entrypoint.sh
```diff
@@ -9,12 +9,19 @@ DB_WAIT_TIMEOUT=${DB_WAIT_TIMEOUT-3}
 MAX_DB_WAIT_TIME=${MAX_DB_WAIT_TIME-30}
 CUR_DB_WAIT_TIME=0
 
+# set NAUTOBOT_DOCKER_ALLOW_UI_BUILD_FAILURE to NAUTOBOT_DOCKER_ALLOW_UI_BUILD_FAILURE, defaulting to "":
+# TODO: Move to just a true/false flag
+UI_BUILD_FAILURE_FLAG=""
+if [[ "$NAUTOBOT_DOCKER_ALLOW_UI_BUILD_FAILURE" == "true" ]]; then
+  UI_BUILD_FAILURE_FLAG="--allow-ui-build-failure"
+fi
+
```
bryanculver, line 18: In development mode (with NodeJS container) the backend fails to start if the UI fails to build. If you are in-flight with making front-end changes this becomes a bit of a chicken-egg problem because you want the NodeJS container up so you can validate the changes but the backend is down and won't answer your API calls.

development/docker-compose.dev.yml
```diff
@@ -25,7 +25,7 @@ services:
     image: node:18
     environment:
       PORT: "3000"
-      # DEBUG: "express:*" # nodejs web server debug logging
+      DEBUG: "express:*" # nodejs web server debug logging
```
bryanculver, line 28: Can probably revert.

nautobot/core/api/views.py
```diff
@@ -418,7 +418,7 @@ class ModelViewSet(
     BulkDestroyModelMixin,
     ModelViewSetMixin,
     ModelViewSet_,
-    MyFormSchemaViewSetMixin,
+    # MyFormSchemaViewSetMixin,  # TODO: This can/should be limited to /api/ui/ schemas
```
bryanculver, line 421: This was causing issues with schema output. I think this should be opt-in via either a header or something else but is a breaking change to the APIs for what amounts to UI data.

nautobot/core/management/commands/build_ui.py
```diff
@@ -37,6 +37,13 @@ def add_arguments(self, parser):
             default=True,
             help="Do not compile UI.",
         )
+        parser.add_argument(
+            "--allow-ui-build-failure",
+            action="store_true",
+            dest="allow_ui_build_failure",
+            default=True,
+            help="Allow build errors in development (UI may fail to build)",
+        )
```
bryanculver, line 46: As stated above, allow the UI build to fail and still start Django in development cases.

nautobot/core/urls.py
```diff
@@ -15,6 +15,7 @@
 urlpatterns = [
     # Base views
     path("", HomeView.as_view(), name="home"),
+    path("csrfrefresh/", CSRFRefreshView.as_view(), name="csrfrefresh"),
```
bryanculver, line 18: Still need to update base to ensure all views get a cookie.

nautobot/ui/package.json
```diff
@@ -30,8 +31,11 @@
     "react-dom": "^18.2.0",
     "react-dotenv": "^0.1.3",
     "react-icons": "^4.7.1",
+    "react-redux": "^8.0.5",
     "react-router-dom": "^6.4.3",
     "react-scripts": "5.0.1",
+    "redux-dynamic-middlewares": "^2.2.0",
```
bryanculver, line 37: Not used in this phase so should be removed. May not be needed depending on how we generate a full API object.

nautobot/ui/src/components/index.js
```diff
@@ -0,0 +1 @@
+// TODO: Let's make components importable from here
```
bryanculver, line 1: Allow

```js
import { SidebarNav, RouterLink, LoadingWidget } from "@components";
```

instead of

```js
import SidebarNav from "@components/common/SidebarNav";
import RouterLink from "@components/common/RouterLink";
import LoadingWidget from "@components/common/LoadingWidget";
```

nautobot/ui/src/index.js
```diff
@@ -1,40 +1,60 @@
...
-const root = ReactDOM.createRoot(document.getElementById('root'));
 root.render(
   <React.StrictMode>
-    <NautobotUIProvider>
-      <BrowserRouter>
-        <Layout>
-          <NautobotRouter />
-        </Layout>
-      </BrowserRouter>
-    </NautobotUIProvider>
```
bryanculver, line 37: Moved to `<App />` pattern to follow with `create-react-app` example.

nautobot/ui/src/router.js
```diff
@@ -1,60 +1,54 @@
...
+    {
+      path: "/:app_name/:model_name",
+      element: <ListView />,
+      children: []
+    },
+    {
+      path: "/:app_name/:model_name/:pk",
+      element: <DetailView />,
+      children: []
+    },
```
bryanculver, line 37: Somehow I changed this, needs to be reverted but DetailView is semi-broken at the moment.

nautobot/ui/src/utils/api.js
```diff
@@ -0,0 +1,76 @@
+// Import the RTK Query methods from the React-specific entry point
+import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/query/react'
+import { API_BASE, API_USER_SESSION_INFO, API_UI_MENU_INFO } from '@constants/apiPath'
+
+/*
+  The one true API!
+
+  Using Redux's RTK Query making fetching and caching a breeze! This will provide hooks and wires automagically.
```
bryanculver, line 8: This follows the documented example from RTK Query: https://redux-toolkit.js.org/rtk-query/overview#basic-usage

nautobot/ui/src/utils/api.js
```diff
@@ -0,0 +1,76 @@
...
+// Otherwise you can retrieve them via navigating baseApi:
+//   import baseApi
+//   baseApi.useGetSessionQuery
+//   // or  
+//   baseApi.endpoints.getSession.getQuery
+export const { useGetSessionQuery, useGetUIMenuQuery } = baseApi
+
+
+
+// TODO: Below is a pattern for taking the menu API and building an entire RTK-Query API for it
```
bryanculver, line 36: 😭

nautobot/users/api/views.py
```diff
@@ -42,6 +42,33 @@ def my_profile(self, request):
         serializer = self.serializer_class(instance=request.user, context={"request": request})
         return Response(serializer.data)
 
+    @action(methods=["GET"], detail=False, url_path="session", permission_classes=[AllowAny])
+    def session(self, request):
```
bryanculver, line 46: Ugly as sin. Would love some thought on what else we would want in here and then I could reformat as necessary.

package-lock.json
bryanculver, line 1: :suspect:

nautobot/ui/src/utils/session.js
bryanculver, line 1: Probably could :yeet: but does show how to store global state if not using RTK Query pattern. Discovered it, may need to continue with it in some cases.

docker/docker-entrypoint.sh
glennmatthews: I'm not a big fan of adding a developer-workflow-specific flag to the docker-entrypoint that's used across the board. Maybe instead a `skip-rebuild-ui` flag or something along those lines might be more generally useful?

nautobot/core/management/commands/build_ui.py
glennmatthews: Actually, how is this different from `--no-npm-build`?

nautobot/ui/src/components/core/Table.jsx
```diff
@@ -1,8 +1,8 @@
 import { Checkbox, Table, Thead, Tr, Th, Tbody, Td } from "@nautobot/nautobot-ui"
 
-import BSTableItem from "@components/core/BSTableItem"
-
+import BSTableItem from "@components/core/TableItem"
```
glennmatthews, line 3:
```suggestion
import TableItem from "@components/core/TableItem"
```

nautobot/ui/src/views/BSCreateViewTemplate.jsx
glennmatthews, line 1: Need to rename this file, if it's still relevant.

glennmatthews (commented):
Just a few quick observations.

docker/docker-entrypoint.sh
bryanculver: I think @gsnider2195 pointed out to me "why don't you just skip `post_upgrade`" which might be sufficient anyway.

nautobot/core/management/commands/build_ui.py
bryanculver: It attempts to build but allows it to fail.

nautobot/ui/src/components/core/Table.jsx
bryanculver: 🤦

nautobot/ui/src/views/BSCreateViewTemplate.jsx
bryanculver: That's what I meant by this except with different words: https://github.com/nautobot/nautobot/pull/3444/files#diff-6707707683d0774b15383962c9716911e5abe2738a3c866668795bbfb7547e22R11

gsnider2195:
> DB connection exhaustion from NodeJS dev container

I believe I was able to resolve this by setting `NAUTOBOT_DB_TIMEOUT=0`

nautobot/ui/src/components/common/SidebarNav.jsx
```diff
@@ -0,0 +1,56 @@
...
+                    Object.entries(group[1].items).map((menu, menu_idx) => (
+                        <SidebarButton key={menu_idx} level={2}>
+                            <RouterLink to={menu[0]}>{menu[1].name}</RouterLink>
+                        </SidebarButton>
+                    ))
+                    }
+                        </AccordionPanel>
+                    </AccordionItem>
+                </Accordion>
+                ))
```
jathanism, line 48: These nested `Object.foo.map()` calls without returns are probably going to make jslint angry.

nautobot/ui/src/views/Home.jsx
```diff
@@ -1,13 +1,11 @@
 import { Alert } from "@chakra-ui/react"
 
-
-export default function Home({ children }) {
+export default function Home() {
```
jathanism, line 3: Wasn't the `children` thing a leave over from embedding plugins (apps)?

nautobot/ui/src/views/Login.jsx
```diff
@@ -1,9 +1,19 @@
...
-import { Card, CardHeader, CardBody, CardFooter } from "@chakra-ui/react"
+import { Button, FormControl, FormLabel, Input, Box } from "@nautobot/nautobot-ui"
 import axios from "axios"
 
+import { useGetSessionQuery } from "@utils/api";
+import { useNavigate } from "react-router-dom";
+
+axios.defaults.withCredentials = true
+axios.defaults.xsrfCookieName = 'csrftoken'
+axios.defaults.xsrfHeaderName = 'X-CSRFToken'
```
jathanism, line 9: This looks redundant based on what is inside of `axios-api.js` which is apparently not yet being used?

nautobot/ui/src/views/Login.jsx
```diff
@@ -12,16 +22,16 @@ export default function Login() {
...
+        refetchSession().then(() => {navigate("/")});
+
       })
       .catch(err => alert(err.detail))
   }
+
+  
   return (
-      <Card>
+      <Box boxShadow='base' p='6' rounded='md' bg='white'>
```
jathanism, line 33: Why is explicitly styling required here? Can this be done using a css import instead?

nautobot/ui/src/views/Logout.jsx
```diff
@@ -0,0 +1,31 @@
+import { Button, FormControl, FormLabel, Input, Link, Box } from "@nautobot/nautobot-ui"
+import axios from "axios"
+
+import { useGetSessionQuery } from "@utils/api";
+import { useNavigate } from "react-router-dom";
+
+axios.defaults.withCredentials = true
+axios.defaults.xsrfCookieName = 'csrftoken'
+axios.defaults.xsrfHeaderName = 'X-CSRFToken'
```
jathanism, line 9: I'm starting to see why you thought to factor this into `axio-api.js`. 😉

nautobot/ui/src/views/generic/ObjectList.jsx
```diff
@@ -1,24 +1,77 @@
...
-  // let pagetitle = "Nautobot"
-  // if (model_name) {
-  //   pagetitle = model_name.replace("-", " ")
-  //   pagetitle = pagetitle[0].toUpperCase() + pagetitle.slice(1)
-  // }
+  const { [0]: searchParams } = useSearchParams();
+
+  // TODO: If we can solve the dynamic API of RTK-Query we won't need to construct the API URLs
+  const list_url = `/api/${app_name}/${model_name}/`
+  const headers_url = list_url + "table-fields/"
```
jathanism, line 20: There's also the `useLocation()` effect hook, and I'm wondering if we might consider using that to grab the URI vs. manually constructing it? Not a big deal, but trying to think of ways to avoid having to hard-code URLs.

nautobot/users/api/views.py
jathanism: `url_path` defaults to match the method name if not provided, so you could at least omit that. :)

nautobot/users/api/views.py
```diff
@@ -42,6 +42,33 @@ def my_profile(self, request):
...
+
+        social_auth_backends = backends(request)["backends"]
+        if sso_enabled:
+            for backend in social_auth_backends["backends"]:
+                _backends.append(reverse("social:begin", kwargs={"backend": backend}))
+
+        resp = {
+            "user": serializer.data,
+            "logged_in": request.user.is_authenticated,
+            "sso_enabled": sso_auth_enabled(django_settings.AUTHENTICATION_BACKENDS),
```
jathanism, line 65:
```suggestion
            "sso_enabled": sso_enabled,
```

jathanism (approved):
I don't see anything show stopping and this is a big step forward.

nautobot/ui/src/views/Login.jsx
bryanculver: This isn't CSS, it's control properties for the React component. But I worked with the design team to get some "standards" around these so I could provide a canonical/default box.

nautobot/ui/src/views/generic/ObjectList.jsx
bryanculver: We shouldn't be crafting this in this way. This is overall just ugly. We should provide a getter that doesn't have to construct the URLs based on an assumption on how the API is constructed underneath.

jathanism: Even better. 😗

nautobot/ui/src/views/Home.jsx
bryanculver: Not expressly. `children` is anything passed on the inside of the component. So:

```js
// App.jsx
<Home>
  <Button />
</Home>
```

`<Button />` become `children`.

However when we build the pluggable homepage these will likely need to be _either_ loaded directly in the `<Home>` component (not passed in) or at least passed in as named arguments, so

```js
// App.jsx
const LeftHomePanel = <Button />
const RightHomePanel = <Text />

<Home left=LeftHomePanel right=RightHomePanel />
```

nautobot/ui/package.json
```diff
@@ -10,7 +10,8 @@
     "@fortawesome/fontawesome-svg-core": "^6.2.1",
     "@fortawesome/free-solid-svg-icons": "^6.2.1",
     "@fortawesome/react-fontawesome": "^0.2.0",
-    "@nautobot/nautobot-ui": "nautobot/nautobot-ui",
+    "@nautobot/nautobot-ui": "git+ssh://git@github.com:nautobot/nautobot-ui.git",
```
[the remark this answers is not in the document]
bryanculver: Bad merge in from upstream. Reverted 👍

Issue #2915, opened 2022-11-28
IPAM Overhaul: Prefix-Prefix Parent Concrete Relationship


bryanculver:
Similar to #1673 in concert with #1032, Make Prefixes have a concrete relationship to each other for performance and get away from derivative assumed parent/children.

Address issues like:

- #780

```[tasklist]
### TODO
- [ ] https://github.com/nautobot/nautobot/issues/3466
- [ ] https://github.com/nautobot/nautobot/issues/3467
- [ ] #3681
- [ ] #3687 
```

jathanism:
Watch this space for better clarity based on uniqueness constraint discussions.

PR #3452, opened 2023-03-20
Implement `Prefix.parent` concrete foreign key to self.

jathanism:
# Closes: #2915 
# What's Changed
- Also broke out IP version info `Prefix.ip_version` concrete field that is non-editable and gets populated on object create.
- Includes various tree methods (descendants/ancestors/siblings/root/leaf)
- Added `index_together` for (`network`, `broadcast`, `prefix_length`)
- Added `force_delete` argument to `Prefix.delete()` to allow a Prefix that has children to be forcibly deleted and to reparent its direct children to the old parent of the deleted Prefix, if there is one.

# TODO

- [ ] Unit, Integration Tests


nautobot/ipam/tests/test_models.py
```diff
@@ -1,4 +1,4 @@
-from unittest import skipIf
+rom unittest import skipIf
```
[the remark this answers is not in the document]
itdependsnetworks: nvm missed this was a draft.

nautobot/ipam/models.py
```diff
@@ -378,15 +381,40 @@ def clean(self):
...
+
+        try:
+            super().delete(**kwargs)
+        except models.ProtectedError as err:
+            # Update protected objects to use the new parent and delete the old parent (self).
+            if force_delete:
+                protected_pks = (po.pk for po in err.protected_objects)
+                protected_objects = Prefix.objects.filter(pk__in=protected_pks)
+                protected_objects.update(parent=self.parent)
+                super().delete(**kwargs)
```
glennmatthews, line 395: `delete()` actually has a return signature, though I think we're not respecting it in a number of other places already.

```suggestion
            return super().delete(**kwargs)
        except models.ProtectedError as err:
            # Update protected objects to use the new parent and delete the old parent (self).
            if force_delete:
                protected_pks = (po.pk for po in err.protected_objects)
                protected_objects = Prefix.objects.filter(pk__in=protected_pks)
                protected_objects.update(parent=self.parent)
                return super().delete(**kwargs)
```

nautobot/ipam/models.py
```diff
@@ -378,15 +381,40 @@ def clean(self):
...
     def save(self, *args, **kwargs):
 
         if isinstance(self.prefix, netaddr.IPNetwork):
 
             # Clear host bits from prefix
             self.prefix = self.prefix.cidr
 
+        supernets = self.supernets(discover_mode=True)
+        if supernets:
+            import operator
```
glennmatthews, line 408: can this be a top-level import instead?

nautobot/ipam/models.py
```diff
@@ -425,18 +453,120 @@ def family(self):
...
+            ip_version=self.ip_version,
+            network__gte=self.network,
+            broadcast__lte=self.broadcast,
+            # namespace=self.namespace,
+        )
+
+        query.update(parent=self)
+
+    def supernets(self, direct=False, discover_mode=False, for_update=False):
+        """Return supernets of this Prefix."""
```
glennmatthews, line 473: Please document the meaning of the named args to this function. :)

nautobot/ipam/models.py
```diff
@@ -425,18 +453,120 @@ def family(self):
...
+
+        return query.filter(
+            ip_version=self.ip_version,
+            prefix_length__lt=self.prefix_length,
+            network__lte=self.network,
+            broadcast__gte=self.broadcast,
+            # namespace=self.namespace,
+        )
+
+    def subnets(self, direct=False, for_update=False):
```
glennmatthews, line 496: Should this have a `discover_mode` flag for parity with `supernets()`?

nautobot/ipam/models.py
```diff
@@ -240,6 +240,20 @@ class Prefix(PrimaryModel, StatusModel, RoleModelMixin):
         choices=choices.PrefixTypeChoices,
         default=choices.PrefixTypeChoices.TYPE_NETWORK,
     )
+    parent = models.ForeignKey(
+        "self",
+        blank=True,
+        null=True,
+        related_name="child_prefixes",  # Maybe child_prefixes? Align with "child_ips"?
```
glennmatthews, line 247: Why not `children` for parity with our other self-referential models?

nautobot/ipam/models.py
```diff
@@ -425,18 +453,120 @@ def family(self):
...
+        """
+        return not self.child_prefixes.exists()
+
+    def is_root_node(self):
+        """
+        Returns whether I am a root node (no parent).
+        """
+        return self.parent is None
+
+    def get_ancestors(self, ascending=False):
```
glennmatthews, line 532: Would be good to have the a similar API signature as our `tree-queries` models, in which case this would be `def ancestors(self, include_self=False)` (possibly still including `ascending=False` as an additional optional kwarg, though that's another one I'm not sure I see the need for at this time.)

nautobot/ipam/models.py
```diff
@@ -425,18 +453,120 @@ def family(self):
...
+        """Return my ancestors."""
+        query = self.supernets()
+        if ascending:
+            query = query.reverse()
+        return query
+
+    def get_children(self):
+        """Return my immediate children."""
+        return self.subnets(direct=True)
```
glennmatthews, line 541: Isn't this redundant with the `child_prefixes` reverse relation?

nautobot/ipam/models.py
jathanism: I'm a fan of that, but I did this because we're also going to have `child_ips` relationship from `IPAddress`. My thought is to avoid the ambiguity of which is which.

nautobot/ipam/models.py
jathanism: Yes, thanks. Good catch.

nautobot/ipam/models.py
jathanism: Oops, yes.

nautobot/ipam/models.py
jathanism: No such concept, so it would just be a no-op.

nautobot/ipam/models.py
jathanism: Yeah, I'll add that.

jathanism: I thought about that, but given the existing supernets algorithm naturally excludes `self` because of `prefix_length__lt`. We could change it to `prefix_length__lte` (note the trailing `e`) and then have `include_self=False` work.

nautobot/ipam/models.py
```diff
@@ -425,18 +453,120 @@ def family(self):
...
+        return self.subnets()
+
+    def get_root(self):
+        """
+        Returns the root node (the parent of all of my ancestors).
+        """
+        ancestors = self.get_ancestors()
+        return ancestors.first()
+
+    def get_siblings(self, include_self=False):
```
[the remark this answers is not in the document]
jathanism: Just tree things. It's helpful if wanting to identify neighbors.

nautobot/ipam/models.py
```diff
@@ -425,18 +453,120 @@ def family(self):
...
+        """
+        Return my siblings. Root nodes are siblings to other root nodes.
+        """
+        query = Prefix.objects.filter(parent=self.parent)
+        if not include_self:
+            query = query.exclude(id=self.id)
+
+        return query
 
     def get_child_prefixes(self):
```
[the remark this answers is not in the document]
jathanism: Agreed with that!

nautobot/ipam/models.py
jathanism: It is. Just had it here for taxonomy completeness for tree methods.

glennmatthews: > child_prefixes relation actually returns all descendants, not merely direct children.

I don't see how that can be possible - non-child descendants don't have `parent = <this instance>` so how can the reverse relation include them?

nautobot/ipam/tests/test_filters.py
```diff
@@ -258,6 +258,11 @@ def test_vrf(self):
     def test_present_in_vrf(self):
         # clear out all the randomly generated route targets and vrfs before running this custom test
         test_prefixes = list(self.queryset.values_list("pk", flat=True)[:10])
+        # With advent of `Prefix.parent`, Prefixes can't just be bulk deleted without clearing their
+        # `parent` first in an `update()` query which doesn't call `save()` or `fire `(pre|post)_save` signals.
+        unwanted_prefixes = self.queryset.exclude(pk__in=test_prefixes)
+        unwanted_prefixes.update(parent=None)
+        unwanted_prefixes.delete()
         self.queryset.exclude(pk__in=test_prefixes).delete()
```
glennmatthews, line 266: Redundant after the above addition?

nautobot/ipam/tests/test_models.py
```diff
@@ -174,17 +182,81 @@ def test_prefix_validation(self):
...
+        self.assertEqual(list(self.root.subnets(include_self=True)), [self.root, self.parent, self.child1, self.child2])
+
+        # is_child_node()
+        self.assertFalse(self.root.is_child_node())
+        self.assertTrue(self.parent.is_child_node())
+        self.assertTrue(self.child1.is_child_node())
+
+        # is_leaf_node()
+        self.assertFalse(self.root.is_leaf_node())
+        self.assertTrue(self.child1.is_leaf_node())
```
glennmatthews, line 206:
```suggestion
        self.assertFalse(self.parent.is_leaf_node())
        self.assertTrue(self.child1.is_leaf_node())
```

nautobot/ipam/tests/test_models.py
```diff
@@ -174,17 +182,81 @@ def test_prefix_validation(self):
...
+
+        # is_root_node()
+        self.assertTrue(self.root.is_root_node())
+        self.assertFalse(self.child1.is_root_node())
```
glennmatthews, line 210:
```suggestion
        self.assertFalse(self.parent.is_root_node())
        self.assertFalse(self.child1.is_root_node())
```

nautobot/ipam/tests/test_models.py
```diff
@@ -174,17 +182,81 @@ def test_prefix_validation(self):
...
+        # Force delete to coerce reparenting
+        num_deleted, _ = self.parent.delete(force_delete=True)
+        self.assertEqual(num_deleted, 1)
+
+        self.assertEqual(list(self.root.get_children()), [self.child1, self.child2])
+        self.child1.refresh_from_db()
+        self.child2.refresh_from_db()
+        self.assertEqual(self.child1.parent, self.root)
+        self.assertEqual(self.child2.parent, self.root)
+        self.assertEqual(list(self.child1.ancestors()), [self.root])
```
glennmatthews, line 257: Can we add to this test to recreate the /25 and verify that the /26s get reparented under it?

nautobot/ipam/tests/test_models.py
```diff
@@ -416,14 +488,14 @@ def test_get_utilization(self):
...
         duplicate_prefix = Prefix(prefix=netaddr.IPNetwork("192.0.2.0/24"))
-        self.assertIsNone(duplicate_prefix.clean())
+        self.assertRaises(ValidationError, duplicate_prefix.clean)
 
-    @override_settings(ENFORCE_GLOBAL_UNIQUE=True)
     def test_duplicate_global_unique(self):
+        """With ENFORCE_GLOBAL_UNIQUE hard-coded, this should raise a ValidationError."""
         Prefix.objects.create(prefix=netaddr.IPNetwork("192.0.2.0/24"))
         duplicate_prefix = Prefix(prefix=netaddr.IPNetwork("192.0.2.0/24"))
         self.assertRaises(ValidationError, duplicate_prefix.clean)
```
glennmatthews, line 501: Looks like these two tests are identical now, remove one?

glennmatthews (commented):
Looks real promising!

nautobot/ipam/models.py
jathanism: You are correct. Let's just pretend this didn't happen. So yeah it's redundant at the moment.

nautobot/ipam/models.py
```diff
@@ -240,6 +241,20 @@ class Prefix(PrimaryModel, StatusModel, RoleModelMixin):
         choices=choices.PrefixTypeChoices,
         default=choices.PrefixTypeChoices.TYPE_NETWORK,
     )
+    parent = models.ForeignKey(
+        "self",
+        blank=True,
+        null=True,
+        related_name="child_prefixes",  # Maybe child_prefixes? Align with "child_ips"?
+        on_delete=models.PROTECT,
```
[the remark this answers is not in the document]
jathanism: Yeah it makes it impossible without the `force_delete` reparenting of the `protected_objects`.

nautobot/ipam/models.py
```diff
@@ -378,15 +397,41 @@ def clean(self):
...
+                raise
+
     def save(self, *args, **kwargs):
 
         if isinstance(self.prefix, netaddr.IPNetwork):
 
             # Clear host bits from prefix
             self.prefix = self.prefix.cidr
 
+        for_update = kwargs.pop("for_update", False)
```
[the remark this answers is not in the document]
jathanism: On large changes where you might want a transaction lock on reparenting. I need to document my thoughts in the next pass.

nautobot/ipam/models.py
```diff
@@ -425,18 +470,141 @@ def family(self):
...
+        if not include_self:
+            query = query.exclude(id=self.id)
+
+        return query.filter(
+            ip_version=self.ip_version,
+            prefix_length__lte=self.prefix_length,
+            network__lte=self.network,
+            broadcast__gte=self.broadcast,
+            # namespace=self.namespace,
+        )
```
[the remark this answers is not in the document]
jathanism: It's implied by meta ordering, so that there is always natural sorting by `network, `broadcast`, `prefix_length`. Once we add `namespace` that will be the first one in the list.

I will document. Thanks.

nautobot/ipam/tests/test_filters.py
jathanism: Yes, thanks. You're catching all my self-review items! 💋

nautobot/ipam/migrations/0027_ipam__prefix__add_parent.py
bryanculver, line 1: Let's stub out the data migration even though it won't be implemented at this time. Saves re-ordering headaches.

nautobot/ipam/models.py
```diff
@@ -536,35 +557,32 @@ def _deconstruct_prefix(self, pre):
...
         return reverse("ipam:prefix", args=[self.pk])
 
+    def get_duplicates(self):
+        return Prefix.objects.net_equals(self.prefix).filter(vrf=self.vrf).exclude(pk=self.pk)
+
     def clean(self):
         super().clean()
 
+        # TODO(jathan): Restored this check for the purpose of unit testing, but this will go away
+        # once we merge Namespaces.
```
glennmatthews, line 576: can this be removed now then?

nautobot/ipam/models.py
```diff
@@ -867,10 +1039,6 @@ def clean(self):
 
         if self.address:
 
-            # /0 masks are not acceptable
-            if self.address.prefixlen == 0:
-                raise ValidationError({"address": "Cannot create IP address with /0 mask."})
```
glennmatthews, line 872: IIRC there's logic in `ipam/factory.py` that's preventing it from creating /0 prefixes, might be able to remove that now?

nautobot/ipam/models.py
```diff
@@ -574,15 +592,46 @@ def clean(self):
...
 
+    def delete(self, *args, **kwargs):
+        """
+        A Prefix with children will be impossible to delete and raise a `ProtectedError`.
+
+        Passing `force_delete=True` will catch the error and explicitly update the
+        `protected_objects` from the exception setting their parent to the old parent of this
+        prefix, and then this prefix will be deleted.
+        """
+        force_delete = kwargs.pop("force_delete", False)
```
glennmatthews, line 603: As raised during the knowledge share today, let's revisit whether this is needed - since deleting a prefix is non-destructive to its descendants and IP addresses, maybe we can just always have this behavior?

nautobot/ipam/models.py
```diff
@@ -621,22 +670,145 @@ def family(self):
...
-        Return all Prefixes within this Prefix and VRF. If this Prefix is a container in the global table, return child
-        Prefixes belonging to any VRF.
+        query = Prefix.objects.select_for_update().filter(
+            ~models.Q(id=self.id),  # Don't include yourself...
+            parent_id=self.parent_id,
+            prefix_length__gt=self.prefix_length,
+            ip_version=self.ip_version,
+            network__gte=self.network,
+            broadcast__lte=self.broadcast,
+            # namespace=self.namespace,
```
glennmatthews, line 687: Uncomment now here and below?

nautobot/ipam/models.py
```diff
@@ -621,22 +670,145 @@ def family(self):
             return self.prefix.version
```
glennmatthews, line 670: Remove `family` property now that we have `ip_version`?

nautobot/ipam/models.py
```diff
@@ -502,8 +519,12 @@ class Meta:
...
             "network",
             "prefix_length",
-        )  # (vrf, prefix) may be non-unique
-        unique_together = ("namespace", "network", "prefix_length")
+        )
+        index_together = [
+            ["network", "broadcast", "prefix_length"],
+            ["namespace", "network", "broadcast", "prefix_length"],
+        ]
+        unique_together = ["namespace", "network", "prefix_length"]
```
glennmatthews, line 527: Should be able to remove the `natural_key` `AttributeRemover` from this model now that it has a proper natural key. Yay!

nautobot/ipam/tests/test_models.py
```diff
@@ -163,8 +163,16 @@ def test_get_db_prep_value_mysql(self):
 class TestPrefix(TestCase):  # TODO change to BaseModelTestCase
```
glennmatthews, line 163: Now that `setUp` is creating several instances, once the natural-key is re-enabled for this model, you should be able to address this TODO:

```suggestion
class TestPrefix(ModelTestCases.BaseModelTestCase):
```

nautobot/ipam/models.py
```diff
@@ -409,6 +411,21 @@ class Prefix(PrimaryModel, StatusModel, RoleModelMixin):
         choices=choices.PrefixTypeChoices,
         default=choices.PrefixTypeChoices.TYPE_NETWORK,
     )
+    parent = models.ForeignKey(
```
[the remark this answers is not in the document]
jathanism: Specifically `ip_version` should replace `ip_family` filter.

nautobot/ipam/models.py
jathanism: It's gone!

nautobot/ipam/models.py
jathanism: In the latest patch this is removed entirely and "it just works".

nautobot/ipam/models.py
jathanism: Definitely not looking forward to revising the factory. 😩

nautobot/ipam/factory.py
```diff
@@ -248,7 +248,7 @@ class Params:
         has_role = NautobotBoolIterator()
         has_tenant = NautobotBoolIterator()
         has_vlan = NautobotBoolIterator()
-        has_vrf = NautobotBoolIterator()
+        # has_vrf = NautobotBoolIterator()
```
HanlinMiao, line 251: Are we leaving this line here intentionally?

Edited: I see we are disabling `vrf` temporarily.

nautobot/ipam/forms.py
```diff
@@ -323,12 +323,6 @@ class Meta:
             "date_allocated": DateTimePicker(),
         }
 
-    def __init__(self, *args, **kwargs):
```
HanlinMiao, line 326: What is this thing doing over here before lol?

nautobot/ipam/querysets.py
```diff
@@ -204,48 +196,6 @@ def string_search(self, search):
 class PrefixQuerySet(BaseNetworkQuerySet):
     """Queryset for `Prefix` objects."""
 
-    def annotate_tree(self):
```
HanlinMiao, line 207: 💯

HanlinMiao (approved):
Noice!

nautobot/ipam/factory.py
jathanism: Right it's going to be fixed in the namespace work since VRFs are unique to namespaces now.

jathanism:
> Noice! Unittests are still failing tho, is that expected?

Yes, they are failing because of the Namespace and VRF changes.

nautobot/ipam/forms.py
jathanism:
```python

        self.fields["vrf"].empty_label = "Global"
```

Issue #3154, opened 2023-01-24
TagFilter created from BaseFilterSet Meta.fields is not using `slug`


gsnider2195:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5-2.0a
* Python version: 3.7-3.10
* Database platform, version: postgres
* Middleware(s):

### Steps to Reproduce
1. Remove `tag = TagFilter()` from a filterset and add `tags` to the `Meta.fields`
2. Try to filter on tag slugs
3. [Example on next demo](https://next.demo.nautobot.com/dcim/devices/?tags=test)

### Expected Behavior

Filter on tag slug

### Observed Behavior

Exception:
```py
Environment:


Request Method: GET
Request URL: http://localhost:8080/dcim/device-types/?&console_ports=&console_server_ports=&power_ports=&power_outlets=&interfaces=&pass_through_ports=&tags=amber&tags=aqua&q=

Django Version: 3.2.16
Python Version: 3.10.9
Installed Applications:
['django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.humanize',
 'cacheops',
 'corsheaders',
 'django_filters',
 'django_jinja',
 'django_tables2',
 'django_prometheus',
 'social_django',
 'taggit',
 'timezone_field',
 'nautobot.core.apps.NautobotConstanceConfig',
 'nautobot.core',
 'django.contrib.admin',
 'django_celery_beat',
 'django_celery_results',
 'rest_framework',
 'db_file_storage',
 'nautobot.circuits',
 'nautobot.dcim',
 'nautobot.ipam',
 'nautobot.extras',
 'nautobot.tenancy',
 'nautobot.users',
 'nautobot.utilities',
 'nautobot.virtualization',
 'drf_spectacular',
 'drf_spectacular_sidecar',
 'graphene_django',
 'health_check',
 'health_check.storage',
 'django_extensions',
 'constance.backends.database',
 'django_ajax_tables',
 'debug_toolbar',
 'example_plugin.ExamplePluginConfig',
 'nautobot.extras.tests.example_plugin_dependency']
Installed Middleware:
['debug_toolbar.middleware.DebugToolbarMiddleware',
 'django_prometheus.middleware.PrometheusBeforeMiddleware',
 'corsheaders.middleware.CorsMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
[...]

PR #3456, opened 2023-03-20
Fixes to TagFilter handling

glennmatthews:
# Closes: #3154 
# What's Changed

- Make `TagFilter` a subclass of `NaturalKeyOrPKMultipleChoiceFilter` instead of a direct subclass of `ModelMultipleChoiceFilter`
   - This fixes the issue reported in #3154, which was due to django-filters explicitly providing a `field_name` when instantiating a filter from `fields = [...]` and therefore creating a `TagFilter` with `field_name="tags"` instead of the desired `field_name="tags__slug"`. Since `NaturalKeyOrPKMultipleChoiceFilter` works with the base `field_name` value, this change in inheritance fixes the issue.
   - As a bonus, you can now filter on tags by UUID as well as by slug!
- Add a `test_tags_filter` generic test to our baseline `FilterTestCase` class, currently gated on whether the filterset model is a `PrimaryModel` or not.
   - The test is similar to `test_filters_generic` but can't use precisely the same setup since Tag filters are conjoined (tag A AND tag B) while the generic filters testing is for non-conjoined (value A OR value B) filters.
   - Convert all Nautobot PrimaryModel FilterSets that weren't already using `fields = [..., "tags"]` to do so
   - Update various `test_filters` that are using non-factory data to ensure that they set some tags on their models. (All of our factory data is already randomly assigned tags so no update is needed there.)
- Changelog fragments and updates to the upgrading document.

# TODO
- [x] Unit, Integration Tests


nautobot/core/filters.py
```diff
@@ -462,6 +445,21 @@ class SearchFilter(MappedPredicatesFilterMixin, django_filters.CharFilter):
...
 
+class TagFilter(NaturalKeyOrPKMultipleChoiceFilter):
+    """
+    Match on one or more assigned tags. If multiple tags are specified (e.g. ?tag=foo&tag=bar), the queryset is filtered
+    to objects matching all tags.
+    """
+
+    def __init__(self, *args, **kwargs):
+        from nautobot.extras import models as extras_models  # avoid circular import
+
```
glennmatthews, line 456: Might need to still `kwargs.setdefault("field_name", "tags")` to keep this working for any apps that are still using the old `tag = TagFilter()` pattern.

nautobot/core/filters.py
```diff
@@ -462,6 +445,21 @@ class SearchFilter(MappedPredicatesFilterMixin, django_filters.CharFilter):
...
 
 
+class TagFilter(NaturalKeyOrPKMultipleChoiceFilter):
+    """
+    Match on one or more assigned tags. If multiple tags are specified (e.g. ?tag=foo&tag=bar), the queryset is filtered
+    to objects matching all tags.
+    """
+
+    def __init__(self, *args, **kwargs):
+        from nautobot.extras import models as extras_models  # avoid circular import
```
timizuoebideri1, line 455: Maybe? as we not using any other class from this module
```suggestion
        from nautobot.extras.models import Tag  # avoid circular import
```

nautobot/ipam/tests/test_filters.py
```diff
@@ -862,42 +862,46 @@ def setUpTestData(cls):
...
+                ports=[2002],
+            ),
+            Service.objects.create(
+                virtual_machine=virtual_machines[2],
+                name="Service 6",
+                protocol=ServiceProtocolChoices.PROTOCOL_UDP,
+                ports=[2003],
+            ),
         )
+        services[0].tags.set(Tag.objects.get_for_model(Service))
```
timizuoebideri1, line 903: Would it be better if assigning of tags to model is done in the generic test function instead of on each test case

glennmatthews: You mean have the test function take responsibility for assigning tags, rather than expecting the tags to be provided by the test case setup? The thing is any models that are using factories already have tags appropriately assigned by the factory, the per-test custom setup is an outlier and should be less and less necessary as we implement more factories.

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -322,49 +323,59 @@ These endpoints `/ipam/roles/`, `/dcim/rack-roles/` and `/dcim/device-roles/` ar
...
 |                         | `local_context_schema_id` | `local_config_context_schema_id` | `/virtualization/virtual-machines/?local_config_context_schema_id=<uuid>` |
 |                         | `local_context_schema`    | `local_config_context_schema`    | `/virtualization/virtual-machines/?local_config_context_schema=<slug>`    |
+|                         | `tag`                     | `tags`                           | `/virtualization/virtual-machines/?tags=<uuid/slug>`                      |
+| VLAN                    | `tag`                     | `tags`                           | `/ipam/vlans/?tags=<uuid/slug>`                                           |
+| VMInterface             | `tag`                     | `tags`                           | `/virtualization/interfaces/?tags=<uuid/slug>`                            |
+| VRF                     | `tag`                     | `tags`                           | `/ipam/vrfs/?tags=<uuid/slug>`                                            |
 
 ### Enhanced Filter Fields
```
HanlinMiao, line 380: Should we document the addition of filtering by `uuids` functionality in this section? It might be a little too much

HanlinMiao (approved):
LGTM!

nautobot/docs/installation/upgrading-from-nautobot-v1.md
glennmatthews: That'd be a set of additions to the following `Enhanced FilterFields` section wouldn't it? I was thinking that given `tags` is a "new" filter rather than an enhancement to the existing `tag` filter that it didn't belong there, but I'm open to convincing otherwise.

nautobot/core/filters.py
glennmatthews:
```suggestion

        kwargs.setdefault("field_name", "tags")
```

nautobot/docs/installation/upgrading-from-nautobot-v1.md
HanlinMiao: I'm good with that

Issue #3457, opened 2023-03-20
Scheduled Job fails to run in next branch


gsnider2195:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0alpha1
* Python version: 3.10
* Database platform, version: postgres 13
* Middleware(s):

Beat scheduler raises an exception with traceback:

```py
nautobot-celery_beat-1    | Traceback (most recent call last):
nautobot-celery_beat-1    |   File "/usr/local/lib/python3.10/site-packages/celery/apps/beat.py", line 105, in start_scheduler
nautobot-celery_beat-1    |     service.start()
nautobot-celery_beat-1    |   File "/usr/local/lib/python3.10/site-packages/celery/beat.py", line 645, in start
nautobot-celery_beat-1    |     interval = self.scheduler.tick()
nautobot-celery_beat-1    |   File "/usr/local/lib/python3.10/site-packages/celery/beat.py", line 341, in tick
nautobot-celery_beat-1    |     not self.schedules_equal(self.old_schedulers, self.schedule)):
nautobot-celery_beat-1    |   File "/usr/local/lib/python3.10/site-packages/django_celery_beat/schedulers.py", line 361, in schedule
nautobot-celery_beat-1    |     self._schedule = self.all_as_schedule()
nautobot-celery_beat-1    |   File "/usr/local/lib/python3.10/site-packages/django_celery_beat/schedulers.py", line 247, in all_as_schedule
nautobot-celery_beat-1    |     s[model.name] = self.Entry(model, app=self.app)
nautobot-celery_beat-1    |   File "/source/nautobot/core/celery/schedulers.py", line 33, in __init__
nautobot-celery_beat-1    |     self.schedule = model.scheduled_job
nautobot-celery_beat-1    | AttributeError: 'ScheduledJob' object has no attribute 'scheduled_job'
```

gsnider2195:
Fixed by #3459

PR #3459, opened 2023-03-20
fix bug preventing scheduled job from running

gsnider2195:
# Closes: #3457
# What's Changed

A typo in `core/celery/schedulers.py` was causing an exception to be raised when beat would try to run a scheduled job

# TODO

### DIFF
--- a/changes/3457.fixed
+++ b/changes/3457.fixed
@@ -1,0 +1,1 @@
+Fixed bug preventing scheduled job from running.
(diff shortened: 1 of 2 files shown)


HanlinMiao (approved):

glennmatthews (approved):
Fix looks good. Wonder if there's any way we could write a basic unit/integration test for scheduling jobs to catch issues like this in the future?

Issue #3206, opened 2023-01-31
Pre-Release Github Action Workflow Doesn't Publish Python Version Specific Builds


bryanculver:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0-alpha.1
* Python version: 3.7-3.10
* Database platform, version: NA
* Middleware(s): NA

### Steps to Reproduce
1. Create Pre-Release Release named `2.0.0-alpha.1`
2. Wait for CI to Run

### Expected Behavior

Docker containers built following or normal release convention `2.0.0-alpha.1` as well as `2.0.0-alpha.1-py3.10`, `2.0.0-alpha.1-py3.9`, etc.

### Observed Behavior

Whichever container is published last "wins" and becomes the `2.0.0-alpha.1` build: https://hub.docker.com/layers/networktocode/nautobot/2.0.0-alpha.1/images/sha256-6f41c067abac092cbed73b5b6176f771e8432e6cbd380b8d5f9013ba521d87a6?context=explore

bryanculver:
Should investigate maybe a couple other CI cleanups such as this resolved issue: https://github.com/nautobot/nautobot/blob/1d5588f83e36498bcb7757aad18b9dbae0900e3d/.github/workflows/ci_integration.yml#L178-L181

PR #3475, opened 2023-03-23
Fix Docker Prerelease Tag Syntax

bryanculver:
# Closes: #3206
# What's Changed
**Changed to use `+` instead of `-` for Python version signifier when using the `semver` format tag**

Previously it was attempting `2.0.0-alpha.1-py3.7` which isn't valid [SemVer](https://semver.org). This label would have been reject and was confirmed in testing. In fact, our current syntax for our stable builds is incorrect. BNF on the SemVer site states that `-alphanum.num` is for pre-release signifiers. `+alphanum.num` is for build numbers which is more applicable for our Python-specific build signifier. We probably should publish `1.5.14+py3.7`. 

**Backup `raw` tag to take the literal value that is passed in, assuming something goes awry with the approach above.**

Suffixed with `r` _just_ in case `docker/build-push-action` or `docker/metadata-action` get sad with two tags that result in the same value.

Also:
- Removed the non-Python-version specific tag as it made discovery a bit opaque as to which tag was failing
- Didn't address https://github.com/nautobot/nautobot/issues/3206#issuecomment-1420663909 because it seems some are still experiencing `dist-info` problems: https://github.com/python-poetry/poetry/issues/6815

# TODO

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -86,6 +86,7 @@ jobs:
             latest=false
           tags: |
             type=semver,pattern={{version}}+py${{ matrix.python-version }}
+            type=semver,pattern={{version}},enable=${{ matrix.python-version == 3.7 }}
             type=raw,value={{version}}-py${{ matrix.python-version }}r
           labels: |
             org.opencontainers.image.title=Nautobot
@@ -113,6 +114,7 @@ jobs:
             latest=false
           tags: |
             type=semver,pattern={{version}}+py${{ matrix.python-version }}
+            type=semver,pattern={{version}},enable=${{ matrix.python-version == 3.7 }}
             type=raw,value={{version}}-py${{ matrix.python-version }}r
           labels: |
             org.opencontainers.image.title=Nautobot
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -85,8 +85,8 @@ jobs:
           flavor: |
             latest=false
           tags: |
-            type=semver,pattern={{version}}-py${{ matrix.python-version }}
-            type=semver,pattern={{version}},enable=${{ matrix.python-version == 3.7 }}
+            type=semver,pattern={{version}}+py${{ matrix.python-version }}
+            type=raw,value={{version}}-py${{ matrix.python-version }}r
           labels: |
             org.opencontainers.image.title=Nautobot
       - name: "Build"
@@ -112,8 +112,8 @@ jobs:
           flavor: |
             latest=false
           tags: |
-            type=semver,pattern={{version}}-py${{ matrix.python-version }}
-            type=semver,pattern={{version}},enable=${{ matrix.python-version == 3.7 }}
+            type=semver,pattern={{version}}+py${{ matrix.python-version }}
+            type=raw,value={{version}}-py${{ matrix.python-version }}r
           labels: |
             org.opencontainers.image.title=Nautobot
       - name: "Build Dev Containers"
--- a/changes/3206.fixed
+++ b/changes/3206.fixed
@@ -1,0 +1,1 @@
+Fixed Docker tag syntax on prerelease workflow.


jathanism:
Seems to me we should fix this in our local builds and in the development containers, too?

.github/workflows/prerelease.yml
```diff
@@ -85,8 +85,8 @@ jobs:
           flavor: |
             latest=false
           tags: |
-            type=semver,pattern={{version}}-py${{ matrix.python-version }}
-            type=semver,pattern={{version}},enable=${{ matrix.python-version == 3.7 }}
+            type=semver,pattern={{version}}+py${{ matrix.python-version }}
+            type=raw,value={{version}}-py${{ matrix.python-version }}r
```
glennmatthews, line 89: Having a hard time parsing this. So before, we would tag as?

- v2.0.0-alpha.1-py3.7
- v2.0.0-alpha.1     (python 3.7 version)
- v2.0.0-alpha.1-py3.8
- v2.0.0-alpha.1-py3.9
- v2.0.0-alpha.1-py3.10

and now we'll tag as:

- v2.0.0-alpha.2+py3.7
- v2.0.0-alpha.2-py3.7r
- (no default "v2.0.0-alpha.2" image)
- v2.0.0-alpha.2+py3.8
- v2.0.0-alpha.2-py3.8r
- v2.0.0-alpha.2+py3.9
- v2.0.0-alpha.2-py3.9r
- v2.0.0-alpha.2+py3.10
- v2.0.0-alpha.2-py3.10r

Is that right?

glennmatthews (commented):
Shouldn't we be making a similar change to the `release.yml`?

bryanculver:
> Shouldn't we be making a similar change to the `release.yml`?

This is meant to address this specific issue. We don't have these problems in `release.yml` because while not _technically_ correct, `1.5.15-py3.7` is valid enough of SemVer per their BNF.

Long-term I think we should discuss moving to `1.5.15+py3.7` but now's not the time to make that significant change for people just to be pedantic.

If we ever shipped a `1.5.10-hotfix.1` _then_ we'd incur a problem with `release.yml`.

.github/workflows/prerelease.yml
bryanculver: > * v2.0.0-alpha.1-py3.7
> * v2.0.0-alpha.1     (python 3.7 version)
> * v2.0.0-alpha.1-py3.8
> * v2.0.0-alpha.1-py3.9
> * v2.0.0-alpha.1-py3.10

We would _attempt_ this but the Docker metadata task would reject the `v2.0.0-alpha.1-py3.X`s because they aren't valid SemVer.

Instead, they all resulted in `v2.0.0-alpha.1`, see: https://github.com/nautobot/nautobot/actions/runs/4057206803/jobs/6983617197#step:8:45

> and now we'll tag as:
> 
> * v2.0.0-alpha.2+py3.7
> * v2.0.0-alpha.2-py3.7r
> * (no default "v2.0.0-alpha.2" image)
> * v2.0.0-alpha.2+py3.8
> * v2.0.0-alpha.2-py3.8r
> * v2.0.0-alpha.2+py3.9
> * v2.0.0-alpha.2-py3.9r
> * v2.0.0-alpha.2+py3.10
> * v2.0.0-alpha.2-py3.10r
> 
> Is that right?

Yes. Because this really isn't easily testable without attempting a release I feel this current syntax gives us the best option for the pre-release workflow not to fail. If for some reason the Docker metadata task doesn't like the `+` annotations, a "raw" tag of `v2.0.0-alpha.2-py3.7r` at least ensures _something_ will be tagged.

glennmatthews: Thanks for the info. I'm not sure I see the benefit/purpose of removing the default (non- `py3.x`) tag but the rest of it makes sense to me.

bryanculver: > Thanks for the info. I'm not sure I see the benefit/purpose of removing the default (non- `py3.x`) tag but the rest of it makes sense to me.

Reverted the removal.

glennmatthews (approved):

jathanism (approved):

Issue #2806, opened 2022-11-11
Change Shadowed Fields on Existing Core Filter Sets


bryanculver:
Adapted from https://gist.github.com/jathanism/7c585954da4cb1a54997be31991cdaac#coreutilities --

## Core/Utilities

Utilities are being included with core since `nautobot.utilities.filters.BaseModelSet` is the bread and butter here and we may very well end up folding all of the primitives from Utilities into Core to streamline the namespace.

Items that have been moved to their own story:
### BaseFilterSet

- The model field to filter field mapping at `BaseFilterSet.FILTER_DEFAULTS` needs to also include the correct UI widget types in the `extra` for the filter mapping
    - #3582
    - #3583
- Filter mappings for ALL bespoke field types should be extended
    - For example `TaggableManager` should always be mapped to `TagFilter` to obviate the need to define `tag`/`tags` on every new filterset that requires tags. (but see also #3154)
    - #3584
- ~The base form used should be included by default on the [`Meta.form`](https://django-filter.readthedocs.io/en/stable/ref/filterset.html#form) option. This base form should be what is used as the base form for our UI including its own [model field to form field mappings](https://docs.djangoproject.com/en/4.0/topics/forms/modelforms/#overriding-the-default-fields). (See `field_classes` on that page).~
    - Most to be done with the JSON schema representation but we may need to somehow provide a way when the schema isn't good enough to specifically provide the react form that represents that model (either by convention in the React packaging or by configuration in the model). Let's start with the schema first.
    - Let's see how much `.form` is used anymore with the new UI
- #3585
- #3587
- #3588

--------

Review:

https://gist.github.com/jathanism/7c585954da4cb1a54997be31991cdaac#extras

for fields to be changed under "Incorrect Fields"

**NOTE:** Let's do this after #2721 ~and #2722~

bryanculver:
Blocked by #2721 and #2722

glennmatthews:
Since we decided not to collapse extras into core for the time being, should this be split into two issues?

bryanculver:
Done because bullet points were either already done or opened as separate issue, many of which are enhancements and not breaking changes.

💪

Issue #1623, opened 2022-04-08
Multiple `q` parameters in query do not use logical OR


david-kn:
### Environment
* Python version:
* Nautobot version: 1.2.11

### Steps to Reproduce
1. Search for a device / send a request with multiple `q` parameters, e.g. `dcim/devices/?q=esw&q=msw`

### Expected Behavior
Results should contain items related to all passed parameters - logical operator OR should be used between the parameters (this approach is used for other type of parameters, e.g. `status`. Query  `dcim/devices/?status=active&status=decommissioning` returns object which are either `active` OR `decommissioning `

### Observed Behavior
Results contains only items related to the latest `q` parameter in the string (in this case _msw_)

itdependsnetworks:
I think this can be avoided with the introduction of https://github.com/nautobot/nautobot/pull/1496. Otherwise, I am not sure this can work as you expect it since 

- We must choose one (either an `AND` or an `OR`)
- It has been an `AND` and changing it would be a breaking change
- This is the logic provided by django-filters, and would have to be applied there
    - https://django-filter.readthedocs.io/en/latest/ref/filters.html#multiplechoicefilter
    - https://django-filter.readthedocs.io/en/latest/ref/filters.html#method

In the end, I think your best bet would be to use the features applied in #1496 or open an issue on the [django-filter](https://github.com/carltongibson/django-filter) project.

david-kn:
Hi, you confused me a bit :-) If I get you correctly, you say that for different parameters (types) you choose different logical operator and for `q` you set `AND`?

- Because for `status` field (which is `multiplechoicefilter` ) it uses logical `OR`. If I stick with my example from previous post, either `active `or `decomissioning` devices are returned. No `AND` logic is applied.

itdependsnetworks:
> If I get you correctly, you say that for different parameters (types) you choose different logical operator and for q you set AND

Mostly correct, but the "you", is not clear in what you are describing. The `you == django-filter` in your statement, and they have an opinion on how it should work--which I agree with.

The current `multiplechoicefilter` is a logical `OR`, which I think we can agree makes sense, however with the the `q` logic (or method), you cannot guarantee you are looking for a specific field and thus cannot guarantee this will always be your intention. Take for example the following 

```
        qs_filter = (
            Q(name__icontains=value)
            | Q(facility__icontains=value)
            | Q(description__icontains=value)
            | Q(physical_address__icontains=value)
            | Q(shipping_address__icontains=value)
[...]

PR #3479, opened 2023-03-23
Core filter code fixes

glennmatthews:
# Closes: N/A (partial progress toward #2806)
# What's Changed

- Enhance `MultiValueFieldFactory` to allow the caller to provide an appropriate widget to override the default `SelectMultiple` widget. (In a future PR we might want to change the default, as it's rarely correct, or perhaps even make this a mandatory parameter.)
  - Provide the appropriate widgets for `MultiValue(UUID/Date/DateTime/Time)Filter` and `RelatedMembershipBooleanFilter`. (Note for the date/time filters, we don't currently have an appropriate multi-input widget, but the single-input widgets we do have are still more correct than before)
  - This allows us to remove some of the code from `get_filterset_parameter_form_field()` that was needed to provide the correct widgets to override the filters' default widgets, now that the default widgets for these are more correct.
  - Future PR(s) should be able to continue along these lines.
- Similarly fix the widgets that are provided by default for custom-field filters and remove the override from `get_filterset_parameter_form_field()`
- Change `CreatedUpdatedModelFilterSetMixin` to provide multi-value filters for `created` and `updated`.
   - [ ] I just realized this is probably going to fail tests without some updates there. Will fix.
   - This required a few changes in `BaseFilterSet` because it wasn't previously correctly handling the case of `MultiValue*Filter` being present on an "abstract" class like `CreatedUpdatedModelFilterSetMixin`.
- Change `VLAN.available_on_device` filter to be a multi-value UUID filter
- In testing the various above fixes I noticed that some aspects of the code that are designed to keep the "basic" and "advanced/dynamic" filter form contents in sync were not working correctly in some cases, mostly due to some of the logic being applied on page load only and not when dynamically refreshing the dynamic filter form. To fix this, I ended up needing to consolidate the JavaScript that was previously split across `object_list.html` and `forms.js` into a single location in `forms.js`. 

image

# TODO
- [ ] Unit, Integration Tests


nautobot/core/tests/test_forms.py
```diff
@@ -868,9 +868,9 @@ def test_dynamic_filter_form_with_data_and_prefix(self):
                 },
             )
 
-        with self.subTest("Test for lookup_value with a ChoiceField and StaticSelect2 widget"):
-            # If `lookup_field` value is a ChoiceField and `lookup_type` lookup expr is `exact` or `in` then,
-            # `lookup_value` field should be a ChoiceField with StaticSelect2 widget
+        with self.subTest("Test for lookup_value with a NullBooleanField and StaticSelect2 widget"):
+            # If `lookup_field` value is a boolean filter and `lookup_type` lookup expr is `exact`, then
```
HanlinMiao, line 872: `lookup_field` is a `boolean field `perhaps?

nautobot/core/tests/test_utils.py
```diff
@@ -373,17 +373,20 @@ def test_slugify_dashes_to_underscores(self):
...
 
-        single_choice_fields = ("created", "created__gte", "has_vlans", "has_clusters", "q")
+        single_choice_fields = ("has_vlans", "has_clusters")
         for field in single_choice_fields:
-            self.assertTrue(requests.is_single_choice_field(filterset_class, field))
+            with self.subTest(f"Single choice field: {field}"):
+                self.assertTrue(requests.is_single_choice_field(filterset_class, field))
 
-        multi_choice_fields = ("status", "tenant", "tags")
+        multi_choice_fields = ("created", "status", "tenant", "tags")
```
HanlinMiao, line 386: Why are we not testing `created__gte` and `q` fields anymore?

HanlinMiao (approved):
🚀

nautobot/core/tests/test_utils.py
glennmatthews: I removed `q` because of #1623 (it probably *shouldn't* be a single-value filter), and `created__gte` is no longer a special-case filter.

Issue #3228, opened 2023-02-04
Jobs Overhaul: Replace `run_job` with `NautobotTask`


jathanism:
- Still need to eliminate `run_job` and move the stuff to `NautobotTask` handler methods
    - Anywhere explicit status changes are being made must go away. We want Celery to make these decisions, along with any explicit calls to `JobResult.save()`, moving those to the `NautobotTask` class or the backend or a mix of both, depending.
    - Review/evaluate `nautobot.extras.datasources.git.log_job_result_final_status` - Can likely be deleted
    - Remove explicit `set_state/save` code from `nautobot.extras.datasources.git.pull_git_repository_and_refresh_data`
    - Review `nautobot.extras.datasources.registry.refresh_datasource_content` to remove explicit `set_state/save` code
- Eliminate the atomic block (#2726) while we're at it.
- Emphasize evolving to the `run()` method being a Celery task once `run_job` is gone, whether we tell users they must decorate it or do so in a metaclass, OR possible even just making the `BaseJob` class a subclass of `NautobotTask` and only registering it in Celery when `enabled=True`. Something like that.

The goal here is to replace `run_job` in exchange for v2 task subclasses. Ideally, v2 `BaseJob` replacement would be a subclass of `NautobotTask` where the `run()` method is still defined. See: https://docs.celeryq.dev/en/stable/reference/celery.app.task.html#celery.app.task.Task.run

```[tasklist]
### TODO
- [x] Move `run_job` Logic to `nautobot.extras.jobs.BaseJob` in `before_start` and `after_return` methods
- [x] Register jobs with Celery app directly (instead of via `@nautobot_task` on `run_job`)
- [x] Update `enqueue_job` to call job directly, instead of `run_job`, to accept `JobModel`
- [x] Don't copy over atomic block handler (#2726)
- [x] Update places (Jobs specifically, not Git Repo sync, et all) that call `enqueue_job` for new signature
```

jathanism:
Deep Thoughts™ by Jathan McCollum:

- Basically everyhting in `run_job` prior to inner  `_run_job` would go into the `NautobotTask.before_start` method (See: https://docs.celeryq.dev/en/stable/userguide/tasks.html#before_start)
- And then `Job.post_run` if we want to keep that would basically just be an analog for `after_return` (See: https://docs.celeryq.dev/en/stable/userguide/tasks.html#after_return)

gsnider2195:
[...]

Issue #3453, opened 2023-03-20
Jobs Overhaul: Re-architect `dryrun` Support


gsnider2195:
To support #2726 removing the atomic transaction in `run_job`.

- "dryrun" logic moves to the job but needs to be a documented pattern supported by core so that approval can be bypassed on jobs with "dryrun" mode selected
- Jobs should manually support/opt-in `dryrun` support by advertising as a BooleanVar

gsnider2195:
blocked by #3228

bryanculver:
Closed with #3593

Issue #3229, opened 2023-02-04
Jobs Overhaul: Refactor `JobResult.result` to `JSONField` from `TextField`


jathanism:

jathanism:
IMO we should also remove the `JobResult.data` field while we're doing this, since that field is no longer used. In practice, instead of adding `JobResult.result` and changing it from text to JSON field, it would be cleaner to just rename `data` to `result`.

bryanculver:
Closed with #4133

Issue #3089, opened 2023-01-05
Jobs Overhaul: Documentation updates for database backend


jathanism:
- [ ] Document how to migrate an existing job to 2.0.
- [ ] Document how to implement dry-run support in a new or existing job (including but not limited to `transaction.atomic()`)
- [ ] Document how to implement equivalent functionality to the 1.x `post_run()` in a job.
- [ ] Add an entry to the 2.0 Release Overview docs summarizing the Job changes.
- [ ] Make sure to call out that `Job.after_return()` MUST be called in order for chords to work

bryanculver:
To be done in #3085

jathanism:
Reopened because the documentation did not get done in #3085 and still needs to be completed.

bryanculver:
Documentation tasks moved up to a task list in #765

PR #3483, opened 2023-03-24
Remove run_job

gsnider2195:
# Closes: #3228, #2979, #2726
# What's Changed

- Moved `run_job` logic to `nautobot.extras.jobs.BaseJob` in `before_start` and `after_return` methods
  - As a result `request`, `commit` and `job_result_pk` are no longer required to run jobs
- Removed backwards compatibility support for netbox scripts and reports
- Register jobs with Celery app directly (instead of via @nautobot_task on run_job)
- Simplified the `JobResult.enqueue_job` signature to `job_model`, `user`, `schedule`, `celery_kwargs`, `job_args` and `job_kwargs`
  - Removed `func`, `name` and `obj_type`. These are now derived from `job_model`
- Removed atomic block handler (https://github.com/nautobot/nautobot/issues/2726)
- Updated places (Jobs specifically, not Git Repo sync, et all) that call enqueue_job for new signature
- Updated runjob management command
- Removed some legacy job features that are no longer supported in 2.0
  - Legacy `/api/extras/jobs/{class_path}/` Rest API endpoints
  - Job `test_` and `post_run` method support
  - Fallback code for jobs that don't have a Job model in the database
- Removed `requires_input` property from Job form. Job forms will always have to display the `task queue` dropdown so this is no longer used.
- Added ability to chain jobs using standard celery primitives:
```py
from celery import chain

addjob=Job.objects.get(name="AddJob").job_task
user_pk=User.objects.first().pk
c=chain(addjob.s(x=1, y=2).set(user_id=user_pk), addjob.s(y=6).set(user_id=user_pk)).apply_async()
c.get()
```

## Follow-up work

This won't be ready to merge in to `next` until the feature complete issues are done. Merge all subsequent PRs into the feature branch [feature/765-jobs-overhaul](https://github.com/nautobot/nautobot/tree/feature/765-jobs-overhaul):

- [x] Figure out why celery thinks `ExampleLoggingJob` `interval` kwarg is a positional arg
  - Can't reproduce. I was probably doing something wrong 
- [x] Figure out how to use the `.apply_async()` method to store the user on the jobresult instead of having to use `enqueue_job`
  - It looks like this is the pattern to set celery kwargs on chained jobs: `chain(task1.s(arg).set(queue='queuename'), task2.s().set(queue='queuename')).apply_async()`
  - Resolved in 7b68da050b92d7da3e2aeaf3a96c2d7cb7a8c6a1
- [ ] #3229
- [ ] #3087
- [ ] #3088 
- [ ] #3089
- [ ] #3453
- [ ] #2853

# TODO

- [ ] Unit, Integration Tests
- [ ] Change logging
- [ ] Remove job logs database?


nautobot/extras/api/views.py
```diff
@@ -971,18 +889,17 @@ def dry_run(self, request, pk):
...
             raise PermissionDenied("You do not have permission to run this job.")
 
-        # Immediately enqueue the job with commit=False
-        job_content_type = get_job_content_type()
+        # Immediately enqueue the job
+        job_kwargs = scheduled_job.kwargs.get("data", {})
+        for key in list(job_kwargs.keys()):
+            attr = getattr(job_model.job_class, key, None)
+            if not isinstance(attr, ScriptVariable):
+                job_kwargs.pop(key)
```
gsnider2195, line 897: Dryrun/Approval will be solved in #3453

gsnider2195:
Marking this as ready for review so the changes so far can be reviewed. This won't be merged into `next` until all the follow-on work is completed (separate pull requests into this branch)

nautobot/core/settings.py
```diff
@@ -688,9 +688,6 @@
 # A value of None or 0 means results will never expire (depending on backend specifications).
 CELERY_RESULT_EXPIRES = None
 
-# If set to True, result messages will be persistent. This means the messages won’t be lost after a broker restart.
-CELERY_RESULT_PERSISTENT = True
```
gsnider2195, line 692: This only applies to the [rpc backend](https://docs.celeryq.dev/en/stable/userguide/configuration.html#std-setting-result_persistent) which Nautobot does not use

nautobot/extras/management/commands/runjob.py
```diff
@@ -71,37 +54,30 @@ def handle(self, *args, **options):
...
-                job_result_pk=job_result.pk,
-            )
+            eager_result = job.job_task.apply(kwargs=data, task_id=str(job_result.task_id))
             job_result.refresh_from_db()
+            job_result.date_done = timezone.now()
+            job_result.status = eager_result.status
+            job_result.result = eager_result.result
+            job_result.traceback = eager_result.traceback
+            job_result.worker = eager_result.worker
+            job_result.save()
```
jathanism, line 77: As commented above, perhaps factor this code into a `JobResult` method?

nautobot/extras/jobs.py
```diff
@@ -104,25 +105,185 @@ class Meta:
...
+        """
+
+        # Cleanup FileProxy objects
+        file_fields = list(self._get_file_vars())
+        file_ids = [kwargs[f] for f in file_fields]
+        if file_ids:
+            self.delete_files(*file_ids)
+
+        self.log_info("Job completed")
+        super().after_return(status, retval, task_id, args, kwargs, einfo=einfo)
```
gsnider2195, line 285: https://docs.celeryq.dev/en/stable/userguide/canvas.html#important-notes

> Note
> 
> If you’re using chords with the Redis result backend and also overriding the Task.after_return() method, you need to make sure to call the super method or else the chord callback won’t be applied.
```

File: examples/example_plugin/example_plugin/jobs.py [hunk 5]
```diff
@@ -130,3 +130,4 @@ def receive_job_button(self, obj):
     ExampleSimpleJobButtonReceiver,
     ExampleComplexJobButtonReceiver,
 )
+register_jobs(*jobs)
```
jathanism, line 133: I think it would good to align the patterns here and in files dropped into `JOBS_ROOT`.

For example if this `jobs` attribute were to just call `register_jobs()` on them internally, then the same pattern used in an App's `jobs.py` or a Git repo, would now be asserted in file-based jobs as well.

The only other consideration would be to obviate this entirely and use a meta class on `BaseJob` to automatically register any subclass into the Celery task registry.

nautobot/extras/tests/example_jobs/test_location_with_custom_field.py
```diff
@@ -11,25 +13,29 @@ class Meta:
         name = "Location and Custom Field Creation"
         description = "Location with a custom field"
 
-    def run(self, data, commit):
-        obj_type = ContentType.objects.get_for_model(Location)
-        cf = CustomField(label="cf1", type=CustomFieldTypeChoices.TYPE_TEXT, default="-")
-        cf.validated_save()
-        cf.content_types.set([obj_type])
+    def run(self):
+        with transaction.atomic():
```
jathanism, line 17: Seeing as you can also decorate a method like so:

```python
@transaction.atomic
def run(self):
    # ...
```

We should probably test that works, too.

jathanism (commented):
Some initial comments/thoughts while I continue to review the bigger stuff.

nautobot/extras/jobs.py
```diff
@@ -104,25 +105,185 @@ class Meta:
...
         self.failed = False
         self._job_result = None
 
-        # Compile test methods and initialize results skeleton
-        self.test_methods = []
-
-        for method_name in dir(self):
-            if method_name.startswith("test_") and callable(getattr(self, method_name)):
-                self.test_methods.append(method_name)
+    def __call__(self, *args, **kwargs):
```
jathanism, line 114: If we're not even processing the `*args` I'm wondering if perhaps instead we should put `*, **kwargs` here to explicitly disallow positional arguments?
```suggestion
    def __call__(self, *, **kwargs):
```

nautobot/extras/jobs.py
```diff
@@ -104,25 +105,185 @@ class Meta:
...
+        # Attempt to resolve serialized data back into original form by creating querysets or model instances
+        # If we fail to find any objects, we consider this a job execution error, and fail.
+        # This might happen when a job sits on the queue for a while (i.e. scheduled) and data has changed
+        # or it might be bad input from an API request, or manual execution.
+        try:
+            deserialized_kwargs = self.deserialize_data(kwargs)
+        except Exception:
+            stacktrace = traceback.format_exc()
+            self.log_failure(f"Error initializing job:\n```\n{stacktrace}\n```")
+        self.active_test = "run"
```
jathanism, line 124: We might consider ditching `active_test` since it's only used in logging? Maybe we could just use the existing task state at the time of logging?

nautobot/extras/jobs.py
```diff
@@ -278,31 +441,10 @@ def _get_file_vars(cls):
 
     @property
     def job_result(self):
+        if self._job_result is None:
+            self._job_result = self.get_job_result()
```
jathanism, line 445: Make this a `@cached_property` and then you can just `return self.get_job_result()` here without having to explicitly save it on the instance.

nautobot/extras/jobs.py
```diff
@@ -104,25 +105,185 @@ class Meta:
...
+            kwargs (Dict): Original keyword arguments for the task to execute.
+
+        Returns:
+            None: The return value of this handler is ignored.
+        """
+        self.active_test = "initialization"
+        self._job_result = None
+
+        try:
+            self.get_job_result()
```
jathanism, line 186: If you make the suggested change to `self.job_result` then you could just attempt to access the attribute here instead for the same result perhaps?

nautobot/extras/jobs.py
```diff
@@ -104,25 +105,185 @@ class Meta:
...
+            )
+
+        soft_time_limit = job_model.soft_time_limit or settings.CELERY_TASK_SOFT_TIME_LIMIT
+        time_limit = job_model.time_limit or settings.CELERY_TASK_TIME_LIMIT
+        if time_limit <= soft_time_limit:
+            self.log_warning(
+                f"The hard time limit of {time_limit} seconds is less than "
+                f"or equal to the soft time limit of {soft_time_limit} seconds. "
+                f"This job will fail silently after {time_limit} seconds.",
+            )
```
jathanism, line 206: I feel like this is something that should be done at `JobResult.enqueue_job()` time and not at runtime? Or perhaps even have it as a part of `JobResult` model validation?

examples/example_plugin/example_plugin/jobs.py
gsnider2195: I think we should align on `register_jobs()` and use the celery task registry instead of the `jobs` variable to register app jobs in the database. This tracks with our earlier discussion on standardizing the natural keys for the job database model and celery task registry.

nautobot/extras/jobs.py
gsnider2195: I don't think we'll be able to align on only `kwargs` since celery task chains rely on passing the result of one task as the first positional arg of the next task. That would also be invalid syntax in this context:

```
nautobot-celery_worker-1  |   File "/source/nautobot/extras/jobs.py", line 114
nautobot-celery_worker-1  |     def __call__(self, *, **kwargs):
nautobot-celery_worker-1  |                        ^
nautobot-celery_worker-1  | SyntaxError: named arguments must follow bare *
```

nautobot/extras/jobs.py
jathanism: This is problematic if we are telling authors to overload this if they need to. Let's make sure this is documented. Also, this appears to have dropped off the docstring:

```
        Returns:
            None: The return value of this handler is ignored.
```

nautobot/extras/models/jobs.py
```diff
@@ -739,68 +743,90 @@ def set_status(self, status):
                 )
 
     @classmethod
-    def enqueue_job(cls, func, name, obj_type, user, *args, celery_kwargs=None, schedule=None, **kwargs):
+    def run_job_synchronously(cls, job_model, user, *job_args, celery_kwargs=None, **job_kwargs):
```
jathanism, line 746: Can we call this `apply` (or `apply_job`) to align with industry terminology?

jathanism (changes requested):
I still have some outstanding comments from the previous review that I would like to see addressed. And some new ones too... :)

nautobot/extras/views.py
```diff
@@ -1336,21 +1322,18 @@ def post(self, request, pk):
...
-                # Immediately enqueue the job with commit=False and send the user to the normal JobResult view
-                job_content_type = get_job_content_type()
-                initial = scheduled_job.kwargs.get("data", {})
-                initial["_commit"] = False
+                # Immediately enqueue the job and send the user to the normal JobResult view
+                job_kwargs = scheduled_job.kwargs.get("data", {})
+                for key in list(job_kwargs.keys()):
+                    attr = getattr(job_model.job_class, key, None)
+                    if not isinstance(attr, ScriptVariable):
+                        job_kwargs.pop(key)
```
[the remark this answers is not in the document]
gsnider2195: we already have `BaseJob._get_vars()`. What about

```py
job_kwargs = {k: v for k, v in job_form.cleaned_data.items() if k in job_model.job_class._get_vars()}
```

nautobot/extras/jobs.py
gsnider2195: This will be covered by https://github.com/nautobot/nautobot/issues/2331

nautobot/extras/models/jobs.py
```diff
@@ -739,68 +743,44 @@ def set_status(self, status):
...
-                        job_result.task_kwargs = job_result_kwargs
-                        job_result.save()
-                else:
-                    logger.error("Neither a Job database record nor a Job source class were found for %s", name)
+        if job_model.soft_time_limit > 0:
+            celery_kwargs["soft_time_limit"] = job_model.soft_time_limit
+        if job_model.time_limit > 0:
+            celery_kwargs["time_limit"] = job_model.time_limit
+        job_result.job_model = job_model
+        job_result.save()
```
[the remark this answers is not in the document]
gsnider2195: Maybe it should be on the `Job` model validation?

nautobot/extras/views.py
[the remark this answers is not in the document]
jathanism: I mean, it's concise but arguably less readable. Even if that were also another helper, IMO that would help.

nautobot/extras/jobs.py
jathanism: Right, so just replace it with the getter and throw a party.

nautobot/core/celery/backends.py
```diff
@@ -10,18 +10,9 @@ class NautobotDatabaseBackend(DatabaseBackend):
...
     TaskModel = JobResult
 
-    def encode_content(self, data):
-        """Pass through encoding since we're storing as JSON explicitly."""
-        return "application/x-nautobot-json", "utf-8", data
-
-    def decode_content(self, obj, content):
-        """Pass through decoding since we're storing as JSON explicitly."""
-        return content
-
```
gsnider2195, line 20: Removed since we're handling `task_args` and `task_kwargs` below. We may need to bypass again if we implement #3229 but we'll need to see if this negatively impacts storing the group result when running advanced workflows.

gsnider2195: We need to define `encode_content` and `decode_content` because the default implementation tries to access `content_encoding` on the `TaskResult` but we're throwing that away on the `JobResult`

nautobot/extras/models/jobs.py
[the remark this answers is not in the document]
gsnider2195: Adding this to the list of follow-on work

changes/3228.removed
```diff
@@ -0,0 +1,2 @@
+Removed atomic transaction from Jobs.
+Removed support for test_* and post_run methods in Jobs.
```
glennmatthews, line 2:
```suggestion
Removed support for `test_*` and `post_run` methods in Jobs.
```

nautobot/core/celery/__init__.py
```diff
@@ -25,7 +28,22 @@
...
 # nautobot.setup()
 
-app = Celery("nautobot", task_cls="nautobot.core.celery.task:NautobotTask")
+
+class NautobotCelery(Celery):
+    task_cls = "nautobot.core.celery.task:NautobotTask"
+
+    def register_task(self, task, **options):
+        """Override the default task name for job classes to allow app provided jobs to use the full module path."""
+        from nautobot.extras.jobs import Job
```
glennmatthews, line 37: Should now be a time to move the `Job` class (not the `Job` database model) from `extras` to `core`? Would that simplify things?

nautobot/core/testing/__init__.py
```diff
@@ -44,63 +39,33 @@
...
     Provide a common interface to run Nautobot jobs as part of unit tests.
 
-    The Celery task result will be stored at the `JobResult.celery_result` attribute.
-
     Args:
       job (Job): Job model instance (not Job class) to run
-      data (dict): Input data values for any Job variables.
-      commit (bool): Whether to commit changes to the database or rollback when done.
       username (str): Username of existing or to-be-created User account to own the JobResult. Ignored if `request.user`
         exists.
```
glennmatthews, line 49:
```suggestion
      username (str): Username of existing or to-be-created User account to own the JobResult.
```

nautobot/docs/additional-features/jobs.md
```diff
@@ -11,8 +11,11 @@ Jobs are a way for users to execute custom logic on demand from within the Nauto
 
 ...and so on. Jobs are Python code and exist outside of the official Nautobot code base, so they can be updated and changed without interfering with the core Nautobot installation. And because they're completely customizable, there's practically no limit to what a job can accomplish.
 
++/- 2.0.0
+    Backwards compatibility with Netbox scripts and reports has been removed.
+
 !!! note
-    Jobs unify and supersede the functionality previously provided in NetBox by "custom scripts" and "reports". Jobs are backwards-compatible for now with the `Script` and `Report` class APIs, but you are urged to move to the new `Job` class API described below. Jobs may be optionally marked as [read-only](#read_only) which equates to the `Report` functionally, but in all cases, user input is supported via [job variables](#variables).
+    Jobs unify and supersede the functionality previously provided in NetBox by "custom scripts" and "reports". Jobs may be optionally marked as [read-only](#read_only) which equates to the `Report` functionally, but in all cases, user input is supported via [job variables](#variables).
```
glennmatthews, line 18: Are we getting rid of `read_only`?

nautobot/extras/api/views.py
```diff
@@ -535,22 +530,16 @@ def _run_job(request, job_model, legacy_response=False):
...
         raise ValidationError({"task_queue": [f'"{task_queue}" is not a valid choice.']})
 
     cleaned_data = None
     try:
-        cleaned_data = job.validate_data(data, files=files)
-        cleaned_data.pop(
-            "_commit", None
-        )  # We don't get commit from the form, instead it's part of the serializer's validated data
+        cleaned_data = job_class().validate_data(data, files=files)
+        cleaned_data = job_model.job_class.prepare_job_kwargs(cleaned_data)
```
glennmatthews, line 542: Feels weird from an API standpoint that `validate_data()` is an instance method but `prepare_job_kwargs` is a classmethod?

nautobot/extras/jobs.py
```diff
@@ -73,17 +76,17 @@
...
-    """Base model for jobs (reports, scripts).
+class RunJobTaskFailed(Exception):
+    """Celery task failed for some reason."""
+
+
+class BaseJob(Task):
+    """Base model for jobs.
 
     Users can subclass this directly if they want to provide their own base class for implementing multiple jobs
     with shared functionality; if no such sharing is required, use Job class instead.
```
glennmatthews, line 87: If we don't have an issue tracking this, we should IMO - let's get rid of the `BaseJob`/`Job` separation and instead support a `Meta.abstract` property that any Job can define.

nautobot/extras/jobs.py
```diff
@@ -104,25 +105,190 @@ class Meta:
...
+    def __call__(self, *args, **kwargs):
+        # Attempt to resolve serialized data back into original form by creating querysets or model instances
+        # If we fail to find any objects, we consider this a job execution error, and fail.
+        # This might happen when a job sits on the queue for a while (i.e. scheduled) and data has changed
+        # or it might be bad input from an API request, or manual execution.
+        try:
+            deserialized_kwargs = self.deserialize_data(kwargs)
+        except Exception:
+            stacktrace = traceback.format_exc()
+            self.log_failure(f"Error initializing job:\n```\n{stacktrace}\n```")
```
glennmatthews, line 123: Since we capture exceptions automatically now, is the try/except here needed or useful to provide additional context?

nautobot/extras/jobs.py
```diff
@@ -104,25 +105,190 @@ class Meta:
...
+        self.active_test = "run"
+        context_class = JobHookChangeContext if isinstance(self, JobHookReceiver) else JobChangeContext
```
glennmatthews, line 125: I'd love for us to rethink this too - can we make JobHookReceiver just "a job whose run() signature matches a certain specific pattern" rather than needing a separate class for this purpose?

nautobot/extras/jobs.py
```diff
@@ -104,25 +105,190 @@ class Meta:
...
+
+                @app.task(shadow_name=shadow_name, serializer='pickle')
+                def apply_function_async(fun, *args, **kwargs):
+                    return fun(*args, **kwargs)
+
+        Arguments:
+            args (Tuple): Task positional arguments.
+            kwargs (Dict): Task keyword arguments.
+            options (Dict): Task execution options.
+        """
```
glennmatthews, line 167: Default implementation returns nothing? Should it instead return an empty string or something like that?

nautobot/extras/jobs.py
```diff
@@ -104,25 +105,190 @@ class Meta:
...
+
+    def before_start(self, task_id, args, kwargs):
+        """Handler called before the task starts.
+
+        .. versionadded:: 5.2
+
```
glennmatthews, line 173:
```suggestion
```

nautobot/extras/jobs.py
```diff
@@ -104,25 +105,190 @@ class Meta:
...
+            None: The return value of this handler is ignored.
+        """
+        self.active_test = "initialization"
+
+        try:
+            self.job_result = self.get_job_result()
+        except TypeError as err:
+            raise RunJobTaskFailed(f"Unable to serialize data for job {task_id}") from err
+        except Exception as err:
+            raise RunJobTaskFailed(f"Unexpected failure in job {self.name}") from err
```
glennmatthews, line 189: Looks like the only exception we'd expect to see from `get_job_result()` would be an `ObjectDoesNotExist` error, no?

nautobot/extras/jobs.py
```diff
@@ -104,25 +105,190 @@ class Meta:
...
+
+        job_model = self.get_job_model()
```
glennmatthews, line 191: This could also raise an `ObjectDoesNotExist`?

nautobot/extras/jobs.py
glennmatthews: Wouldn't be a bad idea to check at model validation time  (prevent explicitly configuring `job_model.time_limit <= job_model.soft_time_limit`) but I think it's appropriate to also warn at job execution time in the case where the settings are misconfigured.

(Actually, maybe that should be a Django check we should add too?)

nautobot/extras/jobs.py
```diff
@@ -104,25 +105,190 @@ class Meta:
...
+        if time_limit <= soft_time_limit:
+            self.log_warning(
+                f"The hard time limit of {time_limit} seconds is less than "
+                f"or equal to the soft time limit of {soft_time_limit} seconds. "
+                f"This job will fail silently after {time_limit} seconds.",
+            )
+
+        self.log_info("Running job")
+
+    def run(self, *args, **kwargs):  # pylint: disable=arguments-differ
```
glennmatthews, line 209: Hmm. What's pylint expect the signature to be?

nautobot/extras/jobs.py
```diff
@@ -104,25 +105,190 @@ class Meta:
...
+        Arguments:
+            exc (Exception): The exception raised by the task.
+            task_id (str): Unique id of the failed task.
+            args (Tuple): Original arguments for the task that failed.
+            kwargs (Dict): Original keyword arguments for the task that failed.
+            einfo (~billiard.einfo.ExceptionInfo): Exception information.
+
+        Returns:
+            None: The return value of this handler is ignored.
+        """
```
glennmatthews, line 260: Does implementing these here as no-ops (as opposed to not implementing them at all) have any consequences we should be aware of?

nautobot/extras/jobs.py
```diff
@@ -354,11 +478,16 @@ def as_form(self, data=None, files=None, initial=None, approval_view=False):
...
-
         return form
 
+    def get_job_model(self):
+        return self.job_result.job_model or JobModel.objects.get(
+            module_name=self.__module__, job_class_name=self.__name__
+        )
+
+    def get_job_result(self):
+        return JobResult.objects.get(task_id=self.request.id)
```
glennmatthews, line 489: Would it make sense to make these `@property` instead of `get_*` methods?

nautobot/extras/jobs.py
```diff
@@ -467,6 +596,12 @@ def validate_data(self, data, files=None):
 
         return f.cleaned_data
 
+    @classmethod
+    def prepare_job_kwargs(cls, job_kwargs):
+        """Process dict and return kwargs that exist as ScriptVariables on this job."""
```
glennmatthews, line 601: We should rename `ScriptVariable` to `JobVariable` throughout.

nautobot/extras/jobs.py
```diff
@@ -467,6 +596,12 @@ def validate_data(self, data, files=None):
...
+        job_vars = cls._get_vars()
+        return {k: v for k, v in job_kwargs.items() if k in job_vars}
```
glennmatthews, line 603: Should we warn or anything if there are things in `job_kwargs` that we're throwing away?

nautobot/extras/jobs.py
```diff
@@ -919,11 +1040,8 @@ def is_job(obj):
     """
     Returns True if the given object is a Job subclass.
```
glennmatthews, line 1041: Would be nice for this to be replaced by a `obj.Meta.abstract` check, per my earlier comment.

nautobot/extras/management/commands/runjob.py
```diff
@@ -70,38 +51,20 @@ def handle(self, *args, **options):
         except json.decoder.JSONDecodeError as error:
             raise CommandError(f"Invalid JSON data:\n{str(error)}")
 
-        job_content_type = get_job_content_type()
+        job = Job.objects.get_for_class_path(job_class.class_path)
```
glennmatthews, line 54: Perhaps `job_model`?

nautobot/extras/scripts.py
glennmatthews, line 1: Should also be able to remove `nautobot/extras/reports.py`. :-)

nautobot/extras/tests/example_jobs/test_atomic_transaction.py
```diff
@@ -0,0 +1,44 @@
...
+    fail = BooleanVar()
+
+    @transaction.atomic
+    def run(self, fail=False):
+        try:
+            Status.objects.create(name="Test database atomic rollback 1")
+            if fail:
+                raise Exception("simulated failure")
+        except Exception:
+            self.log_failure("Job failed, all database changes have been rolled back.")
```
glennmatthews, line 22: Does this work? I'd have thought that since we're catching the exception, the rollback won't actually happen?

nautobot/extras/tests/test_views.py
glennmatthews, line 1: Why the removals from this file?

pyproject.toml
```diff
@@ -246,6 +246,7 @@ good-names = """_,i,j,k,pk"""
 # unused-import is already covered by flake8
 disable=""",
     abstract-method,
+    arguments-differ,
```
glennmatthews, line 249: Revert?

glennmatthews (commented):
Awesome stuff!

nautobot/docs/additional-features/jobs.md
gsnider2195: We don't have to but I was planning on removing `read_only`, `commit_default` and their respective override fields. That's part of the work I'm doing in #3453

nautobot/extras/jobs.py
gsnider2195: Agree. I opened #3448 to track that

nautobot/extras/jobs.py
gsnider2195: I *think* the problem is that pylint doesn't understand that the base class `celery.app.task.Task` is an abstract base class so it doesn't like that individual jobs are redefining the signature. I ended up disabling this check for the entire project temporarily because I couldn't find a way to make pylint happy. This was probably left in mistakenly while trying to figure out a generic fix because this run method never generated a pylint warning.

nautobot/extras/jobs.py
gsnider2195: It shouldn't. They were copy/pasted from the base class in celery.

nautobot/extras/jobs.py
```diff
@@ -1284,15 +1208,12 @@ def scheduled_job_handler(*args, **kwargs):
 
     user_pk = kwargs.pop("user")
     user = User.objects.get(pk=user_pk)
-    name = kwargs.pop("name")
+    kwargs.pop("name")
```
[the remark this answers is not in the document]
gsnider2195: This entire method is going to get deleted in #3453

nautobot/extras/tests/example_jobs/test_atomic_transaction.py
gsnider2195: `log_failure` raises another exception which causes the transaction to roll back. This is tested in `nautobot/extras/tests/test_jobs:JobTest`

nautobot/extras/tests/test_views.py
gsnider2195: The `test_read_only_*` test jobs were removed and `scheduled_job_pk`/`scheduled_job_handler` are going away as part of the dryrun overhaul so I removed the associated tests instead of trying to fix them.

pyproject.toml
gsnider2195: Tracking as follow-up work

nautobot/extras/api/views.py
gsnider2195: `validate_data` should be a class method. The only things it's accessing on `self` are class properties (`task_queues` and `field_order`)

nautobot/extras/jobs.py
gsnider2195: From the [celery docs](https://docs.celeryq.dev/en/stable/userguide/tasks.html?highlight=before_start#before_start) on `before_start`: 
> The return value of this handler is ignored.

nautobot/extras/jobs.py
gsnider2195: Tracked as follow-up item

nautobot/extras/jobs.py
gsnider2195: I don't think so. The reason we need this is to discard extra fields from forms that the job signature doesn't need. The form should take care of the rest of the data validation.

glennmatthews (approved):

PR #3489, opened 2023-03-24
Introduced `IPAddress.parent` concrete foreign key to `Parent`

jathanism:
# Closes: #3464 
# What's Changed
- `IPAddress.parent` relationship has related name to `Prefix.ip_addresses`
- The `IPAddress` constructor now takes two new arguments that are used FOR CREATE ONLY and are discarded (just like `IPAddress.address` is used to populate `network`, `broadcast`, and `prefix_length`):
  - `namespace` - Required to identify the closest parent for this new IP
  - `parent` - If namespace is not provided, parent will be used to determine the namespace and is then DISCARDED to disallow manual changes to the parent Prefix.
- Added `IPAddress.ip_version` field that is non-editable and automatically set on save.
- Added a `Prefix.reparent_ips()` method that will find `IPAddress` objects where the parent is that of the newly-created `Prefix` and set their `parent` to itself.
- `IPAddress.get_child_ips()` now just returns `Prefix.ip_addresses.all()`
- Extended `Prefix.delete()` to be aware of both child `Prefix` and child `IPAddress` objects, re-raising a revised `ProtectedError` if the act of deleting the existing parent would cause the child IPs to become orphaned.
- Added a new manager method `Prefix.objects.get_closest_parent()` to find the closest parent for any `IPAddress` or `Prefix` used in `IPAddress.save()` to automatically set the `parent`.
- Revised `PrefixQuerySet.delete()` to support list view deletions
- Added tree methods to `IPAddress` for ancestors/root/siblings.
- Swapped `family` for `ip_version` and updated filters and filter form
- Added `namespace` field to `IPAddressForm` that gets cleaned into `instance._namespace` to align with the constructor pattern that will discard namespace but use it to determine the correct parent.
- Clicking "Add an IP Address" from Prefix detail view will now include the namespace from the parent Prefix
- Got rid of `duplicate_ips` from detail view

## Screenshots

Child IPs are now directly related to their parent Prefix
Cursor_and_Notification_Center

Creating an IP chooses the namespace from the parent Prefix
Cursor_and_Notification_Center

Duplicate IPs are gone, related IPs are siblings sharing the same Prefix parent regardless of prefix length, and only prefix ancestors from the same Namespace are shown.
Cursor_and_Notification_Center

Creating a new prefix that would contain the IPs automatically re-parents them (/25 is a child of the /24)
image

Observe the ancestor prefixes in the hierarchy
Cursor_and_Notification_Center

IPs must have a parent Prefix and you can't create them without one to contain them
image

# TODO

- [ ] Unit, Integration Tests


nautobot/ipam/models.py
```diff
@@ -656,6 +671,14 @@ def reparent_subnets(self):
 
         return query.update(parent=self)
 
+    def reparent_ips(self):
+        """Determine the list of child IPAddresses and set the parent to self."""
+        query = IPAddress.objects.select_for_update().filter(
+            parent_id=self.parent_id,
+        )
```
jathanism, line 678: This is kind of feeling like it's not complete enough, but I'm not sure what else needs to be considered right now seeing as the namespace and prefix length are coming from the parent, and the parent should still have enough space to contain them.

nautobot/ipam/models.py
```diff
@@ -899,6 +919,21 @@ class IPAddress(PrimaryModel, StatusModel, RoleModelMixin):
     )
     broadcast = VarbinaryIPField(null=False, db_index=True, help_text="IPv4 or IPv6 broadcast address")
     prefix_length = models.IntegerField(null=False, db_index=True, help_text="Length of the Network prefix, in bits.")
+    parent = models.ForeignKey(
+        "ipam.Prefix",
+        blank=True,
+        null=True,
```
jathanism, line 925: I'm not sure how I feel about allowing this to be nullable.

nautobot/ipam/querysets.py
```diff
@@ -271,6 +272,66 @@ def filter(self, *args, prefix=None, **kwargs):
...
+
+        # Prepare the queryset filter
+        lookup_kwargs = {
+            "network__in": networks,
+            "prefix_length__in": prefix_lengths,
+            "ip_version": ip_version,
+            "broadcast__gte": str(broadcast),
+        }
+        if namespace is not None:
+            lookup_kwargs["namespace"] = namespace
```
jathanism, line 323: This is vestigial from initial attempt to make `namespace` optional. It's required, so that should be reflected here.

nautobot/ipam/migrations/0030_ipam__ipaddress__add_parent.py
glennmatthews, line 1: We're going to squash IPAM schema migrations 27, 28, and 30 together before the final merge of this feature, right?

nautobot/ipam/models.py
```diff
@@ -783,11 +806,8 @@ def siblings(self, include_self=False):
         return query
 
     def get_child_ips(self):
-        """
-        Return all IPAddresses within this Prefix and VRF. If this Prefix is a container in the global table, return
-        child IPAddresses belonging to any VRF.
-        """
-        return IPAddress.objects.net_host_contained(self.prefix)
+        """Return all IPAddresses within this Prefix and Namespace."""
```
glennmatthews, line 809:
```suggestion
        """Return all IPAddresses directly contained within this Prefix and Namespace."""
```

With this change we now only return directly descended IPs now, rather than all IPs contained within this prefix?

nautobot/ipam/models.py
```diff
@@ -899,6 +919,21 @@ class IPAddress(PrimaryModel, StatusModel, RoleModelMixin):
...
+        blank=True,
+        null=True,
+        related_name="ip_addresses",  # `IPAddress` to use `related_name="ip_addresses"`
+        on_delete=models.PROTECT,
+        help_text="The parent Prefix of this IPAddress.",
+    )
+    # ip_version is set internally just like network, broadcast, and prefix_length.
+    ip_version = models.IntegerField(
+        choices=choices.IPAddressVersionChoices,
+        null=True,
```
glennmatthews, line 933: why null=true?

nautobot/ipam/models.py
```diff
@@ -948,9 +983,18 @@ class Meta:
         ordering = ("host", "prefix_length")  # address may be non-unique
```
glennmatthews, line 983: Perhaps?

```suggestion
        ordering = ("parent__namespace__name", "ip_version", "host", "prefix_length")
```

Actually, do we even need to order by prefix_length now that host is unique within a namespace?

nautobot/ipam/models.py
```diff
@@ -973,23 +1017,11 @@ def _deconstruct_address(self, address):
...
+            self.ip_version = address.version
 
     def get_absolute_url(self):
         return reverse("ipam:ipaddress", args=[self.pk])
 
-    # TODO: The current IPAddress model has no appropriate natural key available yet.
-    #       However, by default all BaseModel subclasses now have a `natural_key` property;
-    #       but for this model, accessing the natural_key will raise an exception.
-    #       The below is a hacky way to "remove" the natural_key property from this model class for the time being.
-    class AttributeRemover:
```
glennmatthews, line 984: 🥳

nautobot/ipam/models.py
```diff
@@ -1037,6 +1064,19 @@ def clean(self):
...
+    def save(self, *args, **kwargs):
+        if not self.present_in_database:
+            if self._namespace is None:
+                raise ValidationError({"parent": "Namespace could not be determined."})
+            namespace = self._namespace
+        else:
+            namespace = self.parent.namespace
+
+        # Determine the closest parent automatically based on the Namespace.
+        self.parent = Prefix.objects.get_closest_parent(self.host, namespace=namespace)
```
glennmatthews, line 1076: Do we need to recalculate this on every save? Or should this only be done on create?

nautobot/ipam/querysets.py
```diff
@@ -271,6 +272,109 @@ def filter(self, *args, prefix=None, **kwargs):
...
+            protected_objects.update(parent=new_parent)
+            return super().delete(*args, **kwargs)
+
+    def _validate_cidr(self, value):
+        """Validate whether ``value`` is a validr IPv4/IPv6 CIDR."""
+        try:
+            cidr = netaddr.IPNetwork(str(value))
+        except netaddr.AddrFormatError:
+            raise ValidationError({"cidr": f"{value} does not appear to be an IPv4 or IPv6 network."})
+        return cidr
```
glennmatthews, line 324:
```suggestion
        try:
            return netaddr.IPNetwork(str(value))
        except netaddr.AddrFormatError as err:
            raise ValidationError({"cidr": f"{value} does not appear to be an IPv4 or IPv6 network."}) from err
```

nautobot/ipam/models.py
jathanism: Eek I am hesitant to do default, ordering based on namespace merely since that's a table join that won't scale well.

Also I'm thinking that since the addresses are all hierarchically numeric that , v4 and v6 should naturally sort without the need for `ip_version` to be part of ordering.

Lastly, yeah `prefix_length` needs to be reconsidered overall.

nautobot/ipam/models.py
jathanism: Same "it wants a default value if it's a choice field unless it's nullable" problem as always...

It's not user-serviceable and always forcefully set so I'm not worried about this one so much.

nautobot/ipam/models.py
jathanism: At this time it's entirely possible to edit the IP address or change its namespace, so unless we decide to make such properties immutable after create we need this to be recalculated on every save.

nautobot/ipam/models.py
glennmatthews: > So if we were to use `namespace` then its ordering would also be asserted, which is by `name`.

I need to dig a bit more into this then, because `Prefix` orders by `namespace` but in my local testing at least, prefixes in namespace `Gamma` are displayed after those in namespace `Global` when using default sorting, which seems wrong to me.

glennmatthews: Ah. Actually `Namespace` doesn't have any defined `ordering` at present, or even a `Meta` at all. I'll add that as part of #3496 if that's OK with you.

jathanism: In my mind I had added ordering by name. 😑

timizuoebideri1 (approved):

Issue #3429, opened 2023-03-15
IPAM Namespaces: Alpha-completeness


jathanism:
## Branched from #3417 

This is the work required to make the Namespace feature ready for alpha release.

Changes should be merged into #3378.

### To Do

```[tasklist]
### Prefix
- [x] UI: Child Prefixes nav tab in Prefix detail view (PrefixDetailTable) defaults to showing "Global" namespace, not matching that of the Prefix (should be same namespace as parent, once parenting implemented)
- [x] UI: When you go to "add child prefix" from `Prefix` detail view, it should default to the same Namespace as the parent
```

```[tasklist]
### VRF
- [x] Should be able to assign devices from `VRF` edit view
- [x] Ditch `RouteDistinguisher` and fold it back into `VRF.rd`
- [x] Rework `enforce_unique` checks (currently commented out); need to be removed entirely
- [x] `VRF.add_device(device, rd)` and `VRF.remove_device(device)` model methods to discourage use of `vrf.devices.add()`
- [x] `VRF.add_prefix(prefix)` and `VRF.remove_prefix(prefix)` model methods `vrf.prefixes.add()`
```

```[tasklist]
### Device
- [x] Should be able to assign `VRF` from `Device` edit view
```

```[tasklist]
### Interface
- [x] UI: Add `vrf` relationship (the optional FK from `Interface` to `VRF`)
```

```[tasklist]
### IPAddress
- [x] UI: Display `vrf` relationship as inherited from `Prefix`
```

glennmatthews:
### Prefix

> UI: Child Prefixes nav tab in Prefix detail view (PrefixDetailTable) defaults to showing "Global" namespace, not matching that of the Prefix (should be same namespace as parent, once parenting implemented)

Appears to be done as part of parenting work.

> UI: When you go to "add child prefix" from `Prefix` detail view, it should default to the same Namespace as the parent

Appears to be done as part of parenting work.

**Question**: Should it be possible to specify a VRF(s) when editing a Prefix? If so that isn't yet present.

**Question**: Should the Prefix detail view list associated VRF(s)? Currently it always displays VRF "Global" even if a different Prefix-to-VRF association exists.

gsnider2195:
Looking at branch [u/glennmatthews-3429-alpha-completeness](https://github.com/nautobot/nautobot/tree/u/glennmatthews-3429-alpha-completeness) I'm seeing the prefix table overflowing from its panel in the namespace detail view.

[...]

PR #3496, opened 2023-03-27
Fixes to Prefix views and functionality

glennmatthews:
# Closes: #N/A (progress toward #3429)
# What's Changed

- Add working `VRFs` field to Prefix create/edit form
- Rework Prefix ~`unique_together`, `index_together`, and~`ordering` to include `ip_version` as a factor (without this, at least on MySQL, I was seeing the table view interleave IPv4 and IPv6 prefixes confusingly)
   - [x] ~TODO: are the `unique_together` and `index_together` changes actually correct/appropriate/necessary? The `ordering` definitely is.~ Reverted the unique_together and index_together changes.
   - [x] Ordering of Namespace should probably be by `name` instead of PK.
- In `VRFPrefixAssignmentTable` collapse the `VRF` and `name` columns together as they were redundant.
- In `PrefixTable` linkify the `namespace` column and move it to the left of the table for visibility.
    - Also, mark the entire table as `orderable = False` because sorting a tree model just makes for confusion.
        - This also allows me to remove various cases in `ipam/views.py` where we were explicitly instantiating a `PrefixTable(orderable=False)` before
- In the Prefix detail template, prepend the namespace information to the `breadcrumbs` and remove redundant single-VRF information.
- In the related object tables in the Namespace detail view, hide the redundant `namespace` columns.

image

image

image

# TODO

- n/a Unit, Integration Tests

### DIFF
--- a/nautobot/ipam/forms.py
+++ b/nautobot/ipam/forms.py
@@ -324,7 +324,7 @@ class PrefixForm(LocatableModelFormMixin, NautobotModelForm, TenancyForm, Prefix
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         if self.instance is not None:
-            self.initial["vrfs"] = self.instance.vrfs.all().values_list("id", flat=True)
+            self.initial["vrfs"] = self.instance.vrfs.values_list("id", flat=True)
 
     def save(self, *args, **kwargs):
         instance = super().save(*args, **kwargs)
--- a/nautobot/ipam/forms.py
+++ b/nautobot/ipam/forms.py
@@ -271,16 +271,6 @@ class RIRFilterForm(NautobotFilterForm):
 
 
 class PrefixForm(LocatableModelFormMixin, NautobotModelForm, TenancyForm, PrefixFieldMixin):
-    """
-    vrf = DynamicModelChoiceField(
-        empty_label="Global",
-        null_option="Global",
-        queryset=VRF.objects.all(),
-        required=False,
-        label="VRF",
-    )
-    """
-
     vlan_group = DynamicModelChoiceField(
         queryset=VLANGroup.objects.all(),
         required=False,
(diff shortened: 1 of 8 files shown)


jathanism:
> and `ordering` to include `ip_version` as a factor (without this, at least on MySQL, I was seeing the table view interleave IPv4 and IPv6 prefixes confusingly)

I'm actually really surprised by this!

nautobot/ipam/models.py
```diff
@@ -506,14 +506,15 @@ class Prefix(PrimaryModel, StatusModel, RoleModelMixin):
...
             "prefix_length",
         )
         index_together = [
-            ["network", "broadcast", "prefix_length"],
-            ["namespace", "network", "broadcast", "prefix_length"],
+            ["ip_version", "network", "broadcast", "prefix_length"],
+            ["namespace", "ip_version", "network", "broadcast", "prefix_length"],
         ]
-        unique_together = ["namespace", "network", "prefix_length"]
+        unique_together = ["namespace", "ip_version", "network", "prefix_length"]
```
jathanism, line 517: We might want to be cautious about the the `index_together` and `unique_together` that share the same fields since they are both effectively indexing the same fields (unique just adds the constraint as well). The ordering is significant as well, so I think we should probably profile these assuming that you didn't already do this...

nautobot/ipam/forms.py
```diff
@@ -323,6 +321,16 @@ class Meta:
             "date_allocated": DateTimePicker(),
         }
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if self.instance is not None:
+            self.initial["vrfs"] = self.instance.vrfs.all().values_list("id", flat=True)
```
jathanism, line 327: Not that it really matters, but this is a thing:

```suggestion
            self.initial["vrfs"] = self.instance.vrfs.values_list("id", flat=True)
```

jathanism (approved):
Approved but I do think we should discuss the index changes just to make certain we're doing it correctly.

nautobot/ipam/models.py
glennmatthews: Haven't done any performance profiling - just using the `generate_test_data` command for now which is where I saw the ordering issues.

glennmatthews:
FYI here's an example of the interleaving of root IPv4 and IPv6 prefixes that I'm seeing with the baseline branch:

image

which makes sense since 50 (0x32) < (0x)38 < 69 (0x45) and 100 (0x64) < (0x)68.

nautobot/ipam/models.py
glennmatthews: Reverted the index_together/unique_together changes as I'm reasonably sure they're unnecessary.

jathanism:
> FYI here's an example of the interleaving of root IPv4 and IPv6 prefixes that I'm seeing with the baseline branch:
> 
> image
> 
> which makes sense since 50 (0x32) < (0x)38 < 69 (0x45) and 100 (0x64) < (0x)68.

Thanks for showing that. I'll update it in the `IPAddress.parent` branch as well.

Issue #1541, opened 2022-03-23
Reconciliation Story


itdependsnetworks:
### As ...

Nelly - Network Engineer

### I want ...

I want to be able to cleanly compare the actual state vs the intended state. 

### So that ...

So that I can quickly make decisions at what needs to be updated. Such as is the SoT incorrect (someone neglected to update it) or is the actual device incorrect (someone has not pushed out the configurations)

### I know this is done when...

I know this is done when I can see the actual state vs the intended state of the device. 

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

Depends on design

itdependsnetworks:
I believe following the pattern introduced in https://github.com/nautobot/nautobot-plugin-version-control would be a potential good solution.

itdependsnetworks:
@glennmatthews I see you have added the "needs clear requirements", it would be great to expand on what you are looking for. On a high level, need a better reconciliation story, providing explicit requirements would be much more providing a solution--which is why I broke an expected solution in it's own comment.

glennmatthews:
Can you clarify what you mean by "actual state" and "intended state" in this context? I presume "intended state" would be the data as currently stored in Nautobot (correct me if that's wrong, please) -- is "actual state" a separate Nautobot database view (as in nautobot-version-control)? Data gathered directly from devices via the existing NAPALM integration? Some other arbitrary data source (as in https://github.com/nautobot/nautobot-plugin-ssot/) that's being translated into a Nautobot-comparable format? Any and all of the above?

Is there something here that needs to be implemented in Nautobot core, or should this be an extension to the nautobot-ssot plugin's existing dry-run capabilities?

How does this request differ from the existing capabilities provided by the nautobot-golden-config plugin?

What's the scope of intended state here - you mention device configuration specifically several times, but I'm guessing that what's really being asked for is the ability to compare any and all of Nautobot's data, not just devices and config contexts?

itdependsnetworks:
Actual state would be "what is on the actual device". So regardless of how the data gets there (all good examples you had, napalm, ssot, etc.) this thought process.

[...]

PR #3498, opened 2023-03-27
Create Validation Engine for Data Comparison

sdoiron0330:
# Closes: #1541 
# What's Changed

- Creation of Validation Engine for Data Comparison
  - Model, View, and Validation classes for inheriting
  - Adding validations to registry during plugin creation
  - Basic job to run all validations found in registry

# TODO

- [ ] Unit, Integration Tests

# How to Use
All validation engine code would run from a plugin in a file called `validations.py`.  An example would look like this:
```python
from nautobot.extras.validation_engine.validations import ValidationSet

class LocationValidation(ValidationSet):
    name = "Location Validation"
    model = "dcim.location"

    def validate_site_in_correct_tenant(self, obj):
        external_tenant_data = {"na": ["test-site-1"]}
        if obj.tenant is None or obj.tenant == "":
            self.fail(obj, "tenant", obj.tenant, "Site does not have a tenant set.")
            return
        if obj.tenant.slug in external_tenant_data and obj.slug in external_tenant_data.get(obj.tenant.slug, []):
            self.success(obj, "tenant", obj.tenant)
        else:
            self.fail(obj, "tenant", obj.tenant, "Site does not exist in tenant data from external source.")

validations = [LocationValidation]
```
The following updates would be made the plugin's `views.py`, `template_content.py`, and `urls.py`.
```python
# views.py
from nautobot.extras.validation_engine.views import ObjectValidationView
from nautobot.dcim.models import Location

class LocationValidationView(ObjectValidationView):
    queryset = Location.objects.all()
```
```python
# template_content.py
from nautobot.extras.validation_engine.template_content import ValidationTab
class LocationValidationTab(ValidationTab):
    model = "dcim.location"
    reverse_url = "plugins:example_plugin:location_validation"
template_extensions = [LocationValidationTab]
```
```python
# urls.py
urlpatterns = [
    ...
    path("location/<id>", views.LocationValidationView.as_view(), name="location_validation"),
    ...
]
```
Then by running the `RunRegisteredValidations` job, all method that start with `validate_*` will be run for each object of that model, producing a tab on the individual record that looks like this.
Screenshot 2023-03-28 at 9 56 10 AM

# Next Steps
- Working to add UI element to base ObjectDetailView to indicate issue with a particular attribute.
- Optimizing the reverse lookup (how does an object find itself in the `ValidationResult` table).
- Testing/Comments/Linting


itdependsnetworks:
Just an FYI, this was a request by me for a customer project. We could make this a plugin, but the native integration into nautobot core would really take this to the next level. We are pre-pared to make it a plugin, but figured we would show you first before really investing in either way (core or plugin).

nautobot/extras/migrations/0077_validationresult.py
```diff
@@ -0,0 +1,34 @@
...
+        ('extras', '0076_remove_name_field_and_make_label_field_non_nullable'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='ValidationResult',
+            fields=[
+                ('id', models.UUIDField(default=uuid.uuid4, editable=False, primary_key=True, serialize=False, unique=True)),
+                ('class_name', models.CharField(max_length=100)),
+                ('function_name', models.CharField(max_length=100)),
```
itdependsnetworks, line 21: Since it exists on a class, I think `method_name` is more accurate.

nautobot/extras/validation_engine/models.py
```diff
@@ -0,0 +1,18 @@
...
+from django.contrib.contenttypes.models import ContentType
+
+from nautobot.core.models import BaseModel
+
+class ValidationResult(BaseModel):
+    class_name = models.CharField(max_length=100)
+    function_name = models.CharField(max_length=100)
+    last_validation_date = models.DateField()
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_id = models.CharField(max_length=200)
```
itdependsnetworks, line 13: We shouldn't need this, it can be inferred from the `validated_object`

nautobot/extras/validation_engine/models.py
```diff
@@ -0,0 +1,18 @@
+from django.db import models
+from django.shortcuts import reverse
+from django.contrib.contenttypes.fields import GenericForeignKey
+from django.contrib.contenttypes.models import ContentType
+
+from nautobot.core.models import BaseModel
+
+class ValidationResult(BaseModel):
+    class_name = models.CharField(max_length=100)
```
itdependsnetworks, line 9: I think class_name, function|method_name, last_validation_date, and valid should be required. 

We should also have an `expected_attribute_value` that is not required. This would be in situations where another system is indicating it should be such a value. 


validated_object_attribute => validated_attribute
validated_object_value => validated_attribute_value
+ expected_attribute_value

lampwins:
Before we get too far down the design path here I do have a few high level opinions.

- I'd like to avoid creating a yet another new extensibility pattern here, if at all possible (I'm not talking about a new extensibility feature, which this is, but rather the pattern by which it gets implemented).
- To that end, I feel core should provide only the primitive capability to register an external validation, and any higher level functionality, especially historical tracking of validation results should be moved into the Data Validation Engine App. Yes there have been discussions of moving DVE's functionality into core, but I'd rather see that happen wholesale, later down the line.
- In essence, core's external validation primitive would allow ad-hoc, on demand execution, in an object detail or list view. DVE would allow higher level functionality, like historical tracking of validation executions.


@nautobot/core to be clear, this is a longer range side project for which I have been interacting with @itdependsnetworks with, and will be a theme for us later. This is not something bound for 2.0.

itdependsnetworks:
Thanks for the update, I do genuinely appreciate the candor. @sdoiron0330 let's just keep this to the customer's implementation only for now.

Issue #3042, opened 2022-12-22
Replace `NestedFooSerializer` and `brief=True` pattern with DRF built-in `Meta.depth`


jathanism:
### Proposed Changes

The proposed change is:

- Remove all `nested_serializers` modules and `NestedFooSerializer` classes
- Remove all code to process them along with the concept of `?brief=True` and setting `self.brief` on viewsets
- Remove all overloaded `foo = NestedFooSerializer()` fields on API serializers and allow them to fall back to the default field from the model
- Leverage the[ built-in nested serialization feature in DRF](https://www.django-rest-framework.org/api-guide/serializers/#specifying-nested-serialization) e.g. `depth=0` is default (showing PKs), `depth=1` shows first-level objects nested representations, etc.
- Find a way to provide desired nested depth using request query parameters (e.g. `/api/dcim/sites/?depth=1`)

### Justification

This bespoke implementation is hindering our ability to leverage existing community extensions to Django for things like JSON schema generation, API import/export using CSV, among others. Additionally it is extra boilerplate code that must be written for each new model, and is custom code that must continue to be maintained and tested.

We will be better off in the long run 1.) using existing features of DRF, 2.) leveraging community resources where reasonably possible.

bryanculver:
What is the impact to plugins and integrators to the schemas as well as things like pynautobot?

glennmatthews:
https://www.linkedin.com/pulse/friday-quick-tip-dynamic-depth-serialization-django-rest-neidinger is probably relevant as far as dynamic depth based on request query parameters.

One thing we might lose in making this change (and should seek to preserve) is the ability to identify related objects by natural keys or otherwise by properties *other* than PK, e.g.:

```
POST api/dcim/devices/

{
    "name": "MyNewDevice",
    "rack": {
        "site": {
            "name": "Equinix DC6"
        },
        "name": "R204"
    },
    ...
}
```

instead of requiring:

```
POST /api/dcim/devices/

{
    "name": "MyNewDevice",
    "rack": "7f3ca431-8103-45cc-a9ce-b94c1f784a1d",
    ...
}
```

jathanism:
> https://www.linkedin.com/pulse/friday-quick-tip-dynamic-depth-serialization-django-rest-neidinger is probably relevant as far as dynamic depth based on request query parameters.
> 
> One thing we might lose in making this change (and should seek to preserve) is the ability to identify related objects by natural keys or otherwise by properties _other_ than PK, e.g.:
[...]

PR #3500, opened 2023-03-28
Replace `NestedFooSerializer` and `?brief=True` parameters using DRF's built-in `self.Meta.depth`

HanlinMiao:
# Closes: #3042
# What's Changed

## Remaining Items
- [x] Tags serializer representation
- [x] Polymorphic Serializer Functionality
- [x] Do we need `extra_fields` in `class Meta` of the serializers?
- [x] `depth` in OpenAPI Schema
- [x] What should be the default depth 0 or 1, depth should be 0
- [x] Testing & Documentation
    - Generic test bed to test depth to 2 but does not need to be exhaustive
    - Remove all of the `brief` testing functionality
- [x] The Dynamic Group Child Group Edit form is broken?
- [x] `TokenSerializer` whether `user` field is a required field or not.
- [ ] DeviceSerializer and VirtualMachineSerializer
    - `primary_ip` and `location` fields are not inherent fields (how to solve this)
- [ ] `cluster_count` is missing from `TenantSerializer` in next and develop
- [ ] NestedSerializer representation of a non-direct model field e.g `obj.cluster.location`
- [ ] ObjectPermission reinforcement on NestedSerializers

Location Browsable API
`?depth=0 / default depth`
Screen Shot 2023-04-11 at 4 20 02 PM
`?depth = 1`
Screen Shot 2023-04-11 at 4 20 33 PM
`?depth = 2` too much info to fit on one screen but notice that `parent's` `parent` is expanded as well
Screen Shot 2023-04-11 at 4 20 47 PM

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/api/serializers.py
```diff
@@ -170,8 +170,6 @@ class NautobotModelSerializer(
 class StatusModelSerializerMixin(BaseModelSerializer):
     """Mixin to add `status` choice field to model serializers."""
 
-    status = StatusSerializerField(required=True)
```
HanlinMiao, line 173: I need some time to figure out what we are doing with `TaggedModelSerializerMixin` in this file.

nautobot/dcim/api/serializers.py
```diff
@@ -213,7 +213,6 @@ class ConnectedEndpointSerializer(PathEndpointModelSerializerMixin):
 
 class LocationTypeSerializer(NautobotModelSerializer, TreeModelSerializerMixin):
     url = serializers.HyperlinkedIdentityField(view_name="dcim-api:locationtype-detail")
-    parent = NestedLocationTypeSerializer(required=False, allow_null=True, default=None)
```
HanlinMiao, line 216: My concern here is leaving `parent` to drf here which is a self-referential `Foreign Key` field, might cause some inconsistent data format/layout here. for example the `content_types` field here for the `LocationType` itself and its parent. We could just not override the `content_types` field here tho
Screen Shot 2023-03-27 at 11 30 46 PM

glennmatthews: Hmm, this is odd. Is DRF not able to find the `LocationTypeSerializer` to use for the `parent`?

nautobot/core/api/views.py
```diff
@@ -224,42 +222,23 @@ def get_object(self):
...
-            return super().get_queryset().prefetch_related(None).prefetch_related(*self.brief_prefetch_fields)
-
-        return super().get_queryset()
+    def get_serializer_context(self):
+        context = super().get_serializer_context()
+        depth = 0
+        try:
+            depth = int(self.request.query_params.get("depth", 0))
+        except ValueError:
+            pass
```
glennmatthews, line 237: When would we expect a ValueError here? Should we log a message if it happens?

HanlinMiao: When the `?depth` query parameter passed in here is not an integer. and Yeah def

nautobot/core/api/serializers.py
```diff
@@ -132,14 +216,107 @@ def get_field_names(self, declared_fields, info):
...
+            if value.__class__.__name__ in ["NautobotNestedSerializer", "NautobotPrimaryKeyRelatedField"]:
+                sub_data = data.get(key, None)
+                if sub_data is not None:
+                    data[key] = value.to_internal_value(sub_data)
+        return data
+
+    def build_nested_field(self, field_name, relation_info, nested_depth):
+        field = get_serializer_for_model(relation_info.related_model)
+
+        class NautobotNestedSerializer(field):
```
timizuoebideri1, line 244: This class logic looks really similar to `NautobotPrimaryKeyRelatedField`  above, I guess finding a way to utitlize that class here might be a better approach

nautobot/circuits/api/serializers.py
```diff
@@ -40,18 +22,8 @@ class ProviderSerializer(NautobotModelSerializer, TaggedModelSerializerMixin):
...
-            "asn",
-            "account",
-            "portal_url",
-            "noc_contact",
-            "admin_contact",
-            "comments",
-            "circuit_count",
-        ]
+        fields = "__all__"
+        extra_fields = ["circuit_count"]
```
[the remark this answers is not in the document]
glennmatthews: `circuit_count` isn't a model field, it's a queryset annotation, so it does need to be explicitly added IIRC.

nautobot/circuits/api/serializers.py
```diff
@@ -40,18 +22,8 @@ class ProviderSerializer(NautobotModelSerializer, TaggedModelSerializerMixin):
...
-            "slug",
-            "asn",
-            "account",
-            "portal_url",
-            "noc_contact",
-            "admin_contact",
-            "comments",
-            "circuit_count",
-        ]
+        fields = "__all__"
```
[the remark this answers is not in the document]
glennmatthews: In this case the preferred option would be to specify `exclude = [...]` to explicitly opt-out individual fields (https://www.django-rest-framework.org/api-guide/serializers/#specifying-which-fields-to-include)

nautobot/circuits/api/serializers.py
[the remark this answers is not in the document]
timizuoebideri1: I tried removing the `extra_fields` property and it works just as is.

jvanderaa:
What is going to be the overall change on this. What is the goal, what should be tested in SDKs?

jvanderaa:
Some discussion points that will be coming up. 

From my viewpoint, whether there is a default depth of 0 or 1 isn't really a big deal. I like the brevity that comes with 0, just need to make sure that the novel pynautobot and Nautobot API user know how to work with it. 

From my standpoint, if the default is 0, pynautobot would remain having the default be passed in, and provide a mechanism to set the depth on the calls within the instance create time. The Nautobot Ansible modules would create the instance with a depth of 1 to get at the same nested serializer data that is coming along. 

So if there is a vote, I'm a vote for a depth of 0 as default, and that the SDKs take care of providing sane default (Ansible) and access to the system.

nautobot/core/testing/api.py
```diff
@@ -185,13 +187,29 @@ def test_options_object(self):
...
+            depth_fields = []
+            for field in self.model._meta.fields:
+                if isinstance(field, (ForeignKey, ManyToManyField)) and (
+                    field.related_model != ContentType
+                    # user is a model field on Token but not a field on TokenSerializer
+                    and not (field.name == "user" and self.model == users_models.Token)
+                ):
+                    depth_fields.append(field.name)
+                elif field.name == "tags":
+                    depth_fields.append(field.name)
```
glennmatthews, line 207: **why** is `TaggableManager` not a subclass of `ManyToManyField`? :-(

nautobot/core/testing/api.py
```diff
@@ -185,13 +187,29 @@ def test_options_object(self):
...
+                elif field.name == "status" and isinstance(field, extras_models.StatusField):
+                    depth_fields.append(field.name)
```
glennmatthews, line 209: Shouldn't be needed - `StatusField` *is* a ForeignKey?

nautobot/core/testing/api.py
```diff
@@ -214,24 +232,102 @@ def test_list_objects_anonymous(self):
...
-            self.assertEqual(
-                sorted(response.data["results"][0]),
-                self.brief_fields,
-                "In order to test the brief API parameter the brief fields need to be manually added to "
-                "self.brief_fields. If this is already the case, perhaps the serializer is implemented incorrectly?",
-            )
+
+            response_data = response.data["results"][0]
+            for field in depth_fields:
+                self.assertEqual(field in response_data, True)
```
glennmatthews, line 251:
```suggestion
                self.assertIn(field, response_data)
```

nautobot/core/testing/api.py
```diff
@@ -214,24 +232,102 @@ def test_list_objects_anonymous(self):
...
+            response = self.client.get(url, **self.header)
+
+            self.assertHttpStatus(response, status.HTTP_200_OK)
+            self.assertIsInstance(response.data, dict)
+            self.assertIn("results", response.data)
+            self.assertEqual(len(response.data["results"]), self._get_queryset().count())
+
+            response_data = response.data["results"][0]
+            for field in depth_fields:
+                self.assertEqual(field in response_data, True)
```
glennmatthews, line 282:
```suggestion
                self.assertIn(field, response_data)
```

nautobot/core/testing/api.py
```diff
@@ -214,24 +232,102 @@ def test_list_objects_anonymous(self):
...
+                        try:
+                            uuid.UUID(str(entry["id"]))
+                        except ValueError:
+                            self.fail(f"Response data field {field} contains invalid UUID")
+                else:
+                    if response_data[field] is not None:
+                        try:
+                            uuid.UUID(str(response_data[field]["id"]))
+                        except ValueError:
+                            self.fail(f"Response data field {field} contains invalid UUID")
```
glennmatthews, line 294:
```suggestion
                    for entry in response_data[field]:
                        self.assertIsInstance(entry, dict)
                        self.assertTrue(is_uuid(entry["id"]))
                else:
                    if response_data[field] is not None:
                        self.assertIsInstance(response_data[field], dict)
                        self.assertTrue(is_uuid(response_data[field]["id"]))
```

nautobot/core/testing/api.py
```diff
@@ -214,24 +232,102 @@ def test_list_objects_anonymous(self):
...
+
+        @override_settings(EXEMPT_VIEW_PERMISSIONS=[])
+        def test_list_objects_depth_2(self):
+            """
+            GET a list of objects using the "?depth=2" parameter.
+            TreeModel Only
+            """
+            if not issubclass(self.model, TreeModel):
+                self.skipTest(f"Model {self.model._meta.model_name} is not a TreeModel")
```
glennmatthews, line 303: Can we instead make this another testcase mixin class and explicitly add it to the small number of tree-model tests?

nautobot/core/testing/api.py
```diff
@@ -214,24 +232,102 @@ def test_list_objects_anonymous(self):
...
+            url = f"{self._get_list_url()}?depth=2"
+            response = self.client.get(url, **self.header)
+
+            self.assertHttpStatus(response, status.HTTP_200_OK)
+            self.assertIsInstance(response.data, dict)
+            self.assertIn("results", response.data)
+            self.assertEqual(len(response.data["results"]), self._get_queryset().count())
+
+            response_data = response.data["results"]
+            for data in response_data:
```
glennmatthews, line 316: Wonder if the other two tests above should also iterate over all response.data["results"] instead of just the first record returned?

nautobot/dcim/api/serializers.py
```diff
@@ -6,14 +6,17 @@
 from nautobot.core.api import (
     ChoiceField,
     ContentTypeField,
-    SerializedPKRelatedField,
```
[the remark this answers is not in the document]
glennmatthews: We need to decide what we want to do as far as the schema goes - I think it's *probably* desirable to still have asymmetry in which the "read" side uses depth=1 but the "write" side uses depth=0.

nautobot/dcim/api/serializers.py
```diff
@@ -1160,17 +751,19 @@ def _get_termination(self, obj, side):
...
-        data = serializer(termination, context=context).data
-
-        return data
+        depth = int(self.context.get("depth", 0))
+        if depth == 0:
+            return termination.id
+        else:
+            relation_info = get_relation_info_for_nested_serializers(obj, termination, f"termination_{side.lower()}")
+            field_class, field_kwargs = self.build_nested_field(f"termination_{side.lower()}", relation_info, depth)
+            return field_class(termination, context={"request": self.context.get("request")}, **field_kwargs).data
```
glennmatthews, line 760: Is this a pattern that's common enough that we should create a utility function for it?

nautobot/dcim/tests/test_api.py
```diff
@@ -295,7 +293,6 @@ def test_time_zone_field_post_valid(self):
             "status": self.location_statuses[0].pk,
             "time_zone": time_zone,
             "location_type": self.lt1.pk,
-            "location": self.loc1.pk,
```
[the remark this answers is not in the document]
glennmatthews: yeah, it's just getting ignored as it's not a relevant form field. Good to remove it.

nautobot/docs/rest-api/overview.md
```diff
@@ -785,3 +785,478 @@ http://nautobot/api/dcim/sites/ \
 
 !!! note
     The bulk deletion of objects is an all-or-none operation, meaning that if Nautobot fails to delete any of the specified objects (e.g. due a dependency by a related object), the entire operation will be aborted and none of the objects will be deleted.
+
+### Depth Query Parameter
```
glennmatthews, line 789: Also need to remove the `Brief Format` section earlier in this document and add a note that it was removed in 2.0.0; may be other changes needed in this document as well.

nautobot/docs/rest-api/overview.md
```diff
@@ -785,3 +785,478 @@ http://nautobot/api/dcim/sites/ \
...
+        "1e986e5d-099d-4742-bfa5-16363abea5fd",
+        "21afbdad-c782-4fc5-9b3b-7b59273bb08c",
+        "0118b994-9943-4295-946b-92b27efe15db"
+    ]
+}
+```
+
+#### ?depth=1
+
+When `?depth=1` is added to the query parameter, all object-related fields, instead of being represented as UUIDs as they are when `?depth=0`, will be replaced by the related models' serializers.
```
glennmatthews, line 856:
```suggestion
When `?depth=1` is added to the query parameters, all object-related fields, instead of being represented as UUIDs as they are when `?depth=0`, will be replaced by the related models' serializers.
```

nautobot/docs/rest-api/overview.md
```diff
@@ -785,3 +785,478 @@ http://nautobot/api/dcim/sites/ \
...
+    },
+    "tenant": null,
+    "tags": []
+}
+```
+
+### ?depth=2 and beyond
+
+For `TreeModels`, sometimes it is necessary to traverse up more than one tree level and to retrieve an ancestor's information that is not a direct parent of the object.
+In this case, `?depth=2` and beyond would become very useful.
```
glennmatthews, line 989: I don't know that we need to be this specific - any model that has a foreign-key to another model that in turn has foreign-keys of its own (which is most models) can potentially benefit from a higher depth in a query.

nautobot/docs/rest-api/overview.md
```diff
@@ -785,3 +785,478 @@ http://nautobot/api/dcim/sites/ \
...
+
+!!! note
+    Notice that the location and its location_type's parent's parent's serializers are shown as well.
```
glennmatthews, line 992: I find this sentence difficult to parse but I don't have an immediate alternative to suggest.

nautobot/docs/rest-api/overview.md
```diff
@@ -785,3 +785,478 @@ http://nautobot/api/dcim/sites/ \
...
+    "facility": "",
+    "asn": null,
+    "physical_address": "",
+    "shipping_address": "",
+    "latitude": "-49.390000",
+    "longitude": null,
+    "contact_name": "",
+    "contact_phone": "+1-645-720-2291x878",
+    "contact_email": "",
+    "comments": "",
```
glennmatthews, line 1031: Might make the example more readable/clear if we summarize `...` the parts of the response that are unchanged from the previous example?

nautobot/extras/api/fields.py
```diff
@@ -18,11 +15,3 @@ def __init__(self, **kwargs):
...
         set_value = super().to_internal_value(data)
         return sorted(set_value)
-
-
-class RoleSerializerField(LimitQuerysetChoicesSerializerMixin, NestedRoleSerializer):
-    """NestedSerializer field for `Role` object fields."""
-
-
-class StatusSerializerField(LimitQuerysetChoicesSerializerMixin, NestedStatusSerializer):
-    """NestedSerializer field for `Status` object fields."""
```
glennmatthews, line 28: Don't we still need these for the `LimitQuerysetChoicesSerializerMixin` functionality?

nautobot/extras/api/serializers.py
```diff
@@ -266,64 +126,8 @@ class ConfigContextSerializer(ValidatedModelSerializer, NotesSerializerMixin):
         default=None,
     )
     owner = serializers.SerializerMethodField(read_only=True)
-    config_context_schema = NestedConfigContextSchemaSerializer(required=False, allow_null=True)
-    locations = SerializedPKRelatedField(
```
glennmatthews, line 270: Can you do a manual test on a `ConfigContext` REST API and make sure all these M2M fields are correctly represented in the response?

nautobot/extras/api/serializers.py
```diff
@@ -333,43 +137,20 @@ def __init__(self, *args, **kwargs):
...
-            ),
+            serializers=lambda: get_serializers_for_models(FeatureQuery("config_context_owners").list_subclasses()),
             allow_null=True,
         )
     )
     def get_owner(self, obj):
         if obj.owner is None:
             return None
-        serializer = get_serializer_for_model(obj.owner, prefix="Nested")
+        serializer = get_serializer_for_model(obj.owner)
```
glennmatthews, line 153: Don't we need `depth` handling here?

nautobot/extras/api/serializers.py
```diff
@@ -391,31 +172,20 @@ class ConfigContextSchemaSerializer(NautobotModelSerializer):
...
```
glennmatthews, line 188: `depth` handling?

nautobot/extras/api/serializers.py
```diff
@@ -494,56 +249,53 @@ class CustomLinkSerializer(ValidatedModelSerializer, NotesSerializerMixin):
...
-        fields = [
-            "url",
-            "name",
-            "description",
-            "content_type",
-            "filter",
-            "children",
-        ]
+        # renamed it to child_groups here because of
+        # an if statement in select2.min.js
```
glennmatthews, line 293: Ew. I don't like renaming a serializer field just to work around a UI issue. 

If we really must do this, let's make sure it's documented as a change.

nautobot/extras/api/serializers.py
```diff
@@ -565,33 +317,20 @@ class ExportTemplateSerializer(RelationshipModelSerializerMixin, ValidatedModelS
...
-            ),
+            serializers=lambda: get_serializers_for_models(FeatureQuery("export_template_owners").list_subclasses()),
             allow_null=True,
         )
     )
     def get_owner(self, obj):
         if obj.owner is None:
             return None
-        serializer = get_serializer_for_model(obj.owner, prefix="Nested")
+        serializer = get_serializer_for_model(obj.owner)
```
glennmatthews, line 333: depth handling?

nautobot/extras/api/serializers.py
```diff
@@ -702,15 +418,16 @@ def validate(self, data):
...
-                NestedRackSerializer,
+                DeviceSerializer,
+                LocationSerializer,
+                RackSerializer,
             ],
         )
     )
     def get_parent(self, obj):
-        serializer = get_serializer_for_model(obj.parent, prefix="Nested")
+        serializer = get_serializer_for_model(obj.parent)
```
glennmatthews, line 430: depth handling?

nautobot/extras/api/serializers.py
```diff
@@ -1045,36 +670,26 @@ class Meta:
...
+            serializers=lambda: get_serializers_for_models(get_all_concrete_models(NotesMixin)),
             allow_null=True,
         )
     )
     def get_assigned_object(self, obj):
         if obj.assigned_object is None:
             return None
         try:
-            serializer = get_serializer_for_model(obj.assigned_object, prefix="Nested")
+            serializer = get_serializer_for_model(obj.assigned_object)
```
glennmatthews, line 692: Depth handling?

nautobot/extras/api/serializers.py
```diff
@@ -1130,7 +731,7 @@ def get_changed_object(self, obj):
             return None
 
         try:
-            serializer = get_serializer_for_model(obj.changed_object, prefix="Nested")
+            serializer = get_serializer_for_model(obj.changed_object)
```
glennmatthews, line 734: depth handling?

nautobot/extras/tests/test_api.py
```diff
@@ -1480,14 +1479,17 @@ def test_run_job_object_var_lookup(self, mock_get_worker_count):
         self.assertIn("scheduled_job", response.data)
         self.assertIn("job_result", response.data)
         self.assertIsNone(response.data["scheduled_job"])
-        # The urls in a NestedJobResultSerializer depends on the request context, which we don't have
+        # The urls in a JobResultSerializer depends on the request context, which we don't have
         data_job_result = response.data["job_result"]
         del data_job_result["url"]
         del data_job_result["user"]["url"]
-        expected_data_job_result = NestedJobResultSerializer(job_result, context={"request": None}).data
+        del data_job_result["computed_fields"]
```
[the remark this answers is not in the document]
glennmatthews: Yeah, need some manual verification here to make *sure* that computed_fields are *not* showing up by default. Might be something we need to handle in core/api now that we're using `__all__` all over the place?

nautobot/extras/tests/test_tags.py
```diff
@@ -110,6 +110,7 @@ def test_create_invalid_tagged_item(self):
             "slug": "test-location",
             "status": Status.objects.get_for_model(Location).first().pk,
             "tags": [tag.id],
+            "location_type": self.location_type.pk,
```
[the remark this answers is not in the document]
glennmatthews: interesting!! Might be a bug that we need to check out?

nautobot/ipam/api/serializers.py
```diff
@@ -29,24 +26,22 @@
...
+        return str(value)
+
+    def to_internal_value(self, value):
+        """Convert API (string) representation to internal (IPNetwork) representation."""
+        try:
+            return netaddr.IPNetwork(value)
+        except netaddr.AddrFormatError:
+            raise serializers.ValidationError(f"Invalid IP address format: {value}")
+        except (TypeError, ValueError) as e:
+            raise serializers.ValidationError(e)
```
glennmatthews, line 43: Possibly should be in a `nautobot/ipam/api/fields.py` file?

nautobot/ipam/api/serializers.py
```diff
@@ -206,37 +141,18 @@ def validate(self, data):
...
-            "status",
-            "role",
-            "rir",
-            "date_allocated",
-            "description",
-        ]
         read_only_fields = ["family"]
+        # excluded here so that the user do not need to pass in a `prefix_length` value
+        # when doing a POST request.
+        exclude = ["prefix_length"]
```
glennmatthews, line 155: shouldn't it be in `read_only_fields` rather than excluded entirely?

nautobot/ipam/api/serializers.py
```diff
@@ -285,33 +200,26 @@ def to_representation(self, instance):
...
-            "tenant",
-            "status",
-            "role",
-            "nat_inside",
-            "nat_outside_list",
-            "dns_name",
-            "description",
-        ]
         read_only_fields = ["family"]
+        exclude = ["prefix_length"]
```
glennmatthews, line 222: Same comment here

nautobot/ipam/tests/test_api.py
```diff
@@ -610,7 +602,7 @@ def test_ports_regression(self):
             "name": "http",
             "protocol": "tcp",
             "device": str(device.id),
-            "ports": ["80"],
+            "ports": [80],
```
glennmatthews, line 605: Concerned about this change given that the test case appears to be explicitly about checking that a str is okay??

nautobot/project-static/js/forms.js
```diff
@@ -202,8 +202,12 @@ function initializeDynamicChoiceSelection(context, dropdownParent=null){
                     }
 
 
-                    // Allow for controlling the brief setting from within APISelect
-                    parameters.brief = ( $(element).is('[data-full]') ? undefined : true );
+                    // Allow for controlling the depth setting from within APISelect
+                    if ($(element).attr('data-depth') !== undefined){
+                        parameters.depth = parseInt($(element).attr('data-depth'))
+                    }else{
```
glennmatthews, line 208: nit

```suggestion
                    } else {
```

nautobot/tenancy/api/serializers.py
```diff
@@ -4,35 +4,22 @@
 from nautobot.extras.api.serializers import NautobotModelSerializer, TaggedModelSerializerMixin
 from nautobot.tenancy.models import Tenant, TenantGroup
 
-# Not all of these variable(s) are not actually used anywhere in this file, but required for the
-# automagically replacing a Serializer with its corresponding NestedSerializer.
-from .nested_serializers import NestedTenantGroupSerializer, NestedTenantSerializer  # noqa: F401
```
glennmatthews, line 9: I think we *might* have some documentation somewhere about needing to import everything from nested_serializers into serializers? If so that should be revised/removed.

nautobot/tenancy/api/serializers.py
```diff
@@ -45,19 +32,4 @@ class TenantSerializer(NautobotModelSerializer, TaggedModelSerializerMixin):
...
-            "comments",
-            "circuit_count",
-            "device_count",
-            "ipaddress_count",
-            "prefix_count",
-            "rack_count",
-            "virtualmachine_count",
-            "vlan_count",
-            "vrf_count",
-            "cluster_count",
```
glennmatthews, line 62: Have we confirmed manually that the `count` fields get included/populated correctly? I don't have confidence in the thoroughness of our unit tests to test this.

glennmatthews (commented):
Apologies in advance. :-)

glennmatthews (commented):
Apologies in advance. :-)

nautobot/extras/tests/test_api.py
```diff
@@ -547,16 +536,15 @@ def test_create_object_required_fields(self):
         self.assertHttpStatus(response, status.HTTP_400_BAD_REQUEST)
         self.maxDiff = None
         self.assertEqual(
-            response.data,
+            str(response.data["label"][0]),
```
jathanism, line 539: It's not clear what index `0` is here.

nautobot/ipam/api/serializers.py
jathanism: Yes, make it read-only for now. We'll revisit it in the Namespace work.

nautobot/project-static/js/forms.js
jathanism: Newline for `else`! 🔨

nautobot/virtualization/forms.py
```diff
@@ -453,7 +453,7 @@ class VMInterfaceForm(NautobotModelForm, InterfaceCommonForm):
         queryset=VLAN.objects.all(),
         required=False,
         label="Untagged VLAN",
-        brief_mode=False,
+        depth=1,
```
jathanism, line 456: Let's make sure to document that everywhere `brief_mode=False` it must now be `depth=1` on `DynamicModelChoiceField` instances!

jathanism (changes requested):

nautobot/core/api/schema.py
```diff
@@ -84,7 +84,12 @@ def get_operation(self, *args, **kwargs):
             for param in operation["parameters"]:
                 if param["name"] == "id" and "description" not in param:
                     param["description"] = "Unique object identifier, either a UUID primary key or a natural-key slug."
-
+            # TODO #3024 make depth a valid schema query parameters
```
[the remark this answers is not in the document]
HanlinMiao: https://github.com/nautobot/nautobot/actions/runs/4663799363/jobs/8255471066 I believe this is it.

nautobot/core/api/serializers.py
```diff
@@ -132,14 +171,82 @@ def get_field_names(self, declared_fields, info):
...
+                )
+                return self.build_nested_field(field_name, relation_info, nested_depth)
+
+        return super().build_field(field_name, info, model_class, nested_depth)
+
+    def build_property_field(self, field_name, model_class):
+        """
+        Create a property field for model methods and properties.
+        """
+        if field_name == "tags":
```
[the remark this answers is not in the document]
HanlinMiao: `build_property_field` accounts for the `depth=0` case. The previous case is where tags should be presented nested_serializers as denoted by `if nested_depth > 0`

nautobot/core/api/utils.py
```diff
@@ -102,9 +103,14 @@ def get_serializer_for_model(model, prefix=""):
         SerializerNotFound: if the requested serializer cannot be located.
     """
     app_name, model_name = model._meta.label.split(".")
+    if app_name == "contenttypes" and model_name == "ContentType":
+        app_name = "extras"
     # Serializers for Django's auth models are in the users app
     if app_name == "auth":
         app_name = "users"
+    # Special Case where users.Permission needs ObjectPermissionSerializer
```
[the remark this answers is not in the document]
HanlinMiao: In the unittests it seems to be running as `auth.Permission` instead of `users.ObjectPermission`
Screen Shot 2023-04-17 at 10 29 14 AM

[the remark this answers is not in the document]
glennmatthews: That's probably because of using `__all__` on the `GroupSerializer`, it's trying to pull in Django's built-in permissions since it has an M2M field on the built-in Group model. Probably the correct fix is to `exclude = ["permissions"]` on the `GroupSerializer`.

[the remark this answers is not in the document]
glennmatthews: This brings up a more general question of whether we should enhance the API generic test case to take an explicit list of the "expected" serializer fields for any given endpoint and fail if any are missing or any unexpected ones are present. Maybe a follow-on ticket?

nautobot/core/api/serializers.py
```diff
@@ -132,14 +171,82 @@ def get_field_names(self, declared_fields, info):
         fields = list(super().get_field_names(declared_fields, info))  # Meta.fields could be defined as a tuple
         self.extend_field_names(fields, "display", at_start=True)
         self.extend_field_names(fields, "id", at_start=True)
+        self.eliminate_field_names(fields, "_custom_field_data")
+        self.eliminate_field_names(fields, "_name")
```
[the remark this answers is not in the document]
HanlinMiao: do we need some fields to be seen in the API? "_termination_a_device" for example for `Cable`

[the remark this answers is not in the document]
glennmatthews: I wouldn't think so in that case, it should be handled by the `termination_a` and `termination_b` GFKs instead.

(Aside: why do cables have A and B terminations, but circuits have A and Z terminations?)

nautobot/core/api/serializers.py
```diff
@@ -315,3 +421,192 @@ class BulkOperationIntegerIDSerializer(serializers.Serializer):
...
+                            destination_id=instance.id,
+                        )
+                    assoc.validated_save()  # enforce relationship filter logic, etc.
+                    logger.debug("Created %s", assoc)
+
+                for remove_assoc in remove_assocs:
+                    if "request" in self.context and not self.context["request"].user.has_perm(
+                        "extras.delete_relationshipassociation"
+                    ):
+                        raise PermissionDenied("This user does not have permission to delete RelationshipAssociations.")
```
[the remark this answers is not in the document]
HanlinMiao: https://github.com/nautobot/nautobot/issues/3605

nautobot/tenancy/api/serializers.py
HanlinMiao: https://github.com/nautobot/nautobot/issues/3609

nautobot/extras/api/fields.py
HanlinMiao: Im not too sure if we still need `LimitQuerysetChoicesSerializerMixin` or what its original functionality is. (for example is it trying to limit the instances of `Statuses` those that are valid for `Location` for a `POST` request?) If so, it seems like `WritableSerializerMixin` is already doing that.

nautobot/extras/tests/test_api.py
HanlinMiao: It is the drf list of errors `[ErrorDetail(string='This field is required.', code='required')]`. So we are basically getting the first and only item and casting it to an str object.

nautobot/extras/api/serializers.py
HanlinMiao: Came up with a new solution in `forms.js` let me know if it is acceptable.

nautobot/core/api/serializers.py
```diff
@@ -132,14 +171,82 @@ def get_field_names(self, declared_fields, info):
...
+                    fields = field.Meta.fields
+                if hasattr(field.Meta, "exclude"):
+                    exclude = field.Meta.exclude
+
+        # This is a very hacky way to avoid name collisions in OpenAPISchema Generations
+        # The exact error output can be seen in this issue https://github.com/tfranzel/drf-spectacular/issues/90
+        # Apparently drf-spectacular does not support the `?depth` argument that comes with DRF
+        # So auto-generating NestedSerializers with the default class names that are the same when depth > 0
+        # does not make our schema happy.
+        NautobotNestedSerializer.__name__ = "NautobotNestedSerializer" + f"{uuid.uuid1()}"
```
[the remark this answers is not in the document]
HanlinMiao: > We have `field_name` and `relation_info.related_model` both available to us, we should be able to come up with a better name here. If the issue here is to avoid collisions between e.g. "Device.location" and "Cluster.location" both creating a `NestedLocationSerializer`, then the correct fix should be to maintain a registry of nested serializers we've already created and reuse them where appropriate, because they _should_ be the same serializer class - take a look at how `get_writable_class()` in `nautobot.core.api.schema` is implemented.

I do not know if a combination of `related_model` name and `model` name is sufficient to create a unique enough combination. Take `tags` on `location` for example. The list of tags represented by the same name `LocationTagsNautobotNestedSerializers` will still cause a conflict in `invoke check-schema`

nautobot/core/api/schema.py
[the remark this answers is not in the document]
glennmatthews: Looks like instead of `"type": "integer"`, it maybe needs to be `"schema": {"type": "integer"}`, based on looking at the existing schema for fields like `offset` and `limit`.

nautobot/extras/tests/test_tags.py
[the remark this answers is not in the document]
glennmatthews: Oh, because it's a negative test - it's an invalid request because `tag.id` above isn't a tag that's valid for a `Location` anyway. I'll bet if you take a closer look at the response under the base branch, it's probably complaining about the missing `location_type` *and* complaining about the bad `tags`.

nautobot/core/testing/api.py
```diff
@@ -185,13 +187,27 @@ def test_options_object(self):
...
+            for field in self.model._meta.fields:
+                if not field.name.startswith("_"):
+                    if isinstance(field, (ForeignKey, GenericForeignKey, ManyToManyField)) and (
+                        field.related_model != ContentType
+                        # user is a model field on Token but not a field on TokenSerializer
+                        and not (field.name == "user" and self.model == users_models.Token)
+                    ):
+                        depth_fields.append(field.name)
+                    elif field.name == "tags":
+                        depth_fields.append(field.name)
```
glennmatthews, line 208: Would it be possible/better to include `TaggableManager` in the `isinstance()` check above instead?

nautobot/docs/rest-api/overview.md
```diff
@@ -785,3 +714,309 @@ http://nautobot/api/dcim/sites/ \
 
 !!! note
     The bulk deletion of objects is an all-or-none operation, meaning that if Nautobot fails to delete any of the specified objects (e.g. due a dependency by a related object), the entire operation will be aborted and none of the objects will be deleted.
+
+### Depth Query Parameter
+
++++ 2.0.0
+
+The the built-in DRF `?depth` query parameter is introduced in Nautobot 2.0 to replace the `?brief` parameter. It enables [nested serialization](https://www.django-rest-framework.org/api-guide/serializers/#specifying-nested-serialization) functionality and offers a more dynamic and comprehensive browsable API. It is available for both retrieving a single object and a list of objects.
```
glennmatthews, line 722:
```suggestion
A `?depth` query parameter is introduced in Nautobot 2.0 to replace the `?brief` parameter. It enables [nested serialization](https://www.django-rest-framework.org/api-guide/serializers/#specifying-nested-serialization) functionality and offers a more dynamic and comprehensive browsable API. It is available for both retrieving a single object and a list of objects.
```

There's no mention of `DRF` as an acronym prior to this in the document, and I don't think it adds anything here.

nautobot/docs/rest-api/overview.md
```diff
@@ -785,3 +714,309 @@ http://nautobot/api/dcim/sites/ \
 
 !!! note
     The bulk deletion of objects is an all-or-none operation, meaning that if Nautobot fails to delete any of the specified objects (e.g. due a dependency by a related object), the entire operation will be aborted and none of the objects will be deleted.
+
+### Depth Query Parameter
```
glennmatthews, line 718: This should probably be moved earlier in the document, probably in a similar location to the now-removed `Brief Format` section, i.e. adjacent to the list/retrieve object operations since that's what it's most relevant to.

nautobot/extras/api/fields.py
glennmatthews: Interesting. The mixin was added in #2822, I assume there was some need for it. Maybe since we have `ForeignKeyLimitedByContentTypes` we don't need it any more? @timizuoebideri1 any thoughts?

nautobot/extras/api/serializers.py
```diff
@@ -948,10 +565,54 @@ def validate(self, data):
...
+                    raise serializers.ValidationError({"crontab": e})
+
+        return data
+
+
 class JobInputSerializer(serializers.Serializer):
     data = serializers.JSONField(required=False, default=dict)
     commit = serializers.BooleanField(required=False, default=None)
-    schedule = NestedScheduledJobCreationSerializer(required=False)
+    # TODO #3024: How to get rid of this?
```
glennmatthews, line 614: TODO is dealt with now, right?

HanlinMiao: Deleted some of the unnecessary #TODO comments.

nautobot/core/api/schema.py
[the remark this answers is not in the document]
HanlinMiao: Screen Shot 2023-04-19 at 10 52 02 AM
It seems like it is still not happy with it after I added `"schema": {"type": "integer"}`.

nautobot/core/api/serializers.py
```diff
@@ -132,14 +171,82 @@ def get_field_names(self, declared_fields, info):
...
+        """
+        Return a two tuple of (cls, kwargs) to build a serializer field with.
+        """
+        from rest_framework.utils.model_meta import RelationInfo, _get_to_field
+
+        # For tags field, DRF does not recognize the relationship between tags and the model itself (?)
+        # so instead of calling build_nested_field() it will call build_property_field() which
+        # makes the field impervious to the `?depth` parameter.
+        # So we intercept it here to call build_nested_field()
+        # which will make the tags field be rendered with TagSerializer() and respect the `depth` parameter.
```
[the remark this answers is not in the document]
HanlinMiao: `Status.content_types` is specified as `ContentTypeField` on `StatusModelSerializerMixin`. Same for all other `content_types` fields on other models. DRF handles `REAL` `ManyToManyField` automatically, but `tags` again are being weird and it is not getting picked up by DRF.

nautobot/extras/tests/test_api.py
```diff
@@ -1480,13 +1461,8 @@ def test_run_job_object_var_lookup(self, mock_get_worker_count):
         self.assertIn("scheduled_job", response.data)
         self.assertIn("job_result", response.data)
         self.assertIsNone(response.data["scheduled_job"])
-        # The urls in a NestedJobResultSerializer depends on the request context, which we don't have
```
HanlinMiao, line 1483: Turns out we do have the request context. From `response.wsgi_request`

nautobot/extras/tests/test_api.py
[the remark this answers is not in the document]
HanlinMiao: Solved in latest commit

nautobot/core/api/utils.py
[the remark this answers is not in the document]
HanlinMiao: Solved in standup

nautobot/tenancy/api/serializers.py
HanlinMiao: I couldn't really find such documentation. Maybe the comments are the only thing that needs to be deleted?

nautobot/core/api/fields.py
```diff
@@ -96,10 +95,9 @@ def to_internal_value(self, data):
             app_label, model = data.split(".")
             return self.queryset.get(app_label=app_label, model=model)
         except ObjectDoesNotExist:
-            self.fail("does_not_exist", content_type=data)
+            raise ValidationError({"content_type": f"{data} ContentType does not exist"})
```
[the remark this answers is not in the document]
HanlinMiao: It looks like it is working after all without these changes. So Im gonna revert it.

nautobot/tenancy/api/serializers.py
glennmatthews: Think you're right - I can't find any such docs.

nautobot/core/api/serializers.py
```diff
@@ -315,3 +421,192 @@ class BulkOperationIntegerIDSerializer(serializers.Serializer):
...
+        fields = list(super().get_field_names(declared_fields, info))
+        self.extend_field_names(fields, "custom_fields")
+        self.extend_field_names(fields, "computed_fields", opt_in_only=True)
+        return fields
+
+
+# TODO: remove in 2.2
+@class_deprecated_in_favor_of(CustomFieldModelSerializerMixin)
+class CustomFieldModelSerializer(CustomFieldModelSerializerMixin):
+    pass
```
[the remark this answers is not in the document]
glennmatthews: Still need a decision on this point.

nautobot/docs/rest-api/overview.md
```diff
@@ -785,3 +785,478 @@ http://nautobot/api/dcim/sites/ \
...
+
++++ 2.0.0
+
+The the built-in DRF `?depth` query parameter is introduced in Nautobot 2.0 to replace the `?brief` parameter. It enables [nested serialization](https://www.django-rest-framework.org/api-guide/serializers/#specifying-nested-serialization) functionality and offers a more dynamic and comprehensive browsable API.
+This parameter is an positive integer value that can range from 0 to 10. In most use cases, you will only need a maximum `depth` of 2 to get all the information you need.
+
+!!! important
+    The `?depth` query parameter should only be used for `GET` operations in the API. It should not be used in `POST`, `PATCH` and `DELETE` requests. For these requests, only `?depth=0` should be used.
```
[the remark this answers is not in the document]
glennmatthews: Second-guessing myself on this point - since a PUT/POST/PATCH returns the equivalent of a GET response, **should** it be possible to specify the `depth` that you want that response to provide?

nautobot/core/api/serializers.py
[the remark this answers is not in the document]
HanlinMiao: Removed the alias according to discussion in standup.

nautobot/ipam/api/serializers.py
```diff
@@ -344,15 +233,7 @@ def to_representation(self, instance):
 
 class ServiceSerializer(NautobotModelSerializer, TaggedModelSerializerMixin):
     url = serializers.HyperlinkedIdentityField(view_name="ipam-api:service-detail")
-    device = NestedDeviceSerializer(required=False, allow_null=True)
-    virtual_machine = NestedVirtualMachineSerializer(required=False, allow_null=True)
     protocol = ChoiceField(choices=ServiceProtocolChoices, required=False)
```
glennmatthews, line 236: Is this needed or does it "just work"?

nautobot/virtualization/api/serializers.py
```diff
@@ -67,64 +40,35 @@ class ClusterGroupSerializer(NautobotModelSerializer):
...
     url = serializers.HyperlinkedIdentityField(view_name="virtualization-api:virtualmachine-detail")
-    location = NestedLocationSerializer(read_only=True, required=False, allow_null=True)
-    cluster = NestedClusterSerializer()
-    tenant = NestedTenantSerializer(required=False, allow_null=True)
-    platform = NestedPlatformSerializer(required=False, allow_null=True)
-    primary_ip = NestedIPAddressSerializer(read_only=True)
-    primary_ip4 = NestedIPAddressSerializer(required=False, allow_null=True)
-    primary_ip6 = NestedIPAddressSerializer(required=False, allow_null=True)
-    local_config_context_schema = NestedConfigContextSchemaSerializer(required=False, allow_null=True)
+    primary_ip = IPAddressSerializer(read_only=True)
```
glennmatthews, line 63: Same suggestion here - IMHO `primary_ip` is redundant since we have `primary_ip4` and `primary_ip6` as existing serializer fields.

nautobot/virtualization/api/serializers.py
```diff
@@ -67,64 +40,35 @@ class ClusterGroupSerializer(NautobotModelSerializer):
...
 
     class Meta:
         model = VirtualMachine
+        # TODO #3024 keeping this for the append on line 130
```
glennmatthews, line 67: Not clear to me what this is referencing.

nautobot/virtualization/forms.py
glennmatthews: See my other comments, I don't think this is in fact correct/necessary in the vast majority of cases.

glennmatthews (commented):
🚤 🐎 ⏩

nautobot/core/api/serializers.py
```diff
@@ -138,8 +184,55 @@ def get_field_names(self, declared_fields, info):
...
+        """
+        Return a two tuple of (cls, kwargs) to build a serializer field with.
+        """
+        # For tags field, DRF does not recognize the relationship between tags and the model itself (?)
+        # so instead of calling build_nested_field() it will call build_property_field() which
+        # makes the field impervious to the `?depth` parameter.
+        # So we intercept it here to call build_nested_field()
+        # which will make the tags field be rendered with TagSerializer() and respect the `depth` parameter.
+        if field_name == "tags":
+            if nested_depth > 0:
```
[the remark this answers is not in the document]
HanlinMiao: 1.✅
2. I do not think we have enough information to do a `TaggableManager` check on the field per se. But we can check `model_class.tags`

nautobot/dcim/api/serializers.py
```diff
@@ -654,49 +434,28 @@ class Meta:
...
-    cluster = NestedClusterSerializer(required=False, allow_null=True)
-    virtual_chassis = NestedVirtualChassisSerializer(required=False, allow_null=True)
-    device_redundancy_group = NestedDeviceRedundancyGroupSerializer(required=False, allow_null=True)
-    local_config_context_schema = NestedConfigContextSchemaSerializer(required=False, allow_null=True)
+    # TODO #3024 How to get rid of this?
+    primary_ip = IPAddressSerializer(read_only=True)
 
     class Meta:
         model = Device
+        # TODO #3024 Leaving this here so that line 538 does not throw an error
```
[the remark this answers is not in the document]
HanlinMiao: Elaborated on the comment.

nautobot/extras/api/serializers.py
```diff
@@ -266,64 +122,8 @@ class ConfigContextSerializer(ValidatedModelSerializer, NotesSerializerMixin):
...
-        required=False,
-        many=True,
-    )
-    tenants = SerializedPKRelatedField(
-        queryset=Tenant.objects.all(),
-        serializer=NestedTenantSerializer,
-        required=False,
-        many=True,
-    )
     tags = serializers.SlugRelatedField(queryset=Tag.objects.all(), slug_field="slug", required=False, many=True)
```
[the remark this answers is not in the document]
HanlinMiao: replaced this line by adding `TaggedModelSerializerMixin`.

nautobot/virtualization/api/serializers.py
HanlinMiao: Elaborated the comment

nautobot/extras/api/serializers.py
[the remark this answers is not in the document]
glennmatthews: Careful on that one - these aren't the tags *applied* to a ConfigContext, they're the tags that a ConfigContext uses to filter devices/VMs by, so the semantic is different even if the code ends up being the same. It's *probably* fine but be sure to test carefully.

nautobot/core/api/utils.py
```diff
@@ -177,3 +181,76 @@ def rest_api_server_error(request, *args, **kwargs):
...
+        result = []
+        if depth == 0:
+            result = obj_related_field.values_list("pk", flat=True)
+        else:
+            for entry in obj_related_field.all():
+                relation_info = get_relation_info_for_nested_serializers(obj, entry, obj_related_field_name)
+                field_class, field_kwargs = serializer.build_nested_field(obj_related_field_name, relation_info, depth)
+                result.append(
+                    field_class(entry, context={"request": serializer.context.get("request")}, **field_kwargs).data
+                )
```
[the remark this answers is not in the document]
HanlinMiao: We are indirectly calling `nested_serializer_factory` here because `build_nested_field` on `BaseModelSerializer` calls   `nested_serializer_factory`.

nautobot/core/tests/test_api.py
```diff
@@ -420,7 +420,7 @@ def test_related_by_pk_no_match(self):
             response = self.client.post(url, data, format="json", **self.header)
         self.assertHttpStatus(response, status.HTTP_400_BAD_REQUEST)
         self.assertEqual(ipam_models.VLAN.objects.filter(name="Test VLAN 100").count(), 0)
-        self.assertTrue(response.data["location"][0].startswith("Related object not found"))
+        self.assertTrue(response.data["location"]["location"].startswith("Related object not found"))
```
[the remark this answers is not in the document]
HanlinMiao: The second nested "location" is caused by the `ValidationError` which is a dictionary we are raising in the new `WritableSerializerMixin`. Replaced that dictionary with just the error message.

nautobot/extras/api/mixins.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class StatusModelSerializerMixin(BaseModelSerializer):
+    """Mixin to add `status` choice field to model serializers."""
+
+    content_types = ContentTypeField(
+        queryset=ContentType.objects.filter(FeatureQuery("statuses").get_query()),
+        required=False,
+        many=True,
+    )
```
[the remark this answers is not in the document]
HanlinMiao: Yeeted

nautobot/extras/api/serializers.py
```diff
@@ -800,9 +466,6 @@ class JobVariableSerializer(serializers.Serializer):
 class JobRunResponseSerializer(serializers.Serializer):
     """Serializer representing responses from the JobModelViewSet.run() POST endpoint."""
 
-    schedule = NestedScheduledJobSerializer(read_only=True, required=False)
-    job_result = NestedJobResultSerializer(read_only=True, required=False)
```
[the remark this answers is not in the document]
HanlinMiao: Yeah def, I got too excited yeeting code

nautobot/extras/tests/test_plugins.py
```diff
@@ -473,7 +472,7 @@ def test_get_notes_url_on_object_raise_no_reverse_match(self, mock_reverse):
                     f"Notes feature is not available for model {model_name}. "
                     "Please make sure to: "
                     f"1. Include NotesMixin from nautobot.extras.model.mixins in the {model_name} class definition "
-                    f"2. Include NotesViewSetMixin from nautobot.extras.api.views in the {model_name}ViewSet "
+                    f"2. Include NotesViewSetMixin from nautobot.extras.api.mixins in the {model_name}ViewSet "
```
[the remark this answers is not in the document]
HanlinMiao: Revert changes

nautobot/ipam/api/serializers.py
```diff
@@ -206,37 +125,16 @@ def validate(self, data):
...
-            "vlan",
-            "status",
-            "role",
-            "rir",
-            "date_allocated",
-            "description",
-        ]
-        read_only_fields = ["family"]
+        fields = "__all__"
+        read_only_fields = ["family", "prefix_length"]
```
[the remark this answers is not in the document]
HanlinMiao: Probably `read_only_fields` it is much cleaner.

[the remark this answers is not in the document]
HanlinMiao: NVM `extra_kwargs` is more customizable.

nautobot/extras/api/serializers.py
```diff
@@ -842,31 +489,11 @@ class Meta:
 
 class ScheduledJobSerializer(BaseModelSerializer):
     url = serializers.HyperlinkedIdentityField(view_name="extras-api:scheduledjob-detail")
-    user = NestedUserSerializer(read_only=True)
-    job_model = NestedJobSerializer(read_only=True)
-    approved_by_user = NestedUserSerializer(read_only=True)
+    start_time = serializers.DateTimeField(format=None, required=False)
```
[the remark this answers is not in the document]
HanlinMiao: Fixed the test case

nautobot/ipam/api/serializers.py
HanlinMiao: Gonna open a backlog item on this.

nautobot/extras/api/serializers.py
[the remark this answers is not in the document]
HanlinMiao: Tested and `tags` field rendered correctly in the API.

nautobot/dcim/api/serializers.py
```diff
@@ -743,15 +502,26 @@ def validate(self, data):
...
+    @extend_schema_field(str)
     def get_parent_device(self, obj):
         try:
             device_bay = obj.parent_bay
         except DeviceBay.DoesNotExist:
             return None
-        context = {"request": self.context["request"]}
-        data = NestedDeviceSerializer(instance=device_bay.device, context=context).data
-        data["device_bay"] = NestedDeviceBaySerializer(instance=device_bay, context=context).data
+        depth = int(self.context.get("depth", 0))
```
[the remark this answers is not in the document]
HanlinMiao: I think we should keep the `depth = int(self.context.get("depth", 0))` here only because we need `self.Meta.depth` to be 1 for our OpenAPISchema. We should base our depth off of the query parameter depth if that's available, otherwise default to 0. Same comment for similar feedback regarding `self.Meta.depth` vs `self.context.get("depth", 0)`.

nautobot/core/api/serializers.py
```diff
@@ -94,6 +115,24 @@ class BaseModelSerializer(OptInFieldsMixin, serializers.ModelSerializer):
     """
 
     display = serializers.SerializerMethodField(read_only=True, help_text="Human friendly display value")
+    serializer_related_field = NautobotPrimaryKeyRelatedField
+    object_type = ObjectTypeField()
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # If it is not a NautobotNestedSerializer, we should set the depth argument to whatever is in the request's context
+        if "NautobotNestedSerializer" not in self.__class__.__name__:
```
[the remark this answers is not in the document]
HanlinMiao: This check here is in place for the BaseSerializer which is the top-level serializer. It needs to get its `depth` from the request `context`, and then all the other NestedSerializers can be built from `build_nested_field()` where the `nested_depth` argument passed into that function is the BaseSerializer.`Meta.depth`.

nautobot/core/api/serializers.py
```diff
@@ -315,3 +407,192 @@ class BulkOperationIntegerIDSerializer(serializers.Serializer):
...
+# TODO: remove in 2.2
+@class_deprecated_in_favor_of(CustomFieldModelSerializerMixin)
+class CustomFieldModelSerializer(CustomFieldModelSerializerMixin):
+    pass
+
+
+class RelationshipModelSerializerMixin(ValidatedModelSerializer):
+    """Extend ValidatedModelSerializer with a `relationships` field."""
+
+    # TODO # 3024 need to change this as well to show just pks in depth=0
```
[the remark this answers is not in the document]
HanlinMiao: https://github.com/nautobot/nautobot/issues/3630

nautobot/core/api/serializers.py
```diff
@@ -315,3 +407,192 @@ class BulkOperationIntegerIDSerializer(serializers.Serializer):
...
+                    )
+                    existing_objects = [assoc.get_peer(instance) for assoc in existing_associations]
+                else:
+                    existing_associations_1 = relationship.relationship_associations.filter(source_id=instance.pk)
+                    existing_objects_1 = [assoc.get_peer(instance) for assoc in existing_associations_1]
+                    existing_associations_2 = relationship.relationship_associations.filter(destination_id=instance.pk)
+                    existing_objects_2 = [assoc.get_peer(instance) for assoc in existing_associations_2]
+                    existing_associations = list(existing_associations_1) + list(existing_associations_2)
+                    existing_objects = existing_objects_1 + existing_objects_2
+
```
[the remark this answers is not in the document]
HanlinMiao: https://github.com/nautobot/nautobot/issues/3629

nautobot/core/api/utils.py
```diff
@@ -177,3 +181,76 @@ def rest_api_server_error(request, *args, **kwargs):
...
+    This method should only be called in build_nested_field()
+    in which relation_info and nested_depth are already given.
+    """
+    nested_serializer_name = f"{relation_info.related_model._meta.model_name.capitalize()}" + "NautobotNestedSerializer"
+    # If we already have built a suitable NestedSerializer we return the cached serializer.
+    # else we build a new one and store it in the cache for future use.
+    if nested_serializer_name in NESTED_SERIALIZER_CACHE:
+        field_class = NESTED_SERIALIZER_CACHE[nested_serializer_name]
+        field_class.Meta.model = relation_info.related_model
+        field_class.Meta.depth = nested_depth - 1
```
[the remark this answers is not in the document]
HanlinMiao: Implemented `NestedDepth<N><Model>` naming strategy.

nautobot/dcim/api/serializers.py
```diff
@@ -743,15 +502,26 @@ def validate(self, data):
 
         return data
 
-    @extend_schema_field(NestedDeviceSerializer)
+    @extend_schema_field(str)
```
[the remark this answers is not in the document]
HanlinMiao: Change the field to `parent_bay` and `@extend_schema_field(DeviceBaySerializer)`

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -128,6 +128,10 @@ These endpoints `/ipam/roles/`, `/dcim/rack-roles/` and `/dcim/device-roles/` ar
 | `/dcim/rack-roles/`   | `/extras/roles/` |
 | `/ipam/roles/`        | `/extras/roles/` |
 
+### API Query Parameters Changes
+
+Nautobot 2.0 removes the `?brief` query parameter and adds support for the `?depth` query parameter. As a result, the ability to specify `brief_mode` in `DynamicModelChoiceField` , `DynamicModelMultipleChoiceField` or `MultiMatchModelMultipleChoiceField` is also removed. For every occurrence of the aforementioned fields where you have `brief_mode` set to `True` e.g. `brief_mode=True`, please change that to `depth=1`. And leave other occurrences of the fields where you do not have `brief_mode` specified as they are. See more details about the `?depth` query parameter [here](../rest-api/overview.md/#depth-query-parameter)
```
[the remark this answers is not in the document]
HanlinMiao: Deleted the part saying replacing `brief_mode` with `depth` query parameter. Just tell the user to remove `brief_mode=True/False`

nautobot/ipam/api/serializers.py
HanlinMiao: https://github.com/nautobot/nautobot/issues/3631

nautobot/extras/tests/test_api.py
HanlinMiao: Revert the changes

nautobot/core/api/serializers.py
```diff
@@ -132,14 +180,60 @@ def get_field_names(self, declared_fields, info):
...
+        Return a two tuple of (cls, kwargs) to build a serializer field with.
+        """
+        # For tags field, DRF does not recognize the relationship between tags and the model itself (?)
+        # so instead of calling build_nested_field() it will call build_property_field() which
+        # makes the field impervious to the `?depth` parameter.
+        # So we intercept it here to call build_nested_field()
+        # which will make the tags field be rendered with TagSerializer() and respect the `depth` parameter.
+        if (
+            field_name == "tags"
+            and isinstance(model_class.tags, (_NautobotTaggableManager, _TaggableManager))
```
glennmatthews, line 205: Probably just the isinstance check is sufficient here - I think even if we had a _TaggableManager field named something *other* than `tags` we'd still need this case?

nautobot/core/api/serializers.py
```diff
@@ -315,3 +425,194 @@ class BulkOperationIntegerIDSerializer(serializers.Serializer):
 class GraphQLAPISerializer(serializers.Serializer):
     query = serializers.CharField(required=True, help_text="GraphQL query")
     variables = serializers.JSONField(required=False, help_text="Variables in JSON Format")
+
+
+class CustomFieldModelSerializerMixin(ValidatedModelSerializer):
```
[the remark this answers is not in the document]
glennmatthews: nit: should this be in `nautobot.core.api.mixins` instead?

nautobot/core/api/serializers.py
```diff
@@ -315,3 +399,195 @@ class BulkOperationIntegerIDSerializer(serializers.Serializer):
...
+        return fields
+
+
+# TODO: remove in 2.2
+@class_deprecated_in_favor_of(CustomFieldModelSerializerMixin)
+class CustomFieldModelSerializer(CustomFieldModelSerializerMixin):
+    pass
+
+
+class RelationshipModelSerializerMixin(ValidatedModelSerializer):
```
glennmatthews, line 433: nit: move to `nautobot.core.api.mixins`?

nautobot/core/api/serializers.py
```diff
@@ -315,3 +399,195 @@ class BulkOperationIntegerIDSerializer(serializers.Serializer):
...
+                    remove_assoc.delete()
+
+    def get_field_names(self, declared_fields, info):
+        """Ensure that "relationships" is always included as an opt-in field."""
+        fields = list(super().get_field_names(declared_fields, info))
+        self.extend_field_names(fields, "relationships", opt_in_only=True)
+        return fields
+
+
+class NotesSerializerMixin(BaseModelSerializer):
```
glennmatthews, line 555: nit: move to `nautobot.core.api.mixins`?

nautobot/dcim/api/serializers.py
```diff
@@ -743,23 +471,25 @@ def validate(self, data):
 
         return data
 
-    @extend_schema_field(NestedDeviceSerializer)
-    def get_parent_device(self, obj):
+    @extend_schema_field(DeviceBaySerializer)
+    def get_parent_bay(self, obj):
```
glennmatthews, line 475: Need to document this as an API change.

nautobot/dcim/api/serializers.py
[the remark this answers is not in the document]
glennmatthews: I'll be honest, I still don't fully understand this. :-) I see where `Meta.depth` is decremented when we create a nested serializer, but I don't see where `context["depth"]` does this. So it seems like with this code as-is, a query with `depth=1` would result in the Device serializer rendering with depth 1, the nested device-bay serializer rendering with depth 1, its nested device serializer rendering with depth 1, its own nested device-bay serializer rendering with depth 1, etc. to infinite depth. But obviously that isn't happening... so what am I missing? :-)

glennmatthews (approved):
:ship: :it:, I say :shipit:!

nautobot/core/api/serializers.py
HanlinMiao: I did try that before, however `ValidatedModelSerializer` and `BaseModelSerializer` base classes for these mixins seem to create another circular import problem🫠

nautobot/core/api/serializers.py
HanlinMiao: Same comment above

nautobot/dcim/api/serializers.py
HanlinMiao: Added as an entry in `v2-api-renamed-fields.yaml`

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -128,6 +128,10 @@ These endpoints `/ipam/roles/`, `/dcim/rack-roles/` and `/dcim/device-roles/` ar
 | `/dcim/rack-roles/`   | `/extras/roles/` |
 | `/ipam/roles/`        | `/extras/roles/` |
 
+### API Query Parameters Changes
+
+Nautobot 2.0 removes the `?brief` query parameter and adds support for the `?depth` query parameter. As a result, the ability to specify `brief_mode` in `DynamicModelChoiceField` , `DynamicModelMultipleChoiceField` or `MultiMatchModelMultipleChoiceField` is also removed. For every occurrence of the aforementioned fields where you have `brief_mode` set to `True/False` e.g. `brief_mode=True`, please remove the statement. And leave other occurrences of the fields where you do not have `brief_mode` specified as they are. See more details about the `?depth` query parameter [here](../rest-api/overview.md/#depth-query-parameter)
```
jathanism, line 133: We should never link anything in documentation using "here". Better to always link the contextual text.

```suggestion
Nautobot 2.0 removes the `?brief` query parameter and adds support for the `?depth` query parameter. As a result, the ability to specify `brief_mode` in `DynamicModelChoiceField`, `DynamicModelMultipleChoiceField`, and `MultiMatchModelMultipleChoiceField` has also been removed. For every occurrence of the aforementioned fields where you have `brief_mode` set to `True/False` (e.g. `brief_mode=True`), please remove the statement, leaving other occurrences of the fields where you do not have `brief_mode` specified as they are.

Please see the [documentation on the `?depth` query parameter](../rest-api/overview.md/#depth-query-parameter) for more information.
```

nautobot/docs/release-notes/version-2.0.md
```diff
@@ -18,6 +18,10 @@ DeviceRole, RackRole, IPAM Role, and IPAddressRoleChoices have all been merged i
 
 Added Site Model Fields to Location. Location Model now has `asn`, `comments`, `contact_email`, `contact_name`, `contact_phone`, `facility`, `latitude`, `longitude`, `physical_address`, `shipping_address` and `time_zone` fields.
 
+#### Added Depth REST API Query Parameter
+
+Added the `?depth` query parameter in Nautobot v2.X to replace the `?brief` parameter in the REST API. It enables [nested serialization](https://www.django-rest-framework.org/api-guide/serializers/#specifying-nested-serialization) functionality and offers a more dynamic and comprehensive browsable API. It allows users greater control of the API response data and it is available for both retrieving a single object and a list of objects. This parameter is an positive integer value that can range from 0 to 10. Check out more about the `?depth` query parameter in our [REST API Documentation](../rest-api/overview.md/#depth-query-parameter)
```
jathanism, line 23:
```suggestion
Added the `?depth` query parameter in Nautobot v2.X to replace the `?brief` parameter in the REST API. It enables [nested serialization](https://www.django-rest-framework.org/api-guide/serializers/#specifying-nested-serialization) functionality and offers a more dynamic and comprehensive browsable API. It allows users greater control of the API response data and it is available for both retrieving a single object and a list of objects. This parameter is a positive integer value that can range from 0 to 10. To learn more more, check out the [documentation on the `?depth` query parameter](../rest-api/overview.md/#depth-query-parameter).
```

nautobot/docs/release-notes/version-2.0.md
```diff
@@ -236,6 +240,10 @@ The "Container" status will be removed and all prefixes will be migrated to the
 
 ### Removed
 
+#### Removed Brief REST API Query Parameter
+
+Support for `?brief` REST API query parameter and `Nested*Serializers` are being removed in Nautobot v2.X. They are replaced by the new `?depth` query parameter. Check out the specific changes in [REST API documentation](../rest-api/overview.md/#depth-query-parameter)
```
jathanism, line 245:
```suggestion
Support for `?brief` REST API query parameter and `Nested*Serializers` have been removed in Nautobot v2.X. They are replaced by the new [`?depth` query parameter](../rest-api/overview.md/#depth-query-parameter).
```

nautobot/project-static/js/forms.js
```diff
@@ -287,6 +287,12 @@ function initializeDynamicChoiceSelection(context, dropdownParent=null){
                         else {
                             results[idx] = record;
                         }
+                        // DynamicGroupSerializer has a `children` field which fits an inappropriate if condition
+                        // in select2.min.js, which will result in the incorrect rendering of DynamicGroup DynamicChoiceField.
+                        // So we nullify the field here since we do not need this field.
+                        if (record.url.includes("dynamic-groups")){
```
jathanism, line 293: Can't wait to yeet this in v2 UI.

jathanism (approved):

nautobot/core/api/mixins.py
```diff
@@ -12,3 +28,73 @@ def get_queryset(self):
...
+        if hasattr(self, "fields"):
+            for field_name, field_instance in self.fields.items():
+                if field_name in filter_params and field_instance.source == "*":
+                    logger.debug("Discarding non-filter field %s", field_name)
+                    del filter_params[field_name]
+        return filter_params
+
+    def get_queryset_filter_params(self, data, queryset):
+        """
+        Data could be a dictionary and an int (for the User model) or a str that represents the primary key.
```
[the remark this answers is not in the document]
HanlinMiao: It could be `{"location": location.pk}` or a nested dictionary `{"location": {"name": "Campus-01"}}`

nautobot/core/api/mixins.py
```diff
@@ -12,3 +28,73 @@ def get_queryset(self):
...
+        If it is a dictionary, we return it after remove non-filter fields.
+        If it is a primary key, we return a dictionary object formatted like this {"pk": pk}
+        """
+        if isinstance(data, dict):
+            params = dict_to_filter_params(data)
+            return self.remove_non_filter_fields(params)
+        try:
+            pk = int(data) if isinstance(queryset.model._meta.pk, AutoField) else uuid.UUID(str(data))
```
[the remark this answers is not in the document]
HanlinMiao: I believe the `int` case is for the `User` model mentioned in the docstring above. But yeah I can add a comment here as well.

Issue #3432, opened 2023-03-15
Config Contexts Not Applying Correctly


bile0026:
### Environment
* Nautobot version (Docker tag too if applicable): v1.5.12
* Python version: 3.9
* Database platform, version: Postgres
* Middleware(s): k8s/Rancher

### Steps to Reproduce
1. Create 2+ dynamic groups with devices associated with any kind of filters.
2. Create a config context filtered by any other object, or unfiltered.
3. Create config context objects in a git repository applied by dynamic group, and sync to Nautobot.
4. View the "rendered context" in the Device>Config Context view.

### Expected Behavior
Config contexts filtered by dynamic group, should only apply to the devices which are members of that Dynamic Group.

### Observed Behavior
All config contexts will be applied in the rendered context (and then returned with GraphQL or Golden Config intended config generation). The correct contexts are shown in the "Source Contexts", and the "extra" contexts are not shown here, just in the rendered context. This was not an issue prior to this feature being "feature flagged".

bile0026:
Update on this issue. I have determined that this seems to happen when using `Tenant` or custom fields within the Dynamic Group to group the devices you want. There could be other fields, but those are the 2 that I've been able to determine so far that reliably allow the issue to be reproduced.

glennmatthews:
Hmm, that last sounds suspiciously like a caching issue. Just to confirm, cacheops is disabled in your deployment?

bile0026:
> Hmm, that last sounds suspiciously like a caching issue. Just to confirm, cacheops is disabled in your deployment?

Yes it is disabled

abates:
Okay, I think I've narrowed down the problem. It appears that the incorrectly merged contexts are being added by the [config context data annotation](https://github.com/nautobot/nautobot/blob/c82f9294647b2dde600d121c8bfb75cdc7945781/nautobot/extras/querysets.py#L88). We didn't update that queryset to account for dynamic groups, so the context that is applied only by dynamic group ends up matching everything (I think).

I'm not sure what the correct course of action is for fixing this.

bile0026:
I applied similar logic to line 132 to apply dynamic groups to `base_query` and although that appears to properly add the context to the "source contexts" but still is not filtering in the rendered context. Looking for more guidance on what to look at next.

abates:
@johanek we have a concern about potential performance impacts with this solution, especially when retrieving config contexts via API. Do you have any way of testing your use case at scale?

johanek:
[...]

PR #3504, opened 2023-03-28
Disable config context/dynamic groups annotation

abates:
# Closes: #3432
# What's Changed

This change will disable the config context annotation on device queries in order to allow config contexts to be mapped by dynamic group.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/querysets.py
+++ b/nautobot/extras/querysets.py
@@ -94,6 +94,9 @@ class ConfigContextModelQuerySet(RestrictedQuerySet):
         """
         from nautobot.extras.models import ConfigContext
 
+        if settings.CONFIG_CONTEXT_DYNAMIC_GROUPS_ENABLED:
+            return self
+
         return self.annotate(
             config_context_data=Subquery(
                 ConfigContext.objects.filter(self._get_config_context_filters())


bryanculver:
Won’t this mean as soon as you enable the feature, existing config contexts+devices (which may not be using Dynamic Groups) will no longer be annotated?

glennmatthews:
The annotation is only a performance optimization - the code will fallback to a less efficient (but correct) way of calculating the device config contexts in its place.

bryanculver:
> The annotation is only a performance optimization - the code will fallback to a less efficient (but correct) way of calculating the device config contexts in its place.

👍

lampwins:
The reason the annotation was added in the first place was because the device API was extremely slow. I’m not sure we are fully considering that implication here.

bryanculver:
This fix is insufficient at this time and it looks like #3432 can't really be fixed without a massive performance cost. This really needs proper reversibility to solve. Performance may be abated with changes coming into `next` around how API responses are returned (namely #3586). Since this diff is very minor it'll be easy to re-open but no progress is being made towards this at this time.

PR #3505, opened 2023-03-28
VRF namespace enhancements

glennmatthews:
# Closes: #N/A (more progress on #3429)
# What's Changed

- Remove remaining references to `enforce_unique` from VRF model and filter.
- Add VRFs selection to Device create/edit form (likely a placeholder for now given the need for a proper formset to specify RD on the through table)

image

- Show associated VRFs on the Device detail view

image

# TODO

- [x] Unit, Integration Tests


nautobot/dcim/forms.py
```diff
@@ -1755,6 +1761,8 @@ def __init__(self, *args, **kwargs):
                 self.initial["location"] = self.instance.parent_bay.device.location_id
                 self.initial["rack"] = self.instance.parent_bay.device.rack_id
 
+            self.initial["vrfs"] = self.instance.vrfs.values_list("id", flat=True)
```
timizuoebideri1, line 1764: Maybe a test asserting initial vrfs on form?

nautobot/dcim/forms.py
```diff
@@ -1768,6 +1776,11 @@ def __init__(self, *args, **kwargs):
         if position:
             self.fields["position"].widget.choices = [(position, f"U{position}")]
 
+    def save(self, *args, **kwargs):
+        instance = super().save(*args, **kwargs)
+        instance.vrfs.set(self.cleaned_data["vrfs"])
```
timizuoebideri1, line 1781: Here also, I think test asserting this logic?

nautobot/ipam/tables.py
```diff
@@ -264,15 +262,15 @@ class Meta(BaseTable.Meta):
 class VRFDeviceAssignmentTable(BaseTable):
     """Table for displaying VRF Device Assignments with RD."""
 
-    vrf = tables.LinkColumn(verbose_name="VRF")
-    device = tables.LinkColumn()
+    vrf = tables.Column(verbose_name="VRF", linkify=lambda record: record.vrf.get_absolute_url(), accessor="vrf.name")
```
gsnider2195, line 265: Might be nice to have namespace shown here too. The screenshot of this table has VRFs from two namespaces but it's not obvious which VRF belongs to which namespace. Perhaps use `accessor="vrf.display"`? (`f"{self.namespace}: ({self.name})"`)

image

gsnider2195: Also, would it make more sense for `VRF.display` to be `f"{self.name} ({self.namespace})"`

gsnider2195:
@glennmatthews if you pick this back up please pull from github. I merged in the prototype branch and fixed some bugs that were introduced by changing `IPAddress.family` to `IPAddress.ip_version` upstream.

nautobot/dcim/forms.py
gsnider2195, line 1764: I'm wondering why this is necessary here but when I made a similar change to `InterfaceForm` in #3399 I didn't have to do this.

nautobot/dcim/forms.py
gsnider2195, line 1781: Same comment as above. This wasn't necessary for a similar change in #3399

nautobot/dcim/forms.py
gsnider2195: Talking to @jathanism it looks like it's because this is the reverse relationship. The forward relationship on the `VRFform` didn't require these.

gsnider2195 (approved):

jathanism (approved):

Issue #1726, opened 2022-05-03
Migrate Git Repo Sync to be a System Job


bryanculver:
See requirements in #275

bryanculver:
Side-effect of #3545 which is needed for #3228

bryanculver:
Closed with #3545

Issue #275, opened 2021-04-09
Add ability to schedule Git sync


jedelman8:
### Environment

* Python version 3.9
* Nautobot version: 1.0.0b3

### Proposed Functionality

This proposal is to migrate Git repository sync to be a Nautobot Job to be able to leverage the scheduling features therein.

### Use Case

As a user, I want to schedule Git repository syncs so that my export templates, config contexts, and job are always up to date.

### Database Changes

TBD

### External Dependencies

TBD

### Requirements (In-Scope for MVP)
- Schedule Git syncs to be performed on a regular basis
- Migrate Git sync tasks to be execute as a Job
- Establish a pattern for how system Jobs are provided
- System jobs will be enabled by default
- Migrate permissions from Git sync task to job
- Job will take repo as a parameter (therefor a schedule sync can sync a particular repo at a time)

### Non-Requirements (Out of Scope for MVP)
- No "sync all repos" option
- Simplify (or eliminate) the Job Result related object pattern

#### Future
- [x] #1725
- [x] #1726

jedelman8:
Need to breakup into a few issues (in this priority)

- Scheduled Jobs
- Scheduled Git Sync
- Scheduler to be used by plugins
- Templated Jobs 
- Webhooks for Jobs

chadell:
just for clarification,  when you say `Scheduled Jobs` and `Scheduler to be used by plugins`, what about `Scheduled Jobs defined by plugins`?

itdependsnetworks:
While not explicit, it is implied. Since Plugins can create jobs, and if jobs can be scheduled, you would get both.

jedelman8:
Once #374  gets in, we can revisit scheduling other types of operations.

jedelman8:
While this was in `near-term`, I'd like to open it up for discussion to get feedback on priority of scheduling other operations.  Anyone have feedback/input?

glennmatthews:
One possibility here is to create a "system" Job specifically for triggering the sync of a given GitRepository - this job could be built in or could be provided as part of a standard `nautobot-jobs` Git repo of its own if we wanted to. Either way, as a Job like any other, it could be scheduled just like any other job.

jathanism:
In reference to this comment, I wanted to clarify a few things from my perspective.

> Need to breakup into a few issues (in this priority)
> 
> * Scheduled Jobs

This is coming in v1.2.0, so it's already accounted for. 

> * Scheduled Git Sync

[...]

PR #3545, opened 2023-04-05
Got GitRepository sync working with Jobs overhaul

jathanism:
# Closes: DNE (part of the larger work for #3228), also #275 and #1726
# What's Changed

- Introduce the concept of a "system" Job
- System jobs currently live at `nautobot.core.jobs`
- Converted `pull_git_repository_and_refresh_data` into a Job called `GitRepositoryPullAndRefreshData`
- `GitRepositoryPullAndRefreshData` is defined at `nautobot.extras.datasources.jobs.git`
- The `nautobot.extras.datasources.jobs` module has been included in `CELERY_IMPORTS` so that workers will pick it up and tr
eat it like a properly registered task
- Ripped out (almost) all of the `JobResult.set_status()` and `save()` calls and allowed them to just `raise` any errors
- In places where `request` is an argument, it was replaced with `user`
- Implemented `register_jobs()` hook called in `nautobot.core.apps.ready()`
- Added exception handling for imports done by the worker process to align with the same pattern for not letting bad job code block startup
- Eliminated the following from `nautobot.extras.datasources.git`
  - `log_job_result_final_status()`
  - `pull_git_repository_and_refresh_data()`
  - `git_repository_diff_origin_and_local()`
- Moved sync/dry-run code from `GitRepository.save()` to a new `sync()` method, which takes `dry_run` and `user` arguments
- Therefore `GitRepository.save(trigger_resync=False)` is no longer a thing.
- Views have been updated to explicitly call `instance.sync()`
- `GitRepositoryForm.clean()` has been replaced with `save()` which will call `instance.sync()` with the proper arguments after the object has been created.
- 🐛: Fixed a bug in Job edit view template where slug was missing, preventing creation from happening
- Implemented a new `GitRepository.get_latest_sync()` method to return the latest eligible `JobResult` for a given repo; this is being called by views now.

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/datasources/git.py
```diff
@@ -100,38 +104,29 @@ def get_job_result_and_repository_record(repository_pk, job_result_pk, logger):
...
         logger=logger,
     )
-    job_result.save()
 
 
 # TODO(jathan): The state transition stuff here should be deleted in exchange
 # for trusting the database backend.
 @nautobot_task
-def pull_git_repository_and_refresh_data(repository_pk, request, job_result_pk):
+def pull_git_repository_and_refresh_data(repository_pk, user_pk, job_result_pk):
```
jathanism, line 129: This will go away.

nautobot/extras/datasources/git.py
```diff
@@ -180,7 +173,7 @@ def pull_git_repository_and_refresh_data(repository_pk, request, job_result_pk):
 # TODO(jathan): The state transition stuff here should be deleted in exchange
 # for trusting the database backend.
 @nautobot_task
-def git_repository_diff_origin_and_local(repository_pk, request, job_result_pk, **kwargs):
+def git_repository_diff_origin_and_local(repository_pk, user_pk, job_result_pk, **kwargs):
```
jathanism, line 176: So will this...

nautobot/core/apps/__init__.py
```diff
@@ -20,8 +22,25 @@
...
+    """
+    from nautobot.extras.jobs import Job
+
+    for job in class_list:
+        if not inspect.isclass(job):
+            raise TypeError(f"Job class {job} was passed as an instance!")
+        if not issubclass(job, Job):
+            raise TypeError(f"{job} is not a subclass of extras.jobs.Job!")
+
+        registry["system_jobs"].append(job)
```
jathanism, line 43: This was directly lifted from `nautobot.extras.plugins.utils`. I'm thinking this should be centralized as util.

glennmatthews: Is any of this redundant with `register_jobs` in the base branch (https://github.com/nautobot/nautobot/pull/3483/files#diff-e4b3fe39da892dc97291f6a19ae5449758cd5a1353cb61a19cbc6bffe41ade4eR182-R185)?

nautobot/core/celery/__init__.py
```diff
@@ -1,20 +1,20 @@
...
+from celery.utils.log import get_task_logger
 from django.conf import settings
 from django.core.serializers.json import DjangoJSONEncoder
 from django.utils.module_loading import import_string
 from kombu.serialization import register
 from prometheus_client import CollectorRegistry, multiprocess, start_http_server
 
 
-logger = logging.getLogger(__name__)
+logger = get_task_logger(__name__)
```
glennmatthews, line 17: Is this right? This is server-side code not worker-side, isn't it?

nautobot/core/jobs/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+        user = job_result.user
+
+        job_result.log(
+            f'Creating/refreshing local copy of Git repository "{repository.name}"...',
+            logger=self.logger,
+        )
+
+        try:
+            if not os.path.exists(settings.GIT_ROOT):
+                os.makedirs(settings.GIT_ROOT)
```
glennmatthews, line 45: I wonder if we should consider removing this... `GIT_ROOT` should preexist IMHO.

nautobot/core/jobs/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+
+            ensure_git_repository(
+                repository,
+                job_result=job_result,
+                logger=self.logger,
+            )
+
+            job_result.log(
+                f'The current Git repository hash is "{repository.current_head}"',
+            )
```
glennmatthews, line 55: Might make sense to move this into `ensure_git_repository`?

nautobot/core/jobs/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+
+            refresh_datasource_content("extras.gitrepository", repository, user, job_result, delete=False)
+
+        except Exception as exc:
+            job_result.log(
+                f"Error while refreshing {repository.name}: {exc}",
+                level_choice=LogLevelChoices.LOG_FAILURE,
+                logger=self.logger,
+            )
+            raise
```
glennmatthews, line 65: This *might* be redundant - `ensure_git_repository` already catches, logs, and re-raises likely exceptions that occur within it, and `refresh_datasource_content` catches and logs (but does not re-raise) likely exceptions there as well.

nautobot/core/jobs/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+                os.makedirs(settings.GIT_ROOT)
+
+            git_repository_dry_run(repository, job_result=job_result, logger=self.logger)
+
+        except Exception as exc:
+            job_result.log(
+                f"Error while running a dry run on {repository.name}: {exc}",
+                level_choice=LogLevelChoices.LOG_FAILURE,
+            )
+            raise
```
glennmatthews, line 104: Similarly, `git_repository_dry_run` already catches, logs, and re-raises most exceptions that occur.

nautobot/core/testing/__init__.py
```diff
@@ -69,6 +75,38 @@ def run_job_for_testing(job, username="test-user", **kwargs):
     return job_result
 
 
+def create_job_result_and_run_job(module, name, source="local", *args, **kwargs):
+    """Test helper function to call get_job_class_and_model() then and call run_job_for_testing()."""
```
glennmatthews, line 79:
```suggestion
    """Test helper function to call get_job_class_and_model() then call run_job_for_testing()."""
```

nautobot/core/testing/__init__.py
```diff
@@ -69,6 +75,38 @@ def run_job_for_testing(job, username="test-user", **kwargs):
     return job_result
 
 
+def create_job_result_and_run_job(module, name, source="local", *args, **kwargs):
```
glennmatthews, line 78: I know `run_job_for_testing` is already defined directly in `__init__.py`, but should we consider putting these functions in a new submodule like `nautobot/core/testing/jobs.py` and just importing them here if desired?

nautobot/core/apps/__init__.py
jathanism: Kind of? I think we're trying to converge them. For the moment, the above code is for registering the job classes in Nautobot, and the one you linked is registering them in Celery.

nautobot/core/apps/__init__.py
```diff
@@ -50,6 +69,9 @@ def ready(self):
         if menu_items is not None:
             register_menu_items(menu_items)
 
+        jobs = import_string("nautobot.core.jobs.jobs")
+        register_jobs(jobs)
```
jathanism, line 73: This is where `foo_jobs_module.jobs` (the attribute that contains a list of Job classes) is load and could be the point where we align registration for both Nautobot and Celery simultaneously.

nautobot/core/celery/__init__.py
jathanism: It's correct as I see it, because the current `logger` use-cases herein are of the `import_modules` and `worker_ready` signal receivers which are each executed on the worker.

nautobot/core/jobs/__init__.py
jathanism: Yeah, I think it is in fact bootstrapped with `nautobot-server` runs.

nautobot/core/testing/__init__.py
jathanism: I still want to do this. Was focused on the underlying issues with triggering sync/dry-run first.

nautobot/core/apps/__init__.py
jathanism: We are going to unify all Jobs to be registered in the same way, and that most likely means just using the Celery task registry as the registry and reworking the other APIs to pull from that.

nautobot/extras/templates/extras/job_edit.html
```diff
@@ -81,6 +81,7 @@
         <div class="panel-body">
             {% render_field form.enabled %}
             {% include "extras/inc/overridable_field.html" with field=form.name override_field=form.name_override %}
+            {{ form.slug.as_hidden }}
```
jathanism, line 84: This was a side effect from #3439

nautobot/extras/forms/forms.py
```diff
@@ -611,12 +611,14 @@ class Meta:
             "tags",
         ]
 
-    def clean(self):
-        super().clean()
+    def save(self, commit=True):
+        instance = super().save(commit=commit)
```
jathanism, line 615: If there's a use-case for `commit=False`, the code below needs to account for that and probably not run unless `commit=True`.

nautobot/extras/models/datasources.py
```diff
@@ -67,9 +67,8 @@ class Meta:
         verbose_name_plural = "Git repositories"
 
     def __init__(self, *args, **kwargs):
-        # If instantiated from the REST API, the originating Request will be passed as a kwarg:
-        # FIXME(jathan); Replace this with user instead; make request go away.
-        self.request = kwargs.pop("request", None)
+        # If instantiated from the API/UI, the originating User will be passed as a kwarg:
```
jathanism, line 70: This comment isn't entirely true. Other callers have been setting this as an instance variable.

nautobot/extras/models/datasources.py
```diff
@@ -81,6 +80,13 @@ def __str__(self):
     def get_absolute_url(self):
         return reverse("extras:gitrepository", kwargs={"slug": self.slug})
 
+    def get_latest_sync(self):
```
jathanism, line 83: Docstring.

nautobot/extras/signals.py
```diff
@@ -195,6 +195,8 @@ def git_repository_pre_delete(instance, **kwargs):
     """
     from nautobot.extras.datasources import refresh_datasource_content
 
+    # FIXME(jathan): In light of jobs overhaul and Git syncs as jobs, we need to rethink this. We
```
jathanism, line 198: And also consider the worker event pattern for such a thing.

nautobot/core/settings.py
```diff
@@ -682,6 +682,8 @@
 # Celery results backend URL to tell workers where to publish task results
 CELERY_RESULT_BACKEND = "nautobot.core.celery.backends.NautobotDatabaseBackend"
 
+CELERY_IMPORTS = ["nautobot.core.jobs"]
```
jathanism, line 685: Add a comment as to why this is here. This is NOT intended to be a user-serviceable setting as we are using it for system jobs.

nautobot/extras/datasources/registry.py
```diff
@@ -48,20 +45,26 @@ def refresh_datasource_content(model_name, record, request, job_result, delete=F
...
+    # callbacks will result in this task "failing successfully" by raising an
+    # exception.
+    failure_logs = job_result.job_log_entries.filter(log_level=LogLevelChoices.LOG_FAILURE)
+    if failure_logs.exists():
+        msg = f"Failed to refresh data provided by {record}. Please see traceback."
+        job_result.log(msg, level_choice=LogLevelChoices.LOG_FAILURE)
+        raise RuntimeError(msg)
+
+    # Otherwise, log a friendly info message.
+    job_result.log(f"Data refresh from {record} complete!", level_choice=LogLevelChoices.LOG_INFO)
```
jathanism, line 70: @glennmatthews This is the block I was referencing on Slack that had to be added to maintain parity with 1.x behaviors. We can certainly change this, but to keep things moving this is the current solution.

nautobot/extras/managers.py
```diff
@@ -89,6 +89,10 @@ def store_result(
             if fields["status"] in states.READY_STATES:
                 fields["date_done"] = timezone.now()
 
+            # Don't overload the user if it's already been set when the JobResult was created.
+            if obj.user_id is not None:
+                del fields["user_id"]
```
jathanism, line 94: @gsnider2195 This was my interim fix for `JobResult.user` getting nullified. You fixed this with the context changes, right?

nautobot/extras/models/jobs.py
```diff
@@ -725,6 +725,8 @@ def linked_record(self):
     def get_absolute_url(self):
         return reverse("extras:jobresult", kwargs={"pk": self.pk})
 
+    # FIXME(jathan): This needs to go away. Need to think about that the impact
+    # will be in the JOB_RESULT_METRIC and how to compensate for it.
```
jathanism, line 729: @Kircheneer I would appreciate your thoughts on this comment and howe we might approach it.

nautobot/extras/tests/example_jobs/test_atomic_transaction.py
```diff
@@ -17,7 +17,7 @@ def run(self, fail=False):
         try:
             Status.objects.create(name="Test database atomic rollback 1")
             if fail:
-                raise Exception("simulated failure")
+                raise Exception("simulated failure")  # pylint: disable=broad-exception-raised
```
jathanism, line 20: TODO: Replace these naked exceptions and pylint disables with a real exception (e.g. [`JobTaskRunFailed`](https://github.com/nautobot/nautobot/blob/feature/765-jobs-overhaul/nautobot/extras/jobs.py#L80-L81))

nautobot/extras/models/jobs.py
Kircheneer: Where is the job status setting logic moving? Can we just move the metric creation there?

nautobot/extras/managers.py
gsnider2195: Yes the fix was to manually set `user` on the jobresult after running `apply()` [here](https://github.com/nautobot/nautobot/blob/e47608214277dcc4550e2e32c4389807c183fb2e/nautobot/extras/models/jobs.py#L763-L790)

nautobot/extras/forms/forms.py
glennmatthews: Ah, I think I'm starting to remember now why we had used transaction.on_commit in the original implementation - I think there are (or maybe were?) some cases where we were saving the GitRepository with commit=false specifically to prevent triggering an automatic resync or dry-run... I can't remember more specifics than that at the moment.

nautobot/core/jobs/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+            refresh_datasource_content("extras.gitrepository", repository, user, job_result, delete=False)
+        finally:
+            job_result.log(
+                f"Repository synchronization completed in {job_result.duration}",
+                level_choice=LogLevelChoices.LOG_INFO,
+                logger=self.logger,
+            )
+
+
+class GitRepositoryDiffOriginalAndLocal(Job):
```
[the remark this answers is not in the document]
jathanism: And maybe even simplify the other one to `GitRepositorySync`? I like it.

nautobot/extras/api/serializers.py
```diff
@@ -631,7 +631,7 @@ def validate(self, data):
         """
         Add the originating Request as a parameter to be passed when creating/updating a GitRepository.
         """
-        data["request"] = self.context["request"]
+        data["user"] = self.context["request"].user
```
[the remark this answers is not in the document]
jathanism: I don't think so, since one is generated by the change context if not provided. We're trying to get rid of any reason to use or pass requests around entirely.

nautobot/extras/datasources/git.py
```diff
@@ -54,31 +48,29 @@
...
-        repository_pk=repository.pk,
-        request=copy_safe_request(request),
-    )
+    job_model = job_class().job_model
 
+    return JobResult.enqueue_job(job_model, user, repository=repository.pk)
 
-def enqueue_git_repository_diff_origin_and_local(repository, request):
+
+def enqueue_git_repository_diff_origin_and_local(repository, user):
```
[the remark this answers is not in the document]
jathanism: Agreed, but for clarify, which API are you referencing here?

nautobot/extras/tests/test_datasources.py
```diff
@@ -369,13 +377,13 @@ def test_pull_git_repository_and_refresh_data_with_valid_data(self, MockGitRepo)
                 MockGitRepo.return_value.checkout.return_value = self.COMMIT_HEXSHA
 
                 # Run the Git operation and refresh the object from the DB
-                pull_git_repository_and_refresh_data(self.repo.pk, self.mock_request, self.job_result.pk)
-                self.job_result.refresh_from_db()
-
+                job_model = GitRepositoryPullAndRefreshData().job_model
```
glennmatthews, line 380: Why do we need to instantiate the job-class in order to get its job-model? Seems like that should be a classproperty perhaps?

Same question below.

nautobot/extras/managers.py
jathanism: Hmm I think these are different problems, because yours is for eager results, and the issue I encountered was during normal job runs (async). I'll pull this out and see whether it's still necessary.

nautobot/extras/api/serializers.py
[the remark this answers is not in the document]
glennmatthews: Does this pose problems for chained jobs wanting to use a common request id to correlate changelogs together?

glennmatthews (approved):

nautobot/extras/tests/test_datasources.py
jathanism: Maybe. The problem we're having here is coming up with a pattern that allows it to be a cached property of some sort since it's making a database call, and we don't want it to make a database call every single time that attribute is accessed. Some kind of hybrid cached/classproperty would be ideal perhaps.

gsnider2195: Potentially for this line of code

```py
job_model = Job.objects.get(module_name=GitRepositoryPullAndRefreshData.__module__, job_class_name=GitRepositoryPullAndRefreshData.__name__)
```

nautobot/core/jobs/__init__.py
gsnider2195: This is in `nautobot.core.cli._configure_settings()`:

```py
    if not os.path.exists(settings.GIT_ROOT):
        os.makedirs(settings.GIT_ROOT)
```

nautobot/extras/datasources/registry.py
```diff
@@ -29,16 +29,13 @@ def refresh_datasource_content(model_name, record, request, job_result, delete=F
...
     job_result.log(f"Refreshing data provided by {record}...", level_choice=LogLevelChoices.LOG_INFO)
-    job_result.save()
-    if request:
-        change_context = JobChangeContext(user=request.user)
-        # TODO(jathan): Review how we can eliminate these deliberate `set_status()` and `save()`
-        # calls,moving them to Celery-friendly primitives instead.
+    if user:
+        change_context = JobChangeContext(user=user)
         with change_logging(change_context):
             for entry in get_datasource_contents(model_name):
```
gsnider2195, line 40: Should be moved to a helper method so we don't have to maintain 2 copies of the same code

nautobot/extras/datasources/registry.py
```diff
@@ -48,20 +45,26 @@ def refresh_datasource_content(model_name, record, request, job_result, delete=F
...
                     )
-                    job_result.set_status(JobResultStatusChoices.STATUS_FAILURE)
-                job_result.save()
-            job_result.log(f"Data refresh from {record} complete!", level_choice=LogLevelChoices.LOG_INFO)
-            job_result.save()
+                    raise
     else:
-        # TODO(jathan): Review how we can eliminate these deliberate `set_status()` and `save()`
-        # calls,moving them to Celery-friendly primitives instead.
         for entry in get_datasource_contents(model_name):
```
gsnider2195, line 50: Should be moved to a helper method so we don't have to maintain 2 copies of the same code (see comment above)

nautobot/extras/managers.py
gsnider2195: If you're seeing it during normal job runs you need to make sure you're setting the `user_id` kwargs if you're calling `apply_async` manually instead of using `enqueue_job`

nautobot/core/jobs/__init__.py
```diff
@@ -0,0 +1,73 @@
+from nautobot.core.celery import register_jobs
+from nautobot.extras.choices import LogLevelChoices
+from nautobot.extras.datasources import ensure_git_repository, git_repository_dry_run, refresh_datasource_content
+from nautobot.extras.jobs import Job, ObjectVar
+from nautobot.extras.models import GitRepository
+
+name = "System Jobs"
+
+
+class GitRepositorySync(Job):
```
gsnider2195, line 10: Maybe set a `dryrun` BooleanVar here and merge the `GitRepositorySync` and `GitRepositoryDryRun` jobs?

nautobot/extras/api/serializers.py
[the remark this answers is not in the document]
gsnider2195: I believe celery uses a `GroupResult` [model](https://github.com/celery/django-celery-results/blob/28d9101843c146d66a7be35815fbc7a4acc8cb5c/django_celery_results/models.py#L175) to track chains of tasks. We might need to implement that.

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -815,6 +815,17 @@ The below is mostly relevant only to authors of Jobs and Nautobot Apps. End user
 
 The Configuration Contexts Metadata key `schema` has been replaced with `config_context_schema`. This means that any `schema` references in your git repository's data must be updated to reflect this change.
 
+`GitRepository` sync operations are now Jobs. As a result, when creating a new `GitRepository` it is not automatically synchronized. A `GitRepository.sync()` method has been implemented that will execute the sync job on a worker and return the `JobResult` for the operation. This method takes `dry_run` and `user` arguments. The `dry_run` argument defaults to `False`; if set to `True` will cause the sync to dry-run. The `user` argument is required if a sync is performed.
```
gsnider2195, line 818:
```suggestion
`GitRepository` sync operations are now Jobs. As a result, when creating a new `GitRepository` it is **NOT** automatically synchronized. A `GitRepository.sync()` method has been implemented that will execute the sync job on a worker and return the `JobResult` for the operation. This method takes `dry_run` and `user` arguments. The `dry_run` argument defaults to `False`; if set to `True` will cause the sync to dry-run. The `user` argument is required if a sync is performed.
```

nautobot/extras/datasources/git.py
```diff
@@ -282,14 +167,11 @@ def ensure_git_repository(
         head = repo_helper.checkout(from_branch, head)
         if repository_record.current_head != head:
             repository_record.current_head = head
-            # Make sure we don't recursively trigger a new resync of the repository!
-            repository_record.save(trigger_resync=False)
+            repository_record.save()
 
     except Exception as exc:
         if job_result:
```
gsnider2195, line 173: If we let this bubble up to the job the exception will end up in the jobresult.traceback field so this could probably go away?

nautobot/extras/datasources/git.py
```diff
@@ -328,9 +215,7 @@ def git_repository_dry_run(repository_record, job_result=None, logger=None):  #
...
-            job_result.set_status(JobResultStatusChoices.STATUS_FAILURE)
             job_result.log(str(exc), level_choice=LogLevelChoices.LOG_FAILURE, logger=logger)
```
gsnider2195, line 218: Would be nice if `JobResult.log` calls anywhere outside of a job class can be removed.

nautobot/extras/datasources/registry.py
```diff
@@ -29,16 +29,13 @@ def refresh_datasource_content(model_name, record, request, job_result, delete=F
...
         job_result (JobResult): Passed through to the callback functions to use with logging their actions.
         delete (bool): True if the record is being deleted; False if it is being created/updated.
     """
     job_result.log(f"Refreshing data provided by {record}...", level_choice=LogLevelChoices.LOG_INFO)
-    job_result.save()
-    if request:
-        change_context = JobChangeContext(user=request.user)
-        # TODO(jathan): Review how we can eliminate these deliberate `set_status()` and `save()`
-        # calls,moving them to Celery-friendly primitives instead.
+    if user:
```
gsnider2195, line 37: Change context code can go away. The only place that's passing a user to this method is a job and that's already being wrapped in a change context.

nautobot/extras/jobs.py
```diff
@@ -297,12 +297,15 @@ def class_path(cls):  # pylint: disable=no-self-argument
 
         Examples:
         local/my_script/MyScript
+        system/nautobot.core.jobs/MySystemJob
         plugins/my_plugin.jobs/MyPluginJob
         git.my-repository/myjob/MyJob
         """
         # TODO(Glenn): it'd be nice if this were derived more automatically instead of needing this logic
         if cls in registry["plugin_jobs"]:
```
gsnider2195, line 305: If everything is a real python module, is `source_grouping` even necessary?

nautobot/extras/managers.py
gsnider2195: This should be removed before merging unless you're still seeing problems related to the user being removed from the job result

nautobot/extras/models/datasources.py
```diff
@@ -67,8 +67,8 @@ class Meta:
         verbose_name_plural = "Git repositories"
 
     def __init__(self, *args, **kwargs):
-        # If instantiated from the REST API, the originating Request will be passed as a kwarg:
-        self.request = kwargs.pop("request", None)
+        # If instantiated from the API/UI, the originating User will be passed as a kwarg:
+        self.user = kwargs.pop("user", None)
```
gsnider2195, line 71: What exactly is this `__init__` method doing and can it be removed?

nautobot/extras/models/datasources.py
```diff
@@ -94,47 +101,47 @@ def to_csv(self):
...
+            user = self.user
 
-    def save(self, *args, trigger_resync=True, **kwargs):
-        super().save(*args, **kwargs)
+        from nautobot.extras.datasources import (
+            enqueue_pull_git_repository_and_refresh_data,
+            enqueue_git_repository_diff_origin_and_local,
+        )
 
-        def on_commit_callback():
```
gsnider2195, line 108: 💥

nautobot/extras/tests/example_jobs/test_read_only_fail.py
```diff
@@ -0,0 +1,24 @@
+from nautobot.dcim.models import Location, LocationType
```
gsnider2195, line 1: This was removed in the feature branch. Looks like an artifact of a merge?

nautobot/extras/tests/test_datasources.py
```diff
@@ -275,19 +279,24 @@ def create_empty_repo(path, url):
...
+                    source="system",
+                    repository=self.repo.pk,
+                )
 
                 self.assertEqual(
-                    self.job_result.status,
+                    job_result.status,
                     JobResultStatusChoices.STATUS_SUCCESS,
-                    self.job_result.data,
+                    job_result.result,
```
gsnider2195, line 293: the `run` method of this job doesn't return anything so shouldn't `job_result.result` be `None`?

gsnider2195: Might be better to use `job_result.traceback`?

Issue #3151, opened 2023-01-24
Support GitHub App authentication


chadell:
### As ...

Patti - Platform Admin

### I want ...

to support [GitHub Apps authentication](https://docs.github.com/en/developers/apps/building-github-apps/authenticating-with-github-apps)

### So that ...

more secure authentication (with expiration) for server-to-server calls can be used between Nautobot and GitHub
Today, only personal token is supported: https://docs.nautobot.com/projects/core/en/stable/models/extras/gitrepository/#repository-configuration

### I know this is done when...

I can use a secret group that provides the required GitHub App authentication

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [X] Plugins (and other Extensibility)
- [X] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

bryanculver:
Can you provide a workflow example where a GitHub App performs `git` operations against a repo? As I read their App documentation it would appear we would be required to implement the API functions of their integration, not `git`. Integrating Nautbot as a GitHub app is entirely it's own thing.

Also would this be better suited as a https://github.com/nautobot/nautobot-plugin-secrets-providers feature?

bryanculver:
@chadell ☝️

chadell:
My idea is, when integrating Nautobot with Git Repositories to allow a new authenticated method. So:

1. Create a GitHub App integration for Nautobot -> it generates a PEM private key
2. Use this PEM key to sign a JWT token to validate the identity
3. This JWT token would be used to interact with GitHub API https://docs.github.com/en/rest/git?apiVersion=2022-11-28#git-data

The PEM key would be a secret to store in some Nautobot secrets provider, similar to the current approach. 
It seems better for server-to-server, instead of creating a Nautobot Github USER account.

jathanism:
The problem we are faced with is we're using the [`GitPython`](https://gitpython.readthedocs.io/en/stable/) library which only supports SSH or HTTPS for interacting with Git repositories. It's not clear to me how Git clone/pull would be done over the GitHub API. The GitHub App authentication as far as I can tell appears to be for using the REST API.

[...]

PR #3548, opened 2023-04-06
[WIP] Doc Only: Update gitrepository.md about Github Apps

jeffkala:
# Closes: N/A
# What's Changed

Be more explict that the currently functionalities within Git Repos don't support GitHub apps.

# TODO


gsnider2195:
By github apps do you mean something like https://github.com/marketplace/gitguardian

jeffkala:
More like custom apps to try and solve not needing a service account.
https://docs.github.com/en/apps/creating-github-apps/creating-github-apps/about-apps

gsnider2195:
> More like custom apps to try and solve not needing a service account. https://docs.github.com/en/apps/creating-github-apps/creating-github-apps/about-apps

I see. Maybe it would be clearer to just say something like nautobot only supports repositories that can be cloned with the `git` cli app using `git clone`?

bryanculver:
I'm confused here. Along with #3151, where is the use case where your Git Repository is a Github App?

I can see the case for documentation stating we don't support Github App authentication but in review #3151 I'm not seeing how the REST API token can be used to perform a `git clone` onto the filesystem.

bryanculver:
Marking as WIP until feedback is addressed.

jeffkala:
This is all about the management and accessibility to repositories.  GH Apps provides some safeguards for this. Ultimately the goal of this PR is not to provide a fix or recommendation beyond just mentioning it in the docs that we don't support the functionality.  I'm happy to add additional details if needed.

jeffkala:
> > More like custom apps to try and solve not needing a service account. https://docs.github.com/en/apps/creating-github-apps/creating-github-apps/about-apps
> 
> I see. Maybe it would be clearer to just say something like nautobot only supports repositories that can be cloned with the `git` cli app using `git clone`?

I'm open to changing the terminology but do think GH Apps should be directly called out.

nautobot/docs/models/extras/gitrepository.md
```diff
@@ -14,6 +14,9 @@ Some text-based content is more conveniently stored in a separate Git repository
 
 When defining a Git repository for Nautobot to consume, the `name`, `remote URL`, and `branch` parameters are mandatory - the name acts as a unique identifier, and the remote URL and branch are needed for Nautobot to be able to locate and access the specified repository. Additionally, if the repository is private you may specify a `token` and any associated `username` that can be used to grant access to the repository.
 
+!!! note
+    GitHub Apps are not currently supported. GitHub Apps are not specifically focused on the workflow of creating/updating/deleting files. They can indeed perform these actions, but in order to accomplish that the client side MUST hit GitHub APIs instead of using standard command line or Python Git library tooling. At this time this functionality does not exist. 
+
```
jeffkala, line 19:
```suggestion
!!! note
    GitHub Apps are not currently supported. GitHub Apps are not specifically focused on the workflow of creating/updating/deleting files.  Nautobot only supports repositories that can be cloned using standard git command line using `git clone`. 

```

bryanculver, line 19:
```suggestion
!!! note
    Nautobot currently only supports repositories that can be cloned using the standard git command line, `git clone`. This means App-style integrations like GitHub Apps are not currently supported, as their workflow of managing files leverages a REST API.
```


@jeffkala?

jeffkala: committed suggestion. Thanks!

Issue #3345, opened 2023-02-21
Custom Relationships do not show in dynamic filter forms


jathanism:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.x
* Python version: n/a
* Database platform, version: demo.nautobot.com
* Middleware(s):

### Steps to Reproduce
1. Create relationship with source content type `dcim | device` and destination `ipam | VLAN`
2. Go to Device list view and click filter
3. Try to find "VLANs"

### Expected Behavior

The "VLANs" field for custom relationships.

### Observed Behavior

It does not show and cannot be used as a filter. In fact, no relationships will show in the "advanced" view filter view. Default filter view they show up.

jathanism:
It appears the issue is here:

https://github.com/nautobot/nautobot/blob/1b5d62dd861c0c0a0550d49802eb6ac356c02334/nautobot/core/utils/requests.py#L157

It has to do with the fact that `cr` filters are named with their destination e.g. `cr_device_to_vlans__destination`. And they appear to get filtered out incorrectly since it gets split into `cr_device_to_vlans` (which does not exist on the filter nor inner form fields). I would guess that custom relationships fields need to be excluded from this and other "split" logic.

There is some other logic that appears to be filtering out things with `__` in their name as well in the form logic: 

https://github.com/nautobot/nautobot/blob/1b5d62dd861c0c0a0550d49802eb6ac356c02334/nautobot/core/forms/forms.py#L327

Issue #2944, opened 2022-12-03
Relationship from/to object with big population makes object's list unusably slow 


apanagio:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.4 ( beb9c335ae90 )
* Python version: 3.10
* Database platform, version: PostgreSQL 14.5
* Middleware(s): apache2 reverse proxy

### Steps to Reproduce
1. Create one to many relationship from site to site
2. Create many sites (~30000)
3. Make sure relationship is not hidden in source nor destination object
4. Go to sites page (# per page does not matter)

### Expected Behavior
Load list of sites in less than 5 seconds

### Observed Behavior
List of sites loads in more than 60 seconds, sometimes it doesn't load at all.
If the relationship is hidden in source and destination, there is no delay

itdependsnetworks:
I am not the expert to comment if there is an easy or direct fix to the specific github issue opened here, that being said:

I think this is a good point. There are tradeoffs for performance and scalability compared to developer efficiency. While it is clear, there is no "one size fits all" recommendations we can make, I do think we can document better the performance impact in each design choice made.

In a similar issue, for golden config, I documented the potential impact here: https://docs.nautobot.com/projects/golden-config/en/latest/user/app_feature_sotagg/#performance which simply makes it more clear to the user to consider the performance impact for them.

bryanculver:
Computed fields and Relationships were discussed to be performance impacts and there isn’t much we can do to make them more-so; computed fields must be evaluated at display time and cannot be DB optimized, and relationships are generic foreign keys which have non-simple querying logic.

But yes, as Ken pointed out that there should be clear documentation to that effect in our docs. Because your relationships are site-to-site I believe this is impacting your performance significantly because it appears you’re looking up objects on a very large table on every row.

glennmatthews:
Does showing/hiding the table column for the relationship in question affect the performance at all? IMO if the relationship impacts the render performance even when its column is hidden, that feels like a bug to me.

bryanculver:
> Does showing/hiding the table column for the relationship in question affect the performance at all? IMO if the relationship impacts the render performance even when its column is hidden, that feels like a bug to me.

This is easily repeatable in the demo environment on Devices:

- List devices (`~1-2sec load`)
- Create a M2M relationship between Devices & Devices (named "Device Friends")
- Default column view exhibits normal performance (`~1-2sec load`)
[...]

PR #3550, opened 2023-04-07
Batch of things to improve working with relationships.

bryanculver:
# Closes: #3345, #2944
# What's Changed

- 🏎️ Do not evaluate the QuerySet on the FilterForm when we just need to see if one is set #2944
- 🏎️ Allow an instantiated FilterSet to be passed into machinery for DynamicFilterForm because often times we already have one but if you pass in a class (old behavior) instantiate it like we always were (empty). Saves time doing the Relationship lookups and processing 5 times ~#2944
- 💡 Add support for Relationships on the DynamicFilterForm #3345
- 🎨 Make the label on filters Relationships human-readable (missed in #3010 because `.label` would need a side) 

**Filtering by Relationships on the DynamicFilterForm**
Screenshot 2023-04-07 at 12 48 25

**Old Display of Filtered Relationships**
Screenshot 2023-04-07 at 12 47 18

**New Display of Filtered Relationships**
Screenshot 2023-04-07 at 12 46 46

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/core/tests/test_utilities.py
+++ b/nautobot/core/tests/test_utilities.py
@@ -1,7 +1,7 @@
 from django.test import TestCase
 from django.contrib.contenttypes.models import ContentType
 
-from nautobot.core.utilities import check_filter_for_display, get_filter_field_label, field_name_to_display
+from nautobot.core.utilities import check_filter_for_display, get_filter_field_label, _field_name_to_display
 
 from nautobot.dcim.filters import DeviceFilterSet
 from nautobot.dcim.models import Device, DeviceType, DeviceRedundancyGroup
@@ -115,7 +115,7 @@ class GetFilterFieldLabelTest(TestCase):
         device_ct = ContentType.objects.get_for_model(Device)
         cls.peer_relationship = Relationship(
             name="HA Device Peer",
-            slug="ha-device-peer",
+            slug="ha_device_peer",
             source_type=device_ct,
             destination_type=device_ct,
             source_label="Peer",
@@ -131,8 +131,6 @@ class GetFilterFieldLabelTest(TestCase):
     def test_get_filter_field_label(self):
 
         device_filter_set_filters = DeviceFilterSet().filters
-
-        print(device_filter_set_filters)
 
         with self.subTest("Simple field name"):
             self.assertEqual(get_filter_field_label(device_filter_set_filters["id"]), "Id")
@@ -155,21 +153,21 @@ class GetFilterFieldLabelTest(TestCase):
 
 class FieldNameToDisplayTest(TestCase):
     """
-    Validate the operation of field_name_to_display().
+    Validate the operation of _field_name_to_display().
     """
 
-    def test_field_name_to_display(self):
+    def test__field_name_to_display(self):
 
         with self.subTest("id => Id"):
-            self.assertEqual(field_name_to_display("id"), "Id")
+            self.assertEqual(_field_name_to_display("id"), "Id")
 
         with self.subTest("device_type => Device Type"):
-            self.assertEqual(field_name_to_display("device_type"), "Device type")
+            self.assertEqual(_field_name_to_display("device_type"), "Device type")
 
         with self.subTest("_custom_field_data__site_type => Site Type"):
-            self.assertEqual(field_name_to_display("_custom_field_data__site_type"), "Site type")
+            self.assertEqual(_field_name_to_display("_custom_field_data__site_type"), "Site type")
 
         with self.subTest("cr_sister_sites__peer => Peer"):
             # This shouldn't ever be an input because get_filter_field_label
             # will use the label from the custom field instead of the field name
-            self.assertEqual(field_name_to_display("cr_sister_sites__peer"), "Cr_sister_sites peer")
+            self.assertEqual(_field_name_to_display("cr_sister_sites__peer"), "Cr_sister_sites peer")
--- a/nautobot/core/tests/test_utilities.py
+++ b/nautobot/core/tests/test_utilities.py
@@ -1,9 +1,12 @@
 from django.test import TestCase
+from django.contrib.contenttypes.models import ContentType
 
-from nautobot.core.utilities import check_filter_for_display
+from nautobot.core.utilities import check_filter_for_display, get_filter_field_label, field_name_to_display
 
 from nautobot.dcim.filters import DeviceFilterSet
-from nautobot.dcim.models import DeviceType, DeviceRedundancyGroup
+from nautobot.dcim.models import Device, DeviceType, DeviceRedundancyGroup
+from nautobot.extras.choices import RelationshipTypeChoices
+from nautobot.extras.models import Relationship, CustomField
 
 
 class CheckFilterForDisplayTest(TestCase):
@@ -42,7 +45,7 @@ class CheckFilterForDisplayTest(TestCase):
         with self.subTest("Test get field label, none exists (fallback)"):
             expected_output = {
                 "name": "id",
-                "display": "id",
+                "display": "Id",
                 "values": [{"name": "example_field_value", "display": "example_field_value"}],
             }
 
@@ -100,3 +103,73 @@ class CheckFilterForDisplayTest(TestCase):
             self.assertEqual(
                 check_filter_for_display(device_filter_set_filters, "manufacturer", ["fake_slug"]), expected_output
             )
+
+
+class GetFilterFieldLabelTest(TestCase):
+    """
+    Validate the operation of get_filter_field_label().
+    """
+
+    @classmethod
+    def setUpTestData(cls):
+        device_ct = ContentType.objects.get_for_model(Device)
+        cls.peer_relationship = Relationship(
+            name="HA Device Peer",
+            slug="ha-device-peer",
+            source_type=device_ct,
+            destination_type=device_ct,
+            source_label="Peer",
+            destination_label="Peer",
+            type=RelationshipTypeChoices.TYPE_ONE_TO_ONE_SYMMETRIC,
+        )
+        cls.peer_relationship.validated_save()
+
+        cls.custom_field = CustomField(slug="labeled_custom_field", label="Moo!", type="text")
+        cls.custom_field.validated_save()
+        cls.custom_field.content_types.add(device_ct)
+
+    def test_get_filter_field_label(self):
+
+        device_filter_set_filters = DeviceFilterSet().filters
+
+        print(device_filter_set_filters)
+
+        with self.subTest("Simple field name"):
+            self.assertEqual(get_filter_field_label(device_filter_set_filters["id"]), "Id")
+
+        with self.subTest("Semi-complex field name"):
+            self.assertEqual(get_filter_field_label(device_filter_set_filters["has_interfaces"]), "Has interfaces")
+
+        with self.subTest("Relationship field name"):
+            self.assertEqual(
+                get_filter_field_label(device_filter_set_filters[f"cr_{self.peer_relationship.slug}__peer"]),
+                self.peer_relationship.source_label,
+            )
+
+        with self.subTest("Custom field with label"):
+            self.assertEqual(
+                get_filter_field_label(device_filter_set_filters[f"cf_{self.custom_field.slug}"]),
+                "Moo!",
+            )
+
+
+class FieldNameToDisplayTest(TestCase):
+    """
+    Validate the operation of field_name_to_display().
+    """
+
+    def test_field_name_to_display(self):
+
+        with self.subTest("id => Id"):
+            self.assertEqual(field_name_to_display("id"), "Id")
+
+        with self.subTest("device_type => Device Type"):
+            self.assertEqual(field_name_to_display("device_type"), "Device type")
+
+        with self.subTest("_custom_field_data__site_type => Site Type"):
+            self.assertEqual(field_name_to_display("_custom_field_data__site_type"), "Site type")
+
+        with self.subTest("cr_sister_sites__peer => Peer"):
+            # This shouldn't ever be an input because get_filter_field_label
+            # will use the label from the custom field instead of the field name
+            self.assertEqual(field_name_to_display("cr_sister_sites__peer"), "Cr_sister_sites peer")
--- a/nautobot/core/tests/test_utilities.py
+++ b/nautobot/core/tests/test_utilities.py
@@ -10,11 +10,8 @@ from nautobot.extras.models import Relationship, CustomField
 
 
 class CheckFilterForDisplayTest(TestCase):
-    """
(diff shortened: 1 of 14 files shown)


nautobot/core/utilities.py
```diff
@@ -1,5 +1,6 @@
 from django.core.exceptions import FieldError
 
+from nautobot.extras.filters.mixins import RelationshipFilter
```
glennmatthews, line 3: I don't *like* importing from `nautobot.extras` into `nautobot.core` but I guess if it's not introducing a circular dependency it's OK for now.

nautobot/core/utilities.py
```diff
@@ -29,13 +30,19 @@ def check_filter_for_display(filters, field_name, values):
...
+        or field.label
+        or field_name
+    )
+    if len(values) == 0 or not hasattr(field, "queryset") or not is_uuid(values[0]):
         return resolved_filter
     else:
         try:
             new_values = []
-            for value in filters[field_name].queryset.filter(pk__in=values):
+            for value in field.queryset.filter(pk__in=values):
```
glennmatthews, line 45: For clarity, can we change `field` to `filter_field` throughout here?

nautobot/core/utilities.py
```diff
@@ -29,13 +30,19 @@ def check_filter_for_display(filters, field_name, values):
...
 
-    resolved_filter["display"] = filters[field_name].label or field_name
-    if len(values) == 0 or not hasattr(filters[field_name], "queryset") or not is_uuid(values[0]):
+    field = filters[field_name]
+
+    resolved_filter["display"] = (
+        (isinstance(field, RelationshipFilter) and field.relationship.get_label(side=field.side))
+        or field.label
+        or field_name
+    )
```
glennmatthews, line 39: That's a bit of dark magic there relying on `a and b` to return `b` - definitely caused me to do a double-take of "That can't possibly be right, can it?". Maybe clearer even if more verbose:

```suggestion
    resolved_filter["display"] = None
    if isinstance(field, RelationshipFilter):
        resolved_filter["display"] = field.relationship.get_label(side=field.side))
    if not resolved_filter["display"]:
        resolved_filter["display"] = field.label or field_name
```

nautobot/core/views/generic.py
```diff
@@ -255,17 +254,17 @@ def get(self, request):
                 self.queryset = self.queryset.none()
 
             display_filter_params = [
-                check_filter_for_display(filterset_filters, field_name, values)
+                check_filter_for_display(filterset.filters, field_name, values)
```
glennmatthews, line 257: Just checking - is `filterset.filters` equivalent to `filterset.get_filters()`, or are there subtle differences we need to be aware of?

nautobot/core/views/generic.py
```diff
@@ -255,17 +254,17 @@ def get(self, request):
...
                 for field_name, values in filter_params.items()
             ]
 
             if request.GET:
                 factory_formset_params = convert_querydict_to_factory_formset_acceptable_querydict(
-                    request.GET, self.filterset
+                    request.GET, filterset
                 )
-                dynamic_filter_form = DynamicFilterFormSet(filterset_class=self.filterset, data=factory_formset_params)
+                dynamic_filter_form = DynamicFilterFormSet(filterset_class=filterset, data=factory_formset_params)
```
glennmatthews, line 265: Can we rename the `filterset_class` param to `filterset` to reflect its change in usage, or is that an API breakage?

nautobot/utilities/forms/forms.py
```diff
@@ -279,7 +279,9 @@ def __init__(self, *args, filterset_class=None, **kwargs):
         model = self.filterset_class._meta.model
 
         if self.filterset_class is not None:
-            filterset_class = self.filterset_class()
+            filterset_class = self.filterset_class
```
glennmatthews, line 282: From an API standpoint, I don't like passing an *instance* to a param specifically named `filterset_class`. Can we maybe augment the params to include a `filterset` param as a separate option instead?

nautobot/utilities/forms/forms.py
```diff
@@ -324,17 +326,27 @@ def capitalize(field):
...
     def _get_lookup_field_choices(self):
         """Get choices for lookup_fields i.e filterset parameters without a lookup expr"""
+        from nautobot.extras.filters.mixins import RelationshipFilter  # Avoid circular import
+
         filterset_without_lookup = (
-            (name, field.label or self.capitalize(field.field_name))
+            (
+                name,
+                field.label
+                or (isinstance(field, RelationshipFilter) and field.relationship.get_label(side=field.side))
```
glennmatthews, line 335: Same here, personally I'd prefer a more verbose but less "magical" way of setting this value.

nautobot/utilities/forms/forms.py
```diff
@@ -324,17 +326,27 @@ def capitalize(field):
...
-            if "__" not in name and name != "q"
+            if isinstance(field, RelationshipFilter) or ("__" not in name and name != "q")
         )
         return sorted(filterset_without_lookup, key=lambda x: x[1])
 
 
 def dynamic_formset_factory(filterset_class, data=None, **kwargs):
     filter_form = DynamicFilterForm
-    filter_form.filterset_class = filterset_class
+    if isinstance(filterset_class, type):
```
glennmatthews, line 346: Same here, can we let this function take a `filterset` as an argument instead of permitting an instance as the so-called `filterset_class`?

nautobot/utilities/utils.py
```diff
@@ -844,6 +844,13 @@ def get_all_lookup_expr_for_field(model, field_name):
                     "name": build_lookup_label(name, field.lookup_expr),
                 }
             )
+        elif name == field_name and not name.startswith("has_"):
```
glennmatthews, line 847: This is to fix #3345? Can we fix it at its source rather than continuing to add to this function?

nautobot/utilities/utils.py
```diff
@@ -964,7 +971,9 @@ def convert_querydict_to_factory_formset_acceptable_querydict(request_querydict,
         ... }
     """
     query_dict = QueryDict(mutable=True)
-    filterset_class_fields = filterset_class().filters.keys()
+    if isinstance(filterset_class, type):
```
glennmatthews, line 974: blah blah `filterset_class` versus `filterset`...

jathanism (commented):
Comments.

nautobot/core/tests/test_utilities.py
```diff
@@ -42,7 +45,7 @@ def test_check_filter_for_display(self):
         with self.subTest("Test get field label, none exists (fallback)"):
             expected_output = {
                 "name": "id",
-                "display": "id",
+                "display": "Id",
```
glennmatthews, line 48: Any chance we can fix the behavior to correctly display `"ID"`? Not blocking.

nautobot/core/tests/test_utilities.py
```diff
@@ -100,3 +103,73 @@ def test_check_filter_for_display(self):
...
+
+        cls.custom_field = CustomField(slug="labeled_custom_field", label="Moo!", type="text")
+        cls.custom_field.validated_save()
+        cls.custom_field.content_types.add(device_ct)
+
+    def test_get_filter_field_label(self):
+
+        device_filter_set_filters = DeviceFilterSet().filters
+
+        print(device_filter_set_filters)
```
glennmatthews, line 135: Remove?

nautobot/core/tests/test_utilities.py
```diff
@@ -100,3 +103,73 @@ def test_check_filter_for_display(self):
...
+    """
+    Validate the operation of get_filter_field_label().
+    """
+
+    @classmethod
+    def setUpTestData(cls):
+        device_ct = ContentType.objects.get_for_model(Device)
+        cls.peer_relationship = Relationship(
+            name="HA Device Peer",
+            slug="ha-device-peer",
```
glennmatthews, line 118:
```suggestion
            slug="ha_device_peer",
```

nautobot/core/utilities.py
```diff
@@ -30,22 +29,52 @@ def check_filter_for_display(filters, field_name, values):
...
+    """
+    from nautobot.extras.filters.mixins import RelationshipFilter
+    from nautobot.extras.filters.customfields import CustomFieldFilterMixin
+
+    if filter_field.label:
+        return filter_field.label
+    elif isinstance(filter_field, RelationshipFilter):
+        return filter_field.relationship.get_label(side=filter_field.side)
+    elif isinstance(filter_field, CustomFieldFilterMixin):
+        return filter_field.custom_field.label
```
glennmatthews, line 68: Perhaps? Would let us remove the inner import.

```suggestion
    elif hasattr(filter_field, "relationship"):
        return filter_field.relationship.get_label(side=filter_field.side)
    elif hasattr(filter_field, "custom_field"):
        return filter_field.custom_field.label
```

nautobot/core/utilities.py
```diff
@@ -29,16 +29,52 @@ def check_filter_for_display(filters, field_name, values):
...
+    else:
+        return field_name_to_display(filter_field.field_name)
+
+
+def field_name_to_display(field_name):
```
glennmatthews, line 73: Possibly prepend a `_` to this function to mark it as private? Should always get here by going through `get_filter_field_label` rather than calling this function directly, right?

nautobot/core/views/generic.py
glennmatthews, line 257: Same change needed in `nautobot/core/views/renderers.py`? In `get_context()`?

nautobot/utilities/tests/test_forms.py
```diff
@@ -826,23 +820,26 @@ def test_dynamic_filter_form_with_data_and_prefix(self):
         request_querydict.setdefault("created", "2022-09-05")
         request_querydict.setlistdefault("asn", ["4"])
 
+        site_filter_set = SiteFilterSet()
```
glennmatthews, line 823: This batch of changes will be fun to merge into `next`...

nautobot/utilities/utils.py
```diff
@@ -977,29 +988,32 @@ def convert_querydict_to_factory_formset_acceptable_querydict(request_querydict,
...
     request_querydict.pop("q", None)
-    for lookup_type, value in request_querydict.items():
+    for filter_field_name, value in request_querydict.items():
         # Discard fields without values
         if value:
-            if lookup_type in filterset_class_fields:
-                lookup_field = re.sub(r"__\w+", "", lookup_type)
-                lookup_value = request_querydict.getlist(lookup_type)
+            if filter_field_name in filterset_class_fields:
+                if isinstance(filterset.filters[filter_field_name], RelationshipFilter):
```
glennmatthews, line 995: Another case where we might be able to use `hasattr(..., "relationship")` to avoid the need for the inner import?

nautobot/utilities/utils.py
```diff
@@ -977,29 +988,32 @@ def convert_querydict_to_factory_formset_acceptable_querydict(request_querydict,
...
+                    lookup_field = filter_field_name
+                else:
+                    lookup_field = re.sub(r"__\w+", "", filter_field_name)
```
glennmatthews, line 998: Might be good to add an explanatory comment or two here for posterity?

glennmatthews (approved):

nautobot/core/tests/test_utilities.py
bryanculver: Copied from the existing Relationship tests but I'll change here.

nautobot/core/tests/test_utilities.py
bryanculver: Oops! I'll remove.

nautobot/core/utilities.py
bryanculver: 🦆

nautobot/core/utilities.py
bryanculver: Yeah at the moment I don't see other needs to use it.

nautobot/utilities/tests/test_forms.py
bryanculver: Yeah, big `:oof:` when I read these tests. We should have used Location or something else. 🤷

nautobot/core/views/generic.py
bryanculver: Ugh... Just now realized the instantiation of the filterset is in a mixin...

So this: https://github.com/nautobot/nautobot/blob/372d01e47a46838c5816ffc04758934bb162351e/nautobot/core/views/renderers.py#L160


Does this under the hood: https://github.com/nautobot/nautobot/blob/372d01e47a46838c5816ffc04758934bb162351e/nautobot/core/views/mixins.py#L454

nautobot/core/tests/test_utilities.py
```diff
@@ -100,3 +103,71 @@ def test_check_filter_for_display(self):
...
+            destination_label="Peer",
+            type=RelationshipTypeChoices.TYPE_ONE_TO_ONE_SYMMETRIC,
+        )
+        cls.peer_relationship.validated_save()
+
+        cls.custom_field = CustomField(slug="labeled_custom_field", label="Moo!", type="text")
+        cls.custom_field.validated_save()
+        cls.custom_field.content_types.add(device_ct)
+
+    def test_get_filter_field_label(self):
```
jathanism, line 131: Docstring pls.

nautobot/core/tests/test_utilities.py
```diff
@@ -100,3 +103,71 @@ def test_check_filter_for_display(self):
...
+                "Moo!",
+            )
+
+
+class FieldNameToDisplayTest(TestCase):
+    """
+    Validate the operation of _field_name_to_display().
+    """
+
+    def test__field_name_to_display(self):
```
jathanism, line 159: Docstring pls.

nautobot/core/views/renderers.py
```diff
@@ -158,7 +158,10 @@ def get_context(self, data, accepted_media_type, renderer_context):
             if view.action == "list":
                 if view.filterset_class is not None:
                     view.queryset = view.filter_queryset(view.get_queryset())
-                    filterset_filters = view.filterset.get_filters()
+                    if view.filterset is not None:
+                        filterset_filters = view.filterset.filters
+                    else:
+                        filterset_filters = view.filterset.get_filters()
```
jathanism, line 164: This appears as if the `else` will never be called. So it seems like `view.filter_queryset()` has a side effect of setting `view.filterset`... 😬

jathanism (approved):

nautobot/core/views/renderers.py
bryanculver: Side-effect from line 60 but it makes it's own assumptions that `view.filterset` was set... 🤷

Issue #2316, opened 2022-08-26
Drop Python 3.7 Support


bryanculver:
### Proposed Changes
Drop Python 3.7 support.

### Justification
3.7 stops receiving security fixes in July 2023. Dependent libraries may drop support earlier but commonly not soon after.

### TODO

- [x] Remove publish of package & container support
- [x] Switch default container image to newer version (doesn't have to be 3.8)

itdependsnetworks:
Perhaps as an alternate to this, let's just define the process and put this conversation to bed. Something *like* this could be good.

* In the immediate release after Python security will no longer be maintained, Nautobot support will be removed
* The removal will be in semver Minor versions
* They will be announced in the release notes

glennmatthews:
Done in #3554.

PR #3554, opened 2023-04-07
Drop python 3.7 support for 2.0

glennmatthews:
# Closes: #2316 
# What's Changed

- Since Python 3.7 will be EOL as of 2023-06-27, there's no plan at present for Nautobot 2.0 to support Python 3.7.
- Move minimum Python version to 3.8.1. (some of our updated dependencies explicitly don't support 3.8.0).
   - Update default dev environment and default Docker image to use Python 3.8.
- Set an upper bound on the supported Python version as we haven't yet validated or tested against 3.11 or later.
- Update many dependencies to latest versions made possible now that we've dropped 3.7
- Update many development dependencies, including `black`, `flake8`, and `pylint`, to their latest versions, and address any new complaints these linters have.

# TODO
- [x] Unit, Integration Tests


jathanism (approved):

examples/example_plugin/example_plugin/metrics.py
```diff
@@ -4,7 +4,6 @@
 
 
 def metric_example():
-
```
itdependsnetworks, line 7: Does it make sense to run the latest black on develop and pull that into next vs potentially fighting the other way around?

bryanculver: Likely would pose a conflict:

- We can’t install the newer version of Black\Flake8 on `develop` because of 3.7 support
- If we force the install locally and run it likely will complain in CI because the rules change and won’t pass
- If CI doesn’t complain those changes won’t stay and every PR going forward will still have to deal with the black/flake8 changes

itdependsnetworks: > We can’t install the newer version of Black\Flake8 on develop because of 3.7 support

https://github.com/nautobot/nautobot/commit/efa719c3af9ba5298fc6bf735cd8a739d90aa436#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R137

itdependsnetworks: I don’t think we have gone away from them, https://github.com/nautobot/nautobot/blob/develop/pyproject.toml#L189

bryanculver: > @itdependsnetworks Do you see value in moving to a newer version of linters/formatters like pylint and black in the LTM `develop` branch? I don't immediately see a reason to do so, unless the lack of updates is causing challenges for app maintainers?
> 
> If there's a reason/need to do so, I'm happy to do so in a separate PR, at least for `black` and `pylint` - the updated version of `black` touched a lot of files but only in very trivial ways, and making the updated `pylint` happy was fairly trivial as well.
> 
> Note that we **cannot** update `flake8` without dropping 3.7 support as there are conflicts between newer versions of `flake8` and the version of `importlib-metadata` that we need for our 3.7 support.

I don't think we should drop 3.7 support in `develop` for a dev dependency.

glennmatthews: > I don't think we should drop 3.7 support in `develop` for a dev dependency.

To be clear, I wasn't proposing to do so. :-)

itdependsnetworks: I do see value in having black be on the newer version in dev. 

Having different versions of black around tends to be problematic, so I guess that is the only one that I am thinking about. My workflow (and seemingly many other's based on the PR's I see in other projects), people tend to use newer black. This causes a see-saw of going back and forth between merge conflicts and takes a lot of time and overall confusion that can I think we can address. 

Also, I would leave flake8 as a dev dependency, meaning 3.8.1 only seems to be on be an issue for flake, and not the project. This is my understanding of the line `Dropped support for Python 3.7. Python 3.8.1 is now the minimum version required by Nautobot.`, even though I don't see the update in the toml file, but did note that in the towncrier.

glennmatthews: In this morning's commits I changed the minimum version back to 3.8.0 and restricted the `flake8` dev dependency to 3.8.1 and later, but I missed updating the change fragment as well. Will fix, thanks!

gsnider2195 (approved):

examples/example_plugin/example_plugin/metrics.py
glennmatthews: Opened #3561 for making similar dependency updates in `develop` where appropriate.

Issue #3116, opened 2023-01-13
Tests are making assumptions about underlying data


PavelSafronov:
### Environment
* Nautobot version (Docker tag too if applicable): tip of master, tested with commit 0341ca3783b78c9f6550f9871de55420a59e7bea
* Python version: 3.7
* Database platform, version: unsure
* Middleware(s): unsure

### Steps to Reproduce
1. open https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/tests/test_views.py
2. on line 286, change the JSON string in a trivial way, either by adding/removing spaces or by swapping the order of `type` and `properties` fields
  - for example, `"data_schema": '{"type": "object", "properties": {"baz": {"type": "string"}}},'` => `"data_schema": '{"type":"object","properties":{"baz":{"type":"string"}}}',` 
4. run tests, for example `inv unittest -l nautobot.extras.tests.test_views.ConfigContextSchemaTestCase.test_edit_object_with_permission --debug --keepdb --verbose --skip-docs-build`

### Expected Behavior

Tests pass.

### Observed Behavior

Test fails with an error:
```
FAIL: test_edit_object_with_permission (nautobot.extras.tests.test_views.ConfigContextSchemaTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/django/test/utils.py", line 387, in inner
    return func(*args, **kwargs)
  File "/source/nautobot/utilities/testing/views.py", line 447, in test_edit_object_with_permission
    self.assertInstanceEqual(self._get_queryset().get(pk=instance.pk), self.form_data)
  File "/source/nautobot/utilities/testing/mixins.py", line 203, in assertInstanceEqual
    self.assertEqual(new_model_dict, relevant_data)
AssertionError: {'nam[52 chars]{"type": "object", "properties": {"baz": {"type": "string"}}}'} != {'nam[52 chars]{"type":"object","properties":{"baz":{"type":"string"}}}'}
- {'data_schema': '{"type": "object", "properties": {"baz": {"type": "string"}}}',
?                          -         -             -       -        -

+ {'data_schema': '{"type":"object","properties":{"baz":{"type":"string"}}}',
   'name': 'Schema X',
   'slug': 'schema-x'}
```

[...]

PR #3568, opened 2023-04-11
Update assertInstanceEqual to fix JSON comparison of 'data_schema' keys.

PavelSafronov:
This fixes the issue where assertInstanceEqual test code was relying on JSON columns to have the same formatting as the incoming data, which is typically not the case.

# Closes: #3116
# What's Changed
Update assertInstanceEqual to fix JSON comparison of 'data_schema' keys.
This fixes the issue where assertInstanceEqual test code was relying on JSON columns to have the same formatting as the incoming data, which is typically not the case.
This only impacts test code.

### DIFF
--- a/nautobot/extras/tests/test_views.py
+++ b/nautobot/extras/tests/test_views.py
@@ -284,7 +284,7 @@ class ConfigContextSchemaTestCase(
         cls.form_data = {
             "name": "Schema X",
             "slug": "schema-x",
-            "data_schema": '{"type": "object","properties": {"baz": {"type": "string"}}}',
+            "data_schema": '{"type": "object","properties": {"baz": {"type": "string"}}}',  # Intentionally misformatted (missing space) to ensure proper formatting on output
         }
 
         cls.bulk_edit_data = {
--- a/nautobot/extras/tests/test_views.py
+++ b/nautobot/extras/tests/test_views.py
@@ -284,7 +284,7 @@ class ConfigContextSchemaTestCase(
         cls.form_data = {
             "name": "Schema X",
             "slug": "schema-x",
-            "data_schema": '{"type": "object", "properties": {"baz": {"type": "string"}}}',
+            "data_schema": '{"type": "object","properties": {"baz": {"type": "string"}}}',
         }
 
         cls.bulk_edit_data = {
--- a/nautobot/utilities/testing/mixins.py
+++ b/nautobot/utilities/testing/mixins.py
@@ -187,6 +187,9 @@ class NautobotTestCaseMixin:
             if isinstance(v, list):
                 # Sort lists of values. This includes items like tags, or other M2M fields
                 new_model_dict[k] = sorted(v)
+            elif k == "data_schema" and isinstance(v, str):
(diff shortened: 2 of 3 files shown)


nautobot/extras/tests/test_views.py
```diff
@@ -284,7 +284,7 @@ def setUpTestData(cls):
         cls.form_data = {
             "name": "Schema X",
             "slug": "schema-x",
-            "data_schema": '{"type": "object", "properties": {"baz": {"type": "string"}}}',
+            "data_schema": '{"type": "object","properties": {"baz": {"type": "string"}}}',
```
jathanism, line 287: Please revert this line.

nautobot/utilities/testing/mixins.py
```diff
@@ -214,6 +220,11 @@ def assertQuerysetEqualAndNotEmpty(self, qs, values, *args, **kwargs):
     # Convenience methods
     #
 
+    def standardize_json(self, jsonText):
+        obj = json.loads(jsonText)
+        newJsonText = json.dumps(obj, sort_keys=True)
+        return newJsonText
```
jathanism, line 226:
```suggestion
        return json.dumps(obj, sort_keys=True)
```

jathanism (changes requested):
Thanks for the contribution!

nautobot/utilities/testing/mixins.py
```diff
@@ -214,6 +220,11 @@ def assertQuerysetEqualAndNotEmpty(self, qs, values, *args, **kwargs):
     # Convenience methods
     #
 
+    def standardize_json(self, jsonText):
+        obj = json.loads(jsonText)
```
jathanism, line 224: I know I'm being pedantic, but we don't use `camelCase` in Python and our linters will scream.

```suggestion
    def standardize_json(self, data):
        obj = json.loads(data)
```

jathanism (changes requested):

nautobot/utilities/testing/mixins.py
PavelSafronov: Will make this change.

nautobot/utilities/testing/mixins.py
PavelSafronov: No problem, thanks for clarifying, fixing! This is my first time with Python, so I'm happy to learn the conventions.

nautobot/extras/tests/test_views.py
jathanism: Please keep the line and add the comment to explain for posterity. Thank you!

jathanism (approved):
Looking great! Thanks for the straightforward fix. If you could please just address the one comment, this is good to go.

nautobot/extras/tests/test_views.py
PavelSafronov: Looks like @bryanculver got there first and added the comment about formatting, no action required on my part. Thanks!

bryanculver: 😎

glennmatthews (approved):

PR #3571, opened 2023-04-11
[WIP/Prototype] Forking Unittests by App

bryanculver:
# Closes: #DNE
# What's Changed
Hacky attempt at speeding up unittests. Output is garbage but locally it let me run unittests in under 9 minutes on my Macbook Air. I'm sure there are many improvements here (for example pool size should be equal to core count).

Leaving this here to maybe pick up later.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -172,11 +172,14 @@ def run_command(context, command, **kwargs):
             print_command(command, env=env)
         context.run(command, pty=True, env=env, **kwargs)
     else:
+        test_db_name = kwargs.pop("test_db_name", "")
+        if test_db_name:
+            test_db_name = f"-e NAUTOBOT_DB_TEST_NAME={test_db_name}"
         # Check if Nautobot is running; no need to start another Nautobot container to run a command
         docker_compose_status = "ps --services --filter status=running"
         results = docker_compose(context, docker_compose_status, hide="out")
         if "nautobot" in results.stdout:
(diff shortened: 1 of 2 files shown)


tasks.py
```diff
@@ -709,7 +713,39 @@ def unittest(
...
+    else:
+        run_command(context, command)
+
+
+def unittest_forked(fork_context):
+    unittest(
+        fork_context[0],
+        failfast=True,
+        label=f"nautobot.{fork_context[1]}",
+        skip_docs_build=True,
```
jathanism, line 727: This might be problematic for the tests that require docs unless you've asserted that they were built first (or only once) prior to running the forked tests.

tasks.py
```diff
@@ -666,14 +669,15 @@ def unittest(
     keepdb=False,
     label="nautobot",
     failfast=False,
-    buffer=True,
+    buffer=False,
```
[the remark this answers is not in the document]
bryanculver: Again with the above: just some hacks for speed and testing. No way near a final implementation.

tasks.py
bryanculver: Looking way too far into it: This was committed as-is to get a performance comparison. The interface isn't robust at all. But I didn't want every fork rebuilding docs.

jathanism: Hey man. I want this to happen, okay! 😎

Issue #3391, opened 2023-03-06
Unable to select multiple entries on (Multiple) selection custom fields in filter


phoenixswiss:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.11
* Python version: 3.8.10
* Database platform, version: postgresql, Ubuntu 12.12-0ubuntu0.20.04.1
* Middleware(s): -

### Steps to Reproduce
1. Create a custom field (selection, or multiple selection) with multiple options and assign it to interfaces or devices
2. Fill the custom field with some information
3. Open the filter on the device or interface overview and switch to the Default tab

**on custom field of the type selection**

5. move to the custom field and try to select multiple entries

**on custom field of the type multiple selection**

5. move to the custom field and select multiple entries

6. Press Apply

### Expected Behavior
As with the Device field, multiple entries can be selected and all entries matching the filter will be displayed.

### Observed Behavior
**on custom field of the type selection**
Only one entry can be selected.

**on custom field of the type multiple selection**
Multiple entries can be selected, but only the last selected entry will be used for the filter.

bryanculver:
I believe this is related to: https://github.com/nautobot/nautobot/pull/3342

lampwins:
Appears fixed in 2.0.4

PR #3574, opened 2023-04-12
Fix advance filter error on multiple custom filter

timizuoebideri1:
# Closes: #3391 
# What's Changed

Error reported was caused by `is_single_choice_field` returning `True` for `CustomFieldMultiSelectFilter` because `CustomFieldMultiSelectFilter` inherits from `django_filters.Filter` and not `django_filters.MultipleChoiceFilter`.

Also fixed an additional bug where the page freeze after selecting a custom multiple field choice.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/models/customfields.py
+++ b/nautobot/extras/models/customfields.py
@@ -558,14 +558,13 @@ class CustomField(BaseModel, ChangeLoggedModel, NotesMixin):
             choices = [(cfc.value, cfc.value) for cfc in self.choices.all()]
             default_choice = self.choices.filter(value=self.default).first()
 
-            if not required or default_choice is None:
-                choices = add_blank_choice(choices)
-
             # Set the initial value to the first available choice (if any)
             if set_initial and default_choice:
                 initial = default_choice.value
 
             if self.type == CustomFieldTypeChoices.TYPE_SELECT:
+                if not required or default_choice is None:
+                    choices = add_blank_choice(choices)
                 field_class = CSVChoiceField if for_csv_import else forms.ChoiceField
                 field = field_class(
                     choices=choices,
--- a/nautobot/utilities/utils.py
+++ b/nautobot/utilities/utils.py
@@ -1013,9 +1013,11 @@ def convert_querydict_to_factory_formset_acceptable_querydict(request_querydict,
 
 
 def is_single_choice_field(filterset, field_name):
+    from nautobot.extras.filters import CustomFieldMultiSelectFilter  # Avoid circular imports
+
     # Some filter parameters do not accept multiple values, e.g DateTime, Boolean, Int fields and the q field, etc.
     field = get_filterset_field(filterset, field_name)
-    return not isinstance(field, django_filters.MultipleChoiceFilter)
+    return not isinstance(field, (django_filters.MultipleChoiceFilter, CustomFieldMultiSelectFilter))
 
 
 def get_filterable_params_from_filter_params(filter_params, non_filter_params, filterset):


nautobot/extras/models/customfields.py
```diff
@@ -558,14 +558,13 @@ def to_form_field(
...
-                choices = add_blank_choice(choices)
-
             # Set the initial value to the first available choice (if any)
             if set_initial and default_choice:
                 initial = default_choice.value
 
             if self.type == CustomFieldTypeChoices.TYPE_SELECT:
+                if not required or default_choice is None:
+                    choices = add_blank_choice(choices)
                 field_class = CSVChoiceField if for_csv_import else forms.ChoiceField
```
glennmatthews, line 568: This change feels wrong to me - the form field (as opposed to the filter form field) for both Select and MultiSelect custom fields should have a blank choice, shouldn't it? Can you help me understand why this should be specific to Select fields only?

nautobot/utilities/utils.py
```diff
@@ -1013,9 +1013,11 @@ def convert_querydict_to_factory_formset_acceptable_querydict(request_querydict,
 
 
 def is_single_choice_field(filterset, field_name):
+    from nautobot.extras.filters import CustomFieldMultiSelectFilter  # Avoid circular imports
+
     # Some filter parameters do not accept multiple values, e.g DateTime, Boolean, Int fields and the q field, etc.
     field = get_filterset_field(filterset, field_name)
-    return not isinstance(field, django_filters.MultipleChoiceFilter)
+    return not isinstance(field, (django_filters.MultipleChoiceFilter, CustomFieldMultiSelectFilter))
```
glennmatthews, line 1020: Should we instead change it so that `CustomFieldMultiSelectFilter` is a `MultipleChoiceFilter` subclass?

glennmatthews (commented):
Also, this seems to only address the MultiSelect symptom reported in #3391. What about fixing the Select symptom as well?

> on custom field of the type selection
> Only one entry can be selected.

nautobot/extras/models/customfields.py
timizuoebideri1: By default multi select fields have blanks

timizuoebideri1: Adding it makes blank a choice in the multi select, which was causing error

nautobot/extras/filters/customfields.py
```diff
@@ -70,7 +70,7 @@ class CustomFieldJSONFilter(CustomFieldFilterMixin, django_filters.Filter):
     """Custom field single value filter for backwards compatibility"""
 
 
-class CustomFieldMultiSelectFilter(CustomFieldFilterMixin, django_filters.Filter):
+class CustomFieldMultiSelectFilter(CustomFieldFilterMixin, django_filters.MultipleChoiceFilter):
     """Custom field single value filter for backwards compatibility"""
```
glennmatthews, line 74: Is the docstring here no longer accurate/relevant? Do we in fact need this to remain a non-multiple-choice filter for some compatibility reason?

nautobot/extras/filters/customfields.py
```diff
@@ -47,7 +47,7 @@ def filter(self, qs, value):
         # Return custom fields that don't match the value and null custom fields
         if self.exclude:
             qs_null_custom_fields = qs.filter(**{f"{self.field_name}__isnull": True}).distinct()
-            return super().filter(qs, value) | qs_null_custom_fields
+            return super().filter(qs, value).distinct() | qs_null_custom_fields
```
timizuoebideri1, line 50: Had to make this distinct because without it, it caused this error.
Screenshot 2023-04-13 at 5 49 17 PM

nautobot/extras/filters/customfields.py
```diff
@@ -70,7 +70,15 @@ class CustomFieldJSONFilter(CustomFieldFilterMixin, django_filters.Filter):
     """Custom field single value filter for backwards compatibility"""
 
 
-class CustomFieldMultiSelectFilter(CustomFieldFilterMixin, django_filters.Filter):
+class CustomFieldSelectFilter(CustomFieldFilterMixin, django_filters.Filter):
```
glennmatthews, line 73: Do we actually need these two filter classes, or should we just use a `CustomFieldMultiValueCharFilter` for select/multiselect custom fields?

nautobot/extras/filters/customfields.py
```diff
@@ -70,7 +70,15 @@ class CustomFieldJSONFilter(CustomFieldFilterMixin, django_filters.Filter):
...
+    """Custom field single value filter for backwards compatibility"""
```
glennmatthews, line 74: Docstring is inaccurate, isn't it?

nautobot/extras/filters/mixins.py
```diff
@@ -48,6 +50,7 @@ def __init__(self, *args, **kwargs):
             CustomFieldTypeChoices.TYPE_INTEGER: CustomFieldNumberFilter,
             CustomFieldTypeChoices.TYPE_JSON: CustomFieldJSONFilter,
             CustomFieldTypeChoices.TYPE_MULTISELECT: CustomFieldMultiSelectFilter,
+            CustomFieldTypeChoices.TYPE_SELECT: CustomFieldSelectFilter,
```
glennmatthews, line 53: Should this be CustomFieldMultiSelectFilter? Not clear to me why we're using CustomFieldSelectFilter here but using CustomFieldMultiSelectFilter at line 100 below.

nautobot/extras/tests/test_customfields.py
```diff
@@ -376,6 +378,34 @@ def test_regex_validation(self):
...
+                name="select_field",
+            )
+            custom_field_select.save()
+            custom_field_select.content_types.set([ct])
+            CustomFieldChoice.objects.create(field=custom_field_select, value="Foo")
+            CustomFieldChoice.objects.create(field=custom_field_select, value="Bar")
+            CustomFieldChoice.objects.create(field=custom_field_select, value="Baz")
+            filter_field = custom_field_select.to_filter_field()
+            self.assertIsInstance(filter_field, ChoiceField)
+            self.assertIsInstance(filter_field.widget, StaticSelect2)
```
glennmatthews, line 396: Should we assert that the blank choice isn't in the filter_field?

nautobot/extras/filters/customfields.py
timizuoebideri1: Because of the `lookup_expr`, Yes we would need both, `Select` makes use of the `in` lookup expr and `MultiSelect` `contains`

nautobot/extras/filters/customfields.py
timizuoebideri1: 🫠, Mostlikey

nautobot/extras/filters/mixins.py
timizuoebideri1: Each class has a different `lookup expr`

nautobot/extras/forms/mixins.py
```diff
@@ -79,7 +79,7 @@ def __init__(self, *args, **kwargs):
                     set_initial=False, enforce_required=False, simple_json_filter=True
                 )
             else:
-                self.fields[field_name] = cf.to_form_field(set_initial=False, enforce_required=False)
+                self.fields[field_name] = cf.to_filter_field(set_initial=False, enforce_required=False)
```
jathanism, line 82: Why using `to_filter_field` inside of a form mixin? This seems a little confusing.

nautobot/extras/models/customfields.py
```diff
@@ -589,6 +585,16 @@ def to_form_field(
 
         return field
 
+    def to_filter_field(self, *args, **kwargs):
```
jathanism, line 588: This method is named `to_filter_field` and says that it's returning a filter field, but the code appears to actually be returning a form field. Can we please clarify this?

timizuoebideri1: Yes, A filter field is a form field used for filtering.
Maybe it should be called `to_filter_form_field`?

nautobot/extras/filters/mixins.py
```diff
@@ -92,7 +97,8 @@ def _generate_custom_field_lookup_expression_filters(cls, filter_name, custom_fi
         custom_field_type_to_filter_map = {
             CustomFieldTypeChoices.TYPE_DATE: CustomFieldMultiValueDateFilter,
             CustomFieldTypeChoices.TYPE_INTEGER: CustomFieldMultiValueNumberFilter,
-            CustomFieldTypeChoices.TYPE_SELECT: CustomFieldMultiValueCharFilter,
+            CustomFieldTypeChoices.TYPE_SELECT: CustomFieldMultiSelectFilter,
```
glennmatthews, line 100: Should this be `CustomFieldSelectFilter`? Or was the existing use of `CustomFieldMultiValueCharFilter` incorrect for some reason?

nautobot/extras/filters/customfields.py
glennmatthews: Should these be subclasses of `CustomFieldMultiValueCharFilter`?

nautobot/extras/models/customfields.py
glennmatthews: `to_filter_form_field` would be clearer, yes.

nautobot/extras/tests/test_customfields.py
```diff
@@ -1813,16 +1790,6 @@ def test_filter_multi_select(self):
             self.queryset.filter(_custom_field_data__cf9__contains="Bar"),
         )
 
-    def test_filter_null_values(self):
```
timizuoebideri1, line 1816: if `null` filtering is desired, The only way i could get it to show in the UI is by using this `[("null", "None")]` instead of `[(None, "None")]

nautobot/extras/models/customfields.py
```diff
@@ -589,6 +585,16 @@ def to_form_field(
 
         return field
 
+    def to_filter_form_field(self, *args, **kwargs):
+        """Return a filter form field suitable for filtering a CustomField's value for an object."""
+        form_field = self.to_form_field(*args, **kwargs)
+        if self.type == CustomFieldTypeChoices.TYPE_SELECT:
+            # Remove the blank choice, as StaticSelect2Multiple do not need a blank choice
+            # Its blank by default
+            choices = form_field.choices[1:]
```
bryanculver, line 594: Looks like this: https://github.com/nautobot/nautobot/actions/runs/4729451600/jobs/8392681922?pr=3574#step:7:161 is breaking a test case. In some instances, the first choice is not blank.

bryanculver (changes requested):

nautobot/extras/models/customfields.py
jathanism: I think they are only blank if the `add_blank_choice()`  is included in the choices call.

timizuoebideri1: The reason the test failed, is because CustomField Select type is no longer a normal select field, making the normal way of getting its value incorrect, but this has been fixed

glennmatthews: Sure seems like this should have the equivalent logic to the `required or default_choice is None` check that exists above?

nautobot/core/tests/integration/test_filters.py
```diff
@@ -152,7 +152,11 @@ def test_input_field_gets_updated(self):
         self.browser.find_by_xpath("//a[@href='#advanced-filter']").click()
         self.assertEqual(self.browser.find_by_name(text_field_name)[2].value, "test new")
         self.assertEqual(self.browser.find_by_name(integer_field_name)[2].value, "1111")
-        self.assertEqual(self.browser.find_by_name(select_field_name)[2].value, "Option B")
+        # CustomSelect Field is a MultiValueCharField, and the only way to get its values is using this approach.
+        # Its values are the  options available on the select field
+        custom_select_field = self.browser.find_by_name(select_field_name)[2].find_by_tag("option")
```
glennmatthews, line 157: Seems like this test should probably be enhanced to also create and test the UI for a multiselect custom field as well?

nautobot/extras/models/customfields.py
```diff
@@ -589,6 +585,16 @@ def to_form_field(
...
+            form_field.widget = StaticSelect2Multiple(choices=choices)
```
glennmatthews, line 595: This feels wrong to me, actually - for TYPE_SELECT we're using a MultiValueCharFilter and supporting all of the FILTER_CHAR_BASED_LOOKUP_MAP filter variants (`__isw`, `__re`, etc.) - so from a form perspective shouldn't this be a MultiValueCharField ratehr than a StaticSelect?

nautobot/extras/tests/test_customfields.py
```diff
@@ -1740,68 +1771,14 @@ def test_filter_url(self):
...
-            self.queryset.filter(_custom_field_data__cf8="Foo"),
+            self.filterset({"cf_cf8": ["Foo", "AR"]}, self.queryset).qs,
+            self.queryset.filter(_custom_field_data__cf8__in=["Foo", "AR"]),
         )
         self.assertQuerysetEqual(
             self.filterset({"cf_cf8__n": ["Foo"]}, self.queryset).qs,
             self.queryset.exclude(_custom_field_data__cf8="Foo")
             | self.queryset.filter(_custom_field_data__cf8__isnull=True),
         )
-        self.assertQuerysetEqual(
```
glennmatthews, line 1751: Did we lose existing filter capabilities? That seems less than desirable.

nautobot/extras/models/customfields.py
jathanism: Attempting to assert your concern here: `MultiValueCharField` is does not inherit from `ChoiceField` and therefore cannot accept choices.

nautobot/extras/tests/test_customfields.py
jathanism: We can't have it both ways, can we? If we allow `TYPE_SELECT` to be a multiple input filter from choices, then allowing `__isw` in the "Advanced" UI filter form makes no sense because the Value column still wants you to select from the provided choices.

However, to your concern, this still facilitates these filters being usable in the REST API or manually, just not intuitively in the dynamic filter form interface. I would say that for the v1 UI this is fine and we can attempt to fix this in the v2 UI while still allowing the filter to behave as-is.

glennmatthews: I'm assuming by the removal of these tests that this filtering (at the filterset level, not the UI level) is no longer working. Happy to be proven wrong.

nautobot/extras/filters/customfields.py
```diff
@@ -71,14 +72,23 @@ class CustomFieldJSONFilter(CustomFieldFilterMixin, django_filters.Filter):
...
     def __init__(self, *args, **kwargs):
         kwargs.setdefault("lookup_expr", "contains")
         super().__init__(*args, **kwargs)
 
 
+class CustomFieldMultiValueSelectFilter(CustomFieldFilterMixin, django_filters.MultipleChoiceFilter):
+    """This provides functionality for filtering custom fields with select type"""
+
+    def __init__(self, *args, **kwargs):
+        # TODO(jathan): See if I can make this user `kwargs` instead.
```
jathanism, line 87: I think we can ditch this comment. As far as I can tell there is no way to overload the widget more directly except on a FilterSet class.

jathanism (approved):
LGTM. :shipit:

bryanculver (approved):

Issue #765, opened 2021-08-03
Epic: Jobs Overhaul


jedelman8:
## Purpose

Jobs have grown significantly since they were introduced but now we're ready to help it achieve the next level of use cases in both scale and complexity.

## Plan

We look to address several key points with overhauling Jobs:
- Improve state management (currently we must keep tabs on Celery through `JobResult` which is error prone)
    - Lets hope to address things like: #633 
- Simplify APIs (better natural keys, simpler job scheduling, less "magic" methods)

## TODO

```[tasklist]
#### Data Model
- [x] #2611
- [x] https://github.com/nautobot/nautobot/issues/1622
- [x] https://github.com/nautobot/nautobot/issues/3856
```

```[tasklist]
#### Mechanics
- [x] #1622
- [x] https://github.com/nautobot/nautobot/issues/3453
- [x] #2979
- [x] https://github.com/nautobot/nautobot/issues/3692
```

```[tasklist]
#### Logging
- [x] #2331
```

```[tasklist]
### Bugs/Must Fix for 2.0
- [x] Fix `profile` feature (requires celery_kwargs field in job result)
- [x] Ensure scheduled jobs set the `user` when executing the job (Add `user` field to `ScheduledJob` or just leverage `celery_kwargs["user"]`?)
- [x] Update Job Hooks, Job Buttons to account for new enqueueing methods
```

```[tasklist]
### Housekeeping
- [x] https://github.com/nautobot/nautobot/issues/3713
- [ ] #2853
```

```[tasklist]
### Documentation
- [x] #3940
- [x] How to implement dry-run support in a new or existing job (including but not limited to `transaction.atomic()`)
```

### Elsewhere
- [ ] Client code and client documentation (e.g. `pynautobot` and `nautobot-ansible`) should be shored up for v2 changes
- [x] #4095
- [x] See about eliminating `scheduled_job_handler`, altho since that’s internal it’s less scary and may be able to live and could just be used as a dispatcher of a Job's Celery task

### Related
- #2980
- #2726
   - Done via not migrating it in #1622 (#3228)

### Follow-on
- [ ] https://github.com/nautobot/nautobot/issues/3440
- [ ] https://github.com/nautobot/nautobot/issues/3705
- [ ] #4238
- [ ] #4239
- [ ] #4240
- [x] https://github.com/nautobot/nautobot/issues/3928

---
[...]

PR #3593, opened 2023-04-14
Re-architect job dryrun

gsnider2195:
# Closes: #3453 
# What's Changed

- Created `DryRunVar` job variable type to support replacing the `commit` flag with an optional `dryrun` argument on the job
  - The name has been changed to reflect the change in behavior. Previously the `commit` flag would be enforced by Nautobot and only roll back database changes made by the job but the job author would be responsible for making sure that no changes are made to the rest of the environment (device changes or other external systems). With `dryrun` the job author is responsible for determining what a `dryrun` means for their job. Nautobot no longer automatically rolls back database transactions and this argument is only used for bypassing the approval of a job where `approval_required=True`. If a job author would like to implement a dryrun mode without allowing the job to bypass approval, they can declare the dryrun property as a `BooleanVar` instead of a `DryRunVar`.
- Added a `BaseJob.supports_dryrun` property that returns true if the there is a `DryRunVar` named `dryrun` defined on the job class, also cached on the `Job` model for retrieval in list views
- Renamed `Job.commit_default` to `dryrun_default` and `Job.commit_default_override` to `dryrun_default_override`
  - Changed the default from `True` to `False` and created a data migration to invert the value of `dryrun_default` in the database
- Updated `Job.read_only` to set `dryrun=True` instead of setting `commit=False` when running the job
- Updated job approval logic to bypass approval if job supports dryrun and dryrun is set to `True`
  - Now works for both immediately executed jobs and scheduled jobs
- Updated job run view to change the run button from `Run Job Now (DRYRUN)` to `Request to Run Job Now` if approval is required, depending on the state of the `dryrun` argument

# Follow up work (tracking in #765)

- Censor the task args and kwargs in the JobResult manager `store_result` method
- Add `celery_kwargs` field to `JobResult` and `ScheduledJob`
  - Fix `profile` feature (requires celery_kwargs field in job result)
- Add `user` field to `ScheduledJob`?

# TODO

- [x] Unit, Integration Tests
- [x] Add a `supports_dryrun` database field to cache the job class `supports_dryrun` property


nautobot/extras/api/views.py
```diff
@@ -435,169 +426,23 @@ def _create_schedule(serializer, data, job_model, request, celery_kwargs=dict, t
...
+        user=user,
+        approval_required=approval_required,
         crontab=crontab,
         queue=task_queue,
     )
     scheduled_job.save()
     return scheduled_job
 
 
-def _run_job(request, job_model):
```
gsnider2195, line 455: This doesn't need to be a separate method anymore since the legacy job view was removed. Moved this over to the `JobViewSet.run` method

nautobot/extras/jobs.py
```diff
@@ -1195,27 +1242,6 @@ def get_job(class_path):
     return jobs.get(grouping_name, {}).get(module_name, {}).get("jobs", {}).get(class_name, None)
 
 
-# TODO: this probably goes away
```
gsnider2195, line 1198: No longer required since `ScheduledJob` stores the actual task name now

nautobot/extras/templates/extras/job_approval_request.html
```diff
@@ -102,13 +102,15 @@ <h1>{{ object.name }}</h1>
...
                         </div>
                         <div class="pull-right">
-                            <button type="submit" name="_dry_run" id="id__dry_run" class="btn btn-primary"
-                                {% if not perms.extras.run_job or object.job_model is None or not object.job_model.runnable %}
-                                    disabled="disabled"
-                                {% endif %}
-                            >
-                                <i class="mdi mdi-play"></i> Dry Run
-                            </button>
+                            {% if "dryrun" in job_form.fields and object.job_model.job_class.supports_dryrun %}
```
gsnider2195, line 105: TODO: should use `supports_dryrun`

nautobot/extras/jobs.py
```diff
@@ -460,12 +465,29 @@ def as_form(self, data=None, files=None, initial=None, approval_view=False):
...
         form = self.as_form_class()(data, files, initial=initial)
 
-        job_model = JobModel.objects.get_for_class_path(self.class_path)
-        task_queues = job_model.task_queues if job_model.task_queues_override else self.task_queues
+        try:
+            job_model = JobModel.objects.get_for_class_path(self.class_path)
+            read_only = job_model.read_only if job_model.read_only_override else self.read_only
+            dryrun_default = job_model.dryrun_default if job_model.dryrun_default_override else self.dryrun_default
+            task_queues = job_model.task_queues if job_model.task_queues_override else self.task_queues
+        except JobModel.DoesNotExist:
```
gsnider2195, line 473: This had a todo note to remove in 2.0 but it still generates a 500 error when you try to navigate to a scheduled job for a job model that has been deleted. Should this just raise a better error message instead of falling back to the job class?

examples/example_plugin/example_plugin/jobs.py
```diff
@@ -1,16 +1,42 @@
...
 from nautobot.extras.choices import ObjectChangeActionChoices
-from nautobot.extras.jobs import IntegerVar, Job, JobHookReceiver, JobButtonReceiver
+from nautobot.extras.jobs import DryRunVar, IntegerVar, Job, JobHookReceiver, JobButtonReceiver
 
 
 name = "ExamplePlugin jobs"
 
 
+class ExampleDryRunJob(Job):
+    # TODO(gary): assert that only one DryRunVar is defined on Job classes
```
gsnider2195, line 16:
```suggestion
```

Tracking in #3087 now

gsnider2195:
Only one unittest is failing as expected, to be resolved as follow-up PR

`ERROR: test_job_profiling (nautobot.extras.tests.test_jobs.JobTest)`

bryanculver:
> Only one unittest is failing as expected, to be resolved as follow-up PR
> 
> `ERROR: test_job_profiling (nautobot.extras.tests.test_jobs.JobTest)`

Proper fix for that is

bryanculver:
Sorry! Fat-fingered a close of the PR. Re-opened.

nautobot/extras/jobs.py
```diff
@@ -410,6 +411,10 @@ def time_limit(cls):  # pylint: disable=no-self-argument
     def has_sensitive_variables(cls):  # pylint: disable=no-self-argument
         return getattr(cls.Meta, "has_sensitive_variables", True)
 
+    @classproperty
+    def supports_dryrun(cls):  # pylint: disable=no-self-argument
+        return isinstance(getattr(cls, "dryrun", None), DryRunVar)
```
bryanculver, line 416: We should probably throw a warning on import if `dryrun` is not an instance of `DryRunVar`

nautobot/extras/jobs.py
bryanculver: Should we set the FK to delete cascade?

bryanculver (approved):

nautobot/extras/jobs.py
gsnider2195: Tracking in #3087

nautobot/extras/views.py
```diff
@@ -1090,6 +1090,7 @@ def get(self, request, class_path=None, slug=None):
                     # for example "?kwargs_from_job_result=<UUID>&integervar=22&_commit=False"
                     explicit_initial = initial
                     initial = job_result.task_kwargs.copy()
+                    # TODO(gary): task_queue is no longer stored on the task kwargs, may need to add it as a field on JobResult?
```
gsnider2195, line 1093:
```suggestion
```

nautobot/extras/models/jobs.py
```diff
@@ -161,18 +161,16 @@ class Job(PrimaryModel):
...
         db_index=True,
         help_text="Whether the job defaults to not being shown in the UI",
     )
     # Job.Meta.field_order is not overridable in this model
-    read_only = models.BooleanField(
-        default=False, help_text="Whether the job is prevented from making lasting changes to the database"
+    dryrun_default = models.BooleanField(
+        default=False, help_text="Whether the job defaults to running with dryrun argument set to true"
     )
+    read_only = models.BooleanField(default=False, help_text="Whether the job is only allowed to run with dryrun set")
```
gsnider2195, line 173:
```suggestion
    read_only = models.BooleanField(default=False, help_text="Whether the job is only permitted to run with dryrun set")
```

nautobot/extras/api/serializers.py
```diff
@@ -737,8 +737,8 @@ class Meta:
             "has_sensitive_variables_override",
             "approval_required",
             "approval_required_override",
-            "commit_default",
-            "commit_default_override",
+            "dryrun_default",
+            "dryrun_default_override",
             "hidden",
```
gsnider2195, line 742: TODO: add `supports_dryrun` as read only

nautobot/extras/models/jobs.py
```diff
@@ -191,6 +189,10 @@ class Job(PrimaryModel):
         blank=True,
         help_text="Comma separated list of task queues that this job can run on. A blank list will use the default queue",
     )
+    supports_dryrun = models.BooleanField(
```
gsnider2195, line 192: TODO: set `editable=False`

nautobot/extras/views.py
```diff
@@ -1149,13 +1149,20 @@ def post(self, request, class_path=None, slug=None):
                 "This job is flagged as possibly having sensitive variables but is also flagged as requiring approval."
                 "One of these two flags must be removed before this job can be scheduled or run.",
             )
+        elif job_model.supports_dryrun and job_model.read_only and request.POST.get("dryrun", False) is not True:
```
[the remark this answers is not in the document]
gsnider2195: After discussion, `read only` should be something that the job author declares for their job but nautobot doesn't make logic based on the flag (yet)

nautobot/extras/models/jobs.py
gsnider2195: Also set this on `read_only`

PR #3594, opened 2023-04-14
Fix #3592: Better duplicate call detection for populate_model_features_registry()

### LINKED ISSUE #3592: 2.0: Delaring Any Model with `@extras_features("relationships")` Empties Those from #826
### Environment
* Nautobot version (Docker tag too if applicable): `next` or `v2.0.0-alpha.3`
* Python version: 3.X
* Database platform, version: Any
* Middleware(s): N/A

Same applies for Custom Fields.

### Steps to Reproduce
1. Clone the repo
2. Add `@extras_features("relationships")` on any model inheriting from `PrimaryModel`
3. Run Tests/Edit or Create a Relationship

### Expected Behavior
Have access to all models that inherit from `PrimaryModel` to associate a Relationship to.

### Observed Behavior
Only have the decorated model(s).

bryanculver:
# Closes: #3592
# What's Changed
Uses the registry to know if `populate_model_features_registry` has already been called. Provide a way to refresh it if needed (see: https://github.com/nautobot/nautobot/pull/3360#discussion_r1131373797).

# TODO

- [x] Unit, Integration Tests
- [X] Example Plugin Updates (when adding/changing features)

### DIFF
--- a/nautobot/extras/utils.py
+++ b/nautobot/extras/utils.py
@@ -136,11 +136,8 @@ class FeatureQuery:
         # initialization of the application, before `ExtrasConfig.ready` is called.
         # Calling `populate_model_features_registry` in `ExtrasConfig.ready` would lead to an outdated `model_features`
         # `registry` record being used by `FeatureQuery`.
-        # As the `populate_model_features_registry` can be resource-intensive. The check the conditional check is used to
-        # avoid calling the function multiple times and optimize the performance of the application.
-        # TODO(timizuo): Provide a better solution for this; check https://github.com/nautobot/nautobot/pull/3360/files#r1131373797 comment.
-        if not registry["model_features"].get("relationships"):
-            populate_model_features_registry()
+
+        populate_model_features_registry()
         query = Q()
         for app_label, models in registry["model_features"][self.feature].items():
             query |= Q(app_label=app_label, model__in=models)
@@ -223,7 +220,7 @@ def extras_features(*features):
     return wrapper
 
 
-def populate_model_features_registry():
+def populate_model_features_registry(refresh=False):
     """
     Populate the registry model features with new apps.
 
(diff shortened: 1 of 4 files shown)


nautobot/extras/utils.py
```diff
@@ -267,6 +267,9 @@ def populate_model_features_registry():
         feature_name = lookup_conf["feature_name"]
         registry["model_features"][feature_name] = registry_items
 
+    if registry["model_features"].get("populate_model_features_registry_called", False):
+        registry["populate_model_features_registry_called"] = True
```
glennmatthews, line 271: Should this be?

```suggestion
        registry["model_features"]"populate_model_features_registry_called"] = True
```

?

Alternately, should we make this a top-level registry key instead of storing it under the `model_features` key?

bryanculver: I didn't want to clash with this: https://github.com/nautobot/nautobot/blob/next/nautobot/extras/utils.py#L213-L214

Nor is `populate_model_features_registry_called` a model feature. I just wanted someplace to store globally that "you, the runtime engine, have been here before".

glennmatthews: My point was that you're **looking for** `"populate_model_features_registry_called"` inside of `registry["model_features"]`, but then **setting** `"populate_model_features_registry_called"` at the top level of `registry` - one way or the other, it seems wrong?

bryanculver: Doh!

bryanculver: Fixed. I meant for `registry.get("populate_model_features_registry_called", False)`

nautobot/extras/utils.py
```diff
@@ -267,6 +267,9 @@ def populate_model_features_registry():
         feature_name = lookup_conf["feature_name"]
         registry["model_features"][feature_name] = registry_items
 
+    if registry.get("populate_model_features_registry_called", False):
+        registry["populate_model_features_registry_called"] = True
```
glennmatthews, line 271: I don't think the `if` is really needed here:

```suggestion
    registry["populate_model_features_registry_called"] = True
```

glennmatthews (approved):

nautobot/extras/utils.py
bryanculver: https://github.com/nautobot/nautobot/blob/9fe1ffdc6d23bbe86d8d2e9c3ae087233b15e8fc/nautobot/extras/registry.py#L16-L19

Without it I was getting bit by this.

timizuoebideri1 (approved):

PR #3595, opened 2023-04-14
Update the warning provided when a bad reverse entry is not found in …

itdependsnetworks:
…serializer to point to correct import location.

# Closes: DNE
# What's Changed

Update the warning provided when a bad reverse entry is not found in serializer to point to correct import location.

# TODO

### DIFF
--- a/nautobot/extras/api/serializers.py
+++ b/nautobot/extras/api/serializers.py
@@ -148,7 +148,7 @@ class NotesSerializerMixin(BaseModelSerializer):
                     f"Notes feature is not available for model {model_name}. "
                     "Please make sure to: "
                     f"1. Include NotesMixin from nautobot.extras.model.mixins in the {model_name} class definition "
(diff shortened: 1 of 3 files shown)


nautobot/extras/api/serializers.py
```diff
@@ -148,7 +148,7 @@ def get_notes_url(self, instance):
                     f"Notes feature is not available for model {model_name}. "
                     "Please make sure to: "
                     f"1. Include NotesMixin from nautobot.extras.model.mixins in the {model_name} class definition "
-                    f"2. Include NotesViewSetMixin from nautobot.extras.api.mixins in the {model_name}ViewSet "
+                    f"2. Include NotesViewSetMixin from nautobot.extras.api.views in the {model_name}ViewSet "
```
glennmatthews, line 151: To be fair, it *should* be in a `.mixins` module. :-)

glennmatthews (approved):

glennmatthews:
Looks like `nautobot.extras.tests.test_plugins.PluginAPITest` needs to be updated to check for the corrected verbiage.

Issue #3518, opened 2023-03-30
Failing migration to v2.0.0-alpha.2


mzbroch:
### Environment
* Nautobot version: v2.0.0-alpha.2
* Python version: 3.7
* Database platform, version: Postgres
* Middleware(s):

### Steps to Reproduce
1. Import existing 1.5.14 dataset into Postgres
2. Run `invoke debug`
3.

### Expected Behavior

Migrations work

### Observed Behavior

Migrations fail with following output:
```
nautobot_1       |   Applying extras.0060_role_and_alter_status... OK
db_1             | 2023-03-30 09:48:38.151 UTC [129] ERROR:  duplicate key value violates unique constraint "extras_role_slug_key"
db_1             | 2023-03-30 09:48:38.151 UTC [129] DETAIL:  Key (slug)=(loopback) already exists.
db_1             | 2023-03-30 09:48:38.151 UTC [129] STATEMENT:  INSERT INTO "extras_role" ("id", "created", "last_updated", "_custom_field_data", "name", "slug", "color", "description", "weight") VALUES ('5f4d5a5a-0cb9-44cc-bad8-b4a9c9c95863'::uuid, '2023-03-30'::date, '2023-03-30T09:48:38.150962+00:00'::timestamptz, '{}', 'Loopback', 'loopback', '9e9e9e', '', NULL)
nautobot_1       |   Applying extras.0061_collect_roles_from_related_apps_roles...Traceback (most recent call last):
nautobot_1       |   File "/usr/local/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
nautobot_1       |     return self.cursor.execute(sql, params)
nautobot_1       |   File "/usr/local/lib/python3.7/site-packages/django_prometheus/db/common.py", line 71, in execute
nautobot_1       |     return super().execute(*args, **kwargs)
nautobot_1       | psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "extras_role_slug_key"
nautobot_1       | DETAIL:  Key (slug)=(loopback) already exists.
nautobot_1       |
nautobot_1       |
nautobot_1       | The above exception was the direct cause of the following exception:
nautobot_1       |
nautobot_1       | Traceback (most recent call last):
nautobot_1       |   File "/usr/local/bin/nautobot-server", line 6, in <module>
nautobot_1       |     sys.exit(main())
nautobot_1       |   File "/source/nautobot/core/cli.py", line 61, in main
[...]

PR #3596, opened 2023-04-14
Improve Role migrations

glennmatthews:
# Closes: #3518 
# What's Changed

- Rewrite a couple of 2.0 `Role` migrations so that we don't unnecessarily create a `Role.slug` field then turn around and delete it in a later migration. (This eliminated a case we otherwise had to handle on the intermediate extras.Role data migrations where we might try to create records with the same name but differing slugs or vice versa)
- Total rewrite of the `migrate_role_data` function in ways that (I hope!) improve clarity as well as (I think!) efficiency of the migration. Probably would be good to test at scale to be sure.
- Rework the `bulk_create_roles` function in extras.0061 migration to correctly use `Role.objects.get_or_create` so that we don't throw an error (#3518) in the case where we have duplicate Roles across several of the source models.
- Fix a bug I noticed in `config_context.html` while manually testing this.

# Manual tests

- With ipam.Role and DeviceRole and RackRole objects with the same names, migrate forward from develop thru extras.0061 migration and verify #3518 isn't encountered any more.
- Migrate forward and backward between dcim.0023 and dcim.0027
- Migrate forward and backward between extras.0061 and extras.0065
- Migrate forward and backward between ipam.0009 and ipam.0013
- Migrate forward and backward between virtualization.0011 and virtualization.0014

# TODO
- n/a Unit, Integration Tests

### DIFF
--- a/nautobot/extras/migrations/0060_role_and_alter_status.py
+++ b/nautobot/extras/migrations/0060_role_and_alter_status.py
@@ -37,12 +37,6 @@ class Migration(migrations.Migration):
                     models.JSONField(blank=True, default=dict, encoder=django.core.serializers.json.DjangoJSONEncoder),
                 ),
                 ("name", models.CharField(max_length=100, unique=True)),
-                (
-                    "slug",
-                    nautobot.core.models.fields.AutoSlugField(
-                        blank=True, max_length=100, populate_from="name", unique=True
-                    ),
-                ),
                 ("color", nautobot.core.models.fields.ColorField(default="9e9e9e", max_length=6)),
                 ("description", models.CharField(blank=True, max_length=200)),
                 ("weight", models.PositiveSmallIntegerField(blank=True, null=True)),
--- a/nautobot/extras/utils.py
+++ b/nautobot/extras/utils.py
@@ -524,80 +524,94 @@ def check_if_key_is_graphql_safe(model_name, key):
 
 def migrate_role_data(
     model_to_migrate,
-    legacy_role_field_name="legacy_role",
-    legacy_role_model=None,
-    legacy_role_choiceset=None,
-    new_role_field_name="new_role",
-    new_role_model=None,
-    new_role_choiceset=None,
+    *,
+    from_role_field_name,
+    from_role_model=None,
+    from_role_choiceset=None,
+    to_role_field_name,
+    to_role_model=None,
+    to_role_choiceset=None,
     is_m2m_field=False,
 ):
     """
-    Update all `model_to_migrate` with a new value for `new_role_field_name` based on `legacy_role_field_name` value.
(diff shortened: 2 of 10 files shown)


bryanculver (approved):
Semantically I like `previous` and `next` instead of `legacy` and `new` for the migration but that's not something to hold this up on. Since `migrate_role_data` is also used for the reverse migration it's a bit of a misnomer to call the generic Role model `legacy` since it is cemented that is the newer one.

glennmatthews:
How about `from_` and `to_`? Happy to do a quick search-and-replace if it improves understandability.

bryanculver:
> How about `from_` and `to_`? Happy to do a quick search-and-replace if it improves understandability.

Even better.

nautobot/extras/migrations/0060_role_and_alter_status.py
```diff
@@ -37,12 +37,6 @@ class Migration(migrations.Migration):
                     models.JSONField(blank=True, default=dict, encoder=django.core.serializers.json.DjangoJSONEncoder),
                 ),
                 ("name", models.CharField(max_length=100, unique=True)),
-                (
-                    "slug",
-                    nautobot.core.models.fields.AutoSlugField(
-                        blank=True, max_length=100, populate_from="name", unique=True
-                    ),
-                ),
```
timizuoebideri1, line 45: Is it a good practice, to alter a migration file not created by this PR?

glennmatthews: Since this migration is specific to `next`, which hasn't shipped a non-alpha release yet, it's permissible. We generally wouldn't want to do this in `develop` to affect existing released migrations without extensive testing for compatibility.

nautobot/extras/utils.py
```diff
@@ -523,95 +523,81 @@ def check_if_key_is_graphql_safe(model_name, key):
...
+        is_m2m_field (bool): True if the role fields are both ManyToManyFields, else False
+    """
+    if legacy_role_model is not None:
+        assert legacy_role_choiceset is None
+        if new_role_model is not None:
+            assert new_role_choiceset is None
+            # Mapping legacy model instances to corresponding new model instances
+            legacy_to_new_roles = {
+                # Use .filter().first(), not .get() because new role might not exist, especially on reverse migrations
+                legacy_role: new_role_model.objects.filter(name=legacy_role.name).first()
```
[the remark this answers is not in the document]
timizuoebideri1: Or could we create a `safe_get` method that does the `try` and `except`:
`Model.objects.safe_get(name="sample")`

timizuoebideri1 (approved):
Most of the questions i have are out of curiosity 🫠

bryanculver (approved):
😎

Issue #3576, opened 2023-04-13
Provide HTML docstring-based documentation


Kircheneer:
### As ...

Austin - Network Automation Engineer

### I want ...

everything in the code base to be available as HTML documentation online. 

### So that ...

available functions and classes are more discoverable and can be indexed and found by/via search engines.

### I know this is done when...

https://docs.nautobot.com provides documentation for all the classes/functions via their respective docstrings.

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [X] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

None

### External Dependencies

- https://mkdocstrings.github.io/ could be a good option

itdependsnetworks:
trying to understand how this is different then: https://github.com/nautobot/nautobot/blob/develop/mkdocs.yml#L190-L203

As far as adding all code, I am not sure I am a fan just yet. In my mind, it will flood the search and I find the integrations with github to find where code is referenced to be better.

jathanism:
> trying to understand how this is different then: https://github.com/nautobot/nautobot/blob/develop/mkdocs.yml#L190-L203
> 
> As far as adding all code, I am not sure I am a fan just yet. In my mind, it will flood the search and I find the integrations with github to find where code is referenced to be better.

Having a full documentation of our developer API I think is really the point. But the underlying primitives will all be required to have pedantic-level docstrings for this to work. Check out the source code of Celery for an example of how a fully-marked up code project can be used to generate the API documentation.

Celery uses RST + Sphinx, but he premise remains the same.

For example, compare:

- Source for `Task`: https://github.com/celery/celery/blob/main/celery/app/task.py#L164
- Generated docs for `Task`: https://docs.celeryq.dev/en/stable/reference/celery.app.task.html

jathanism:
Also, I wanted to comment that we are already including `mkdocstrings` as a dependency for our documentation. So the door is open for us to start marking up all of our code.

We haven't formalized it, but have softly agreed to using Google style doc strings, and [`mkdocstrings` includes a parser for this](https://mkdocstrings.github.io/python/usage/#supported-docstrings-styles). A part of this body of work should also make a formal decision on our docstring style and commit to it.

Issue #1570, opened 2022-03-30
Epic: Developer Documentation Overhaul


lampwins:
## TODO

- [ ] #2981
     - Mostly based on the current `docs/plugins/development.md`, but removing the parts that are about setting up a development environment and such.
- [ ] #5217
     - How best to author models, etc.
     - Cross-reference: https://blog.networktocode.com/post/developing-nautobot-plugins-1/
- [ ] Model auto-docs
     - Probably get this for free with Django admin model docs
- [x] #2800 
- [x] #2982
    - Job Author
- [ ] #2983
    - Not Nautobot itself, but setting up your environment for success for Jobs or App

glennmatthews:
Taking inspiration from the blog series starting at https://blog.networktocode.com/post/developing-nautobot-plugins-1/ would be an excellent idea.

itdependsnetworks:
Please consider and/or at least be aware of the "classy" views. There is django classy https://ccbv.co.uk/ and drf-classy 
https://www.cdrf.co/. It helps provide the MRO of classes as well as make it easier to navigate the Class objects and show where the code originates from. The code used to generate classy views for django and drf could be used to do the same for Nautobot as well.


![image](https://user-images.githubusercontent.com/9260483/216630516-0bb8b437-ea30-488a-914b-2ba0e5fd128a.png)

bryanculver:
Proposing changes to top-level sections:

- Overview
- User Guide (How to use Nautobot)
- Administrator Guide (How to deploy, admin Nautobot)
- Developer Guide (Jobs, Apps, etc, _maybe_ shove Maintainers guide in here)
- Release Notes (fix Overview page)
- Nautobot Apps
    - Future work could show links to the docs for the apps you have installed, as well as listing out all of the other ones

bryanculver:
- Improve cross-linking between different sections of the documentation
    - Potentially investigate a plugin to help do this for us
- The target audience is mainly directed towards internal development, and very light on user guides, medium on feature guide
    - Feature Guide: "How to use `Namespaces`"
    - User Guide: "How to manage IPs in Nautobot"
        - Our current user guides are more so model/feature documentation
- Look at other excellent documentation as examples
    - Zero-to-hero

bryanculver:
# Overview

- Move app stack to administrator guide
   - Should also be accessible from the developer guide
- A very basic "try it out" for getting started then segway into the full install
   - Link to demo
   - Link to docker compose
   - link to nautobot lab
[...]

PR #1657, opened 2022-04-18
Vendoring drf-spectacular to work around package PyPI Issue

bryanculver:
# Closes: #DNE
# What's Changed

"Vendor"-ed def-spectacular into `/nautobot/third_party/` until a new package is published to get around blocking 1.3 and requiring fix for https://github.com/tfranzel/drf-spectacular/issues/688.

# TODO

- [x] Explanation of Change(s)
- [x] Attached Screenshots, Payload Example
- [x] Unit, Integration Tests
- [x] Documentation Updates (when adding/changing features)
- [x] Example Plugin Updates (when adding/changing features)
- [x] Outline Remaining Work, Constraints from Design

PR #3614, opened 2023-04-19
Add mkdocs python classy to document classes

itdependsnetworks:
# Closes: Partial #1570 and potentially #3576
# What's Changed

This is POC, it is only a POC, not intended to go in any form as shown. 

This is a POC around building "classy" docs, that show you the nested inheritance of Python Classes. I created a mkdocs plugin to build this out.

The library provides the following features

* The clear definition of the MRO (Method Resolution Order)
* The code in each method
    * The code even when over written, such as when you would like to see `super` being called
    * The code is provided in an ordered (highest is preferred) manner
* The code provided with line numbers from where the code was take from
* The definition of where the code is preferred
* The code is provided in a collapsible block
* Each attribute, and when not preferred shown via a ~~strike-through~~
* Link-able ancestors, when that class is also covered
* Link-able descendants, if applicable 
* Import statement convenience method

ui-view1
ui-view2
ui-view3

# TODO

- [ ] Unit, Integration Tests

There is a few challenges, most importantly is it currently requires you to start the application to run. While I have mostly worked around that when "serving the docs", I haven't when building. There are likely other issues, but again, wanted to show rather than tell.


bryanculver:
I'm curious to know if it will be possible to get this to work when it comes time to publish http://docs.nautobot.com: https://readthedocs.org/projects/nautobot/builds/20160099/

Kircheneer:
As to what I was aiming at with #3756, the docstring plays a major role there as it already contains valuable documentation a lot of the time, but is currently "hidden" in the source code

nautobot/extras/plugins/validators.py
```diff
@@ -40,6 +41,8 @@ def wrap_model_clean_methods():
     """
     Helper function that wraps plugin model validator registered clean methods for all applicable models
     """
+    if os.getenv("NAUTOBOT_NO_DB"):
+        return
     for model in ContentType.objects.filter(FeatureQuery("custom_validators").get_query()):
```
bryanculver, line 46: Might be unncesscesary with #3633

itdependsnetworks:
The workflow has been figured out, and is now building and rendering to read the docs: https://docs-landing-page--3614.org.readthedocs.build/projects/core/en/3614/development/classy/views.html

The things pending would be:
* Get agreement to move forward
* Get mkdocs-python-classy on pypi
* Fix anything left not passing in tests

.readthedocs.yaml
```diff
@@ -11,13 +11,17 @@ build:
   os: "ubuntu-20.04"
   tools:
     python: "3.9"
+  commands:
+    - pip install poetry
```
bryanculver, line 15: If this lets us ditch `requirements.txt` in factor of something like `poetry install -E docs`... 🔥

pyproject.toml
```diff
@@ -207,6 +207,9 @@ watchdog = "~2.1.9"
 [tool.poetry.scripts]
 nautobot-server = "nautobot.core.cli:main"
 
+[tool.poetry.group.dev.dependencies]
+mkdocs-python-classy = {git = "https://github.com/itdependsnetworks/mkdocs-python-classy.git", rev = "main"}
```
bryanculver, line 211: Depending on how Poetry handles this, the biggest blocker is this must be available in PyPI.

1.3 got held up due to Git imports. PyPI doesn't allow it: https://github.com/nautobot/nautobot/pull/1657, https://github.com/nautobot/nautobot/actions/runs/2185417492/jobs/3209161850 (logs gone but this was the failed release).

docker/Dockerfile
```diff
@@ -312,6 +310,9 @@ USER nautobot
 WORKDIR /opt/nautobot
 
 COPY --from=final-dev --chown=nautobot /opt/nautobot/nautobot_config.py /opt/nautobot/nautobot_config.py
+RUN mkdocs build --strict=False -q
```
glennmatthews, line 313: Why `--strict=False`?

docker/Dockerfile
```diff
@@ -312,6 +310,9 @@ USER nautobot
...
+# Copy in the built docs to bundle with the wheel
+COPY /source/nautobot/project-static/docs /source/nautobot/project-static/docs
```
glennmatthews, line 315: This doesn't seem right to me?
1) Same source and target paths?
2) Wheel has already been built at this point so this is too late to bundle in, in any case.

docker/Dockerfile
```diff
@@ -191,18 +191,16 @@ FROM dependencies-dev-python as build-docs
 # Copy docs dependencies to build the rendered docs
 COPY mkdocs.yml /source/
 COPY docs /source/docs
+COPY scripts /source/scripts
 
 # Build the rendered docs, this ensures that the docs are in the final image.
-RUN mkdocs build --no-directory-urls
```
glennmatthews, line 196: Is the entire `build-docs` layer irrelevant now?

docker/Dockerfile
```diff
@@ -191,18 +191,16 @@ FROM dependencies-dev-python as build-docs
...
 
 FROM dependencies as build-nautobot
 # Copy in the Nautobot source code to build the wheel from
 COPY nautobot /source/nautobot
 
-# Copy in the built docs to bundle with the wheel
-COPY --from=build-docs /source/nautobot/project-static/docs /source/nautobot/project-static/docs
 
 # Build the wheel
 RUN poetry build
```
glennmatthews, line 206: Do we not still need to render the docs before building the wheel?

mkdocs.yml
```diff
@@ -201,6 +217,11 @@ nav:
           - nautobot.apps.ui: "code-reference/nautobot/apps/ui.md"
           - nautobot.apps.urls: "code-reference/nautobot/apps/urls.md"
           - nautobot.apps.views: "code-reference/nautobot/apps/views.md"
```
glennmatthews, line 219: Does the addition of classy make the above `mkdocstrings` output irrelevant/redundant? It seems confusing to me to have two different sets of rendered docs for the same APIs.

glennmatthews (commented):
Interesting stuff!

docker/Dockerfile
itdependsnetworks: Need to fix, I had other issues and was trying to get around it.

docker/Dockerfile
itdependsnetworks: I don't know... I am not really sure what the layering buys us. I see the docs on top of the docs, but tbh don't understand it all.

mkdocs.yml
itdependsnetworks: I don't think so... The mkdocstrings is to get that code in those files, while this provides attributes and methods within MRO in mind.

bryanculver:
@jathanism didn't close this. It's because the target branch got deleted.

jathanism:
> @jathanism didn't close this. It's because the target branch got deleted.

👀

jathanism:
@itdependsnetworks Can you please update this to point to `develop`?

itdependsnetworks:
I cannot change target or re-open, created new one here: #4202

PR #3618, opened 2023-04-19
Fix corrupted package-lock.json

bryanculver:
# Closes: #3618
# What's Changed

In trying to create a reproduction of https://github.com/oklas/react-app-alias/issues/95 (the reason I created a semi-vendored `CracoAliasPlugin`) it turned out to be an issue with our `package-lock.json` which had somehow become corrupted due to the nature of how we built out `package.json` and our UI. Switching design systems, backends, etc. NPM believed packages were there that weren't. Since this is essentially cached resolution it's no surprise. I was unable to reproduce it by remaking a `create-react-app` project and adding back in our dependencies. Then used that as the seed for this `package-lock.json`.

Since we no longer needed to patch `CracoAliasPlugin` I removed our patch for it. Glad it worked but one less thing to maintain and one less demon to trip over in the future.

# TODO

- [x] Unit, Integration Tests


nautobot/ui/package.json
```diff
@@ -22,6 +22,7 @@
         "@testing-library/user-event": "^13.5.0",
         "axios": "^1.2.1",
         "framer-motion": "^9.0.3",
+        "nautobot-ui": "github:nautobot/nautobot-ui",
```
glennmatthews, line 25: Added inadvertently? We already have this at line 15 above.

bryanculver: So when I do `npm add https://github.com/nautobot/nautobot-ui.git` or `npm add nautobot/nautobot-ui` it adds this line instead of the one above. Maybe `"@nautobot/nautobot-ui": "nautobot/nautobot-ui",` has come backward capability handling?

bryanculver: Fixed @glennmatthews

glennmatthews (approved):

Issue #3612, opened 2023-04-19
Unable to filter Dynamic Groups


tanleach:
### Environment
* Nautobot version (Docker tag too if applicable): v1.5.16
* Python version: 3.10.11
* Database platform, version: Postgres 15.1
* Middleware(s): N/A

### Steps to Reproduce
1. <nautobot_url>/extras/dynamic-groups/?&content_type=ipam.prefix&q=

### Expected Behavior
Dynamic groups would show only entries where content_type is 'ipam.prefix' (not specific to `ipam.prefix`)

### Observed Behavior
Server Error:
``` 
There was a problem with your request. Please contact an administrator.

The complete exception is provided below:

<class 'KeyError'>

'dynamic groups'

Python version: 3.7.16
Nautobot version: 1.5.7
```
FWIW: also able to reproduce here: https://demo.nautobot.com/extras/dynamic-groups/?&content_type=dcim.device&q=

glennmatthews:
Thanks for the report! Looks like a bug in the [`get_filterset_parameter_form_field` function ](https://github.com/nautobot/nautobot/blob/develop/nautobot/utilities/utils.py#L909-L920)- it's looking for content-types corresponding to the feature `dynamic groups` but the feature is registered as `dynamic_groups`.

jathanism:
> Thanks for the report! Looks like a bug in the [`get_filterset_parameter_form_field` function ](https://github.com/nautobot/nautobot/blob/develop/nautobot/utilities/utils.py#L909-L920)- it's looking for content-types corresponding to the feature `dynamic groups` but the feature is registered as `dynamic_groups`.

I did a little triage and determined that this occurs with the `model._meta.verbose_name_plural` being passed through without being slugified. I suspect we would run into this issue on other objects but it's somewhat hidden from other models of `content_type` relationships because most of them have single-word names.

PR #3625, opened 2023-04-20
Fixed a 500 error when filtering by `content_type` in Dynamic Groups list view.

jathanism:
# Closes: #3612 
# What's Changed

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/utilities/utils.py
+++ b/nautobot/utilities/utils.py
@@ -916,7 +916,7 @@ def get_filterset_parameter_form_field(model, parameter, filterset=None):
             # `MultipleContentTypeField` employs `registry["model features"][feature]`, which may
             # result in an error if `feature` is not found in the `registry["model features"]` dict.
             # In this case use queryset
-            queryset_map = {"tags": TaggableClassesQuery, "job hooks": ChangeLoggedModelsQuery}
+            queryset_map = {"tags": TaggableClassesQuery, "job_hooks": ChangeLoggedModelsQuery}
             form_field = MultipleContentTypeField(
                 choices_as_strings=True, queryset=queryset_map[plural_name]().as_queryset()
             )
--- a/nautobot/utilities/utils.py
+++ b/nautobot/utilities/utils.py
@@ -907,7 +907,9 @@ def get_filterset_parameter_form_field(model, parameter, filterset=None):
 
         form_field = DynamicModelMultipleChoiceField(**form_attr)
     elif isinstance(field, ContentTypeMultipleChoiceFilter):
-        plural_name = model._meta.verbose_name_plural
+        # While there are other objects using `ContentTypeMultipleChoiceFilter`, the case where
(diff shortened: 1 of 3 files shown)


nautobot/utilities/utils.py
```diff
@@ -907,7 +907,9 @@ def get_filterset_parameter_form_field(model, parameter, filterset=None):
 
         form_field = DynamicModelMultipleChoiceField(**form_attr)
     elif isinstance(field, ContentTypeMultipleChoiceFilter):
-        plural_name = model._meta.verbose_name_plural
+        from nautobot.utilities.utils import slugify_dashes_to_underscores
```
glennmatthews, line 910: Why are we importing from `nautobot.utilities.utils` in `nautobot.utilities.utils`? :-)

nautobot/utilities/utils.py
```diff
@@ -907,7 +907,9 @@ def get_filterset_parameter_form_field(model, parameter, filterset=None):
...
+
+        plural_name = slugify_dashes_to_underscores(model._meta.verbose_name_plural)
```
glennmatthews, line 912: This is still a hack, tbh - we're *assuming* that the slugified model verbose_name_plural matches the feature name in the model features registry... but probably fine for now.

nautobot/utilities/utils.py
bryanculver: Because we need utils in our utils!

jathanism: I'm a genius. Don't question it.

nautobot/utilities/utils.py
jathanism: Yeah, I commented on the issue about that, and this appears to be the only case where a verbose_name_plural of a model also matches an extras feature, so it's actually specific to dynamic groups... apparently.

glennmatthews (approved):

nautobot/utilities/utils.py
```diff
@@ -907,7 +907,9 @@ def get_filterset_parameter_form_field(model, parameter, filterset=None):
 
         form_field = DynamicModelMultipleChoiceField(**form_attr)
     elif isinstance(field, ContentTypeMultipleChoiceFilter):
-        plural_name = model._meta.verbose_name_plural
+        # While there are other objects using `ContentTypeMultipleChoiceFilter`, the case where
+        # models that have sucha  filter and the `verbose_name_plural` has multiple words is ony one: "dynamic groups".
+        plural_name = slugify_dashes_to_underscores(model._meta.verbose_name_plural)
         try:
             form_field = MultipleContentTypeField(choices_as_strings=True, feature=plural_name)
         except KeyError:
```
glennmatthews, line 915: Looks to have b0rked the test - probably line 919 below needs to change to `"job_hooks"` instead of `"job hooks"` now.

jathanism: Yep. Thanks!

Issue #3529, opened 2023-04-03
UI 2.0: GenericView Base


bryanculver:
Provide a `GenericView` base template that can be built on top of (see #3528) that provides breadcrumbs, top-menu, and attributes for defining tabs, page headings, buttons, etc:

See:

<img width="1366" alt="image" src="https://user-images.githubusercontent.com/31187/229587078-f7a77daf-8e32-4002-b146-90d97ad8cad2.png">

bryanculver:
Implemented with #3626

PR #3626, opened 2023-04-20
UI 2.0 Implement GenericView component

norbert-mieczkowski-codilime:
# Closes: #3529
# What's Changed

- Add global `height: 100%; width 100%;` styles to `html`, `body` and `#root` elements.
- Change `Layout` component content and style to allow better scrolling experience.
- Implement `GenericView` component providing all the basic components needed for a generic layout.
- Use newly implemented `GenericView` component in `ObjectCreate`,  `ObjectList` and `ObjectRetrieve` views.

image

# TODO

- [x] Wait for `package-lock.json` to be updated in #3618 to avoid conflicts related to upgrading `@nautobot/nautobot-ui` dependency.


nautobot/ui/src/views/generic/ObjectCreate.js
```diff
@@ -18,14 +20,22 @@ export default function GenericObjectCreateView({ list_url }) {
     const { data, error } = useSWR(list_url, fetcher);
 
     if (!app_name || !model_name) {
-        return <></>;
+        return <GenericView />;
```
norbert-mieczkowski-codilime, line 23: We might want to consider `react-router` layout nesting not to duplicate so many `<GenericView>` usages. It is a feature invented exactly for this kind of layouts, but I didn't want to implement it in this PR to focus on just one thing at a time.

bryanculver: 👍 When I was looking at how we will do the login screen I figured we'll have to do just that.

glennmatthews (commented):
I pulled it down locally and the object-list views look great. I couldn't get any content to appear in the object-detail or object-create views but I **think** this is expected currently?

bryanculver (approved):

nautobot/ui/src/views/generic/GenericView.js
```diff
@@ -0,0 +1,247 @@
...
+    rows,
+}) {
+    const { pathname } = useLocation();
+
+    const menu = useGetUIMenuQuery();
+    const session = useGetSessionQuery();
+
+    const breadcrumbs = useMemo(
+        () =>
+            (menu?.data ?? []).reduce((breadcrumbs, root) => {
```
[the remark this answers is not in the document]
norbert-mieczkowski-codilime: the data structure created here is very component-specific, so I wouldn't move it anywhere outside. if so required, perhaps it would be a good idea to split it into functions returning one menu item and/or breadcrumb at a time, e.g. `getUIMenuItem(pathname, level)`, `getBreadcrumb(uiMenuItem)`, and composing according to specific requirements. also, enabling breadcrumbs overrides through `GenericView` props of course, but that's the simplest thing to do. would that work for you?

PR #3615, opened 2023-04-19
Add New UI Toggle

bryanculver:
# Closes: #3615
# What's Changed
- Add simple toggle buttons to toggle between new UIs and not. Let's make it closer to merge the new UI into `next` so we can cut down on drift and work faster

image
image

# TODO

- [x] Explanation of Change(s)
- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [NA] Attached Screenshots, Payload Example
- [NA] Unit, Integration Tests
- [NA] Documentation Updates (when adding/changing features)
- [NA] Example Plugin Updates (when adding/changing features)
- [x] Outline Remaining Work, Constraints from Design

PR #3632, opened 2023-04-21
Get tests passing for UI 2.0

### LINKED ISSUE #2901: Implement `get_absolute_url()` on `BaseModel`
### Proposed Changes

Base models should always provide `get_absolute_url() `so that each model doesn’t have to hard-code this as well. 

e.g. This should suffice:

```python
class BaseModel(models.Model):
    def get_absolute_url(self):
        return reverse(get_route_for_model(self), args=[self.pk])
```

### Justification

Since we’re aligning around pk for all objects, then this is even easier in that we know the app_label and model_name for the model, so we can automatically have something like above. We could just as easily align around the `natural_key_slug` (see: #2900) as well (e.g. `args=self.natural_key_slug` should we so desire). 

More importantly, this is another boiler plate reduction. The majority of model objects are using `{app_label}.{model}` as their detail route name, and `pk` as the unique identifier in detail routes, so this would eliminate most places where `get_absolute_url()` is required to be defined in the common case.

bryanculver:
# Closes: #DNE
# What's Changed

- Unmake breaking changes since many of the API changes are going to be opt-in (OPTIONS header, additional endpoint, etc) so we should be able to get the UI to pass our existing Python tests
- UI 2.0 already introduced the `web_url` payload on the serializer which may need to be tweaked once #3500 is merged
- To solve test and schema failures, close #2901

# TODO

- [x] Unit, Integration Tests
- [X] Example Plugin Updates (when adding/changing features)

### DIFF
--- a/nautobot/core/api/urls.py
+++ b/nautobot/core/api/urls.py
@@ -56,7 +56,7 @@ urlpatterns = [
     path("graphql/", GraphQLDRFAPIView.as_view(), name="graphql-api"),
     # Plugins
     path("plugins/", include((plugin_api_patterns, "plugins-api"))),
-    # Core (keeping for backwards compatibility for the moment)
+    # Core, keeping for backwards compatibility of the legacy UI (Dynamic Filter Form)
     path("core/", include((core_api_patterns, "core-api"))),
     # UI
     path("ui/", include((ui_api_patterns, "ui-api"))),
--- a/nautobot/core/api/urls.py
+++ b/nautobot/core/api/urls.py
@@ -56,6 +56,8 @@ urlpatterns = [
     path("graphql/", GraphQLDRFAPIView.as_view(), name="graphql-api"),
     # Plugins
     path("plugins/", include((plugin_api_patterns, "plugins-api"))),
+    # Core (keeping for backwards compatibility for the moment)
+    path("core/", include((core_api_patterns, "core-api"))),
     # UI
     path("ui/", include((ui_api_patterns, "ui-api"))),
 ]
--- a/nautobot/core/api/views.py
+++ b/nautobot/core/api/views.py
@@ -210,49 +210,49 @@ class ModelViewSetMixin:
         return super().get_serializer(*args, **kwargs)
 
     # TODO: the below needs to be either fixed or removed as part of issue #3042.
-    # def get_serializer_class(self):
-    #     logger = logging.getLogger("nautobot.core.api.views.ModelViewSet")
+    def get_serializer_class(self):
+        logger = logging.getLogger("nautobot.core.api.views.ModelViewSet")
 
-    #     # If using 'brief' mode, find and return the nested serializer for this model, if one exists
-    #     if self.brief:
-    #         logger.debug("Request is for 'brief' format; initializing nested serializer")
-    #         try:
-    #             serializer = get_serializer_for_model(self.queryset.model, prefix="Nested")
-    #             logger.debug(f"Using serializer {serializer}")
-    #             return serializer
-    #         except SerializerNotFound:
-    #             logger.debug(f"Nested serializer for {self.queryset.model} not found!")
+        # If using 'brief' mode, find and return the nested serializer for this model, if one exists
+        if self.brief:
+            logger.debug("Request is for 'brief' format; initializing nested serializer")
+            try:
+                serializer = get_serializer_for_model(self.queryset.model, prefix="Nested")
+                logger.debug(f"Using serializer {serializer}")
+                return serializer
+            except SerializerNotFound:
+                logger.debug(f"Nested serializer for {self.queryset.model} not found!")
 
-    #     # Fall back to the hard-coded serializer class
-    #     return self.serializer_class
+        # Fall back to the hard-coded serializer class
+        return self.serializer_class
 
     # TODO: this is part of issue #3042.
-    def get_serializer_context(self):
-        ctx = super().get_serializer_context()
-        ctx["request"] = None
-        try:
-            depth = int(self.request.query_params.get("depth", 0))
-        except ValueError:
-            depth = 0  # Ignore non-numeric parameters and keep default 0 depth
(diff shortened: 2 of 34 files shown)


nautobot/core/api/views.py
```diff
@@ -210,49 +210,49 @@ def get_serializer(self, *args, **kwargs):
...
-            depth = 0  # Ignore non-numeric parameters and keep default 0 depth
-        ctx["depth"] = depth
+    # def get_serializer_context(self):
+    #     ctx = super().get_serializer_context()
+    #     ctx["request"] = None
+    #     try:
+    #         depth = int(self.request.query_params.get("depth", 0))
+    #     except ValueError:
+    #         depth = 0  # Ignore non-numeric parameters and keep default 0 depth
+    #     ctx["depth"] = depth
```
bryanculver, line 237: Likely to be replaced with changes in #3500

nautobot/core/context_processors.py
```diff
@@ -36,7 +36,7 @@ def settings(request):
 
     return {
         "settings": django_settings,
-        "base_template": "base_react.html" if request.COOKIES.get("newui", False) else "base_django.html",
+        "root_template": "base_react.html" if request.COOKIES.get("newui", False) else "base_django.html",
```
bryanculver, line 39: #3615 Introduced a bug on some views that were using the `base_template` context. Missed because we didn't have passing tests, which this should fix!!!

nautobot/core/testing/views.py
```diff
@@ -772,7 +772,9 @@ def test_list_objects_with_constrained_permission(self):
             if hasattr(self.model, "name"):
                 self.assertIn(instance1.name, content, msg=content)
                 self.assertNotIn(instance2.name, content, msg=content)
-            elif hasattr(self.model, "get_absolute_url"):
+            elif (
+                hasattr(self.model, "get_absolute_url") and instance1.get_absolute_url() is not None
+            ):  # TODO fix RelationshipAssociation
```
bryanculver, line 777: RelationshipAssociation Doesn't play well with this test case so may need some tests overloaded since this and like `DynamicGroupMembers` don't make sense to have their own list vies.

nautobot/core/utils/lookup.py
```diff
@@ -76,7 +76,9 @@ def get_route_for_model(model, action, api=False):
 
     suffix = "" if not api else "-api"
     prefix = f"{model._meta.app_label}{suffix}:{model._meta.model_name}"
-    sep = "_" if not api else "-"
+    sep = ""
+    if action != "":
+        sep = "_" if not api else "-"
```
bryanculver, line 81: Many of our routes right now do not end in `_detail` or `_retrieve`. For now, this works and we can choose to clean up our routes whenever instead of scope-creaping this.

nautobot/core/views/generic.py
```diff
@@ -126,23 +126,22 @@ def get(self, request, *args, **kwargs):
...
+
+            temp_fake_context = {
+                "object": instance,
+                "request": request,
+                "settings": {},
+                "csrf_token": "",
+                "perms": {},
+            }
+
+            plugin_tabs = _get_registered_content(instance, "detail_tabs", temp_fake_context, return_html=False)
```
bryanculver, line 144: When get to a polished implementation for plugin views we will likely overhaul this.

nautobot/core/api/urls.py
```diff
@@ -56,6 +56,8 @@
     path("graphql/", GraphQLDRFAPIView.as_view(), name="graphql-api"),
     # Plugins
     path("plugins/", include((plugin_api_patterns, "plugins-api"))),
+    # Core (keeping for backwards compatibility for the moment)
```
glennmatthews, line 59: backwards compatibility with what? (i.e. please add more context to the comment :-))

nautobot/core/models/__init__.py
```diff
@@ -34,6 +36,29 @@ def present_in_database(self):
...
+
+            for action in actions:
+                route = get_route_for_model(self, action)
+
+                try:
+                    return reverse(route, kwargs={field: getattr(self, field)})
+                except NoReverseMatch:
+                    continue
+
+        return None
```
glennmatthews, line 60: Wonder if it might be better from a Django API standpoint to raise `AttributeError` (as if the model didn't have a `get_absolute_url` function at all) in the failure case rather than returning `None`? I know it may make for needing more exception handling in our code but it might be worth it.

nautobot/extras/models/relationships.py
```diff
@@ -738,6 +735,9 @@ def _get_genericforeignkey(self, name):
 
         return None
 
+    def get_absolute_url(self):
+        self.relationship.get_absolute_url()
```
glennmatthews, line 739:
```suggestion
        return self.relationship.get_absolute_url()
```

glennmatthews (approved):

nautobot/extras/models/relationships.py
bryanculver: I'm wondering if that was what was causing https://github.com/nautobot/nautobot/pull/3632#discussion_r1173202511 to still fail...

nautobot/core/views/mixins.py
```diff
@@ -181,7 +181,13 @@ def get_return_url(self, request, obj=None):
         # Note that the use of both `obj.present_in_database` and `obj.pk` is correct here because this conditional
         # handles all three of the create, update, and delete operations. When Django deletes an instance
         # from the DB, it sets the instance's PK field to None, regardless of the use of a UUID.
-        if obj is not None and obj.present_in_database and obj.pk and hasattr(obj, "get_absolute_url"):
+        if (
+            obj is not None
+            and obj.present_in_database
+            and obj.pk
+            and hasattr(obj, "get_absolute_url")
+            and obj.get_absolute_url() is not None
```
bryanculver, line 189: Handle error now

PR #3633, opened 2023-04-21
Apply Custom Validators without the need of a DB lookup for ContentType

bryanculver:
# Closes: #DNE
# What's Changed
Observing the need for this: https://github.com/nautobot/nautobot/pull/3614/files#diff-cca2a364f59877c49898f8d14d311f7567adfb80b5aec6fa51748be4f79ef9b5R44-R46

This might be one of the weird DB calls we see on startup pre-fork in Celery 🤷 

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/plugins/validators.py
+++ b/nautobot/extras/plugins/validators.py
@@ -40,7 +40,7 @@ def wrap_model_clean_methods():
     """
     Helper function that wraps plugin model validator registered clean methods for all applicable models
     """
-    for app_label, models in FeatureQuery("custom_validators").get_dict():
+    for app_label, models in FeatureQuery("custom_validators").as_dict():
         for model in models:
             model_class = apps.get_model(app_label=app_label, model_name=model)
             model_class.clean = custom_validator_clean(model_class.clean)
--- a/nautobot/extras/plugins/validators.py
+++ b/nautobot/extras/plugins/validators.py
@@ -1,6 +1,6 @@
 from functools import wraps
 
-from django.contrib.contenttypes.models import ContentType
+from django.apps import apps
 
 from nautobot.extras.plugins import CustomValidator
 from nautobot.extras.registry import registry
@@ -40,6 +40,7 @@ def wrap_model_clean_methods():
     """
     Helper function that wraps plugin model validator registered clean methods for all applicable models
     """
-    for model in ContentType.objects.filter(FeatureQuery("custom_validators").get_query()):
-        model_class = model.model_class()
-        model_class.clean = custom_validator_clean(model_class.clean)
+    for app_label, models in FeatureQuery("custom_validators").get_dict():
+        for model in models:
+            model_class = apps.get_model(app_label=app_label, model_name=model)
+            model_class.clean = custom_validator_clean(model_class.clean)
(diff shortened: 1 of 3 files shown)


nautobot/extras/plugins/validators.py
```diff
@@ -40,6 +40,7 @@ def wrap_model_clean_methods():
     """
     Helper function that wraps plugin model validator registered clean methods for all applicable models
     """
-    for model in ContentType.objects.filter(FeatureQuery("custom_validators").get_query()):
-        model_class = model.model_class()
-        model_class.clean = custom_validator_clean(model_class.clean)
+    for app_label, models in FeatureQuery("custom_validators").get_dict():
+        for model in models:
+            model_class = apps.get_model(app_label=app_label, model_name=model)
+            model_class.clean = custom_validator_clean(model_class.clean)
```
bryanculver, line 46: I _assume_ in this case we don't need the ContentType for some FK lookup in the DB but we just need to wrap the class' clean method, probably better to go straight to the Django for that. Cut out the middle man I _think_.

itdependsnetworks:
Brilliant!

nautobot/extras/plugins/validators.py
```diff
@@ -40,6 +40,7 @@ def wrap_model_clean_methods():
     """
     Helper function that wraps plugin model validator registered clean methods for all applicable models
     """
-    for model in ContentType.objects.filter(FeatureQuery("custom_validators").get_query()):
-        model_class = model.model_class()
-        model_class.clean = custom_validator_clean(model_class.clean)
+    for app_label, models in FeatureQuery("custom_validators").get_dict():
+        for model in models:
+            model_class = apps.get_model(app_label=app_label, model_name=model)
```
glennmatthews, line 45: Possibly might need/want the `wrap_model_clean_methods` function to take an `apps` parameter (defaulting to the global `apps`) so that for example we could call this function during a migration or a `nautobot_database_ready` signal handler where `apps` is a different context? Not sure, just thinking aloud.

nautobot/extras/utils.py
```diff
@@ -123,11 +123,17 @@ def get_query(self):
         Given an extras feature, return a Q object for content type lookup
         """
         query = Q()
-        for app_label, models in registry["model_features"][self.feature].items():
+        for app_label, models in self.get_dict():
             query |= Q(app_label=app_label, model__in=models)
 
         return query
 
+    def get_dict(self):
```
glennmatthews, line 131: Maybe `items` or `as_dict` instead? Not blocking.

glennmatthews (approved):

nautobot/extras/plugins/validators.py
bryanculver: @glennmatthews your thought is so if a custom validator is available we can use it in migrations now? Because I assume that doesn't work at present?

glennmatthews: Yeah, seems pretty clear it wouldn't work at present. We *probably* don't want them to apply during migrations in general (that would be troublesome) but I feel like there *might* be a use case for having them apply during the database-ready callback signals? As I said, not sure. :-)

bryanculver: Required for this PR or future housekeeping/feature?

glennmatthews: Definitely not required.

Issue #1633, opened 2022-04-12
Remove monkey-mixin for `TaggableManager`


jathanism:
### Proposed Changes

For v2, per [this PR comment](https://github.com/nautobot/nautobot/pull/1505/files/fdfd0df621f195b45f19ee922e7b5d7c9bf2bd05) 

- [ ] Remove the `monkey_mix` from `nautobot.core.apps.CoreConfig.ready()`
- [ ] Convert `nautobot.extras.models.mixins.TaggableManagerMonkeyMixin` into a subclass of `taggit.managers.TaggableManager`  instead of a mixin and move it to `nautobot.extras.models.managers.TaggableManager` (assuming this stays in `extras`)
- [ ] Replace all `from taggit.managers import TaggableManager` references to `from nautobot.extras.models.managers import TaggableManager`
- [ ] Generate database migrations to account for the underlying model code changes (no schema changes required)

### Justification

This should be removed in v2 since we will no longer care about backwards-incompatibility and we can assert forward migration to move away from the need to monkey-mix `taggit.managers.TaggableManager`.

glennmatthews:
Implemented in #3641.

PR #3641, opened 2023-04-21
Remove TaggableManager monkey-mixin and update migrations

glennmatthews:
# Closes: #1633 
# What's Changed

- Instead of monkey-patching `taggit.managers.TaggableManger`, create `nautobot.core.models.fields.TagsField` as a proper subclass and use it throughout.
   - This inevitably results in a bunch more schema migrations, but what else is new for 2.0?
- Similarly, create `nautobot.core.models.managers.TagsManager` as a subclass of `taggit.managers._TaggableManager` and use it as appropriate as well.
- Change `VMInterface` to be a proper `PrimaryModel` rather than a cobbled-together model (that is "almost" a PrimaryModel) with its own bespoke `tags` field. This introduces `created` and `last_updated` fields to the model as a bonus.

# TODO
- [x] Unit, Integration Tests


nautobot/core/testing/mixins.py
```diff
@@ -150,14 +149,13 @@ def assertHttpStatus(self, response, expected_status, msg=None):
...
-                err = response.data
-            else:
-                # Attempt to extract form validation errors from the response HTML
-                form_errors = testing.extract_form_failures(response.content.decode(response.charset))
-                err = form_errors or response.content.decode(response.charset) or "No data"
-            err_message = f"Expected HTTP status(es) {expected_status}; received {response.status_code}: {err}"
+                err_message += f"\n{response.data}"
+            # Attempt to extract form validation errors from the response HTML
+            form_errors = testing.extract_form_failures(response.content.decode(response.charset))
+            err_message += "\n" + str(form_errors or response.content.decode(response.charset) or "No data")
```
glennmatthews, line 158: In troubleshooting some test failures, I found that for `NautobotUIViewSet`, the response includes a REST-style `data` but also has information in `response.content` - so I changed this from an if/else case to a cumulative construction of the `err_message`.

HanlinMiao (approved):
Seems like `pylint` isn't happy about something, but otherwise looks good!

PR #3645, opened 2023-04-24
Fix #1721 - remove django-cacheops

### LINKED ISSUE #1721: Remove django-cacheops dependency/usage

glennmatthews:
# Closes: #1721 
# What's Changed

- Remove dependency on `django-cacheops`.
- Replace our limited usage of django-cacheops with direct Django caching.
- Update docs and tests, etc.

# TODO
- [x] Unit, Integration Tests


install.sh
```diff
@@ -104,11 +104,6 @@ COMMAND="nautobot-server clearsessions"
 echo "Removing expired user sessions ($COMMAND)..."
 eval $COMMAND || exit 1
 
-# Clear all cached data
-COMMAND="nautobot-server invalidate all"
-echo "Clearing cache data ($COMMAND)..."
-eval $COMMAND || exit 1
-
```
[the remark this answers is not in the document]
glennmatthews: Could possibly do a `cache.clear()` though that's a bit of a nuclear option.

nautobot/core/settings.py
```diff
@@ -648,34 +647,14 @@
...
-
 # The django-redis cache is used to establish concurrent locks using Redis.
 CACHES = {
     "default": {
         "BACKEND": os.getenv(
             "NAUTOBOT_CACHES_BACKEND",
             "django_prometheus.cache.backends.redis.RedisCache" if METRICS_ENABLED else "django_redis.cache.RedisCache",
         ),
-        "LOCATION": parse_redis_connection(redis_database=0),
+        "LOCATION": parse_redis_connection(redis_database=1),
```
glennmatthews, line 657: I'm not entirely sure about the correctness and implications of this change. It "feels" right to me that CACHES should use a different redis DB than Celery though. Probably needs to be documented as a change at least. Tagging @nniehoff for some visibility.

nniehoff: This makes sense to me, most of our deployments are explicit and we separate the database indexes... Also it's important to note that redis clusters (not sentinel) only support 1 DB, in which case it might be nice to use the same one but I think it's still cleaner to not

PR #3669, opened 2023-05-01
Add Time, Status Indexes to JobResult

bryanculver:
# Closes: #DNE
# What's Changed

- Added indexes to `JobResult` to speed up list and homepage performance.

# TODO


nautobot/extras/models/jobs.py
```diff
@@ -586,6 +586,32 @@ def __init__(self, *args, **kwargs):
...
         get_latest_by = "created"
+        indexes = [
+            models.Index(
+                name="extras_jobresult_created_idx",
+                fields=["created"],
+            ),
+            models.Index(
+                name="extras_jobresult_rcreated_idx",
+                fields=["-created"],
+            ),
```
glennmatthews, line 597: Ignorant question - is there actually a need/value to indexing the same field in both directions like this? I would have naively assumed that one index would cover both directions. (and in that case, wouldn't adding `db_index=True` on the `created` field solve this more succinctly?)

jathanism: I would rather see the forward indexes done on the field using `db_index=True`, only leaving the reverse indexes here in the `Meta.indexes`.

bryanculver: Sorry it took me a while to update. @glennmatthews @jathanism updated from feedback.

glennmatthews (approved):
Needs blackening, otherwise looks good. Thanks!

bryanculver:
> Needs blackening, otherwise looks good. Thanks!

🤦

Issue #3663, opened 2023-04-27
NautobotUIViewset with lookup_filed set to 'pk' but slug exist on model causes issues


jeffkala:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5-1.5.17
* Python version: 3.8
* Database platform, version: postgres
* Middleware(s): n/a

While migrating an existing application from oldering `generic.<x>views` to `NautobotUiViewset` hitting a bug where `edit` and `delete` buttons don't show up only when `lookup_field = "pk"` is used but `slug` field does exist on model.

 I have a model where the retrieve view doesn’t show edit/delete buttons. Is there any chance having a slug on a model but modifying everything else including lookup_field = "pk" to use PK could some how cause issues?  Another model that is basically identical is working fine (it does not have a slug field at all). And we’re using the same lookup_field = "pk" on it. Tried versions from 1.5.3 through 1.5.16 with the same result. This is the only model that has a slug field in GC so pulling at straws to figure it out.  Everything else is identical template wise/ model wises/ view wise etc.

### Steps to Reproduce
1. Migrate a models view from the older generic views to NautobotUiViewset.
2. Set `lookup_field = "pk"` on the model.
3. Change the `get_absolute_url` to using `args=[self.pk]`
4. Load the detail(retrieve) view.
5. Buttons (edit,delete) will not display at all.

To get the buttons to show up I can add below to the view.
```
    lookup_field = "pk"
    lookup_url_kwarg = "slug"
```

But the buttons themselves when clicked will fail since they're trying to resolve to the /blah/<slug> url.

### Expected Behavior
Eventhough the model has the slug field, I should be able to use `lookup_field` as a pk to be uniform throughout the app.

### Observed Behavior
buttons don't show up.  If workaround to get buttons to display is implemented the button links fail.

jeffkala:
related to https://github.com/nautobot/nautobot-plugin-golden-config/pull/471

bryanculver:
Thanks for the report @jeffkala

HanlinMiao:
Hi @jeffkala can you try the following steps to see if it works for your case?

1. Migrate a models view from the older generic views to NautobotUiViewset.
2. Set lookup_field = "pk" on the model.
3. Change the get_absolute_url to using args=[self.pk]
4. Load the detail(retrieve) view.
[...]

PR #3671, opened 2023-05-01
modified `delete_button` and `edit_button` template tags

HanlinMiao:
# Closes: #3663 
# What's Changed

Modified `delete_button` and `edit_button` template tags to lookup `pk` and `slug` in that order without the need to specify the lookup `key`.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/templates/dcim/rackgroup.html
+++ b/nautobot/dcim/templates/dcim/rackgroup.html
@@ -7,10 +7,10 @@
             {% clone_button object %}
         {% endif %}
         {% if perms.dcim.change_rackgroup %}
-            {% edit_button object key="pk" %}
+            {% edit_button object %}
         {% endif %}
         {% if perms.dcim.delete_rackgroup %}
-            {% delete_button object key="pk" %}
+            {% delete_button object %}
         {% endif %}
 {% endblock buttons %}
 
--- a/nautobot/dcim/templates/dcim/rackgroup.html
+++ b/nautobot/dcim/templates/dcim/rackgroup.html
@@ -2,18 +2,6 @@
 {% load buttons %}
 {% load helpers %}
 
-{% block buttons %}
-        {% if perms.dcim.add_rackgroup %}
-            {% clone_button object %}
-        {% endif %}
-        {% if perms.dcim.change_rackgroup %}
-            {% edit_button object %}
-        {% endif %}
-        {% if perms.dcim.delete_rackgroup %}
-            {% delete_button object %}
-        {% endif %}
-{% endblock buttons %}
-
 {% block content_left_page %}
         <div class="panel panel-default">
             <div class="panel-heading">
--- a/nautobot/dcim/templates/dcim/rackrole.html
+++ b/nautobot/dcim/templates/dcim/rackrole.html
@@ -7,10 +7,10 @@
             {% clone_button object %}
         {% endif %}
         {% if perms.dcim.change_rackrole %}
-            {% edit_button object key="pk" %}
+            {% edit_button object %}
         {% endif %}
         {% if perms.dcim.delete_rackrole %}
-            {% delete_button object key="pk" %}
+            {% delete_button object %}
         {% endif %}
 {% endblock buttons %}
 
--- a/nautobot/dcim/templates/dcim/rackrole.html
+++ b/nautobot/dcim/templates/dcim/rackrole.html
@@ -2,18 +2,6 @@
 {% load buttons %}
 {% load helpers %}
 
-{% block buttons %}
-        {% if perms.dcim.add_rackrole %}
-            {% clone_button object %}
-        {% endif %}
-        {% if perms.dcim.change_rackrole %}
-            {% edit_button object %}
-        {% endif %}
-        {% if perms.dcim.delete_rackrole %}
-            {% delete_button object %}
-        {% endif %}
-{% endblock buttons %}
-
 {% block content_left_page %}
         <div class="panel panel-default">
             <div class="panel-heading">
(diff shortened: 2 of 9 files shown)


nautobot/utilities/templatetags/buttons.py
```diff
@@ -30,59 +30,57 @@ def clone_button(instance):
...
     """
     viewname = get_route_for_model(instance, "edit")
 
-    # Assign kwargs
-    if hasattr(instance, key) and not use_pk:
-        kwargs = {key: getattr(instance, key)}
-    else:
-        kwargs = {"pk": instance.pk}
+    # We try different lookups to get a valid reverse url
+    lookups = ["pk", "slug"]
```
HanlinMiao, line 43: I might be missing a lookup key here. But I do not see any urls that are using `name` or `key` for lookup in develop

nautobot/dcim/templates/dcim/rackgroup.html
```diff
@@ -7,10 +7,10 @@
             {% clone_button object %}
         {% endif %}
         {% if perms.dcim.change_rackgroup %}
-            {% edit_button object key="pk" %}
+            {% edit_button object %}
         {% endif %}
         {% if perms.dcim.delete_rackgroup %}
-            {% delete_button object key="pk" %}
+            {% delete_button object %}
         {% endif %}
```
glennmatthews, line 14: Should be able to remove the `buttons` block from this template entirely now.

nautobot/dcim/templates/dcim/rackrole.html
```diff
@@ -7,10 +7,10 @@
...
         {% endif %}
         {% if perms.dcim.change_rackrole %}
-            {% edit_button object key="pk" %}
+            {% edit_button object %}
         {% endif %}
         {% if perms.dcim.delete_rackrole %}
-            {% delete_button object key="pk" %}
+            {% delete_button object %}
         {% endif %}
 {% endblock buttons %}
```
glennmatthews, line 15: ditto removing `block buttons`

nautobot/dcim/templates/dcim/region.html
```diff
@@ -7,10 +7,10 @@
...
         {% endif %}
         {% if perms.dcim.change_region %}
-            {% edit_button object key="pk" %}
+            {% edit_button object %}
         {% endif %}
         {% if perms.dcim.delete_region %}
-            {% delete_button object key="pk" %}
+            {% delete_button object %}
         {% endif %}
 {% endblock buttons %}
```
glennmatthews, line 15: ditto

nautobot/extras/templates/extras/customfield.html
```diff
@@ -8,10 +8,10 @@
...
         {% endif %}
         {% if perms.extras.change_customfield %}
-            {% edit_button object key="slug" %}
+            {% edit_button object %}
         {% endif %}
         {% if perms.extras.delete_customfield %}
-            {% delete_button object key="slug" %}
+            {% delete_button object %}
         {% endif %}
 {% endblock buttons %}
```
glennmatthews, line 16: ditto

nautobot/ipam/templates/ipam/vlangroup.html
```diff
@@ -9,10 +9,10 @@
...
         {% endif %}
         {% if perms.ipam.change_vlangroup %}
-            {% edit_button object key="pk" %}
+            {% edit_button object %}
         {% endif %}
         {% if perms.ipam.delete_vlangroup %}
-            {% delete_button object key="pk" %}
+            {% delete_button object %}
         {% endif %}
 {% endblock buttons %}
```
glennmatthews, line 17: ditto

nautobot/utilities/templatetags/buttons.py
glennmatthews: Maybe include `key` just because we know it'll be needed in `next` for the time being?

nautobot/utilities/templatetags/buttons.py
```diff
@@ -30,59 +30,57 @@ def clone_button(instance):
 
 
 @register.inclusion_tag("buttons/edit.html")
-def edit_button(instance, use_pk=False, key="slug"):
+def edit_button(instance):
```
glennmatthews, line 33: This is a breaking API change and should be avoided. Better to keep these two params for now but make them no-ops.

nautobot/utilities/templatetags/buttons.py
```diff
@@ -30,59 +30,57 @@ def clone_button(instance):
...
+    for lookup in lookups:
+        if hasattr(instance, lookup):
+            kwargs = {lookup: getattr(instance, lookup)}
+        else:
+            continue
+        try:
+            url = reverse(viewname, kwargs=kwargs)
+            return {"url": url}
+        except NoReverseMatch:
+            continue
```
glennmatthews, line 54:
```suggestion
        if hasattr(instance, lookup):
            kwargs = {lookup: getattr(instance, lookup)}
            try:
                url = reverse(viewname, kwargs=kwargs)
                return {"url": url}
            except NoReverseMatch:
                continue
```

nautobot/utilities/templatetags/buttons.py
```diff
@@ -30,59 +30,57 @@ def clone_button(instance):
...
 
-    return {
-        "url": url,
-    }
+    return {"url": None}
 
 
 @register.inclusion_tag("buttons/delete.html")
-def delete_button(instance, use_pk=False, key="slug"):
+def delete_button(instance):
```
glennmatthews, line 60: same API breakage concern here.

glennmatthews (changes requested):

nautobot/dcim/templates/dcim/rackgroup.html
HanlinMiao: Removing the block buttons removed the `clone button`. However these objects do not `clone_fields` to begin with so I do not think it was supposed to be there.

PR #3675, opened 2023-05-01
fix job edit form silently failing

gsnider2195:
# Closes: DNE
# What's Changed

The `slug` field was removed from the form template but not the form, so when editing a job the form would always fail because it expected a slug that cannot be provided.

# TODO

- [ ] Unit, Integration Tests


glennmatthews:
The `Job` model still has a `slug` field - should the correct fix be to add the form field back to the template, instead of removing it from the form?

glennmatthews:
Is this the reason why UT is failing in `next-ui` as well?

gsnider2195:
> The `Job` model still has a `slug` field - should the correct fix be to add the form field back to the template, instead of removing it from the form?

I think so. We'll eventually be able to remove the slug field on Job but since it's still there I agree that the appropriate fix is to re-add the field to the html template.

> Is this the reason why UT is failing in `next-ui` as well?

It is. I made this change on that branch so that we could enable/run jobs to populate the job results table on the new home page. I'll fix it over there too.

glennmatthews (approved):

bryanculver (approved):

PR #3679, opened 2023-05-02
Change REST API to use hyperlinks instead of raw PKs for related objects with depth 0

glennmatthews:
# Related to #3691 
# What's Changed

Change the base REST API serializer inheritance from `ModelSerializer` to `HyperlinkedModelSerializer`. This changes our rendering of related objects at depth 0 from raw PK UUIDs (which don't provide enough useful information, such as the content-type of the related object) to API URLs (which intrinsically show the type of the related object, and are also walkable by API clients to easily retrieve details of specific related objects if desired. Note that on the "write" side of the API, it's still possible to describe related objects by raw PK or by nested serializer data - this change only impacts the "read" side of the API.

This wasn't quite a drop-in replacement because for whatever reason, DRF's HyperlinkedModelSerializer, HyperlinkedRelatedField, and HyperlinkedIdentityField classes are **not** aware of URL namespacing, and so they default to trying to automatic reverse urls like `<modelname>-detail` instead of what we need of `<applabel>-api:<modelname>-detail`. I've overridden what I believe to be all of the places where this was being used incorrectly.

This lets us remove the explicit `url` field from most of our serializers as a bonus, because it's now provided automatically by the serializer now that we have correct logic in place for finding the URL for a given object instance.

image

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/core/api/serializers.py
+++ b/nautobot/core/api/serializers.py
@@ -1,7 +1,6 @@
 import logging
 import uuid
 
-from abc import abstractmethod
 from django.core.exceptions import (
     FieldError,
     MultipleObjectsReturned,
--- a/nautobot/core/api/serializers.py
+++ b/nautobot/core/api/serializers.py
@@ -107,7 +107,14 @@ class NautobotHyperlinkedRelatedField(WritableSerializerMixin, serializers.Hyper
         DRF defaults to '{model_name}-detail' instead of '{app_label}:{model_name}-detail'.
         """
         if "view_name" not in kwargs or (kwargs["view_name"].endswith("-detail") and ":" not in kwargs["view_name"]):
-            kwargs["view_name"] = get_route_for_model(kwargs["queryset"].model, "detail", api=True)
+            if "queryset" not in kwargs:
+                logger.warning(
+                    '"view_name=%r" is probably incorrect for this related API field; '
+                    'unable to determine the correct "view_name" as "queryset" wasn\'t specified',
+                    kwargs["view_name"],
+                )
+            else:
+                kwargs["view_name"] = get_route_for_model(kwargs["queryset"].model, "detail", api=True)
         super().__init__(*args, **kwargs)
 
 
--- a/nautobot/core/api/serializers.py
+++ b/nautobot/core/api/serializers.py
@@ -99,11 +99,20 @@ class OptInFieldsMixin:
         return self.__pruned_fields
 
 
-class NautobotPrimaryKeyRelatedField(WritableSerializerMixin, serializers.PrimaryKeyRelatedField):
-    """DRF's built-in PrimaryKeyRelatedField combined with custom to_internal_value() function"""
+class NautobotHyperlinkedRelatedField(WritableSerializerMixin, serializers.HyperlinkedRelatedField):
+    """DRF's built-in HyperlinkedRelatedField combined with custom to_internal_value() function"""
+
+    def __init__(self, *args, **kwargs):
+        """Override DRF's namespace-unaware default view_name logic for HyperlinkedRelatedField.
+
+        DRF defaults to '{model_name}-detail' instead of '{app_label}:{model_name}-detail'.
+        """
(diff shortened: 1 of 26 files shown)


itdependsnetworks:
Can you confirm/deny that the effect of this is that if as a plugin developer, I use `BaseModelSerializer`, it will automatically generate a field called "url" and it will correctly link.

glennmatthews:
> Can you confirm/deny that the effect of this is that if as a plugin developer, I use `BaseModelSerializer`, it will automatically generate a field called "url" and it will correctly link.

That should be correct, yes.

glennmatthews:
To be clear that is not the primary intent of this PR, but it is a bonus. I'll add more details tomorrow.

nautobot/core/tests/test_api.py
```diff
@@ -401,7 +401,9 @@ def test_related_by_pk(self):
 
         response = self.client.post(url, data, format="json", **self.header)
         self.assertHttpStatus(response, status.HTTP_201_CREATED)
-        self.assertEqual(str(response.data["location"]), str(self.location1.pk))
+        self.assertEqual(
+            str(response.data["location"]), f"http://nautobot.example.com{self.location1.get_absolute_url(api=True)}"
```
jathanism, line 405: Instead of hard-coding `http://nautobot.example.com{self.location1.get_absolute_url(api=True)}` in a bunch of places, it would be cleaner to use a request factory, and stash this in a helper method on the base test class.

e.g. 

```python
from django.test import RequestFactory

def get_api_url(self, obj):
    request = RequestFactory().request(self.client._base_environ()["SERVER_NAME"])
    return request.build_absolute_uri(obj.get_absolute_url(api=True))
```

And then have the test code call `self.get_api_url(obj)`

jathanism (changes requested):

nautobot/core/api/serializers.py
```diff
@@ -116,13 +131,12 @@ class BaseModelSerializer(OptInFieldsMixin, serializers.ModelSerializer):
...
     PolymorphicProxySerializer class defined below, relies upon this field as a way to identify to the client
     which of several possible serializers are in use for a given attribute.
-    - ensures that `url` field has to be explicitly declared with a valid view_name on each serializer subclass.
     - supports `?depth` query parameter. It is passed in as `nested_depth` to the `build_nested_field()` function to enable the
     dynamic generation of nested serializers.
     """
 
     display = serializers.SerializerMethodField(read_only=True, help_text="Human friendly display value")
-    serializer_related_field = NautobotPrimaryKeyRelatedField
+    serializer_related_field = NautobotHyperlinkedRelatedField
```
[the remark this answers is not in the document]
glennmatthews: I *swear* I grepped for all instances of `NautobotPrimaryKey` but clearly I missed this. Thanks for the catch!

nautobot/core/api/utils.py
```diff
@@ -335,7 +335,9 @@ def return_nested_serializer_data_based_on_depth(serializer, depth, obj, obj_rel
     if obj_related_field.__class__.__name__ == "RelatedManager":
         result = []
         if depth == 0:
-            result = obj_related_field.values_list("pk", flat=True)
+            result = [entry.get_absolute_url(api=True) for entry in obj_related_field.all()]
+            if serializer.context.get("request"):
+                result = [serializer.context.get("request").build_absolute_uri(url) for url in result]
```
HanlinMiao, line 340: Why the distinction here with `request` vs `no request`? Is `get_absolute_url` here not enough to make the BrowableAPI walkable?

HanlinMiao (approved):

nautobot/core/api/utils.py
glennmatthews: It's for self-consistency. `get_absolute_url` alone doesn't know about the protocol/host/port info (it returns e.g. `/api/dcim/devices/<uuid>`), `request.build_absolute_uri` is what adds that information (`http://localhost:3000/api/dcim/devices/<uuid>`).

PR #3680, opened 2023-05-03
Changed device component instantiation to be a separate method.

Kircheneer:
# Closes: #DNE
# What's Changed

`Device.save` now calls `Device.create_components` to create its components. The advantage lies in the fact that now that method can be used to populate the components when the `save` method is not available, such as with `bulk_create`.

# TODO

### DIFF
--- a/nautobot/docs/models/dcim/device.md
+++ b/nautobot/docs/models/dcim/device.md
@@ -18,9 +18,9 @@ For Devices forming a group (Failover, Load-Sharing, Redundacy or similar) refer
 
 ## Developer API
 
-The `Device` Django model class supports a method called `create_components()`. This method is normally called during `Device.save()`, which is called whenever you save create a Device via the GUI or the REST API, but if you are working directly in the ORM and encounter one of the two following scenarios, `Device.save()` is not called:
+The `Device` Django model class supports a method called `create_components()`. This method is normally called during `device_instance.save()`, which is called whenever you save create a Device via the GUI or the REST API, but if you are working directly in the ORM and encounter one of the two following scenarios, `device_instance.save()` is not called:
 
-- Usage of `Device.objects.bulk_create()` to perform a bulk creation of Device objects
-- Usage of `Device.save()` during handling of the `nautobot_database_ready` signal (which uses [historical models](https://docs.djangoproject.com/en/3.2/topics/migrations/#historical-models))
+- Usage of `device_instance.objects.bulk_create()` to perform a bulk creation of Device objects
+- Usage of `device_instance.save()` during handling of the `nautobot_database_ready` signal (which uses [historical models](https://docs.djangoproject.com/en/3.2/topics/migrations/#historical-models))
 
-In these cases you will have to manually run `Device.create_components()` in order to instantiate the [device type's](devicetype.md) component templates (interfaces, power ports, etc.).
+In these cases you will have to manually run `device_instance.create_components()` in order to instantiate the [device type's](devicetype.md) component templates (interfaces, power ports, etc.).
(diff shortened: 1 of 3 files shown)


Kircheneer:
Is there a place where I should document the availability of this API to developers?

glennmatthews (commented):
Definitely a step in the right direction, thank you!

As far as documentation goes, there isn't a great place at present, but I might suggest just adding a note in `nautobot/docs/models/dcim/device.md` for now?

Kircheneer:
Also this would be a prime use case for `mkdocstrings` imo - but I found it wasn't set up properly. Do we have a working version of that somewhere where the `nautobot` package is available to the `mkdocs` container?

nautobot/docs/models/dcim/device.md
```diff
@@ -15,3 +15,12 @@ Device names must be unique within a site, unless the device has been assigned t
...
 For Devices forming a group (Failover, Load-Sharing, Redundacy or similar) refer to [Device Redundancy Groups](deviceredundancygroup.md) model documentation.
+
+## Developer API
+
+The `Device` Django ORM class supports a method called `create_components`. This method is normally called during `Device.save`, which is called whenever you save create a device via the GUI or the REST API, but if you are working directly in the ORM and encounter one of the two following scenarios, `Device.save` is not used:
+
+- Usage of `Device.bulk_create`
+- Usage of `Device.save` during handling of the `nautobot_database_ready` signal (uses [historical models](https://docs.djangoproject.com/en/3.2/topics/migrations/#historical-models))
+
+In these cases you will have to 'manually' re-run `Device.create_components` in order to instantiate the [device type's](devicetype.md) component templates (interfaces, power ports, etc.).
```
jathanism, line 26:
```suggestion
The `Device` Django model class supports a method called `create_components()`. This method is normally called during `Device.save()`, which is called whenever you save create a Device via the GUI or the REST API, but if you are working directly in the ORM and encounter one of the two following scenarios, `Device.save()` is not called:

- Usage of `Device.objects.bulk_create()` to perform a bulk creation of Device objects
- Usage of `Device.save()` during handling of the `nautobot_database_ready` signal (which uses [historical models](https://docs.djangoproject.com/en/3.2/topics/migrations/#historical-models))

In these cases you will have to manually run `Device.create_components()` in order to instantiate the [device type's](devicetype.md) component templates (interfaces, power ports, etc.).
```

jathanism (changes requested):

nautobot/docs/models/dcim/device.md
```diff
@@ -15,3 +15,12 @@ Device names must be unique within a site, unless the device has been assigned t
...
 For Devices forming a group (Failover, Load-Sharing, Redundacy or similar) refer to [Device Redundancy Groups](deviceredundancygroup.md) model documentation.
+
+## Developer API
+
+The `Device` Django model class supports a method called `create_components()`. This method is normally called during `Device.save()`, which is called whenever you save create a Device via the GUI or the REST API, but if you are working directly in the ORM and encounter one of the two following scenarios, `Device.save()` is not called:
+
+- Usage of `Device.objects.bulk_create()` to perform a bulk creation of Device objects
+- Usage of `Device.save()` during handling of the `nautobot_database_ready` signal (which uses [historical models](https://docs.djangoproject.com/en/3.2/topics/migrations/#historical-models))
+
+In these cases you will have to manually run `Device.create_components()` in order to instantiate the [device type's](devicetype.md) component templates (interfaces, power ports, etc.).
```
bryanculver, line 26: My only objection to this is the documentation reads that this is a class function, not an instance function. Could you instead do something like `device_instance.create_components()`?

Kircheneer: Done!

bryanculver:
@jathanism please re-review when you can

jathanism (approved):

PR #3696, opened 2023-05-04
Implemented list_display_fields for IPAM serializers.

jathanism:
# Closes: #3691 
# What's Changed

This PR goes back to using `Meta.fields = "__all__"` on serializers and instead makes correct use of [`uiSchema["ui:order"]`](https://rjsf-team.github.io/react-jsonschema-form/docs/usage/objects#specifying-property-order).

- Renamed `Meta.list_display` to `list_display_fields` for improved readability.
- Fixed a bug in UI `ObjectCreate` where uiSchema was `undefined`.
- Updated WritableSerializerMixin to support URLs as pk
- Introduced new `is_url()` helper that uses Django's `URLValidator` under the hood.
- PK will be extracted from the URL before passed onto PK validation.
- Reverted a change to `readOnly` and `format` JSON schema hints that was breaking validation
- Enabled `OPTIONS` in the browsable API.

# TODO

- [ ] Unit, Integration Tests


nautobot/core/api/metadata.py
```diff
@@ -144,16 +192,25 @@ def determine_actions(self, request, view):
                 view.request = request
         return actions
 
-    def determine_view_options(self, request, view, serializer):
+    def get_list_display_fields(self, serializer):
+        """Try to get the list display fields or default to an empty list."""
+        serializer_meta = getattr(serializer, "Meta", None)
+        return getattr(serializer_meta, "list_display_fields", None) or []
```
glennmatthews, line 198: Not sure name-wise about `list_display_fields`, would something like `default_list_display_fields` be clearer as to its meaning?

jathanism: We could just call them `default_columns` like they are in the tables?

nautobot/core/api/metadata.py
```diff
@@ -188,13 +246,21 @@ def determine_metadata(self, request, view):
         # If there's a serializer, do the needful to bind the schema/uiSchema.
         if hasattr(view, "get_serializer"):
             serializer = view.get_serializer()
+            # TODO(jathan): Bit of a WIP here. Will likely refactor. There might be cases where we
+            # want to explicitly override the UI field ordering, but that's not yet accounted for
+            # here. For now the assertion is always put the `list_display_fields` first, and then
+            # include the rest in whatever order.
```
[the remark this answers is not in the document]
itdependsnetworks: > we'd be better off to explicitly specify the ordering of all fields in every case

My unsolicited vote would be yes compared to the other logic

glennmatthews (approved):
Progress!

nautobot/core/api/metadata.py
```diff
@@ -144,16 +192,25 @@ def determine_actions(self, request, view):
                 view.request = request
         return actions
 
-    def determine_view_options(self, request, view, serializer):
+    def get_list_display_fields(self, serializer):
+        """Try to get the list display fields or default to an empty list."""
+        serializer_meta = getattr(serializer, "Meta", None)
+        return list(getattr(serializer_meta, "list_display_fields", []))
```
glennmatthews, line 198: Probably should default to something like `serializer.get_fields()` instead of an empty list?

glennmatthews (approved):

PR #3697, opened 2023-05-04
Enhance Add UI View

timizuoebideri1:
# What's Changed

- Check `DeviceTypeSerializer` for a shoddy method to manage field ordering and grouping on the serializer. Any serializer that does not explicitly declare a field_order or grouping on its own would only use the default field order.
- Additionally, working on a way to style the layout of your forms from the serializer; most likely, we can create a layout that can be used for all form views by default, and users can define their own layout styles or select from a template of layouts when creating and editing forms.

# TODO

- [ ] Unit, Integration Tests


nautobot/ui/src/styles/shared.css
```diff
@@ -0,0 +1,4 @@
+.nautobot-add-form>.form-group>.chakra-form-control>.css-qq8aff>.css-0 {
```
glennmatthews, line 1: I strongly suspect that this is fragile, not sure where `qq8aff` is coming from but I feel like it might change at any time.

nautobot/ui/src/views/generic/schema_sample.json
glennmatthews, line 1: Is this file used somewhere? I don't quite see where that is.

nautobot/ui/src/views/generic/ui_schema_sample.js
```diff
@@ -0,0 +1,66 @@
...
+    if (isMulti) {
+        extra_props.isMulti = true;
+    }
+
+    return <AsyncSelect cacheOptions loadOptions={loadOptions} defaultOptions {...extra_props} styles={colourStyles} />
+}
+
+const DynamicMultipleChoiceField = (props) => {
+    return <DynamicChoiceField isMulti={true} {...props} />
+}
```
glennmatthews, line 39: Should these be components rather than part of this view definition?

nautobot/ui/src/views/generic/ObjectCreate.js
```diff
@@ -1,13 +1,34 @@
...
 import Form from "@rjsf/chakra-ui";
 import validator from "@rjsf/validator-ajv8";
 import axios from "axios";
-import { useState } from "react";
 import { useLocation, useNavigate, useParams } from "react-router-dom";
 import useSWR from "swr";
+import { useEffect, useState } from "react";
 
 import GenericView from "@views/generic/GenericView";
+import { uiSchema, widget } from "./ui_schema_sample.js";
```
glennmatthews, line 11: Where is `uiSchema` used?

PR #3702, opened 2023-05-04
Don't render M2M fields on nested serializers, and other depth=1 performance improvements

glennmatthews:
# Closes: #DNE
# What's Changed

- On nested serializers, exclude ManyToMany fields, Tag fields, computed fields/relationships, and custom fields, in order to improve performance and reduce the number of SQL queries needed.
    - Added an `is_nested` property to the BaseModelSerializer class to facilitate implementing this sort of logic.
- Add caching to `TreeModel.display` because this was calculated and re-calculated quite frequently when listing objects with related Locations.
- Add some missing `select_related` entries on `DeviceViewSet` and `PrefixViewSet` that I noticed while profiling with DjDT.

Before (GET /api/dcim/locations/?depth=1):

Screenshot 2023-05-04 at 5 25 31 PM

After:

Screenshot 2023-05-04 at 5 25 46 PM

# TODO
- [ ] **Document this clearly in the REST API documentation!**
- [ ] Examine whether any of these (particularly custom fields) should be re-enabled and quantify the impact
- [ ] Unit, Integration Tests


itdependsnetworks:
Can you confirm my understanding:

* depth 0 gets ManyToMany fields, Tag fields, computed fields/relationships, and custom fields
* depth 1 and above fields can never get ManyToMany fields, Tag fields, computed fields/relationships, and custom fields

glennmatthews:
> Can you confirm my understanding:
> 
>     * depth 0 gets ManyToMany fields, Tag fields, computed fields/relationships, and custom fields
> 
>     * depth 1 and above fields can never get ManyToMany fields, Tag fields, computed fields/relationships, and custom fields

Just to be explicitly clear:

- at depth 0, the top-level object will include FK as well as M2M/Tag related objects by URL only, i.e.:

```json
        {
            "id": "9bc9abcb-8ac0-4ad4-bba5-15d22a5d0299",
            "url": "http://localhost:8080/api/dcim/locations/9bc9abcb-8ac0-4ad4-bba5-15d22a5d0299/",
            "display": "AMER → USA → BLV",
            "created": "2023-04-28T14:47:10.351659Z",
            "last_updated": "2023-05-04T17:53:35.983034Z",
            "name": "BLV",
            "slug": "amer-usa-blv",
...
            "status": "http://localhost:8080/api/extras/statuses/cd3f3acc-56e7-47da-8711-bea391edd635/",
            "parent": "http://localhost:8080/api/dcim/locations/6428666a-c839-41b2-ae0e-048f7902011f/",
            "tags": [
                "http://localhost:8080/api/extras/tags/6d7f17bb-b050-4fd5-a910-3c22cf6e8cf8/"
            ]
        },
```

- at depth 1, the top-level object will include FK, M2M and tag related objects as nested objects. The nested objects will include their own FK related objects by URL but will **NOT** (as changed by this PR) include M2M, Tags, Relationships, Computed Fields, even by URL:

```json
        {
            "id": "9bc9abcb-8ac0-4ad4-bba5-15d22a5d0299",
            "url": "http://localhost:8080/api/dcim/locations/9bc9abcb-8ac0-4ad4-bba5-15d22a5d0299/",
...
            "location_type": {
                "id": "2c4b28ad-0bcd-43a0-98db-fc4a9cb1348d",
                "url": "http://localhost:8080/api/dcim/location-types/2c4b28ad-0bcd-43a0-98db-fc4a9cb1348d/",
                "display": "region → country → site",
                "notes_url": "http://localhost:8080/api/dcim/location-types/2c4b28ad-0bcd-43a0-98db-fc4a9cb1348d/notes/",
                "tree_depth": null,
                "created": "2023-04-28T14:46:29.483447Z",
                "last_updated": "2023-04-28T14:47:38.556059Z",
                "name": "site",
                "slug": "site",
                "description": "",
                "nestable": false,
                "parent": "http://localhost:8080/api/dcim/location-types/7240674f-984d-46dc-9d8d-4ffad964b44f/"
            },  <----<< NOTE: no `tags` on the `location_type` nested object
            "tenant": {
                "id": "523cbc1b-7c83-411e-b1e5-db07741caac4",
                "url": "http://localhost:8080/api/tenancy/tenants/523cbc1b-7c83-411e-b1e5-db07741caac4/",
                "display": "IT Facility AMER",
                "notes_url": "http://localhost:8080/api/tenancy/tenants/523cbc1b-7c83-411e-b1e5-db07741caac4/notes/",
                "created": "2023-04-28T14:46:48.494676Z",
                "last_updated": "2023-04-28T14:47:19.730230Z",
                "name": "IT Facility AMER",
                "description": "IT Facility for AMER region",
                "comments": "",
                "tenant_group": null
            },   <---<< NOTE: no `tags` on the `tenant` nested object
            "tags": [
                {
                    "id": "6d7f17bb-b050-4fd5-a910-3c22cf6e8cf8",
                    "url": "http://localhost:8080/api/extras/tags/6d7f17bb-b050-4fd5-a910-3c22cf6e8cf8/",
                    "display": "Remote",
                    "notes_url": "http://localhost:8080/api/extras/tags/6d7f17bb-b050-4fd5-a910-3c22cf6e8cf8/notes/",
                    "name": "Remote",
                    "slug": "remote",
                    "created": "2023-05-04T17:52:56.242959Z",
                    "last_updated": "2023-05-04T17:52:56.243007Z",
                    "color": "ff9800",
                    "description": ""
                }  <----<< NOTE: no `content_types` on the `tags` related objects
            ]
        },
```

How exactly we handle Relationships on the base object at depth 0 vs depth 1 is I think something we're still sorting out.

nautobot/core/models/tree_queries.py
```diff
@@ -41,16 +43,24 @@ class Meta:
...
-            for ancestor in self.ancestors():
-                display_str += ancestor.name + " → "
+            if self.parent is not None:
+                display_str = self.parent.display + " → "
         except self.DoesNotExist:
             # Expected to occur at times during bulk-delete operations
             pass
         finally:
             display_str += self.name
+            cache.set(cache_key, display_str, 5)
```
gsnider2195, line 65: Would it make sense to cache the display name on the model and update the cache for the instance and all descendants any time the `parent` or `name` fields change? If the parent and name fields don't change often would that be more performant?

gsnider2195:
> The nested objects will include their own FK related objects by URL but will NOT (as changed by this PR) include M2M, Tags, Relationships, Computed Fields, even by URL

Could this be opt-in behavior? I could see this causing confusion when the top level object includes the m2m fields but they're missing from the nested objects.

glennmatthews:
> > The nested objects will include their own FK related objects by URL but will NOT (as changed by this PR) include M2M, Tags, Relationships, Computed Fields, even by URL
> 
> Could this be opt-in behavior? I could see this causing confusion when the top level object includes the m2m fields but they're missing from the nested objects.

That's definitely a valid concern. The issue is that looking up these related objects by default makes `depth=1` quite a bit slower than our nested-serializer approach was in v1.x. 

Possibly an alternate approach would be, rather than having things behave differently based on their depth/nestedness, explicitly making all M2M fields (including `tags`) opt-in fields that aren't included in the API at all by default? Happy to discuss further.

gsnider2195:
> > > The nested objects will include their own FK related objects by URL but will NOT (as changed by this PR) include M2M, Tags, Relationships, Computed Fields, even by URL
> > 
> > 
> > Could this be opt-in behavior? I could see this causing confusion when the top level object includes the m2m fields but they're missing from the nested objects.
> 
> That's definitely a valid concern. The issue is that looking up these related objects by default makes `depth=1` quite a bit slower than our nested-serializer approach was in v1.x.
> 
> Possibly an alternate approach would be, rather than having things behave differently based on their depth/nestedness, explicitly making all M2M fields (including `tags`) opt-in fields that aren't included in the API at all by default? Happy to discuss further.

The absence of the key should be sufficient information to know that the serializer isn't including that key in the data instead of `"tags": []` if tags were actually empty on the nested object.

nautobot/core/api/fields.py
```diff
@@ -104,25 +104,6 @@ def to_representation(self, obj):
         return f"{obj.app_label}.{obj.model}"
 
 
-class ObjectTypeField(serializers.CharField):
```
glennmatthews, line 107: I don't know why this stopped working when we switched to `fields = "__all__"`, but changing it to a SerializerMethodField seems good enough to get it included again - unfortunately this seems to have made drf-spectacular unhappy, so I'll need to change it back and figure out what's going on instead.

glennmatthews:
Need a fix from #3696 here it looks like.

nautobot/core/api/fields.py
```diff
@@ -113,14 +113,15 @@ def __init__(self, *args, read_only=True, **kwargs):  # pylint: disable=useless-
...
+        model = self.parent.Meta.model
+        return model._meta.label_lower
 
-        Implemented this way because `_value` may be None when generating the schema.
-        """
+    def to_representation(self, _value):
+        """Get the content-type of this serializer's model."""
         model = self.parent.Meta.model
-        return f"{model._meta.app_label}.{model._meta.model_name}"
+        return model._meta.label_lower
```
bryanculver, line 124: Took me a minute to parse. These seem identical. I'm sure we're overloading the bases but should we not have one call the other or make a common function for them?

nautobot/core/models/tree_queries.py
bryanculver: Cascading cache updates might be something we want to look into for dynamic group reversibility performance but for now, I think this is a valuable add.

If we like the pattern for using Redis for caching it makes it easier to do invalidation but I could also see a JSONField "cache" entry on objects for DB-backed caching:

```
>>> location._cache
{
    "display": ["Grad-grandparent Location", "Grandparent Location", "Parent Location", "Location"]
}
```

bryanculver (approved):

nautobot/core/api/fields.py
glennmatthews: I mean it's two lines of code :-). I'm honestly not sure why I needed to add a `get_attribute()` method to this class when it wasn't needed before, but object-type fields weren't being included properly until I added it. My best guess is that since it's not an actual model field, it wasn't being included by default in `__all__`?

nautobot/core/api/serializers.py
```diff
@@ -578,8 +612,8 @@ class NotesSerializerMixin(BaseModelSerializer):
     def get_field_names(self, declared_fields, info):
         """Ensure that fields includes "notes_url" field if applicable."""
         fields = list(super().get_field_names(declared_fields, info))
-        if hasattr(self.Meta.model, "notes"):
-            self.extend_field_names(fields, "notes_url")
```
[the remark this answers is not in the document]
glennmatthews: It seemed reasonable to me to assume that all models that use `NotesSerializerMixin` should be notable, but maybe that's a mistake - should models/apps be allowed to use `NautobotModelSerializer` (which includes this mixin) as their base class if they don't support notes? Happy to revert this if so.

nautobot/core/api/fields.py
bryanculver: I wonder if that is the cause of some of my weird sort behavior in #3708

bryanculver:
> Need a fix from #3696 here it looks like.

Does this need to be rebased now that #3696 is merged?

glennmatthews:
> > Need a fix from #3696 here it looks like.
> 
> Does this need to be rebased now that #3696 is merged?

I fixed it differently by explicitly adding the `display` and other fields back to the base class `get_field_names` in ad8d68f098388096bd5cbfc91c42e1734110e87d

nautobot/core/api/serializers.py
```diff
@@ -564,9 +595,14 @@ def _save_relationships(self, instance, relationships):
...
-        """Ensure that "relationships" is always included as an opt-in field."""
+        """Ensure that "relationships" is included as an opt-in field on root serializers."""
         fields = list(super().get_field_names(declared_fields, info))
-        self.extend_field_names(fields, "relationships", opt_in_only=True)
+        if not self.is_nested:
+            # Only include relationships as opt-in.
+            self.extend_field_names(fields, "relationships", opt_in_only=True)
+        else:
+            # As relationships are expensive, do not include them on nested serializers even if opted in.
+            fields.remove("relationships")
```
gsnider2195, line 605: Same as above, should this account for `ValueError` if `relationships` does not exist in `fields`?

gsnider2195 (approved):

nautobot/core/api/serializers.py
```diff
@@ -441,10 +466,16 @@ def get_computed_fields(self, obj):
...
         fields = list(super().get_field_names(declared_fields, info))
+        # Ensure that custom_fields field appears at the end, not the start, of the fields
         self.extend_field_names(fields, "custom_fields")
-        self.extend_field_names(fields, "computed_fields", opt_in_only=True)
+        if not self.is_nested:
+            # Only include computed_fields as opt-in.
+            self.extend_field_names(fields, "computed_fields", opt_in_only=True)
+        else:
+            # As computed fields are expensive, do not include them in nested serializers even if opted-in at the root
+            fields.remove("computed_fields")
```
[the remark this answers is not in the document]
glennmatthews: If a serializer uses `CustomFieldModelSerializer` but doesn't do `fields = "__all__"` then that could be a possibility, actually. Good thought.

PR #3704, opened 2023-05-05
Fix `related_name` performance for Git Repo Jobs

bryanculver:
# Closes: #3704
# What's Changed

Calling `related_name` on Job Results of GitRepository-based Jobs ultimately called `ensure_git_repository`. Scenarios needed to be addressed:

- If the `Job` is installed, we almost certainly have a `JobModel` for it, and therefore just grab the name from there
- If the `Job` has been removed (`GitRepository` removed for example) we should still have a hidden `JobModel` for it, so grab the name from there
- If the `Job` has been removed and so has the `JobModel` the `get_jobs` code shouldn't be doing a sync so we probably won't get it if we call `ensure_git_repository` anyway, in fact, it returns `None` and we fall back to `self.name`. Just do that without needing to call `ensure_git_repository`.

The side benefit of this change is that in scenario two above, we don't return the less-friendly class path when we have the JobModel anyway:

Screenshot 2023-05-04 at 23 28 40

instead of

Screenshot 2023-05-04 at 23 28 14

**Demo Before**

https://user-images.githubusercontent.com/31187/236374548-56a2b47e-92f0-4452-9d9d-347451eb6b9c.mov

**Demo After**

https://user-images.githubusercontent.com/31187/236374569-d4b024eb-7f75-4250-bc75-24da7619246a.mov

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/3704.fixed
+++ b/changes/3704.fixed
@@ -1,0 +1,1 @@
+Fixed GitRepository fetching on Home Page when getting repo-based Job's name.
--- a/nautobot/extras/models/jobs.py
+++ b/nautobot/extras/models/jobs.py
@@ -673,7 +673,7 @@ class JobResult(BaseModel, CustomFieldModel):
         if self.job_model:
             # No need to call related_object(), getting the Job class name if we already have a job_model
             return self.job_model.name
-        elif not self.job_model and self.obj_type == get_job_content_type():
+        elif self.obj_type == get_job_content_type():
             # Related object is an extras.Job subclass but the Job Model has been deleted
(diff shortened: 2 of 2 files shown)


glennmatthews (commented):
This reads as a bit of a behavior change to me in that `related_name` is supposed to prefer `self.related_object.name` over `self.name` - now it seems like there are cases where we use `self.name` even if a `self.related_object` exists?  Should we just be using `self.name` on the homepage instead of `self.related_name`?

glennmatthews (approved):

Issue #2909, opened 2022-11-23
Correct Existing Natural Keys 3 of 3 [Advanced] - Boundary Uniqueness


jathanism:
## Blocked by: #2900, #3244

```[tasklist]
### VLAN - 3
- [ ] **VLAN Group:** `name` becomes unique with an enumeration for collisions, drop unique with `location`
- [ ] **VLAN:** make `vlan_group` required, keep existing unique together, make default `vlan_group`
```

```[tasklist]
### User - 2
- [ ] **ObjectPermission:** `name` unique, move on. For collisions just enumerate.
```

- **Provider:** `name` is sufficient at this time
- **VRF:** Prefer `(namespace, name)` but `(namespace, rd)` also available
- **Virtualization/Cluster:** Would be good to consider in a future for making `group` part of it but fine for now since `name` is unique

glennmatthews:
> #### Tenancy vs. Global
> 
>     * Duplicate objects can only be in Tenants

Should this read "in *different* Tenants"?

> 
> #### IPAM Duplicates
> 
>     * Use a VRF with `enforce_unique=False`

Personally I'd like to get rid of this as even being an option. Can we instead enumerate the cases where "duplicate" entries are needed and figure out how to design the data model and uniqueness constraints accordingly?

>     * IPAM "Namespaces" should still be a thing, so you could have multiple namespaces per Tenant

Is this being proposed as a new/distinct model from `VRF`?

> 
>     * IPAddress shouldn’t have a relationship to Tenant. It should be inherited from the Prefix.

Not sure I understand why we want to remove this as a feature?

> ## Circuits
> ### Provider
> 
>     * Similar to what we’ve discussed with `Prefix` and `IPAddress`, the same uniqueness “boundaries” for other relations should be considered here, e.g:
>       
>       * `tenant`
>       * `role`

Neither of these are currently fields on a Provider. I'm also not clear on why `role` would be a uniqueness boundary?

> ## IPAM
> 
>     * Tenant/name might not be good for L2 objects (such as VRF, VLAN)
> 
>     * VLAN/location might make sense in this context, but not Tenant?
>       
>       * We might want `name/location` or `vid/location` as an example for VLAN?

I'm honestly not entirely clear what's being proposed for any of these three.

> ### VLAN
> 
>     * Constraints:
> 
> ```python
>         unique_together = [
>             # 2.0 TODO: since group is nullable and NULL != NULL, we can have multiple non-group VLANs with
[...]

Issue #2908, opened 2022-11-23
Correct Existing Natural Keys 2 of 3 [Intermediate] - Conditional Uniqueness


jathanism:
Correct or assert existing natural keys on objects where conditional uniquness is a requirement or concern.
---

```[tasklist]
### Tasks
- [ ] **Service:** Unique together: `(device, name)` and `(vm, name)` with natural key `(name, device, vm)`; for duplication use other fields to help add to name uniqueness - 2
- [ ] **Device Type:** A stretch goal is to rename `model` to `name`
- [ ] **Device:** Configurable natural key/unique, default to PK
- [x] **Computed Field:** Make `key` unique
- [ ] **VirtualChassis:** Make `name` unique, reject migration if duplicate (don't want to rename VCs) - 2
- [ ] **ScheduledJob:** Make `name` unique, enumerate if duplicate, update help text and auto-assign name (the uuid) for one-off jobs
- [ ] **JobResult:** Ensure `pk` is called out as the natural key
- [ ] **Job:** (see below)
- [ ] **ExportTemplate:** Move to `unique_together` on just `content_type` and `name`, reject migration if duplicate
- [ ] **ConfigContextSchema** Move to `name` unique, reject migration if duplicate
- [ ] **ConfigContext** Move to `name` unique, reject migration if duplicate
```
---
**Job**
- Relax unique on `name` because it will be make unique together with `grouping`: `unique_together = ["grouping", "name"]`
- Make the natural key `class_path` which is unique
- Other fields (git repository, source, etc) being removed in #765

---

- **VLAN Group:** discussed in #2909
- **Circuit:** Good enough for now with `["provider", "cid"]`
- **Circuit Termination:** Good enough with `["circuit", "term_side"]`
- **Cable:** Good enough for now, just could use some UX improvements for describing it
- **GitRepository** `name` good enough for now

bryanculver:
Remaining stories carved into #3845, #3846, #3847, #3848, #3849

PR #3708, opened 2023-05-06
Add Ordering Support to API

bryanculver:
# Closes: #DNE
# What's Changed

Introduce sorting of API responses. Needs tests but manual testing shows it's working.

# TODO

- [ ] Unit, Integration Tests


bryanculver:
In doing some manual testing there is still some fields that aren't sorting correctly. For example `name` on `Location` or nested attributes which `.order_by` should be able to accept. I suspect something is up with the field validation in the `OrderingFilter` implementation: https://github.com/encode/django-rest-framework/blob/e08e606c82afd0d5ec82b2c58badec11a4ce825e/rest_framework/filters.py#L245

HanlinMiao (approved):
LGTM!

nautobot/core/tests/test_api.py
```diff
@@ -488,3 +488,33 @@ def test_related_by_invalid(self):
...
+
+    @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
+    def test_descending_sort(self):
+        """Tests that results are returned in the expected descending order."""
+        response = self.client.get(f"{self.url}?ordering=-name&limit=10", **self.header)
+        self.assertHttpStatus(response, 200)
+        self.assertEqual(
+            list(map(lambda p: p["name"], response.data["results"])),
+            list(Provider.objects.order_by("-name").values_list("name", flat=True)[:10]),
+        )
```
glennmatthews, line 520: Would be good to see this as a generic test in `APIViewTestCases.ListObjectsViewTestCase` that we can apply to each list endpoint.

nautobot/core/api/filter_backends.py
```diff
@@ -22,6 +22,7 @@ def get_filterset_kwargs(self, request, queryset, view):
             "include",  # used to include computed fields (excluded by default)
             "limit",  # pagination
             "offset",  # pagination
+            "ordering",  # ordering of results
```
glennmatthews, line 25: Looks like the name of this param is configurable in settings.py - should we call it `sort` instead of `ordering` as the 1.x UI uses `sort`?

bryanculver: Or do we follow what the ORM calls it: `order_by`?

nautobot/core/tests/test_api.py
bryanculver: How would we identify which fields are sortable? In base we should at least test all direct attributes. I could see the need to add additional for FKs and their common sorting operation `name`.

glennmatthews: Does this filter support ordering by nested fields out-of-the-box? I didn't see anything to make me think so.

Could probably make a pretty decent first approximation by iterating over serializer.fields and skipping any that are RelatedField or SerializerMethodField instances.

bryanculver: I suspect so when I was digging into the code: https://github.com/encode/django-rest-framework/blob/e08e606c82afd0d5ec82b2c58badec11a4ce825e/rest_framework/filters.py#L255

nautobot/core/api/filter_backends.py
bryanculver: ☝️ Please see PR description

nautobot/core/tests/test_api.py
bryanculver: ☝️ Generic test base added, further discussion above in PR description

nautobot/core/tests/test_api.py
```diff
@@ -488,3 +488,33 @@ def test_related_by_invalid(self):
             response = self.client.post(url, data, format="json", **self.header)
         self.assertHttpStatus(response, status.HTTP_400_BAD_REQUEST)
         self.assertEqual(ipam_models.VLAN.objects.filter(name="Test VLAN 100").count(), 0)
+
+
+class APIOrderingTestCase(testing.APITestCase):
+    """
+    Testing integration with DRF's OrderingFilter.
```
glennmatthews, line 495: Is this test redundant with the generic tests in the above file now? Certainly seems very similar.

bryanculver: Yeah could probably safely delete it. I _might_ want to use it though to ensure coverage on different field types (datetime, text, int, etc) instead of exhaustively testing different models to ensure good support.

bryanculver:
The reason a few of the tests have commented-out tests are due to our `natural_key_field_lookups` being triggered by `OrderingFilter` introspecting class properties:

https://github.com/encode/django-rest-framework/blob/99e8b4033efa44930ace40fb48a4d7bcd224f9fb/rest_framework/filters.py#L229-L233

Calling

https://github.com/nautobot/nautobot/blob/45424c6b48f4fede83e213dc74a4f53722457b70/nautobot/core/models/__init__.py#L145-L151

To be fixed by #2908 and #2909

glennmatthews:
> The reason a few of the tests have commented-out tests are due to our `natural_key_field_lookups` being triggered by `OrderingFilter` introspecting class properties:
> 
> https://github.com/encode/django-rest-framework/blob/99e8b4033efa44930ace40fb48a4d7bcd224f9fb/rest_framework/filters.py#L229-L233
> 
> Calling
> 
> https://github.com/nautobot/nautobot/blob/45424c6b48f4fede83e213dc74a4f53722457b70/nautobot/core/models/__init__.py#L145-L151
> 
> To be fixed by #2908 and #2909

Hmm, starts to feel more and more like we might want to subclass or override the generic `OrderingFilter` to better suit our specific requirements?

bryanculver:
> 



> > The reason a few of the tests have commented-out tests are due to our `natural_key_field_lookups` being triggered by `OrderingFilter` introspecting class properties:
> > https://github.com/encode/django-rest-framework/blob/99e8b4033efa44930ace40fb48a4d7bcd224f9fb/rest_framework/filters.py#L229-L233
> > Calling
> > https://github.com/nautobot/nautobot/blob/45424c6b48f4fede83e213dc74a4f53722457b70/nautobot/core/models/__init__.py#L145-L151
> > 
> > To be fixed by #2908 and #2909
> 
> Hmm, starts to feel more and more like we might want to subclass or override the generic `OrderingFilter` to better suit our specific requirements?

Let's please keep the scope of this to a minimum. You can see the exceptions in the PR now and it's very seldom.

We will have to subclass for the other events but I don't think it should block being able to write the UI code to implement API-driven sort.

glennmatthews (approved):

HanlinMiao (approved):
💯

Issue #3711, opened 2023-05-08
UI 2.0: Plugin Provided Model View


bryanculver:

HanlinMiao:
Closed by #3721

PR #3721, opened 2023-05-09
Re-introduce app provided model view override

timizuoebideri1:
# Closes: #3711 
# What's Changed

Screenshot 2023-05-10 at 10 00 23 PM

The latest commit has the following:
1. `ExampleModel` uses the Nautobot Default `ObjectRetrieveView`.
Screen Shot 2023-05-12 at 3 24 33 PM
2. `AnotherExampleModel` uses a Customized `ModelView` from `example_plugin`.
Screen Shot 2023-05-12 at 3 24 44 PM
3. Finally, the `Location` model uses `OverridenView` provided by `example_plugin_with_view_override`.
Screen Shot 2023-05-12 at 3 24 23 PM

# TODO

- [ ] Unit, Integration Tests


bryanculver:
w8... wot?

What is the code in the retrieve to render the plugin's view instead of the core one? I would have thought we accidentally removed that.

Two points:

- Let's use `nautobot-ui` components in our example: https://github.com/nautobot/nautobot/pull/3721/files#diff-d4a3993cc893833a26c36f02b1e8774b12a594c8b1dbba604613fc98d4eb14b7R1
- Can you have it provide a view for its own model too?

examples/example_plugin/example_plugin/ui/index.js
```diff
@@ -1,11 +1,16 @@
...
 const app_config = {
     detail_tabs: {},
     full_width_components: {
-        "dcim:sites": ["ExampleAppFullWidthPageComponent"]
+        "dcim:devices": ["ExampleAppFullWidthPageComponent"]
     },
-    view_overrides: {}
+    view_overrides: {
+        "dcim:platforms": {
+            "retrieve": "ExamplePluginRetrieveViewOverride"
```
glennmatthews, line 11: We can't have `example_plugin` override any core views as this will break tests for the core view (and rightly so). This is what the `example_plugin_with_view_override` plugin is explicitly for, and that's where this override should be implemented.

timizuoebideri1:
> * Can you have it provide a view for its own model too?
Example Plugin has a model view

bryanculver (changes requested):
😵 OK so I think I understand how this is working now.

What I'd like to see is three examples when starting the app with the `example_plugin` is installed:

- A model provided by `example_plugin` that uses what we provide for it's detail view (what should be working today)
- A model providing its own view for its own model (I suppose this is working) but isn't technically an override. It just wants to define its own view.
- A model providing an overridden view for another model that is not its own (the thing we have to use `example_plugin_with_view_override` for as @glennmatthews stated above)

For the latter two cases they should have their own discrete dictionary items in the `index.js`
- Here are a list of my views for my models (maybe something like `views: {"example_plugin:example_models": Component}`)
- Here are a list of the views I want to override elsewhere (like the existing `view_overrides`)

development/nautobot_config.py
```diff
@@ -33,6 +33,7 @@
 
 PLUGINS = [
     "example_plugin",
+    "example_plugin_with_view_override",
```
glennmatthews, line 36: Shouldn't be enabled by default; again, it will break tests. It should only be enabled ad-hoc for demo purposes.

timizuoebideri1: Ooooh i understand now 😔

HanlinMiao:
The latest commit has the following:
1. `ExampleModel` uses the Nautobot Default `ObjectRetrieveView`.
Screen Shot 2023-05-12 at 3 24 33 PM
2. `AnotherExampleModel` uses a Customized `ModelView` from `example_plugin`.
Screen Shot 2023-05-12 at 3 24 44 PM
3. Finally, the `Location` model uses `OverridenView` provided by `example_plugin_with_view_override`

nautobot/ui/src/utils/utils.js
jathanism, line 1: Can we please try not to perpetuate the `utils.utils`/`utilities.utils` namespace problem here? Rename this to something like `core.js` or `general.js`, etc.?

jathanism (changes requested):
📖 📓 📚

examples/example_plugin/example_plugin/ui/FullWidthPage.js
```diff
@@ -0,0 +1,7 @@
+import { Box } from '@nautobot/nautobot-ui';
+
+export function MyFancyCard(props) {
+    return <Box body>I am a full-fledge component provided by the default example App. The ID of the object I am displaying is: {props.id}</Box>;
```
glennmatthews, line 4:
```suggestion
    return <Box body>I am a full-fledged component provided by the Example App. The ID of the object I am displaying is: {props.id}</Box>;
```

examples/example_plugin/example_plugin/ui/NonModelView.jsx
glennmatthews, line 1: Doesn't this belong in #3736 instead? Seems like there's a lot of duplicated code between the two PRs, can you clarify?

HanlinMiao:
Removed changes in `example_plugin_with_view_override`, since we already proved that we can override the Location Detail View from an app.

HanlinMiao:
> Code looks good to me. At some point we need to document this pattern in the app developer guide - is that in scope here?

I think this story only entails the re-introduction of the app overrides. I will open a follow-up issue that tracks this.

HanlinMiao:
> > Code looks good to me. At some point we need to document this pattern in the app developer guide - is that in scope here?
> 
> I think this story only entails the re-introduction of the app overrides. I will open a follow-up issue that tracks this.

https://github.com/nautobot/nautobot/issues/3773

jathanism (approved):

examples/example_plugin/example_plugin/ui/index.js
```diff
@@ -1,11 +1,16 @@
...
     full_width_components: {
-        "dcim:sites": ["ExampleAppFullWidthPageComponent"]
+        "dcim:devices": ["ExampleAppFullWidthPageComponent"]
     },
-    view_overrides: {}
+    view_overrides: {
+        "example-plugin:other-models": {
+            "retrieve": "ExampleAppOverrideModelView"
+        },
+    }
```
bryanculver, line 13: Can you just add one paragraph to our existing docs to this convention? Not perfect but _starts_ getting these conventions documented.

bryanculver (approved):

Issue #2569, opened 2022-10-04
Implement CSV export of model data using REST API


jathanism:
### Proposed Changes

The following changes are proposed to implement support for export (rendering) and import (parsing) of CSV data using the REST API:

#### Implementation

- We will use [`djangorestframework-csv`](https://github.com/mjumbewu/django-rest-framework-csv)
- The renderer `rest_framework_csv.renderers.PaginatedCSVRenderer` must be used 
    - This returns the current page of results without the `next`, `previous`, `count`
    - We will still need to solve how to get all objects vs. just the current page of objects for this file format.
- This enables any API view to render CSV by passing the `?format=csv` query parameter or setting `Accept: text/csv` in the request headers.

#### Nested object representations

- By default these are flattened and indexed with dotted notation
    - (e.g. `Status.content_types` gets exploded into `content_types.0`, `content_types.1`, 
    - This is not desirable. We want flat representations as a comma-separated string e.g.
        - `["dcim.device", "dcim.interface"]` becomes…
        - `"dcim.device,dcim.interface"` in a CSV field
- Bulk import done (using `bulk_create`)  and bulk update (using `bulk_update`) existing API endpoints and should “just work” but this needs to be tested.
    - This is done using the `rest_framework_csv.parsers.CSVParser` 
- Custom fields (being a nested representation) need to be accounted for
    - Current implementation is 1 header per e.g. `cf_{custom_field.slug}`

#### Models

- Remove `csv_headers` attribute
- Remove `to_csv()` method

#### Views

- Remove `queryset_to_csv()` method
- The correct API call needs to be accounted for to export ALL objects via CSV in keeping with current UI behaviors.

### Justification

This has been decided to be the path forward for eliminating bespoke import/export code in the Nautobot core in exchange for something closer to the way we're already doing import/export of objects using JSON via the REST API. Ideally this will require us to maintain little to no extra code to provide support for CSV inputs.

Additionally, because the REST API already supports bulk-create and bulk-update using the REST API (predicated on the serializer objects that know how to parse and render model data) these operations should also work relatively out of the box with little changes required, outside of those highlighted above in the proposed changes.

### Keep in Mind
- FK nested representations can/should stop at just exporting a PK and will be replaced with a natural key after #1574 is complete

lampwins:
> The renderer rest_framework_csv.renderers.PaginatedCSVRenderer must be used
>   - This returns the current page of results without the next, previous, count
[...]

Issue #3438, opened 2023-03-16
Create superuser through the API


gsnider2195:
### Discussed in https://github.com/nautobot/nautobot/discussions/3436

<div type='discussions-op-text'>

<sup>Originally posted by **rithikyennam** March 16, 2023</sup>
Is there a way to create a superuser using Nautobot API
I see there is no is_superuser flag in the API for creating users.

Is there any other way of creating nautobot superuser using a python script?

If  python script is not possible can you suggest a way to create it using ansible module?
We are using ansible module for deployment, using ansible it is allowing to create a super user but we cannot set a password for it </div>

Unable to set the `is_superuser` field through the API nor set the password on an existing user. Should either/both of these be supported?

gsnider2195:
`is_superuser` is already being tracked in #2208

glennmatthews:
Implemented for 2.0 by #3722.

Issue #3610, opened 2023-04-18
Cached Computed Fields


lampwins:
### As ...

Austin - Network Automation Engineer

### I want ...

A computed field that supported caching the rendered result

### So that ...

the value can be used in filtering, and recall of the value is very performant

### I know this is done when...

- I either have a new extensible field type, or an option on existing Computed Fields to cache their rendered value

For now it will be documented that the rendering and caching of the value will only occur on object create/update. A convience method/API endpoint will also allow users/admins to recalculate the value.

Such fields will then be automatically supported in filtering, API responses, table columns, etc.

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

Either a new field model, or changes to Computed Fields.

lampwins:
Obviously their are tradeoffs and downsides but with proper documentation and application, there do exist use cases where a value changes very infrequently, if at all, and thus a cached value that offers other benefits (filtering) would be nice.

Issue #2208, opened 2022-08-15
Return `is_superuser` and `last_login` fields for User API endpoints


joewesch:
### As ...

Austin - Network Automation Engineer

### I want ...

To be able to get the `is_superuser` and `last_login` fields for my User models via API.

### So that ...

I can create a report that will show which of our users are superusers in order to find new superusers that may have been created by other superusers without our knowledge. In addition, to be able to determine when a superuser hasn't logged in for a while and their permissions could be pruned.

### I know this is done when...

- I can call the User API endpoint and get back the values for the `is_superuser` and `last_login` fields.

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [X] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [X] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

I don't believe so as these fields already exist on the Django User model.

### External Dependencies

N/A

jrlee16:
additionally it would be nice to be able to enable/disable superuser status for user(s) via api to enable automation to enforce superuser access to a specified list of users only.

glennmatthews:
Resolved for 2.0 by #3500.

PR #3722, opened 2023-05-09
CSV export and import via REST API

### LINKED ISSUE #3715: Implement CSV import of model data using REST API
Building on #2569, allow CSV import to be based on REST API/serializer

glennmatthews:
# Closes: #2569

~This is not ready to merge at this time, as the flip-side of CSV import will obviously be broken by this changeset. Most likely this will be the target branch for #3715 as a follow-on PR before this can be merged down to `next`.~

Now includes changes from #3741, so this closes #3715 as well.

# What's Changed

- Implement `NautobotCSVRenderer` for DRF and wire it up to the generic REST API list and retrieve views.
  - This takes the queryset or object instance from the view, plus the associated serializer class, and renders it to CSV instead of JSON or HTML.
  - I didn't end up adding a dependency on `django-rest-framework-csv` as I found enough opinionated differences between that library and our specific requirements to where it was simpler to implement our own from scratch with that library as inspiration.
  - Most related objects are serialized as their natural keys
  - GenericForeignKeys are serialized as content-type + ~PK~ natural key.
  - M2M relations and tags are serialized as their natural-key-slugs (so that we don't have to deal with trying to represent a list of lists in a CSV column)
  - Custom field values are serialized as individual columns
- Export URLs now take the form `/api/app/model/?format=csv` instead of `/app/model/?export`; I've updated the legacy `export_button` template tag to handle this accordingly.
- Remove `csv_headers` and `to_csv` from all models
  - This does totally break the existing CSV import implementation since the CSVForm classes typically relied on `csv_headers` to define their fields. That will be addressed by ripping these forms out in #3715.
- Remove CSV export test from generic View test case and add `@skip` decorator to some related tests that will also need to be rethought in #3715.
- Add basic CSV export test to generic API test case
  - Revise a couple of serializer implementations to address issues I noticed while running this test.
- Remove `queryset_to_csv` from generic legacy views as well as from NautobotUIViewSet

## Example CSV

[nautobot_locations.csv](https://github.com/nautobot/nautobot/files/11436406/nautobot_locations.csv)

# TODO

- [x] Unit, Integration Tests
- [x] Maybe change the GenericForeignKey representation to content-type + natural-key-slug? This would require some extra DB lookups though I think.
- [ ] Add support for exporting computed field or relationship values?
- [x] Test handling of JSON fields (`User.user_config`, `ConfigContextSchema.schema`, JSON custom fields) as there are probably some bugs here
- [ ] Do some performance profiling - it "feels" noticeably slower to render the CSV compared to the speed of the JSON rendering.


nautobot/core/testing/api.py
```diff
@@ -386,6 +390,63 @@ def test_options_objects(self):
...
+            reader = csv.DictReader(StringIO(response_1.content.decode(response_1.charset)))
+            rows = list(reader)
+            # Should only have one entry (instance1) since we filtered out instance2 and permissions block instance3
+            self.assertEqual(1, len(rows))
+            self.assertEqual(rows[0]["id"], str(instance1.pk))
+            self.assertEqual(rows[0]["display"], getattr(instance1, "display", str(instance1)))
+            # TODO what other generic tests should we run on the data?
+
+            # Headers should match those constructed by the NautobotCSVRenderer
+            expected_headers = NautobotCSVRenderer.get_headers(get_serializer_for_model(self._get_queryset().model)())
```
glennmatthews, line 447:
```suggestion
            expected_headers = NautobotCSVRenderer.get_headers(get_serializer_for_model(self.model)())
```

nautobot/core/testing/api.py
```diff
@@ -386,6 +390,63 @@ def test_options_objects(self):
...
+            )
+
+            # Load the csv data back into a list of object dicts
+            reader = csv.DictReader(StringIO(response_1.content.decode(response_1.charset)))
+            rows = list(reader)
+            # Should only have one entry (instance1) since we filtered out instance2 and permissions block instance3
+            self.assertEqual(1, len(rows))
+            self.assertEqual(rows[0]["id"], str(instance1.pk))
+            self.assertEqual(rows[0]["display"], getattr(instance1, "display", str(instance1)))
+            # TODO what other generic tests should we run on the data?
```
glennmatthews, line 444: Would probably be straightforward to add a check for custom fields at the least.

nautobot/core/views/generic.py
```diff
@@ -157,6 +157,7 @@ def get_filter_params(self, request):
     def get_required_permission(self):
         return get_permission_for_model(self.queryset.model, "view")
 
+    # TODO: remove this as well?
```
glennmatthews, line 160: Maybe we need/want a NautobotYAMLRenderer too? Thought that might necessitate a hard fork from the device-types library repo.

nautobot/dcim/views.py
```diff
@@ -2393,56 +2392,7 @@ class CableBulkDeleteView(generic.BulkDeleteView):
 
 
 class ConnectionsListView(generic.ObjectListView):
```
glennmatthews, line 2394: The subclasses of this view had bespoke export implementations (below) - need to figure out whether we have equivalent REST endpoints that would provide equivalent data, or possibly add them. These also probably need to be reimplemented for the new UI in any case (out of scope for this PR).

nautobot/extras/models/secrets.py
```diff
@@ -144,6 +125,8 @@ class SecretsGroupAssociation(BaseModel):
     access_type = models.CharField(max_length=32, choices=SecretsGroupAccessTypeChoices)
     secret_type = models.CharField(max_length=32, choices=SecretsGroupSecretTypeChoices)
 
+    natural_key_field_names = ["secrets_group", "access_type", "secret_type", "secret"]
```
glennmatthews, line 128: Added because the default natural-key derived from `unique_together` didn't include the actual `secret` as a field.

nautobot/extras/templates/extras/jobresult.html
```diff
@@ -46,7 +46,9 @@
             </a>
         {% endif %}
     {% endif %}
-    {% export_button %}
+    <a href="{% url 'extras-api:joblogentry-list' %}?job_result={{ result.pk }}&format=csv" class="btn btn-success">
+        <span class="mdi mdi-database-export" aria-hidden="true"></span> Export
+    </a>
```
glennmatthews, line 51: The `export` button on the JobResult detail view doesn't export the JobResult itself; it exports the JobLogEntries associated with this result. Previously (see a few files below) this was done via a custom `export` handler on the JobResultView; now it's cleaner. :-)

nautobot/dcim/views.py
glennmatthews: Should be doable through the `/api/dcim/cables/` endpoint though it currently lacks a filter for "termination A type OR termination B type matches ...". Might have to add that.

nautobot/circuits/forms.py
```diff
@@ -65,7 +65,7 @@ class Meta:
 class ProviderCSVForm(CustomFieldModelCSVForm):
     class Meta:
         model = Provider
-        fields = Provider.csv_headers
+        fields = []  # TODO
```
HanlinMiao, line 68: These empty fields are for CSV import right?

nautobot/core/api/renderers.py
```diff
@@ -24,3 +36,168 @@ class NautobotJSONRenderer(JSONRenderer):
...
+                # Tricky since a SerializerMethodField can theoretically return anything.
+                value = serializer.data[key]
+                # However, in *many* cases in Nautobot core, it's just a wrapper around a GenericForeignKey field.
+                # So let's try that:
+                try:
+                    related_object = getattr(obj, key)
+                    if related_object is None:
+                        value = None
+                    elif isinstance(related_object, models.Model):
+                        value = [related_object._meta.label_lower] + self.get_natural_key(related_object)
```
HanlinMiao, line 159: Why do we need to wrap a `[]` around `related_object._meta.label_lower`?

nautobot/core/api/renderers.py
```diff
@@ -24,3 +36,168 @@ class NautobotJSONRenderer(JSONRenderer):
...
+                value = serializer.data[key]
+
+            # Coerce the value to a format to make the CSV renderer happy (i.e. a string)
+            if value is None:
+                value = ""
+            elif isinstance(value, dict):
+                if "id" in value and "object_type" in value:
+                    # A related object from a polymorphic serializer that we weren't able to catch above?
+                    # Serialize it as if it were a GenericForeignKey
+                    value = ",".join([value["object_type"], value["id"]])
```
HanlinMiao, line 173: Do we need to turn `value` into a `str` here and replace the comma with "%2c"?

nautobot/core/settings.py
```diff
@@ -188,6 +188,7 @@
     "DEFAULT_RENDERER_CLASSES": (
         "nautobot.core.api.renderers.NautobotJSONRenderer",
         "nautobot.core.api.renderers.FormlessBrowsableAPIRenderer",
+        "nautobot.core.api.renderers.NautobotCSVRenderer",
```
HanlinMiao, line 191: nit: Probably alpha sort here

nautobot/extras/api/serializers.py
```diff
@@ -795,23 +789,17 @@ class Meta:
...
         model = SecretsGroup
         fields = "__all__"
+        # TODO: it would be **awesome** if we could create/update SecretsGroupAssociations
+        # alongside creating/updating the base SecretsGroup, but since this is a ManyToManyField with
+        # a `through` table, that appears very non-trivial to implement. For now we have this as a
+        # read-only field; to create/update SecretsGroupAssociations you must make separate calls to the
+        # api/extras/secrets-group-associations/ REST endpoint as appropriate.
+        extra_kwargs = {
+            "secrets": {"source": "secrets_group_associations", "read_only": True},
+        }
```
HanlinMiao, line 802: WOO I didn't know you can do that. Can we do something similar for `DynamicGroupSerializer`'s `children` field? And what does the field look like now?

HanlinMiao (commented):
Some nitpick and questions but Sheeesh.
Screen Shot 2023-05-11 at 11 56 07 AM
Also would be helpful to see an example CSV file with some of the special case handling provided by `NautobotCSVRenderer`.

nautobot/core/api/renderers.py
```diff
@@ -24,3 +36,168 @@ class NautobotJSONRenderer(JSONRenderer):
...
+        return buffer.getvalue()
+
+    @classmethod
+    def get_headers(cls, serializer):
+        """Identify the appropriate CSV headers corresponding to the given serializer."""
+        base_headers = list(serializer.fields)
+
+        # Remove specific headers that we know are irrelevant
+        for undesired_header in [
+            "computed_fields",
```
bryanculver, line 95: I like the restraint of this for the first go-around but I'd like to consider opt-in ability of some of these fields (also noting #3610).

nautobot/core/api/renderers.py
```diff
@@ -24,3 +36,168 @@ class NautobotJSONRenderer(JSONRenderer):
...
+            "tree_depth",
+            "url",
+        ]:
+            if undesired_header in base_headers:
+                base_headers.remove(undesired_header)
+
+        # Remove all write-only and object-count headers
+        for header in list(base_headers):
+            if serializer.fields[header].write_only:
+                base_headers.remove(header)
```
bryanculver, line 110: Can we provide an example of write-only fields for those less aware such as myself?

nautobot/core/api/renderers.py
```diff
@@ -24,3 +36,168 @@ class NautobotJSONRenderer(JSONRenderer):
...
+            if header.endswith("_count"):
+                base_headers.remove(header)
```
bryanculver, line 112: I understand that things like this aren't round-trip-able, but I see value in exporting fields that are ignored on import. Maybe `ro_device_count` for `read only`? Not a blocker.

nautobot/core/api/renderers.py
```diff
@@ -24,3 +36,168 @@ class NautobotJSONRenderer(JSONRenderer):
...
+        # Add individual headers for each relevant custom field
+        if "custom_fields" in serializer.fields:
+            cf_keys = CustomField.objects.get_for_model(serializer.Meta.model).values_list("key", flat=True)
+            cf_headers = sorted([f"cf_{key}" for key in cf_keys])
+        else:
+            cf_headers = []
+
+        # TODO: relationships? computed fields?
+
+        headers = base_headers + cf_headers
```
bryanculver, line 123: How do we plan to import Custom Fields if it doesn't exist on import (or the name has changed?) I figure this is going to have to be addressed in #3715

bryanculver (commented):
👍 Don't know if this is going to become the PR for #2569 and #3715 but this looks good enough for export. I would suggest getting it merged into a `feature/` or `prototype/` branch so the import code is easier to review.

Regardless I think you're good to continue.

nautobot/circuits/forms.py
glennmatthews: Right, I'm *hoping* that the `*CSVForm` classes will be able to go away entirely in #3715. Haven't gotten there yet. :-)

nautobot/core/api/renderers.py
glennmatthews: The natural key is a list of values, so this is prepending the content-type to that list. I may need to adjust this as part of #3715 as I'm finding that mixing "natural keys as a comma-separated list" and "natural keys as a natural-key-slug" is challenging to handle consistently on the import side.

nautobot/core/settings.py
glennmatthews: This specifies the priority order of renderers, AIUI - I think putting NautobotCSVRenderer at the top would make it so that the default API response (if you don't specify a format or content-type) would become CSV, probably not desirable :-).

nautobot/core/api/renderers.py
glennmatthews: I can probably just let `config_context` and `local_config_context_data` be serialized as JSON (ugly though CSV-escaped JSON is!), as it seems to be working well enough for other JSON fields like `User.config_data`.

nautobot/core/api/renderers.py
glennmatthews: Yeah, I can probably re-enable the `count` fields if we think they might be desirable. They'll just be ignored on re-import, as they should be.

nautobot/core/api/renderers.py
glennmatthews: Given that the REST API normally silently ignores/discards any JSON keys that aren't understood and writable, I think it makes sense to do the same for custom fields that don't actually exist. In fact that might "just work".

nautobot/extras/api/serializers.py
glennmatthews: It **seems** to render the same way on output now. I'll take a look at DynamicGroupSerializer as well, thanks!

nautobot/extras/models/secrets.py
glennmatthews: We probably need an issue after #2908/#2909 to audit the auto-derived natural keys for all models and make sure they're as desired. Some of the auto-keys are pretty gnarly at present.

nautobot/ipam/api/serializers.py
```diff
@@ -229,23 +226,14 @@ class Meta:
...
-            obj,
-            nat_outside_list,
-            "nat_outside_list",
-        )
-        return data
+    def get_field_names(self, declared_fields, info):
+        """Add nat_outside_list reverse relation to the automatically discovered fields."""
+        fields = list(super().get_field_names(declared_fields, info))
+        self.extend_field_names(fields, "nat_outside_list")
+        return fields
```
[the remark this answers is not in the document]
glennmatthews: Might be a tricky one to lint around. This change works for MtoM (like `SecretsGroup.secrets_group_associations` above) and for Mto1 reverse relations (like `IPAddress.nat_outside_list` here)because DRF sees them as serializable fields, it just doesn't include such fields in `__all__` by default.

For GFKs and such I think we'll continue to need to handle them separately.

nautobot/core/views/generic.py
jathanism: I'm thinking yes. For what it's worth there's `djangorestframework-yaml` from the same author (whose work you didn't end up using), of note specifically is the encoder class: https://github.com/jpadilla/django-rest-framework-yaml/blob/master/rest_framework_yaml/encoders.py

nautobot/dcim/api/views.py
```diff
@@ -622,22 +623,23 @@ class InventoryItemViewSet(NautobotModelViewSet):
 
 #
 # Connections
+# TODO: remove these in favor of using the ConsolePort/PowerPort/Interface API endpoints and/or Cable endpoint.
 #
 
 
-class ConsoleConnectionViewSet(ListModelMixin, GenericViewSet):
+class ConsoleConnectionViewSet(ModelViewSetMixin, ListModelMixin, GenericViewSet):
```
jathanism, line 630: I *think* you can probably get rid of `GenericViewSet` since you're adding `ModelViewSetMixin`.

jathanism (commented):
This is so great to see! I have a lot to say as I'm wondering somewhat about if we can reuse a lot of the same code for other renderers (such as YAML).

nautobot/core/api/renderers.py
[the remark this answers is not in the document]
glennmatthews: Removed in latest diffs, no reason not to export count fields as they'll just be ignored on import.

nautobot/core/api/renderers.py
```diff
@@ -24,3 +36,168 @@ class NautobotJSONRenderer(JSONRenderer):
...
+                    construct_natural_key_slug(self.get_natural_key(related_object))
+                    for related_object in related_queryset.all()
+                ]
+            elif isinstance(serializer_field, serializers.Serializer):
+                # A related object; we want to render its natural key if at all possible.
+                related_object = serializer_field.get_attribute(obj)
+                value = self.get_natural_key(related_object)
+            elif isinstance(serializer_field, serializers.SerializerMethodField):
+                # Tricky since a SerializerMethodField can theoretically return anything.
+                value = serializer.data[key]
```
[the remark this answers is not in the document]
glennmatthews: Removed some of this code in latest diffs. Still a valid point/concern.

I'd love for us to come up with a custom field class (or perhaps factory) to handle our GenericForeignKey cases in a more elegant way than individual bespoke SerializerMethodFields - I spent an hour or two this afternoon trying to prototype something here but was ultimately unsuccessful.

nautobot/core/api/views.py
```diff
@@ -263,6 +263,32 @@ def dispatch(self, request, *args, **kwargs):
...
+        if "text/csv" in request.accepted_media_type:
+            instance = self.get_object()
+            serializer_class = self.get_serializer_class()
+            filename = f"{settings.BRANDING_PREPENDED_FILENAME}{instance._meta.verbose_name}.csv"
+            return Response(
+                data={"instance": instance, "serializer_class": serializer_class},
+                headers={"Content-Disposition": f"attachment; filename={filename}"},
+            )
+
+        return super().retrieve(request, *args, format=format, **kwargs)
```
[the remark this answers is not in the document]
glennmatthews: The main reason here is actually so that I can pass the instance/queryset and the serializer_class to the renderer. By default a renderer only gets the `data` dict derived from the serializer.

It's possible that it might be doable to revise this pattern, as the biggest reason for this was to be able to convert foreign keys and the like from PK/URL strings to natural-keys - since in the latest diffs I've added `natural_key_slug` as a field on BaseModelSerializer, we **might** be able to just get everything from the serializer data at depth 1?

nautobot/dcim/api/views.py
glennmatthews: Thanks for asking - I checked and our `ModelViewSetMixin` is a pure mixin class, no parent class. I added it to these viewsets because it's what provides the custom `list` override to render to CSV; if we can get rid of that override (as mentioned above) then these can get rid of the mixin again.

nautobot/extras/api/serializers.py
HanlinMiao: Screen Shot 2023-05-11 at 4 18 08 PM
This is the culprit I guess.

nautobot/core/api/views.py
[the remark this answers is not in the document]
glennmatthews: Oh right, it's also to bypass the pagination that the default `list()` implementation will enforce.

nautobot/extras/api/serializers.py
glennmatthews: Weird, that just makes me wonder why in my case it's *not* using the DG URL?

nautobot/core/api/views.py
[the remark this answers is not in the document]
glennmatthews: Looks like maybe I can/should override our `paginate_queryset()` implementation instead, will do some further digging.

nautobot/extras/api/serializers.py
HanlinMiao: I think if you leave it up to DRF to render that field, it will use `DGMembershipSerializer` from `BaseModelSerializer` which in turn uses `NautobotHyperlinkedRelatedField`. And `NautobotHyperlinkedRelatedField` uses `get_route_for_model` instead of `get_absolute_url` from `DynamicGroupMemebrship` model class. And That gives us the correct URL.

glennmatthews:
Updated csv:

[locations.csv](https://github.com/nautobot/nautobot/files/11456550/locations.csv)

Note the addition of a `natural_key_slug` column, as well as the use of natural-key-slugs for all foreign key fields (status, parent, location_type, tenant) and for tags

glennmatthews:
One issue with the latest implementation is that since we're no longer overriding the `list` and `retrieve` API endpoints, we don't currently have a way to specify the CSV filename, and so by default it gets a random name assigned by the browser like `WcUkQR3Y.csv`. This is a bit ugly, it would be good to find a fix for this.

nautobot/core/views/generic.py
glennmatthews: That looks like a larger body of work, so I've opened #3772.

nautobot/extras/models/secrets.py
glennmatthews: I opened #3752 for this work.

nautobot/extras/api/serializers.py
```diff
@@ -795,23 +789,17 @@ class Meta:
...
-        )
-
     class Meta:
         model = SecretsGroup
         fields = "__all__"
+        # TODO: it would be **awesome** if we could create/update SecretsGroupAssociations
+        # alongside creating/updating the base SecretsGroup, but since this is a ManyToManyField with
+        # a `through` table, that appears very non-trivial to implement. For now we have this as a
+        # read-only field; to create/update SecretsGroupAssociations you must make separate calls to the
+        # api/extras/secrets-group-associations/ REST endpoint as appropriate.
```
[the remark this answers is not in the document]
glennmatthews: Relates to #3772 actually...

nautobot/core/api/renderers.py
glennmatthews: Addressed - `cf_foobar` will be ignored if no `foobar` custom field exists, similar to how the REST API ignores any other fields/columns that it doesn't understand or are read-only.

jathanism (approved):

changes/3715.fixed
```diff
@@ -0,0 +1,4 @@
+Fixed inability to set the `failover_strategy` field as unspecified/blank when creating/updating `DeviceRedundancyGroup` via the REST API.
+Fixed incorrect interim natural key definition for `Device` and `VRF` models.
+Fixed the possibility of setting values for nonexistent custom fields via the REST API.
+Fixed incorrect password-handling logic when creating/updating a `User` via the REST API.
```
gsnider2195, line 4: Does this resolve #3438

glennmatthews: Good catch! I think it does. Mind taking a look at the tests I've added in `users/test_api.py` and let me know if anything is missing there?

gsnider2195: I verified in your branch that:
- Able to create a user with `is_superuser` set to True or False, and `is_staff` set to True or False
- Able to modify an existing user password
This should resolve #3438

Also #2208 should be resolved by this PR because:
- GET http://localhost:8080/api/users/users/{id}/ returns `is_superuser` and `last_login` fields

nautobot/users/tests/test_api.py
```diff
@@ -39,9 +39,9 @@ class UserTest(APIViewTestCases.APIViewTestCase):
         },
         {
             "username": "User_6",
-            "password": "password6",
         },
     ]
+    update_data = {"username": "User_22"}
```
gsnider2195, line 44: Do we need to add the rest of the user fields (`is_superuser`, `last_login`, `password`, etc.) to the `create_data`/`update_data` properties on this test?

changes/3715.fixed
glennmatthews: 3500 gets the credit for resolving #2208 by switching the UserSerializer fields to `__all__`.  :-) Thanks for checking these!

nautobot/users/tests/test_api.py
glennmatthews: I don't think it's critical, since with the new round-trip csv test case we're exercising every field that declares itself as writable for the `create` case, but it would be good to add just to assert what fields we *expect* to be writable. I'll try to remember to tackle this on Monday - it does fall into a broader category of expanded API test data that's probably needed/desirable after #3500 expanded so many of the serializers' fields.

Issue #3712, opened 2023-05-08
UI 2.0: App Provided Non-Model View


bryanculver:
Unline #3711 where we already have the route for model-based views, this is for non-model-based views an App would like to register with the new UI.

Take: https://github.com/nautobot/nautobot/commit/2f99d676cf7e83c81bed5947096270556ed0ed6a#diff-8a0f49e806240486e1ae00f66455d102918cd81fe273de9642bcd346ffa61875L10-L14 further and register a route to be merged into here: https://github.com/nautobot/nautobot/blob/73ef56e3b4fdc2946cfb1de7fdea33f1b3600849/nautobot/ui/src/router.js#L44-L50 namespaced with the app name. We already template out pieces of the UI for imports so this will be the next extension of that pattern (see: https://github.com/nautobot/nautobot/blob/73ef56e3b4fdc2946cfb1de7fdea33f1b3600849/nautobot/ui/app_imports.js.j2#L2-L4)

PR #3736, opened 2023-05-11
App Provided Non Model View -View Without an associated API

timizuoebideri1:
# Closes: #3712 
# What's Changed

Screenshot 2023-05-11 at 8 52 04 PM

- Developed the '`utils/nav/get_routes()`' function to retrieve declared plugin routes.
- '`useGetUIMenuQuery`' has been updated to now take into account the plugin routes created by 'get_routes()'
- Made and added a View Component to the '`example_plugin_with_view_override`' plugin.
- Changed the ObjectList component to load either the `GenericObjectListView` or a plugin component view depending on specific criteria's

Limitations with this approach:

- Custom routes outside of these two are not feasible since the app depends on the dynamic routes made on `nautobot-ui` to load custom plugin provided views. Plugin developers can only construct the routes `/<app_name>/ `and `/<app_name>/<model>/`. Where app_name and model could be anything.

In other PR

- [ ] View with an associated API

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/examples/example_plugin_with_view_override/example_plugin_with_view_override/ui/CustomView.jsx
+++ b/examples/example_plugin_with_view_override/example_plugin_with_view_override/ui/CustomView.jsx
@@ -1,30 +1,35 @@
-import Table from 'react-bootstrap/Table';
+import { Table, Box, TableContainer, Tbody, Thead, Tr, Th, Td } from '@nautobot/nautobot-ui';
+import GenericView from "@views/generic/GenericView";
 
-export function CustomTableView(props) {
+export default function CustomView(props) {
   return (
-    <Table striped bordered hover>
-      <thead>
-        <tr>
-          <th>Key</th>
-          <th>Value</th>
-        </tr>
-      </thead>
-      <tbody>
-        <tr>
-          <td>ID</td>
-          <td>{props.id}</td>
-        </tr>
-        <tr>
-          <td>Display</td>
-          <td>{props.display}</td>
-        </tr>
-        <tr>
-          <td>URL</td>
-          <td>{props.url}</td>
-        </tr>
-      </tbody>
-    </Table>
+  <GenericView>
+    <Box background="white-0" borderRadius="md" padding="md">
+      <TableContainer>
+        <Table>
+              <Thead>
+                <Th>Key</Th>
+                <Th>Value</Th>
+              </Thead>
+            <Tbody>
+              <Tr>
+                <Td>ID</Td>
+                <Td>{props.id}</Td>
+              </Tr>
+              <Tr>
+                <Td>Display</Td>
+                <Td>{props.display}</Td>
+              </Tr>
+              <Tr>
+                <Td>URL</Td>
+                <Td>{props.url}</Td>
+              </Tr>
+            </Tbody>
+          </Table>
+      </TableContainer>
+
+    </Box>
+
+    </GenericView>
   );
 }
-
-export default CustomTableView;
--- a/examples/example_plugin_with_view_override/example_plugin_with_view_override/ui/FullWidthPage.jsx
+++ b/examples/example_plugin_with_view_override/example_plugin_with_view_override/ui/FullWidthPage.jsx
@@ -1,7 +1,9 @@
-import Card from 'react-bootstrap/Card';
+import { Box } from '@nautobot/nautobot-ui';
 
 export function MyFancyCard() {
-  return <Card body>I am from another plugin!</Card>;
+  return (
+    <Box body>I am from another plugin!</Box>
+  );
 }
 
 export default MyFancyCard;
--- a/nautobot/ui/src/components/Apps.js
+++ b/nautobot/ui/src/components/Apps.js
@@ -1,7 +1,6 @@
 import { lazy } from "react";
 
 import NautobotApps from "../app_imports";
-import { convertPluginMenuIntoNavMenuFormat } from "@utils/nav"
 import { slugify } from "@utils/string";
 
 function EmptyElement() {
@@ -22,6 +21,7 @@ function get_components() {
     var base = {};
     base["FullWidthComponents"] = {};
     base["CustomViews"] = {};
+    base["Views"] = {};
 
     for (const [app_name, import_promise] of Object.entries(NautobotApps)) {
         import_promise.then((value) => {
@@ -65,31 +65,20 @@ function get_components() {
                     }
                 );
             }
+            // Add Plugin Provided View Components
+            if (value?.default?.routes) {
+                // eslint-disable-next-line
+                value?.default?.routes.forEach(({ namespace, component }) => {
+                    const route = `${slugify(app_name)}:${namespace}`;
+                    base["Views"][route] = lazy(() =>
+                        my_import_as_function(app_name, component)
+                    );
+                });
+            }
         });
     }
 
     return base;
 }
 
-
-
-
-function get_routes(){
-    let data = {}
-    for (const [app_name, import_promise] of Object.entries(NautobotApps)) {
-        // eslint-disable-next-line no-loop-func
-        import_promise.then((value) => {
-            const routes = value?.default?.routes
-            if (routes) {
-                data = convertPluginMenuIntoNavMenuFormat(routes, slugify(app_name), data)
-            }
-        })
-    }
-    return data;
-
-}
-const pluginRoutes = get_routes()
-
-
 export default get_components();
-export { pluginRoutes }
--- a/nautobot/ui/src/components/Apps.js
+++ b/nautobot/ui/src/components/Apps.js
@@ -1,6 +1,8 @@
 import { lazy } from "react";
 
 import NautobotApps from "../app_imports";
+import { convertPluginMenuIntoNavMenuFormat } from "@utils/nav"
+import { slugify } from "@utils/string";
 
 function EmptyElement() {
     return <></>;
@@ -69,4 +71,25 @@ function get_components() {
     return base;
 }
 
+
+
+
+function get_routes(){
+    let data = {}
+    for (const [app_name, import_promise] of Object.entries(NautobotApps)) {
+        // eslint-disable-next-line no-loop-func
+        import_promise.then((value) => {
+            const routes = value?.default?.routes
+            if (routes) {
+                data = convertPluginMenuIntoNavMenuFormat(routes, slugify(app_name), data)
+            }
+        })
+    }
+    return data;
+
+}
+const pluginRoutes = get_routes()
+
+
 export default get_components();
+export { pluginRoutes }
--- a/nautobot/ui/src/utils/api.js
+++ b/nautobot/ui/src/utils/api.js
@@ -6,6 +6,7 @@ import {
     API_USER_SESSION_INFO,
     API_UI_MENU_INFO,
 } from "@constants/apiPath";
+import { updateMenuitemsWithPluginMenu } from "./nav";
 
 /*
   The one true API!
@@ -16,6 +17,22 @@ import {
   in React and Redux's sites.
 */
 
+/**
+ * 
+ * getUIMenu: builder.query({
+    query: async () => {
+        const response = await staggeredBaseQuery.fetch(API_UI_MENU_INFO);
+        const data = response.data;
+        // Add custom data to the returned data
+        const updatedData = {
+            ...data,
+            customData: "your custom data here",
+        };
+        return updatedData;
+    },
+    providesTags: ["AppData"],
+}),
+ */
 const staggeredBaseQuery = retry(fetchBaseQuery({ baseUrl: API_BASE }), {
     maxRetries: 5,
 });
@@ -35,6 +52,9 @@ export const baseApi = createApi({
         getUIMenu: builder.query({
             query: () => API_UI_MENU_INFO,
             providesTags: ["AppData"],
+            // Update API return Menu with menu from menu data from plugins
+            transformResponse: (response) =>
+                updateMenuitemsWithPluginMenu(response),
         }),
         getObjectCounts: builder.query({
             query: () => API_OBJECT_COUNTS,
--- a/nautobot/ui/src/utils/string.js
+++ b/nautobot/ui/src/utils/string.js
@@ -4,3 +4,23 @@ export function toTitleCase(str, separator = " ") {
         .map((x) => (x ? x[0].toUpperCase() + x.slice(1) : ""))
         .join(" ");
 }
+
+export function slugify(str) {
+    let data = str.replace(/^\s+|\s+$/g, ""); // trim
+    data = data.toLowerCase();
+
+    // Remove accents, swap ñ for n, etc
+    var from = "àáäâèéëêìíïîòóöôùúüûñç·/_,:;";
+    var to = "aaaaeeeeiiiioooouuuunc------";
+    for (var i = 0, l = from.length; i < l; i++) {
+        data = data.replace(new RegExp(from.charAt(i), "g"), to.charAt(i));
+    }
+
+    data = data
+        .replace(/[^a-z0-9 -]/g, "") // remove invalid chars
+        .replace(/\s+/g, "-") // collapse whitespace and replace by -
+        .replace(/-+/g, "-") // collapse dashes
+        .replace(/_+/g, "-"); // replace underscores with hyphens
+
+    return data;
+}
--- a/development/nautobot_config.py
+++ b/development/nautobot_config.py
@@ -33,6 +33,7 @@ LOGGING["loggers"]["nautobot"]["level"] = LOG_LEVEL  # noqa: F405
 
 PLUGINS = [
     "example_plugin",
+    "example_plugin_with_view_override",
 ]
 
 CORS_ALLOWED_ORIGINS = ["http://localhost:3000"]
--- a/examples/example_plugin_with_view_override/example_plugin_with_view_override/ui/NonModelView.jsx
+++ b/examples/example_plugin_with_view_override/example_plugin_with_view_override/ui/NonModelView.jsx
@@ -1,0 +1,10 @@
+
+
+
+export default function NonModelView(){
+    return (
+        <div>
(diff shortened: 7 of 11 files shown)


examples/example_plugin/example_plugin/ui/NonModelView.jsx
```diff
@@ -0,0 +1,10 @@
+import GenericView from "@views/generic/GenericView";
+
+
+export default function NonModelView(){
+    return (
+        <GenericView gridBackground="white">
+            <div> Im a humble 🙈 Non Model Plugin View.</div>
+        </GenericView>
+    )
```
glennmatthews, line 9: "Non-Model view" is not super self-explanatory. Perhaps something like "client-side-only view" might be clearer?

examples/example_plugin/example_plugin/ui/NonModelView.jsx
```diff
@@ -0,0 +1,10 @@
+import GenericView from "@views/generic/GenericView";
+
+
+export default function NonModelView(){
+    return (
+        <GenericView gridBackground="white">
+            <div> Im a humble 🙈 Non Model Plugin View.</div>
```
glennmatthews, line 7: Let's use `app` instead of `plugin` for anything user-facing at this point please.

examples/example_plugin/example_plugin/ui/index.js
```diff
@@ -1,11 +1,32 @@
...
+                {
+                    name: "Example App",
+                    weight: 150,
+                    items: [
+                        {
+                            name: "Non Model View",
+                            weight: 100,
+                            path: "/non-model-view/",
+                            component: "ExamplePluginNonModelView",
+                            namespace: "non-model-view",
```
glennmatthews, line 23: Please add some explanatory comments where appropriate, at least:

- is the `path` absolute, or relative to the `plugins/example-plugin/` base path? (Do we support both?)
- why is `component` a string instead of an object reference?
- what purpose does `namespace` serve?

examples/example_plugin_with_view_override/example_plugin_with_view_override/ui/CustomView.jsx
```diff
@@ -1,30 +1,35 @@
-import Table from 'react-bootstrap/Table';
+import { Table, Box, TableContainer, Tbody, Thead, Tr, Th, Td } from '@nautobot/nautobot-ui';
```
glennmatthews, line 1: Are the changes in `example_plugin_with_view_override` unrelated to this PR?

nautobot/ui/src/components/Apps.js
```diff
@@ -63,6 +53,20 @@ function get_components() {
                     }
                 );
             }
+            // Add Plugin Provided View Components
+            if (value?.default?.routes) {
+                // eslint-disable-next-line
```
glennmatthews, line 58: why?

nautobot/ui/src/components/Apps.js
```diff
@@ -63,6 +53,20 @@ function get_components() {
...
+                value?.default?.routes.forEach(({ groups }) => {
+                    groups.forEach(({ items }) => {
+                        items.forEach(({ namespace, component }) => {
+                            const route = `${slugify(app_name)}:${namespace}`;
```
glennmatthews, line 62: I'm not a fan of using `slugify(app_name)` here - I'd prefer that the route be explicitly specified in some other fashion rather than assuming that a given app's routes follow a particular arbitrary pattern.

nautobot/ui/src/components/Apps.js
```diff
@@ -63,6 +53,20 @@ function get_components() {
...
+                            base["OverrideViews"][route] = lazy(() =>
```
glennmatthews, line 63: Why `OverrideViews`? This PR doesn't (AIUI) have anything to do with overriding base views, it's about adding entirely new views right?

nautobot/ui/src/utils/string.js
```diff
@@ -4,3 +4,23 @@ export function toTitleCase(str, separator = " ") {
...
+    var to = "aaaaeeeeiiiioooouuuunc------";
+    for (var i = 0, l = from.length; i < l; i++) {
+        data = data.replace(new RegExp(from.charAt(i), "g"), to.charAt(i));
+    }
+
+    data = data
+        .replace(/[^a-z0-9 -]/g, "") // remove invalid chars
+        .replace(/\s+/g, "-") // collapse whitespace and replace by -
+        .replace(/-+/g, "-") // collapse dashes
+        .replace(/_+/g, "-"); // replace underscores with hyphens
```
glennmatthews, line 23: Ok yeah, I'm definitely not a fan of this function. It's both overly complex and also overly naive - there's all sorts of non-ASCII characters that it doesn't handle, it feels really inefficient (*how* many regexp replaces are performed each time this function is invoked?), and again I think it relies on assumptions about app route names that aren't necessarily going to be true in all cases.

examples/example_plugin_with_view_override/example_plugin_with_view_override/ui/CustomView.jsx
timizuoebideri1: Yup unrelated, Came across them, and decided to fix them

nautobot/core/management/commands/build_ui.py
```diff
@@ -37,18 +41,85 @@ def add_arguments(self, parser):
...
+                    path_regex = urlpattern.pattern._compile(urlpattern.pattern._route)
+                else:
+                    path_regex = urlpattern.pattern._compile(urlpattern.pattern._regex)
+                url_path = self.convert_django_url_regex_to_react_route_path(path_regex)
+                data.append({ "path": base_url + url_path, "component": component })
+        return data
+
     def render_app_imports(self):
-        """Render `app_imports.js` and update `jsconfig.json` to map to the path for each."""
+        """Render `app_imports.js`, update `jsconfig.json` to map to the path for each and update `urls.json` to match url imports."""
```
timizuoebideri1, line 102:
```suggestion
        """Render `app_imports.js`, update `jsconfig.json` to map to the path for each and render `app_routes.json` to register all app routes."""
```

nautobot/ui/src/utils/api.js
```diff
@@ -7,13 +7,13 @@ import {
     API_UI_MENU_INFO,
 } from "@constants/apiPath";
 
-/*
-  The one true API!
+/*	
+  The one true API!	
 
-  Using Redux's RTK Query making fetching and caching a breeze! This will provide hooks and wires automagically.
+  Using Redux's RTK Query making fetching and caching a breeze! This will provide hooks and wires automagically.
```
timizuoebideri1, line 13:
```suggestion
  Using Redux's RTK Query making fetching and caching a breeze! This will provide hooks and wires automagically.
```

nautobot/ui/src/utils/api.js
```diff
@@ -7,13 +7,13 @@ import {
...
+/*	
+  The one true API!	
 
-  Using Redux's RTK Query making fetching and caching a breeze! This will provide hooks and wires automagically.
+  Using Redux's RTK Query making fetching and caching a breeze! This will provide hooks and wires automagically.	
 
-  A standardized convention for retrieving data which should make developer's lives easier and is already extensively documented
-  in React and Redux's sites.
+  A standardized convention for retrieving data which should make developer's lives easier and is already extensively documented	
+  in React and Redux's sites.
```
timizuoebideri1, line 16:
```suggestion
  in React and Redux's sites.
```

examples/example_plugin_with_view_override/example_plugin_with_view_override/ui/CustomView.jsx
glennmatthews: Aren't these in conflict with the changes being made in #3721? I don't understand how these two PRs relate to one another.

examples/example_plugin/example_plugin/ui/ModelRetrieveView.jsx
glennmatthews, line 1: This should be a .js file instead of .jsx correct? Isn't the standard we decided on? 

Same question for the other files in this PR.

examples/example_plugin/example_plugin/ui/ModelRetrieveView.jsx
```diff
@@ -0,0 +1,9 @@
+import { useParams } from "react-router-dom";
+
+
+export default function ModelRetrieveView(props){
+    const { pk } = useParams();
+    return (
+            <div>Im a App provided route view with pk: { pk }.</div>
```
glennmatthews, line 7: Please fix spelling/grammar as well as making this a bit more self-documenting as to what this is demonstrating.

examples/example_plugin/example_plugin/ui/index.js
```diff
@@ -1,11 +1,15 @@
...
 
 const app_config = {
     detail_tabs: {},
     full_width_components: {
         "dcim:sites": ["ExampleAppFullWidthPageComponent"]
     },
-    view_overrides: {}
+    view_overrides: {},
+    routes_view_components: {
+        "examplemodel": "ExampleAppModelRetrieveView",
```
glennmatthews, line 11: Can you please document how this is used and what its structure is?

It seems weird to me that this is still associated to the `examplemodel` given the issue this is addressing - maybe a better example might be to use this to demonstrate how the example app can provide a "dashboard/home" view or a "config" view, either of which is entirely unrelated to the DB models?

examples/example_plugin_with_view_override/example_plugin_with_view_override/ui/FullWidthPage.jsx
```diff
@@ -1,7 +1,9 @@
-import Card from 'react-bootstrap/Card';
+import { Box } from '@nautobot/nautobot-ui';
 
 export function MyFancyCard() {
-  return <Card body>I am from another plugin!</Card>;
+  return (
+    <Box body>I am from another plugin!</Box>
```
glennmatthews, line 5:
```suggestion
    <Box body>I am from another app!</Box>
```

nautobot/core/management/commands/build_ui.py
```diff
@@ -37,19 +40,78 @@ def add_arguments(self, parser):
...
+        """
+        Extracts the view component associated with the specified route name from the App index.js file.
+
+        Parameters:
+            file_path (str): The path to the JavaScript file to read from.
+            route_name (str): The name of the route to extract the view component for.
+
+        Returns:
+            str: The view component associated with the specified route name, or None if not found.
+        """
```
glennmatthews, line 69: Example please?

nautobot/core/management/commands/build_ui.py
```diff
@@ -37,19 +40,78 @@ def add_arguments(self, parser):
...
+        with open(file_path, "r") as f:
+            js_content = f.read()
+
+        # Construct a regular expression that matches the specified key and extracts the associated view component
+        pattern = rf'routes_view_components:\s*{{[^}}]*"{re.escape(route_name)}"\s*:\s*"([^"]+)"'
+        view_component_match = re.search(pattern, js_content)
+        return view_component_match[1] if view_component_match else None
+
+    def render_routes_imports(self, app_base_path, app_name, app_config):
```
glennmatthews, line 78: Docstring please.

nautobot/core/management/commands/build_ui.py
```diff
@@ -37,19 +40,78 @@ def add_arguments(self, parser):
...
+        data = []
+        module = __import__(f"{app_name}.urls", {}, {}, [""])
```
glennmatthews, line 80: Maybe use `importlib.import_module` instead of `__import__`?

nautobot/core/management/commands/build_ui.py
```diff
@@ -37,19 +40,78 @@ def add_arguments(self, parser):
...
+        base_url = app_config.base_url or app_config.label
+        for urlpattern in module.urlpatterns:
+            if component := self.get_app_component(
+                app_base_path / "ui/index.js",
+                urlpattern.name,
+            ):
+                if isinstance(urlpattern.pattern, RoutePattern):
+                    path_regex = urlpattern.pattern._compile(urlpattern.pattern._route)
+                else:
+                    path_regex = urlpattern.pattern._compile(urlpattern.pattern._regex)
```
glennmatthews, line 90: So many private attribute accesses... no cleaner way to do this?

nautobot/core/management/commands/build_ui.py
```diff
@@ -73,6 +136,9 @@ def render_app_imports(self):
         with open(jsconfig_file_path, "w", encoding="utf-8") as generated_config_file:
             json.dump(jsconfig, generated_config_file, indent=4)
 
+        with open(app_routes_file_path, "w", encoding="utf-8") as app_rotes_file:
+            json.dump(app_routes, app_rotes_file, indent=4)
```
glennmatthews, line 140:
```suggestion
        with open(app_routes_file_path, "w", encoding="utf-8") as app_routes_file:
            json.dump(app_routes, app_routes_file, indent=4)
```

nautobot/ui/src/utils/navigation.js
```diff
@@ -0,0 +1,33 @@
...
+import { LoadingWidget } from "@components/LoadingWidget";
+
+export function getPluginRoutes() {
+    let react_routes = [];
+    let app_routes = [];
+    try {
+        app_routes = require("../app_routes.json");
+    } catch (error) {
+        console.warn("app_routes.json not found");
+    }
```
timizuoebideri1, line 13: use of a try-and-catch method due to test-ui failures

bryanculver: We should probably have an empty file or address this directly. The example plugin is loaded for tests so we shouldn't ignore this workflow.

.github/workflows/ci_pullrequest.yml
```diff
@@ -292,6 +292,9 @@ jobs:
       - name: "Setup base app_imports for mock"
         working-directory: ./nautobot/ui
         run: "echo 'export const NautobotApps = {}; export default NautobotApps;' > src/app_imports.js"
+      - name: "Setup base app_routes.json for mtestock"
```
bryanculver, line 295:
```suggestion
      - name: "Setup base app_routes.json for test"
```

examples/example_plugin/example_plugin/ui/index.js
```diff
@@ -1,11 +1,19 @@
...
     detail_tabs: {},
     full_width_components: {
         "dcim:devices": ["ExampleAppFullWidthPageComponent"]
     },
+    /**
+     * The key of `routes_view_components` is the name of the url path,
+     * and the value is the View Component that should be associated with that url path.
+     */
+    routes_view_components: {
+        "home": "ExampleAppCustomRouteView",
```
bryanculver, line 15: For clarification: This ultimately becomes `example_plugin:home`, correct?

bryanculver: This route: https://github.com/nautobot/nautobot/blob/develop/examples/example_plugin/example_plugin/urls.py#L19 ?

bryanculver: OK, let's clear up some conventions here. We include their own app label for overridden views but not for this. For now let's include it here, even if it's redundant.

.github/workflows/ci_pullrequest.yml
```diff
@@ -292,6 +292,9 @@ jobs:
       - name: "Setup base app_imports for mock"
         working-directory: ./nautobot/ui
         run: "echo 'export const NautobotApps = {}; export default NautobotApps;' > src/app_imports.js"
+      - name: "Setup base app_routes.json for test"
+        working-directory: ./nautobot/ui
+        run: "echo '{}' > src/app_routes.json"
```
glennmatthews, line 297: Are these steps that a developer needs to do before running `invoke unittest-ui` locally? If so let's make sure we have that well documented.

Are these steps that a user needs to do when installing Nautobot from pypi? Same concern.

glennmatthews (approved):

.github/workflows/ci_pullrequest.yml
bryanculver: @glennmatthews What if instead we added said file in the repo bit `.gitignore` and ignored all changes?

So:
```
echo "{}" > nautobot/ui/src/app_routes.json
git add -f nautobot/ui/src/app_routes.json
```

glennmatthews: That's probably fine, I think?

bryanculver: Womp womp...

> Editing .gitignore will only ignore files so they will not be added to the git repository. However files that are tracked already will not be ignored.

Testing locally it still picked up future changes to the file. I'll update the PR to create the file if it doesn't exist.

bryanculver: https://github.com/nautobot/nautobot/pull/3736/files#diff-5c50d475237a0b724fe8d90c7862d19b2a48d8cf8f0c334b8e7d8d336db219f8R6-R9

How about this instead @glennmatthews?

nautobot/ui/src/utils/navigation.js
```diff
@@ -0,0 +1,32 @@
+import { Suspense } from "react";
+import { lazy } from "react";
+import { getComponentFromModule } from "./app-import";
+import { LoadingWidget } from "@components/LoadingWidget";
+
+let app_routes = {};
+try {
+    app_routes = require("../app_routes.json");
+} catch {}
```
bryanculver, line 9: Because JavaScript I can't do:

```js
try {
    const app_routes = require("../app_routes.json");
} catch {
    const app_routes = {};
}
```

.github/workflows/ci_pullrequest.yml
glennmatthews: I don't have a strong opinion about how we do this, just want to make sure that if it's something a developer needs to do, they are informed about it. :-)

bryanculver: They shouldn't have to do that anymore. Exception handling FTW!

PR #3754, opened 2023-05-16
Added Loading Widget to the left of the ObjectListTable Buttons after initial data load.

HanlinMiao:
# Closes: N/A
# What's Changed

# TODO

- [ ] Unit, Integration Tests


nautobot/ui/src/components/ObjectListTable.js
```diff
@@ -174,6 +181,13 @@ export default function ObjectListTable({
                 mt="3"
                 isLoaded={data_loaded}
             >
+                {!data_fetched && data_loaded ? (
+                    <Box background="white-0" borderRadius="md" padding="md">
+                        <LoadingWidget name={tableTitle}></LoadingWidget>
```
timizuoebideri1, line 186:
```suggestion
                        <LoadingWidget name={tableTitle} />
```

timizuoebideri1: why does the `LoadingWidget` needs the name props?

HanlinMiao: `tableTitle` in this case is a capitalized name of the models. So for Locations, the widget would display `Loading Locations ...`. For IPAddresses, it would display `Loading Ip Addresses ...`.

nautobot/ui/src/components/ObjectListTable.js
```diff
@@ -128,7 +130,12 @@ export default function ObjectListTable({
     });
 
     return (
-        <Box background="white-0" borderRadius="md" padding="md">
+        <Box
+            background="white-0"
+            borderRadius="md"
```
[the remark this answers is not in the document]
HanlinMiao: Yeah at least for the props here, they are CSS style attributes. There is a bad pattern here in that we are mixing and matching inherent prop attributes and CSS style attributes inline. (See the <Heading> tag below). 
But at least this PR did not perpetuate the bad pattern. The only thing I added here is `id="ObjectListContainer"`.

nautobot/ui/src/components/ObjectListTable.js
```diff
@@ -174,6 +181,13 @@ export default function ObjectListTable({
                 mt="3"
                 isLoaded={data_loaded}
             >
+                {!data_fetched && data_loaded ? (
+                    <Box background="white-0" borderRadius="md" padding="md">
```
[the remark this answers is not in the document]
HanlinMiao: Will open an issue to clean this up.

nautobot/ui/src/components/ObjectListTable.js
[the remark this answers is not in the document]
bryanculver: They aren't true CSS style attributes. Chakra doesn't reinvent the wheel here: https://chakra-ui.com/docs/styled-system/style-props#background

nautobot/ui/src/components/ObjectListTable.js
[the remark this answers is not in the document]
bryanculver: https://github.com/nautobot/nautobot/pull/3754/files#r1200911866

nautobot/ui/src/components/paginator_form.js
```diff
@@ -7,6 +7,9 @@ export default function PaginatorForm({ start, end, total_count }) {
     let paginator_string = `Showing ${start} - ${end} of ${total_count}`;
     function onPageSizeChange(event) {
         let offset = searchParams.get("offset");
+        document
+            .getElementById("ObjectListContainer")
```
bryanculver, line 11: I wonder if we can access, or have one passed in, a parent to scroll to instead of having to use a hardcoded ID.

bryanculver (approved):
Net improvement but I'd like to see us figure out how we can infer what to scroll to instead of a hard-coded name.

nautobot/ui/src/components/paginator_form.js
HanlinMiao: I can do
```
document
            .getElementsByClassName("chakra-table")
            .scrollIntoView({ alignToTop: true });
```
Which is functionally the same when we have **one and only one** `ObjectListTable` present on the page if that is better.

bryanculver: If you set this: https://github.com/nautobot/nautobot/blob/b811c5a346ad3dcf60acba5341e0ae5c3a6cbe7c/nautobot/ui/src/components/ObjectListTable.js#L191-L194 to something like `const TheTable` you could pass that into Paginator (`<Paginator table={TheTable} #...`) and then you should be able to fire `table.scrollIntoView()`.

HanlinMiao: It seems like passing a react component as a prop does not work here. Because `scrollIntoView`() is not a function of the react component. Instead, I used the `useRef` react hook, which does not need a hard-coded ID and is also the standard way to do this in react.
https://github.com/nautobot/nautobot/pull/3754/commits/b98fe1269b618be54edbaa7266a3c7bdb276d997

bryanculver (approved):

PR #3780, opened 2023-05-19
Fix #851 - add CircuitTermination list/import/export views

### LINKED ISSUE #851: Allow Import/Export of CircuitTerminations
### User Story

As Austin the Network Automation Engineer,

I regularly take partial exports from our production Nautobot instance to import them in a separate local development instance.  There are reasons for which I can't take a full database dump from production at this time, for example, I'm working on integrating a new feature in testing and I have quite a bit of data which is new and doesn't exist in production yet.  I need to be able to export/import CircuitTermination data somehow into Nautobot.  A better approach than providing list views for the CircuitTermination model might be to have a writeable graphql endpoint to which I can export/import any data I would like ;).  

I know this is done when:

* I can export CircuitTermination data (any format is acceptable, even if the export is json from a GraphQL query)
* I can import the exported data from above into Nautobot somehow

 
### Database Changes (Optional)

None

### External Dependencies (Optional)

None

### Requires

- [x] #2276

glennmatthews:
# Closes: #851
# What's Changed

Building atop #3722, we now have the ability to do CSV import and export of CircuitTermination records. To wire this up to the legacy UI, I had to add a list view for circuit-terminations (we already have a detail view) so that users can get to the Import and Export buttons. This in turn required implementing a `CircuitTerminationTable` class. I've also added generic view test coverage.

image

image

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/circuits/navigation.py
+++ b/nautobot/circuits/navigation.py
@@ -18,9 +18,17 @@ menu_items = (
                         ],
                     ),
                     NavMenuItem(
+                        link="circuits:circuittermination_list",
+                        name="Circuit Terminations",
+                        weight=200,
+                        permissions=[
+                            "circuits.view_circuittermination",
+                        ],
+                    ),
+                    NavMenuItem(
                         link="circuits:circuittype_list",
                         name="Circuit Types",
-                        weight=200,
+                        weight=300,
                         permissions=[
                             "circuits.view_circuittype",
                         ],
@@ -28,7 +36,7 @@ menu_items = (
                     NavMenuItem(
                         link="circuits:provider_list",
                         name="Providers",
-                        weight=300,
+                        weight=400,
                         permissions=[
                             "circuits.view_provider",
                         ],
@@ -36,7 +44,7 @@ menu_items = (
                     NavMenuItem(
                         link="circuits:providernetwork_list",
                         name="Provider Networks",
-                        weight=400,
+                        weight=500,
                         permissions=[
                             "circuits.view_providernetwork",
                         ],


nautobot/circuits/views.py
```diff
@@ -59,23 +60,38 @@ def get_extra_context(self, request, instance):
...
-    def get_return_url(self, request, obj):
-        return obj.circuit.get_absolute_url()
+    def get_return_url(self, request, obj=None):
+        # First, see if `return_url` was specified as a query parameter or form data. Use this URL only if it's
+        # considered safe.
+        query_param = request.GET.get("return_url") or request.POST.get("return_url")
+        if query_param and is_safe_url(url=query_param, allowed_hosts=request.get_host()):
+            return query_param
+
+        if obj is not None and obj.present_in_database and obj.pk:
```
HanlinMiao, line 91: Are `obj is not None` and `obj.pk` and `obj.present_in_database` all necessary for this check?

nautobot/circuits/tables.py
```diff
@@ -141,3 +141,41 @@ class Meta(BaseTable.Meta):
...
+
+#
+# Circuit Terminations
+#
+
+
+class CircuitTerminationTable(BaseTable):
+    pk = ToggleColumn()
+    circuit = tables.Column(linkify=True)
+    term_side = tables.Column(linkify=True)
```
HanlinMiao, line 154: Do we need to linkify `term_side` here? Where does the link take us?

HanlinMiao (approved):
Long-overdue ListView for CircuitTerminations LGTM!

nautobot/circuits/views.py
glennmatthews: It's copied logic from the base viewset mixin. IIRC it's there to prevent us from accidentally redirecting to an object that's in the process of being deleted.

nautobot/circuits/tables.py
glennmatthews: It links to the detail view for the circuittermination itself. Since they don't have a `name` or anything like that it was the only place I could reasonably add this link.

PR #3785, opened 2023-05-19
Make status fields non-nullable and fixup any existing invalid nulls

glennmatthews:
# Closes: #DNE
# What's Changed

All objects with `status` fields had an inconsistency in that the field was `blank=False` (requiring the UI or API to always provide a value for these fields) but `null=True` (meaning that the ORM could, and by default would, create objects with NULL `status` fields). The intent was always for these fields to be required/non-nullable; this PR fixes the code and DB to actually match that intent, including data migrations to create a `NULL` Status and use it if any null `status`es are found in the database.

I've left the `StatusModel` mixin in place, and kept its default behavior of `blank=False, null=True` because I know we have various apps using that mixin; however I've changed all of our core models to simply include a `StatusField` directly rather than using an unnecessary mixin class. 

I've removed the `RoleModelMixin` and `RoleRequiredRoleModelMixin` classes as well (those were `next`-only and so have no adoption outside of core to worry about) and replaced their usage with direct inclusion of `RoleField`. This did result in a migration of the `Device.role` field to be (correctly) non-nullable; I haven't bothered with a data migration here because its nullability was only a temporary bug introduced in `next` and shouldn't be encountered in the wild.

~I haven't run tests yet; it's quite possible that some number of tests were creating bad data and may now fail as a result; if so I'll fix those as well.~ Tests should all be passing now. Had to make so many fixes to bad test data...

# TODO

- [x] Unit, Integration Tests


bryanculver:
I see the benefit of making a model "Status-able" and "Role-able" with just a mixin. Are you leaving it open so people can let them be nullable in their models instead of us enforcing it?

glennmatthews:
> I see the benefit of making a model "Status-able" and "Role-able" with just a mixin. 

It's one line of code to add a `status` or `role` field to a class, there's no other functionality attached to the mixins. Having two different `RoleModelMixin` classes just to distinguish between `role = RoleField(blank=True, null=True)` and `role = RoleField(blank=False, null=False)` seemed like overkill to me.

> Are you leaving it open so people can let them be nullable in their models instead of us enforcing it?

Partly that, but also because it looks like overriding or setting defaults for `blank` or `null` in the `__init__()` of these fields is oddly problematic (seems to have been why `Device.role` was nullable despite using the `RoleRequiredRoleModelMixin`?)

nautobot/circuits/migrations/0017_fixup_null_statuses.py
```diff
@@ -0,0 +1,22 @@
+from django.db import migrations
+
+from nautobot.extras.utils import fixup_null_statuses
+
+
+def migrate_null_statuses(apps, schema):
+    status_model = apps.get_model("extras", "Status")
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    for model_name in ("Circuit",):
```
HanlinMiao, line 9: Is the for loop necessary here?

nautobot/dcim/migrations/0041_fixup_null_statuses.py
```diff
@@ -0,0 +1,30 @@
+from django.db import migrations
+
+from nautobot.extras.utils import fixup_null_statuses
+
+
+def migrate_null_statuses(apps, schema):
+    status_model = apps.get_model("extras", "Status")
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    for model_name in (
```
HanlinMiao, line 9: Wonder if we can just do a `FeatureQuery("status").get_query()` or something similar here instead of hardcoding the names into the for loop?

nautobot/dcim/tests/test_api.py
```diff
@@ -83,12 +83,20 @@ def test_trace(self):
...
                 device_type=DeviceType.objects.first(),
                 role=Role.objects.get_for_model(Device).first(),
+                status=Status.objects.get_for_model(Device).first(),
                 name="Peer Device",
             )
             if self.peer_termination_type is None:
                 raise NotImplementedError("Test case must set peer_termination_type")
-            peer_obj = self.peer_termination_type.objects.create(device=peer_device, name="Peer Termination")
-            cable = Cable(termination_a=obj, termination_b=peer_obj, label="Cable 1")
+            if self.peer_termination_type is Interface:
```
HanlinMiao, line 91: Only `Interface` has a `status` huh? interesting.

nautobot/ipam/migrations/0029_fixup_null_statuses.py
```diff
@@ -0,0 +1,26 @@
+from django.db import migrations
+
+from nautobot.extras.utils import fixup_null_statuses
+
+
+def migrate_null_statuses(apps, schema):
+    status_model = apps.get_model("extras", "Status")
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    for model_name in (
```
HanlinMiao, line 9: Same comment about `FeatureQuery`

HanlinMiao (approved):

nautobot/circuits/migrations/0017_fixup_null_statuses.py
glennmatthews: No, just laziness that I could try to justify as self-consistency. :-)

nautobot/dcim/migrations/0041_fixup_null_statuses.py
glennmatthews: Could do that if we're okay with migrating models from multiple apps in a single migration, which I *think* we would want to avoid? Otherwise I'd have to take the FeatureQuery and filter it down again by app_label which seems like an unnecessary layer of indirection.

nautobot/dcim/tests/test_api.py
glennmatthews: VMInterface does too, but it can't be cabled.

nautobot/extras/models/statuses.py
```diff
@@ -104,10 +101,12 @@ def _get_FIELD_color(self, field):
 
 class StatusModel(models.Model):
```
bryanculver, line 102: Can we throw a deprecation warning decorator here?

glennmatthews: I can see if `class_deprecated_in_favor_of` can be used here as-is.

bryanculver:
> 



> > I see the benefit of making a model "Status-able" and "Role-able" with just a mixin.
> 
> It's one line of code to add a `status` or `role` field to a class, there's no other functionality attached to the mixins. Having two different `RoleModelMixin` classes just to distinguish between `role = RoleField(blank=True, null=True)` and `role = RoleField(blank=False, null=False)` seemed like overkill to me.
> 
> > Are you leaving it open so people can let them be nullable in their models instead of us enforcing it?
> 
> Partly that, but also because it looks like overriding or setting defaults for `blank` or `null` in the `__init__()` of these fields is oddly problematic (seems to have been why `Device.role` was nullable despite using the `RoleRequiredRoleModelMixin`?)

Any concern about developers not setting the `StatusField` to the `status` property on a model? It would make it harder on the UI to magically put the status value in the header if a model has it:

image


```py
burrito = StatusField(blank=False, null=False)
```

glennmatthews:
> > 
> 
> > > I see the benefit of making a model "Status-able" and "Role-able" with just a mixin.
> > 
> > 
> > It's one line of code to add a `status` or `role` field to a class, there's no other functionality attached to the mixins. Having two different `RoleModelMixin` classes just to distinguish between `role = RoleField(blank=True, null=True)` and `role = RoleField(blank=False, null=False)` seemed like overkill to me.
> > > Are you leaving it open so people can let them be nullable in their models instead of us enforcing it?
> > 
> > 
> > Partly that, but also because it looks like overriding or setting defaults for `blank` or `null` in the `__init__()` of these fields is oddly problematic (seems to have been why `Device.role` was nullable despite using the `RoleRequiredRoleModelMixin`?)
> 
> Any concern about developers not setting the `StatusField` to the `status` property on a model? It would make it harder on the UI to magically put the status value in the header if a model has it:
> image
> 
> ```python
> burrito = StatusField(blank=False, null=False)
> ```

If an app has a use case for a `StatusField` that isn't semantically equivalent to `status`, I don't think that's necessarily something we need to worry about preventing or automatically handling. (I'd be fine with a pylint warning since those can always be disabled by someone if it's a false positive.) I think we can safely assert in general that "if you have a `status` field, it must be a `StatusField`" (though `JobResult` already violates that, IIRC) but not "if you are using `StatusField`, you must name it `status`."

bryanculver (approved):
Decorated and lint rule request opened. LGTM

Issue #3787, opened 2023-05-20
Out of sort memory errors in Job and JobResult list views with MySQL backend


grelleum:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.16
* Python version: 3.8.13
* Database platform, version: MySQL 8.0.29

### Steps to Reproduce
1. Run a significant number of large SSoT jobs.
2. Navigate to the Jobs list view: `/extras/jobs`
3. Navigate to the Job-results list view: `/extras/job-results`

### Expected Behavior
Job or job result list view should load.

### Observed Behavior
```
OperationalError at /extras/jobs/
(1038, 'Out of sort memory, consider increasing server sort buffer size')

Error during template rendering
{% include 'extras/inc/job_label.html' with result=result %}
```

glennmatthews:
I wonder to what extent the changes in #3669 already mitigate this issue.

PR #3788, opened 2023-05-21
Fix MySQL Out of sort memory error in Job list views (#3787)

grelleum:
# Closes: #3787
# What's Changed
For the `JobResultListView` the `defer("data")` has been added to the queryset. This avoids the sort memory issue and speeds up the page load.

For  the `JobListView` the issue is with the `Last Status` column in the table.  This column uses the `latest_result` accessor, which is a property on the `Job` model. This property returns the complete object, when it only requires the `status` field.  This property is not used elsewhere in the Nautobot code, aside from the tests.

# TODO
- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/models/jobs.py
+++ b/nautobot/extras/models/jobs.py
@@ -341,7 +341,12 @@ class Job(PrimaryModel):
 
     @property
     def latest_result(self):
-        """Return the most recent JobResult object associated with this Job."""
+        """
+        Return the most recent JobResult object associated with this Job.
+        
+        Note that, as a performance optimization for this function's repeated use in
+        JobListview, the returned object only includes its `status` field.
+        """
         if self._latest_result is None:
             self._latest_result = self.results.only("status").first()
         return self._latest_result
--- a/nautobot/extras/models/jobs.py
+++ b/nautobot/extras/models/jobs.py
@@ -343,7 +343,7 @@ class Job(PrimaryModel):
     def latest_result(self):
         """
         Return the most recent JobResult object associated with this Job.
-        
+
         Note that, as a performance optimization for this function's repeated use in
         JobListview, the returned object only includes its `status` field.
         """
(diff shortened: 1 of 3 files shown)


nautobot/extras/views.py
```diff
@@ -1491,7 +1491,7 @@ class JobResultListView(generic.ObjectListView):
     List JobResults
     """
 
-    queryset = JobResult.objects.select_related("job_model", "obj_type", "user").prefetch_related("logs")
+    queryset = JobResult.objects.defer("data").select_related("job_model", "obj_type", "user").prefetch_related("logs")
```
glennmatthews, line 1494: I wonder if a similar change is needed in `JobResultBulkDeleteView`?

nautobot/extras/models/jobs.py
```diff
@@ -342,7 +342,7 @@ def class_path(self):
     @property
     def latest_result(self):
         if self._latest_result is None:
-            self._latest_result = self.results.first()
+            self._latest_result = self.results.only("status").first()
```
glennmatthews, line 345: Would you mind adding a docstring for `latest_result` documenting this restriction/requirement?

nautobot/extras/models/jobs.py
```diff
@@ -341,8 +341,9 @@ def class_path(self):
 
     @property
     def latest_result(self):
+        """Return the most recent JobResult object associated with this Job."""
```
glennmatthews, line 344:
```suggestion
        """
        Return the most recent JobResult object associated with this Job.
        
        Note that, as a performance optimization for this function's repeated use in
        JobListview, the returned object only includes its `status` field.
        """
```

bryanculver:
@grelleum Can you address Glenn's feedback around the `JobResultBulkDeleteView` case: https://github.com/nautobot/nautobot/pull/3788#discussion_r1200460100 as well as fix black & flake8 failure. Would like to see this included in Monday's patch release.

nautobot/extras/views.py
grelleum: Added `.defer("data")` to queryset in `JobResultBulkDeleteView`.

grelleum:
Fixed whitespace issue in docstring and added `.defer("data")` to queryset in `JobResultBulkDeleteView`.

glennmatthews (approved):

Issue #2331, opened 2022-08-29
Jobs Overhaul: Improve Logging Utilities


itdependsnetworks:
### As ...

Patti - Platform Admin

### I want ...

The ability to view all of the logs for a lob, not just the ones that are shown in the logging facility (e.g.  self.log_success() ) 

### So that ...

so that I can see full stacktraces and similar issues that I may not otherwise have easy access to. 

### I know this is done when...

There is a UI and API that provides feedback that is the same as I would see on the console of a worker in debug mode. 

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

unknown

---

```[tasklist]
### Tasks
- [ ] Ditch bespoke logging levels and align standard logging levels (see: https://docs.python.org/3/library/logging.html#levels, we don't currently support `critical` and `success` does not exist)
- [ ] Use built-in Celery logging utility `get_task_logger` instead of `self.log_success()` as it's just a wrapper around normal Python logging conventions (see: https://docs.celeryq.dev/en/stable/userguide/tasks.html?highlight=get_task_logger#logging)
- [ ] Intercept Celery logging utilities to send to DB (see: https://docs.celeryq.dev/en/stable/userguide/signals.html?highlight=logging#after-setup-task-logger)
```

### Considerations

- We don't need log levels outside the standard Python ones above
- stdout/stderr are redirected by default: https://docs.celeryq.dev/en/stable/userguide/configuration.html#std-setting-worker_redirect_stdouts so may "just work" but making them work should not be considered in-scope for this
  - The default behavior redirects to a new `celery.get_logger` instance which does not implement our custom task database handler. Will probably need to use the `task_prerun` signal to implement the `sys.stdout/sys.stderr` redirect.
- Structlog support is a stretch goal and a separate issue
  - If we can use the stdlib logger to keep the `log_object` field functionality, structlog support can be pushed back to a 2.x or beyond feature request (maybe use the `extra` kwarg in the stdlib logger methods to pass the object)
- Controlling emitted log-levels is in scope for #2071

glennmatthews:
[...]

PR #3611, opened 2023-04-18
Jobs overhaul: Replace run_job with NautobotTask

gsnider2195:
# Closes: #765 
# What's Changed

- #3483
- #3545
- #3593
- #3723
- #3800
- #3799
- #3815
- #3818
- #3823
- #3824
- #3840

# TODO

- [ ] Explanation of Change(s)
- [ ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [ ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example Plugin Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design
- [ ] Remove references to `test_` and `post_run` functions in docs

PR #3799, opened 2023-05-23
Overhaul job logging

gsnider2195:
# Closes: #2331
# What's Changed

- Removed bespoke logging methods on `BaseJob` and extended celery task loggers to log to Nautobot `JobLogEntry` database entries
  - Currently logs all message levels to the database. Setting the `--loglevel` argument when starting the worker with `nautobot-server celery worker` will only filter the log messages output to the console on the worker
  - Jobs may either use `self.logger` or retrieve the logger from `celery.utils.log.get_task_logger(__name__)`, they should be identical
- Changed logging levels from previous custom levels: `default`, `info`, `success`, `warning`, `failure` to python default logging levels: `debug`, `info`, `warning`, `error` and `critical`
  - Migrated `success` and `default` to `info` and `failure` to `error`
- Extended celery stdout/stderr redirect logging to also log to Nautobot `JobLogEntry` database entries if `CELERY_WORKER_REDIRECT_STDOUTS` is enabled (enabled by default) 
  - STDOUT and STDERR log messages honor the `CELERY_WORKER_REDIRECT_STDOUTS_LEVEL` setting, which defaults to outputting these messages as level `warning`
- Updated `runjob` management command and `JobResult.enqueue_job` to properly setup logging when running synchronously on the nautobot server as the nautobot server does not trigger the celery signals that set up logging on the workers
- Removed `active_test` property from `BaseJob` class. Jobs may specify the grouping they would like to use in their log messages by passing it in the [`extra` kwarg](https://docs.python.org/3/library/logging.html#logging.Logger.debug) when logging their message. If no grouping is provided, the name of the method that logged the message will be used.
- Removed `failed` property from `BaseJob` class. Celery tracks the state of the job now.

# TODO

- [ ] Unit, Integration Tests


gsnider2195:
Still needs some more migration docs for job authors but marking as ready for review to review the implementation

nautobot/extras/jobs.py
```diff
@@ -769,8 +768,7 @@ def save_file(uploaded_file):
         fp = FileProxy.objects.create(name=uploaded_file.name, file=uploaded_file)
         return fp.pk
 
-    @staticmethod
```
bryanculver, line 772: Why no longer a static method?

nautobot/core/celery/log.py
```diff
@@ -0,0 +1,30 @@
...
+        from nautobot.extras.models.jobs import JobResult
+
+        try:
+            self.format(record)
+
+            if record.task_id == "???":
+                return
+            job_result = JobResult.objects.filter(id=record.task_id)
+            if not job_result.exists():
+                return
```
bryanculver, line 22: Are there cases where `???` would be frequent enough to not do the search always? Seems redundant.

nautobot/extras/management/commands/runjob.py
```diff
@@ -62,13 +62,7 @@ def handle(self, *args, **options):
...
-            job_result = JobResult.execute_job(
-                job_model=job_model,
-                user=user,
-                profile=options["profile"],
-                **data,
-            )
-
+            job_result = JobResult.execute_job(job_model, user, profile=options["profile"], **data)
         else:
             job_result = JobResult.enqueue_job(job_model, user, profile=options["profile"], **data)
```
bryanculver, line 67: Any reason not to stick to specifying with kwargs?

nautobot/extras/tests/example_jobs/api_test_job.py
```diff
@@ -17,10 +22,10 @@ class Meta:
...
 
     def run(self, var1, var2, var3, var4):
-        self.log_debug(message=var1)
-        self.log_info(message=var2)
-        self.log_success(message=var3)
-        self.log_warning(message=var4)
+        logger.debug(var1)
+        logger.info(var2)
+        logger.info(var3)
+        logger.warning(var4)
```
bryanculver, line 28: Not that we want to document this too heavily but for external calls and testing, we should be able to call functions in the job to test locally, and log messages won't :boom: because of a failure to access the DB. Thinking of testing functions that call to external systems without having to run the whole job.

nautobot/core/celery/log.py
```diff
@@ -0,0 +1,30 @@
...
+            job_result.first().log(
```
bryanculver, line 23: I worry this will come back to bite us in the concurrency case. Are `task_id`s unique globally, even if two workers pick up the same task?

bryanculver (approved):

nautobot/extras/tests/example_jobs/test_atomic_transaction.py
```diff
@@ -22,9 +27,10 @@ def run(self, fail=False):
             Status.objects.create(name="Test database atomic rollback 1")
             if fail:
                 raise SimulatedError("simulated failure")
-        except Exception:
-            self.log_failure("Job failed, all database changes have been rolled back.")
-        self.log_success("Job succeeded.")
+        except Exception as err:
+            logger.error("Job failed, all database changes have been rolled back.")
+            raise err
```
glennmatthews, line 32: Maybe here and below:

```suggestion
        except Exception:
            logger.error("Job failed, all database changes have been rolled back.")
            raise
```

nautobot/extras/tests/example_jobs/test_fail.py
```diff
@@ -13,8 +18,8 @@ def run(self):
         """
         Job function.
         """
-        self.log_success(obj=None)
-        self.log_failure("Test failure")
+        logger.info("Success")
+        raise RunJobTaskFailed("Test failure")
```
glennmatthews, line 22: Task failed successfully? 🤪

nautobot/extras/tests/example_jobs/test_file_upload_fail.py
```diff
@@ -21,7 +26,7 @@ class Meta:
 
     def run(self, file):
         contents = str(file.read())
-        self.log_warning(message=f"File contents: {contents}")
+        logger.warning("File contents: %s", contents)
```
glennmatthews, line 29: Should any/all of these test jobs use `self.logger` instead of `logger`?

nautobot/extras/tests/test_filters.py
```diff
@@ -1000,7 +1000,7 @@ class JobLogEntryTestCase(FilterTestCases.FilterTestCase):
     def setUpTestData(cls):
         cls.job_result = JobResult.objects.create(name="test")
 
-        for log_level in ("debug", "info", "success", "warning"):
+        for log_level in ("debug", "info", "error", "warning", "critical"):
```
glennmatthews, line 1003:
```suggestion
        for log_level in ("debug", "info", "warning", "error", "critical"):
```

nautobot/extras/management/commands/runjob.py
gsnider2195: The signature for both of these methods is the same so I just wanted to make them consistent here

nautobot/core/celery/log.py
gsnider2195: We would have to wrap this in a try/except block because if you pass a string that's not a uuid to this filter it raises an exception: `django.core.exceptions.ValidationError: ['“???” is not a valid UUID.']`

nautobot/extras/jobs.py
gsnider2195: I needed access to the object to be able to call `self.logger`

nautobot/extras/tests/example_jobs/api_test_job.py
gsnider2195: If the job isn't running on a celery worker or using `JobResult.execute_job`, the database logger handler won't be present. Directly running any code that sends messages to the celery task logger should work.

nautobot/core/celery/log.py
gsnider2195: The `task_id` is generated when enqueuing the job by celery's `apply_async` using `uuid()` so it has the same likelihood of collision that it had before when our `enqueue_job` was doing the same.

examples/example_plugin/example_plugin/jobs.py
```diff
@@ -1,5 +1,7 @@
+import sys
 import time
 
+from celery.utils.log import get_task_logger
```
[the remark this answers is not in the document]
gsnider2195: It should already be possible to import from there since we have `from celery.utils.log import get_task_logger` in `nautobot.extras.jobs.py`? We'd just need to document it?

nautobot/core/celery/log.py
```diff
@@ -0,0 +1,30 @@
...
+        if current_task is None:
+            return
+
+        from nautobot.extras.models.jobs import JobResult
+
+        try:
+            self.format(record)
+
+            if record.task_id == "???":
+                return
```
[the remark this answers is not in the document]
gsnider2195: That's the only expected non-uuid value that we would get here. Anything else should raise an exception when trying to filter on the JobResult manager below.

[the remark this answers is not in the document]
glennmatthews: When is `"???"` expected? I'm missing that context.

nautobot/core/celery/log.py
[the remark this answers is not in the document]
gsnider2195: It would probably have to be an exception or a write to stdout/stderr since trying to log here might go into an infinite loop. It's probably safe to raise an exception here since I can't think of a way to get to this line of code without the job result being created.

bryanculver: I think a try/except might be more robust and easier to read.

nautobot/docs/additional-features/jobs.md
```diff
@@ -880,17 +882,23 @@ class DeviceConnectionsReport(Job):
...
         for console_port in ConsolePort.objects.prefetch_related('device').filter(device__status=STATUS_ACTIVE):
             if console_port.connected_endpoint is None:
-                self.log_failure(
-                    obj=console_port.device,
-                    message="No console connection defined for {}".format(console_port.name)
+                self.logger.error(
+                    "No console connection defined for %s",
+                    console_port.name,
+                    extra={"object": console_port.device},
                 )
```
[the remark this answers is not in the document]
gsnider2195: I think for our example here we should raise an exception. We don't have an equivalent for `log_failure` anymore, the failed flag would have to be tracked by the job author and an exception raised at the end of the job.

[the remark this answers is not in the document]
glennmatthews: The downside is that (thinking of this one as a "report") raising an exception makes it end immediately on the first failure encountered, whereas it'd probably be desirable to get a list of all failures in one run.

nautobot/docs/additional-features/jobs.md
```diff
@@ -880,17 +882,23 @@ class DeviceConnectionsReport(Job):
         # Check that every console port for every active device has a connection defined.
```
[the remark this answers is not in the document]
gsnider2195: This needs to be updated. Added as a todo for #3611

nautobot/extras/migrations/0087_joblogentry__log_level_data_migration.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    Migrate job log levels from 1.0 values to 2.0 values.
+    default => info
+    success => info
+    failure => error
+    """
+    JobLogEntry = apps.get_model("extras", "JobLogEntry")
+
+    for entry in JobLogEntry.objects.all():
+        if entry.log_level in LOG_LEVEL_MAPPING:
```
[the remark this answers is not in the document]
gsnider2195: Would it be more efficient to do `JobLogEntry.objects.filter(log_level="failure").update(log_level="error")` for each of the levels?

nautobot/extras/migrations/0087_joblogentry__log_level_data_migration.py
```diff
@@ -0,0 +1,36 @@
...
+    "default": "info",
+    "success": "info",
+    "failure": "error",
+}
+
+
+def migrate_job_log_entry_log_levels(apps, schema_editor):
+    """
+    Migrate job log levels from 1.0 values to 2.0 values.
+    default => info
```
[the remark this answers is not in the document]
gsnider2195: I think default was rendered as a debug message on the job result page. The new default is `info` but should it be `debug` instead? If we migrate `default` to `debug` it would make it possible to reverse the migration.

nautobot/docs/models/extras/jobbutton.md
```diff
@@ -115,24 +115,24 @@ class ExampleComplexJobButtonReceiver(JobButtonReceiver):
...
             else:
                 self._run_site_job(obj)
         if isinstance(obj, Device):
             if not user.has_perm("dcim.add_device"):
-                self.log_failure(obj=obj, message=f"User '{user}' does not have permission to add a Device.")
+                self.logger.error("User '%s' does not have permission to add a Device.", user, extra={"object": obj})
             else:
                 self._run_device_job(obj)
-        self.log_failure(obj=obj, message=f"Unable to run Job Button for type {type(obj).__name__}.")
+        self.logger.error("Unable to run Job Button for type %s.", type(obj).__name__, extra={"object": obj})
```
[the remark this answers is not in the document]
gsnider2195: I can update the example to track failure with a flag

nautobot/extras/tests/example_jobs/test_file_upload_fail.py
gsnider2195: Both of them are supported since they both return the same logger. We need to set `self.logger` to use on some of our internal methods in `BaseJob` so it needs to stay but I think we should document the usage of `get_task_logger` since it's closer to the stdlib logging api.

nautobot/extras/models/jobs.py
```diff
@@ -784,7 +809,7 @@ def log(
             log.save(using=JOB_LOGS)
 
         if logger:
-            if level_choice == LogLevelChoices.LOG_FAILURE:
+            if level_choice == LogLevelChoices.LOG_ERROR:
                 log_level = logging.ERROR
             elif level_choice == LogLevelChoices.LOG_WARNING:
                 log_level = logging.WARNING
```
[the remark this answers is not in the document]
gsnider2195: It could change to 

```py
log_method = getattr(logger, level_choice, None)
if log_method is not None and callable(log_method):
    log_method(message)
else:
    raise Exception(f"this should never happen")
```

changes/2331.changed
```diff
@@ -0,0 +1,2 @@
+Changed `JobLogEntry.log_level` choices from `default`, `info`, `success`, `warning`, `failure` to python default logging levels `debug`, `info`, `warning`, `error` and `critical`.
```
glennmatthews, line 1:
```suggestion
Changed `JobLogEntry.log_level` choices from `default`, `info`, `success`, `warning`, `failure` to Python default logging levels `debug`, `info`, `warning`, `error` and `critical`.
```

nautobot/docs/additional-features/jobs.md
```diff
@@ -447,43 +447,45 @@ Again, defining user variables is totally optional; you may create a job with a
...
 
-!!! note
-    Using `self.log_failure()`, in addition to recording a log message, will flag the overall job as failed, but it will **not** stop the execution of the job, nor will it result in an automatic rollback of any database changes made by the job. To end a job early, you can use a Python `raise` or `return` as appropriate. Raising any exception (e.g. `ValueError` for malformed input values) will ensure that any database changes are rolled back as part of the process of ending the job. `AbortTransaction` from Nautobot, which was recommended in past versions of the docs, should explicitly **not** be used, as it is only intended for use by Nautobot's internal job handling.
++/- 2.0.0
+    The convenience method to mark a job as failed, `log_failure()` has been removed. To replace the functionality of this method, you can log an error message with `self.logger.error()` and then raise an exception to fail the job. Note that it is no longer possible to manually set the job result status as failed without raising an exception in the job.
```
glennmatthews, line 471:
```suggestion
    The convenience method to mark a job as failed, `log_failure()`, has been removed. To replace the functionality of this method, you can log an error message with `self.logger.error()` and then raise an exception to fail the job. Note that it is no longer possible to manually set the job result status as failed without raising an exception in the job.
```

nautobot/extras/migrations/0087_joblogentry__log_level_data_migration.py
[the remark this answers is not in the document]
glennmatthews: I think it's fine for the `log()` API to now default to `info`, but I do think it would be more correct to past usage to map `default` entries to `debug`.

nautobot/extras/tests/example_jobs/api_test_job.py
glennmatthews, line 28: Perhaps:

```suggestion
        logger.warning(var3)
        logger.error(var4)
```

nautobot/docs/additional-features/jobs.md
```diff
@@ -447,42 +447,75 @@ Again, defining user variables is totally optional; you may create a job with a
...
 
-!!! note
-    Using `self.log_failure()`, in addition to recording a log message, will flag the overall job as failed, but it will **not** stop the execution of the job, nor will it result in an automatic rollback of any database changes made by the job. To end a job early, you can use a Python `raise` or `return` as appropriate. Raising any exception (e.g. `ValueError` for malformed input values) will ensure that any database changes are rolled back as part of the process of ending the job. `AbortTransaction` from Nautobot, which was recommended in past versions of the docs, should explicitly **not** be used, as it is only intended for use by Nautobot's internal job handling.
++/- 2.0.0
+    The logging functions (example: `self.log(message)`, `self.log_success(obj=None, message=message)`, etc) have been removed. Also, the convenience method to mark a job as failed, `log_failure()`, has been removed. To replace the functionality of this method, you can log an error message with `self.logger.error()` and then raise an exception to fail the job. Note that it is no longer possible to manually set the job result status as failed without raising an exception in the job.
```
glennmatthews, line 482:
```suggestion
    The Job class logging functions (example: `self.log(message)`, `self.log_success(obj=None, message=message)`, etc) have been removed. Also, the convenience method to mark a job as failed, `log_failure()`, has been removed. To replace the functionality of this method, you can log an error message with `self.logger.error()` and then raise an exception to fail the job. Note that it is no longer possible to manually set the job result status as failed without raising an exception in the job.
```

glennmatthews (approved):

Issue #3798, opened 2023-05-23
UI 2.0: Redirect to Legacy View for Create/Edit Views


bryanculver:

bryanculver:
Closed with #3804

PR #3804, opened 2023-05-24
Redirect all routes ending with add and edit to legacyUI

timizuoebideri1:
# Closes: #3798 
# What's Changed

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/core/context_processors.py
+++ b/nautobot/core/context_processors.py
@@ -33,10 +33,11 @@ def settings(request):
     """
     Expose Django settings in the template context. Example: {{ settings.DEBUG }}
     """
+    use_base_template = request.path.endswith("/edit/") or request.path.endswith("/add/") or not request.COOKIES.get("newui", False)
 
     return {
         "settings": django_settings,
-        "root_template": "base_react.html" if request.COOKIES.get("newui", False) else "base_django.html",
+        "root_template": "base_django.html" if use_base_template else "base_react.html",
     }
 
 
--- a/nautobot/ui/src/components/ObjectListTable.js
+++ b/nautobot/ui/src/components/ObjectListTable.js
@@ -158,14 +158,20 @@ export default function ObjectListTable({
                             Actions
                         </UIButton>
                         <Icon.TbMinusVertical />
-
-                        <RouterButton
+                        
+                        <UIButton
                             to={`${location.pathname}add/`}
                             size="sm"
                             leftIcon={<PlusIcon />}
+                            onClick={(e) => {
+                                e.preventDefault();
+                                if (process.env.NODE_ENV === "production") {
+                                    document.location.href += "add/"
+                                }
+                            }}
                         >
                             Add {tableTitle}
-                        </RouterButton>
+                        </UIButton>
                     </ButtonGroup>
                 </Box>
             )}


nautobot/ui/src/components/ObjectListTable.js
```diff
@@ -158,14 +158,20 @@ export default function ObjectListTable({
                             Actions
                         </UIButton>
                         <Icon.TbMinusVertical />
-
-                        <RouterButton
+
```
timizuoebideri1, line 161:
```suggestion

```

glennmatthews:
a thought I had this morning - for the time being, we should probably invert the scope of the logic - instead of redirecting add/edit specifically to the legacy UI, we should use the legacy UI for all routes *except* home, list, and detail views.

nautobot/core/context_processors.py
```diff
@@ -34,9 +35,22 @@ def settings(request):
     Expose Django settings in the template context. Example: {{ settings.DEBUG }}
     """
 
+    try:
+        view_class = request.resolver_match.func.view_class
+    except AttributeError:
+        # Use this method to import the view class as apps (like example_plugin)
+        # do not have the 'view_class' attribute.
```
glennmatthews, line 42: Can you elaborate on this? I don't understand offhand why plugin apps would behave differently from core apps in this regard.

nautobot/core/views/mixins.py
```diff
@@ -131,6 +131,7 @@ class ObjectPermissionRequiredMixin(AccessMixin):
     """
 
     additional_permissions = []
+    use_new_ui = False
```
glennmatthews, line 134: Would you mind documenting this parameter in the class docstring?


I'm also not entirely comfortable with defining this in `ObjectPermissionRequiredMixin` at all - I get that it's probably our most widely-used mixin so that's convenient, but the UI toggle really doesn't have anything to do with permissions enforcement so it's kind of a mismatch. Should we add it to the generic view base classes directly instead?

nautobot/ui/src/components/ObjectListTable.js
```diff
@@ -158,14 +157,19 @@ export default function ObjectListTable({
...
                         <Icon.TbMinusVertical />
-
-                        <RouterButton
+                        <UIButton
                             to={`${location.pathname}add/`}
                             size="sm"
                             leftIcon={<PlusIcon />}
+                            onClick={(e) => {
+                                e.preventDefault();
+                                if (process.env.NODE_ENV === "production") {
```
glennmatthews, line 166: What's the purpose of / need for this check? Could you maybe add some explanatory comments?

nautobot/core/context_processors.py
```diff
@@ -1,3 +1,4 @@
+from importlib import import_module
 from django.conf import settings as django_settings
```
HanlinMiao, line 2: Nit 
```suggestion
from django.conf import settings as django_settings
from importlib import import_module
```

nautobot/core/views/generic.py
```diff
@@ -67,6 +67,7 @@ class ObjectView(ObjectPermissionRequiredMixin, View):
 
     queryset = None
     template_name = None
+    use_new_ui = True
```
HanlinMiao, line 70: Should this be in `ObjectListView` and other read-only mixins for `NautobotUIViewSet`?

nautobot/core/context_processors.py
```diff
@@ -34,9 +35,22 @@ def settings(request):
     Expose Django settings in the template context. Example: {{ settings.DEBUG }}
     """
 
+    try:
+        view_class = request.resolver_match.func.view_class
+    except AttributeError:
+        # Use this method to import the view class views that inherits from
+        # NautobotUIViewSet do not have the 'view_class' attribute.
```
timizuoebideri1, line 42:
```suggestion
        # NautobotUIViewSet, as this views do not have the 'view_class' attribute.
```

nautobot/ui/src/components/ObjectListTable.js
```diff
@@ -158,14 +157,22 @@ export default function ObjectListTable({
...
+                        <UIButton
                             to={`${location.pathname}add/`}
                             size="sm"
                             leftIcon={<PlusIcon />}
+                            onClick={(e) => {
+                                e.preventDefault();
+                                // Because there is currently no support for Add view in the new UI for production,
+                                // the code below checks if the app is running in production and redirects the user to
+                                // the Add page; after the page is reloaded, nautobot takes care of rendering the legacy UI.
+                                if (process.env.NODE_ENV === "production") {
```
bryanculver, line 169: Does this get resolved at `npm build` time because I don't think there is a `process.env.NODE_ENV` available to the browser in the static environment.

timizuoebideri1: Yup, During build

nautobot/core/context_processors.py
```diff
@@ -34,9 +35,22 @@ def settings(request):
...
+    except AttributeError:
+        # Use this method to import the view class views that inherits from
+        # NautobotUIViewSet, as this views do not have the 'view_class' attribute.
+        view_func = request.resolver_match.func
+        module_path = view_func.__module__
+        view_name = view_func.__name__
+        module = import_module(module_path)
+        view_class = getattr(module, view_name, None)
+    use_new_ui = request.COOKIES.get("newui", False)
+    use_new_ui = use_new_ui and getattr(view_class, "use_new_ui", False)
```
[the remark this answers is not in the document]
timizuoebideri1: True

bryanculver (changes requested):
Thinking about this over the weekend, I think having a new `BaseView` is unnecessary. This `BaseView` seems to only support model views and we have many non-model views, especially in plugins.

I'm not opposed to a `ModelBaseView` but I think that's unnecessary at this time. Instead if the check in `context_processors` assumes a view doesn't support `use_new_ui` if it's not declared should be sufficient.

bryanculver (approved):
If the tests pass, looks good to me.

nautobot/core/views/mixins.py
```diff
@@ -593,6 +596,7 @@ class ObjectListViewMixin(NautobotViewSetMixin, mixins.ListModelMixin):
         "per_page",  # used by get_paginate_count
         "sort",  # table sorting
     )
+    use_new_ui = True
```
glennmatthews, line 599: Does this result in all of `NautobotUIVIewSet` (e.g. the Circuits views) getting `use_new_ui=True` (even for its create/edit/delete views)? It seems like it would.

nautobot/core/views/mixins.py
```diff
@@ -578,6 +578,17 @@ class ObjectDetailViewMixin(NautobotViewSetMixin, mixins.RetrieveModelMixin):
...
     """
 
+    def retrieve(self, request, *args, **kwargs):
+        """
+        Retrieve a model instance.
+        """
+        instance = self.get_object()
+        serializer = self.get_serializer(instance)
+
+        context = serializer.data
```
bryanculver, line 588: Keeps the existing behavior with DRF: https://github.com/encode/django-rest-framework/blob/376a5cbbba3f8df9c9db8c03a7c8fa2a6e6c05f4/rest_framework/mixins.py#L51-L58

glennmatthews (approved):

Issue #3767, opened 2023-05-18
UI 2.0: detail views never populate if /detail-view-config endpoint returns 404


glennmatthews:
### Environment
* Nautobot version (Docker tag too if applicable): next
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce

Example: https://next.demo.nautobot.com/extras/object-changes/ce20d17c-ed11-455b-93a3-d7efc59594be/

### Expected Behavior

Page to render with some default layout

### Observed Behavior

![image](https://github.com/nautobot/nautobot/assets/5603551/3308c271-451f-4dd9-9a58-b6cff2fbe72a)

![image](https://github.com/nautobot/nautobot/assets/5603551/f9b73ce4-72c3-4f55-9e17-1c9b9b512723)

glennmatthews:
The underlying trigger appears to be that `ObjectChangeViewSet` is not a `NautobotModelViewSet` subclass and so it doesn't implement the `detail-view-config` API action. Regardless, the front-end needs to account for this possibility.

HanlinMiao:
Should this be blocked by https://github.com/nautobot/nautobot/issues/3526?

HanlinMiao:
Closed by #3805

Issue #3526, opened 2023-04-03
UI 2.0: /api/ui/ Endpoint for Detail View Layout


bryanculver:
For the 2.0 UI, data will be retrieved from the existing REST API endpoints. However, the UI will need to have some way to determine where the attributes should live in a section on the view.

For example here: 
<img width="2036" alt="image" src="https://user-images.githubusercontent.com/31187/229576891-8673f85f-c725-4878-8903-4d0f1b305d0d.png">

The UI will need to know the grouping, ordering, and potentially tab location for what is returned from the REST API response.

This should be viewed as the 80% case (models without heavy view customization) as anything sufficiently more nuanced will need to register its own view.

Models that don't define their own, should still get a response from this endpoint that is fairly generic (all models fields under one section, things like tags can be broken out). Things like Status, Created, Last Updated appears in the title bar and are special cases. API response should inform you of those.

## TODO
- [ ] Design and define the convention, and provide `Device` as an example that overrides the default

## Follow-on:
- Update views to take advantage of this endpoint exists (currently would be blocked by the base views and presentation not existing at the moment)
- Create backlog for defining overloaded model view definitions
- Need to address/be able to extend this endpoint for the related objects on detail views (Images, Services, etc from above)

bryanculver:
Implemented with #3816

PR #3805, opened 2023-05-24
Enhanced error handling for fetcher in React UI `ObjectRetrieveView`.

HanlinMiao:
# Closes: #3767 
# What's Changed

When `/detail-view-config` is loaded correctly.
Screen Shot 2023-05-25 at 10 55 07 PM

When `/detail-view-config` is not loaded correctly.
Screen Shot 2023-05-25 at 10 55 36 PM

# TODO

- [ ] Unit, Integration Tests


nautobot/core/api/views.py
```diff
@@ -236,9 +238,9 @@ def restrict_queryset(self, request, *args, **kwargs):
         """
         # Restrict the view's QuerySet to allow only the permitted objects for the given user, if applicable
         if request.user.is_authenticated:
-            action = HTTP_ACTIONS[request.method]
-            if action:
-                self.queryset = self.queryset.restrict(request.user, action)
+            http_action = HTTP_ACTIONS[request.method]
```
HanlinMiao, line 241: pylint is not happy with `action` here because it thinks it is a rename for the `@action` decorator

nautobot/ui/src/views/generic/ObjectRetrieve.js
```diff
@@ -175,18 +181,22 @@ export default function ObjectRetrieve({ api_url }) {
...
     const ui_url = objectData?.url
-        ? `${objectData.url}detail-view-config/`
+        ? `${objectData.url}detail-view-onfig/`
         : null;
-    var { data: appConfig } = useSWR(() => ui_url, fetcher);
+    const {
+        data: appConfig,
+        error: appConfigError,
+        isLoading: appConfigLoading,
+    } = useSWR(() => ui_url, fetcher);
```
glennmatthews, line 194: Can/should we be using `getRESTAPI` instead of `useSWR` here? Since it was already recently updated to handle 4xx errors itself, it feels weird to be having to add equivalent logic in this file instead of just using that. What am I missing? Is this something that we can't change until we do #3526 and move it to an OPTIONS query?

nautobot/ui/src/views/generic/ObjectRetrieve.js
```diff
@@ -175,18 +181,22 @@ export default function ObjectRetrieve({ api_url }) {
     // Object Data
     const {
         data: objectData,
-        isError: error,
+        error: objectDataError,
         isLoading: objectDataLoading,
     } = useSWR(() => api_url, fetcher);
     const ui_url = objectData?.url
-        ? `${objectData.url}detail-view-config/`
+        ? `${objectData.url}detail-view-onfig/`
```
glennmatthews, line 188: revert?

nautobot/ui/src/views/generic/ObjectRetrieve.js
```diff
@@ -305,74 +320,119 @@ export default function ObjectRetrieve({ api_url }) {
...
+                                                            />
+                                                        ))}
+                                                    </Tbody>
+                                                </Table>
+                                            </TableContainer>
+                                        </NautobotGridItem>
+                                    </NautobotGrid>
+                                </Card>
+                            </TabPanel>
+                        ) : (
```
glennmatthews, line 387: Instead of these massive ternary statements, can we extract some of this logic/structure into a few helper functions?

nautobot/ui/src/views/generic/ObjectRetrieve.js
HanlinMiao: Yeah that and also `useGetRESTAPIQuery` right now does not have the capability of adding url `query params` or just straight out specifying an `url` to get the information from. So we are still using `useSWR()` here for `DetailViewConfig`, `ChangeLog` and `Notes`.

nautobot/ui/src/views/generic/ObjectRetrieve.js
```diff
@@ -164,6 +164,134 @@ function RenderRow(props) {
...
+        );
+    return tabs;
+}
+
+function construct_detail_view_tabpanels({ isAvailable, configuration, data }) {
+    {
+        /* If instance data is available, but detail view layout endpoint is not loaded correctly */
+    }
+    let tabpanels =
+        !isAvailable && data ? (
```
glennmatthews, line 187: Can't this instead become something like:

```
if (!isAvailable && data) {
    return <TabPanel....
} else {
    return Object.keys(configuration).map((tab, idx) => ...
}

glennmatthews (approved):

PR #3814, opened 2023-05-26
Fix #3586 - change REST API config_context to opt-in

### LINKED ISSUE #3586: Change REST API config_context from opt-out to opt-in
### Proposed Changes

Instead of `config_context` being included by default on Device and VirtualMachine REST API endpoints unless you specify `?exclude=config_context`, let's change it so that (like other expensive fields such as `computed_fields` and `relationships`) the config context is *excluded by default* unless you specify `?include=config_context`.

### Justification

Avoid a common API performance pitfall.

glennmatthews:
# Closes: #3586 
# What's Changed

- Remove support for `exclude=config_context` REST API query parameter.
- Replace it with treating `config_context` as an opt-in field (same as `computed_fields` and `relationships`), i.e. omit it by default but allow it to be requested by specifying `include=config_context` REST API query parameter.
- Update docs and tests accordingly.

# TODO
- [x] Unit, Integration Tests


bryanculver (approved):
If the tests pass here there's a minimal amount changed. I'd like to see this and our other improvements for the next beta in the demo environment.

PR #3816, opened 2023-05-29
Detail View Config

timizuoebideri1:
# Closes: #3526
# What's Changed

# TODO

- [ ] Unit, Integration Tests


nautobot/core/api/metadata.py
```diff
@@ -71,6 +74,14 @@ def _get_field_properties(self, field: schema.SerializerType, name: str) -> Dict
...
             result["items"] = self._get_field_properties(field.child_relation, "")
             result["uniqueItems"] = True
+        elif isinstance(field, NautobotHyperlinkedRelatedField):
+            result["format"] = type_map_obj["format"]
+            model_options = field.queryset.model._meta
+            # Custom Keyword: modelName and appName
+            # This Keyword represents the model name of the uuid model
+            # and appName represents the app_name of the model
+            result["modelName"] = model_options.model_name
+            result["appName"] = model_options.app_label
```
glennmatthews, line 84: Do `modelName` and `appName`/`appLabel` need to also be added to the `ManyRelatedField` case?

nautobot/dcim/api/serializers.py
```diff
@@ -199,6 +199,18 @@ class Meta:
...
+                "Location": {"fields": ["name", "display", "rack_count", "parent"]},
+                "Management": {"fields": ["status", "prefix_count", "device_count"]},
+                "Others": {"fields": ["tenant", "natural_key_slug", "url"]},
+            },
+            {
+                "Tags": {"fields": ["tags"]},
+                "Comments": {"fields": ["comments"]},
+                "Contact": {"fields": ["contact_name", "contact_phone", "contact_email"]},
+            },
+        ]
```
glennmatthews, line 213: We need to be sure to document the schema for this attribute in the docs.

nautobot/dcim/api/serializers.py
```diff
@@ -199,6 +199,18 @@ class Meta:
...
+            {
+                "Location": {"fields": ["name", "display", "rack_count", "parent"]},
+                "Management": {"fields": ["status", "prefix_count", "device_count"]},
+                "Others": {"fields": ["tenant", "natural_key_slug", "url"]},
+            },
+            {
+                "Tags": {"fields": ["tags"]},
+                "Comments": {"fields": ["comments"]},
+                "Contact": {"fields": ["contact_name", "contact_phone", "contact_email"]},
+            },
```
glennmatthews, line 212: Shouldn't at least some of these have `template: list`?

nautobot/dcim/api/serializers.py
```diff
@@ -199,6 +199,18 @@ class Meta:
         list_display_fields = ["name", "status", "parent", "tenant", "description", "tags"]
         # https://www.django-rest-framework.org/api-guide/validators/#optional-fields
         validators = []
+        detail_view_config = [
+            {
+                "Location": {"fields": ["name", "display", "rack_count", "parent"]},
+                "Management": {"fields": ["status", "prefix_count", "device_count"]},
+                "Others": {"fields": ["tenant", "natural_key_slug", "url"]},
```
glennmatthews, line 206: We should think about whether we can have a pattern for "standard" fields to not need to be explicitly called out in the detail_view_config (i.e. `natural_key_slug`, `url`, `display`, `status` should always be handled consistently, so it's redundant and error-prone to have to call them out in every serializer).

timizuoebideri1: Excellent idea, but where would these fields be located?
I'm presuming it's typically on Col 1 Group 1?

nautobot/dcim/api/serializers.py
timizuoebideri1: As of now, just the list template can be defined; I don't see the need in explicitly declaring it again.

nautobot/core/api/metadata.py
```diff
@@ -256,11 +268,18 @@ def determine_metadata(self, request, view):
...
             metadata["view_options"] = self.determine_view_options(request, serializer)
 
         return metadata
+
+    def determine_detail_view_schema(self, serializer):
+        view_config = getattr(
+            serializer.Meta,
+            "detail_view_config",
+            SerializerDetailViewConfig(serializer).view_config(),
+        )
```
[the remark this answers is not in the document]
timizuoebideri1: Not sure i understand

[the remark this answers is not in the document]
glennmatthews:
```suggestion
        if hasattr(serializer.Meta, "detail_view_config"):
            view_config = serializer.Meta.detail_view_config
        else:
            view_config = SerializerDetailViewConfig(serializer).view_config()
```

nautobot/dcim/api/serializers.py
glennmatthews: It's going to depend on the field:

- Status and Display appear in the page header already and probably can be simply ignored in the detail_view_config
- ~natural key slug~ ("composite key") and URL and ID probably should all appear at the end of column 1 group 1.

nautobot/dcim/api/serializers.py
glennmatthews: OK, just seems asymmetrical to me that `view_config()` returns a list of dicts with `fields` and `template` as keys, while a statically declared `detail_view_config` only contains `fields`?

nautobot/ui/src/utils/url.js
```diff
@@ -2,3 +2,15 @@
 export function uiUrl(apiUrl) {
     return apiUrl.replace("/api/", "/");
 }
+
+/* Build url form app_label, model_name and or pk */
+export function buildUrl(appLabel, modelName, pk = null, isPlugin = false) {
+    let url = `/${appLabel}/${modelName}/`;
+    if (isPlugin) {
+        url = "/plugin" + url;
```
glennmatthews, line 10:
```suggestion
        url = "/plugins" + url;
```

nautobot/core/api/metadata.py
timizuoebideri1: `ManyRelatedField` would recursively call the `_get_field_properties`, which would end up here `elif isinstance(field, drf_serializers.RelatedField):`

nautobot/ui/src/components/ObjectDetail/RenderChangeLogsTab.js
```diff
@@ -0,0 +1,64 @@
+import useSWR from "swr";
```
HanlinMiao, line 1: NOICE

nautobot/ui/src/components/Table.js
```diff
@@ -1,48 +1,99 @@
...
+            size="xs"
+            variant="table"
+            onClick={() => alert(`Clicked ${cellContext.row.original.name}!`)}
+        />
+    );
+}
+
 // A standard Nautobot table. This _may_ be beneficial to move into nautobot-ui
-export default function NautobotTable({ data, headers }) {
+export default function NautobotTable({
```
HanlinMiao, line 59: Sorry for bothering you with this but can I get some comments on what some of the react hooks are doing?

timizuoebideri1: Im attaching link to the documentation for each hook:

1. [useMemo](https://react.dev/reference/react/useMemo)
2. [useCallback](https://react.dev/reference/react/useCallback)

Its basically a copy paste code from what was in ObjectRetrieve.js initially

nautobot/dcim/api/serializers.py
timizuoebideri1: Do you have any suggestions about where this document should be added?

nautobot/core/testing/api.py
```diff
@@ -862,6 +862,11 @@ def test_options_returns_expected_choices(self):
                         field_choices.add(field_name)
             else:
                 self.fail(f"Neither PUT nor POST are available actions in: {data['actions']}")
+                
+            print("========")
+            print(set(self.choices_fields))
+            print(field_choices)
+            print("========")
```
bryanculver, line 869:
```suggestion
```

nautobot/ui/src/components/ObjectDetail/RenderCol.js
```diff
@@ -0,0 +1,25 @@
+import { NautobotGrid, NautobotGridItem } from "@nautobot/nautobot-ui";
+
+import RenderGroup from "./RenderGroup";
+
+function RenderCol({ tabData, data, schema }) {
```
bryanculver, line 5: I think many of these are good enough to be used generically. Maybe `components/render` instead of `components/ObjectDetail`?

nautobot/ui/src/views/generic/ObjectRetrieve.js
```diff
@@ -362,190 +44,45 @@ export default function ObjectRetrieve({ api_url }) {
...
+        tabs: {
+            Location: schemaData.view_options.retrieve,
+        },
+    };
+
+    return (
+        <GenericView objectData={data}>
+            <Box background="white-0" borderRadius="md">
+                <RenderHeader data={data} />
+                <RenderTabs
```
bryanculver, line 76: Much better 👏 

Let's update the names to be more descriptive, something like `RetreiveViewHeader` and `RetreiveViewTabs`.

bryanculver (commented):
Loving the direction this is going in multiple ways. Some PR feedback, ensure tests are passing, then I'd say this is GTG.

bryanculver:
@timizuoebideri1 I pulled this down and want to know if you're seeing this in your environment. Looks like `id`, `natural_key_slug`, and `url` are being added multiple times.


image

image

nautobot/core/api/utils.py
```diff
@@ -349,3 +350,71 @@ def return_nested_serializer_data_based_on_depth(serializer, depth, obj, obj_rel
...
+            ]
+
+        Returns:
+            A list representing the view config.
+        """
+        m2m_fields, other_fields = self.get_m2m_and_non_m2m_fields()
+        model_verbose_name = self.serializer.Meta.model._meta.verbose_name
+        return [
+            {
+                model_verbose_name.capitalize(): {
```
bryanculver, line 415: `verbose_name` should already properly capitilized: https://github.com/nautobot/nautobot/blob/496ce1000c8c5cb5962dbd58ca2ed190fbeaf028/nautobot/ipam/models.py#L102

Re-capitalizing this is causing the names to appear incorrect.

nautobot/core/api/utils.py
```diff
@@ -386,6 +386,10 @@ def get_m2m_and_non_m2m_fields(self):
 
         return m2m_fields, non_m2m_fields
 
+    def capitalize(self, value: str):
+        """Capitalize value only if not uppercase."""
+        return value if (value.isupper()) else value.capitalize()
```
bryanculver, line 391: This won't work as well for cases like `IP Address`

bryanculver (changes requested):

nautobot/core/api/utils.py
glennmatthews: Looks like my comment on the commit itself isn't showing up here - should be able to use the existing `bettertitle` function for this purpose.

nautobot/dcim/api/serializers.py
```diff
@@ -199,6 +199,19 @@ class Meta:
         list_display_fields = ["name", "status", "parent", "tenant", "description", "tags"]
         # https://www.django-rest-framework.org/api-guide/validators/#optional-fields
         validators = []
+        # TODO: Add documentation of this schema
+        detail_view_config = [
+            {
+                "Location": {"fields": ["name", "rack_count", "parent"]},
+                "Management": {"fields": ["status", "prefix_count", "device_count"]},
```
glennmatthews, line 206: `"status"` isn't necessary here right?

glennmatthews, line 206: why is `rack_count` in `Location` but `prefix_count` and `device_count` in Management? Any specific reasoning?

nautobot/dcim/api/serializers.py
```diff
@@ -199,6 +199,19 @@ class Meta:
...
+                "Others": {"fields": ["tenant"]},
+            },
+            {
+                "Tags": {"fields": ["tags"]},
```
glennmatthews, line 210: As we move toward more and more models having `tenant` and `tags` fields, it'd be good to have these handled in a standardized way too, I think?

nautobot/ui/src/components/Retrieve/RenderChangeLogsTab.js
```diff
@@ -0,0 +1,64 @@
...
+    const defaultHeaders = schema.view_options.list_display_fields;
+    const tableHeaders = schema.view_options.fields;
+    const tableData = data?.results || [];
+
+    return (
+        <NautobotGrid columns="4">
+            <NautobotGridItem>
+                <Heading display="flex" alignItems="center" gap="5px">
+                    <NtcThumbnailIcon width="25px" height="30px" /> Change Logs
+                </Heading>
```
glennmatthews, line 55: Not blocking, but it might be useful to create a reusable component for this `<Heading ...><NtcThumbnailIcon/> {title} </Heading>` pattern since you're using it in at least three places in this PR alone.

nautobot/ui/src/components/RetrieveViewHeader.js
glennmatthews, line 1: Should this be `components/Retrieve/ViewHeader.js`? Not clear to me the rationale why some components are in the subdirectory and others aren't.

nautobot/ui/src/components/RetrieveViewTabs.js
glennmatthews, line 1: Same question here - `components/Retrieve/ViewTabs.js`?

nautobot/ui/src/components/Table.js
glennmatthews, line 1: Not blocking but it would be worth thinking about the name of our `Table` and `TableItem` components, as they're a bit confusing against the `nautobot-ui/Table` component and I don't *think* they're meant as generic components - they have a fairly specific set of applications, don't they?

nautobot/core/api/metadata.py
```diff
@@ -72,41 +76,47 @@ def _get_field_properties(self, field: schema.SerializerType, name: str) -> Dict
...
         return result
 
+    @staticmethod
+    def _filter_fields(all_fields):
+        """
+        Override super._filter_fields to return all fields, including read-only fields,
+        as read-only fields have to be displayed in Detail View.
+        """
+        return tuple((name, field) for name, field in all_fields)
+
```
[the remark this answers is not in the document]
glennmatthews: I *think* this results in read-only fields now appearing in the create/edit forms by default, which seems undesirable?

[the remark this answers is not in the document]
timizuoebideri1: In the creat/edit field we could hide read-only fields

nautobot/core/api/metadata.py
```diff
@@ -256,11 +266,56 @@ def determine_metadata(self, request, view):
...
+                    value["fields"].extend(fields_to_add)
+        return view_config
+
+    def determine_detail_view_schema(self, serializer):
+        """Determine the layout option that would be used for the detail view"""
+        if hasattr(serializer.Meta, "detail_view_config"):
+            view_config = serializer.Meta.detail_view_config
+        else:
+            view_config = SerializerDetailViewConfig(serializer).view_config()
+        return self.restructure_view_config(view_config)
```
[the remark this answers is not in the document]
timizuoebideri1: According to our discussions, the idea was to leave it as a json schema for the time being.

nautobot/dcim/api/serializers.py
timizuoebideri1: Nope, Could not think of a better grouping,
Illd use the grouping on the legacy UI 👍

nautobot/ui/src/components/Table.js
timizuoebideri1: It could be used as a generic Table component, for ChangeLogs, Notes, or in the ListView.

nautobot/core/api/metadata.py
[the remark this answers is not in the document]
glennmatthews: I'm not suggesting changing the schema, just how it's implemented in code.

nautobot/ui/src/components/Table.js
glennmatthews: So maybe something like `ObjectTable` or `RecordTable`?

nautobot/core/api/metadata.py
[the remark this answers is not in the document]
timizuoebideri1: Nope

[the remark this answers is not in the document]
timizuoebideri1: It's in the schema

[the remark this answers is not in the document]
timizuoebideri1: The schema tells you the field type to employ, such as CharField, DynamicSelect, and so on. At that point, you can determine if a field is readonly and hide it if needed.

nautobot/core/api/metadata.py
```diff
@@ -72,41 +77,47 @@ def _get_field_properties(self, field: schema.SerializerType, name: str) -> Dict
...
+            if isinstance(field, drf_serializers.RelatedField):
+                result["uniqueItems"] = True
+                if hasattr(field, "queryset") and hasattr(field.queryset, "model"):
+                    model_options = field.queryset.model._meta
+                    result["required"] = field.required
+                    # Custom Keyword: modelName and appLabel
+                    # This Keyword represents the model name of the uuid model
+                    # and appLabel represents the app_name of the model
+                    result["modelName"] = model_options.model_name
+                    result["appLabel"] = model_options.app_label
```
glennmatthews, line 89: Do we need to add these keys for the `ManyRelatedField` case above as well?

nautobot/core/api/metadata.py
```diff
@@ -72,41 +77,47 @@ def _get_field_properties(self, field: schema.SerializerType, name: str) -> Dict
...
-            if format_:
+                    result["enum"] = [item[0] for item in enum]
+                    result["enumNames"] = [item[1] for item in enum]
+            if format_ := type_map_obj["format"]:
                 result["format"] = format_
 
-            # Process "readOnly"
-            read_only = type_map_obj["readOnly"]
-            if read_only:
+            if read_only := field.read_only:
```
glennmatthews, line 103: Why the change from `type_map_obj["readOnly"]` to `field.read_only`?

nautobot/core/api/metadata.py
```diff
@@ -256,11 +267,111 @@ def determine_metadata(self, request, view):
...
+                    Location: {fields: ["name","id","natural_key_slug","url"]},
+                    Others: {fields: ["tenant", "tenant_group"]}
+                },
+                ...
+            ]
+        """
+
+        # TODO(timizuo): Add a standardized way of handling `tenant` and `tags` fields, Possible should be on last items on second col.
+        fields_to_remove = ["natural_key_slug", "url", "display", "status", "id"]
+        fields_to_add = ["id", "natural_key_slug", "url"]
```
glennmatthews, line 306: This will of course conflict with #3881.

nautobot/core/api/metadata.py
```diff
@@ -256,11 +267,111 @@ def determine_metadata(self, request, view):
...
-                    # "uiSchema": NautobotUiSchemaProcessor(serializer, request.parser_context).get_ui_schema(),
                     "uiSchema": ui_schema,
                 }
             )
 
             metadata["view_options"] = self.determine_view_options(request, serializer)
 
         return metadata
+
+    def restructure_view_config(self, view_config):
```
glennmatthews, line 278: Could you please add some unit tests for the functionality added in this PR?

nautobot/dcim/api/serializers.py
```diff
@@ -200,6 +200,21 @@ class Meta:
...
         validators = []
 
+        detail_view_config = [
+            {
+                "Location": {
+                    "fields": ["name", "location_type", "tenant", "description", "facility", "asn", "time_zone"]
+                },
+                "Contact Info": {
+                    "fields": ["physical_address", "shipping_address", "contact_name", "contact_phone", "contact_email"]
+                },
```
glennmatthews, line 210: Missing `latitude` and `longitude` fields. Is there anything we can add in the detail_view_config processing to automatically detect if fields are missing from the config?

nautobot/ui/src/components/ObjectTable.js
```diff
@@ -0,0 +1,99 @@
...
+                const column_data =
+                    idx === 0
+                        ? props.row.original
+                        : props.row.original[props.column.id];
+                return (
+                    <ObjectTableItem
+                        name={key}
+                        obj={props.getValue()}
+                        url={getTableItemLink(idx, column_data)}
+                    />
```
glennmatthews, line 39: Is the comment inaccurate here? I don't see how `column_data = {"display": "Active"}` then leads to a useful `getTableItemLink(idx, {"display": "Active"})` call when it seems like `getTableItemLink()` expects an object with an `id` or `url` key.

nautobot/ui/src/components/ObjectTableItem.js
```diff
@@ -58,8 +58,9 @@ function TableColumnDisplay({ obj }) {
     }
 }
 
-export default function TableItem({ name, obj, url }) {
-    if (url) {
+// TODO(timizuo): create a ObjectTableItem folder and move ObjectTable, ObjectTableItem and ObjectListTable to the folder.
```
glennmatthews, line 61: Still a TODO? If so why not do it now?

nautobot/ui/src/components/Retrieve/RenderChangeLogsTab.js
```diff
@@ -0,0 +1,64 @@
...
+    if (error || schemaIsError) {
+        return <div>Failed to load Change Log Data</div>;
+    }
+
+    const defaultHeaders = schema.view_options.list_display_fields;
+    const tableHeaders = schema.view_options.fields;
+    const tableData = data?.results || [];
+
+    return (
+        <NautobotGrid columns="4">
```
glennmatthews, line 51: why `columns="4"` when there's only a single `NautobotGridItem`?

nautobot/ui/src/components/Retrieve/RenderNotesTab.js
```diff
@@ -0,0 +1,66 @@
...
+    if (error || schemaIsError) {
+        return <div>Failed to load Notes Data</div>;
+    }
+
+    const defaultHeaders = schema.view_options.list_display_fields;
+    const tableHeaders = schema.view_options.fields;
+    const tableData = data?.results || [];
+
+    return (
+        <NautobotGrid columns="4">
```
glennmatthews, line 53: Same question here re `columns="4"` with a single NautobotGridItem?

nautobot/dcim/api/serializers.py
timizuoebideri1: What of the user do not want to show all model fields?

nautobot/core/api/metadata.py
timizuoebideri1: Nope, as ManyRelated field, would is a recursive call, and would end up getting to this point

nautobot/core/api/metadata.py
```diff
@@ -56,10 +60,11 @@ def _get_field_properties(self, field: schema.SerializerType, name: str) -> Dict
 
         This has been overloaded with an `elif` to account for `ManyRelatedField`.
         """
-        result = {}
         type_map_obj = self._get_type_map_value(field)
-        result["type"] = type_map_obj["type"]
-        result["title"] = self._get_title(field, name)
+        result = {
+            "type": type_map_obj["type"] or "string",
```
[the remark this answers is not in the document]
timizuoebideri1: For some readonly fields, previously readonly fields where not part of schema.

nautobot/core/api/metadata.py
timizuoebideri1: Its not always accurate: below response of `print("====> " + name, (type_map_obj["readOnly"], field.read_only))`

```bash
====> id (None, True)
====> object_type (None, True)
====> display (None, True)
====> url (None, True)
====> natural_key_slug (None, True)
====> tree_depth (None, True)
====> time_zone (None, False)
====> circuit_count (None, True)
====> device_count (None, True)
====> name (None, False)
====> slug (None, False)
====> description (None, False)
====> facility (None, False)
====> asn (None, False)
```

nautobot/core/api/metadata.py
timizuoebideri1: Its now `composite_key`?

nautobot/ui/src/components/Retrieve/RenderChangeLogsTab.js
timizuoebideri1: I'm not sure why, but if I don't set the column, the table doesn't take up the entire width.

nautobot/core/api/metadata.py
glennmatthews: Yes.

nautobot/dcim/api/serializers.py
glennmatthews: Let's discuss this in standup tomorrow then. I think right now it's too easy to inadvertently omit fields that should be present.

nautobot/core/api/metadata.py
jathanism: I had added this and higher up being processed by `_get_type_map_value()`, which was relying on it being set in the schema somehow, which explains why it wasn't always accurate.

jathanism (changes requested):
Minor asks.

nautobot/core/testing/api.py
```diff
@@ -191,6 +191,37 @@ def test_options_object(self):
             response = self.client.options(url, **self.header)
             self.assertHttpStatus(response, status.HTTP_200_OK)
 
+        @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
+        def test_options_object_detail_view_config(self):
```
glennmatthews, line 195: Feels like this and the above `test_options_object` could be merged into a single test?

nautobot/core/api/metadata.py
```diff
@@ -256,11 +265,111 @@ def determine_metadata(self, request, view):
...
+            ]
+        """
+
+        # TODO(timizuo): Add a standardized way of handling `tenant` and `tags` fields, Possible should be on last items on second col.
+        fields_to_remove = ["composite_key", "url", "display", "status", "id"]
+        fields_to_add = ["id", "composite_key", "url"]
+
+        for section_idx, section in enumerate(view_config):
+            for idx, data in enumerate(section.items()):
+                _, value = data
```
gsnider2195, line 308:
```suggestion
            for idx, value in enumerate(section.values()):
```

nautobot/ui/src/components/ObjectTable/index.js
```diff
@@ -0,0 +1,3 @@
+export { default as ObjectListTable } from "./ObjectListTable"
+export { default as ObjectTable } from "./ObjectTable"
+export {default as ObjectTableItem } from "./ObjectTableItem"
```
gsnider2195, line 3:
```suggestion
export { default as ObjectTableItem } from "./ObjectTableItem"
```

nautobot/ui/src/utils/url.js
```diff
@@ -2,3 +2,15 @@
 export function uiUrl(apiUrl) {
     return apiUrl.replace("/api/", "/");
 }
+
+/* Build url form app_label, model_name and or pk */
```
gsnider2195, line 6:
```suggestion
/* Build url from app_label, model_name and optional pk */
```

nautobot/ui/src/components/ObjectTable/ObjectListTable.js
```diff
@@ -16,11 +16,12 @@ import {
...
     useToast,
 } from "@nautobot/nautobot-ui";
-import Paginator from "@components/paginator";
 import { useCallback, useMemo } from "react";
 
-import TableItem from "@components/TableItem";
-import LoadingWidget from "./LoadingWidget";
+import Paginator from "../paginator";
+import ObjectTableItem from "./ObjectTableItem";
+import LoadingWidget from "../LoadingWidget";
```
HanlinMiao, line 23: Probably should alpha-sort these

nautobot/ui/src/utils/url.js
```diff
@@ -2,3 +2,15 @@
 export function uiUrl(apiUrl) {
     return apiUrl.replace("/api/", "/");
 }
+
+/* Build url from app_label, model_name and optional pk */
+export function buildUrl(appLabel, modelName, pk = null, isPlugin = false) {
```
HanlinMiao, line 7: Should this function account for the few `slugs` that still exist in the url after https://github.com/nautobot/nautobot/pull/3914?

HanlinMiao (approved):
I pulled it down and it looks noice! Just some minor changes.

nautobot/ui/src/components/Retrieve/RenderCol.js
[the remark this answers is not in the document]
timizuoebideri1: 🦥

jathanism (approved):
I'm good to let this go, I just want to think about readability and consistency. :)

nautobot/core/testing/api.py
```diff
@@ -191,6 +191,37 @@ def test_options_object(self):
...
+                    # Assert specific fields(`id`, `composite_key`, `url`) are present only at the end of the first col group fields.
+                    special_fields = ["id", "composite_key", "url"]
+                    for col_idx, col in enumerate(response_view_config):
+                        for group_idx, (group_title, group) in enumerate(col.items()):
+                            group_fields = group["fields"]
+                            # Config set on the serializer
+                            fields = detail_view_config[col_idx][group_title]["fields"]
+                            if group_idx == 0 == col_idx:
+                                self.assertEqual(special_fields, group_fields[-3:])
+                                self.assertFalse(any(field in special_fields for field in group_fields[:-3]))
```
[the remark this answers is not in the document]
timizuoebideri1: I'm asserting that special_fields only exists at the end of the group fields and that it doesn't exist anywhere else in the group fields.

Group fields are the serializer fields that should be present in a particular column group

nautobot/ui/src/utils/url.js
timizuoebideri1: Not really, if a slug is passed in to the buildUrl it would still work as expected.
Do you have any in mind, just to verify

bryanculver (approved):

nautobot/docs/development/react-ui.md
```diff
@@ -44,6 +44,38 @@ The libraries and tools used to build the React UI are:
...
+                    "Group Name 3": {"fields": [...]},
+                    "Group Name 4": {"fields": [...]},
+                },
+            ],
+            "include_others": False
+        }
+```
+
+In the above example, we add the `detail_view_config` attribute to the Serializer's inner `Meta` class. The value of this attribute is a dict containing `layout` and `include_others`(optional). the `layout` is a list containing two dictionaries, each representing the two columns of the detail view. The first dictionary represents the fields in the first column, while the second dictionary represents the fields in the second column. Each dictionary consists of a key-value pair, where the key is the name of the grouping, and the value is a list of the model fields that should be included in that grouping.
+The optional key `include_others`, when set to `True`, adds missing serializer fields from the `detail_view_config["layout"]` to the view_config layout.
```
glennmatthews, line 72: I'd still like us to revisit this in the near future - my feeling is the number of cases where we want to *omit* fields from the UI is the minority, and in the vast majority of cases we'll want every serializer field to be represented in the UI.

nautobot/ui/src/components/__tests__/Table.test.js
```diff
@@ -1,36 +1,39 @@
...
+// const mockHeaders = [
+//     { key: "name", dataIndex: "name", title: "Name" },
+//     { key: "description", dataIndex: "description", title: "Description" },
+// ];
+// const mockData = [
+//     { id: 1, name: "Item 1", description: "Description 1" },
+//     { id: 2, name: "Item 2", description: "Description 2" },
+// ];
 
+// FixMe(timizuo): Cannot test ObjectTable as it implements @nautobot-ui TableRenderer component which
```
glennmatthews, line 14: Is this test going to be addressed in a follow-on issue, or as part of this PR?

glennmatthews (approved):
I still have some reservations, but let's get this merged and iterate on it.

PR #3818, opened 2023-05-30
jobs overhaul run_task docs updates

gsnider2195:
# Closes: #DNE
# What's Changed

# TODO

- [ ] Unit, Integration Tests


nautobot/docs/additional-features/jobs.md
```diff
@@ -377,7 +376,7 @@ device = ObjectVar(
 )
 ```
 
-Multiple values can be specified by assigning a list to the dictionary key. It is also possible to reference the value of other fields in the form by prepending a dollar sign (`$`) to the variable's name. The keys you can use in this dictionary are the same ones that are available in the REST API - as an example it is also possible to filter the `Site` `ObjectVar` for its `tenant_group`.
+Multiple values can be specified by assigning a list to the dictionary key. It is also possible to reference the value of other fields in the form by prepending a dollar sign (`$`) to the variable's name. The keys you can use in this dictionary are the same ones that are available in the REST API - as an example it is also possible to filter the `Location` `ObjectVar` for its `location_type` and `tenant_group`.
 
 ```python
 region = ObjectVar(
```
glennmatthews, line 382: Based on the example, probably need to replace this with a `location_type` ObjectVar too.

nautobot/docs/additional-features/jobs.md
```diff
@@ -676,12 +680,12 @@ While individual methods within your Job can and should be tested in isolation,
...
 
 +/- 2.0.0
     `run_job_for_testing` was moved from the `nautobot.utilities.testing` module to `nautobot.core.testing`.
 
-Because of the way `run_job_for_testing` and more specifically `run_job()` works, which is somewhat complex behind the scenes, you need to inherit from `nautobot.core.testing.TransactionTestCase` instead of `django.test.TestCase` (Refer to the [Django documentation](https://docs.djangoproject.com/en/stable/topics/testing/tools/#provided-test-case-classes) if you're interested in the differences between these classes - `TransactionTestCase` from Nautobot is a small wrapper around Django's `TransactionTestCase`).
+Because of the way `run_job_for_testing` and more specifically celery tasks work, which is somewhat complex behind the scenes, you need to inherit from `nautobot.core.testing.TransactionTestCase` instead of `django.test.TestCase` (Refer to the [Django documentation](https://docs.djangoproject.com/en/stable/topics/testing/tools/#provided-test-case-classes) if you're interested in the differences between these classes - `TransactionTestCase` from Nautobot is a small wrapper around Django's `TransactionTestCase`).
```
glennmatthews, line 688:
```suggestion
Because of the way `run_job_for_testing` and more specifically Celery tasks work, which is somewhat complex behind the scenes, you need to inherit from `nautobot.core.testing.TransactionTestCase` instead of `django.test.TestCase` (Refer to the [Django documentation](https://docs.djangoproject.com/en/stable/topics/testing/tools/#provided-test-case-classes) if you're interested in the differences between these classes - `TransactionTestCase` from Nautobot is a small wrapper around Django's `TransactionTestCase`).
```

nautobot/docs/additional-features/jobs.md
```diff
@@ -808,90 +738,72 @@ This will print the 10 functions that the job execution spent the most time in -
...
-        STATUS_PLANNED = Status.objects.get(name='Planned')
+    def run(self, location_name, switch_count, switch_model):
+        STATUS_PLANNED = Status.objects.get(name="Planned")
 
-        # Create the new site
-        site = Site(
-            name=site_name,
-            slug=slugify(site_name),
+        # Create the new location
+        root_type = LocationType.objects.create(name="Campus")
```
glennmatthews, line 776: Possibly `get` or `get_or_create` instead?

glennmatthews (approved):

PR #3835, opened 2023-05-31
Cleanup Namespaces Detail, Related Views

bryanculver:
# Closes: #DNE
# What's Changed

- Update the Namespaces views to load with tabs for related objects

(probably should have been against the prototype branch but that doesn't have the working data migrations which made this easier to test and implement)

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/ipam/models.py
+++ b/nautobot/ipam/models.py
@@ -66,6 +66,11 @@ class Namespace(PrimaryModel):
         null=True,
     )
 
+    @property
+    def ip_addresses(self):
+        """Return all IPAddresses in this Namespace."""
+        return IPAddress.objects.filter(parent__namespace=self).distinct()
+
     class Meta:
         ordering = ("name",)
 
--- a/nautobot/ipam/models.py
+++ b/nautobot/ipam/models.py
@@ -68,7 +68,7 @@ class Namespace(PrimaryModel):
 
     @property
     def ip_addresses(self):
-        """Return all IPAddresses in this Namespace."""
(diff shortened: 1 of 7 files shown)


nautobot/ipam/models.py
```diff
@@ -66,6 +66,11 @@ class Namespace(PrimaryModel):
         null=True,
     )
 
+    @property
+    def ip_addresses(self):
+        """Return all IPAddresses in this Namespace."""
```
gsnider2195, line 71:
```suggestion
        """Return all IPAddresses associated to this Namespace through their parent Prefix."""
```

nautobot/ipam/views.py
```diff
@@ -52,17 +61,133 @@ def get_extra_context(self, request, instance):
...
+        context = super().get_extra_context(request, instance)
+        context.update(
+            {
+                "ip_address_table": ip_address_table,
+                "permissions": permissions,
+                "bulk_querystring": bulk_querystring,
+                "active_tab": "ip-addresses",
+            }
+        )
+        context.update(get_namespace_related_counts(instance, request))
```
gsnider2195, line 106: Why isn't this being performed by `super().get_extra_context`? Is it because of the `if self.action == "retrieve":` line in the super method?

gsnider2195 (approved):

HanlinMiao:
Migration Test strikes again Im sorry🫠

nautobot/ipam/views.py
bryanculver: Because these views don't inherit from the parent.

PR #3837, opened 2023-05-31
Ensure GitRepostory.slug is a valid Python package name

bryanculver:
# Closes: #DNE
# What's Changed

- Slug `GitRepository.slug` with `slugify_dashes_to_underscores`, same as the other places we use slugs for programmatic purposes
- Add test to ensure the slug is a valid Python package name
- Migration to convert existing repo slugs to valid package names

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/models/datasources.py
+++ b/nautobot/extras/models/datasources.py
@@ -2,6 +2,7 @@
 import os
 
 from django.conf import settings
+from django.core.exceptions import ValidationError
 from django.core.serializers.json import DjangoJSONEncoder
 from django.core.validators import URLValidator
 from django.db import models
@@ -84,7 +85,7 @@ class GitRepository(PrimaryModel):
             check_if_key_is_graphql_safe(self.__class__.__name__, self.slug, "slug")
 
             if self.present_in_database and self.slug != self.__initial_slug:
-                raise ValueError(
+                raise ValidationError(
                     f"Slug cannot be changed once set. Current slug is {self.__initial_slug}, "
                     f"requested slug is {self.slug}"
                 )
--- a/nautobot/extras/models/datasources.py
+++ b/nautobot/extras/models/datasources.py
@@ -6,7 +6,7 @@ from django.core.serializers.json import DjangoJSONEncoder
 from django.core.validators import URLValidator
 from django.db import models
 
-from nautobot.core.models.fields import AutoSlugField
+from nautobot.core.models.fields import AutoSlugField, slugify_dashes_to_underscores
 from nautobot.core.models.generics import PrimaryModel
 from nautobot.extras.utils import extras_features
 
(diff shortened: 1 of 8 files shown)


nautobot/extras/models/datasources.py
```diff
@@ -24,7 +24,11 @@ class GitRepository(PrimaryModel):
         max_length=100,
         unique=True,
     )
-    slug = AutoSlugField(populate_from="name")
+    slug = AutoSlugField(
+        populate_from="name",
+        help_text="Internal field name. Please use underscores rather than dashes in this key.",
+        slugify_function=slugify_dashes_to_underscores,
```
glennmatthews, line 30: Is this reflected in the UI or does the UI still default to dashes?

nautobot/extras/models/datasources.py
```diff
@@ -24,7 +24,11 @@ class GitRepository(PrimaryModel):
...
+    )
 
     remote_url = models.URLField(
         max_length=255,
```
glennmatthews, line 34: Comparing against #3426, isn't this missing `clean()` logic to enforce validation of the slug?

bryanculver: My bad, I assumed the field did the validation. Adding.

glennmatthews (commented):
Actually, we may also want to pull in the code from custom-fields that makes `slug` immutable once set without deleting/recreating the repository? Since changing it will affect import paths for jobs?

glennmatthews:
> > Actually, we may also want to pull in the code from custom-fields that makes `slug` immutable once set without deleting/recreating the repository? Since changing it will affect import paths for jobs?
> 
> What if I refactor my jobs? Do I have to delete and re-add the repo?

I wouldn't think so? I'm working on handling this in my current work.

> Also it would be pretty fragile to do full package name import since the author of the job may not have access to the GitRepository configuration.

Jobs shouldn't do absolute imports within a Git repository, but on the Nautobot side we have to have an absolute path to import the jobs from. That path *would* change when re-slugging a GitRepository.

glennmatthews (approved):

nautobot/project-static/js/forms.js
```diff
@@ -4,14 +4,16 @@
 
 // Slugify
 function slugify(s, num_chars) {
-    s = s.replace(/[^\-\.\w\s]/g, '');          // Remove unneeded chars
+    s = s.replace(/[^_A-Za-z0-9\-\s\.]/g, '');  // Remove non-ascii chars
     s = s.replace(/^[\s\.]+|[\s\.]+$/g, '');    // Trim leading/trailing spaces
     s = s.replace(/[\-\.\s]+/g, '-');           // Convert spaces and decimals to hyphens
```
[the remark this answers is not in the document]
bryanculver: Trying to closely match how the Python slugging is happening which does it in two steps: Replace multiple instances of these characters (in Django `slugify`), then replace the `-` with `_` in our `slugify_dashes_to_underscores`.

timizuoebideri1 (approved):

Issue #3465, opened 2023-03-21
IPAddress: Addresss `host` modifiable, `broadcast`, `prefix_length` removal


bryanculver:
See #1673 for context.

```[tasklist]
### Tasks
- [x] Make `host` not modifiable
- [x] Remove `broadcast`
- [x] Rename `prefix_length` to `mask_length`
- [ ] ~~`mask_length`, which is currently a derived attribute from `prefix_length`, should now be set at `IPAddressToInterface` M2M instance. (Suggest a default value for `mask_length` from the parent prefix's `prefix_length`. )~~
```

**Other:**
Data migrations: #3681
Unit tests: #3688

bryanculver:
For 2.0.0: Must remove the fields above. `host` modifiable may be easy enough to enforce in the clean method for now, but not required for 2.0.0.

HanlinMiao:
Closed by #3842

PR #3842, opened 2023-06-02
Make `IPAddress` `host` non-modifiable, remove `broadcast` field and rename `prefix_length` field to `mask_length` field.

HanlinMiao:
# Closes: #3465 
# What's Changed

1. Make `IPAddress` host non-modifiable.
Screen Shot 2023-06-02 at 1 35 13 PM
2. Remove `broadcast` field.
3. Rename `prefix_length` field to `mask_length` field.

# TODO

- [x] Unit, Integration Tests


bryanculver (commented):
Would be good to add here: https://docs.nautobot.com/projects/core/en/stable/models/ipam/ipaddress/

Something along the lines of:

```
+/- 2.0.0 

`prefix_length` becomes `mask_length` and is intended to describe the desired subnet mask when configured on interface(s).
```

changes/3465.removed
```diff
@@ -0,0 +1 @@
+Remove `broadcast` field from IPAddress.
```
glennmatthews, line 1:
```suggestion
Removed `broadcast` field from IPAddress.
```

nautobot/docs/installation/migrating-from-netbox.md
```diff
@@ -133,10 +133,10 @@ All IPAM objects with network field types (`ipam.IPAddress`, and `ipam.Prefix`)
 
 Below is a summary of the underlying technical changes to network fields. These will be explained in more detail in the following sections.
 
-- For `IPAddress`, the `address` field was exploded out to `host`, `broadcast`, and `prefix_length` fields; `address` was converted into a computed field.
+- For `IPAddress`, the `address` field was exploded out to `host`, and `mask_length` fields; `address` was converted into a computed field.
```
glennmatthews, line 136:
```suggestion
- For `IPAddress`, the `address` field was exploded out to `host` and `mask_length` fields; `address` was converted into a computed field.
```

nautobot/ipam/filters.py
```diff
@@ -404,7 +400,7 @@ class IPAddressFilterSet(
 
     class Meta:
         model = IPAddress
-        fields = ["id", "ip_version", "dns_name", "tags"]
+        fields = ["id", "ip_version", "dns_name", "tags", "mask_length"]
```
glennmatthews, line 403: This makes `mask_length` into a proper MultiValueNumberFilter right? Should that be documented in the migration tables?

nautobot/ipam/models.py
```diff
@@ -936,7 +937,7 @@ class IPAddress(PrimaryModel, StatusModel, RoleModelMixin):
     objects = BaseManager.from_queryset(IPAddressQuerySet)()
 
     class Meta:
-        ordering = ("ip_version", "host", "prefix_length")  # address may be non-unique
+        ordering = ("ip_version", "host", "mask_length")  # address may be non-unique
```
glennmatthews, line 940: Should `ordering` also include something like `parent__namespace`? Or are we OK with IPAddresses in different Namespaces being interleaved?

nautobot/ipam/models.py
```diff
@@ -969,10 +970,8 @@ def _deconstruct_address(self, address):
             #    We store this address as both the host and the "broadcast".
             # This variance is intentional in both cases as we use the "broadcast" primarily for filtering and grouping
             # of addresses and prefixes, not for packet forwarding. :-)
```
glennmatthews, line 972: Should this entire comment block be removed?

nautobot/ipam/tests/test_views.py
```diff
@@ -179,7 +179,7 @@ def test_empty_queryset(self):
             self.assertNotIn(prefix.get_absolute_url(), content, msg=content)
 
 
-@skip("Updated for Namespaces but still failing. Remove broadcast/prefix_length and revisit.")
+@skip("Updated for Namespaces but still failing. Remove broadcast and rename prefix_length to mask_length and revisit.")
```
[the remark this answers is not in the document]
HanlinMiao: reenabled tests with some updates.

nautobot/ipam/models.py
HanlinMiao: Yeah I think that makes sense. However, there is no way of seeing `parent__namespace` from `IPAddressListTable`. Would it make sense to add a column for that attribute?

nautobot/ipam/forms.py
```diff
@@ -603,11 +603,11 @@ class IPAddressFilterForm(NautobotFilterForm, TenancyFilterForm, StatusModelFilt
         label="IP version",
         widget=StaticSelect2(),
     )
-    mask_length = forms.ChoiceField(
+    mask_length = forms.MultipleChoiceField(
```
HanlinMiao, line 606: Change the field and widget attributes to account for `mask_length` filter becoming a `MultiValueNumberFilter`.

nautobot/core/testing/views.py
```diff
@@ -442,6 +443,10 @@ def test_edit_object_with_permission(self):
             # Try GET with model-level permission
             self.assertHttpStatus(self.client.get(self._get_url("edit", instance)), 200)
 
+            # We cannot edit IPAddress's host field once it is already set.
+            # but we do need the address field to be not empty
+            if isinstance(instance, IPAddress):
+                self.form_data["address"] = instance.address
```
glennmatthews, line 449: I'm not a big fan of having single-model-specific logic in the core generic tests. Can we solve this in a more flexible way? IIRC @bryanculver had some similar challenges in #3837 (see https://github.com/nautobot/nautobot/pull/3837/files#diff-63390a9316fbabcd9c87d7ea0980d9ff2738d5da7d7def31e8f152d5fcf92996R627) that were worked around by overloading these two test methods locally. That'd be my preferred approach for this PR; longer term I think we should look at borrowing the pattern that already exists in `nautobot.core.testing.api` where by default (as today) the same data is used for both create and edit tests, but you have the option to override one or the other if needed.

glennmatthews (changes requested):

nautobot/ipam/tests/test_querysets.py
```diff
@@ -22,13 +22,13 @@ def setUpTestData(cls):
...
         cls.ips = {
             "10.0.0.1/24": IPAddress.objects.create(address="10.0.0.1/24", namespace=cls.namespace, tenant=None),
-            "10.0.0.1/25": IPAddress.objects.create(address="10.0.0.1/25", namespace=cls.namespace2, tenant=None),
             "10.0.0.2/24": IPAddress.objects.create(address="10.0.0.2/24", namespace=cls.namespace, tenant=None),
             "10.0.0.3/24": IPAddress.objects.create(address="10.0.0.3/24", namespace=cls.namespace, tenant=None),
             "10.0.0.4/24": IPAddress.objects.create(address="10.0.0.4/24", namespace=cls.namespace, tenant=None),
             "2001:db8::1/64": IPAddress.objects.create(address="2001:db8::1/64", namespace=cls.namespace, tenant=None),
             "2001:db8::2/64": IPAddress.objects.create(address="2001:db8::2/64", namespace=cls.namespace, tenant=None),
             "2001:db8::3/64": IPAddress.objects.create(address="2001:db8::3/64", namespace=cls.namespace, tenant=None),
+            "10.0.0.1/25": IPAddress.objects.create(address="10.0.0.1/25", namespace=cls.namespace2, tenant=None),
```
HanlinMiao, line 31: Move this `IPAddress` to the end because we added ordering `parent__namespace` to `class Meta `of the `IPAddress` model.

nautobot/ipam/models.py
```diff
@@ -936,7 +937,7 @@ class IPAddress(PrimaryModel, StatusModel, RoleModelMixin):
     objects = BaseManager.from_queryset(IPAddressQuerySet)()
 
     class Meta:
-        ordering = ("ip_version", "host", "prefix_length")  # address may be non-unique
+        ordering = ("parent__namespace", "ip_version", "host", "mask_length")  # address may be non-unique
```
jathanism, line 940: Eek cross-table ordering is going to be potentially very expensive at scale. Just something to be aware of.

nautobot/ipam/models.py
```diff
@@ -995,6 +986,12 @@ def STATUS_SLAAC(cls):  # pylint: disable=no-self-argument
     def clean(self):
         super().clean()
 
+        # Validate that host is not being modified
+        if self.present_in_database:
+            ip_address = IPAddress.objects.get(id=self.id)
+            if ip_address.host != self.host:
+                raise ValidationError({"address": "Host cannot be modified once set"})
```
jathanism, line 993:
```suggestion
                raise ValidationError({"address": "Host address cannot be changed once created"})
```

nautobot/ipam/models.py
```diff
@@ -882,8 +882,9 @@ class IPAddress(PrimaryModel, StatusModel, RoleModelMixin):
         db_index=True,
         help_text="IPv4 or IPv6 host address",
     )
-    broadcast = VarbinaryIPField(null=False, db_index=True, help_text="IPv4 or IPv6 broadcast address")
-    prefix_length = models.IntegerField(null=False, db_index=True, help_text="Length of the Network prefix, in bits.")
+    mask_length = models.IntegerField(
+        null=False, db_index=True, help_text="Length of the Network Configuration Mask, in bits."
+    )
```
jathanism, line 887: I'm not sure how I feel about this help text.
```suggestion
        null=False, db_index=True, help_text="Length of the network mask, in bits."
    )
```

jathanism (changes requested):
Minor changes.

jathanism (approved):

nautobot/ipam/models.py
glennmatthews: This was my suggestion, I hadn't considered the performance implication. I'd be fine with reverting this as long as we're OK with IPAddress querysets potentially interleaving records from different namespaces?

glennmatthews (approved):

Issue #3845, opened 2023-06-03
Change `Service` Uniqueness Constraint


bryanculver:
_Carved out of #2908_
### Proposed Changes
Unique together: `(device, name)` and `(vm, name)` with natural key `(name, device, vm)`; for duplication use other fields to help add to name uniqueness

HanlinMiao:
Closed by #3863

PR #3863, opened 2023-06-06
Added two uniqueness constraints, ("name", "device") and ("name", "virtual_machine"), to `Service` model.

HanlinMiao:
# Closes: #3845 
# What's Changed

1. Added two uniqueness constraints, ("name", "device") and ("name", "virtual_machine"), to Service model.
2. Added data migration to ensure all existing service instances fit those uniqueness constraints.

# TODO

- [ ] Unit, Integration Tests


nautobot/ipam/forms.py
```diff
@@ -776,6 +788,14 @@ def __init__(self, *args, **kwargs):
         else:
             self.fields["ip_addresses"].choices = []
 
+    def clean(self, *args, **kwargs):
+        if self.instance:
+            if self.instance.device:
+                self.cleaned_data["device"] = self.instance.device
+            else:
+                self.cleaned_data["virtual_machine"] = self.instance.virtual_machine
```
glennmatthews, line 796: What's this needed for?

HanlinMiao: I think because I added `device` and `virtual_machine` fields on `ServiceForm` and in `service_edit.html`, we only statically render `obj.device` or `obj.virtual_machine` in EditView. Because of that, self.cleaned_data["device"] and self.cleaned_data["virtual_machine"] will always be `None`, this will cause a `ValidationError` in `clean()`.

HanlinMiao: On second thought, can `Device/VirtualMachine` be modified in `ServiceEditView`? If so, we can always render the fields dynamically.

nautobot/ipam/tests/test_api.py
```diff
@@ -617,6 +617,7 @@ def test_ports_regression(self):
         self.assertEqual(response.json()["ports"], expected)
 
         # And do it again, but with ports as int.
+        data["name"] = "http-1"
```
HanlinMiao, line 620: Change the name here because of the new constraint added.

nautobot/ipam/migrations/0035_ensure_all_services_fit_uniqueness_constraint.py
```diff
@@ -0,0 +1,57 @@
...
+
+from django.db import migrations
+
+
+def make_sure_service_name_is_unique(service_list, parent, cache, max_name_length):
+    """
+    service_list: list; Service instances that are associated to Devices or Virtual Machines.
+    parent: str; could be either "device" or "virtual_machine".
+    cache: dictionary; dictionary that contains Device/Virtual Machine instances as Key and associated Service instances' names as Value.
+    max_name_length: int; max_length for the name attribute of Service model.
```
jathanism, line 11:
```suggestion
    Args:
        service_list (list): Service instances that are associated to Devices or Virtual Machines.
        parent (str): could be either "device" or "virtual_machine".
        cache (dict): dictionary that contains Device/Virtual Machine instances as Key and associated Service instances' names as Value.
        max_name_length (int): max_length for the name attribute of Service model.
```

nautobot/ipam/migrations/0035_ensure_all_services_fit_uniqueness_constraint.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    Service = apps.get_model("ipam", "Service")
+    MAX_NAME_LENGTH = 100
+    device_services = Service.objects.filter(device__isnull=False)
+    DEVICE_TO_SERVICE = {}
+    make_sure_service_name_is_unique(device_services, "device", DEVICE_TO_SERVICE, MAX_NAME_LENGTH)
+
+    vm_services = Service.objects.filter(virtual_machine__isnull=False)
+    VM_TO_SERVICE = {}
+    make_sure_service_name_is_unique(vm_services, "virtual_machine", VM_TO_SERVICE, MAX_NAME_LENGTH)
```
jathanism, line 47: Just a minor style nit since these aren't module-level constants having them be UPPERCASED isn't necessary. Also since `MAX_NAME_LENGTH` is hard-coded, seems like it should just be built into `make_sure_service_name_is_unique()` instead of passed as an argument?

jathanism (approved):
Minor comments. Everything looks good.

Issue #2916, opened 2022-11-28
IPAM Overhaul: Refactor utilization/allocation calculation logic


lampwins:
- [ ] Based on #2915, make calculations work
- [ ] Separate UI context for prefix vs IP address calculations
- [ ] Address #2648

bryanculver:
Need to do #2531 so solving this with that story (and IPAddress) while moving the existing code.

gsnider2195:
Adding a second utilization column slows down the list view even more (over 5 seconds to load 46 prefixes on my dev machine). I'm going to combine the columns for now.

Issue #2248, opened 2022-08-17
Prefix utilization column rendering slowly


gsnider2195:
### Environment
* Python version: v3.7.13
* Nautobot version: v1.4.0

### Steps to Reproduce
1. Create 1000 prefixes
2. Disable django debug toolbar
3. Navigate to ipam prefixes page with per_page set to 1000
4. Click `configure` and remove the utilization column

### Expected Behavior
Page load times shouldn't increase by 10+ seconds to call `Prefix.get_utilization` 1000 times

### Observed Behavior
Page load times increase by 10+ seconds to call `Prefix.get_utilization` 1000 times

Appears to be the template code in [utilization_graph.html](https://github.com/nautobot/nautobot/blob/488bcbb27d970d20ea22e15615f9715066a37051/nautobot/utilities/templates/utilities/templatetags/utilization_graph.html#L1)

gsnider2195:
Rendering the same template in jinja2 reduced the render time from 16 seconds to 5 seconds for 1000 rows.

```diff --git a/nautobot/ipam/tables.py b/nautobot/ipam/tables.py
index a9cdbbf2a..832d0736e 100644
--- a/nautobot/ipam/tables.py
+++ b/nautobot/ipam/tables.py
@@ -340,10 +340,29 @@ class PrefixTable(StatusTableMixin, BaseTable):
 
 
 class PrefixDetailTable(PrefixTable):
-    utilization = tables.TemplateColumn(template_code=UTILIZATION_GRAPH, orderable=False)
+    utilization = tables.Column(accessor="get_utilization")
     tenant = TenantColumn()
     tags = TagColumn(url_name="ipam:prefix_list")
 
+    def render_utilization(self, value):
+        from django.utils.html import format_html
+        from nautobot.utilities.templatetags.helpers import utilization_graph
+        import jinja2
+
+        template_code = """
+            <div {% if utilization_count and total_count %}title="Used: {{ utilization_count }}&#13;Count: {{ total_count }}" {% endif %}class="progress text-center">
+                {% if utilization < 30 %}<span style="font-size: 12px;">{{ utilization }}%</span>{% endif %}
+                <div class="progress-bar progress-bar-{% if utilization >= danger_threshold %}danger{% elif utilization >= warning_threshold %}warnin
g{% else %}success{% endif %}"
[...]

PR #3873, opened 2023-06-07
Update prefix utilization calculation

gsnider2195:
# Closes: #2916 #3894
# What's Changed

- If the `Prefix.type` is `Container`, the utilization is calculated as the sum of the total address space of all child prefixes.
- If the `Prefix.type` is `Pool`, the utilization is calculated as the sum of the total number of IP addresses within the pool's range.
- If the `Prefix.type` is `Network`:
    - The utilization is calculated as the sum of the total address space of all child `Pool` prefixes plus the total number of child IP addresses.
    - For IPv4 networks larger than /31, if neither the first or last address is occupied by either a pool or an IP address, they are subtracted from the total size of the prefix.

## Example

- 192.168.0.0/16          `Container - 1024/65536 utilization`
    - 192.168.1.0/24      `Network - 1/254 utilization`
        - 192.168.1.1     `IP Address`
    - 192.168.2.0/24      `Network - 4/256 utilization`
        - 192.168.2.0/30  `Pool - 1/4 utilization`
            - 192.168.2.1 `IP Address`
    - 192.168.3.0/24      `Network - 5/254 utilization`
        - 192.168.3.1     `IP Address`
        - 192.168.3.64/30 `Pool - 0/4 utilization`
    - 192.168.4.0/24      `Network - 1/256 utilization`
        - 192.168.4.255   `IP Address`

## ⚠️ NOTE

It's no longer possible to assign IP addresses to a pool and pools must exist within a network with a shorter prefix length. So in order to create a /24 pool of 256 addresses, you must first create a network to contain it that is /23 or larger.

# TODO

- [ ] Unit, Integration Tests


nautobot/ipam/models.py
```diff
@@ -709,7 +709,6 @@ def subnets(self, direct=False, include_self=False, for_update=False):
 
         return query.filter(
             ip_version=self.ip_version,
-            prefix_length__gte=self.prefix_length,
```
gsnider2195, line 712: With uniqueness constraints it would never be possible to have an overlapping network with the same prefix length and if the prefix length is shorter, either `network__gte` or `broadcast__lte` would not match.

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -91,14 +91,38 @@ The following changes have been made to the `Prefix` model.
...
-#### IPAddressd Parenting Concrete Relationship
+### IPAddress Parenting Concrete Relationship
 
 The `ipam.IPAddress` model has been modified to have a foreign key to `ipam.Prefix` as the `parent` field. Parenting of IP addresses is now automatically managed at the database level to greatly improve performance especially when calculating tree hierarchy and utilization.
 
 | Removed                | Replaced With   |
 |------------------------|-----------------|
 | `get_child_prefixes()` | `descendants()` |
 
+### Prefix get_utilization Method
```
glennmatthews, line 102: Would it make sense to duplicate some/all of this section into `nautobot/docs/models/ipam/prefix.md`?

nautobot/ipam/models.py
```diff
@@ -836,29 +835,42 @@ def get_first_available_ip(self):
...
+            child_ips = netaddr.IPSet(self.ip_addresses.values_list("host", flat=True))
+
+        child_prefixes = netaddr.IPSet(p.prefix for p in self.children.only("network", "prefix_length").iterator())
+        numerator = child_prefixes | child_ips
+
+        # Exclude network and broadcast address from the denominator unless they've been assigned to an IPAddress or child pool.
+        if all(
+            [
+                not self.children.exclude(type=choices.PrefixTypeChoices.TYPE_POOL).exists(),
+                denominator > 2,
```
glennmatthews, line 865: This line is for /31, /32 prefixes? A comment might be good to document this.

nautobot/ipam/models.py
```diff
@@ -836,29 +835,42 @@ def get_first_available_ip(self):
...
+                self.type != choices.PrefixTypeChoices.TYPE_POOL,
+                self.ip_version == 4,
```
glennmatthews, line 867: I saw your comment on Slack about the behavior of `all()`, but can we make this any more efficient by doing the non-DB-accessing checks first? I.e not even look at `self.children` if we don't pass the other three checks?

nautobot/ipam/models.py
```diff
@@ -836,29 +835,42 @@ def get_first_available_ip(self):
...
-                    self.type != choices.PrefixTypeChoices.TYPE_POOL,
-                ]
-            ):
-                prefix_size -= 2
-            child_count = prefix_size - self.get_available_ips().size
-            return UtilizationData(numerator=child_count, denominator=prefix_size)
+            child_ips = netaddr.IPSet(self.ip_addresses.values_list("host", flat=True))
+
+        child_prefixes = netaddr.IPSet(p.prefix for p in self.children.only("network", "prefix_length").iterator())
+        numerator = child_prefixes | child_ips
```
glennmatthews, line 859: Perhaps `numerator_set` for clarity?

nautobot/ipam/models.py
```diff
@@ -836,29 +835,42 @@ def get_first_available_ip(self):
         return f"{next(available_ips.__iter__())}/{self.prefix_length}"
 
     def get_utilization(self):
```
glennmatthews, line 837: Does this help at all with #2248?

nautobot/ipam/tables.py
```diff
@@ -410,8 +410,8 @@ class PrefixDetailTable(PrefixTable):
     class Meta(PrefixTable.Meta):
         fields = (
             "pk",
-            "namespace",
             "prefix",
+            "namespace",
```
glennmatthews, line 414: Was there an open issue regarding this ordering? I thought there was but I can't find it.

nautobot/ipam/tests/test_models.py
```diff
@@ -522,57 +522,53 @@ def test_get_utilization(self):
...
+            prefix="10.0.0.128/30", type=PrefixTypeChoices.TYPE_POOL, status=self.status, namespace=self.namespace
+        )
+        self.assertEqual(slash25.get_utilization(), (4, 128))
+
+        # When the pool does not overlap with broadcast or network address, the denominator decrements by 2
+        pool.network = "10.0.0.132"
+        pool.save()
+        self.assertEqual(slash25.get_utilization(), (4, 126))
 
         # IPv4 Non-container Prefix /31, network and broadcast addresses count toward utilization
```
glennmatthews, line 557: Can we add a test for /32 and /128 prefixes, just making sure we aren't setting ourselves up for a divide-by-zero or anything like that?

nautobot/ipam/tests/test_models.py
```diff
@@ -581,13 +577,10 @@ def test_get_utilization(self):
...
-            Prefix(prefix="22.0.0.0/12", status=self.status, namespace=self.namespace),
-            Prefix(prefix="22.16.0.0/12", status=self.status, namespace=self.namespace),
-            Prefix(prefix="22.32.0.0/12", status=self.status, namespace=self.namespace),
-            Prefix(prefix="22.48.0.0/12", status=self.status, namespace=self.namespace),
-        ]
-        [p.save() for p in prefixes]  # pylint: disable=expression-not-assigned
+        Prefix.objects.create(prefix="22.0.0.0/12", status=self.status, namespace=self.namespace)
+        Prefix.objects.create(prefix="22.16.0.0/12", status=self.status, namespace=self.namespace)
+        Prefix.objects.create(prefix="22.32.0.0/12", status=self.status, namespace=self.namespace)
+        Prefix.objects.create(prefix="22.48.0.0/12", status=self.status, namespace=self.namespace)
```
glennmatthews, line 583: IIRC the reason we were doing `save()` instead of `create()` was because the logic to decompose `prefix` into its components wasn't happening in some cases - is that a non-issue here?

nautobot/ipam/tables.py
gsnider2195: I did too and I also can't find it.

nautobot/ipam/tests/test_models.py
gsnider2195: It looks like it's doing almost the same thing by using create:

https://github.com/django/django/blob/722e9f8a38f5b34f2423059a75f8a59bb8eb931a/django/db/models/query.py#L446-L454
```py
    def create(self, **kwargs):
        """
        Create a new object with the given kwargs, saving it to the database
        and returning the created object.
        """
        obj = self.model(**kwargs)
        self._for_write = True
        obj.save(force_insert=True, using=self.db)
        return obj
```

nautobot/ipam/tables.py
gsnider2195: Got it now #3894

nautobot/ipam/models.py
gsnider2195: - With this utilization calculation in place it takes 94 seconds to load the list view with 1000 prefixes. 
- If I short circuit the utilization calculation and just return `UtilizationData(numerator=numerator_set.size, denominator=denominator)` it takes 47 seconds. 
- With the jinja template from #2248 and short circuiting get_utilization it takes 46 seconds.
- With the jinja template and calculating utilization it takes 88 seconds
- When I switch back to `next` it takes 71 seconds

These times are with DJDT enabled

nautobot/ipam/querysets.py
```diff
@@ -327,52 +327,38 @@ def _validate_cidr(self, value):
         except netaddr.AddrFormatError as err:
             raise ValidationError({"cidr": f"{value} does not appear to be an IPv4 or IPv6 network."}) from err
 
-    def get_closest_parent(self, cidr, namespace, max_prefix_length=0):
```
gsnider2195, line 330: Refactored `get_closest_parent` with @jathanism to fix case where /32 IPv4 address was not matching /32 prefix

nautobot/ipam/models.py
glennmatthews: Not entirely sure I follow all that. :-) Sounds like the gist is that the jinja template gives only a negligible performance gain at this point?

nautobot/ipam/tests/test_models.py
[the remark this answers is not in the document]
gsnider2195: I brought the changes for `test_querysets.py` over from [u/gas-prefix-ip-constraints](https://github.com/nautobot/nautobot/tree/u/gas-prefix-ip-constraints). The test in `test_models` didn't need to be updated.

nautobot/ipam/models.py
gsnider2195: Bottom line is this change makes the utilization calculation slower and it doesn't appear that #2248 is going to help much.

nautobot/ipam/models.py
```diff
@@ -832,29 +831,43 @@ def get_first_available_ip(self):
...
-                    self.prefix_length < 31,
-                    self.type != choices.PrefixTypeChoices.TYPE_POOL,
-                ]
-            ):
-                prefix_size -= 2
-            child_count = prefix_size - self.get_available_ips().size
-            return UtilizationData(numerator=child_count, denominator=prefix_size)
+            child_ips = netaddr.IPSet(self.ip_addresses.values_list("host", flat=True))
+
+        child_prefixes = netaddr.IPSet(p.prefix for p in self.children.only("network", "prefix_length").iterator())
```
glennmatthews, line 854: Could skip this lookup if self.type != TYPE_CONTAINER, right?

nautobot/ipam/models.py
```diff
@@ -832,29 +831,43 @@ def get_first_available_ip(self):
...
-                [
-                    self.ip_version == 4,
-                    self.prefix_length < 31,
-                    self.type != choices.PrefixTypeChoices.TYPE_POOL,
-                ]
-            ):
-                prefix_size -= 2
-            child_count = prefix_size - self.get_available_ips().size
-            return UtilizationData(numerator=child_count, denominator=prefix_size)
+            child_ips = netaddr.IPSet(self.ip_addresses.values_list("host", flat=True))
```
glennmatthews, line 852: Could skip this lookup if self.type == TYPE_CONTAINER, right?

glennmatthews (approved):

jathanism (approved):

Issue #3850, opened 2023-06-03
Change `VLAN Group` Uniqueness Constraints


bryanculver:
_Carved from #2909_
### Proposed Changes
- [x] **VLAN Group:** `name` becomes unique with an enumeration for collisions, drop unique with `location`

bryanculver:
We're going to split `VLAN` from `VLAN Group`.

Issue #2662, opened 2022-10-20
Remove the usage of slug fields


glennmatthews:
### Discussed in https://github.com/nautobot/nautobot/discussions/2312

<div type='discussions-op-text'>

<sup>Originally posted by **lampwins** August 25, 2022</sup>
I propose ditching the usage of slug fields altogether. I realize this might be a contentious idea, so here are my thoughts.

#### Slugs in their present form create bad UX when managing data
Today we have slugs all over the place as fields that users can edit. Many folks have no idea what the actual function of the slug is and why it exists. I can't count the number of times I have answered the question "what is the slug for" or "what is the purpose of this field since it duplicates the name?"

We could certainly address some of these concerns, but I truly don't see the point in favor of just ditching them.

#### Slugs cause surprises when permalinking
If a slug is updated for whatever reason, that record's UI link changes which is surprising to the user if they don't understand the purpose of the slug. This becomes very obvious when you have objects with simple names (e.g., one word) in which the name field and the slug contain the exact same value. The user is presented with two fields with the same value, and when their intention is to change the name, it is natural for them to also change the slug, even though it is not automatically updated for an existing record.

Also, consider automation like SSoT use cases which may change the value of the slug (either inadvertently or via a valid use case).

#### Slugs are great if you actually care about fancy URLs
IMO slugs in Django are great if you are running a blog where there is absolutely no reason to change the title of a post, and you want a nice-looking URL, which, more importantly, is SEO friendly. Nautobot is a data-intensive, internal IT application; our data changes, and we don't care about SEO.

Also, it's 2022, and modern browsers are removing focus from the structure and content of URLs. Safari does not even show the URL. Chrome puts focus on only the domain.

Link sharing between humans has become much more advanced. Many modern chat applications will attempt to include additional link context, like the page's title while hiding the actual text of the URL.

#### Our usage of slugs vs. PKs is entirely inconsistent
Sometimes we use slugs, sometimes we use the UUID PK. Our REST API uses the UUID exclusively. I know we have discussed moving towards natural keys, which alone is great, but we know in many (potentially most) cases, these will be composite keys which presents many challenges in creating workable URLs that play nice within the Django ecosystem of 3rd party apps that we rely on.

#### Bottom line, let's move to all UUIDs
[...]

PR #3875, opened 2023-06-08
Changed VLANGroup `name` to unique and made VLAN `vlan_group` required.

timizuoebideri1:
# Closes: #3850 
# What's Changed

# TODO

- [ ] Unit, Integration Tests


jathanism (approved):
Pretty straightforward fix. Only have the one question. Obviously we need the tests passing, too! 🙈

nautobot/ipam/migrations/0035_vlan_vlan_group_default_values.py
```diff
@@ -0,0 +1,58 @@
...
+
+def vlan_vlan_group_uniqueness_constrains_revise(apps, schema_editor):
+    # Resolve VLAN-Group name uniqueness
+    VLAN = apps.get_model("ipam", "vlan")
+    VLANGroup = apps.get_model("ipam", "vlangroup")
+
+    increment_names_of_records_with_similar_names(VLANGroup)
+
+    # Hope fully vlan group with name `Default VLAN Group` would not exist
+    default_vlan_group = VLANGroup.objects.create(name="Default VLAN Group")
```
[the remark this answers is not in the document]
gsnider2195: I believe the plan is to make it the same as namespace

nautobot/ipam/migrations/0035_vlan_vlan_group_default_values.py
```diff
@@ -0,0 +1,58 @@
...
+    )
+    records_to_update = []
+
+    for duplicate_record in duplicate_records:
+        records = model.objects.filter(name=duplicate_record["name"])[1:]
+        for idx, record in enumerate(records):
+            # Starting from 2 e.g Example Name 2, Example Name 3
+            record.name = f"{record.name} {idx + 2}"
+            records_to_update.append(record)
+    model.objects.bulk_update(records_to_update, ["name"])
```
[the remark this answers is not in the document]
HanlinMiao: Indeed, there are so many cases of this. This algorithm seems more efficient, however, it does not seem to take into account the `max_length` of the field.

glennmatthews:
Looks like migration needs to be renumbered?

nautobot/ipam/models.py
```diff
@@ -1146,9 +1146,6 @@ class Meta:
             "name",
         )  # (location, name) may be non-unique
         unique_together = [
-            # 2.0 TODO: since location is nullable, and NULL != NULL, this means that we can have multiple non-Location VLANGroups
-            # with the same name. This should probably be fixed with a custom validate_unique() function!
-            ["location", "name"],
             # 2.0 TODO: Remove unique_together to make slug globally unique. This would be a breaking change.
             ["location", "slug"],
```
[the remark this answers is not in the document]
glennmatthews: If doing so, please also be sure to update any NaturalKeyOrPKMultipleChoiceFilters that reference a VLANGroup to use `to_field_name="name"`.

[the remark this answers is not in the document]
glennmatthews: I'm also happy to tackle `VLANGroup.slug` removal as part of #2662 if you haven't already started working on this.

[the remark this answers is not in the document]
timizuoebideri1: That would be great

nautobot/extras/tests/test_api.py
```diff
@@ -2738,32 +2739,30 @@ def send_bulk_data(http_method, data):
...
                 # Add required fields for PUT method:
                 if method == "put":
-                    vlan2_json_data.update({"vid": vlan2.vid, "name": vlan2.name})
+                    vlan2_json_data.update({"vid": "5", "name": vlan2.name})
 
             # Try method without specifying required relationships for either vlan1 or vlan2 (fails)
             json_data = [vlan1_json_data, vlan2_json_data]
             response = send_bulk_data(method, json_data)
             self.assertHttpStatus(response, 400)
+            print("==> ", response.json())
```
timizuoebideri1, line 2765:
```suggestion
```

nautobot/ipam/factory.py
```diff
@@ -142,13 +142,36 @@ class Params:
         "has_location", random_instance(lambda: Location.objects.get_for_model(VLANGroup), allow_null=False), None
     )
 
+    @factory.post_generation
+    def children(self, create, extracted, **kwargs):
+        """Creates child prefixes and ip addresses within the prefix IP space."""
```
glennmatthews, line 147:
```suggestion
        """Creates child VLANs within the VLANGroup."""
```

nautobot/ipam/migrations/0035_vlan_vlan_group_default_values.py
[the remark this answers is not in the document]
glennmatthews: Probably better to use an approach similar to `nautobot/extras/migrations/0091_uniqueness_data_migration.py`, in which we explicitly iterate over all records in the DB and check for duplicate names as we go, and also enforce name length limits.

nautobot/ipam/migrations/0037_vlan_vlan_group_default_values.py
```diff
@@ -0,0 +1,41 @@
...
+def vlan_vlan_group_uniqueness_constrains_revise(apps, schema_editor):
+    # Resolve VLAN-Group name uniqueness
+    VLAN = apps.get_model("ipam", "vlan")
+    VLANGroup = apps.get_model("ipam", "vlangroup")
+
+    increment_names_of_records_with_similar_names(VLANGroup)
+
+    # Hope fully vlan group with name `Default VLAN Group` would not exist
+    _, default_vlan_group = VLANGroup.objects.get_or_create(name="Default VLAN Group")
+    VLAN.objects.filter(vlan_group__isnull=True).update(vlan_group=default_vlan_group)
```
glennmatthews, line 17: This is **probably** going to result in uniqueness constraint validations, as presently you can create multiple VLANs with null `vlan_group` and the same `vid` and `name`, but they will not be permitted to coexist under the same VLANGroup once moved. (https://github.com/nautobot/nautobot/blob/develop/nautobot/ipam/models.py#L1215-L1217)

Fixing this may require a similar approach to that taken in the IPAddress/Prefix namespace migrations, where we might need to create up to N distinct `VLANGroups`, probably using `vlan.location` as a way to determine which such group any given VLAN gets assigned to. Sorry to be the bearer of bad news!

glennmatthews (changes requested):

nautobot/ipam/migrations/0037_vlan_vlan_group_default_values.py
```diff
@@ -0,0 +1,52 @@
...
+                continue
+            counter += 1
+    VLAN.objects.bulk_update(vlans_to_update, ["vlan_group"])
+
+
+def revert_vlan_vlan_group_uniqueness_constrains_revise(apps, schema_editor):
+    VLAN = apps.get_model("ipam", "vlan")
+    VLANGroup = apps.get_model("ipam", "vlangroup")
+
+    # TODO(timizuo): Find a way to revert changes made to VLAN and VLANGroup
```
timizuoebideri1, line 38: Any ideas on how I may revert this, since I don't want to destroy a VLAN group that was there before the migration with similar names(highly unlikey),
or should it be none reversible. Same for VLANGroup.

😔

HanlinMiao: I honestly do not think there is a need to reverse this migration. Since we are only adding data in the forward migration. Rolling it back forcefully might cause data loss as you said. but Just my opinion tho.

glennmatthews: Probably non-reversible is fine.

timizuoebideri1:
I believe that not reverting modifications made to the Factory and test cases would have a good impact on the VLAN issue that would be created.

nautobot/core/api/serializers.py
```diff
@@ -159,7 +159,7 @@ def to_representation(self, value):
             model = self.queryset.model
 
         # foreign key relationship when the destination field is referenced by string not by class
-        elif getattr(self.parent.Meta.model, self.source, False):
+        elif getattr(self.parent, "Meta", None) and getattr(self.parent.Meta.model, self.source, False):
```
glennmatthews, line 162: Is this needed after #3916? It looks like a reasonable change, but is it fixing an issue that is actually encountered?

nautobot/core/tests/test_api.py
```diff
@@ -630,10 +642,11 @@ def test_related_by_attributes(self):
 
     def test_related_by_attributes_no_match(self):
         data = {
-            "vid": 100,
-            "name": "Test VLAN 100",
+            "vid": 120,
+            "name": "Test VLAN 120",
```
glennmatthews, line 646: This change appears to make the check at line 644/657 below invalid.

These are a weird set of tests in any case though, seems like very basic data validation testing...?

nautobot/core/tests/test_api.py
```diff
@@ -646,14 +659,15 @@ def test_related_by_attributes_no_match(self):
 
     def test_related_by_attributes_multiple_matches(self):
         data = {
-            "vid": 100,
-            "name": "Test VLAN 100",
+            "vid": 130,
+            "name": "Test VLAN 130",
```
glennmatthews, line 663: Similarly this check appears to make line 664/678 invalid.

nautobot/core/tests/test_api.py
```diff
@@ -666,10 +680,11 @@ def test_related_by_attributes_multiple_matches(self):
 
     def test_related_by_invalid(self):
         data = {
-            "vid": 100,
-            "name": "Test VLAN 100",
+            "vid": 140,
+            "name": "Test VLAN 140",
```
glennmatthews, line 684: Similarly, see line 680/695.

nautobot/dcim/tests/test_filters.py
```diff
@@ -284,14 +284,16 @@ def common_test_data(cls):
     Prefix.objects.create(prefix=netaddr.IPNetwork("192.168.2.0/24"), location=loc1, status=prefix_status)
 
     # TODO: remove these once we have a Sites fixture; for now SiteTestCase needs VLANGroups and VLANs with Sites
```
glennmatthews, line 286: Comment is probably obsolete now; not blocking but would be interesting to see whether the below can be removed now that we have appropriate factories in place for VLANGroup and VLAN.

nautobot/dcim/tests/test_migrations.py
```diff
@@ -256,9 +256,11 @@ def populateDataBeforeMigration(self, installed_apps):
...
         self.vlans = [
-            self.vlan.objects.create(name="VLAN 1", vid=1, site=sites[1]),
-            self.vlan.objects.create(name="VLAN 2", vid=2, site=sites[2]),
-            self.vlan.objects.create(name="VLAN 3", vid=3, site=sites[3], location=locations[2]),
+            self.vlan.objects.create(name="VLAN 1", vid=1, site=sites[1], vlan_group=self.vlan_groups[0]),
+            self.vlan.objects.create(name="VLAN 2", vid=2, site=sites[2], vlan_group=self.vlan_groups[0]),
+            self.vlan.objects.create(
+                name="VLAN 3", vid=3, site=sites[3], location=locations[2], vlan_group=self.vlan_groups[0]
+            ),
         ]
```
glennmatthews, line 264: Since this is "before-migration" data, it would probably make sense to revert this change so that the migration (once we implement VLAN-requires-VLANGroup) would be exercised?

nautobot/ipam/factory.py
```diff
@@ -191,14 +214,13 @@ class Meta:
     has_description = NautobotBoolIterator()
     description = factory.Maybe("has_description", factory.Faker("text", max_nb_chars=200), "")
 
-    has_vlan_group = NautobotBoolIterator()
-    vlan_group = factory.Maybe("has_vlan_group", random_instance(VLANGroup, allow_null=False), None)
+    vlan_group = random_instance(VLANGroup, allow_null=False)
```
glennmatthews, line 217: Revert for now?

nautobot/ipam/migrations/0037_data_migration_vlan_group_name_uniqueness.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("ipam", "0036_add_uniqueness_constraints_to_service"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            vlan_group_name_uniqueness_constrains,
```
glennmatthews, line 24:
```suggestion
def vlan_group_name_uniqueness_constraints(apps, schema_editor):
    VLANGroup = apps.get_model("ipam", "vlangroup")

    increment_names_of_records_with_similar_names(VLANGroup)


class Migration(migrations.Migration):
    dependencies = [
        ("ipam", "0036_add_uniqueness_constraints_to_service"),
    ]

    operations = [
        migrations.RunPython(
            vlan_group_name_uniqueness_constraints,
```

nautobot/ipam/utils/migrations.py
```diff
@@ -731,3 +731,27 @@ def ensure_correct_prefix_broadcast(apps):
...
             prefix.save()
+
+
+def increment_names_of_records_with_similar_names(model: models.Model):
+    """
+    This function increments the names of records with similar names in a given model.
+    """
+    cache = set()
+    records_to_update = []
+    for instance in model.objects.all():
```
glennmatthews, line 742: with the potential for a large number of records, should we be using `iterator` here? Compare `nautobot/dcim/migrations/0030_migrate_region_and_site_data_to_locations.py`

nautobot/ipam/utils/migrations.py
```diff
@@ -731,3 +731,27 @@ def ensure_correct_prefix_broadcast(apps):
...
+        name = instance.name
+        counter = 1
+        while name in cache:
+            max_name_length = model.name.field.max_length
+            name = f"{instance.name[:max_name_length]} {counter}"
```
glennmatthews, line 747: This won't do the right thing if instance.name is already at max-length - the resulting constructed `name` string will be too long and will result in a DB error. Compare `nautobot/extras/migrations/0091_uniqueness_data_migration.py`.

nautobot/ipam/utils/migrations.py
```diff
@@ -731,3 +731,27 @@ def ensure_correct_prefix_broadcast(apps):
...
+            counter += 1
+
+        if name != instance.name:
+            print(f'{model._meta.verbose_name} instance {instance.id} is being renamed to "{name}" for uniqueness')
```
glennmatthews, line 751: Since this will typically be seen in the context of a migration, might be good to prepend `    ` to the printed string so that it appears nested under the migration message.

nautobot/dcim/tests/test_api.py
```diff
@@ -51,7 +51,7 @@
     VirtualChassis,
 )
 from nautobot.extras.models import ConfigContextSchema, Role, SecretsGroup, Status
-from nautobot.ipam.models import IPAddress, VLAN, Namespace, Prefix
+from nautobot.ipam.models import IPAddress, VLAN, Namespace, Prefix, VLANGroup
```
HanlinMiao, line 54: lets go ahead and alpha sort this line.

nautobot/dcim/tests/test_views.py
```diff
@@ -81,7 +81,7 @@
     Status,
     Tag,
 )
-from nautobot.ipam.models import VLAN, IPAddress, Namespace, Prefix
+from nautobot.ipam.models import VLAN, IPAddress, Namespace, Prefix, VLANGroup
```
HanlinMiao, line 84: Same here, can we alpha-sort this line :)?

nautobot/core/api/serializers.py
timizuoebideri1: Nope

nautobot/dcim/tests/test_filters.py
timizuoebideri1: Yup, It can be removed but it causes test failures where `VLAN` and `VLANGroup` location are required to be one of the above locations

glennmatthews (approved):

Issue #3725, opened 2023-05-10
RESTful API adherence to schema


itdependsnetworks:
### Environment
* Nautobot version (Docker tag too if applicable): next
* Python version: 3.9
* Database platform, version: Postgres
* Middleware(s): N/A

### Steps to Reproduce
1. Run an API query on a list like api with query of depth=0
2. Run an API query on a list like api with query of depth=1
3. Capture both outputs

### Expected Behavior

The data structure to be of the same schema. Note: This is not to state the data should be the same, query params can and are built to add and remove data, but in my experience this is a part of RESTful APIs I agree with. APIs that change structure on query params

### Observed Behavior

depth=0
```
        {
            "id": "9bc9abcb-8ac0-4ad4-bba5-15d22a5d0299",
            "url": "http://localhost:8080/api/dcim/locations/9bc9abcb-8ac0-4ad4-bba5-15d22a5d0299/",
            "tags": [
                "http://localhost:8080/api/extras/tags/6d7f17bb-b050-4fd5-a910-3c22cf6e8cf8/"
            ]
        },
```

depth=1
```
        {
            "id": "9bc9abcb-8ac0-4ad4-bba5-15d22a5d0299",
            "url": "http://localhost:8080/api/dcim/locations/9bc9abcb-8ac0-4ad4-bba5-15d22a5d0299/",
            "tags": [
                {
                    "id": "6d7f17bb-b050-4fd5-a910-3c22cf6e8cf8",
                    "url": "http://localhost:8080/api/extras/tags/6d7f17bb-b050-4fd5-a910-3c22cf6e8cf8/",
                    "display": "Remote",
                    "notes_url": "http://localhost:8080/api/extras/tags/6d7f17bb-b050-4fd5-a910-3c22cf6e8cf8/notes/",
                    "name": "Remote",
                    "slug": "remote",
                    "created": "2023-05-04T17:52:56.242959Z",
                    "last_updated": "2023-05-04T17:52:56.243007Z",
                    "color": "ff9800",
                    "description": ""
                } 
            ]
        },
```

glennmatthews:
This relates to #3042 / #3500, the `uiSchema` functionality added in #3651 et al to support generating create/edit forms from the REST API serializers, and the CSV import/export rework being done for #254, and so any change here needs to take those functionalities into account.

One possibility we've previously proposed would be to change the `depth=0` case from a bare URL or PK to a minimal nested dict, i.e. the example above would change its `tags` rendering to something like:

[...]

PR #3876, opened 2023-06-08
Self-Consistent Data-Types Hyperlinked Relation Fields

bryanculver:
# Closes: #3725
# What's Changed

- Build a nested brief serializer for relation hyperlinked fields instead of a URL

**Before**
Screenshot 2023-06-07 at 22 05 22

**After**
Screenshot 2023-06-07 at 22 05 17

# TODO

- [ ] Unit, Integration Tests
    -  `local_config_context_data` changed from `{}` to `null`, special case handling


nautobot/core/api/utils.py
```diff
@@ -337,9 +375,15 @@ def return_nested_serializer_data_based_on_depth(serializer, depth, obj, obj_rel
...
+        url = obj_related_field.get_absolute_url(api=True)
         if serializer.context.get("request"):
-            result = serializer.context.get("request").build_absolute_uri(result)
+            url = serializer.context.get("request").build_absolute_uri(url)
+
+        result = {
+            "id": obj_related_field.pk,
+            "object_type": obj_related_field._meta.app_label + ":" + obj_related_field._meta.model_name,
+            "url": url,
+        }
```
bryanculver, line 386: This was be from an earlier prototype which might be able to be removed.

nautobot/core/api/utils.py
```diff
@@ -322,11 +322,49 @@ class Meta:
...
+        field_kwargs = get_nested_relation_kwargs(relation_info)
+    else:
+        base_serializer_class = get_serializer_for_model(relation_info.related_model)
+
+        class NautobotBriefNestedSerializer(base_serializer_class):
+            def get_field_names(self, declared_fields, info):
+                """Ensure id, object_type, and url are only in the brief serializer."""
+                base_fields = list(super().get_field_names(declared_fields, info))
+                desired_fields = set(["id", "object_type", "url"])
+                return [field for field in base_fields if field in desired_fields]
```
bryanculver, line 348: Because we use this method in our mixins to bolt on fields not defined in `Meta.fields` and we should inherit from the related models base serializer, we need to prevent them from carrying through.

bryanculver:
**Depth 1**
Screenshot 2023-06-07 at 22-18-50 Device – Django REST framework

nautobot/core/api/utils.py
```diff
@@ -322,11 +322,49 @@ class Meta:
...
+    # else we build a new one and store it in the cache for future use.
+    if brief_serializer_name in BRIEF_NESTED_SERIALIZER_CACHE:
+        field_class = BRIEF_NESTED_SERIALIZER_CACHE[brief_serializer_name]
+        field_kwargs = get_nested_relation_kwargs(relation_info)
+    else:
+        base_serializer_class = get_serializer_for_model(relation_info.related_model)
+
+        class NautobotBriefNestedSerializer(base_serializer_class):
+            def get_field_names(self, declared_fields, info):
+                """Ensure id, object_type, and url are only in the brief serializer."""
```
HanlinMiao, line 345: Perhaps?
```suggestion
                """Ensure only id, object_type, and url are in the brief serializer."""
```

nautobot/core/api/utils.py
HanlinMiao, line 348: Is there a case where "id", "object_type" or "url" is not available in `base_fields`? If not, we think we can just return `desired_fields` straight up.

HanlinMiao (commented):
I like the approach and made some comments. Let's see what others think.

nautobot/extras/tests/test_tags.py
```diff
@@ -52,7 +52,7 @@ def test_create_tagged_item(self):
         response = self.client.post(url, data, format="json", **self.header)
         self.assertHttpStatus(response, status.HTTP_201_CREATED)
         # response with default depth is a list of tag URLs
-        self.assertEqual(sorted(response.data["tags"]), sorted([self.absolute_api_url(t) for t in self.tags]))
+        self.assertEqual([rt["url"] for rt in response.data["tags"]], [self.absolute_api_url(t) for t in self.tags])
```
bryanculver, line 55: TODO: restore sorted

nautobot/core/api/utils.py
```diff
@@ -337,9 +337,15 @@ def return_nested_serializer_data_based_on_depth(serializer, depth, obj, obj_rel
...
     if depth == 0:
-        result = obj_related_field.get_absolute_url(api=True)
+        url = obj_related_field.get_absolute_url(api=True)
         if serializer.context.get("request"):
-            result = serializer.context.get("request").build_absolute_uri(result)
+            url = serializer.context.get("request").build_absolute_uri(url)
+
+        result = {
+            "id": obj_related_field.pk,
+            "object_type": obj_related_field._meta.app_label + ":" + obj_related_field._meta.model_name,
```
glennmatthews, line 346:
```suggestion
            "object_type": obj_related_field._meta.app_label + "." + obj_related_field._meta.model_name,
```

nautobot/core/api/serializers.py
```diff
@@ -119,6 +136,28 @@ def __init__(self, *args, **kwargs):
...
+        url = super().to_representation(value)
+        if self.queryset:
+            model = self.queryset.model
+        elif isinstance(self.parent, ManyRelatedField) and getattr(self.parent.parent.Meta.model, self.source, False):
+            model = getattr(self.parent.parent.Meta.model, self.source).field.model
+        elif getattr(self.parent.Meta.model, self.source, False):
+            model = getattr(self.parent.Meta.model, self.source).field.model
+        else:
+            return {"id": value.pk, "object_type": "unknown", "url": url}
+        return {"id": value.pk, "object_type": model._meta.app_label + ":" + model._meta.model_name, "url": url}
```
glennmatthews, line 159:
```suggestion
        return {"id": value.pk, "object_type": model._meta.app_label + "." + model._meta.model_name, "url": url}
```

nautobot/core/testing/api.py
```diff
@@ -253,13 +253,20 @@ def test_list_objects_depth_0(self):
...
+                            self.assertIsInstance(response_data[field], dict)
+                            url = response_data[field]["url"]
+                            pk = response_data[field]["id"]
+                            # The response should be a brief API object, containing an ID, object_type, and URL ending in the UUID of the relevant object
                             # http://nautobot.example.com/api/circuits/providers/<uuid>/
                             #                                                    ^^^^^^
-                            self.assertTrue(is_uuid(response_data[field].split("/")[-2]))
+                            self.assertTrue(is_uuid(url.split("/")[-2]))
+                            self.assertTrue(is_uuid(pk))
+                            # app_label, model_name = response_data[field]["object_type"].split(":")
```
glennmatthews, line 268:
```suggestion
                            # app_label, model_name = response_data[field]["object_type"].split(".")
```

nautobot/core/api/serializers.py
```diff
@@ -100,6 +101,22 @@ def fields(self):
...
+            },
+            "object_type": {"type": "string"},
+            "url": {
+                "type": "string",
+                "format": "uri",
+            },
+        },
+    }
+)
 class NautobotHyperlinkedRelatedField(WritableSerializerMixin, serializers.HyperlinkedRelatedField):
```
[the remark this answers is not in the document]
bryanculver: Tried that. Ran into many more issues with them being a serializer: https://github.com/nautobot/nautobot/pull/3876/commits/feb77dd02cb7269b3bc28f059a5b3e9383ace92b#diff-ffd2d83bcbb181f327be8d0433d68fe6237d39a6b45b1740c4e3424592a50382R328

nautobot/extras/tests/test_api.py
```diff
@@ -3388,6 +3388,10 @@ def test_update_tag_content_type_unchanged(self):
         self.assertEqual(tag.color, ColorChoices.COLOR_LIME)
         self.assertEqual(list(tag.content_types.all()), tag_content_types)
 
+    @skip("Something weird is happening with the returned order of content_types")
+    def test_recreate_object_csv(self):
+        pass
```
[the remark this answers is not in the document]
bryanculver: Looks like this may have been a local issue due to migrations.

nautobot/core/api/serializers.py
```diff
@@ -119,6 +136,28 @@ def __init__(self, *args, **kwargs):
...
+        """Convert URL representation to a brief nested representation."""
+        url = super().to_representation(value)
+        if self.queryset:
+            model = self.queryset.model
+        elif isinstance(self.parent, ManyRelatedField) and getattr(self.parent.parent.Meta.model, self.source, False):
+            model = getattr(self.parent.parent.Meta.model, self.source).field.model
+        elif getattr(self.parent.Meta.model, self.source, False):
+            model = getattr(self.parent.Meta.model, self.source).field.model
+        else:
+            return {"id": value.pk, "object_type": "unknown", "url": url}
```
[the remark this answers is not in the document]
bryanculver: It's a dead stop on the API response if we do. I'd rather have someone open a bug for "unknown" and we can fix than not being able to use the APIs for that.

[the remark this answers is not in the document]
glennmatthews: Fair enough - let's at least have it log a warning message too so we can detect it server-side as well?

nautobot/core/api/utils.py
```diff
@@ -337,9 +337,15 @@ def return_nested_serializer_data_based_on_depth(serializer, depth, obj, obj_rel
...
+        url = obj_related_field.get_absolute_url(api=True)
         if serializer.context.get("request"):
-            result = serializer.context.get("request").build_absolute_uri(result)
+            url = serializer.context.get("request").build_absolute_uri(url)
+
+        result = {
+            "id": obj_related_field.pk,
+            "object_type": obj_related_field._meta.app_label + "." + obj_related_field._meta.model_name,
+            "url": url,
+        }
```
glennmatthews, line 348: Pretty sure this will necessitate some updates to the `docs/rest-api/overview.md`. :-)

nautobot/core/api/utils.py
```diff
@@ -326,7 +326,7 @@ def return_nested_serializer_data_based_on_depth(serializer, depth, obj, obj_rel
     """
     Handle serialization of GenericForeignKey fields at an appropriate depth.
 
-    When depth = 0, return the URL for the related object.
+    When depth = 0, return a brief representation of the related object, containing URL, PK, and object_type.
```
glennmatthews, line 329: Wish we could include `composite_key` in this case too but calculating that possibly involves N additional database lookups per object, so that's to be avoided.

nautobot/core/api/serializers.py
```diff
@@ -100,6 +101,22 @@ def fields(self):
...
 
+@extend_schema_field(
+    {
+        "type": "object",
+        "properties": {
+            "id": {
+                "type": "string",
+                "format": "uuid",
+            },
+            "object_type": {"type": "string"},
```
glennmatthews, line 112: Non-blocking, but could consider adding a `pattern` regex to the `object_type` schema for greater precision.

nautobot/core/api/serializers.py
```diff
@@ -119,6 +136,34 @@ def __init__(self, *args, **kwargs):
...
+            model = self.queryset.model
+        elif isinstance(self.parent, ManyRelatedField) and getattr(self.parent.parent.Meta.model, self.source, False):
+            model = getattr(self.parent.parent.Meta.model, self.source).field.model
+        elif getattr(self.parent.Meta.model, self.source, False):
+            model = getattr(self.parent.Meta.model, self.source).field.model
+        else:
+            logger.warning(
+                "Unable to determine model for related field %r; "
+                "ensure that either 'queryset' is set on the field "
+                "or defines the related model in the Meta.",
```
glennmatthews, line 161: Perhaps?

```suggestion
                "ensure that either the field defines a 'queryset' "
                "or the Meta defines the related 'model'.",
```

nautobot/core/api/serializers.py
```diff
@@ -119,6 +136,34 @@ def __init__(self, *args, **kwargs):
...
+                self.field_name,
+            )
+            return {"id": value.pk, "object_type": "unknown", "url": url}
```
glennmatthews, line 164: Perhaps:

```suggestion
            return {"id": value.pk, "object_type": "unknown.unknown", "url": url}
```

Just so we stick to the `app_label.modelname` structure?

nautobot/core/api/serializers.py
```diff
@@ -119,6 +136,34 @@ def __init__(self, *args, **kwargs):
...
+        return {"id": value.pk, "object_type": model._meta.app_label + "." + model._meta.model_name, "url": url}
```
glennmatthews, line 165:
```suggestion
        return {"id": value.pk, "object_type": model._meta.label_lower, "url": url}
```

nautobot/core/api/utils.py
```diff
@@ -337,9 +337,15 @@ def return_nested_serializer_data_based_on_depth(serializer, depth, obj, obj_rel
...
     if depth == 0:
-        result = obj_related_field.get_absolute_url(api=True)
+        url = obj_related_field.get_absolute_url(api=True)
         if serializer.context.get("request"):
-            result = serializer.context.get("request").build_absolute_uri(result)
+            url = serializer.context.get("request").build_absolute_uri(url)
+
+        result = {
+            "id": obj_related_field.pk,
+            "object_type": obj_related_field._meta.app_label + "." + obj_related_field._meta.model_name,
```
glennmatthews, line 346:
```suggestion
            "object_type": obj_related_field._meta.label_lower,
```

nautobot/docs/rest-api/overview.md
```diff
@@ -783,6 +828,37 @@ http://nautobot/api/ipam/prefixes/ \
...
+  },
+  "role": null,
+  "type": "network",
+  "description": "",
+  "tags": [],
+  "custom_fields": {},
+  "created": "2020-08-04T20:08:39.007125Z",
+  "last_updated": "2020-08-04T20:08:39.007125Z"
+}
+```
```
glennmatthews, line 860: Note that the preceding JSON blob is the API *response* not the *request*. The request body is simply `--data '{"prefix": "192.0.2.0/24", "location": 8df9e629-4338-438b-8ea9-06114f7be08e}'`. So perhaps:

````suggestion
```no-highlight
curl -s -X POST \
-H "Authorization: Token $TOKEN" \
-H "Accept: application/json; version=2.0" \
http://nautobot/api/ipam/prefixes/ \
--data '{"prefix": "192.0.2.0/24", "location": {"id": "8df9e629-4338-438b-8ea9-06114f7be08e", "object_type": "dcim.location"}' | jq '.'
```
````

File: nautobot/docs/rest-api/overview.md [hunk 19]
```diff
@@ -783,6 +828,37 @@ http://nautobot/api/ipam/prefixes/ \
...
+
```
glennmatthews, line 861: Also realizing that both of these examples here are incorrect/outdated since 1) `status` is a mandatory field on a prefix create and 2) so is `namespace` now.

tasks.py
```diff
@@ -627,7 +627,7 @@ def hadolint(context):
 @task
 def markdownlint(context):
     """Lint Markdown files."""
-    command = "markdownlint --ignore nautobot/project-static --ignore nautobot/ui/node_modules --config .markdownlint.yml --rules scripts/use-relative-md-links.js nautobot examples *.md"
+    command = "npm exec -- markdownlint --ignore nautobot/project-static --ignore nautobot/ui/node_modules --config .markdownlint.yml --rules scripts/use-relative-md-links.js nautobot examples *.md"
```
glennmatthews, line 630: Maybe add a `3876.fixed` changelog entry about this?

nautobot/ipam/tests/test_api.py
```diff
@@ -222,7 +222,7 @@ def test_create_single_available_prefix(self):
             response = self.client.post(url, data, format="json", **self.header)
             self.assertHttpStatus(response, status.HTTP_201_CREATED)
             self.assertEqual(response.data["prefix"], str(prefixes_to_be_created[i]))
-            self.assertEqual(str(response.data["namespace"]), self.absolute_api_url(prefix.namespace))
+            self.assertEqual(str(response.data["namespace"]["url"]), self.absolute_api_url(prefix.namespace))
```
HanlinMiao, line 225: optional, but shall we check the `object_type` and the `id`  attributes are rendered correctly as well?
Edit: This response is `depth=0` correct? If it is `depth=1`, then nvm.

nautobot/ipam/tests/test_api.py
```diff
@@ -343,7 +343,7 @@ def test_create_single_available_ip(self):
             }
             response = self.client.post(url, data, format="json", **self.header)
             self.assertHttpStatus(response, status.HTTP_201_CREATED)
-            self.assertEqual(str(response.data["parent"]), self.absolute_api_url(prefix))
+            self.assertEqual(str(response.data["parent"]["url"]), self.absolute_api_url(prefix))
```
HanlinMiao, line 346: optional, but shall we check the `object_type` and the `id`  attributes are rendered correctly as well?

nautobot/ipam/tests/test_api.py
bryanculver: The base test case `test_list_objects_depth_0` should be covering most of this. This is just validating the data that was returned. I feel that's redundant.

nautobot/ipam/tests/test_api.py
bryanculver: Same feedback as above

nautobot/core/api/serializers.py
```diff
@@ -100,6 +101,22 @@ def fields(self):
         return self.__pruned_fields
 
 
+@extend_schema_field(
+    {
+        "type": "object",
+        "properties": {
+            "id": {
+                "type": "string",
+                "format": "uuid",
```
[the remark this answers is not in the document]
glennmatthews: Might be able to use:

```python
"id": {
    "oneOf": [
        {"type": "string", "format": "uuid"},
        {"type": "integer"},
    ]
}
```

glennmatthews (approved):

nautobot/core/api/serializers.py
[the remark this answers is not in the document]
HanlinMiao: Agreed, `oneOf` should work. https://stackoverflow.com/questions/24023536/how-do-i-require-one-field-or-another-or-one-of-two-others-but-not-all-of-them

HanlinMiao (approved):

tasks.py
jathanism, line 630: Just for my own knowledge can we not use `npx` here?

nautobot/core/settings.py
```diff
@@ -814,7 +814,7 @@ def UI_RACK_VIEW_TRUNCATE_FUNCTION(device_display_name):
     # "NautobotHyperlinkedRelatedField": {"type": "string", "readOnly": True},
     # type=string results in a free text field; also not what we want. For now,
     # however, this will keep things moving so the unit tests pass.
-    "NautobotHyperlinkedRelatedField": {"type": "string", "format": "uuid"},
+    "NautobotHyperlinkedRelatedField": {"type": "object"},
```
jathanism, line 817: My only concern here is what implications this will have for the field-level rendering and validation in RJSF and how it behaves in the React UI. We very likely will have to do some sort of custom template for this field mapping.

nautobot/core/testing/api.py
```diff
@@ -253,13 +253,23 @@ def test_list_objects_depth_0(self):
...
+                            # The response should be a brief API object, containing an ID, object_type, and URL ending in the UUID of the relevant object
                             # http://nautobot.example.com/api/circuits/providers/<uuid>/
                             #                                                    ^^^^^^
-                            self.assertTrue(is_uuid(response_data[field].split("/")[-2]))
+                            self.assertTrue(is_uuid(url.split("/")[-2]))
+                            self.assertTrue(is_uuid(pk))
+
+                            with self.subTest(f"Assert object_type {object_type} is valid"):
+                                app_label, model_name = object_type.split(".")
+                                ContentType.objects.get(app_label=app_label, model=model_name)
```
jathanism, line 272:
```suggestion
                                ContentType.objects.get_for_model(model)
```

jathanism (approved):
One suggestion to remove a line of code. But this is seaworthy! 🏴‍☠️

nautobot/core/testing/api.py
bryanculver: @jathanism Where do I get `model` from here?

jathanism: Oops. Disregard. I crossed wires there.

tasks.py
bryanculver: From my finding `npx` can sometimes install the utility for you. In this case I want to ensure the version provided in `package.json` is used or fail otherwise (since I don't want to run mismatched versions of markdownlint). Might be preferred to use `npm exec --` or `npx --no-install` in all cases.

Issue #3882, opened 2023-06-09
Remove deprecated distutils dependency


snaselj:
### Proposed Changes

Remove deprecated `distutils` dependency.

### Justification

`distutils` are deprecated in Python 3.10, to be removed in Python 3.12.

PR #3884, opened 2023-06-09
Remove deprecated distutils dependency

snaselj:
# Closes: #3882 

`distutils` was used only in `is_truthy` implementation.

# What's Changed

Refactored `is_truthy` to use the same logic as deprecated `distutils.strtobool`.

### DIFF
--- a/changes/3882.fixed
+++ b/changes/3882.fixed
@@ -1,0 +1,1 @@
+Remove deprecated distutils dependency.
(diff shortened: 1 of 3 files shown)


glennmatthews (approved):
Thank you!

jvanderaa:
Would this make sense to put into netutils ?

Issue #3635, opened 2023-04-21
IP-Interface M2M: Validate GraphQL Representation


bryanculver:
GraphQL changes (see: https://github.com/nautobot/nautobot/pull/3399/files#diff-9687b347d7a3698e4a385d4fd99ceea3b22424c81f9ec9cdd74bc9a558ac30d1R25)

- M2Ms "Just Work ™️ "
- Need to clean up TODO comments here for no longer relevant fields

glennmatthews:
Closed by #3899.

PR #3899, opened 2023-06-12
IPAM API and GraphQL fixes

glennmatthews:
# Closes: #3635 
# What's Changed

- Remove `interface` and `vminterface` fields from IPAddress GraphQL type (replaced now by automatic representation of `interfaces` and `vm_interfaces` M2Ms).
- Remove `ip_addresses` fields from Interface and VMInterface GraphQL types (verified that they're automatically represented correctly now that they're M2M)
- Add explicit cast of `ip_version` to an int in GraphQL because by default it becomes an enum, and enums can't start with digits, so `4` and `6` were getting mapped to `"A_4"` and `"A_6"` which is undesirable.
- Re-enable a couple of GraphQL tests that were temporarily disabled, and update them where needed
- Document a couple more IPAM API/ORM changes that I noticed were missing from the tables
- Add missing `interfaces` and `vm_interfaces` to the REST API representation of `IPAddress`.

image

# TODO

- [x] Unit, Integration Tests


bryanculver (approved):

nautobot/ipam/graphql/types.py
```diff
@@ -9,22 +9,13 @@ class IPAddressType(gql_optimizer.OptimizedDjangoObjectType):
     """Graphql Type Object for IPAddress model."""
 
     address = graphene.String()
-    interface = graphene.Field("nautobot.dcim.graphql.types.InterfaceType")
-    vminterface = graphene.Field("nautobot.virtualization.graphql.types.VMInterfaceType")
+    ip_version = graphene.Int()
```
jathanism, line 12: I'm actually surprised this needs to be added. I was pretty sure this would come through from the model just fine.

jathanism (approved):
:shipit:

nautobot/ipam/graphql/types.py
glennmatthews: I was surprised too. All I can figure is that graphene takes the fact that the field has `choices` as an indicator that it should be treated as an enum type instead of an integer type.

Issue #3898, opened 2023-06-12
Bug when trying to overload `natural_key_field_names` with nested related fields


jathanism:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0b1
* Python version: 3.9
* Database platform, version: n/a
* Middleware(s): n/a

### Steps to Reproduce
1. Replace `IPAddress.natural_key_field_names` with `["parent__namespace__name", "host"]`
2. Get an instance of an `IPAddress`
3. Attempt to call `IPAddress.objects.get_by_natural_key(*instance.natural_key())` to perform a symmetric lookup

### Expected Behavior

Should return the same instance that was used to make the call.

### Observed Behavior

A `FieldDoesNotExist` error is raised:

```python
In [1]: instance = IPAddress.objects.first()

In [2]: instance.natural_key()
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
File ~/Library/Caches/pypoetry/virtualenvs/nautobot-b2ttWva6-py3.10/lib/python3.10/site-packages/django/db/models/options.py:608, in Options.get_field(self, field_name)
    605 try:
    606     # Retrieve field instance by name from cached or just-computed
    607     # field map.
--> 608     return self.fields_map[field_name]
    609 except KeyError:

KeyError: 'parent__namespace__name'

During handling of the above exception, another exception occurred:

FieldDoesNotExist                         Traceback (most recent call last)
Cell In[2], line 1
----> 1 instance.natural_key()

File ~/sandbox/src/nautobot/nautobot/core/models/__init__.py:90, in BaseModel.natural_key(self)
     83 """
     84 Smarter default implementation of natural key construction.
     85
     86 1. Handles nullable foreign keys (https://github.com/wq/django-natural-keys/issues/18)
     87 2. Handles variadic natural-keys (e.g. Location model - [name, parent__name, parent__parent__name, ...].)
     88 """
     89 vals = []
---> 90 for lookups in [lookup.split("__") for lookup in self.natural_key_field_lookups]:
     91     val = self
     92     for lookup in lookups:

File ~/Library/Caches/pypoetry/virtualenvs/nautobot-b2ttWva6-py3.10/lib/python3.10/site-packages/django/utils/functional.py:61, in classproperty.__get__(self, instance, cls)
     60 def __get__(self, instance, cls=None):
---> 61     return self.fget(cls)

[...]

Issue #3752, opened 2023-05-15
Audit natural key fields for all models


glennmatthews:
After #3423, Nautobot's data models almost all have natural keys, most of which are derived automatically from the `unique` or `unique_together` settings on each model. As a sanity measure after completing #2908 and #2909, but before releasing 2.0.0, we should audit the natural-key fields for every model and make sure they are the most appropriate set of fields to use as a natural key. Among other considerations:

- a single `unique=True` field cannot be the natural key by itself if it is also `null=True`, so in such a case a composite natural key needs to be identified and explicitly specified in the model definition
- in cases where a model has multiple `unique_together` constraints, the first such constraint may not in fact be the most logically appropriate natural key, especially if any of its components are `null=True` (are we seeing a pattern here?)
```[tasklist]
### Tasks
- [ ] https://github.com/nautobot/nautobot/issues/3972
- [ ] https://github.com/nautobot/nautobot/issues/3973
- [ ] https://github.com/nautobot/nautobot/issues/3974
- [ ] https://github.com/nautobot/nautobot/issues/3975
- [ ] https://github.com/nautobot/nautobot/issues/3976
- [ ] https://github.com/nautobot/nautobot/issues/3977
```

bryanculver:
Someone can condense the natural key fields for all models into a simple spreadsheet to review then we can discuss them dogpile-style as to if they are the best fields or not.

glennmatthews:
| Model | natural_key_fields | Comments |
| - | - | - |
| circuits.Circuit | | |
| circuits.CircuitTermination | | |
| circuits.CircuitType | | |
| circuits.Provider | | |
| circuits.ProviderNetwork | | |
| dcim.Cable | | |
| dcim.CablePath | | |
| dcim.ConsolePort | | |
| dcim.ConsolePortTemplate | | |
| dcim.ConsoleServerPort | | |
| dcim.ConsoleServerPortTemplate | | |
| dcim.Device | | |
| dcim.DeviceBay | | |
| dcim.DeviceBayTemplate | | |
| dcim.DeviceRedundancyGroup | | |
| dcim.DeviceType | | |
| dcim.FrontPort | | |
| dcim.FrontPortTemplate | | |
| dcim.Interface | | |
| dcim.InterfaceTemplate | | |
| dcim.InventoryItem | | |
| dcim.Location | | |
| dcim.LocationType | | |
| dcim.Manufacturer | | |
| dcim.Platform | | |
[...]

PR #3903, opened 2023-06-12
Natural key definition fixes and fix IPAddress.natural_key_field_names

glennmatthews:
# Closes: #3898 
# What's Changed

- Add logic in `natural_key_field_lookups()` generic method to handle `natural_key_field_names` definitions that include traversal of related fields (e.g. `natural_key_field_names = ["parent__namespace__name", "host"]`)
- Remove logic in `get_by_natural_key()` generic method that was unnecessarily enforcing that related objects could *only* be referenced by their natural key and never by other fields.
- Add logic in `get_by_natural_key()` to handle the usage error of `get_by_natural_key(obj.natural_key())` (it's supposed to be `get_by_natural_key(*obj.natural_key())`) with a helpful warning message.
- Change `IPAddress.natural_key_field_names` to `["parent__namespace", "host"]` (which is implicitly equivalent to `["parent__namespace__name", "host"]` since the natural key for a Namespace is just its `name`).

```python
>>> IPAddress.objects.first().natural_key()
['Global', '8.8.8.8']
>>> IPAddress.natural_key_args_to_kwargs(['Global', '8.8.8.8'])
{'parent__namespace__name': 'Global', 'host': '8.8.8.8'}
>>> IPAddress.objects.get_by_natural_key(*IPAddress.objects.first().natural_key())
<IPAddress: 8.8.8.8/32>
>>> Prefix.objects.get_by_natural_key(Prefix.objects.first().natural_key())
19:08:38.121 WARNING nautobot.core.models.managers managers.py                 get_by_natural_key() :
  Prefix.objects.get_by_natural_key() was called with a single list as its args, instead of a list of individual args. Did you forget a '*' in your call?
<Prefix: 0.0.0.0/0>
>>> Prefix.objects.get_by_natural_key(*Prefix.objects.first().natural_key())
<Prefix: 0.0.0.0/0>
```

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/ipam/models.py
+++ b/nautobot/ipam/models.py
@@ -972,8 +972,7 @@ class IPAddress(PrimaryModel):
             self.mask_length = address.prefixlen
             self.ip_version = address.version
 
-    # TODO: The current IPAddress model has no appropriate natural key available yet.
-    natural_key_field_names = ["id"]
+    natural_key_field_names = ["parent__namespace", "host"]
 
     def clean(self):
         super().clean()
--- a/changes/3898.changed
+++ b/changes/3898.changed
@@ -1,0 +1,1 @@
+Changed the `IPAddress` natural key definition (`IPAddress.natural_key_field_names`) to `[parent__namespace, host]`.
--- a/changes/3898.fixed
+++ b/changes/3898.fixed
@@ -1,0 +1,1 @@
+Fixed several errors that could occur when defining a model's `natural_key_field_names` to include related object lookups.
--- a/nautobot/core/models/__init__.py
+++ b/nautobot/core/models/__init__.py
@@ -153,7 +153,14 @@ class BaseModel(models.Model):
         # Next, for any natural key fields that have related models, get the natural key for the related model if known
         natural_key_field_lookups = []
         for field_name in natural_key_field_names:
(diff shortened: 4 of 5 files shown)


nautobot/ipam/models.py
```diff
@@ -972,8 +972,7 @@ def _deconstruct_address(self, address):
             self.mask_length = address.prefixlen
             self.ip_version = address.version
 
-    # TODO: The current IPAddress model has no appropriate natural key available yet.
-    natural_key_field_names = ["id"]
+    natural_key_field_names = ["parent__namespace", "host"]
```
HanlinMiao, line 975: I noticed that `Prefix` does not have this attribute `natural_key_field_names`. Do you think we need one?

jathanism (approved):
Brilliant and almost net zero code! :)

HanlinMiao (approved):

nautobot/ipam/models.py
glennmatthews: That falls within scope of #3752, but I think it's OK for now - Prefix has `unique_together = ["namespace", "network", "prefix_length"]` which will automatically be used as its default natural-key field names.

Issue #3634, opened 2023-04-21
IP-Interface M2M: IPAddressFilterSet Filter on Interface Relationships


bryanculver:
```[tasklist]
### Tasks
- [ ] Restore `IPAddressFilterSet.assigned_to_interface` as a `RelatedMembershipBooleanFilter` on `interfaces` and `vm_interfaces` field as `has_interface_assignments`
- [ ] Restore `present_in_vrf` filter on `IPAddressFilterSet`
```

HanlinMiao:
Closed by #3915

Issue #3858, opened 2023-06-05
2.0 legacy view dynamic filter form is broken


gsnider2195:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0a2
* Python version: 3.10.11
* Database platform, version: MySQL
* Middleware(s):

### Steps to Reproduce
1. Go to https://next.demo.nautobot.com/dcim/devices/
2. Open up the filter form and click on the advanced tab
3. Select any filter field and try to select a lookup type

### Expected Behavior

The lookup type dropdown populates with valid options

### Observed Behavior

Nothing loads in the dropdown and the browser generates a JQuery error:

```
jquery-3.6.0.min.js:2 jQuery.Deferred exception: Cannot read properties of undefined (reading 'includes') TypeError: Cannot read properties of undefined (reading 'includes')
    at https://next.demo.nautobot.com/static/js/forms.js?v2.0.0a3:293:40
    at Array.reduce (<anonymous>)
    at n.processResults (https://next.demo.nautobot.com/static/js/forms.js?v2.0.0a3:263:39)
    at Object.<anonymous> (https://next.demo.nautobot.com/static/select2-4.0.13/select2.min.js:2:38054)
    at e (https://next.demo.nautobot.com/static/jquery/jquery-3.6.0.min.js:2:30038)
    at t (https://next.demo.nautobot.com/static/jquery/jquery-3.6.0.min.js:2:30340) undefined
```

bryanculver:
Culprit: https://github.com/nautobot/nautobot/compare/develop...next#diff-8906c95050cdc47ee55226257d68c10271005536e061de29fc6a256d4bbfe102R290-R295

HanlinMiao:
Closed by #3634

PR #3915, opened 2023-06-13
IP-Interface M2M: IPAddressFilterSet Filter on Interface Relationships

HanlinMiao:
# Closes: #3634, #3858
# What's Changed

1. Restored `assigned_to_interface` as a `RelatedMembershipBooleanFilter` as `has_interface_assignments` on `IPAddressFilterSet`.
2. Restored `present_in_vrf` and `present_in_vrf_id` filters on `IPAddressFilterSet`.
3. Re-enabled `IPAddressFilterSet` Test Case.
4. Added logic to `test_boolean_filters_generic()` that if a custom `method` is defined, do not use the default logic.
5. Fixed broken Dynamic Filter Form in Legacy UI in form.js.

# TODO

- [X] Unit, Integration Tests


nautobot/core/testing/filters.py
```diff
@@ -124,6 +124,8 @@ def test_boolean_filters_generic(self):
             for filter_name, filter_object in self.filterset.get_filters().items():
                 if not isinstance(filter_object, RelatedMembershipBooleanFilter):
                     continue
+                if filter_object.method is not None:
```
HanlinMiao, line 127: Custom method e.g `has_interface_assignments` does not fit the default boolean logic here. So we ignore that and redefine it in the test case itself.

nautobot/project-static/js/forms.js
```diff
@@ -292,7 +292,7 @@ function initializeDynamicChoiceSelection(context, dropdownParent=null){
                         // DynamicGroupSerializer has a `children` field which fits an inappropriate if condition
                         // in select2.min.js, which will result in the incorrect rendering of DynamicGroup DynamicChoiceField.
                         // So we nullify the field here since we do not need this field.
-                        if (record.url.includes("dynamic-groups")){
+                        if (record?.url ? record.url.includes("dynamic-groups") : false){
```
HanlinMiao, line 295: Check if url exists in the record first.

nautobot/core/api/serializers.py
```diff
@@ -152,6 +152,8 @@ def to_representation(self, value):
         url = super().to_representation(value)
         if self.queryset:
             model = self.queryset.model
+        elif isinstance(value, Model):
```
HanlinMiao, line 155: This fixes the mishandling of ManyRelatedField from the slack thread https://networktocode-llc.slack.com/archives/C01RMAGBD29/p1686671551128129

nautobot/ipam/tests/test_filters.py
```diff
@@ -443,32 +448,28 @@ def setUpTestData(cls):
...
-            vrf=vrfs[2],
             status=statuses[1],
             role=roles[1],
             dns_name="ipaddress-d",
             namespace=cls.namespace,
         )
         interfaces[2].add_ip_addresses(ip2)
         IPAddress.objects.create(
-            address="10.0.0.1/25",
+            address="10.0.0.5/24",
```
HanlinMiao, line 458: `parent_id` and `host` uniqueness constraint

changes/3634.fixed
```diff
@@ -0,0 +1 @@
+Fixed broken dynamic filter form in legacy view.
```
glennmatthews, line 1: Should this be 3858.fixed?

nautobot/core/api/serializers.py
glennmatthews: Fixed more generally in #3916.

nautobot/ipam/filters.py
```diff
@@ -397,11 +392,24 @@ class IPAddressFilterSet(
...
+        field_name="interfaces",
+        method="_assigned_to_interface",
+        label="Has Interface Assignments",
+    )
 
     class Meta:
         model = IPAddress
         fields = ["id", "ip_version", "dns_name", "type", "tags", "mask_length"]
 
+    def _assigned_to_interface(self, queryset, name, value):
```
glennmatthews, line 405: Does this need to follow the `generate_query__...` pattern for reversibility by dynamic-groups?

nautobot/ipam/tests/test_filters.py
```diff
@@ -603,25 +601,25 @@ def test_filter_address(self):
 
     def test_mask_length(self):
         # Test filtering by a single integer value
-        params = {"mask_length": self.queryset.first().mask_length}
+        params = {"mask_length": [self.queryset.first().mask_length]}
         self.assertQuerysetEqualAndNotEmpty(
-            self.filterset(params, self.queryset).qs, self.queryset.filter(mask_length=params["mask_length"])
+            self.filterset(params, self.queryset).qs, self.queryset.filter(mask_length__in=params["mask_length"])
         )
```
glennmatthews, line 607: Is this set of changes necessary? Does it represent a loss of functionality (before could be a single value or a list, now must always be a list)?

nautobot/ipam/tests/test_filters.py
```diff
@@ -674,6 +672,71 @@ def test_vm_interfaces(self):
...
+            self.filterset(params, self.queryset).qs,
+            self.queryset.filter(Q(interfaces__isnull=False) | Q(vm_interfaces__isnull=False)),
+        )
+        params = {"has_interface_assignments": False}
+        self.assertQuerysetEqualAndNotEmpty(
+            self.filterset(params, self.queryset).qs,
+            self.queryset.filter(Q(interfaces__isnull=True) & Q(vm_interfaces__isnull=True)),
+        )
+
+    def test_present_in_vrf(self):
```
glennmatthews, line 687: Huh. Did we just not have any tests for `present_in_vrf` before?

HanlinMiao: Yeah I dont think so.

nautobot/ipam/tests/test_filters.py
HanlinMiao: Enhanced `multivalue_field_factory()` to handle passing in a single integer as well since this field is mapped to `MutliValueNumberFilter`

nautobot/ipam/filters.py
```diff
@@ -397,10 +393,29 @@ class IPAddressFilterSet(
         to_field_name="name",
         label="Namespace (name or ID)",
     )
+    has_interface_assignments = RelatedMembershipBooleanFilter(
+        field_name="interfaces",
+        method="_assigned_to_interface",
+        label="Has Interface Assignments",
+    )
+    ip_version = MultiValueNumberFilter()
```
HanlinMiao, line 401: For some strange reason, `ip_version` is not mapped to `MultiValueNumberFilter`. Instead it is mapped to `MultipleChoiceFilter`. Maybe because its `models.IntegerField` has a `choices` attribute specified, so `django_filters's` default lookup mapped it to a `MultipleChoiceFilter`. We have to override it here to have the correct filter behavior.

nautobot/ipam/filters.py
```diff
@@ -397,10 +393,29 @@ class IPAddressFilterSet(
...
+        if value is not None:
+            if value:
+                return Q(interfaces__isnull=False) | Q(vm_interfaces__isnull=False)
+            else:
+                return Q(interfaces__isnull=True) & Q(vm_interfaces__isnull=True)
+        return Q()
+
+    def _assigned_to_interface(self, queryset, name, value):
+        params = self.generate_query__interface_assignments(value)
+        return queryset.filter(params)
```
bryanculver, line 418:
```suggestion
    def generate_query___has_interface_assignments(self, value):
        """Helper method used by DynamicGroups and by _assigned_to_interface method."""
        if value is not None:
            if value:
                return Q(interfaces__isnull=False) | Q(vm_interfaces__isnull=False)
            else:
                return Q(interfaces__isnull=True) & Q(vm_interfaces__isnull=True)
        return Q()

    def _has_interface_assignments(self, queryset, name, value):
        params = self.generate_query___has_interface_assignments(value)
        return queryset.filter(params)
```

nautobot/ipam/filters.py
```diff
@@ -397,10 +393,29 @@ class IPAddressFilterSet(
         to_field_name="name",
         label="Namespace (name or ID)",
     )
+    has_interface_assignments = RelatedMembershipBooleanFilter(
+        field_name="interfaces",
+        method="_assigned_to_interface",
+        label="Has Interface Assignments",
+    )
```
bryanculver, line 400:
```suggestion
    has_interface_assignments = RelatedMembershipBooleanFilter(
        field_name="interfaces",
        method="_has_interface_assignments",
        label="Has Interface Assignments",
    )
```

nautobot/ipam/filters.py
glennmatthews: +1. It's an enum rather than a true number filter.

glennmatthews (approved):

Issue #3930, opened 2023-06-15
Inconsistent test behavior in 2.0


gsnider2195:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0b1
* Python version: 3.10
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Run `invoke unittest -l nautobot.extras.tests.test_jobs`
2. Run `invoke unittest -l nautobot.extras.tests.test_jobs.JobHookTest`

### Expected Behavior

Both tests behave the same

### Observed Behavior

`JobHookTest` fails when running independently

glennmatthews:
Fixed by #3933 and #3931.

PR #3931, opened 2023-06-15
Unit test speed optimization

gsnider2195:
# Closes: #DNE
# What's Changed

- While testing fixes for #3930 I got annoyed that running tests from one file took almost 13 minutes so I moved some of the tests out of the `TransactionTestCase` test classes that didn't need to be there, reducing the run time by 5 1/2 minutes on my machine

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/tests/test_jobs.py
```diff
@@ -35,9 +35,147 @@
 from nautobot.extras.models.models import JobLogEntry
 
 
-class JobTest(TransactionTestCase):
+class JobTest(TestCase):
     """
-    Test basic jobs to ensure importing works.
+    Test job features that don't require a transaction test case.
```
jathanism, line 40: How can we reasonably determine which tests do or do not require transaction test cases?

gsnider2195: I left anything that uses `create_job_result_and_run_job` in a transaction test case

glennmatthews:
Latest CI runs don't look notably faster, but that may just fall within normal variance?

glennmatthews (approved):

gsnider2195:
> Latest CI runs don't look notably faster, but that may just fall within normal variance?

I only tested it once locally but it was significantly faster. 780 seconds down to 460

PR #3933, opened 2023-06-16
Hopefully address noisy job output about logs

glennmatthews:
# Closes: #n/a
# What's Changed

Instead of letting `CELERY_TASK_ALWAYS_EAGER` affect the code flow of how we enqueue jobs, just let it affect how we set up logging for jobs. Tested locally with a few sets of tests and it appears to fix the issue.

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/models/jobs.py
+++ b/nautobot/extras/models/jobs.py
@@ -19,7 +19,6 @@ from django_celery_beat.clockedschedule import clocked
 from prometheus_client import Histogram
 
 from nautobot.core.celery import (
-    add_nautobot_log_handler,
     app,
     NautobotKombuJSONEncoder,
     setup_nautobot_job_logging,
--- a/nautobot/extras/models/jobs.py
+++ b/nautobot/extras/models/jobs.py
@@ -653,7 +653,12 @@ class JobResult(BaseModel, CustomFieldModel):
         if celery_kwargs is not None:
             job_celery_kwargs.update(celery_kwargs)
 
+        # When running synchronously (explicitly on this call or globally), the celery worker
+        # setup_nautobot_job_logging() doesn't get called by a worker signal, so we need to do it here.
         if synchronous or getattr(settings, "CELERY_TASK_ALWAYS_EAGER", False):
+            setup_nautobot_job_logging(None, None, app.conf)
+
+        if synchronous:
             # synchronous tasks are run before the JobResult is saved, so any fields required by
             # the job must be added before calling `apply()`
             job_result.celery_kwargs = job_celery_kwargs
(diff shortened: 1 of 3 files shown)


nautobot/extras/models/jobs.py
```diff
@@ -653,16 +652,19 @@ def enqueue_job(
...
+            setup_nautobot_job_logging(None, None, app.conf)
+
+        if synchronous:
             # synchronous tasks are run before the JobResult is saved, so any fields required by
             # the job must be added before calling `apply()`
             job_result.celery_kwargs = job_celery_kwargs
             job_result.save()
 
             # setup synchronous task logging
             redirect_logger = get_logger("celery.redirected")
```
gsnider2195, line 667: This comment can be removed and the following line moved down below the next comment

glennmatthews: Good thought. Will do once CI verifies that the fix is otherwise valid.

nautobot/extras/models/jobs.py
```diff
@@ -653,16 +652,19 @@ def enqueue_job(
         if celery_kwargs is not None:
             job_celery_kwargs.update(celery_kwargs)
 
+        # When running synchronously (explicitly on this call or globally), the celery worker
+        # setup_nautobot_job_logging() doesn't get called by a worker signal, so we need to do it here.
         if synchronous or getattr(settings, "CELERY_TASK_ALWAYS_EAGER", False):
+            setup_nautobot_job_logging(None, None, app.conf)
```
gsnider2195, line 658: My only concern with this instead of calling it in the test runner setup is this runs for every call of `enqueue_job` during testing

gsnider2195: It's not an expensive process but it is something that we should consider when thinking about optimizing our test run time

glennmatthews: That's a thought... I can see whether we can move it there.

gsnider2195 (approved):

gsnider2195:
Closes #3930

nautobot/extras/models/jobs.py
```diff
@@ -653,18 +652,17 @@ def enqueue_job(
...
-        if synchronous or getattr(settings, "CELERY_TASK_ALWAYS_EAGER", False):
+        if synchronous:
             # synchronous tasks are run before the JobResult is saved, so any fields required by
             # the job must be added before calling `apply()`
             job_result.celery_kwargs = job_celery_kwargs
             job_result.save()
 
             # setup synchronous task logging
-            redirect_logger = get_logger("celery.redirected")
-            add_nautobot_log_handler(redirect_logger)
```
jathanism, line 664: So is `add_nautobot_log_handler()` no longer needed for anything?

Issue #3935, opened 2023-06-16
2.0 IPAM Overhaul: Add constraints to Prefix.parent


gsnider2195:
- [ ] Prefix type `network` can only contain IP addresses and prefix of type `pool`
- [ ] Prefix type `container` can only contain child prefixes of type `container` or `network`
- [ ] Prefix `type` is user serviceable and defaults to Container
- [ ] ~Bring back the `is_pool` property~
- [ ] IP Address parent property cannot be set to a prefix with type `pool`, it will parent to the nearest network prefix
- [ ] Pools cannot contain child prefixes or ip addresses

### Prefix save/delete methods

- [ ] Network contains IP addresses but we want to subdivide the network (/24 converted to /25)
  - [ ] Newly created prefix will contain all of the parent prefix's IP addresses, parent will be converted to a container automatically
  - [ ] Newly created prefix will NOT contain all of the parent prefix's IP addresses. Do we fail on creation of the new prefix?
- [ ] Network contains IP addresses but we want to delete the network and reparent all of the IPs to the parent (/25 converted to /24)

### IP Address save method

- [ ] IP Address created in a container prefix that does not contain any prefixes. Does the creation of the IP address automatically convert the parent container to a network?

gsnider2195:
Completed by #3942

PR #3942, opened 2023-06-16
IPAM 2.0: Add prefix parent constraints

gsnider2195:
# Closes: #3935
# What's Changed

- Added constraints to the `parent` field on `Prefix` based on the prefix's `type` and the parent's `type`.

# TODO

- [ ] Unit, Integration Tests


changes/3935.added
```diff
@@ -0,0 +1 @@
+Added constraints to the `parent` field on `Prefix` based on the prefixs and the parents `type`.
```
glennmatthews, line 1:
```suggestion
Added constraints to the `parent` field on `Prefix` based on the prefix's `type` and the parent's `type`.
```

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -91,6 +91,15 @@ The following changes have been made to the `Prefix` model.
 |------------------------|-----------------|
 | `get_child_prefixes()` | `descendants()` |
 
+#### Prefix Parenting Constraints
+
+The following constraints have been added to the `Prefix` model:
```
glennmatthews, line 96: Maybe?

```suggestion
The following constraints have been added to the `Prefix` model in order to ensure more accurate network modeling:
```

nautobot/ipam/models.py
```diff
@@ -18,6 +18,7 @@
 from nautobot.extras.models import RoleField, StatusField
 from nautobot.extras.utils import extras_features
 from nautobot.ipam import choices
+from nautobot.ipam import constants
```
jathanism, line 21: Can this not just be on the line above, comma-separated?

nautobot/ipam/models.py
```diff
@@ -558,26 +559,44 @@ def delete(self, *args, **kwargs):
...
+                    and self.parent is not None
+                    and constants.PREFIX_ALLOWED_PARENT_TYPES[instance.type] != self.parent.type
+                ):
+                    raise models.ProtectedError(
+                        msg=(
+                            f"Cannot delete Prefix {self} because it has child Prefix objects that "
+                            "would no longer have a valid parent."
+                        ),
+                        protected_objects=err.protected_objects,
+                    ) from err
```
glennmatthews, line 598: Do we need an `else` case here or are there certain `protected_objects` that we're OK with dropping on the floor?

nautobot/ipam/models.py
```diff
@@ -596,6 +615,30 @@ def save(self, *args, **kwargs):
             parent = max(supernets, key=operator.attrgetter("prefix_length"))
             self.parent = parent
 
+        # Validate that creation of this prefix does not create an invalid parent/child relationship
+        if self.parent and self.parent.type != constants.PREFIX_ALLOWED_PARENT_TYPES[self.type]:
+            err_msg = f"{self.type.title()} prefixes cannot be created in {self.parent.type} prefixes"
```
jathanism, line 619: This error is a tad ambiguous. Perhaps something like?
```suggestion
            err_msg = f"{self.type.title()} prefixes cannot be children of {self.parent.type.title()} prefixes"
```

nautobot/ipam/models.py
```diff
@@ -596,6 +614,30 @@ def save(self, *args, **kwargs):
...
+        invalid_children = Prefix.objects.filter(
+            ~models.Q(id=self.id),
+            ~models.Q(type__in=constants.PREFIX_ALLOWED_CHILD_TYPES[self.type]),
+            parent_id=self.parent_id,
+            prefix_length__gt=self.prefix_length,
+            ip_version=self.ip_version,
+            network__gte=self.network,
+            broadcast__lte=self.broadcast,
+            namespace=self.namespace,
+        )
```
glennmatthews, line 631: Fun stuff!

nautobot/ipam/models.py
```diff
@@ -596,6 +614,30 @@ def save(self, *args, **kwargs):
...
+
+        if invalid_children.exists():
+            invalid_child_prefixes = [
+                f"{child.cidr_str} ({child.type})" for child in invalid_children.only("network", "prefix_length")
+            ]
+            err_msg = f'Creating prefix "{self.prefix}" in namespace "{self.namespace}" with type "{self.type}" '
```
glennmatthews, line 637: Is this only an issue on `create` or do we need to worry about changing the `type` of an existing `Prefix`?

nautobot/ipam/models.py
```diff
@@ -596,6 +615,30 @@ def save(self, *args, **kwargs):
             parent = max(supernets, key=operator.attrgetter("prefix_length"))
             self.parent = parent
 
+        # Validate that creation of this prefix does not create an invalid parent/child relationship
+        if self.parent and self.parent.type != constants.PREFIX_ALLOWED_PARENT_TYPES[self.type]:
+            err_msg = f"{self.type.title()} prefixes cannot be created in {self.parent.type} prefixes"
+            raise ValidationError({"type": err_msg})
+
+        invalid_children = Prefix.objects.filter(
```
jathanism, line 622: Can you please add a brief comment explaining the intent of this query for posterity? Line 625 is the one that might be a gotcha (e.g. "NOT including prefixes of my type")...

nautobot/ipam/models.py
HanlinMiao, line 631: Quick question for myself, what is this query filtering on?

nautobot/ipam/models.py
```diff
@@ -596,6 +615,30 @@ def save(self, *args, **kwargs):
...
+            broadcast__lte=self.broadcast,
+            namespace=self.namespace,
+        )
+
+        if invalid_children.exists():
+            invalid_child_prefixes = [
+                f"{child.cidr_str} ({child.type})" for child in invalid_children.only("network", "prefix_length")
+            ]
+            err_msg = f'Creating prefix "{self.prefix}" in namespace "{self.namespace}" with type "{self.type}" '
+            err_msg += f"would create an invalid parent/child relationship with prefixes {invalid_child_prefixes}"
```
jathanism, line 638: Can you change this to a multi-line string instead?
```suggestion
            err_msg = (
                f'Creating prefix "{self.prefix}" in namespace "{self.namespace}" with type "{self.type}" '
                f"would create an invalid parent/child relationship with prefixes {invalid_child_prefixes}"
            )
```

changes/3935.added
gsnider2195: Thanks. Apostrophes got me again!

``` echo 'Added constraints to the `parent` field on `Prefix` based on the prefix's `type` and the parent's `type`.' > changes/3935.added ```

**Review by jathanism — CHANGES_REQUESTED** [review]

File: nautobot/ipam/models.py (hunk 4 above)
Comment (gsnider2195) [reply nautobot/ipam/models.py:598]: I think the else should be: 
```py
elif protected_model not in (Prefix, IPAddress):
    raise
``` 

Need to do more testing though

nautobot/ipam/models.py
gsnider2195: This is filtering on:

- prefixes that share my parent
`parent_id=self.parent_id,`

- and will be reparented to me
```
            prefix_length__gt=self.prefix_length,
            ip_version=self.ip_version,
            network__gte=self.network,
            broadcast__lte=self.broadcast,
            namespace=self.namespace,
```

- but are not the correct type for this parent/child relationship
`~models.Q(type__in=constants.PREFIX_ALLOWED_CHILD_TYPES[self.type]),`

nautobot/ipam/tests/test_querysets.py
```diff
@@ -462,6 +462,8 @@ class PrefixQuerysetTestCase(TestCase):
 
     @classmethod
     def setUpTestData(cls):
+        IPAddress.objects.all().delete()
```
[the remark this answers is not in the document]
gsnider2195: Broken merge I think

nautobot/ipam/models.py
```diff
@@ -558,26 +559,43 @@ def delete(self, *args, **kwargs):
...
+                        ),
+                        protected_objects=err.protected_objects,
+                    ) from err
+
+                # Prefix objects must have a valid parent
+                elif (
+                    protected_model == Prefix
+                    and self.parent is not None
+                    and constants.PREFIX_ALLOWED_PARENT_TYPES[instance.type] != self.parent.type
+                ):
```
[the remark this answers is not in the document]
jathanism: Nooo because if there's no parent then `self.parent.type` would raise an error.

nautobot/ipam/models.py
jathanism: I would just make it an `else: raise` since the previous two are covered above. This is a good call though since there may very likely be protected relationships to other models that we don't want to be subject to any of this logic.

nautobot/ipam/models.py
jathanism: Make this a comment and 💋

nautobot/ipam/tests/test_querysets.py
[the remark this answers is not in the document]
jathanism: Delete them twice for good measure!!

nautobot/ipam/models.py
gsnider2195: `else: raise` means it would always raise. We want to allow the parent to be deleted if reparenting would result in a valid parent/child relationship

nautobot/ipam/models.py
gsnider2195: Added tests that modify `type` in e89a98e72

nautobot/ipam/models.py
```diff
@@ -562,26 +563,45 @@ def delete(self, *args, **kwargs):
...
-                    protected_objects=err.protected_objects,
-                ) from err
+            for instance in err.protected_objects:
+                # This will be either IPAddress or Prefix.
+                protected_model = instance._meta.model
+
+                # IPAddress objects must have a valid parent.
+                if protected_model == IPAddress and (
+                    self.parent is None or self.parent.type != choices.PrefixTypeChoices.TYPE_NETWORK
+                ):
```
gsnider2195, line 581: Since it will never be possible to delete a network prefix with IP addresses I was thinking this could be changed to just `if protected_model == IPAddress:`. Would this cause problems with the workflows described in #3970?

jathanism (approved):

nautobot/ipam/models.py
```diff
@@ -600,6 +620,34 @@ def save(self, *args, **kwargs):
             parent = max(supernets, key=operator.attrgetter("prefix_length"))
             self.parent = parent
 
+        # Validate that creation of this prefix does not create an invalid parent/child relationship
+        if self.parent and self.parent.type != constants.PREFIX_ALLOWED_PARENT_TYPES[self.type]:
+            err_msg = f"{self.type.title()} prefixes cannot be children of {self.parent.type.title()} prefixes"
```
glennmatthews, line 625: What if anything do we need to do in the data migrations to ensure that existing data follows these constraints? Or is that future work (which I'm fine with if so, so long as it's tracked)?

glennmatthews (approved):

nautobot/ipam/models.py
gsnider2195: We didn't account for these constraints when creating the migrations so we'll unfortunately need to circle back to make sure we're not assigning IP addresses to pool prefixes, make sure parent prefixes are containers and likely create new network prefixes to contain IP addresses if necessary. I opened #4004 for this work.

Issue #3859, opened 2023-06-05
Protect against Git-repo jobs clobbering sys.modules


glennmatthews:
### As ...

Patti - Platform Admin

### I want ...

To add Jobs via Git repositories without worrying about them impacting the overall stability of the platform.

After #3840 , there is a risk that a misnamed (or maliciously-named) Git repository could potentially clobber existing Python code. For example, creating a repository and assigning it the slug `nautobot` causes all sorts of havoc as it results in Nautobot unloading itself and then attempting to reimport all of Nautobot's code from the Git repository by that name.

### So that ...

The application is resilient to user error as well as mischievous/malicious actions by authenticated users.

### I know this is done when...

- Creating a GitRepository whose `slug` matches any currently loaded Python module in the Nautobot environment is rejected with an appropriate error message.

### Optional - Feature groups this request pertains to.

- [X] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [X] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

None

### External Dependencies

None

glennmatthews:
As @gsnider2195 pointed out, since we add `GIT_ROOT` to the *end* of `sys.path`, we should also ensure that the git repo slug isn't shadowed by any already-importable module that would come from an earlier `sys.path` entry (e.g. standard library packages like `math` or `traceback`). Possibly `importlib.util.find_spec(gitrepo.slug)` or similar would be an appropriate check.

bryanculver:
Add check in `clean` (see others here: https://github.com/nautobot/nautobot/blob/956c954cb9ff7e913c04e3293bd4d9aabd8fdf17/nautobot/extras/models/datasources.py#L86) implementing the above https://github.com/nautobot/nautobot/issues/3859#issuecomment-1579045188

PR #3943, opened 2023-06-16
Add validation of GitRepository.slug against Python modules

glennmatthews:
# Closes: #3859 
# What's Changed

- Add a check on creating a new GitRepository to make sure its `slug` won't be shadowed by an existing/available Python module.
- Add test coverage
- Change the base class of a couple of other tests in the same file to broaden their test coverage.

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/models/datasources.py
+++ b/nautobot/extras/models/datasources.py
@@ -1,4 +1,5 @@
 """Models for representing external data sources."""
+from importlib.util import find_spec
 import os
 
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


nautobot/extras/models/datasources.py
```diff
@@ -81,13 +82,24 @@ def __str__(self):
...
+        if not self.present_in_database:
+            check_if_key_is_graphql_safe(self.__class__.__name__, self.slug, "slug")
+            # Check on create whether the proposed slug conflicts with a module name already in the Python environment.
+            # Because we add GIT_ROOT to the end of sys.path, trying to import this repository will instead
+            # import the earlier-found Python module in its place, which would be undesirable.
+            if find_spec(self.slug) is not None:
                 raise ValidationError(
-                    f"Slug cannot be changed once set. Current slug is {self.__initial_slug}, "
-                    f"requested slug is {self.slug}"
+                    f'Please choose a different slug, as "{self.slug}" is an existing Python package or module.'
```
bryanculver, line 102: Don't know what would be a better message here but since we aren't inspecting all potential packages or modules I'm wondering if a suggestion that the resulting slug is a protected name would be better.

bryanculver (approved):

nautobot/extras/models/datasources.py
glennmatthews: Would changing `existing` to `installed` be an improvement?

gsnider2195 (approved):

Issue #3953, opened 2023-06-20
Update the "Home" link in Nautobot 2.0 UI


DistantVoyager:
### As ...

Nelly - Network Engineer

### I want ...

to be able to use a working "Home" button and see a more visually-appealing home link

### So that ...

I can navigate to Home in a clear way.

### I know this is done when...

- The current "Home" text is replaced with a working Home icon and link
- The padding/spacing is consistent under the navbar

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

glennmatthews:
See also #3998.

bryanculver:
The link in earlier designs was a fallback from broken breadcrumbs. That was fixed with #3998. For now, "Home" is clicking the logo.

glennmatthews:
We should remove the "Home" breadcrumb from the homepage/dashboard though, as it serves no useful purpose there:

![image](https://github.com/nautobot/nautobot/assets/5603551/db60c743-ac0c-4d00-bb76-3f66da0960ae)

PR #3957, opened 2023-06-21
Update the home link in Nautobot 2.0 UI

DistantVoyager:
# Closes: #3953 <ISSUE NUMBER GOES HERE>
# What's Changed

- Added HomeIcon from `nautobot-ui` and linked to "/".
- Removed static "Home" text.
- Fixed padding to be consistent.

### New:

image

image

### Before:

image

# TODO

### DIFF
--- a/changes/3953.changed
+++ b/changes/3953.changed
@@ -1,0 +1,3 @@
+Added HomeIcon and linked to "/".
+Removed static "Home" text.
+Fixed padding to be consistent.
--- a/nautobot/ui/src/components/Navbar.js
+++ b/nautobot/ui/src/components/Navbar.js
@@ -2,10 +2,12 @@ import { useSelector } from "react-redux";
 import {
     Link as ReactRouterLink,
     NavLink as ReactRouterNavLink,
+    useMatch,
 } from "react-router-dom";
 import {
     AutomationIcon,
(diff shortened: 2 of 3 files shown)


jathanism (approved):
I'm a fan. There was some talk about breadcrumbs and home being there as this home button is not in the mockup for the design, but given we don't address this I think this update is sane.

bryanculver:
@lampwins Can you provide your 2️⃣🪙 here?

glennmatthews:
The `Home` text in the current UI is *supposed* to be breadcrumbs to the current view, e.g.:

image

but looks like that got broken along the way. I think I have a local fix for it, will open an alternative PR soon.

lampwins:
@DistantVoyager I know we had the small "Home" text before, but what is your feeling on needing a home link at all? From my perspective it seems to be standard to just click on the site icon to go to the home page? My concern is simply in taking up space in the context/search bar, knowing more content will be present up there when apps start integrating.

lampwins:
I just don't think it is necessary. Plus, I'd worry about the spacing with branded logos there. Let's not do this for now.

bryanculver:
Combination of #3998 and just developing new habits, logo is sufficient for "Go To Home"

Issue #3900, opened 2023-06-12
Add Migration Testing CI Step


bryanculver:
### Proposed Changes

- Introduce a CI step to start from a database snapshot and perform the migrations included in the branch/PR and perform `nautobot-server migrate`. Step/check passes if the migrations run without error.

- Stretch to get this to work locally

### Justification

Existing migration tests are flaky and aren't capturing errors like #3889 without sufficient rework. This test would more realistically test what users are expected to experience.

PR #3960, opened 2023-06-21
Implement 'migration-test' task, add test dataset for postgres, add CI step

glennmatthews:
# Closes: #3900
# What's Changed

- Add `invoke migration-test` task, verified locally with Docker Compose workflow against both mysql and postgres.
- Add example dataset for postgres + 1.5.16 ~(I don't currently have a passing dataset for mysql, only one that fails migration due to some of the data validation checks we have in IPAM, but will want to add one before this is merged)~ and mysql + 1.5.16
- Make an attempt at writing a CI step using this test (I expect there will be some trial and error to get this right)

# TODO
- [x] Add passing mysql dataset when available
- [x] Add ci_integration workflow step(s) once the ci_pullrequest workflow is confirmed passing
- [x] Unit, Integration Tests


tasks.py
```diff
@@ -799,6 +799,47 @@ def integration_test(
     )
 
 
+@task(
+    help={
+        "dataset": "File (.sql.tar.gz) to start from that will untar to 'nautobot.sql'",
+        "db_engine": "mysql or postgres",
```
gsnider2195, line 805: Instead of requiring `--db_engine` could we introspect the output of either `docker-compose config --format json` or `docker-compose config --images`? In the json format you could look at `json["services"]["db"]["image"]`:

```json
{
  "name": "nautobot",
  "networks": {
    "default": {
      "name": "nautobot_default",
      "ipam": {},
      "external": false
    }
  },
  "services": {
  -- snipped
    "db": {
      "command": null,
      "entrypoint": null,
      "environment": {
        -- snipped
      },
      "healthcheck": {
        "test": [
          "CMD-SHELL",
          "pg_isready"
        ],
        "timeout": "5s",
        "interval": "5s",
        "retries": 50
      },
      "image": "postgres:13",
      "networks": {
        "default": null
      },
      "volumes": [
        {
          "type": "volume",
          "source": "pgdata_nautobot",
          "target": "/var/lib/postgresql/data",
          "volume": {}
        }
      ]
    },
    -- snipped
}
```

The `--images` output is simpler but perhaps more fragile:

```
local/nautobot-dev:local-py
local/nautobot-dev:local-py
postgres:13
local/nautobot-dev:local-py
node:18-slim
redis:6-alpine
selenium/standalone-firefox:4.1.2
```

glennmatthews: That'll work for the local docker workflow, but won't it be a problem for non-docker workflows and for CI?

gsnider2195: Yeah :disappointed:. I always forget about the non docker workflows.

jathanism (approved):

timizuoebideri1 (approved):

Issue #3937, opened 2023-06-16
Pre-Migration Helper for Uniqueness Enforcement Changes on VC, ExportTemplate, ConfigContext, ConfigContextSchema


bryanculver:
See: #2908

A management command in 1.X to check data won't cause migration failures in 2.0 for the above-mentioned models.

PR #3962, opened 2023-06-21
Added a Nautobot 2.0 pre-migration management command aptly named `pr…

jathanism:
…e_migrate`.

# Closes: #3937 
# What's Changed

# TODO

- [ ] Unit, Integration Tests


nautobot/core/management/commands/pre_migrate.py
```diff
@@ -0,0 +1,149 @@
...
+        checks = [
+            check_interface_vrfs,
+            check_configcontext_uniqueness,
+            check_exporttemplate_uniqueness,
+            check_virtualchassis_uniqueness,
+        ]
+
+        for check, args in checks:
+            try:
+                print(f">>> Running check: {check}...")
```
glennmatthews, line 142: Management commands are [recommended](https://docs.djangoproject.com/en/4.2/howto/custom-management-commands/) to use `self.stdout.write` rather than `print`

nautobot/core/management/commands/pre_migrate.py
```diff
@@ -0,0 +1,149 @@
...
+                check(*args)
+            except ValidationError as err:
+                # self.stderr.write(self.style.ERROR(str(err)))
+                # raise CommandError("Pre-migration checks failed.")
+                raise CommandError(str(err))
```
glennmatthews, line 147: Definitely would be good to have it run through all checks before exiting, even if an early check fails.

glennmatthews (commented):
Looks like a solid start! It'd be great to have unit tests exercising this command.

nautobot/core/management/commands/pre_migrate.py
jathanism: Yep, that's coming next.

nautobot/core/management/commands/pre_migrate.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+HELP_TEXT = """
+Performs pre-migration validation checks for Nautobot 2.0.
+
+If the Nautobot 1.5 instance cannot be upgraded, this command will exit uncleanly.
+"""
+
+
+def check_virtualchassis_uniqueness():
```
HanlinMiao, line 21: I wonder if we can use `check_for_duplicates_with_natural_key_fields_in_migration` in `nautobot.core.utils.migrations.py` for these checks?

jathanism: I thought about using some existing migration utils, but since this is going into `develop` I would have to just forklift the code. This is helpful! I could just forklift that and then use it across the board here.

HanlinMiao: My bad i didnt even see it is based from `develop` but yeah should be somewhat useful!

nautobot/core/management/commands/pre_migrate.py
```diff
@@ -0,0 +1,114 @@
+# TOOD(jathan): This file MUST NOT be merged into Nautobot v2 (next).
```
jathanism, line 1: TOOD!!!

nautobot/core/management/commands/pre_migrate.py
```diff
@@ -0,0 +1,114 @@
...
+            func_name = check.__code__.co_name
+            try:
+                self.stdout.write(self.style.WARNING(f">>> Running check: {func_name}..."))
+                check()
+            except ValidationError as err:
+                errors[func_name].append(err)
+
+        if errors:
+            msg = "One or more pre-migration checks failed:\n"
+            for err_item in itertools.chain.from_iterable(errors.values()):
```
glennmatthews, line 107: Pardon my ignorance here - what's this doing that's different from just `for err_item in errors.values()`?

nautobot/core/tests/test_commands.py
```diff
@@ -0,0 +1,66 @@
...
+        """Test that duplicate ConfigContext[Schema] objects result in a failure."""
+        out, err = self.run_command()
+
+        self.assertIn("All pre-migration checks passed.", out)
+        self.assertEqual("", err)
+
+    def test_configcontext_failure(self):
+        """Test that duplicate ConfigContext objects result in a failure."""
+        ConfigContext.objects.create(name="cc1", data={})
+        ConfigContext.objects.create(name="cc1", data={})
```
glennmatthews, line 38: A more realistic test here (as I'm pretty sure this would fail if you use `validated_save()` instead of `create()`) would be to define a GitRepository and do:

```python
ConfigContext(name="cc1", data={}).validated_save()
ConfigContext(name="cc1", data={}, owner=git_repo).validated_save()
```

Similar suggestion on the ConfigContextSchema and ExportTemplate tests.

nautobot/core/tests/test_commands.py
```diff
@@ -0,0 +1,66 @@
...
+            "pre_migrate",
+            *args,
+            stdout=out,
+            stderr=err,
+        )
+
+        return (out.getvalue(), err.getvalue())
+
+    def test_success(self):
+        """Test that duplicate ConfigContext[Schema] objects result in a failure."""
```
glennmatthews, line 29:
```suggestion
        """Test that the command passes with most common data."""
```

nautobot/core/management/commands/pre_migrate.py
jathanism: It puts them in one big list instead of a list of lists.

glennmatthews (commented):
Needs documentation of some sort (possibly in nautobot/docs/administration/nautobot-server.md?) I think? Otherwise looks solid.

glennmatthews (approved):

PR #3968, opened 2023-06-22
Only strip domain.tld part from lldp_device

jlixfeld:
Any device hostname returned by LLDP that contains a subdomain will appear in the LLDP Neighbors tab as a mismatch.  This is due to `lldp_device` being parsed with an assumption that the hostname will never contain a subdomain part and only ever contain a domain name part.

IEEE 802.1AB does not specify exactly what the System Name TLV should contain, nor does it specify one way or the other whether the how the domain name should be treated.

This PR does not account for cases where a second level domain is used, so that would still be a corner case which would get mis-handled.  On the other hand, it is being completely stripped previously currently, so in either case, if a device hostname with a second level domain name also contained a subdomain, it would report incorrectly in the `LLDP Neighbors` tab.

An alternative approach could be to not strip anything, and leave it up to the LLDP implementation and/or knobs on the device itself to correctly format the TLV and leave Nautobot out of that decision process.

# What's Changed

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/3968.changed
+++ b/changes/3968.changed
@@ -1,0 +1,1 @@
+Changed lldp_device parser to only strip domain and tld parts from neighbor_host allowing a device subdomain to be correctly matched.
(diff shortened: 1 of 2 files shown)


jlixfeld:
Proposed changes broke output entirely because I have no idea how to write javascript :)

Issue #3666, opened 2023-04-28
Database dump with `--natural-primary` is problematic for Tag and other GenericForeignKeys


glennmatthews:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0alpha
* Python version: any
* Database platform, version: any
* Middleware(s): n/a

### Steps to Reproduce
1. Have a database populated with various Tags applied to various objects
2. Create a database dump using `nautobot-server dumpdata --natural-primary --natural-foreign ...` (as recommended in [several](https://docs.nautobot.com/projects/core/en/stable/administration/nautobot-server/#dumpdata) [places](https://docs.nautobot.com/projects/core/en/stable/installation/migrating-from-postgresql/#export-data-from-postgresql) in our documentation at present)
3. Import this database dump into a fresh instance using `nautobot-server loaddata ...`
4. Check whether Tags are correctly associated to objects

### Expected Behavior

Tags to be correctly applied in the imported instance.

### Observed Behavior

For any model that supports natural keys (which is most of them after #3423), `dumpdata --natural-primary` will omit the `pk` UUID from the database dump for such models, resulting in the imported models receiving new, random UUID `pk` values. When using `--natural-foreign` as well (as recommended in our docs) this is OK for foreign-keys as the foreign-key dumps will also use the natural key rather than a UUID to identify the target object. Unfortunately, GenericForeignKey (as used in `TaggedItem` along with various other code points) does **not** support natural keys out of the box, and so any exported `TaggedItem` instances (and others) will still reference the UUID of the original object, which is not preserved in the exported data and hence not in the imported database.

The net effect is that the imported data will contain effectively "garbage" `TaggedItem` records (and other GenericForeignKey field values as well), which is clearly not desirable.

See also #2374, which has a similar root cause.

One possible fix may be to use a custom serializer/deserializer for the `dumpdata` and `loaddata` commands, such as described in https://stackoverflow.com/questions/11159377/is-it-possible-to-use-a-natural-key-for-a-genericforeignkey-in-django

gsnider2195:
related to https://github.com/jazzband/django-taggit/issues/708

glennmatthews:
ObjectChange records are another model that make extensive use of GFKs and will be affected by this issue.

glennmatthews:
[...]

PR #3984, opened 2023-06-23
Fix 2374 - update guidance on use of dumpdata/loaddata/trace_paths

### LINKED ISSUE #2374: Cable paths likely to become corrupted after dumpdata/loaddata, such as postgres/mysql migration
### Environment
* Nautobot version (Docker tag too if applicable): 1.4.2
* Python version: any
* Database platform, version: postgres, mysql
* Middleware(s): n/a

### Steps to Reproduce
1. Have a Postgres database with connected cables and traced cable paths
2. Export data via `nautobot-server dumpdata ...` [as described in our docs](https://nautobot.readthedocs.io/en/stable/installation/migrating-from-postgresql/#export-data-from-postgresql)
3. Import exported data into MySQL as described in the docs
4. View the interfaces of a device that has cables with traces

### Expected Behavior

Cable traces to be available and correct

### Observed Behavior

The obvious symptom is a KeyError being thrown from `CablePath.get_path()`, however that specific error is going to be tracked and fixed as a separate GitHub issue (adding error handling in `get_path()` to address the case where a cable path is corrupt or invalid. I'm filing this issue to focus on the *cause* of the path corruption, which is a result of two details interacting:

1. In our dumpdata/loaddata, we do not include `contenttypes.ContentType`, expecting that Django will handle creation of contenttypes records in the target system instead. As a result, content-type PKs are *not* guaranteed to be the same between the two databases
2. The `CablePath.path` field stores denormalized content-type PKs, using the style `["<content-type PK>:<CableTermination PK>", "..."]`, for example `['2:7b33b9c2-2f4b-4f2d-8732-befa4d90776e', '48:bfbf506a-d4cf-4f37-96a0-3f7e2cd38c00', '62:bb12a028-5a83-481e-971d-2a5fb6396fa9', '2:6aae157c-2d61-4c52-a5ab-2376851d7fbf']` Since these are strings, and not actual foreign-keys at the database level, they are not validated and are not converted to natural-keys (e.g. ["dcim", "interface"]) when the data is dumped.

Hence, post-import, the CablePath may contain invalid content-types and/or invalid combinations of content-type PK and instance UUID.

Offhand, the idea that comes to mind as a possible fix is to migrate things to use natural-keys for the content-type in the `path` database field, i.e. instead of `'2:7b33b9c2-2f4b-4f2d-8732-befa4d90776e'`, store the path node as `'dcim.interface:7b33b9c2-2f4b-4f2d-8732-befa4d90776e'`. (Once we have natural keys for the CableTermination models as well, we can take this the rest of the way as something like `dcim.interface:device-1,ethernet0`...)

glennmatthews:
# Closes: #2374 
# What's Changed

- Removed problematic guidance to use `--natural-primary` flag when running `nautobot-server dumpdata`. (see #3666)
- Added guidance to run `nautobot-server trace_paths` after `nautobot-server loaddata`. (fixes #2374).

# TODO

- n/a Unit, Integration Tests

### DIFF
--- a/nautobot/docs/administration/nautobot-server.md
+++ b/nautobot/docs/administration/nautobot-server.md
@@ -151,12 +151,14 @@ nautobot=> \q
     - `extras.job` should now be included in the dump (removed `--exclude extras.job` from the example usage)
     - `django_rq` should now be excluded from the dump (added `--exclude django_rq` to the example usage)
 
++/- 1.5.22
+    - We do not recommend at this time using `--natural-primary` as this can result in inconsistent or incorrect data for data models that use GenericForeignKeys, such as `Cable`, `Note`, `ObjectChange`, and `Tag`.
+
 ```no-highlight
 nautobot-server dumpdata \
   --natural-foreign \
-  --natural-primary \
-  --exclude contenttypes \
   --exclude auth.permission \
+  --exclude contenttypes \
   --exclude django_rq \
   --format json \
   --indent 2 \
@@ -165,7 +167,7 @@ nautobot-server dumpdata \
 
 Use this command to generate a JSON dump of the database contents.
 
-One example of using this command would be to [`export data from PostgreSQL`](../installation/migrating-from-postgresql.md#export-data-from-postgresql).
+One example of using this command would be to [export data from PostgreSQL](../installation/migrating-from-postgresql.md#export-data-from-postgresql) and then [import the data dump into MySQL](../installation/migrating-from-postgresql.md#import-the-database-dump-into-mysql).
 
 ### `fix_custom_fields`
 
@@ -306,8 +308,9 @@ There are a number of other options not covered here.
 
 To import the data that was exported with `nautobot-server dumpdata ...` see the following documentation:
 
-- [`Remove the auto-populated Status records from the database`](../installation/migrating-from-postgresql.md#remove-the-auto-populated-status-records-from-the-mysql-database)
-- [`Import the database dump`](../installation/migrating-from-postgresql.md#import-the-database-dump-into-mysql)
+- [Remove auto-populated records from the database](../installation/migrating-from-postgresql.md#remove-auto-populated-records-from-the-mysql-database)
+- [Import the database dump](../installation/migrating-from-postgresql.md#import-the-database-dump-into-mysql)
+- [Rebuild cached cable path traces](../installation/migrating-from-postgresql.md#rebuild-cached-cable-path-traces)
 
 ### `migrate`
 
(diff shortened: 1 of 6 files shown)


itdependsnetworks:
@mzbroch just want to call your intention to this.

bryanculver (approved):

nautobot/docs/administration/nautobot-server.md
```diff
@@ -151,12 +151,14 @@ nautobot=> \q
     - `extras.job` should now be included in the dump (removed `--exclude extras.job` from the example usage)
     - `django_rq` should now be excluded from the dump (added `--exclude django_rq` to the example usage)
 
++/- 1.5.22
+    - We do not recommend at this time using `--natural-primary` as this can result in inconsistent or incorrect data for data models that use GenericForeignKeys, such as `Cable`, `Note`, `ObjectChange`, and `Tag`.
+
 ```no-highlight
 nautobot-server dumpdata \
   --natural-foreign \
-  --natural-primary \
```
bryanculver, line 157: Are there any instances where exporting with `natural-foreign` but without `natural-primary` might result in import issues?

bryanculver: If we remove `natural-foreign` will we have to add `ContentType` to export too?

glennmatthews: If a plugin implements model tests inheriting from `nautobot.core.testing.models.ModelTestCases.BaseModelTestCase` then it does include testing of the natural and composite keys for the model. This is a relatively new addition though and not all models even in core have tests of this sort yet.

glennmatthews: > If we remove `natural-foreign` will we have to add `ContentType` to export too?

I think so, yes. Which will also mean we need to document explicitly deleting all existing ContentTypes before importing.

bryanculver: > > If we remove `natural-foreign` will we have to add `ContentType` to export too?
> 
> I think so, yes. Which will also mean we need to document explicitly deleting all existing ContentTypes before importing.

I think this would be preferred if we intend on encouraging (or at least just denoting that it is the primary use case) dump/loaddata being used only for wholesale migration and not merging datasets. We could call out that using `natural-foreign` can be used to make other migrations easier, but not 100% battle-tested.

If I'm not mistaken, the only risk for importing `ContentType` into an empty DB is if an app with model(s) is missing or new ones are added, but both should be cleaned up in `post_upgrade`.

jathanism (approved):

bryanculver (approved):

Issue #3782, opened 2023-05-19
2.0: CSV export doesn't provide a clearly-named CSV file.


glennmatthews:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0

### Steps to Reproduce
1. Export a data set to CSV from the legacy UI

### Expected Behavior

Clearly named file such as `nautobot_locations.csv`

### Observed Behavior

File has a `.csv` extension but the remainder of the filename is randomly generated by the browser, e.g. `EjU0iJSk.csv`

This is a consequence of the shift to a REST API-based solution for CSV import/export (#3722) but it's ugly and would be good to come up with a fix for this.

glennmatthews:
Fixed by #3993.

PR #3993, opened 2023-06-26
Add default filename for CSV export from the UI

glennmatthews:
# Closes: #3782 
# What's Changed

Add a `download=""` declaration to the CSV export link, so that when downloading a CSV file from the UI, you get a filename like `nautobot_location_data.csv` instead of something random like `EjU0iJSk.csv`.

# TODO

- n/a Unit, Integration Tests

### DIFF
--- a/changes/3782.fixed
+++ b/changes/3782.fixed
@@ -1,0 +1,1 @@
+Fixed filename for CSV exports.
--- a/nautobot/core/api/views.py
+++ b/nautobot/core/api/views.py
@@ -268,6 +268,13 @@ class ModelViewSetMixin:
             self.logger.warning(msg)
             return self.finalize_response(request, Response({"detail": msg}, status=409), *args, **kwargs)
 
+    def finalize_response(self, request, response, *args, **kwargs):
+        # In the case of certain errors, we might not even get to the point of setting request.accepted_media_type
+        if hasattr(request, "accepted_media_type") and "text/csv" in request.accepted_media_type:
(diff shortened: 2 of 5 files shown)


gsnider2195:
Would it make any difference if this was accomplished by setting the `Content-Disposition` header on the list view instead of using the html tag attribute? This worked on `ModelViewSet.list`:

```py
    def list(self, request, *args, **kwargs):
        response = super().list(request, *args, **kwargs)
        if isinstance(request.accepted_renderer, NautobotCSVRenderer):
            response["Content-Disposition"] = 'attachment; filename="somefilename.csv"'
        return response
```

glennmatthews:
That might work too, I hadn't considered it. What do you see as the pros/cons?

gsnider2195:
> That might work too, I hadn't considered it. What do you see as the pros/cons?

If we set the filename in the API response it would be easier to implement in the new ui

bryanculver (approved):
Based on @gsnider2195's comment above, should we also use the content-disposition method for SVG downloads at some point?

glennmatthews:
> Based on @gsnider2195's comment above, should we also use the content-disposition method for SVG downloads at some point?

*Maybe*? Need to make sure that doesn't break embedding the SVGs into the UI.

gsnider2195 (approved):

bryanculver:
> > Based on @gsnider2195's comment above, should we also use the content-disposition method for SVG downloads at some point?
> 
> _Maybe_? Need to make sure that doesn't break embedding the SVGs into the UI.

Not blocking or required for this PR but maybe a query param `?disposition=download` would suffice.

jathanism (approved):

Issue #1854, opened 2022-06-06
Tree model tables should sort sensibly or not be sortable at all


glennmatthews:
### Environment
* Python version:
* Nautobot version: 1.3.5

### Steps to Reproduce
1. Click on any hyperlinked column header in the Region, RackGroup, or TenantGroup tables to resort the table based on this column
    - For example, https://demo.nautobot.com/dcim/regions/?sort=-site_count

### Expected Behavior

Some sort of intelligent ordering of the tree, or else for this sorting to be disallowed

### Observed Behavior

The table is sorted by row based on the given column, but rows are still labeled based on their hierarchy within the tree, resulting in a fairly nonsensical display where rows appear to be "children" of the wrong parent rows:

![image](https://user-images.githubusercontent.com/5603551/172200690-02b03881-af5a-4c4b-9c1a-2e6a85a0e8f5.png)

PR #3995, opened 2023-06-26
Remove name nesting on MTPP Model tables when sorting

timizuoebideri1:
# Closes: #1854 
# What's Changed

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/1854.changed
+++ b/changes/1854.changed
@@ -1,0 +1,1 @@
+Make Tree Models list ui view table non sortable.
--- a/changes/1854.changed
+++ b/changes/1854.changed
@@ -1,1 +1,1 @@
-Make Tree Models list ui view table non sortable.
+Disabled sorting on tree model tables as it doesn't make sense to sort these.
--- a/changes/1854.changed
+++ b/changes/1854.changed
@@ -1,1 +1,1 @@
-Disabled sorting on tree model tables as it doesn't make sense to sort these.
+Disabled name nesting on MPTT model tables as it doesn't make sense to have nesting on these.
--- a/changes/1854.changed
+++ b/changes/1854.changed
@@ -1,1 +1,1 @@
-Disabled name nesting on MPTT model tables as it doesn't make sense to have nesting on these.
+When sorting tables for MPTT models, nesting/indentation of the model name display is disabled as it was misleading.
--- a/changes/1854.changed
+++ b/changes/1854.changed
@@ -1,1 +1,2 @@
 When sorting tables for MPTT models, nesting/indentation of the model name display is disabled as it was misleading.
+Disabled sorting on TreeNode model tables as TreeNode do not support sorting.
--- a/nautobot/utilities/tables.py
+++ b/nautobot/utilities/tables.py
@@ -33,6 +33,8 @@ class BaseTable(tables.Table):
         }
 
     def __init__(self, *args, user=None, **kwargs):
+        from nautobot.dcim.tables.template_code import MPTT_LINK_WITHOUT_NESTING  # Avoid Circular Import
+
         # Add custom field columns
         obj_type = ContentType.objects.get_for_model(self._meta.model)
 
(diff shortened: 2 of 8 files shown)


changes/1854.changed
```diff
@@ -0,0 +1 @@
+Make Tree Models list ui view table non sortable.
```
glennmatthews, line 1:
```suggestion
Disabled sorting on tree model tables as it doesn't make sense to sort these.
```

glennmatthews (approved):

itdependsnetworks:
Why doesn't it make sense? Having nesting when sorting doesn't make sense, but being able to sort makes sense to me.

glennmatthews:
I suppose that could be an alternate approach - when sorting, disable all nesting and render the sorted data entirely flat? Not sure that's what a user would necessarily expect though.

itdependsnetworks:
`when sorting, disable all nesting and render the sorted data entirely flat`, that is what I thought it was tbh.

nautobot/utilities/tables.py
```diff
@@ -58,6 +60,12 @@ def __init__(self, *args, user=None, **kwargs):
                 )
             # symmetric relationships are already handled above in the source_type case
 
+        if model := getattr(self.Meta, "model", None):
+            # Disable ordering on these models(Tree Models) as sorting on UI results in rows appears
+            # to be children of the wrong parent rows.
+            if issubclass(model, (MPTTModel, TreeNode)):
```
bryanculver, line 66: This would remove functionality for `Region` which works* today, as `MPTTModel` supports it.

This would make it more obvious for `Location` and other `TreeNode` based models that they do no support it, because under the hood you can't perform `order_by`: 
- https://github.com/matthiask/django-tree-queries/issues/23
- https://github.com/matthiask/django-tree-queries/issues/33


*Subjective to what any person would expect sorting to behave like here with the way we return nested results.

bryanculver:
> `when sorting, disable all nesting and render the sorted data entirely flat`, that is what I thought it was tbh.

Not how it works today: https://demo.nautobot.com/dcim/regions/?sort=-site_count

lampwins:
We will move forward with supporting sorting and when the column is sorted, the indentation that is rendered via the template will be stripped.

nautobot/utilities/tables.py
```diff
@@ -61,6 +65,11 @@ def __init__(self, *args, user=None, **kwargs):
         # Init table
         super().__init__(*args, **kwargs)
 
+        if model := getattr(self.Meta, "model", None):
+            # The name nesting is removed when sorting on the UI because it results in rows appearing as children of the wrong parent rows.
+            if issubclass(model, (MPTTModel, TreeNode)) and self.order_by:
```
timizuoebideri1, line 70:
```suggestion
            if issubclass(model, MPTTModel) and self.order_by:
```

nautobot/utilities/tables.py
```diff
@@ -10,6 +10,8 @@
 from django.utils.text import Truncator
 from django_tables2.data import TableQuerysetData
 from django_tables2.utils import Accessor
+from mptt.models import MPTTModel
+from tree_queries.models import TreeNode
```
timizuoebideri1, line 14:
```suggestion
```

nautobot/utilities/tables.py
```diff
@@ -61,6 +64,11 @@ def __init__(self, *args, user=None, **kwargs):
         # Init table
         super().__init__(*args, **kwargs)
 
+        if model := getattr(self.Meta, "model", None):
+            # The name nesting is removed when sorting on the UI because it results in rows appearing as children of the wrong parent rows.
+            if issubclass(model, MPTTModel) and self.order_by:
+                self.columns["name"].column.template_code = MPTT_LINK_WITHOUT_NESTING
+
```
glennmatthews, line 71: Don't we still need to add code to *disable* sorting for tables based on `TreeModel`? I thought we'd determined that sorting didn't work for those. Or is that already disabled case-by-case in the individual table definitions?

glennmatthews (approved):

Issue #3778, opened 2023-05-19
2.0: Swagger UI broken for POST /dcim/cables/, possibly others


glennmatthews:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Navigate to `https://next.demo.nautobot.com/api/docs/` (you must be using the legacy UI or the page fails to load at all; this is probably a separate bug)
2. Click to expand the `POST /dcim/cables/` subheading.

### Expected Behavior

Section to expand and render information about how to POST to this endpoint - this works in `demo.nautobot.com` (1.5.7) by comparison:

![image](https://github.com/nautobot/nautobot/assets/5603551/4635672e-42c2-421c-a564-28b19d4c7131)

### Observed Behavior

Loading spinner appears indefinitely, eventually the browser may complain about a long-running script and offer to stop it for you:

![image](https://github.com/nautobot/nautobot/assets/5603551/1d6090ba-fba9-4901-9de2-b1c2fa1775c1)

Maybe related to #3500?

gsnider2195:
I'm seeing similar schema sizes for demo and next:


```
https://demo.nautobot.com/api/swagger/?api_version=1.2

Allow: GET, HEAD, OPTIONS
Content-Disposition: inline; filename="API Documentation (1.2).json"
Content-Length: 12467165
Content-Type: application/json
Date: Tue, 20 Jun 2023 16:43:02 GMT
Referrer-Policy: same-origin
Vary: Accept, Cookie, Origin
X-Content-Type-Options: nosniff
X-Frame-Options: DENY
```

```
https://next.demo.nautobot.com/api/swagger/?api_version=2.0

Allow:
GET, HEAD, OPTIONS

Content-Disposition: inline; filename="API Documentation (2.0).json"
Content-Length: 10640435
Content-Type: application/json
Date: Tue, 20 Jun 2023 16:44:52 GMT
Referrer-Policy: same-origin
Server: Caddy
Server: WSGIServer/0.2 CPython/3.9.16
Vary: Accept, Cookie, Origin
X-Content-Type-Options: nosniff
X-Frame-Options: DENY
```

```bash
$ wc -l *.json
  292742 1.2.json
  251157 2.0.json

$ sort 1.2.json | uniq -c | sort -n | tail -20
   2472                                                     },
   2617                                 "format": "uuid"
   2952                                 "type": "string",
   3049                                 },
[...]

PR #4000, opened 2023-06-26
OpenAPI schema / Swagger UI fixes, remove unused REST API endpoints

glennmatthews:
# Closes: #3778 
# What's Changed

- Remove `app_full_width_fragment`, `detail-view-config`, `form-fields`, and `table-fields` REST API endpoints that were added during initial prototyping of the new UI but are currently unused.
- Set `ENUM_GENERATE_CHOICE_DESCRIPTION` to `False` in settings.py so that filter field descriptions in the OpenAPI schema do not unnecessarily/incorrectly list all possible enum values for each filter.
- Change the REST API `depth` used to render the OpenAPI schema from 1 to 0.
- The root cause of #3778 was that our `PolymorphicProxySerializer` fields in the schema were not respecting depth or generating nested serializers, so the schema in several cases was infinitely self-referential (a `CircuitTermination` has a `cable_peer` of type CableTermination; a `CableTermination` is polymorphic but might be a `CircuitTermination`, and repeat). I fixed this by replacing `get_serializers_for_models` (which is only used to define the schema for these fields) with `nested_serializers_for_models` which works similarly, but *only* generates nested serializers, avoiding the self-reference problem.
    - I'm not using `nested_serializer_factory` here because that requires a `RelationInfo` namedtuple as an input, which I don't have and can't easily construct.

Net result is that the Swagger UI once again loads and renders relatively quickly.

Polymorphic field on Cable, showing that nested serializers are used for this field in the schema:

image

All FKs also default to depth 0 now as well:

image

# TODO

- n/a Unit, Integration Tests


glennmatthews:
Looks like the CSV export is now falling back to depth 0 instead of depth 1, which won't work. Investigating.

nautobot/dcim/api/views.py
```diff
@@ -251,10 +251,6 @@ class RackReservationViewSet(NautobotModelViewSet):
     serializer_class = serializers.RackReservationSerializer
     filterset_class = filters.RackReservationFilterSet
 
-    # Assign user from request
-    def perform_create(self, serializer):
-        serializer.save(user=self.request.user)
-
```
glennmatthews, line 257: I don't know why tests were passing before, but this was definitely wrong and causing test failures now. We already handle the case in `RackReservationSerializer.to_internal_value()` where `user` is omitted and we want to use `request.user`, but this was ensuring that on create we would *always* use `request.user` even if a specific other `user` was provided.

glennmatthews: This might be because of the change in default serializer depth to 0; `test_recreate_object_csv()` directly instantiates serializers and I think at depth 1 any nested serializers get flagged as read-only by DRF and so were skipped over in the data validation.

jathanism (approved):

Issue #4005, opened 2023-06-27
2.0: Registering new Job from plugin cannot be done from jobs.py


jakubkrysl:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0-beta.1
* Python version: 3.10.11
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Create a new Job in plugin with `read_only = True`
2. make sure the Job is not in DB yet (its 1st registration)
3. use `nautobot.core.celery.register_jobs` to register it from `jobs.py`

### Expected Behavior
Job registered and working

### Observed Behavior
Job does not exist

```Running migrations:
  No migrations to apply.
[2023-06-27 20:00:39,562] DEBUG nautobot.core.celery: Importing system Jobs
[2023-06-27 20:00:39,639] INFO nautobot.extras.utils: Refreshed Job "System Jobs: Git Repository: Sync" from <GitRepositorySync>
[2023-06-27 20:00:39,656] INFO nautobot.extras.utils: Refreshed Job "System Jobs: Git Repository: Dry-Run" from <GitRepositoryDryRun>
[2023-06-27 20:00:39,671] INFO nautobot.extras.utils: Refreshed Job "User: Export owner changes" from <ExportOwnerChanges>
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/django/db/models/query.py", line 581, in get_or_create
    return self.get(**kwargs), False
  File "/usr/local/lib/python3.10/site-packages/django/db/models/query.py", line 435, in get
    raise self.model.DoesNotExist(
__fake__.Job.DoesNotExist: Job matching query does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/nautobot-server", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/site-packages/nautobot/core/cli/__init__.py", line 54, in main
    run_app(
  File "/usr/local/lib/python3.10/site-packages/nautobot/core/runner/runner.py", line 263, in run_app
    management.execute_from_command_line([runner_name, command] + command_args)
  File "/usr/local/lib/python3.10/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.10/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
[...]

PR #4009, opened 2023-06-28
Harden Job class API a bit and gracefully handle errors when registering Jobs to the DB

glennmatthews:
# Closes: #4005
# What's Changed

- Correct an admonition in the jobs documentation that instructed app authors to call `register_jobs()` from `__init__.py`; it should be called from `jobs.py`.
- Added `register_jobs()` call to documentation about developing jobs in an app.
- Added `@final` decorator to Job classproperties (https://peps.python.org/pep-0591/); this doesn't *prevent* subclasses from overriding these methods but at least serves as documentation that doing so is discouraged.
- Add type hints to Job classproperties
- Add type ~coercion~ enforcement to Job classproperties (e.g., force `read_only` property to return a `bool` or else raise a TypeError, rather than potentially return a non-bool value)
- In `refresh_job_model_from_job_class()`, add `transaction.atomic` wrapper around the creation and updating of each Job DB record, and catch and log any errors encountered in doing so, so that a single invalid Job doesn't cause the entire function to abort.

I tested this manually by adding a Job to the example plugin that defines invalid values for some or all of its `Meta` properties, i.e.:

```python
class ExampleBuggyJob(Job):
    class Meta:
        name = "Example job with invalid parameters"
        dryrun_default = "no"
        hidden = ""
        has_sensitive_variables = "who knows?"
        ...

    def run(self):
        pass
```

and verifying that an unhandled `ValidationError` is no longer encountered but instead an appropriate error message is logged and the invalid job is not registered to the database. Probably wouldn't hurt to automate this but I'm not sure the value is there - let me know what you think.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/core/celery/__init__.py
+++ b/nautobot/core/celery/__init__.py
@@ -217,5 +217,6 @@ nautobot_task = shared_task
 def register_jobs(*jobs):
     """Helper method to register jobs with Celery"""
     for job in jobs:
+        # TODO: should we only register a job if it corresponds to a Job database record?
         logger.debug("Registering job %s.%s", job.__module__, job.__name__)
         app.register_task(job)
--- a/nautobot/docs/additional-features/jobs.md
+++ b/nautobot/docs/additional-features/jobs.md
@@ -35,7 +35,7 @@ In any case, each module holds one or more Jobs (Python classes), each of which 
 For example, we can create a module named `devices.py` to hold all of our jobs which pertain to devices in Nautobot. Within that module, we might define several jobs. Each job is defined as a Python class inheriting from `extras.jobs.Job`, which provides the base functionality needed to accept user input and log activity.
 
 +/- 2.0.0
-    All job classes must now be registered with `nautobot.core.celery.register_jobs` on module import. For plugins providing jobs, the module containing the jobs must be imported by the plugin's `__init__.py` and the `register_jobs` method called at import time. The `register_jobs` method accepts one or more job classes as arguments.
+    All job classes must now be registered with `nautobot.core.celery.register_jobs` on module import. For plugins providing jobs, the `register_jobs` method must called from the plugin's `jobs.py` file/submodule at import time. The `register_jobs` method accepts one or more job classes as arguments.
 
 !!! warning
     Make sure you are *not* inheriting `extras.jobs.models.Job` instead, otherwise Django will think you want to define a new database model.
--- a/nautobot/extras/jobs.py
+++ b/nautobot/extras/jobs.py
@@ -422,14 +422,22 @@ class BaseJob(Task):
         return getattr(module, "name", module.__name__)
 
     @final
+    @classmethod
+    def _get_meta_attr_and_assert_type(cls, attr_name, default, expected_type):
+        result = getattr(cls.Meta, attr_name, default)
+        if not isinstance(result, expected_type):
+            raise TypeError(f"Meta.{attr_name} should be {expected_type}, not {type(result)}")
+        return result
+
+    @final
     @classproperty
     def name(cls) -> str:  # pylint: disable=no-self-argument
-        return getattr(cls.Meta, "name", cls.__name__)
+        return cls._get_meta_attr_and_assert_type("name", cls.__name__, expected_type=str)
 
     @final
     @classproperty
     def description(cls) -> str:  # pylint: disable=no-self-argument
-        return dedent(getattr(cls.Meta, "description", "")).strip()
+        return dedent(cls._get_meta_attr_and_assert_type("description", "", expected_type=str)).strip()
 
     @final
     @classproperty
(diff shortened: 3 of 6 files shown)


nautobot/extras/utils.py
```diff
@@ -409,42 +410,53 @@ def refresh_job_model_from_job_class(job_model_class, job_class):
...
+                    # It was inherited and not overridden
+                    setattr(job_model, field_name, getattr(job_class, field_name))
 
-    for field_name in JOB_OVERRIDABLE_FIELDS:
-        # Was this field directly inherited from the job before, or was it overridden in the database?
-        if not getattr(job_model, f"{field_name}_override", False):
-            # It was inherited and not overridden
-            setattr(job_model, field_name, getattr(job_class, field_name))
+            # set supports_dryrun to match the property on the job class
+            job_model.supports_dryrun = job_class.supports_dryrun
```
glennmatthews, line 444: remove, redudant with `if not created` code below.

nautobot/extras/utils.py
```diff
@@ -409,42 +410,53 @@ def refresh_job_model_from_job_class(job_model_class, job_class):
...
+                    "is_job_button_receiver": issubclass(job_class, JobButtonReceiver),
+                    "read_only": job_class.read_only,
+                    "supports_dryrun": job_class.supports_dryrun,
+                    "installed": True,
+                    "enabled": False,
+                },
+            )
+
+            if job_name != default_job_name:
+                job_model.name_override = True
```
glennmatthews, line 431: should also probably set `grouping_override` if `job_class.grouping` was truncated above?

glennmatthews: on reflection perhaps not - we don't have the same concept of renumbering a grouping in case of conflict.

glennmatthews:
Feedback from the submitter of #4005 

> I can confirm your fix allows me to load the Job without issues. Though it hides the error completely now and if you set the value to something not `True|False` but for example `(False,)` or `"False"` it gets changed to `True` because `bool((False,)) == True` 🙂 

Maybe we should do type checking instead of type coercion in the classproperties?

nautobot/extras/jobs.py
```diff
@@ -394,83 +397,108 @@ def class_path(cls):  # pylint: disable=no-self-argument
...
+    @final
     @classproperty
-    def grouping(cls):  # pylint: disable=no-self-argument
+    def grouping(cls) -> str:  # pylint: disable=no-self-argument
         module = inspect.getmodule(cls)
         return getattr(module, "name", module.__name__)
 
+    @final
+    @classmethod
+    def _get_meta_attr_and_assert_type(cls, attr_name, default, expected_type):
```
itdependsnetworks, line 426: Perhaps this should be as a mixin or decorator, so if perform type enforcement later it is centralized, well defined, and more easily replaceable?

glennmatthews: This is solving a fairly narrow problem, that of user-defined code in Jobs not necessarily being conformant to the API for Jobs. I don't think we have any plans now or in the near future to add more generalized type enforcement throughout the code base.

nautobot/core/celery/__init__.py
```diff
@@ -217,5 +217,6 @@ def _loads(obj):
 def register_jobs(*jobs):
     """Helper method to register jobs with Celery"""
     for job in jobs:
+        # TODO: should we only register a job if it corresponds to a Job database record?
```
gsnider2195, line 220: Don't we use the celery registry to create the database entries? We don't want to go back to trying to find jobs manually, right?

glennmatthews: Yeah, it might get circular if we try to do this. I'm just not a big fan of the current behavior where we register it into Celery even if it failed validation when we tried to create the database entry.

gsnider2195: We could prune the celery registry based on the database models but it would get complicated on the worker(s) since they don't know when the database is ready

nautobot/docs/additional-features/jobs.md
```diff
@@ -35,7 +35,7 @@ In any case, each module holds one or more Jobs (Python classes), each of which
 For example, we can create a module named `devices.py` to hold all of our jobs which pertain to devices in Nautobot. Within that module, we might define several jobs. Each job is defined as a Python class inheriting from `extras.jobs.Job`, which provides the base functionality needed to accept user input and log activity.
 
 +/- 2.0.0
-    All job classes must now be registered with `nautobot.core.celery.register_jobs` on module import. For plugins providing jobs, the module containing the jobs must be imported by the plugin's `__init__.py` and the `register_jobs` method called at import time. The `register_jobs` method accepts one or more job classes as arguments.
```
gsnider2195, line 38: Is the note about loading the job module in the app's `__init__.py` unnecessary because the job module always gets loaded by [the plugin ready callback](https://github.com/nautobot/nautobot/blob/8746766e141cd5d5c84d260f5d629d331c12da06/nautobot/extras/plugins/__init__.py#L144C2-L144C2)?

Issue #3973, opened 2023-06-22
Refine natural keys for VLANGroup and Prefix


glennmatthews:
- [x] Remove overridden natural-key for VLANGroup, allow it to just be `name`
- [x] Investigate feasibility of using `prefix` instead of `network, prefix_length` as the composite-key for Prefix

glennmatthews:
Done by #4027.

PR #4027, opened 2023-06-29
Update IPAM natural keys

glennmatthews:
# Closes: #3973 
# What's Changed

- Changed natural-key for `Prefix` model to `[namespace, prefix]` instead of `[namespace, network, prefix_length]`.
   - This required a few code changes in BaseModel to ensure that the netaddr.Network object is coerced to a string and to handle natural-key components that are a property rather than a field.
   - This required a change in PrefixQuerySet to ensure that `composite_key` lookup deconstructs the composite key before deconstructing the `prefix` parameter
- Changed natural-key for `Service` model to `[name, virtual_machine, device]` rather than `[name, device, virtual_machine]` since `device` natural-key is variable-length.
- Changed natural-key for `VLANGroup` model to simply `[name]` rather than `[name, location]` since `name` is now globally unique.
- Changed natural-key for `VLAN` model to `[pk]` for now because we don't currently have a true uniqueness constraint on `VLAN`.
- Added at least minimal model test cases for these models where they were absent.

# TODO
- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/ipam/tests/test_models.py
+++ b/nautobot/ipam/tests/test_models.py
@@ -11,12 +11,20 @@ from nautobot.core.testing.models import ModelTestCases
 from nautobot.dcim import choices as dcim_choices
 from nautobot.dcim.models import Device, DeviceType, Interface, Location, LocationType
 from nautobot.extras.models import Role, Status
-from nautobot.ipam.choices import IPAddressTypeChoices, PrefixTypeChoices
-from nautobot.ipam.models import IPAddress, IPAddressToInterface, Namespace, Prefix, VLAN, VLANGroup
+from nautobot.ipam.choices import IPAddressTypeChoices, PrefixTypeChoices, ServiceProtocolChoices
+from nautobot.ipam.models import (
+    IPAddress,
+    IPAddressToInterface,
+    Namespace,
+    Prefix,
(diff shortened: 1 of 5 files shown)


nautobot/ipam/tests/test_models.py
```diff
@@ -990,3 +1030,8 @@ def test_vlan_validation(self):
             f'The assigned group belongs to a location that does not include location "{location.name}"',
             str(cm.exception),
         )
+
+
+class TestVRF(ModelTestCases.BaseModelTestCase):
+    model = VRF
+    # TODO(jathan): Add VRF model tests.
```
jathanism, line 1037: Is this comment still relevant with the addition of this test case?

jathanism (approved):

nautobot/ipam/tests/test_models.py
glennmatthews: You tell me :-) All the base class adds currently are some tests around natural-key and composite-key; I'm guessing that you had something beyond that in mind?

jathanism: I suppose we might want to exercise the new stuff, but I also this base test class didn't exist before...

Issue #4004, opened 2023-06-27
IPAM 2.0: Revisit prefix/ipaddress type constraints


gsnider2195:
### Proposed Changes

Migrations need to be updated to account for the prefix/ip address type constraints added in #3935:

- `nautobot/ipam/migrations/0031_ipam___data_migrations.py`

### Justification

Follow-up work from #3935

glennmatthews:
implemented by #4036.

PR #4036, opened 2023-06-30
Change prefix type enforcement to advisory for 2.0

glennmatthews:
# Closes: #4004
# What's Changed

- Comment out a bunch of code from #3873 and #3942 as `# 3.0 TODO`, net effects are:
   - IPAddresses are once again for now allowed to parent directly to Pool prefixes, instead of to the containing prefix
   - A Prefix can again be deleted even if this would cause its child IPs to be parented to a prefix other than type Network
   - A Prefix can again be deleted even if this would cause its child prefixes to be parented to a prefix of the "wrong" type
   - A Prefix can again be created or edited even if its parent or children would be the "wrong" type
   - An IPAddress can again be created or edited under a Container prefix.
- Update docs to reflect that all of the above cases are recommended against and will be enforced invalid in a future release.
- Add a callback API `successful_post()` to ObjectEditView so that I can use this to add post-POST logic to `IPAddressEditView` and `PrefixEditView` without needing to copy-and-paste the entire `post()` function from the generic view.
- Add missing `parent` filters to `PrefixFilterSet` and `IPAddressFilterSet` because I needed to use them in remediation guidance.
   - `IPAddressFilterSet` had a legacy string-based `parent` filter, which I've kept but renamed to `prefix`.
- In `PrefixEditView` and `IPAddressEditView`, add `successful_post()` implementations that check for all of the above recommended-against cases when saving a Prefix or an IPAddress respectively, and use the `messages` API to post a banner to the user advising them of possible ways to correct their data for forwards-compatibility.

image

image

image

etc.

- Add test cases for these warnings.

# TODO

- [x] Unit, Integration Tests


glennmatthews:
It just occurred to me - would it be desirable for the Prefix and IPAddress `save()` and `delete()` to uncomment and perform the various checks, but log warning messages to the syslog for now instead of raising ValidationErrors? Or is it better to just disable them entirely as I've done so far?

bryanculver:
> It just occurred to me - would it be desirable for the Prefix and IPAddress `save()` and `delete()` to uncomment and perform the various checks, but log warning messages to the syslog for now instead of raising ValidationErrors? Or is it better to just disable them entirely as I've done so far?

For 2.0 it would be fine to keep this as-is, since we don't support editing in the new UI at the moment. I think we need to consider some way to do warnings in the API to support this in the new UI. Warnings in the ORM/nbshell would be ideal at some point too.

timizuoebideri1 (approved):

nautobot/docs/installation/upgrading-from-nautobot-v1.md
```diff
@@ -138,13 +140,15 @@ The `ipam.IPAddress` model has been modified to have a foreign key to `ipam.Pref
...
 
-#### IPAddress Parenting Constraints
+#### IPAddress Parenting Guidance
+
+The following guidance has been added to the `IPAddress` model:
 
-The following constraints have been added to the `IPAddress` model:
+- An `IPAddress` should have a parent `Prefix` of type `Network`
+- An `IPAddress` should not be created if a suitable parent `Prefix` of type `Network` does not exist
+- An `IPAddress` can be a member of a `Pool` but only if the `Pool` is a child of a `Network`.
```
gsnider2195, line 149:
```suggestion
- An `IPAddress` can be a member of a `Pool` but only if the `Pool` is a child of a `Network`
```
For consistency

nautobot/docs/models/ipam/ipaddress.md
```diff
@@ -47,8 +47,11 @@ An IP address can be designated as the network address translation (NAT) inside
...
 
-The following constraints are enforced on the `IPAddress.parent` field:
+The following guidance has been added for the `IPAddress.parent` field:
 
-* An `IPAddress` must have a parent `Prefix` of type `Network`
-* An `IPAddress` cannot be created if a suitable parent `Prefix` of type `Network` does not exist
-* An `IPAddress` can be a member of a `Pool` but only if the `Pool` is a child of a `Network` prefix. This is because the `IPAddress` must have a concrete relationship to a `Network` and the `Pool` membership is derived from the IP address being within the `Pool`'s range, not from the IP's `parent` field.
+* An `IPAddress` should have a parent `Prefix` of type `Network`
+* An `IPAddress` should not be created if a suitable parent `Prefix` of type `Network` does not exist
+* An `IPAddress` can be a member of a `Pool` but only if the `Pool` is a child of a `Network` prefix.
```
gsnider2195, line 54:
```suggestion
* An `IPAddress` can be a member of a `Pool` but only if the `Pool` is a child of a `Network` prefix
```

nautobot/ipam/tests/test_views.py
```diff
@@ -294,6 +424,41 @@ def test_move_ip_addresses_between_namespaces(self):
...
+
+        with self.subTest("Creating an IP as a child of a same-size Container prefix raises a different warning"):
+            self.form_data["address"] = "192.0.2.2/25"
+            response = self.client.post(self._get_url("add"), data=post_data(self.form_data), follow=True)
+            self.assertHttpStatus(response, 200)
+            content = extract_page_body(response.content.decode(response.charset))
+            self.assertIn(
+                "IP address 192.0.2.2/25 currently has prefix 192.0.2.0/25 as its parent, which is a Container. "
+                "This will be considered invalid data in a future release. "
+                "Consider changing the prefix to type Pool to resolve this issue.",
```
gsnider2195, line 458: Why only suggest changing the container to a pool? Would a network not be a suitable type for this situation?

glennmatthews: See https://github.com/nautobot/nautobot/pull/4036/files#diff-caf8a412a9f30295cacfae888c0fbc1bdc2a7cd2643df55f444414f363113a7bR730-R733 - my reasoning was that the parent of the Container prefix is itself a Network, so changing the Container to a Network would still be invalid data unless you also change the parent Network to a Container. Maybe that's too fiddly though and should be simplified - thoughts?

nautobot/ipam/views.py
```diff
@@ -596,6 +695,54 @@ class IPAddressEditView(generic.ObjectEditView):
...
+                messages.warning(
+                    request,
+                    mark_safe(
+                        f"IP address {obj} currently has prefix {parent_link} as its parent, which is a Container. "
+                        f'{warning_msg} Consider <a href="{create_url}">creating an intermediate /{obj.mask_length} '
+                        "prefix of type Network</a> to resolve this issue."
+                    ),
+                )
+            else:
+                if obj.parent.parent.type == choices.PrefixTypeChoices.TYPE_NETWORK:
```
glennmatthews, line 730: Need to handle `obj.parent.parent is None` possibility here.

nautobot/ipam/tests/test_views.py
gsnider2195: I must have missed that `successful_post` logic. I think it might be better to either change the message to "Network or Pool" on the IPAddress object save or explain why you're suggesting `Pool` or `Network`. Otherwise, the message changes based on conditions that aren't obvious when saving the IPAddress and I think that could be confusing.

gsnider2195 (approved):

HanlinMiao (approved):

Issue #4021, opened 2023-06-28
Incorrect warning banner when MAX_PAGE_SIZE is set as 0


glennmatthews:
### Environment
* Nautobot version (Docker tag too if applicable): b7ec7c397f2b0516a2757dfa5bd543cf1caaeece

### Steps to Reproduce
1. Set `settings.MAX_PAGE_SIZE` to `0` (or `None`) to allow a user to specify arbitrarily large page sizes
2. Visit any list view

### Expected Behavior

No banner

### Observed Behavior

![image](https://github.com/nautobot/nautobot/assets/5603551/7ccb4895-7583-4480-8182-dbf1b60862ae)

Looks like a missed special case in #3988

PR #4046, opened 2023-07-05
The MAX_PAGE_SIZE_EXCEEDED notice is no longer displayed. When MAX_PAGE_SIZE is set to zero.

timizuoebideri1:
# Closes: #4021 
# What's Changed

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/4021.fixed
+++ b/changes/4021.fixed
@@ -1,0 +1,1 @@
+The MAX_PAGE_SIZE_EXCEEDED notice is no longer displayed. When MAX_PAGE_SIZE is set to zero.
--- a/nautobot/core/views/generic.py
+++ b/nautobot/core/views/generic.py
@@ -326,7 +326,7 @@ class ObjectListView(ObjectPermissionRequiredMixin, View):
             RequestConfig(request, paginate).configure(table)
             table_config_form = TableConfigForm(table=table)
             max_page_size = get_settings_or_config("MAX_PAGE_SIZE")
-            if paginate["per_page"] > max_page_size:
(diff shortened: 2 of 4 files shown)


changes/4021.fixed
```diff
@@ -0,0 +1 @@
+The MAX_PAGE_SIZE_EXCEEDED notice is no longer displayed. When MAX_PAGE_SIZE is set to zero.
```
glennmatthews, line 1:
```suggestion
Fixed erroneous warning banner on list views when `MAX_PAGE_SIZE` is set to zero.
```

glennmatthews:
Looks like the docs build is failing because there's been a new version of `griffe` package (which we don't currently pin in docs/requirements.txt) released that breaks compability with the old version of `mkdocstrings-python` that we currently pin. Probably the correct fix is to update `mkdocstrings-python` to `1.1.2` if possible and add a pin for `griffe` as well. I can tackle that in a separate PR if you'd prefer.

timizuoebideri1:
> Looks like the docs build is failing because there's been a new version of `griffe` package (which we don't currently pin in docs/requirements.txt) released that breaks compability with the old version of `mkdocstrings-python` that we currently pin. Probably the correct fix is to update `mkdocstrings-python` to `1.1.2` if possible and add a pin for `griffe` as well. I can tackle that in a separate PR if you'd prefer.

 👍 That would be great, Thanks

glennmatthews (approved):

gsnider2195 (approved):

PR #4049, opened 2023-07-05
Update mkdocstrings and mkdocstrings-python, pin griffe, restructure poetry dependencies

glennmatthews:
# Closes: #N/A

## Problems:

- ReadTheDocs build was broken the last couple of days because we don't pin our `griffe` dependency and a new version of griffe was released that breaks compatibility with the older versions of `mkdocstrings` and `mkdocstrings-python` we were pinned to.
- `[tool.poetry.dev-dependencies]` in pyproject.toml has been deprecated by Poetry since version 1.2.0 in favor of `[tool.poetry.group.dev.dependencies]`, and newer versions of Poetry don't add things to the deprecated section any more, requiring manual cleanup of pyproject.toml.

## Solutions

- Put in the time to restructure pyproject.toml to match Poetry's newer expectations. I've split our "dev" dependencies into the sub-groups of "dev", "docs", "linting", and "testing", all of which are still installed by default by Poetry but could in future be installed separately or together as best suits needs.
- Pin `griffe` docs dependency in both `pyproject.toml` and `docs/requirements.txt` to the latest version that still supports Python 3.7. (The new version just released, 0.31, drops support for 3.7, so we can't use it just yet)
- Update `mkdocstrings` and `mkdocstrings-python` docs dependencies in both `pyproject.toml` and `docs/requirements.txt` to latest versions for compatibility with this version of `griffe`.

# TODO

- n/a Unit, Integration Tests


pyproject.toml
```diff
@@ -154,55 +154,65 @@ remote_storage = ["django-storages"]
...
 # Tool for debugging Django
 django-debug-toolbar = "~3.7.0"
-# DiscoverSlowestTestsRunner for running CI performance tests and benchmarking.
-django-slowtests = "^1.1.1"
 # Nautobot example plugin used for testing
 example-plugin = {path = "examples/example_plugin", develop = true}
 # Nautobot test plugin used for testing core view overrides
 example-plugin-with-view-override = { path = "examples/example_plugin_with_view_override", develop = true }
 # Random data generation
 factory-boy = "~3.2.1"
```
HanlinMiao, line 165: Should `factory-boy` be under testing?

Issue #3646, opened 2023-04-25
Restrict access to what a user sees when not logged into Nautobot


jedelman8:
```[tasklist]
### TODO
- [ ] When `HIDE_RESTRICTED_UI` is enabled and a user is not logged in, all routes lead to the login page
- [ ] When `HIDE_RESTRICTED_UI` is enabled and a user is not logged in, the footer only contains the current time (remove hostname, links)
- [ ] When `HIDE_RESTRICTED_UI` is enabled and a user is not logged in, navigating to the browsable API, Swagger, and GraphQL pages redirects to the login page
- [ ] When `HIDE_RESTRICTED_UI` is enabled and a user is not logged in, banners (except for LOGIN_BANNER) are not displayed
```
---
### As ...

Patti - Platform Admin

### I want ...

to restrict what a user can see when not logged into Nautobot to improve the security of the platform. 

### So that ...

I am able to confidently expose Nautobot in a large organization or even on the Internet without users knowing more details than they need to know (even giving the option to the Nautobot identity when not logged in).

### I know this is done when...

- [ ] There is an option remove the footer and/or individually (a) the "hostname" that is displayed in the bottom left (given "nautobot" may be used in the name quite often) and (b) Not show Theme, Docs, API, GraphQL, Code, and Help.  This removes some of the "Nautobot identity" when not logged in. 
- [ ] Ensure API docs are now browsable when not logged in.  For one, it gives the identity of the application, but it also exposes which plugins/Apps are installed without logging in. 
- [ ] Ensure GraphiQL is not available for users not logged in.  I just tested in the demo.nautobot.com sandbox and I tried a valid devices query when not logged in and I got a valid response with an empty list as a result.  It _seems_ like I got a 200 and I'd _at least_ expect a 400. Regardless, this is proposing not viewing GraphiQL when not logged. 
- [ ] Do not display top and bottom banners when not logged in (not the login banner)
- [ ] ~Remove the Nautobot logo (in the top left) when not logged in.~ (willing to discuss, but not doing for now)
- [ ] Remove the search bar in top right.
- [ ] ~I have the option to ONLY see a Login panel with username and password and the timestamp in the footer (at best).~ (HIDE_RESTRICTED_UI covers this for as far as we are going to go with it)

[...]

PR #4068, opened 2023-07-07
Redirect unauthenticated users to login if HIDE_RESTRICTED_UI

timizuoebideri1:
# Closes: #3646 
# What's Changed

Screenshot 2023-07-07 at 8 42 22 AM

# TODO

- [ ] Unit, Integration Tests


nautobot/core/middleware.py
```diff
@@ -16,10 +17,24 @@
...
     ldap_auth_enabled,
 )
+from nautobot.core.utils.config import get_settings_or_config
 from nautobot.core.views import server_error
 from nautobot.extras.context_managers import change_logging, WebChangeContext
 
 
+class LoginRedirectMiddleware(MiddlewareMixin):
+    def __call__(self, request):
+        login_required_paths = ["/graphql/", "/api/docs/"]
```
timizuoebideri1, line 27: Any suggestions on how I can tell if '/graphql/' is being called from the browser or the API, because the API ought not to redirect?

bryanculver:
@timizuoebideri1 We want this in 1.6. Mind cherry-picking off a branch of `develop-1.6`?

nautobot/core/views/__init__.py
```diff
@@ -221,6 +221,8 @@ def csrf_failure(request, reason="", template_name="403_csrf_failure.html"):
     return HttpResponseForbidden(t.render(context), content_type="text/html")
 
 
+# LoginRequiredMixin cannot be added here as this is both available to the UI and API
+# if added on UI a redirect tot login occurs which is good but on API it also redirects to Login page which is bad
```
timizuoebideri1, line 225: Also, redirecting the docs when not LoggedIn is not straight forward because Django does not really handle static files, and in production it may be managed by ngnix or something else.

nautobot/core/settings.py
```diff
@@ -833,3 +833,5 @@ def UI_RACK_VIEW_TRUNCATE_FUNCTION(device_display_name):
     :type: str
     """
     return str(device_display_name).split(".")[0]
+
+HIDE_RESTRICTED_UI = True
```
bryanculver, line 837: Was this from testing? It's already a Constance config setting which should be retreived with `settings_or_config`.

nautobot/core/views/__init__.py
bryanculver: What would you like to do here? Heuristic on the `render_graphiql` to check to see if the browser has `accept: html`?

nautobot/core/tests/test_views.py
```diff
@@ -255,6 +255,25 @@ def test_installed_plugins_disabled_to_non_staff_user_with_hide_restricted_ui_fa
 
 
 class LoginUI(TestCase):
+    def setUp(self):
+        super().setUp()
+
+        self.footer_links = """
```
bryanculver, line 261: This seems fragile. Instead can we just use the `assertContains` to search for the strings we'd like to avoid?

Also, "New UI" does not exist in 1.6.

nautobot/core/templates/inc/footer.html
```diff
@@ -6,29 +6,34 @@
...
-                        {% endif %}
-                        &middot;
-                        <i class="mdi mdi-cloud-braces text-primary"></i> <a href="{% url 'api_docs' %}">API</a> &middot;
-                        <i class="mdi mdi-graphql text-primary"></i> <a href="{% url 'graphql' %}">GraphQL</a> &middot;
-                        <i class="mdi mdi-xml text-primary"></i> <a href="{{ settings.BRANDING_URLS.code }}">Code</a> &middot;
-                        <i class="mdi mdi-lifebuoy text-primary"></i> <a href="{{ settings.BRANDING_URLS.help }}">Help</a>
-                    </p>
+                    {% if request.user.is_authenticated or not "HIDE_RESTRICTED_UI"|settings_or_config %}
+                        <p class="text-muted">
+                            <i class="mdi mdi-new-box text-primary"></i> <a href="#" onclick="viewNewUI();">View in New UI</a> &middot;
```
bryanculver, line 22: No new UI in 1.6

bryanculver (changes requested):

changes/3646.added
```diff
@@ -0,0 +1,2 @@
+Redirect unauthenticated users on either `/api/docs/` or `/graphql/` to login page if `HIDE_RESTRICTED_UI` is True.
```
glennmatthews, line 1: My read of the first list item in #3646 is that with HIDE_RESTRICTED_UI, *all* views should redirect to the login for unauthenticated users. Does that match your understanding?

changes/3646.added
glennmatthews, line 1: Both of these items read as `changed` rather  than `added` to me, FWIW.

nautobot/core/settings.py
glennmatthews: Yeah definitely should revert this.

nautobot/core/templates/inc/footer.html
```diff
@@ -6,29 +6,34 @@
...
             <div class="row">
                 <div class="col-xs-4">
-                    <p class="text-muted">
-                        {% if settings.BRANDING_FILEPATHS.logo and settings.BRANDING_POWERED_BY_URL %}<a href="{{ settings.BRANDING_POWERED_BY_URL }}"><img src="{% static 'img/nautobot_logo.svg' %}" height="20" /> Powered</a> &middot;{% endif %}
-                        {{ settings.HOSTNAME }} {% if request.user.is_authenticated %} (v{{ settings.VERSION }}) {% endif %}
-                    </p>
+                    {% if request.user.is_authenticated or not "HIDE_RESTRICTED_UI"|settings_or_config %}
+                        <p class="text-muted">
+                            {% if settings.BRANDING_FILEPATHS.logo and settings.BRANDING_POWERED_BY_URL %}<a href="{{ settings.BRANDING_POWERED_BY_URL }}"><img src="{% static 'img/nautobot_logo.svg' %}" height="20" /> Powered</a> &middot;{% endif %}
+                            {{ settings.HOSTNAME }} {% if request.user.is_authenticated %} (v{{ settings.VERSION }}) {% endif %}
```
glennmatthews, line 12: @bryanculver @lampwins I'd be real tempted for us to just flat-out remove the `settings.HOSTNAME` from the footer - its meaningless in a Docker deployment and I'm curious if anyone actually finds it ever useful. Thoughts?

nautobot/core/tests/test_views.py
glennmatthews: Agreed, let's use some more narrowly targeted `assertNotContains`/`assertNotIn` checks instead.

nautobot/core/tests/test_views.py
```diff
@@ -278,6 +297,31 @@ def test_sso_login_button_visible(self):
...
+        self.client.logout()
+        url = "/api/docs/"
+        response = self.client.get(url)
+        self.assertHttpStatus(response, 302)
+        self.assertEqual(response.url, f"/login/?next={url}")
+        response_content = response.content.decode(response.charset).replace("\n", "")
+        # Assert `self.footer_links` not found in `response_content`
+        # Since self.assertNotInHTML is not available, Hence the workaround
+        with self.assertRaises(AssertionError):
+            self.assertInHTML(self.footer_links, response_content)
```
glennmatthews, line 312: Not sure this is testing what you think here - since it's a 302 response, I don't think it has any content to check. Would need to do `follow=true` on the `get()` call, I think?

nautobot/core/views/__init__.py
glennmatthews: `render_graphiql` should be UI-specific, I can't think that it would be used in the API?

nautobot/core/templates/inc/footer.html
bryanculver: Probably better to use the banner header for instance validation but I can some nominal value with the Docker deployment model for multi-instance. Help pin down where your logs might be.

changes/3646.added
timizuoebideri1: Yr, but those has been implemented before

glennmatthews (approved):

glennmatthews (changes requested):
Sorry, just reviewed #3646 and it doesn't look like this currently addresses the item about suppressing banners when not logged in?

glennmatthews (approved):

bryanculver (approved):

PR #4071, opened 2023-07-07
Fix Pre-release Tag Format in GitHub Action

bryanculver:
# Closes: #DNE
# What's Changed
- Change Docker pre-release tag to use `match` format to deal with `+` not being a valid character in Docker tags
- Remove invalid raw format that caused previous pre-release workflow to fail

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -85,9 +85,8 @@ jobs:
           flavor: |
             latest=false
           tags: |
-            type=semver,pattern={{version}}+py${{ matrix.python-version }}
+            type=match,pattern=v(.*),group=1,suffix=-py${{ matrix.python-version }}
             type=semver,pattern={{version}},enable=${{ matrix.python-version == 3.8 }}
-            type=raw,value={{version}}-py${{ matrix.python-version }}r
           labels: |
             org.opencontainers.image.title=Nautobot
       - name: "Build"
@@ -113,9 +112,8 @@ jobs:
           flavor: |
             latest=false
           tags: |
-            type=semver,pattern={{version}}+py${{ matrix.python-version }}
+            type=match,pattern=v(.*),group=1,suffix=-py${{ matrix.python-version }}
             type=semver,pattern={{version}},enable=${{ matrix.python-version == 3.8 }}
-            type=raw,value={{version}}-py${{ matrix.python-version }}r
           labels: |
             org.opencontainers.image.title=Nautobot
       - name: "Build Dev Containers"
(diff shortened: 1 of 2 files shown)


.github/workflows/prerelease.yml
```diff
@@ -85,9 +85,8 @@ jobs:
           flavor: |
             latest=false
           tags: |
-            type=semver,pattern={{version}}+py${{ matrix.python-version }}
+            type=match,pattern=v(.*),group=1,suffix=-py${{ matrix.python-version }}
             type=semver,pattern={{version}},enable=${{ matrix.python-version == 3.8 }}
-            type=raw,value={{version}}-py${{ matrix.python-version }}r
```
bryanculver, line 90: This caused the previous run to fail since there are no template arguments for `type=raw`
- Previous run: https://github.com/nautobot/nautobot/actions/runs/5193439184/jobs/9365530741#step:8:50
- Docs: https://github.com/docker/metadata-action#typeraw

.github/workflows/prerelease.yml
```diff
@@ -85,9 +85,8 @@ jobs:
           flavor: |
             latest=false
           tags: |
-            type=semver,pattern={{version}}+py${{ matrix.python-version }}
+            type=match,pattern=v(.*),group=1,suffix=-py${{ matrix.python-version }}
             type=semver,pattern={{version}},enable=${{ matrix.python-version == 3.8 }}
```
bryanculver, line 89: This works as-is because it's not of a Python-specific format (`1.2.3-alpha.1-py3.8`) and only applied to the "default" Python version

.github/workflows/prerelease.yml
```diff
@@ -85,9 +85,8 @@ jobs:
           flavor: |
             latest=false
           tags: |
-            type=semver,pattern={{version}}+py${{ matrix.python-version }}
```
bryanculver, line 88: `{{version}}-py3.8` for example would always output `1.2.3-alpha.1` because `1.2.3-alpha.1-py3.8` is not valid SemVer, but Docker doesn't support the valid SemVer of `1.2.3-alpha.1+py3.9` (see docs: https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions, https://docs.docker.com/engine/reference/commandline/tag/#description)

.github/workflows/prerelease.yml
```diff
@@ -85,9 +85,8 @@ jobs:
...
+            type=match,pattern=v(.*),group=1,suffix=-py${{ matrix.python-version }}
```
bryanculver, line 88: This follows the `type=match` pattern which will strip the `v` off our tag: https://github.com/docker/metadata-action#typematch

HanlinMiao (approved):

Issue #3269, opened 2023-02-08
Dynamic Groups: Improve Reversibility Performance


bryanculver:
Improve reversibility performance via caching (Django's built-in, not CacheOps). Either on the membership instances themselves or the groups. The goal is to cut down reverse lookup time to something reasonable for the user. As an example with current performance, 10k Devices with 10 Dynamic Groups can cause the Device Dynamic Group Detail view (Devices > Device > Dynamic Groups) can sometimes fail to load.

User (likely developers) should be able to choose if they can use the cached lists or need to skip the cache for 100% accuracy.

Cache should have reasonable expiration behavior as well as refreshed with basic operations (modifications of the Dynamic Group's filter definition).

---
_Other ideas have been moved up to #3268_

glennmatthews:
@lampwins Is this addressed in full by #5473 or is there more to be done for this issue still?

PR #4092, opened 2023-07-11
Add Dynamic Group membership caching.

jfach:
---
## Design

The non `_list` methods return a `DynamicGroupQuerySet` which requires a `DynamicGroup.objects.filter(pk__in)` wrap around an internal list.

```
>>> d.dynamic_groups
<DynamicGroupQuerySet [<DynamicGroup: Group A>, <DynamicGroup: Group C>]>
>>> d.dynamic_groups_cached
<DynamicGroupQuerySet [<DynamicGroup: Group A>, <DynamicGroup: Group C>]>
```

The `_list` methods expose the internal list to return the results above if you just need a list. Maybe not necessary to expose it? Saves an additional DB call though.
```
>>> d.dynamic_groups_list
[<DynamicGroup: Group A>, <DynamicGroup: Group C>]
>>> d.dynamic_groups_list_cached
[<DynamicGroup: Group A>, <DynamicGroup: Group C>]
```

## Testing

I conducted a quick test for the performance gains here:

### Setup

I used the demo dataset and created the three following Dynamic Groups:

*Group A:* (31 members)
```
(
  (
    site__slug='ang01' OR site__slug='del01' OR site__slug='hnd02'
  ) AND status__slug='active'
)
```
*Group B:* (0 members)
```
(
  (
    site__slug='hnd01' OR site__slug='jcy' OR site__slug='sin01'
  ) AND status__slug='planned'
)
```
*Group C:* (31 members)
```
group_a OR group_b
```
Results in
```
(
  (
    (
      site__slug='ang01' OR site__slug='del01' OR site__slug='hnd02'
    ) AND status__slug='active'
  ) OR (
    (
      site__slug='hnd01' OR site__slug='jcy' OR site__slug='sin01'
    ) AND status__slug='planned'
  )
)
```

#### Web UI on a Device's "Dynamic Groups" tab (reversibility)

Consistently 3.2 seconds for every access to this page. This is a common use case for dynamic group reversibility.

```
nautobot on  develop [⇣$]
➜ time curl --location 'http://localhost:8080/dcim/devices/3db65457-309f-4c8f-8546-f09781b823fa/dynamic-groups/' \
--header 'Cookie: csrftoken=pJ1g1FEejDvWmYdqQuwCho3KavWQQ7iUm1vSe4ggbtCzWkdDkznTNmRwNRrNQzj7; sessionid=o4zsjjaxq7qh41zx69k3emhmhe8evuii' > /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  231k  100  231k    0     0  73577      0  0:00:03  0:00:03 --:--:-- 73633
curl --location  --header  > /dev/null  0.00s user 0.00s system 0% cpu 3.226 total

nautobot on  develop [⇣$]
➜ time curl --location 'http://localhost:8080/dcim/devices/3db65457-309f-4c8f-8546-f09781b823fa/dynamic-groups/' \
--header 'Cookie: csrftoken=pJ1g1FEejDvWmYdqQuwCho3KavWQQ7iUm1vSe4ggbtCzWkdDkznTNmRwNRrNQzj7; sessionid=o4zsjjaxq7qh41zx69k3emhmhe8evuii' > /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  231k  100  231k    0     0  72227      0  0:00:03  0:00:03 --:--:-- 72330
curl --location  --header  > /dev/null  0.00s user 0.01s system 0% cpu 3.291 total

nautobot on  develop [⇣$]
➜ time curl --location 'http://localhost:8080/dcim/devices/3db65457-309f-4c8f-8546-f09781b823fa/dynamic-groups/' \
--header 'Cookie: csrftoken=pJ1g1FEejDvWmYdqQuwCho3KavWQQ7iUm1vSe4ggbtCzWkdDkznTNmRwNRrNQzj7; sessionid=o4zsjjaxq7qh41zx69k3emhmhe8evuii' > /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  231k  100  231k    0     0  73061      0  0:00:03  0:00:03 --:--:-- 73133
curl --location  --header  > /dev/null  0.00s user 0.00s system 0% cpu 3.253 total
```

**Logs:**

This is from one page access but every request did a full processing of every group, de/constructing it to perform the request.

```
12:46:59.311 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:46:59.312 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:46:59.312 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:46:59.312 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:46:59.312 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:46:59.312 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:46:59.405 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:46:59.405 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:46:59.405 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:46:59.405 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:46:59.405 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:46:59.405 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:46:59.450 DEBUG   nautobot.extras.models.groups groups.py                       generate_query() :
  Processing group Group A...
12:46:59.450 DEBUG   nautobot.extras.models.groups groups.py                       generate_query() :
  Query: Group A -> {'site': ['ang01', 'del01', 'hnd02'], 'status': ['active']} -> union
12:46:59.512 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:46:59.512 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:46:59.512 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:46:59.512 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:46:59.512 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:46:59.512 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:46:59.563 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Non-UUID value detected: Filtering using natural key
12:46:59.563 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Non-UUID value detected: Filtering using natural key
12:46:59.564 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Non-UUID value detected: Filtering using natural key
12:46:59.569 DEBUG   nautobot.extras.models.groups groups.py                       generate_query() :
  Processing group Group B...
12:46:59.569 DEBUG   nautobot.extras.models.groups groups.py                       generate_query() :
  Query: Group B -> {'site': ['hnd01', 'jcy', 'sin01'], 'status': ['planned']} -> union
12:46:59.627 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:46:59.628 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:46:59.628 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:46:59.628 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:46:59.628 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:46:59.628 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:46:59.682 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Non-UUID value detected: Filtering using natural key
12:46:59.682 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Non-UUID value detected: Filtering using natural key
12:46:59.682 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Non-UUID value detected: Filtering using natural key
12:47:00.067 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:47:00.067 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:47:00.068 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:47:00.068 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:47:00.068 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:47:00.068 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:47:00.124 DEBUG   nautobot.extras.models.groups groups.py                       generate_query() :
  Processing group Group A...
12:47:00.124 DEBUG   nautobot.extras.models.groups groups.py                       generate_query() :
  Query: Group A -> {'site': ['ang01', 'del01', 'hnd02'], 'status': ['active']} -> union
12:47:00.194 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:47:00.194 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:47:00.194 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:47:00.194 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:47:00.194 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:47:00.194 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:47:00.252 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Non-UUID value detected: Filtering using natural key
12:47:00.252 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Non-UUID value detected: Filtering using natural key
12:47:00.252 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Non-UUID value detected: Filtering using natural key
12:47:00.259 DEBUG   nautobot.extras.models.groups groups.py                       generate_query() :
  Processing group Group B...
12:47:00.259 DEBUG   nautobot.extras.models.groups groups.py                       generate_query() :
  Query: Group B -> {'site': ['hnd01', 'jcy', 'sin01'], 'status': ['planned']} -> union
12:47:00.327 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:47:00.327 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:47:00.327 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:47:00.327 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:47:00.327 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Model instance detected. Casting to a PK.
12:47:00.327 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  UUID detected: Filtering using field name
12:47:00.386 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Non-UUID value detected: Filtering using natural key
12:47:00.386 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Non-UUID value detected: Filtering using natural key
12:47:00.386 DEBUG   nautobot.utilities.filters filters.py                get_filter_predicate() :
  Non-UUID value detected: Filtering using natural key
12:47:02.444 INFO    django.server :
  "GET /dcim/devices/3db65457-309f-4c8f-8546-f09781b823fa/dynamic-groups/ HTTP/1.1" 200 237027
```

#### ORM

Current cache methods just caches the results on the instance which is fine unless you are going to be iterating over devices because every DG gets re-evaluated for each device:

```py
import time
def timeit(callable):
    start = time.time()
    callable()
    end = time.time()
    print(f"{callable.__name__} took {end - start} seconds")

def get_device_dgs():
    return d.dynamic_groups

>>> d = Device.objects.get(pk="3db65457-309f-4c8f-8546-f09781b823fa")
>>> timeit(get_device_dgs)
get_device_dgs took 0.24410486221313477 seconds
>>> timeit(get_device_dgs)
get_device_dgs took 0.0004742145538330078 seconds
>>> del(d)
>>> d = Device.objects.get(pk="3db65457-309f-4c8f-8546-f09781b823fa")
>>> timeit(get_device_dgs)
get_device_dgs took 0.1245725154876709 seconds
>>> d = Device.objects.get(pk="56b14e70-dd45-4fef-8a58-b5d0fc8bfde6")
>>> timeit(get_device_dgs)
get_device_dgs took 0.13463044166564941 seconds
```

### With Caching

Web Requests: 3.2 seconds -> ~1 second (3x faster)
ORM Access:
- Accessing every devices' DGs: 116 seconds -> 0.86 seconds (145x faster)
- Accessing a single device's DGs: 0.125 seconds (best I saw, not average) -> 0.016 (worst I saw) seconds (7x faster)

#### Web UI on a Device's "Dynamic Groups" tab

First access of the page naturally takes as long as the current implementation, but subsequent requests are much faster. And because the group member list is cached, all instances benefit from the cache.

```
nautobot on  dyanmic-group-caching [⇣$]
➜ time curl --location 'http://localhost:8080/dcim/devices/3db65457-309f-4c8f-8546-f09781b823fa/dynamic-groups/' \
--header 'Cookie: csrftoken=pJ1g1FEejDvWmYdqQuwCho3KavWQQ7iUm1vSe4ggbtCzWkdDkznTNmRwNRrNQzj7; sessionid=o4zsjjaxq7qh41zx69k3emhmhe8evuii' > /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  231k  100  231k    0     0  76168      0  0:00:03  0:00:03 --:--:-- 76263
curl --location  --header  > /dev/null  0.00s user 0.01s system 0% cpu 3.122 total

nautobot on  dyanmic-group-caching [$]
➜ time curl --location 'http://localhost:8080/dcim/devices/3db65457-309f-4c8f-8546-f09781b823fa/dynamic-groups/' \
--header 'Cookie: csrftoken=pJ1g1FEejDvWmYdqQuwCho3KavWQQ7iUm1vSe4ggbtCzWkdDkznTNmRwNRrNQzj7; sessionid=o4zsjjaxq7qh41zx69k3emhmhe8evuii' > /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  231k  100  231k    0     0   199k      0  0:00:01  0:00:01 --:--:--  200k
curl --location  --header  > /dev/null  0.00s user 0.00s system 0% cpu 1.164 total

nautobot on  dyanmic-group-caching [$]
➜ time curl --location 'http://localhost:8080/dcim/devices/3db65457-309f-4c8f-8546-f09781b823fa/dynamic-groups/' \
--header 'Cookie: csrftoken=pJ1g1FEejDvWmYdqQuwCho3KavWQQ7iUm1vSe4ggbtCzWkdDkznTNmRwNRrNQzj7; sessionid=o4zsjjaxq7qh41zx69k3emhmhe8evuii' > /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  231k  100  231k    0     0   221k      0  0:00:01  0:00:01 --:--:--  222k
curl --location  --header  > /dev/null  0.00s user 0.00s system 0% cpu 1.052 total

nautobot on  dyanmic-group-caching [$]
➜ time curl --location 'http://localhost:8080/dcim/devices/56b14e70-dd45-4fef-8a58-b5d0fc8bfde6/dynamic-groups/' \
--header 'Cookie: csrftoken=pJ1g1FEejDvWmYdqQuwCho3KavWQQ7iUm1vSe4ggbtCzWkdDkznTNmRwNRrNQzj7; sessionid=o4zsjjaxq7qh41zx69k3emhmhe8evuii' > /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  229k  100  229k    0     0   269k      0 --:--:-- --:--:-- --:--:--  270k
curl --location  --header  > /dev/null  0.00s user 0.01s system 1% cpu 0.864 total
```

#### ORM

```cache.clear()``` was called before starting the test to ensure the cache was empty and bleed from the web tests weren't affecting this.

Of note:

- Looks like there might be a bug because expected instance cache does not exist anymore
- In branch comparison to accessing with cache and without: ~30x improvement
- `develop` to `dynamic-group-caching` comparison: ~7x improvement (lower bound, best from `develop`, worse from cache in `dynamic-group-caching`)

```py
import time
def timeit(callable):
    start = time.time()
    callable()
    end = time.time()
    print(f"{callable.__name__} took {end - start} seconds")

def get_device_dgs():
    return d.dynamic_groups

def get_device_dgs_cached():
    return d.dynamic_groups_cached

>>> d = Device.objects.get(pk="3db65457-309f-4c8f-8546-f09781b823fa")
>>> timeit(get_device_dgs)
get_device_dgs took 0.48552536964416504 seconds
>>> timeit(get_device_dgs)
get_device_dgs took 0.18681025505065918 seconds  # Instance cache appears to be missing
>>> timeit(get_device_dgs_cached)
get_device_dgs_cached took 0.19591760635375977 seconds  # First time eval
>>> timeit(get_device_dgs_cached)
get_device_dgs_cached took 0.00012564659118652344 seconds  # Instance cache
>>> timeit(get_device_dgs_cached)
get_device_dgs_cached took 0.0001304149627685547 seconds  # Instance cache
>>> del(d)
>>> d = Device.objects.get(pk="3db65457-309f-4c8f-8546-f09781b823fa")
>>> timeit(get_device_dgs)
get_device_dgs took 0.30399060249328613 seconds
>>> timeit(get_device_dgs_cached)
get_device_dgs_cached took 0.018523454666137695 seconds  # Restored from cache
>>> timeit(get_device_dgs_cached)
get_device_dgs_cached took 0.00023674964904785156 seconds  # Instance cache
>>> d = Device.objects.get(pk="56b14e70-dd45-4fef-8a58-b5d0fc8bfde6")
>>> timeit(get_device_dgs)
get_device_dgs took 0.4865992069244385 seconds  # New instance, new query
>>> timeit(get_device_dgs)
get_device_dgs took 0.17659974098205566 seconds  # Again appears instance cache is missing
>>> timeit(get_device_dgs_cached)
get_device_dgs_cached took 0.015529632568359375 seconds  # New instance benefits from cache set from previous instance (fetches from cache) 31x faster for worst case, 8x faster then best case 
>>> timeit(get_device_dgs_cached)
get_device_dgs_cached took 0.0003426074981689453 seconds  # Instance cache
```

### Stressing It

Getting the dynamic groups for all devices in the database, 145x improvement.

```py
def all_dgs_uncached():
    for device in Device.objects.all():
        list(device.dynamic_groups)

def all_dgs_cached():
    for device in Device.objects.all():
        list(device.dynamic_groups_cached)


>>> timeit(all_dgs_uncached)
all_dgs_uncached took 115.79997563362122 seconds
>>> timeit(all_dgs_cached)
all_dgs_cached took 0.8603224754333496 seconds
```

---

# Adds caching for membership lists of Dynamic Groups, configurable and invalidate-able.

## New BaseModel class properties:
`_content_type`
`_content_type_cached`

## New settings:
- `CONTENT_TYPE_CACHE_TIMEOUT`
- `DYNAMIC_GROUPS_MEMBER_CACHE_TIMEOUT`

## New management commands:
- `refresh_content_type_cache`
- `refresh_dynamic_group_member_caches`

## New DynamicGroup model properties:
- `members_cached`

## New DynamicGroup model instance methods:
- `am_i_a_member`
- `update_cached_members`

## New DynamicGroupMixin model mixin properties:
- `dynamic_groups_cached`
- `dynamic_groups_list`
- `dynamic_groups_list_cached`

## New DynamicGroupQuerySet methods:
- `get_list_for_object`
- `_get_eligible_dynamic_groups`

## New signal handlers:
- `dynamic_group_eligible_groups_changed` connected to `post_save` and `post_delete` for `DynamicGroup` model
- `dynamic_group_update_cache_members` connected to `post_save` for `DynamicGroup` and `DynamicGroupMembership` models


nautobot/extras/templates/extras/configcontext_edit.html
```diff
@@ -28,7 +28,7 @@
             {% render_field form.device_redundancy_groups %}
             {% render_field form.tags %}
             {% if settings.CONFIG_CONTEXT_DYNAMIC_GROUPS_ENABLED %}
-            {% render_field form.dynamic_groups %}
+            {% render_field form.dynamic_groups_cached %}
```
bryanculver, line 31: I think `dynamic_groups` here is ConfigContext.dynamic_groups. Possibly not good test coverage here.
```suggestion
            {% render_field form.dynamic_groups %}
```

nautobot/extras/templates/extras/configcontext.html
```diff
@@ -219,9 +219,9 @@
                     <tr>
                         <td>Dynamic Groups</td>
                         <td>
-                            {% if object.dynamic_groups.all %}
+                            {% if object.dynamic_groups_cached.all %}
                                 <ul>
-                                    {% for dynamic_group in object.dynamic_groups.all %}
+                                    {% for dynamic_group in object.dynamic_groups_cached.all %}
```
bryanculver, line 224:
```suggestion
                                    {% for dynamic_group in object.dynamic_groups.all %}
```

nautobot/extras/templates/extras/configcontext.html
```diff
@@ -219,9 +219,9 @@
                     <tr>
                         <td>Dynamic Groups</td>
                         <td>
-                            {% if object.dynamic_groups.all %}
+                            {% if object.dynamic_groups_cached.all %}
```
bryanculver, line 222:
```suggestion
                            {% if object.dynamic_groups.all %}
```

nautobot/core/models/__init__.py
```diff
@@ -34,6 +38,31 @@ def present_in_database(self):
...
+    def _content_type_cached(cls):  # pylint: disable=no-self-argument
+        """
+        Return the ContentType of the object, cached.
+        """
+        cache_key = f"{cls._meta.label_lower}._content_type"
+
+        return cache.get_or_set(cache_key, cls._content_type, get_settings_or_config("CONTENT_TYPE_CACHE_TIMEOUT"))
+
+    @classmethod
+    def _get_content_type(cls, skip_cache=False):
```
glennmatthews, line 58: What's the reasoning behind `_`-prefixing these? They don't seem to be private methods/properties since we're using them outside this class/module - would unprefixed names collide with something, or anything like that?

nautobot/core/tests/test_models.py
```diff
@@ -12,3 +18,47 @@ def clean(self):
     def test_validated_save_calls_full_clean(self):
         with self.assertRaises(ValidationError):
             self.FakeBaseModel().validated_save()
+
+    def test__content_type(self):
+        """
+        Verify that the ContentType of the object is cached.
+        """
+        self.assertEqual(self.FakeBaseModel._get_content_type(), self.FakeBaseModel._get_content_type(skip_cache=True))
```
glennmatthews, line 26: What content-type does `FakeBaseModel` have, anyway? :) Should we test this with a real model instead?

nautobot/core/tests/test_models.py
```diff
@@ -12,3 +18,47 @@ def clean(self):
...
+            self.FakeBaseModel._content_type_cached
+            self.assertEqual(mock__content_type.call_count, 2)
+
+        # Clean-up after ourselves
+        cache.delete(f"{self.FakeBaseModel._meta.label_lower}._content_type")
+
+    @override_settings(CONTENT_TYPE_CACHE_TIMEOUT=0)
+    def test__content_type_caching_disabled(self):
+        """
+        Verify that the ContentType of the object is cached.
```
glennmatthews, line 55:
```suggestion
        Verify that the ContentType of the object is not cached.
```

nautobot/docs/models/extras/dynamicgroup.md
```diff
@@ -602,3 +602,57 @@ To delete a Dynamic Group Membership you would send a `DELETE` request to the de
...
+A Dynamic Group object in the ORM exposes two (2) properties for retrieving the members of that group:
+
+- `members` - The evaluated QuerySet defined by the Dynamic Group and it's potential child groups. This will always perform database queries.
+- `members_cached` - A cached instance of the `members` property. This will only perform database queries if the cache is expired.
+
+Additionally a method is exposed for retrieving the members of a Dynamic Group:
+
+- `get_members()` - A way of retrieving the members of a Dynamic Group and also forcibly refreshing the cache. The arguments are:
+    - `skip_cache` - A boolean value that will ignore the cache and will always return the `members` QuerySet if `True`, defaults to `False`.
+    - `force_update_cache` - A boolean value that will force the cache to be updated with the current membership of the Dynamic Group if `True`, defaults to `False`.
```
glennmatthews, line 623: I'm not sure I understand the use cases and possible permutations of having these as two separate parameters. Especially if `.get_members(skip_cache=True)` is identical to `.members`, what purpose does it serve?

nautobot/docs/models/extras/dynamicgroup.md
```diff
@@ -602,3 +602,57 @@ To delete a Dynamic Group Membership you would send a `DELETE` request to the de
...
+
+An model instance that supports Dynamic Groups will expose the following properties:
+
+- `dynamic_groups` - Full reverse membership of instance to `DynamicGroup` objects, performs the most database queries.
+- `dynamic_groups_cached` - Full reverse membership of instance to `DynamicGroup` objects, uses cached member list if available. Ideal for most use cases.
+- `dynamic_groups_list` - List of membership of instance to `DynamicGroup` objects, performs one less database query than `dynamic_groups`.
```
glennmatthews, line 629: Can you flesh this out more? It's not at all clear to me what the difference is between "Full reverse membership" and "List of membership". Maybe some concrete examples might help?

nautobot/docs/models/extras/dynamicgroup.md
```diff
@@ -602,3 +602,57 @@ To delete a Dynamic Group Membership you would send a `DELETE` request to the de
...
+- `dynamic_groups_list_cached` - List of membership of instance to `DynamicGroup` objects, uses cached member list if available. Performs no database queries in optimal conditions.
+
+The model instance will also expose the following methods:
+
+- `get_dynamic_groups` - A way of retrieving the Dynamic Groups the instance is a member of. The arguments are:
```
glennmatthews, line 634: When/why would I want to use `get_dynamic_groups()` instead of one of the four properties above? Seems like two ways to do the exact same thing?

nautobot/docs/models/extras/dynamicgroup.md
```diff
@@ -602,3 +602,57 @@ To delete a Dynamic Group Membership you would send a `DELETE` request to the de
...
+    - `skip_cache` - A boolean value that will ignore the cache(s) if `True`, defaults to `False`.
+    - `as_queryset` - A boolean value that will return the `dynamic_groups` QuerySet instead of the `dynamic_groups_list` list if `True`, defaults to `False`.
+
+### Invalidating/Refreshing the Cache
+
+If you need to invalidate the membership cache for a Dynamic Group, you can do so by running the management command: `nautobot-server refresh_dynamic_group_member_caches`. This will invalidate the cache for all Dynamic Groups.
```
glennmatthews, line 640: Might be useful to think about adding a "Refresh" button on the Dynamic Group detail view? I know editing and saving the group will have the same effect, but that is more clicks for the user and also results in unnecessarily creating a changelog entry.

bryanculver (commented):
Looks like this could close #3269

Can you add a changelog fragment @jfach? Also this seems like a feature so we should target this against `develop-1.6`

nautobot/extras/models/groups.py
```diff
@@ -1,10 +1,12 @@
 """Dynamic Groups Models."""
 
 import logging
+import pickle
```
glennmatthews, line 4: We've generally been trying to avoid using `pickle` due to its known security concerns. I'm guessing it's necessary here since we're wanting to store and return a fully evaluated QuerySet rather than something like a simple list of PKs? Worth considering though.

nautobot/extras/models/groups.py
```diff
@@ -313,13 +316,78 @@ def get_queryset(self):
...
+
+        Args:
+            skip_cache (bool): Whether to skip the cache and run the query directly. Optional. Default: False.
+            force_update_cache (bool): Whether to force an update of the cache. Optional. Default: False. Ignored if `skip_cache` is True.
+        """
+
+        if skip_cache:
+            return self.members
+
+        if force_update_cache:
```
glennmatthews, line 358: More concerns about having these as two separate params - it's not obvious without reading the code that `force_update_cache` is completely ignored if `skip_cache==True`, for example, and again I wonder why you'd ever want to call `.get_members(skip_cache=True)` instead of just `.members`?

nautobot/extras/models/groups.py
```diff
@@ -313,13 +316,78 @@ def get_queryset(self):
...
+            cache_key = f"{self.__class__.__name__}.{self.id}.cached_members"
+            cache.delete(cache_key)
+
+        return self.members_cached
+
+    def am_i_a_member(self, obj, skip_cache=False):
```
glennmatthews, line 364: Should this be documented in `dynamicgroup.md`? I'm not sure I understand the purpose of this API and whether it's meant for internal use only or consumption.

nautobot/extras/models/groups.py
```diff
@@ -313,13 +316,78 @@ def get_queryset(self):
...
+        Args:
+            obj (django.db.models.Model): The object to check for membership.
+            skip_cache (bool): Whether to skip the cache and run the query directly. Optional. Default: False.
+
+        Returns:
+            bool: True if the object is a member of this group, otherwise False.
+        """
+
+        # Object's class may have content type cached, so check that first.
+        if not skip_cache and hasattr(type(obj), "_content_type"):
```
glennmatthews, line 379: I feel like we *ought* to be able to assert that any object we're checking for DG membership *must* be a `BaseModel` subclass and therefore should always have the `_content_type` property?

nautobot/extras/models/groups.py
```diff
@@ -313,13 +316,78 @@ def get_queryset(self):
...
+            if type(obj)._content_type.id != self.content_type_id:
+                return False
+        else:
+            if ContentType.objects.get_for_model(obj).id != self.content_type_id:
+                return False
+
+        if skip_cache:
+            return obj in list(self.members.all())
```
glennmatthews, line 387: Is this more or less efficient than "self.members.filter(pk=obj.pk).exists()`?

nautobot/extras/models/mixins.py
```diff
@@ -14,19 +14,101 @@
 
 class DynamicGroupMixin:
     """
-    Adds a `dynamic_groups` property that returns a queryset of `DynamicGroup` membership.
+    Adds properties to a model to facilitate reversing DynamicGroup membership:
+
+    - `dynamic_groups` - Full reverse membership of instance to `DynamicGroup` objects, performs the most database queries.
+    - `dynamic_groups_cached` - Full reverse membership of instance to `DynamicGroup` objects, uses cached member list if available. Ideal for most use cases.
+    - `dynamic_groups_list` - List of membership of instance to `DynamicGroup` objects, performs one less database query than `dynamic_groups`.
+    - `dynamic_groups_list_cached` - List of membership of instance to `DynamicGroup` objects, uses cached member list if available. Performs no database queries in optimal conditions.
```
glennmatthews, line 22: Same comment here requesting clarification and/or concrete examples.

nautobot/extras/models/mixins.py
```diff
@@ -14,19 +14,101 @@
...
+        This will use a cached member list of the dynamic groups if available.
+
+        In optimal conditions this will incur a single database query to convert internal list into a queryset which is reasonably performant.
+        """
+        from nautobot.extras.models.groups import DynamicGroup
+
+        if not hasattr(self, "_dynamic_group_cached_queryset"):
             queryset = DynamicGroup.objects.get_for_object(self)
-            self._dynamic_group_queryset = queryset
+            self._dynamic_group_cached_queryset = queryset
```
glennmatthews, line 55: When if ever does this cache get cleared?

nautobot/extras/models/mixins.py
```diff
@@ -14,19 +14,101 @@
...
+        if as_queryset:
+            if skip_cache:
+                return self.dynamic_groups
+            else:
+                return self.dynamic_groups_cached
+        else:
+            if skip_cache:
+                return self.dynamic_groups_list
+            else:
+                return self.dynamic_groups_list_cached
```
glennmatthews, line 111: Same comment here as above - why have two different ways to do the same thing? Why would I want to call `.get_dynamic_groups(as_queryset=False, skip_cache=False)` instead of `.dynamic_groups_list_cached`?

nautobot/extras/tests/test_models.py
```diff
@@ -376,6 +376,11 @@ def test_dynamic_group_assignment_uniqueness(self):
             device_role=self.devicerole,
             device_type=self.devicetype,
         )
+
+        # Ensure group membership cache is up to date
+        self.dynamic_groups.get_members(skip_cache=False, force_update_cache=True)
+        self.dynamic_group_2.get_members(skip_cache=False, force_update_cache=True)
```
glennmatthews, line 382: Would it be more appropriate to override_settings so that caching of members is disabled?

glennmatthews (changes requested):
Thanks for the PR. I get what you're going for here but I have a number of concerns about the implementation and validation.

nautobot/extras/models/groups.py
```diff
@@ -313,13 +316,78 @@ def get_queryset(self):
...
+
+        return unpickled_query
+
+    def get_members(self, skip_cache=False, force_update_cache=False):
+        """
+        Return the member objects for this group.
+
+        Args:
+            skip_cache (bool): Whether to skip the cache and run the query directly. Optional. Default: False.
+            force_update_cache (bool): Whether to force an update of the cache. Optional. Default: False. Ignored if `skip_cache` is True.
```
gsnider2195, line 352:
```suggestion
            skip_cache (bool, optional): Whether to skip the cache and run the query directly. Defaults to False.
            force_update_cache (bool, optional): Whether to force an update of the cache. Ignored if `skip_cache` is True. Defaults to False. 
```

nautobot/extras/models/groups.py
```diff
@@ -313,13 +316,78 @@ def get_queryset(self):
...
+
+    def am_i_a_member(self, obj, skip_cache=False):
+        """
+        Return True if the given object is a member of this group.
+
+        Does check if object's content type matches this group's content type.
+
+        Args:
+            obj (django.db.models.Model): The object to check for membership.
+            skip_cache (bool): Whether to skip the cache and run the query directly. Optional. Default: False.
```
gsnider2195, line 372:
```suggestion
            skip_cache (bool, optional): Whether to skip the cache and run the query directly. Defaults to False.
```

gsnider2195:
Thanks for the PR @jfach. I added a skeleton of the major changes I noticed in the description of this PR. It would help some of us who weren't involved with the design of this feature if you could add a little bit of context for these changes.

bryanculver:
I tested with a bit more complex groups added and the scale became more pronounced:

**Group D**
```
(
  device_role__slug='edge' AND status__slug='active'
)
```

**Group E**
```
(
  status__slug='active' AND device_type_id='2b77a82c-543a-45fe-8fc8-5b9c60eeb004'
)
```
**Group F**
```
group_d AND group_e
```
```
(
  device_role__slug='edge' AND status__slug='active' AND status__slug='active' AND device_type_id='2b77a82c-543a-45fe-8fc8-5b9c60eeb004'
)
```
**Group G**
```
group_c NOT group_f
```

```
(
  (
    (
      (
        site__slug='ang01' OR site__slug='del01' OR site__slug='hnd02'
      ) AND status__slug='active'
    ) OR (
      (
        site__slug='hnd01' OR site__slug='jcy' OR site__slug='sin01'
      ) AND status__slug='planned'
    )
  ) AND (
    NOT (device_role__slug='edge' AND status__slug='active' AND status__slug='active' AND device_type_id='2b77a82c-543a-45fe-8fc8-5b9c60eeb004')
  )
)
```

**Without Cache: 5.5s**

```
nautobot on  develop [⇣$]
➜ time curl --location 'http://localhost:8080/dcim/devices/8424306c-b034-43a6-b49d-bf3a03249702/dynamic-groups/' \
--header 'Cookie: csrftoken=pJ1g1FEejDvWmYdqQuwCho3KavWQQ7iUm1vSe4ggbtCzWkdDkznTNmRwNRrNQzj7; sessionid=o4zsjjaxq7qh41zx69k3emhmhe8evuii' > /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  230k  100  230k    0     0  43135      0  0:00:05  0:00:05 --:--:-- 72320
curl --location  --header  > /dev/null  0.00s user 0.01s system 0% cpu 5.488 total
```

**With Cache: 0.8s**
```
nautobot on  dyanmic-group-caching [$]
➜ time curl --location 'http://localhost:8080/dcim/devices/8424306c-b034-43a6-b49d-bf3a03249702/dynamic-groups/' \
--header 'Cookie: csrftoken=pJ1g1FEejDvWmYdqQuwCho3KavWQQ7iUm1vSe4ggbtCzWkdDkznTNmRwNRrNQzj7; sessionid=o4zsjjaxq7qh41zx69k3emhmhe8evuii' > /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  230k  100  230k    0     0   306k      0 --:--:-- --:--:-- --:--:--  308k
curl --location  --header  > /dev/null  0.00s user 0.00s system 0% cpu 0.759 total
```

nautobot/extras/models/mixins.py
```diff
@@ -14,19 +14,101 @@
...
+        This will NOT use a cached member list of the dynamic groups and will always query the database for each DynamicGroup.
+
+        Additionally, this performs a final database query to turn the internal list into a queryset.
+        """
         from nautobot.extras.models.groups import DynamicGroup
 
-        if not hasattr(self, "_dynamic_group_queryset"):
+        if not hasattr(self, "_dynamic_group_uncached_queryset"):
+            queryset = DynamicGroup.objects.get_for_object(self, skip_cache=True)
+            self._dynamic_group_uncached_queryset = queryset
```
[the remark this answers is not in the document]
bryanculver: In my testing this looks like a bug. This is an instance cache we already use bit we don't return it if it's there.

nautobot/extras/signals.py
```diff
@@ -286,6 +287,53 @@ def dynamic_group_membership_created(sender, instance, raw=False, **kwargs):
 pre_save.connect(dynamic_group_membership_created, sender=DynamicGroupMembership)
 
 
+def dynamic_group_eligible_groups_changed(sender, instance, **kwargs):
+    """
+    When a DynamicGroup is created or deleted, refresh the cache of eligible groups for the associated ContentType.
+
+    Can't change content_type_id on an existing instance, so no need to check for that.
+    """
+    if kwargs.get("created", None) is False:  # Using the same handler for post_save and post_delete
```
jathanism, line 296: Is there a specific need to only explicitly check for `False` or is evaluating falsiness good enough?

```suggestion
    if not kwargs.get("created", False):  # Using the same handler for post_save and post_delete
```

nautobot/extras/signals.py
```diff
@@ -286,6 +287,53 @@ def dynamic_group_membership_created(sender, instance, raw=False, **kwargs):
...
+
+    def _update_parent_cache(parent_group):
+        parent_group.get_members(skip_cache=False, force_update_cache=True)
+        for ancestor in list(parent_group.parents.all()):
+            _update_parent_cache(ancestor)
+
+    _update_parent_cache(instance.parent_group)
+
+
+post_save.connect(dynamic_group_membership_cache_members, sender=DynamicGroupMembership)
```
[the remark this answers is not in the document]
jathanism: Enumerating the graph members here is probably ideal, but it certainly comes back to the "where do we stop" problem? Might be valuable to have the instance traverse up to its parents (if they exist) and then have each parent refresh its descendants, assuming they are unique. There's not a simple solution to this without considering that any given group could have both parents and children, and further each ancestor or descendant could also have permutations of the same groups that we've already traversed.

nautobot/core/tests/test_models.py
bryanculver: Adding
```py
print(self.FakeBaseModel.get_content_type())
self.assertEqual(self.FakeBaseModel._content_type.id, 0)
```

Results in:
```
Stdout:
core | fake base model

AssertionError: 13 != 0
```

nautobot/core/tests/test_models.py
```diff
@@ -12,3 +18,47 @@ def clean(self):
...
+        with patch.object(self.FakeBaseModel, "_content_type", return_value=True) as mock__content_type:
+            self.FakeBaseModel._content_type_cached
+            self.FakeBaseModel._content_type_cached
+            self.FakeBaseModel._content_type_cached
+            self.assertEqual(mock__content_type.call_count, 1)
+
+            time.sleep(2)  # Let the cache expire
+
+            self.FakeBaseModel._content_type_cached
+            self.assertEqual(mock__content_type.call_count, 2)
```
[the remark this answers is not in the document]
bryanculver: It appears `assertNumQueries` requires a transaction test case: https://docs.djangoproject.com/en/3.2/topics/testing/tools/#django.test.TransactionTestCase.assertNumQueries

nautobot/extras/models/groups.py
bryanculver: So following the style rulings: https://stackoverflow.com/questions/21024913/python-isinstance-vs-hasattr-vs-try-except-what-is-better

The preference (it seems) is if you can assume something should work, it's better to do it and handle the failure (save the computational check I guess).

Also, it seems torn on the `isinstance` vs `hasatter`; some seem to suggest duck typing is better than inheritance checking.

Can we get an internal style guide on this?

nautobot/extras/signals.py
bryanculver: Yes, because "created" is not provided in the `post_delete` signal but is in the `post_save` case. Therefore the scenarios we have is `(True, False, Unset)` and we only want to return in the `False` case.

nautobot/extras/tests/test_dynamicgroups.py
```diff
@@ -937,6 +942,61 @@ def test_filter_search(self):
...
+            group.members_cached
+            self.assertEqual(mock_get_queryset.call_count, 2)
+
+        # Clean-up after ourselves
+        cache.delete(f"{group.__class__.__name__}.{group.id}.cached_members")
+
+    @override_settings(DYNAMIC_GROUPS_MEMBER_CACHE_TIMEOUT=0)
+    def test_member_caching_disabled(self):
+        """
+        Verify that the members list of the DynamicGroup is cached.
```
bryanculver, line 984:
```suggestion
        Verify that the members list of the DynamicGroup is not cached.
```

nautobot/extras/tests/test_dynamicgroups.py
```diff
@@ -937,6 +942,61 @@ def test_filter_search(self):
...
+
+        # Ensure the cache is empty from previous tests
+        cache.delete(f"{group.__class__.__name__}.{group.id}.cached_members")
+
+        self.assertEqual(sorted(list(group.members)), sorted(list(group.members_cached)))
+
+    @override_settings(DYNAMIC_GROUPS_MEMBER_CACHE_TIMEOUT=2)
+    def test_member_caching_enabled(self):
+        """
+        Verify that the members list of the DynamicGroup is cached.
```
bryanculver, line 956:
```suggestion
        Verify that the members list of the DynamicGroup is cached and expires.
```

nautobot/extras/tests/test_models.py
```diff
@@ -371,6 +371,7 @@ def test_dynamic_group_assignment_uniqueness(self):
             device_role=self.devicerole,
             device_type=self.devicetype,
         )
+
```
bryanculver, line 374:
```suggestion
```

nautobot/docs/administration/nautobot-server.md
```diff
@@ -480,6 +480,12 @@ Removing expired sessions...
 Invalidating cache...
 ```
 
+### `refresh_dynamic_group_member_caches`
+
+`nautobot-server refresh_dynamic_group_member_caches`
+
+Refresh the cached members of all dynamic groups. This is useful to periodically update the cached list of members of a dynamic group without having to wait for caches to expire, which defaults to one hour.
```
jathanism, line 487:
```suggestion
Refresh the cached members of all Dynamic Groups. This is useful to periodically update the cached list of members of a Dynamic Group without having to wait for caches to expire, which defaults to one hour.
```

I don't see the new settings documented. If you could please add optional settings documentation for `DYNAMIC_GROUPS_MEMBER_CACHE_TIMEOUT` and `CONTENT_TYPE_CACHE_TIMEOUT` and their defaults, and then link to `DYNAMIC_GROUPS_MEMBER_CACHE_TIMEOUT` here in the "which defaults to one hour" text, that would be superfly.

nautobot/docs/models/extras/dynamicgroup.md
```diff
@@ -602,3 +602,66 @@ To delete a Dynamic Group Membership you would send a `DELETE` request to the de
 ```json
 DELETE /api/extras/dynamic-group-memberships/{uuid}/
 ```
+
+## Membership and Caching
+
+Since looking up the members of a Dynamic Group can be a very expensive operation, Nautobot caches the results of these lookups for a configurable amount of time. By default, this cache is set to one (1) hour.This means that if you make a change a member (or potential member) of a Dynamic Group, it may take up to an hour for that change to be reflected in the group's membership. You can change this default value by changing the `DYNAMIC_GROUP_CACHE_TIMEOUT` in the administration panel. This value is in seconds.
```
jathanism, line 608: Any place the new settings are referenced in this document should also link to the forthcoming optional settings documentation for them.

```suggestion
Since looking up the members of a Dynamic Group can be a very expensive operation, Nautobot caches the results of these lookups for a configurable amount of time. By default, this cache is set to one (1) hour. This means that if you make a change a member (or potential member) of a Dynamic Group, it may take up to an hour for that change to be reflected in the group's membership. You can change this default value by changing the `DYNAMIC_GROUP_CACHE_TIMEOUT` in the administration panel. This value is in seconds.
```

nautobot/docs/models/extras/dynamicgroup.md
```diff
@@ -602,3 +602,66 @@ To delete a Dynamic Group Membership you would send a `DELETE` request to the de
...
+
+Creating, updating, or deleting a Dynamic Group will automatically invalidate the cache for that group. This means that the next time the group is evaluated, the cache will be refreshed with the new membership information.
+
+This greatly speeds up the reverse association of an object to the Dynamic Group(s) it is a part of.
```
jathanism, line 612:
```suggestion
This greatly speeds up the reverse association of any object to any Dynamic Group(s) to which it may be associated.
```

nautobot/docs/models/extras/dynamicgroup.md
```diff
@@ -602,3 +602,66 @@ To delete a Dynamic Group Membership you would send a `DELETE` request to the de
...
+
+A Dynamic Group object in the ORM exposes two (2) properties for retrieving the members of that group:
+
+- `members` - The evaluated QuerySet defined by the Dynamic Group and it's potential child groups. This will always perform database queries.
+- `members_cached` - A cached instance of `members.all()`. This will only perform database queries if the cache is expired. You can continue to perform `.filter()` and other QuerySet operations but this will incur additional database queries.
+
+Additionally methods for working with group membership and caching are:
+
+- `update_cached_members` - A way of forcing an update to the cached members of a Dynamic Group. This will always perform database queries. It will also return the updated `members_cached` property.
+
```
jathanism, line 622:
```suggestion
```

nautobot/docs/models/extras/dynamicgroup.md
```diff
@@ -602,3 +602,66 @@ To delete a Dynamic Group Membership you would send a `DELETE` request to the de
...
+- `am_i_a_member` - A way of checking if an object is a member of a Dynamic Group. The arguments are:
+    - `obj` - An instance of an object to check if it is a member of the given group.
+    - `skip_cache` - A boolean value that will use the preferred method of checking an instances member depending if using the cached list (`obj in list`) or the uncached queryset (`members.filter(pk=obj.pk).exists()`) is desired, defaults to `False`.
+
+An model instance that supports Dynamic Groups will expose the following properties:
```
jathanism, line 627:
```suggestion
A model instance that supports Dynamic Groups will expose the following properties:
```

nautobot/docs/models/extras/dynamicgroup.md
```diff
@@ -602,3 +602,66 @@ To delete a Dynamic Group Membership you would send a `DELETE` request to the de
...
+
+- `dynamic_groups` - A QuerySet of `DynamicGroup` objects, performs the most database queries.
```
jathanism, line 629:
```suggestion
- `dynamic_groups` - A QuerySet of `DynamicGroup` objects; performs the most database queries.
```

nautobot/docs/models/extras/dynamicgroup.md
```diff
@@ -602,3 +602,66 @@ To delete a Dynamic Group Membership you would send a `DELETE` request to the de
...
+    - Iterates over all Dynamic Groups that are applicable to the instance's content type, resulting in a list (what is available as `dynamic_groups_list`) of applicable Dynamic Groups.
+    - A final query (`DynamicGroup.objects.filter(pk__in=dynamic_groups_list)`) is necessary to retrieve a QuerySet of `DynamicGroup` objects.
+    - Always performs `N+1` queries where `N` is the number of Dynamic Groups that are applicable to the instance's content type
+    - Evaluation of `instance_1.dynamic_groups` add no benefit to `instance_2.dynamic_groups`: each instance will perform `N+1` queries.
```
jathanism, line 633:
```suggestion
    - Evaluation of `instance_1.dynamic_groups` adds no benefit to `instance_2.dynamic_groups`: each instance will perform `N+1` queries.
```

nautobot/docs/models/extras/dynamicgroup.md
```diff
@@ -602,3 +602,66 @@ To delete a Dynamic Group Membership you would send a `DELETE` request to the de
...
+- `dynamic_groups_cached` - A QuerySet of instance to `DynamicGroup` objects, uses cached member list if available. Ideal for most use cases.
```
jathanism, line 634:
```suggestion
- `dynamic_groups_cached` - A QuerySet of `DynamicGroup` objects; uses cached member list if available. Ideal for most use cases.
```

nautobot/docs/models/extras/dynamicgroup.md
```diff
@@ -602,3 +602,66 @@ To delete a Dynamic Group Membership you would send a `DELETE` request to the de
...
+    - Uses cached member lists for each Dynamic Group that is applicable to the instance's content type. Membership check returns a list (what is available as `dynamic_groups_list_cached`) of applicable Dynamic Groups.
+    - A query (`DynamicGroup.objects.filter(pk__in=dynamic_groups_list_cached)`) is necessary to retrieve a QuerySet of `DynamicGroup` objects.
+    - Ideal for most use cases, performing only `1` query if membership lists are cached.
+    - Evaluation of `instance_1.dynamic_groups_cached` benefits `instance_2.dynamic_groups_cached` as all dynamic group membership lists are cached: `instance_1.dynamic_groups_cached` may perform `N+1` queries, but `instance_2.dynamic_groups_cached` will perform `1` query.
+- `dynamic_groups_list` - List of membership of instance to `DynamicGroup` objects, performs one less database query than `dynamic_groups`.
```
jathanism, line 639:
```suggestion
- `dynamic_groups_list` - List of membership to `DynamicGroup` objects; performs one less database query than `dynamic_groups`.
```

nautobot/docs/models/extras/dynamicgroup.md
```diff
@@ -602,3 +602,66 @@ To delete a Dynamic Group Membership you would send a `DELETE` request to the de
...
+    - The internal list used by `dynamic_groups` to retrieve a QuerySet of `DynamicGroup` objects, but saves the final query.
+    - Beneficial if you don't need QuerySet instance of `DynamicGroup` objects, but want to use uncached membership lists on a large amount of objects.
+    - Always performs `N` queries where `N` is the number of Dynamic Groups that are applicable to the instance's content type
+- `dynamic_groups_list_cached` - List of membership of instance to `DynamicGroup` objects, uses cached member list if available. Performs no database queries in optimal conditions.
```
jathanism, line 643:
```suggestion
- `dynamic_groups_list_cached` - List of membership to `DynamicGroup` objects; uses cached member list if available. Performs no database queries in optimal conditions.
```

nautobot/docs/models/extras/dynamicgroup.md
```diff
@@ -602,3 +602,66 @@ To delete a Dynamic Group Membership you would send a `DELETE` request to the de
...
+    dynamic_group = ObjectVar(
+        model=DynamicGroup
+    )
+
+    class Meta:
+        name = "Update Dynamic Group Member Cache"
+
+    def run(self, data, commit):
+        DynamicGroup.objects.get(pk=data['dynamic_group']).update_cached_members()
+```
```
jathanism, line 667: We might consider publishing this as a system job in keeping with Glenn's suggestion about having a "refresh" button. Perhaps that could be a Job Button? Just spitballing here.

jathanism (changes requested):

nautobot/docs/models/extras/dynamicgroup.md
bryanculver: We don't have system jobs in 1.6 AFAIK.

nautobot/extras/management/commands/refresh_dynamic_group_member_caches.py
```diff
@@ -0,0 +1,23 @@
...
+class Command(BaseCommand):
+    help = "Update the member caches for all DynamicGroups."
+
+    def handle(self, *args, **kwargs):
+        """Run through all Dynamic Groups and ensure their member caches are up to date."""
+
+        if get_settings_or_config("DYNAMIC_GROUPS_MEMBER_CACHE_TIMEOUT") == 0:
+            self.stdout.write(
+                self.style.NOTICE("DYNAMIC_GROUPS_MEMBER_CACHE_TIMEOUT is set to 0; skipping cache refresh")
+            )
```
glennmatthews, line 16: Same thought here about whether we should clear any existing caches in this case?

nautobot/extras/models/groups.py
```diff
@@ -312,13 +315,73 @@ def get_queryset(self):
...
+                unpickled_query = pickle.loads(cached_query)
+        except pickle.UnpicklingError:
+            logger.warning("Failed to unpickle cached members for %s", self)
+        finally:
+            if unpickled_query is None:
+                unpickled_query = self.members.all()
+                cached_query = pickle.dumps(unpickled_query)  # Explicitly pickle the query to evaluate it.
+                cache.set(cache_key, cached_query, get_settings_or_config("DYNAMIC_GROUPS_MEMBER_CACHE_TIMEOUT"))
+
+        return unpickled_query
```
glennmatthews, line 343: Perhaps?

```suggestion
        try:
            cached_query = cache.get(cache_key)
            if cached_query is not None:
                return pickle.loads(cached_query)
        except pickle.UnpicklingError:
            logger.warning("Failed to unpickle cached members for %s", self)

        unpickled_query = self.members.all()
        cached_query = pickle.dumps(unpickled_query)  # Explicitly pickle the query to evaluate it.
        cache.set(cache_key, cached_query, get_settings_or_config("DYNAMIC_GROUPS_MEMBER_CACHE_TIMEOUT"))

        return unpickled_query
```

nautobot/extras/querysets.py
```diff
@@ -163,34 +165,71 @@ def _get_config_context_filters(self):
...
-        for dynamic_group in eligible_groups.iterator():
-            if dynamic_group.members.filter(pk=obj.pk).values_list("pk", flat=True).exists():
-                my_groups.append(dynamic_group.pk)
+        for dynamic_group in list(eligible_groups):
+            if dynamic_group.am_i_a_member(obj, skip_cache=skip_cache):
+                my_groups.append(dynamic_group)
+
+        return my_groups
+
+    def get_for_object(self, obj, skip_cache=False):
```
glennmatthews, line 193: Any concern about breaking changes in that `get_for_object` used to always perform the query, whereas now it'll use the cache unless you explicitly specify `skip_cache=True`? (In other words, should we invert the default for `skip_cache` to stay backwards-compatible?)

nautobot/extras/querysets.py
```diff
@@ -163,34 +165,71 @@ def _get_config_context_filters(self):
...
+            skip_cache: If True, skip the cache and query the database directly.
+        """
+        return self.filter(pk__in=[dg.pk for dg in self.get_list_for_object(obj, skip_cache=skip_cache)])
 
     def get_by_natural_key(self, slug):
         return self.get(slug=slug)
 
+    def _get_eligible_dynamic_groups(self, obj, skip_cache=False):
+        """
+        Return a queryset of `DynamicGroup` objects that are eligible to filter the given object.
```
glennmatthews, line 208:
```suggestion
        Return a queryset of `DynamicGroup` objects that are eligible to potentially contain the given object.
```

nautobot/extras/signals.py
```diff
@@ -286,6 +287,66 @@ def dynamic_group_membership_created(sender, instance, raw=False, **kwargs):
...
 
+def dynamic_group_eligible_groups_changed(sender, instance, **kwargs):
+    """
+    When a DynamicGroup is created or deleted, refresh the cache of eligible groups for the associated ContentType.
+
+    Can't change content_type_id on an existing instance, so no need to check for that.
+    """
+
+    if get_settings_or_config("DYNAMIC_GROUPS_MEMBER_CACHE_TIMEOUT") == 0:
+        # Caching is disabled, so there's nothing to do
```
glennmatthews, line 298: Here too should we clear any existing cache in this case?

nautobot/extras/signals.py
```diff
@@ -286,6 +287,66 @@ def dynamic_group_membership_created(sender, instance, raw=False, **kwargs):
...
+        return
+
+    if kwargs.get("created", None) is False:
+        # We do not care about updates
+        # "created" is not a kwarg for post_delete signals, so is unset or None
+        # "created" is a kwarg for post_save signals, but you cannot change content types of existing groups
+        #   therefor we can ignore cache updates on DynamicGroups updates
+        return
+
+    content_type = instance.content_type
```
glennmatthews, line 308: can/should we use cached content type here?

nautobot/extras/signals.py
```diff
@@ -286,6 +287,66 @@ def dynamic_group_membership_created(sender, instance, raw=False, **kwargs):
...
+        DynamicGroup.objects.filter(content_type_id=instance.content_type_id),
+        get_settings_or_config("DYNAMIC_GROUPS_MEMBER_CACHE_TIMEOUT"),
+    )
+
+
+post_save.connect(dynamic_group_eligible_groups_changed, sender=DynamicGroup)
+post_delete.connect(dynamic_group_eligible_groups_changed, sender=DynamicGroup)
+
+
+def dynamic_group_update_cache_members(sender, instance, **kwargs):
```
glennmatthews, line 321: Perhaps here and below?

```suggestion
def dynamic_group_update_cached_members(sender, instance, **kwargs):
```

nautobot/docs/models/extras/dynamicgroup.md
jathanism: @bryanculver Riiiight. 🧠

nautobot/core/tests/test_models.py
```diff
@@ -12,3 +18,45 @@ def clean(self):
     def test_validated_save_calls_full_clean(self):
         with self.assertRaises(ValidationError):
             self.FakeBaseModel().validated_save()
+
+    def test__content_type(self):
```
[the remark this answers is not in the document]
bryanculver: What about `test_validated_save_calls_full_clean`?

nautobot/docs/models/extras/dynamicgroup.md
```diff
@@ -602,3 +602,65 @@ To delete a Dynamic Group Membership you would send a `DELETE` request to the de
 ```json
 DELETE /api/extras/dynamic-group-memberships/{uuid}/
 ```
+
+## Membership and Caching
+
+Since looking up the members of a Dynamic Group can be a very expensive operation, Nautobot caches the results of these lookups for a configurable amount of time. By default, this cache is set to one (1) hour. This means that if you make a change a member (or potential member) of a Dynamic Group, it may take up to an hour for that change to be reflected in the group's membership. You can change this default value by changing the `DYNAMIC_GROUP_CACHE_TIMEOUT` in the administration panel. This value is in seconds.
```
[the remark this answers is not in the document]
bryanculver: It's actually defaulted to off now per your earlier suggestion.

nautobot/extras/models/groups.py
bryanculver: This will cause it to always evaluate the query if I'm not mistaken, negating the cache. The former ensures we call it if unpickling was unsuccessful.

nautobot/core/tests/test_models.py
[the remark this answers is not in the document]
glennmatthews: Disregard, I was thinking of `nautobot.core.testing.models.ModelTestCases.BaseModelTestCase` in `next`, not this `BaseModelTest` class. Sorry for any confusion.

nautobot/extras/signals.py
bryanculver: 🐛

nautobot/extras/models/groups.py
glennmatthews: Not sure I see that, but I guess my change would cause problems if `cached_query` unpickles to `None` (is that likely?) Fine to leave as-is.

nautobot/extras/signals.py
```diff
@@ -286,6 +287,66 @@ def dynamic_group_membership_created(sender, instance, raw=False, **kwargs):
...
+    """
+    When a DynamicGroup or DynamicGroupMembership is updated, update the cache of members.
+    """
+
+    if get_settings_or_config("DYNAMIC_GROUPS_MEMBER_CACHE_TIMEOUT") == 0:
+        # Caching is disabled, so there's nothing to do
```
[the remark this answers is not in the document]
bryanculver: This would be hard to do because the cache keys aren't easily known.

YAGNI?

nautobot/extras/querysets.py
bryanculver: This added some drop-in improvements to some of the places we use it (GraphQL resolver, etc). But I'm not opposed. @nautobot/core?

nautobot/extras/management/commands/refresh_dynamic_group_member_caches.py
bryanculver: Cache keys aren't readily "exposed" and we'd have to to unset/clear them here (their keys including their PKs). Worth the effort or YAGNI?

nautobot/ipam/graphql/types.py
```diff
@@ -72,7 +72,7 @@ def resolve_vminterface(self, args):
         return None
 
     def resolve_dynamic_groups(self, args):
-        return DynamicGroup.objects.get_for_objectƒ(self)
+        return DynamicGroup.objects.get_for_object(self)
```
bryanculver, line 75: 🐛

nautobot/docs/models/extras/dynamicgroup.md
bryanculver: @lampwins said last week we can add as follow-on

nautobot/core/models/__init__.py
```diff
@@ -34,6 +38,22 @@ def present_in_database(self):
...
+        Return the ContentType of the object, never cached.
+        """
+        return ContentType.objects.get_for_model(cls)
+
+    @classproperty  # https://github.com/PyCQA/pylint-django/issues/240
+    def _content_type_cached(cls):  # pylint: disable=no-self-argument
+        """
+        Return the ContentType of the object, cached.
+        """
+        cache_key = f"{cls._meta.label_lower}._content_type"
```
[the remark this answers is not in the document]
bryanculver: Management command should have used `type(instance)._content_type`.

nautobot/extras/signals.py
bryanculver: Reverted. This is the FK on DynamicGroup to the applicable content type (`DynamicGroup.content_type`)

nautobot/extras/management/commands/refresh_dynamic_group_member_caches.py
glennmatthews: Just trying to think of the case where we have a long non-zero timeout set and then change it to zero - worried about potential confusion/misleading behavior if we keep returning the cached values until the old timeout expires. Maybe not worth the effort.

nautobot/core/management/commands/refresh_content_type_cache.py
```diff
@@ -0,0 +1,25 @@
...
+            return
+
+        content_types = ContentType.objects.all()
+
+        for content_type in content_types:
+            self.stdout.write(self.style.SUCCESS(f"Processing {content_type}"))
+
+            cache.set(
+                content_type.model_class()._content_type_cache_key, content_type, settings.CONTENT_TYPE_CACHE_TIMEOUT
+            )
```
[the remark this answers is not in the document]
bryanculver: Missed the `cache.delete` change here.

glennmatthews (approved):
As long as CI passes, LGTM :)

nautobot/extras/tests/test_dynamicgroups.py
```diff
@@ -937,6 +942,61 @@ def test_filter_search(self):
         expected = [str(device)]
         self.assertEqual(sorted(dg.members.values_list("name", flat=True)), expected)
 
+    def test_member_caching_output(self):
+        group = self.first_child
+
+        # Ensure the cache is empty from previous tests
+        cache.delete(f"{group.__class__.__name__}.{group.id}.cached_members")
```
bryanculver, line 949:
```suggestion
        cache.delete(group.members_cache_key)
```

nautobot/extras/tests/test_dynamicgroups.py
```diff
@@ -937,6 +942,61 @@ def test_filter_search(self):
...
+        Verify that the members list of the DynamicGroup is cached and expires.
+        """
+        group = self.first_child
+
+        class FakeQuerySet:
+            def all(self):
+                return []
+
+        # Ensure the cache is empty from previous tests
+        cache.delete(f"{group.__class__.__name__}.{group.id}.cached_members")
```
bryanculver, line 965:
```suggestion
        cache.delete(group.members_cache_key)
```

nautobot/extras/tests/test_dynamicgroups.py
```diff
@@ -937,6 +942,61 @@ def test_filter_search(self):
...
+        Verify that the members list of the DynamicGroup is not cached.
+        """
+        group = self.first_child
+
+        class FakeQuerySet:
+            def all(self):
+                return []
+
+        # Ensure the cache is empty from previous tests
+        cache.delete(f"{group.__class__.__name__}.{group.id}.cached_members")
```
bryanculver, line 993:
```suggestion
        cache.delete(group.members_cache_key)
```

Issue #3015, opened 2022-12-16
Multiple jobhooks and webhooks are being triggered by a single change


gsnider2195:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5
* Python version: 3.7
* Database platform, version: Postgres
* Middleware(s):

### Steps to Reproduce
1. Start a webserver to watch for requests from nautobot webhooks (`npx http-server test_webhook_server -p 8888`)
2. Create a new webhook for an interface change action (payload url `http://host.docker.internal:8888/test` for docker dev env)
3. Add a new tagged vlan to an interface
4. Watch the logs for your webserver

### Expected Behavior

One webhook is sent to the server

### Observed Behavior

Two webhooks are sent to the server

This is due to the way Nautobot handles change logging for many to many changes. This also affects job hooks and will affect relationship associations after the completion of #2951

gsnider2195:
If we take this approach for firing hooks we'll need to consider what happens when multiple changes are made to an object in a single "transaction" (a single request, or within a context manager, not database transaction). For example, if we delay webhook execution to the end of the transaction and in the transaction the object is updated and then deleted, do we fire a webhook for the update and the delete? If so, we'll need to cache the object to fire the webhook for the update since the object has already been deleted by the time we're firing the webhook.

PR #4109, opened 2023-07-11
[WIP] Patch webhook and jobhook triggered multiple times for single change

MeganerdDev:
# Closes: #3015
# What's Changed

This change patches the multiple firings of a jobhook/webhook that occurs when there is a singular change in Nautobot
- Moves enqueue_job_hooks call from ./extras/signals.py to ./core/middleware.py

# Screenshots
image
image

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/3015.fixed
+++ b/changes/3015.fixed
@@ -1,0 +1,1 @@
+Fixed issue of multipe jobhooks and webhooks being triggered by a single change.
--- a/nautobot/core/middleware.py
+++ b/nautobot/core/middleware.py
@@ -88,7 +88,7 @@ class ObjectChangeMiddleware:
         self.get_response = get_response
 
     def __call__(self, request):
-        from nautobot.extras.jobs import enqueue_job_hooks # avoid circular import
+        from nautobot.extras.jobs import enqueue_job_hooks  # avoid circular import
         from nautobot.extras.models import ObjectChange
 
         # Determine the resolved path of the request that initiated the change
--- a/nautobot/core/middleware.py
+++ b/nautobot/core/middleware.py
@@ -1,3 +1,5 @@
+import uuid
+
 from django.conf import settings
 from django.contrib.auth.middleware import RemoteUserMiddleware as RemoteUserMiddleware_
 from django.db import ProgrammingError
@@ -86,6 +88,9 @@ class ObjectChangeMiddleware:
         self.get_response = get_response
 
     def __call__(self, request):
+        from nautobot.extras.jobs import enqueue_job_hooks # avoid circular import
+        from nautobot.extras.models import ObjectChange
+
         # Determine the resolved path of the request that initiated the change
         try:
             change_context_detail = resolve(request.path).view_name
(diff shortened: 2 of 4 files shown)


MeganerdDev:
- I have tested this to be working on Nautobot v1.5.23b1
- I was assisted on identifying this change, contribution credit to @gsnider2195 !

MeganerdDev:
- Tested on both jobhooks and webhooks

MeganerdDev:
- Ran black code formatter
- Removed unused import in ./extras/signals.py _handle_changed_object per flake8

glennmatthews (commented):
From the test failures it looks like there's an issue on object deletion.

MeganerdDev:
Also thanks @HanlinMiao !

gsnider2195:
> * Tested on both jobhooks and webhooks

Wouldn't the `enqueue_webhooks` call also need to be moved out of the change logging code into the middleware?

glennmatthews (commented):
So on reflection I'm not sure this approach works - you've made it so that a jobhook **only** triggers in response to a web UI request, instead of triggering on any change to an object. That's why the unit tests are failing - the test is using the `web_request_context` context manager around an ORM change, rather than going through the web UI, and is expecting a jobhook to be triggered in response, but it isn't.

gsnider2195:
> So on reflection I'm not sure this approach works - you've made it so that a jobhook **only** triggers in response to a web UI request, instead of triggering on any change to an object. That's why the unit tests are failing - the test is using the `web_request_context` context manager around an ORM change, rather than going through the web UI, and is expecting a jobhook to be triggered in response, but it isn't.

Maybe we should also add it into the `web_request_context` context manager? Would that cover all supported use cases for change logging (UI, REST, ORM with `web_request_context`)?

gsnider2195:
I propose we move this logic out of the middleware and into the `web_request_context` and change the middleware to use `web_request_context` instead of going directly to the `change_logging` context.

gsnider2195:
I tried moving this over to the context manager but now I'm having problems with django not releasing the database connections. I tried moving the database access to a separate method that's called with `transaction.on_commit` and as a signal that gets fired but neither of these fixes the issue.

https://github.com/nautobot/nautobot/compare/ltm-1.6...u/gas-3015-webhook-jobhook?expand=1

gsnider2195:
It turns out the health check was causing the database exhaustion issue. [Related code is here](https://github.com/revsys/django-health-check/blob/593594f6ba9db7cd7fa02a30868133376cc302e4/health_check/mixins.py#L41)

If we take this approach for firing hooks we'll need to consider what happens when multiple changes are made to an object in a single "transaction" (a single request, or within a context manager, not database transaction). For example, if we delay webhook execution to the end of the transaction and in the transaction the object is updated and then deleted, do we fire a webhook for the update and the delete? If so, we'll need to cache the object to fire the webhook for the update since the object has already been deleted by the time we're firing the webhook.

lampwins:
My preference would be that we focus on the final state of the object at the end of a "transaction." As you point out, today that is a web request or a manual invocation of the context manager. We have talked about adding support for the concept of long running web transactions in the future and I think this point is all the more reason to focus on the final state.

We just need to document it this way and users can then adjust their client access patterns if they need to.

gsnider2195:
> My preference would be that we focus on the final state of the object at the end of a "transaction." As you point out, today that is a web request or a manual invocation of the context manager. We have talked about adding support for the concept of long running web transactions in the future and I think this point is all the more reason to focus on the final state.
> 
> We just need to document it this way and users can then adjust their client access patterns if they need to.

We discussed this earlier and came up with these tasks to get this across the finish line:

- [ ] The jobhook/webhook enqueue functions would need minor updates to no longer need access to the instance since all of the required data is already in the object change instance. (1 story point)
- [ ] Update the change logging to always look for an existing objectchange to update before creating a new one. So if an object is updated and then deleted inside of one web request, we just create a “deleted” objectchange instead of creating one for the update and one for the delete. (2 story points or potentially a bit more when thinking about test failures)

This would satisfy change logging only the "end state" and also give us a good path to implementing #4706. We would base these changes against develop as a bugfix, not planning to backport to 1.6. Do you see these tasks needing separate stories or implemented under #3015?

gsnider2195:
Closing in favor of #4810

PR #4112, opened 2023-07-12
Run unittests in parallel

snaselj:
# Closes: #DNE

## What's Changed

- Added `invoke unittest-parallel`.
    - Test outputs are stored in `.test-results` directory.
    - There is a `tests-times.json` file used for distributing tests. Use `invoke sum-tests-times` to update it.
- Added `invoke ps` to display docker compose services.
- Added `invoke export-docker-compose` to export docker compose configuration to `compose.yaml` file.
- Added `invoke logs` to display docker compose logs.
- Fixed `api-status` API call when using higher Redis DB index.

## Considerations

- [Non-durable Django Postgres Testing](https://docs.djangoproject.com/en/3.2/ref/databases/#speeding-up-test-execution-with-non-durable-settings).
- Parallelize other Jobs (pylint takes a lot of time) to spare GH workers.

## TO-DO

- [x] Is there a ticket for this PR?
- [x] Display merged test results.
- [x] Implement to CI.


tasks.py
```diff
@@ -718,7 +735,11 @@ def unittest(
         for individual_exclude_tag in exclude_tag:
             command += f" --tag {individual_exclude_tag}"
 
-    run_command(context, command)
+    env = {"NAUTOBOT_TEST_DB_NAME": test_db_name}
+    if xml_output:
+        env["NAUTOBOT_TEST_OUTPUT_DIR"] = _TEST_RESULTS_DOCKER_DIR
```
glennmatthews, line 740: Probably need alternative logic for `context.nautobot.local` test environment case, as `_TEST_RESULTS_DOCKER_DIR` won't exist on a local filesystem.

pr.md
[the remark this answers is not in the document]
snaselj: fixed

pyproject.toml
```diff
@@ -171,6 +171,7 @@ rich = "~12.5.1"
 towncrier = "~22.8.0"
 # Watch files for celery and mkdocs reload
 watchdog = "~2.1.9"
+unittest-xml-reporting = "~3.2.0"
```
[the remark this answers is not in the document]
snaselj: fixed

tasks.py
```diff
@@ -856,3 +877,111 @@ def tests(context, lint_only=False, keepdb=False):
...
+    if follow:
+        command += "--follow "
+    if tail:
+        command += f"--tail={tail} "
+
+    docker_compose(context, command, service=service)
+
+
+@task
+def export(context):
```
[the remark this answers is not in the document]
snaselj: fixed

changes/4112.added
```diff
@@ -0,0 +1,7 @@
+- Added `invoke unittest-parallel`.
+    - Test outputs are stored in `.test-results` directory.
+    - There is a `tests-times.json` file used for distributing tests. Use `invoke sum-tests-times` to update it.
+- Added `invoke ps` to display docker compose services.
+- Added `invoke export-docker-compose` to export docker compose configuration to `compose.yaml` file.
+- Added `invoke logs` to display docker compose logs.
+- Fixed `api-status` API call when using higher Redis DB index.
```
glennmatthews, line 7: towncrier automatically splits each line in this file into a separate bullet-list item.

```suggestion
Added `invoke unittest-parallel`, which runs unit tests in parallel with results written to `.test-results` directory.
Added `test-times.json` file which is used to distribute tests when running in parallel. This file can be updated by running `invoke sum-tests-times`.
Added `invoke ps` to display docker compose services.
Added `invoke export-docker-compose` to export docker compose configuration to `compose.yaml` file.
Added `invoke logs` to display docker compose logs.
Fixed `api-status` API call when using higher Redis DB index.
```

Also if that last item is a bugfix, it should be split into a separate `4112.fixed` file for consumption by towncrier.

nautobot/core/tests/nautobot_config.py
```diff
@@ -26,9 +26,19 @@
 
 # Redis variables
 
+_GROUP_INDEX = os.getenv("NAUTOBOT_TEST_GROUP_INDEX", "")
+if _GROUP_INDEX:
+    _redis_index = int((int(_GROUP_INDEX) + 2) * 2)
```
glennmatthews, line 31: Please add a comment explaining this calculation?

nautobot/core/tests/nautobot_config.py
```diff
@@ -49,3 +59,5 @@
 
 # Metrics need to enabled in this config as overriding them with override_settings will not actually enable them
 METRICS_ENABLED = True
+
+TEST_OUTPUT_DIR = os.getenv("NAUTOBOT_TEST_OUTPUT_DIR")
```
glennmatthews, line 63: Should this have a fallback if that variable is undefined?

nautobot/core/tests/runner.py
```diff
@@ -1,11 +1,13 @@
-import yaml
+from time import sleep
 
-from django.core.management import call_command
+import yaml
 from django.conf import settings
-from django.test.runner import DiscoverRunner
+from django.core.management import CommandError
+from django.core.management import call_command
```
glennmatthews, line 6:
```suggestion
from django.core.management import call_command, CommandError
```

nautobot/core/tests/runner.py
```diff
@@ -55,7 +57,25 @@ def setup_databases(self, **kwargs):
...
+                while retry_count:
+                    try:
+                        print(f'Flushing test database "{db_name}"...')
+                        call_command("flush", "--no-input", "--database", db_name)
+                        break
+                    except CommandError as e:
+                        retry_count -= 1
+                        sleep(retry_wait)
+                        retry_wait *= 2
+                        print(f"Flush command failed with error: {e}, will retry {retry_count} more times.")
```
glennmatthews, line 76: Might make for better UX if the print command happens before the sleep?

nautobot/core/tests/runner.py
```diff
@@ -55,7 +57,25 @@ def setup_databases(self, **kwargs):
...
+
```
glennmatthews, line 77: Don't we need an `else` on the `while` loop to fail the test as a whole if we retry `retry_count` times and still fail every time?

nautobot/core/tests/runner.py
```diff
@@ -72,7 +92,7 @@ def teardown_databases(self, old_config, **kwargs):
         if settings.TEST_USE_FACTORIES and old_config:
             for connection in old_config:
                 db_name = connection[0].alias
-                print(f'Emptying test database "{db_name}"...')
+                print(f'Flushing test database "{db_name}"...')
                 call_command("flush", "--no-input", "--database", db_name)
```
glennmatthews, line 96: Does this also need a retry loop added?

pyproject.toml
```diff
@@ -171,6 +171,8 @@ rich = "~12.5.1"
 towncrier = "~22.8.0"
 # Watch files for celery and mkdocs reload
 watchdog = "~2.1.9"
+# Store tests results in XML format
+unittest-xml-reporting = "~3.2.0"
```
glennmatthews, line 175: Looking at the pypi page for this library, it has this caveat:

>unittest has the concept of sub-tests for a unittest.TestCase; this doesn't map well to an existing xUnit concept, so you won't find it in the schema. What that means, is that you lose some granularity in the reports for sub-tests.
>
> unittest also does not report successful sub-tests, so the accounting won't be exact.

We do use subTest in a number of our tests, does this pose any problems/concerns?

tasks.py
```diff
@@ -12,10 +12,22 @@
...
 
+import json
 import os
 import re
-
-from invoke import Collection, task as invoke_task
+import subprocess
+import time
+from enum import Enum
+from enum import auto
```
glennmatthews, line 21:
```suggestion
from enum import Enum, auto
```

tasks.py
```diff
@@ -12,10 +12,22 @@
...
+from functools import partial
+from multiprocessing import Pool
+from pathlib import Path
+from typing import List
+from typing import Tuple
```
glennmatthews, line 26:
```suggestion
from typing import List, Tuple
```

tasks.py
```diff
@@ -31,6 +43,40 @@
...
+    "users",
+    "utilities",
+    "virtualization",
+]
+
+_SKIP_TESTS_NAMES = [
+    "/example_jobs/",
+    "/integration/",
+    "nautobot/core/tests/test_cli.py",
+]
```
glennmatthews, line 72: Please add some comments documenting these various constants.

tasks.py
```diff
@@ -31,6 +43,40 @@
...
+    "tenancy",
+    "users",
+    "utilities",
+    "virtualization",
+]
+
+_SKIP_TESTS_NAMES = [
+    "/example_jobs/",
+    "/integration/",
+    "nautobot/core/tests/test_cli.py",
```
glennmatthews, line 71: Why's this test excluded?

tasks.py
```diff
@@ -856,3 +925,300 @@ def tests(context, lint_only=False, keepdb=False):
     build_and_check_docs(context)
     if not lint_only:
         unittest(context, keepdb=keepdb)
+
+
+def remove_ansi_escapes(text: str) -> str:
+    return _ANSI_ESCAPE_RE.sub("", text)
+
+
+def _invoke_unittest_group(default_exec: str, args):
```
glennmatthews, line 934: Docstring please.

tasks.py
```diff
@@ -856,3 +925,300 @@ def tests(context, lint_only=False, keepdb=False):
...
+        process.communicate()
+        if process.returncode != 0:
+            print(50 * "=")
+            print(f"ERROR: group {index} failed with return code {process.returncode}")
+            print(50 * "=")
+
+        return index, process.returncode, time.time() - start_time
+
+
+def _print_tests_results(path: Path) -> dict:
```
glennmatthews, line 971: Docstring please.

tasks.py
```diff
@@ -856,3 +925,300 @@ def tests(context, lint_only=False, keepdb=False):
...
+            return
+        groups[group_index].append(filename[:-3].replace("/", "."))
+        estimations[group_index] += estimation
+
+    if distribution == DistributionType.NAMES:
+        # Sort more specific items first
+        names = sorted(_NAMES_DISTRIBUTION, reverse=True)
+        tests_times_len = len(tests_times)
+        for index, name in enumerate(names):
+            for item in tests_times[:]:
```
glennmatthews, line 1047: I'm not familiar with the `[:]` notation, what does it do?

tasks.py
```diff
@@ -856,3 +925,300 @@ def tests(context, lint_only=False, keepdb=False):
...
+    if default_exec is None:
+        default_exec = "local" if is_truthy(context.nautobot.local) else "exec"
+    if default_exec != "local":
+        stop(context)
+        if default_exec == "exec":
+            docker_compose(context, "up --detach -- nautobot")
+        elif default_exec == "run":
+            docker_compose(context, "up --detach -- db redis selenium")
+        else:
+            raise ValueError(f"Invalid --default-exec: {default_exec}")
```
glennmatthews, line 1118: This is being done to ensure that the worker container is *not* running? Or some other reason? A comment would be appreciated. :)

tests-times.json
glennmatthews, line 1: Would it be possible/worthwhile to move this file to a subdirectory, perhaps `nautobot/core/tests/` so that it would live alongside `nautobot/core/tests/performance_baselines.yml` and not add noise to the project root directory?

glennmatthews:
In https://github.com/nautobot/nautobot/actions/runs/5580198978 (`develop`) unit tests are taking between 26 and 36 minutes to complete. This looks pretty comparable to the CI runtime of the tests in this PR. Am I overlooking something?

snaselj:
> In https://github.com/nautobot/nautobot/actions/runs/5580198978 (`develop`) unit tests are taking between 26 and 36 minutes to complete. This looks pretty comparable to the CI runtime of the tests in this PR. Am I overlooking something?

This one took 16 minutes: https://github.com/nautobot/nautobot/actions/runs/5589264560/jobs/10217398047
It probably depends on current load on GH Actions. I can parallelize tests to multiple GH Jobs, however @bryanculver wanted to have it in single Job only.

nautobot/core/tests/nautobot_config.py
snaselj: If not specified, the repo root directory is used for XML files. Do you have some proposal?

nautobot/core/tests/runner.py
snaselj: Teardown is done in different times, I didn't experience issues with that.

tasks.py
snaselj: It's a copy of the list, used because the algorithm is removing items from the original one. What's your preferred notation? Would `[*original_list]` be better readable?

tasks.py
```diff
@@ -12,10 +12,22 @@
...
+import subprocess
+import time
+from enum import Enum
+from enum import auto
+from functools import partial
+from multiprocessing import Pool
+from pathlib import Path
+from typing import List
+from typing import Tuple
+from xml.etree import ElementTree
```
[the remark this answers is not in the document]
snaselj: xml is native Python library https://docs.python.org/3/library/xml.etree.elementtree.html

I'm not using lxml directly, it's a dependency of `unittest-xml-reporting`

changes/4112.added
snaselj: Fixed

nautobot/core/tests/nautobot_config.py
snaselj: Fixed

nautobot/core/tests/runner.py
snaselj: Fixed

nautobot/core/tests/runner.py
snaselj: Fixed

tasks.py
snaselj: Fixed

tasks.py
snaselj: Explained

tasks.py
snaselj: Fixed

tasks.py
snaselj: Fixed

tasks.py
```diff
@@ -856,3 +925,300 @@ def tests(context, lint_only=False, keepdb=False):
...
+
+    # For NAMES distribution, group count is specified by the number of names in the distribution
+    group_count = len(_NAMES_DISTRIBUTION) if distribution == DistributionType.NAMES else workers
+
+    groups = [[] for _ in range(group_count)]
+    estimations = [0.0 for _ in range(group_count)]
+
+    obsoleted = []
+
+    def add_file(filename: str, estimation: float, group_index=None) -> None:
```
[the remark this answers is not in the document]
snaselj: Fixed.

tasks.py
snaselj: Comment with explanation added.

tasks.py
```diff
@@ -856,3 +925,300 @@ def tests(context, lint_only=False, keepdb=False):
...
+    print(50 * "=")
+
+
+@task
+def sum_tests_times(_context):
+    """Summarize the time it takes to run each test file.
+
+    Use this to rebuild the file in `_TESTS_TIMES_PATH` file after successfully running `invoke unittest-parallel`.
+    Commit this file for later use.
+    This file is used for DistributionType.TIMES and contains the times for each test file.
```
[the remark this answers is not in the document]
snaselj: Explained better.

tests-times.json
```diff
@@ -0,0 +1,318 @@
...
+        "file": "nautobot/circuits/tests/test_api.py",
+        "time": 5
+    },
+    {
+        "file": "nautobot/circuits/tests/test_filters.py",
+        "time": 1
+    },
+    {
+        "file": "nautobot/circuits/tests/test_models.py",
+        "time": 0
```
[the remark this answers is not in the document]
snaselj: Updated the file to round to the nearest second to avoid unnecessary changes. So the min value is 1 s now.

tests-times.json
snaselj: Moved

snaselj:
Thanks for valuable review @glennmatthews . Most comments are addressed now, the only remaining thing is to check subtests reporting. Could you pls. resolve comments that doesn't need to be processed further?

changes/4112.added
```diff
@@ -0,0 +1,6 @@
+Added `invoke unittest-parallel`.
+    - Test outputs are stored in `.test-results` directory.
+    - There is a `tests-times.json` file used for distributing tests. Use `invoke sum-tests-times` to update it.
```
glennmatthews, line 3: Still need to change these to single-line items:

```suggestion
Added `invoke unittest-parallel`, which runs unit tests in parallel with results written to `.test-results` directory.
Added `test-times.json` file which is used to distribute tests when running in parallel. This file can be updated by running `invoke sum-tests-times`.
```

pyproject.toml
glennmatthews, line 175: nit, alphabetization:

```suggestion
# Store tests results in XML format
unittest-xml-reporting = "~3.2.0"
# Watch files for celery and mkdocs reload
watchdog = "~2.1.9"
```

nautobot/core/tests/nautobot_config.py
snaselj: Yup, it's working with both options.

tasks.py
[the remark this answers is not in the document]
snaselj: I see, not sure about the performance, files we are working with are not huge, so the performance should not be an issue IMHO. And noticed this: https://docs.python.org/3/library/xml.etree.elementtree.html `Changed in version 3.3: This module will use a fast implementation whenever available.`

tasks.py
snaselj: Yes, these are excluded in `invoke unittest` by Django discovery (due to tags), and we just mimic this behavior by listing patterns to skip. Not the best solution, just don't see any better way.

glennmatthews (approved):

tasks.py
```diff
@@ -12,10 +12,20 @@
...
+from enum import Enum, auto
+from functools import partial
+from multiprocessing import Pool
+from pathlib import Path
+from typing import List, Tuple
+from xml.etree import ElementTree
+
+from invoke import Collection
+from invoke import task as invoke_task
 from invoke.exceptions import Exit
```
timizuoebideri1, line 29: Reorder imports alphabetically?

timizuoebideri1 (approved):

tasks.py
snaselj: Hi @timizuoebideri1 , `xml` is native Python library, those are sorted in the first group. `invoke` is 3rd party library, sorted in the second group.

glennmatthews (approved):

PR #3674, opened 2023-05-01
Bump napalm

jmcgill298:
# Closes: #3672 
# What's Changed

# TODO

- [x] Explanation of Change(s)
- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [ ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example Plugin Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #3027, opened 2022-12-19
Batch Update Next Dependenices

bryanculver:
# Closes: #3021, #3020, #2902, #2903, #2996, #2997, #3003, #3004
# What's Changed
Renovate is going to be slow one-by-one. Batch updating minor updates which should be GTG.

# TODO

- [x] Explanation of Change(s)
- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [ ] Attached Screenshots, Payload Example
- [x] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example Plugin Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #4118, opened 2023-07-13
[1.6] Add support for Python 3.11, drop support for Python 3.7

### LINKED ISSUE #3561: Uprev pinned dependencies in `develop-1.6` where possible
### Proposed Changes

Review dependencies in `develop` that have newer versions than what we're currently pinned to, and update to any new ones that don't have breaking changes that impact us. Similar to updates made in `next` in #3554.

### Justification

Keeping LTM from getting too stale.

glennmatthews:
In addition to explicitly adding Python 3.11, this is basically a backport of #3554 and #3027 and #3674 to `develop-1.6` branch, with a few additional dependency updates to latest versions made possible by and/or required for 3.11 compatibility.

Closes #3561, though I haven't exhaustively checked all of our dependencies to see what else we can uprev - just the latest versions of the same set of dependencies that were touched in those three PRs.

# What's Changed

- Lots of dependency updates, including some major version bumps (I've reviewed the changelogs a bit and didn't see any red flags, but am mostly relying on testing and CI to catch any breakage)
- Lots of Python file formatting changes due to upreving black (virtually identical to #3554)
- A few more substantive Python file changes due to upreving pylint (virtually identical to #3027)
- Updates to workflows, etc. to drop 3.7 (virtually identical to #3554) and then add 3.11 (new)

# TODO

- [x] Unit, Integration Tests


bryanculver:
Can we also uprev `napalm`.

u1735067:
If that can be useful, `celery<5.3` (specifically `billiard<4.1.0`) is affected by https://github.com/celery/billiard/issues/377 with Python 3.11, so you might want to upgrade Celery too (one way to trigger this issue is by having a job that fails in a hard way, like the DB going down when writing a log entry).

glennmatthews:
Thank you @u1735067! I'll look into updating celery, it looks like it should be feasible here.

nautobot/core/tests/test_releases.py
```diff
@@ -99,6 +99,7 @@ def test_pre_releases(self, mock_cache_get: Mock, mock_cache_set: Mock, mock_req
             "https://localhost/unittest/releases",
             headers={"Accept": "application/vnd.github.v3+json"},
             proxies=settings.HTTP_PROXIES,
+            timeout=15,
```
HanlinMiao, line 102: Just for my own curiosity, why did we add the time out here?

HanlinMiao (approved):
Looks good going through the list of dependency changes and upgrades!

nautobot/core/tests/test_releases.py
glennmatthews: Complaint by pylint that `requests` can potentially hang forever if there's no timeout provided. We already did the same change in `next` in #3027.

PR #4123, opened 2023-07-14
Fixed custom fields not auto-populating when creating objects through…

Kircheneer:
… the ORM.

# Closes: #3312 
# What's Changed

`CustomFieldModel.clean` now auto-populates missing, non-required custom fields in `model.custom_field_data`.

# TODO
- [x] Unit, Integration Tests


nautobot/extras/models/customfields.py
```diff
@@ -230,11 +230,14 @@ def clean(self):
...
+        # Check for missing values, erroring on required ones and populating non-required ones automatically
         for cf in custom_fields.values():
             # 2.0 TODO: #824 replace cf.name with cf.slug
-            if cf.required and cf.name not in self._custom_field_data:
-                raise ValidationError(f"Missing required custom field '{cf.name}'.")
+            if cf.name not in self._custom_field_data:
+                if cf.required:
+                    raise ValidationError(f"Missing required custom field '{cf.name}'.")
+                else:
+                    self._custom_field_data[cf.name] = cf.default
```
glennmatthews, line 240: I wonder if this should only be done on create and not on update? On the other hand, explicitly deleting a field from `self._custom_field_data` in an update probably **should** revert it to its default value if applicable. So maybe it's fine as-is. :-)

nautobot/extras/tests/test_customfields.py
```diff
@@ -29,6 +29,25 @@ def setUp(self):
...
+        custom_field = CustomField.objects.create(
+            # 2.0 TODO: #824 remove name field
+            name=name,
+            slug="custom_field",
+            type=CustomFieldTypeChoices.TYPE_TEXT,
+        )
+        custom_field.validated_save()
+        custom_field.content_types.set([ContentType.objects.get_for_model(Site)])
+
+        site = Site.objects.create(name="Test", status=Status.objects.get(name="Active"))
```
glennmatthews, line 44: Please avoid using `Site` in new code where possible since it makes more work for us in 2.0. :-) I might suggest `Provider` as an alternate simple model to use instead.

nautobot/extras/tests/test_customfields.py
```diff
@@ -29,6 +29,25 @@ def setUp(self):
...
+        site.validated_save()
+
+        self.assertIn(
+            name, site.custom_field_data.keys(), "Custom fields aren't being set properly on a model on save."
+        )
```
glennmatthews, line 49: Possible additional tests:

- on create, omit required custom field, make sure ValidationError is raised (I'd have thought we'd have a test for this but I don't see it!)
- on update, remove custom field, make sure it reverts to default value
- on update, remove required custom field, make sure ValidationError is raised

changes/4123.fixed
```diff
@@ -0,0 +1 @@
+Fixed custom fields not auto-populating when creating objects through the ORM.
```
glennmatthews, line 1: Might be good to add a note in `nautobot/docs/models/extras/customfield.md` explicitly calling out the need to use `validated_save` rather than `save` when creating model instances that may have custom fields?

Actually I see there's already a note to this effect in `nautobot/docs/user-guides/custom-fields.md`. Duplication of documentation strikes again. Still might be good to copy-paste it into the other doc as well.

glennmatthews (commented):
That was fast! :smile:

changes/4123.fixed
glennmatthews, line 1: Nit: suggest changing this to `3312.fixed` since there's an existing issue for this PR.

bryanculver:
If I'm not mistaken this will result in some interesting changes in behavior. When accessed in the UI we pull a CF default and display it if it's missing in the instance's cf_data.

A hypothetical order of operations:

**Today:**
- CF "taco_kind" is created with the default being "chicken"
- Object that would get the CF is created via the ORM
- In the UI, viewing the object will show `taco_kind: chicken`
- CF "taco_kind"'s default is changed to "beef"
- In the UI, viewing the object will show `taco_kind: beef`

This is because it is missing being set on the `cf_data` and has a graceful fallback in certain scenarios

**Tomorrow: (this PR)**
- CF "taco_kind" is created with the default being "chicken"
- Object that would get the CF is created via the ORM
- In the UI, viewing the object will show `taco_kind: chicken`
- CF "taco_kind"'s default is changed to "beef"
- In the UI, viewing the object will show `taco_kind: chicken`

This is because that object will no longer fall back to the previous behavior.

I would suggest at least a migration to ensure we update the `cf_data`s with all existing defaults if missing, and a call out for this change.

nautobot/extras/models/customfields.py
Kircheneer: Take this for example:

```
# Assume Site has a custom field
site = Site.objects.create(name="Banana", description: "Banana Land")
site.validated_save()
```

If I understand the proposal correctly, then `site.custom_field_data` would be empty after this, which IMO is not desirable.

glennmatthews: Agreed. This comment was a half-baked thought and I convinced myself that the way you have it is fine by the time I was done writing it. :-)

glennmatthews:
> If I'm not mistaken this will result in some interesting changes in behavior. When accessed in the UI we pull a CF default and display it if it's missing in the instance's cf_data.

Can you point out where we're doing this, for my reference? I'm looking at `CustomFieldModel.get_custom_fields` and `CustomFieldModel.get_custom_field_groupings` and both of them just do a `self.cf.get()` without providing a default value for missing fields other than `None`.

nautobot/extras/tests/test_customfields.py
Kircheneer: Makes sense.

nautobot/extras/tests/test_customfields.py
Kircheneer: Done and also moved `CustomFieldModelTest` as that seemed a better fit, let me know if that's not the case.

Kircheneer:
To this point:

# After

```
>>> Site.objects.create(name="Bremen", status=Status.objects.get(name="Active")).validated_save()
```

leads to

Screenshot 2023-07-14 at 16 52 45

with this change. This is expected.

# Before

The same ORM command leads to

Screenshot 2023-07-14 at 16 54 23

Changing the default doesn't change this at all.

glennmatthews:
Looks like at least one unit test is failing due to the example-plugin's custom field appearing now unexpectedly in test data.

nautobot/extras/tests/test_customfields.py
```diff
@@ -1289,6 +1290,84 @@ def setUp(self):
...
+            type=CustomFieldTypeChoices.TYPE_TEXT,
+        )
+        custom_field.validated_save()
+        custom_field.content_types.set([ContentType.objects.get_for_model(Provider)])
+
+        provider = Provider.objects.create(name="Test")
+        provider.validated_save()
+
+        self.assertIn(
+            name, provider.custom_field_data.keys(), "Custom fields aren't being set properly on a model on save."
```
glennmatthews, line 1309: Shouldn't this be?

```suggestion
            name, provider._custom_field_data.keys(), "Custom fields aren't being set properly on a model on save."
```

changes/4123.fixed
Kircheneer: What do you think about moving any ORM interaction from the docs into the user guide and then saying "if you want to interact with the ORM, check out the corresponding `[User Guide](...)`"?

Kircheneer:
Everything addressed but the docs, left a question for the docs topic.

changes/4123.fixed
glennmatthews: Sure, cross-referencing instead of duplicating is definitely good. :)

glennmatthews:
Other than the docs update, this looks solid! Thanks again.

nautobot/extras/models/customfields.py
```diff
@@ -230,11 +230,14 @@ def clean(self):
             except ValidationError as e:
                 raise ValidationError(f"Invalid value for custom field '{field_name}': {e.message}")
 
-        # Check for missing required values
+        # Check for missing values, erroring on required ones and populating non-required ones automatically
         for cf in custom_fields.values():
             # 2.0 TODO: #824 replace cf.name with cf.slug
```
HanlinMiao, line 235: Not blocking but we should delete this comment.

nautobot/extras/tests/test_changelog.py
```diff
@@ -4,6 +4,7 @@
 from django.utils.html import escape
 from rest_framework import status
 
+from example_plugin.signals import EXAMPLE_PLUGIN_CUSTOM_FIELD_NAME, EXAMPLE_PLUGIN_CUSTOM_FIELD_DEFAULT
```
HanlinMiao, line 7: Let's go ahead and alpha-sort the imports here.

nautobot/extras/models/customfields.py
HanlinMiao: Wait nvm this is base off `develop`

HanlinMiao (approved):
Looks great, thanks for the PR!

changes/4123.fixed
Kircheneer: Should be addressed, take a look if it works for you.

Issue #3974, opened 2023-06-22
Refine natural keys for various Extras models (+ users.Token)


glennmatthews:
- [x] Add `pk` as natural-key for FileAttachment and ImageAttachment
- [x] Change natural-key for ObjectChange to `pk` for simplicity
- [x] Change natural-key for Token to `pk`
- [x] Change natural-key for ComputedField and Relationship to `key`
- [x] Simplify natural-key for RelationshipAssociation (remove redundant content-type fields)

glennmatthews:
Implemented in #4131

PR #4131, opened 2023-07-17
Natural key definitions for Token and various extras models

glennmatthews:
# Closes: #3974 
# What's Changed

- Corrected the natural-key definitions for `ComputedField`, `CustomField`, `FileAttachment`, `ImageAttachment`, `ObjectChange`, `Relationship`, `RelationshipAssociation`, and `Token` models.
- Re-enabled some Relationship/RelationshipAssociation tests that had been disabled during the IPAM Namespaces work.
- Fixed the problem that while we could do `model.objects.get(composite_key=...)` and `model.objects.filter(composite_key=...)`, we didn't have code in place for `model.objects.exclude(composite_key=...)`.
- Similarly, fixed lack of support for `Prefix.objects.exclude(prefix=...)` and `IPAddress.objects.exclude(address=...)`.

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/changes/3974.fixed
+++ b/changes/3974.fixed
@@ -1,1 +1,1 @@
-Corrected the natural-key definitions for `ComputedField`, `CustomField`, `FileAttachment`, `ImageAttachment`, `ObjectChange`, `Relationship`, `RelationshipAssociation, and `Token` models.
+Corrected the natural-key definitions for `ComputedField`, `CustomField`, `FileAttachment`, `ImageAttachment`, `ObjectChange`, `Relationship`, `RelationshipAssociation`, and `Token` models.
--- a/changes/3974.fixed
+++ b/changes/3974.fixed
@@ -1,0 +1,1 @@
+Corrected the natural-key definitions for `ComputedField`, `CustomField`, `FileAttachment`, `ImageAttachment`, `ObjectChange`, `Relationship`, `RelationshipAssociation, and `Token` models.
--- a/nautobot/core/models/querysets.py
+++ b/nautobot/core/models/querysets.py
@@ -3,6 +3,7 @@ from django.db.models.functions import Coalesce
 
 from nautobot.core.models.utils import deconstruct_composite_key
 from nautobot.core.utils import permissions
+from nautobot.core.utils.data import merge_dicts_without_collision
 
 
 def count_related(model, field):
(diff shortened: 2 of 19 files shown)


changes/3974.fixed
```diff
@@ -0,0 +1 @@
+Corrected the natural-key definitions for `ComputedField`, `CustomField`, `FileAttachment`, `ImageAttachment`, `ObjectChange`, `Relationship`, `RelationshipAssociation, and `Token` models.
```
HanlinMiao, line 1:
```suggestion
Corrected the natural-key definitions for `ComputedField`, `CustomField`, `FileAttachment`, `ImageAttachment`, `ObjectChange`, `Relationship`, `RelationshipAssociation`, and `Token` models.
```

nautobot/extras/tests/test_relationships.py
```diff
@@ -14,24 +13,24 @@
...
-class RelationshipBaseTest(TestCase):
+class RelationshipBaseTest:
     @classmethod
     def setUpTestData(cls):
         cls.location_ct = ContentType.objects.get_for_model(Location)
         cls.rack_ct = ContentType.objects.get_for_model(Rack)
         cls.vlan_ct = ContentType.objects.get_for_model(VLAN)
 
-        cls.locations = Location.objects.all()[:5]
+        cls.locations = Location.objects.get_for_model(Rack).get_for_model(VLAN)[:5]
```
HanlinMiao, line 33: Should we assert something like `len(cls.locations) == 5` or something after this line?

HanlinMiao (approved):
It looks like `RequiredRelationshipTestMixin` is not inheriting from `RelationshipBaseTest`. That might be causing the test failure. But otherwise, it looks good.

nautobot/core/models/querysets.py
```diff
@@ -16,7 +16,25 @@ def count_related(model, field):
...
-class RestrictedQuerySet(QuerySet):
+class CompositeKeyQuerySetMixin:
+    """
+    Mixin to extend a base queryset class with support for filtering by `composite_key=...`.
+    """
+
+    def filter(self, *args, **kwargs):
+        """
+        This is an enhanced version of natural-key slug support from django-natural-keys.
+        Counterpart to BaseModel.composite_key property.
```
itdependsnetworks, line 27: Is it possible to provide an example in the comment? I don't know what an example of what `composite_key` is going to be, so hard for me to imagine what is going on here.

nautobot/extras/tests/test_relationships.py
glennmatthews: Yeah, we should. Since this is a `classmethod` we can't do `self.assertEqual()` but I can maybe just add a plain `assert`?

nautobot/core/models/querysets.py
glennmatthews: Added documentation - this is mostly a mixin for internal use rather than anything a plugin would use, FWIW (as it's included in the `RestrictedQuerySet` class) but documentation is never a bad thing. :)

glennmatthews:
In addressing Ken's feedback, I uncovered that we didn't have a working implementation for `exclude(composite_key=...)`, and in fixing that, found that we also didn't support `Prefix.objects.exclude(prefix=...)` or `IPAddress.objects.exclude(address=...)`, so I fixed those too.

itdependsnetworks:
Thanks!!! I appreciate your thoroughness and this really helps me understand what is going on.

HanlinMiao (approved):
🚢

nautobot/ipam/querysets.py
```diff
@@ -247,39 +247,62 @@ def net_contains_or_equals(self, prefix):
...
-    def get(self, *args, prefix=None, **kwargs):
+    def get(self, *args, **kwargs):
         """
         Provide a convenience for `.get(prefix=<prefix>)`
+
+        Only needed here (normally `get()` just redirects to `filter()` under the hood) because we specifically want
+        in the `get()` case (only) to *not* automatically convert `get(prefix="10.1.1.1/8")` to
+        `get(network="10.0.0.0", prefix_length=8)', i.e. discard the host bits.
+
+        TODO: why *shouldn't* we have similar enforcement on `filter()` and `exclude()`?
```
jathanism, line 258: It looks like this TODO was addressed in the code in the places `merge_dicts_without_collision` is being called? Is that accurate?

jathanism (approved):
Looking pretty sane. Only the one question for that TODO comment. Good to go either way.

nautobot/ipam/querysets.py
glennmatthews: I'm referring to the check at lines 265-266 in this function:

```python
            if str(_prefix) != str(prefix):
                raise self.model.DoesNotExist()
```

There's no equivalent check in the `filter()` method, so I didn't add it in the `exclude()` implementation either.

Issue #3972, opened 2023-06-22
Refine natural keys for various DCIM models


glennmatthews:
- [x] Fix natural-keys for `ComponentModel` base class to be `[name, device]` due to variadic device-location key.
- [x] Fix natural-key field ordering for PowerOutlet and PowerPanel likewise
- [x] Change Cable natural-key to simply `pk` as this is one of the rare cases where the PK is a user-facing identifier for the model.
- [x] Change natural-keys for CablePath and TaggedItem to `pk` as these are not user-facing models
- [x] Add `[units, rack]` natural-key for RackReservation
- [x] VirtualChassis natural-key should just be `name`

bryanculver:
Closed with #4133

PR #4132, opened 2023-07-18
Refine natural keys for various DCIM models

timizuoebideri1:
# Closes: #3972 
# What's Changed

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/models/device_components.py
+++ b/nautobot/dcim/models/device_components.py
@@ -64,6 +64,8 @@ class ComponentModel(PrimaryModel):
     label = models.CharField(max_length=64, blank=True, help_text="Physical label")
     description = models.CharField(max_length=200, blank=True)
 
+    natural_key_field_names = ["name", "device"]
+
     class Meta:
         abstract = True
 
--- a/nautobot/dcim/models/device_components.py
+++ b/nautobot/dcim/models/device_components.py
@@ -400,6 +400,8 @@ class PowerOutlet(CableTermination, PathEndpoint, ComponentModel):
         help_text="Phase (for three-phase feeds)",
     )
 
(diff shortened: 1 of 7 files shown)


nautobot/dcim/models/racks.py
```diff
@@ -477,8 +477,7 @@ class RackReservation(PrimaryModel):
     class Meta:
         ordering = ["created"]
 
-    # TODO: no appropriate natural-key for RackReservation?
-    natural_key_field_names = ["id"]
+    natural_key_field_names = ["units", "rack"]
```
itdependsnetworks, line 480: I would guess this is better off reverse?

glennmatthews: no, it needs to be this way because the natural-key for a `Rack` itself is variable-length.

timizuoebideri1:
Do i still need to write test for these models with changed `natural_key`? 🤷🏽‍♂️
Since there are test already for `natural_key`

nautobot/dcim/models/device_components.py
```diff
@@ -400,6 +402,8 @@ class PowerOutlet(CableTermination, PathEndpoint, ComponentModel):
         help_text="Phase (for three-phase feeds)",
     )
 
+    natural_key_field_names = ["name", "device"]
```
glennmatthews, line 405: Shouldn't be needed I think since `PowerOutlet` inherits from `ComponentModel`?

glennmatthews (approved):

nautobot/dcim/models/device_components.py
timizuoebideri1, line 405:
```suggestion
```

Issue #4081, opened 2023-07-10
2.0: Migration extras.0094 breaks when user created multiple notes on same object in a day


Thetacz:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0.beta2
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Create 2 (or more) Notes in single day (same date) on same object with same user
2. Run extras.0094 migration to remove slug from Notes

Note: the message on these notes can be different, it is not included in the new constraint `unique_together={("assigned_object_type", "assigned_object_id", "user_name", "created")}`

### Expected Behavior
migration passes

### Observed Behavior
```
  Applying extras.0094_remove_slugs...Traceback (most recent call last):                                                                                                                                           
  File "/usr/local/lib/python3.10/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)                                                              
  File "/usr/local/lib/python3.10/site-packages/django_prometheus/db/common.py", line 71, in execute
    return super().execute(*args, **kwargs)                                                              
psycopg2.errors.UniqueViolation: could not create unique index "extras_note_assigned_object_type_id__e1504524_uniq"
DETAIL:  Key (assigned_object_type_id, assigned_object_id, user_name, created)=(1, 258ccdb0-1b29-44aa-a4ba-6468281de6ad, test_user, 2022-09-02 00:00:00+00) is duplicated.
```

glennmatthews:
Thanks for the report!

bryanculver:
I suspect this is because pre-2.0 `created` was a Date field instead of Datetime.

@Thetacz would it be acceptable to create a random second/microsecond to add to the `00:00:00+00` (since we never had the exact time anyway) to ensure uniqueness?

Thetacz:
> I suspect this is because pre-2.0 `created` was a Date field instead of Datetime.
> 
> @Thetacz would it be acceptable to create a random second/microsecond to add to the `00:00:00+00` (since we never had the exact time anyway) to ensure uniqueness?

I already fixed the issue in our data, so it should not apply to us. But generally speaking you can either go the easy way of creating random time OR extract the time from ObjectChange which got created with the Note :)
Or do both with the random time as fallback.

glennmatthews:
resolved by #4134.

PR #4134, opened 2023-07-18
Added data migration to ensure `Note` objects `created` timestamps are unique

HanlinMiao:
# Closes: #4081 
# What's Changed

Manually tested the migration and it works as expected

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/migrations/0094_ensure_note_timestamps_are_unique.py
```diff
@@ -0,0 +1,54 @@
...
+            note.save()
+
+    # If there are still duplicate Note objects exist, we append some random milliseconds of time.
+    duplicate_records = (
+        Note.objects.values(*natural_key_fields)
+        .order_by()
+        .annotate(count=models.Count(natural_key_fields))
+        .filter(count__gt=1)
+    )
+    for duplicate_record in duplicate_records:
```
timizuoebideri1, line 38: Will it not be better if you just added the milliseconds in the first place, Instead of having to check and do it

HanlinMiao: Yeah this approach is from the discussion thread in the issue [itself](https://github.com/nautobot/nautobot/issues/4081#issuecomment-1629411565). But you are right, adding milliseconds in the first place is a cleaner and easier way.

nautobot/extras/migrations/0094_ensure_note_timestamps_are_unique.py
```diff
@@ -0,0 +1,40 @@
...
+        duplicate_notes = Note.objects.filter(
+            assigned_object_type=duplicate_record.get("assigned_object_type"),
+            assigned_object_id=duplicate_record.get("assigned_object_id"),
+            user_name=duplicate_record.get("user_name"),
+        )
```
HanlinMiao, line 26: I had to explicitly filter on each field here instead of doing `**duplicate_record` because ** expression fails on `mysql` for some reason.

nautobot/extras/migrations/0094_ensure_note_timestamps_are_unique.py
```diff
@@ -0,0 +1,40 @@
...
+
+def ensure_note_created_timestamps_are_unique(apps, schema_editor):
+    Note = apps.get_model("extras", "Note")
+    natural_key_fields = ["assigned_object_type", "assigned_object_id", "user_name", "created"]
+
+    # We append some random milliseconds of time to avoid duplicate timestamps for Note objects' created field.
+    duplicate_records = (
+        Note.objects.values(*natural_key_fields)
+        .order_by()
+        .annotate(count=models.Count(natural_key_fields))
```
glennmatthews, line 18: IIRC the reason this is failing on mysql is the use of `Count()` on an array. You might try `Count("pk")` instead?

timizuoebideri1 (approved):

nautobot/extras/migrations/0095_ensure_note_timestamps_are_unique.py
glennmatthews, line 1: Needs to be renumbered after #4146 goes in.

glennmatthews (approved):

Issue #4154, opened 2023-07-24
Last page in Pagination not updating when page size changes


DistantVoyager:
### Environment
* Nautobot version: 2.0.0 next https://github.com/nautobot/nautobot/commit/612f05347dc211dfef5555c9a6186087a09adb73
* Python version: 3.10

### Steps to Reproduce
1. Switch to the new UI in v2.0
2. View an object list such as Devices
3. Change the number of rows per page
4. The last page on "You are on page # out of {last_page}" does not change anymore

### Expected Behavior
The last page gets updated dynamically if the page size changes and increases/decreases the number of pages.

### Observed Behavior
The last page stays the same as it was and does not change to the new number. This seems to have gotten broken from the initial PR. 

![image](https://github.com/nautobot/nautobot/assets/111259311/3d339787-96fa-4496-817e-64d274082ce4)

![image](https://github.com/nautobot/nautobot/assets/111259311/556341f5-80a7-482b-8fb7-2e8db428f67d)

lampwins:
Looks great! How hard would it be to add an arrow that takes you to the last page?

lampwins:
@DistantVoyager works for me.

lampwins:
Great, love it!

HanlinMiao:
Closed by #4155

PR #4155, opened 2023-07-24
Pagination upgrades

DistantVoyager:
# Closes: #4154 
# What's Changed

- Fixes a problem where the last page in the Pagination footer was not updating when the page size changed.
- Added handleOnKeyDown to Pagination to now update the page number when Enter key is pressed (in addition to the existing onBlur events where a user clicks outside of the input box).
- Added double arrow icons to Pagination for navigating to first/last page.

image

image

# TODO

### DIFF
--- a/changes/4154.added
+++ b/changes/4154.added
@@ -1,0 +1,1 @@
+Added handleOnKeyDown to Pagination to now update the page number when Enter key is pressed.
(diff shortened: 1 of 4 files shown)


HanlinMiao (approved):
It looks awesome! Thank you for the PR. Pulled it down and did some manual testing. One thing I noticed is that I can't delete the page number from `PageNumberForm` by hitting the backspace button. But once I started to type in the form, the number changed to what I intended it to be. I think that is a little quirky but a huge improvement regardless.

HanlinMiao (approved):
Very nice! Thank you

Issue #3582, opened 2023-04-14
UI 2.0: /api/ui/ Endpoint for Model, Filterset Definition to Schema


bryanculver:
Carved from #2806, Related (and blocking): #3583

**Context:**
- [ ] The model field to filter field mapping at `BaseFilterSet.FILTER_DEFAULTS` needs to also include the correct UI widget types in the `extra` for the filter mapping
    - [ ] Prototype a JSON representation of the FilterSet to see if we can get away of needing widgets at all for the new UI. JSON Schema would be a good place to start/model from
    - [ ] Serve the above representation on the UI API by some convention to the react UI, Ask for a model and get it's filters and their classes and args. We likely will need to advance it to allow for plugin views which define their own filterset (so maybe the view may need to hint that it needs a different filterset than the default one by the model).
    - [ ] Build the filterset => widget logic in the React UI, similar to what has been done here: https://github.com/nautobot/nautobot/blob/0fffa8ee6d4168fdb11b27019e500e0e1ef9fc1c/nautobot/utilities/utils.py#L866, but in React

**This Story:**

Addresses that first two points:
- Serializing field definitions into a schema, something like JSONSchema. So:
```json
# For Device's Model Fields
{
   "device_type": {
       "type": "ForeignKey",
       "to": "dcim.DeviceType",
   },
   "name": {
       "type": "CharField"
   }
   "..."
}
```
- Providing an API endpoint that returns the representation for a provided content type.

PR #4162, opened 2023-07-25
Filterset field's DOM element in JSON format.

timizuoebideri1:
# Closes: #3582 
# What's Changed

The ability to return the JSON equivalent of a filter field DOM element has been added.

Filtering in the NewUI should be done in the following ways.
1. Get all model fields from the `schema` in the OPTIONS request
2. Get the field types
3. Get the DOM element in JSON representation

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/core/api/views.py
+++ b/nautobot/core/api/views.py
@@ -958,17 +958,17 @@ class GetFilterSetFieldDOMElementAPIView(NautobotAPIVersionMixin, APIView):
             model_form_instance = TempForm(auto_id="id_for_%s")
 
         bound_field = form_field.get_bound_field(model_form_instance, field_name)
-        if "as_json" in request.GET:
+        if request.META.get("HTTP_ACCEPT") == "application/json":
             data = {
                 "field_type": form_field.__class__.__name__,
                 "attrs": bound_field.field.widget.attrs,
-                # The choices attr is present in all field types; choices can be ignored in the UI depending on the `field_type`
-                "choices": list(bound_field.field.widget.choices),
                 # `is_required` is redundant here as it's not used in filterset;
                 # Just leaving it here because this would help when building create/edit form for new UI,
                 # This logic(as_json representation) should be extracted into a helper function at that time
                 "is_required": bound_field.field.widget.is_required,
             }
+            if hasattr(bound_field.field.widget, "choices"):
+                data["choices"] = list(bound_field.field.widget.choices)
         else:
-            data = {"dom_element": bound_field.as_widget()}
+            data = bound_field.as_widget()
         return Response(data)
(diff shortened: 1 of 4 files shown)


nautobot/core/api/views.py
```diff
@@ -959,4 +958,17 @@ def get(self, request):
             model_form_instance = TempForm(auto_id="id_for_%s")
 
         bound_field = form_field.get_bound_field(model_form_instance, field_name)
-        return Response({"dom_element": bound_field.as_widget()})
+        if "as_json" in request.GET:
```
bryanculver, line 961: Does jQuery's request provide an `Accept` header? If it's asking for `application/json` then I'd like to suggest a potential refactor here:

- The legacy UI returns just plain HTML (ie `Response(bound_field.as_widget())`) and jQuery passes the `Accept: */*` or `Accept: text/html` header
- The new UI requests `Accept: application/json` and you get the data below

nautobot/core/tests/test_api.py
```diff
@@ -413,6 +413,18 @@ def test_get_lookup_value_dom_element(self):
             },
         )
 
+        # Also assert JSON representation
+        response = self.client.get(f"{url}?content_type=dcim.location&field_name=name&as_json", **self.header)
+        self.assertEqual(response.status_code, 200)
+        print(response.data)
```
HanlinMiao, line 419: Just a reminder to delete this line.

HanlinMiao (approved):

glennmatthews:
I'm not sure that this fully addresses #3582 - if I'm reading that issue correctly, we need a JSON representation of the *filter fields* (e.g. `MultiValueCharFilter`), not necessarily a JSON representation of the *filter form fields/widgets* (e.g. `MultiValueCharField`). I'm not really seeing how the current JSON representation addresses the needs of the new UI -- but perhaps I'm missing something?

timizuoebideri1:
> I'm not sure that this fully addresses #3582 - if I'm reading that issue correctly, we need a JSON representation of the _filter fields_ (e.g. `MultiValueCharFilter`), not necessarily a JSON representation of the _filter form fields/widgets_ (e.g. `MultiValueCharField`). I'm not really seeing how the current JSON representation addresses the needs of the new UI -- but perhaps I'm missing something?

At the end of the day, all we need is the field name representation that would be mapped in react into its corresponding Input Component. In my opinion, this is much better because it would serve for create/edit forms as well as the fact that we would not need to also create logic to handle create/edit form field mapping.

Additionally, we agreed to limit the scope of this PR to just building the endpoint and excluding the UI component.

glennmatthews (approved):

PR #4170, opened 2023-07-26
Post 1.5.25 develop to next

gsnider2195:
# Do not squash

Sync develop up to next after releasing v1.5.25


poetry.lock
```diff
@@ -1,10 +1,9 @@
-# This file is automatically @generated by Poetry 1.4.2 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.5.1 and should not be changed by hand.
 
 [[package]]
 name = "amqp"
 version = "5.1.1"
 description = "Low-level AMQP client for Python (fork of amqplib)."
-category = "main"
```
gsnider2195, line 7: Looks like the categories got removed from the lockfile when I ran `poetry update`. Is this a problem?

HanlinMiao: https://github.com/python-poetry/poetry/issues/5702#issuecomment-1139512812 It seems like `categories` are going to be dropped in a future release anyways so shouldnt matter much? Do double check me on this.

HanlinMiao (approved):

bryanculver (approved):

nautobot/extras/tests/test_customfields.py
```diff
@@ -1185,6 +1186,84 @@ def setUpTestData(cls):
             weight=200,
         )
 
+    def test_custom_field_dict_population(self):
+        """Test that custom_field_data is properly populated when no data is passed in."""
+        name = "Custom Field"
+        custom_field = CustomField.objects.create(
+            # 2.0 TODO: #824 remove name field
+            name=name,
+            slug="custom_field",
```
bryanculver, line 1195: name=>label
slug=>key

?

PR #4178, opened 2023-07-28
Fix DynamicGroup Serialization Bug

bryanculver:
# Closes: #DNE
# What's Changed

Worked with @lampwins to reproduce this bug.

TLDR: When a Dynamic Group is being saved, there is an edge case where a Filter Form and a _FilterSet_ Form are not identical. A way to reproduce this is with: https://github.com/nautobot/nautobot/tree/u/bryanculver-dg-bug and creating the necessary relationship to filter on it. What happened is the nested filter through the relationship and ultimately is filtering on the `name` of an FK, the FilterSet Form expects inputs to be a list of strings (it's a MultiValueCharFilter after all). But for UX we want to have a selection of names of object instances. We can do this by using the `to_field_name` and processing accordingly. We have logic in Dynamic Groups that does this if the FilterSetForm filed is of the right type. The FilterForm is helping us to this today, but the FilterSetForms need some work to be returning/use the smarter Form field types.

This fix allows to use the FilterForm field first if it's available and falls back if it's not.

Special handling had to be done for Tags because our Tag filter form field isn't a ModelMultipleChoiceField; the FilterSetForm field before was using a ModelMultipleChoiceField.

I'm open to ideas on how to write a test for this which doesn't require a bunch of complex setup like in the branch referenced above.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/models/groups.py
+++ b/nautobot/extras/models/groups.py
@@ -371,7 +371,7 @@ class DynamicGroup(OrganizationalModel):
             field = declared_form.declared_fields.get(field_name, filterset_form.fields[field_name])
             field_value = filterset_form.cleaned_data[field_name]
 
-            if isinstance(field, forms.ModelMultipleChoiceField) or isinstance(field, TagFilterField):
+            if isinstance(field, (forms.ModelMultipleChoiceField, TagFilterField)):
                 field_to_query = field.to_field_name or "pk"
                 new_value = [getattr(item, field_to_query) for item in field_value]
 
--- a/nautobot/extras/models/groups.py
+++ b/nautobot/extras/models/groups.py
@@ -371,6 +371,10 @@ class DynamicGroup(OrganizationalModel):
             field = declared_form.declared_fields.get(field_name, filterset_form.fields[field_name])
             field_value = filterset_form.cleaned_data[field_name]
 
+            # `tag` on FilterForms is not a ModelMultipleChoiceField, so we need to handle it.
+            # On `FilterSetForm.tag`, it is a ModelMultipleChoiceField.
+            # TODO: This could/should check for both "convenience" FilterForm fields (ex: DynamicModelMultipleChoiceField)
+            # and literal FilterSet fields (ex: MultiValueCharFilter).
             if isinstance(field, (forms.ModelMultipleChoiceField, TagFilterField)):
                 field_to_query = field.to_field_name or "pk"
                 new_value = [getattr(item, field_to_query) for item in field_value]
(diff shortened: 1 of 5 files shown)


nautobot/extras/models/groups.py
```diff
@@ -354,6 +355,9 @@ def set_filter(self, form_data):
         # Use the auto-generated filterset form perform creation of the filter dictionary.
         filterset_form = filterset.form
 
+        # Get the declared form for any overloaded form field definitions.
+        declared_form = get_form_for_model(filterset._meta.model, form_prefix="Filter")
```
glennmatthews, line 359: Do we need to fallback to `filterset_form` if `get_form_for_model` fails?

nautobot/extras/models/groups.py
```diff
@@ -354,6 +355,9 @@ def set_filter(self, form_data):
...
+
         # It's expected that the incoming data has already been cleaned by a form. This `is_valid()`
         # call is primarily to reduce the fields down to be able to work with the `cleaned_data` from the
         # filterset form, but will also catch errors in case a user-created dict is provided instead.
```
glennmatthews, line 363: Should we check `declared_form.is_valid` instead of `filterset_form.is_valid` below?

nautobot/extras/models/groups.py
```diff
@@ -364,10 +368,10 @@ def set_filter(self, form_data):
         # empty/null value fields.
         new_filter = {}
         for field_name in filter_fields:
-            field = filterset_form.fields[field_name]
+            field = declared_form.declared_fields.get(field_name, filterset_form.fields[field_name])
             field_value = filterset_form.cleaned_data[field_name]
 
-            if isinstance(field, forms.ModelMultipleChoiceField):
+            if isinstance(field, (forms.ModelMultipleChoiceField, TagFilterField)):
```
glennmatthews, line 374: What's special about TagFilterField? Perhaps add a comment?

nautobot/utilities/forms/fields.py
```diff
@@ -454,6 +454,8 @@ def get_choices():
         # Choices are fetched each time the form is initialized
         super().__init__(label="Tags", choices=get_choices, required=False, *args, **kwargs)
 
+    to_field_name = "slug"
```
glennmatthews, line 457: What'll we need to do in 2.0 to avoid reintroducing the bug, given that tags don't have slugs any more?

glennmatthews (commented):
Would it make sense to add a unit/integration test to check for this bug?

nautobot/extras/models/groups.py
bryanculver: We do when we lookup the field below.

nautobot/extras/models/groups.py
bryanculver: We lean against form validation when filtering AFAICT, and ultimately the FilterSetForm should be doing the enforcement since it's what is based on the FilterSet.

nautobot/utilities/forms/fields.py
bryanculver: IIRC `name` since that's the unique field in 2.0. This keeps things unbreaking.

glennmatthews (approved):
Needs change fragment, otherwise looks solid.

timizuoebideri1 (approved):

Issue #3976, opened 2023-06-22
Enhance CSV import to allow referencing by UUID as an alternative to composite-key


glennmatthews:
Similar to how the JSON REST API and the NaturalKeyOrPKMultipleChoiceFilter work, allow for CSV import to identify related objects by *either*:

- composite-key string (currently supported in 2.x)
- UUID string

The underlying REST serializer knows how to handle UUIDs, of course, so this probably just needs some logic in the NautobotCSVParser class.

glennmatthews:
Implemented by #4188.

PR #4188, opened 2023-08-01
Add support for PKs in CSV import and composite-keys in JSON REST API

glennmatthews:
# Closes: #3976 
# What's Changed

- Slim down `NautobotCSVParser` and augment `NautobotHyperlinkedRelatedField` by moving the composite-key deconstruction and processing from the former to the latter.
  - By passing data through from the parser to the serializer-field with less parser-side processing, this makes it possible for CSV import to use the existing serializer-field support for related object specification by PK
    - Adjust existing test data in dcim/test_views to exercise this functionality
  - As a bonus, this makes it so that related objects can be specified by composite-key in the JSON REST API as well!
     - Add test coverage in core/test_api
- Fix a bug I discovered in manual testing - CSV/REST creation of VRF and Prefix indicate the `namespace` as an optional field, with the intent that if unspecified, the `Global` namespace is implied and should be used; however, this wasn't actually working in the REST (and hence CSV) case because it was defaulting to the UUID of the global namespace instead of the global namespace object itself. This is now fixed.
- Docs updates, UI enhancement.

image

# TODO
- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/core/api/serializers.py
+++ b/nautobot/core/api/serializers.py
@@ -7,7 +7,7 @@ from django.core.exceptions import (
     ObjectDoesNotExist,
     ValidationError as DjangoValidationError,
 )
-from django.db.models import AutoField, ManyToManyField, Model
+from django.db.models import AutoField, ManyToManyField
 from django.db.models.fields.related_descriptors import ManyToManyDescriptor
 from django.urls import NoReverseMatch
 from drf_spectacular.types import OpenApiTypes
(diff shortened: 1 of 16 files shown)


nautobot/core/api/serializers.py
```diff
@@ -137,13 +138,41 @@ def __init__(self, *args, **kwargs):
...
+        # Foreign key where the destination is referenced by string rather than by Python class
+        if getattr(self.parent.Meta.model, self.source, False):
+            return getattr(self.parent.Meta.model, self.source).field.model
+
+        logger.warning(
+            "Unable to determine model for related field %r; "
+            "ensure that either the field defines a 'queryset' or the Meta defines the related 'model'.",
+            self.field_name,
+        )
+        return None
```
jathanism, line 155: It seems like this warning might be redundant seeing as `RelatedField` has this:

https://github.com/encode/django-rest-framework/blob/15c613a9eb645c63102b9e894199bcf1c9bf4d65/rest_framework/relations.py#L107-L115

nautobot/ipam/migrations/0030_ipam__namespaces.py
```diff
@@ -132,7 +132,7 @@ class Migration(migrations.Migration):
             model_name="prefix",
             name="namespace",
             field=models.ForeignKey(
-                default=nautobot.ipam.models.get_default_namespace,
+                default=nautobot.ipam.models.get_default_namespace_pk,
```
jathanism, line 135: I'm a bit surprised that this change was required seeing as it was absolutely working before.

jathanism (changes requested):
I'm asking for one change, but not only if there's not an explicit reason for it to remain how it is!

nautobot/ipam/migrations/0030_ipam__namespaces.py
glennmatthews: This isn't actually a functional change - I renamed the existing `get_default_namespace` function to `get_default_namespace_pk` for clarity since the new function returns the namespace object and the existing function returns the pk.

nautobot/core/api/serializers.py
```diff
@@ -137,13 +138,41 @@ def __init__(self, *args, **kwargs):
                 kwargs["view_name"] = get_route_for_model(kwargs["queryset"].model, "detail", api=True)
         super().__init__(*args, **kwargs)
 
+    @property
```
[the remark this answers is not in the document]
glennmatthews: Done.

jathanism (approved):

HanlinMiao (approved):
🚢

PR #4202, opened 2023-08-02
[WIP] Added mkdocs python classy docs.

itdependsnetworks:
# Closes: Partial #1570 and potentially #3576
# What's Changed

This is POC, it is only a POC, not intended to go in any form as shown. 

This is a POC around building "classy" docs, that show you the nested inheritance of Python Classes. I created a mkdocs plugin to build this out.

The library provides the following features

* The clear definition of the MRO (Method Resolution Order)
* The code in each method
    * The code even when over written, such as when you would like to see `super` being called
    * The code is provided in an ordered (highest is preferred) manner
* The code provided with line numbers from where the code was take from
* The definition of where the code is preferred
* The code is provided in a collapsible block
* Each attribute, and when not preferred shown via a ~~strike-through~~
* Link-able ancestors, when that class is also covered
* Link-able descendants, if applicable 
* Import statement convenience method

ui-view1
ui-view2
ui-view3

# TODO

- [ ] Unit, Integration Tests

There is a few challenges, most importantly is it currently requires you to start the application to run. While I have mostly worked around that when "serving the docs", I haven't when building. There are likely other issues, but again, wanted to show rather than tell.


itdependsnetworks:
One thing to note is the Dockerfile is likely not correct, but I am super confused at this point in how it works, so would need some guidance

lampwins:
Going to close this for now, but still very much want to think about this as a part of the current docs effort.

Issue #4164, opened 2023-07-25
dev container is unable to build docs in bind mounted directory


gsnider2195:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0b2
* Python version: 3.10.12 (3.8.17 in container)
* Database platform, version: N/A
* Middleware(s):

### Steps to Reproduce
1. Build docker container with latest version of `next` (after #4130)
2. Try to run `inv build-and-check-docs`

### Expected Behavior
Docs are built

### Observed Behavior
Permission denied. mkdocs tries to build the docs in `$PWD/project-static/docs` which is owned by the user on the host system (uid 1000 for me) but is running as uid 999 in the container

My quick workaround was to run `build_nautobot_docs` and `build_example_plugin_docs` as the current user:

```diff
diff --git a/tasks.py b/tasks.py
index 9dc59ee2d..e60fb0eaa 100644
--- a/tasks.py
+++ b/tasks.py
@@ -536,7 +536,11 @@ def build_and_check_docs(context):
 def build_nautobot_docs(context):
     "Build Nautobot docs."
     command = "mkdocs build --no-directory-urls --strict"
-    run_command(context, command)
+    if is_truthy(context.nautobot.local):
+        run_command(context, command)
+    else:
+        docker_command = f"run --user {os.getuid()}:{os.getgid()} --entrypoint '{command}' nautobot"
+        docker_compose(context, docker_command, pty=True)
 
 
 def build_example_plugin_docs(context):
@@ -547,7 +551,7 @@ def build_example_plugin_docs(context):
         print_command(local_command)
         context.run(local_command, pty=True)
     else:
-        docker_command = f"run --workdir='/source/examples/example_plugin' --entrypoint '{command}' nautobot"
+        docker_command = f"run --user {os.getuid()}:{os.getgid()} --workdir='/source/examples/example_plugin' --entrypoint '{command}' nautobot"
         docker_compose(context, docker_command, pty=True)
```

I'm still getting permission denied when unittest tries to create `.coverage` so we may need to run unittest as the local user also.

gsnider2195:
My preferred alternate option is to continue to run as root in the `dev` stage. Thoughts @glennmatthews?

glennmatthews:
I'm not necessarily opposed to that option. I was kinda worried that we might hit issues like this with #4130 given the difference in how Docker Desktop works on Mac versus other platforms. Whatever the cleanest/quickest way to fix this turns out to be is fine with me.

[...]

PR #4213, opened 2023-08-03
Modified docker dev stage to run as root user 

gsnider2195:
# Closes: #4164
# What's Changed

Modified docker dev stage to run as root user to prevent permission issues with bind mounts.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/4164.changed
+++ b/changes/4164.changed
@@ -1,0 +1,1 @@
+Modified docker dev stage to run as root user to prevent permission issues with bind mounts.
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -195,9 +195,6 @@ RUN mkdocs build --no-directory-urls && \
 
 FROM python-dev-dependencies AS dev
 
-# hadolint ignore=DL3002
-USER root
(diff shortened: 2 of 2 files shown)


jathanism (approved):

gsnider2195:
I ended up leaving the rest of the build steps in the dev stage alone because other stages are benefiting from the docker cache in the `poetry install` step even though no other stage is based off of the dev stage. In order to run `poetry install` as the nautobot user and continue to use its pip cache, we needed to also use the nautobot user to run the `COPY` directives earlier in the dev stage. Tldr: all we want to change here is the user that's running the container.

glennmatthews (approved):
Seems lower risk this way. If it works locally for you I have reasonable confidence in it. :-)

gsnider2195:
> Seems lower risk this way. If it works locally for you I have reasonable confidence in it. :-)

Unit tests and doc builds were the pain points I was having and those are working for me locally with these changes. I also tested the new UI build and the old and new UI work correctly.

Issue #4217, opened 2023-08-04
Duplicate Git repos with jobs breaks `post_upgrade`


whitej6:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.24 (likely affects >=1.3)
* Python version: 3.8
* Database platform, version: psql
* Middleware(s):n/a

### Steps to Reproduce
1. Create a duplicate of a git repository that provides jobs
2. run `migrate` or `post_upgrade`

### Expected Behavior
Duplicate git repository fails validation on submit of create form (unique with url & provides X)
**_OR_**
Duplicate jobs catches exceptions & provides a critical log message identifying the duplicate jobs

### Observed Behavior

1. Second repository clones then errors out on creation of duplicate job
2. `post_upgrade`/`migrate` cause an `IntegrityError`

PR #4219, opened 2023-08-04
Closes: #4217 - Prevent duplicate repo with duplicated `provided_content`

whitej6:
Not sure this is the best way to accomplish this, tests work but the clean method on the model feels inefficient.

# Closes: #4217
# What's Changed

Provides validation to prevent a duplicate repo (same `remote_url`) providing at least 1 duplicated content in `provided_contents`

### DIFF
--- a/changes/4217.fixed
+++ b/changes/4217.fixed
@@ -1,0 +1,1 @@
+Fixed duplicated git repos providing duplicated contents in the `provided_contents` attribute, duplicate repos with the same `remote_url` cannot provide a duplicate content.
--- a/changes/4217.fixed
+++ b/changes/4217.fixed
@@ -1,1 +1,0 @@
-Fixed duplicated git repos providing duplicated contents in the `provided_contents` attribute, duplicate repos with the same `remote_url` cannot provide a duplicate content.
--- a/nautobot/extras/models/datasources.py
+++ b/nautobot/extras/models/datasources.py
@@ -179,5 +179,5 @@ class GitRepository(PrimaryModel):
             q |= models.Q(provided_contents__contains=item)
         duplicate_repos = GitRepository.objects.filter(remote_url=self.remote_url).exclude(id=self.id).filter(q)
         if duplicate_repos.exists():
-            raise ValidationError(f"Duplicate GitRepository detected providing one or more duplicated content")
+            raise ValidationError("Duplicate GitRepository detected providing one or more duplicated content")
         super().clean()
--- a/nautobot/extras/models/datasources.py
+++ b/nautobot/extras/models/datasources.py
@@ -174,10 +174,11 @@ class GitRepository(PrimaryModel):
 
     def clean(self):
         """Overload clean to enforce a repo with duplicate url does not have duplicate provided_contents."""
-        q = models.Q()
-        for item in self.provided_contents:
-            q |= models.Q(provided_contents__contains=item)
-        duplicate_repos = GitRepository.objects.filter(remote_url=self.remote_url).exclude(id=self.id).filter(q)
-        if duplicate_repos.exists():
-            raise ValidationError("Duplicate GitRepository detected providing one or more duplicated content")
+        if self.provided_contents:
+            q = models.Q()
+            for item in self.provided_contents:
+                q |= models.Q(provided_contents__contains=item)
+            duplicate_repos = GitRepository.objects.filter(remote_url=self.remote_url).exclude(id=self.id).filter(q)
+            if duplicate_repos.exists():
+                raise ValidationError("Duplicate GitRepository detected providing one or more duplicated content")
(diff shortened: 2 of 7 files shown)


glennmatthews:
Looks pretty reasonable to me. Thanks!

nautobot/extras/models/datasources.py
```diff
@@ -170,3 +171,12 @@ def on_commit_callback():
...
 
         transaction.on_commit(on_commit_callback)
+
+    def clean(self):
+        """Overload clean to enforce a repo with duplicate url does not have duplicate provided_contents."""
+        duplicate_repos = GitRepository.objects.filter(remote_url=self.remote_url).exclude(id=self.id)
+        if duplicate_repos.exists():
+            for content in self.provided_contents:
+                if duplicate_repos.filter(provided_contents__contains=content).exists():
+                    raise ValidationError(f"Duplicate GitRepository detected providing {content}")
```
jathanism, line 181: I feel like this error message should also include the URL. E.g. "Duplicate exists with this URL that provides this content"?

whitej6: You are creating the repo so you are already passing in the single attr for URL. The issue is less about duplicate URL but more about 2 to N GitRepositories providing the same content. You could absolutely have different GitRepository objects tracking different branches and each one provides different content. This is mainly to block the ability for a duplicate repo to provide `extras.job` because it will cause failures for `migrate` and `post_upgrade` until the duplicate repo is removed.

glennmatthews: Should the duplicate check only apply to repos that have the same specified `branch` as well as URL?

whitej6: I see that as a possible scenario BUT that would mean in the scenario I ran into today a user created 2 repos with different branches, one branch had all the original jobs plus an additional job so url + branch wouldn't have caught it.

glennmatthews:
Looks like tests need to be updated so as to not try to do the exact thing this is protecting against. 😄

gsnider2195:
Can you add a line or 2 to the docs for GitRepository that points out this constraint?

nautobot/extras/models/datasources.py
```diff
@@ -170,3 +171,14 @@ def on_commit_callback():
...
+
+    def clean(self):
+        """Overload clean to enforce a repo with duplicate url does not have duplicate provided_contents."""
+        if self.provided_contents:
+            q = models.Q()
+            for item in self.provided_contents:
+                q |= models.Q(provided_contents__contains=item)
+            duplicate_repos = GitRepository.objects.filter(remote_url=self.remote_url).exclude(id=self.id).filter(q)
+            if duplicate_repos.exists():
+                raise ValidationError("Duplicate GitRepository detected providing one or more duplicated content")
```
gsnider2195, line 183:
```suggestion
                raise ValidationError("Duplicate GitRepository detected with this remote url and provided content")
```

HanlinMiao, line 183: Perhaps?
```suggestion
                raise ValidationError("Duplicate GitRepository detected providing one or more duplicated content types")
```

HanlinMiao (approved):

nautobot/extras/models/datasources.py
```diff
@@ -170,3 +171,17 @@ def on_commit_callback():
...
 
         transaction.on_commit(on_commit_callback)
+
+    def clean(self):
+        """Overload clean to enforce a repo with duplicate url does not have duplicate provided_contents."""
+        if self.provided_contents:
+            q = models.Q()
+            for item in self.provided_contents:
+                q |= models.Q(provided_contents__contains=item)
+            duplicate_repos = GitRepository.objects.filter(remote_url=self.remote_url).exclude(id=self.id).filter(q)
```
timizuoebideri1, line 181:
```suggestion
           duplicate_repos = GitRepository.objects.exclude(id=self.id).filter(
                provided_contents__contains=self.provided_contents, remote_url=self.remote_url
            )
```

How about this?

timizuoebideri1: Oooh that's true.

timizuoebideri1 (approved):

changes/4217.fixed
```diff
@@ -0,0 +1 @@
+Fixed duplicated git repos providing duplicated contents in the `provided_contents` attribute, duplicate repos with the same `remote_url` cannot provide a duplicate content.
```
gsnider2195, line 1:
```suggestion
Fixed duplicated git repos providing duplicated contents in the `provided_contents` attribute, multiple repos with the same `remote_url` cannot provide a duplicate content.
```
My brain is losing context for the word "duplicate"

gsnider2195 (approved):

PR #4220, opened 2023-08-04
Added `workflow_dispatch` to Upstream Testing workflow.

jathanism:
# Closes: #DNE
# What's Changed
- This will allow it to be manually executed by providing the `invoke_context_name` and `plugin_name` arguments in the UI.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/.github/workflows/plugin_upstream_testing_base.yml
+++ b/.github/workflows/plugin_upstream_testing_base.yml
@@ -14,11 +14,11 @@ on:
     inputs:
       invoke_context_name:
         required: true
-        description: "Invoke context name (This will be `UPPERCASE_WITH_UNDERSCORES`)"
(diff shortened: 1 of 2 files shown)


.github/workflows/plugin_upstream_testing_base.yml
```diff
@@ -10,6 +10,16 @@ on:
...
         type: string
+  workflow_dispatch:
+    inputs:
+      invoke_context_name:
+        required: true
+        description: "Invoke context name (This will be `UPPERCASE_WITH_UNDERSCORES`)"
+        type: string
+      plugin_name:
+        required: true
+        description: "Plugin name (This will be `lowercase-with-hyphens`)"
```
glennmatthews, line 21: Maybe add concrete examples to each of these descriptions?

glennmatthews (approved):

jathanism:
On closer review, I am realizing that this is is not going to do what I want it to do as the context is in each of the app repos, not here.

For now, we will have to update each app repo instead. Longer term moving the execution logic into a Nautobot repo workflow would be better.

Issue #4222, opened 2023-08-04
Dynamic filter form population can alter presentation of ChoiceFields elsewhere in the UI


glennmatthews:
### Environment
* Nautobot version (Docker tag too if applicable): 1.6.0rc1

### Steps to Reproduce
1. Log in to Nautobot
2. In one tab, open the `/dcim/devices/` list view, open the `Filter` modal, and switch to the `Advanced` tab.
3. In another browser tab, open the "run" view for a Job that defines a `ChoiceVar`. Observe that the ChoiceVar is correctly represented as a single-value Select2 widget in the UI.
4. In the first tab, in the Filter modal, select "Rack Face" in the first column, then "exact" in the second column.
5. Refresh the second tab.

### Expected Behavior

No change on page refresh

### Observed Behavior

The `ChoiceVar` now gets incorrectly rendered as a multi-value Select2 widget.

This is due to incorrect code in `get_filterset_parameter_form_field()` which is inadvertently modifying the `forms.ChoiceField` class directly instead of only changing a single instance of that class.

glennmatthews:
Thank you to the many folks who helped identify, reproduce, and troubleshoot this issue, including @Dav-C, @jtdub, @joewesch. @matt852.

glennmatthews:
Workaround to temporarily "unbreak" the Job form after this issue is encountered:

- Navigate again to `/dcim/devices`, open the Filter modal, and switch to the Advanced tab.
- Select "Has a primary IP" and then select "exact"
- Refresh the job tab.

This is because there's a second code path that also incorrectly modifies `forms.ChoiceField` when making a different kind of selection from the advanced filter form; we're just lucky in that this code path modifies it to somewhat closer to the original, desired behavior of the form field. Both code paths will be fixed in #4223.

PR #4223, opened 2023-08-04
Fix inadvertent modification of `forms.ChoiceField` class definition

glennmatthews:
# Closes: #4222 
# What's Changed

Avoid directly modifying the `forms.ChoiceField` class with a custom `widget` default; instead, pass the widget class through when instantiating specific instances of ChoiceField.

Verified through manual testing following the steps described in #4222 and confirming that filter fields like `face` and `has_interfaces` are still represented by the correct widgets but Job variables do not change presentation as a result.

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/utilities/utils.py
+++ b/nautobot/utilities/utils.py
@@ -915,17 +915,9 @@ def get_filterset_parameter_form_field(model, parameter, filterset=None):
                 choices_as_strings=True, queryset=queryset_map[plural_name]().as_queryset()
             )
     elif isinstance(field, (filters.MultipleChoiceFilter, filters.ChoiceFilter)) and "choices" in field.extra:
-        form_field_class = forms.ChoiceField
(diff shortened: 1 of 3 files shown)


nautobot/utilities/utils.py
```diff
@@ -915,17 +915,9 @@ def get_filterset_parameter_form_field(model, parameter, filterset=None):
                 choices_as_strings=True, queryset=queryset_map[plural_name]().as_queryset()
             )
     elif isinstance(field, (filters.MultipleChoiceFilter, filters.ChoiceFilter)) and "choices" in field.extra:
```
glennmatthews, line 917: This may also be a bit incorrect - I don't think we should be presenting a multiple-choice formfield/widget for a `filters.ChoiceFilter` - but it's outside the narrow scope of fixing this bug.

jathanism (approved):

Issue #3636, opened 2023-04-21
IP-Interface M2M: "Assign IP Address(s)" Button on Interface List View


bryanculver:
### Tasks
- [ ] Re-add "Assign IP Address(s)" Button On Device's Interface List View
    - Needs some investigation with the changes in #3244, current UI support, 2.0 modal support
    - Also should consider the steps needed with #3244 to create an IP address from an Interface form
    - Similarly, `IPAddressEditView` `interface` GET param (see: https://github.com/nautobot/nautobot/pull/3399/files#diff-caf8a412a9f30295cacfae888c0fbc1bdc2a7cd2643df55f444414f363113a7bR479)
        - Again if we don't re-introduce the assign an Interface from IPAddress form this goes away. If instead we have a modal UX for creating an IPAddress from editing an Interface, the address is created first and then the Interface form gets the resulting PK of the IPAddress to add to it's multi-select

gsnider2195:
Completed by #4224 4224

PR #4224, opened 2023-08-05
ReIntroduce "Assign IPAddress" on Device Interface List View.

timizuoebideri1:
# Closes: #3636

- ReIntroduce the Add icon in the Device Interface List View table
- Change the "Assign an IP Address" View Table as follows: Add a checkbox to the table so that users may choose which IP addresses to assign to the interface.

Screenshot 2023-08-08 at 1 13 42 PM

# What's Changed

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/3636.changed
+++ b/changes/3636.changed
@@ -1,1 +1,1 @@
-ReIntroduce Assign IPAddress on Device Interface List View.
+Reintroduced "Assign IP Address" button to Device Interfaces list view.
--- a/changes/3636.changed
+++ b/changes/3636.changed
@@ -1,0 +1,1 @@
+ReIntroduce Assign IPAddress on Device Interface List View.
--- a/nautobot/dcim/tests/test_views.py
+++ b/nautobot/dcim/tests/test_views.py
@@ -1360,7 +1360,7 @@ class DeviceTestCase(ViewTestCases.PrimaryObjectViewTestCase):
         response_body = response.content.decode(response.charset)
         # Count the number of occurrences of "Add IP address" in the response_body
         count = response_body.count("Add IP address")
-        # Assert that "Add IP address" appears three times for each of the interfaces
+        # Assert that "Add IP address" appears for each of the three interfaces
         self.assertEqual(count, 3)
 
         with self.subTest("Assert Create and Assign IPAddress"):
--- a/nautobot/dcim/tests/test_views.py
+++ b/nautobot/dcim/tests/test_views.py
@@ -1345,7 +1345,7 @@ class DeviceTestCase(ViewTestCases.PrimaryObjectViewTestCase):
     @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
     def test_device_interfaces(self):
         device = Device.objects.first()
-        self.add_permissions("ipam.add_ipaddress")
+        self.add_permissions("ipam.add_ipaddress", "dcim.change_interface")
 
         intf_status = Status.objects.get_for_model(Interface).first()
         interfaces = (
@@ -1363,33 +1363,53 @@ class DeviceTestCase(ViewTestCases.PrimaryObjectViewTestCase):
         # Assert that "Add IP address" appears three times for each of the interfaces
         self.assertEqual(count, 3)
 
-        # Assest assigning ipadress to interface
-        namespace = Namespace.objects.create(name="ipam_test_views_ip_address_test")
-        prefix_status = Status.objects.get_for_model(Prefix).first()
-        Prefix.objects.get_or_create(
-            prefix="192.0.2.0/24",
-            defaults={"namespace": namespace, "status": prefix_status, "type": "network"},
-        )
-        form_data = {
-            "namespace": namespace.pk,
-            "address": IPNetwork("192.0.2.99/24"),
-            "tenant": None,
-            "status": Status.objects.get_for_model(IPAddress).first().pk,
-            "type": IPAddressTypeChoices.TYPE_DHCP,
-            "role": None,
-            "nat_inside": None,
-            "dns_name": None,
-            "description": None,
-            "tags": [],
-        }
+        with self.subTest("Assert Create and Assign IPAddress"):
+            # Assest assigning ipadress to interface
+            namespace = Namespace.objects.create(name="ipam_test_views_ip_address_test")
+            prefix_status = Status.objects.get_for_model(Prefix).first()
+            Prefix.objects.get_or_create(
+                prefix="192.0.2.0/24",
+                defaults={"namespace": namespace, "status": prefix_status, "type": "network"},
+            )
+            form_data = {
+                "namespace": namespace.pk,
+                "address": IPNetwork("192.0.2.99/24"),
+                "tenant": None,
+                "status": Status.objects.get_for_model(IPAddress).first().pk,
+                "type": IPAddressTypeChoices.TYPE_DHCP,
+                "role": None,
+                "nat_inside": None,
+                "dns_name": None,
+                "description": None,
+                "tags": [],
+            }
 
-        request = {
-            "path": reverse("ipam:ipaddress_add") + f"?interface={interfaces[0].id}",
-            "data": post_data(form_data),
-        }
-        self.assertHttpStatus(self.client.post(**request), 302)
-        interfaces[0].refresh_from_db()
-        self.assertEqual(str(interfaces[0].ip_addresses.all().first().address), "192.0.2.99/24")
+            request = {
+                "path": reverse("ipam:ipaddress_add") + f"?interface={interfaces[0].id}",
+                "data": post_data(form_data),
+            }
+            self.assertHttpStatus(self.client.post(**request), 302)
+            interfaces[0].refresh_from_db()
+            self.assertEqual(str(interfaces[0].ip_addresses.all().first().address), "192.0.2.99/24")
+
+        with self.subTest("Assertx Assign IPAddress"):
+            ipaddresses = [str(ipadress) for ipadress in IPAddress.objects.values_list("pk", flat=True)[:3]]
+            form_data = {"pk": ipaddresses}
+            device_list_url = reverse("dcim:device_interfaces", args=(interfaces[1].pk,))
+            url = reverse("ipam:ipaddress_assign") + f"?interface={interfaces[1].id}&return_url={device_list_url}"
+            request = {
+                "path": url,
+                "data": post_data(form_data),
+            }
+            response = self.client.post(**request)
+            self.assertHttpStatus(response, 302)
+            interfaces[1].refresh_from_db()
+            self.assertEqual(interfaces[1].ip_addresses.count(), 3)
+            interface_ips = [str(ip) for ip in interfaces[1].ip_addresses.values_list("pk", flat=True)]
+            self.assertEqual(
+                sorted(ipaddresses),
+                sorted(interface_ips),
+            )
 
     @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
     def test_device_rearports(self):
--- a/nautobot/dcim/tests/test_views.py
+++ b/nautobot/dcim/tests/test_views.py
@@ -1373,7 +1373,7 @@ class DeviceTestCase(ViewTestCases.PrimaryObjectViewTestCase):
             )
             form_data = {
                 "namespace": namespace.pk,
-                "address": IPNetwork("192.0.2.99/24"),
+                "address": "192.0.2.99/24",
                 "tenant": None,
                 "status": Status.objects.get_for_model(IPAddress).first().pk,
                 "type": IPAddressTypeChoices.TYPE_DHCP,
--- a/nautobot/dcim/tests/test_views.py
+++ b/nautobot/dcim/tests/test_views.py
@@ -1244,6 +1244,14 @@ class DeviceTestCase(ViewTestCases.PrimaryObjectViewTestCase):
             IPAddress.objects.create(address="3.3.3.3/32", namespace=namespace, status=cls.ipaddr_status),
         )
 
+        intf_status = Status.objects.get_for_model(Interface).first()
+
+        cls.interfaces = (
+            Interface.objects.create(device=devices[0], name="Interface 1", status=intf_status),
+            Interface.objects.create(device=devices[0], name="Interface 2", status=intf_status),
+            Interface.objects.create(device=devices[0], name="Interface 3", status=intf_status),
+        )
+
         for device, ipaddress in zip(devices, ipaddresses):
             RelationshipAssociation(
                 relationship=cls.relationships[0], source=device, destination=ipaddress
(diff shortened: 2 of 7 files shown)


changes/3636.changed
```diff
@@ -0,0 +1 @@
+ReIntroduce Assign IPAddress on Device Interface List View.
```
glennmatthews, line 1:
```suggestion
Reintroduced "Assign IP Address" button to Device Interfaces list view.
```

nautobot/ipam/views.py
```diff
@@ -752,6 +752,13 @@ def successful_post(self, request, obj, created, logger):
                     ),
                 )
 
+        # Add IpAddress to interface if interface is in query_params
+        if interface_id := request.GET.get("interface"):
```
glennmatthews, line 756: Seems wrong to me to be checking `request.GET` on a `successful_post` function - should this be part of `request.POST` or something instead?

nautobot/ipam/views.py
```diff
@@ -752,6 +752,13 @@ def successful_post(self, request, obj, created, logger):
...
+            try:
+                interface = Interface.objects.get(id=interface_id)
+                interface.ip_addresses.add(obj)
+            except Interface.DoesNotExist:
+                messages.warning("Interface with id `{interface_id}` no found.")
```
glennmatthews, line 761:
```suggestion
                messages.warning(f'Interface with id "{interface_id}" not found.')
```

nautobot/dcim/tests/test_views.py
```diff
@@ -1344,14 +1345,51 @@ def test_device_poweroutlets(self):
...
+            "type": IPAddressTypeChoices.TYPE_DHCP,
+            "role": None,
+            "nat_inside": None,
+            "dns_name": None,
+            "description": None,
+            "tags": [],
+        }
+
+        request = {
+            "path": reverse("ipam:ipaddress_add") + f"?interface={interfaces[0].id}",
```
glennmatthews, line 1387: Why is this a query parameter instead of part of the `form_data`?

nautobot/ipam/views.py
timizuoebideri1: The purpose of this issue is to reintroduce the "Assign IPAddress" button to the Device Interface List page's table.
Clicking the button takes you to the ip-address edit page, and the query param is `interface=uuiid>`.

nautobot/ipam/views.py
```diff
@@ -752,6 +752,20 @@ def successful_post(self, request, obj, created, logger):
                     ),
                 )
 
+        # Add IpAddress to interface if interface is in query_params
+        if interface_id := request.GET.get("interface"):
+            try:
+                interface = Interface.objects.get(id=interface_id)
+                interface.ip_addresses.add(obj)
+            except Interface.DoesNotExist:
+                messages.warning(f'Interface with id "{interface_id}" not found.')
```
glennmatthews, line 761:
```suggestion
                messages.warning(request, f'Interface with id "{interface_id}" not found.')
```

nautobot/dcim/tests/test_views.py
```diff
@@ -1344,14 +1345,71 @@ def test_device_poweroutlets(self):
...
+        )
 
         url = reverse("dcim:device_interfaces", kwargs={"pk": device.pk})
-        self.assertHttpStatus(self.client.get(url), 200)
+        response = self.client.get(url)
+        self.assertHttpStatus(response, 200)
+        response_body = response.content.decode(response.charset)
+        # Count the number of occurrences of "Add IP address" in the response_body
+        count = response_body.count("Add IP address")
+        # Assert that "Add IP address" appears three times for each of the interfaces
```
glennmatthews, line 1363:
```suggestion
        # Assert that "Add IP address" appears for each of the three interfaces
```

nautobot/dcim/tests/test_views.py
```diff
@@ -1344,14 +1345,71 @@ def test_device_poweroutlets(self):
...
+            # Assest assigning ipadress to interface
+            namespace = Namespace.objects.create(name="ipam_test_views_ip_address_test")
+            prefix_status = Status.objects.get_for_model(Prefix).first()
+            Prefix.objects.get_or_create(
+                prefix="192.0.2.0/24",
+                defaults={"namespace": namespace, "status": prefix_status, "type": "network"},
+            )
+            form_data = {
+                "namespace": namespace.pk,
+                "address": IPNetwork("192.0.2.99/24"),
```
glennmatthews, line 1376: Does `"address": "192.0.2.99/24"` not work? I'd expect the post data to just be a string.

nautobot/ipam/views.py
glennmatthews: Wonder if it'd be possible and/or more elegant to have the GET to the ip-address edit page render/populate the interface as a form field value (hidden or otherwise) so that here, in the POST handler, it'd be part of the form data instead of a separate query parameter?

nautobot/ipam/views.py
```diff
@@ -818,6 +840,14 @@ def post(self, request):
             },
         )
 
+    def assign_ipaddresses_to_interface(self, pks, request):
+        ip_addresses = IPAddress.objects.filter(pk__in=pks)
+        interface_model = Interface if "interface" in request.GET else VMInterface
+        interface_id = request.GET.get("interface") or request.GET.get("vminterface")
+        interface = interface_model.objects.get(id=interface_id)
```
glennmatthews, line 847: Do we need to apply queryset `restrict()` to `interface_model.objects` to prevent a user from assigning an IPAddress to an interface that they don't have edit permissions for? Or is that handled magically?

glennmatthews, line 847: Relatedly, probably need an except handler for for the case of a DoesNotExist exception?

nautobot/ipam/views.py
```diff
@@ -818,6 +840,14 @@ def post(self, request):
...
+        interface.ip_addresses.set(ip_addresses)
```
glennmatthews, line 848: Should this be `.add` instead of `.set`?

nautobot/ipam/views.py
timizuoebideri1: This case has been handled by the `dispatch` method

nautobot/ipam/views.py
timizuoebideri1: `set` since `ip_addresses` is an `list`.

glennmatthews: Right, but `set()` will replace any existing interface.ip_addresses won't it? `add()` would add to any existing ip_addresses instead of replacing them.

nautobot/ipam/views.py
timizuoebideri1: Would simply hiding the 'Assign IPAddress' button in the Device List View Table suffice if the user does not have permission to edit the interface?

nautobot/ipam/views.py
timizuoebideri1: it means this would
`if interface_id := request.GET.get("interface"):`
to
`if interface_id := request.POST.get("interface"):`
or 
illd modify the `IPAddressForm` to check for `interface`.

Any of the above would require the addition of more code, which makes me wonder if it's worth the effort to archive the same thing?

gsnider2195:
The title for the assign IP address view is showing up as `<title>Add a new  - Nautobot</title>` for me. Probably need to change it to say "Assign an IP address"?

gsnider2195:
When searching for IP Addresses to assign to interfaces, should the returned list be filtered to only return IP Addresses that are not already assigned to the interface?

nautobot/ipam/tables.py
```diff
@@ -507,12 +507,14 @@ class Meta(IPAddressTable.Meta):
 
 
 class IPAddressAssignTable(StatusTableMixin, BaseTable):
```
gsnider2195, line 509: This should also show the namespace column

nautobot/ipam/views.py
```diff
@@ -779,11 +796,20 @@ class IPAddressAssignView(generic.ObjectView):
 
     queryset = IPAddress.objects.all()
 
+    def _get_interface_type_and_model(self, request):
+        interface_model = Interface if "interface" in request.GET else VMInterface
+        interface_id = request.GET.get("interface") or request.GET.get("vminterface")
+        return interface_id, interface_model
```
jathanism, line 802: Since you have this method, can you please replace the calls on lines 715-716 and lines 767-768 with this?

nautobot/ipam/views.py
```diff
@@ -710,6 +710,16 @@ class IPAddressEditView(generic.ObjectEditView):
     model_form = forms.IPAddressForm
     template_name = "ipam/ipaddress_edit.html"
 
+    def dispatch(self, request, *args, **kwargs):
+        if "interface" in request.GET or "vminterface" in request.GET:
+            interface_model = Interface if "interface" in request.GET else VMInterface
+            interface_id = request.GET.get("interface") or request.GET.get("vminterface")
+            if not interface_model.objects.restrict(request.user, "change").filter(id=interface_id).exists():
```
jathanism, line 717: This block is nearly identical to lines 766-769. Can we write it once?

nautobot/ipam/views.py
```diff
@@ -752,6 +762,13 @@ def successful_post(self, request, obj, created, logger):
                     ),
                 )
 
+        # Add IpAddress to interface if interface is in query_params
+        if "interface" in request.GET or "vminterface" in request.GET:
+            interface_model = Interface if "interface" in request.GET else VMInterface
+            interface_id = request.GET.get("interface") or request.GET.get("vminterface")
+            interface = interface_model.objects.restrict(request.user, "change").get(id=interface_id)
```
jathanism, line 769: This block is nearly identical to lines 714-717. Can we write it once?

nautobot/ipam/views.py
```diff
@@ -779,11 +796,20 @@ class IPAddressAssignView(generic.ObjectView):
...
+        interface_id = request.GET.get("interface") or request.GET.get("vminterface")
+        return interface_id, interface_model
+
     def dispatch(self, request, *args, **kwargs):
         # Redirect user if an interface has not been provided
         if "interface" not in request.GET and "vminterface" not in request.GET:
             return redirect("ipam:ipaddress_add")
 
+        interface_id, interface_model = self._get_interface_type_and_model(request)
+        if not interface_model.objects.restrict(request.user, "change").filter(id=interface_id).exists():
```
jathanism, line 810: Another place that is almost identical to the blocks in 714-717 and 766-769. Can we please try to use this same code in all places?

nautobot/ipam/views.py
```diff
@@ -799,11 +825,15 @@ def get(self, request, *args, **kwargs):
...
     def post(self, request):
+        if pks := request.POST.getlist("pk"):
+            return self.assign_ipaddresses_to_interface(pks, request)
         form = forms.IPAddressAssignForm(request.POST)
         table = None
 
         if form.is_valid():
-            addresses = self.queryset.select_related("tenant")
+            interface_id, interface_model = self._get_interface_type_and_model(request)
+            obj = interface_model.objects.get(id=interface_id)
```
jathanism, line 835: Another place that is almost identical to the blocks in 714-717 and 766-769. Can we please try to use this same code in all places?

nautobot/ipam/views.py
```diff
@@ -818,6 +848,14 @@ def post(self, request):
             },
         )
 
+    def assign_ipaddresses_to_interface(self, pks, request):
+        ip_addresses = IPAddress.objects.filter(pk__in=pks)
+        interface_model = Interface if "interface" in request.GET else VMInterface
+        interface_id = request.GET.get("interface") or request.GET.get("vminterface")
+        interface = interface_model.objects.get(id=interface_id)
```
jathanism, line 855: Another place that is almost identical to the blocks in 714-717 and 766-769. Can we please try to use this same code in all places?

jathanism (changes requested):

gsnider2195:
Here's some weird behavior I'm noticing. Can you try to replicate?

- go to the assign IP address view
- don't add anything to the search box
- hit search to return all IP addresses
- click the select all checkbox to select all ip addresses
- change the paginator dropdown

All IP Addresses that were selected get assigned to the interface

gsnider2195:
It looks like the paginator on the assign view (after searching) isn't working?

nautobot/ipam/templates/ipam/ipaddress_assign.html
```diff
@@ -3,12 +3,20 @@
...
-    <form action="{% querystring request %}" method="post" class="form form-horizontal">
-        {% csrf_token %}
+    <form action="#" method="get" class="form form-horizontal">
         {% for field in form.hidden_fields %}
             {{ field }}
         {% endfor %}
+        {% for k, v_list in request.GET.lists %}
+            {% if k != 'q' %}
+                {% for v in v_list %}
+                    <input type="hidden" name="{{ k }}" value="{{ v }}" />
```
glennmatthews, line 16: Can we use more meaningful variable names than `k` and `v`? It's not clear to me what these actually are. I *think* `k` is going to be "pk" and `v_list` is going to be a list of PK values?

nautobot/ipam/templates/ipam/ipaddress_assign.html
```diff
@@ -39,10 +46,21 @@ <h3>Assign an IP Address</h3>
...
+                    <div class="pull-left noprint">
+                        {% if perms.dcim.change_interface %}
+                            <button type="submit" name="_edit" formaction="#" class="btn btn-primary btn-sm">
+                                <span class="mdi mdi-clipboard-tick" aria-hidden="true"></span> Assign Selected IPAddresses to Interface
+                            </button>
+                        {% endif %}
+                    </div>
+                </div>
+            </form>
+            {% include 'inc/paginator.html' with paginator=table.paginator page=table.page %}
```
glennmatthews, line 63: Why do we have/need a paginator if `disable_pagination` is True?

nautobot/ipam/views.py
```diff
@@ -815,36 +833,55 @@ def dispatch(self, request, *args, **kwargs):
...
+                "paginator_class": EnhancedPaginator,
+                "per_page": get_paginate_count(request),
+            }
+            RequestConfig(request, paginate).configure(table)
+            max_page_size = get_settings_or_config("MAX_PAGE_SIZE")
+            if max_page_size and paginate["per_page"] > max_page_size:
+                messages.warning(
+                    request,
+                    f'Requested "per_page" is too large. No more than {max_page_size} items may be displayed at a time.',
+                )
```
glennmatthews, line 861: Why are we paginating if we render the table with `disable_pagination=True`?

nautobot/ipam/views.py
```diff
@@ -815,36 +833,55 @@ def dispatch(self, request, *args, **kwargs):
...
-        table = None
+        interface, _ = retrieve_interface_or_vminterface_from_request(request)
 
-        if form.is_valid():
-            addresses = self.queryset.select_related("tenant")
-            # Limit to 100 results
-            addresses = filters.IPAddressFilterSet(request.POST, addresses).qs[:100]
-            table = tables.IPAddressAssignTable(addresses)
+        if pks := request.POST.getlist("pk"):
+            ip_addresses = IPAddress.objects.filter(pk__in=pks)
```
glennmatthews, line 877: Do we need to `restrict()` the IPAddress queryset here?

nautobot/ipam/templates/ipam/ipaddress_assign.html
gsnider2195: We had to do `{% include 'utilities/obj_table.html' with table_template='panel_table.html' disable_pagination=True %}` and then render the paginator ourselves later in the file. Otherwise we ended up with nested `<form>` elements which caused the paginator to submit the wrong form.

nautobot/ipam/views.py
gsnider2195: We had to do `{% include 'utilities/obj_table.html' with table_template='panel_table.html' disable_pagination=True %}` and then render the paginator ourselves later in the file. Otherwise we ended up with nested `<form>` elements which caused the paginator to submit the wrong form.

glennmatthews (approved):

jathanism (approved):

PR #4232, opened 2023-08-09
Fix #4149 - correct logic in RackForm

### LINKED ISSUE #4149: Unable to re-name Racks
### Environment
* Nautobot version (Docker tag too if applicable): v1.5.23
* Python version: 3.10
* Middleware(s):

### Steps to Reproduce
1. Go to dcim/racks/
2. select any rack to edit
3. change the name of the rack

### Expected Behavior
Page would reload and name would be reflected. 

### Observed Behavior
When submitting an error shows up below the `name` field:
```
Device with `name` in ['2.5 960GB SSD (PCIE Gen3) Samsung', '40GBE Active Fiber Cable - Aqua - 03M', '40GBE Active Fiber Cable - Aqua - 05M', ... '2.5 960GB SSD  Intel'] and site=A52Plus already exists.
```

From [Slack Conversation with Glenn M.](https://networktocode.slack.com/archives/C01NWPK6WHL/p1689953425513139?thread_ts=1689951178.929749&cid=C01NWPK6WHL) it is suspected that #4025 may have introduced this.

glennmatthews:
# Closes: #4149
# What's Changed

- Logic in `RackForm` introduced in #4025 was incorrect - it would cause form validation to fail if any Device in the target site shared a non-globally-unique name with any other device in the database, which was not the intent.
  - Fixed the logic so that it will only fail if a device *in the rack being edited* has a name *and tenant* conflict with an existing device *in the target site*
  - Furthermore, we only need to run this check at all if the form is changing a Rack's `site` attribute, so I added a check for that as well.
- Added a check for #4149 in the `dcim.test_forms.RackTestCase` and verified that it failed before my changes and passes now.

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/forms.py
+++ b/nautobot/dcim/forms.py
@@ -696,7 +696,7 @@ class RackForm(LocatableModelFormMixin, NautobotModelForm, TenancyForm):
         cleaned_data = self.cleaned_data
         site = cleaned_data.get("site")
 
-        if self.instance and site != self.instance.site:
+        if self.instance and self.instance.present_in_database and site != self.instance.site:
             # If the site is changed, the rack post save signal attempts to update the rack devices,
             # which may result in an Exception if the updated devices conflict with existing devices at this site.
             # To avoid an unhandled exception in the signal, check for this scenario here.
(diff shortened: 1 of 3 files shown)


nautobot/dcim/forms.py
```diff
@@ -696,21 +696,21 @@ def clean(self):
         cleaned_data = self.cleaned_data
         site = cleaned_data.get("site")
 
-        if self.instance:
+        if self.instance and site != self.instance.site:
```
HanlinMiao, line 699: This is causing issues in the unit tests, perhaps we need to check if `self.instance.site` exists as well?

glennmatthews: Should be fixed now, was missing a `self.instance.present_in_database` check

HanlinMiao (approved):

Issue #4197, opened 2023-08-02
2.0.0b2: namespaces legacy list view reports "Missing views for action(s) import"


glennmatthews:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0b2

### Steps to Reproduce
1. Navigate to legacy UI view for `/ipam/namespaces/`

### Expected Behavior

No UI errors

### Observed Behavior

![image](https://github.com/nautobot/nautobot/assets/5603551/5fa7a82c-d846-4f4e-8716-b5f48f44d78e)

We probably should just go ahead and implement bulk-import for namespaces as the fix for this issue, it shouldn't be particularly complex.

bryanculver:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Nautobot-Core-Engineering-64a875e74842500d9e80e23e/issues/nautobot/nautobot/4197?planning-poker) with Zenhub @glennmatthews @gsnider2195 @HanlinMiao @jathanism @timizuoebideri1

PR #4263, opened 2023-08-15
Added Bulk Import View for Namespace.

timizuoebideri1:
# Closes: #4197 
# What's Changed

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/ipam/tests/test_views.py
+++ b/nautobot/ipam/tests/test_views.py
@@ -40,6 +40,32 @@ from nautobot.users.models import ObjectPermission
 from nautobot.virtualization.models import Cluster, ClusterType, VirtualMachine
 
 
+class NamespaceTestCase(
+    ViewTestCases.GetObjectViewTestCase,
+    ViewTestCases.GetObjectChangelogViewTestCase,
+    ViewTestCases.GetObjectNotesViewTestCase,
+    ViewTestCases.CreateObjectViewTestCase,
+    ViewTestCases.EditObjectViewTestCase,
+    ViewTestCases.DeleteObjectViewTestCase,
+    ViewTestCases.ListObjectsViewTestCase,
+    ViewTestCases.BulkImportObjectsViewTestCase,
+):
+    model = Namespace
+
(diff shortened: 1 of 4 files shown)


nautobot/ipam/views.py
```diff
@@ -90,6 +90,11 @@ def get_extra_context(self, request, instance):
         return context
 
 
+class NamespaceImportView(generic.BulkImportView):
```
glennmatthews, line 93: why isn't this done as part of the `NamespaceUIViewSet`?

nautobot/ipam/tests/test_views.py
```diff
@@ -40,6 +40,32 @@
...
 
+class NamespaceTestCase(
+    ViewTestCases.GetObjectViewTestCase,
+    ViewTestCases.GetObjectChangelogViewTestCase,
+    ViewTestCases.GetObjectNotesViewTestCase,
+    ViewTestCases.CreateObjectViewTestCase,
+    ViewTestCases.EditObjectViewTestCase,
+    ViewTestCases.DeleteObjectViewTestCase,
+    ViewTestCases.ListObjectsViewTestCase,
+    ViewTestCases.BulkImportObjectsViewTestCase,
```
glennmatthews, line 51: Can we use `PrimaryObjectViewTestCase` instead of these mixins? Or do `BulkEdit` and `BulkDelete` tests not apply to Namespaces for some reason? (And if that's the case, should we change that?)

timizuoebideri1: Namespace do not currently support BulkEdit/BulkDelete, and fixing that would be out of scope for this issue

glennmatthews: Ok, let's get a follow-up issue opened then.

timizuoebideri1: https://github.com/nautobot/nautobot/issues/4272

jathanism (approved):

glennmatthews (approved):

Issue #4294, opened 2023-08-19
Nautobot 2.0.0rc1 Docs: References NPM Where It Should Be Node


jvanderaa:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0rc1
* Python version: N/A
* Database platform, version: N/A
* Middleware(s):

### Steps to Reproduce
1. Review the docs around installing Node
2. Install docs indicate to install NPM, which should be Node
3.

### Expected Behavior

Expected to request install of Node

### Observed Behavior

Requests to install NPM

gsnider2195:
Are you referencing the ubuntu install guide [here](https://docs.nautobot.com/projects/core/en/next/user-guide/administration/installation/ubuntu/?h=npm#install-nodejs)?

gsnider2195:
`nodejs` is required for installation of the `npm` package on ubuntu lunar. What version are you using? https://packages.ubuntu.com/lunar/npm

jvanderaa:
When I installed for Ubuntu 22.04 just installing NPM I didn't get Node. I also had to remove something as well, so I may need to get that documented.

gsnider2195:
It looks like `apt install npm` doesn't work on 22.04. You have to do `apt install nodejs` which installs npm as well.

gsnider2195:
Anyway, sorry for the trouble. I thought I tested on 22.04!

jvanderaa:
No worries here, I'm the one that read later than `22.10` as `22.04` earlier. So this may work out, and it moves the node install to the very beginning to do all system stuff at the same spot.

jathanism:
Sounds like there is some confusion here. Node is not required for production deployments, only NPM. Node is required for development only, and even then, only for dynamic rendering.

We should at least revise the documentation to be explicit about the required Node.js version, which is 18 LTS. We have had issues with "later" versions of Node, so until we can certify them, let's keep it crisp.

jdrew82:
I wanted to mention also that the upgrade docs [here](https://docs.nautobot.com/projects/core/en/stable/user-guide/administration/upgrading/from-v1/upgrading-from-nautobot-v1/) also makes no mention of needing NodeJS/npm installed for the upgrade.

PR #4287, opened 2023-08-18
Collapses Docs and Updates Dev Environment for Docker Compose Plugin

jvanderaa:
# Closes: #2821
# What's Changed

- Updates tasks.py for Docker Compose plugin since system wasn't able to work with a fresh install
- Creates an `install_system.md` file to handle all system installation steps that were in two divergent files previously
  - Removes `centos.md` file that is now all in `install_system.md`
  - Removes `ubuntu.md` file that is now all in `install_system.md`
- Adds navigation footer to move along the install steps, may at future time remove in document navigation 

The last bullet above is a docs design decision. I think that the Navigation footer will also help the 2.0 migration docs that were seen.

https://github.com/nautobot/nautobot/assets/22228848/d7faec50-b2fc-4bff-95e2-fe565e3437c2

# TODO

- [x] Explanation of Change(s)
- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [x] Attached Screenshots, Payload Example
- [x] Unit, Integration Tests
- [x] Documentation Updates (when adding/changing features)
- [x] Example Plugin Updates (when adding/changing features)
- [x] Outline Remaining Work, Constraints from Design

PR #4295, opened 2023-08-19
Update 2 docs node

jvanderaa:
# Closes: #4294 
# What's Changed

- Added in doc updates from PR to 1.6
- Updated docs to put Node install in with system install instructions
- Added verification docs

# TODO

- [ ] Unit, Integration Tests


bryanculver:
@jvanderaa It seems some of this should be resolved with merging `develop` up into this branch.

tasks.py
```diff
@@ -146,15 +146,15 @@ def print_command(command, env=None):
...
 
     Args:
         context (obj): Used to run specific commands
-        command (str): Command string to append to the "docker-compose ..." command, such as "build", "up", etc.
+        command (str): Command string to append to the "docker compose ..." command, such as "build", "up", etc.
         **kwargs: Passed through to the context.run() call.
     """
     compose_command_tokens = [
-        "docker-compose",
+        "docker compose",
```
gsnider2195, line 157: This is a significant change. We should probably discuss

bryanculver: It's been opened many times. It's gone around and around based on how you install Docker and Docker compose.

jvanderaa: Well, `docker-compose` CLI is EOL as of July - https://docs.docker.com/compose/release-notes/

So it's time to move on to this.

jvanderaa:
> @jvanderaa It seems some of this should be resolved with merging `develop` up into this branch.

Yes I agree. Is that something that is upcoming?

tasks.py
glennmatthews: I see you're making the same changes in #4287 so whatever we decide let's make sure we are consistent.

changes/2821.changed
```diff
@@ -0,0 +1,4 @@
+Adds navigation.footer to mkdocs features to enable navigation in the footer.
+Updates installation documentation into a single path, using tabs for MySQL/PostgeSQL & Ubuntu/Debian/RHEL.
+Removes centos.md and ubuntu.md as these are collapsed into single `install_system.md` file.
+Updates docker-compose standalone to docker compose plugin to assist in development environment.
```
glennmatthews, line 4: Was it intentional to include all of #4287 in this PR? Should we close that PR in favor of this one, or can they be separated for individual review?

jvanderaa: If I can move to using #4287 instead, I am good. The update that is needed for this is basically on top of #4287 due to the nature of where the requirement to do the install of NodeJS to. I will gladly re-submit a new PR if we are good that 4287 is set for the release.

jvanderaa: Which is probably better to get that merged in and then at this point have me submit new work on top of that. Since the linting failures are solved there.

tasks.py
jathanism: We need to assert a base Docker version that includes the `docker compose` sub-command and make sure that people are aware of what that is. 

The linked article about Compose V1 being EOL as of July 2023 references Docker Desktop. But what about Linux native Docker users that don't use Docker DesktoP? Does the Docker Engine command also include `compose`? We need these answers and need to make sure they are documented.

nautobot/docs/user-guide/administration/installation/nautobot.md
```diff
@@ -94,30 +94,29 @@ Example output:
 
 Because the `nautobot` user was created with `NAUTOBOT_ROOT` set as its home directory and we had you set the shell to `/bin/bash`, the binary path `$NAUTOBOT_ROOT/bin` is automatically added to the beginning of the `$PATH` environment variable:
 
-In Ubuntu 20.04:
+=== "Ubuntu 20.04"
```
jathanism, line 97: What's this `===` syntax? I haven't seen that before in Markdown...

tasks.py
timizuoebideri1: According to this docs https://github.com/docker/compose, I don't think `compose` is included by default for linux OS.

> ### Where to get Docker Compose

    #### Windows and macOS
    
    Docker Compose is included in [Docker Desktop](https://www.docker.com/products/docker-desktop) for Windows and macOS.
    
    #### Linux
    
    You can download Docker Compose binaries from the [release page](https://github.com/docker/compose/releases) on this repository.
    
    Rename the relevant binary for your OS to docker-compose and copy it to $HOME/.docker/cli-plugins
    
    Or copy it into one of these folders to install it system-wide:
    
    /usr/local/lib/docker/cli-plugins OR /usr/local/libexec/docker/cli-plugins
    /usr/lib/docker/cli-plugins OR /usr/libexec/docker/cli-plugins
    (might require making the downloaded file executable with chmod +x)

nautobot/docs/user-guide/administration/installation/nautobot.md
jvanderaa: This is what enables the tabbed interface within the mkdocs theme. So each `=== "Section Title"` is the section title. Everything indented at that point falls under the section until either another `=== "Section 2 Title"` is found, or it is de-indented, which then is back outside of the tab.

jvanderaa: An example with it is here of the output - https://docs.nautobot.com/projects/core/en/stable/installation/http-server/#obtain-an-ssl-certificate

tasks.py
gsnider2195: The [docker ubuntu install docs](https://docs.docker.com/engine/install/ubuntu/) do instruct you to install the docker-compose-plugin package:

`sudo apt-get install docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin`

Also, the latest deb package of `docker-ce` depends on `docker-ce-cli` which recommends `docker-compose-plugin` so a basic `apt install docker-ce` *should* install the plugin. It doesn't look like this would cause any problems for folks developing for nautobot.

glennmatthews:
Needs re-review after #4287 is merged as a pre-requisite.

glennmatthews:
#4287 has been merged to `develop`; it may be easiest to pull a new branch with the changes specific to this PR rather than attempting to merge or rebase the existing branch for this PR?

Issue #4300, opened 2023-08-21
nautobot-dev:2.0.0-rc.1 image leaves the USER as nautobot


jmcgill298:
### Environment
* Nautobot version (Docker tag too if applicable):
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. pull nautobot-dev:2.0.0-rc.1 image from available source
2. build container with that image
3. try to have that container write to filesystem (e.g. makemigrations)

### Expected Behavior
expected migration file to exist

### Observed Behavior
container failed due to system permissions (USER nautobot does not have permission to write to my filesystem)

jathanism:
As I understand it, this is specific to the `final-dev` container build and the `USER` needs to be set to `root` at the end in order to alleviate this.

glennmatthews:
#4082 introduced this issue; need to review carefully to determine which parts of that change need to be revised/rolled back.

jathanism:
Fixed by #4302

PR #4302, opened 2023-08-21
Fixed a permissions issue when using `final-dev` containers 

jathanism:
# Closes: #4300 
# What's Changed

This addresses potential filesystem permissions issues on mounted volumes for Nautobot App developers using Linux native Docker workflows. 

A `USER root` statement is added to the `final-dev` build stage before the `EXPOSE  8080` command identically to how it was solved for the `dev stage.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/4300.fixed
+++ b/changes/4300.fixed
@@ -1,1 +1,1 @@
-Fixed a permissions issue when using `final-dev` containers by  switching to root user before exposing port and entrypoint.
+Fixed a permission issue when using `final-dev` containers by switching to root user before exposing port and entrypoint.
--- a/changes/4300.fixed
+++ b/changes/4300.fixed
@@ -1,0 +1,1 @@
+Fixed a permissions issue when using `final-dev` containers by  switching to root user before exposing port and entrypoint.
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -245,6 +245,10 @@ ENV NAUTOBOT_INSTALLATION_METRICS_ENABLED=false
 RUN nautobot-server init && \
     nautobot-server build_ui
 
+# switch to root user for final-dev stage: https://github.com/nautobot/nautobot/issues/4300
+# hadolint ignore=DL3002
+USER root
+
 # Run Nautobot development server by default
 EXPOSE 8080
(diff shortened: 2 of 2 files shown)


changes/4300.fixed
```diff
@@ -0,0 +1 @@
+Fixed a permissions issue when using `final-dev` containers by  switching to root user before exposing port and entrypoint.
```
HanlinMiao, line 1:
```suggestion
Fixed a permission issue when using `final-dev` containers by switching to root user before exposing port and entrypoint.
```

HanlinMiao (approved):

gsnider2195 (approved):

timizuoebideri1 (approved):

docker/Dockerfile
```diff
@@ -245,6 +245,10 @@ ENV NAUTOBOT_INSTALLATION_METRICS_ENABLED=false
 RUN nautobot-server init && \
     nautobot-server build_ui
 
+# switch to root user for final-dev stage: https://github.com/nautobot/nautobot/issues/4300
+# hadolint ignore=DL3002
+USER root
+
```
bryanculver, line 251: Are there any concerns/issues with these commands here: [https://github.com/nautobot/nautobot-plugin-golden-config/blob/develop/development/Dockerfile#L72-L76 being run](https://github.com/nautobot/nautobot-plugin-golden-config/blob/9cc0dbbb5d54c74dce539cc7ae986b1e0dd6b2b6/development/Dockerfile#L72-L76) being run as a different user than the `pip` command installing Nautobot and other packages above?

gsnider2195: Looks like we'll be fine. We're using `ENV VIRTUAL_ENV="/opt/nautobot"` to tell poetry and pip to install python packages to `/opt/nautobot` so even as root when you run a `pip install xyz` in the `final-dev` image, it installs to `/opt/nautobot`

jathanism: @bryanculver This was something we explicitly tested as well FYI, which was building the `Dockerfile` you linked from the revised `final-dev` container. It was a win.

Issue #4293, opened 2023-08-19
Nautobot 2 - NewUI - ChangeLog Does Not Load


jvanderaa:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0rc1
* Python version: 3.10.12
* Database platform, version: PostgreSQL 14
* Middleware(s):

### Steps to Reproduce
1. Create an object
2. Navigate to Governance -> Change Log
3. Click on a change log (only one in this instance)
4. Get a blank screen

### Expected Behavior

To either provide a new UI view of the change or to load the original UI with the view

### Observed Behavior

Blank Screen, the blank screen continues on the back button until you refresh
[Screencast from 2023-08-19 08-10-30.webm](https://github.com/nautobot/nautobot/assets/22228848/5b43275c-f4e0-4fca-9c1e-5ee853942261)

timizuoebideri1:
This error is brought on by the client's outdated cache, which means that you probably viewed this page before the changelogs were added.

glennmatthews:
Do we need to add some cache-busting to the files?

jvanderaa:
How long is the typical client cache? There was some minor navigation that happened between looking at it?

timizuoebideri1:
I guess so, 
1. We could implement category-based caching with different expiration times. For models like changelogs that are frequently updated, use a short cache of 30 seconds or no caching at all. For other models that change less often, we can set a longer cache time like 1 minute.
2. Alternatively, we could set a general cache expiration time that applies across all categories. However, this may not be as optimized as using category-specific times since some content changes more frequently than others.

glennmatthews:
Wait, hang on. I thought you were talking about the client/browser file cache, not the data itself.

timizuoebideri1:
> Wait, hang on. I thought you were talking about the client/browser file cache, not the data itself.

With redux-persist(what we use), every api you call and its response are stored in the local storage, and if you call that same api again that data is loaded from the local storage.

jathanism:
@jvanderaa Can you please try to clear the local storage in your browser and try again? 

https://intercom.help/scoutpad/en/articles/3478364-how-to-clear-local-storage-of-web-browser

Per the reply by @timizuoebideri1 we had issues in earlier builds with the internal application state not properly being flushed and sounds like you may have encountered that. Thanks in advance.

jvanderaa:
@jathanism do I need to install a new version?

jvanderaa:
[...]

Issue #4307, opened 2023-08-23
Nautobot UI 2.0: Notes and Change Log Tables are not hyperlinked correctly in ObjectDetailView.


HanlinMiao:
### Environment
* Nautobot version (Docker tag too if applicable): v2.0.0-rc
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Go to any object's detail view
2. Go to the Notes or ChangeLogs Tab of the object.
3. Click on any item in the first column of the table.

### Expected Behavior
The page redirects the user to ChangeLog/Notes Detail View with the correct URL (extras/object-changes/<uuid>/ or extras/notes/<uuid>)

### Observed Behavior
Blank page with the incorrect URL (Object Detail URL + ChangeLog/Note's UUID) rendered.

PR #4312, opened 2023-08-23
Fixed the issue of ObjectChange Detail View not rendering in the new UI.

HanlinMiao:
# Closes: #4293, #4307 
# What's Changed

1. Fixed the issue of ObjectChange Detail View not rendering in the new UI.
2. Fixed incorrect hyperlinks that exist in the Notes and ChangeLog Tables in the ObjectDetailView of new UI.

https://github.com/nautobot/nautobot/assets/46973263/6d4c2708-a576-443d-95eb-a3eaf943d3f9

# TODO

- [ ] Unit, Integration Tests


nautobot/ui/src/views/generic/ObjectRetrieve.js
```diff
@@ -70,7 +70,7 @@ export default function ObjectRetrieve({ api_url }) {
     };
 
     return (
-        <GenericView objectData={data}>
+        <GenericView objectData={data} key={window.location.pathname}>
```
HanlinMiao, line 73: This key value is added so that when the user clicks on the link of the changelog/note entry from `ObjectRetrieveView`. The component is re-rendered and the very first tab will become active. Without it, the URL will redirect but the page won't re-render, we will still remain on the Changelog tab/Note tab which is not ideal.

nautobot/ui/src/components/ObjectTable/ObjectTable.js
```diff
@@ -10,10 +10,6 @@ import {
 import ObjectTableItem from "./ObjectTableItem";
 
 function getTableItemLink(idx, obj) {
-    if (idx === 0) {
-        // TODO: ui-schema should be providing the name of the field used to describe the object (name)
-        return window.location.pathname + obj.id + "/";
```
HanlinMiao, line 15: Appending the changelog/note UUID directly to the current `window.location.pathname` for example, `http://localhost:3000/dcim/locations/659efd0a-0778-496e-b9b7-c3d56c8fbb93/` does not take us to the correct URL. We should let this fall through and let line 21 handle all cases of URL rendering.

nautobot/extras/navigation.py
```diff
@@ -701,6 +701,15 @@
                         link="extras:customlink_list",
                         permissions=["extras.view_customlink"],
                     ),
+                    NavItem(
+                        name="Notes",
+                        weight=500,
+                        # TODO figure out what link we should put here
+                        # This NavItem is required for the breadCrumb on NotesDetailView to render.
```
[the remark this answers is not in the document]
HanlinMiao: Yes indeed! However, the `link=` here is wrong, because `extras/notes/` which is a list view for Note does not exist.

bryanculver (approved):

Issue #4216, opened 2023-08-04
Slow page rendering with thousands of Tags present on objects


glennmatthews:
## Environment
* Nautobot version (Docker tag too if applicable): 1.6.0rc1, 2.0.0b2
* Python version:
* Database platform, version:
* Middleware(s): django-debug-toolbar (optional)

### Steps to Reproduce
1. Configure 2000+ Tags that apply to a particular model (or all models)
2. Create instances of that model such that each tag is used one or more times
3. Navigate to the list view for that model

nbshell example:

```python
>>> for i in range(1, 2001):
...     t = Tag.objects.create(name=f"Tag {i}", slug=f"tag-{i}")
...     t.content_types.add(ContentType.objects.get_for_model(Prefix))
... 
>>> for i in range(1, 2001):
...     p = Prefix.objects.create(prefix=f"2001:{i}::/64")
...     p.tags.add(Tag.objects.get(name=f"Tag {i}"))
... 
>>>
```

### Expected Behavior

Page to load relatively quickly regardless of the number of tags.

### Observed Behavior

Page loads substantially slower due to the large number of tags.

Investigation suggests that this is due to server-side rendering of the list of all possible tags as part of `TagFilterField`. When django-debug-toolbar is enabled, it can be seen that thousands of copies of the `selectwithdisabled_option.html` template have been included in the page; DJDT's memory consumption can grow into the gigabytes as a result due to it caching the complete template rendering context for each such copy.

List view of Prefix with page size of 50 and no Tags applied:

![Screenshot 2023-08-04 at 10 40 27 AM](https://github.com/nautobot/nautobot/assets/5603551/21bb51f5-4a22-40e9-b69e-653e76bfb6ba)
![Screenshot 2023-08-04 at 10 40 32 AM](https://github.com/nautobot/nautobot/assets/5603551/144dec68-707f-4895-a0d3-7a7084d844d2)

Same list view with 2000 Tags in use:

![Screenshot 2023-08-04 at 10 38 16 AM](https://github.com/nautobot/nautobot/assets/5603551/d2fce90f-0831-47a7-a483-2c578f96b55e)
![Screenshot 2023-08-04 at 10 38 20 AM](https://github.com/nautobot/nautobot/assets/5603551/6ac7c6bc-60e9-4e3d-bd84-9acaf3ee7f9e)

[...]

PR #4322, opened 2023-08-24
Fix rendering of TagFilterField with large number of tags (#4216)

glennmatthews:
# Closes: #4216
# What's Changed

- Change `TagFilterField` to a `DynamicModelMultipleChoiceField` so that it retrieves and renders tags 50-at-a-time from the API like most other filter fields, instead of statically rendering **all** applicable tags during page render time.
- Add integration tests to confirm this behavior.

*This will be a minor pain to merge up into `next` due to code reorganization. Sorry!*

image

image

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/utilities/forms/fields.py
+++ b/nautobot/utilities/forms/fields.py
@@ -10,7 +10,7 @@ from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
 from django.contrib.postgres.forms import SimpleArrayField
 from django.core.exceptions import MultipleObjectsReturned, ObjectDoesNotExist, ValidationError
-from django.db.models import Count, Q
+from django.db.models import Q
 from django.forms.fields import BoundField, JSONField as _JSONField, InvalidJSONInput
 from django.urls import reverse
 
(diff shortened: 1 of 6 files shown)


nautobot/utilities/forms/fields.py
```diff
@@ -788,3 +768,25 @@ def _check_values(self, values):
...
+
+class TagFilterField(DynamicModelMultipleChoiceField):
+    """
+    A filter field for the tags of a model. Only the tags used by a model are displayed.
+
+    :param model: The model of the filter
+    """
+
+    def __init__(self, model, *args, query_params=None, queryset=None, **kwargs):
+        queryset = queryset or model.tags.all()
```
gsnider2195, line 781: I'm not sure if there's a use case where this would be encountered but if `queryset` is an empty queryset (`Tag.objects.filter(name="nonexistent")`), it would fall back to `model.tags.all()`. Is that the behavior we want or should it be `if queryset is None:`?

glennmatthews: `if None` would be better. Good catch!

glennmatthews: also saves us evaluating the queryset at init time.

timizuoebideri1 (approved):

timizuoebideri1 (approved):

gsnider2195 (approved):

Issue #4201, opened 2023-08-02
InterfaceRedundancyGroup Model Documentation


HanlinMiao:
### Proposed Changes
We need to document `InterfaceRedundancyGroup` model and its capabilities at `docs/models/dcim/interfaceredundancygroup.md`

### Justification
All the other dcim models have their own respective documentation.

PR #4323, opened 2023-08-24
Added docs for InterfaceRedundancyGroup

gsnider2195:
# Closes: #4201 
# What's Changed

image
image

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/docs/models/dcim/interfaceredundancygroup.md
+++ b/nautobot/docs/models/dcim/interfaceredundancygroup.md
@@ -8,7 +8,7 @@ Interface Redundancy Groups must be created before interfaces can be assigned to
 
 ## Required Fields
 
-When adding Interfaces to the Interface Redundancy Group, a priority integer value must be set. This value will depend on the redundancy protocol being used. For example, HSRP uses a priority value between 1 and 255.
+When adding Interfaces to the Interface Redundancy Group, a priority integer value must be set for each interface in the group. This value will depend on the redundancy protocol being used. For example, HSRP uses a priority value between 1 and 255.
 
 ## Optional Fields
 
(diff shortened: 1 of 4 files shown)


nautobot/docs/models/dcim/interfaceredundancygroup.md
```diff
@@ -0,0 +1,21 @@
+# Interface Redundancy Groups
+
++++ 1.6.0
+
+Interface Redundancy Groups represent groups of interfaces that share a single virtual address. This model is intended to represent redundancy protocols such as HSRP or VRRP that allow multiple devices to provide a fault-tolerant default gateway for a network.
```
itdependsnetworks, line 5: I am lacking the language, but I think we should be clear that it is for any interface that is considered redundant, regardless if there is a protocol tying it together. As an example, my two uplinks may just run bgp, but I consider it in my interface redundancy strategy.

timizuoebideri1 (approved):

nautobot/docs/models/dcim/interfaceredundancygroup.md
```diff
@@ -0,0 +1,21 @@
...
+
+Interface Redundancy Groups must be created before interfaces can be assigned to the group.
+
+## Required Fields
+
+When adding Interfaces to the Interface Redundancy Group, a priority integer value must be set. This value will depend on the redundancy protocol being used. For example, HSRP uses a priority value between 1 and 255.
```
glennmatthews, line 11: Perhaps?

```suggestion
When adding Interfaces to the Interface Redundancy Group, a priority integer value must be set for each interface in the group. This value will depend on the redundancy protocol being used. For example, HSRP uses a priority value between 1 and 255.
```

HanlinMiao (approved):

nautobot/docs/models/dcim/interfaceredundancygroup.md
gsnider2195: Thanks @itdependsnetworks 

I added a note to call that out

glennmatthews (approved):

Issue #4327, opened 2023-08-25
Webhook causes traceback with M2M changes after ObjectChange has been deleted/pruned


joewesch:
### Environment
* Nautobot version (Docker tag too if applicable): 1.6.2b1
* Python version: 3.8.17
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Have an object with no ObjectChange entries either by manually deleting them or they get pruned via CHANGELOG_RETENTION
2. Have a Webhook with `Type update` enabled that includes the content type of the object
3. Modify a M2M relationship on that object (such as tags) via a Job so that it triggers the `_handle_changed_object` signal.

### Expected Behavior
I expect the tag to be set without any errors

### Observed Behavior
_Replicated via ORM with `web_request_context` to show error_
```
root@590eb78745e9:/source# nautobot-server nbshell
### Nautobot interactive shell (590eb78745e9)
### Python 3.8.17 | Django 3.2.20 | Nautobot 1.6.2b1
### lsmodels() will show available models. Use help(<model>) for more info.
>>> from nautobot.extras.context_managers import web_request_context
>>> user = User.objects.get(username="admin")
>>>
>>> with web_request_context(user, context_detail="manual-fix"):
...     device = Device.objects.first()
...     tag = Tag.objects.first()
...     device.tags.add(tag)
...
Traceback (most recent call last):
  File "<console>", line 4, in <module>
  File "/usr/local/lib/python3.8/site-packages/taggit/utils.py", line 124, in inner
    return func(self, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/taggit/managers.py", line 186, in add
    signals.m2m_changed.send(
  File "/usr/local/lib/python3.8/site-packages/django/dispatch/dispatcher.py", line 180, in send
    return [
  File "/usr/local/lib/python3.8/site-packages/django/dispatch/dispatcher.py", line 181, in <listcomp>
    (receiver, receiver(signal=self, sender=sender, **named))
  File "/source/nautobot/extras/signals.py", line 113, in _handle_changed_object
    enqueue_webhooks(instance, change_context_state.get().get_user(), change_context_state.get().change_id, action)
  File "/source/nautobot/extras/webhooks.py", line 41, in enqueue_webhooks
    snapshots = most_recent_change.get_snapshots()
AttributeError: 'NoneType' object has no attribute 'get_snapshots'
```

Issue #3321, opened 2023-02-15
ObjectChange Created Regardless of Change


jdrew82:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.10
* Python version: 3.7
* Database platform, version: PostgreSQL/MySQL
* Middleware(s): Golden Config

### Steps to Reproduce
1. Stand up instance of Nautobot.
2. Fill instance with devices and associated data.
3. Edit a Device and save without changing anything.
4. Review changelog for Device and see another change log entry was created with no diff filled in as nothing changed on the Device object.

### Expected Behavior
Expect no change log entry to be created as nothing changed.

### Observed Behavior
It appears that regardless of whether any fields were changed on an object an ObjectChange, ie changelog, entry is created upon object save. This means that any objects that are saved, even if nothing changes, has an entry. This leads to a large number of garbage logs that are unnecessary in the database. This becomes even more of a concern with Apps like Golden Config where each compliance feature has a changelog and an entry is created everytime a compliance Job is ran. This means that if there are 30 features to be evaluated for 1000 Devices, even if nothing has changed for these Devices there are now 30k more change log entries in the database.

bryanculver:
@jdrew82 Does the last updated field get updated though...

Kircheneer:
Changing this could be considered a breaking change as currently this also triggers job hooks.

jdrew82:
@bryanculver It does appear that the `last_updated` field is being updated to be close to the `time` field for the ObjectChange object.

lampwins:
So the question is, do we want the `last_updated` to be updated with no other object data changed? Sure, that's what django does in this case, but is it really what we want?

jdrew82:
IMO, if no other data about the object has changed, ie no fields have changed, then there's no reason for a changelog to be created cause nothing about it changed. The object wasn't actually "updated" in this cause so the `last_updated` being changed to the save data is a bit of a misnomer cause there was no update.

lampwins:
@nautobot/core just to confirm, with the recent work in this area, this issue still exists, right?

glennmatthews:
To the best of my knowledge, yes.

PR #4328, opened 2023-08-25
Fixes trying to send a webhook when there isn't a previous change

joewesch:
# Closes: #4327
# What's Changed
When you are modifying a M2M relationship on an object it will trigger any applicable webhooks for that object. Before it sends that webhook, it gathers the last ObjectChange for it to add to the webhook payload in the form of a `snapshots` dict.

In the case that all of the ObjectChange records for that object have been removed (either by manually deleting them to save DB space or they were pruned automatically via the `CHANGELOG_RETENTION` setting) there will be no previous records. This change allows the webhook to still trigger, but with an empty `snapshots` value.


glennmatthews:
Change looks reasonable to me. Would be nice to add a unit test for this scenario, probably in `nautobot/extras/tests/test_webhooks.py`.

nautobot/extras/tests/test_webhooks.py
```diff
@@ -261,3 +268,54 @@ class FakeResponse:
...
+        location.save()
+
+        tag = Tag.objects.create(name="Tag 1")
+
+        all_changes = get_changes_for_model(location)
+        # Mimicking when all changes have been pruned via CHANGELOG_RETENTION
+        all_changes.delete()
+
+        with web_request_context(self.user, change_id=request_id):
+            location.tags.add(tag)
```
glennmatthews, line 311: Doesn't/shouldn't this result in a new ObjectChange being created and therefore `most_recent_change` (below) being non-None? Not sure I fully understand the code flow involved here.

joewesch: Maybe the bug here is that modifying M2M fields on an object does not trigger an actual change log to be created, but it **does** trigger the webhook:
```
In [1]: Location.objects.first()
Out[1]: <Location: Australia>

In [2]: location = _

In [3]: location.tags.all()
Out[3]: <TagQuerySet []>

In [4]: Tag.objects.first()
Out[4]: <Tag: Amber>

In [5]: tag = _

In [6]: from nautobot.core.utils.lookup import get_changes_for_model

In [7]: changes = get_changes_for_model(location)

In [8]: changes.delete()
Out[8]: (1, {'extras.ObjectChange': 1})

In [9]: from nautobot.extras.context_managers import web_request_context

In [10]: user = User.objects.get(username="admin")

In [11]: with web_request_context(user, context_detail="manual-fix"):
    ...:     location.tags.add(tag)
    ...:

In [12]: get_changes_for_model(location)
Out[12]: <RestrictedQuerySet []>
```

joewesch: My fix is merely the bandaid.

glennmatthews: That *sort of* makes sense, as via the UI or API you're only going to change `tags` or any other M2M via an operation that also saves the object itself (even if nothing about the object has otherwise changed). Once #3321 is fixed though that would probably expose this issue anew. 

This might be possible to fix even now, I suspect the following code in `nautobot/extras/signals.py` to be the issue:

```python
        if object_m2m_changed:
            related_changes = ObjectChange.objects.filter(
                changed_object_type=ContentType.objects.get_for_model(instance),
                changed_object_id=instance.pk,
                request_id=change_context_state.get().change_id,
            )
            m2m_changes = instance.to_objectchange(action)
            related_changes.update(object_data=m2m_changes.object_data, object_data_v2=m2m_changes.object_data_v2)   # <-----<< will do nothing if there's no existing objectchange
            objectchange = related_changes.first() if related_changes.exists() else None  # <-----<< else case should probably create a new objectchange instead of using None.
```

I also note the following inconsistent behavior in that same function:

```python
        # Enqueue job hooks
        if objectchange is not None:
            enqueue_job_hooks(objectchange)

    # Enqueue webhooks
    enqueue_webhooks(instance, change_context_state.get().get_user(), change_context_state.get().change_id, action)
```

I suspect you'll find that calling `object.tags.set(...)` when there's no existing objectchange will trigger webhooks but will not trigger jobhooks. 😿 

Would you like to take a stab at addressing either or both of the above issues as a part of this PR?

glennmatthews:
This overlaps with the changes being made in #4109 - I'm going to see if I can reconcile the two into a single PR branch.

joewesch:
Thanks, Glenn. I got pulled into something more urgent and haven't gotten a chance to get back to this yet.

jathanism (approved):
Great work you guys!

Issue #4329, opened 2023-08-25
Filtering By Related Tree Models With Non-Unique Names Can Return Unexpected Results


bryanculver:
Additional unexpected behavior transforms filtering of parent locations into the parent name and the name of all descendants in the tree (as opposed to the actual descendants in the tree). For example, having the following Location Tree:

```
Region 1
  Site 1
Region 2
  Site 1
  Site 2
```

and using the filter `{"location": ["Region 2"]}`

Would result in `location__name__in=["Region 2", "Site 1", "Site 2"]`

I would expect this to be `location__in=[<LocationRegion1>,<LocationSite1>,<LocationSite2>]`

_Originally posted by @jmcgill298 in https://github.com/nautobot/nautobot/issues/4207#issuecomment-1693821232_

bryanculver:
Example: 
<img width="1492" alt="image" src="https://github.com/nautobot/nautobot/assets/31187/90fdcfd4-f458-4ef4-a04f-8f0f7ecbbb57">

bryanculver:
Closed via #4330

PR #4330, opened 2023-08-25
Use PK of Descendants in TreeNodeMultipleChoiceFilter

bryanculver:
# Closes: #4329
# What's Changed

- Used the `pk` of the descendants instead of `name` which is non-unique. This _could_ be backported to 1.6 but we used `slug` there which was globally unique.

# TODO

- [x] Unit, Integration Tests


bryanculver:
Using the dataset constructed in #4329, fixed: 
image

nautobot/core/filters.py
```diff
@@ -501,7 +501,7 @@ def generate_query(self, value, qs=None, **kwargs):
         predicates = []
         for obj in value:
             # Try to get the `to_field_name` (e.g. `name`) or just pass the object through.
-            val = getattr(obj, self.field.to_field_name, obj)
```
HanlinMiao, line 504: For my own edification, how is the filter still able to filter on `name` if we make this change here?

bryanculver: @glennmatthews Might be able to point to the line where this is happening but by the time we get to `generate_query` we already have actual instances (hence the `node.descendants(include_self=True)` above). When you intend to search for something by name, you want to find all instances that match that name. Without this fix, it is using the name of the descendants to then perform the query again (find all instances that match the name of my descendants) you expand the resulting query.

Something like:

```
Blues
  Teal
  Aquamarine
  Sky
Reds
  Crimson
  Strawberry
  Burgundy
```

If you search for "Devices in Locations with the name 'Red'", the resulting filter becomes "Devices in Locations with the names ['Red', 'Crimson', 'Strawberry', 'Burgundy']". Fine but if you add "Purple" as a Location in both Blues and Reds, Devices in Blues > Purple would show up too.

Instead what is performed is "Give me all Locations with the name 'Red'" and then to process the tree, it then performs "Give me all Locations matching this PK which are the descendants of Locations with the name 'Red'".

nautobot/core/filters.py
```diff
@@ -501,7 +501,7 @@ def generate_query(self, value, qs=None, **kwargs):
         predicates = []
         for obj in value:
             # Try to get the `to_field_name` (e.g. `name`) or just pass the object through.
```
bryanculver, line 503:
```suggestion
            # Get the exact instance by PK as we are nested from the original query,
            #   or just pass the object through, commonly the null case.
```
@HanlinMiao?

nautobot/core/tests/test_filters.py
```diff
@@ -87,26 +87,62 @@ def setUp(self):
...
+            status=status,
+        )
         self.queryset = dcim_models.Location.objects.filter(name__icontains="Test Child Location")
 
     def test_filter_single_name(self):
         kwargs = {"parent": [self.parent_location_1.name]}
         qs = self.LocationFilterSet(kwargs, self.queryset).qs
 
-        self.assertQuerysetEqual(qs, [self.child_location_1])
+        self.assertQuerysetEqual(qs, [self.child_location_1, self.child_location_same_name_1])
```
glennmatthews, line 124: Hmm, something seems off here - why isn't `self.parent_location_same_name_1` expected to be part of this result too? It's a child of `child_location_1` and a parent of `child_location_same_name_1` so why are they both returned but not it?

bryanculver: Note the name filter of the base QuerySet.

HanlinMiao: This line 90/118 filters out all the locations with name "Test Parent Location" etc.

glennmatthews (approved):

Issue #4311, opened 2023-08-23
Update `Pull Request` CI Workflow


bryanculver:
### Proposed Changes

- Run all linters (including pylint) and `check-schema` and `check-migrations` in parallel
    - Pylint can be excluded from the "depends on" for unit tests
- Change the unit test matrix to test:
    - Max supported Python version on MySQL DB
    - Min supported Python version on Postgres DB
    - (the above are inverse of Integration for sufficient coverage, these are aimed at speed: fastest Python with slowest DB and vise versa)
- Run integration tests in parallel with unit tests
- Run migration tests in parallel with unit tests
- Add a container build in parallel with unit tests
    - Max supported Python version
    - x86 only
    - Don't publish

### Justification

- All linters are reasonably performant and schema/migration checks will cause unittest to fail anyway
- We rarely find intermediate-supported Python failures
    - Either you used too new of a language feature and that would be caught by min test, or
    - You used a potentially deprecated/removed language feature and that would be caught by the max test
- Integration tests test different things than unit tests and with fail-fast can be grouped together, no point in waiting 30min for unittest only to find out you'll fail integration and need to re-run again
- Migration tests test different things and are always good to have, but same as integration: you'll have to fix it and might as well cancel unittests if they're going to fail
- Only way we can check container builds currently is by merging them into an integration branch

gsnider2195:
> Run migration tests in parallel with unit tests

Migration tests aren't currently run in `CI - Pull Request`. Did we decide to add this test?

gsnider2195:
Never mind. That's only in `next`.

bryanculver:
Closed with #4340

PR #4340, opened 2023-08-28
WIP - Update CI pull request workflow

gsnider2195:
# Closes: #4311
# What's Changed

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/.github/workflows/ci_pullrequest_test.yml
+++ b/.github/workflows/ci_pullrequest_test.yml
@@ -1,0 +1,329 @@
+---
+# A lighter-weight version of CI specifically for pull requests
+name: "CI - Pull Request"
+concurrency:  # Cancel any existing runs of this workflow for this same PR
+  group: "${{ github.workflow }}-${{ github.ref }}"
+  cancel-in-progress: true
+on:  # yamllint disable
+  pull_request:
+  workflow_call:
+
+jobs:
+  black:
+    runs-on: "ubuntu-20.04"
+    env:
+      INVOKE_NAUTOBOT_LOCAL: "True"
+    steps:
+      - name: "Do nothing"
+        run: "echo 'no-op success job'"
+  flake8:
+    runs-on: "ubuntu-20.04"
+    env:
+      INVOKE_NAUTOBOT_LOCAL: "True"
+    steps:
+      - name: "Do nothing"
+        run: "echo 'no-op success job'"
+  prettier:
+    runs-on: "ubuntu-20.04"
+    env:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_pullrequest.yml
```diff
@@ -341,19 +324,20 @@ jobs:
...
-
+      - "black"
+      - "check-migrations"
+      - "check-schema"
+      - "flake8"
+      - "hadolint"
+      - "markdownlint"
+      - "yamllint"
   migration-tests-postgres:
+    name: "Run migrations against test dataset (PostgreSQL and Python 3.10)"
```
HanlinMiao, line 337: Quick question, Should we run this using Python 3.11? Since that is what we are doing with the unit tests.

.github/workflows/ci_pullrequest.yml
```diff
@@ -487,3 +435,52 @@ jobs:
         run: |
           git fetch --no-tags origin +refs/heads/${{ github.base_ref }}:refs/remotes/origin/${{ github.base_ref }}
           poetry run towncrier check --compare-with origin/${{ github.base_ref }}
+  container-build-test:
+    name: "Test Container Build (amd64 only on Python 3.11)"
+    runs-on: "ubuntu-22.04"
```
HanlinMiao, line 440: Any reason why this is the only test that runs on ubuntu-22.04 instead of ubuntu-20.04?

changes/4311.housekeeping
```diff
@@ -0,0 +1 @@
+Updated CI pull request workflow for `next` branch to remove some redundant tests, run more tests in parallel and test the container build.
```
HanlinMiao, line 1: ChangeLog test still failing for some reason, maybe we need to merge `next` into this branch?

.github/workflows/ci_pullrequest.yml
gsnider2195: Probably. Not sure if this was missed or intentional when we upped the max version.

.github/workflows/ci_pullrequest.yml
gsnider2195: I think @snaselj updated this to use the new version when he updated the container build process. We're probably ok to update the others but we need to account for the time it will take to research and test in a separate issue.

changes/4311.housekeeping
gsnider2195: No, that change is still in develop. I think we're ok to merge with failing changelog as long as we plan on merging develop into next before the next 2.0 release.

HanlinMiao (approved):

Issue #4017, opened 2023-06-28
Refine Device Type View in New UI


bryanculver:
As a Nautobot 2.0 user, I want to be able to interact with the Device Type list and detail views in the new UI.

Strive for parity with the legacy UI, for top level, concrete fields:

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBbWdoIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--c79da50328d51cdae7b5353c1599c48cd36e755e/Screenshot%202023-08-21%20at%201.05.43%20PM.png)

- Show the image fields with just the image file names (we will worry about rendering the images later)
- Move object type to advanced
- Remove the notes URL entirely
- Ignore tags, custom fields, computed fields, relationships, etc (they will be solved generically)

glennmatthews:
Implemented in #4352.

PR #4352, opened 2023-08-29
Refine new-UI DeviceType detail view

glennmatthews:
# Closes: #4017
# What's Changed

- Move `object_type` field to the Advanced tab in the new UI
- Exclude `notes_url` from the new UI altogether
  - Notes tab still renders correctly in the new UI using this information, it's just not rendered as a table field any more
- Refine DeviceType view in new UI
  - Caveats:
     - Legacy UI used "Parent/Child" for the `subdevice_role` field, which the new UI is auto-labeling as "Subdevice role" - I think the latter is clearer but it is a difference.
     - Front/rear images are currently presented as raw URLs rather than rendering as images; per #4017 this is fine for now
     - The `Device count` value is not currently hyperlinked to a filtered Device list view; we don't currently have a pattern for attributing a hyperlink to what (in the REST API) is just a raw number value. Future work?

image

image

# TODO

- n/a Unit, Integration Tests

### DIFF
--- a/changes/4017.changed
+++ b/changes/4017.changed
@@ -1,0 +1,2 @@
+Refined new-UI DeviceType detail view.
+Moved `object_type` to the Advanced tab of new-UI detail views in general.
--- a/changes/4017.removed
+++ b/changes/4017.removed
@@ -1,0 +1,1 @@
+Removed `notes_url` from new-UI object detail views.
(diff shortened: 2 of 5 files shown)


HanlinMiao (approved):
After merging `next` into this branch we are good to go!

Issue #1905, opened 2022-06-17
Add `__is_null` as a filter expression


whitej6:
### As ...

Austin - Network Automation Engineer

### I want ...

To filter a list of objects based on if an attribute is `null`

### So that ...

I can get a list of objects from the API that has an attribute set to `null`

### I know this is done when...

- `null` is set as a filter expression that maps to `<attr>__isnull`
- Can filter based on `<attr>__null` 

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [X] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

None

### External Dependencies

None

glennmatthews:
Implemented for 2.1 by #4635.

PR #4378, opened 2023-09-03
Add `isnull` Filter on Nullable Fields via Filter Field Factory

itdependsnetworks:
# Closes: #1905
# What's Changed

Added the ability to automatically apply `isnull` filters when model field is nullable

# TODO

- [x] Unit, Integration Tests

@glennmatthews I know we spoke about another approach, it quickly got complicated to consider what to do in various situations, such as an empty string vs null check on a char or 0 on an integer. This seemed to be a reasonable approach considering that it is basically 5 lines of code (not including comments, blackified spacing, tests, etc.), follows an established pattern, and works for plugins, api, graphql immediately. 

That being said, I didn't want to spend the time building tests without confirmation that this approach makes sense.

### DIFF
--- a/nautobot/core/filters.py
+++ b/nautobot/core/filters.py
@@ -642,10 +642,14 @@ class BaseFilterSet(django_filters.FilterSet):
         if field is None:
             return magic_filters
 
+        field_type = None
         try:
-            field_type = cls._meta.model._meta.get_field(field_name)
+            if hasattr(cls._meta.model, field_name):
+                field_type = cls._meta.model._meta.get_field(field_name)
         except FieldDoesNotExist:
-            field_type = None
+            # Will get error like:
+            # `FieldDoesNotExist: RelationshipAssociation has no field named 'relationship__key'`
+            pass
 
         # If the field allows null values, add an `isnull`` check
         if field.null and isinstance(field_type, NULLABLE_CLASSES):
--- a/nautobot/core/filters.py
+++ b/nautobot/core/filters.py
@@ -8,6 +8,7 @@ from django.conf import settings
 from django.core.exceptions import FieldDoesNotExist
 from django.db import models
 from django.forms.utils import ErrorDict, ErrorList
+from django.utils.module_loading import import_string
 import django_filters
 from django_filters.constants import EMPTY_VALUES
 from django_filters.utils import get_model_field, resolve_field
@@ -644,8 +645,12 @@ class BaseFilterSet(django_filters.FilterSet):
 
         field_type = None
         try:
-            if hasattr(cls._meta.model, field_name):
-                field_type = cls._meta.model._meta.get_field(field_name)
+            field_type = cls._meta.model._meta.get_field(field_name)
+            # TODO: This should not be needed, there seems to be some behavior changes with
+            # RoleFilter filters that haven't been migrated, which is beyond the scope of this change
+            # Working around it for now
+            if isinstance(filter_field, import_string("nautobot.extras.filters.mixins.RoleFilter")):
+                field_type = None
         except FieldDoesNotExist:
             # Will get error like:
             # `FieldDoesNotExist: RelationshipAssociation has no field named 'relationship__key'`
--- a/nautobot/core/filters.py
+++ b/nautobot/core/filters.py
@@ -5,6 +5,7 @@ import uuid
 
 from django import forms as django_forms
 from django.conf import settings
+from django.core.exceptions import FieldDoesNotExist
 from django.db import models
 from django.forms.utils import ErrorDict, ErrorList
 import django_filters
@@ -19,6 +20,17 @@ from nautobot.core.models import fields as core_fields
 from nautobot.core.utils import data as data_utils
 
 logger = logging.getLogger(__name__)
+
+NULLABLE_CLASSES = (
+    models.CharField,
+    models.DateField,
+    models.DecimalField,
+    models.ForeignKey,
+    models.FloatField,
+    models.IntegerField,
+    models.TextField,
+    models.UUIDField,
+)
 
 
 def multivalue_field_factory(field_class, widget=django_forms.SelectMultiple):
@@ -630,15 +642,28 @@ class BaseFilterSet(django_filters.FilterSet):
         if field is None:
             return magic_filters
 
+        try:
+            field_type = cls._meta.model._meta.get_field(field_name)
+        except FieldDoesNotExist:
+            field_type = None
+
+        # If the field allows null values, add an `isnull`` check
+        if field.null and isinstance(field_type, NULLABLE_CLASSES):
+            # Use this method vs extend as the `lookup_map` variable is generally one of
+            # the constants which we do not want to update
+            lookup_map = dict(lookup_map, **{"isnull": "isnull"})
+
         # Create new filters for each lookup expression in the map
         for lookup_name, lookup_expr in lookup_map.items():
             new_filter_name = f"{filter_name}__{lookup_name}"
 
             try:
-                if filter_name in cls.declared_filters:
-                    # The filter field has been explicity defined on the filterset class so we must manually
+                if filter_name in cls.declared_filters and lookup_expr not in {"isnull"}:
+                    # The filter field has been explicitly defined on the filterset class so we must manually
                     # create the new filter with the same type because there is no guarantee the defined type
-                    # is the same as the default type for the field
+                    # is the same as the default type for the field. This does not apply to apply if the filter
+                    # should retain the original lookup_expr type, such as `isnull` using a boolean field on a
+                    # char or date object.
                     resolve_field(field, lookup_expr)  # Will raise FieldLookupError if the lookup is invalid
                     new_filter = type(filter_field)(
                         field_name=field_name,
@@ -649,7 +674,7 @@ class BaseFilterSet(django_filters.FilterSet):
                         **filter_field.extra,
                     )
                 else:
-                    # The filter field is listed in Meta.fields so we can safely rely on default behaviour
+                    # The filter field is listed in Meta.fields so we can safely rely on default behavior
                     # Will raise FieldLookupError if the lookup is invalid
                     new_filter = cls.filter_for_field(field, field_name, lookup_expr)
             except django_filters.exceptions.FieldLookupError:
--- a/nautobot/docs/user-guide/platform-functionality/rest-api/filtering.md
+++ b/nautobot/docs/user-guide/platform-functionality/rest-api/filtering.md
@@ -73,10 +73,28 @@ Custom fields can be mixed with built-in fields to further narrow results. When 
 
 ## Lookup Expressions
 
-Certain model fields (including, in Nautobot 1.4.0 and later, custom fields of type `text`, `url`, `select`, `integer`, and `date`) also support filtering using additional lookup expressions. This allows
-for negation and other context-specific filtering.
+Certain model fields (including, in Nautobot 1.4.0 and later, custom fields of type `text`, `url`, `select`, `integer`, and `date`) also support filtering using additional lookup expressions. This allows for negation and other context-specific filtering.
 
 These lookup expressions can be applied by adding a suffix to the desired field's name, e.g. `mac_address__n`. In this case, the filter expression is for negation and it is separated by two underscores. Below are the lookup expressions that are supported across different field types.
+
+### Nullable Fields
+
+The `isnull` lookup expression is automatically added when the following three conditions are met:
+
+1. A FilterSet that inherits `nautobot.apps.filters.BaseFilterSet`
+2. The model field is set to `Null = True`
+3. A model field that is or inherits from one of:
+    - `django.db.models.CharField`
+    - `django.db.models.DateField`
+    - `django.db.models.DecimalField`
+    - `django.db.models.ForeignKey`
+    - `django.db.models.FloatField`
+    - `django.db.models.IntegerField`
+    - `django.db.models.TextField`
+    - `django.db.models.UUIDField`
+
++/- 2.0.0
+    This feature was only added in Nautobot 2.0.0.
 
 ### Numeric Fields
 
--- a/changes/4378.added
+++ b/changes/4378.added
@@ -1,0 +1,1 @@
+Added the ability to automatically apply `isnull` filters when model field is nullable.
--- a/nautobot/core/tests/test_filters.py
+++ b/nautobot/core/tests/test_filters.py
@@ -1,3 +1,5 @@
+import datetime
+
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.contenttypes.models import ContentType
@@ -9,6 +11,8 @@ from django.test.utils import isolate_apps
 import django_filters
 from tree_queries.models import TreeNodeForeignKey
 
+from nautobot.circuits.filters import CircuitFilterSet
+from nautobot.circuits.models import Circuit, CircuitType, Provider
 from nautobot.core import filters, testing
 from nautobot.core.models import fields as core_fields
 from nautobot.core.utils import lookup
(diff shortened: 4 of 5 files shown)


glennmatthews:
See also #1905.

itdependsnetworks:
@glennmatthews, thanks for the reference, didn't know about that PR. This should be different since we are doing a `if field.null and field.get_internal_type() in constants.FILTER_LOCAL_FIELDS:` check, so only if the model field is nullable will this show up.

nautobot/docs/user-guide/platform-functionality/rest-api/filtering.md
```diff
@@ -73,11 +73,31 @@ Custom fields can be mixed with built-in fields to further narrow results. When
...
 
-Certain model fields (including, in Nautobot 1.4.0 and later, custom fields of type `text`, `url`, `select`, `integer`, and `date`) also support filtering using additional lookup expressions. This allows
-for negation and other context-specific filtering.
+Certain model fields (including, in Nautobot 1.4.0 and later, custom fields of type `text`, `url`, `select`, `integer`, and `date`) also support filtering using additional lookup expressions. This allows for negation and other context-specific filtering.
 
 These lookup expressions can be applied by adding a suffix to the desired field's name, e.g. `mac_address__n`. In this case, the filter expression is for negation and it is separated by two underscores. Below are the lookup expressions that are supported across different field types.
 
+### Nullable Fields
+
+The `isnull` lookup expression is automatically added to any of the below field types for all core Nautobot fields and any Nautobot app FilterSet's that inherit from `nautobot.apps.filters.BaseFilterSetBaseFilterSet`.
```
jathanism, line 82:
```suggestion
The `isnull` lookup expression is automatically added to any of the below field types for all core Nautobot fields and any Nautobot app FilterSet's that inherit from `nautobot.apps.filters.BaseFilterSet`.
```

nautobot/core/constants.py
```diff
@@ -32,6 +32,24 @@
...
+    "IntegerField",
+    "PositiveIntegerField",
+    "PositiveSmallIntegerField",
+    "SlugField",
+    "SmallIntegerField",
+    "TextField",
+    "UUIDField",
+}
+
+FILTER_LOOKUP_EXPR = {"isnull"}
```
jathanism, line 51: These constants appear to be ambiguously named to me when they are currently only used to determine if a field should support `isnull` filtering. Were you thinking there would be future extensions to this?

nautobot/core/filters.py
```diff
@@ -629,15 +629,23 @@ def _generate_lookup_expression_filters(cls, filter_name, filter_field):
         if field is None:
             return magic_filters
 
+        # If the field allows null values, add an `isnull`` check
+        if field.null and field.get_internal_type() in constants.FILTER_LOCAL_FIELDS:
+            # Use this method vs extend as the `lookup_map` variable is generally one of
+            # the constants which we do not want to update
+            lookup_map = dict(lookup_map, **{"isnull": "isnull"})
```
jathanism, line 636: Similarly here since the code is trying to discern and set the lookup value also be the string `"isnull"`, it feels like it would be more readable if we didn't just reference "local fields".

glennmatthews:
For reference, the alternate approach that I'd personally prefer would be to extend the existing "exact" filters (e.g. `name`, `name__n`, `ip_version`/`ip_version__n`, etc.) to recognize the existing [`NULL_CHOICE_VALUE`](https://django-filter.readthedocs.io/en/stable/ref/settings.html?highlight=null#filters-null-choice-value) setting, so that the filter would be `?name=null`/`?name__n=null` rather than `?name__isnull=true`/`?name__isnull=false`.

My preference for that approach is that we already have a tremendous number of uniquely named filter variants, and not wanting to worsen that problem by adding yet more of them.

nautobot/core/constants.py
```diff
@@ -32,6 +32,24 @@
...
+    "DecimalField",
+    "FloatField",
+    "IntegerField",
+    "PositiveIntegerField",
+    "PositiveSmallIntegerField",
+    "SlugField",
+    "SmallIntegerField",
+    "TextField",
+    "UUIDField",
+}
```
glennmatthews, line 49: Instead of explicitly looking for exact matches on `field.get_internal_type()` strings, can we instead use `isinstance()` checks? That would automatically (and more correctly) handle subclasses of any of these types, and would also simplify this listing substantially (e.g. `isinstance(IntegerField)` would also automatically cover `SmallIntegerField`, `PositiveSmallIntegerField`, etc. since they're just subclasses of it).

nautobot/core/filters.py
itdependsnetworks: premature optimization, should be good now.

itdependsnetworks:
> It would pose some problems on the filter form presentation though, as there's no way for a number input field to show a "null" value string, etc. So this approach has its advantages as well.

This was the issue. It seemed like a rabbit hole, and at least we have documented this in the future to fix. All in all, I actually working with it, I did genuinely prefer this approach.

nautobot/docs/user-guide/platform-functionality/rest-api/filtering.md
```diff
@@ -73,11 +73,25 @@ Custom fields can be mixed with built-in fields to further narrow results. When
...
+- `DateField`
+- `DecimalField`
+- `FloatField`
+- `IntegerField`
+- `TextField`
+- `UUIDField`
+
++/- 2.0.0
+    This feature was only added in Nautobot 2.0.0.
+
```
glennmatthews, line 94:
```suggestion
+++ 2.0.0

The `isnull` lookup expression is automatically added to any of the below field types or their subclasses, if instantiated with `null=True`, by all core Nautobot / Nautobot App FilterSet's that inherit from `nautobot.apps.filters.BaseFilterSet`.

- `CharField`
- `DateField`
- `DecimalField`
- `FloatField`
- `IntegerField`
- `TextField`
- `UUIDField`

```

nautobot/core/filters.py
```diff
@@ -629,15 +640,28 @@ def _generate_lookup_expression_filters(cls, filter_name, filter_field):
         if field is None:
             return magic_filters
 
+        try:
+            field_type = cls._meta.model._meta.get_field(field_name)
+        except FieldDoesNotExist:
+            field_type = None
```
glennmatthews, line 646: What's the case where we'd hit this exception? Is this for non-model filter-fields (SerializerMethodField, etc.) or something else?

nautobot/core/filters.py
```diff
@@ -629,15 +640,28 @@ def _generate_lookup_expression_filters(cls, filter_name, filter_field):
...
+
+        # If the field allows null values, add an `isnull`` check
+        if field.null and isinstance(field_type, NULLABLE_CLASSES):
+            # Use this method vs extend as the `lookup_map` variable is generally one of
+            # the constants which we do not want to update
+            lookup_map = dict(lookup_map, **{"isnull": "isnull"})
```
glennmatthews, line 652: Would be good to add tests for this logic, likely in `nautobot/core/tests/test_filters.py` `BaseFilterSetTest` class.

nautobot/core/filters.py
itdependsnetworks:
```bash


  Registering job nautobot.core.jobs.GitRepositoryDryRun
  File "/opt/nautobot/lib/python3.8/site-packages/django/db/models/options.py", line 608, in get_field
    return self.fields_map[field_name]
KeyError: 'relationship__key'

During handling of the above exception, another exception occurred:

  File "/opt/nautobot/bin/nautobot-server", line 6, in <module>
    sys.exit(main())
  File "/source/nautobot/core/cli/__init__.py", line 54, in main
    run_app(
  File "/source/nautobot/core/runner/runner.py", line 297, in run_app
    management.execute_from_command_line([runner_name, command] + command_args)
  File "/opt/nautobot/lib/python3.8/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/opt/nautobot/lib/python3.8/site-packages/django/core/management/__init__.py", line 395, in execute
    django.setup()
  File "/opt/nautobot/lib/python3.8/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/opt/nautobot/lib/python3.8/site-packages/django/apps/registry.py", line 122, in populate
    app_config.ready()
  File "/source/nautobot/circuits/apps.py", line 14, in ready
    super().ready()
  File "/source/nautobot/core/apps/__init__.py", line 57, in ready
    register_menu_items(menu_items)
  File "/source/nautobot/core/apps/__init__.py", line 104, in register_menu_items
    item.link = reverse(item.link, args=item.args, kwargs=item.kwargs)
  File "/opt/nautobot/lib/python3.8/site-packages/django/urls/base.py", line 54, in reverse
    app_list = resolver.app_dict[ns]
  File "/opt/nautobot/lib/python3.8/site-packages/django/urls/resolvers.py", line 530, in app_dict
    self._populate()
  File "/opt/nautobot/lib/python3.8/site-packages/django/urls/resolvers.py", line 464, in _populate
    for url_pattern in reversed(self.url_patterns):
  File "/opt/nautobot/lib/python3.8/site-packages/django/utils/functional.py", line 48, in __get__
    res = instance.__dict__[self.name] = self.func(instance)
  File "/opt/nautobot/lib/python3.8/site-packages/django/urls/resolvers.py", line 602, in url_patterns
    patterns = getattr(self.urlconf_module, "urlpatterns", self.urlconf_module)
  File "/opt/nautobot/lib/python3.8/site-packages/django/utils/functional.py", line 48, in __get__
    res = instance.__dict__[self.name] = self.func(instance)
  File "/opt/nautobot/lib/python3.8/site-packages/django/urls/resolvers.py", line 595, in urlconf_module
    return import_module(self.urlconf_name)
  File "/usr/local/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 843, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/source/nautobot/core/urls.py", line 7, in <module>
    from nautobot.extras.plugins.urls import (
  File "/source/nautobot/extras/plugins/urls.py", line 32, in <module>
    urlpatterns = import_object(f"{plugin_path}.urls.urlpatterns")
  File "/source/nautobot/extras/plugins/utils.py", line 46, in import_object
    spec.loader.exec_module(module)
  File "/source/examples/example_plugin/example_plugin/urls.py", line 6, in <module>
    from nautobot.extras.views import ObjectDynamicGroupsView
  File "/source/nautobot/extras/views.py", line 42, in <module>
    from . import filters, forms, tables
  File "/source/nautobot/extras/filters/__init__.py", line 813, in <module>
    class RelationshipAssociationFilterSet(BaseFilterSet):
  File "/opt/nautobot/lib/python3.8/site-packages/django_filters/filterset.py", line 62, in __new__
    new_class.base_filters = new_class.get_filters()
  File "/source/nautobot/core/filters.py", line 733, in get_filters
    cls._generate_lookup_expression_filters(
  File "/source/nautobot/core/filters.py", line 644, in _generate_lookup_expression_filters
    field_type = cls._meta.model._meta.get_field(field_name)
  File "/opt/nautobot/lib/python3.8/site-packages/django/db/models/options.py", line 610, in get_field
    raise FieldDoesNotExist("%s has no field named '%s'" % (self.object_name, field_name))
django.core.exceptions.FieldDoesNotExist: RelationshipAssociation has no field named 'relationship__key'

```

nautobot/core/filters.py
```diff
@@ -630,15 +642,28 @@ def _generate_lookup_expression_filters(cls, filter_name, filter_field):
         if field is None:
             return magic_filters
 
+        try:
+            field_type = cls._meta.model._meta.get_field(field_name)
```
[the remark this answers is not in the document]
lampwins: What's the harm in having `isnull` generally available? I.e. just adding it to `FILTER_CHAR_BASED_LOOKUP_MAP` and `FILTER_NUMERIC_BASED_LOOKUP_MAP` and dispensing with this conditional field logic? It's a boolean so fields that aren't allowed to be null will still technically work with the filter.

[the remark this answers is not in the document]
itdependsnetworks: I don't know? It seemed like we would only want on nullable fields, but not sure.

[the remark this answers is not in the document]
itdependsnetworks: I am not sure it makes sense to have it on "isnull", as empty and null are different. 

```python
In [2]: Device.objects.filter(serial__isnull=True)
Out[2]: <ConfigContextModelQuerySet []>

In [3]: Device.objects.first().serial
Out[3]: ''

In [4]: Device.objects.filter(serial__isnull=False)
Out[4]: <ConfigContextModelQuerySet [<Device: ams01-dist-01>, <Device: ams01-edge-01>,'...(remaining elements truncated)...']>
```

An isnull on the empty string I would have expected to return something, and my bet is others would have similar confusion.

nautobot/core/filters.py
```diff
@@ -630,15 +643,36 @@ def _generate_lookup_expression_filters(cls, filter_name, filter_field):
         if field is None:
             return magic_filters
 
+        field_type = None
+        try:
+            field_type = cls._meta.model._meta.get_field(field_name)
+            # TODO: This should not be needed, there seems to be some behavior changes with
+            # RoleFilter filters that haven't been migrated, which is beyond the scope of this change
+            # Working around it for now
```
glennmatthews, line 651: Would like to better understand the issue/symptom here so we can figure out the correct fix.

nautobot/core/filters.py
```diff
@@ -630,15 +643,36 @@ def _generate_lookup_expression_filters(cls, filter_name, filter_field):
...
+            if isinstance(filter_field, import_string("nautobot.extras.filters.mixins.RoleFilter")):
+                field_type = None
+        except FieldDoesNotExist:
+            # Will get error like:
+            # `FieldDoesNotExist: RelationshipAssociation has no field named 'relationship__key'`
+            pass
+
+        # If the field allows null values, add an `isnull`` check
+        if field.null and isinstance(field_type, NULLABLE_CLASSES):
```
glennmatthews, line 660: Is there a case you encountered where we have a field that is `.null` but *isn't* one of `NULLABLE_CLASSES`? Just wanting to understand what fields we're excluding with the `isinstance` check.

itdependsnetworks: I do not know, I asked ChatGPT which ones are, that gave me a list, I added those, you let me know to just use the inherited class, and I figured it was good preventive measure.

nautobot/core/filters.py
itdependsnetworks: https://github.com/nautobot/nautobot/actions/runs/6215156129/job/16867760641

glennmatthews: Ugh, I hadn't realized that `RoleFilter` and `StatusFilter` had such completely different code structures; I would have expected both or neither to show this behavior. It's *probably* incorrect for `RoleFilter` to restrict its queryset (https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/filters/mixins.py#L304-L305) based on content-type (StatusFilter doesn't do this).

nautobot/core/filters.py
glennmatthews: Just not sure what we're preventing here. If a field has a `null=True` flag on it, it *seems* like we should have a corresponding `isnull` filter on it, regardless of what type of field it is?

nautobot/core/filters.py
itdependsnetworks: >  I would have expected both or neither to show this behavior. 

That was my exact comment to John yesterday

nautobot/core/filters.py
itdependsnetworks: works for me, I will make the update.

itdependsnetworks: It failed on some other similar issue. Can we put this in for now? I think it makes sense to:

* Merge for now
* Develop a system that sanely fails before this becomes an issue, so we can stop such future issues from happening
* Remove the nullable_class check

itdependsnetworks: Just change from `if field.null and isinstance(field_type, NULLABLE_CLASSES):` to `if field.null:` and re-run tests to see what I was referring to.

snaselj: The solution with `NULLABLE_CLASSES` works because it skips certain fields. E.g. `dcim.Rack.role` field has `null=True`, but it's skipped when using `NULLABLE_CLASSES`.

Issue is a race condition, as we are trying to fetch data from the database to configure filters before the database is ready.

https://github.com/nautobot/nautobot/blob/develop/nautobot/core/models/fields.py#L190

IMHO it's better to fix this race condition first, and then implement this feature using duck-typing approach. WDYT @glennmatthews?

itdependsnetworks: It sounds like my suspicion was correct and we should probably: 

> Develop a system that sanely fails before this becomes an issue, so we can stop such future issues from happening

I realize that I say that without understanding how difficult it may be... but seems like the correct direction.

glennmatthews: > Issue is a race condition, as we are trying to fetch data from the database to configure filters before the database is ready.
> 
> https://github.com/nautobot/nautobot/blob/develop/nautobot/core/models/fields.py#L190
> 
> IMHO it's better to fix this race condition first, and then implement this feature using duck-typing approach. WDYT @glennmatthews?

Agreed, see also my comment above https://github.com/nautobot/nautobot/pull/4378#discussion_r1334387786

snaselj: Something like this can resolve it for now, tests passes:

```
diff --git a/nautobot/core/filters.py b/nautobot/core/filters.py
index 0851d68fd..ae9b5b017 100644
--- a/nautobot/core/filters.py
+++ b/nautobot/core/filters.py
@@ -5,10 +5,8 @@ import uuid
 
 from django import forms as django_forms
-from django.core.exceptions import FieldDoesNotExist
 from django.db import models
 from django.forms.utils import ErrorDict, ErrorList
-from django.utils.module_loading import import_string
 import django_filters
 from django_filters.constants import EMPTY_VALUES
 from django_filters.utils import get_model_field, resolve_field
@@ -22,17 +20,6 @@ from nautobot.core.utils import data as data_utils
 
-NULLABLE_CLASSES = (
-    models.CharField,
-    models.DateField,
-    models.DecimalField,
-    models.ForeignKey,
-    models.FloatField,
-    models.IntegerField,
-    models.TextField,
-    models.UUIDField,
-)
-
 
 def multivalue_field_factory(field_class, widget=django_forms.SelectMultiple):
     """
@@ -643,21 +630,8 @@ class BaseFilterSet(django_filters.FilterSet):
         if field is None:
             return magic_filters
 
-        field_type = None
-        try:
-            field_type = cls._meta.model._meta.get_field(field_name)
-            # TODO: This should not be needed, there seems to be some behavior changes with
-            # RoleFilter filters that haven't been migrated, which is beyond the scope of this change
-            # Working around it for now
-            if isinstance(filter_field, import_string("nautobot.extras.filters.mixins.RoleFilter")):
-                field_type = None
-        except FieldDoesNotExist:
-            # Will get error like:
-            # `FieldDoesNotExist: RelationshipAssociation has no field named 'relationship__key'`
-            pass
-
         # If the field allows null values, add an `isnull`` check
-        if field.null and isinstance(field_type, NULLABLE_CLASSES):
+        if getattr(field, "null", None):
             # Use this method vs extend as the `lookup_map` variable is generally one of
             # the constants which we do not want to update
             lookup_map = dict(lookup_map, **{"isnull": "isnull"})
diff --git a/nautobot/core/models/fields.py b/nautobot/core/models/fields.py
index 49f2bf525..822babb8e 100644
--- a/nautobot/core/models/fields.py
+++ b/nautobot/core/models/fields.py
@@ -187,7 +187,12 @@ class ForeignKeyLimitedByContentTypes(ForeignKeyWithAutoRelatedName):
     """
 
     def get_limit_choices_to(self):
-        return {"content_types": ContentType.objects.get_for_model(self.model)}
+        try:
+            return {"content_types": ContentType.objects.get_for_model(self.model)}
+        except:
+            import logging
+            logging.error(5 * "UNHANDLED", exc_info=True)
+            return {}
 
     def formfield(self, **kwargs):
         """Return a prepped formfield for use in model forms."""
```

Issue #4338, opened 2023-08-28
Establish Long-Term Maintenance Branch


bryanculver:
- Branch `ltm/1.6` off of `develop`
- Update CI to protect and test LTM branches
- Enable for upstream testing workflows for apps
- Document the changes
- Consider merging `next` down into `develop`

bryanculver:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Nautobot-Core-Product-64a8783b7a3f4b1225973a0c/issues/nautobot/nautobot/4338?planning-poker) with Zenhub @glennmatthews @gsnider2195 @HanlinMiao @jathanism @timizuoebideri1

gsnider2195:
The weekly report job on the dev server will need to be updated for the upstream testing workflow

bryanculver:
Renovate Config: https://github.com/nautobot/nautobot/blob/next/renovate.json
Integration Workflow: https://github.com/nautobot/nautobot/blob/next/.github/workflows/ci_integration.yml
PR Workflow: https://github.com/nautobot/nautobot/blob/next/.github/workflows/ci_integration.yml
Branch Management: https://github.com/nautobot/nautobot/settings/branches

PR #4390, opened 2023-09-05
[Review Only Do Not Merge] Long Term Maintenance Branch 1.6

HanlinMiao:
# Closes: #4338 
# What's Changed

# TODO

- [x] update ci_integration.yml
- [x] update ci_pullrequest.yml
- [x] update renovate.json
- [x] update plugin_upstream_testing_base.yml
- [x] establish branch protection rule for `ltm/*` branches


.github/workflows/ci_pullrequest.yml
```diff
@@ -322,7 +322,7 @@ jobs:
       - "tests-mysql"
 
   changelog:
-    if: contains(fromJson('["develop","next","develop-1.6"]'), github.base_ref)
+    if: contains(fromJson('["develop","next","develop-1.6", "ltm/1.6"]'), github.base_ref)
```
glennmatthews, line 325: let's remove `"develop-1.6"` while we're in here.

.github/workflows/ci_pullrequest.yml
```diff
@@ -322,7 +322,7 @@ jobs:
       - "tests-mysql"
 
   changelog:
-    if: contains(fromJson('["develop","next","develop-1.6"]'), github.base_ref)
+    if: contains(fromJson('["develop","next","ltm/1.6"]'), github.base_ref)
```
bryanculver, line 325: Since this is `contains` it should be good enough to have `"ltm/"`

.github/workflows/ci_integration.yml
```diff
@@ -1,12 +1,13 @@
...
-# Full CI for commits to the integration branches (main, develop, next)
+# Full CI for commits to the integration branches (main, develop, next, "ltm/1.6")
 name: "CI - Integration Branch"
 on:  # yamllint disable
   push:
     branches:
       - "main"
       - "develop"
       - "next"
+      - "ltm/1.6"
```
bryanculver, line 10: Does Github actions support wildcards? `"ltm/*"`

renovate.json
```diff
@@ -18,6 +18,16 @@
...
+    {
+      "description": "Group patch-only updates to develop weekly to cut down on PR noise.",
+      "groupName": "Lock-file Patch Releases for LTM branches",
+      "matchBaseBranches": ["ltm/*"],
+      "schedule": [
+        "before 4am on Friday"
+      ],
+      "major": {"enabled": false},
+      "minor": {"enabled": false}
+    },
```
bryanculver, line 30: We discussed before about hard-pinning our dependencies so as no surprise upgrades affect this but I think that might be too hard of a line and our already restrictive requirements has shown problematic for sensible use.

Since we won't be planning bi-weekly releases of our LTM, should we instead move this check to a monthly update?

We might need to find an alternative to Dependabot (or maybe this configuration option: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#target-branch) to get security updates about our LTMs.

.github/workflows/plugin_upstream_testing_base.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        nautobot-version: ["next", "develop"]
+        nautobot-version: ["next", "develop", "ltm/1.6"]
```
bryanculver, line 23: Line 29 here will probably need to be reviewed as well.

Issue #3029, opened 2022-12-20
Missing lookup function for Values in Advanced Filter


martin-networker:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.2
* Python version: 3.10
* Database platform, version: -
* Middleware(s): -

### Steps to Reproduce
1. Click the "Filter" button for example in Rack Elevation (it affects also Device and others)
2. Click the "Advanced" tab
3. Select for example name in (same issue also for custom field in Device) 
4. Select contain in Lookup type
5. Value does not offer anything, you have to open a new tab and find values from there.

### Expected Behavior
Value field will offer possible values.

### Observed Behavior
For some filters the lookup function does not work.

bryanculver:
Contains is a substring search. By nature it won't populate exact names for things as it would be perceived that you would be looking for the name containing a string, such as "105":


https://user-images.githubusercontent.com/31187/208682003-dbb5928d-c5fa-4483-b0ff-020a6c4fd418.mov

bryanculver:
However "name" with "exact" lookup expression could be improved to return the list of device names to provide as options.

Many of the advanced filters take advantage of user-supplied strings for searches that previously weren't possible for the default filters before. As an example, the contains and regex lookup expressions weren't possible to put in a form because you would have only been able to search on one provided string. Multiple values entered into the "value" column are `OR` conditions.

martin-networker:
You are right that for `contains` and `regex` the lookup can't work. Sorry for bad example.
However for `exact` it should work and we will be glad if this will be available.

PR #4392, opened 2023-09-05
Fix #4377 - incorrect logic in Device and VirtualMachine API endpoints

### LINKED ISSUE #4377: DeviceFilterSet attributes being stripped of options
### Environment
* Nautobot version (Docker tag too if applicable): 2.0-rc2
* Python version: 3.11
* Database platform, version: latest
* Middleware(s): N/A

I attempted to look on 1.5 or 1.6 but the demo instances weren't immediately loading to see if this is 1.x or not. 

### Steps to Reproduce
1. Go to https://next.demo.nautobot.com/api/docs/#/dcim/dcim_devices_list
2. Observe has_device_bays, created, last_updated, etc. 

### Expected Behavior

The attributes to match up to their type, such as DateTime, Boolean, Integer, etc.. as shown in the device-types api

image

### Observed Behavior

Everything is defaulting to char type.

image

glennmatthews:
# Closes: #4377
# What's Changed

- Because of the way `drf_spectacular` extensions work, they apply to a base class when specified (such as `NautobotFilterBackend`) but do **not** therefore automatically apply to subclasses of that base class (such as `ConfigContextFilterBackEnd`), which is why, as reported in #4377, the Device and VirtualMachine REST API endpoints (only) did not correctly list their filters in the OpenAPI schema, as these two were using `ConfigContextFilterBackend` rather than `NautobotFilterBackend`.
- I could have fixed this by adding an explicit drf_spectacular extension to `ConfigContextFilterBackend` itself, but as it turns out, after #3814 this class is no longer necessary; in fact it should have been deleted as a part of #3814 but was missed. Therefore I've deleted it.
- It turns out that there was another oversight in #3814 in adjacent code (`ConfigContextQuerySetMixin`) in this same file; this was still annotating the queryset with config-context data unless `?exclude=config_context` was present in the request, even though the logic elsewhere has been changed to omit the config-context unless `?include=config_context` is present. The impact of this error was that all Device/VM REST API queries were still taking the performance hit of calculating the config context, even though it generally wasn't actually included in the API response. I fixed this logic here and, using django-debug-toolbar, confirmed that the config-context annotations are only included in the SQL query now when `?include=config_context` is explicitly specified in the request.
- Add some basic regression tests against this portion of the OpenAPI schema; it would be good to flesh out this set of tests as and when we fix other schema inconsistencies over time.

image

image

Note to @itdependsnetworks - it doesn't appear that this fixes the `face` filter to be enum-based as you reported was the case in 1.5.x - it's showing as an array-of-strings filter here:

image

Please feel free to file a followup issue if this one is considered problematic.

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/changes/4377.fixed
+++ b/changes/4377.fixed
@@ -1,0 +1,2 @@
+Fixed incorrect OpenAPI schema for filters available on Device and VirtualMachine REST API endpoints.
+Fixed incorrect logic for queryset annotation on Device and VirtualMachine REST API views.
--- a/nautobot/core/tests/test_openapi.py
+++ b/nautobot/core/tests/test_openapi.py
@@ -1,0 +1,64 @@
+from drf_spectacular.renderers import OpenApiYamlRenderer
+from drf_spectacular.settings import spectacular_settings
+
+import yaml
+
+from nautobot.core.testing import TestCase
+
+
+class OpenAPITest(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        generator_class = spectacular_settings.DEFAULT_GENERATOR_CLASS
+        generator = generator_class()  # TODO: in future we may want to specify an `api_version=...` here
+        schema = generator.get_schema(request=None, public=True)
+        # We probably could stop here but let's round-trip it through the YAML renderer just for the heck of it
+        renderer = OpenApiYamlRenderer()
+        cls.binary_output = renderer.render(schema, renderer_context={})
+        cls.yaml_output = cls.binary_output.decode("utf-8")
+        cls.schema = yaml.safe_load(cls.yaml_output)
+
+    def test_filter_boolean_type(self):
+        """
+        Test that a boolean filter is correctly represented as a boolean.
+
+        Testing for regression of https://github.com/nautobot/nautobot/issues/4377.
+        """
+        query_params = self.schema["paths"]["/dcim/devices/"]["get"]["parameters"]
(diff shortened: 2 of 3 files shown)


gsnider2195 (approved):

itdependsnetworks:
@glennmatthews I do not have consider `face` as problematic, I would defer to you, what should it be? I am just observing the behavior and I think we can all agree that it shouldn't change between minor versions, but I do not know what it should be in general?

glennmatthews:
> @glennmatthews I do not have consider `face` as problematic, I would defer to you, what should it be? I am just observing the behavior and I think we can all agree that it shouldn't change between minor versions, but I do not know what it should be in general?

It probably *should* be reported in the schema as an enum *for the "exact" variants of the filter* (`face`, `face__n`) but **not** for the others (`face__ic`, `face__isw`, `face__ire`, etc.). So I'm fine with an issue being opened for this but it'll be a separate (and likely larger) effort to address. See also #3029.

itdependsnetworks:
ok, just spent a few on this we are making the choiceset a callable  

https://github.com/nautobot/nautobot/blob/8e00b8f30ef63e13613a11d3dcdd197b65e09047/nautobot/utilities/choices.py#L6 

which get's caught here:

https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/contrib/django_filters.py#L227

Issue #4388, opened 2023-09-05
Config contexts applied to parent locations missing from rendered config context


glennmatthews:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0rc1 (next.demo.nautobot.com)

### Steps to Reproduce
1. Based on the demo data on next.demo.nautobot.com, create the following config-contexts:
    a. "americas", applicable to location Americas
    b. "united-states", applicable to location Americas -> United States
    c. "ATL01", applicable to location Americas -> United States -> ATL01
2. View the config context tab for any device in ATL01

### Expected Behavior

Rendered config context to include data from all three of these config contexts

### Observed Behavior

Although all three contexts appear as "Source Contexts" on the right, the rendered config context only includes ATL01 and not the other two:

![image](https://github.com/nautobot/nautobot/assets/5603551/2bee3c7e-9e20-44bc-93d5-149ed7e00434)

gsnider2195:
We need to make sure we have test coverage for this use case

PR #4417, opened 2023-09-07
Fixed a bug that makes ConfigContexts applied to parent locations missing from rendered config context of child location Devices/Virtual Machines.

HanlinMiao:
# Closes: #4388 
# What's Changed

The root cause of the issue is in `ConfigContextModelQuerySet._get_config_context_filters()`. The function only takes into account the location that the Device/VM is in and not its parent locations. Device and VirtualMachine's Location has its own ConfigContext and its parent Locations' ConfigContext, if any, should
also be applied. However, since moving from `mptt` to `django-tree-queries `https://github.com/nautobot/nautobot/issues/510, we lost the ability to query the ancestors for a particular tree node for subquery https://github.com/matthiask/django-tree-queries/issues/54.
So instead of constructing the location-related query in `ConfigContextModelQueryset._get_config_context_filters()` with `OuterRef` and `Subquery`, which is complicated if not impossible across multiple databases. We append the location query process in `ConfigContextModel.get_config_context()` as a patch.
Screen Shot 2023-09-07 at 5 13 18 PM

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/docs/user-guide/core-data-model/extras/configcontext.md
+++ b/nautobot/docs/user-guide/core-data-model/extras/configcontext.md
@@ -67,6 +67,9 @@ When the context data for a device at this location is rendered, the second, hig
 
 Data from the higher-weight context overwrites conflicting data from the lower-weight context, while the non-conflicting portion of the lower-weight context (the list of NTP servers) is preserved.
 
+!!! warning
+    ConfigContexts can be applied to parents and descendants of TreeModels such as Locations and RackGroups. The inheritance of ConfigContext will always be determined by the value of the weight attribute. You may see unexpected behavior if you have ConfigContext of the same weight applied to the parents and the descendants of TreeModels.
+
 ## Local Context Data
 
 Devices and virtual machines may also have a local config context defined. This local context will _always_ take precedence over any separate config context objects which apply to the device/VM. This is useful in situations where we need to call out a specific deviation in the data for a particular object.
--- a/nautobot/extras/models/models.py
+++ b/nautobot/extras/models/models.py
@@ -205,12 +205,12 @@ class ConfigContextModel(models.Model, ConfigContextSchemaValidationMixin):
             location_config_context_queryset = ConfigContext.objects.none()
             if self._meta.model_name == "device":
                 location_config_context_queryset = ConfigContext.objects.filter(
-                    Q(locations__in=self.location.ancestors(include_self=True))
+                    locations__in=self.location.ancestors(include_self=True)
                 ).distinct()
             else:
                 if self.cluster and self.cluster.location:
                     location_config_context_queryset = ConfigContext.objects.filter(
-                        Q(locations__in=self.cluster.location.ancestors(include_self=True))
+                        locations__in=self.cluster.location.ancestors(include_self=True)
                     ).distinct()
 
             # Annotation has keys "weight" and "name" (used for ordering) and "data" (the actual config context data)
--- a/nautobot/extras/models/models.py
+++ b/nautobot/extras/models/models.py
@@ -204,12 +204,12 @@ class ConfigContextModel(models.Model, ConfigContextSchemaValidationMixin):
             # We append the missing parent location query here as a patch.
             if self._meta.model_name == "device":
                 location_config_context_queryset = ConfigContext.objects.filter(
-                    Q(locations__in=self.location.ancestors(include_self=True)) | Q(locations=None),
+                    Q(locations__in=self.location.ancestors(include_self=True))
                 ).distinct()
             else:
                 if self.cluster:
                     location_config_context_queryset = ConfigContext.objects.filter(
-                        Q(locations__in=self.cluster.location.ancestors(include_self=True)) | Q(locations=None),
+                        Q(locations__in=self.cluster.location.ancestors(include_self=True))
                     ).distinct()
 
             # Annotation has keys "weight" and "name" (used for ordering) and "data" (the actual config context data)
--- a/nautobot/extras/models/models.py
+++ b/nautobot/extras/models/models.py
@@ -9,6 +9,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.core.serializers.json import DjangoJSONEncoder
 from django.core.exceptions import ValidationError
 from django.db import models
+from django.db.models import Q
 from django.http import HttpResponse
 from graphene_django.settings import graphene_settings
 from graphql import get_default_backend
@@ -196,7 +197,24 @@ class ConfigContextModel(models.Model, ConfigContextSchemaValidationMixin):
             config_context_data = ConfigContext.objects.get_for_object(self).values_list("data", flat=True)
         else:
             config_context_data = self.config_context_data or []
+            # Device and VirtualMachine's Location has its own ConfigContext and its parent Locations' ConfigContext, if any, should
+            # also be applied. However, since moving from mptt to django-tree-queries https://github.com/nautobot/nautobot/issues/510,
+            # we lost the ability to query the ancestors for a particular tree node for subquery https://github.com/matthiask/django-tree-queries/issues/54.
+            # So instead of constructing the location related query in ConfigContextModelQueryset._get_config_context_filters(), which is complicated across databases
+            # We append the missing parent location query here as a patch.
+            if self._meta.model_name == "device":
+                location_config_context_queryset = ConfigContext.objects.filter(
+                    Q(locations__in=self.location.ancestors(include_self=True)) | Q(locations=None),
+                ).distinct()
+            else:
+                if self.cluster:
+                    location_config_context_queryset = ConfigContext.objects.filter(
+                        Q(locations__in=self.cluster.location.ancestors(include_self=True)) | Q(locations=None),
+                    ).distinct()
+
             # Annotation has keys "weight" and "name" (used for ordering) and "data" (the actual config context data)
+            for cc in location_config_context_queryset:
+                config_context_data.append({"data": cc.data, "name": cc.name, "weight": cc.weight})
             config_context_data = [
                 c["data"] for c in sorted(config_context_data, key=lambda k: (k["weight"], k["name"]))
             ]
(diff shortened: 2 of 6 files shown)


nautobot/extras/models/models.py
```diff
@@ -196,7 +197,24 @@ def get_config_context(self):
...
         else:
             config_context_data = self.config_context_data or []
+            # Device and VirtualMachine's Location has its own ConfigContext and its parent Locations' ConfigContext, if any, should
+            # also be applied. However, since moving from mptt to django-tree-queries https://github.com/nautobot/nautobot/issues/510,
+            # we lost the ability to query the ancestors for a particular tree node for subquery https://github.com/matthiask/django-tree-queries/issues/54.
+            # So instead of constructing the location related query in ConfigContextModelQueryset._get_config_context_filters(), which is complicated across databases
+            # We append the missing parent location query here as a patch.
+            if self._meta.model_name == "device":
+                location_config_context_queryset = ConfigContext.objects.filter(
+                    Q(locations__in=self.location.ancestors(include_self=True)) | Q(locations=None),
```
gsnider2195, line 207: Why the `| Q(locations=None)`? This is causing config contexts with no location to apply to all devices. I created a config context with tenant=xyz and no location and it is being rendered for a device with a location and no tenant

HanlinMiao: It seems like all the constructed queries from `nautobot/extras/queryset.py` in line 41 have this `Q(locations=None)`. So I thought it was necessary to include it.

gsnider2195: I'm actually seeing all contexts rendered even if the location is completely different. Here's my location hierarchy:

- Campus-01
  - Building-02
    - Floor-03
      - Room-05
        - Aisle-13
        - Aisle-06

My device is in Aisle-13 but it's rendering a context applied to Aisle-06

image

image

HanlinMiao: The latest commit should fix that issue if you want to test it out on your end.

nautobot/extras/models/models.py
```diff
@@ -196,7 +197,24 @@ def get_config_context(self):
             config_context_data = ConfigContext.objects.get_for_object(self).values_list("data", flat=True)
         else:
             config_context_data = self.config_context_data or []
+            # Device and VirtualMachine's Location has its own ConfigContext and its parent Locations' ConfigContext, if any, should
+            # also be applied. However, since moving from mptt to django-tree-queries https://github.com/nautobot/nautobot/issues/510,
+            # we lost the ability to query the ancestors for a particular tree node for subquery https://github.com/matthiask/django-tree-queries/issues/54.
+            # So instead of constructing the location related query in ConfigContextModelQueryset._get_config_context_filters(), which is complicated across databases
+            # We append the missing parent location query here as a patch.
```
gsnider2195, line 204: Lets make we document this in `ConfigContextModelQueryset._get_config_context_filters()`

nautobot/extras/querysets.py
```diff
@@ -127,10 +127,11 @@ def _get_config_context_filters(self):
                 (Q(device_redundancy_groups=OuterRef("device_redundancy_group")) | Q(device_redundancy_groups=None)),
                 Q.AND,
             )
-            base_query.add((Q(locations=OuterRef("location")) | Q(locations=None)), Q.AND)
-
+            # This is necessary to prevent location related config context to be applied now.
```
gsnider2195, line 135:
```suggestion
            # This is necessary to prevent location related config context to be applied now. The location hierarchy cannot be processed by the database and must be added by `ConfigContextModel.get_config_context`
```

nautobot/extras/models/models.py
```diff
@@ -196,7 +197,24 @@ def get_config_context(self):
...
+            if self._meta.model_name == "device":
+                location_config_context_queryset = ConfigContext.objects.filter(
+                    Q(locations__in=self.location.ancestors(include_self=True))
+                ).distinct()
+            else:
+                if self.cluster:
+                    location_config_context_queryset = ConfigContext.objects.filter(
+                        Q(locations__in=self.cluster.location.ancestors(include_self=True))
```
gsnider2195, line 212: `Location` is not required for `Cluster` so this could raise an `AttributeError`

nautobot/extras/models/models.py
HanlinMiao: Did that in this commit https://github.com/nautobot/nautobot/pull/4417/commits/24f41c70e1ccfc7934c384f89c1d63078179b18d

gsnider2195:
I'm getting some unexpected behavior. I have a location and parent location that both provide a config context key ("location" in this case). The child location should take priority over the parent location, but that isn't happening here.

HanlinMiao:
> I'm getting some unexpected behavior. I have a location and parent location that both provide a config context key ("location" in this case). The child location should take priority over the parent location, but that isn't happening here.

What kind of behavior are you seeing, both "location" keys show up in the device config context?

gsnider2195:
> > I'm getting some unexpected behavior. I have a location and parent location that both provide a config context key ("location" in this case). The child location should take priority over the parent location, but that isn't happening here.
> 
> What kind of behavior are you seeing, both "location" keys show up in the device config context?

Just the key for the parent but I'm assuming it could be random depending on the default order of locations?

HanlinMiao:
> > > I'm getting some unexpected behavior. I have a location and parent location that both provide a config context key ("location" in this case). The child location should take priority over the parent location, but that isn't happening here.
> > 
> > 
> > What kind of behavior are you seeing, both "location" keys show up in the device config context?
> 
> Just the key for the parent but I'm assuming it could be random depending on the default order of locations?

I think it might be the `weight` of the data? on line 224 for `get_config_context`
# Compile all config data, overwriting lower-weight values with higher-weight values where a collision occurs
        data = OrderedDict()
        for context in config_context_data:
            data = deepmerge(data, context)

gsnider2195:
Mine have the same weight so in the case I would expect the more specific location to win. You may be able to sort the location query on tree_depth ascending and it might just work?

HanlinMiao:
I did some experiments here:

```
location_config_context_queryset = ConfigContext.objects.filter(
                    Q(locations__in=self.location.ancestors(include_self=True))
                )
print(location_config_context_queryset.order_by("locations"))
print(location_config_context_queryset.order_by("-locations"))
>> <ConfigContextQuerySet [<ConfigContext: americas>, <ConfigContext: atl-01>, <ConfigContext: united-states>]>
>> <ConfigContextQuerySet [<ConfigContext: united-states>, <ConfigContext: atl-01>, <ConfigContext: americas>]>
```
They are all the same weight and atl-01 is the most specific, but its never first.

gsnider2195:
@bryanculver was saying order_by doesn't work for tree models so we'll have to sort it in python

nautobot/extras/models/models.py
```diff
@@ -196,7 +197,25 @@ def get_config_context(self):
...
             config_context_data = self.config_context_data or []
+            # Device and VirtualMachine's Location has its own ConfigContext and its parent Locations' ConfigContext, if any, should
+            # also be applied. However, since moving from mptt to django-tree-queries https://github.com/nautobot/nautobot/issues/510,
+            # we lost the ability to query the ancestors for a particular tree node for subquery https://github.com/matthiask/django-tree-queries/issues/54.
+            # So instead of constructing the location related query in ConfigContextModelQueryset._get_config_context_filters(), which is complicated across databases
+            # We append the missing parent location query here as a patch.
+            location_config_context_queryset = ConfigContext.objects.none()
+            if self._meta.model_name == "device":
+                location_config_context_queryset = ConfigContext.objects.filter(
+                    Q(locations__in=self.location.ancestors(include_self=True))
```
gsnider2195, line 208:
```suggestion
                    locations__in=self.location.ancestors(include_self=True)
```

nautobot/extras/models/models.py
```diff
@@ -196,7 +197,25 @@ def get_config_context(self):
...
+                ).distinct()
+            else:
+                if self.cluster and self.cluster.location:
+                    location_config_context_queryset = ConfigContext.objects.filter(
+                        Q(locations__in=self.cluster.location.ancestors(include_self=True))
```
gsnider2195, line 213:
```suggestion
                        locations__in=self.cluster.location.ancestors(include_self=True)
```

gsnider2195:
> Mine have the same weight so in that case I would expect the more specific location to win. You may be able to sort the location query on tree_depth ascending and it might just work?

Since the filter is multivalue I don't think there's any expectation that we can sort config contexts by anything BUT weight.

nautobot/docs/user-guide/core-data-model/extras/configcontext.md
```diff
@@ -67,6 +67,9 @@ When the context data for a device at this location is rendered, the second, hig
 
 Data from the higher-weight context overwrites conflicting data from the lower-weight context, while the non-conflicting portion of the lower-weight context (the list of NTP servers) is preserved.
 
+!!! warning
+    ConfigContexts can be applied to parents and descendants of TreeModels such as Locations and RackGroups. The inheritance of ConfigContext will always be determined by the value of the weight attribute. You may see unexpected behavior if you have ConfigContext of the same weight applied to the parents and the descendants of TreeModels.
```
gsnider2195, line 71:
```suggestion
    ConfigContexts can be applied to parents and descendants of TreeModels such as Locations and RackGroups. The inheritance of ConfigContext will always be determined by the value of the weight attribute. You may see unexpected behavior if you have ConfigContexts of the same weight applied to TreeModel parents and their descendants.
```

nautobot/extras/querysets.py
```diff
@@ -83,6 +83,9 @@ def annotate_config_context_data(self):
 
         Order By clause in Subquery is not guaranteed to be respected within the aggregated JSON array, which is why
         we include "weight" and "name" into the result so that we can sort it within Python to ensure correctness.
+
+        TODO This method is broken because of the reasons stated in _get_config_context_filters()
```
gsnider2195, line 87:
```suggestion
        TODO This method does not accurately reflect location inheritance because of the reasons stated in _get_config_context_filters()
```

gsnider2195 (approved):
:shipit:

PR #4418, opened 2023-09-09
Added `netutilsparser` to `network_driver`

itdependsnetworks:
# Closes: DNE
# What's Changed

Add netutilsparser

image

# TODO

- [x] Unit, Integration Tests


glennmatthews:
Please target this PR to the `develop` branch rather than `ltm/1.6`. We should always be making code changes in `develop` first then making the decision as to whether it's justifiable to back-port them to the LTM branch after they've been accepted and merged.

nautobot/docs/release-notes/version-1.6.md
```diff
@@ -43,7 +43,7 @@ The following is an example of the data that is sent:
...
+The [Platform](../user-guide/core-data-model/dcim/platform.md) model has been enhanced to include a `network_driver` database field and a `network_driver_mappings` derived property based on the [`netutils`](https://netutils.readthedocs.io/en/latest/) library. For example, if you set a Platform to have a `network_driver` value of `"cisco_ios"`, the `platform.network_driver_mappings` property will return a dictionary containing `ansible`, `hier_config`, `napalm`, `netmiko`, `netutils_parser`, `ntc_templates`, `pyats`, `pyntc`, and `scrapli` keys corresponding to this entry. These properties can be referenced via the REST API and GraphQL to assist in developing and maintaining Apps, Jobs, or third-party code that interact with devices by using any of these libraries.
```
glennmatthews, line 46: This PR is for 2.0 not for 1.6, and this specific change is already covered by `changes/4396.added`.

```suggestion
The [Platform](../user-guide/core-data-model/dcim/platform.md) model has been enhanced to include a `network_driver` database field and a `network_driver_mappings` derived property based on the [`netutils`](https://netutils.readthedocs.io/en/latest/) library. For example, if you set a Platform to have a `network_driver` value of `"cisco_ios"`, the `platform.network_driver_mappings` property will return a dictionary containing `ansible`, `hier_config`, `napalm`, `netmiko`, `ntc_templates`, `pyats`, `pyntc`, and `scrapli` keys corresponding to this entry. These properties can be referenced via the REST API and GraphQL to assist in developing and maintaining Apps, Jobs, or third-party code that interact with devices by using any of these libraries.
```

pyproject.toml
```diff
@@ -109,7 +109,7 @@ netaddr = "~0.8.0"
 # Library of network related utilities including network driver mappings and Jinja Filters
 # Note: netutils is limited in scope, dependencies, and observes semver, as such
 #       we permit a looser (^) version constraint here.
-netutils = "^1.5.0"
+netutils = "^1.6.0"
```
glennmatthews, line 112: Let's add a `changes/4418.dependencies` documenting this change.

changes/4396.added
glennmatthews, line 1: Should be `changes/4418.added`.

nautobot/dcim/utils.py
```diff
@@ -98,6 +99,7 @@ def get_all_network_driver_mappings():
         ("hier_config", HIERCONFIG_LIB_MAPPER_REVERSE),
         ("napalm", NAPALM_LIB_MAPPER_REVERSE),
         ("netmiko", NETMIKO_LIB_MAPPER_REVERSE),
+        ("netutils_parser", NETUTILSPARSER_LIB_MAPPER_REVERSE),
```
glennmatthews, line 102: Let's document this in `nautobot/docs/user-guide/core-data-model/dcim/platform.md` as a 2.0 addition to the `Platform` model functionality.

nautobot/docs/release-notes/version-1.6.md
itdependsnetworks: > This PR is for 2.0 not for 1.6, and this specific change is already covered by changes/4396.added.

I am not sure I follow, this was for 1.6, you asked me to set to develop and not to ltm branch, are you saying "we will not accept this for 1.6"? If so, fair enough, would be good to understand why though so in the future I can better understand what does/does not get into ltm.

glennmatthews: `develop` is for 2.0 at this time, so this PR is currently making code changes to the upcoming 2.0 release. If there's a use case for including this feature in a 1.6.x patch release as well, we can deal with that after this PR is merged into `develop`, either as a cherry-pick effort by the Nautobot core team or by having someone open a second PR.

https://docs.nautobot.com/projects/core/en/stable/development/#maintenance-release-ltm is the current documentation regarding the plan for maintenance of the LTM train - short version is that LTM will be guaranteed security and data-loss bug fixes, everything else will be case-by-case. This PR is technically speaking a new feature rather than a bug fix, so personally I think that makes it less likely to be accepted for backporting to a 1.6.x patch release per our usual SemVer expectations, but all things are possible if there's a strong enough use case or need.

glennmatthews (approved):

itdependsnetworks:
Safe to assume you will have the poetry conflict? This is a moving target if someone from core doesn't handle and then merge instantly.

PR #4419, opened 2023-09-10
Updates for NavMenu, startplugin, and docs

itdependsnetworks:
# Closes: DNW
# What's Changed

Added documentation on `natobot.apps` import locations.
Added documentation about the supported public interfaces
Fixed the documentation referring to NavMenu removal.
Fixed a missing import of `NavMenuButton` into `nautobot.apps`.
Fixed plugin template creation to point to `nautobot.apps`.

# TODO


nautobot/docs/user-guide/administration/upgrading/from-v1/tables/v2-code-nautobot-app-location.yaml
glennmatthews, line 1: I'm not sure how much value this table adds compared to the existing Apps API reference (https://docs.nautobot.com/projects/core/en/latest/code-reference/nautobot/apps/__init__/ and related pages) which is autogenerated from source and so will always be more complete and up-to-date than this static table.

I don't see the value in documenting the internal location of each of these functions/classes since that is explicitly subject to change in future releases. This would be better structured (if even desirable) as:

```yaml
---
- Function or Class: "nautobot_database_ready"
  Public Interface: "nautobot.apps.nautobot_database_ready"
- Function or Class: "ConstanceConfigItem"
  Public Interface: "nautobot.apps.ConstanceConfigItem"
- Function or Class: "NautobotAppConfig"
  Public Interface: "nautobot.apps.NautobotAppConfig"
```

or even just:


```yaml
---
- Public Interface: "nautobot.apps.nautobot_database_ready"
- Public Interface: "nautobot.apps.ConstanceConfigItem"
- Public Interface: "nautobot.apps.NautobotAppConfig"
```

but that again comes back to "does this really add value compared to either linking directly to the source or to the existing App developer docs?"

changes/4419.fixed
```diff
@@ -0,0 +1,3 @@
+Fixed the documentation referring to NavMenu removal.
```
glennmatthews, line 1:
```suggestion
Removed some incorrect content from the documentation about nav menu changes that were reverted during 2.0 development.
```

changes/4419.fixed
```diff
@@ -0,0 +1,3 @@
...
+Fixed a missing import of `NavMenuButton` into `nautobot.apps`.
```
glennmatthews, line 2:
```suggestion
Re-added `NavMenuButton` back into `nautobot.apps.ui` module.
```

nautobot/docs/user-guide/administration/upgrading/from-v1/upgrading-from-nautobot-v1.md
```diff
@@ -36,6 +36,10 @@ All pre-migration checks passed.
 - Nautobot no longer uses or supports the use of `django-mptt`.
 - Nautobot no longer uses or supports the use of `django-rq`.
 
+## App Developer Interface
+
+All imports for app developers should now be sourced from `nautobot.apps.*` or provided models, as this is the supported public interface. This will help ensure proper SemVer is kept in tact and provide a place to add new features without breaking existing integrations.
```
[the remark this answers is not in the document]
itdependsnetworks: > I'm not sure we're at a point yet where we can dogmatically assert this

This seems like table stakes for 2.0, no?

nautobot/docs/user-guide/administration/upgrading/from-v1/tables/v2-code-nautobot-app-location.yaml
itdependsnetworks: > I don't see the value in documenting the internal location of each of these functions/classes since that is explicitly subject to change in future releases.

Yes, sorry, I didn't finish this thought. This should be combined with the prior table, I realized that about half way through. Honestly, it was very difficult to try and figure out where everything was. 

I have this code
```python
    tenant_group = utilities_forms.DynamicModelMultipleChoiceField(queryset=TenantGroup.objects.all(), required=False)
```

I want to lookup where I should import that from now, I search `DynamicModelMultipleChoiceField` in https://docs.nautobot.com/projects/core/en/v2.0.0-rc.2/development/apps/migration/code-updates, I do not get a relevant hit. I figure it out by understanding that:

* I have an import of -> `import nautobot.utilities.forms as utilities_forms`
* I search on the migration guide, and I see is now -> `nautobot.core.forms`
* I then think to myself... do I want to do this again in the future? I say no, let me find the correct `nautobot.core` path
* I search for `DynamicModelMultipleChoiceField`, which has 4 results I need to change

With this, at a minimum, I can just `ctrl+F` and find my class/function vs those 4-8 complicated steps (depending if I click on the right page) I had to figure out vs everything in the migration guide.

I would argue that both the new and current tables are lacking in what i need as migrating the app. From my perspective, and as said in the other comment, it is table stakes to have this correct imho.

glennmatthews: Did searching the docs not find https://docs.nautobot.com/projects/core/en/v2.0.0-rc.2/code-reference/nautobot/apps/forms/#nautobot.apps.forms.DynamicModelMultipleChoiceField?

nautobot/docs/user-guide/administration/upgrading/from-v1/upgrading-from-nautobot-v1.md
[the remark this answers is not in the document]
glennmatthews: Perhaps it should be, but it currently isn't being tracked as such, will likely be a series of incremental improvements throughout 2.x.

nautobot/docs/user-guide/administration/upgrading/from-v1/tables/v2-code-nautobot-app-location.yaml
itdependsnetworks: > I search for DynamicModelMultipleChoiceField, which has 4 results I need to change

Yes, it found 4 results I needed to compute which one was correct. 

Then there is sometimes 0 results

image

Or even more results

image

itdependsnetworks: Again, it is not to say it can't be done, it is just trying to make it as easy as possible.

nautobot/core/templates/plugin_template/models.py-tpl
```diff
@@ -2,7 +2,7 @@
```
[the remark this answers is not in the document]
itdependsnetworks: agreed, but will presume that is outside the scope of this PR.

nautobot/docs/development/core/index.md
```diff
@@ -69,6 +69,30 @@ Nautobot adheres to the Semantic Versioning ("SemVer") strategy, which gives us
 
 For more information, please visit [SemVer.org](https://semver.org/).
 
+#### Public Interface
```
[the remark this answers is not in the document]
itdependsnetworks: This is under `Release Management -> Versioning -> Public Interface` to define what is SemVer'd vs not. Taking inspiration from here: https://regret.readthedocs.io/en/stable/before-you-deprecate/#document-your-public-api, I think this is the correct place for this. 

This is simply saying, "hey, not every docstring and private function is semver'd, we reserve the right to change the function signature on a private function during every patch release, which to me belongs under release management.

nautobot/docs/user-guide/administration/upgrading/from-v1/tables/v2-code-nautobot-app-location.yaml
itdependsnetworks: After looking at it, I think the current strategy is actually the best. The only thing I could find explicitly matching was `TaggedModelSerializerMixin`, we could match on the `(all)`, but then would have to list all those methods as well

nautobot/docs/user-guide/administration/upgrading/from-v1/upgrading-from-nautobot-v1.md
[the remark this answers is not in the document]
itdependsnetworks: @lampwins I think we should have this for the 2.0 migration, otherwise we will be doing this again when we want to do the same thing and can't update anything until 3.0 again, plus it makes it should be easier for the app developers to know where to go to find things.

nautobot/docs/development/apps/migration/code-updates.md
```diff
@@ -2,11 +2,16 @@
 
 ## Update Code Import Locations
 
-Most changes in code location arise from the merging of the `nautobot.utilities` module into the `nautobot.core` module.
+Most changes in code location arise from the merging of the `nautobot.utilities` module into the `nautobot.core` module. Note that in most cases, the recommended approach for Jobs and Apps is to import from `nautobot.apps` rather than `nautobot.core` (see below).
 
 ??? info "Full table of code location changes"
     {data-table user-guide/administration/upgrading/from-v1/tables/v2-code-location-changes.yaml}
 
+Other than modes, most other imports from Nautobot should be imported from `nautobot.apps.*` as described here.
```
glennmatthews, line 10:
```suggestion
Other than models, most other imports from Nautobot should be imported from `nautobot.apps.*` as described here.
```

nautobot/docs/development/core/best-practices.md
```diff
@@ -232,7 +232,7 @@ The following best practices must be considered when establishing new `FilterSet
 
 ### Mapping Model Fields to Filters
 
-- FilterSets **must** inherit from `nautobot.extras.filters.NautobotFilterSet` (which inherits from `nautobot.core.filters.BaseFilterSet`)
+- FilterSets **must** inherit from `nautobot.apps.filters.NautobotFilterSet`
```
glennmatthews, line 235: Please revert all `nautobot.apps` references in this file.

nautobot/docs/development/core/index.md
```diff
@@ -69,6 +69,30 @@ Nautobot adheres to the Semantic Versioning ("SemVer") strategy, which gives us
...
+
+The guiding principals behind Nautobot's SemVer strategy is scoped to adhering to the prescribed public interface. This includes both the application operators and application developer. Anything not explicitly covered here is presumed to not be covered under SemVer.
+
+* The REST API interface within a Version, will not change, but may add to
```
glennmatthews, line 76:
```suggestion
* Behavior of a given X.Y version of the REST API. New versions of the REST API may be added, which may introduce changes to
```

nautobot/docs/development/core/index.md
```diff
@@ -69,6 +69,30 @@ Nautobot adheres to the Semantic Versioning ("SemVer") strategy, which gives us
...
+    * URL structures
+    * Expected request and response formats
+    * Behavior of each endpoint
+    * Status and error codes
+    * The HTTP verbs
+* Graphql data structures
```
glennmatthews, line 82:
```suggestion
* GraphQL data structures
```

nautobot/docs/development/core/index.md
```diff
@@ -69,6 +69,30 @@ Nautobot adheres to the Semantic Versioning ("SemVer") strategy, which gives us
...
+* URL structures for primary navigation and feature access
+* Any Django model and their relationships for models provided by Nautobot
```
glennmatthews, line 84: I don't have specific verbiage to suggest as an alternative, but this feels a bit vague to me. We absolutely might introduce new fields or new behaviors to Django models.

nautobot/docs/development/core/index.md
```diff
@@ -69,6 +69,30 @@ Nautobot adheres to the Semantic Versioning ("SemVer") strategy, which gives us
...
+* Functions, classes, and modules in `nautobot.apps.*` and their associated signature
+* Any management commands provided by Nautobot
+* Documented configuration options
+
+While anything not explicitly stated above should be not be considered scoped to SemVer, the following should be strictly avoided.
+
+* Functions or methods that start with an underscore (e.g., `_private_func`)
+* Any functions, classes, and modules **not** in `nautobot.apps.*`.
+* Any modules, classes, functions, or feature that are explicitly marked as "internal" or "experimental" in the documentation
+* Any features or functionalities that aren't part of the official documentation should not be relied upon.
```
glennmatthews, line 94: This bullet list, at least, still feels like it belongs in the `apps` and `jobs` development documentation rather than here. Maybe at a minimum those docs should cross-link to this section?

nautobot/docs/development/core/index.md
```diff
@@ -69,6 +69,30 @@ Nautobot adheres to the Semantic Versioning ("SemVer") strategy, which gives us
...
+    * Status and error codes
+    * The HTTP verbs
+* Graphql data structures
+* URL structures for primary navigation and feature access
+* Any Django model and their relationships for models provided by Nautobot
+* Functions, classes, and modules in `nautobot.apps.*` and their associated signature
+* Any management commands provided by Nautobot
+* Documented configuration options
+
+While anything not explicitly stated above should be not be considered scoped to SemVer, the following should be strictly avoided.
```
glennmatthews, line 89:
```suggestion
While anything not explicitly stated above should be not be considered scoped to SemVer, the following should be strictly avoided when developing non-core Nautobot Apps and Jobs.
```

nautobot/docs/user-guide/administration/upgrading/from-v1/upgrading-from-nautobot-v1.md
glennmatthews, line 41:
```suggestion
All imports for app developers should now be sourced from `nautobot.apps.*` or provided models, as this is the supported public interface. This will help ensure proper SemVer is kept intact and provide a place to add new features without breaking existing integrations.
```

nautobot/docs/development/core/index.md
itdependsnetworks: i added an info statement below, does that help?

bryanculver:
@itdependsnetworks are you OK if we address this via #4459 and #4460?

itdependsnetworks:
I think we need an equivalent of the v2-code-nautobot-app-location.yaml file, something that is simple and I can find with a ctrl+F, is that happening? If so can close this in favor of those other two.


Figured getting the public interface in quickly would not be easy, so if need to hold off on that as well, at least the conversation has started.

nautobot/docs/user-guide/administration/upgrading/from-v1/upgrading-from-nautobot-v1.md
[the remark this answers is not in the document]
glennmatthews: Addressed in #4533

nautobot/docs/user-guide/administration/upgrading/from-v1/tables/v2-code-nautobot-app-location.yaml
```diff
@@ -0,0 +1,207 @@
+---
+- Nautobot Location: "nautobot.core.signals.nautobot_database_ready"
+  Public Interface: "nautobot.apps.nautobot_database_ready"
```
glennmatthews, line 3: This is going to be a pain to keep updated - every time we add something new to nautobot.apps, or relocate it behind the scenes, we will need to update this? I'm just not sure the value is here.

lampwins: @itdependsnetworks Is the intent that this data table is used to build a pylint rule? If so, since we have moved on in time, perhaps this one just moves solely into pylint?

itdependsnetworks: > I'm just not sure the value is here.

There is well over 100 known nautobot apps that should be able to benefit. 

> perhaps this one just moves solely into pylint?

Seems like a reasonable solution, I do think there should be a link to the list, as my process was to just pull up this file and do a ctrl-f and that was rather helpful.

itdependsnetworks: I missed this, thank you sir!!! (and no one would ever say that about you!!)

glennmatthews (approved):

PR #4420, opened 2023-09-10
Update jobs migration docs

itdependsnetworks:
# Closes: DNE
# What's Changed

NA

# TODO

NA


changes/4420.changed
```diff
@@ -0,0 +1 @@
+Update jobs migration docs to reflect alternate method of job params migration.
```
glennmatthews, line 1:
```suggestion
Updated jobs migration docs to reflect alternate method of job params migration.
```

I wonder - should this probably be a `.added` rather than a `.changed`?

glennmatthews (approved):

Issue #4010, opened 2023-06-28
Specifying an invalid parent when creating an IP address should result in an error


lampwins:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0b1
* Python version: 3.10
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Have a namespace with 10.0.0.0/8 and 192.168.0.0/16 as networks
2. Use the API or ORM to create an IP Address of 192.168.0.1/16 but specify a parent of 10.0.0.0/8

### Expected Behavior
An error should be returned saying that the request to use the parent cannot be honored. Context that an available parent does exist and what it is, should be included in the error.

### Observed Behavior
The parenting algorithm will silently discard the requested parent and assign the valid parent. In the contrived use case described above, one could argue this is a net good thing, but consider much larger data sets and this then becomes an issue with the principal of least surprise. We should rely on the explicit nature of the user's request, and validate it on that basis alone.

bryanculver:
@lampwins Would it be acceptable for`parent` to be removed from an accepted input when creating an IP address (you must always pass a Namespace)?

lampwins:
Yes, that is fine for now.

lampwins:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Nautobot-Core-Product-64a8783b7a3f4b1225973a0c/issues/nautobot/nautobot/4010?planning-poker) with Zenhub @bryanculver

PR #4432, opened 2023-09-12
Fixed error not being raised when an invalid parent is provided 

timizuoebideri1:
# Closes: #4010 
# What's Changed

1.  Raise an exception in `IPAddress.clean` if `parent` is specified but it's not the closest `prefix` to the `IPAddress`.
2. Changed `IPAddressSerializer.validate` to use "IPAddress.clean" to eliminate code duplication.

# TODO

- [ ] Unit, Integration Tests


bryanculver (approved):
@timizuoebideri1 please review. It seems with the model save changes something broke in the other test cases.

bryanculver (dismissed):
@timizuoebideri1 please review. It seems with the model save changes something broke in the other test cases.

nautobot/ipam/api/serializers.py
```diff
@@ -265,6 +265,16 @@ def validate(self, data):
         if self.instance is None and not any([namespace, parent]):
             raise ValidationError({"__all__": "One of parent or namespace must be provided"})
 
+        # DRF does not call Model clean method by default; however, to avoid repeating validation checks in Model clean, call clean method here.
```
timizuoebideri1, line 268:
```suggestion
        # DRF does not call Model clean method by default; however, to avoid rewriting validation checks in Model clean, call clean method here.
```

nautobot/ipam/api/serializers.py
```diff
@@ -265,6 +265,16 @@ def validate(self, data):
...
+        instance = self.instance or self.Meta.model(**data)
+        try:
+            instance.clean()
+        except ValidationError as e:
+            raise serializers.ValidationError(e.args[0])
+
+        # `dns_name` and `parent` were been modified in the Model clean method, pass those modified data into `data`
```
timizuoebideri1, line 275:
```suggestion
        # `dns_name` and `parent` has been modified in the Model clean method, pass those modified data into `data`
```

nautobot/ipam/models.py
```diff
@@ -1079,30 +1097,26 @@ def clean(self):
         if self.type == choices.IPAddressTypeChoices.TYPE_SLAAC and self.ip_version != 6:
             raise ValidationError({"type": "Only IPv6 addresses can be assigned SLAAC type"})
 
+        # Validate `parent` can be used as a parent for this ipaddress
```
timizuoebideri1, line 1100:
```suggestion
        # Validate `parent` can be used as the parent for this ipaddress
```

nautobot/ipam/models.py
```diff
@@ -1079,30 +1097,26 @@ def clean(self):
...
-            .get_closest_parent(self.host, include_self=True)
-        )
-
         # 3.0 TODO: uncomment the below to enforce this constraint
         # if self.parent.type != choices.PrefixTypeChoices.TYPE_NETWORK:
         #     err_msg = f"IP addresses cannot be created in {self.parent.type} prefixes. You must create a network prefix first."
         #     raise ValidationError({"address": err_msg})
 
+        # validated_save is not always called especially in Test environment;
+        # but we still need to set the self.parent attr in this cases
```
timizuoebideri1, line 1117:
```suggestion
        # but we still need to set the `self.parent` attr in these cases
```

nautobot/ipam/api/serializers.py
```diff
@@ -265,6 +265,16 @@ def validate(self, data):
         if self.instance is None and not any([namespace, parent]):
             raise ValidationError({"__all__": "One of parent or namespace must be provided"})
 
+        # DRF does not call Model clean method by default; however, to avoid rewriting validation checks in Model clean, call clean method here.
```
glennmatthews, line 268: DRF doesn't call model clean by default, but Nautobot's `ValidatedModelSerializer` (parent of `NautobotModelSerializer` and therefore of this serializer) does do so. Probably a more correct fix here would be to call `super.validate(data)` at some point, rather than calling `instance.clean()` directly?

nautobot/ipam/models.py
```diff
@@ -1043,8 +1043,7 @@ class Meta:
     def __init__(self, *args, **kwargs):
         address = kwargs.pop("address", None)
         namespace = kwargs.pop("namespace", None)
-        # We don't want users providing their own parent since it will be derived automatically.
-        parent = kwargs.pop("parent", None)
+        parent = kwargs.get("parent", None)
```
glennmatthews, line 1046: Why the change from `pop` to `get`? Does the comment about not allowing users to directly set `parent` no longer apply because of the `clean` changes?

nautobot/ipam/models.py
```diff
@@ -1066,6 +1065,25 @@ def _deconstruct_address(self, address):
 
     natural_key_field_names = ["parent__namespace", "host"]
 
+    def _get_closest_parent(self):
+        if self.present_in_database:
+            namespace = self._namespace or self.parent.namespace
```
glennmatthews, line 1070: I can't see a case in `__init__()` where parent is non-None but self._namespace is None - is the "or" actually needed here?

nautobot/ipam/models.py
```diff
@@ -1066,6 +1065,25 @@ def _deconstruct_address(self, address):
...
+        elif self._namespace is None:
+            raise ValidationError({"parent": "Namespace could not be determined."})
```
glennmatthews, line 1072: When would this error be triggered? I don't see it in the tests.

nautobot/ipam/models.py
```diff
@@ -1066,6 +1065,25 @@ def _deconstruct_address(self, address):
...
+            namespace = self._namespace
+
+        try:
+            closes_parent = (
+                Prefix.objects.filter(namespace=namespace)
+                # 3.0 TODO: disallow IPAddress from parenting to a TYPE_POOL prefix, instead pick closest TYPE_NETWORK
+                # .exclude(type=choices.PrefixTypeChoices.TYPE_POOL)
+                .get_closest_parent(self.host, include_self=True)
+            )
+            return closes_parent
```
glennmatthews, line 1083:
```suggestion
            closest_parent = (
                Prefix.objects.filter(namespace=namespace)
                # 3.0 TODO: disallow IPAddress from parenting to a TYPE_POOL prefix, instead pick closest TYPE_NETWORK
                # .exclude(type=choices.PrefixTypeChoices.TYPE_POOL)
                .get_closest_parent(self.host, include_self=True)
            )
            return closest_parent
```

nautobot/ipam/models.py
```diff
@@ -1079,30 +1097,26 @@ def clean(self):
         if self.type == choices.IPAddressTypeChoices.TYPE_SLAAC and self.ip_version != 6:
             raise ValidationError({"type": "Only IPv6 addresses can be assigned SLAAC type"})
 
+        # Validate `parent` can be used as the parent for this ipaddress
+        if self.address and self.parent:
+            closes_parent = self._get_closest_parent()
+            if self.parent != closes_parent:
+                raise ValidationError({"parent": f"{self.parent} cannot be assigned as a parent."})
+            self.parent = closes_parent
```
glennmatthews, line 1105:
```suggestion
            closest_parent = self._get_closest_parent()
            if self.parent != closest_parent:
                raise ValidationError({"parent": f"{self.parent} cannot be assigned as a parent."})
            self.parent = closest_parent
```

nautobot/ipam/models.py
```diff
@@ -1079,30 +1097,26 @@ def clean(self):
         if self.type == choices.IPAddressTypeChoices.TYPE_SLAAC and self.ip_version != 6:
             raise ValidationError({"type": "Only IPv6 addresses can be assigned SLAAC type"})
 
+        # Validate `parent` can be used as the parent for this ipaddress
+        if self.address and self.parent:
```
glennmatthews, line 1101: when would `self.address` evaluate as falsey?

nautobot/ipam/models.py
```diff
@@ -1079,30 +1097,26 @@ def clean(self):
...
-
         # 3.0 TODO: uncomment the below to enforce this constraint
         # if self.parent.type != choices.PrefixTypeChoices.TYPE_NETWORK:
         #     err_msg = f"IP addresses cannot be created in {self.parent.type} prefixes. You must create a network prefix first."
         #     raise ValidationError({"address": err_msg})
 
+        # validated_save is not always called especially in Test environment;
+        # but we still need to set the `self.parent` attr in these cases
+        if self.address and not self.parent:
+            self.parent = self._get_closest_parent()
```
glennmatthews, line 1119: Does this mean that I can now do something like `IPAddress.objects.create(address="10.1.1.1", parent=Prefix.objects.get(network="172.16.0.0/16"))` and as long as I never call clean() it'll be accepted by the DB? We definitely should still prevent that.

nautobot/ipam/models.py
timizuoebideri1: Yes, because if it is popped, we will no longer be able to validate it in the clean method.

nautobot/ipam/models.py
timizuoebideri1: if neither `namespace` nor `parent` is specified when an IPAddress is created e.g `IPAddress.objects.create(address="...", status="...")`

nautobot/ipam/models.py
timizuoebideri1: This would means we would still have to silently discard invalid parent, which is what the issue is reporting against.
Except we force validation save for `IPAddress`

nautobot/ipam/forms.py
```diff
@@ -493,6 +493,8 @@ def clean_namespace(self):
         setattr(self.instance, "_namespace", namespace)
 
     def clean(self):
+        # Pass address to the instance, because this is required to be accessible in the IPAddress.clean method
+        self.instance.address = self.cleaned_data.get("address")
```
glennmatthews, line 497: Is this specifically for "create", for "edit", or both cases that this is needed?

timizuoebideri1: Both cases, cause address might/can be updated

nautobot/ipam/models.py
```diff
@@ -1079,30 +1089,31 @@ def clean(self):
...
-            Prefix.objects.filter(namespace=namespace)
-            # 3.0 TODO: disallow IPAddress from parenting to a TYPE_POOL prefix, instead pick closest TYPE_NETWORK
-            # .exclude(type=choices.PrefixTypeChoices.TYPE_POOL)
-            .get_closest_parent(self.host, include_self=True)
-        )
+        # Validate `parent` can be used as the parent for this ipaddress
+        if self.parent:
+            closest_parent = self._get_closest_parent()
+            if self.parent != closest_parent:
+                raise ValidationError({"parent": f"{self.parent} cannot be assigned as a parent."})
```
glennmatthews, line 1100: Perhaps something more explicit? Like:

```suggestion
                raise ValidationError(
                    {
                        "parent": (
                            f"{self.parent} cannot be assigned as the parent of {self}. "
                            f" In namespace {self._namespace}, the expected parent would be {closest_parent}."
                        )
                    }
                )
```

nautobot/ipam/tests/test_models.py
```diff
@@ -954,6 +957,28 @@ def test_parenting_constraints(self):
...
+        with self.assertRaises(ValidationError) as err:
+            ipaddress = IPAddress(address="192.168.0.1/16", parent=prefixes[0], status=self.status)
+            ipaddress.validated_save()
+        self.assertEqual(f"{prefixes[0]} cannot be assigned as a parent.", err.exception.message_dict["parent"][0])
+
+    def test_creating_an_ipaddress_without_namespace_or_parent(self):
+        with self.assertRaises(ValidationError) as err:
+            ip = IPAddress(address="1976:2023::1/128", status=self.status)
+            ip.validated_save()
+        self.assertEqual(err.exception.message_dict["parent"][0], "Namespace could not be determined.")
```
glennmatthews, line 980: Maybe this error should be changed to something like `"Either a parent or a namespace must be provided"`?

glennmatthews (approved):
A few nits about message wording, otherwise looks solid.

Issue #3583, opened 2023-04-14
Parse Filter Schema Definitions to React Filter Panel Components


bryanculver:
With #3983 @norbert-mieczkowski-codilime added the filtering side panel to the new UI.

With #4162 @timizuoebideri1 extended the Dynamic Filter form to fully emit JSON representation of the fields, lookup types, and field definitions.

In this story should preset in the filtering side panel the choice dropdown of:

- Which field would you like to filter on
- Which lookup expression would you like to choose that's available for the field chosen above
- Present an input box to enter values to filter on

We will create a separate story to map the output of #4162 to each different field type (multi-choice select, dropdown, text input similar to #2381)

---
Carved from #2806, Related (and blocked by): #3582

**Context:**
- [ ] The model field to filter field mapping at `BaseFilterSet.FILTER_DEFAULTS` needs to also include the correct UI widget types in the `extra` for the filter mapping
    - [x] Prototype a JSON representation of the FilterSet to see if we can get away of needing widgets at all for the new UI. JSON Schema would be a good place to start/model from
    - [x] Serve the above representation on the UI API by some convention to the react UI, Ask for a model and get it's filters and their classes and args. We likely will need to advance it to allow for plugin views which define their own filterset (so maybe the view may need to hint that it needs a different filterset than the default one by the model).
    - [ ] Build the filterset => widget logic in the React UI, similar to what has been done here: https://github.com/nautobot/nautobot/blob/0fffa8ee6d4168fdb11b27019e500e0e1ef9fc1c/nautobot/utilities/utils.py#L866, but in React

**This Story:**

Addresses that last point which is taking the schema and rendering the components it in the UI.

bryanculver:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Nautobot-Core-Engineering-64a875e74842500d9e80e23e/issues/nautobot/nautobot/3583?planning-poker) with Zenhub @jathanism

bryanculver:
@norbert-mieczkowski-codilime for the final input field I think it would be valuable to start with
- https://github.com/csandman/chakra-react-select
- https://github.com/JedWatson/react-select

norbert-mieczkowski-codilime:
Yeah, I already implemented `react-select` in Nautobot UI https://github.com/nautobot/nautobot-ui/tree/develop/src/components/ReactSelect. Regarding `chakra-react-select`, the library history shows that it has rather rare maintenance cycles and inactive community, so I decided not to use it.

[...]

PR #4434, opened 2023-09-12
feature: implement object list view filters

norbert-mieczkowski-codilime:
# Closes: #3583

# What's Changed
Implement `ObjecList` view filters in `FiltersPanelContent` component. Additionally, fix some minor existing component lifecycle and style issues for the feature to work correctly.

filters_panel

# TODO
- [x] Use valid `contentType` instead of the one temporarily hardcoded in `nautobot/ui/src/components/FiltersPanel.js:159`
- [ ] For `lookup_field` ("Field" field), use allowed filter fields instead of table headers. See #4462

### DIFF
--- a/nautobot/ui/src/components/FiltersPanel.js
+++ b/nautobot/ui/src/components/FiltersPanel.js
@@ -1,15 +1,28 @@
 import {
     Box,
+    Button,
     calc,
     CloseButton,
+    Divider,
     Flex,
+    FormContainer,
+    FormControl,
+    FormLabel,
     forwardRef,
     getCssVar,
     Heading,
+    Input,
     NtcThumbnailIcon,
+    ReactSelect,
+    Tag,
+    TagCloseButton,
+    TagLabel,
 } from "@nautobot/nautobot-ui";
 import React, {
+    cloneElement,
     createContext,
+    Fragment,
+    isValidElement,
     useCallback,
     useContext,
     useEffect,
@@ -17,6 +30,12 @@ import React, {
     useRef,
     useState,
 } from "react";
+import { useSearchParams } from "react-router-dom";
+
+import { useLazyGetFieldLookupChoicesQuery } from "@utils/api";
+
+export const FILER_RESET_QUERY_PARAMS = ["offset"];
+export const NON_FILTER_QUERY_PARAMS = ["limit", "offset", "sort"];
 
 const FiltersPanel = forwardRef(
     (
@@ -53,6 +72,7 @@ const FiltersPanel = forwardRef(
                     getCssVar("space.md").reference
                 )}
                 marginLeft="md"
+                overflow="auto"
                 position="absolute"
                 right={right}
                 top={0}
@@ -76,13 +96,18 @@ const FiltersPanel = forwardRef(
                     children
                 ) : (
                     <>
-                        <Flex align="center" gap="sm" margin="md">
+                        <Flex align="center" flex="none" gap="sm" margin="md">
                             <NtcThumbnailIcon height="auto" width={24} />
                             <Heading lineHeight="tall">{title}</Heading>
                             <CloseButton marginLeft="auto" onClick={onClose} />
                         </Flex>
-                        <Box paddingBottom="md" paddingX="md">
-                            {content}
+                        <Box flex="1" paddingBottom="md" paddingX="md">
+                            {isValidElement(content)
+                                ? cloneElement(content, {
+                                      onClose,
+                                      ...content?.props,
+                                  })
+                                : content}
                         </Box>
                     </>
                 )}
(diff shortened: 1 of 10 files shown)


bryanculver:
@norbert-mieczkowski-codilime Wherever there is something that needs to be reviewed (your `TODO` comments above) since we may feature branch this

nautobot/ui/src/components/FiltersPanel.js
```diff
@@ -1,22 +1,41 @@
...
     useMemo,
     useRef,
     useState,
 } from "react";
+import { useSearchParams } from "react-router-dom";
+
+import { useLazyGetFieldLookupChoicesQuery } from "@utils/api";
+
+export const FILER_RESET_QUERY_PARAMS = ["offset"];
+export const NON_FILTER_QUERY_PARAMS = ["limit", "offset", "sort"];
```
norbert-mieczkowski-codilime, line 38: Is it a right call to list query params, which are not filters, and assume all others are filters? If so, are these 3 correct?

bryanculver: This _might_ be fine for now, but I think a better way to do this is to emit the FilterSet as a JSON schema along with the model schema, something like

```json
{
    "name":
        "__e": "Exact",
        "__ic": "Contains (Case-Insensitive)",
   #...
}
```

And you can check query params against the dictionary.

OR

In the API response of a filtered query, return a `_meta` property with interpreted filtered params:

```json
{
    "_meta": {
        "filters": {
            "name__ic": "ams"
        }
    },
    "count": 11,
    "next": "https://next.demo.nautobot.com/api/dcim/devices/?limit=5&name__ic=ams&offset=5",
    "previous": null,
    "results": [ #... ]
}
```

bryanculver (commented):
Looks good. Please add a changelog fragment.

While you're at it, can you fix the spacing on the panel: 
image

nautobot/ui/src/utils/api.js
```diff
@@ -104,6 +115,21 @@ export const baseApi = createApi({
             },
             providesTags: ["APIData"],
         }),
+        getFieldLookupChoices: builder.query({
+            query: ({ contentType, lookupField }) => {
+                const queryParams = new URLSearchParams({
+                    content_type: contentType,
+                    field_name: lookupField,
+                    limit: 50,
+                    offset: 0,
```
glennmatthews, line 124: Do we need to do anything to account for filtersets with more than 50 results?

nautobot/ui/src/components/FiltersPanel.js
norbert-mieczkowski-codilime: That makes sense, I'll update the list. Though, since they are already hardcoded in backend, I do not see any reason not to do that in the UI. That is, unless they can be emitted via HTTP endpoint which the UI can be hoooked up to

nautobot/ui/src/utils/api.js
norbert-mieczkowski-codilime: I have no opinion here, I honestly just took the query from the old UI

nautobot/ui/src/components/FiltersPanel.js
```diff
@@ -128,6 +153,384 @@ export const FiltersPanelContainer = forwardRef(
...
+
+        return (
+            <Box height="full" ref={ref}>
+                <FormContainer as="form" onSubmit={onSubmitFilter}>
+                    <Heading as="h3" size="H3">
+                        Add new filter
+                    </Heading>
+
+                    <FormControl>
+                        <FormLabel>Field</FormLabel>
```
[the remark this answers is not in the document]
norbert-mieczkowski-codilime: I took the name from the old UI, but since we're aiming to generally improve things in the new UI, I'm happy to update the labels here as well

nautobot/ui/src/components/FiltersPanel.js
```diff
@@ -128,6 +153,384 @@ export const FiltersPanelContainer = forwardRef(
 
 FiltersPanelContainer.displayName = "FiltersPanelContainer";
 
+export const FiltersPanelContent = forwardRef(
+    ({ contentType, lookupFields, onClose, onSave }, ref) => {
+        // FIXME(norbert-mieczkowski-codilime): this is a temporary hack until passed `contentType` is valid
```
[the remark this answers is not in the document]
norbert-mieczkowski-codilime: Yeah, I'm not sure about it. Hardcoding or mapping is obviously not an option, but if the device type is coming from list elements, when users apply filters which match no objects, the whole filters panel will blow up. I agree that UI has to get the content type somewhere on the meta or schema level

nautobot/ui/src/components/FiltersPanel.js
norbert-mieczkowski-codilime: Anyway, if these names are static, perhaps we do not even have complicate things with additional API endpoints, and instead move them to a shared JSON or YAML file which both API and UI can import.

nautobot/ui/src/components/FiltersPanel.js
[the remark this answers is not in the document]
norbert-mieczkowski-codilime: The `object_type` is now dynamically set from the HTTP API response data

nautobot/ui/src/components/FiltersPanel.js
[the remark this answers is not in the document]
bryanculver: Asking for user feedback

bryanculver:
> Right, the spacing you mentioned is something I'd like to discuss further before changing it here.

Being addressed in #4458

nautobot/ui/src/utils/api.js
bryanculver: @norbert-mieczkowski-codilime if I'm not mistaken this is the lookup for available lookup choices (so `__ie`, `__e`, etc) for the given field. I think 50 is fine for now.

nautobot/ui/src/components/ObjectTable/ObjectListTable.js
```diff
@@ -113,7 +120,13 @@ export default function ObjectListTable({
     );
 
     const filtersPanel = useFiltersPanel({
-        content: <Text>You have successfully opened filters panel.</Text>,
+        content: (
+            <FiltersPanelContent
+                // TODO(norbert-mieczkowski-codilime): use actual lookup fields rather than table headers
```
glennmatthews, line 125: Do you need help from others on implementing the REST API to retrieve the lookup fields?

nautobot/ui/src/utils/api.js
norbert-mieczkowski-codilime: Correct.

nautobot/ui/src/components/ObjectTable/ObjectListTable.js
norbert-mieczkowski-codilime: Likely, I don't feel proficient in the Nautobot backend enough to do it myself

bryanculver: Tracking as a new story

glennmatthews (approved):

bryanculver:
We'll work on https://github.com/nautobot/nautobot/issues/4462 to take this the rest of the way. Merge this into a feature branch to move forward.

Issue #3934, opened 2023-06-16
Add `Namespace` Column to Related Model Tables, Such as Device "Assigned VRFs"


glennmatthews:
### As ...

Nelly - Network Engineer

### I want ...

To easily be able to distinguish between similarly-named VRFs in different namespaces whenever they appear in the UI.

One example where this is not apparent is in the Device detail view "Assigned VRFs" table, where only the VRF `name`, `rd`, and `tenant` appear, which may be ambiguous without also adding a `namespace` column. There are probably any number of other similar cases.

### So that ...

I can effectively make use of namespaced VRFs.

### I know this is done when...

All relevant UI tables include VRF namespace information.

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [X] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

none

### External Dependencies

none

PR #4435, opened 2023-09-12
Add `Namespace` attribute to relevant tables and views.

HanlinMiao:
# Closes: #3934 
# What's Changed

Screen Shot 2023-09-12 at 11 11 38 AM

Screen Shot 2023-09-12 at 11 11 52 AM

Screen Shot 2023-09-13 at 1 02 01 PM

Screen Shot 2023-09-13 at 1 02 33 PM

Screen Shot 2023-09-13 at 1 02 12 PM

Fixed validation logic for primary_ips in `VirtualMachine.clean()` method

# TODO

- [ ] Unit, Integration Tests


bryanculver:
Since we scoped this as a 2, should we consider adding the Namespace to things like Primary IP fields and the IP Address column on Device Interface views?

image

image

bryanculver (commented):
Suggest reviewing some other tables where IPAddress/Prefix is displayed without Namespace in context.

changes/3934.added
```diff
@@ -0,0 +1,4 @@
+Add `Namespace` column to VRFDeviceAssignmentTable and VRFPrefixAssignmentTable to display assigned vrfs' namespace property.
+Append namespace attribute to IPAddress columns of relevant Interface and InterfaceRedundancyGroup tables.
+Append namespace attribute to primary_ip fields in DeviceDetailView and VirtualMachineDetailView.
+Add hover copy button and functionality to primary_ip fields in VirtualMachineDetailView.
```
glennmatthews, line 4:
```suggestion
Added `Namespace` column to `VRFDeviceAssignmentTable` and `VRFPrefixAssignmentTable` to display assigned VRFs' `namespace` property.
Added `namespace` attribute to rendering of "IP Addresses" columns of relevant `Interface` and `InterfaceRedundancyGroup` tables.
Added `namespace` attribute to rendering of `primary_ip` fields in `DeviceDetailView` and `VirtualMachineDetailView`.
Added hover copy button and functionality to `primary_ip` field in `VirtualMachineDetailView`.
```

nautobot/virtualization/models.py
```diff
@@ -273,10 +273,10 @@ def clean(self):
             ip = getattr(self, field)
             if ip is not None:
                 if field == "primary_ip4":
-                    if ip.family != 4:
+                    if ip.ip_version != 4:
                         raise ValidationError({f"{field}": f"{ip} is not an IPv4 address."})
                 else:
-                    if ip.family != 6:
+                    if ip.ip_version != 6:
                         raise ValidationError({f"{field}": f"{ip} is not an IPv6 address."})
```
glennmatthews, line 280: Oof, how'd we miss this? Can you add a test case exercising this code?

nautobot/virtualization/models.py
```diff
@@ -273,10 +273,10 @@ def clean(self):
             ip = getattr(self, field)
             if ip is not None:
                 if field == "primary_ip4":
-                    if ip.family != 4:
+                    if ip.ip_version != 4:
```
glennmatthews, line 276: Not blocking for this issue, but I noticed via grep that the IPAM querysets still have an `ip_family` method, wonder if we should be renaming that to `version` or `ip_version`? Or maybe actually removing it altogether as unnecessary now that `version` is a concrete field?

nautobot/virtualization/templates/virtualization/virtualmachine.html
```diff
@@ -64,8 +64,9 @@
                 <tr>
                     <td>Primary IPv4</td>
                     <td>
+                        <span class="hover_copy">
                         {% if object.primary_ip4 %}
```
glennmatthews, line 68: Perhaps?

```suggestion
                        {% if object.primary_ip4 %}
                        <span class="hover_copy">
```

nautobot/virtualization/templates/virtualization/virtualmachine.html
```diff
@@ -65,17 +65,22 @@
...
-                            {% elif object.primary_ip4.nat_outside_list.exists %}
-                                <br /><span>NAT: </span>
-                                <ul>
-                                    {% for nat in object.primary_ip4.nat_outside_list.all %}
-                                        <li>{{ nat|hyperlinked_object }}</li>
-                                    {% endfor %}
-                                </ul>
-                            {% endif %}
+                            <span class="hover_copy">
+                                {{ object.primary_ip4|hyperlinked_object|add_html_id:"ipv4" }} ({{ object.primary_ip4.parent.namespace|hyperlinked_object }})
```
glennmatthews, line 69: Not at all blocking, but I wonder if it would be a useful nicety to add and use something like the following on `IPAddress` model:

```python
@property
def namespace(self):
    return self.parent.namespace
```

nautobot/ipam/tables.py
```diff
@@ -277,20 +282,25 @@ class VRFDeviceAssignmentTable(BaseTable):
...
 class VRFPrefixAssignmentTable(BaseTable):
     """Table for displaying VRF Prefix Assignments."""
 
     vrf = tables.Column(verbose_name="VRF", linkify=lambda record: record.vrf.get_absolute_url(), accessor="vrf.name")
+    namespace = tables.Column(
+        verbose_name="Namespace",
+        linkify=lambda record: record.vrf.namespace.get_absolute_url(),
+        accessor="vrf.namespace.name",
+    )
     prefix = tables.Column(linkify=True)
```
glennmatthews, line 297: Actually I'm wondering if this one (used only on the VRF detail view, which I don't see a screenshot of in the PR) is needed after all. Don't we enforce that Prefix/VRF assignments must always belong to the same common namespace? So won't the namespace be the same for every row in this table?

glennmatthews (approved):

nautobot/ipam/tables.py
HanlinMiao: Yeetus Deletus

nautobot/virtualization/templates/virtualization/virtualmachine.html
HanlinMiao: Trying to add that @property header makes the API unittests fail for some reason. https://github.com/nautobot/nautobot/actions/runs/6190492584/job/16806921893?pr=4435

bryanculver: Probably because it isn’t settable. Not blocking. Let’s open an issue to add improvements to accessing namespace from IPAddress

HanlinMiao: https://github.com/nautobot/nautobot/issues/4452

Issue #4346, opened 2023-08-29
The Natural Slug Property


itdependsnetworks:
## TODO

_From @lampwins comment below:_

- The natural key field(s) are still the input to the composite key function
- Change the output of the composite key function to the django slugify function, or some derivative of it (look at our `slugify_dashes_to_underscores` function)
- CSV Export will move to PK for the time being
    - We will create a story to address export side of #4274
- Provide a clear acknowledgement that the resulting composite key strings will not have a guarantee of uniqueness.
- The composite key strings will no longer be useable for direct lookup via any mechanism
- Access this property via `.natural_slug` and should be accessible via the REST and GraphQL APIs
- Append the first 4 characters of the PK to the end of the generated slug (if `314b709c-fa1e-46af-8c49-dfedd1f4081c` append `_314b`) to aid in minimizing collisions

---
### As ...

P.D. - Plugin Developer

### I want ...

The ability to have a standard way to that "slugify" is provided. The request is to have a property that is called slug, based on name, and ~uniqueness is enforced~ based on this as well. 

### So that ...

* My migration is easier
* There is only one official way to "slugify" apps

### I know this is done when...

There is a slug property on any model that has the "name" attribute on it.

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [X] Circuits
- [X] DCIM
- [X] IPAM
- [X] Misc (including Data Sources)
- [ ] Organization
- [X] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

Yes, add properties

itdependsnetworks:
Just to provide some color and concrete use cases, all of which could be handled in other manners, but trying to provide a reasonable user experience:

* I use the slug for my file and folder structure, which has issues with special characters
* I use my slug in my json/yaml/csv or other output formats that have issues with special characters
* I use my slug in my url/html or other output tools that have issues with special characters
* I prefer my shortened name slugified vs natural key since it is user intuitive, friendly and generally (but not always) is unique and handles 80%+ of my use cases. Meaning I do not have to think about what my device role, device type, or manufacturer strategy is going be, as I will rarely if ever have a name space issue here, but will have special characters.

lampwins:
After a few conversations, we are going to change some of the behavior of "composite keys" to to meet these usability concerns.

[...]

Issue #4368, opened 2023-09-01
Allow non-english characters in Composite Key


apast0r:
### Environment
* Nautobot version (Docker tag too if applicable):  v2.0.0rc2
* Python version: python 3.11
* Database platform, version: 8.0.33
* Middleware(s):

### Steps to Reproduce
1. Create any location with non English character. Example: "España".
2. Create a rack and set his location to "España".
3. Create a device and assign to the rack.
4. Go to Advanced Tab of the device to see the Composite Key.

### Expected Behavior
The Composite Key sould be: "device;rack;España"

### Observed Behavior
The Composite Key is: "device;%00;Espa%C3%B1a"

This also happen in the Export generator.

bryanculver:
@apast0r We are considering making two changes related to this:

- #4346 (note it would remove composite key and not be usable for lookup)
- #4367 (inverse of #3794)

The second story should not perform any URL-safe encoding of strings so should support non-English characters, but since we will be eliminating the Composite Key itself, I wanted to ensure your needs would be addressed.

jathanism:
@apast0r Hello! In the working I'm doing in #4439, it will be a bit of a compromise, where instead of stripping the characters, they have accents removed. So your example would result in `España` becoming `espana` instead of `Espa%C3%B1a`.

bryanculver:
@apast0r Please take a look at RC3 which now implements `natural_slug` and better handles non-English characters. Exporting the non-URL encoded payload by breaking out the natural keys as discrete columns will be in the final 2.0 release.

apast0r:
hi @bryanculver, I will update our test environment with the last 2.0 RC3.

PR #4439, opened 2023-09-12
Implemented `.natural_slug` property on all models that inherit from `BaseModel`.

jathanism:
# Closes: #4346 
# What's Changed

- Implemented a new utility called `construct_natural_slug()` which is kind of like `construct_composite_key()` but not reversible, is lossy, and is prioritizing human-readability over uniqueness.
- Added a `.natural_slug` property to all models that inherit from `BaseModel` which generates the value from the natural key for the object.
- Effectively `slugify_dots_to_dashes()` is used on each member item in the list of natural key values for an instance, and then joined together with underscores.

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/core/api/serializers.py
+++ b/nautobot/core/api/serializers.py
@@ -24,9 +24,9 @@ from nautobot.core.api.utils import (
     dict_to_filter_params,
     nested_serializer_factory,
 )
-from nautobot.core.models.constants import COMPOSITE_KEY_SEPARATOR
+from nautobot.core.models.constants import NATURAL_SLUG_SEPARATOR
 from nautobot.core.models.managers import TagsManager
-from nautobot.core.models.utils import construct_composite_key
+from nautobot.core.models.utils import construct_natural_slug
 from nautobot.core.utils.lookup import get_route_for_model
 from nautobot.core.utils.requests import normalize_querydict
 from nautobot.extras.api.relationships import RelationshipsDataField
@@ -121,7 +121,7 @@ class BaseModelSerializer(OptInFieldsMixin, serializers.HyperlinkedModelSerializ
 
     display = serializers.SerializerMethodField(read_only=True, help_text="Human friendly display value")
     object_type = ObjectTypeField()
-    composite_key = serializers.SerializerMethodField()
+    natural_slug = serializers.SerializerMethodField()
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -144,12 +144,12 @@ class BaseModelSerializer(OptInFieldsMixin, serializers.HyperlinkedModelSerializ
     @extend_schema_field(
         {
             "type": "string",
-            "example": COMPOSITE_KEY_SEPARATOR.join(["attribute1", "attribute2"]),
+            "example": NATURAL_SLUG_SEPARATOR.join(["attribute1", "attribute2"]),
         }
     )
-    def get_composite_key(self, instance):
+    def get_natural_slug(self, instance):
         try:
-            return getattr(instance, "composite_key", construct_composite_key(instance.natural_key()))
+            return getattr(instance, "natural_slug", construct_natural_slug(instance.natural_key(), pk=instance.pk))
         except (AttributeError, NotImplementedError):
             return "unknown"
 
--- a/nautobot/core/api/serializers.py
+++ b/nautobot/core/api/serializers.py
@@ -24,9 +24,9 @@ from nautobot.core.api.utils import (
     dict_to_filter_params,
     nested_serializer_factory,
 )
-from nautobot.core.models.constants import NATURAL_SLUG_SEPARATOR
+from nautobot.core.models import constants
 from nautobot.core.models.managers import TagsManager
-from nautobot.core.models.utils import construct_natural_slug
+from nautobot.core.models.utils import construct_composite_key, construct_natural_slug
 from nautobot.core.utils.lookup import get_route_for_model
 from nautobot.core.utils.requests import normalize_querydict
 from nautobot.extras.api.relationships import RelationshipsDataField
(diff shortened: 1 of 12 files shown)


nautobot/core/tests/test_models.py
```diff
@@ -59,6 +59,13 @@ def test_composite_key(self):
         dt = DeviceType.objects.first()
         self.assertEqual(dt.composite_key, construct_composite_key(dt.natural_key()))
 
+    def test_natural_slug(self):
+        """Test the natural_slug default implementation with some represenatitive models."""
```
glennmatthews, line 63:
```suggestion
        """Test the natural_slug default implementation with some representative models."""
```

glennmatthews:
I raised this on #4346, but is it in scope for this PR to **remove** all `composite_key` support as well? That seemed to be implied but was not explicitly stated as such so far as I can see.

glennmatthews:
Given the emphasis on human-readability, should we consider using something like https://github.com/un33k/python-slugify instead of django-slugify to handle non-ASCII characters more gracefully than just discarding them entirely?

nautobot/core/models/utils.py
```diff
@@ -203,3 +204,22 @@ def deconstruct_composite_key(composite_key):
...
+    - Spaces or repeated dashes are converted to single dashes.
+    - Remove characters that are not alphanumerics, underscores, or hyphens.
+    - Converted to lowercase.
+    - Strips leading/trailing whitespace, dashes, and underscores.
+    - Each natural key value in the list is separated by underscores.
+
+    This value is not reversible, lossy, and is not guaranteed to be unique.
+    """
+    values = [str(value) if value is not None else "\0" for value in values]
+    return "_".join(slugify_dots_to_dashes(value) for value in values)
```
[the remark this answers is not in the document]
jathanism: Yes. To the latter question it was just copy pasta from `construct_composite_key` to have a unicode space as a filler. It gets normalized out anyhow.

jathanism:
> I raised this on #4346, but is it in scope for this PR to **remove** all `composite_key` support as well? That seemed to be implied but was not explicitly stated as such so far as I can see.

I'm not sure. I still feel like given the work already done the composite key still has tangible value. It's probably another delightful bike shedding conversation about why, but I feel like all preceding justification for it is still valid.

jathanism:
> Given the emphasis on human-readability, should we consider using something like https://github.com/un33k/python-slugify instead of django-slugify to handle non-ASCII characters more gracefully than just discarding them entirely?

Thanks for this. I will check it out.

nautobot/core/models/utils.py
```diff
@@ -203,3 +205,24 @@ def deconstruct_composite_key(composite_key):
...
+
+def construct_natural_slug(values):
+    """
+    Convert the given list of natural key values to a single human-readable string.
+
+    A third-party lossy `slugify()` function is used to convert each natural key value to a
+    slug, and then they are joined with an underscore.
+
+    - Spaces or repeated dashes are converted to single dashes.
+    - Remove characters that are not alphanumerics, underscores, or hyphens.
```
glennmatthews, line 218: This is more nuanced now that we're using `python-slugify` right? Non-ASCII characters are converted to closest-match alphanumeric characters now?

jathanism: Yes. 

```python
In [2]: w = "Über Ålpha"

In [3]: slugify(w)
Out[3]: 'uber-alpha'
```

lampwins: Will also point out https://github.com/nautobot/nautobot/issues/4368. Sounds like this will reasonably solve that ask as it relates to natural slug, at least for now.

jathanism: > Will also point out #4368. Sounds like this will reasonably solve that ask as it relates to natural slug, at least for now.

Yes. The revised slugify pattern is aware of accents/ligatures, so the example in #4368 would be:

```python
'device-rack-espana'
```

glennmatthews (commented):
Code looks solid. Let's get some documentation written and then :shipit:.

nautobot/core/api/serializers.py
```diff
@@ -122,6 +122,7 @@ class BaseModelSerializer(OptInFieldsMixin, serializers.HyperlinkedModelSerializ
     display = serializers.SerializerMethodField(read_only=True, help_text="Human friendly display value")
     object_type = ObjectTypeField()
     composite_key = serializers.SerializerMethodField()
+    natural_slug = serializers.SerializerMethodField()
```
glennmatthews, line 125: Can/should we have a test in the generic API test case that checks to make sure `natural_slug` is present and correctly populated on all serializers? Maybe just an extra check in `APIViewTestCases.GetObjectViewTestCase.test_get_object`?

nautobot/core/models/__init__.py
```diff
@@ -148,6 +149,16 @@ def composite_key(self) -> str:
         """
         return construct_composite_key(self.natural_key())
 
+    @property
+    def natural_slug(self) -> str:
+        """
+        Automatic "slug" string derived from this model's natural key. This differs from composite
+        key in that it must human-readable and therefore lossy. This value is not guaranteed to be
```
glennmatthews, line 156:
```suggestion
        key in that it must be human-readable and comply with a very limited character set, and is therefore lossy.
        This value is not guaranteed to be
```

nautobot/core/models/utils.py
```diff
@@ -200,6 +203,44 @@ def deconstruct_composite_key(composite_key):
...
+
+    - Spaces or repeated dashes are converted to single dashes.
+    - Accents and ligatures from Unicode characters are reduced to ASCII.
+    - Remove remaining characters that are not alphanumerics, underscores, or hyphens.
+    - Converted to lowercase.
+    - Strips leading/trailing whitespace, dashes, and underscores.
+    - Each natural key value in the list is separated by underscores.
+    - Emojis will be converted to their registered name.
+
+    This value is not reversible, lossy, and is not guaranteed to be unique.
```
glennmatthews, line 229:
```suggestion
    This value is not reversible, is lossy, and is not guaranteed to be unique.
```

bryanculver (approved):

PR #4440, opened 2023-09-13
Fixed an issues with ModelForm's that used   on  and  displayed a JSO…

itdependsnetworks:
…N blob

# Closes: DNE
# What's Changed

Fixed an issues with ModelForm's that used `__all__`  on `Meta.fields` and `custom_fields` displayed a JSON blob

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/forms/mixins.py
+++ b/nautobot/extras/forms/mixins.py
@@ -90,6 +90,12 @@ class CustomFieldModelFormMixin(forms.ModelForm):
         super().__init__(*args, **kwargs)
 
         self._append_customfield_fields()
+        try:
+            if self.Meta.fields == "__all__":
+                # Above we accounted for all cf_* data, so can safely remove _custom_field_data.
+                self.fields.pop("_custom_field_data", None)
+        except AttributeError:
+            pass
 
     def _append_customfield_fields(self):
         """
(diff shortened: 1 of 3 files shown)


nautobot/extras/forms/mixins.py
```diff
@@ -107,6 +107,13 @@ def _append_customfield_fields(self):
             # Annotate the field in the list of CustomField form fields
             self.custom_fields.append(field_name)
 
+        try:
+            if self.Meta.fields == "__all__":
```
glennmatthews, line 111: Not sure this check is necessary, TBH - I see you're testing this below, but is there ever actually a valid use case for trying to present `_custom_field_data` as a form field?

itdependsnetworks: I cannot think of a valid use case, and I approached it that way at first, but I figured if there was one I couldn't think of, then not considering this extra check would not follow the principal of least astonishment.

nautobot/extras/tests/test_forms.py
```diff
@@ -1195,3 +1195,27 @@ def test_without_dynamic_groups(self):
...
+
+        class TestForm(CustomFieldModelFormMixin):
+            class Meta:
+                model = dcim_models.InterfaceRedundancyGroup
+                fields = "__all__"
+
+        custom_field_form = TestForm()
+        self.assertNotIn("_custom_field_data", custom_field_form.fields)
+
+    def test_custom_field_data_removed_in_kept(self):
```
glennmatthews, line 1212: Perhaps?

```suggestion
    def test_custom_field_data_kept_if_explicit(self):
```

nautobot/extras/forms/mixins.py
glennmatthews: Fair point. That said it's a bit surprising itself that this removal happens in `_append_customfield_fields()` since nothing about that function name suggests that it's going to remove anything - maybe this logic should be located directly in `__init__` above instead?

glennmatthews (approved):

Issue #4033, opened 2023-06-29
Filter params not being sent to ObjectBulkUpdateViewMixin and ObjectBulkDestroyViewMixin


joewesch:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.20
* Python version: 3.8.16
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Filter a model's list view that uses the NautobotUIViewSet (like Circuits) that returns a list of items less than the total number of items but still more than the page number (example: https://demo.nautobot.com/circuits/circuits/?&tenant=nautobot-airports) - this causes the "select all" dialog popup
2. Click the select all checkbox at the top of the list
3. Select the checkbox that says "Select all 160 circuits matching query"
4. Click either Edit All or Delete All

### Expected Behavior
I expect the number of items that are edited or deleted to be the original filtered queryset (i.e. 160 circuits)

### Observed Behavior
Instead it shows:
- When Editing: "Editing **184 Circuits**"
- When Deleting: "Warning: The following operation will delete **184 circuits**. Please carefully review the circuits to be deleted and confirm below."

### Analysis
I believe this is due to the `request.POST` data [getting sent to the filterset](https://github.com/nautobot/nautobot/blob/develop/nautobot/core/views/mixins.py#L926), but since the filters are query params they are typically in the `request.GET` data.

lampwins:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Nautobot-Core-Product-64a8783b7a3f4b1225973a0c/issues/nautobot/nautobot/4033?planning-poker) with Zenhub @bryanculver @glennmatthews @gsnider2195 @HanlinMiao @jathanism @timizuoebideri1

PR #4443, opened 2023-09-13
Fix bulk edit and bulk delete filtering

snaselj:
# Closes: #4033, #4478

## What's Changed

Fixed filtering objects in bulk edit and bulk delete.

## To Do

- [x] Add bulk edit all with filter test.
    - [x] Fix skipped models.
- [x] Add bulk delete all with filter test.
    - [x] Fix skipped models.


snaselj:
> Changes look good. Anything we can/should add to the generic view test cases that would have caught this bug when it was introduced?

Thanks, I can add tests covering this if desirable.

nautobot/core/testing/views.py
```diff
@@ -1054,6 +1054,52 @@ def test_bulk_edit_form_contains_all_pks(self):
             for pk in pk_list:
                 self.assertIn(f'<input type="hidden" name="pk" value="{pk}"', response_body)
 
+        @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
+        def test_bulk_edit_form_contains_all_filtered(self):
```
snaselj, line 1058: Added a test for bulk edit with the filter, some models fails, so skipping them so far. WDYT about that @glennmatthews pls?

nautobot/core/testing/views.py
```diff
@@ -1054,6 +1054,52 @@ def test_bulk_edit_form_contains_all_pks(self):
...
+            # We are testing the intermediary step of bulk_edit with pagination applied and additional filter.
+            # i.e. "_all" passed in the form and filter using query params.
+            # Prequisity is not to have `bulk_edit_data` values in objects created by `setUpTestData`.
+            pk_list = self._get_queryset().values_list("pk", flat=True)
+            self.assertTrue(len(pk_list) > 1)
+            obj_perm = users_models.ObjectPermission(name="Test permission", actions=["change"])
+            obj_perm.save()
+            obj_perm.users.add(self.user)
+            obj_perm.object_types.add(ContentType.objects.get_for_model(self.model))
```
glennmatthews, line 1067: Should be able to replace these lines with something like `self.add_permissions(f"{self.model._meta.app_label}.change_{self.model._meta.model_name}")`

nautobot/core/testing/views.py
```diff
@@ -1054,6 +1054,52 @@ def test_bulk_edit_form_contains_all_pks(self):
...
+            # Then we bulk update objects with filter applied.
+            #     We only pass in another pk to test the functionality of "_all" and filter,
+            #     which should grab instance updated in the previous step.
+            update_data = {
+                "pk": pk_list[1:2],
+                "_all": "on",
+                **post_data,
+            }
+            query_string = urlencode(post_data, doseq=True)
+            response = self.client.post(f"{self._get_url('bulk_edit')}?{query_string}", update_data)
```
glennmatthews, line 1094: Need to either:

1. restrict the constructed `query_string` to only include valid filter parameters
2. `@override_settings(STRICT_FILTERING=False)`

Probably the latter is simpler.

snaselj: Not necessary after switching to pk filtering.

nautobot/core/testing/views.py
snaselj: Thanks applied.

snaselj:
Ready for review now.

jathanism:
> Ready for review now.

Looking good overall but I would like to see the tests passing so I can commit to final review!

jathanism (approved):
:shipit:

Issue #3372, opened 2023-02-27
ObjectPermission Data Migration Strategy


HanlinMiao:
### Proposed Changes

For any models that are being removed or collapsed into models (`Site` and `Region` -> `Location`, `Aggregate` -> `Prefix`),
we need to keep some sort of records to trace back the old model instances that are removed along with the old model classes specifically for `ObjectPermission` constraints constructed by the UUIDs of old model instances:

```json
{
    "id": "d266c50f-fdcc-4bb1-b768-2492034e1b86",
    "id__in": ["d266c50f-fdcc-4bb1-b768-2492034e1b86", "5318f8c3-c0dc-4534-9b92-458a378a0fdc"],
}
```

### Justification
Since we decided to let admin users themselves resolve and change the constraints of ObjectPermission instances to use the new model filters **after the old model classes are removed**, we need to assist the user in any way possible in referencing the old model instances.

lampwins:
As discussed, I think we should migrate the UUIDs to the new objects, i.e. a record keeps its same UUID through the migration. This would help not only in this case, but also external/3rd party references to records via the UUID.

Additionally, I would like to see us produce log files are part of these migrations to assist administrators if they need to trace an object.

bryanculver:
> As discussed, I think we should migrate the UUIDs to the new objects, i.e. a record keeps its same UUID through the migration. This would help not only in this case, but also external/3rd party references to records via the UUID.
> 
> Additionally, I would like to see us produce log files are part of these migrations to assist administrators if they need to trace an object.

This is being done for Location, but isn't for Role, Aggregate, and potentially others. @lampwins would it be preferred to go back and do that for all models which are being collapsed into others?

gsnider2195:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Nautobot-Core-Engineering-64a875e74842500d9e80e23e/issues/nautobot/nautobot/3372?planning-poker) with Zenhub @gsnider2195 @HanlinMiao @jathanism @timizuoebideri1

PR #4445, opened 2023-09-13
Added ObjectPermission constraints check to pre_migrate management command

gsnider2195:
# Closes: #3372
# What's Changed

This issue still needs docs added to the develop branch. (#4450) This is just the management command helper for 1.x

image

# TODO

- [ ] Unit, Integration Tests


nautobot/core/management/commands/pre_migrate.py
```diff
@@ -75,6 +86,115 @@ def check_configcontext_uniqueness():
...
+                continue
+
+            for lookup in qs._query.where.children:
+                related_model = lookup.lhs.target.related_model or lookup.lhs.target.model
+
+                if related_model in deleted_models:
+                    cls = f"{related_model.__module__}.{related_model.__name__}"
+                    warnings.append(
+                        f"ObjectPermission '{perm}' (id: {perm.id}) has a constraint that references "
+                        f"a model ({cls}) that will be deleted by the Nautobot 2.0 migration."
```
glennmatthews, line 156: perhaps `deleted or replaced` or something like that? E.g. a constraint on `device_role__name` might just need to become a constraint on `role__name` in 2.0?

Actually, will this catch `device_role__name` or only just `device_role`?

nautobot/core/management/commands/pre_migrate.py
```diff
@@ -75,6 +86,115 @@ def check_configcontext_uniqueness():
         )
 
 
+def check_permissions_constraints(command):
+    """
+    Check for permission constraints that are referencing an object field
+    that will be migrated to a different model/field/value in 2.0 migrations.
+
+    Objects that will be deleted in 2.0 are:
```
glennmatthews, line 94: s/deleted/replaced/?

nautobot/core/management/commands/pre_migrate.py
```diff
@@ -75,6 +86,115 @@ def check_configcontext_uniqueness():
...
+    - dcim.DeviceRole
+    - dcim.RackRole
+    - ipam.Aggregate
+    - ipam.Role
```
glennmatthews, line 98: What about Site and Region?

nautobot/core/management/commands/pre_migrate.py
```diff
@@ -75,6 +86,115 @@ def check_configcontext_uniqueness():
...
+    ]
+
+    pk_change_models = [
+        IPAddress,
+        Prefix,
+        ObjectPermission,
+        TaggedItem,
+    ]
+
+    field_change_models = {
```
glennmatthews, line 126: Should this or something similar cover all of the field changes documented in 2.0 in `nautobot/docs/user-guide/administration/upgrading/from-v1/tables/v2-database-renamed-fields.yaml` and/or `v2-database-behavior-changes.yaml` and/or `v2-database-removed-fields.yaml`?

E.g. `Vlan.group` --> `VLAN.vlan_group`, `ConfigContext.regions` -> `ConfigContext.locations`, etc etc.

nautobot/core/management/commands/pre_migrate.py
gsnider2195: This only works on the last model referenced in the qs so `device_role__name` works but it wouldn't catch `device_role__devices__name` because the related model in that case is `Device`. I'm not sure if we can get that granular here without blowing up the scope.

nautobot/core/management/commands/pre_migrate.py
gsnider2195: I thought we covered those in the migrations but if not I'll add them here

nautobot/core/management/commands/pre_migrate.py
gsnider2195: That's what I was hoping to accomplish. Do the best we can here but also document that it's not 100%. The field renames and region/site shouldn't be too problematic because new random UUIDs aren't being assigned but for Aggregate, Prefix and IPAddress it's going to be important to document the old primary keys.

smk4664: This caught me, I read through the migration guide and missed the group -> vlan_group or the temrination_a|z -> circuit_termination_a|z. Also an issue with `DeviceType.instances` -> `DeviceType.devices`.

nautobot/core/management/commands/pre_migrate.py
```diff
@@ -75,6 +118,259 @@ def check_configcontext_uniqueness():
...
+        DeviceRole,
+        RackRole,
+        Region,
+        Role,
+        Site,
+    ]
+
+    pk_change_models = [
+        IPAddress,
+        Prefix,
```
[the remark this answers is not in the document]
gsnider2195: The migrations aren't changing the PKs or doing any deduplication but I still think it's a good idea to bring any constraints using those models' PKs to the users' attention because their prefixes and ip addresses are likely going to need to be deduplicated manually after the migration and the instance that they choose to keep may not match the instance that they originally pointed the constraint to. It's probably highly unlikely that a constraint is pointing to one instance of a duplicate IP/Prefix so maybe it's ok to leave it out. I don't think it hurts to leave it in either though?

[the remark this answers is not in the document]
glennmatthews: Makes sense, thanks for the context. I worry that the output of this command will already be potentially overwhelming for users with any realistically complex set of defined permissions, but it's probably better to err on the side of a few false-positive warnings than potentially miss important changes.

glennmatthews (approved):

bryanculver (approved):

Issue #4367, opened 2023-08-31
Natural key multi column csv export


lampwins:
As Ozzy the Operator, I want to be able to export Nautobot data that contains relationships to other data in a format that is clear and easy for me to understand.

I will know this is complete when CSV export supports multi column natural keys by allowing the various fields in a natural key to be separate fields in the export.
- Fields should be named with double underscore deleters, I.e when exporting devices, the tenant field would be `tenant__name`, so they are round-trippable.
- Device and location should honor the simplified natural key settings, e.g. `DEVICE_NAME_NATURAL_KEY`

This is the inverse of #3794

lampwins:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Nautobot-Core-Product-64a8783b7a3f4b1225973a0c/issues/nautobot/nautobot/4367?planning-poker) with Zenhub @bryanculver @glennmatthews @gsnider2195 @HanlinMiao @jathanism @timizuoebideri1

HanlinMiao:
Should this be blocked by https://github.com/nautobot/nautobot/issues/4346?

bryanculver:
> Should this be blocked by #4346?

If this gets done first it negates the need to change the export of #4346 from `composite_key` to `pk`. More likely this can just be a proper communication depending on the order they are picked up.

bryanculver:
Due to M2M->CSV->M2M ugliness we are going to document not supporting M2Ms (except for tags) in CSV round trip and exclude testing for them for 2.0.0.

Future follow on may find a way to re-introduce them, or provide multi-part forms to handle this (see how we've documented addressing CSV import of IPAddress-to-Interface assignments via #4103)

PR #4448, opened 2023-09-14
Add natural_key_field_lookups to to csv export

timizuoebideri1:
# Closes: #4367
# What's Changed

- Modify the `BaseModelSerializer`:
    - in the `__init__` method,  when handling CSV export requests, we efficiently retrieve the values of `field.model. natural_key_field_lookups` for each of the related fields within the Serializer Model using `Model.objects.filter(...).values(<all_related_fields_models_natural_key_field_lookups>)`. 
    - Override the `to_representation` method to replace related field key-value pairs with their corresponding `natural_key_field_lookups` key-value pairs

# TODO

- [ ] Unit, Integration Tests
- [ ] Unittest
- [ ] Handling ManyToMany Field


nautobot/core/api/serializers.py
```diff
@@ -122,13 +123,50 @@ class BaseModelSerializer(OptInFieldsMixin, serializers.HyperlinkedModelSerializ
...
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         # If it is not a Nested Serializer, we should set the depth argument to whatever is in the request's context
         if not self.is_nested:
             self.Meta.depth = self.context.get("depth", 0)
 
+        # Check if the request is related to CSV export;
+        if self._is_csv_request() and self.instance:
+            related_fields_natural_key_field_lookups = self._get_related_fields_natural_key_field_lookups()
+            if isinstance(self.instance, QuerySet):
```
glennmatthews, line 138: When is this the case?

nautobot/core/api/serializers.py
```diff
@@ -122,13 +123,50 @@ class BaseModelSerializer(OptInFieldsMixin, serializers.HyperlinkedModelSerializ
...
+                queryset = self.instance
+            else:
+                # We would only need to run one additional query, making this a more efficient method of
+                # obtaining all the natural key values for this instance;
+                queryset = self.instance.__class__.objects.filter(pk=self.instance.pk)
```
glennmatthews, line 143: Perhaps?

```suggestion
                queryset = self.Meta.model.objects.filter(pk=self.instance.pk)
```

nautobot/core/api/serializers.py
```diff
@@ -122,13 +123,50 @@ class BaseModelSerializer(OptInFieldsMixin, serializers.HyperlinkedModelSerializ
...
+        retrieves the natural_key_field_lookups for each related model, and prepends the field name
+        to create a list of field lookups.
+        """
+        model = self.Meta.model
+        field_lookups = []
+        fields = [field for field in model._meta.get_fields() if field.is_relation]
+        self.related_model_fields_names = [field.name for field in fields]
+        for field in fields:
+            # Get each related field model's natural_key_fields and prepend field name
+            with contextlib.suppress(Exception):
```
glennmatthews, line 159: What exception class(es) are you expecting to encounter here? It worries me to see such a broad ignoring of exceptions.

nautobot/core/api/serializers.py
```diff
@@ -233,6 +271,39 @@ def build_field(self, field_name, info, model_class, nested_depth):
...
+        altered_data = {}
+
+        # if `self.natural_keys_values` is not empty means that this is a csv request then add natural_kay key and values
+        if self.natural_keys_values:
+            natural_key_field_instance = [item for item in self.natural_keys_values if item["pk"] == instance.pk][0]
+            for key, value in data.items():
+                if natural_key_field_lookups_for_field := self._get_natural_key_field_lookups_for_field(
+                    key, natural_key_field_instance
+                ):
+                    altered_data |= natural_key_field_lookups_for_field
```
glennmatthews, line 300: Per the Python docs, the `dict |= other_dict` syntax was only added in Python 3.9, so we can't use it yet.

nautobot/core/api/serializers.py
```diff
@@ -122,13 +123,50 @@ class BaseModelSerializer(OptInFieldsMixin, serializers.HyperlinkedModelSerializ
...
+        model = self.Meta.model
+        field_lookups = []
+        fields = [field for field in model._meta.get_fields() if field.is_relation]
+        self.related_model_fields_names = [field.name for field in fields]
+        for field in fields:
+            # Get each related field model's natural_key_fields and prepend field name
+            with contextlib.suppress(Exception):
+                field_lookups.extend(
+                    f"{field.name}__{lookup}" for lookup in field.related_model.natural_key_field_lookups
+                )
```
glennmatthews, line 162: Perhaps related to the above comment, do we need to have special-case handling for relations to non-Nautobot models (`Group`, `ContentType`) that don't have a `natural_key_field_lookups` attribute?

nautobot/core/api/serializers.py
timizuoebideri1: When `many=True` i.e `DeviceSerializer(queryset, many=True)`

nautobot/core/api/serializers.py
```diff
@@ -122,13 +127,108 @@ class BaseModelSerializer(OptInFieldsMixin, serializers.HyperlinkedModelSerializ
...
+            - If `device.tenant` is not None, the 'tenant__name' field is set to the actual tenant name.
+
+        Args:
+            lookups: List of natural key lookups
+        """
+        case_query = {}
+        for lookup_field in lookups:
+            # Get the field name of the lookup, e.g the field_name for this lookup `location__parent__parent__name` is `location__parent__parent`
+            *field_name, _ = lookup_field.split("__")
+            field_name = "__".join(field_name)
```
jathanism, line 190: So you're really just getting ride of the last field in the lookup expression?

nautobot/core/api/serializers.py
```diff
@@ -233,6 +338,45 @@ def build_field(self, field_name, info, model_class, nested_depth):
...
+                    data[key] = CSV_NON_TYPE
+                else:
+                    data[key] = value
+        return data
+
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+        altered_data = {}
+
+        if self._is_csv_request() and self.natural_keys_values:
```
jathanism, line 365: Since `natural_keys_values` is defaulted to `None` this is more dogmatic.

```suggestion
        if self._is_csv_request() and self.natural_keys_values is not None:
```

nautobot/core/api/views.py
```diff
@@ -208,8 +208,8 @@ def get_serializer(self, *args, **kwargs):
     def get_serializer_context(self):
         context = super().get_serializer_context()
         if "text/csv" in self.request.accepted_media_type:
-            # CSV rendering should always use depth 1
-            context["depth"] = 1
+            # CSV rendering should always use depth 0
+            context["depth"] = 0
```
jathanism, line 212: I'm not clear on why this changed. Can you please provide a brief explanation?

jathanism (commented):
Awaiting further updates. Just some basic sentiment for now.

nautobot/core/api/serializers.py
timizuoebideri1: Yup

nautobot/core/api/views.py
timizuoebideri1: I suppose it was required in the past so that `CSVParser` could access data from FK fields like "url," "composite_key," and so on. But as we no longer require access to these fields, we no longer need `depth=1`

Before:
```
device,status,name
device_composite_key,Active,Interface
```

After:
```
name,status,device__name,device__location__name,...
Interface One,Active,Device Name,device location name
```

nautobot/core/tests/test_csv.py
```diff
@@ -0,0 +1,286 @@
...
+from nautobot.dcim.models.devices import Device, DeviceType
+from nautobot.dcim.models.locations import Location
+from nautobot.extras.models.roles import Role
+from nautobot.extras.models.statuses import Status
+from nautobot.extras.models.tags import Tag
+from nautobot.tenancy.models import Tenant
+from nautobot.users.factory import UserFactory
+
+
+class CSVParsingRelatedTestCase(TestCase):
```
timizuoebideri1, line 17: This Test covers all use case I can think of:

- FK field with some missing `natural_key_lookup` values i.e `ObjectNotFound`
- FK field = None i.e all FK field lookups are `ObjectNotFound`
- FK fields without any missing lookup value

nautobot/core/api/parsers.py
```diff
@@ -103,6 +105,34 @@ def insert_nested_dict(keys, value, current_dict):
...
+        grouped_data = {}
+        for lookup, lookup_value in data.items():
+            field_name = lookup.split("__", 1)[0]
+            grouped_data.setdefault(field_name, {}).update({lookup: lookup_value})
+
+        valid_data = {}
+
+        for lookup_group in grouped_data.values():
+            for lookup, lookup_value in lookup_group.items():
+                if any(value != CSV_OBJECT_NOT_FOUND for value in lookup_group.values()):
```
jathanism, line 130: Is this right? After iterating on `lookup_group.items()` you're still iterating over `lookup_group.values()` for each item? I assume it's right, but can you please add some commenting for this block on why you need to do this? It's not clear from the block of code exactly why. Thank you.

nautobot/core/api/parsers.py
```diff
@@ -156,16 +187,17 @@ def row_elements_to_data(self, counter, row, serializer):
...
-                elif value:
-                    value = value.split(",")
-                    try:
-                        # We have some cases where it's a list of integers, such as in RackReservation.units
-                        value = [int(v) for v in value]
-                    except ValueError:
-                        # Guess not!
-                        pass
+                if value is not None:
+                    if "{" in value or "[" in value:
```
jathanism, line 191: This seems like it's possibly very fragile such as if a string that isn't JSON has either of these characters. Should it perhaps be a `startswith()` check or some other type of type check for the content?

nautobot/core/api/serializers.py
```diff
@@ -130,6 +135,100 @@ def __init__(self, *args, **kwargs):
...
+
+        Args:
+            lookups: List of natural key lookups
+        """
+        case_query = {}
+        for lookup_field in lookups:
+            # Get the field name of the lookup, e.g the field_name for this lookup `location__parent__parent__name` is `location__parent__parent`
+            *field_name, _ = lookup_field.split("__")
+            field_name = "__".join(field_name)
+            default_value = Value(constants.CSV_OBJECT_NOT_FOUND)
```
jathanism, line 191: This is problematic because this `constants` module is meant to refer to `nautobot.core.constants` but the one in the module scope here is `nautobot.core.models.constants`. 

This is confusing at best. We should merge the two constants modules IMO. `nautobot.core.models.constants` should be folded into `nautobot.core.constants` and then we can reference these all from the same place.

I discovered this by trying to export from DeviceType list view, and got this traceback:

```python
  File "/Users/jathan/Library/Caches/pypoetry/virtualenvs/nautobot-b2ttWva6-py3.10/lib/python3.10/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/Users/jathan/Library/Caches/pypoetry/virtualenvs/nautobot-b2ttWva6-py3.10/lib/python3.10/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/Users/jathan/Library/Caches/pypoetry/virtualenvs/nautobot-b2ttWva6-py3.10/lib/python3.10/site-packages/sentry_sdk/integrations/django/views.py", line 67, in sentry_wrapped_callback
    return callback(request, *args, **kwargs)
  File "/Users/jathan/Library/Caches/pypoetry/virtualenvs/nautobot-b2ttWva6-py3.10/lib/python3.10/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/Users/jathan/Library/Caches/pypoetry/virtualenvs/nautobot-b2ttWva6-py3.10/lib/python3.10/site-packages/rest_framework/viewsets.py", line 125, in view
    return self.dispatch(request, *args, **kwargs)
  File "/Users/jathan/sandbox/src/nautobot/nautobot/core/api/views.py", line 260, in dispatch
    return super().dispatch(request, *args, **kwargs)
  File "/Users/jathan/Library/Caches/pypoetry/virtualenvs/nautobot-b2ttWva6-py3.10/lib/python3.10/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "/Users/jathan/Library/Caches/pypoetry/virtualenvs/nautobot-b2ttWva6-py3.10/lib/python3.10/site-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/Users/jathan/Library/Caches/pypoetry/virtualenvs/nautobot-b2ttWva6-py3.10/lib/python3.10/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "/Users/jathan/Library/Caches/pypoetry/virtualenvs/nautobot-b2ttWva6-py3.10/lib/python3.10/site-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "/Users/jathan/Library/Caches/pypoetry/virtualenvs/nautobot-b2ttWva6-py3.10/lib/python3.10/site-packages/rest_framework/mixins.py", line 46, in list
    serializer = self.get_serializer(queryset, many=True)
  File "/Users/jathan/sandbox/src/nautobot/nautobot/core/api/views.py", line 206, in get_serializer
    return super().get_serializer(*args, **kwargs)
  File "/Users/jathan/Library/Caches/pypoetry/virtualenvs/nautobot-b2ttWva6-py3.10/lib/python3.10/site-packages/rest_framework/generics.py", line 110, in get_serializer
    return serializer_class(*args, **kwargs)
  File "/Users/jathan/Library/Caches/pypoetry/virtualenvs/nautobot-b2ttWva6-py3.10/lib/python3.10/site-packages/rest_framework/serializers.py", line 122, in __new__
    return cls.many_init(*args, **kwargs)
  File "/Users/jathan/Library/Caches/pypoetry/virtualenvs/nautobot-b2ttWva6-py3.10/lib/python3.10/site-packages/rest_framework/serializers.py", line 149, in many_init
    child_serializer = cls(*args, **kwargs)
  File "/Users/jathan/sandbox/src/nautobot/nautobot/core/api/serializers.py", line 146, in __init__
    case_query = self._build_query_case_for_natural_key_field_lookup(all_related_fields_natural_key_lookups)
  File "/Users/jathan/sandbox/src/nautobot/nautobot/core/api/serializers.py", line 191, in _build_query_case_for_natural_key_field_lookup
    default_value = Value(constants.CSV_OBJECT_NOT_FOUND)

Exception Type: AttributeError at /api/dcim/device-types/
Exception Value: module 'nautobot.core.models.constants' has no attribute 'CSV_OBJECT_NOT_FOUND'
```

jathanism (changes requested):

glennmatthews:
Needs to be retargeted to `develop` rather than `next`.

nautobot/core/api/parsers.py
```diff
@@ -156,16 +192,17 @@ def row_elements_to_data(self, counter, row, serializer):
...
```
timizuoebideri1, line 196:
```suggestion
                    if value.startswith("{") or value.startswith("["):
```

nautobot/core/api/renderers.py
```diff
@@ -140,18 +140,22 @@ def object_to_row_elements(self, record, *, headers):
...
                 if isinstance(value, set):
                     value = sorted(value)
-                if value and isinstance(value[0], dict) and "composite_key" in value[0]:
+                if value and isinstance(value[0], dict) and ("composite_key" in value[0] or "id" in value[0]):
                     # Multiple nested related objects
                     if value[0].get("generic_foreign_key"):
                         # Multiple *generic* nested related obects
                         value = [COMPOSITE_KEY_SEPARATOR.join([v["object_type"], v["composite_key"]]) for v in value]
-                    else:
+                    elif value[0].get("composite_key"):
```
timizuoebideri1, line 155: Should composite key be removed from CSV in this PR?

nautobot/core/api/serializers.py
```diff
@@ -130,6 +136,116 @@ def __init__(self, *args, **kwargs):
         if not self.is_nested:
             self.Meta.depth = self.context.get("depth", 0)
 
+        # Check if the request is related to CSV export;
+        if self._is_csv_request() and self.instance:
+            # Retrieve the natural key values of related fields in an optimized way.
+            # This is done to ensure smooth CSV export and import processes:
+            #   1. Remove lookup fields that have no instance, e.g., if the Location Parent FK field is missing,
+            #    we remove 'location_parent*' from the CSV.
+            #   2. Replace fields with 'None' values with "NaN" to differentiate between empty and None values during import.
```
timizuoebideri1, line 145:
```suggestion
```

nautobot/core/api/renderers.py
glennmatthews: Yes, I believe so.

nautobot/core/api/parsers.py
jathanism: We could even make it simpler (`startswith()` can take a tuple):

```python
                    if value.startswith(("{", "[")):
```

nautobot/core/constants.py
```diff
@@ -44,3 +44,29 @@
 
 NAUTOBOT_BOOL_ITERATOR_DEFAULT_LENGTH = 8
 NAUTOBOT_BOOL_ITERATOR_DEFAULT_PROBABILITY = 50
+
+
+#
+# CSV Import/Export
+#
+
+CSV_NON_TYPE = "NULL"
```
gsnider2195, line 53: Should this be `CSV_NONE_TYPE`? Or perhaps `CSV_NULL_TYPE`?

nautobot/core/api/parsers.py
```diff
@@ -103,6 +105,39 @@ def insert_nested_dict(keys, value, current_dict):
 
         return result_dict
 
+    def _field_lookups_not_empty(self, field_lookups):
+        """Check if all values of the field lookups dict are not all ObjectNotFound"""
+        return any(value != CSV_OBJECT_NOT_FOUND for value in field_lookups.values())
+
+    def _remove_object_not_found_values(self, data):
+        """Remove all `ObjectNotFound` field lookups from the given data, and swap out 'NaN' and
```
gsnider2195, line 113: Should `NaN` here be the same as the `CSV_NONE_VALUE` constant?

nautobot/core/api/serializers.py
```diff
@@ -130,6 +137,116 @@ def __init__(self, *args, **kwargs):
...
+            else:
+                # We would only need to run one additional query, making this a more efficient method of
+                # obtaining all the natural key values for this instance;
+                queryset = self.Meta.model.objects.filter(pk=self.instance.pk)
+            self.natural_keys_values = queryset.annotate(**case_query).values(
+                *all_related_fields_natural_key_lookups, "pk"
+            )
+
+    def _get_lookup_field_name_and_output_field(self, lookup_field):
+        """Get lookup field name and its corresponding output_field which would be used in building this lookup Case in `_build_query_case_for_natural_key_field_lookup`"""
```
glennmatthews, line 156:
```suggestion
        """
        Get lookup field name and its corresponding output_field.
        
        Used in building this lookup Case in `_build_query_case_for_natural_key_field_lookup`.
        """
```

glennmatthews: Mind also adding an example of the input and output of this function?

nautobot/core/api/serializers.py
```diff
@@ -248,6 +373,47 @@ def build_field(self, field_name, info, model_class, nested_depth):
 
         return super().build_field(field_name, info, model_class, nested_depth)
 
+    def _get_natural_key_lookups_value_for_field(self, field_name, natural_key_field_instance):
+        """Extract natural key field lookups for a specific field name.
+
+        Args:
+            field_name (str): The field name to extract lookups for.
+            natural_key_field_instance (dict): The dict containing natural key field values.
+        """
```
glennmatthews, line 382: Example of input/output here would be helpful.

nautobot/core/api/serializers.py
```diff
@@ -130,6 +137,116 @@ def __init__(self, *args, **kwargs):
...
+            )
+        return case_query
+
+    def _get_related_fields_natural_key_field_lookups(self):
+        """Retrieve a list of field lookups for natural key fields of related models.
+
+        This method iterates through the related fields of the Serializer model,
+        retrieves the natural_key_field_lookups for each related model, and prepends the field name
+        to create a list of field lookups.
+        """
```
glennmatthews, line 224: Example of output here would be helpful.

nautobot/core/views/utils.py
```diff
@@ -140,7 +140,9 @@ def get_csv_form_fields_from_serializer_class(serializer_class):
                 '<a href="https://en.wikipedia.org/wiki/List_of_tz_database_time_zones">available options</a>'
             )
         elif isinstance(field, serializers.ManyRelatedField):
-            if isinstance(field.child_relation, ContentTypeField):
+            if field.field_name == "tags":
+                field_info["format"] = mark_safe('<code>"name,name"</code> or <code>"UUID,UUID"</code>')
+            elif isinstance(field.child_relation, ContentTypeField):
                 field_info["format"] = mark_safe('<code>"app_label.model,app_label.model"</code>')
             else:
                 field_info["format"] = mark_safe(
```
glennmatthews, line 148: Since we're removing composite-key from CSV import/export entirely for now, don't likes 149 and 155 also need updates?

timizuoebideri1: I don't intend to address that in this PR; instead, it should be addressed in a separate issue to remove any references to composite-key.

nautobot/core/api/parsers.py
```diff
@@ -103,6 +105,39 @@ def insert_nested_dict(keys, value, current_dict):
...
+        'CSV_OBJECT_NOT_FOUND' values for `None`.
+
+        If all the lookups for a field are 'CSV_OBJECT_NOT_FOUND', it indicates that the field does not exist,
+        and it needs to be removed to prevent unnecessary database queries.
+
+        Args:
+            data (dict): A dictionary containing field natural key lookups and their corresponding values.
+
+        Returns:
+            dict: A modified dictionary with field lookups of 'CSV_OBJECT_NOT_FOUND' values removed, and 'CSV_NULL_TYPE' and 'CSV_OBJECT_NOT_FOUND' swapped for `None`.
```
glennmatthews, line 123: Should all of these `OBJECT_NOT_FOUND` become `NO_OBJECT`?

nautobot/core/constants.py
```diff
@@ -44,3 +44,29 @@
...
+
+
+#
+# CSV Import/Export
+#
+
+CSV_NULL_TYPE = "NULL"
+CSV_NO_OBJECT = "NoObject"
+# VarbinaryIPField Represents b'NoObject' as `0:4f62:6a65:6374:4e6f:7446:6f75:6e64`
+VARBINARY_IP_FIELD_REPR_OF_CSV_NO_OBJECT = "0:4f62:6a65:6374:4e6f:7446:6f75:6e64"
```
glennmatthews, line 56: This magic value needs to change I think after changing `ObjectNotFound` to `NoObject`?

Though I'd really like to understand why we're even *getting* to the point where Nautobot is trying to interpret a string as a byte sequence... feels like we should be intercepting things before it gets there?

timizuoebideri1: We could intercept it here `VarbinaryIPField.to_python`

```python
class VarbinaryIPField(models.BinaryField):
    ...
    def to_python(self, value):
        """Converts `value` to Python (str)."""
        if value == b'NoObject':
            return "NoObject"

        if isinstance(value, netaddr.IPAddress):
            return str(value)

        if value is None:
            return value

        return str(self._parse_address(value))
```

But I'm just wondering whether this is right because ordinarily something like to this would never happen.

glennmatthews: That still feels lower level than it should be - we should be catching this value in the parser/serializer, not in the model field, IMO.

timizuoebideri1: We currently intercept it in the serializer converting it to NoObject

nautobot/core/api/renderers.py
glennmatthews: Can/should we remove this at this time?

nautobot/core/api/serializers.py
```diff
@@ -121,7 +127,8 @@ class BaseModelSerializer(OptInFieldsMixin, serializers.HyperlinkedModelSerializ
 
     display = serializers.SerializerMethodField(read_only=True, help_text="Human friendly display value")
     object_type = ObjectTypeField()
-    composite_key = serializers.SerializerMethodField()
+    # composite_key = serializers.SerializerMethodField()  # TODO: Revisit if we reintroduce composite keys
```
glennmatthews, line 130: Let's add a `changes/4448.removed` to document this removal.

glennmatthews (approved):

jathanism (approved):
With the tests passing and my branch based off of this, it all seems to be ready to go! Thanks @timizuoebideri1!

Issue #4416, opened 2023-09-07
`CablePath matching query does not exist.` when deleting a device with multiple types of connected interfaces.


ebusto:
### Environment
* Nautobot version: 1.6.2
* Python version: 3.11.5
* Database platform, version: PostgreSQL 14
* Middleware(s): N/A

### Steps to Reproduce
1. Create two devices, along with at least two types of interfaces (network, console + console server) in each device.
2. Create cables between both types of interfaces.
3. Try to delete one device.

### Expected Behavior
The device to be successfully deleted.

### Observed Behavior
An exception:
```
<class 'nautobot.dcim.models.cables.CablePath.DoesNotExist'>

CablePath matching query does not exist.
```

ebusto:
Assuming the database doesn't get reset with any regularity, you can see this issue on the public demo instance by trying to delete [this](https://demo.nautobot.com/dcim/devices/c88a2ef4-3de8-4387-a99e-36f77959268e/?tab=main) device.

lampwins:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Nautobot-Core-Product-64a8783b7a3f4b1225973a0c/issues/nautobot/nautobot/4416?planning-poker) with Zenhub @bryanculver @glennmatthews @gsnider2195 @HanlinMiao @jathanism @timizuoebideri1

PR #4449, opened 2023-09-14
Fixed a bug that raises `CablePath matching query does not exist` exception when deleting a device with multiple types of connected interfaces.

HanlinMiao:
# Closes: #4416 
# What's Changed

Added `try...except..`. block to catch `Cable.DoesNotExist` error per discussion in standup.

# TODO

- [ ] Unit, Integration Tests


glennmatthews (commented):
Change looks good to me. Any chance of adding a unit test that would catch this issue?

nautobot/dcim/api/serializers.py
```diff
@@ -173,25 +173,34 @@ class PathEndpointModelSerializerMixin(ValidatedModelSerializer):
 
     @extend_schema_field(serializers.CharField(allow_null=True))
     def get_connected_endpoint_type(self, obj):
-        if obj._path is not None and obj._path.destination is not None:
-            return f"{obj._path.destination._meta.app_label}.{obj._path.destination._meta.model_name}"
+        try:
+            if obj._path is not None and obj._path.destination is not None:
+                return f"{obj._path.destination._meta.app_label}.{obj._path.destination._meta.model_name}"
+        except CablePath.DoesNotExist:
+            pass
```
glennmatthews, line 180: Consider (as @timizuoebideri1 has been doing in some of his recent code changes) using [contextlib.suppress](https://docs.python.org/3/library/contextlib.html#contextlib.suppress) for these, I think that would be for example:

```suggestion
        with contextlib.suppress(CablePath.DoesNotExist):
            if obj._path is not None and obj._path.destination is not None:
                return f"{obj._path.destination._meta.app_label}.{obj._path.destination._meta.model_name}"
```

nautobot/dcim/tests/test_models.py
```diff
@@ -1423,6 +1429,61 @@ def test_create_cable_with_missing_status_connected(self):
 
         self.assertTrue(Cable.objects.filter(id=cable.pk).exists())
 
+    def test_deleting_device_with_multiple_types_of_connected_interfaces_successful(self):
```
glennmatthews, line 1432: Would be good to add a docstring noting that this is testing for #4416.

nautobot/dcim/tests/test_models.py
```diff
@@ -1423,6 +1429,61 @@ def test_create_cable_with_missing_status_connected(self):
...
+        Cable.objects.all().delete()
+        interface1 = Interface.objects.create(device=self.device1, name="ethernet1")
+        interface2 = Interface.objects.create(device=self.device2, name="ethernet2")
+        interface3 = Interface.objects.create(device=self.device1, name="ethernet3")
+        interface4 = Interface.objects.create(device=self.device2, name="ethernet4")
```
glennmatthews, line 1437: Missing a `status` on these Interfaces. Can/should we use the `self.interfaceX` interfaces already created in `setUpTestData` instead of creating new ones?

nautobot/dcim/tests/test_models.py
```diff
@@ -1423,6 +1429,61 @@ def test_create_cable_with_missing_status_connected(self):
 
         self.assertTrue(Cable.objects.filter(id=cable.pk).exists())
 
+    def test_deleting_device_with_multiple_types_of_connected_interfaces_successful(self):
+        Cable.objects.all().delete()
```
glennmatthews, line 1433: Why is this needed?

glennmatthews (commented):
This initial PR should probably be targeted at `develop` rather than `ltm/1.6`; we can then make the call as to whether to backport/cherrypick the fix to the LTM branch after it's merged in `develop`.

nautobot/dcim/tests/test_models.py
HanlinMiao: This is more or less for me trying to start the test with a clean slate. It was necessary because I ran into duplicate cables when creating new cables.

nautobot/dcim/api/serializers.py
```diff
@@ -145,17 +148,18 @@ def get_connected_endpoint(self, obj):
...
-        if obj._path is not None and obj._path.destination is not None:
-            depth = get_nested_serializer_depth(self)
-            return return_nested_serializer_data_based_on_depth(
-                self, depth, obj, obj._path.destination, "connected_endpoint"
-            )
+        with contextlib.suppress(CablePath.DoesNotExist):
+            if obj._path is not None and obj._path.destination is not None:
+                serializer = get_serializer_for_model(obj._path.destination, prefix="Nested")
+                context = {"request": self.context["request"]}
+                return serializer(obj._path.destination, context=context).data
```
glennmatthews, line 155: Changes here look suspect - please double-check.

HanlinMiao: Sorry for the oversight when rebasing, should be fixed now.

glennmatthews (approved):

EdificomSA:
Thanks for the fix, I face the issue on 1.6.4, would it be possible to cherrypick/backport it ?

milan2655:
Hallo!

I'm still facing this bug with the nautobot v1.6.22:

PR #4463, opened 2023-09-15
New UI - Add Live Search

bryanculver:
# Closes: #4474 
# What's Changed

- Add live search of the `q` filter on list view

https://github.com/nautobot/nautobot/assets/31187/791dc0cf-89ce-449a-9790-b0117ae0262e

# TODO

- [ ] Unit, Integration Tests


nautobot/ui/src/components/Navbar.js
```diff
@@ -41,6 +45,35 @@ export function Navbar() {
     const location = useLocation();
     const { data: readyRoutes } = useGetNewUIReadyRoutesQuery();
 
+    const [searchParams, setSearchParams] = useSearchParams();
+
+    const changeHandler = (event) => {
+        let filters = [];
+        if (event.target.value !== "") {
+            filters = [["q", event.target.value]];
+        }
```
norbert-mieczkowski-codilime, line 54: Let's talk code style a bit:

1. Unless really required, most of the time it's better to use `event.currentTarget` rather than `target`. `currentTarget` references the node which the event was hooked up to and, thanks to that, it is guaranteed to be stable. `target`, on the other hand, holds a reference to a node which the event was fired on, which resolves during the application runtime, and makes it unstable. Even though it doesn't make much of a difference in `onChange` handlers, at the very least it guarantees type safety (that is, if TypeScript was used).

2. JavaScript coerces `""` to `false`, so the conditional statement can be simplified to `if (event.currentTarget.value)`.

3. Opinions are divided on this one, but I'd start with a known fact that functional programming doesn't allow variable reassignments or mutations (immutability). While JavaScript is not fully complaint with functional paradigm, functional is still a preferred way of writing JS code for the vast majority of engineers, simply because the lesser mutations and side effects there are in the source code, the easier it is to reason about. So, to make this statement look like functional programming, one could use a `const` and a ternary operator:
   ```
   const filters = event.currentTarget.value ? [["q", event.currentTarget.value]] : [];
   ```

These are just code style comments, the behavior remains unchanged. I do not insist on anything, because it's ultimately up to the team to agree on the code style to keep the source code cohesive, but this is just how a frontend engineer would've wrote it.

nautobot/ui/src/components/Navbar.js
```diff
@@ -41,6 +45,35 @@ export function Navbar() {
...
+        setSearchParams([
+            ...filters,
+            ...[...searchParams].filter(
+                ([searchParamLabel]) =>
+                    !FILTER_RESET_QUERY_PARAMS.includes(searchParamLabel) &&
+                    searchParamLabel !== "q"
```
norbert-mieczkowski-codilime, line 60: That expression would've looked nicer if `"q"` was a constant or part of a constant, as suggested in a previous comment.

nautobot/ui/src/components/Navbar.js
```diff
@@ -41,6 +45,35 @@ export function Navbar() {
...
+            ),
+        ]);
+    };
+
+    const debouncedOnChangeSearchBox = useMemo(() => {
+        return debounce(changeHandler, 100);
+    }, []); // eslint-disable-line react-hooks/exhaustive-deps -- For some reason, eslint doesn't like this dependency array, but it's needed to prevent the debounce from being recreated on every render.
```
norbert-mieczkowski-codilime, line 67: There is a good reason for that - every time the component renders, i.e. it's called (because it's a function), a new `changeHandler` function is created. Function creation can be avoided with memoization provided by some of the React hooks, `useCallback` in this case. Moreover, with dependency array set to `[]`, this code now has a bug, which is caused by this debounce function holding the initial `changeHandler` reference and results in `searchParams` not being updated in it. You can test it, for example, first by changing the page or page size and then searching - you should see the old (initial) params restored. To avoid it, you need to set a valid dependency array here

nautobot/ui/src/components/Navbar.js
norbert-mieczkowski-codilime: Actually, nevermind, after staring at it for a while, I prefer it the way you implemented it

nautobot/ui/src/components/Navbar.js
bryanculver: @norbert-mieczkowski-codilime Can we talk about this tomorrow morning? I ran into an issue with addressing these lint issues all the way down.

nautobot/ui/src/components/Navbar.js
bryanculver: @norbert-mieczkowski-codilime Can we discuss tomorrow morning?

glennmatthews (approved):

timizuoebideri1 (changes requested):
Since the search functionality in the list or detail views is limited to that model, it would be clearer to use `"Search Devices..."`.
From the current implementation I would think `"Search whole app"`

nautobot/ui/src/components/Navbar.js
```diff
@@ -22,25 +28,83 @@ import {
...
+                    filters = [["q", event.target.value]];
+                }
+                setSearchParams([
+                    ...filters,
+                    ...[...searchParams].filter(
+                        ([searchParamLabel]) =>
+                            !FILTER_RESET_QUERY_PARAMS.includes(
+                                searchParamLabel
+                            ) && searchParamLabel !== "q"
+                    ),
```
timizuoebideri1, line 85:
```suggestion
                    ...searchParams.filter(
                        ([searchParamLabel]) =>
                            !FILTER_RESET_QUERY_PARAMS.includes(
                                searchParamLabel
                            ) && searchParamLabel !== "q"
                    ),
```

nautobot/ui/src/components/Navbar.js
```diff
@@ -22,25 +28,83 @@ import {
...
+        const changeHandler = (event) => {
+            if (object_id) {
+                navigate(
+                    `/${app_label}/${model_name}/?q=${event.target.value}`
+                );
+            } else {
+                let filters = [];
+                if (event.target.value !== "") {
+                    filters = [["q", event.target.value]];
+                }
```
norbert-mieczkowski-codilime, line 77: What about immutability here? That part of https://github.com/nautobot/nautobot/pull/4463#discussion_r1328442865 (2. and 3.) hasn't been addressed.

nautobot/ui/src/components/Navbar.js
```diff
@@ -22,25 +28,83 @@ import {
...
+         * and the user will be navigated to the new URL.
+         *
+         * If the user is on a detail view, the user will be navigated to the list view
+         * with the search params in the URL.
+         */
+        const changeHandler = (event) => {
+            if (object_id) {
+                navigate(
+                    `/${app_label}/${model_name}/?q=${event.target.value}`
+                );
```
norbert-mieczkowski-codilime, line 72: It's safer to use `URLSearchParams` prototype to add query params, or at least format them with `encodeURIComponent` function. That way they are ensured to always be valid URL encoded strings, as opossed to raw input values, e.g. `&` vs. `%26`, etc.

```suggestion
                navigate(
                    `/${app_label}/${model_name}/?${new URLSearchParams([
                        ["q", event.target.value],
                    ])}`
                );
```

nautobot/ui/src/components/Navbar.js
```diff
@@ -22,25 +28,83 @@ import {
...
+        object_id,
+        navigate,
+    ]);
+
+    const SearchBox = (
+        <Input
+            placeholder="Search..."
+            defaultValue={searchParams.get("q")}
+            onChange={debouncedOnChangeSearchBox}
+            disabled={app_label && model_name ? false : true} // Disable search box if not on a list or detail view (global search is not yet implemented)
```
norbert-mieczkowski-codilime, line 104: Bool cast is a more elegant way of resolving given condition to either `true` or `false` in JavaScript.

```suggestion
            disabled={!(app_label && model_name)} // Disable search box if not on a list or detail view (global search is not yet implemented)
```

nautobot/ui/src/components/Navbar.js
```diff
@@ -96,13 +160,18 @@ export function Navbar() {
...
-            <InputGroup flex="1" size="lg" variant="navbar">
-                <InputLeftElement>
-                    <SearchIcon />
-                </InputLeftElement>
-                <Input placeholder="Search..." />
-            </InputGroup>
+            <Tooltip
+                label="Global search is not yet implemented."
+                placement="bottom"
+                isDisabled={app_label && model_name ? true : false} // Disable tooltip if on a list or detail view (contextual search is implemented)
```
norbert-mieczkowski-codilime, line 166: Bool cast is a more elegant way of resolving given condition to either `true` or `false` in JavaScript.

```suggestion
                isDisabled={Boolean(app_label && model_name)} // Disable tooltip if on a list or detail view (contextual search is implemented)
```

timizuoebideri1 (approved):

Issue #4396, opened 2023-09-06
Rack object creation fails with mandatory custom field (web form data handling issue)


david-kn:
### Environment
* Nautobot version (Docker tag too if applicable): 1.6.1
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Let's have a custom field that is required for `dcim.rack` model (let's say custom field "owner")
2. Try to create a new rack via web GUI and fill in all mandatory data (including that mandatory custom field).

BUT 
- adding new rack works in `nbshell` when using `validated_save()` (it seems that the problem is not within the validators)
- adding new device works via API also works
e.g. following minimalistic API payload works just fine:
```
{
  "name": "rack1234",
  "site": "a87d0362-3a21-5cee-a147-57dee074e9ab",
  "status": "Active",
  "custom_fields": {
       "owner" : "ownerABC_of_the_rack"
  }
}
```

Hence it seems that the problem is somewhere in the handing the data between a form to the backend. 

Tested also on your demo.nautobot.com instance and received the same error.

### Expected Behavior
A new rack to be created.

### Observed Behavior
Object creation fails with an error message `Missing required custom field 'owner'`

Thetacz:
Note: *any* custom field value from edit form gets discarded, not only required fields

gsnider2195:
Replicated on `demo.nautobot.com` and `next.demo.nautobot.com`

bryanculver:
It appears not to affect all models. Validated in Demo that IP Address works fine.

lampwins:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Nautobot-Core-Product-64a8783b7a3f4b1225973a0c/issues/nautobot/nautobot/4396?planning-poker) with Zenhub @bryanculver @glennmatthews @gsnider2195 @HanlinMiao @jathanism @timizuoebideri1

PR #4466, opened 2023-09-15
Fixed rack form silently dropping custom field values

gsnider2195:
# Closes: #4396
# What's Changed

Small bug fix in the form clean method. This should probably be back ported to 1.6 

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/forms.py
+++ b/nautobot/dcim/forms.py
@@ -459,7 +459,7 @@ class RackForm(LocatableModelFormMixin, NautobotModelForm, TenancyForm):
                         "would conflict with same-named devices in this rack."
                     }
                 )
-        return cleaned_data
+        return super().clean()
 
 
 class RackBulkEditForm(
--- a/changes/4396.fixed
+++ b/changes/4396.fixed
@@ -1,0 +1,1 @@
+Fixed rack form silently dropping custom field values.


nautobot/dcim/forms.py
```diff
@@ -459,7 +459,7 @@ def clean(self):
                         "would conflict with same-named devices in this rack."
                     }
                 )
-        return cleaned_data
+        return super().clean()
```
glennmatthews, line 462: In a lot of cases `form.clean()` doesn't actually return anything, as the "outcome" of the function is just to populate `self.cleaned_data`. What happens if we just remove this line?

glennmatthews: I think the actual issue is that we were never calling `super.clean()` in this function.

glennmatthews (approved):

nautobot/dcim/forms.py
gsnider2195: Yep. The `CustomFieldModelFormMixin` clean method is what is actually taking the form custom field data and adding it to the instance. The `RackForm` clean method didn't exist. When it was added without `super().clean()` it introduced this bug.

glennmatthews:
also :+1: on backporting this fix to 1.6.

PR #4470, opened 2023-09-17
Enforce upper on constance and lower on plugin values

itdependsnetworks:
# Closes: DNE
# What's Changed

Changed `get_app_settings_or_config` to enforce upper on constance and lower on plugin values per standards.

The standards have been in all plugins I have checked for lower in plugin settings and constance enforces upper. This seemed liked the preferred opinionated approach from NTC. Others can build their own function, but this seems correct to me for how we operate.  

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/apps/config.py
+++ b/nautobot/apps/config.py
@@ -7,6 +7,6 @@ from django.conf import settings
 def get_app_settings_or_config(app_name, variable_name):
     """Get a value from Django settings (if specified there) or Constance configuration (otherwise)."""
     # Explicitly set in settings.py or nautobot_config.py takes precedence, for now
-    if variable_name in settings.PLUGINS_CONFIG[app_name]:
-        return settings.PLUGINS_CONFIG[app_name][variable_name]
-    return getattr(config, f"{app_name}__{variable_name}")
+    if variable_name.lower() in settings.PLUGINS_CONFIG[app_name]:
+        return settings.PLUGINS_CONFIG[app_name][variable_name.lower()]
+    return getattr(config, f"{app_name}__{variable_name.upper()}")
--- a/changes/4470.changed
+++ b/changes/4470.changed
@@ -1,0 +1,1 @@
+Changed `get_app_settings_or_config` to enforce upper on constance and lower on plugin values per standards.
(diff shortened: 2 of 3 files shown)


glennmatthews (commented):
If we're going to code-enforce this convention, it needs to be clearly documented as such, probably in `nautobot/docs/development/apps/api/database-backend-config.md`.

nautobot/apps/config.py
```diff
@@ -7,6 +7,6 @@
 def get_app_settings_or_config(app_name, variable_name):
     """Get a value from Django settings (if specified there) or Constance configuration (otherwise)."""
     # Explicitly set in settings.py or nautobot_config.py takes precedence, for now
-    if variable_name in settings.PLUGINS_CONFIG[app_name]:
-        return settings.PLUGINS_CONFIG[app_name][variable_name]
-    return getattr(config, f"{app_name}__{variable_name}")
+    if variable_name.lower() in settings.PLUGINS_CONFIG[app_name]:
+        return settings.PLUGINS_CONFIG[app_name][variable_name.lower()]
+    return getattr(config, f"{app_name}__{variable_name.upper()}")
```
glennmatthews, line 12: Should this be `f"{app_name}__{variable_name}".upper()` instead? Seems weird to me to force upper case for half of the name and expect lower case for the other half.

itdependsnetworks: I would defer to you, I thought this was a constance thing that we as users didn't have control over. When I tested creating it, this is the format it used.

* Constance says, "must be capital"
* I never set anything on the app_name, other than potentially what `nautobot_golden_config.confi.name` is??

itdependsnetworks:
> If we're going to code-enforce this convention, it needs to be clearly documented as such, probably in nautobot/docs/development/apps/api/database-backend-config.md

There is no documentation on this function. The only reason I mention this is, this is not exactly correct `code-enforce this convention`, we code-enforce this convention `when you use get_app_settings_or_config`, which you are free to not use. More than happy to add the documentation, but just want to ensure that is understood.

bryanculver:
We can re-open this PR up later but we should address #4530 first.

PR #4479, opened 2023-09-18
Fix to Failing ARM CI `npm install`

snaselj:
# Closes: #NaN

Potential fix for failing npm install on CI.

https://github.com/nautobot/nautobot/actions/runs/6202886484/job/16851734010#step:9:3048

## What's Changed

- Added 30s timeout to `npm` configuration
    - Probably just a good practice anyway
- Install `npm` 9.x instead of `latest`
    - This was the smoking gun with regards to our issue. Something with the way the new network agent in npm 10 works. Worked when the runs were faster, but for now (see Actions history) it solved the issue. (see related discussions here: https://github.com/npm/cli/issues/4652)
- Move to `npm ci` instead of `npm install`
    - This is because `npm ci` installs against the lock file vs the `package.json` file which can cause non-reproducible builds (see: https://docs.npmjs.com/cli/v9/commands/npm-ci?v=true#description)


changes/4479.changed
```diff
@@ -0,0 +1 @@
+Configured `npm` in Docker image to use 30s timeout, pin `npm` to 9.X, and use `npm ci` instead of `npm install` to improve builds.
```
glennmatthews, line 1:
```suggestion
Configured `npm` in Docker image to use 30s timeout, pinned `npm` to 9.X, and changed Docker build to use `npm ci` instead of `npm install` to improve builds.
```

docker/Dockerfile
```diff
@@ -67,8 +67,12 @@ RUN --mount=type=cache,target="/var/cache/apt",sharing=locked \
     apt-get install --no-install-recommends -y git mime-support libxml2 libmariadb3 openssl nodejs && \
     apt-get autoremove -y
 
+# Ensure the npm install is cached, updated, and has a long timeout for slow connections
+# The Node.JS install aboce should install npm 9, but we pin it here just in case
+# npm 10 has shown issues with slow CI environments
```
glennmatthews, line 72: In our installation docs should we recommend users to stay with npm 9 for Nautobot 2.0?

bryanculver: I can update that

changes/4479.changed
```diff
@@ -0,0 +1 @@
+Configured `npm` in Docker image to use 30s timeout, pinned `npm` to 9.X, and changed Docker build to use `npm ci` instead of `npm install` to improve builds.
```
gsnider2195, line 1: Should this go into the housekeeping section of the release notes?

docker/Dockerfile
```diff
@@ -67,8 +67,12 @@ RUN --mount=type=cache,target="/var/cache/apt",sharing=locked \
     apt-get install --no-install-recommends -y git mime-support libxml2 libmariadb3 openssl nodejs && \
     apt-get autoremove -y
 
+# Ensure the npm install is cached, updated, and has a long timeout for slow connections
+# The Node.JS install aboce should install npm 9, but we pin it here just in case
```
gsnider2195, line 71:
```suggestion
# The Node.JS install above should install npm 9, but we pin it here just in case
```

changes/4479.changed
bryanculver: 🤷 It is a change to something the user will use but I am OK with moving it to housekeeping.

jathanism (approved):
:shipit:

glennmatthews (approved):

gsnider2195 (approved):

Issue #3689, opened 2023-05-03
IPAM Documentation: Visual - What are Namespaces? What's changed with other models?


bryanculver:
User story explaining the design idea behind Namespaces and how they are to be used in practice.

bryanculver:
Much of this work was done in #4151, but adding a graph/diagram to help visualize the change would be a benefit.

bryanculver:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Nautobot-Core-Engineering-64a875e74842500d9e80e23e/issues/nautobot/nautobot/3689?planning-poker) with Zenhub @gsnider2195 @HanlinMiao @jathanism @timizuoebideri1

PR #4485, opened 2023-09-19
Added namespace diagram to docs

gsnider2195:
# Closes: #3689 
# What's Changed

image

# TODO

- [ ] Unit, Integration Tests


glennmatthews (commented):
It might be clearer to extend the "MPLS Backbone" namespace to the right so that it includes the entirety of the 100.64.x.y links and partially intersects with the set of devices on the right. This would make it clearer that multiple Namespaces may be used on a single Device, as well.

Also, should there be some links between the Corp and MPLS Backbone namespaces?

jathanism (approved):
:shipit:

Issue #4458, opened 2023-09-15
Improve New UI Container Behavior, Filter Panel Padding


bryanculver:
### Proposed Changes

- Page content container is always scrollable, even when the content does not take the whole browser viewport. This is pictured on the first two images. To fix it, it would either require sizing the content container to exactly fit the content or just compensating the pixel differences not to display the scroll when not necessary. Note that the generic layout components can be configurable and the same solution does not have to apply to all pages, eg. home vs. list view.
- Page header and content do not have any space between them and when the content is scrolled down, the two may look glued to each other. This is pictured on the 3rd image. To fix it, we can add a standard 20px gap between the header and the content.
- ~Tables should have sticky headers, so when the lists are scrolled down, table headers are still visible. The problem is picutred on the 4th image.~ Moved -> https://github.com/nautobot/nautobot-ui/issues/39

### Justification

Polish the new UI from original implementation.

![image_4](https://github.com/nautobot/nautobot/assets/31187/b210cf25-c0ea-4f4a-8f41-8248c5c4dcd5)
![image_3](https://github.com/nautobot/nautobot/assets/31187/39362694-6246-46da-abff-d328fec7899a)
![image_2](https://github.com/nautobot/nautobot/assets/31187/43124840-5e24-457e-85a1-5f2a157676e3)
![image](https://github.com/nautobot/nautobot/assets/31187/6d67386e-5a50-4eb2-b56e-764e65fad440)

PR #4487, opened 2023-09-20
fix: general ui layout sizing and spacing

norbert-mieczkowski-codilime:
# Closes: #4458
# What's Changed

Fix general UI layout sizing and spacing on Home (`Home`), List (`ObjectList`) and Object (`ObjectRetrieve`) views.

| Home (`Home`) | List (`ObjectList`) | Object (`ObjectRetrieve`) |
| - | - | - |
| image | image_2 | image_3 |


nautobot/ui/src/components/Retrieve/RenderHeader.js
```diff
@@ -21,47 +22,63 @@ import { humanFriendlyDate } from "@utils/date";
 
 export default function RenderHeader({ data }) {
     return (
-        <Box display="flex" justifyContent="space-between" padding="md">
-            <Heading display="flex" alignItems="center" gap="5px">
+        <Flex justify="space-between">
```
HanlinMiao, line 25: Should we be using `Flex` instead of `Box` from now on? What are the differences?

nautobot/ui/src/components/Retrieve/RenderTabs.js
```diff
@@ -12,8 +19,8 @@ function RenderTabs({
     isPluginView,
 }) {
     return (
-        <Tabs>
-            <TabList pl="md">
+        <Tabs variant="outline">
+            <TabList marginX={negateCssVar("md")} paddingX="md">
```
HanlinMiao, line 23: Can you offer a short explanation for `negateCssVar`?

HanlinMiao (approved):
Changes looked good when I pulled it down.

nautobot/ui/src/components/Retrieve/RenderHeader.js
norbert-mieczkowski-codilime: The difference between `Flex` and `Box` is esentially that `Box` has, by default, `display: block;`, whereas `Flex` has `display: flex;`. Outside (i.e. for their immediate parents) `block` and `flex` are the same, but inside, their children flow is fundamentally different. MDN can offer more detailed explanation: https://developer.mozilla.org/en-US/docs/Learn/CSS/CSS_layout/Flexbox

nautobot/ui/src/components/Retrieve/RenderTabs.js
norbert-mieczkowski-codilime: If raw values were used, negation would be as simple as adding a `-` sign in front of a number (e.g. `-20px`). However, in case of space tokens (which resolve to CSS custom properties), such as `xs`, `sm`, `md`, etc., the negation needs to be somewhat more sophisticated. So what `negateCssVar` does is, it takes space token and multiplies it by `-1` with CSS `calc` function. To put it simple, you can think of `negateCssVar("md")` as something like `-md` or `md * -1`. In reality it is `calc(var(--chakra-space-md), -1)`.

PR #4513, opened 2023-09-22
Add a meta class for BaseJob so that jobs no longer need to be manually registered (2.0)

gsnider2195:
# Closes: #DNE
# What's Changed

```diff
-class BaseJob(Task):
+class JobMeta(type):
+    """Job metaclass to automatically register subclasses in celery."""
+
+    def __init__(cls, name, bases, dct):
+        if not cls.__dict__.get("abstract", False):
+            register_job(cls)
+
+
+class BaseJob(Task, metaclass=JobMeta):
    """Base model for jobs.
    Users can subclass this directly if they want to provide their own base class for implementing multiple jobs
    with shared functionality; if no such sharing is required, use Job class instead.
    Jobs must define at minimum a run method.
    """

+    abstract = True
+
    class Meta:
```

Maybe also get rid of the plugin_jobs registry here since:

- Plugins don't need to iterate over their jobs to run `register_jobs`
- The celery task registry is effectively replacing `registry["plugin_jobs"]`

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/jobs.py
```diff
@@ -85,7 +85,15 @@ class RunJobTaskFailed(Exception):
     """Celery task failed for some reason."""
 
 
-class BaseJob(Task):
+class JobMeta(type):
+    """Job metaclass to automatically register subclasses in celery."""
+
+    def __init__(cls, name, bases, dct):
+        if not cls.__dict__.get("abstract", False):
```
glennmatthews, line 92: We should document this as a feature -- apps like SSOT will want/need to be able to define their own abstract job subclasses as well, and now that job registration has changed from opt-in to opt-out this will need to be handled by them.

gsnider2195: I wonder if we should update this to look for `nautobot.extras.Job` in `bases` so that the abstract attr isn't required?

glennmatthews: That'd be closer to the existing behavior but I like the idea of flagging things explicitly rather than relying on base class selection to inform magic behavior (I think `abstract` would actually make it so we can get rid of `BaseJob` class entirely in a later release?)

gsnider2195: Sounds like a good plan to me. Should we consider this a breaking change and push it back to 3.0 or target to next for 2.1?

timizuoebideri1 (approved):

PR #4512, opened 2023-09-22
Resolved KeyError on custom actions 

whitej6:
# Closes: #4511
# What's Changed
* provides ability to overload permissions on individual actions

# TODO

- [ ] Explanation of Change(s)
- [ ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [ ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example Plugin Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #4514, opened 2023-09-25
Use `alter_queryset` in All Uses of QuerySet in `NautobotHTMLRenderer`

itdependsnetworks:
# Closes: #DNE
# What's Changed

* Fixed an issue with alter_queryset not being implemented when constructing a table.
* While #4528 exists that would be a breaking change. This PR makes us self-consistent for eventually moving toward only `get_queryset` being the way to get the QuerySet and overloading it to change it..

# TODO


itdependsnetworks:
I can add docs if so desired for the DynamicRoute update, please let me know if that will be required.

Can someone take a look at the alter_queryset logic? This is the second issue I ran into and probably worth 15 minutes to review other places this needs to be applied.

HanlinMiao:
> I can add docs if so desired for the DynamicRoute update, please let me know if that will be required.
> 
> Can someone take a look at the alter_queryset logic? This is the second issue I ran into and probably worth 15 minutes to review other places this needs to be applied.

Docs would be super helpful if you have time to add it. A test case would also be desirable although I am not sure how to test the DyanmicRoute.

itdependsnetworks:
I did not see any tests on the prior so I presume it was problematic. I will add the docs.

nautobot/core/views/routers.py
```diff
@@ -52,6 +52,13 @@ class NautobotUIViewSetRouter(SimpleRouter):
             detail=False,
             initkwargs={"suffix": "Bulk Delete"},
         ),
+        # DynamicRoute that will handle custom actions where detail=False
+        DynamicRoute(
+            url=r"^{prefix}/{url_path}{trailing_slash}$",
```
glennmatthews, line 57: Would be nice to add a custom action to the example-plugin that demonstrates this functionality - would probably be useful in test writing too.

nautobot/core/views/renderers.py
```diff
@@ -99,7 +99,7 @@ def construct_table(self, view, **kwargs):
             return RequestConfig(request, paginate).configure(table)
         else:
             pk_list = kwargs.get("pk_list", [])
-            table = table_class(queryset.filter(pk__in=pk_list), orderable=False)
+            table = table_class(view.get_queryset().filter(pk__in=pk_list), orderable=False)
```
glennmatthews, line 102: Why not `alter_queryset` here too?

nautobot/core/views/routers.py
itdependsnetworks: sounds good, can you give me some hints where and what the test should look like? Struggling to think about how to go about such a test.

nautobot/core/views/renderers.py
gsnider2195: The else statement here would apply to the `bulk_destroy` and `bulk_update` actions. In the legacy views `alter_queryset` is only called on the `ObjectListView` view and it's not being called on the detail view notes or changelog tabs. However, `alter_queryset` is defined on `NautobotViewSetMixin` so it is available for all of the actions (`retrieve`, `list`, `destroy`, `create`, `update`, `bulk_destroy`, `bulk_create`, `bulk_update`, `changelog` and `notes`). I'm leaning toward calling it for every action here?

Also, the only documentation I'm seeing for `alter_queryset` is on [this line](https://github.com/nautobot/nautobot/blob/5db9249dad4a18cc66ca6794dfbf8c797c1e2cef/nautobot/core/views/generic.py#L254). We should probably document the `NautobotViewSetMixin.alter_queryset()` method with its intended use cases.

nautobot/core/views/routers.py
HanlinMiao: A custom view for `ExampleModel` should probably look something like this https://github.com/nautobot/nautobot/compare/u/hanlin-demo?expand=1. However, this does not work there because it seems like predefined static routes are taking precedence over dynamic routes. This is going to `{prefix}/{lookup}` instead of `{prefix}/{url_path}`.
Screen Shot 2023-09-25 at 11 56 01 PM

Also a test case should look similar to this, basically asserting response code 200 and that the content is rendered correctly.
```
def test_list_objects_with_permission(self):
            # Add model-level permission
            obj_perm = users_models.ObjectPermission(name="Test permission", actions=["view"])
            obj_perm.save()
            obj_perm.users.add(self.user)
            obj_perm.object_types.add(ContentType.objects.get_for_model(self.model))

            # Try GET with model-level permission
            response = self.client.get(self._get_url("list"))

            list_url = self.get_list_url()
            title = self.get_title()

            # Check if breadcrumb is rendered correctly
            self.assertIn(
                f'<a href="{list_url}">{title}</a>',
                response_body,
            )

```

itdependsnetworks: Are you getting that error when not setting url_path?

HanlinMiao: Yes

itdependsnetworks: ok, took me a bit, I didn't realize this wasn't based on my PR. Notice the difference in the order of your dynamicroute vs mine. Once it goes from `details=False` to `details=True` you will run into the issue you are seeing. Move the dynamicroute up one and you should be fine.

HanlinMiao: I'm running into the same issue here with the reordering. https://github.com/nautobot/nautobot/compare/u/hanlin-demo?expand=1.

HanlinMiao: Sorry about the confusion, I added the custom action to the APIViewSet instead of the UIViewSet. Running into a different issue now.

glennmatthews: Tagging #4512 for traceability.

bryanculver:
I carved out the DynamicRoute feature over to https://github.com/nautobot/nautobot/tree/u/itdependsnetworks-4529-custom-actions-dynamic-routes, and we should group with https://github.com/nautobot/nautobot/pull/4512/files to address #4529

changes/4514.fixed
```diff
@@ -0,0 +1 @@
+Fixed an issue with alter_queryset not being implemented when constructing a table.
```
glennmatthews, line 1:
```suggestion
Fixed an issue with `alter_queryset` not being called when constructing a table.
```

glennmatthews (approved):

Issue #4014, opened 2023-06-28
Refine Prefix View in New UI


bryanculver:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBblZiIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--a4ced34542c91e88247c79fb995a2a58125722fe/Screenshot%202023-08-31%20at%202.40.48%20PM.png)

Replicate the 2.0 legacy list view but do not include the child indentation or the child counts.

For the detail view:
- Change "Ip version" > "IP version"
- Ignore tags, custom fields, computed fields, relationships, etc (they will be solved generically)

Detail view field order:
- Prefix
- Parent
- Namespace
- Type
- Network
- Broadcast
- Prefix length
- IP Version
- Description
- Role
- VLAN
- Location
- Tenant
- RIR
- Date allocated

bryanculver:
@bryanculver Get a mockup uploaded for this view and #4015, #4016

bryanculver:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Nautobot-Core-Product-64a8783b7a3f4b1225973a0c/issues/nautobot/nautobot/4014?planning-poker) with Zenhub @glennmatthews @gsnider2195 @HanlinMiao @jathanism @timizuoebideri1

PR #4520, opened 2023-09-25
Refine Prefix Views in new UI.

HanlinMiao:
# Closes: #4014 
# What's Changed

List View
Screen Shot 2023-09-25 at 4 07 06 PM
Detail View
Screen Shot 2023-09-25 at 4 31 01 PM
1. added `namespace`, `children` and `utilization` columns to PrefixListView table.
2. render `ip_version` field correctly in PrefixDetailView.
3. I took the liberty to make the first column of the table to have capitalized letters immediately after the separator. It looks nicer and it is the same way in the old new UI: `Child location type` -> `Child Location Type`.
4. Reordered PrefixDetailView table Fields.
# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/ipam/api/serializers.py
+++ b/nautobot/ipam/api/serializers.py
@@ -1,5 +1,6 @@
 from collections import OrderedDict
 
+from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
 from rest_framework.exceptions import ValidationError
 from rest_framework.validators import UniqueTogetherValidator
@@ -144,14 +145,19 @@ class VLANSerializer(NautobotModelSerializer, TaggedModelSerializerMixin):
 class PrefixSerializer(NautobotModelSerializer, TaggedModelSerializerMixin):
     prefix = IPFieldSerializer()
     type = ChoiceField(choices=PrefixTypeChoices, default=PrefixTypeChoices.TYPE_NETWORK)
+    children = serializers.SerializerMethodField(read_only=True)
+    utilization = serializers.SerializerMethodField(read_only=True)
 
     class Meta:
         model = Prefix
         fields = "__all__"
         list_display_fields = [
             "prefix",
+            "namespace",
             "type",
             "status",
+            "children",
+            "utilization",
             "vrf",
             "tenant",
             "location",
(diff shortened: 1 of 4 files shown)


nautobot/ipam/api/serializers.py
```diff
@@ -165,6 +167,50 @@ class Meta:
...
+    @extend_schema_field(serializers.CharField)
+    def get_utilization(self, instance):
+        """
+        Return the child count of a prefix
+        """
+        utilization_data = instance.get_utilization()
+        numerator = utilization_data[0]
+        denominator = utilization_data[1]
+        utilization = round((numerator / denominator * 100), 1)
+        return str(utilization) + " %"
```
glennmatthews, line 212: How's the performance impact of these fields on the REST API at scale? Should we make them opt-in fields instead of always including them?

nautobot/ui/src/utils/string.js
```diff
@@ -1,6 +1,12 @@
 export function toTitleCase(str, separator = " ") {
     return str
         .split(separator)
-        .map((x) => (x ? x[0].toUpperCase() + x.slice(1) : ""))
+        .map((x) =>
+            x
+                ? x === "Ip"
```
glennmatthews, line 6: Is that the best approach we can do? We'll have to solve the same problem for `vlan`, `vrf`, `vm interface`, etc. Would it be better to do something through the detail_view_config or list_display_fields in some way?

nautobot/ipam/api/serializers.py
HanlinMiao: I didnt realize performance was the reason we excluded them from the story, so I removed those two fields.

nautobot/ui/src/utils/string.js
HanlinMiao: Let me know if this is an acceptable way of doing this. https://github.com/nautobot/nautobot/pull/4520/commits/aafb031fc12ded8c1c80863ae51c09e7ec14aa06#diff-7eb4de91128d426a0c3b51d50705d7ba4b1033169ec88617031657896728d8e1R73-R75.
Basically by using the "label" attribute on the serializer field for display in the UI.

glennmatthews: Yes! definitely the direction we should head.

timizuoebideri1 (changes requested):
Ulld have to add `use_new_ui = True` to the `PrefixView` and `PrefixListView` class, so that in `prod` it would not be redirected to legacy UI

timizuoebideri1 (approved):

nautobot/ipam/api/serializers.py
```diff
@@ -160,11 +161,37 @@ class Meta:
             "description",
         ]
         extra_kwargs = {
-            "ip_version": {"read_only": True},
+            "ip_version": {"read_only": True, "label": "IP Version"},
```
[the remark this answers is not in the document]
glennmatthews: Did you try setting on the model field instead of the serializer? I think that's how `vlan` and `rir` are currently set up on Prefix and it looks like those are working in your screenshot without needing this here.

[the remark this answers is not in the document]
HanlinMiao: I see yeah setting `verbose_name` worked!

nautobot/ipam/api/serializers.py
```diff
@@ -165,6 +166,32 @@ class Meta:
...
+                            "description",
+                            "role",
+                            "vlan",
+                            "location",
+                            "tenant",
+                            "rir",
+                            "date_allocated",
+                        ]
+                    },
+                },
```
glennmatthews, line 191: Do we need to add this to get tags to appear for now?

```
              {"Tags": {"fields": ["tags"]}}
```

HanlinMiao: I tweaked `validated_view_config` method to add `Tags` field automatically if it is not added already.

glennmatthews: I'm doing something very similar in #4532 as part of a broader refactor, let's stay in sync here.

HanlinMiao:
Blocked by https://github.com/nautobot/nautobot/pull/4532

glennmatthews (approved):

PR #4524, opened 2023-09-26
Fix regression introduced in #4517 and add test coverage

glennmatthews:
# Closes: #n/a
# What's Changed

#4517 introduced a bug into the REST API object `/notes/` endpoints in that they would error out when called *unless* the viewset in question explicitly declared a `lookup_url_kwarg` attribute. (JobViewSet does this, so I didn't catch this issue in my testing on #4517, but most other views do not). The simple solution here is to (as DRF does in most of its own code) fail through to `lookup_field` if `lookup_url_kwarg` is unset.

However this uncovered that we don't actually *have* any automated test coverage of the `/notes/` REST API endpoints yet, so I added a new test case to exercise that feature as part of our generic REST API testing.

This uncovered that the existing `test_notes_url_on_object` test case in the generic REST API testing was actually never testing anything, because it was checking whether `model.notes` is a `ForeignKey` (which it never is) before executing the test, instead of checking whether `model_instance.notes` is a `QuerySet` (which it should be). So I fixed that as well.

I also updated the permissions associated with the REST API object `/notes/` endpoint to check for note-related permissions in addition to object-related permissions, and to check for `view_<object>` instead of `add_<object>` permission when creating a Note on that object.

# TODO

- [x] Unit, Integration Tests


nautobot/extras/api/views.py
```diff
@@ -80,18 +80,37 @@ def get_view_name(self):
...
+            self.queryset = self.queryset.restrict(request.user, "view")
+        else:
+            super().restrict_queryset(request, *args, **kwargs)
+
+    class CreateNotePermissions(TokenPermissions):
+        """As nautobot.core.api.authentication.TokenPermissions, but enforcing add_note permission."""
+
+        perms_map = {
+            "GET": ["%(app_label)s.view_%(model_name)s", "extras.view_note"],
+            "POST": ["%(app_label)s.view_%(model_name)s", "extras.add_note"],
```
HanlinMiao, line 97: Do we need `"%(app_label)s.view_%(model_name)s"` for the `POST` action here? (Do we need permission to view the model in order to add notes to it?)

glennmatthews: I was thinking so - you shouldn't be able to add notes to an object that you can't see? Open to discussion on this point as this was just my best guess as to how it "should" work.

Issue #4425, opened 2023-09-11
Further refine device detail view


lampwins:
As a Network Engineer, I want the "other" grouping of fields in the device detail view to be redistributed and/or hidden, so that I am not confused by their meaning, or the meaning of the "other" grouping in general.

I will know this is complete when the "other" field grouping is removed the visible fields more closely match the legacy device detail view.

lampwins:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Nautobot-Core-Engineering-64a875e74842500d9e80e23e/issues/nautobot/nautobot/4425?planning-poker) with Zenhub @bryanculver @glennmatthews @gsnider2195 @jathanism @timizuoebideri1

PR #4532, opened 2023-09-26
Refactor list/detail view schema logic to serializers. Refine Device detail view

glennmatthews:
# Closes: #4425 
# What's Changed

- Move most of the logic for constructing the new-UI schemas for "list" and "detail" views from the `NautobotMetadata` class to the `BaseModelSerializer` class. 
  - This makes it possible for individual serializers to extend or override this logic where needed
  - I think it conceptually makes a lot more sense, as well, as most of this logic required passing around a Serializer instance within the Metadata class while not having any particular dependency on the Metadata class itself. 
  - Most of the methods within this logic are private and are flagged (`_`-prefixed) as such
  - Model serializers can override `get_additional_detail_view_tabs` if they want to move some model fields to a tab other than the main/"Advanced" tabs provided by default.
  - `tags` and `comments` fields are automatically included (if present on the serializer) in the right column of the main tab, without needing to be explicitly specified in the `detail_view_config` for each serializer.
  - Changed `list.fields` and `list.list_display_fields` schema keys to `list.all_fields` and `list.default_fields` for clarity
  - Moved `view_options.retrieve` and `view_options.advanced` schema keys under a common `view_options.tabs` key that can now contain additional tabs if defined.
- Modify Device detail view schema
  - Add `name`, `face`, `cluster`, `parent_bay`, `device_redundancy_group_priority` fields to the appropriate groupings
  - Move virtual-chassis information to a separate tab
  - Change `include_others` to False so that config-context-related fields are omitted from the device detail UI for now.

image

image

image

# TODO

- [ ] Unit, Integration Tests


HanlinMiao (approved):
Looks great! Wonder if we should add testing for adding additional tabs and overriding advanced tab functionalities.

HanlinMiao (commented):
Also should probably re-enable Device Views with `use_new_ui` see the comment here. https://github.com/nautobot/nautobot/pull/4520#pullrequestreview-1643856388. After that, modify `test_new_ui_ready_routes()`.

glennmatthews:
> Also should probably re-enable Device Views with `use_new_ui` see the comment here. [#4520 (review)](https://github.com/nautobot/nautobot/pull/4520#pullrequestreview-1643856388). After that, modify `test_new_ui_ready_routes()`.

Looks like `use_new_ui` was already added to these in cf3053a7b30934a34d1794e0a14c9dea7b7b7eab but thanks for the check!

timizuoebideri1 (approved):

glennmatthews:
> Looks great! Wonder if we should add testing for adding additional tabs and overriding advanced tab functionalities.

Good call, added.

Issue #4459, opened 2023-09-15
Second pass at nautobot.apps exports


lampwins:
As a Patti the Plugin Developer, I want the most relevant items from the core codebase to be exported for official use in the nautobot.apps namespace, so I know I am using a blessed interface and so that my imports are consolidated.

I will know this is complete when the core team takes a fresh pass over the core code base to decern items that should reasonably be exported to the apps namespace. The goal will be to export the broadest set of features, leaving only those things that the team has a strong, justifiable reason for not exporting, out of the apps namespace. Obviously if an item is positively known to be currently used by an app, it should be included, but the point of this story is to use good judgement to include things that are either not positively known, or that may reasonably be used in the future.

The app developer and migration documentation should also be updated to clarify the state and purpose of the apps namespace.

There has been some discussion about including models in the namespace and it was not entirely clear if that would pose challenges with FKs or migrations. Unless we feel confident at this time that those issues are resolved, models will be the primary exception to the namespace exports and the docs will clearly spell this out. Model dependent items, such as Choice classes will remain with models and my preference for the docs would be to push app developers to import those items from the model's module to simplify things for app developers.

lampwins:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Nautobot-Core-Engineering-64a875e74842500d9e80e23e/issues/nautobot/nautobot/4459?planning-poker) with Zenhub @glennmatthews @gsnider2195 @HanlinMiao @jathanism @timizuoebideri1

bryanculver:
As discussed in standup:

Look through Core and Extras and export all functions and classes unless there is a significant reason it should not be exported in apps (ex: Models).

itdependsnetworks:
In theory nothing is calling nautobot.apps EXCEPT a plugin, so a the key import (e.g. ConstanceConfigItem, NautobotAppConfig) should ensure the application is ready before importing anything else or simply ensure the app is ready (maybe `nautobot_database_ready`?) before any other import in the nautobot.apps directory.

PR #4533, opened 2023-09-26
Update the apps module namespace

gsnider2195:
# Closes: #4459 
# What's Changed

# TODO

- [ ] Unit, Integration Tests


examples/example_plugin/example_plugin/jobs.py
```diff
@@ -4,13 +4,18 @@
...
+from nautobot.apps.jobs import (
+    DryRunVar,
+    IntegerVar,
+    Job,
+    JobButtonReceiver,
+    JobHookReceiver,
+    register_jobs,
+)
 from nautobot.dcim.models import Device, Location
 from nautobot.extras.choices import ObjectChangeActionChoices
```
glennmatthews, line 16: Should `ObjectChangeActionChoices` be added to apps namespace?

nautobot/apps/__init__.py
```diff
@@ -1,11 +1,14 @@
...
 from nautobot.core.signals import nautobot_database_ready
 from nautobot.core.settings_funcs import ConstanceConfigItem
 from nautobot.extras.plugins import NautobotAppConfig
 
+
 __all__ = (
     "ConstanceConfigItem",
     "nautobot_database_ready",
     "NautobotAppConfig",
+    "NautobotConfig",
```
glennmatthews, line 13: I don't think `NautobotConfig` should be included - apps should always be using its subclass `NautobotAppConfig` that's already here.

nautobot/apps/api.py
```diff
@@ -9,26 +9,100 @@
...
+from nautobot.core.api.fields import (
+    ChoiceField,
+    ContentTypeField,
+    NautobotHyperlinkedRelatedField,
+    ObjectTypeField,
+    SerializedPKRelatedField,
+    TimeZoneSerializerField,
+)
+from nautobot.core.api.mixins import WritableSerializerMixin
+from nautobot.core.api.pagination import OptionalLimitOffsetPagination
```
glennmatthews, line 21: Not sure this is needed - it should be part of the default configuration of Nautobot, not something apps should ever need to relate to directly.

nautobot/apps/api.py
```diff
@@ -9,26 +9,100 @@
...
+from nautobot.core.api.parsers import NautobotCSVParser
+from nautobot.core.api.renderers import FormlessBrowsableAPIRenderer, NautobotCSVRenderer, NautobotJSONRenderer
```
glennmatthews, line 23: Also not sure these are needed by apps - they're infra. I can't immediately picture a case where an app would need to access any of these.

nautobot/apps/api.py
```diff
@@ -9,26 +9,100 @@
...
+    get_api_version_serializer,
+    get_nested_serializer_depth,
+    get_relation_info_for_nested_serializers,
+    get_serializer_for_model,
+    get_view_name,
+    is_api_request,
+    nested_serializer_factory,
+    nested_serializers_for_models,
+    rest_api_server_error,
+    return_nested_serializer_data_based_on_depth,
```
glennmatthews, line 42: Also hesitant about some of the `nested` stuff, do these need to be called directly by apps at any point or are they part of the internal magic that apps don't need to care about?

nautobot/apps/choices.py
```diff
@@ -1,5 +1,21 @@
 """Utilities for implementing choices."""
 
-from nautobot.core.choices import ChoiceSet
+from nautobot.core.choices import (
+    ButtonActionColorChoices,
+    ButtonActionIconChoices,
+    ChoiceSet,
+    ChoiceSetMeta,
```
glennmatthews, line 7: Do apps need `ChoiceSetMeta`?

nautobot/apps/filters.py
```diff
@@ -1,12 +1,54 @@
...
+    CustomFieldDateFilter,
+    CustomFieldFilterMixin,
+    CustomFieldJSONFilter,
+    CustomFieldMultiSelectFilter,
+    CustomFieldMultiValueCharFilter,
+    CustomFieldMultiValueDateFilter,
+    CustomFieldMultiValueNumberFilter,
+    CustomFieldMultiValueSelectFilter,
+    CustomFieldNumberFilter,
+)
```
glennmatthews, line 43: Do apps need these?

nautobot/apps/graphql.py
```diff
@@ -0,0 +1,61 @@
...
+    generate_custom_field_resolver,
+    generate_filter_resolver,
+    generate_list_resolver,
+    generate_list_search_parameters,
+    generate_null_choices_resolver,
+    generate_relationship_resolver,
+    generate_restricted_queryset,
+    generate_schema_type,
+    generate_single_item_resolver,
+)
```
glennmatthews, line 16: I would have thought all of these are internals.

nautobot/apps/graphql.py
```diff
@@ -0,0 +1,61 @@
...
+    extend_schema_type,
+    extend_schema_type_computed_field,
+    extend_schema_type_config_context,
+    extend_schema_type_custom_field,
+    extend_schema_type_filter,
+    extend_schema_type_null_field_choice,
+    extend_schema_type_relationships,
+    extend_schema_type_tags,
+    generate_query_mixin,
+)
```
glennmatthews, line 27: These too.

nautobot/apps/choices.py
gsnider2195: It doesn't appear to be used by any app in the `nautobot` org

nautobot/apps/api.py
gsnider2195: In 2.0 I don't see why the apps would care to mess with nesting serializers. FWIW I don't see any of these being imported in the apps in the `nautobot` org

nautobot/apps/api.py
gsnider2195: Agreed. Not used anywhere by the `nautobot` org apps.

nautobot/apps/filters.py
gsnider2195: Doesn't look like they use any of these

glennmatthews (approved):

PR #4534, opened 2023-09-26
New UI Filtering

bryanculver:
# What's Changed

- Adds filtering capability to the drawer on list views
- Adds a schema for the model's FilterSet
- Adds live searching against the `q` filter in the navbar

(This is a combined feature branch of several stories)

# TODO

- [ ] Unit, Integration Tests


nautobot/ui/src/App.js
```diff
@@ -11,6 +11,9 @@ import { updateAuthStateWithSession, updateNavigation } from "@utils/store";
 import NautobotRouter from "./router";
 
 const theme = {
+    // TODO(norbert-mieczkowski-codilime): This should be fixed in Nautobot UI.
```
glennmatthews, line 14: Fixed in https://github.com/nautobot/nautobot-ui/pull/42 correct? Can we update to a new version of `nautobot-ui` to remove the need for this?

glennmatthews (approved):

gsnider2195:
I'm noticing that the `q` filter isn't getting the remove/add elements in the filter panel. Is this intentional?

image

gsnider2195:
Nitpick: This message on the home page search box is a bit ambiguous. With this message I would assume that the search box doesn't work anywhere, but it's really just the home page where it's not implemented yet.

image

timizuoebideri1 (approved):

nautobot/core/api/metadata.py
```diff
@@ -284,6 +286,10 @@ def determine_metadata(self, request, view):
         """This is the metadata that gets returned on an `OPTIONS` request."""
         metadata = super().determine_metadata(request, view)
 
+        # Include the object type label for this model.
+        object_type = view.queryset.model._meta.label_lower if hasattr(view, "queryset") else "unknown"
```
gsnider2195, line 290: The default for some of our generic views is `queryset = None` so this should be changed to:

```suggestion
        object_type = view.queryset.model._meta.label_lower if getattr(view, "queryset", None) else "unknown"
```

nautobot/core/api/metadata.py
```diff
@@ -304,9 +310,47 @@ def determine_metadata(self, request, view):
             )
 
             metadata["view_options"] = self.determine_view_options(request, serializer)
+            metadata["filters"] = self.get_filter_info(view)
```
gsnider2195, line 313: This probably doesn't need to be in the block `if hasattr(view, "get_serializer")`? The `get_filter_info` method below checks for the `filterset_class` attr on the view and returns early if it's not present.

nautobot/core/api/metadata.py
```diff
@@ -304,9 +310,47 @@ def determine_metadata(self, request, view):
...
+            key=lambda x: get_filter_field_label(x[1]),
+        ):
+            filter_key = filter_name.rsplit("__", 1)[0]
+            label = get_filter_field_label(filter_instance)
+            lookup_label = self._filter_lookup_label(filter_name, filter_instance)
+            filters.setdefault(filter_key, {"label": label})
+            filters[filter_key].setdefault("lookup_types", []).append({"value": filter_name, "label": lookup_label})
+        return filters
+
+    def _filter_lookup_label(self, filter_name, filter_instance):
```
gsnider2195, line 348:
```suggestion
    # TODO: move this into `build_lookup_label` when the legacy UI is removed
    def _filter_lookup_label(self, filter_name, filter_instance):
```

nautobot/ui/src/components/FiltersPanel.js
```diff
@@ -128,6 +160,393 @@ export const FiltersPanelContainer = forwardRef(
...
+                setLastUsedFilters,
+            ]
+        );
+
+        useEffect(() => {
+            setLookupField(null);
+        }, [lookupFields, setLookupField]);
+
+        useEffect(() => {
+            setLookupType(null);
```
gsnider2195, line 349: To automatically populate the `Lookup Type` if there's only one option:

```suggestion
            fieldLookupChoices.length === 1 ? setLookupType(fieldLookupChoices[0].value) : setLookupType(null);
```

nautobot/ui/src/App.js
bryanculver: I'll address this once #4538 is merged

PR #4553, opened 2023-09-28
feature: implement table sticky headers

norbert-mieczkowski-codilime:
# Implements: https://github.com/nautobot/nautobot-ui/issues/39

# What's Changed
Implement table sticky headers from `nautobot-ui` library. Apart from the obvious, the changes are following:
* List view layout components are now explicitly sized and they don't exceed the viewport size
* Skeleton loader slightly changed to better reflect the content being loaded
* Loader widget has inlined spinner and text
* The following React error with rendering invalid component is fixed:
  image
* The tags are no longer getting excessive margins which made their containing table cells too large:
  image

| Loading | Table |
| - | - |
| image | image |

### DIFF
--- a/nautobot/ui/src/components/ObjectTable/ObjectListTable.js
+++ b/nautobot/ui/src/components/ObjectTable/ObjectListTable.js
@@ -3,8 +3,10 @@ import { useLocation } from "react-router-dom";
 import { useEffect, useRef, useState } from "react";
 import {
     Box,
+    calc,
     Divider,
     Flex,
+    getCssVar,
     Heading,
     NtcThumbnailIcon,
     MeatballsIcon,
@@ -172,7 +174,7 @@ export default function ObjectListTable({
     );
 
     return (
-        <Box borderRadius="md" ref={topRef}>
+        <Box borderRadius="md" height="full" ref={topRef}>
             {!include_button ? null : (
                 <Flex align="center">
                     <Heading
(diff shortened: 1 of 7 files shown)


nautobot/ui/src/components/ObjectTable/ObjectListTable.js
```diff
@@ -187,16 +189,13 @@ export default function ObjectListTable({
...
                     <Spacer />
                     {!data_fetched ? (
-                        <Box pr="sm">
+                        <Box marginRight="md">
                             <LoadingWidget name={tableTitle} />
                         </Box>
-                    ) : (
-                        () => {}
-                    )}
+                    ) : null}
```
timizuoebideri1, line 195:
```javascript
{!data_fetched && (
         <Box marginRight="md">
             <LoadingWidget name={tableTitle} />
         </Box>
   )}
```
I guess no need of the `null`

timizuoebideri1 (approved):

nautobot/ui/src/components/ObjectTable/ObjectListTable.js
norbert-mieczkowski-codilime: It certainly is possible and well supported by React, though for safety and consistency the community prefers to use ternary operators anyway. Here's a blog post on that subject by Kent C. Dodds: https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx. I really recommend reading his blog as well as watching his talks, most of the time his thought process is brilliant.

glennmatthews (approved):

glennmatthews:
Interesting. This appears to work for me in Chrome but not in Firefox:

image

image

Issue #4349, opened 2023-08-29
ERROR syslog regarding 'note_list' during Nautobot startup


glennmatthews:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0rc2
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
Run any `nautobot-server` command

### Expected Behavior

Nautobot to run without errors

### Observed Behavior

```
18:55:28.739 ERROR   nautobot.core.apps   __init__.py                                url() :
  Error in link construction for Notes: Reverse for 'note_list' not found. 'note_list' is not a valid view function or pattern name.
```

HanlinMiao:
This occurs https://github.com/nautobot/nautobot/blob/1f5c38eee6a0926f1bc769a2bf667e827e5cec9c/nautobot/extras/navigation.py#L704C1-L713C1. And the TODO comment addresses why the link is there. Related to https://github.com/nautobot/nautobot/issues/4309

skumarmallampeta:
Hi, I am running into a similar issue running nautobot version 1.5.11 in a docker, can you please let me know how to resolve this issue

nautobot_1         |                                     [--no-refresh-dynamic-group-member-caches]
nautobot_1         | nautobot-server post_upgrade: error: unrecognized arguments: --no-invalidate-all
nautobot_1         | ⏳ Waiting on DB... (6s / 30s)
nautobot-worker_1  | Grafana ChatOps integration is not available.
nautobot-worker_1  | NoneType: None
nautobot_1         | Grafana ChatOps integration is not available.
nautobot_1         | NoneType: None
nautobot-worker_1  | 20:44:32.003 ERROR   nautobot.core.apps :
nautobot-worker_1  |   Error in link construction for Notes: Reverse for 'note_list' not found. 'note_list' is not a valid view function or pattern name.
nautobot-worker_1  | usage: nautobot-server post_upgrade [-h] [--version] [-v {0,1,2,3}]
nautobot-worker_1  |                                     [--settings SETTINGS]
nautobot-worker_1  |                                     [--pythonpath PYTHONPATH] [--traceback]
nautobot-worker_1  |                                     [--no-color] [--force-color]
nautobot-worker_1  |                                     [--skip-checks] [--no-build-ui]
nautobot-worker_1  |                                     [--no-clearsessions] [--no-collectstatic]
nautobot-worker_1  |                                     [--no-migrate]
[...]

Issue #4309, opened 2023-08-23
NoteDetailView is not accessible due to missing template.


HanlinMiao:
### Environment
* Nautobot version (Docker tag too if applicable): v1.6.0 and v2.0.0-rc
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Go to demo.nautobot.com
2. Try accessing extras/notes/[slug]/ in v1.6.0 or extras/notes/[uuid]/ in v2.0.0

### Expected Behavior
NoteDetailView renders correctly.

### Observed Behavior
![Screen Shot 2023-08-23 at 1 36 29 PM](https://github.com/nautobot/nautobot/assets/46973263/efd9686d-f9aa-45df-b5a4-0dbcc253d219)

We probably need to add a template to render a single Note instance since the NoteDetailView URL does exist.

bryanculver:
@lampwins @smk4664 Does it make sense for an individual Note to have a detail view or should it be redirected to the list view of the associated object?

gsnider2195:
This is also an issue when creating a new note. The notice presented to the ui links to this non-existent route:

![image](https://github.com/nautobot/nautobot/assets/75227981/a5c84103-aec5-4ea7-b67d-046e952c8965)

![image](https://github.com/nautobot/nautobot/assets/75227981/48663cd1-2771-40ca-9d24-e6c74f9e5fea)

smk4664:
> This is also an issue when creating a new note. The notice presented to the ui links to this non-existent route:
> 
> ![image](https://user-images.githubusercontent.com/75227981/279699185-a5c84103-aec5-4ea7-b67d-046e952c8965.png)
> 
> ![image](https://user-images.githubusercontent.com/75227981/279699335-48663cd1-2771-40ca-9d24-e6c74f9e5fea.png)

#4555 Addresses this as well, but I don't know if this is going to be the accepted solution.

gsnider2195:
Fixed by #4555

PR #4555, opened 2023-09-28
Add Note List View to resolve issue with url missing.

smk4664:
# Closes: #4349
# What's Changed

Add Note List View.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/4349.added
+++ b/changes/4349.added
@@ -1,0 +1,1 @@
+Added Note List View.
--- a/nautobot/extras/forms/forms.py
+++ b/nautobot/extras/forms/forms.py
@@ -128,6 +128,7 @@ __all__ = (
     "LocalContextModelForm",
     "LocalContextModelBulkEditForm",
     "NoteForm",
+    "NoteFilterForm",
     "ObjectChangeFilterForm",
     "PasswordInputWithPlaceholder",
     "RelationshipForm",
@@ -1045,6 +1046,20 @@ class NoteForm(BootstrapMixin, forms.ModelForm):
             "assigned_object_type": forms.HiddenInput,
             "assigned_object_id": forms.HiddenInput,
         }
+
+
+class NoteFilterForm(BootstrapMixin, forms.Form):
+    model = Note
+    q = forms.CharField(required=False, label="Search")
+
+    assigned_object_type_id = DynamicModelMultipleChoiceField(
+        queryset=ContentType.objects.all(),
+        required=False,
+        label="Object Type",
+        widget=APISelectMultiple(
+            api_url="/api/extras/content-types/",
+        ),
+    )
 
 
 #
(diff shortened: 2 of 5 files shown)


glennmatthews:
I wonder if, given that we don't presently support `NotesDetailView` in the new UI, if it would be better to just delete https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/navigation.py#L704-L712 rather than adding these additional but not particularly useful views?

timizuoebideri1 (dismissed):

glennmatthews:
@timizuoebideri1 Thoughts on https://github.com/nautobot/nautobot/pull/4555#issuecomment-1739793672?

timizuoebideri1:
> @timizuoebideri1 Thoughts on [#4555 (comment)](https://github.com/nautobot/nautobot/pull/4555#issuecomment-1739793672)?

Yes, that makes sense. Because we don't support a `NotesDetailView` a breadcrumb for the Notes link is useless because there is no page where this would be displayed.

glennmatthews (commented):
Based on https://github.com/nautobot/nautobot/issues/4309#issuecomment-1789178239, I'm now thinking this full fix is necessary/appropriate.

nautobot/extras/templates/extras/note.html
```diff
@@ -0,0 +1,53 @@
...
+{% load plugins %}
+{% load perms %}
+{% load helpers %}
+
+{% block breadcrumbs %}
+    <li><a href="{% url 'extras:note_list' %}">Notes</a></li>
+    {% if object.assigned_object.get_absolute_url %}
+        <li><a href="{{ object.assigned_object.get_absolute_url }}changelog/">{{ object.assigned_object }}</a></li>
+    {% endif %}
+    <li>{{ object }}</li>
```
glennmatthews, line 12: Is this correct? Feels weird to have the changelog URL in the breadcrumbs for a Note.

smk4664: Copy/Paste issue

smk4664: Fixed!

glennmatthews (approved):

nautobot/extras/templates/extras/note.html
```diff
@@ -0,0 +1,53 @@
...
+                        </td>
+                    </tr>
+                </table>
+                <div class="panel panel-default">
+                    <div class="panel-heading">
+                        <strong>Text</strong>
+                    </div>
+                    <div class="panel-body rendered-markdown">
+                        {{ object.note|render_markdown }}
+                    </div>
```
gsnider2195, line 48: Can you move this panel outside the other panel? This is how it looks now:

image

smk4664: Thanks for catching this! Fixed now.

Issue #4530, opened 2023-09-26
Establish Convention/Best-Practices for Constance & Plugin Settings


bryanculver:
### Proposed Changes

Define things like "should be all uppercase in top-level keys" and "all lowercase in nested dictionary keys" etc. (That is not a suggestion as to what the convention should be, but just an example as to what it could be).

### Justification

- Self-consistency for how setting can be provided and accessed
- Address things like this https://github.com/nautobot/nautobot/pull/4470 and the comments within.

itdependsnetworks:
I would ask to re-consider, we can only do this per breaking change and the current version is:

1. Not usable for any existing plugins without being non-backwards compatible
2. Would not follow our unwritten, but clearly defined pattern for lower case on settings. 

I think it makes sense to either:

1. Remove the `get_app_settings_or_config` function, as it is **not** solving the 80% use case, and will cause confusion. This would allow us to finish this issue before making the change. 
2. Commit #4470 as at least following the standards that have been implied, though not reached. 

It does not make sense to me to de-facto create another undocumented standard by leaving in `get_app_settings_or_config` as is.

smk4664:
Agreed that this should be discussed. When I implemented, I put it in with all UPPERCASE which was what was in the example_plugin and all Nautobot Config Items as well. But looking at the plugin I was planning on using this in, I see all the config is lowercase.

smk4664:
That said, I might be missing something, but I don't believe there is anything forcing the get_app_settings_or_config to use all UPPERCASE.

PR #4556, opened 2023-09-28
Add examples of using lowercase variables in Constance Config

smk4664:
# What's Changed

Document the use of lowercase variables in App provided Constance Config.

This is a stepping stone in addressing #4530, it does not try to define a standard, but shows how we can support existing conventions.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/4556.added
+++ b/changes/4556.added
@@ -1,0 +1,1 @@
+Add documentation for using lowercase variables in App provided Constance Config.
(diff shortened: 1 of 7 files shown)


glennmatthews (approved):
Looks reasonable to me. Did you manually test to confirm that the lowercase_example actually works? (I ask because all of the testing is naturally using `@override_config`, which I worry could have subtle differences from the actual Constance behavior.)

smk4664:
> Looks reasonable to me. Did you manually test to confirm that the lowercase_example actually works? (I ask because all of the testing is naturally using `@override_config`, which I worry could have subtle differences from the actual Constance behavior.)

Yep, I tested manually, and even tested in the plugin that I am implementing this in.

PR #4583, opened 2023-10-02
Fix #4436 - allow non-physical interfaces as parents

### LINKED ISSUE #4436: Can't use LAG as Parent to Virtual Interface
### Environment
* Nautobot version (Docker tag too if applicable): 1.6.0
* Python version: 3.9
* Database platform, version: Postgres 14
* Middleware(s):

### Steps to Reproduce
1. Create LAG interface ae0
2. Create Virtual Interface ae0.0
3. While creating Virtual interface, try to select LAG Interface as Parent

### Expected Behavior
I was hoping that the LAG interface would be in the list of valid parents, but that does not seem to be the case.

### Observed Behavior
 It appears that only physical interfaces can be selected as parents. Is this expected? Or am I modeling the the subinterface of a LAG incorrectly? In this case, the sub interface is an IP interface not a switched interface. So using the Bridge model doesn't seem appropriate.

glennmatthews:
# Closes: #4436 
# What's Changed

- Remove Interface form restriction that blocked non-physical interfaces from appearing as options for parent_interface
- Remove interface model validation check that blocked Virtual interfaces specifically from being parent interfaces
- Add testing

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/models/device_components.py
+++ b/nautobot/dcim/models/device_components.py
@@ -604,16 +604,11 @@ class Interface(CableTermination, PathEndpoint, ComponentModel, BaseInterface):
             # An interface cannot be its own parent
             if self.parent_interface_id == self.pk:
                 raise ValidationError({"parent_interface": "An interface cannot be its own parent."})
+
             # A physical interface cannot have a parent interface
             if hasattr(self, "type") and self.type != InterfaceTypeChoices.TYPE_VIRTUAL:
                 raise ValidationError(
                     {"parent_interface": "Only virtual interfaces may be assigned to a parent interface."}
-                )
-
-            # A virtual interface cannot be a parent interface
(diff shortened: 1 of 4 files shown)


nautobot/dcim/models/device_components.py
```diff
@@ -604,18 +604,13 @@ def clean(self):
             # An interface cannot be its own parent
             if self.parent_interface_id == self.pk:
                 raise ValidationError({"parent_interface": "An interface cannot be its own parent."})
+
```
HanlinMiao, line 607:
```suggestion
```

HanlinMiao (approved):

nautobot/dcim/models/device_components.py
glennmatthews: Was trying to make this function more self-consistent formatting - all the other "if" blocks have a space between them.

PR #4646, opened 2023-10-13
Implement Django admin log entries in admin.

jathanism:
# Closes: #DNE
# What's Changed

- Added `LogEntryAdmin` that asserts a read-only view of `LogEntry` admin logs.
- Revised Change Logging documentation to describe both admin and change logging.
- Updated `ObjectPermission.object_types.limit_choices_to` to allow `auth.LogEntry`
- Fixed a bug in `ObjectPermission.object_types.limit_choices_to` where filtering on `auth.User` should be `users.User`

- [x] Unit, Integration Tests


glennmatthews (commented):
Looks like a solid start. Let's get some tests added.

Do we need to do anything special to get LogEntry records created for our custom admin models like ObjectPermission, or is that all auto-magic?

jathanism:
> Looks like a solid start. Let's get some tests added.
> 
> Do we need to do anything special to get LogEntry records created for our custom admin models like ObjectPermission, or is that all auto-magic?

As long as they are registered in the admin panel (e.g. `admin_site.register(FooModel)`), they will have admin logs.

nautobot/core/admin.py
```diff
@@ -60,3 +57,53 @@ class NautobotModelAdmin(ModelAdmin):
...
+    def has_view_permission(self, request, obj=None):
+        return request.user.is_superuser
+
+    def object_link(self, obj):
+        """Disable delete permission."""
+        if obj.action_flag == DELETION:
+            link = escape(obj.object_repr)
+        else:
+            ct = obj.content_type
+            change_link = reverse(f"admin:{ct.app_label}_{ct.model}_change", args=[obj.object_id])
```
glennmatthews, line 100: Any chance we need to handle a NoReverseMatch here? Maybe if there's an object that has `create/delete` admin views but no `change` view?

nautobot/core/utils/lookup.py
```diff
@@ -78,7 +79,7 @@ def get_route_for_model(model, action, api=False):
     suffix = "" if not api else "-api"
     # The `contenttypes` and `auth` app doesn't provide REST API endpoints,
     # but Nautobot provides one for the ContentType model in our `extras` and Group model in `users` app.
```
glennmatthews, line 81: Mind updating the comment as well?

nautobot/extras/api/serializers.py
```diff
@@ -624,6 +625,26 @@ class Meta:
...
+
+class LogEntrySerializer(NautobotModelSerializer):
+    url = serializers.HyperlinkedIdentityField(view_name="extras-api:logentry-detail")
+    action_flag = ChoiceField(choices=ACTION_FLAG_CHOICES, read_only=True)
+    change_message = serializers.SerializerMethodField()
+    content_type = ContentTypeField(queryset=ContentType.objects.all())
+
+    class Meta:
+        model = LogEntry
+        fields = "__all__"
```
glennmatthews, line 641: What fields are included here? Any concern about sensitive information exposure?

nautobot/extras/filters/__init__.py
```diff
@@ -727,6 +729,24 @@ class LocalContextFilterSet(LocalContextModelFilterSetMixin):
...
 
+#
+# Log Entries
+#
+
+
+class LogEntryFilterSet(BaseFilterSet):
+    q = SearchFilter(
+        filter_predicates={
+            "object_repr": "icontains",
```
glennmatthews, line 740: can/should this also include the `change_message` to match the modeladmin behavior?

nautobot/extras/tests/test_api.py
```diff
@@ -2236,6 +2238,84 @@ def test_dry_run_not_supported(self):
         self.assertHttpStatus(response, status.HTTP_405_METHOD_NOT_ALLOWED)
 
 
+class LogEntryTest(
+    APIViewTestCases.GetObjectViewTestCase,
+    APIViewTestCases.ListObjectsViewTestCase,
+    APIViewTestCases.NotesURLViewTestCase,
```
glennmatthews, line 2244: Seems weird for this model to include a notes URL?

nautobot/core/admin.py
jathanism: This is specifically for `admin` namespace, so all of those views/routes are automatically generated when a model is registered in the admin site. So it's not a concern IMO.

nautobot/extras/tests/test_api.py
jathanism: Yeah good point.

nautobot/extras/filters/__init__.py
jathanism: You mean to search on the `change_message`? Problem is it's stored as an integer and translated to text, so not sure how we'd make it useful without some extras steps.

glennmatthews: Yeah, since it looks like the modeladmin is set up to search on it above? If not feasible, that's fine, just asking.

nautobot/core/admin.py
glennmatthews: What about install/uninstall apps case? (App installed with a ModelAdmin, log entries are created for it, then app is uninstalled)

nautobot/extras/filters/__init__.py
jathanism: Oh right, and I'm realizing that we never implemented the search box for the admin integration so that isn't even doing anything. I'll remove it for clarity from the `ModelAdmin`.

jathanism: Actually nevermind, I was mixing up `change_message` and `action_flag`. I'll add `change_message` to search.

nautobot/extras/api/serializers.py
jathanism:
```json
        {
            "id": 4,
            "object_type": "admin.logentry",
            "display": "Deleted “user2.”",
            "url": "http://localhost:6789/api/extras/admin-logs/4/",
            "natural_slug": "unknown",
            "action_flag": {
                "value": 3,
                "label": "Deletion"
            },
            "change_message": "",
            "content_type": "users.user",
            "action_time": "2023-10-18T11:36:28.864585-07:00",
            "object_id": "0334ba48-c37a-4a42-b47c-66ba5bafc87b",
            "object_repr": "user2",
            "user": {
                "id": "78cc0fef-844c-4359-a4b1-163916f4f85b",
                "object_type": "users.user",
                "url": "http://localhost:6789/api/users/users/78cc0fef-844c-4359-a4b1-163916f4f85b/"
            }
        },
```

No concerns on my end.

nautobot/core/admin.py
jathanism: I've never had an issue in the admin panel for that. I'm guessing since the object would no longer be registered in the admin, as it's not installed, it wouldn't even show.

jvanderaa:
It would be great to also incorporate this into 1.6.x LTM.

nautobot/docs/user-guide/platform-functionality/change-logging.md
```diff
@@ -1,5 +1,19 @@
...
+
+## Administrative Changes
+
+Administrative changes are those made under the "Admin" section of the user interface. This is the primary view for Users, Groups, Object Permissions, and other objects core to the administration of Nautobot. Any changes made to objects using this interface will be displayed as "Log entries" under the "Administration" section of the Admin list view. This is a read-only view that disallows manual creation, updating, or deletion of these objects.
+
+These records are commonly referred to as "admin logs" for short and are provided by default by the Django web framework.  
+
+You may access these records by navigating to `/admin/` or by clicking your username in the top right, then "Admin".
+
+Admin logs are exposed in the API via the read-only endpoint `/api/extras/admin-logs/`.
```
glennmatthews, line 13:
```suggestion
You may access these records (if logged in as an admin/staff user) by navigating to `/admin/` or by clicking your username in the navigation bar, then "Admin".

Admin logs are also retrievable (by admin/staff users only) through the REST API via the read-only endpoint `/api/extras/admin-logs/`.
```

nautobot/extras/tests/test_api.py
```diff
@@ -2236,6 +2238,83 @@ def test_dry_run_not_supported(self):
...
+        # Fields that should be present in *ALL* model serializers:
+        self.assertIn("id", response.data)
+        self.assertEqual(str(response.data["id"]), str(instance1.pk))  # coerce to str to handle both int and uuid
+        self.assertIn("url", response.data)
+        self.assertIn("display", response.data)
+        self.assertIn("natural_slug", response.data)
+        self.assertIsInstance(response.data["display"], str)
+
+    def test_list_objects_csv(self):
+        """Disabled."""
```
glennmatthews, line 2293: Not supported?

nautobot/extras/tests/test_filters.py
```diff
@@ -954,6 +956,52 @@ def test_search(self):
...
+
+    def test_search(self):
+        params = {"q": "admin"}
+        self.assertQuerysetEqualAndNotEmpty(
+            self.filterset(params, self.queryset).qs, self.queryset.filter(user__username__icontains="admin")
+        )
+        params = {"q": "nautobotuser"}
+        self.assertQuerysetEqualAndNotEmpty(
+            self.filterset(params, self.queryset).qs, self.queryset.filter(object_repr__icontains="nautobotuser")
+        )
```
glennmatthews, line 998: These two are going to match **all** LogEntry instances created in `setUp` above, won't they? Not ideal for a filtering test case since it's identical to fail-open behavior for the filterset.

nautobot/extras/tests/test_api.py
jathanism: Didn't think it was a good idea but open to suggestions.

glennmatthews: I didn't see anything in the code that would disable CSV export via REST - just the test case here. I would imagine that with the current implementation any admin user would be able to do CSV export just by passing `?format=csv` to the REST call.

jathanism: It's because of `is_staff=True` or not permissions, it fails:

```
FAIL: test_list_objects_csv (nautobot.extras.tests.test_api.LogEntryTest)
GET a list of objects in CSV format as an authenticated user with permission to view some objects.
  File "/Users/jathan/Library/Caches/pypoetry/virtualenvs/nautobot-b2ttWva6-py3.10/lib/python3.10/site-packages/django/test/utils.py", line 387, in inner
    return func(*args, **kwargs)
  File "/Users/jathan/sandbox/src/nautobot/nautobot/core/testing/api.py", line 566, in test_list_objects_csv
    self.assertEqual(1, len(rows))
AssertionError: 1 != 2

```

Need to think this one through.

timizuoebideri1 (approved):

nautobot/extras/tests/test_api.py
```diff
@@ -2236,6 +2238,91 @@ def test_dry_run_not_supported(self):
...
+
+    def test_list_objects_csv(self):
+        """
+        Disabled for now.
+
+        The reasoning is that the base method combines test of access to all objects and filtering
+        them out in the same test case. This complicates trying to pass this through as-is due to
+        LogEntry gating on whether a user has `is_staff=True`. Splitting those into two tests might
+        help, but we also need to think about whether limiting admin-logs to `is_staff=True` only is
+        sane, or if we should rely strictly on model permissions to access `admin.logentry`.
```
glennmatthews, line 2300: When are we going to do this thinking about? :-)

jathanism: Right meow.

jvanderaa:
@jathanism will this be included on sending the logs to the Python Django logging facility not just the UI and API?

glennmatthews:
> @jathanism will this be included on sending the logs to the Python Django logging facility not just the UI and API?

That hadn't previously been considered in scope here. We'd have to add a signal handler on the `LogEntry` model or something, I think, if that's needed. Trying to keep the scope small here as this is uncommitted work at this time.

jathanism:
> @jathanism will this be included on sending the logs to the Python Django logging facility not just the UI and API?

No.

nautobot/core/admin.py
```diff
@@ -60,3 +57,49 @@ class NautobotModelAdmin(ModelAdmin):
...
+    def has_change_permission(self, request, obj=None):
+        """Disable change permission."""
+        return False
+
+    def has_delete_permission(self, request, obj=None):
+        """Disable delete permission."""
+        return False
+
+    def has_view_permission(self, request, obj=None):
+        return request.user.is_staff
```
glennmatthews, line 88: Something seems wrong/missing here - I created a non-superuser `is_staff` user and they don't have a `Log Entries` section on their admin home page:

image

Although they do appear to have access to http://localhost:8080/admin/admin/logentry/.

This is different behavior than I see for non-superuser staff users for other admin pages - for `groups`, `tokens`, and `permissions` the home page items only appear, and the URL is only accessible, if the user explicitly has object permissions for those content-types.

nautobot/core/admin.py
```diff
@@ -60,3 +57,49 @@ class NautobotModelAdmin(ModelAdmin):
         models.ForeignKey: {"widget": widgets.StaticSelect2},
         models.ManyToManyField: {"widget": widgets.StaticSelect2Multiple},
     }
+
+
+class LogEntryAdmin(NautobotModelAdmin):
```
glennmatthews, line 62: Wonder if this should go in the `users` app instead of `core` just for symmetry with `UserAdmin`, `GroupAdmin`, `ObjectPermissionAdmin`, and `TokenAdmin`?

nautobot/core/admin.py
jathanism: Yeah this should just not be overloaded and allowed to use the default permissions controls in the `ModelAdmin`. My attempt to gate by `is_staff` appears to have been flawed.

timizuoebideri1 (approved):

nautobot/users/models.py
```diff
@@ -245,6 +245,7 @@ class ObjectPermission(BaseModel):
                     "users",
                 ]
             )
+            | Q(app_label="admin", model__in=["logentry"])
             | Q(app_label="auth", model__in=["group", "user"])
             | Q(app_label="users", model__in=["objectpermission", "token"])
```
glennmatthews, line 250: I noticed the other day that this is actually wrong - we don't use `auth.user`, we use `users.user`; the most obvious side effect I noticed is that you can't actually define objectpermissions around `users.user` at present. Might make sense to fix that as part of this PR?

nautobot/users/admin.py
```diff
@@ -70,6 +74,47 @@ def user_count(obj):
...
+    def has_change_permission(self, request, obj=None):
+        """Disable change permission."""
+        return False
+
+    def has_delete_permission(self, request, obj=None):
+        """Disable delete permission."""
+        return False
+
+    def object_link(self, obj):
+        """Disable delete permission."""
```
glennmatthews, line 104:
```suggestion
        """Link to the related object, unless it was deleted."""
```

glennmatthews:
Shouldn't this go to `next` rather than `develop`?

jathanism:
> Shouldn't this go to `next` rather than `develop`?

Done.

glennmatthews (commented):
Sorry - mind removing a73b9a9347fd08bd52c8b64e2129ffac66ed5426, 50243ff13d0aa0071bfa9aeb96eecc1d7c75c64b, 969dcf4c361d52d4bf31920e9c39ff0cf3acf941, 70d82de295e948c61a998d15fe1716dad7d5d9cf, and f4ecd8b166daf4bf5b95131f57a99a105728596b from your branch so that we don't have merge conflicts down the line when we squash-merge this into next?

glennmatthews (approved):

jathanism:
> Sorry - mind removing [a73b9a9](https://github.com/nautobot/nautobot/commit/a73b9a9347fd08bd52c8b64e2129ffac66ed5426), [50243ff](https://github.com/nautobot/nautobot/commit/50243ff13d0aa0071bfa9aeb96eecc1d7c75c64b), [969dcf4](https://github.com/nautobot/nautobot/commit/969dcf4c361d52d4bf31920e9c39ff0cf3acf941), [70d82de](https://github.com/nautobot/nautobot/commit/70d82de295e948c61a998d15fe1716dad7d5d9cf), and [f4ecd8b](https://github.com/nautobot/nautobot/commit/f4ecd8b166daf4bf5b95131f57a99a105728596b) from your branch so that we don't have merge conflicts down the line when we squash-merge this into next?

Fixed.

Issue #4673, opened 2023-10-17
git repo sync with failure can expose token value in logging


jeffkala:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.2
* Python version: py3.8
* Database platform, version: Postgres
* Middleware(s): n/a

### Steps to Reproduce
1. Create gitrepo and assign it to a secrets group with with type `https/token`.
2. Try to sync to a repository like azure(ADO) that requires a user and password for basic auth.
3. The new `Return Value` section in the job result logs the actual failure with the access token

### Expected Behavior
redact the token from the return value 

### Observed Behavior
Return value shows the actual token in the failure.

Some of the token was obfuscated but left some for clarity.

![nautobot_issue](https://github.com/nautobot/nautobot/assets/48843785/003cc5b7-2b70-4c37-8f21-031dbc7c0e0e)

PR #4683, opened 2023-10-19
Fixed token exposure in `JobResult` traceback and result output when `GitRepositorySync` job fails.

jathanism:
# Closes: #4673 
# What's Changed

- Updated `NautobotDatabaseBackend._get_extended_properties` to always sanitize the traceback output if it is set.
- Overloaded `NautobotDatabaseBackend.prepare_exception()` to always sanitize the CLI error output message when a exception's message payload is a list or tuple. The output from this method is what gets stored as the `result` of the job in such error cases.
 

## Example fetch
Cursor_and_Notification_Center

## Example clone
Cursor_and_Notification_Center

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/4673.fixed
+++ b/changes/4673.fixed
@@ -1,0 +1,1 @@
+Fixed token exposure in `JobResult` traceback and result output when a `GitRepositorySync` job fails.
--- a/changes/4673.fixed
+++ b/changes/4673.fixed
@@ -1,1 +1,0 @@
-Fixed token exposure in `JobResult` traceback and result output when a `GitRepositorySync` job fails.
--- a/changes/4673.security
+++ b/changes/4673.security
@@ -1,1 +1,1 @@
-Fixed token exposure in `JobResult` traceback and result output when a `GitRepositorySync` job fails in scertain ways.
+Fixed token exposure in `JobResult` traceback and result output when a `GitRepositorySync` job fails in certain ways.
--- a/changes/4673.security
+++ b/changes/4673.security
@@ -1,0 +1,1 @@
+Fixed token exposure in `JobResult` traceback and result output when a `GitRepositorySync` job fails in scertain ways.
--- a/nautobot/core/celery/backends.py
+++ b/nautobot/core/celery/backends.py
@@ -74,11 +74,27 @@ class NautobotDatabaseBackend(DatabaseBackend):
         """Overload default to explicitly sanitize the traceback message result."""
         exc_info = super().prepare_exception(exc, serializer=serializer)
 
-        # If this is a GitCommandError, attempt to sanitize the message error.
-        # e.g. b"fatal: could not read Password for 'https://abc123@github.com': terminal prompts disabled"
-        if exc_info["exc_type"] == "GitCommandError":
-            exc_msg = list(exc_info["exc_message"])  # It starts as a tuple.
-            exc_msg[-1] = sanitize(exc_msg[-1].decode("utf-8"))  # The message will be bytes.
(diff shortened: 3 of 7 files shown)


changes/4673.fixed
```diff
@@ -0,0 +1 @@
+Fixed token exposure in `JobResult` traceback and result output when a `GitRepositorySync` job fails.
```
glennmatthews, line 1: Perhaps?

```suggestion
Fixed token exposure in `JobResult` traceback and result output when a `GitRepositorySync` job fails in certain ways.
```

glennmatthews: Should this be a `.security` item?

nautobot/core/celery/backends.py
```diff
@@ -64,3 +69,16 @@ def _get_extended_properties(self, request, traceback):
...
+    def prepare_exception(self, exc, serializer=None):
+        """Overload default to explicitly sanitize the traceback message result."""
+        exc_info = super().prepare_exception(exc, serializer=serializer)
+
+        # If this is a GitCommandError, attempt to sanitize the message error.
+        # e.g. b"fatal: could not read Password for 'https://abc123@github.com': terminal prompts disabled"
+        if exc_info["exc_type"] == "GitCommandError":
+            exc_msg = list(exc_info["exc_message"])  # It starts as a tuple.
+            exc_msg[-1] = sanitize(exc_msg[-1].decode("utf-8"))  # The message will be bytes.
+            exc_info["exc_message"] = tuple(exc_msg)  # Turn it back into a tuple.
```
glennmatthews, line 82: This is a lot more narrowly targeted than I'd personally prefer. Can we not just sanitize every part of `exc_message` unconditionally?

changes/4673.fixed
jathanism: Probably.

nautobot/core/celery/backends.py
jathanism: For the fetch/clone error cases, `exc_message` coming from `GitCommandError` is a list following this pattern:

```
# ([shlex.split(), retval, error_msg])
(["git", "fetch", "other", "args"], 128, b"fatal: could not read Password for 'https://abc123@github.com': terminal prompts disabled"])
```

I was trying `json.loads(sanitize(json.dumps(exc_info["exc_message"])))` to do it in one shot but that stupid error message is `bytes` for some reason.

The other thing is that I have no assertion that other errors would be in this same format, so I kept it super specific to the reported issue.

jathanism: Also I figure now that there is a pattern for sanitizing exceptions coming from Celery, that we might end up with mappings on how to do each one as-needed, unless we can come-up with a byte-steamrolling-JSON-encoder.

changes/4673.security
```diff
@@ -0,0 +1 @@
+Fixed token exposure in `JobResult` traceback and result output when a `GitRepositorySync` job fails in scertain ways.
```
glennmatthews, line 1:
```suggestion
Fixed token exposure in `JobResult` traceback and result output when a `GitRepositorySync` job fails in certain ways.
```

nautobot/core/celery/backends.py
```diff
@@ -64,3 +69,32 @@ def _get_extended_properties(self, request, traceback):
             )
 
         return extended_props
+
+    def prepare_exception(self, exc, serializer=None):
```
glennmatthews, line 73: Anything we can do as a test case for this functionality? I'm thinking one of our test Jobs that intentionally throws an exception?

glennmatthews (approved):

Issue #4704, opened 2023-10-24
Role screen related object tables pagination selection overlap with the table below it. 


scetron:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0
* Python version: 3.11
* Database platform, version: postgres/na
* Middleware(s):

### Steps to Reproduce
1. Create a role
2. View the role detail screen
3.

### Expected Behavior

UI elements to not overlap

### Observed Behavior

Pagination selector for each table overlaps with the table below it. See screenshot 
![Screenshot 2023-10-24 at 16 48 43](https://github.com/nautobot/nautobot/assets/10200477/ffae8f30-4c06-4755-acd2-349c6bd67b04)

timizuoebideri1:
The issue here is caused by the `pull-right` CSS class, which makes a `div` float and affects the `div` positioned below it. Here are two solutions:

1. Remove `pull-right` from the top `div`. The pagination box will still be on the right, and the parent `div` will occupy 100% width.
https://github.com/nautobot/nautobot/blob/f56e49ff54b1bd088945c2765acbfe023e139f44/nautobot/core/templates/inc/paginator.html#L3-L6


2. Apply `clear: both` to all `div`'s below the pagination to prevent overlapping.

https://github.com/nautobot/nautobot/blob/5da5c04c8e2c77b81a88cf42d5e753eb5a49c72a/nautobot/extras/templates/extras/role_retrieve.html#L62-L63

i.e 
```html
    {% include 'inc/paginator.html' with paginator=vlan_table.paginator page=device_table.page %}
    <div class="panel panel-default clear" style="clear: both">
```

scetron:
Are these the only two acceptable solutions? Both feel a bit off. Removing the pull-right will also affect everywhere else the paginator template is used. 

I also noticed that each table, the `paginator=vlan_table.paginator `instead of a `paginator` for the specific table, is `device_table.paginator` an invalid `paginator`?

scetron:
Ahh well, looks like removing the pull-right doesn't effect other tables on other pages...

Issue #4026, opened 2023-06-29
2.0: Role view shows irrelevant content


mzbroch:
### Environment
* Nautobot version (Docker tag too if applicable): `next` (20230629)
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Create Role
2. Assign Role with Content Types
3.

### Expected Behavior
Role view displays only relevant information

### Observed Behavior
Role view shows irrelevant content (wrt to role's Content-Type) like tables including Device, IP Address information:

<img width="1433" alt="Screenshot 2023-06-29 at 14 49 24" src="https://github.com/nautobot/nautobot/assets/16935122/4f58abf1-d3c3-428d-ada6-bd3240b3d9c5">

PR #4707, opened 2023-10-25
Fix Role View

scetron:
# Closes: #4704 #4026
# What's Changed

Fixed pagination controls floating over UI elements below. 
Before: 
Screenshot 2023-10-25 at 10 42 12

After: 
Screenshot 2023-10-25 at 10 42 59

Fixed Role view showing content types that aren't enabled for the Role.
Screenshot 2023-10-25 at 10 44 14

# TODO


nautobot/extras/templates/extras/role_retrieve.html
```diff
@@ -46,20 +46,23 @@
 {% endblock content_left_page %}
 
 {% block content_full_width_page %}
+    {% if "dcim | device" in content_types %}
```
timizuoebideri1, line 49: Could you add a test case asserting only expected tables are shown

scetron: I've updated the view to provide those flags.  For the test, do you have any example code to assert some html is not in the body of a page?

scetron: Actually, I think we can assert this via extra_context using the view

timizuoebideri1: `self.assertNotIn` should do the trick

timizuoebideri1: > Actually, I think we can assert this via extra_context using the view

Im not sure this would test if the table is shown in HTML

HanlinMiao (approved):

changes/4704.fixed
```diff
@@ -0,0 +1,3 @@
+Fixed a display issue with table pagination controls floating over other UI elements.
+Fixed Role detail view displaying tables for content types that aren't in the Role's allowed content types.
+Fixed a broken link in CONTRIBUTING.md to point to the correct contributing guide.
```
gsnider2195, line 3: This line should be in housekeeping

nautobot/extras/templates/extras/role_retrieve.html
scetron: Can you please point out an example that tests for html in a rendered view?

nautobot/extras/templates/extras/role_retrieve.html
```diff
@@ -46,75 +46,101 @@
...
-             </div>
-         {% endif %}
-     </div>
+    {% if device_table %}
+        <div class="panel panel-default">
+            <div class="panel-heading">
+                <strong>Devices</strong>
+            </div>
+            {% include 'inc/table.html' with table=device_table %}
+            {% if perms.ipam.add_device %}
```
gsnider2195, line 55:
```suggestion
            {% if perms.dcim.add_device %}
```

nautobot/extras/templates/extras/role_retrieve.html
```diff
@@ -46,75 +46,101 @@
...
+        </div>
+    {% include 'inc/paginator.html' with paginator=device_table.paginator page=device_table.page %}
+    {% endif %}
+    {% if rack_table %}
+        <div class="panel panel-default">
+            <div class="panel-heading">
+                <strong>Racks</strong>
+            </div>
+            {% include 'inc/table.html' with table=rack_table %}
+            {% if perms.ipam.add_rack %}
```
gsnider2195, line 71:
```suggestion
            {% if perms.dcim.add_rack %}
```

gsnider2195 (approved):

Issue #3994, opened 2023-06-26
Show user that created and last updated object in advanced tab


lampwins:
### As ...

Dan - Data Owner

### I want ...

I easily see at a glance the user that created and last updated an object in nautobot

### So that ...

I can quickly audit this important detail without having to go to the changelog

### I know this is done when...

- The user name that created and last updated the object are shown in the advanced tab
- If the user name for the object in either case is not know, simple show an em dash
- The advanced tab should show this data in a new "Data Provenance" section

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

bryanculver:
The quickest way to address this would be to look up the first and most recent `ObjectChange` record but those may have been cleaned up, especially the first.

Should we add a `created_by`, `changed_by` (and do similar username caching we do for `ObjectChange` in the event of a deleted user) to `ChangeLoggedModel?

What is the "user" if we don't have historical information?

What is the "user" in the direct ORM change case?

lampwins:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Nautobot-Core-Product-64a8783b7a3f4b1225973a0c/issues/nautobot/nautobot/3994?planning-poker) with Zenhub @bryanculver @gsnider2195 @HanlinMiao @jathanism

bryanculver:
> The quickest way to address this would be to look up the first and most recent `ObjectChange` record but those may have been cleaned up, especially the first.
> 
> Should we add a `created_by`, `changed_by` (and do similar username caching we do for `ObjectChange` in the event of a deleted user) to `ChangeLoggedModel?
This should "Just Work :tm:" with `ChangeLoggedModel` (app developers shoudln't be required to add migrations)

Either attempt performant indexing of the ObjectChange records or use Redis caching

> What is the "user" if we don't have historical information?

Do not show or em-dash.
 
> What is the "user" in the direct ORM change case?

Outside of scope.

gsnider2195:
Also needs to be presented in the API for new UI to display

bryanculver:
> Also needs to be presented in the API for new UI to display

[...]

PR #4710, opened 2023-10-25
Show user that created and last updated object in advanced tab

HanlinMiao:
# Closes: #3994 
# What's Changed

Screen Shot 2023-10-25 at 5 38 41 PM

# TODO

- [x] Unit, Integration Tests


nautobot/extras/models/change_logging.py
```diff
@@ -28,10 +28,36 @@ class ChangeLoggedModel(models.Model):
...
+    def save(self, *args, **kwargs):
+        from nautobot.extras.choices import ObjectChangeActionChoices
+
+        # call save() first so that the ObjectChange record exists
+        super().save(*args, **kwargs)
+
+        most_recent_change = get_changes_for_model(self).first()
+        if most_recent_change and most_recent_change.action == ObjectChangeActionChoices.ACTION_CREATE:
+            self.created_by = most_recent_change.user
+        self.last_updated_by = most_recent_change.user
```
glennmatthews, line 58: Would it make any sense to instead do this as part of the change_logging decorator, instead of the object save()? Just thinking about bulk operations and avoiding the need to repeatedly look up the ObjectChange and associated User for each and every object created/modified.

nautobot/core/testing/views.py
```diff
@@ -459,6 +461,7 @@ def test_edit_object_with_permission(self):
                 objectchanges = lookup.get_changes_for_model(instance)
                 self.assertEqual(len(objectchanges), 1)
                 self.assertEqual(objectchanges[0].action, extras_choices.ObjectChangeActionChoices.ACTION_UPDATE)
+                self.assertEqual(instance.last_updated_by, self.user)
```
glennmatthews, line 464: Might need/want to use a different user account specifically for this test - since all of our generic tests pretty much use the same single user account, this isn't a very meaningful test as currently implemented.

nautobot/extras/models/change_logging.py
```diff
@@ -28,10 +28,36 @@ class ChangeLoggedModel(models.Model):
...
+    )
 
     class Meta:
         abstract = True
 
+    def save(self, *args, **kwargs):
+        from nautobot.extras.choices import ObjectChangeActionChoices
+
+        # call save() first so that the ObjectChange record exists
+        super().save(*args, **kwargs)
```
glennmatthews, line 53: save() doesn't create an ObjectChange, the post_save signal handler does.

nautobot/core/templates/inc/object_details_advanced_panel.html
```diff
@@ -57,6 +59,23 @@
...
         </table>
     </div>
+    <div class="panel panel-default">
+        <div class="panel-heading">
+            <strong>Data Provenance</strong>
+        </div>
+        <table class="table table-hover panel-body attr-table">
+            <tbody>
+                <tr>
+                    <td>Created By</td>
```
glennmatthews, line 69: Looks weird to me to have "Created" in one panel and "Created By" in a different panel. Wonder if we can/should move the "Created" and "Last Updated" entries into the "Data Provenance" panel as well - any thoughts @lampwins?

glennmatthews (commented):
This should be targeted at `next`, not `develop`. Looks like a promising start!

PR #4711, opened 2023-10-26
Show user that created and last updated object in advanced tab

HanlinMiao:
# Closes: #3994 
# What's Changed

Screen Shot 2023-10-25 at 5 38 41 PM

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/models/change_logging.py
```diff
@@ -28,10 +28,35 @@ class ChangeLoggedModel(models.Model):
...
     class Meta:
         abstract = True
 
+    def save(self, *args, **kwargs):
+        from nautobot.extras.choices import ObjectChangeActionChoices
+
+        # call save() first so that the ObjectChange record exists
+        super().save(*args, **kwargs)
+        most_recent_change = get_changes_for_model(self).first()
+        if most_recent_change and most_recent_change.action == ObjectChangeActionChoices.ACTION_CREATE:
```
HanlinMiao, line 55: From Glenn: Would it make any sense to instead do this as part of the change_logging decorator, instead of the object save()? Just thinking about bulk operations and avoiding the need to repeatedly look up the ObjectChange and associated User for each and every object created/modified.

nautobot/core/templates/inc/object_details_advanced_panel.html
```diff
@@ -42,6 +44,15 @@
                         </td>
                     </tr>
                 {% endif %}
+            </tbody>
+        </table>
+    </div>
+    <div class="panel panel-default">
```
HanlinMiao, line 50: From Glenn: Looks weird to me to have "Created" in one panel and "Created By" in a different panel. Wonder if we can/should move the "Created" and "Last Updated" entries into the "Data Provenance" panel as well - any thoughts @lampwins?

HanlinMiao (commented):
I have to close the last PR because I messed up the git operations so opening a new one.

From Glenn: Might need/want to use a different user account specifically for this test - since all of our generic tests pretty much use the same single user account, this isn't a very meaningful test as currently implemented.

nautobot/extras/models/change_logging.py
timizuoebideri1: Why do we need to the `ObjectChange` record to exist first before attempting to change the `created_by` or `last_updated_by`?

examples/example_plugin/example_plugin/migrations/0007_add_new_user_fields_to_changelogged_models.py
glennmatthews, line 1: Can we put some thought into how this will impact apps? I'm guessing they'll need migrations to all of their models as well in order to support this new feature. @whitej6

nautobot/extras/models/change_logging.py
glennmatthews: Because as currently implemented we're using the ObjectChange to find the user that needs to be set into those fields as we don't natively have that information in `save()`.

nautobot/circuits/migrations/0020_add_new_user_fields_to_changelogged_models.py
glennmatthews, line 1: Are we going to need/want data migrations to populate initial values for `created_by` and `last_updated_by` for all existing objects? Feels like we should.

HanlinMiao: The only way I think we can do that is through querying ObjectChange records, is that what you have in mind?

nautobot/core/views/generic.py
```diff
@@ -101,6 +103,19 @@ def get(self, request, *args, **kwargs):
...
+        object_change_records = get_changes_for_model(instance)
+        created_by = None
+        last_updated_by = None
+        try:
+            created_by_record = object_change_records.get(action=ObjectChangeActionChoices.ACTION_CREATE)
+            created_by = created_by_record.user
+        except ObjectChange.DoesNotExist:
+            pass
+
+        last_updated_by_record = object_change_records.filter(action=ObjectChangeActionChoices.ACTION_UPDATE).first()
```
glennmatthews, line 116: Wonder if the `filter` is actually needed here. If the newest record is a `Delete` we'll never see this page, and if it was a `Create` only, then that's still the "last updated" information for this model?

nautobot/core/views/generic.py
```diff
@@ -101,6 +103,19 @@ def get(self, request, *args, **kwargs):
         Generic GET handler for accessing an object.
         """
         instance = get_object_or_404(self.queryset, **kwargs)
+        # Get the ObjectChange records to populate the advanced tab information
+        object_change_records = get_changes_for_model(instance)
+        created_by = None
+        last_updated_by = None
+        try:
+            created_by_record = object_change_records.get(action=ObjectChangeActionChoices.ACTION_CREATE)
+            created_by = created_by_record.user
```
glennmatthews, line 112: It might be fine to use `.user_name` instead of `.user` (the former is a string, and will persist even if the user account itself is later deleted)?

nautobot/core/testing/views.py
```diff
@@ -333,7 +334,20 @@ def test_create_object_with_permission(self):
                 # Verify ObjectChange creation
                 objectchanges = lookup.get_changes_for_model(instance)
                 self.assertEqual(len(objectchanges), 1)
+                # Assert that Created By table row is updated with the user that created the object
                 self.assertEqual(objectchanges[0].action, extras_choices.ObjectChangeActionChoices.ACTION_CREATE)
+                # Validate if detail view exists
+                validate = URLValidator()
+                try:
+                    validate(instance.get_absolute_url())
```
timizuoebideri1, line 342: Maybe store `instance.get_absolute_url()` in a variable and use it?

nautobot/core/testing/views.py
```diff
@@ -459,6 +473,20 @@ def test_edit_object_with_permission(self):
                 objectchanges = lookup.get_changes_for_model(instance)
                 self.assertEqual(len(objectchanges), 1)
                 self.assertEqual(objectchanges[0].action, extras_choices.ObjectChangeActionChoices.ACTION_UPDATE)
+                # Validate if detail view exists
+                validate = URLValidator()
+                try:
+                    validate(instance.get_absolute_url())
```
timizuoebideri1, line 479: Maybe store instance.get_absolute_url() in a variable and use it?

nautobot/core/views/renderers.py
```diff
@@ -229,6 +231,21 @@ def get_context(self, data, accepted_media_type, renderer_context):
             "verbose_name_plural": queryset.model._meta.verbose_name_plural,
         }
         if view.action == "retrieve":
+            # Get the ObjectChange records to populate the advanced tab information
+            object_change_records = get_changes_for_model(instance)
```
timizuoebideri1, line 235: Since this is a duplicate, would it make sense to create a util function 
```python
created_by, last_updated_by = get_created_and_last_updated_usernames_for_model(instance)
```

timizuoebideri1 (approved):

nautobot/core/testing/views.py
```diff
@@ -333,7 +334,21 @@ def test_create_object_with_permission(self):
...
+                # Assert that Created By table row is updated with the user that created the object
                 self.assertEqual(objectchanges[0].action, extras_choices.ObjectChangeActionChoices.ACTION_CREATE)
+                # Validate if detail view exists
+                validate = URLValidator()
+                try:
+                    detail_url = instance.get_absolute_url()
+                    validate(detail_url)
+                except ValidationError:
+                    self.skipTest("Model does not have a detail view")
+                response = self.client.get(instance.get_absolute_url())
```
timizuoebideri1, line 346:
```suggestion
                response = self.client.get(detail_url)
```

nautobot/core/testing/views.py
```diff
@@ -333,7 +334,21 @@ def test_create_object_with_permission(self):
...
+                response_body = testing.extract_page_body(response.content.decode(response.charset))
+                advanced_tab_href = f"{instance.get_absolute_url()}#advanced"
```
timizuoebideri1, line 348:
```suggestion
                advanced_tab_href = f"{detail_url}#advanced"
```

nautobot/core/testing/views.py
```diff
@@ -333,7 +334,21 @@ def test_create_object_with_permission(self):
...
                 self.assertEqual(len(objectchanges), 1)
+                # Assert that Created By table row is updated with the user that created the object
                 self.assertEqual(objectchanges[0].action, extras_choices.ObjectChangeActionChoices.ACTION_CREATE)
+                # Validate if detail view exists
+                validate = URLValidator()
+                try:
+                    detail_url = instance.get_absolute_url()
+                    validate(detail_url)
+                except ValidationError:
+                    self.skipTest("Model does not have a detail view")
```
timizuoebideri1, line 345:
```suggestion
                detail_url = instance.get_absolute_url()
                try:
                    validate(detail_url)
                except ValidationError:
                    self.skipTest("Model does not have a detail view")
```

nautobot/core/testing/views.py
```diff
@@ -459,6 +474,21 @@ def test_edit_object_with_permission(self):
...
+
+                # Assert that Last Updated By table row is updated with the user that most recently modified the object
+                response = self.client.get(instance.get_absolute_url())
+                response_body = testing.extract_page_body(response.content.decode(response.charset))
+                advanced_tab_href = f"{instance.get_absolute_url()}#advanced"
```
timizuoebideri1, line 488:
```suggestion
                detail_url = instance.get_absolute_url()
                try:
                    validate(detail_url)
                except ValidationError:
                    self.skipTest("Model does not have a detail view")

                # Assert that Last Updated By table row is updated with the user that most recently modified the object
                response = self.client.get(detail_url)
                response_body = testing.extract_page_body(response.content.decode(response.charset))
                advanced_tab_href = f"{detail_url}#advanced"
```

nautobot/core/utils/lookup.py
```diff
@@ -192,3 +192,31 @@ def get_table_for_model(model):
...
+
+def get_created_and_last_updated_usernames_for_model(instance):
+    """
+    Args:
+        instance: A model class instance
+
+    Returns:
+        created_by: Username of the user that created the instance
+        last_updated_by: Username of the user that last modified the instance
+    """
```
timizuoebideri1, line 205: 😅

timizuoebideri1: TBH, I dont think you need a doctoring for this, looks quite self-explanatory to me, but if you choose I think you should have a description

timizuoebideri1 (approved):

nautobot/core/testing/views.py
```diff
@@ -333,7 +334,21 @@ def test_create_object_with_permission(self):
...
                 self.assertEqual(len(objectchanges), 1)
+                # Assert that Created By table row is updated with the user that created the object
                 self.assertEqual(objectchanges[0].action, extras_choices.ObjectChangeActionChoices.ACTION_CREATE)
+                # Validate if detail view exists
+                validate = URLValidator()
+                detail_url = instance.get_absolute_url()
+                try:
+                    validate(detail_url)
+                except ValidationError:
+                    self.skipTest("Model does not have a detail view")
```
glennmatthews, line 345: This will report the whole `test_create_object_with_permission` as skipped, which is probably not what you want here. Same concern below.

nautobot/core/views/generic.py
```diff
@@ -112,6 +115,8 @@ def get(self, request, *args, **kwargs):
             temp_fake_context = {
                 "object": instance,
                 "request": request,
+                "created_by": created_by,
+                "last_updated_by": last_updated_by,
```
glennmatthews, line 119: Doesn't *look* to me like this is actually needed here - this is for getting view configuration, not rendering the view, if I'm reading this right?

glennmatthews (approved):

Issue #4680, opened 2023-10-19
Dynamic Group with Tag-Filter without Member-Device


gygetho:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.1 on virtual machine
* Python version: 3.10.12
* Database platform, version: postgres v14.9
* Middleware(s): none

### Steps to Reproduce
1. all steps are done in GUI
2. create TAG "testtag" with scope Devices
3. make sure no device has this tag applied
4. create a dynamic group with filter based on tag from step 2
5. error occures: Select a valid choice. testtag is not one of the available choices.

6. repeat step1-2
7. make sure a device has applied the testtag
8. create a dynamic group with filter based on tag from step 2
9. everything works fine

### Expected Behavior
to have dynamic groups based on tags even if theres curently no device with this tag

### Observed Behavior
the dynamic group with filter based on tag can only be created if an device with this tag applied already exist

glennmatthews:
Thanks for the report! This may be related to the code changes made in #4322, specifically https://github.com/nautobot/nautobot/pull/4322/files#diff-2b8d45fb821f9852b3de63ce5ce1a14214ba942e34466c386578500b5e42e735R782

PR #4722, opened 2023-10-27
Fixed the bug that prevented the user from adding tags without member objects as DynamicGroup `tags` filters.

HanlinMiao:
# Closes: #4680 
# What's Changed

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/core/forms/fields.py
+++ b/nautobot/core/forms/fields.py
@@ -784,7 +784,7 @@ class TagFilterField(DynamicModelMultipleChoiceField):
         from nautobot.extras.models import Tag
 
         if queryset is None:
-            queryset = Tag.objects.filter(content_types=ContentType.objects.get_for_model(model))
+            queryset = Tag.objects.get_for_model(model)
         query_params = query_params or {}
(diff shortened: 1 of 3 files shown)


nautobot/core/forms/fields.py
```diff
@@ -781,8 +781,10 @@ class TagFilterField(DynamicModelMultipleChoiceField):
     """
 
     def __init__(self, model, *args, query_params=None, queryset=None, **kwargs):
+        from nautobot.extras.models import Tag
+
         if queryset is None:
-            queryset = model.tags.all()
+            queryset = Tag.objects.filter(content_types=ContentType.objects.get_for_model(model))
```
glennmatthews, line 787: I forget - can we do `Tag.objects.get_for_model(model)` instead?

Looks like evaluating `ContentType.objects.get_for_model()` here is no bueno when migrations haven't been run yet.

HanlinMiao: Interesting, thanks for the help!

glennmatthews (approved):

Issue #3352, opened 2023-02-22
Allow Jobs to "output" a File (ie Download to Browser)


lampwins:
As a Nelly the Network Automation Engineer I want to write a job which munges data and produces a file that I can download from the UI when the job completes so that perform some unnatural UX operation to retrieve that output (e.g. obtain the file from some other location, copy paste output text, etc).

Kircheneer:
I have something that somewhat works, just putting this here for visibility. Not recommending the use of this.

```python
from django.urls import reverse

...

# Put the following stuff into a Jobs `run` method
def run(...):
  ...
  file_name = f"{self.job_result.pk}.txt"
  encoding = "utf-8"
  with io.BytesIO(b"text_to_put_into_file") as buffer:
      # Upload file to Nautobot's database
      in_memory_uploaded_file = InMemoryUploadedFile(
          file=buffer, field_name=None, name=file_name, content_type="text/csv", size=None, charset=encoding
      )
      file_proxy_pk = self.save_file(in_memory_uploaded_file)
  
  # Make file available for download
  file_name = FileProxy.objects.get(pk=file_proxy_pk).file.name
  self.log_info(
      obj=None,
      message=f"Download CSV [here]({reverse('db_file_storage.download_file')}?name={file_name})."
      f" It is recommended that you delete this file"
      f" [here]({reverse('admin:extras_fileproxy_change', args=[file_proxy_pk])}) once you have downloaded"
      f" it.",
  )
  ```

bryanculver:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Nautobot-Core-Product-64a8783b7a3f4b1225973a0c/issues/nautobot/nautobot/3352?planning-poker) with Zenhub @glennmatthews @gsnider2195 @HanlinMiao @jathanism @timizuoebideri1

bryanculver:
We should document what the file size limit is with `FileProxy` since at present it is using the database as a "file system".

jathanism:
> We should document what the file size limit is with `FileProxy` since at present it is using the database as a "file system".

I think we had discussed this but it's not captured here, but seeing as the underlying model for `FileProxy.file` is `FileAttachment.bytes`, which is a `BinaryField` without a `max_length`, the file size will be limited by the database backend.

[...]

PR #4723, opened 2023-10-27
Job file output

glennmatthews:
# Closes: #3352
# What's Changed

- Add `Job.create_file(filename, content)` API, which uses our existing `FileProxy` model to create stored files.
   - Generated output files are accessible through the JobResult detail view.
- Add `JOB_STORAGE_BACKEND` settings variable which controls the behavior of `FileProxy`, defaults to `DatabaseFileStorage`, but can be set to others like `FileSystemStorage` and `S3Boto3Storage` if desired.

image

# TODO

- [x] Unit, Integration Tests


nautobot/docs/release-notes/version-2.1.md
```diff
@@ -0,0 +1,14 @@
...
+
+This document describes all new features and changes in Nautobot 2.1.
+
+## Release Overview
+
+### Added
+
+#### Job file outputs ([#3352](https://github.com/nautobot/nautobot/issues/3352))
+
+The `Job` base class now includes a [`create_file(filename, content)`](../development/jobs/index.md#file-output) method which can be called by a Job to create a persistent file with the provided content when run. This file will linked from the Job Result detail view for subsequent download by users.
```
HanlinMiao, line 11:
```suggestion
The `Job` base class now includes a [`create_file(filename, content)`](../development/jobs/index.md#file-output) method which can be called by a Job to create a persistent file with the provided content when run. This file will be linked from the Job Result detail view for subsequent download by users.
```

nautobot/extras/templates/extras/jobresult.html
```diff
@@ -118,6 +118,30 @@
                         {% include 'extras/inc/json_data.html' with data=result.result format="json" %}
                     </div>
                 </div>
+                <div class="panel panel-default">
```
HanlinMiao, line 121: Can we get a screenshot of the JobResult page with downloadable files?

HanlinMiao (commented):
It looks great! One question: what is the behavior if we re-run the jobs that create files with the same name? I think we should document that as well somewhere.

glennmatthews:
> It looks great! One question: what is the behavior if we re-run the jobs that create files with the same name? I think we should document that as well somewhere.

Both the database backend and the filesystem backend handle this just fine - the latter by including the PK of the FileAttachment as a part of the filename to ensure uniqueness. 

One problem I should have called out that I haven't solved yet is that the `download=` attribute on the file download link (which is supposed to let us specify the filename that it will be saved under) doesn't appear to work correctly with the filesystem backend - the downloaded file still gets the actual filename from the backend (e.g. `<uuid>-output.txt` rather than just `output.txt`)

development/docker-compose.yml
```diff
@@ -47,8 +53,10 @@ services:
         - "-c"
         - "nautobot-server celery inspect ping --destination celery@$$HOSTNAME"
     depends_on:
-      - nautobot
-      - redis
+      nautobot:
+        condition: service_healthy
```
glennmatthews, line 57: This keeps the `celery_worker` container from attempting to startup while the main `nautobot` container is still in the midst of running initial migrations.

examples/s3_static_files/README.md
```diff
@@ -10,7 +10,7 @@ TLDR:
 
 ```shell
 echo "nautobot[remote_storage]" >> $NAUTOBOT_ROOT/local_requirements.txt
-pip3 install "nautobot[remote_storage]"
+pip3 install "nautobot[remote_storage]" boto3
```
glennmatthews, line 13: I noticed that installing `nautobot[remote_storage]` doesn't actually include `boto3` by default, I guess because `django-storages` supports many different backends, not just S3.

nautobot/core/urls.py
```diff
@@ -31,7 +31,7 @@
     path("api/", include("nautobot.core.api.urls")),
     # GraphQL
     path("graphql/", CustomGraphQLView.as_view(graphiql=True), name="graphql"),
-    # Serving static media in Django
+    # Serving static media in Django (TODO: should be DEBUG mode only - "This view is NOT hardened for production use")
     path("media/<path:path>", serve, {"document_root": settings.MEDIA_ROOT}),
```
glennmatthews, line 35: 😬

nautobot/docs/release-notes/version-2.1.md
```diff
@@ -0,0 +1,14 @@
+# Nautobot v2.1
+
+This document describes all new features and changes in Nautobot 2.1.
+
+## Release Overview
+
+### Added
```
glennmatthews, line 7: We need to add the other significant 2.1 features already merged into the release overview:

- Bootstrap/CSS changes
- Admin log view
- isnull filters

nautobot/extras/jobs.py
```diff
@@ -757,20 +760,20 @@ def prepare_job_kwargs(cls, job_kwargs):
         return {k: v for k, v in job_kwargs.items() if k in job_vars}
 
     @staticmethod
-    def load_file(pk):
+    def _load_file_from_proxy(pk):
```
glennmatthews, line 763: *technically* a breaking change to rename these methods but AFAICT no one outside of core is using them, nor should they - these should have been marked as private methods from the start.

nautobot/extras/jobs.py
```diff
@@ -823,6 +826,11 @@ def load_json(self, filename):
 
         return data
 
+    def create_file(self, filename, content):
```
glennmatthews, line 829: Whoops, need a docstring here!

nautobot/extras/models/models.py
```diff
@@ -473,29 +474,55 @@ class Meta:
...
 
-    The `file` field can be used like a file handle. The file contents are stored and retrieved from
-    `FileAttachment` objects.
+    The `file` field can be used like a file handle.
 
-    The associated `FileAttachment` is removed when `delete()` is called. For this reason, one
-    should never use bulk delete operations on `FileProxy` objects, unless `FileAttachment` objects
-    are also bulk-deleted, because a model's `delete()` method is not called during bulk operations.
+    The file contents are stored and retrieved from `FileAttachment` objects,
+    if DEFAULT_FILE_STORAGE is set to use DatabaseFileStorage,
```
glennmatthews, line 511:
```suggestion
    if settings.JOB_STORAGE_BACKEND is set to use DatabaseFileStorage,
```

nautobot/docs/release-notes/version-2.1.md
glennmatthews: #4734, #4735, #4736

HanlinMiao (approved):
Do we have/do we need an issue open for the duplicate filenames download?

PR #4755, opened 2023-11-06
Removes no-else-raise Pylint exception

jvanderaa:
# Closes: Ongoing Pylint
# What's Changed

Removes pylint exception for `no-else-raise`.

# TODO

- [ ] Unit, Integration Tests


gsnider2195:
I'm of the opinion that this is one of those linters that's technically correct at the expense of readability with no clear benefit.

gsnider2195:
Please let me know if I'm missing some context for this change. I don't want to block this over readability issues if you need it merged.

glennmatthews:
I'm inclined to agree - I appreciate the effort and we absolutely should look at reducing the set of disabled rules to only those that we actually want to disable/ignore, but this particular one doesn't feel like a useful rule to enable to me.

Issue #4709, opened 2023-10-25
InterfaceForm does not filter VRFs by device


grahamjohnston:
### Environment
* Nautobot version (Docker tag too if applicable): 2.x
* Python version: 3.9
* Database platform, version: Postgres 13
* Middleware(s): N/A

### Steps to Reproduce
1. Have some number of VRFs defined.
2. Have a subnet of those VRFs applied to a device.
3. Edit an existing interface and see the full list of VRFs, not just what is assigned to the device.

### Expected Behavior
I should only see VRFs that are already assigned to the device. The expect behavior is present in InterfaceCreateForm.

### Observed Behavior
I see every VRF in the system.

PR #4758, opened 2023-11-06
Add correct VRF field to InterfaceForm, other DCIM form fixes

glennmatthews:
# Closes: #4709 
# What's Changed

- Change a number of DCIM forms `ModelChoiceField` to `DynamicModelChoiceField`.
- Make device-component edit forms include the device as a regular form field so that it can be used in query_params.
    - This is the base change needed to address #4709.
- Add correctly filtered VRF field (referencing the device field via query_params) to InterfaceForm.
- Make VMInterface edit form include the VM as a regular form field so that it can be used in query_params.
- Add missing ip_addresses field to InterfaceCreateForm.
- Add missing label/help_text to `mtu` field in InterfaceCreateForm.
- Change order of fields in interface edit template to match the order in the interface create view.

image

image

# TODO

- [x] Unit, Integration Tests


itdependsnetworks:
> Change a number of DCIM forms ModelChoiceField to DynamicModelChoiceField.

Should this be added to our best practices and/or nautobot-pylint? If so, can you do that, since if I do that I am not able to understand the nuance if this is always preferred. 


> Change order of fields in interface edit template to match the order in the interface create view.

Same as previous.

glennmatthews:
> > Change a number of DCIM forms ModelChoiceField to DynamicModelChoiceField.
> 
> Should this be added to our best practices and/or nautobot-pylint? If so, can you do that, since if I do that I am not able to understand the nuance if this is always preferred.

Not a bad idea. There will always be exceptions, but in general yes the DynamicModelChoiceField and DynamicModelMultipleChoiceField should preferred as they avoid the server-side overhead of pre-rendering all possible options of a non-dynamic Model(Multiple)ChoiceField.

> 
> > Change order of fields in interface edit template to match the order in the interface create view.
> 
> Same as previous.

This is a fairly narrow special case (we don't often have a different template/form for "create" versus "edit" operations) so I don't think so.

nautobot/dcim/forms.py
```diff
@@ -1816,6 +1813,16 @@ class ComponentCreateForm(ComponentForm):
...
 
 
+class ComponentEditForm(NautobotModelForm):
+    """
+    Base class for editing device components (models subclassed from ComponentModel).
+
+    Distinct from ComponentCreateForm in that it has a name/label instead of a name_pattern/label_pattern.
+    """
+
+    device = DynamicModelChoiceField(queryset=Device.objects.all())
```
gsnider2195, line 1823: Could there be any negative impact of being able to change the device of an existing component?

glennmatthews: Yeah, it's not something that normally would/should be done. That said, I'm not sure it's necessarily *harmful* to permit it for that 1 in 1000 use case?

An alternate approach that I tried initially was making `device` be a hidden form field, but then I had to change the logic in `forms.js` that only permits population of `query_params` from visible fields, which I was also concerned about possible side effects from.

Thoughts? Back to the drawing board on this?

gsnider2195: We could disable that form field?

```py
    device = DynamicModelChoiceField(queryset=Device.objects.all(), disabled=True)
```

gsnider2195:
From your summary:

> Add missing label/help_text to mtu field in InterfaceCreateForm

That should be `mode` right?

gsnider2195 (approved):
One minor nitpick. I do think the disabled `device`/`virtual_machine` field in the `interface`/`vm_interface` edit forms makes more sense but it's not blocking.

glennmatthews:
> From your summary:
> 
> > Add missing label/help_text to mtu field in InterfaceCreateForm
> 
> That should be `mode` right?

Yup!

nautobot/dcim/forms.py
glennmatthews: Good thought. Had to do this conditionally in `__init__` because the "edit" form is actually used as the `model_form` for the "create" use case as well, so disabling it unconditionally would make component creation fail. But this appears to work.

Issue #4750, opened 2023-11-03
CSS refinements to Bootstrap 3 UI


lampwins:
As a Nautobot User, I want the reskinned UI, to approximate the standard branded design as much as reasonably possible, so that the reskinning effort brings the most brand value.

After #19, refine any custom CSS in the Bootstrap 3 UI to further bring the UI inline with brand identity standards.

The goal here is not to get 100% perfection, rather it is to identify and implement any "quick wins" that can address glaring discrepancies.

Explicitly time-boxed to not blow this out of the water. Go after the quick, easy wins that are strait forward to implement. Also avoid any changes that present breaking or inconsistencies with apps.

lampwins:
- `>>>` on page headings
- Rounded corners on list tables
- Blue text on badges (foreground/background color)
- List table row/cell padding (data density)
- The blue text in general is too blue

HanlinMiao:
Closed by https://github.com/nautobot/nautobot/pull/4775

PR #4765, opened 2023-11-07
feature: move navbar to the left

norbert-mieczkowski-codilime:
# Closes: #4774

# What's Changed
* Moved navbar from the top of the page to the left side.
* Changed navbar color scheme to navy/orange.
* Removed "Import" buttons from navbar dropdown menus.
* Restricted dropdown menus width to the navbar width and applied word wrapping to long text items.
* Adjusted footer styles to better match new page layout.

#### User not logged in:
image

#### User logged in:
image_1

#### User profile dropdown:
image_2

#### Devices dropdown:
image_3

#### Virtualization dropdown:
image_4

#### Footer:
image_5

PR #4775, opened 2023-11-09
[WIP] CSS Refinement to CSS UI

HanlinMiao:
# Closes: #4750
# What's Changed

# TODO

- [ ] Unit, Integration Tests


nautobot/core/templates/generic/object_bulk_create.html
```diff
@@ -25,7 +26,7 @@
 {% block tabs %}{% endblock %}
     <div class="row">
         <div class="col-md-8 col-md-offset-2">
-            <h1>{% block title %}{{ obj_type|bettertitle }} Bulk Import{% endblock %}</h1>
+            <h1><img src="{% static 'img/nautobot_chevron_header.svg' %}"/>{% block title %}{{ obj_type|bettertitle }} Bulk Import{% endblock %}</h1>
```
glennmatthews, line 29: As mentioned in standup - try the following css as an alternative to updating a bunch of templates:

```css
h1::before {
    content: url("/static/img/nautobot_chevron_header.svg");
    position: relative;
    top: 15px;
}
```

Note that for "correctness" we'd need a templated CSS file somewhere (since we shouldn't be hard-coding `/static/`) as a way to inject this (perhaps https://stackoverflow.com/a/55225412?) but it's a thought...!

nautobot/core/templates/generic/object_retrieve.html
```diff
@@ -58,7 +58,12 @@
     </div>
 
     {% block masthead %}
-        <h1>{% block title %}{{ object }}{% endblock %}</h1>
+        <h1>
+            <span class="hover_copy"><span id="copy_title">{% block title %}{{ object.display|default:object }}{% endblock %}</span><button type="button" class="btn btn-inline btn-default hover_copy_button" data-clipboard-target="#copy_title">
```
glennmatthews, line 62: nit, perhaps:

```suggestion
            <span class="hover_copy">
                <span id="copy_title">{% block title %}{{ object.display|default:object }}{% endblock %}</span>
                <button type="button" class="btn btn-inline btn-default hover_copy_button" data-clipboard-target="#copy_title">
```

nautobot/core/templates/inc/media.html
```diff
@@ -83,6 +83,11 @@
             src: url('{% static 'fonts/UbuntuMono-Bold.woff2' %}') format('woff2');
             unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+0304, U+0308, U+0329, U+2000-206F, U+2074, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
         }
+        h1::before {
+            content: url("{% static 'img/nautobot_chevron_header.svg' %}");
+            position: relative;
+            top: 15px;
+        }
```
glennmatthews, line 90: Clever place to put this! Not sure it's a pattern we want to propagate though... it'd probably be better in the long term to have a templated css file to contain this and the font-faces above rather that sticking it all directly in `media.html`, maybe?

nautobot/core/templates/generic/object_retrieve.html
```diff
@@ -58,7 +58,12 @@
     </div>
 
     {% block masthead %}
-        <h1>{% block title %}{{ object }}{% endblock %}</h1>
+        <h1>
```
glennmatthews, line 61: Would it make sense to have a `.added` change fragment? Something like:

```
Added "copy" button support to titles of all object retrieve views.
```

nautobot/core/templates/template.css
```diff
@@ -0,0 +1,87 @@
+{% load static %}
```
HanlinMiao, line 1: I have to put this in the core/template subfolder so that the django template engine can find it.
[Here, I prepended a 'css/' prefix to the url, but this is not necessary, and you don't need a "css" folder in your project; just make sure that the template engine can find "home_global.css"; that is, put it in the /template/ subfolder of any installed app, or even in the project if it is listed among the installed apps:](https://stackoverflow.com/questions/55144698/use-django-template-syntax-in-a-css-file/55225412#55225412)

nautobot/circuits/templates/circuits/circuittype_retrieve.html
```diff
@@ -26,7 +26,7 @@
             <div class="panel-heading">
                 <strong>Circuits</strong>
             </div>
-            {% include 'inc/table.html' with table=circuits_table %}
+            {% include 'inc/table.html' with detail_view=True table=circuits_table %}
```
glennmatthews, line 29: If this is something app detail views will need to adopt in order to have correct rendering, let's make sure we document it thoroughly.

What are the cases where we are using `inc/table.html` *without* `detail_view=True`? If those are less common and/or less-commonly-used-by-apps, maybe we could invert the default here?

nautobot/dcim/templates/dcim/location.html
```diff
@@ -249,7 +249,7 @@ <h2><a href="{% url 'virtualization:virtualmachine_list' %}?location={{ object.p
             <div class="panel-heading">
                 <strong>Children</strong>
             </div>
-            {% include 'inc/table.html' with table=children_table %}
+            {% include 'inc/table.html' with detail_view=True table=children_table %}
```
HanlinMiao, line 252: If detail_view=True, unwrap the table of the `panel` class.

nautobot/circuits/templates/circuits/circuittype_retrieve.html
HanlinMiao: invert the setting in the latest commit

HanlinMiao: The app detail view wouldn't need to do anything to adopt this change I believe

nautobot/core/templates/responsive_table.html
```diff
@@ -1,5 +1,5 @@
 {% load render_table from django_tables2 %}
 
 <div class="table-responsive">
-    {% render_table table table_inc_template|default:'inc/table.html' %}
+    {% include table_inc_template|default:'inc/table.html' with list_view=True %}
```
HanlinMiao, line 4: After manual testing, this is functionally the same as `render_table table`

glennmatthews: Looking at the `render_table` definition, the main concern I see is with regard to how it handles TemplateColumn: (https://github.com/jieter/django-tables2/blob/master/django_tables2/templatetags/django_tables2.py#L158-L164)

HanlinMiao: Would it be acceptable to use a separate template for list view table then? It is difficult to pass in a django template variable with the `with` keyword in this line `{% render_table table table_inc_template|default:'inc/table.html' %}.`

glennmatthews: I'm not sure how important that bit of code is TBH - `blame` shows it as having been untouched in the last 12 (!) years. :-) If you've tested with some various list views and the data all look OK then I'm willing to have us take that risk, personally.

glennmatthews: Otherwise yeah creating a new template file might be appropriate.

glennmatthews: Actually, there's already `nautobot/core/templates/panel_table.html`... maybe we can/should be using it in the list view?

HanlinMiao: Wow didnt see that at all, Im going to try using that.

HanlinMiao: The only noticable difference using`panel_table.html` is that when a table has no entries. It is rendered as such:
Screen Shot 2023-11-13 at 1 52 37 PM

nautobot/core/templates/template.css
```diff
@@ -0,0 +1,87 @@
...
+@font-face {
+    font-family: 'Ubuntu Mono';
+    font-style: normal;
+    font-weight: 700;
+    font-display: swap;
+    src: url('{% static 'fonts/UbuntuMono-Bold.woff2' %}') format('woff2');
+    unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+0304, U+0308, U+0329, U+2000-206F, U+2074, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
+}
+h1::before {
+    content: url("{% static 'img/nautobot_chevron_header.svg' %}");
```
glennmatthews, line 84: Let's make this configurable via the `custom_branding_or_static` templatetag - you can refer to https://github.com/nautobot/nautobot/pull/4765/commits/e158eb3035c8cece7c133225a53b8a87ae1889db for an example of the changes needed.

Also, should we:

1. specify an exact height for this image in CSS (allowing the width to vary if overridden with a different image)
2. use the same default `nautobot_chevron.svg` file as is being added in #4765 rather than a separate new file (I know it's a tiny file, but might as well?)

nautobot/core/templates/responsive_table.html
glennmatthews: I'm not a huge fan of that but I can accept it. We have so many fiddly variants of a basic object table that I'd rather we not add another new one. :-(

nautobot/core/testing/views.py
```diff
@@ -704,7 +704,7 @@ def test_list_objects_unknown_filter_strict_filtering(self):
             # TODO: it'd make test failures more readable if we strip the page headers/footers from the content
             self.assertIn("Unknown filter field", content, msg=content)
             # There should be no table rows displayed except for the empty results row
-            self.assertIn(f"No {self.model._meta.verbose_name_plural} found", content, msg=content)
+            self.assertIn(f"None", content, msg=content)
```
glennmatthews, line 707: Similar change needed at line 732?

nautobot/core/templates/template.css
HanlinMiao: 1.✅
2. The dimensions for the svg file is not modifiable as a `content:` argument in the `h1::before` tag. But as a `background-image: ` argument it can be. However, that requires us to specify both width and height to align it properly.

HanlinMiao: So width cannot be left unspecified

glennmatthews: Ah okay, good to know. If we need to retain a separate image file to get the right default dimensions that's OK by me.

nautobot/docs/user-guide/administration/configuration/optional-settings.md
```diff
@@ -114,6 +115,7 @@ These environment variables may be used to specify the values:
 * `NAUTOBOT_BRANDING_FILEPATHS_ICON_MASK`
 
 +++ 2.1.0
+    * `NAUTOBOT_BRANDING_FILEPATHS_HEADER_BULLET`
     * `NAUTOBOT_BRANDING_FILEPATHS_NAV_BULLET`
```
glennmatthews, line 119: Markdownlint seems unhappy with this though the rendered page looks fine to me. May be simplest just to disable rule `MD037/no-space-in-emphasis` inline?

glennmatthews (approved):

HanlinMiao:
Bootstrap changes https://github.com/nautobot/nautobot-bootstrap/pull/7

PR #4779, opened 2023-11-10
avoid crash when syncing a git repo without a worker

cardoe:
If a worker is not available and you attempt to sync a git repo, a crash will happen because you fall through to the case to redirecting based on the job_result variable. But there is no job_result variable in scope at this point. Instead return an HTTP 400 Bad Request to the user.

### DIFF
--- a/nautobot/extras/views.py
+++ b/nautobot/extras/views.py
@@ -904,6 +904,7 @@ def check_and_call_git_repository_function(request, pk, func):
     # Allow execution only if a worker process is running.
     if not get_worker_count():
         messages.error(request, "Unable to run job: Celery worker process not running.")
+        return redirect(request.get_full_path(), permanent=False)
     else:
         repository = get_object_or_404(GitRepository, pk=pk)
         job_result = func(repository, request.user)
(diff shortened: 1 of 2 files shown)


nautobot/extras/views.py
```diff
@@ -903,7 +903,9 @@ def check_and_call_git_repository_function(request, pk, func):
 
     # Allow execution only if a worker process is running.
     if not get_worker_count():
-        messages.error(request, "Unable to run job: Celery worker process not running.")
+        msg = "Unable to run job: Celery worker process not running."
+        messages.error(request, msg)
+        return HttpResponseBadRequest(msg)
```
glennmatthews, line 908: It might be a nicer user experience if we instead redirected back to the original view (GitRepositorySyncView/GitRepositoryDryRunView) rather than displaying a 400 error page?

glennmatthews (commented):
Thank you for the pull request!

nautobot/extras/views.py
cardoe: I agree I think that is better given the flash capabilities of the Django messages. But I'll be real honest, I don't know how to do it correctly. Taking a guess in my most recent push.

glennmatthews (approved):

PR #4799, opened 2023-11-14
Exclude docs source and animated gifs from built package

glennmatthews:
# Closes: #N/A
# What's Changed

- Update mkdocs to version 1.5.3 (needed because v1.5 adds support for the `exclude_docs` parameter)
- Update `mkdocs.yml` to exclude the `media/*.gif` files from the built docs (as the rendered docs don't actually reference these files, only `README.md` does, and they're quite large files)
- Update `mkdocs.yml` with recommended strict `validation` options added in mkdocs 1.5
- Update `pyproject.toml` to not package `CONTRIBUTING.md`, `CHANGELOG.md`, `NOTICE` files in the built package (since these aren't in the `nautobot/` directory, they end up in the `site-packages` root directory as clutter when pip installing nautobot from the package)
- Update `pyproject.toml` to not package `nautobot/docs/**/*` in the built package since it's redundant with the built documentation and resulted in a duplicate copy of every documentation image in the built package.

**Net-net here is to reduce the sdist and wheel files built for Nautobot from 85/86 MB in size down to a mere 30/31 MB.**

# TODO

- n/a Unit, Integration Tests


glennmatthews:
Will need double-commit to ltm/1.6 as well.

gsnider2195:
Since we're on mkdocs v1.5 now we can suppress this warning message using the [`not_in_nav`](https://www.mkdocs.org/user-guide/configuration/#not_in_nav) config:

```
INFO     -  The following pages exist in the docs directory, but are not included in the "nav" configuration:
              - models/examplemodel.md
```

Not sure if you want to include it in this PR, it's just something I noticed when looking at the docs.

gsnider2195:
I'm seeing this warning now when building docs:

```
INFO    -  DeprecationWarning: A plugin has set File.page to an instance of Page and it got overwritten. The behavior of this will change in MkDocs 1.6.
             File "/usr/local/lib/python3.8/site-packages/mkdocs/structure/nav.py", line 183, in <listcomp>
               _data_to_navigation((key, value), files, config)
             File "/usr/local/lib/python3.8/site-packages/mkdocs/structure/nav.py", line 209, in _data_to_navigation
               warnings.warn(
```

It's not an issue yet, but we may get questions about it.

glennmatthews:
Yes, I noticed that too. I wish it was clearer about *which* plugin is doing that :-)

I'll dig into it and see if I can find out where it's coming from and if an update is already available.

gsnider2195:
> Yes, I noticed that too. I wish it was clearer about _which_ plugin is doing that :-)
> 
> I'll dig into it and see if I can find out where it's coming from and if an update is already available.

I did a grep through the `mkdocs*` plugins and couldn't find anything :(

glennmatthews:
The interesting thing is that it looks like in the development branch for mkdocs 1.6 they've just removed that warning altogether - so it may be that there's not actually any action required in this case?

HanlinMiao (approved):

Issue #4791, opened 2023-11-13
Regression - Dynamic filter form population can alter presentation of ChoiceFields elsewhere in the UI


grahamjohnston:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.3

### Steps to Reproduce
1. Log in to Nautobot
2. In one tab, open the /dcim/devices/ list view, open the Filter modal, and switch to the Advanced tab.
3. In another browser tab, open the "run" view for a Job that defines a ChoiceVar. Observe that the ChoiceVar is correctly represented as a single-value Select2 widget in the UI.
4. In the first tab, in the Filter modal, select "Rack Face" in the first column, then "exact" in the second column.
5. Refresh the second tab.

### Expected Behavior
No change on page refresh

### Observed Behavior
The `ChoiceVar` now gets incorrectly rendered as a multi-value Select2 widget.

There is a belief that the issue lies here: https://github.com/nautobot/nautobot/blob/develop/nautobot/core/utils/filtering.py#L152-L155 - compare against https://github.com/nautobot/nautobot/blob/ltm-1.6/nautobot/utilities/utils.py#L857-L936 and specifically https://github.com/nautobot/nautobot/pull/4223/files#diff-1163f5bad3157eeb8af809244f6b84ad3b27580d2ba6ba4907e4c32525025a2c

joewesch:
Assumed regression of #4222

PR #4801, opened 2023-11-14
Fixed a bug where some form fields may change unexpectedly

gsnider2195:
# Closes: #4791
# What's Changed

# TODO

- [ ] Unit, Integration Tests


nautobot/core/tests/test_utils.py
```diff
@@ -442,58 +442,58 @@ def test_get_filterset_parameter_form_field(self):
             location_fields = ["comments", "name", "contact_email", "physical_address", "shipping_address"]
             for field_name in location_fields:
                 form_field = filtering.get_filterset_parameter_form_field(dcim_models.Location, field_name)
-                self.assertIsInstance(form_field, django_forms.CharField)
+                self.assertIs(type(form_field), forms.MultiValueCharField)
```
gsnider2195, line 445: I changed all of these because `isinstance(field, ChoiceField)` was matching when the field was a `MultipleChoiceField` since it's a subclass of `ChoiceField`. Matching subclasses doesn't seem to be what we were trying to achieve here.

glennmatthews: Makes sense for the test to be more specific, agreed.

nautobot/core/utils/filtering.py
```diff
@@ -150,11 +150,7 @@ def get_filterset_parameter_form_field(model, parameter, filterset=None):
                 choices_as_strings=True, queryset=queryset_map[plural_name]().as_queryset()
             )
     elif isinstance(field, (MultipleChoiceFilter, ChoiceFilter)) and "choices" in field.extra:
-        form_field_class = forms.ChoiceField
-        form_field_class.widget = StaticSelect2Multiple()
-        form_attr = {"choices": field.extra.get("choices")}
-
-        form_field = form_field_class(**form_attr)
+        form_field = forms.MultipleChoiceField(choices=field.extra.get("choices"), widget=StaticSelect2Multiple)
```
glennmatthews, line 153: In the [original fix](https://github.com/nautobot/nautobot/commit/4175e8b6ab4592c4ddd249ac1a7f5b1b43038c7e) a similar change was made to an `elif isinstance(field, BooleanFilter)` check that seems to not even exist in 2.0. Would you mind for my curiosity chasing down why that check is no longer present/needed?

gsnider2195: It was removed completely in https://github.com/nautobot/nautobot/pull/3479/files#diff-56c997d3b0c57888c53b79e77be7d85c6302706fbf9f829aba3ad3bd2cd373d5

I assume because the default behavior was determined to be the desired behavior? We don't have many BooleanFilters but I can try to add a test for one.

gsnider2195: Looks like that one is broken. When you change values in the "has local config context data" filter in the device list view you get this JS error in the console. It looks like our form code is expecting everything to use select2 but I can't figure out where that's happening.

image

gsnider2195: Could be [here](https://github.com/nautobot/nautobot/blob/cafab53d861cc625cdde9da11121895d7e94a8a0/nautobot/project-static/js/forms.js#L539)

glennmatthews: Darn that @glennmatthews, always causing problems...

glennmatthews (approved):

nautobot/core/utils/filtering.py
gsnider2195: The original fix had a ChoiceField for booleans but NullBooleanField looks a little more flexible with the values it accepts. It doesn't seem to make a difference in this context though, so should we stick with ChoiceField for consistency with ltm-1.6?

glennmatthews: I *think* ChoiceField rather than NullBooleanField is appropriate in the filter context, as I *think* the user shouldn't be able to set a filter of `has_local_config_context_data=null`?

PR #4810, opened 2023-11-15
Fix multiple jobhooks and/or webhooks triggered by a single change

gsnider2195:
# Closes: #3015
# What's Changed

Also fixes change logging behavior to properly combine multiple changes to an object in the same request

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/context_managers.py
+++ b/nautobot/extras/context_managers.py
@@ -144,14 +144,15 @@ def web_request_context(
         request = RequestFactory().request(SERVER_NAME="web_request_context")
         request.user = user
     change_context = valid_contexts[context](request=request, context_detail=context_detail, change_id=change_id)
-    with change_logging(change_context):
-        yield request
-
-    # enqueue jobhooks and webhooks
-    for oc in (
-        ObjectChange.objects.filter(request_id=change_context.change_id)
-        .exclude(action=ObjectChangeActionChoices.ACTION_DELETE)
-        .iterator()
-    ):
-        enqueue_job_hooks(oc)
-        enqueue_webhooks(oc.changed_object, oc.user, oc.request_id, oc.action)
+    try:
+        with change_logging(change_context):
+            yield request
+    finally:
+        # enqueue jobhooks and webhooks
+        for object_change in (
+            ObjectChange.objects.filter(request_id=change_context.change_id)
+            .exclude(action=ObjectChangeActionChoices.ACTION_DELETE)
+            .iterator()
+        ):
+            enqueue_job_hooks(object_change)
+            enqueue_webhooks(object_change)
--- a/nautobot/extras/context_managers.py
+++ b/nautobot/extras/context_managers.py
@@ -5,7 +5,7 @@ from django.contrib.auth import get_user_model
 from django.contrib.auth.models import AnonymousUser
 from django.test.client import RequestFactory
 
-from nautobot.extras.choices import ObjectChangeActionChoices, ObjectChangeEventContextChoices
+from nautobot.extras.choices import ObjectChangeEventContextChoices
 from nautobot.extras.models import ObjectChange
 from nautobot.extras.signals import change_context_state
 from nautobot.extras.webhooks import enqueue_webhooks
@@ -149,10 +149,6 @@ def web_request_context(
             yield request
     finally:
         # enqueue jobhooks and webhooks
-        for object_change in (
-            ObjectChange.objects.filter(request_id=change_context.change_id)
-            .exclude(action=ObjectChangeActionChoices.ACTION_DELETE)
-            .iterator()
-        ):
+        for object_change in ObjectChange.objects.filter(request_id=change_context.change_id).iterator():
             enqueue_job_hooks(object_change)
             enqueue_webhooks(object_change)
--- a/nautobot/extras/context_managers.py
+++ b/nautobot/extras/context_managers.py
@@ -2,10 +2,13 @@ import uuid
 from contextlib import contextmanager
 
 from django.contrib.auth import get_user_model
+from django.contrib.auth.models import AnonymousUser
 from django.test.client import RequestFactory
 
-from nautobot.extras.choices import ObjectChangeEventContextChoices
+from nautobot.extras.choices import ObjectChangeActionChoices, ObjectChangeEventContextChoices
+from nautobot.extras.models import ObjectChange
 from nautobot.extras.signals import change_context_state
+from nautobot.extras.webhooks import enqueue_webhooks
 
 
 class ChangeContext:
@@ -95,7 +98,9 @@ def change_logging(change_context):
 
 
 @contextmanager
-def web_request_context(user, context_detail="", change_id=None):
+def web_request_context(
+    user, context_detail="", change_id=None, context=ObjectChangeEventContextChoices.CONTEXT_ORM, request=None
+):
     """
     Emulate the context of an HTTP request, which provides functions like change logging and webhook processing
     in response to data changes. This context manager is for use with low level utility tooling, such as the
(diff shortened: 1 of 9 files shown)


nautobot/extras/context_managers.py
```diff
@@ -116,13 +121,34 @@ def web_request_context(user, context_detail="", change_id=None):
...
+    if request is None:
+        request = RequestFactory().request(SERVER_NAME="web_request_context")
+        request.user = user
+    change_context = valid_contexts[context](request=request, context_detail=context_detail, change_id=change_id)
+    try:
+        with change_logging(change_context):
+            yield request
+    finally:
+        # enqueue jobhooks and webhooks
+        for object_change in ObjectChange.objects.filter(request_id=change_context.change_id).iterator():
```
glennmatthews, line 152: Is `change_context.change_id` different from `change_id`, or should we use the latter for simplicity?

nautobot/extras/jobs.py
```diff
@@ -1178,12 +1177,12 @@ def enqueue_job_hooks(object_change):
...
-    model_type = object_change.changed_object._meta.model
-    if model_type not in ChangeLoggedModelsQuery().list_subclasses():
+    model_type = object_change.changed_object_type
+    if model_type not in ChangeLoggedModelsQuery().as_queryset():
         return
 
     # Retrieve any applicable job hooks
-    content_type = ContentType.objects.get_for_model(object_change.changed_object)
+    content_type = object_change.changed_object_type
     action_flag = {
```
glennmatthews, line 1186: Looks like there's some room for cleanup here as now `model_type` and `content_type` are equivalent variables.

nautobot/extras/signals.py
```diff
@@ -55,55 +61,50 @@ def _handle_changed_object(sender, instance, raw=False, **kwargs):
...
-                objectchange.user = _get_user_if_authenticated(change_context_state.get().get_user(), objectchange)
-                objectchange.request_id = change_context_state.get().change_id
-                objectchange.change_context = change_context_state.get().context
-                objectchange.change_context_detail = change_context_state.get().context_detail[
-                    :CHANGELOG_MAX_CHANGE_CONTEXT_DETAIL
-                ]
-                objectchange.save()
+
+        # if a change already exists for this change_id, user, and object, update it instead of creating a new one
+        # if the object was deleted and then recreated with the same pk (don't do this), change the action to update
```
glennmatthews, line 90: Readability?

```suggestion
        # If a change already exists for this change_id, user, and object, update it instead of creating a new one.
        # If the object was deleted then recreated with the same pk (don't do this), change the action to update.
```

nautobot/extras/signals.py
```diff
@@ -55,55 +61,50 @@ def _handle_changed_object(sender, instance, raw=False, **kwargs):
...
+        related_changes = ObjectChange.objects.filter(
+            changed_object_type=ContentType.objects.get_for_model(instance),
+            changed_object_id=instance.pk,
+            user=user,
+            request_id=change_context_state.get().change_id,
+        )
+        objectchange = instance.to_objectchange(action)
+        if related_changes.exists():
```
glennmatthews, line 98: Wonder if it would be interesting/useful to at least check whether `related_changes.count() > 1` (shouldn't ever be the case now, right?) and log a message if it somehow is the case?

nautobot/extras/tests/test_context_managers.py
```diff
@@ -56,6 +56,101 @@ def test_change_log_created(self):
...
+    def test_create_then_delete(self):
+        """Test that a create followed by a delete is logged as two changes"""
+        with web_request_context(self.user):
+            location_type = LocationType.objects.get(name="Campus")
+            location_status = Status.objects.get_for_model(Location).first()
+            location = Location(name="Test Location 1", location_type=location_type, status=location_status)
+            location.save()
+        with web_request_context(self.user):
+            location_pk = location.pk
+            location.delete()
```
glennmatthews, line 68: Should the create-then-delete case be tested in a single `web_request_context` as well?

nautobot/extras/tests/test_context_managers.py
```diff
@@ -56,6 +56,101 @@ def test_change_log_created(self):
         self.assertEqual(oc_list[0].changed_object, location)
         self.assertEqual(oc_list[0].action, ObjectChangeActionChoices.ACTION_CREATE)
 
+    def test_create_then_delete(self):
+        """Test that a create followed by a delete is logged as two changes"""
+        with web_request_context(self.user):
+            location_type = LocationType.objects.get(name="Campus")
+            location_status = Status.objects.get_for_model(Location).first()
```
glennmatthews, line 63: For consistency/readability with the following tests:

```suggestion
        location_type = LocationType.objects.get(name="Campus")
        location_status = Status.objects.get_for_model(Location).first()
        with web_request_context(self.user):
```

nautobot/extras/tests/test_context_managers.py
```diff
@@ -56,6 +56,101 @@ def test_change_log_created(self):
...
+            location = Location(name="Test Location 1", location_type=location_type, status=location_status)
+            location.save()
+        with web_request_context(self.user):
+            location_pk = location.pk
```
glennmatthews, line 67: For consistency with the following tests:

```suggestion
            location_pk = location.pk
        with web_request_context(self.user):
```

nautobot/extras/tests/test_context_managers.py
```diff
@@ -56,6 +56,101 @@ def test_change_log_created(self):
...
+    def test_delete_then_create(self):
+        """Test that a delete followed by a create is logged as a single update"""
+        location_type = LocationType.objects.get(name="Campus")
+        location_status = Status.objects.get_for_model(Location).first()
+        with web_request_context(self.user):
+            location = Location(
+                name="Test Location 1",
+                location_type=location_type,
+                status=location_status,
+            )
```
glennmatthews, line 129: self-consistency?

```suggestion
            location = Location(name="Test Location 1", location_type=location_type, status=location_status)
```

nautobot/extras/tests/test_context_managers.py
```diff
@@ -56,6 +56,101 @@ def test_change_log_created(self):
...
+        location = Location.objects.filter(pk=location_pk)
+        self.assertFalse(location.exists())
+        oc_list = get_changes_for_model(Location).filter(changed_object_id=location_pk)
+        self.assertEqual(len(oc_list), 2)
+        self.assertEqual(oc_list[0].action, ObjectChangeActionChoices.ACTION_DELETE)
+        snapshots = oc_list[0].get_snapshots()
+        self.assertIsNotNone(snapshots["prechange"])
+        self.assertIsNone(snapshots["postchange"])
+        self.assertIsNone(snapshots["differences"]["added"])
+        self.assertEqual(snapshots["differences"]["removed"]["description"], "")
```
glennmatthews, line 99: Interesting, good test.

nautobot/extras/tests/test_webhooks.py
```diff
@@ -100,11 +100,13 @@ class FakeResponse:
...
 
-            with web_request_context(self.user, change_id=request_id):
+            # create the object to be updated in a separate context
+            with web_request_context(self.user):
                 location_type = LocationType.objects.get(name="Campus")
                 location = Location(name="Location 1", status=self.statuses[0], location_type=location_type)
                 location.save()
 
+            with web_request_context(self.user, change_id=request_id):
                 location.name = "Location Update"
```
glennmatthews, line 110: Should we also have tests for the "create and update in a single web_request_context" case to make sure the webhooks are correctly triggered or not triggered in that case?

nautobot/extras/webhooks.py
```diff
@@ -1,56 +1,52 @@
...
-        serializer_context = {
-            "request": None,
-            "depth": 1,
-        }
-        serializer = serializer_class(instance, context=serializer_context)
-        most_recent_change = get_changes_for_model(instance).first()
+        # fall back to object_data if object_data_v2 is not available
+        serialized_data = object_change.object_data_v2
+        if serialized_data is None:
+            serialized_data = object_change.object_data
```
glennmatthews, line 33: Wonder if this is even needed at this point?

nautobot/extras/webhooks.py
```diff
@@ -1,56 +1,52 @@
...
+
+        most_recent_change = ObjectChange.objects.filter(
```
glennmatthews, line 35: Wait... how is `most_recent_change` different than the provided `object_change`? Shouldn't `object_change` already be the most recent change?

nautobot/extras/context_managers.py
```diff
@@ -116,13 +121,34 @@ def web_request_context(user, context_detail="", change_id=None):
...
 
-    request = RequestFactory().request(SERVER_NAME="web_request_context")
-    request.user = user
-    change_context = ORMChangeContext(request=request, context_detail=context_detail, change_id=change_id)
-    with change_logging(change_context):
-        yield request
+    if context not in valid_contexts:
+        raise TypeError(f"{context} is not a valid context")
+
+    if not isinstance(user, (get_user_model(), AnonymousUser)):
```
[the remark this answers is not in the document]
gsnider2195: We're not but since every web request is being funneled through this we have to account for AnonymousUser. We could exit early if the user is AnonymousUser but from a security perspective I'd rather the change be logged with user=None

nautobot/extras/signals.py
gsnider2195: There could be more than one in the case where an object is deleted and recreated with the same pk

nautobot/extras/tests/test_context_managers.py
gsnider2195: It should. Thanks for catching that

nautobot/extras/webhooks.py
gsnider2195: `object_change` is the object change we want to fire this webhook for! This was from the previous signature where the instance was being passed in. Good catch!

nautobot/extras/webhooks.py
gsnider2195: We should probably have a generic test that iterates through `registry["model_features"]["webhooks"]` and checks for `to_objectchange` and an api serializer for the model?

nautobot/extras/context_managers.py
gsnider2195: It could be. If `change_id` is `None` the change_context generates one.

glennmatthews: Thanks for the info! Maybe add a comment to this effect?

nautobot/extras/webhooks.py
glennmatthews: Sounds like a great idea.

nautobot/extras/webhooks.py
```diff
@@ -1,56 +1,52 @@
...
-    app_label = instance._meta.app_label
-    model_name = instance._meta.model_name
+    app_label = object_change.changed_object_type.app_label
+    model_name = object_change.changed_object_type.model
     if model_name not in registry["model_features"]["webhooks"].get(app_label, []):
         return
 
     # Retrieve any applicable Webhooks
-    content_type = ContentType.objects.get_for_model(instance)
+    content_type = object_change.changed_object_type
```
HanlinMiao, line 21: Could probably move this up to line 14 and use it for `app_label` and `model_name`

nautobot/extras/signals.py
```diff
@@ -31,11 +37,11 @@
 
 
 #
-# Change logging/webhooks
+# Change logging
 #
 
 
-def _get_user_if_authenticated(user, objectchange):
+def _get_user_if_authenticated(user, instance):
```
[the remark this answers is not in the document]
gsnider2195: This isn't exposed through the apps api or referenced in the docs and it's also prepended with an underscore which indicates a private method.

nautobot/extras/webhooks.py
```diff
@@ -1,56 +1,52 @@
...
-from nautobot.extras.models import Webhook
+from nautobot.extras.choices import ObjectChangeActionChoices
+from nautobot.extras.models import ObjectChange, Webhook
 from nautobot.extras.registry import registry
 from nautobot.extras.tasks import process_webhook
-from .choices import ObjectChangeActionChoices
 
 
-def enqueue_webhooks(instance, user, request_id, action):
+def enqueue_webhooks(object_change):
```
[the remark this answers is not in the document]
gsnider2195: This isn't exposed through the apps api or referenced in the docs so we can safely consider it a private api.

glennmatthews (approved):
Looks solid to me. Let's get pylint happy and :shipit:.

Issue #4813, opened 2023-11-16
gen_secret_key.py uses system (urandom) PRNG


byteborg:
### Environment
* Nautobot version (Docker tag too if applicable): develop
* Python version: 3.9.6
* Database platform, version: n/a
* Middleware(s): n/a

### Steps to Reproduce
1. gen_secret_key.py uses "random" instead of "secrets" module
2.
3.

### Expected Behavior
gen_secret_key.py outputs a unique 50 char secret

### Observed Behavior
gen_secret_key.py uses random.SystemRandom() which does not guarantee randomness for cryptohraphic tokens and secrets

PR #4814, opened 2023-11-16
Use secrets modules for key generation

byteborg:
# Closes: #4813
# What's Changed
    - replace "random" with "secrets" module in generate_secret_key.py
    - pull secret generation implementation from example from Python module reference

Output:
```
$ ./generate_secret_key.py
x7f@iv%fX31aGqGZ(gV=cpCdhMr_Jb3=^Rq!Ce3!q-87GIb)eP
$ ./generate_secret_key.py
M%SWjF4Jd!)luIRYl)YPN_jc^n6Ug#v#xHDu@JS4#zZDZskCgm
```

### DIFF
--- a/nautobot/generate_secret_key.py
+++ b/nautobot/generate_secret_key.py
@@ -1,8 +1,15 @@
 #!/usr/bin/env python
 # This script will generate a random 50-character string suitable for use as a SECRET_KEY.
-from random import SystemRandom
+# shamelessly pulled from https://docs.python.org/3/library/secrets.html#recipes-and-best-practices
+import string
+import secrets
 
-choice = SystemRandom().choice
-
-charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789!@#$%^&*(-_=+)"
-print("".join(choice(charset) for _ in range(50)))
+length = 50
+alphabet = string.ascii_letters + string.digits + "!@#$%^&*(-_=+)"  # string.punctuation would be an option
+while True:
+    password = ''.join(secrets.choice(alphabet) for i in range(length))
+    if (any(c.islower() for c in password)
+            and any(c.isupper() for c in password)
+            and sum(c.isdigit() for c in password) >= 3):
+        break
+print(password)


nautobot/generate_secret_key.py
```diff
@@ -1,8 +1,15 @@
...
-
-charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789!@#$%^&*(-_=+)"
-print("".join(choice(charset) for _ in range(50)))
+length = 50
+alphabet = string.ascii_letters + string.digits + "!@#$%^&*(-_=+)"  # string.punctuation would be an option
+while True:
+    password = ''.join(secrets.choice(alphabet) for i in range(length))
+    if (any(c.islower() for c in password)
+            and any(c.isupper() for c in password)
+            and sum(c.isdigit() for c in password) >= 3):
```
glennmatthews, line 13: Looking at the Django code, the actual check that it performs at startup to confirm that the SECRET_KEY is good is a `len(set(password)) >= 5` to confirm that there's a decent number of unique characters in the key. Would it make sense to add that, or perhaps even replace this set of checks?

glennmatthews (commented):
Thanks for the contribution!

nautobot/generate_secret_key.py
byteborg: The check ensures a minimum amount of lowercase, uppercase and digits in the generated string. 

From my point of view it serves two purposes:
1. Ensure a certain mix of character classes
2. Prevent generation of substandard secrets due to a broken PRNG implementation

byteborg: Sure

byteborg:
What do I need to do to pass the changelog check?

bryanculver:
> What do I need to do to pass the changelog check?

Please create a changelog fragment: https://docs.nautobot.com/projects/core/en/stable/development/core/#creating-changelog-fragments

byteborg:
Thanks @gsnider2195 for the finishing touches! I was on leave and didn't get around to complete the commit :/

Issue #4647, opened 2023-10-13
Add device factory


gsnider2195:
We have everything that's required to create a device. It should be relatively easy to get a device factory implemented to make setting up a test data set that much easier.

PR #4825, opened 2023-11-17
Added Device Factory

HanlinMiao:
# Closes: #4647 
# What's Changed

Screen Shot 2023-11-17 at 2 39 21 PM

# TODO

- [ ] Unit, Integration Tests


nautobot/dcim/factory.py
```diff
@@ -99,6 +99,112 @@ def get_rack_reservation_units(obj):
...
+    )
+    name = factory.LazyAttributeSequence(lambda o, n: f"{o.device_type.model} Device - {o.location} - {n + 1}")
+
+    has_tenant = NautobotBoolIterator()
+    tenant = factory.Maybe("has_tenant", random_instance(Tenant))
+    has_platform = NautobotBoolIterator()
+    platform = factory.Maybe("has_platfrom", random_instance(Platform))
+
+    has_serial = NautobotBoolIterator()
+    serial = factory.Maybe("has_part_number", factory.Faker("ean", length=10), "")
```
glennmatthews, line 135:
```suggestion
    serial = factory.Maybe("has_serial", factory.Faker("ean", length=10), "")
```

nautobot/dcim/factory.py
```diff
@@ -99,6 +99,112 @@ def get_rack_reservation_units(obj):
...
+    # primary_ip4 = factory.Maybe(
+    #     "has_primary_ip4",
+    #     random_instance(lambda: IPAddress.objects.filter(ip_version=4), allow_null=False),
+    # )
+
+    # has_primary_ip6 = NautobotBoolIterator()
+    # primary_ip6 = factory.Maybe(
+    #     "has_primary_ip6",
+    #     random_instance(lambda: IPAddress.objects.filter(ip_version=6), allow_null=False),
+    # )
```
glennmatthews, line 205: Would be awesome to get these enabled, but definitely not blocking for the first pass.

nautobot/dcim/tests/test_views.py
```diff
@@ -632,6 +633,7 @@ def setUpTestData(cls):
     # factories.
     @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
     def test_bulk_edit_objects_with_constrained_permission(self):
+        Device.objects.all().delete()
```
glennmatthews, line 636: Are this one and those below redundant? We already deleted all Devices in `setUpTestData`, and the line below this is deleting all Devices except those of a specific test model?

nautobot/dcim/factory.py
```diff
@@ -99,6 +99,112 @@ def get_rack_reservation_units(obj):
...
+    location = random_instance(
+        lambda: Location.objects.filter(location_type__content_types=ContentType.objects.get_for_model(Device)),
+        allow_null=False,
+    )
+    name = factory.LazyAttributeSequence(lambda o, n: f"{o.device_type.model} Device - {o.location} - {n + 1}")
+
+    has_tenant = NautobotBoolIterator()
+    tenant = factory.Maybe("has_tenant", random_instance(Tenant))
+    has_platform = NautobotBoolIterator()
+    platform = factory.Maybe("has_platfrom", random_instance(Platform))
```
gsnider2195, line 132:
```suggestion
    platform = factory.Maybe("has_platform", random_instance(Platform))
```

The Platform manufacturer should match the DeviceType manufacturer

nautobot/dcim/factory.py
```diff
@@ -99,6 +99,156 @@ def get_rack_reservation_units(obj):
...
+    has_tenant = NautobotBoolIterator()
+    tenant = factory.Maybe("has_tenant", random_instance(Tenant))
+    has_platform = NautobotBoolIterator()
+    platform = factory.Maybe(
+        "has_platform",
+        factory.SubFactory(
+            PlatformFactory,
+            manufacturer=factory.LazyAttribute(lambda o: o.factory_parent.device_type.manufacturer),
+        ),
+    )
```
glennmatthews, line 176: This makes it so we create a new Platform for each Device that gets created, doesn't it? Would be better to use one of the existing Platforms instead, I think?

nautobot/dcim/factory.py
```diff
@@ -99,6 +99,115 @@ def get_rack_reservation_units(obj):
...
+        allow_null=False,
+    )
+    name = factory.LazyAttributeSequence(lambda o, n: f"{o.device_type.model} Device - {o.location} - {n + 1}")
+
+    has_tenant = NautobotBoolIterator()
+    tenant = factory.Maybe("has_tenant", random_instance(Tenant))
+    has_platform = NautobotBoolIterator()
+    platform = factory.Maybe(
+        "has_platform",
+        factory.LazyAttribute(lambda o: Platform.objects.filter(manufacturer=o.device_type.manufacturer).first()),
```
HanlinMiao, line 134:
```
factory.LazyAttribute(lambda o: random_instance(Platform.objects.filter(manufacturer=o.device_type.manufacturer))),
```
I wanted to use random_instance() but it would not let me. This is working for now.
This is the error I got:
```
ValueError: Cannot assign "<factory.declarations.LazyFunction object at 0x7fc76e256a40>": "Device.platform" must be a "Platform" instance.
```

glennmatthews: If you want a random Platform from the filtered queryset, you can use `.order_by("?").first()` - it's slow at scale but might be fine here.

glennmatthews (approved):

nautobot/dcim/factory.py
gsnider2195: I think you need to do it like this:

```py
def get_random_platform_for_manufacturer(manufacturer):
    qs = Platform.objects.filter(manufacturer=manufacturer)
    return factory.random.randgen.choice(qs) if qs.exists() else None

...

    platform = factory.Maybe(
        "has_platform",
        factory.LazyAttribute(lambda o: get_random_platform_for_manufacturer(o.device_type.manufacturer)),
        None,
    )
```

glennmatthews (approved):

PR #4832, opened 2023-11-21
[2.0] Fix unsafe `mark_safe` use in various features

glennmatthews:
# Closes: GHSA-cf9f-wmhp-v4pr (for 2.0.x)
# What's Changed

- Audit all uses of `mark_safe` in our code base and change to `format_html` where appropriate for security and integrity.
  - This fixes several cases where `CustomLink`, `ComputedField`, and `JobButton` were provably vulnerable to XSS
  - This also hardens a number of other cases that I wasn't able to trivially exploit but should be considered insecure as previously implemented.
- Add test cases for CustomLink, ComputedField, and JobButton rendering. These are not exhaustive but provide at least some automated validation of the fixes and protection against regression.
- As I originally discovered this issue by experimenting with using `ruff` for linting, I've added `ruff` to the invoke tasks and also to the CI (not yet validated).
   - I've enabled a few basic categories of lint rules (mostly redundant with `flake8`, but inexpensive to cover with `ruff` as well and laying groundwork for eventually removing `flake8` less urgently)
   - I've enabled the `S308` security rule that specifically flags usage of `mark_safe` (there are a lot of other rules in the S category that we should plan to enable and audit over time)
   - I've added `# noqa: S308` to the remaining cases of `mark_safe` in our code that I believe to be in fact safe. *I'D STRONGLY APPRECIATE A SANITY CHECK ON ALL OF THESE!*

# TODO

- [x] Unit, Integration Tests


HanlinMiao (approved):

nautobot/core/utils/data.py
```diff
@@ -100,7 +100,11 @@ def render_jinja2(template_code, context):
     """
     rendering_engine = engines["jinja"]
     template = rendering_engine.from_string(template_code)
-    return template.render(context=context)
+    # For reasons unknown to me, django-jinja2 `template.render()` implicitly calls `mark_safe()` on the rendered text.
```
gsnider2195, line 103: We probably want to update the docstring for this method to state that this is for templating untrusted template code. The typical use case for jinja template rendering is to trust the template code but not the user input (the context variable here). Since this is not a typical use case we should document it.

Also, if we update the docstring we could get rid of this comment and just run the rendered content through  `django.utils.html.escape`

gsnider2195: Actually, I prefer your method here in case a developer needs to work with the html in the rendered template. We don't want to prematurely escape the content.

nautobot/extras/templatetags/job_buttons.py
```diff
@@ -154,23 +158,34 @@ def job_buttons(context, obj):
                 if text_rendered:
                     template_args["button_text"] = text_rendered
                     if jb.confirmation:
-                        buttons_rendered += "<li>" + CONFIRM_BUTTON.format(**template_args) + "</li>"
-                        forms_html += CONFIRM_MODAL.format(**template_args)
+                        buttons_rendered += (
+                            mark_safe("<li>")  # noqa: S308
+                            + format_html(CONFIRM_BUTTON, **template_args)
+                            + mark_safe("</li>")  # noqa: S308
+                        )
```
gsnider2195, line 165:
```suggestion
                        buttons_rendered += format_html(f"<li>{CONFIRM_BUTTON}</li>", **template_args)
```

Not blocking, just more concise and accomplishes the same thing?

nautobot/extras/tests/test_views.py
```diff
@@ -130,6 +131,63 @@ def setUpTestData(cls):
         cls.slug_test_object = "Computed Field Five"
 
 
+class ComputedFieldRenderingTestCase(TestCase):
```
gsnider2195, line 134: Great tests!

gsnider2195 (approved):

nautobot/extras/templatetags/job_buttons.py
glennmatthews: Probably so, good thought.

PR #4845, opened 2023-11-22
Fixes to support multiple Apps per package

snaselj:
# Closes: NaN

## What's Changed

- Fixed the following methods to use to use `app_config.name` instead of `app_label` + conditional logic.
    - `nautobot.core.api.utils.get_serializer_for_model`
    - `nautobot.core.utils.lookup.get_route_for_model`
    - `nautobot.core.utils.lookup.get_related_class_for_model`
    - `nautobot.extras.plugins.tables.InstalledPluginsTable`
    - `nautobot.extras.plugins.views.InstalledPluginsView`

## TODO


nautobot/core/api/utils.py
```diff
@@ -103,20 +103,21 @@ def get_serializer_for_model(model, prefix=""):
...
-        serializer_name = f"nautobot.{serializer_name}"
+    module_parts = model.__module__.split(".")
+    name = ""
+    while name != "models":
+        name = module_parts.pop()
+    # There could be double models in the module path, so we need to pop again
+    while module_parts[-1] == "models":
+        module_parts.pop()
+
+    serializer_name = f"{'.'.join(module_parts)}.api.serializers.{prefix}{model.__name__}Serializer"
```
glennmatthews, line 114: It's not well documented but `model` can be a model class **or** a model _instance_. I think the latter is why the tests are presently failing, as an instance would need to do `model.__class__.__name__` instead of `model.__name__`?

snaselj:
> @snaselj What else is there to be done to get it across the finish line?

This PR is currently taken over by @whitej6, there will probably be more fixes comming here. Or would you rather finish this one and add other fixes to separate PRs? @HanlinMiao

whitej6:
@HanlinMiao it is ready for review.

nautobot/core/api/utils.py
```diff
@@ -103,20 +104,20 @@ def get_serializer_for_model(model, prefix=""):
...
+    if app_label == "contenttypes" and model_name == "ContentType":
+        app_path = "nautobot.extras"
     # Serializers for Django's auth models are in the users app
-    if app_name == "auth":
-        app_name = "users"
-    serializer_name = f"{app_name}.api.serializers.{prefix}{model_name}Serializer"
-    if app_name not in settings.PLUGINS:
-        serializer_name = f"nautobot.{serializer_name}"
+    elif app_label == "auth":
+        app_path = "nautobot.users"
```
snaselj, line 112: I prefer not to use this kind of conditions in the code, as it needs future care. Implementation using `__module__` was more generic. Not sure, whether this works with e.g. `Group`:

```
invoke nbshell

In [1]: Group._meta.label.split('.')
Out[1]: ['auth', 'Group']

In [2]: Group.__module__
Out[2]: 'django.contrib.auth.models'
```

nautobot/extras/plugins/views.py
```diff
@@ -30,24 +30,25 @@ class InstalledPluginsView(AdminRequiredMixin, View):
...
-                    "version": plugin.version,
-                    "actions": {
-                        "home": plugin.home_view_name,
-                        "configure": plugin.config_view_name,
-                        "docs": plugin.docs_view_name,
-                    },
-                }
-            )
+        for plugin in apps.get_app_configs():
+            if plugin.name in settings.PLUGINS:
```
glennmatthews, line 35: I wonder if we should also add a `and issubclass(plugin, NautobotAppConfig)` as protection against errors, since most of the below attributes aren't present in the base `AppConfig` class?

whitej6: All apps should be in the plugins list and all apps should have the config a subclass of either `NautobotAppConfig` or `nautobot.extras.plugins.PluginConfig` (later is a subclass of the former with a deprecation warning). Not opposed to adding what value does it have for the view? I would think the startup should error out if an app isn't a subclass of one of the above ancestors right?

glennmatthews: I haven't actually tried to see what happens if you have something in PLUGINS that's using AppConfig instead of NautobotAppConfig. :-) You may well be correct that it's not needed at this point in the code.

whitej6: @glennmatthews Just checked, so _technically_ there's no validation on the ancestor BUT it requires the following attrs that aren't part of the the base `AppConfig` so it throws an `PluginImproperlyConfigured` exception on startup. I think that there's merit in doing validation before [this line](https://github.com/nautobot/nautobot/blob/512b1ee9062638ea726173abc910fc6bbc68bcfb/nautobot/extras/plugins/utils.py#L74)

* `config`
* `min_version`
* `max_version`
* `default_settings`
* `installed_apps`
* `middleware`
* `required_settings`

glennmatthews: Thanks for checking! Let's get an issue in the backlog to improve that validation; definitely doesn't need to be part of this PR.

glennmatthews (approved):

nautobot/extras/plugins/views.py
whitej6: @glennmatthews I opened #4880

Issue #4796, opened 2023-11-14
Add InterfaceRedundancyGroupAssociation (and priority) to GraphQL


milan2655:
### As ...

Ozzie - Operator

### I want ...

enhance the graphql schema, because `interface priority` is not exposed via graphql on the Interface Redundancy Group Association.

### So that ...

when `interface priority` is exposed via graphql , this value can be used in jinja2 templates.

### I know this is done when...

graphql is executed and the `interface priority` is missing.

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

glennmatthews:
Thanks for the report! Indeed, we need a pattern for exposing all of the various custom "through" tables in GraphQL:

- InterfaceRedundancyGroupAssociation
- IPAddressToInterface
- VRFDeviceAssignment
- VRFPrefixAssignment

milan2655:
without exposing `priority` value it is no possible to make intended config for a redundant interface.

lampwins:
Extending my thoughts from #4797, this is another good example of feature sets that make sense for these M2M through models.

lampwins:
We will increase the scope of this to include all M2M through models to the schema.

glennmatthews:
Added for 2.1 by #4861

PR #4861, opened 2023-11-27
Added InterfaceRedundancyGroupAssociation to GraphQL.

HanlinMiao:
# Closes: #4796 
# What's Changed

1. Added `InterfaceRedundancyGroupAssociation` to GraphQL.
2. Added try...except... block to handle the `AttributeError` raised by `get_absolute_url()` when a valid absolute url is not found.

- [x] I should open a PR against develop as well for the second item.
- [x] IPAddressToInterface
- [x] VRFDeviceAssignment
- [x] VRFPrefixAssignment

Screen Shot 2023-11-27 at 2 50 29 PM
Screen Shot 2023-11-27 at 6 19 19 PM

# TODO

- [ ] Unit, Integration Tests


changes/4796.added
```diff
@@ -0,0 +1,2 @@
+Added InterfaceRedundancyGroupAssociation to GraphQL.
+Added try...except... block to handle the `AttributeError` raised by `get_absolute_url()` when a valid absolute url is not found.
```
glennmatthews, line 2: Not sure I see the connection between these two items. Since as you mentioned the try/except logic may be needed in `develop`, is there a reason it shouldn't just be part of a separate PR that can bubble up to `next` in due time?

nautobot/dcim/models/device_components.py
```diff
@@ -849,7 +849,9 @@ def remove_interface(self, interface):
         )
         return instance.delete()
 
-
+@extras_features(
+    "graphql"
+)
```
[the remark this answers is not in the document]
glennmatthews: This seems correct for what it is (allowing this model to be queried directly in GraphQL). Wonder if it's sufficient though - we probably also need to make sure there's a pattern in place for traversing in GraphQL from an Interface to its InterfaceRedundancyGroupAssociations, from an InterfaceRedundancyGroup to its InterfaceRedundancyGroupAssociations, and vice versa - may require some custom GraphQL types, or might not.

nautobot/dcim/views.py
```diff
@@ -2910,3 +2911,15 @@ class InterfaceRedundancyGroupAssociationUIViewSet(ObjectEditViewMixin, ObjectDe
     form_class = forms.InterfaceRedundancyGroupAssociationForm
     template_name = "dcim/interfaceredundancygroupassociation_create.html"
     lookup_field = "pk"
+
+    def get_return_url(self, request, obj=None):
+        # First, see if `return_url` was specified as a query parameter or form data. Use this URL only if it's
+        # considered safe.
```
glennmatthews, line 2917: What's this about? What case are we protecting against?

nautobot/dcim/models/device_components.py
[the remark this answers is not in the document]
HanlinMiao: It seems good as far as traversal goes.
Screen Shot 2023-11-27 at 6 05 11 PM
Screen Shot 2023-11-27 at 6 07 27 PM

changes/4796.added
HanlinMiao: Yeah confirmed the error in develop as well, will just open a PR in `develop`

nautobot/dcim/views.py
HanlinMiao: That was for debugging purposes, removed now.

glennmatthews (commented):
Changes look reasonable, anything we can add in unit tests?

glennmatthews (approved):

Issue #4787, opened 2023-11-13
Remove `HIDE_RESTRICTED_UI` toggle (make default and always-on)


glennmatthews:
### As ...

Patti - Platform Admin

### I want ...

Nautobot to always require login by unauthenticated users, without even presenting an option for unauthenticated users to access any Nautobot content.

### So that ...

Nautobot is secure by default, rather than providing unsecure options.

### I know this is done when...

- `HIDE_RESTRICTED_UI` setting is removed  from all code, tests, and documentation
- Unauthenticated users can never access the REST API or GraphQL
- Unauthenticated users attempting to access the OpenAPI (Swagger) docs are always redirected to login
- Unauthenticated users attempting to access GraphiQL are always redirected to login
- Unauthenticated users attempting to access any Nautobot view are always redirected to login.

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [X] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

none

### External Dependencies

none

glennmatthews:
This will likely close the following issues if implemented:

- #1646 
- #3328 
- #3732

PR #4865, opened 2023-11-28
Removed support for `HIDE_RESTRICTED_UI` config.

timizuoebideri1:
# Closes: #4787 
# What's Changed

# TODO

- [ ] Unit, Integration Tests


changes/4787.removed
```diff
@@ -0,0 +1 @@
+Removed support for `HIDE_RESTRICTED_UI` config.
```
glennmatthews, line 1: Would be good to elaborate here - specifically make it clear whether the effect is 1) restricted UI elements are never hidden or 2) restricted UI elements are always hidden. (I believe it's the latter now?)

nautobot/core/api/views.py
```diff
@@ -721,14 +721,13 @@ class GetMenuAPIView(NautobotAPIVersionMixin, APIView):
...
 
-    def format_and_remove_hidden_menu(self, request, data, hide_restricted_ui):
+    def format_and_remove_hidden_menu(self, request, data, hide_restricted_ui=False):
         """
         Formats the menu data and removes hidden menu items based on user permissions.
 
         Args:
             request (HttpRequest): The request object.
             data (Union[dict, str]): The menu data to format and filter. Can be either a dictionary or a string.
-            hide_restricted_ui (bool): Flag indicating whether to hide restricted menu items.
```
glennmatthews, line 731: Let's discuss in standup - I don't know if we'd been thinking about the role of `HIDE_RESTRICTED_UI` in selectively disabling individual items based on user permissions; the focus of #4787 was really more around removing the ability for unauthenticated users to access the UI at all. We might want to keep `HIDE_RESTRICTED_UI` around but narrow its scope to only hiding/disabling UI elements for authenticated users based on their permissions?

changes/4787.removed
```diff
@@ -0,0 +1 @@
+Removed support for `HIDE_RESTRICTED_UI`, resulting in the inability of unauthorized users to view UI elements requiring specific permissions. Additionally, users not logged in will now be automatically redirected to the login page.
```
glennmatthews, line 1:
```suggestion
Removed support for `HIDE_RESTRICTED_UI`. UI elements requiring specific permissions will now always be hidden from users lacking those permissions. Additionally, users not logged in will now be automatically redirected to the login page.
```

This probably deserves a mention in the "Release Overview" section in `nautobot/docs/release-notes/version-2.1.md` as well?

nautobot/core/api/views.py
```diff
@@ -721,14 +721,13 @@ class GetMenuAPIView(NautobotAPIVersionMixin, APIView):
 
     permission_classes = [IsAuthenticated]
 
-    def format_and_remove_hidden_menu(self, request, data, hide_restricted_ui):
+    def format_and_remove_hidden_menu(self, request, data, hide_restricted_ui=False):
```
glennmatthews, line 724: Let's just remove the `hide_restricted_ui` parameter altogether - this isn't a documented/guaranteed-stable API as far as I know.

nautobot/core/api/views.py
```diff
@@ -747,15 +746,18 @@ def format_and_remove_hidden_menu(self, request, data, hide_restricted_ui):
             Output:
             {"Devices": "data value"}
         """
-        if isinstance(data, dict):
```
glennmatthews, line 750: Why is the `isinstance` check no longer needed? Is `data` now always required/guaranteed to be a dict? (If so, please update the docstring above accordingly.)

nautobot/core/tests/integration/test_plugin_home.py
```diff
@@ -143,3 +126,18 @@ def test_homepage_render_limit_permissions(self):
                 else:
                     item_html = columns_html.first.find_by_xpath(f".//h4[contains(text(), '{item_name}')]")
                     self.assertTrue("mdi mdi-lock" in item_html.find_by_xpath("./../span")["innerHTML"])
+
+    def test_examplemodel_custom_panel(self):
```
glennmatthews, line 130: Why was this relocated?

nautobot/core/tests/test_api.py
```diff
@@ -72,6 +76,9 @@ def setUp(self):
         self.cf_text.save()
 
     def test_api_docs(self):
+        user = User.objects.create_user(username="nautobotuser")
+        self.client.force_login(user)
```
glennmatthews, line 80:
```suggestion
        self.client.force_login(self.user)
```

nautobot/core/tests/test_api.py
```diff
@@ -810,6 +817,9 @@ def test_descending_sort(self):
 class NewUIGetMenuAPIViewTestCase(testing.APITestCase):
     def test_get_menu(self):
         """Asset response from new ui nav menu api returns a well formatted registry["new_ui_nav_menu"] expected by nautobot-ui."""
+        self.user.is_superuser = True
+        self.user.save()
+
         url = reverse("ui-api:get-menu")
         response = self.client.get(url, **self.header)
```
glennmatthews, line 824: Shouldn't this test case require the user to be logged in? Do we have a bug here?

nautobot/core/tests/test_api.py
```diff
@@ -912,6 +922,8 @@ def test_get_menu(self):
                 },
             },
         }
+
+        print(response.data)
```
glennmatthews, line 926: remove

nautobot/core/tests/test_views.py
```diff
@@ -230,48 +229,12 @@ def test_installed_apps_visible_to_staff_with_hide_restricted_ui_true(self):
...
-        )
-
-    @override_settings(HIDE_RESTRICTED_UI=True)
-    def test_installed_apps_not_visible_to_non_staff_user_with_hide_restricted_ui_true(self):
-        """The "Installed Apps" menu item should be hidden from a non-staff user when HIDE_RESTRICTED_UI=True."""
+    def test_installed_apps_not_visible_to_non_staff_user_without_permission(self):
+        """The "Installed Apps" menu item should be hidden from a non-staff user without permission."""
         response_content = self.make_request()
 
         self.assertNotRegex(response_content, r"Installed\s+Apps")
```
glennmatthews, line 236: Hm. Should this be `Installed\s+Plugins` instead?

nautobot/core/tests/test_views.py
```diff
@@ -309,46 +272,6 @@ def test_sso_login_button_visible(self):
...
-            redirect_chain = [(f"/login/?next={url}", 302)]
-            self.assertEqual(response.redirect_chain, redirect_chain)
-            response_content = response.content.decode(response.charset).replace("\n", "")
-            # Assert Footer items(`self.footer_elements`), Banner and Banner Top is hidden
-            for footer_text in self.footer_elements:
-                self.assertNotIn(footer_text, response_content)
-            # Only API Docs implements BANNERS
-            if url == urls[0]:
-                self.assertNotIn("Hello, Banner Top", response_content)
-                self.assertNotIn("Hello, Banner Bottom", response_content)
```
glennmatthews, line 330: Isn't this test case still valid/meaningful?

nautobot/core/api/views.py
timizuoebideri1: Because I set `the'return_value`' instead of recalling the function on line `754`, and that's the only place where data may be anything other than `dict`.

nautobot/core/tests/test_api.py
timizuoebideri1: User access token is provided in `self.header`

nautobot/core/tests/test_views.py
timizuoebideri1, line 236:
```suggestion
        self.assertNotRegex(response_content, r"Installed\s+Plugins")
```

nautobot/core/tests/test_api.py
timizuoebideri1: `APIDocsTestCase(TestCase)` inherits from Django TestCase, which do not have `self.user`

nautobot/core/api/views.py
```diff
@@ -721,14 +721,13 @@ class GetMenuAPIView(NautobotAPIVersionMixin, APIView):
...
         """
         Formats the menu data and removes hidden menu items based on user permissions.
 
         Args:
             request (HttpRequest): The request object.
             data (Union[dict, str]): The menu data to format and filter. Can be either a dictionary or a string.
-            hide_restricted_ui (bool): Flag indicating whether to hide restricted menu items.
 
         Returns:
             (Union[dict, str]): The formatted menu data without hidden items. Returns a dict if `data` is a
```
glennmatthews, line 733: If this now only returns a dict, please update this accordingly.

nautobot/core/api/views.py
```diff
@@ -721,14 +721,13 @@ class GetMenuAPIView(NautobotAPIVersionMixin, APIView):
...
     permission_classes = [IsAuthenticated]
 
-    def format_and_remove_hidden_menu(self, request, data, hide_restricted_ui):
+    def format_and_remove_hidden_menu(self, request, data):
         """
         Formats the menu data and removes hidden menu items based on user permissions.
 
         Args:
             request (HttpRequest): The request object.
             data (Union[dict, str]): The menu data to format and filter. Can be either a dictionary or a string.
```
glennmatthews, line 730: If this now only accepts a dict, please update this accordingly.

nautobot/core/tests/integration/test_plugin_home.py
```diff
@@ -142,4 +140,8 @@ def test_homepage_render_limit_permissions(self):
                         self.assertEqual(counter, counter_html)
                 else:
                     item_html = columns_html.first.find_by_xpath(f".//h4[contains(text(), '{item_name}')]")
-                    self.assertTrue("mdi mdi-lock" in item_html.find_by_xpath("./../span")["innerHTML"])
+                    # Assert Layout items without permission is not visible in DOM
+                    self.assertFalse(item_html)
+
+
+# secret
```
glennmatthews, line 147: ?

nautobot/core/api/views.py
timizuoebideri1, line 730:
```suggestion
            data (dict): The menu data to format and filter.
```

nautobot/core/api/views.py
timizuoebideri1, line 733:
```suggestion
            (dict): The formatted menu data without hidden items.
```

glennmatthews (approved):

HanlinMiao (approved):

Issue #4805, opened 2023-11-15
The Jobs Developer Guide is ambiguous when it comes to register_jobs()


dylanbob:
### Environment
* Nautobot version (Docker tag too if applicable): documentation of the v2.0.4
* Python version: n/a
* Database platform, version: n/a
* Middleware(s): n/a

### Steps to Reproduce
1. Head to [the jobs developer guide](https://docs.nautobot.com/projects/core/en/latest/development/jobs/)
2. Discover how custom jobs are developed and what changed since earlier nautobot versions
3. Be mislead by the documentation and fail to develop a job 👎

### Expected Behavior
The documentation should :
1. clearly state how, when and where register_jobs() must be used
2. match the example with the explanatory text : either both `jobs.py` and `jobs/` options should be explained and shown with an example, or only one of both (in that case, the `jobs/` seems better)
3. be written from the 2.X standpoint, but a more general issue, [#4760](https://github.com/nautobot/nautobot/issues/4760), already covers that

### Observed Behavior
In the documentation :
- the only occurence of register_jobs() is inside a "Changed in version 2.0.0" block, and it mixes in developing jobs and developing plugins providing jobs
- the example of register_jobs() does not match the documentation text : while the example is for `jobs.py`, the text is for `jobs/`, which changes where the register_jobs() should be put

### Additional information
register_jobs() is now mandatory, therefore its documentation should be crystal clear.

See my [corresponding message](https://networktocode.slack.com/archives/C01NWPK6WHL/p1699957240955239) in the Slack.

dylanbob:
This issue is also somewhat related to [#2727](https://github.com/nautobot/nautobot/issues/2727).

PR #4872, opened 2023-11-28
Update docs for jobs registration

gsnider2195:
# Closes: #4805 
# What's Changed

I also found that `JobSourceChoices` is no longer being used so I removed it. `JOBS_ROOT` no longer requires a `__init__.py` so I got rid of that in the docs and the code.

TODO (future?): condense some of these docs into a single file for developers with links to the development docs in the user docs instead of copy/pasting the same thing everywhere


glennmatthews (approved):
LGTM. Thanks for taking this on!

HanlinMiao (approved):

PR #4883, opened 2023-11-30
IPAddress default namespace

snaselj:
# Closes: NaN

## What's Changed

- Assign default namespace to `IPAddress`, when no namespace is provided.


snaselj:
Without this fix, I was not able to create `IPAddress` instance without the namespace defined.

cc.: @whitej6

glennmatthews:
Are you trying to create an IPAddress without specifying either a `namespace` or a `parent` as context? I'm not entirely convinced that's a desirable workflow to permit. Can you provide some more explanation as to why this is needed?

snaselj:
Seems like expected behavior is to raise an exception `Either a parent or a namespace must be provided.`.

However, this exception is not raised when using `IPAddress.save()`.

Probably not necessary to do anything with that, closing this PR.

glennmatthews:
Reopening this as it's been pointed out that `Prefix` creation automatically does a `get_default_namespace()` call if none is provided, so it's odd/inconsistent that `IPAddress` creation doesn't do the same.

gsnider2195:
> > > Reopening this as it's been pointed out that `Prefix` creation automatically does a `get_default_namespace()` call if none is provided, so it's odd/inconsistent that `IPAddress` creation doesn't do the same.
> > 
> > 
> > An IPAddress can't be created without a valid parent. This would be similar to setting a default form value on a required ForeignKey field and I don't think we do that anywhere else?
> 
> We do this already on `Prefix` and on `VRF`:
> 
> ```python
>     namespace = models.ForeignKey(
>         to="ipam.Namespace",
>         on_delete=models.PROTECT,
>         related_name="prefixes",
>         default=get_default_namespace_pk,    <----------<<
>     )
> ```
> 
> ```python
>     namespace = models.ForeignKey(
>         "ipam.Namespace",
>         on_delete=models.PROTECT,
>         related_name="vrfs",
>         default=get_default_namespace_pk,   <---------<<
>     )
> ```
> 
> It's also a usability issue for Jobs and Apps that are written for a single-namespace environment, having to either explicitly find the parent Prefix before creating the IP or else having to explicitly call `get_default_namespace()` on every IP creation.

Makes sense. We'll need to do a `get_closest_parent` for the namespace that's returned by `get_default_namespace` and if a suitable parent isn't found it will still fail.

PR #4906, opened 2023-12-03
fix: devcontainer

erjac77:
# Closes: N/A
# What's Changed

The `devcontainer.json` file was updated to fix the `devcontainer` workflow. I also took the liberty of adding an `.editorconfig` file in order to maintain consistent coding styles between different editors and IDEs (since I saw some discrepancies between files). I did not reformat the whole code base to keep this PR minimal. Some tweaks will certainly be necesssary.

The setup is not perfect, but it's usable. I had some difficulties (or we can say unexpected behavior) with the integration of certain tools with the devcontainer.

**isort**

* I had to comment out `source.organizeImports` on code save, since changes were made.

**pre-commit**

* I was not able to integrate the pre-commit script (`scripts/git-hooks/pre-commit`) in the `.devcontainer/postCreateCommand.sh`, since the pre-commit hook attempts to start conflicting containers.

**pylint**

* I had to disable `django-not-configured` error. This must be related to how nautobot loads its settings.
* "astroid-error" message are thrown with some models. Not sure what to do with this one.

Other possible improvements:
* [Add non-root user to a container](https://code.visualstudio.com/remote/advancedcontainers/add-nonroot-user)
* [Persist bash history](https://code.visualstudio.com/remote/advancedcontainers/persist-bash-history)

# TODO

- [ ] Unit, Integration Tests


development/docker-compose.yml
```diff
@@ -5,6 +5,7 @@ services:
     build:
       args:
         PYTHON_VER: "${PYTHON_VER}"
+        TARGETARCH: "${TARGETARCH}"
```
glennmatthews, line 8: Would have thought that `TARGETARCH` would be auto-detected by Docker - doesn't seem like something that we should require the user to specify?

erjac77: You're right. But I think I'm hitting this [bug](https://github.com/devcontainers/cli/issues/275). It seems that the automatic platform ARGs in the global scope are not recognized / detected by the `devcontainer`.

```bash
[2023-12-11T14:18:23.991Z] Start: Run: docker inspect --type image system-dev-dependencies-
[2023-12-11T14:18:24.064Z] Stop (73 ms): Run: docker inspect --type image system-dev-dependencies-
[2023-12-11T14:18:24.064Z] Path 'library/system-dev-dependencies-' for input 'docker.io/library/system-dev-dependencies-' failed validation.  Expected path to match regex '/^[a-z0-9]+([._-][a-z0-9]+)*(\/[a-z0-9]+([._-][a-z0-9]+)*)*$/'.
[2023-12-11T14:18:24.064Z] Error fetching image details: Could not parse image name 'system-dev-dependencies-'
[2023-12-11T14:18:24.065Z] Start: Run: docker pull system-dev-dependencies-
[2023-12-11T14:18:24.132Z] invalid reference format
[2023-12-11T14:18:24.133Z] Stop (68 ms): Run: docker pull system-dev-dependencies-
[2023-12-11T14:18:24.134Z] []
[2023-12-11T14:18:24.134Z] Error response from daemon: invalid reference format
```

The only workaround I found is to manually specified the environment variable.

.devcontainer/devcontainer.json
```diff
@@ -1,19 +1,79 @@
 {
-    "dockerComposeFile": ["../development/docker-compose.yml", "../development/docker-compose.debug.yml"],
+    "name": "Nautobot Dev Container",
+    "dockerComposeFile": [
+        "../development/docker-compose.yml",
+        "../development/docker-compose.postgres.yml",
+        "../development/docker-compose.dev.yml",
+        "../development/docker-compose.override.yml"
```
glennmatthews, line 7: Does this make it so that a `docker-compose.override.yml` *must* be provided by the user? Or will this still work if no such file is provided?

.devcontainer/devcontainer.json
```diff
@@ -1,19 +1,79 @@
...
+                "isort.args": [
+                    "--settings-file",
+                    "${workspaceFolder}/pyproject.toml"
+                ],
+                "isort.importStrategy": "fromEnvironment",
+                "pylint.args": [
+                    "--rcfile=${workspaceFolder}/pyproject.toml",
+                    "--disable=django-not-configured"
+                ],
+                "pylint.importStrategy": "fromEnvironment"
```
glennmatthews, line 73: Pylint is tricky because `pylint-django` only works correctly when DJANGO_SETTINGS_MODULE is configured. Might be better to just leave pylint out of the devcontainer setup for now and revisit later?

.devcontainer/devcontainer.json
erjac77: It will still work without this file. I use custom `docker-compose.override.yml` and `override.env` files to automatically bootstrap Nautobot with a user and token, as explained [here](https://docs.nautobot.com/projects/core/en/stable/development/core/docker-compose-advanced-use-cases/#automatically-creating-a-superuser). I can remove them if you think they are unnecessary.

.devcontainer/devcontainer.json
erjac77: Ok, I'll leave `pylint` out for now.

.devcontainer/devcontainer.json
erjac77: Is there any reason why you don't want superuser information in `dev.env`? For development purpose, I find incredibly useful to automatically start Nautobot with a superuser, so I don't have to worry recreating it every time my setup is destroyed/rebuilt.

development/docker-compose.yml
erjac77: CI was failing, so instead of a build argument, I simply specified a default value in the dockerfile.
See: https://github.com/nautobot/nautobot/blob/20f7bf9e624d1093af996f673eedc8b97e171b05/docker/Dockerfile#L4
Tell me if it's ok with you.

.devcontainer/devcontainer.json
gsnider2195: I cloned a clean copy of this branch and the devcontainer fails to start with this error:

```
[752829 ms] Start: Run: docker ps -q -a --filter label=devcontainer.local_folder=/home/gary/github/nautobot_devcontainer --filter label=devcontainer.config_file=/home/gary/github/nautobot_devcontainer/.devcontainer/devcontainer.json
[752857 ms] Start: Run: docker ps -q -a --filter label=devcontainer.local_folder=/home/gary/github/nautobot_devcontainer
[752882 ms] Resolving Feature dependencies for 'ghcr.io/devcontainers/features/docker-outside-of-docker:1'...
[752882 ms] * Processing feature: ghcr.io/devcontainers/features/docker-outside-of-docker:1
[753107 ms] Start: Run: docker-credential-secret get
[753606 ms] * Processing feature: ghcr.io/devcontainers/features/common-utils
[753853 ms] Soft-dependency 'ghcr.io/devcontainers/features/common-utils' is not required.  Removing from installation order...
[753856 ms] * Fetching feature: docker-outside-of-docker_0_oci
[754196 ms] * Fetched feature: docker-outside-of-docker_0_oci version 1.3.1
[754200 ms] Start: Run: docker-compose version --short
[754206 ms] Start: Run: docker compose version --short
[754318 ms] Docker Compose version: 2.21.0
[754318 ms] Start: Run: docker compose -f /home/gary/github/nautobot_devcontainer/development/docker-compose.yml -f /home/gary/github/nautobot_devcontainer/development/docker-compose.postgres.yml -f /home/gary/github/nautobot_devcontainer/development/docker-compose.dev.yml -f /home/gary/github/nautobot_devcontainer/development/docker-compose.override.yml --profile * config
[754433 ms] 
[754434 ms] stat /home/gary/github/nautobot_devcontainer/development/docker-compose.override.yml: no such file or directory

[754434 ms] Exit code 14
[754434 ms] Error: Command failed: docker compose -f /home/gary/github/nautobot_devcontainer/development/docker-compose.yml -f /home/gary/github/nautobot_devcontainer/development/docker-compose.postgres.yml -f /home/gary/github/nautobot_devcontainer/development/docker-compose.dev.yml -f /home/gary/github/nautobot_devcontainer/development/docker-compose.override.yml --profile * config
    at Wg (/home/gary/.vscode/extensions/ms-vscode-remote.remote-containers-0.327.0/dist/spec-node/devContainersSpecCLI.js:460:889)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async TJ (/home/gary/.vscode/extensions/ms-vscode-remote.remote-containers-0.327.0/dist/spec-node/devContainersSpecCLI.js:390:23529)
    at async _eA (/home/gary/.vscode/extensions/ms-vscode-remote.remote-containers-0.327.0/dist/spec-node/devContainersSpecCLI.js:612:35354)
[754442 ms] Exit code 1
```

so it looks like the `docker-compose.override.yml` is required

erjac77: @gsnider2195 Sorry, I was not clear enough. The devcontainer will start without the`docker-compose.override.yml` file, but you need to remove it from `.devcontainer/devcontainer.json`.

.devcontainer/devcontainer.json
```diff
@@ -1,19 +1,73 @@
...
+                "editor.formatOnSave": true,
+                // Python settings
+                "[python]": {
+                    "editor.codeActionsOnSave": {
+                        // "source.organizeImports": "explicit"
+                    },
+                    // When rebuilding the devcontainer, `Developer: Reload Window` must be
+                    // executed for the Black formatter to be recognized by the settings.
+                    // See: https://github.com/microsoft/vscode/issues/189839
+                    "editor.defaultFormatter": "ms-python.black-formatter"
```
gsnider2195, line 47: We actually just changed the formatter to use the ruff extension `charliermarsh.ruff`. Since I'm already testing this workflow for the review I can make any necessary changes to get the new formatter working unless you'd like to do that?

erjac77: Great! I really like `ruff`, I'll do the change.

gsnider2195: Thanks! While you're at it you should merge in the latest changes from upstream. There's a fix in the docker file that's needed after Debian made some changes last week

erjac77: Did you also replaced `flake8` and `isort` with `ruff`?

gsnider2195: We did. Sorry I forgot to mention that earlier

gsnider2195: #5055 is still open to replace isort but the plan is to merge that soon, perhaps tomorrow

erjac77: Ok, I pushed the changes. Let me know if it's ok with you.

I guess the only question left is what should I do with the `docker-compose.override.yml`? See my comment about providing Nautobot with a superuser.

gsnider2195: Let me discuss that with @glennmatthews tomorrow. I'm not opposed to adding the superuser environment variables to the development environment by default. 

I'll finish reviewing the changes tomorrow as well to try to get this merged.

glennmatthews: I have no objection to adding a superuser to the dev environment.

.devcontainer/devcontainer.json
```diff
@@ -1,19 +1,66 @@
...
+    "shutdownAction": "stopCompose",
+    "customizations": {
+        "vscode": {
+            "extensions": [
+                "charliermarsh.ruff",
+                "DavidAnson.vscode-markdownlint",
+                "dbaeumer.vscode-eslint",
+                "eamodio.gitlens",
+                "EditorConfig.EditorConfig",
+                "esbenp.prettier-vscode",
```
gsnider2195, line 22:
```suggestion
                "eamodio.gitlens",
                "EditorConfig.EditorConfig",
```
Since we removed the js frontend, prettier is complaining about not finding a package.json and eslint isn't needed anymore.

.devcontainer/devcontainer.json
erjac77: Thanks! Now, where do you think it makes more sense to add it? In `dev.env` or in `override.env` / `docker-compose.override.yml`?

development/.env.example
```diff
@@ -0,0 +1 @@
+PYTHON_VER=3.11
```
gsnider2195, line 1:
```suggestion
# Required for vscode Dev Containers support
PYTHON_VER=3.11
```

I tested using the `invoke build`/`invoke debug` container workflow while a `development/.env` file was present and this file does not seem to negatively impact that workflow.

erjac77: I'm curious, where do you put this variable usually?

.devcontainer/devcontainer.json
erjac77: I'll put the superuser information in `dev.env`. Keep the override for user specific needs...

gsnider2195: I think `dev.env`. We'll also need to update `nautobot/docs/development/core/docker-compose-advanced-use-cases.md` "Automatically Creating a Superuser" to state that the superuser is now automatically created and that you'll need to set `NAUTOBOT_CREATE_SUPERUSER=false` in `override.env` to disable it. I can help with the doc updates if you'd like.

erjac77: Yeah, my english is not very good, I prefer to let you update the doc. Thanks.

development/.env.example
gsnider2195: The default is set in tasks.py but can be overridden in `invoke.yml`

.devcontainer/devcontainer.json
erjac77: I think the changelog is also missing...

development/docker-compose.yml
gsnider2195: We'll need some docs for how to change this for arm64 users. There are 3 options that all require changing a git tracked file:

- update the default in `docker/Dockerfile`
- add an `ARCH: arm64` variable to the docker-compose.yml `services.nautobot.build.args`
- create a `docker-compose.override.yml` with the above build arg and add the `docker-compose.override.yml` to  your devcontainer.json `dockerComposeFile` list (preferred IMO)

development/.env.example
erjac77: Ok, I see. But it won't work with the devcontainer workflow. I tried setting the var in `remoteEnv` and `containerEnv`, but unfortunately it doesn't work.

development/docker-compose.yml
gsnider2195: With the change we're making to the superuser creation I don't think we should include the `docker-compose.override.yml` in the `devcontainer.json` file by default

.devcontainer/devcontainer.json
```diff
@@ -1,19 +1,62 @@
...
+                },
+                // Python settings
+                "[python]": {
+                    "editor.codeActionsOnSave": {
+                        "source.fixAll": "explicit",
+                        "source.organizeImports": "explicit"
+                    },
+                    // When rebuilding the devcontainer, `Developer: Reload Window` must be
+                    // executed for the Black formatter to be recognized by the settings.
+                    // See: https://github.com/microsoft/vscode/issues/189839
```
gsnider2195, line 48: Still required when using ruff? If so, update the comment to change `Black` to `Ruff`

development/.env.example
gsnider2195: Yeah I tried that too. I just wanted to make sure that this file (`development/.env`) wasn't going to break the other development workflows.

.devcontainer/devcontainer.json
gsnider2195: After this is fixed I'll update the docs and we should be good to merge

.devcontainer/devcontainer.json
gsnider2195: I can add that as well. We probably need more than one

.devcontainer/devcontainer.json
erjac77: Yes, still required. I'll update the comment.

development/docker-compose.yml
erjac77: Has been removed already.

.devcontainer/devcontainer.json
```diff
@@ -1,19 +1,62 @@
...
+            "extensions": [
+                "charliermarsh.ruff",
+                "DavidAnson.vscode-markdownlint",
+                "eamodio.gitlens",
+                "EditorConfig.EditorConfig",
+                "GitHub.vscode-pull-request-github",
+                "ms-python.python",
+                "ms-python.vscode-pylance",
+                "samuelcolvin.jinjahtml",
+                "tamasfe.even-better-toml"
```
glennmatthews, line 24: Just curious - should we add all of these extensions to `nautobot.code-workspace` as well?

.devcontainer/devcontainer.json
```diff
@@ -1,19 +1,62 @@
...
+                // Markdown settings
+                "[markdown]": {
+                    "editor.defaultFormatter": "DavidAnson.vscode-markdownlint"
+                },
+                // Python settings
+                "[python]": {
+                    "editor.codeActionsOnSave": {
+                        "source.fixAll": "explicit",
+                        "source.organizeImports": "explicit"
+                    },
```
glennmatthews, line 45: Same question on these, actually.

changes/4906.changed
glennmatthews, line 1: Wonder if both of these change fragments should be moved to a `4906.housekeeping` instead since they're not end-user-facing changes?

nautobot/docs/development/core/docker-compose-advanced-use-cases.md
```diff
@@ -76,23 +74,31 @@ services:
...
+```bash
+NAUTOBOT_SUPERUSER_PASSWORD=changeme
+NAUTOBOT_SUPERUSER_API_TOKEN=1111111111111111111111111111111111111111
+```
+
+This file can be used to add new environment variables or override default environment variables found in `development/dev.env` but it cannot override any environment variables defined in any docker compose yml files.
+
+### Automatically Creating a Superuser
+
++/- 2.0.7
```
glennmatthews, line 88: Probably at this point:

```suggestion
+/- 2.1.2
```

nautobot/docs/development/core/docker-compose-advanced-use-cases.md
```diff
@@ -144,11 +150,58 @@ Keycloak admin console is reachable via `http://localhost:8087/admin/` with user
...
 
-- `.devcontainers/devcontainer.json` - Dev. container definition
+- `.devcontainer/devcontainer.json` - Configuration for the [`Dev Containers`](https://code.visualstudio.com/docs/devcontainers/containers) extension
 - `nautobot.code-workspace` - VS Code workspace configuration for Nautobot
 - `development/docker-compose.debug.yml` - Docker Compose file with debug configuration for VS Code
 
-After opening the project directory in VS Code in a supported environment, you will be prompted by VS Code to **Reopen in Container** and **Open Workspace**. Select **Reopen in Container** to build and start the development containers. Once your window is connected to the container, you can open the workspace which enables support for Run/Debug.
+### Dev Containers Setup
+
++/- 2.0.7
```
glennmatthews, line 159:
```suggestion
+/- 2.1.2
```

nautobot/docs/development/core/docker-compose-advanced-use-cases.md
```diff
@@ -144,11 +150,58 @@ Keycloak admin console is reachable via `http://localhost:8087/admin/` with user
...
+
+#### ARM64 Build Argument
+
+Due to a bug in Dev Containers, when using VS Code to build your docker containers, docker cannot automatically determine the CPU architecture. To work around this bug, We have set a default value of `amd64` (x86_64). If you're running on another architecture like Apple Silicon or Raspberry Pi you will need to override this setting.
```
glennmatthews, line 163:
```suggestion
Due to a bug in Dev Containers, when using VS Code to build your docker containers, docker cannot automatically determine the CPU architecture. To work around this bug, we have set a default value of `amd64` (x86_64). If you're running on another architecture like Apple Silicon or Raspberry Pi you will need to override this setting.
```

nautobot/docs/development/core/getting-started.md
```diff
@@ -247,21 +249,24 @@ Available tasks:
 
 #### Using Docker with Invoke
 
++/- 2.0.7
```
glennmatthews, line 252:
```suggestion
+/- 2.1.2
```

nautobot/docs/development/core/getting-started.md
```diff
@@ -247,21 +249,24 @@ Available tasks:
...
+    The docker dev environment creates a superuser account by default with the username and password "admin". To disable this behavior, you must set the environment variable `NAUTOBOT_CREATE_SUPERUSER=false` using an [override file](./docker-compose-advanced-use-cases.md#docker-compose-overrides).
```
glennmatthews, line 253:
```suggestion
    The Docker dev environment creates a superuser account by default with the username and password "admin". To disable this behavior, you must set the environment variable `NAUTOBOT_CREATE_SUPERUSER=false` using an [override file](./docker-compose-advanced-use-cases.md#docker-compose-overrides).
```

nautobot/docs/development/core/getting-started.md
```diff
@@ -466,6 +471,9 @@ Below are common commands for working your development environment.
 
 ### Creating a Superuser
 
++/- 2.0.7
```
glennmatthews, line 474:
```suggestion
+/- 2.1.2
```

nautobot/docs/development/core/getting-started.md
```diff
@@ -466,6 +471,9 @@ Below are common commands for working your development environment.
...
+    The docker dev environment creates a superuser account by default with the username and password "admin". To disable this behavior, you must set the environment variable `NAUTOBOT_CREATE_SUPERUSER=false` using an [override file](./docker-compose-advanced-use-cases.md#docker-compose-overrides).
```
glennmatthews, line 475:
```suggestion
    The Docker dev environment creates a superuser account by default with the username and password "admin". To disable this behavior, you must set the environment variable `NAUTOBOT_CREATE_SUPERUSER=false` using an [override file](./docker-compose-advanced-use-cases.md#docker-compose-overrides).
```

.devcontainer/devcontainer.json
gsnider2195: I'd be ok with it however I haven't used `DavidAnson.vscode-markdownlint`, `EditorConfig.EditorConfig`, `GitHub.vscode-pull-request-github`, `samuelcolvin.jinjahtml` or `tamasfe.even-better-toml`. I don't think `ms-python.vscode-pylance` is required in `nautobot.code-workspace` since `ms-python.python` installs it automatically. It may be necessary here because this is the list of extensions that will be automatically installed in the container. The `nautobot.code-workspace` is just a list of recommended extensions.

.devcontainer/devcontainer.json
gsnider2195: I'm a little bit hesitant to add the settings to `nautobot.code-workspace` since workspace settings take precedence over user settings so it makes it more difficult to override these if you don't want them. However, I don't use the workspace file because I want to control my settings so maybe that is the right way to do it? https://code.visualstudio.com/docs/getstarted/settings#_settings-precedence

glennmatthews (approved):

Issue #2149, opened 2022-08-05
Homepage cards can be customized


chadell:
### As ...

Austin - Network Automation Engineer

### I want ...

to have the option to choose which family cards are showed in the Nautobot homepage, to provide a more relevant view of Nautobot status.

> Eventually, it would be great to also be able to add new cards from a plugin (it would go to a different issue if this makes sense) 

### So that ...

the Nautobot homepage would show only relevant information

### I know this is done when...

I have an option to customize the homepage via settings, ideally per user, but it could also work defining it globally in the nautobot_config.py

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

don't think so

glennmatthews:
Plugins can already add new homepage content (https://nautobot.readthedocs.io/en/stable/development/homepage/) and you can use the `HIDE_RESTRICTED_UI` config setting if you want content that a user doesn't have access to to be removed from the home page.

Is the request here to be able to remove/hide homepage content that core provides, and that the user has permissions to access?

chadell:
> Is the request here to be able to remove/hide homepage content that core provides, and that the user has permissions to access?

Yes, the request is to remove/hide content not limited to the permissions to access

glennmatthews:
Initial implementation in #4908. We can refine further in future.

PR #4892, opened 2023-12-01
Use custom template tag to resolve versioning of static files

timizuoebideri1:
# Closes: #4747 
# What's Changed

# TODO

- [ ] Explanation of Change(s)
- [ ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [ ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example Plugin Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #4908, opened 2023-12-03
UI/UX - Updates to homepage tiles

matt852:
# Closes: #2149
# What's Changed

Updates homepage panels to be more customizable on a per-user basis.
Enables each panel to be individually collapsable.  Settings retained via local user cookie.
Enables panels to be reordered via drag and drop.  Panel order retained via browser cache.
Introduces a short fade in of homepage tiles.

Example screenshots in light/dark mode of homepage tiles collapsed and in a user-customized order.
Screenshot 2023-12-03 at 11 07 52 AM
Screenshot 2023-12-03 at 11 08 00 AM

GIF of fade in effect when refreshing the homepage
homepage-fade

# TODO

- [ ] Unit, Integration Tests


matt852:
> This is really cool! We're trying not to introduce any new jquery to make it easier when we cut over to a new UI. Do you think it would be possible to rewrite this to vanilla JS?

Collapsable panels are updated to raw JavaScript.  The draggable part will take me a little longer to figure out.

matt852:
@gsnider2195 Draggable panels have been converted to pure JavaScript.
One thing I noticed is in dark mode, the "ghost" image of the panel being dragged reverts to the default light theme panel.  I don't see this as an issue, and if anything was a nice contrast highlighting the panel being moved, but did want to point that out.
Everything else seems good.  I added in the fade in to the panels in JS after the new order is loaded to prevent a popping effect.

Later on, I'd like to add the ability to remove panels from view via a little "x" icon, and potentially add saving the panel layout to the database on a per-user basis, instead of on a local-browser basis.

LMK thoughts and if anything else needs changed.

glennmatthews:
Storing/retrieving user settings in the DB can use the `/api/users/config/` API endpoint, just as a note.

lampwins:
@matt852 I like this, but to make sure we can wrap it up and get it into a release, can you define the remaining scope and the time to achieve that? I'm tempted to say let's draw the line here and add user persistence later?

matt852:
> @matt852 I like this, but to make sure we can wrap it up and get it into a release, can you define the remaining scope and the time to achieve that? I'm tempted to say let's draw the line here and add user persistence later?

@lampwins Thanks.  I agree on user persistence later as that's a whole other beast.

Remaining scope is your approval and adding integration tests.  I haven't started on the tests yet, but if you have a target date I'll work to get them in by then.  The collapsible panels shouldn't be too bad to test, but the draggable panels may be a bit more challenging.

nautobot/core/templates/home.html
```diff
@@ -23,87 +23,95 @@
...
+                        {% endif %}
+                    {% endfor %}
+                </div>
             </div>
         </div>
     </div>
 {% endblock %}
+{% block javascript %}
+{{ block.super }}
+{% include 'inc/homepage_javascript.html' %}
```
glennmatthews, line 116: Using the new templatetag added in #4892, and avoiding the need for a separate `inc/` file that's only used once:

```suggestion
<script src="{% versioned_static 'js/homepage_layout.js' %}"
        onerror="window.location='{% url 'media_failure' %}?filename=js/homepagelayout.js'"></script>
```

nautobot/core/templates/inc/homepage_javascript.html
glennmatthews, line 1: Can remove this file if implementing the previous comment.

nautobot/core/templates/inc/javascript.html
```diff
@@ -18,6 +18,7 @@
         onerror="window.location='{% url 'media_failure' %}?filename=js/forms.js'"></script>
 <script src="{% static 'js/theme.js' %}"
         onerror="window.location='{% url 'media_failure' %}?filename=js/theme.js'"></script>
+</script>
```
glennmatthews, line 21: Extraneous/incorrect addition?

nautobot/project-static/js/homepage_layout.js
```diff
@@ -0,0 +1,130 @@
...
+        // Remove 'toggle-' in the ID to get the localStorage key the toggle btn references
+        elementId = elementId.replace("toggle-", "");
+        var collapsibleDiv = document.getElementById(elementId);
+
+        // Toggle the collapsed class
+        if (collapsibleDiv.classList.contains('collapsed')) {
+            collapsibleDiv.classList.remove('collapsed');
+        } else {
+            collapsibleDiv.classList.add('collapsed');
+        }
```
glennmatthews, line 14: I think you could use `isCollapsed = collapsibleDiv.classList.toggle('collapsed')` here as a shorter alternative?

nautobot/project-static/js/homepage_layout.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    // Function to swap two elements
+    function swapElements(obj1, obj2) {
+        var temp = document.createElement("div");
+        obj2.parentNode.insertBefore(temp, obj2);
+        obj1.parentNode.insertBefore(obj2, obj1);
+        temp.parentNode.insertBefore(obj1, temp);
+        temp.parentNode.removeChild(temp);
+    }
```
glennmatthews, line 84: It "feels" a bit weird to me that drag-and-drop works as a "swap" operation. I'd have expected it to be more of a "move" or "insert" operation, i.e the dragged element gets inserted before/after the element it's dropped onto, not that the two elements exchange positions. Maybe just me?

nautobot/project-static/css/base.css
```diff
@@ -703,3 +703,24 @@ td span.hover_copy:hover .hover_copy_button {
...
+
+/* Collapse icon on homepage panels */
+.collapse-icon {
+    float: right;
+    cursor: pointer !important;
+}
+
+/* Draggable homepage panels fade in effect on page load */
+#draggable-homepage-panels {
+    opacity: 0;
```
glennmatthews, line 715: This means these elements will start as and remain invisible if JS is disabled or if there's an error in loading `homepage_layout.js`, right? Not saying that's a 100% no-go but it does concern me.

nautobot/core/templates/home.html
```diff
@@ -23,87 +23,95 @@
...
-                                <strong>{{ panel_name }}</strong>
-                            </div>
-                            <div class="list-group">
-                                {% if  panel_details.rendered_html %}
-                                    {% autoescape off %}{{ panel_details.rendered_html }}{% endautoescape %}
-                                {% elif panel_details.items %}
-                                    {% for item_name, item_details in panel_details.items.items %}
-                                        {% if  item_details.rendered_html %}
-                                            <div class="list-group-item" data-item-weight="{{ item_details.weight }}">
+                <div id="draggable-homepage-panels">
```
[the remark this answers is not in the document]
matt852: Yup this works.  Updated

nautobot/core/templates/home.html
```diff
@@ -23,87 +23,95 @@
...
+                    {% for panel_name, panel_details in registry.homepage_layout.panels.items %}
+                        {% if request.user|has_one_or_more_perms:panel_details.permissions %}
+                            <div class="panel panel-default" id="homepanel-list-panel{{ forloop.counter }}" style="break-inside: avoid" data-panel-weight="{{ panel_details.weight }}" draggable="true">
```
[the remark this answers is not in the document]
matt852: This is a much better idea.  I think I was using a `for` loop earlier in my JS to identify the panels by increments, but aren't using it anymore.  I updated it to slugify.

nautobot/core/templates/home.html
matt852: Done.  Thanks I didn't know this was possible.

nautobot/core/templates/inc/homepage_javascript.html
matt852: Done

nautobot/core/templates/inc/javascript.html
matt852: 🤦  yup

nautobot/project-static/js/homepage_layout.js
matt852: Yup!

nautobot/project-static/css/base.css
matt852: Those are good points.  The reason for this is since JS is loaded client side, the panels load in their initial state, then immediately "pop" into the custom ordering (similar to that dark mode flash of white a while ago).

This and the fade in hide the panels briefly to mask the popping effect.  I will need to think on another way to handle this.  Any ideas?

nautobot/project-static/js/homepage_layout.js
matt852: I found other ways using jQuery, but this was a common way I found online when looking this up.  Functionally it does what I need, so I'm a little confused.  Is this way a concern, or do you know of a better way?  I'm not as experienced with JS and didn't find many other ways to do so online, so guide me Obi Wan Kenobi.

nautobot/project-static/css/base.css
matt852: I found a way using <noscript> tags on the homepage.  It is working well in my local testing.  Does this seem like a viable solution?

nautobot/project-static/js/homepage_layout.js
glennmatthews: My concern wasn't with the code but with the behavior - it feels off to me that the element you drop onto also gets relocated, I would have expected just the dragged element would be relocated (other elements would shift to make room for it, but the order of the other elements wouldn't change).

Maybe replace this function with something as simple as like `draggedElement.parentNode.insertBefore(targetElement)` perhaps - not tested, I'll see if I can make time to try it out.

matt852: I actually tried that this morning and it worked, but I think the behavior wasn't exactly the same.  I will test it again later today and let you know the results.

glennmatthews: I was hoping for different behavior, in case that wasn't clear. :smile: Just my opinion, of course!

nautobot/project-static/js/homepage_layout.js
```diff
@@ -0,0 +1,126 @@
...
+    function loadSavedPanelOrder() {
+        var savedOrder = localStorage.getItem('homepage-panels-order');
+
+        if (savedOrder) {
+            savedOrder = JSON.parse(savedOrder);
+
+            // Append the panels in the saved order
+            for (var i = 0; i < savedOrder.length; i++) {
+                var panel = document.getElementById(savedOrder[i]);
+                draggableHomepagePanels.appendChild(panel);
```
glennmatthews, line 100:
```suggestion
                if (panel) {
                    draggableHomepagePanels.appendChild(panel);
                }
```

glennmatthews (commented):
Try this alternative and see if you like it:

```js
    // Function to handle a drop event
    function handleDrop(e) {
        e.preventDefault();

        let dragged = document.getElementById(e.dataTransfer.getData("text/plain"));
        let insertBefore = null;

        /* Were we dropped onto another panel? */
        let droppedOn = e.target.closest(".panel");
        if (droppedOn) {
            /* Were we dropped in the top half or the bottom half of the target panel? */
            boundingClientRect = droppedOn.getBoundingClientRect();
            if (e.clientY < boundingClientRect.top + (boundingClientRect.height / 2)) {
                /* Top half - insert before that panel */
                insertBefore = droppedOn;
            } else {
                /* Bottom half - insert after that panel */
                insertBefore = droppedOn.nextSibling;
            }
        } else {
            /* We were dropped into empty space - find the closest panel by geometry */
            for (let child of this.children) {
                /* Are we in the correct column? */
                if (child.offsetLeft > e.offsetX) {
                    /* Found the first child that is too far to the right, so we insert before that child */
                    insertBefore = child;
                    break;
                } else if (child.offsetLeft + child.offsetWidth >= e.offsetX) {
                    /* Child is in the correct column */
                    if (child.offsetTop >= e.offsetY) {
                        /* Found the first child in this column we were dropped above, so we insert before that child */
                        insertBefore = child;
                        break;
                    }
                }
            }
        }

        if (insertBefore) {
            this.insertBefore(dragged, insertBefore);
        } else {
            /* Add to end of the list */
            this.append(dragged);
        }

        savePanelOrder();

        // Remove the "dragging" class from the panel that was dragged
        dragged.classList.remove("dragging");
    }
```

gsnider2195:
I think we should open an issue to iterate on this so that panels remain in their original column when you collapse panels to the left


https://github.com/nautobot/nautobot/assets/75227981/823a3f47-47aa-4d63-9771-83283fbf85e4

matt852:
@glennmatthews your suggestion works great and looks a lot cleaner.  I'll update the code with it.  Thank you.

glennmatthews:
> I think we should open an issue to iterate on this so that panels remain in their original column when you collapse panels to the left

Not disagreeing, just noting that currently the column wrapping is done automatically by the browser based on available space (no JS involved, just CSS) so this may be quite tricky to implement.

matt852:
@glennmatthews Since your last review, I added in a new JS function called `enablePanelDragging()`, set the panels to default to not drag, and some <noscript> styling to the top of home.html.
The idea is so if JS is disabled, the panels aren't draggable or collapsable.

Ideally I'd like to move the <noscript> settings to a separate no-js.css file but I'm not sure if you want a new CSS file created, so please let me know.

nautobot/project-static/js/homepage_layout.js
```diff
@@ -0,0 +1,170 @@
...
+    function enablePanelDragging() {
+        var panelOrder = Array.from(draggableHomepagePanels.children).map(function(panel) {
+            return panel.id;
+        });
+        for (var i = 0; i < panelOrder.length; i++) {
+            var panel = document.getElementById(panelOrder[i]);
+            if (panel) {
+                panel.draggable = true;
+            }
+        }
```
glennmatthews, line 146:
```suggestion
        for (let panel of draggableHomepagePanels.children) {
            panel.draggable = true;
        }
```

nautobot/project-static/js/homepage_layout.js
```diff
@@ -0,0 +1,170 @@
...
+
+    }
+
+    // Fade in the panels after they have been rearranged
+    var opacity = 0; 
+    var intervalID = 0; 
+    window.onload = fadeIn; 
+      
+    function fadeIn() { 
+        setInterval(show, 50);
```
glennmatthews, line 156:
```suggestion
        intervalID = setInterval(show, 50); 
```

glennmatthews (commented):
This is getting there! A couple of code suggestions, and please also let's document this feature in `nautobot/docs/release-notes/version-2.1.md` under `Release Overview -> Added`.

HanlinMiao:
Weird behavior I found after I dragged and dropped the panels a couple of times. Data sources panel's opacity is off I think 🤔Screen Shot 2023-12-15 at 12 51 57 PM

HanlinMiao:
> Weird behavior I found after I dragged and dropped the panels a couple of times. Data sources panel's opacity is off I think 🤔Screen Shot 2023-12-15 at 12 51 57 PM

@gsnider2195 found that we are missing a dragend_event (https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dragend_event) which might be the cause of the issue.
Also here is a screen-recording on how to trigger this behavior.

https://github.com/nautobot/nautobot/assets/46973263/6ecb0602-fa49-46c8-b9b0-d23376253683

gsnider2195:
> > Weird behavior I found after I dragged and dropped the panels a couple of times. Data sources panel's opacity is off I think 🤔Screen Shot 2023-12-15 at 12 51 57 PM
> 
> @gsnider2195 found that we are missing a dragend_event (https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dragend_event) which might be the cause of the issue. Also here is a screen-recording on how to trigger this behavior.
> 
>  Screen.Recording.2023-12-15.at.1.04.36.PM.mov

This happens when you cancel a drag event by either pressing escape or trying to drag the element somewhere invalid.

nautobot/project-static/js/homepage_layout.js
glennmatthews: This is needed so that the function will be stopped once the fade-in is completed. Otherwise it keeps running infinitely.

nautobot/project-static/js/homepage_layout.js
matt852: Marking as resolved after implementing Glenn's `handleDrop()` function rewrite

matt852:
> This is getting there! A couple of code suggestions, and please also let's document this feature in `nautobot/docs/release-notes/version-2.1.md` under `Release Overview -> Added`.

@glennmatthews Docs updated!

glennmatthews:
Added `dragend` handler.

glennmatthews (approved):

gsnider2195 (approved):

Issue #4757, opened 2023-11-06
Postgres 10: 2.0 Migration error: `cannot ALTER TABLE "extras_jobresult" because it has pending trigger events`


joewesch:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.3
* Python version: 3.9
* Database platform, version: PostgreSQL v10.17
* Middleware(s):

### Steps to Reproduce
1. Have JobResult objects in Nautobot 1.6.2
2. Upgrade to Nautobot 2.0.3
3. Run `nautobot-server post_upgrade`

### Expected Behavior
Migrations to complete successfully

### Observed Behavior
```
Applying extras.0087_job__celery_task_fields__dryrun_support...Traceback (most recent call last):
  File "/opt/nautobot/lib64/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
    psycopg2.errors.ObjectInUse: cannot ALTER TABLE "extras_jobresult" because it has pending trigger events
 
 
The above exception was the direct cause of the following exception:
 
Traceback (most recent call last):
  File "/opt/nautobot/bin/nautobot-server", line 8, in <module>
    sys.exit(main())
  File "/opt/nautobot/lib64/python3.9/site-packages/nautobot/core/cli/__init__.py", line 54, in main
    run_app(
  File "/opt/nautobot/lib64/python3.9/site-packages/nautobot/core/runner/runner.py", line 297, in run_app
    management.execute_from_command_line([runner_name, command] + command_args)
  File "/opt/nautobot/lib64/python3.9/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/opt/nautobot/lib64/python3.9/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/opt/nautobot/lib64/python3.9/site-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/opt/nautobot/lib64/python3.9/site-packages/django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
  File "/opt/nautobot/lib64/python3.9/site-packages/nautobot/core/management/commands/post_upgrade.py", line 100, in handle
    call_command(
  File "/opt/nautobot/lib64/python3.9/site-packages/django/core/management/__init__.py", line 181, in call_command
[...]

PR #4910, opened 2023-12-04
[2.1] Drop support for PostgreSQL versions before 12.0

glennmatthews:
# Closes: #4757 
# What's Changed

- Update documentation to indicate Postgres 12.0 is the minimum supported version for Nautobot 2.1
- Add a system check to enforce this requirement when running `post_upgrade` or `migrate`.

Testing by setting the minimum version to 14.0 temporarily (as our dev environment runs version 13 currently):

```
nautobot-nautobot-1  | Performing database migrations...
nautobot-nautobot-1  | SystemCheckError: System check identified some issues:
nautobot-nautobot-1  | 
nautobot-nautobot-1  | ERRORS:
nautobot-nautobot-1  | connections[default]: (nautobot.core.E006) PostgreSQL version less than 140000 (i.e. 14.0) is not supported by this version of Nautobot
nautobot-nautobot-1  | 	HINT: Detected version is 130002 (major version 13)
```

# TODO

- n/a Unit, Integration Tests

### DIFF
--- a/nautobot/core/management/commands/post_upgrade.py
+++ b/nautobot/core/management/commands/post_upgrade.py
@@ -99,6 +99,7 @@ class Command(BaseCommand):
             self.stdout.write("Performing database migrations...")
             call_command(
                 "migrate",
+                skip_checks=False,  # make sure Postgres version check and others are applied
                 interactive=False,
                 traceback=options["traceback"],
                 verbosity=options["verbosity"],
--- a/nautobot/core/templates/exceptions/programming_error.html
+++ b/nautobot/core/templates/exceptions/programming_error.html
@@ -12,6 +12,6 @@
     <p>
         <i class="mdi mdi-alert"></i> <strong>Unsupported PostgreSQL version</strong> - Ensure that PostgreSQL version 12.0 or higher is in use. You
         can check this by connecting to the database using Nautobot's credentials and issuing a query for
-        <code>SELECT VERSION()</code>.
+        <code>SELECT VERSION();</code>.
     </p>
 {% endblock %}
--- a/nautobot/core/templates/exceptions/programming_error.html
+++ b/nautobot/core/templates/exceptions/programming_error.html
@@ -10,7 +10,7 @@
         <code>nautobot-server migrate</code> from the command line.
     </p>
     <p>
-        <i class="mdi mdi-alert"></i> <strong>Unsupported PostgreSQL version</strong> - Ensure that PostgreSQL version 9.6 or higher is in use. You
+        <i class="mdi mdi-alert"></i> <strong>Unsupported PostgreSQL version</strong> - Ensure that PostgreSQL version 12.0 or higher is in use. You
         can check this by connecting to the database using Nautobot's credentials and issuing a query for
         <code>SELECT VERSION()</code>.
     </p>
--- a/nautobot/docs/release-notes/version-2.1.md
+++ b/nautobot/docs/release-notes/version-2.1.md
@@ -37,7 +37,11 @@ The Nautobot UI has been updated with a customized theme, giving it a brand new 
 
 ### Removed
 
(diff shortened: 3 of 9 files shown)


nautobot/core/management/commands/post_upgrade.py
```diff
@@ -99,6 +99,7 @@ def handle(self, *args, **options):
             self.stdout.write("Performing database migrations...")
             call_command(
                 "migrate",
+                skip_checks=False,  # make sure Postgres version check and others are applied
```
gsnider2195, line 102: Is this a noop? It looks like this is already the default for `migrate.py`

nautobot/core/templates/exceptions/programming_error.html
```diff
@@ -10,7 +10,7 @@
         <code>nautobot-server migrate</code> from the command line.
     </p>
     <p>
-        <i class="mdi mdi-alert"></i> <strong>Unsupported PostgreSQL version</strong> - Ensure that PostgreSQL version 9.6 or higher is in use. You
+        <i class="mdi mdi-alert"></i> <strong>Unsupported PostgreSQL version</strong> - Ensure that PostgreSQL version 12.0 or higher is in use. You
         can check this by connecting to the database using Nautobot's credentials and issuing a query for
         <code>SELECT VERSION()</code>.
```
gsnider2195, line 15: Since you're already here can you update this to the following? Postgres waits for you to enter a semicolon before finishing the query so if you type the command as written it looks like nothing happened.

```suggestion
        <code>SELECT VERSION();</code>.
```

nautobot/core/management/commands/post_upgrade.py
glennmatthews: No, it's needed. When you use `call_command` it sets `skip_checks=True` if it wasn't already explicitly set by the caller - I'm guessing this is to avoid, in the more common case, re-calling checks that were already called by the parent command? However, by default most management commands (such as our `post_upgrade`) do *not* call `database` checks, but `migrate` does, so I do want to tell it to run its checks.

https://github.com/django/django/blob/stable/3.2.x/django/core/management/__init__.py#L178-L179

nautobot/docs/release-notes/version-2.1.md
```diff
@@ -37,7 +37,11 @@ The Nautobot UI has been updated with a customized theme, giving it a brand new
 
 ### Removed
 
-#### Remove HIDE_RESTRICTED_UI toggle ([#4787](https://github.com/nautobot/nautobot/issues/4787))
+#### Drop Support for Legacy PostgreSQL Versions ([#4757](https://github.com/nautobot/nautobot/issues/4757))
+
+Support for versions of PostgreSQL prior to 12.0 has been removed as these versions are no longer maintained. The `nautobot-server migrate` or `nautobot-server post_upgrade` commands will now abort when detecting an unsupported PostgreSQL version.
```
gsnider2195, line 42: Perhaps?

```suggestion
Support for versions of PostgreSQL prior to 12.0 has been removed as these versions are no longer maintained and contain bugs that prevent migrations from running in certain scenarios. The `nautobot-server migrate` or `nautobot-server post_upgrade` commands will now abort when detecting an unsupported PostgreSQL version.
```

nautobot/core/management/commands/post_upgrade.py
gsnider2195: I see. Thanks for the clarification!

nautobot/core/templates/exceptions/programming_error.html
glennmatthews: I'm actually not sure when/if we actually encounter ProgrammingError in any real case, and maybe this block should be removed entirely now that we're enforcing the version at migration time? Might need to do some testing on what happens if I start up with a newer version of Nautobot without running `migrate` and see if I can trip this error.

Issue #4891, opened 2023-12-01
Nautobot API Issue: Multiple Prefix Query Returns a Single Prefix


gbs098:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.5
* Python version:3.11.6
* Database platform, version: PostgreSQL 15.5
* Middleware(s):

### Steps to Reproduce
1.Send a GET request to the Nautobot API to retrieve multiple IP prefixes using the following format:
```
GET /api/ipam/prefixes/?prefix=10.112.56.0%2F25&prefix=10.112.56.128%2F28&prefix=10.112.56.192%2F29
```
2. Tried from nautobot web
3. Go to the section ipam->prefixes
4.In this section, locate and click on the 'Advanced Filter' tab.
In the advanced filter, enter the following prefixes to filter the data:
After entering the prefixes, submit the filter/search request.

### Expected Behavior
The API should return all requested prefixes (10.112.56.0/25, 10.112.56.128/28, and 10.112.56.192/29).

### Observed Behavior
The API only returns one prefix (e.g., 10.112.56.0/25) instead of all the specified prefixes.

glennmatthews:
Looks like the `PrefixFilterSet.prefix` filter (and several others on this FilterSet) are defined as a `CharFilter` rather than a `MultiValueCharFilter`. See also #1730.

gsnider2195:
There are only 5 filters that are still single value CharFilters. Lets go ahead and fix those as part of this

`PrefixFilterSet.prefix`
`PrefixFilterSet.within`
`PrefixFilterSet.within_include`
`PrefixFilterSet.contains`
`IPAddressFilterSet.prefix`

itdependsnetworks:
Is there concrete logic that we can add to pylint-nautobot to find these?

PR #4916, opened 2023-12-05
Fixed Prefix API and UI incorrectly filtering

timizuoebideri1:
# Closes: #4891 
# What's Changed

# TODO

- [ ] Unit, Integration Tests


nautobot/ipam/filters.py
```diff
@@ -222,53 +222,31 @@ class Meta:
...
 
     def search_within(self, queryset, name, value):
-        value = value.strip()
-        if not value:
-            return queryset
-        try:
-            return queryset.net_contained(netaddr.IPNetwork(value))
-        except (AddrFormatError, ValueError):
-            return queryset.none()
+        prefixes = [value.strip() for value in value]
```
glennmatthews, line 229: Here and below, do we need to throw away values that strip to an empty string as we were doing in the old code? Or is that being handled somewhere else?

nautobot/ipam/models.py
```diff
@@ -530,23 +530,37 @@ def __init__(self, *args, **kwargs):
     def __str__(self):
         return str(self.prefix)
 
+    @staticmethod
+    def deconstruct_prefix(prefix):
```
glennmatthews, line 534: Why this refactor? I don't see anywhere else in the diffs that you're calling this new method, and it feels strange to me that you're taking what was otherwise a private method and making it public if it's not needed.

nautobot/core/tests/test_graphql.py
```diff
@@ -1449,8 +1449,8 @@ def test_query_ip_addresses_filter(self):
                 IPAddress.objects.net_host_contained("10.0.0.0/16").count(),
             ),
             (
-                'prefix: "10.0.2.0/24"',
-                IPAddress.objects.net_host_contained("10.0.2.0/24").count(),
+                'prefix: ["10.0.1.0/24", "10.0.2.0/24"]',
+                IPAddress.objects.net_host_contained("10.0.1.0/24", "10.0.2.0/24").count(),
             ),
```
glennmatthews, line 1454: If we can still support single-value filtering, let's keep the original test data here as well.

nautobot/ipam/querysets.py
```diff
@@ -191,46 +192,50 @@ def string_search(self, search):
...
-            broadcast__gte=last_ip,
-        )
+    def net_equals(self, *prefixes):
+        query = Q()
+        for prefix in prefixes:
+            with contextlib.suppress(netaddr.AddrFormatError, ValueError):
+                prefix = netaddr.IPNetwork(prefix)
+                last_ip = self._get_last_ip(prefix)
+                query |= Q(prefix_length=prefix.prefixlen, network=prefix.network, broadcast=last_ip)
+        return self.filter(query)
```
glennmatthews, line 202: Did this change the previous behavior where if any input is invalid we return `queryset.none()`? It looks like now if there's an exception raised we just skip that prefix and fail "open" instead?

nautobot/ipam/querysets.py
```diff
@@ -191,46 +192,50 @@ def string_search(self, search):
...
+
+    def net_contains_or_equals(self, *prefixes):
+        query = Q()
+        for prefix in prefixes:
+            with contextlib.suppress(netaddr.AddrFormatError, ValueError):
+                prefix = netaddr.IPNetwork(prefix)
+                last_ip = self._get_last_ip(prefix)
+                query |= Q(prefix_length__lte=prefix.prefixlen, network__lte=prefix.network, broadcast__gte=last_ip)
+        return self.filter(query)
```
gsnider2195, line 238: I think if we're going to update these queryset methods to accept multiple values they need to behave the same way as django. This method is always doing an OR on multiple arguments which is not how django filtering works. Instead of accepting multiple args we probably need to accept a Q object and use the deconstruct() method to pick apart the Q object and recreate it with the updated parameters.

gsnider2195: May need to discuss this with @glennmatthews

nautobot/ipam/querysets.py
```diff
@@ -191,46 +192,50 @@ def string_search(self, search):
...
-        last_ip = self._get_last_ip(prefix)
-        return self.filter(
-            prefix_length__lte=prefix.prefixlen,
-            network__lte=prefix.network,
-            broadcast__gte=last_ip,
-        )
+    def net_equals(self, *prefixes):
+        query = Q()
+        for prefix in prefixes:
+            with contextlib.suppress(netaddr.AddrFormatError, ValueError):
```
gsnider2195, line 198: I think this is too significant of a behavior change for a bug fix. Previously these methods would fail if you gave them invalid data.

nautobot/ipam/querysets.py
gsnider2195: We don't need to make these changes here

nautobot/core/tests/test_graphql.py
timizuoebideri1: Yup line 1448 uses single value filtering

nautobot/ipam/filters.py
```diff
@@ -383,14 +372,8 @@ def _has_interface_assignments(self, queryset, name, value):
...
-        value = value.strip()
-        if not value:
-            return queryset
-        try:
-            query = netaddr.IPNetwork(value.strip()).cidr
-            return queryset.net_host_contained(query)
-        except (AddrFormatError, ValueError):
-            return queryset.none()
+        ips = [value.strip() for value in value if value.strip()]
+        return queryset.net_host_contained(*ips)
```
gsnider2195, line 376:
```suggestion
        prefixes = [prefix.strip() for prefix in value if prefix.strip()]
        return queryset.net_host_contained(*prefixes)
```

glennmatthews (approved):

Issue #4917, opened 2023-12-05
`location_hierarchy` inc template renders slowly


Kircheneer:
### Environment
* Nautobot version: 1.6.6
* Python version: 3.8
* Database platform, version: PostgresQL version 16
* Middleware(s): -

### Steps to Reproduce
1. Have a big enough location hierarchy with lots of layers (regions + normal locations
2. Load for example the device detail page
3.

### Expected Behavior

The page loads reasonably quickly.

### Observed Behavior

The page takes 25s to load.

glennmatthews:
Fixed for 1.6 by #4918 - as mentioned in https://github.com/nautobot/nautobot/pull/4918#discussion_r1419547034 it would be good to forward-port part of the fix to `develop` for 2.x as well.

PR #4918, opened 2023-12-05
fixes slow performance on location hierarchy HTML template

Kircheneer:
# Closes: #4917, #4921
# What's Changed
- Employs variables in template to avoid re-calculation of location tree/region tree multiple times
- Avoid calculating the entire location tree via `Location.ancestors` when instead we can iterate over the foreign key relationships

# TODO
- [ ] Unit, Integration Tests

### DIFF
--- a/changes/4917.fixed
+++ b/changes/4917.fixed
@@ -1,0 +1,1 @@
+Fixed slow performance on location hierarchy HTMl template.
--- a/changes/4917.fixed
+++ b/changes/4917.fixed
@@ -1,1 +1,1 @@
-Fixed slow performance on location hierarchy HTMl template.
+Fixed slow performance on location hierarchy html template.
--- a/nautobot/dcim/models/locations.py
+++ b/nautobot/dcim/models/locations.py
@@ -237,7 +237,13 @@ class Location(TreeModel, StatusModel, PrimaryModel):
     @property
     def base_site(self):
         """The site that this Location belongs to, if any, or that its root ancestor belongs to, if any."""
-        return self.site or self.ancestors().first().site
+        if self.site:
+            return self.site
+        current_location = self
+        while current_location := current_location.parent:
+            if current_location.site:
+                return current_location.site
+        return None
 
     def validate_unique(self, exclude=None):
         # Check for a duplicate name on a Location with no parent.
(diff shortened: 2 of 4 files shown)


Kircheneer:
Some of this is also applicable for 2.0, there is still a duplicate `location.ancestors()` call there. If you want, I am happy to port the applicable changes over as well.

Kircheneer:
Update: Removed variable for region ancestors as
- it didn't work
- the queries are already optimized enough with MPTT

nautobot/dcim/models/locations.py
```diff
@@ -237,7 +237,13 @@ def to_csv(self):
     @property
     def base_site(self):
         """The site that this Location belongs to, if any, or that its root ancestor belongs to, if any."""
```
gsnider2195, line 239:
```suggestion
        """The site that this Location belongs to, if any, or that its nearest ancestor belongs to, if any."""
```

changes/4917.fixed
```diff
@@ -0,0 +1 @@
+Fixed slow performance on location hierarchy HTMl template.
```
gsnider2195, line 1:
```suggestion
Fixed slow performance on location hierarchy html template.
```

nautobot/dcim/templates/dcim/inc/location_hierarchy.html
```diff
@@ -12,32 +13,33 @@
...
     </style>
     <ul class="location-hierarchy">
-        {% for region in location.base_site.region.get_ancestors %}
+        {% for region in site.region.get_ancestors %}
             <li>{{ region|hyperlinked_object }} (Region)
                 <ul>
         {% endfor %}
-        {% if location.base_site.region %}
-            <li>{{ location.base_site.region|hyperlinked_object }} (Region)
+        {% if site %}
```
gsnider2195, line 20:
```suggestion
        {% if site and site.region %}
```

gsnider2195 (approved):

nautobot/dcim/templates/dcim/inc/location_hierarchy.html
```diff
@@ -1,5 +1,6 @@
 {% load helpers %}
 {% if location is not None %}
+{% with site=location.base_site ancestors=location.ancestors %}
```
glennmatthews, line 3: Although `base_site` and `region` are gone in 2.0, it looks like 2.0 could still benefit from `with ancestors=location.ancestors` to make it so that we would only generate that list once instead of twice. Would you be willing to open the equivalent `develop` PR?

glennmatthews (approved):

Issue #4854, opened 2023-11-27
Computed Fields with invalid slugs are not converted to valid keys during 2.0 migration


joewesch:
### Environment
* Nautobot version (Docker tag too if applicable): 1.6.4 -> 2.0.5
* Python version: 3.10.12
* Database platform, version: Postgres 14.9
* Middleware(s):

### Steps to Reproduce
1. On Nautobot 1.6.4 create a computed field with dashes in the slug instead of underscores (e.g. `bad-cpf`)
2. Upgrade Nautobot to 2.0
3. Try and load the GraphQL UI

### Expected Behavior
GraphQL UI loads

### Observed Behavior
```
AssertionError: Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but "cpf_bad-cpf" does not.
```

HanlinMiao:
It is the intended behavior of computed field slugs/keys. Changing the dash to underscore should fix the issue.
We should have enforced the pattern in the migrations but for now can you try `cpf.key.replace("-", "_")` and then `cpf.save()`

glennmatthews:
The issue here is at least twofold:

- Validation of the `key` for ComputedField, CustomField, and Relationship is only done in `clean()` so any direct ORM interaction (`ComputedField.objects.create()` or `ComputedField(...).save()`) can result in invalid keys being set. Possible fix here: move that enforcement from `clean()` to `save()`?
- When invalid keys are present, the GraphQL UI throws an exception and fails to load; it should instead log a warning/error message to syslog but not prevent the UI from loading at all.

lampwins:
Check if `get_or_create()` calls save.

PR #4920, opened 2023-12-05
Fixed GraphQL UI unable to load when some ComputedField instances have non-graphql-safe keys.

HanlinMiao:
# Closes: #4854
# What's Changed

Fixed GraphQL UI unable to load when some ComputedField instances have non-graphql-safe keys.
Moved `ComputedField` key string validation from clean() to save()

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/4854.changed
+++ b/changes/4854.changed
@@ -1,0 +1,1 @@
+Moved `ComputedField` key string validation from clean() to save()
--- a/nautobot/core/graphql/schema.py
+++ b/nautobot/core/graphql/schema.py
@@ -4,6 +4,7 @@ import logging
 
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
+from django.core.validators import ValidationError
 from django.db.models.fields.reverse_related import ManyToOneRel, OneToOneRel
 
 import graphene
@@ -43,6 +44,7 @@ from nautobot.extras.registry import registry
 from nautobot.extras.models import ComputedField, CustomField, Relationship
 from nautobot.extras.choices import CustomFieldTypeChoices, RelationshipSideChoices
 from nautobot.extras.graphql.types import TagType, DynamicGroupType
+from nautobot.extras.utils import check_if_key_is_graphql_safe
 from nautobot.ipam.graphql.types import IPAddressType, PrefixType
 from nautobot.virtualization.graphql.types import VirtualMachineType, VMInterfaceType
 
@@ -460,6 +462,15 @@ def generate_query_mixin():
                 # Find the model class based on the content type
                 ct = ContentType.objects.get(app_label=app_name, model=model_name)
                 model = ct.model_class()
+                if model == ComputedField:
+                    for i, cpf in enumerate(ComputedField.objects.all()):
+                        try:
+                            check_if_key_is_graphql_safe("Computed Field", cpf.key)
+                        except ValidationError:
+                            cpf.key = cpf.key.replace(" ", "_")
+                            cpf.key = cpf.key.replace("-", "_")
+                            cpf.key = f"a{i}" + cpf.key
+                            cpf.save()
             except ContentType.DoesNotExist:
                 logger.warning(
                     'Unable to generate a schema type for the model "%s.%s" in GraphQL, '
--- a/nautobot/extras/models/customfields.py
+++ b/nautobot/extras/models/customfields.py
@@ -112,8 +112,8 @@ class ComputedField(BaseModel, ChangeLoggedModel, NotesMixin):
             logger.warning("Failed to render computed field %s: %s", self.key, exc)
             return self.fallback_value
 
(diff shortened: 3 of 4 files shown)


nautobot/extras/models/customfields.py
```diff
@@ -112,8 +112,8 @@ def render(self, context):
             logger.warning("Failed to render computed field %s: %s", self.key, exc)
             return self.fallback_value
 
-    def clean(self):
-        super().clean()
+    def save(self, *args, **kwargs):
+        super().save(*args, **kwargs)
```
glennmatthews, line 116: similar change needed/desired for relationships and custom fields?

Also, probably want to reverse the order here - it makes sense to call super().clean() first, but super.save() should only be called *after* we do our own logic, I think?

nautobot/core/graphql/schema.py
```diff
@@ -460,6 +462,15 @@ def already_present(model):
...
                 model = ct.model_class()
+                if model == ComputedField:
+                    for i, cpf in enumerate(ComputedField.objects.all()):
+                        try:
+                            check_if_key_is_graphql_safe("Computed Field", cpf.key)
+                        except ValidationError:
+                            cpf.key = cpf.key.replace(" ", "_")
+                            cpf.key = cpf.key.replace("-", "_")
+                            cpf.key = f"a{i}" + cpf.key
+                            cpf.save()
```
glennmatthews, line 473: Risk of collisions with existing computed fields here when doing replacements? Probably would be better to skip over such fields instead of trying to fix them up.

Also, this seems like a weird place to do this check. I would have thought it'd be something we can/should do in `extend_schema_type_computed_field()` and similar functions.

changes/4854.changed
```diff
@@ -0,0 +1 @@
+Moved `ComputedField` key string validation from clean() to save()
```
glennmatthews, line 1:
```suggestion
Moved `ComputedField`, `CustomField`, and `Relationship` `key` string validation from `clean()` to `save()` to protect against creation of invalid data.
```

nautobot/core/graphql/schema.py
```diff
@@ -280,6 +282,17 @@ def extend_schema_type_computed_field(schema_type, model):
         field_name = f"{prefix}{field.key}"
         resolver_name = f"resolve_{field_name}"
 
+        try:
+            check_if_key_is_graphql_safe("Computed Field", field.key)
+        except ValidationError:
+            logger.warning(
+                'Unable to add the computed field "%s" to %s because ' 'computed field key "%s" is not GraphQL safe',
```
glennmatthews, line 289:
```suggestion
                'Unable to add the computed field "%s" to %s because computed field key "%s" is not GraphQL safe',
```

nautobot/core/graphql/schema.py
```diff
@@ -280,6 +282,17 @@ def extend_schema_type_computed_field(schema_type, model):
         field_name = f"{prefix}{field.key}"
         resolver_name = f"resolve_{field_name}"
 
+        try:
```
glennmatthews, line 285: Why not do this *before* the `field_name` and `resolver_name` calculation?

HanlinMiao:
I wonder if we should keep the validation in the `clean()` methods. It lets the validation error bubble up to the create/edit form in the UI, whereas validation in `save()` alone does not do that and it just throws a traceback. Is that acceptable to you guys @glennmatthews @gsnider2195 ?

changes/4854.changed
```diff
@@ -0,0 +1 @@
+Move invocation of `ComputedField`, `CustomField`, and `Relationship` `key` string validation `clean()` methods to `save()` to protect against creation of invalid data.
```
glennmatthews, line 1: Probably can/should rephrase this now?

nautobot/extras/tests/test_customfields.py
```diff
@@ -408,14 +408,14 @@ def setUpTestData(cls):
...
-            type=CustomFieldTypeChoices.TYPE_TEXT, label="Text Field", key="text_cf", default="foo"
+            type=CustomFieldTypeChoices.TYPE_TEXT, label="Text Field", key="text_cf", default="FOO"
         )
         cls.cf_text.save()
         cls.cf_text.content_types.set([content_type])
 
         # Integer custom field
         cls.cf_integer = CustomField(
-            type=CustomFieldTypeChoices.TYPE_INTEGER, label="Number Field", key="number_cf", default=123
+            type=CustomFieldTypeChoices.TYPE_INTEGER, label="Number Field", key="number_cf", default=12
```
glennmatthews, line 418: Why these changes?

HanlinMiao: Since we are calling `clean()` in `save()` now. This was making the tests fail on lines 874-875 and on lines 824-826.

Issue #4895, opened 2023-12-01
Enhancements to External Integration Model


lampwins:
As a Network Automation Engineer, I want jinja2 template support and additional concrete fields on the External Integration Model, so that I can more effectively use this new model.

Add fields:
- HTTP Method: choice field (GET, POST, PATCH, PUT, DELETE), not required
- Headers: json field, not required
- CA File Path: char field, not required

We can hide these under an "advanced accordion" in the form.

Add support for templating in the URL, headers, and extra config fields. Provide a method on the model that users can call to render the model instance with user provided template data.

glennmatthews:
Implemented by #4929.

PR #4929, opened 2023-12-06
Added `http_method`, `headers` and `ca_file_path` fields to `ExternalIntegration` model.

HanlinMiao:
# Closes: #4895 
# What's Changed

1. Added http_method, headers, and ca_file_path fields to ExternalIntegration model. 
2. Added accordion toggle to ExternalIntegrationEdit/CreateForm 
3. Added helper method to render remote_url, headers, and extra_config in jinja2 templates.

https://github.com/nautobot/nautobot/assets/46973263/d9796892-1944-4436-ace7-1373c9119343

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/migrations/0103_add_extra_fields_to_external_integration_model.py
glennmatthews, line 1: Would it be preferable to make this an addition to the existing `0101` migration, rather than adding a new one, since we haven't shipped a final release with that migration in it yet?

nautobot/extras/forms/forms.py
```diff
@@ -605,10 +609,17 @@ class ExternalIntegrationBulkEditForm(NautobotBulkEditForm):
     verify_ssl = forms.NullBooleanField(required=False, label="Verify SSL", widget=BulkEditNullBooleanSelect)
     timeout = forms.IntegerField(required=False, min_value=0)
     extra_config = forms.JSONField(required=False)
+    http_method = forms.ChoiceField(
+        required=False,
+        label="HTTP Method",
+        choices=add_blank_choice(HTTPMethodClassChoices),
+    )
+    headers = forms.JSONField(required=False, label="HTTP Request headers")
+    ca_file_path = forms.CharField(required=False, label="CA File Path")
```
glennmatthews, line 618: Is this order of fields in the form the best/most logical way to structure them? I would have thought `extra_config` at least should be the last in the form.

nautobot/extras/forms/forms.py
```diff
@@ -605,10 +609,17 @@ class ExternalIntegrationBulkEditForm(NautobotBulkEditForm):
...
 
     class Meta:
         model = ExternalIntegration
-        nullable_fields = ["extra_config", "secrets_group"]
+        nullable_fields = ["extra_config", "secrets_group", "http_method", "headers", "ca_file_path"]
```
glennmatthews, line 622: Looking at the model, `http_method` and `ca_file_path` are not nullable, nor should they be.

nautobot/extras/choices.py
```diff
@@ -125,6 +125,27 @@ class DynamicGroupOperatorChoices(ChoiceSet):
     )
 
 
+#
+# HTTP Methods
```
glennmatthews, line 129: Why not re-use the existing WebhookHttpMethodChoices instead of duplicating it?

nautobot/extras/models/models.py
```diff
@@ -489,13 +490,55 @@ class ExternalIntegration(PrimaryModel):
         null=True,
         help_text="Optional user-defined JSON data for this integration",
     )
+    http_method = models.CharField(
+        max_length=10,
+        choices=HTTPMethodClassChoices,
+        default=HTTPMethodClassChoices.METHOD_GET,
```
glennmatthews, line 496: Given that it's blankable, does it need a default other than ""?

nautobot/extras/models/models.py
```diff
@@ -489,13 +490,55 @@ class ExternalIntegration(PrimaryModel):
...
+        blank=True,
+    )
+    headers = models.JSONField(
+        encoder=DjangoJSONEncoder,
+        blank=True,
+        null=True,
+        help_text="Headers for the HTTP request",
+    )
+    ca_file_path = models.CharField(
+        max_length=100,
```
glennmatthews, line 507: Any reason for this limit or can we use something larger like 255?

nautobot/extras/models/models.py
```diff
@@ -489,13 +490,55 @@ class ExternalIntegration(PrimaryModel):
...
+        """
+        Render extra_config and return a dict of Header: Value pairs.
+        """
+        if not self.extra_config:
+            return {}
+        data = render_jinja2(self.extra_config, context)
+        return data
+
+    def render_remote_url(self, context):
+        if not self.remote_url:
```
glennmatthews, line 537: remote_url should never be blank

nautobot/extras/tables.py
```diff
@@ -423,6 +425,8 @@ class Meta(BaseTable.Meta):
             "secrets_group",
             "verify_ssl",
             "timeout",
+            "http_method",
+            "ca_file_path",
         )
```
glennmatthews, line 430: Looks like we're missing `tags` as a column, would you mind adding it?

nautobot/extras/models/models.py
```diff
@@ -489,13 +490,55 @@ class ExternalIntegration(PrimaryModel):
...
+    )
+    headers = models.JSONField(
+        encoder=DjangoJSONEncoder,
+        blank=True,
+        null=True,
+        help_text="Headers for the HTTP request",
+    )
+    ca_file_path = models.CharField(
+        max_length=100,
+        blank=True,
```
gsnider2195, line 508: Can you add `verbose_name = "CA file path",`?

nautobot/extras/models/models.py
```diff
@@ -489,13 +490,55 @@ class ExternalIntegration(PrimaryModel):
...
+        data = render_jinja2(self.headers, context)
+        return data
+
+    def render_extra_config(self, context):
+        """
+        Render extra_config and return a dict of Header: Value pairs.
+        """
+        if not self.extra_config:
+            return {}
+        data = render_jinja2(self.extra_config, context)
```
gsnider2195, line 533: `render_jinja2` returns a string so I don't think we should return an empty dict if extra_config is blank

nautobot/extras/models/models.py
```diff
@@ -489,13 +490,55 @@ class ExternalIntegration(PrimaryModel):
...
     class Meta:
         ordering = ["name"]
 
+    def render_headers(self, context):
+        """
+        Render headers and return a dict of Header: Value pairs.
+        """
+        if not self.headers:
+            return {}
+        data = render_jinja2(self.headers, context)
```
gsnider2195, line 524: Same comment as below

nautobot/extras/templates/extras/externalintegration_update.html
```diff
@@ -0,0 +1,65 @@
...
+        <div class="panel-heading"><strong>External Integration</strong></div>
+        <div class="panel-body">
+            {% render_field form.name %}
+            {% render_field form.remote_url %}
+            {% render_field form.secrets_group %}
+            {% render_field form.timeout %}
+            {% render_field form.extra_config %}
+        </div>
+    </div>
+    <div id="accordion" class="panel panel-default">
```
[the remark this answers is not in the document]
HanlinMiao: What do you think @gsnider2195?

HanlinMiao:
Note to self, need to add those fields to ExternalIntegrationFactory as well.

nautobot/extras/models/models.py
```diff
@@ -489,13 +489,53 @@ class ExternalIntegration(PrimaryModel):
...
         ordering = ["name"]
 
+    def render_headers(self, context):
+        """
+        Render headers and return a dict of Header: Value pairs.
+        """
+        if not self.headers:
+            return ""
+        data = render_jinja2(self.headers, context)
+        return data
```
glennmatthews, line 524: Docstring/behavior mismatch here. Also given `self.headers` is a JSON field, do we need to stringify it before rendering as jinja2? Perhaps something like:

```suggestion
        Render headers and return a dict of Header: Value pairs.
        """
        if not self.headers:
            return {}
        data = json.loads(render_jinja2(json.dumps(self.headers, ensure_ascii=False), context)
        return data
```

Probably need to also add error handling in case render_jinja2 or json.loads throws an exception.

nautobot/extras/models/models.py
```diff
@@ -489,13 +489,53 @@ class ExternalIntegration(PrimaryModel):
...
+
+    def render_extra_config(self, context):
+        """
+        Render extra_config and return a dict of Header: Value pairs.
+        """
+        if not self.extra_config:
+            return ""
+        data = render_jinja2(self.extra_config, context)
+        return data
```
glennmatthews, line 533: Same comments here.

nautobot/extras/models/models.py
```diff
@@ -489,13 +489,53 @@ class ExternalIntegration(PrimaryModel):
...
+
+    def render_remote_url(self, context):
```
glennmatthews, line 535: Missing docstring?

nautobot/extras/models/models.py
```diff
@@ -489,13 +489,53 @@ class ExternalIntegration(PrimaryModel):
...
+        verbose_name="CA file path",
+    )
 
     def __str__(self):
         return f"{self.name} ({self.remote_url})"
 
     class Meta:
         ordering = ["name"]
 
+    def render_headers(self, context):
```
glennmatthews, line 517: Need some model tests for these `render_*` APIs including negative tests.

nautobot/extras/tables.py
```diff
@@ -431,6 +437,8 @@ class Meta(BaseTable.Meta):
             "secrets_group",
             "verify_ssl",
             "timeout",
+            "http_method",
```
glennmatthews, line 440: should this be moved up?

changes/4895.added
```diff
@@ -0,0 +1 @@
+Added `http_method`, `headers` and `ca_file_path` fields to `ExternalIntegration` model.
```
glennmatthews, line 1:
```suggestion
Added `http_method`, `headers`, and `ca_file_path` fields, as well as `render_headers`, `render_extra_config`, and `render_remote_url` APIs, to the `ExternalIntegration` model.
```

nautobot/extras/models/models.py
```diff
@@ -489,13 +489,66 @@ class ExternalIntegration(PrimaryModel):
...
+        """
+        Render headers and return a dict of Header: Value pairs.
+        Raises:
```
glennmatthews, line 520: Docstring convention is to have a blank line after the first line of the docstring.

```suggestion
        Render headers and return a dict of Header: Value pairs.

        Raises:
```

nautobot/extras/models/models.py
```diff
@@ -489,13 +489,66 @@ class ExternalIntegration(PrimaryModel):
...
+        if not self.headers:
+            return {}
+
+        data = json.loads(render_jinja2(json.dumps(self.headers, ensure_ascii=False), context))
+        return data
+
+    def render_extra_config(self, context):
+        """
+        Render extra_config and return a dict of Header: Value pairs.
+        Raises:
```
glennmatthews, line 534:
```suggestion
        Render extra_config and return a dict of Header: Value pairs.

        Raises:
```

nautobot/extras/models/models.py
```diff
@@ -489,13 +489,66 @@ class ExternalIntegration(PrimaryModel):
...
+        """
+        if not self.extra_config:
+            return {}
+
+        data = json.loads(render_jinja2(json.dumps(self.extra_config, ensure_ascii=False), context))
+        return data
+
+    def render_remote_url(self, context):
+        """
+        Render remote_url in jinja2 templates.
```
glennmatthews, line 547:
```suggestion
        Render remote_url in jinja2 templates.

        Raises:
            TemplateAssertionError: Raised when an invalid template helper function exists in remote_url.
            TemplateSyntaxError: Raised when an invalid template variable exists in remote_url.
```

glennmatthews (commented):
Please be sure to update `nautobot/docs/user-guide/platform-functionality/externalintegration.md` as well to document these new fields and APIs.

nautobot/users/factory.py
```diff
@@ -13,27 +13,9 @@
...
-        ),
-        factory.Maybe(
-            "has_last_name",
-            factory.LazyAttribute(lambda u: u.last_name.lower()),
-            factory.Faker("user_name"),
-        ),
-    )
+        exclude = "has_email"
+
+    username = factory.Faker("user_name")
```
glennmatthews, line 18: Why this change? Removing the first_name/last_name properties from the generated user feels like a step backwards to me.

nautobot/extras/tests/test_models.py
```diff
@@ -809,6 +810,50 @@ def test_timeout_validation(self):
...
+        ei = ei_with_extra_config.first()
+        ei.extra_config = {"context": "{% foo %}"}
+        ei.save()
+        with self.assertRaises(TemplateSyntaxError):
+            ei.render_extra_config({})
+
+        ei.extra_config = "{{ data | notvalid }}"
+        ei.save()
+        with self.assertRaises(TemplateAssertionError):
+            ei.render_extra_config({})
```
glennmatthews, line 833: Am I missing something or do we not have any positive tests here (valid J2 template, data gets substituted as expected)?

glennmatthews (approved):

nautobot/extras/forms/forms.py
```diff
@@ -605,10 +609,16 @@ class ExternalIntegrationBulkEditForm(NautobotBulkEditForm):
     verify_ssl = forms.NullBooleanField(required=False, label="Verify SSL", widget=BulkEditNullBooleanSelect)
     timeout = forms.IntegerField(required=False, min_value=0)
     extra_config = forms.JSONField(required=False)
+    http_method = forms.ChoiceField(
+        required=False,
+        label="HTTP Method",
+        choices=add_blank_choice(WebhookHttpMethodChoices),
+    )
+    headers = forms.JSONField(required=False, label="HTTP Request headers")
```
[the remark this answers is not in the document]
HanlinMiao: Sorry, I should have responded more promptly. My reasoning is that since we are using a custom template for Edit/Create ExternalIntegration, the form field's order are hard-coded in the template anyway. So I thought the order here on the form did not matter. Please correct me if I am wrong.

nautobot/users/factory.py
HanlinMiao: Yeah, I think our old way of constructing a username was flaky and it was running into duplicate name issues in MySQL https://github.com/nautobot/nautobot/actions/runs/7149742147/job/19472323800. So I thought it might be more straightforward to just generate a user_name directly.

nautobot/extras/forms/forms.py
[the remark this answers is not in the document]
glennmatthews: Makes sense. Thanks :-)

nautobot/users/factory.py
glennmatthews: Would be interesting to know what name the faker was running into issues with. I wonder if just changing it so that first_name and last_name were always present instead of randomly absent would be a cleaner fix. Not critical though.

Issue #4529, opened 2023-09-26
Add Support for Custom Actions in `NautobotUIViewSet`


bryanculver:
### As ...

P.D. - Plugin Developer

### I want ...

to be able to define custom actions for my models on my existing NautobotUIViewSet.

### So that ...

I do not have to maintain a ViewSet along with discrete Views and Routes.

### I know this is done when...

- I can add a method on my `NautobotUIViewSet` decorated with the DRF `@action` decorator
- My method(s) can be either detail- or list-view-based actions
- I can specify the roles necessary to use said action

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [X] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

NA

### External Dependencies

NA

lampwins:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Nautobot-Core-Product-64a8783b7a3f4b1225973a0c/issues/nautobot/nautobot/4529?planning-poker) with Zenhub @bryanculver @gsnider2195 @HanlinMiao @jathanism

jathanism:
Looks like the missing link are the `DynamicRoute` actions being created with `detail=False`. From the DRF source code:

```python
        # Dynamically generated list routes. Generated using
        # @action(detail=False) decorator on methods of the viewset.
        DynamicRoute(
            url=r'^{prefix}/{url_path}{trailing_slash}$',
            name='{basename}-{url_name}',
            detail=False,
            initkwargs={}
        ),
```

itdependsnetworks:
@jathanism this was the solution provided already, it was not clear to me what the challenge was with it. 
https://github.com/nautobot/nautobot/pull/4514/commits/993674d19e547cf31395edcb6547a1ba4efbc93a#diff-b6155147f3bbc9256ba909bcd27607a8a93896a7945cc05fb24b511d448be5b0R55

itdependsnetworks:
My guess was there was confusion in the testing, and it was lost that the order I put the route in was important, see here for related comment: https://github.com/nautobot/nautobot/pull/4514/commits/993674d19e547cf31395edcb6547a1ba4efbc93a#r1337318755

[...]

PR #4942, opened 2023-12-08
Dynamic Route for ExamplePlugin

HanlinMiao:
# Closes: #N/A
# What's Changed

# TODO

- [ ] Unit, Integration Tests


lampwins:
If this is all it is, that's fine. Do want to make sure this doesn't grow, otherwise we will need to punt to 2.2.

nautobot/core/views/mixins.py
```diff
@@ -252,7 +252,7 @@ def get_required_permission(self):
         """
         queryset = self.get_queryset()
         try:
-            actions = [PERMISSIONS_ACTION_MAP[self.action]]
+            actions = [PERMISSIONS_ACTION_MAP.get(self.action, "view")]
```
glennmatthews, line 255: This is a different approach than what's being done in #4512 to address #4529, so we should consider that carefully. Also I'm not sure `view` is the appropriate default permission to use here.

examples/example_plugin/example_plugin/views.py
```diff
@@ -77,6 +78,18 @@ class ExampleModelUIViewSet(views.NautobotUIViewSet):
     serializer_class = serializers.ExampleModelSerializer
     table_class = tables.ExampleModelTable
 
+    @action(detail=False, methods=["get"], url_path="custom_action_get_all_example_model_names")
+    def custom_action_get_all_example_model_names(self, request):
```
glennmatthews, line 82: Usually the action corresponds to the URL endpoint, so maybe something like `def all_names(...)` might be more typical?

changes/4942.added
```diff
@@ -0,0 +1 @@
+Added DynamicRoute() to NautbotUIViewSet and added a custom view using DynamicRoute() in Example App.
```
glennmatthews, line 1:
```suggestion
Added DynamicRoute() to NautobotUIViewSet to support custom actions.
```

Separately, 4942.housekeeping:

```
Added example of a custom action on a NautobotUIViewSet to the Example App.
```

nautobot/core/views/mixins.py
```diff
@@ -241,8 +241,6 @@ def get_permissions_for_model(self, model, actions):
         """
         model_permissions = []
         for action in actions:
-            if action not in ("view", "add", "change", "delete"):
-                raise ValueError(f"Unsupported action: {action}")
```
HanlinMiao, line 245: Removing this part since we are supporting custom actions that might not be in these typical action list.

nautobot/core/views/mixins.py
```diff
@@ -473,7 +471,11 @@ def get_queryset(self):
         Override the original `get_queryset()` to apply permission specific to the user and action.
         """
         queryset = super().get_queryset()
-        return queryset.restrict(self.request.user, PERMISSIONS_ACTION_MAP[self.action])
+        return queryset.restrict(self.request.user, self.get_action())
```
glennmatthews, line 474: Might be good to add a unit test verifying the permissions enforcement for `/plugins/example-plugin/example-models/all-names/` (or whatever) is correctly using the custom action permission?

examples/example_plugin/example_plugin/views.py
```diff
@@ -77,6 +78,18 @@ class ExampleModelUIViewSet(views.NautobotUIViewSet):
     serializer_class = serializers.ExampleModelSerializer
     table_class = tables.ExampleModelTable
 
+    @action(detail=False, methods=["get"], url_path="all_names")
+    def all_names(self, request):
+        """
+        Returns a list of all the example model names.
+        """
+        all_example_models = ExampleModel.objects.all()
```
glennmatthews, line 86: For testing purposes, this *probably* should be using `self.get_queryset()`?

itdependsnetworks (approved):

glennmatthews (approved):

nautobot/extras/tests/test_plugins.py
```diff
@@ -485,6 +485,69 @@ def test_relationship_association_validator_raises_exception(self):
...
+    """Test for custom action view `all_names` added to Example App"""
+
+    model = ExampleModel
+
+    @classmethod
+    def setUpTestData(cls):
+        ExampleModel.objects.create(name="Example 1", number=100)
+        ExampleModel.objects.create(name="Example 2", number=200)
+        ExampleModel.objects.create(name="Example 3", number=300)
+        cls.custom_view_url = reverse("plugins:example_plugin:examplemodel_all-names")
```
glennmatthews, line 502: Not a huge fan of the mixed dashes/underscores here... anything we can or should do about that without making a breaking change?

examples/example_plugin/example_plugin/views.py
```diff
@@ -77,6 +78,18 @@ class ExampleModelUIViewSet(views.NautobotUIViewSet):
     serializer_class = serializers.ExampleModelSerializer
     table_class = tables.ExampleModelTable
 
+    @action(detail=False, methods=["get"], url_path="all_names")
```
glennmatthews, line 81: For consistency the url_path should be (if possible) `all-names` since we normally use hyphens rather than underscores in our URLs.

nautobot/extras/tests/test_plugins.py
HanlinMiao: Specifying `url_name` on the` @action` header fixed the issue.

Issue #4736, opened 2023-10-31
Document __isnull filters in 2.1 Release Overview


glennmatthews:
### Proposed Changes

Follow up to #4635 - this feature should be documented in the "Release Overview" section of the 2.1 release notes

### Justification

Significant feature enhancement.

glennmatthews:
Implemented by #4971.

PR #4971, opened 2023-12-12
Documented `isnull` filter expression in v2.1 Release Overview

HanlinMiao:
# Closes: #4736 
# What's Changed

# TODO

- [ ] Unit, Integration Tests


nautobot/docs/release-notes/version-2.1.md
```diff
@@ -8,6 +8,13 @@ This document describes all new features and changes in Nautobot 2.1.
 
 ### Added
 
+#### isnull Filter on Nullable Fields
```
glennmatthews, line 11: Perhaps?

```suggestion
#### UI/API `isnull` Filter on Nullable Fields
```

in which case this would alphabetically belong at the end of the `Added` section rather than the beginning of it?

nautobot/docs/release-notes/version-2.1.md
```diff
@@ -8,6 +8,13 @@ This document describes all new features and changes in Nautobot 2.1.
...
+
+`<attr>__isnull` filter expression has been added to support filtering objects based on nullable model fields (i.e. model fields with `null=True`). Now users can obtain a filtered list of objects from the API with `<attr>__isnull=True/False`.
```
glennmatthews, line 13: Perhaps?

```suggestion
Models with nullable fields (i.e. model fields with `null=True`) can now be filtered in the UI and the API with `<field>__isnull=true/false` filters. These filters are automatically added to all appropriate fields.
```

nautobot/docs/release-notes/version-2.1.md
HanlinMiao: I think the issue is ordered by their issue number perhaps? This issue is 1905 and should be in front

glennmatthews: The previous minor versions' *release overviews* (not the summary release-notes for each patch) are ordered alphabetically.

glennmatthews (approved):

Issue #4620, opened 2023-10-07
Install issue 2.0.1 -> collect static fails to create directories


travisjohnsonga:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.1
* Python version: 3.10.12
* Database platform, version: postgres
* Middleware(s):none

### Steps to Reproduce
1. New installation of nautobot
2. Following instructions at https://blog.networktocode.com/post/installing-nautobot/
3. Issue command nautobot-server collectstatic

### Expected Behavior
Expected directories to be created and static files to be collected

Error occurred

nautobot@server:~$ nautobot-server collectstatic
16:21:56.992 ERROR   nautobot.core.apps :
  Error in link construction for Notes: Reverse for 'note_list' not found. 'note_list' is not a valid view function or pattern name.
Traceback (most recent call last):
  File "/opt/nautobot/bin/nautobot-server", line 8, in <module>
    sys.exit(main())
  File "/opt/nautobot/lib/python3.10/site-packages/nautobot/core/cli/__init__.py", line 54, in main
    run_app(
  File "/opt/nautobot/lib/python3.10/site-packages/nautobot/core/runner/runner.py", line 297, in run_app
    management.execute_from_command_line([runner_name, command] + command_args)
  File "/opt/nautobot/lib/python3.10/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/opt/nautobot/lib/python3.10/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/opt/nautobot/lib/python3.10/site-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/opt/nautobot/lib/python3.10/site-packages/django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
  File "/opt/nautobot/lib/python3.10/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 187, in handle
    collected = self.collect()
  File "/opt/nautobot/lib/python3.10/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 105, in collect
    for path, storage in finder.list(self.ignore_patterns):
  File "/opt/nautobot/lib/python3.10/site-packages/django/contrib/staticfiles/finders.py", line 130, in list
    for path in utils.get_files(storage, ignore_patterns):
[...]

PR #4976, opened 2023-12-13
Collectstatic failing due to no such file or directory

whitej6:
# Closes: #4620
# What's Changed

Added ability to initialize the directory and prevent unintended failures when running collectstatic

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/core/cli/__init__.py
+++ b/nautobot/core/cli/__init__.py
@@ -115,6 +115,7 @@ def _configure_settings(config):
     os.makedirs(settings.MEDIA_ROOT, exist_ok=True)
     os.makedirs(os.path.join(settings.MEDIA_ROOT, "devicetype-images"), exist_ok=True)
     os.makedirs(os.path.join(settings.MEDIA_ROOT, "image-attachments"), exist_ok=True)
+    os.makedirs(os.path.join(settings.NAUTOBOT_UI_DIR, "build", "static"), exist_ok=True)
     os.makedirs(settings.STATIC_ROOT, exist_ok=True)
 
     #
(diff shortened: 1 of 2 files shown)


nautobot/core/cli/__init__.py
```diff
@@ -115,6 +115,7 @@ def _configure_settings(config):
     os.makedirs(settings.MEDIA_ROOT, exist_ok=True)
     os.makedirs(os.path.join(settings.MEDIA_ROOT, "devicetype-images"), exist_ok=True)
     os.makedirs(os.path.join(settings.MEDIA_ROOT, "image-attachments"), exist_ok=True)
+    os.makedirs(os.path.join(settings.NAUTOBOT_UI_DIR, "build", "static"), exist_ok=True)
     os.makedirs(settings.STATIC_ROOT, exist_ok=True)
```
glennmatthews, line 119: Looks solid,  but I wonder if instead we should do something like:

```suggestion
    for directory in settings.STATICFILES_DIRS:
        os.makedirs(directory, exist_ok=True)
```

What do you think?

glennmatthews: Oh wait, I'm confusing myself. `STATIC_ROOT` is a destination directory, while `NAUTOBOT_UI_DIR/build/static` is a source directory (but only if we've run `nautobot-server build_ui`). So the current implementation is correct I think.

glennmatthews (approved):

nautobot/core/cli/__init__.py
whitej6: Confirmed locally that the correct dir is being created.

PR #4977, opened 2023-12-13
Process CSV import one row at a time to permit back-references

glennmatthews:
# Closes: #4998
# What's Changed

Instead of processing all rows of a CSV in a single ListSerializer instance, process each row sequentially. This may (**may**) result in a performance hit, but it permits self-referential models such as TreeModels to be defined in a CSV where later rows reference back to objects defined/created in earlier rows, which was previously not possible in a single CSV import, e.g. for LocationTypes:

```csv
name,nestable,parent__name
Region,TRUE,NoObject
Site,FALSE,Region
Building,FALSE,Site
```

# TODO

- [x] Unit, Integration Tests


nautobot/extras/tests/test_views.py
```diff
@@ -766,7 +766,7 @@ def setUpTestData(cls):
             "name,slug,remote_url,branch,secrets_group,provided_contents",
             "Git Repository 5,git_repo_5,https://example.com,main,,extras.configcontext",
             "Git Repository 6,git_repo_6,https://example.com,develop,Secrets Group 2,",
-            'Git Repository 7,git_repo_7,https://example.com,next,Secrets Group 2,"extras.job,extras.configcontext"',
+            'Git Repository 7,git_repo_7,https://example.com,next,Secrets Group 2,"extras.job,extras.exporttemplate"',
```
HanlinMiao, line 769: Why this change here?

HanlinMiao (approved):

nautobot/extras/tests/test_views.py
glennmatthews: Actually this was a bug caught by this change. Previously, since all of the rows were being created in a single operation, the `clean()` check for multiple GitRepositories with the same `remote_url` and overlapping `provided_contents` wasn't catching the conflict between "Git Repository 5" and "Git Repository 7" both providing `extras.configcontext` from `https://example.com`. With this change, that conflict is now detected and causes the import to fail, hence why I had to change the data.

glennmatthews:
> Is there any performance impact for this change?

Almost certainly. I'll see if I can gather some data to quantify it.

gsnider2195 (approved):

glennmatthews:
Importing 71 IPAddress records via CSV took approximately 15 seconds in `develop` and approximately 19 seconds in this branch, both with `DEBUG = False` to turn off DjDT. So it's slower, but not excessively so IMHO.

We definitely should plan to move CSV import to a background task / system job, in any case.

PR #4981, opened 2023-12-14
added Serial industrial types

Torstein-Eide:
# What's Changed

Add choices for some used serial interfaces:
- Generic RS232, RS422, RS423, R485
- Modbus
- Profibus
- HART

# TODO


nautobot/dcim/choices.py
```diff
@@ -821,6 +821,17 @@ class InterfaceTypeChoices(ChoiceSet):
...
+    TYPE_EIA422 = "eia422"
+    TYPE_EIA423 = "eia423"
+    TYPE_EIA485 = "eia485"
+    TYPE_EIA485_modbus = "modbus"
+    TYPE_EIA485_profibus = "profibus"
+    TYPE_fiber_profibus = "profibus-fiber"
+    TYPE_HART_RS232 = "hart-rs232"
+    TYPE_HART_RS422 = "hart-rs422"
+    TYPE_HART_RS485 = "hart-rs485"
+    TYPE_HART_bell202 = "hart-bell202"
```
glennmatthews, line 834: Our general convention would be for each of the constants to be all-caps, rather than mixed case, i.e. `TYPE_EIA485_MODBUS`, etc.

nautobot/dcim/choices.py
```diff
@@ -821,6 +821,17 @@ class InterfaceTypeChoices(ChoiceSet):
     TYPE_E1 = "e1"
     TYPE_T3 = "t3"
     TYPE_E3 = "e3"
+    TYPE_EIA232= "eia232"
```
glennmatthews, line 824: Would it be more self-consistent and make better sense for these to be the connector type rather than the protocol? E.g. `ConsolePortTypeChoices` has `TYPE_DE9 = "de-9"` and `TYPE_DB25 = "db-25"` rather than `eia232`.

Eideen: For serial port there is 4 elements:
* The physical connector, there is a lot.
* Is the connector a point to point, og bus.
* OSI model 1, how bits are send, EIA233, EIA485, BELL202, fiber,
* OSI model 2-6 is not used.
* OSI model 7, how the data is formatted, Profibus, MODbus, HART, and you can make your own.

So based on that i think a mix of OSI L1 and OSI L7, is what will  give us the best result, this is most inline with the other elements in the table.

And option is to have physical connector, as it own table.

glennmatthews: Tagging @lampwins to provide guidance here.

lampwins: Let's focus on the physical type and a custom field can be used for transport protocol where appropriate. In the future we can look at adding additional context to the data model here.

gsnider2195 (approved):

Issue #4988, opened 2023-12-15
Job and job-button constraints permissions have not been applied


abdikanipd:
### Environment
* Nautobot version (Docker tag too if applicable): 1.6.7
* Python version: 3.8
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1 .Create permission for the group and apply a constraint to allow them to run a single job button.
2.
3.

### Expected Behavior
Just group have permission of single job and job-button, nothing else and still user can run all job-buttons. Expected to view and run just allowed job

### Observed Behavior
Still user can run all job-buttons

glennmatthews:
Thanks for the report! Haven't tested it yet but looking at the code it does appear that way. We're only checking `context["user"].has_perms(("extras.run_jobbutton", "extras.run_job"))` when rendering the job_buttons templatetag, nothing more specific per Job/JobButton instance, and it similarly looks like the view code is only checking for generic `extras.run_job` permission and failing to enforce per-job-button permissions.

abdikanipd:
@glennmatthews It seems to only check if the user has the `context 'user'. has_perms (("extras. run_jobbutton", "extras. run_job"))` Correct.  Not for specific jobs, I tried to remove the `extras. run_jobbutton` perms and users can see buttons but cannot run them. It is possible to check `job_pk and job-button_id` for every job context. Constraints can also be based on multiple jobs and job buttons.

PR #4993, opened 2023-12-15
Fix object permissions enforcement on Job Buttons

glennmatthews:
# Closes: #4988 (will also be needed in 1.6.x)
# What's Changed

- The `job_buttons` template tag wasn't considering object-level permissions when determining whether to enable or disable each JobButton it rendered. I added the appropriate `has_perms` check.
  - There was a lot of duplicate code and logic in this function between the "ungrouped buttons" and "grouped buttons" cases, so I've extracted it into a helper function to reduce duplication.
- Similarly, the `JobButtonRunView` implementation was also not considering object-level permissions. In this case this view is redundant as a JobButton's underlying Job can be invoked just as easily through the `/extras/jobs/<id>/run/` view, which is already enforcing permissions correctly. I've therefore *removed `JobButtonRunView`* rather than attempt to fix it.
   - To keep the existing behavior where clicking a JobButton redirects back to the current page rather than redirecting to the JobResult page, I've added support for a `_return_url` parameter to the Job run view.
- Enforcing `extras.run_jobbutton` permission on top of `extras.run_job` permission was redundant, especially since a user with `extras.run_job` permission alone could still run the underlying job without going through the JobButton. I've therefore *removed `extras.run_jobbutton` as a required permission*; `extras.run_job` suffices.
- In manual testing, I found that the example plugin `ExampleComplexJobButtonReceiver` hadn't been completely/correctly updated to work under 2.0, so I fixed it.

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/templatetags/job_buttons.py
+++ b/nautobot/extras/templatetags/job_buttons.py
@@ -6,7 +6,7 @@ from django.urls import reverse
 from django.utils.html import format_html
 from django.utils.safestring import mark_safe
 
-from nautobot.extras.models import JobButton
+from nautobot.extras.models import Job, JobButton
 from nautobot.core.utils.data import render_jinja2
 
 
@@ -27,7 +27,8 @@ HIDDEN_INPUTS = """
 <input type="hidden" name="csrfmiddlewaretoken" value="{csrf_token}">
 <input type="hidden" name="object_pk" value="{object_pk}">
 <input type="hidden" name="object_model_name" value="{object_model_name}">
-<input type="hidden" name="redirect_path" value="{redirect_path}">
+<input type="hidden" name="_schedule_type" value="immediately">
+<input type="hidden" name="_return_url" value="{redirect_path}">
 """
 
 NO_CONFIRM_BUTTON = """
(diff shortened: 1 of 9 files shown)


nautobot/extras/templatetags/job_buttons.py
```diff
@@ -69,115 +70,108 @@
 </div>
 """
 
+SAFE_EMPTY = mark_safe("")  # noqa: S308
```
gsnider2195, line 73:
```suggestion
SAFE_EMPTY_STR = mark_safe("")  # noqa: S308
```
Make it easier to determine what this variable is doing without context later in the file.

nautobot/extras/views.py
```diff
@@ -1223,6 +1226,19 @@ def post(self, request, class_path=None, pk=None):
                     **job_model.job_class.serialize_data(job_kwargs),
                 )
 
+                return_url = request.POST.get("_return_url")
```
gsnider2195, line 1229: There's a change in behavior here that we'll want to make sure is desired. Previously job buttons ignored any of the error cases above this `else` block and were always redirected back to the object detail page.

timizuoebideri1 (approved):

nautobot/extras/views.py
glennmatthews: Good catch. Let me see if there's a straightforward way I can make the error cases also respect `_return_url`.

gsnider2195 (approved):
💯

PR #4994, opened 2023-12-15
more flexible default sanitizer pattern

jeffkala:
# Closes:  NA
# What's Changed

Default sanitizer patterns don't capture the same patterns wrapped in a python dictionary. Also secret / secrets isn't a pattern that is currently caught.  This is a small update to just be a slighly more flexible default sanitizer.

# TODO

- [ ] Unit, Integration Tests


timizuoebideri1:
Could you add a test case entry to `nautobot.core.tests.test_logging.LoggingUtilitiesTest.DIRTY_CLEAN`

nautobot/core/settings.py
```diff
@@ -124,7 +124,10 @@
 SANITIZER_PATTERNS = [
     # General removal of username-like and password-like tokens
     (re.compile(r"(https?://)?\S+\s*@", re.IGNORECASE), r"\1{replacement}@"),
-    (re.compile(r"(username|password|passwd|pwd)((?:\s+is.?|:)?\s+)\S+", re.IGNORECASE), r"\1\2{replacement}"),
+    (
+        re.compile(r"(username|password|passwd|pwd|secret|secrets)([\"']?(?:\s+is.?|:)?\s+)\S+[\"']?", re.IGNORECASE),
```
gsnider2195, line 128: Probably also a good idea to provide an example of what kind of cases this covers in a comment

jeffkala: @gsnider2195  In a comment in the PR or in the doc string?

glennmatthews: Personally I'd vote for a comment here and maybe a note in the documentation as well.

jeffkala:
Opened an issue #5031 for proper changelog entry.

nautobot/core/settings.py
glennmatthews: @jeffkala The docs update is the only thing remaining before we can merge this, I think?

jeffkala: will get the docs added today.

jeffkala: @glennmatthews fit best to put some additional docs in the optional-settings existing docs? Or create a new page of sorts?  Not seeing anywhere else that fits in the current docs structure.

glennmatthews: Yeah I was thinking in the optional-settings docs for SANITIZER_PATTERNS.

glennmatthews (approved):

nautobot/core/settings.py
jeffkala: Just updated optional settings with some examples.

Issue #4996, opened 2023-12-18
External Integration UI changes


glennmatthews:
### As ...

Austin - Network Automation Engineer

### I want ...

- "Extra config" to be moved to the bottom of the "Advanced" section
- Remove redundant "headers" key from help text example of "Headers" field data
- Revise example "Extra config" help text to be not identical to headers?
- Add `<code>` tags in help text around "render_*()" methods
- Remove redundant "CA file path" help text

### So that ...

We are responsive to user feedback from sprint review.

### I know this is done when...

All of the above changes are implemented.

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [X] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

glennmatthews:
Implemented by #5008.

PR #5008, opened 2023-12-20
External Integration edit form updates

gsnider2195:
# Closes: #4996
# What's Changed

image

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/forms/forms.py
```diff
@@ -584,29 +584,26 @@ class Meta:
...
             Optional user-defined <a href="https://json.org/">JSON</a> data for this integration. Example:
             <pre><code>{
-                "headers": {
-                    "Accept": "application/json",
-                    "Content-Type": "application/json",
-                },
+                "Accept": "application/json",
+                "Content-Type": "application/json"
             }</code></pre>
-            You can render the field value in Jinja2 template as well with the .render_extra_config() helper method.
```
gsnider2195, line 595: Removed mentions of Jinja2 since the `render_*` methods are model methods and the edit UI for the model feels like the wrong place to document these methods. Also, since using the render methods is optional and core doesn't have control over the context variables passed to `render_jinja2`, we can't even provide an example.

timizuoebideri1 (approved):

nautobot/extras/forms/forms.py
glennmatthews: Makes sense - I see that we document that these APIs exist in the feature documentation, would it be desirable to add an example there? Even something simple like `remote_url = "https://example.com/api/devices?name={{ device.name }}"` with `{"device": <device instance>}` as the template context?

nautobot/extras/forms/forms.py
```diff
@@ -584,29 +584,26 @@ class Meta:
         model = ExternalIntegration
         fields = "__all__"
 
-        EXTRA_CONFIG_HELP_TEXT = """
+        HEADERS_HELP_TEXT = """
             Optional user-defined <a href="https://json.org/">JSON</a> data for this integration. Example:
             <pre><code>{
```
glennmatthews, line 589: `<pre><code>` seems redundant to me - does it render fine if we remove `<code>` tags?

glennmatthews (approved):

Issue #4961, opened 2023-12-12
UI Feedback from 2.1.0beta1


cmsirbu:
### Environment
* Nautobot version (Docker tag too if applicable): 2.1.0b1
* https://next.demo.nautobot.com/
* Browser: Chrome (120.0.6099.71) on Ubuntu 20.04

## User menu

![image](https://github.com/nautobot/nautobot/assets/9288571/0d33fa87-95d1-495e-a611-f8617ea91d77)

User pop-up menu is out of place at the top or bottom of the nav - suggestion: move it to the bottom of the nav panel as a fixed line anchored to the bottom of the page with always visible Profile/Logout buttons. Something like this (just a quickie image edit):

![image](https://github.com/nautobot/nautobot/assets/9288571/19acd15e-6978-4914-b119-198a66cc61e8)

## Large nav menu pop-ups cause a scrollbar to appear

![image](https://github.com/nautobot/nautobot/assets/9288571/43eb9c63-36a4-4f08-9207-7cbb4041c839)

Click on `DEVICES` in the nav. The drop-down menu overflows the page and causes a scroll-bar to appear, which causes the usable space in the nav to shrink.

Expected behaviour: the scrollbar does not affect the layout of the nav column by reducing its width.

## Vertical side-bar nav does not behave as one would expect

Using drop-down pop-ups in a vertical side nav bar is not a common pattern and completely unexpected UI wise.

It's not entirely obvious that it indeed is a temporary drop-down instead of a menu expansion inline, until one opens a smaller category, like `POWER`:

![image](https://github.com/nautobot/nautobot/assets/9288571/0834e145-c17f-42be-b7c1-5c91e3a9ffec)

A side-effect of this temporary drop-down is that when navigating to a particular page (e.g. the Devices page within the `DEVICES` category), the nav resets to the default collapsed mode not showing the location of the current page within the larger menu.

![image](https://github.com/nautobot/nautobot/assets/9288571/2bdf4867-1119-452d-b00d-59c09cfafd9c)

Expected behaviour: the left-hand side nav expands inline (no drop-downs) and tracks the current location of the main view, highligting the "Power Feeds" menu item when viewing the respective page.

## The nav is too narrow

When expanding categories with longer names, the scrollbar makes the layout overflow on a new line and a horizontal scroll-bar appears:

![image](https://github.com/nautobot/nautobot/assets/9288571/fcdc7d4a-7c5c-4267-a7e5-2e90af9ace83)

[...]

PR #5013, opened 2023-12-21
2.1 UI updates

gsnider2195:
# Closes: #4961 
# What's Changed

- Updated navbar width to 240px
- Added divider between user navbar menu item and rest of the menu items above
- Added ellipses for text overflow in all navbar top level menus
- Fixed bug in dark mode where the navbar menu would turn white when losing focus

navbar_light
navbar_dark

# TODO

- [ ] Unit, Integration Tests


changes/4961.changed
```diff
@@ -0,0 +1,2 @@
+Widened left navigation menu and made longer menus truncate instead of overflowing into the next line.
```
glennmatthews, line 1: Is this for menu items too or only for top-level menu tabs? (E.g. does "Interface Redundancy Groups" still get wrapped, or is it now truncated?)

glennmatthews (approved):

changes/4961.changed
gsnider2195: This is only for top level items:

image

This is what it would look like if applied to all items:

image

Here's what device lifecycle would look like, notice the 4 reports at the bottom:

image

glennmatthews:
Looks like maybe some integration test changes are needed?

PR #5024, opened 2023-12-23
A few performance updates

lampwins:
A random assortment of quick performance related fixes.

Screenshot 2023-12-23 at 1 26 20 AM


nautobot/core/models/querysets.py
```diff
@@ -6,17 +6,23 @@
 from nautobot.core.utils.data import merge_dicts_without_collision
 
 
-def count_related(model, field):
+def count_related(model, field, filter_dict=None):
```
glennmatthews, line 9: As a nit, we might want to make filter_dict a keyword-only argument for clarity? I.e.:

```suggestion
def count_related(model, field, *, filter_dict=None):
```

and

```python
debug_log_count=count_related(JobLogEntry, "job_result", filter_dict={"log_level": LogLevelChoices.LOG_DEBUG}),
```

etc.

nautobot/core/models/querysets.py
```diff
@@ -6,17 +6,23 @@
...
     """
     Return a Subquery suitable for annotating a child object count.
+
+    Args:
+        model: The related model to aggregate
+        field: The file don the related model which points back to the OuterRef model
```
glennmatthews, line 15:
```suggestion
        field: The field on the related model which points back to the OuterRef model
```

nautobot/extras/views.py
```diff
@@ -1512,7 +1513,18 @@ class JobResultListView(generic.ObjectListView):
...
+    queryset = (
+        JobResult.objects.defer("result")
+        .select_related("job_model", "user")
+        .annotate(
+            debug_log_count=count_related(JobLogEntry, "job_result", {"log_level": LogLevelChoices.LOG_DEBUG}),
+            info_log_count=count_related(JobLogEntry, "job_result", {"log_level": LogLevelChoices.LOG_INFO}),
+            warning_log_count=count_related(JobLogEntry, "job_result", {"log_level": LogLevelChoices.LOG_WARNING}),
+            error_log_count=count_related(
+                JobLogEntry, "job_result", {"log_level__in": [LogLevelChoices.LOG_ERROR, LogLevelChoices.LOG_CRITICAL]}
+            ),
```
glennmatthews, line 1525: Very nice!

changes/5024.fixed
```diff
@@ -0,0 +1,5 @@
+Improved performance of the Job Result list view by optimizing the way JobLogEntry records are queried.
+Improved performance of the Device list view by including the manufacturer name in the table queryset.
+Improved performance of most ObjectListViews by optimizing how Custom fields, Computed fields, and Relationships are queried.
+Fixed a bug that causes IPAddress objects to query their parent Prefix and Namespace every time they are instantiated.
```
glennmatthews, line 4:
```suggestion
Fixed a bug that caused IPAddress objects to query their parent Prefix and Namespace every time they were instantiated.
```

nautobot/extras/models/relationships.py
```diff
@@ -329,16 +330,54 @@ def required_related_objects_errors(
...
 
-    def get_for_model(self, model):
+    @lru_cache(maxsize=128)
+    def get_for_model(self, model, hidden=False):
         """
         Return all Relationships assigned to the given model.
+
+        Args:
+            model: The django model to which relationships are registered
+            hidden: Boolean to look for hidden relationships
```
glennmatthews, line 340: Feels weird to me that `hidden` is always either `True` or `False` here, so in either case you only get a subset of all relationships -- shouldn't there be a way to get all relationships regardless of their `hidden` flag?

nautobot/ipam/models.py
```diff
@@ -1103,10 +1095,6 @@ def clean(self):
         if self.type == choices.IPAddressTypeChoices.TYPE_SLAAC and self.ip_version != 6:
             raise ValidationError({"type": "Only IPv6 addresses can be assigned SLAAC type"})
 
-        # If neither `parent` or `namespace` was provided; raise this exception
```
gsnider2195, line 1106: I think `self._get_closest_parent` is performing this function adequately

glennmatthews: `_get_closest_parent` doesn't check for the `self._namespace = None` case currently as I read it, so there may need to be some change made there.

nautobot/ipam/models.py
```diff
@@ -1120,6 +1108,7 @@ def clean(self):
                     }
                 )
             self.parent = closest_parent
+            self._namespace = None
```
glennmatthews, line 1111: Shouldn't this be updating to `self.parent.namespace` instead of `None`? Same question below.

(edit) Oh, this is because setting `self._namespace = None` actually means "use self.parent.namespace"? That feels pretty unintuitive to me.

nautobot/ipam/models.py
```diff
@@ -1181,6 +1171,19 @@ def siblings(self, include_self=False):
...
         return query
 
+    @property
+    def _namespace(self):
+        # if a namespace was explicitly set, use it
+        if getattr(self, "_provided_namespace", None):
+            return self._provided_namespace
+        if self.parent is not None:
+            return self.parent.namespace
+        return get_default_namespace()
```
glennmatthews, line 1181: This also "feels" off to me - so if we have an IPAddress with no parent, it's always going to be implicitly in the global namespace?

nautobot/ipam/models.py
gsnider2195: Currently `self._namespace` defaults to `get_default_namespace()` when `self._provided_namespace = None` and `self.parent` isn't set.

nautobot/ipam/models.py
gsnider2195: Actually, I don't think we should set `self._namespace = self.parent.namespace` because the parent's namespace can be changed. If we cache the namespace and then the parent's namespace changes, our cache is incorrect.

nautobot/ipam/models.py
gsnider2195: If you try to create an IPAddress without specifying a namespace or parent (in the ORM only) it will try to find a valid parent in the default namespace and if one can be found it assigns that parent. This isn't changing that behavior, just moving it out of the model `__init__` so that creating a new instance doesn't hit the database.

The IPAddress namespace/parent logic is a non-standard workflow. In the UI if you're creating an IPAddress you can only select the namespace but we use that namespace to try to find a valid parent in that namespace and if we don't find one we fail validation in `clean`/`save`.

nautobot/ipam/models.py
gsnider2195: If we implement the `parent.setter` this will probably be easier to understand

gsnider2195: It may also be better to move that logic out of `self._namespace` and into the places where it's actually used.

glennmatthews (approved):

gsnider2195 (changes requested):
We need to clear the lru cache on object delete as well as object save

gsnider2195:
> We need to clear the lru cache on object delete as well as object save

Also need to invalidate the cache when CustomField.content_types changes, and the same for the other relevant models.

gsnider2195:
Let me know if https://github.com/nautobot/nautobot/pull/5024/commits/928d012bdf5893c5744002d79036bcba95a57b4b looks ok for cache invalidation.

nautobot/extras/tests/test_customfields.py
```diff
@@ -385,15 +385,51 @@ def test_to_filter_field(self):
...
         self.assertEqual(CustomField.objects.get_for_model(Location).count(), 2)
         self.assertEqual(CustomField.objects.get_for_model(VirtualMachine).count(), 0)
 
+    def test_get_for_model_lru_cache(self):
+        """Test that the lru cache is properly invalidated when CustomFields are created or deleted."""
+
+        qs1 = CustomField.objects.get_for_model(Location)
+        self.assertIsNone(qs1._result_cache)
+        list(qs1)  # force queryset evaluation
+        self.assertIsNotNone(qs1._result_cache)
```
glennmatthews, line 403: This is surprising to me that calling get_for_model doesn't populate the cache, only evaluating the queryset does. I guess this is an interaction between lru_cache and Django's lazy queryset evaluation? It feels a bit weird to do the initial assertIsNone as that feels like an implementation detail. Perhaps?

```suggestion
        qs1 = CustomField.objects.get_for_model(Location)
        self.assertEqual(qs1.count(), 2)
        self.assertIsNotNone(qs1._result_cache)
```

Actually I'm not a big fan of checking `_result_cache` either, would it be possible/better to check the return from `qs1.cache_info()` instead?

Similar comments on the other checks below as well.

nautobot/extras/tests/test_customfields.py
```diff
@@ -385,15 +385,51 @@ def test_to_filter_field(self):
...
+
+        # Assert that the cache is invalidated on object save
```
glennmatthews, line 405: Should we have at least one test asserting that the cache is used when no change has been made?

nautobot/extras/tests/test_customfields.py
gsnider2195: > This is surprising to me that calling get_for_model doesn't populate the cache, only evaluating the queryset does

I'm just evaluating the queryset so that I can see if the cache was used but I *might* be able to just say `qs1 is qs2` to determine if the cache is being used?

nautobot/extras/tests/test_customfields.py
gsnider2195: I actually had this before but removed it because it made the test difficult to read. I'll write another test for that.

nautobot/extras/tests/test_customfields.py
gsnider2195: > Actually I'm not a big fan of checking _result_cache either, would it be possible/better to check the return from qs1.cache_info() instead?

I'm not seeing a `cache_info` anywhere in Django

gsnider2195 (approved):

Issue #3877, opened 2023-06-08
Filter Job Result log entries after the first page


gsnider2195:
### As ...

Nelly - Network Engineer

### I want ...

To be able to filter job result "Logs" table and see all entries that match the filter, not just those on the first page.

### So that ...

I can easily find job log entries matching specific criteria.

### I know this is done when...

I can enter a search term on the job result page and have results returned from pages 2+.

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

None

### External Dependencies

None

## Prototype

https://github.com/nautobot/nautobot/commit/f2eed8ba596fa10a41a14956b691c72d95f74eb1

gsnider2195:
@lampwins can we look into bringing this in to the product backlog?

gsnider2195:
Prototype will need to be rewritten in vanilla js

Issue #1289, opened 2022-01-21
Jobs result live pagination resets to page 1


lampwins:
### Environment
* Python version: 3.8.6
* Nautobot version: 1.2.4

### Steps to Reproduce
1. Have a job that produces a lot of continuous logs and runs for several minutes
2. Run the job
3. In the results page, wait for more than 10 pages of log results to populate
4. Click on page 8

### Expected Behavior
The result set should remain on the user-selected log results page.

### Observed Behavior
On the next log poll cycle, the user is redirected back to page 1 of the log results.

bryanculver:
Include the query string when auto refreshing the page.

PR #5029, opened 2024-01-02
Added global filtering to Job Result log table

timizuoebideri1:
# Closes: #3877 
# What's Changed

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/tests/integration/test_jobs.py
```diff
@@ -82,14 +85,16 @@ def get_cell_value(row, column):
         # Test for log level (one row should be visible)
         filter_element.fill("")
         filter_element.type(log_entries[3].log_level.title())
+        time.sleep(1) # Wait for API call to return a response
```
gsnider2195, line 88: Instead of using sleep are you able to use splinter's [`is_element_present`](https://splinter.readthedocs.io/en/stable/matchers.html#matchers) with a `wait_time` kwarg?

nautobot/extras/views.py
```diff
@@ -1559,7 +1559,12 @@ class JobLogEntryTableView(View):
 
     def get(self, request, pk=None):
         instance = self.queryset.get(pk=pk)
-        log_table = tables.JobLogEntryTable(data=instance.job_log_entries.all(), user=request.user)
+        filter_q = request.GET.get("q")
+        if request.GET.get("q", None):
```
gsnider2195, line 1563:
```suggestion
        if filter_q:
```

nautobot/project-static/js/log_level_filtering.js
```diff
@@ -1,30 +1,6 @@
...
-        let message = $(log).find('td').eq(messageColumn).text();
-
-        // Perform a regex test on the log level and message column text
-        if (filter.test(logLevel) || filter.test(message)) {
-            $(log).show();
-        } else {
-            $(log).hide();
-        }
-    }
+document.querySelector('input.log-filter').addEventListener('input', function() {
```
gsnider2195, line 2: How difficult would it be to rewrite this to vanilla js instead of jquery?

timizuoebideri1: This is vanilla javascript

nautobot/extras/views.py
```diff
@@ -1559,7 +1559,12 @@ class JobLogEntryTableView(View):
 
     def get(self, request, pk=None):
         instance = self.queryset.get(pk=pk)
-        log_table = tables.JobLogEntryTable(data=instance.job_log_entries.all(), user=request.user)
+        filter_q = request.GET.get("q")
+        if filter_q:
+            queryset = instance.job_log_entries.filter(Q(message__iregex=filter_q) | Q(log_level__iregex=filter_q))
```
glennmatthews, line 1564: I'm concerned about the potential for a ReDoS now that we're doing the filtering server-side rather than client-side. I wonder if it would be acceptable to reduce this down to simple `icontains` checks instead of `iregex`?

nautobot/project-static/js/job_result.js
```diff
@@ -23,20 +23,22 @@ function updatePendingStatusLabel(status) {
...
 }
 
-function updateLogTable(result_id) {
+function updateLogTable(result_id, qs) {
     // Calls `update_log_table` to refresh the jobs table from the `/log-table/` endpoint
     // Grab the query string from session storage and pass it through to the log table.
-    let qs = window.sessionStorage.getItem(session_key) || "";
+    if (!qs){
+        qs = window.sessionStorage.getItem(session_key) || "";
+    }
```
glennmatthews, line 31: This feels wrong given that the sessionStorage was introduced to save what page of results were being displayed (#1289). With this change, are we regressing on that issue? (I.e. if we're filtering logs *and* viewing page N of results, are both the filter and the page number being respected?

glennmatthews:
I see this is targeted at `next` since #3877 was marked as a feature - personally I'd argue that this should be considered a bug and fixed in `develop`. Thoughts?

nautobot/extras/views.py
gsnider2195: I have the same concern but that would be a loss of functionality. Maybe we could add a delay so it's not sending requests on every single keystroke?

glennmatthews: I'm OK with losing functionality here personally.

nautobot/project-static/js/job_result.js
timizuoebideri1: On filtering it resets pagination, IMO it makes no sense to remain on the current pagination page when filtering cause we are not just filtering that page alone.
But after filtering is applied switching between filtered pages works well

nautobot/extras/views.py
gsnider2195: I'm concerned that users might rely on this heavily and this will negatively impact one of our core automation features.

timizuoebideri1: @lampwins

gsnider2195: As discussed, since this previously only worked on the first page of results it wasn't that useful anyway so we're going to change to icontains.

nautobot/project-static/js/log_level_filtering.js
glennmatthews, line 1: Wonder why we even made this a separate file from `job_result.js` in the first place. Ah well.

glennmatthews (approved):

nautobot/extras/tests/integration/test_jobs.py
```diff
@@ -71,9 +73,15 @@ def get_cell_value(row, column):
         # Sanity check
         self.assertEqual(len(LogLevelChoices.values()), len(visible_rows()))
 
+        # Trial 1: Give selenium some time to attach event listeners
```
timizuoebideri1, line 76:
```suggestion
        # Give selenium some time to attach event listeners
```

glennmatthews:
Failing in CI again - something fragile/timing-dependent still?

bintangf:
Hi, I would like to backport the changes from this PR to the `ltm-1.6` branch as we also need this feature/fix there. Is there any objection or a specific process I should follow before submitting a backport PR?

Issue #5012, opened 2023-12-21
ObjectChange missing indexes for common use cases


lampwins:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.5
* Python version:
* Database platform, version: mysql
* Middleware(s):

### Steps to Reproduce
1. Have a very large change log table across all object types
2. Query the changelog for common use cases:

--- 
1. Device detail object change view
[This view](https://github.com/nautobot/nautobot/blob/84fba7bf2c020658878be9961c258cc70316cad6/nautobot/extras/views.py#L1618C7-L1618C26) uses this query:
https://github.com/nautobot/nautobot/blob/84fba7bf2c020658878be9961c258cc70316cad6/nautobot/extras/views.py#L1636-L1641

Which uses a combination of (`changed_object_type`,`changed_object_id`) and (`related_object_type`, `related_object_id`), however those combinations are not indexed together. They do exist with other combinations of fields, but we need an index that the query in this view can use. This should be two new indexes for the pairs.

2. Lookup by user
Both the user FK and the `user_name` fields should be indexes. Today only the user FK is indexed.

3. Lookup by user & object
There should be a triple index on `user`, `changed_object_type`, and `changed_object_id`. Likewise, the same triple should exist for `user_name`.

lampwins:
Will target 2.1.X

glennmatthews:
Implemented in #5041

PR #5041, opened 2024-01-03
Added database indexes to the ObjectChange model

timizuoebideri1:
# Closes: #5012 
# What's Changed

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/models/change_logging.py
```diff
@@ -133,6 +133,22 @@ class Meta:
                 name="extras_objectchange_rtime_idx",
                 fields=["-time"],
             ),
+            models.Index(
+                name="changed_object_double_idx",
```
timizuoebideri1, line 137: The 30 character limit prevents me from using the field naming standard used above, i.e., `extras_objectchange_....`

changes/5012.added
```diff
@@ -0,0 +1,7 @@
+Added database indexes to the ObjectChange model:
+
+- Added index on field `user_name`
+- Added `changed_object_double_idx` index on fields: `changed_object_type`, `changed_object_id`
+- Added `related_object_double_idx` index on fields: `related_object_type`, `related_object_id`
+- Added `user_changed_object_triple_idx` index on fields: `user`, `changed_object_type`, `changed_object_id`
+- Added `name_changed_object_triple_idx` index on fields: `user_name`, `changed_object_type`, `changed_object_id`
```
glennmatthews, line 7: Reminder that towncrier will automatically convert each line of the change fragment to a distinct bullet list item - multi-line formatting in the change fragment will almost certainly not be formatted the way you might expect. 

Additionally, remember that the change log is primarily meant to be user-facing, so I think this is probably more detail than we need.

I'd suggest:

```suggestion
Added database indexes to the ObjectChange model to improve performance when filtering by `user_name`, `changed_object`, or `related_object`, and also by `changed_object` in combination with `user` or `user_name`.
```

nautobot/extras/models/change_logging.py
glennmatthews: Personally I don't think that `double` or `triple` are particularly useful terms to include in these names, though I'm not a big fan of including `extras_objectchange` in the name either. I might suggest for the new indices:

- `changed_object_idx`
- `related_object_idx`
- `user_changed_object_idx`
- `user_name_changed_object_idx`

nautobot/extras/models/change_logging.py
```diff
@@ -133,6 +133,22 @@ class Meta:
...
+                fields=["related_object_type", "related_object_id"],
+            ),
+            models.Index(
+                name="user_changed_object_triple_idx",
+                fields=["user", "changed_object_type", "changed_object_id"],
+            ),
+            models.Index(
+                name="name_changed_object_triple_idx",
+                fields=["user_name", "changed_object_type", "changed_object_id"],
+            ),
```
glennmatthews, line 151: Do we need/want a combo index on `user/related_object` and `user_name/related_object` as well? I know it wasn't called out in #5012 but it "feels" like a gap to me. @lampwins any input?

lampwins: Maybe. The related objects are very rarely used and thus even less accessed. I'd be fine leaving them out for now, unless you guys really want to add them.

glennmatthews (approved):
LGTM

PR #5039, opened 2024-01-02
Add more ruff rules, remove flake8 and black

glennmatthews:
# Closes: n/a
# What's Changed

- Updated `ruff` development dependency to `~0.1.10` (latest as of today)
- Add `ruff format` to `invoke ruff` task (before we were only doing `ruff check`; `ruff format` replaces `black`)
- Removed `black` and `flake8` as development dependencies as they're fully replaced by `ruff` now.
- Removed `black` and `flake8` steps from CI.
- Enabled `DJ` Ruff rules (`flake8-django`) and addressed all warnings raised.
  - Some of these resulted in me adding TODO comments as the correct resolution wasn't immediately obvious to me
- Enabled `PIE` Ruff rules except for `PIE808` (`flake8-pie`) and addressed all warnings raised.
  - PIE808 warns about doing `range(0, n)` as it's technically redundant with simply `range(n)`, but personally I prefer the clarity of the former style.
- Enabled `RUF` Ruff native rules except for `RUF012` (which results in a lot of noise, see also https://github.com/astral-sh/ruff/issues/5243) and addressed all warnings raised.
- Enabled remaining `S` Ruff rules (`flake8-bandit`) and addressed all warnings raised.

If this gets accepted and merged, I can open a sequel PR to remove `isort` (which is included as a dev dependency, but not actually run by our CI presently) and enable its equivalent rules in ruff; I've already done the work locally but it results in a lot more diffs and I wanted to keep this PR from being **too** unmanageable to review. 😀 

# TODO

- [x] Explanation of Change(s)
- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- n/a Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [x] Documentation Updates (when adding/changing features)
- [x] Example Plugin Updates (when adding/changing features)
- [x] Outline Remaining Work, Constraints from Design

PR #5048, opened 2024-01-03
Remove alpha UI from the main code base for now

### LINKED ISSUE #5033: Move alpha UI to prototype branch
As a core team member, I want the alpha UI code and infrastructure to be moved to a prototype branch, so that we can maintain the code, but remove the issues it causes with dependancies, etc.

glennmatthews:
# Closes #5033
# What's Changed

Removed most of the alpha UI from the main code base for now (it still exists in prototype branch `prototype/5033-alpha-ui`) to reduce the burden of maintaining its dependencies in the meantime.

Some candidates for related removal *not* currently addressed by this PR:

- `get_all_new_ui_ready_routes()`, `get_only_new_ui_ready_routes()`, `is_route_new_ui_ready()` -- these were (mistakenly IMHO) exported to the `nautobot.apps.utils` namespace and so removing them would technically be a breaking change
- `registry` entries `"new_ui_nav_menu"` and `"new_ui_ready_routes"` and associated code
- the `use_new_ui` flag on various view classes
- dependency on `drf-react-template-framework` and the associated `schema` generation code.
- probably others as well...

This **will** conflict with #5039 to some extent (removing prettier and eslint vs removing black and flake8).

# TODO

- [x] Unit, Integration Tests


gsnider2195:
Just a thought, if we decide to bring this back in it'll probably be easier to just revert the commit rather than trying to use the outdated branch.

glennmatthews:
> Just a thought, if we decide to bring this back in it'll probably be easier to just revert the commit rather than trying to use the outdated branch.

Agreed 100%.

nautobot.code-workspace
```diff
@@ -7,16 +7,10 @@
...
 		]
 	},
 	"settings": {
-		"eslint.format.enable": true,
-		"editor.codeActionsOnSave": {
-			"source.fixAll.eslint": true
-		},
 		"python.formatting.provider": "black",
 		"python.linting.enabled": true,
 		"python.linting.flake8Enabled": true,
```
gsnider2195, line 16: Looks like we missed some of these in the ruff PR. If you want to fix it here, it should look like this:

```json
    "[python]": {
        "editor.formatOnSave": true,
        "editor.defaultFormatter": "charliermarsh.ruff"
    },
```

nautobot.code-workspace
```diff
@@ -26,9 +20,6 @@
 		"[javascript]": {
 			"editor.formatOnSave": true
 		},
-		"[javascriptreact]": {
-			"editor.formatOnSave": true
-		},
 		"[jsx-tags]": {
 			"editor.formatOnSave": true
 		}
```
gsnider2195, line 25: Do we still need the `jsx-tags` section? I don't think this is even valid because `jsx-tags` is an embedded language inside javascript and javascriptreact.

nautobot/core/api/urls.py
```diff
@@ -36,10 +34,8 @@
 ui_api_patterns = [
     # Lookup Expr
     path("core/", include((core_api_patterns, "core-api"))),
-    path("new-ui-ready-routes/", NewUIReadyRoutesAPIView.as_view(), name="new-ui-ready-routes"),
     path("get-menu/", GetMenuAPIView.as_view(), name="get-menu"),
     path("get-object-counts/", GetObjectCountsView.as_view(), name="get-object-counts"),
```
gsnider2195, line 38: I think `GetObjectCountsView` was also implemented specifically for the React UI

glennmatthews: Agreed; I'd left `GetMenuAPIView` and `GetObjectCountsView` in place for now because I was thinking that we could update the current nav-menu rendering and home-page population to be API-driven at some point even in the current UI paradigm, in which case these would become relevant again. Happy to rip them out for now if you think that'd be cleaner though.

gsnider2195: Nope that sounds like a good reason to leave them in

gsnider2195 (approved):

bintangf:
Do you have any plans to modify or remove this documentation?

https://docs.nautobot.com/projects/core/en/stable/release-notes/version-2.0/?h=build_ui#alpha-ui-20-3142
https://github.com/nautobot/nautobot/blob/f30d16061c41c4b47d85381fdb620c1b1796bafc/nautobot/docs/release-notes/version-2.0.md?plain=1#L19

Because I tried it on Nautobot v2.2.3 but the build_ui command is missing as per PR #5048.

PR #5054, opened 2024-01-04
Improve IPAddressEditView and IPAddressAssignView logic

glennmatthews:
# Closes: #n/a
# What's Changed

- Snyk flagged a few more cases where we weren't validating the `return_url` query parameter; I've fixed those here.
- In testing the fix, I noticed that an invalid UUID (as opposed to a valid but nonexistent one) for the `interface` or `vm_interface` query parameter was raising a ValidationError that got bubbled up to the user; I added logic so that it gets handled gracefully now.
- In the "Assign an IP Address" view, we weren't actually *using* the `q` query parameter to search for relevant IP addresses; I've added the appropriate `search_string()` call on the queryset now. 

I don't *believe* any of this needs back-porting to ltm-1.6, as the code in question was added in 2.0 only via #4224.

# TODO

- n/a Unit, Integration Tests


nautobot/ipam/utils/__init__.py
```diff
@@ -271,5 +273,5 @@ def retrieve_interface_or_vminterface_from_request(request):
     try:
         obj = interface_model.objects.restrict(request.user, "change").get(id=interface_id)
         return obj, None
-    except interface_model.DoesNotExist:
+    except (interface_model.DoesNotExist, ValidationError):
```
HanlinMiao, line 276: How do we trigger a ValidationError here and do we need a test case for it?

nautobot/core/views/mixins.py
```diff
@@ -193,6 +193,9 @@ def get_return_url(self, request, obj=None):
             # Model has no get_absolute_url() method or no reverse match
             pass
 
+        if default_return_url is not None:
```
HanlinMiao, line 196: Is there a case where `default_return_url` is not None?

HanlinMiao (approved):

nautobot/ipam/utils/__init__.py
glennmatthews: I did it manually by passing in an invalid (non-UUID) `interface` or `vminterface` query parameter (`?interface=123foobar`). I don't think it's important enough to need a test case for, personally.

nautobot/core/views/mixins.py
glennmatthews: https://github.com/nautobot/nautobot/pull/5054/files#diff-caf8a412a9f30295cacfae888c0fbc1bdc2a7cd2643df55f444414f363113a7bR790 - I didn't want to change the default return url at the view level here, just for this one specific case.

timizuoebideri1 (approved):

PR #5055, opened 2024-01-04
Remove isort, enable ruff's isort rules and apply them

glennmatthews:
# Closes: #n/a
# What's Changed

- Remove `isort` dev dependency (was never automatically used or enforced) - note that it actually stays in poetry.lock because it's a transitive dependency of `pylint`, but we can't do anything about that for now...
- Enable Ruff `I` rule set and configure it in `pyproject.toml` to most closely approximate our current conventions (**I'm absolutely open to suggested changes here**, see https://docs.astral.sh/ruff/settings/#isort for options)
- In the example_plugin, configure `nautobot` as its own group of imports to get placed *before* imports from within the example plugin, as this seems to match our current convention most closely.
- Run `invoke ruff --fix` and let it rearrange imports to its heart's content.
   - Due to some known circular dependencies between `nautobot.core.models` and `nautobot.extras.models`, I had to add `# isort:skip` to prevent a very small number of changes in `nautobot.circuits.models` and `nautobot.dcim.models.cables` that would have triggered circular import errors if moved.

# TODO

- n/a Unit, Integration Tests


HanlinMiao (approved):
In YOLOs we trust. The dependency changes look good, but I only looked through a quarter of the change incurred due to the applications of isort rules.

examples/example_plugin/pyproject.toml
```diff
@@ -16,3 +16,14 @@ python = ">=3.8,<3.12"
...
 requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
+
+[tool.ruff]
+line-length = 120
+target-version = "py38"
+
+[tool.ruff.lint.isort]
+known-first-party = ["example_plugin"]
+section-order = ["future", "standard-library", "nautobot", "first-party", "local-folder", "third-party"]
```
itdependsnetworks, line 26: any reason the example is different than nautobot core?

glennmatthews: The `section-order` and `sections` below were to force imports from `nautobot.*` to be their own section in the imports listing. Probably would be good to add the other settings from core here as well though. Thanks!

glennmatthews: This is actually wrong - will fix :)

examples/example_plugin/pyproject.toml
```diff
@@ -16,3 +16,33 @@ python = ">=3.8,<3.12"
...
+]
+
+[tool.ruff.lint.isort]
+combine-as-imports = true
+force-sort-within-sections = true
+order-by-type = false
+section-order = ["future", "standard-library", "third-party", "nautobot", "first-party", "local-folder"]
+
+[tool.ruff.lint.isort.sections]
+"nautobot" = ["nautobot"]
```
glennmatthews, line 48: No idea why GitHub is considering this invalid - as far as I can tell it's correct TOML?

itdependsnetworks: Github does not like a comment on the last line of the list within toml... seems like a github bug, will see if I can open an issue. 

image

itdependsnetworks: Opened github support ticket 2514042 on it

gsnider2195:
Can you update `docs/development/core/style-guide.md` `## Importing Python Packages` section?

glennmatthews:
> Can you update `docs/development/core/style-guide.md` `## Importing Python Packages` section?

Oof, I'd forgotten that we had such detailed prescriptivism in the docs that isn't currently enforced by any tooling. At a quick glance, I don't see exactly what needs to be updated for this PR though - any specific pointers?

gsnider2195:
> > Can you update `docs/development/core/style-guide.md` `## Importing Python Packages` section?
> 
> Oof, I'd forgotten that we had such detailed prescriptivism in the docs that isn't currently enforced by any tooling. At a quick glance, I don't see exactly what needs to be updated for this PR though - any specific pointers?

I was thinking just an annotation on how to use ruff to format your imports somewhere near the top.

glennmatthews:
Added a note to the docs and improved the `invoke ruff` task based on John's feedback.

gsnider2195 (approved):

Issue #5067, opened 2024-01-08
ExternalIntegration missing Q filter for API


whitej6:
### Environment
* Nautobot version (Docker tag too if applicable): 2.1.0
* Python version: 3.11
* Database platform, version: NA
* Middleware(s):

### Steps to Reproduce
1. Create job using ObjectVar pointing ExternalIntegration
2. Attempt to select an existing ExternalINtegration on the job form
3.

### Expected Behavior
API call returns 200 with paginated instances

### Observed Behavior
API call returns 400 with an error similar to `q filter unknown`

PR #5068, opened 2024-01-08
Add generic test cases for "q" filters.

whitej6:
# Closes: #5067 
# What's Changed

Implemented missing q filter for ExternalIntegration causing 400 response of form fields sourcing from DynamicModelChoiceField.

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/filters/__init__.py
```diff
@@ -493,6 +493,12 @@ class Meta:
 
 
 class ExternalIntegrationFilterSet(NautobotFilterSet):
+    q = SearchFilter(
```
glennmatthews, line 496: Would be good to add a test case in `nautobot.extras.tests.test_filters` to exercise this.

changes/5067.fixed
```diff
@@ -0,0 +1 @@
+ExternalIntegration missing q search filter causing failure when using DynamicModelChoiceField OR ObjectVar.
```
glennmatthews, line 1: Perhaps:

```suggestion
Fixed missing `q` (search) filter for ExternalIntegration model.
```

nautobot/extras/filters/__init__.py
whitej6: Implemented 2 tests, one to always check if q filter is implemented (I would prefer to find something to skip based on an existing attr instead of hard coding a skip) and a second that validates icontains & iexact by dynamically appending a random string. Not sure I am terribly thrilled with either test but did bubble up some other places to update. Happy to rewrite if you have a suggestion.

nautobot/core/testing/filters.py
```diff
@@ -58,6 +59,14 @@ class FilterTestCase(BaseFilterTestCase):
         queryset = None
         filterset = None
 
+        # dictionary of attributes and lookup expression implemented in q filter
+        # meant to skip attributes that use icontains but cannot append a random string to attribute
```
glennmatthews, line 63: I'm not sure I understand what this comment means - can you perhaps clarify?

Would probably also be good to document that if you're using a `SearchFilter` you probably don't need to define this?

nautobot/core/testing/filters.py
```diff
@@ -155,6 +170,40 @@ def test_tags_filter(self):
...
+            if not self.filterset.declared_filters.get("q"):
+                self.skipTest("`q` filter not implemented")
+
+            obj = self.queryset.first()
+            for attr, lookup_method in self.get_q_filter().items():
+                if lookup_method not in ["icontains", "iexact"]:
+                    # only testing icontains and iexact filter lookups
+                    continue
+                lookup = str(obj.id)
+                if not attr == "id":
```
glennmatthews, line 190: Do we have any models for which `id` **is** part of the q filter? That would seem odd to me.

nautobot/core/testing/filters.py
```diff
@@ -155,6 +170,40 @@ def test_tags_filter(self):
...
+                    if not hasattr(obj, attr):
+                        # only testing direct attrs, does not support nested lookups
+                        continue
```
glennmatthews, line 193: Feels like this could be a potential bug, if the q filter references something that isn't actually an object attribute. Could we instead *specifically* check for nested lookups via something like `"__" in attr`?

nautobot/extras/tests/test_filters.py
```diff
@@ -652,6 +653,9 @@ class GraphQLTestCase(FilterTestCases.NameOnlyFilterTestCase):
     queryset = GraphQLQuery.objects.all()
     filterset = GraphQLQueryFilterSet
 
+    # skip testing query attribute for q filter test
+    q_filter = {"name": "icontains"}
```
glennmatthews, line 657: I don't understand from the comment alone why this is needed here - can you perhaps elaborate?

nautobot/extras/tests/test_filters.py
```diff
@@ -1308,8 +1312,12 @@ def test_peer_id(self):
         params = {"peer_id": [self.devices[2].pk]}
         self.assertEqual(self.filterset(params, self.queryset).qs.count(), 1)
 
+    @skip("Skipping RelationshipAssociation is a through model.")
+    def test_q_filter_exists(self):
+        """Test is not applicable for model."""
+
 
-class RelationshipModelFilterSetTestCase(FilterTestCases.FilterTestCase):
+class RelationshipAssociationFilterSetTestCase(FilterTestCases.FilterTestCase):
```
glennmatthews, line 1320: good spot :)

nautobot/extras/tests/test_filters.py
```diff
@@ -1524,6 +1532,10 @@ def test_regression_distinct_2963(self):
             2,
         )
 
+    @skip("Skipping RelationshipAssociation is a through model.")
```
glennmatthews, line 1535: I think it might be perfectly reasonable to expect through models to also support q filters - thoughts?

nautobot/ipam/tests/test_filters.py
```diff
@@ -59,6 +61,10 @@ class VRFTestCase(FilterTestCases.FilterTestCase, FilterTestCases.TenancyFilterT
     queryset = VRF.objects.all()
     filterset = VRFFilterSet
     tenancy_related_name = "vrfs"
+    q_filter = {
+        "name": "icontains",
+        "description": "icontains",
+    }
```
glennmatthews, line 67: To make sure I understand - this is being declared so that we *don't* test for the `rd` field that is otherwise part of the searchfilter? Why so?

nautobot/ipam/tests/test_filters.py
```diff
@@ -103,6 +109,9 @@ class RouteTargetTestCase(FilterTestCases.FilterTestCase, FilterTestCases.Tenanc
     queryset = RouteTarget.objects.all()
     filterset = RouteTargetFilterSet
     tenancy_related_name = "route_targets"
+    q_filter = {
+        "description": "icontains",
+    }
```
glennmatthews, line 114: Why needed? Can we not change the `name` of a RouteTarget?

nautobot/users/filters.py
```diff
@@ -100,6 +100,11 @@ class Meta:
 
 
 class ObjectPermissionFilterSet(BaseFilterSet):
+    q = SearchFilter(
+        filter_predicates={
+            "name": "icontains",
+        }
+    )
```
glennmatthews, line 107: Can we use the `NameSearchFilterSet` mixin here instead?

nautobot/extras/filters/__init__.py
```diff
@@ -590,6 +596,11 @@ class Meta:
 
 
 class ImageAttachmentFilterSet(BaseFilterSet):
+    q = SearchFilter(
+        filter_predicates={
+            "name": "icontains",
+        },
+    )
```
glennmatthews, line 603: Can we use the `NameSearchFilterSet` mixin here instead?

glennmatthews (commented):
Thanks for the updates!

nautobot/core/testing/filters.py
```diff
@@ -155,6 +170,40 @@ def test_tags_filter(self):
             qs_result = self.queryset.filter(tags=tags[0]).filter(tags=tags[1]).distinct()
             self.assertQuerysetEqualAndNotEmpty(filterset_result, qs_result)
 
+        def test_q_filter_exists(self):
+            """Test the `q` filter exists on a filterset, does not validate the filter works as expected."""
+            self.assertTrue(self.filterset.declared_filters.get("q"))
+
+        def test_q_filter_valid(self):
+            """Test the `q` filter based on attributes in `filter_predicates`."""
+            if not isinstance(self.filterset, SearchFilter):
```
gsnider2195, line 179:
```suggestion
            if not isinstance(self.filterset.declared_filters.get("q"), SearchFilter):
```

nautobot/extras/tests/test_filters.py
gsnider2195: This is actually an almost identical copy of the previous test case. We should really combine the two.

nautobot/core/testing/filters.py
whitej6: When testing it appears that `id` was dynamically getting inserted even when not declared.

glennmatthews: Ah right. SearchFilter always includes an `id: iexact` predicate. I'd forgotten that.

nautobot/ipam/tests/test_filters.py
whitej6: The test's is written to dynamically add a random string to the end of the attr which does not conform to the the data expected.

nautobot/extras/tests/test_filters.py
whitej6: The other field in the q filter is the `query` and the dynamic updates to the fields are not compatible with the graphql query format

nautobot/core/testing/filters.py
whitej6: Do we want to continue on both conditions or replace with `"__" in attr`?

nautobot/extras/tests/test_filters.py
glennmatthews:
```suggestion
    # skip testing "query" attribute for generic q filter test as it's not trivially modifiable
    q_filter = {"name": "icontains"}
```

gsnider2195:
@whitej6 I merged the fix to the ExternalIntegration q filter in #5181 so this PR can be changed to just implementing the unittest and related changes.

nautobot/core/testing/filters.py
```diff
@@ -155,6 +164,76 @@ def test_tags_filter(self):
...
+            field = obj._meta.get_field(filter_field_name)
+
+            if isinstance(field, ManyToOneRel) or isinstance(field, ManyToManyRel):
+                rel_obj = getattr(obj, filter_field_name).first()
+            else:
+                rel_obj = getattr(obj, filter_field_name)
+
+            if not rel_obj:
+                # Raise an error stating no instance found to test on
+                # ?: Should we require an instance to be available or just skip
```
timizuoebideri1, line 188: ?

nautobot/extras/tests/test_filters.py
timizuoebideri1: What should be the filter predicate for `RelationshipAssociation` q filter?

nautobot/dcim/filters/__init__.py
```diff
@@ -1570,7 +1570,7 @@ def filter_virtual_ip(self, queryset, name, value):
         return queryset.filter(virtual_ip__in=ip_queryset)
 
 
-class InterfaceRedundancyGroupAssociationFilterSet(BaseFilterSet, NameSearchFilterSet):
```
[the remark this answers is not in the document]
glennmatthews: Good catch. Any analysis on why this wasn't failing before?

nautobot/extras/tests/test_filters.py
glennmatthews: Mabye something like `relationship__label, relationship__key`?

nautobot/circuits/tests/test_filters.py
```diff
@@ -15,7 +15,7 @@
 from nautobot.extras.models import Role, Status, Tag
 
 
-class ProviderTestCase(FilterTestCases.NameOnlyFilterTestCase):
+class ProviderTestCase(FilterTestCases.FilterTestCase):
```
glennmatthews, line 18: Why this change? Provider has a name field, right?

nautobot/core/testing/filters.py
```diff
@@ -155,6 +165,116 @@ def test_tags_filter(self):
             qs_result = self.queryset.filter(tags=tags[0]).filter(tags=tags[1]).distinct()
             self.assertQuerysetEqualAndNotEmpty(filterset_result, qs_result)
 
+        def test_q_filter_exists(self):
+            """Test the `q` filter exists on a filterset, does not validate the filter works as expected."""
+            self.assertTrue(self.filterset.declared_filters.get("q"))
```
glennmatthews, line 170: Possibly?

```suggestion
            self.assertIn("q", self.filterset.declared_filters)
```

nautobot/core/testing/filters.py
```diff
@@ -155,6 +165,116 @@ def test_tags_filter(self):
...
+
+        def _assert_nested_q_filter(self, obj, model_field_name, filter_field_name, lookup_method):
+            """
+            Asserts the validity of a nested filter in the queryset.
+
+            Args:
+                obj: The object on which the filter is applied.
+                model_field_name: The name of the field in the model.
+                filter_field_name: The name of the field being filtered.
+                lookup_method: The method used for the lookup.
```
glennmatthews, line 186: Mind adding a concrete example of this method being called to the docstring? Same request on `_assert_non_nested_q_filter`.

nautobot/core/testing/filters.py
```diff
@@ -155,6 +165,116 @@ def test_tags_filter(self):
...
+
+            if isinstance(field, (ManyToOneRel, ManyToManyRel, ManyToManyField)):
+                rel_obj = getattr(obj, filter_field_name).first()
+            else:
+                rel_obj = getattr(obj, filter_field_name)
+
+            if "__" not in nested_field:
+                self._assert_non_nested_q_filter(rel_obj, model_field_name, nested_field, lookup_method)
+            else:
+                self._assert_nested_q_filter(rel_obj, model_field_name, nested_field, lookup_method)
```
glennmatthews, line 199: Maybe consider doing this with a `while` loop instead of recursion? Not blocking.

nautobot/core/testing/filters.py
```diff
@@ -155,6 +165,116 @@ def test_tags_filter(self):
...
+                raise ValueError(
+                    f"Cannot find valid test data for {queryset.model.__name__} with params {filter_params}"
+                )
+            return queryset
+
+        def test_q_filter_valid(self):
+            """Test the `q` filter based on attributes in `filter_predicates`."""
+
+            if not isinstance(self.filterset.declared_filters.get("q"), SearchFilter):
+                self.skipTest("`q` filter is not a SearchFilter")
```
glennmatthews, line 254: IMO this should be a failure of the test. Though I recognize that may cause apps to start failing upstream testing.

nautobot/core/testing/filters.py
```diff
@@ -155,6 +165,116 @@ def test_tags_filter(self):
...
+
+            if not self.filterset.declared_filters.get("q"):
+                self.skipTest("`q` filter not implemented")
```
glennmatthews, line 257: Maybe this should be a failure too?

nautobot/extras/tests/test_filters.py
```diff
@@ -1679,6 +1591,10 @@ def test_secret_type(self):
         params = {"secret_type": [SecretsGroupSecretTypeChoices.TYPE_PASSWORD]}
         self.assertEqual(self.filterset(params, self.queryset).qs.count(), 2)
 
+    @skip("Skipping SecretsGroupAssociation is a through model.")
+    def test_q_filter_exists(self):
+        """Test is not applicable for model."""
```
glennmatthews, line 1596: Probably could add a q filter that checks `secrets_group__name`, `secret__name`, etc.?

changes/5067.added
```diff
@@ -0,0 +1 @@
+Added `q` (search) filter to `ProviderFilterSet`, `ProviderNetworkFilterSet`, `ImageAttachmentFilterSet` and `ObjectPermissionFilterSet`.
```
glennmatthews, line 1: Let's document the new generic test cases as an addition as well.

I wonder, should we add the tests in `next` rather than `develop` just given the potential impact to apps?

nautobot/core/testing/filters.py
timizuoebideri1: I don't believe `SearchFilter` would be very usefull in some `FilterSets`, such as `IPAddressFIlterset` and `PrefixFIlterset`.

nautobot/dcim/filters/__init__.py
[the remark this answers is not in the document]
timizuoebideri1: Maybe because there was no generic `q` filter test case

nautobot/core/testing/filters.py
glennmatthews: Fair point, I hadn't thought of that. Is there a subset of `Filter` types that we *could* assert that includes those cases?

changes/5067.added
```diff
@@ -0,0 +1,2 @@
+Added `q` (SearchFilter) filter to all filterset missing it.
+Added two generic test cases for `q` filter`test_q_filter_exists` and `test_q_filter_valid`.
```
glennmatthews, line 2:
```suggestion
Added two generic test cases for `q` filter: `test_q_filter_exists` and `test_q_filter_valid`. 
```

nautobot/core/testing/filters.py
timizuoebideri1: There are currently test cases for each of these `q` fields in their respective Test classes.

Issue #5076, opened 2024-01-09
Nautobot package `packaging` version incompatible with Pynautobot `packaging` version


housepbass:
### Environment
* Nautobot version (Docker tag too if applicable): 2.1.2b1
* Python version: 3.9.17
* Database platform, version: postgres:13-alpine
* Middleware(s):

### Steps to Reproduce
1. Install Nautobot and Pynautobot packages:

```bash
poetry add nautobot 
poetry add pynautobot
```

### Expected Behavior
Both packages install and have compatible versioning.

### Observed Behavior
The Pynautobot package failed to install due to version incompatibility of the `packaging` package.

```bash
Because no versions of pynautobot match >2.0.2,<3.0.0
 and pynautobot (2.0.2) depends on packaging (>=23.2,<24.0), pynautobot (>=2.0.2,<3.0.0) requires packaging (>=23.2,<24.0).
And because nautobot (2.1.1) depends on packaging (>=23.1,<23.2)
 and no versions of nautobot match >2.1.1,<3.0.0, pynautobot (>=2.0.2,<3.0.0) is incompatible with nautobot (>=2.1.1,<3.0.0).
So, because bug-test depends on both nautobot (^2.1.1) and pynautobot (^2.0.2), version solving failed.
```

PR #5077, opened 2024-01-09
bump packaging

housepbass:
# Closes: #5076 
# What's Changed

Bump `packaging` version from `~23.1` to `^23.2`

# TODO

- [ ] Unit, Integration Tests


pyproject.toml
```diff
@@ -113,7 +113,7 @@ netaddr = "~0.8.0"
 #       we permit a looser (^) version constraint here.
 netutils = "^1.6.0"
 # Handling of version numbers
-packaging = "~23.1"
+packaging = "^23.2"
```
glennmatthews, line 116: I'm a bit hesitant over raising the minimum version here as we've had doing so for `packaging` specifically cause conflicts with downstream apps before. Also, I see that `packaging` uses CalVer rather than SemVer, so this will still prevent installation of the next minor release of `packaging` since that will be a `24.x` version number at this point even if there are no breaking changes.

Taking all of that into account, I might suggest  that we just go with `poetry add 'packaging>=23.1'` for now, and *hope* that that package doesn't release any breaking changes.

housepbass: That seems reasonable to me.

glennmatthews (approved):

pyproject.toml
housepbass: @glennmatthews I'm reconsidering whether this is actually an issue. The original "issue" is `packaging` version between pynautobot and nautobot... One could simply run them from different environments rather than installing them in the same environment. Thoughts?

glennmatthews: In general that would be the more common scenario, but I can also see that there might be developer use cases where you want to have both installed and not have to manage two different venvs.

PR #5079, opened 2024-01-09
Fixed bug with invoke cli and invoke nbshell.

gsnider2195:
# Closes: #DNE
# What's Changed

After #5055 when running `invoke cli` or `invoke nbshell` you get this error:

`TypeError: tasks.docker_compose() got multiple values for keyword argument 'pty'`

This fixes that bug.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/5079.fixed
+++ b/changes/5079.fixed
@@ -1,1 +1,0 @@
-Fixed bug with invoke cli and invoke nbshell.
--- a/changes/5079.fixed
+++ b/changes/5079.fixed
@@ -1,0 +1,1 @@
+Fixed bug with invoke cli and invoke nbshell.
--- a/tasks.py
+++ b/tasks.py
@@ -436,7 +436,7 @@ def nbshell(context):
     """Launch an interactive Nautobot shell."""
     command = "nautobot-server nbshell"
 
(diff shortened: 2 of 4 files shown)


tasks.py
```diff
@@ -436,7 +436,7 @@ def nbshell(context):
     """Launch an interactive Nautobot shell."""
     command = "nautobot-server nbshell"
 
-    run_command(context, command, pty=True)
+    run_command(context, command)
```
glennmatthews, line 439: I'd fixed this a different way locally, by changing `run_command`:

```diff
-        return docker_compose(context, compose_command, pty=True, **kwargs)
+        kwargs["pty"] = True
+        return docker_compose(context, compose_command, **kwargs)
```

Thoughts on one versus the other?

changes/5079.fixed
glennmatthews, line 1: Probably should be `.housekeeping`?

tasks.py
gsnider2195: I don't think it matters either way

HanlinMiao (approved):

HanlinMiao:
Seems to be broken in `next` as well, should we port this to `next`?

PR #5087, opened 2024-01-11
update gitpython to 3.1.41 (ltm)

gsnider2195:
Backport of #5083

### DIFF
--- a/nautobot/docs/requirements.txt
+++ b/nautobot/docs/requirements.txt
@@ -1,5 +1,5 @@
 griffe==0.31.0
-Jinja2==3.1.2
+Jinja2==3.1.3
 Markdown==3.3.7
 mkdocs==1.5.3
 mkdocs-gen-files==0.5.0
--- a/poetry.lock
+++ b/poetry.lock
@@ -1303,13 +1303,13 @@ doc = ["Sphinx", "sphinx-rtd-theme", "sphinxcontrib-spelling"]
 
 [[package]]
 name = "faker"
-version = "22.2.0"
+version = "22.1.0"
 description = "Faker is a Python package that generates fake data for you."
 optional = false
 python-versions = ">=3.8"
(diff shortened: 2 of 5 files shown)


poetry.lock
```diff
@@ -2988,54 +2988,54 @@ files = [
 
 [[package]]
 name = "pyuwsgi"
-version = "2.0.23"
+version = "2.0.23.post0"
```
glennmatthews, line 2991: A bit hesitant about this change as there doesn't seem to be much documentation around it. I *think* this is primarily to fix a compatibility issue with Python 3.12

poetry.lock
```diff
@@ -1633,13 +1633,13 @@ colors = ["colorama (>=0.4.6)"]
 
 [[package]]
 name = "jinja2"
-version = "3.1.2"
+version = "3.1.3"
```
glennmatthews, line 1636: Want to also update this in pyproject.toml (#5086 / CVE-2024-22195)?

poetry.lock
gsnider2195: Yeah I forgot to do `--no-update` when I updated the lockfile. Fixed now

nautobot/docs/requirements.txt
```diff
@@ -1,5 +1,5 @@
 griffe==0.31.0
-Jinja2==3.1.2
+Jinja2==3.1.3
```
glennmatthews, line 2: does the example_plugin also need this update or is that 2.x-only?

glennmatthews (approved):

nautobot/docs/requirements.txt
gsnider2195: I missed it. It's updated now

glennmatthews (approved):

whitej6 (approved):

PR #5102, opened 2024-01-12
More job fixes

glennmatthews:
Closes: #n/a
# What's Changed

- Fixed missing log messages when errors occur during `Job.__call__()` initial setup.
- Fixed misleading "Job completed" message from being logged when a Job aborted.
- Fixed an error in `nautobot-server runjob` if a job returned data other than a dict.
- Fixed misleading "SUCCESS" message when `nautobot-server runjob` resulted in any JobResult status other than "FAILED".
- Fixed incorrect JobResult data when using `nautobot-server runjob --local` or `JobResult.execute_job()`.

# TODO

- n/a Unit, Integration Tests


nautobot/extras/models/jobs.py
```diff
@@ -663,12 +663,15 @@ def enqueue_job(
...
-                job_model.job_task.apply(
+                eager_result = job_model.job_task.apply(
                     args=job_args, kwargs=job_kwargs, task_id=str(job_result.id), **job_celery_kwargs
                 )
 
             # copy fields from eager result to job result
             job_result.refresh_from_db()
+            job_result.result = eager_result.result
+            job_result.status = eager_result.status
+            job_result.traceback = eager_result.traceback
```
gsnider2195, line 674: These were removed in #4683 to fix a security vulnerability. This doesn't reintroduce that vulnerability does it?

glennmatthews: Hmm, good catch and good question. I'll admit I don't quite see how that particular change in that PR (https://github.com/nautobot/nautobot/pull/4683/commits/621bf8df9ed020caab3cea75857b396f83bed381)  is directly related to the vulnerability fix. At a glance I feel like we still need to re-add this code but maybe we need to explicitly sanitize the result and traceback as well? Will do some digging.

glennmatthews: Ok, the test added in that PR is failing now, so let me see what I need to do to fix it.

nautobot/core/utils/logging.py
```diff
@@ -10,20 +10,36 @@
...
+    if isinstance(string, (list, tuple)):
+        new_string = []
+        for item in string:
+            if isinstance(item, (list, tuple, bytes, str)):
+                new_string.append(sanitize(item))
+            else:
+                # Pass through anything that isn't a string or iterable of strings
+                new_string.append(item)
+        if isinstance(string, tuple):
+            new_string = tuple(string)
```
gsnider2195, line 28:
```suggestion
            new_string = tuple(new_string)
```

glennmatthews: Good catch. Clearly this function isn't tested very thoroughly at present.

nautobot/core/utils/logging.py
```diff
@@ -10,20 +10,36 @@
 
 def sanitize(string, replacement="(redacted)"):
```
gsnider2195, line 11: Should we rename the input arg to `dirty` or something besides string?

nautobot/core/utils/logging.py
```diff
@@ -10,20 +10,36 @@
...
+        if isinstance(string, tuple):
+            new_string = tuple(string)
+        return new_string
+
+    if isinstance(string, bytes):
+        return sanitize(string.decode("utf-8")).encode("utf-8")
+
+    if isinstance(string, str):
+        # Don't allow regex match groups to be referenced in the replacement string!
+        if re.search(r"\\\d|\\g<\d+>", replacement):
```
gsnider2195, line 36: We should have a test for this case but I don't see any

nautobot/core/utils/logging.py
glennmatthews: I was hesitant to do so just on the off chance that some app might be explicitly calling `sanitize(string=...)` but it may be desirable.

gsnider2195 (approved):

PR #5109, opened 2024-01-16
Remove `files/get` endpoint which is not used

gertzakis:
# Closes: #N/A
# What's Changed

Removes the `files/get` URL endpoint which is not used. Also, the way it was configured without `"add_attachment_headers": True` it was missing the `Content-Disposition: attachment` HTTP header which causes a security vulnerability. 

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/core/tests/test_views.py
+++ b/nautobot/core/tests/test_views.py
@@ -452,10 +452,8 @@ class DBFileStorageViewTestCase(TestCase):
         obj_perm.save()
         obj_perm.users.add(self.user)
         obj_perm.object_types.add(ContentType.objects.get_for_model(FileProxy))
-        with self.subTest(self.url):
-            response = self.client.get(self.url)
-            self.assertHttpStatus(response, 200)
+        response = self.client.get(self.url)
+        self.assertHttpStatus(response, 200)
         url = f"{reverse('db_file_storage.download_file')}?name={self.file_proxy_2.file.name}"
-        with self.subTest(url):
-            response = self.client.get(url)
-            self.assertHttpStatus(response, 404)
+        response = self.client.get(url)
+        self.assertHttpStatus(response, 404)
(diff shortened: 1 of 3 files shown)


nautobot/core/tests/test_views.py
```diff
@@ -424,30 +424,24 @@ def setUp(self):
...
 
     def test_get_object_with_permission(self):
         self.add_permissions(get_permission_for_model(FileProxy, "view"))
-        for url in self.urls:
-            with self.subTest(url):
-                response = self.client.get(url)
-                self.assertHttpStatus(response, 200)
+        with self.subTest(self.url):
+            response = self.client.get(self.url)
+            self.assertHttpStatus(response, 200)
```
glennmatthews, line 444: In each of these, since there's only a single URL being tested now, the `with self.subTest()` is probably redundant and can be removed.

nautobot/core/tests/test_views.py
```diff
@@ -458,14 +452,10 @@ def test_get_object_with_constrained_permission(self):
...
-            with self.subTest(url):
-                response = self.client.get(url)
-                self.assertHttpStatus(response, 404)
+        with self.subTest(self.url):
+            response = self.client.get(self.url)
+            self.assertHttpStatus(response, 200)
+        url = f"{reverse('db_file_storage.download_file')}?name={self.file_proxy_2.file.name}"
+        with self.subTest(url):
+            response = self.client.get(url)
+            self.assertHttpStatus(response, 404)
```
glennmatthews, line 461: Same idea here:

```suggestion
        response = self.client.get(self.url)
        self.assertHttpStatus(response, 200)
        url = f"{reverse('db_file_storage.download_file')}?name={self.file_proxy_2.file.name}"
        response = self.client.get(url)
        self.assertHttpStatus(response, 404)
```

glennmatthews (commented):
Suggest adding a `changes/5109.security` change fragment along the lines of:

```
Removed `/files/get/` URL endpoint (for viewing FileAttachment files in the browser), as it was unused and could potentially pose security issues.
```

nautobot/core/tests/test_views.py
gertzakis: Nice, changed it!

glennmatthews (approved):

glennmatthews:
I hadn't noticed that this was a PR to `ltm-1.6`. This should really go into `develop` first and then be backported to the LTM branch.

gertzakis:
Oh of course! Opened https://github.com/nautobot/nautobot/pull/5115 for adding it into `develop`.

timizuoebideri1 (approved):

PR #5110, opened 2024-01-16
Fix #4075 - sortability of Device Bays list view by installed device status

### LINKED ISSUE #4075: Sort by Status on dcim/device-bays
### Environment
* Nautobot version (Docker tag too if applicable): v1.5.21
* Python version: Python 3.10.12
* Database platform, version: PostgreSQL 13.8
* Middleware(s): None

When accessing Device Bays and sorting, sorting by status does not work.
### Steps to Reproduce
1. Navigate to Devices > Device bays (`dcim/device-bays/?q=&sort=name`)
2. Select "Status" at the top of the page

### Expected Behavior
Sort the device bays by status

### Observed Behavior
Django Error
image

glennmatthews:
# Closes #4075 
# What's Changed

- Replace bespoke template code for rendering statuses in the `DeviceBayTable`, `DeviceDeviceBayTable`, and `InterfaceRedundancyGroupAssociationTable` with our general-purpose `ColoredLabelColumn` class.
  - This does have the minor behavior change that the Device Bay tables now show our standard — marker instead of the special-purpose text "Vacant" in the Status column, but IMHO that's an improvement.
- Rename `status` column to `installed_device__status` in the Device Bay tables, which has no effect on the column header (it remains simply `Status`) but allows for correct lookup and sortability.

image

Sorting by reverse status:

image

# TODO

- n/a Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/tables/devices.py
+++ b/nautobot/dcim/tables/devices.py
@@ -37,15 +37,12 @@ from .template_code import (
     CONSOLESERVERPORT_BUTTONS,
     DEVICE_LINK,
     DEVICEBAY_BUTTONS,
-    DEVICEBAY_STATUS,
     FRONTPORT_BUTTONS,
     INTERFACE_BUTTONS,
     INTERFACE_IPADDRESSES,
     INTERFACE_REDUNDANCY_GROUP_INTERFACES,
     INTERFACE_REDUNDANCY_GROUP_INTERFACES_IPADDRESSES,
-    INTERFACE_REDUNDANCY_GROUP_STATUS,
     INTERFACE_REDUNDANCY_INTERFACE_PRIORITY,
-    INTERFACE_REDUNDANCY_INTERFACE_STATUS,
     INTERFACE_TAGGED_VLANS,
     LINKED_RECORD_COUNT,
     PATHENDPOINT,
@@ -772,7 +769,7 @@ class DeviceRearPortTable(RearPortTable):
 
 
 class DeviceBayTable(DeviceComponentTable):
-    status = tables.TemplateColumn(template_code=DEVICEBAY_STATUS)
+    installed_device__status = ColoredLabelColumn()
     installed_device = tables.Column(linkify=True)
     tags = TagColumn(url_name="dcim:devicebay_list")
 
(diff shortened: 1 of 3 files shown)


nautobot/dcim/tables/devices.py
```diff
@@ -783,7 +780,7 @@ class Meta(DeviceComponentTable.Meta):
             "device",
             "name",
             "label",
-            "status",
+            "installed_device__status",
```
itdependsnetworks, line 783: Is this something that we could build a pylint rule for?

glennmatthews: I'm not sure offhand how we would spell out such a rule, TBH.

itdependsnetworks: Fair enough, if I could bother you a bit more, can you explain what this does? Like the why and what of the PR.

glennmatthews: Instead of manually defining an HTML template for rendering this column (which, like any TemplateColumn, receives `record` as a context variable, and in this case manually evaluates `record.installed_device.status` as part of the template), we now:

1. use a ColoredLabelColumn, which is a Nautobot-specific class that knows how to render Status, Role, and similar objects in a consistent way (obviating the need for a custom HTML template)
2. define the column as being based on `installed_device__status`, which django-tables2 knows to mean that by default it uses that field lookup (i.e. `record.installed_device.status`) to provide the `value` for rendering this column.
    - This is necessary in combination with item 1, because by default a column called `status` would try to derive its value from `record.status`, which isn't a field on a DeviceBay record.
    - This fixes the original reported issue, in that now that this column natively understands its own `value`, django-tables2 can now intelligently sort the table based on this column's `value`.

There are absolutely use cases for TemplateColumn to handle various other advanced scenarios, so a blanket rule of "don't use TemplateColumn" wouldn't be useful. It would have to be something able to detect that:
1. You're using a TemplateColumn whose template is doing a nested lookup to get the actual value of interest
2. Using an appropriate column name (or specifying an `accessor` on the column definition) would let you get the actual value more effectively than doing the lookup in the template code itself.

HanlinMiao (approved):

PR #5052, opened 2024-01-04
Added `locations` field to VLAN.

timizuoebideri1:
# Closes: #4412 
# What's Changed

# TODO

- [ ] Explanation of Change(s)
- [ ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [ ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example Plugin Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #5124, opened 2024-01-18
Fixes #5058: use location instead of location_id in vm forms

### LINKED ISSUE #5058: Bug: VM Interface vlan list in UI uses invalid option
### Environment
* Nautobot version (Docker tag too if applicable): 2.1.0
* Python version: 3.10
* Database platform, version: postgres, 14.10
* Middleware(s):

### Steps to Reproduce
1. Create a VLAN
2. Create a Cluster Type
3. Create a Cluster
4. Create a VM in the cluster
5. Create a interface on the vm
6. Selected Access, Tagged or Tagged (all)

### Expected Behavior
Show a list of vlans

### Observed Behavior
Dropdown shows "The results could not be loaded"

Was able to replicate on [demo.nautobot.com](https://demo.nautobot.com)

It appears that this dropdown tries to use an API query attribute/option called `location_id` however this seems to generate an 400. 

Looking at the API browser with `GET /api/ipam/vlans/?q=&limit=50&offset=0&depth=0&location_id=<<GUID>>` or even `GET /api/ipam/vlans/?q=&limit=50&offset=0&depth=0&location_id=null` returns 

```
HTTP 400 Bad Request
API-Version: 2.1
Allow: GET, POST, PUT, PATCH, DELETE, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "location_id": [
        "Unknown filter field"
    ]
}
```

tlourey:
Fixes #5058

# Closes: #5058 
# What's Changed

Changed filter parameters from location_id to location in virtualization/forms.py.

# TODO

- [ ] Unit, Integration Tests


tlourey:
For whatever reason, I couldn't get either docker or virtualenv working so I wasn't able to do much testing before submitting PR.

tlourey:
@glennmatthews could you please allow checks to run and review? many thanks.

glennmatthews (commented):
Thanks for the PR!

@timizuoebideri1 let's make sure this fix is updated correctly for 2.2 as part of #5052.

changes/5058.fixed
```diff
@@ -0,0 +1 @@
+Changed filter parameters from location_id to location in virtualization/forms.py.
```
HanlinMiao, line 1:
```suggestion
Changed filter query parameters from location_id to location in virtualization/forms.py.
```

HanlinMiao (approved):

tlourey:
I intentionally didn't change those as I _thought_ that query was looking for the primary key of the record that has location location_id.

I'm still getting familiar with the model but saying the above out loud, I think that statement is wrong and I should have changed those as well. 

@glennmatthews do we re open this PR for the next version or just use the same issue and submit a new PR?

glennmatthews:
Good catch and your analysis is spot on @Lennahrt - those also need to use `location` (as do lines 537/538 and 607/608 as well).

@tlourey Another PR against the same issue would be totally fine. Thank you!

Issue #3349, opened 2023-02-22
Scheduled Jobs > Crontab corresponds to UTC


mattmiller87:
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.10
* Python version: 3.10.10
* Database platform, version: Postgres: 13-alpine
* Middleware(s): None

### Steps to Reproduce
1. Change TIME_ZONE to "America/New_York"
2. Schedule a job using "Recurring custom"
3. Use the Crontab to run a job at a specific time each day ( 0 9 * * *) for 9am

### Expected Behavior
The job should be scheduled to run at 9am EST

### Observed Behavior
The job runs at 9am UTC

gsnider2195:
Looks like this could be as simple as

```py
from django.utils import timezone
return schedules.crontab(
    minute=minute,
    hour=hour,
    day_of_month=day_of_month,
    month_of_year=month_of_year,
    day_of_week=day_of_week,
    nowfun=timezone.now,
)
```

gsnider2195:
@mattmiller87 how are you setting `TIME_ZONE` in your environment? Are you using the same method for your Nautobot container and your celery beat container?

lampwins:
Let's document all the places where the timezone settings come into play and a note that we recommend UTC for consistency.

itdependsnetworks:
> Let's document all the places where the timezone settings come into play and a note that we recommend UTC for consistency.

Just off the top of my head
* system timezone on:
    * Nautobot
    * Nautobot Worker
    * Redis
    * Database
    * Celery Beat
* Application Level
    * nautobot-server: TIME_ZONE variable
    * psql "SET TIME ZONE 'UTC';"
    * mysql my.cnf "default-time-zone='+00:00'"

lampwins:
This sounds like it will be deep inside of beat. If there is no easy solution here, I am inclined to document this as an edge case and say that if you change the system timezone, you need to update your scheduled jobs, because changing the system timezone is just such a rare thing to do in the first place.

gsnider2195:
Spoke with @mattmiller87 today. It looks like an old cookie was missing the volume mapping `nautobot_config.py` for the celery beat service so the test deployment that he was using was experiencing the issue where beat's clock was not using the same time zone as nautobot. Will use this issue to update the docs to state that you must mount your nautobot_config.py into your celery beat and celery worker and ensure the time on all of those servers is correct.

I'm going to do one more test to simulate a distributed deployment to see how beat handles that:
[...]

PR #5125, opened 2024-01-18
Fixed some recurring scheduled jobs not using configured time zone

gsnider2195:
# Closes: #3349
# What's Changed

The root cause of this issue is `celery.schedulers.crontab` is trying to compare the `.hour` of two datetime objects:
- `nautobot.core.celery.app.now()` which is in the current tz
- `ScheduledJob.last_run_at` which is stored in the database as utc
 
`.hour` is going to be different for objects in different timezones even when the normalized time is the same. This is actually fixed in [django-celery-beat](https://github.com/celery/django-celery-beat/blob/00f0a5c5b2ea4cff100e3c0c9013f4dc891db3e7/django_celery_beat/schedulers.py#L132C1-L136C55) but we didn't bring this logic over when we created `ScheduledJob`.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/tests/test_jobs.py
+++ b/nautobot/extras/tests/test_jobs.py
@@ -1055,8 +1055,11 @@ class ScheduledJobIntervalTestCase(TestCase):
         with mock.patch("django.utils.timezone.now") as mock_now:
             mock_now.return_value = self._strptime("2 Jan 2024 8:00AM")
             is_due, _ = scheduled_job.schedule.is_due(scheduled_job.last_run_at)
+            mock_now.assert_called()
             self.assertFalse(is_due)
 
+            mock_now.reset_mock()
             mock_now.return_value = self._strptime("2 Jan 2024 10:00AM")
             is_due, _ = scheduled_job.schedule.is_due(scheduled_job.last_run_at)
+            mock_now.assert_called()
             self.assertTrue(is_due)
--- a/nautobot/extras/tests/test_jobs.py
+++ b/nautobot/extras/tests/test_jobs.py
@@ -1036,10 +1036,7 @@ class ScheduledJobIntervalTestCase(TestCase):
         scheduled_job_weekday = self.cron_days[schedule_day_of_week]
         self.assertEqual(scheduled_job_weekday, requested_weekday)
 
-
-@override_settings(USE_TZ=True, TIME_ZONE="America/New_York")
-class ScheduledJobCronTestCase(TestCase):
-    @timezone.override("America/New_York")
+    @override_settings(USE_TZ=True, TIME_ZONE="America/New_York")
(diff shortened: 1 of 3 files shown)


nautobot/extras/tests/test_jobs.py
```diff
@@ -1019,6 +1020,11 @@ class ScheduledJobIntervalTestCase(TestCase):
     # datetime weekday starts on Monday (Sunday = 6)
     datetime_days = ["Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday", "Sunday"]
 
+    def _strptime(self, time_string):
+        """Return a utc datetime object from a string with the format: 01 Jan 2024 09:00AM"""
+        time_format = r"%d %b %Y %I:%M%p"
+        return datetime.datetime.strptime(time_string, time_format).replace(tzinfo=timezone.utc)
```
glennmatthews, line 1026: Is the `.replace` converting the datetime *from* an initial `America/New_York` timezone to UTC? Trying to understand why we're using UTC here instead of `America/New_York`.

gsnider2195: We're using utc because that's what `django.utils.timezone.now` returns, which is required to fix this issue. The default is to use celery's `app.now` which returns a datetime in the local timezone. The `.replace` is adding the timezone without datetime trying to convert it from the system's local time to utc.

nautobot/extras/tests/test_jobs.py
```diff
@@ -1033,3 +1039,27 @@ def test_weekly_interval(self):
...
+    def test_cron_time_zone(self):
+        self.assertEqual(app.timezone.key, "America/New_York")
+        scheduled_job = models.ScheduledJob.objects.create(
+            name="test_cron_time_zone",
+            task="pass.TestPass",
+            interval=JobExecutionType.TYPE_CUSTOM,
+            user=self.user,
+            start_time=self._strptime("1 Jan 2024 12:00AM"),
+            last_run_at=self._strptime("1 Jan 2024 9:00AM"),
+            crontab="0 9 * * *",
```
glennmatthews, line 1053: Apologies for ignorant questions, but with `TIME_ZONE="America/New_York"`, is this crontab 0900 UTC or 0900 EST? I thought that per the original issue the user was expecting this to be evaluated as 0900 EST, but if I understand your previous reply, the `_strptime` function is evaluating the provided strings as UTC datetimes, not EST?

What part am I misunderstanding? :smile:

gsnider2195: Yeah this isn't fixing the problem it's just making it consistently wrong. Instead of changing `now` to `utc` we need to change `last_run_at` and `start_time` to the local time (which is how they fixed it in django celery beat)

PR #5128, opened 2024-01-18
Fixed Sync Git Repository requires non-matching permissions for UI vs API

### LINKED ISSUE #4444: Sync Git Repository requires different permissions for UI vs API
### Environment
* Nautobot version (Docker tag too if applicable): v1.6.2b1
* Python version: 3.11
* Database platform, version: pg15
* Middleware(s): n/a

From the API it requires “add” and “change” permissions for “extras > Git repository” to sync git repositories. However from the UI a user with only the “change” permission can perform a sync git repository operation
### Steps to Reproduce
For UI
1. Create object permission for `extras > git repository`
2. Create one with "can change" permission item checked ("can view" as well, but not related to reproducing issue)
4. Assign to Group and assign that to User
5. You can perform sync from UI

For API
1. Create object permission for `extras > git repository`
2. Create one with "can change" permission item checked ("can view" as well, but not related to reproducing issue)
4. Assign to Group and assign that to User
5. You cannot perform sync from API
6. Edit permission for the group, adding "can add"
7. You can now perform the sync from the API

### Expected Behavior
Required permissions to be the same for action

### Observed Behavior
Required permissions were not the same for action

HanlinMiao:
# Closes #4444
# What's Changed

Since syncing git repositories in the UI only requires `extras:change_gitrepository` whereas `extras:add_gitrepository` is needed in the API in addition to `extras:change_gitrepository`.
So what I did was:
Since we are explicitly checking for `extras:change_gitrepository` in the API `sync()` method, I explicitly set the `permission_classes` to `IsAuthenticated` in the `@action` decorator, bypassing the default DRF permission check for `extras:add_gitrepository` and the permission check fall through to the function itself.

# Screenshots

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/api/views.py
+++ b/nautobot/extras/api/views.py
@@ -14,7 +14,7 @@ from rest_framework import mixins, status, viewsets
 from rest_framework.decorators import action
 from rest_framework.exceptions import MethodNotAllowed, PermissionDenied, ValidationError
 from rest_framework.parsers import JSONParser, MultiPartParser
-from rest_framework.permissions import IsAuthenticated
+from rest_framework.permissions import AllowAny, IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.routers import APIRootView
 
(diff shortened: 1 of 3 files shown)


nautobot/extras/api/views.py
```diff
@@ -375,7 +375,7 @@ class GitRepositoryViewSet(NautobotModelViewSet):
     filterset_class = filters.GitRepositoryFilterSet
 
     @extend_schema(methods=["post"], request=serializers.GitRepositorySerializer)
-    @action(detail=True, methods=["post"])
+    @action(detail=True, methods=["post"], permission_classes=[IsAuthenticated])
```
glennmatthews, line 378: Would be good to add an explanatory comment about the permission_classes, I think?

glennmatthews (approved):

PR #5130, opened 2024-01-18
Add experimental `--parallel` option to `invoke unittest`

glennmatthews:
# What's Changed

- Add option to run `invoke unittest` with Django's native `--parallel` flag
- Rework our custom test runner by, instead of calling `super().setup_databases()` and `super().teardown_databases()`, instead vendoring the entire Django 3.2 implementation of these two methods so that we can inject our factory setup and cleanup at the appropriate point in these two functions to work properly when `--parallel` is in use.

In my local testing, most *but not all* test cases still work with `--parallel`; we'll need to chase down the remaining failures (mostly in `nautobot.core.tests.test_graphql`, but not exclusively there) before we can enable this in CI. However I think it's worth adding as an experimental option even in its current state so that others can try it out.

# Screenshots

```
Ran 7525 tests in 737.626s

FAILED (failures=9, errors=2, skipped=173, expected failures=1)
```

# TODO

- [x] Unit, Integration Tests - investigate failures in parallel execution and figure out what we need to do to fix (interim solution *might* be to add `SerializeMixin` to certain test cases as a last resort...)


nautobot/core/tests/runner.py
```diff
@@ -55,31 +57,79 @@ def setup_test_environment(self, **kwargs):
...
-            if self.cache_test_fixtures:
-                command += ["--cache-test-fixtures"]
-            for connection in result:
-                db_name = connection[0].alias
-                print(f'Pre-populating test database "{db_name}" with factory data...')
-                db_command = [*command, "--database", db_name]
-                call_command(*db_command)
-
-        return result
+        time_keeper = self.time_keeper
```
HanlinMiao, line 60: Does vendoring django's methods here mean more work for us when we upgrade to 4.0? We should probably keep a note somewhere.

HanlinMiao (approved):
Out of curiosity, is there a significant performance improvement running tests with this flag? If so, is it quantifiable?

glennmatthews:
> Out of curiosity, is there a significant performance improvement running tests with this flag? If so, is it quantifiable?

Yeah, that's what I was trying to show with the "screenshot" in the PR description. Running the full test suite with `--parallel` took "only" 737 seconds locally (though as noted, it did show some failures and errors that need investigation). I didn't have a comparable screenshot for my most recent non-parallel full run (I try to avoid running the full suite locally :wink:) but it's usually on the order of 2800-3000 seconds IIRC.

nautobot/core/tests/runner.py
glennmatthews: It definitely does make for something we need to maintain, though I did a quick compare in github between 3.2.x and 4.2.x implementations and it looks like the only significant change is https://github.com/django/django/commit/3089018e951dcca568574c0e0ebf9d8aab112389#diff-03e86b809c687971a91f0f1aeee4e5731251c2f05956f8c46d7a08e63655b67fL159-L186

HanlinMiao:
> > Out of curiosity, is there a significant performance improvement running tests with this flag? If so, is it quantifiable?
> 
> Yeah, that's what I was trying to show with the "screenshot" in the PR description. Running the full test suite with `--parallel` took "only" 737 seconds locally (though as noted, it did show some failures and errors that need investigation). I didn't have a comparable screenshot for my most recent non-parallel full run (I try to avoid running the full suite locally 😉) but it's usually on the order of 2800-3000 seconds IIRC.

Ahh sorry, I did not read the screenshot carefully. That's awesome tho!

glennmatthews:
```
❯ time invoke unittest --skip-docs-build --keepdb --cache-test-fixtures --parallel --no-buffer
...
Ran 7595 tests in 637.517s

OK (skipped=175, expected failures=1)
...
11:21.99 total
```

versus

```
❯ time invoke unittest --skip-docs-build --keepdb --cache-test-fixtures                       
...
Ran 7595 tests in 2424.286s

OK (skipped=175, expected failures=1)
...
41:13.00 total
```

glennmatthews:
Looks like my latest "fix" (hack) for getting all of the parallel tests to pass has broken CI. I'll look into it this week.

HanlinMiao (approved):

glennmatthews:
Hold off on merging this until we get 2.1.2 out the door and merge `main` back up into `develop`.

PR #5131, opened 2024-01-18
Add `TreeManager.max_depth` cachedproperty as an alternative to `max_tree_depth()`

glennmatthews:
# What's Changed

- Add `TreeManager.max_depth` cachedproperty as a wrapper around `TreeQuerySet.max_tree_depth()` method
- Update `Location.natural_key_field_lookups` implementation to use the cachedproperty
   - This saves us from re-evaluating `Location.objects.max_tree_depth()` every time we calculate a Location instance's `natural_slug`, for example in a REST API list response.
- Update other uses of `max_tree_depth` to also use the cachedproperty as well.

# Screenshots

Debug toolbar for `/api/dcim/locations/` before:
Screenshot 2024-01-18 at 4 40 08 PM

and after:
Screenshot 2024-01-18 at 4 41 20 PM

# TODO

- [x] Unit, Integration Tests


Kircheneer:
Is it okay if I backport this to the LTM branch as well once done? We are still facing some performance issues on locations in a 1.6 install with lots of locations and I think this should help.

timizuoebideri1 (dismissed):

nautobot/core/models/tree_queries.py
```diff
@@ -28,6 +30,10 @@ class TreeManager(TreeManager_, BaseManager.from_queryset(TreeQuerySet)):
     _with_tree_fields = True
     use_in_migrations = True
 
+    @cached_property
```
timizuoebideri1, line 33: There's a chance that old data will be retrieved if a new location is added, which could change `max_tree_depth`.

Is it possible whenever we create a location with a descendent length larger than the current `max_depth`, we should invalidate the cache.

timizuoebideri1:
I also suggest adding a test case that asserts `max_depth` returns the correct value after a new Location is added that alters the depth.

HanlinMiao (approved):

PR #5133, opened 2024-01-19
Sanitize `render_markdown()` output with `nh3` library

glennmatthews:
# What's Changed

- Add `nh3` (Python wrapper around `ammonia`) as an HTML sanitization library and use it in `render_markdown`.
    - This provides a more robust defense against potential XSS and obfuscated script-injection attacks than Django's built-in `strip_tags()` that we were previously using.
    - As a bonus, we are no longer stripping all HTML tags before rendering Markdown to HTML - "safe" tags and attributes defined in an allow-list constant are now preserved, meaning that the various Markdown fields we have now permit explicit HTML tags and attributes so long as they comply with the allow-lists.
- Update documentation
- Add test coverage for a specific reported XSS example and several other examples.

# Screenshots

Some notes that I authored in manual testing. The "Malicious link" would run JS code when rendered in develop and clicked, but now is just a no-op link to nowhere. The "Home" link, horizontal rule, table, and "Hello World" heading are all written directly as HTML rather than Markdown.

develop:
image

this branch:
image

# TODO

- [x] Unit, Integration Tests


nautobot/docs/user-guide/platform-functionality/template-filters.md
```diff
@@ -216,12 +216,86 @@ Render a dictionary as formatted JSON.
 
 ### render_markdown
 
-Render text as Markdown.
+Render and sanitize Markdown text into HTML. A limited subset of HTML tags and attributes are permitted in the text as well; non-permitted HTML will be stripped from the output for security.
```
HanlinMiao, line 219:
```suggestion
Render and sanitize Markdown text into HTML. A limited subset of HTML tags and attributes are permitted in the text; non-permitted HTML will be stripped from the output for security.
```

nautobot/core/settings.py
```diff
@@ -682,7 +682,7 @@
     ),
     "SUPPORT_MESSAGE": ConstanceConfigItem(
         default="",
-        help_text="Help message to include on 4xx and 5xx error pages. Markdown is supported.\n"
+        help_text="Help message to include on 4xx and 5xx error pages. Markdown is supported, as are some HTML tags.\n"
```
HanlinMiao, line 685: Is this accurate?
```suggestion
        help_text="Help message to include on 4xx and 5xx error pages. Markdown is supported, as are some HTML tags and attributes.\n"
```

glennmatthews:
> I'm a little confused as to why `rel="noopener noreferrer"` is present in some of the rendered markdown text.

It's added automatically by `nh3`/`ammonia`; my limited understanding is that it prevents external links, when clicked on, from being able to introspect information from the previous (i.e. Nautobot) page context.

- https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noopener
- https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noreferrer

nautobot/core/tests/test_views.py
```diff
@@ -392,16 +392,16 @@ def test_500_default_support_message(self, mock_get):
...
+            '<a href="https://slack.networktocode.com/" rel="noopener noreferrer">Network to Code\'s '
+            "Slack community</a> and post your question.",
             response_content,
         )
 
     @override_settings(DEBUG=False, SUPPORT_MESSAGE="Hello world!")
     @mock.patch("nautobot.core.views.HomeView.get", side_effect=Exception)
     def test_500_custom_support_message(self, mock_get):
-        """Nautobot's custom 500 page should be used and should include a default support message."""
+        """Nautobot's custom 500 page should be used and should include a custom support message if defined."""
```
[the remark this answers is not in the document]
glennmatthews: In this PR - I had to fix the test, and in so doing, noticed that the docstring was inaccurate.

nautobot/docs/user-guide/platform-functionality/template-filters.md
glennmatthews: Reads more accurately/completely to me as is - but I'm open to convincing?

nautobot/core/settings.py
glennmatthews: But then I have to wrap the help_text to another line to make black/ruff happy... :-(

Will do.

HanlinMiao (approved):

nautobot/docs/user-guide/platform-functionality/template-filters.md
glennmatthews: I was thinking "as well as Markdown".

PR #5146, opened 2024-01-22
Update sidebar navigation menu styling and behavior

bradh11:
# What's Changed
This PR includes updates to improve the side navbar behavior that was introduced in Nautobot 2.1. There are numerous small improvements that make the user experience better than it was with the previous iteration. Below is an enumeration of the changes introduced in this PR:
1. **Menu Animation:** Menu items now collapse in a smooth animated way rather than jumping 100% open immediately and covering menu items below.
2. **Menu Behavior:** As a menu is opened, previously opened menu items collapse in a visually pleasing and responsive way.  Menu expansion now pushes other menus below it downward rather than covering them (z axis). Lastly, previously the expanded menus would close automatically when a sub-item was clicked, effectively causing the user to lose place in their navigation.
3. **Menu Header Improvements:** Menu headers are now represented with a color indication based on menu expanded rather than menu focus. Previously the menu header would lose color after a link is clicked.
4. **Expanded Menus State Persistence:** Behavior (expanded or closed) is kept in local storage which allows it to be maintained on page refresh, link clicked, or page reload. Previously the expanded menus would close automatically when a subitem was clicked, effectively causing the user to lose place in their navigation.
5. **Navbar Sroll Position State Persistence:** Navbar scroll up/down position is stored in local storage and restored to previous position upon page refresh, link clicked, or page reload. Previously the scrollbar position would completely reset after a link was clicked or the page refreshed, effectively causing the user to lose place in their navigation.
6. **Debounce Optimization on Scrollbar State Storage:** Scrollbar events can happen many times per second. Performance is managed by leveraging a debounce function to ensure the storage happens at a reasonable clip.
7. **Removal of Thick Scrollbar:** Issues were being caused with search bar resizing due to the thickness of the scrollbar. This PR removes the scrollbar itself, but not the functionality that allows the navbar to be scrolled up and down.

Overall, I think this PR will bring improved user experience to the Left Navigation menu.

# Screenshots
https://github.com/nautobot/nautobot/assets/18483910/46a25d1a-e29e-4ab7-a95f-cf79b356b15d

# TODO
- [ ] Unit, Integration Tests


bradh11:
I may have one more update to improve the performance and readability of the Javascript portion of this PR. I will test and validate it shortly. Please do not merge or close until I have added this in a new commit.

changes/5146.fixed
```diff
@@ -0,0 +1,8 @@
...
+3. **Menu Header Improvements:** Menu headers are now represented with a color indication based on menu expanded rather than menu focus. Previously the menu header would lose color after a link is clicked.
+4. **Expanded Menus State Persistence:** Behavior (expanded or closed) is kept in local storage which allows it to be maintained on page refresh, link clicked, or page reload. Previously the expanded menus would close automatically when a subitem was clicked, effectively causing the user to lose place in their navigation.
+5. **Navbar Sroll Position State Persistence:** Navbar scroll up/down position is stored in local storage and restored to previous position upon page refresh, link clicked, or page reload. Previously the scrollbar position would completely reset after a link was clicked or the page refreshed, effectively causing the user to lose place in their navigation.
+6. **Debounce Optimization on Scrollbar State Storage:** Scrollbar events can happen many times per second. Performance is managed by leveraging a debounce function to ensure the storage happens at a reasonable clip.
```
glennmatthews, line 8: Appreciate the detail here! However, the processing of this change fragment into the eventual release-notes is fairly naive in that it just converts each line of the fragment into a bullet list item, so the formatting here won't actually be correct in the end. Additionally we have a convention of stating change fragment entries in the past tense. Perhaps:

```suggestion
Changed nav menu items to collapse in a smooth animated way, rather than jumping 100% open immediately and covering menu items below. Previously opened menu items now collapse smoothly as well.
Changed nav menu so that menu expansion now pushes other menus below it downward rather than covering them (z axis).
Changed nav menu headers to not lose their color after a link was clicked.
Added nav menu state (expanded or closed) and scroll position to local storage, allowing it to be maintained on page refresh, link clicked, or page reload.
```

nautobot/core/templates/inc/nav_menu.html
```diff
@@ -2,6 +2,59 @@
 {% load registry %}
 {% load helpers %}
 {% registry %}
+
+<style>
```
glennmatthews, line 6: Should/could any of this styling be added to the `nautobot/project-static/css/base.css` stylesheet instead of being inline in this file? I know we do a fair amount of styling of the nav menus in that file already and am concerned about possible conflicts between the two.

nautobot/core/templates/inc/nav_menu.html
```diff
@@ -2,6 +2,59 @@
...
+        overflow: hidden;
+        background-color: #646464;
+    }
+
+    html[data-theme="dark"] .nav-dropdown-menu, html[data-theme="dark"] .swagger-ui .servers>label select html[data-theme="dark"] .select2-dropdown {
+        background-color: rgb(222, 222, 222) !important;
+    }
+    html[data-theme="dark"] .nav-dropdown-menu>li>a {
+        color: rgb(51, 51, 51) !important;
+    }
```
glennmatthews, line 52: Similarly, should these belong in `nautobot/project-static/css/dark.css`?

glennmatthews (commented):
Looks impressive! Just a few style/self-consistency nits to start with. :-)

nautobot/core/templates/inc/nav_menu.html
bradh11: Happy to move the CSS into the `base.css` file.  Will make those updates.

bradh11:
> I may have one more update to improve the performance and readability of the Javascript portion of this PR. I will test and validate it shortly. Please do not merge or close until I have added this in a new commit.

I believe i have everything in good order now.  Cleaned up some old CSS that was no longer in use as a result of this PR.

cmsirbu:
This looks great @bradh11! And it independently addresses some of the UX feedback I'd given about 2.1's changes 😃

nautobot/core/templates/inc/nav_menu.html
```diff
@@ -111,3 +112,73 @@
...
+                timeout = setTimeout(later, wait);
+                if (callNow) func.apply(context, args);
+            };
+        }
+
+        // Close all dropdowns except the one specified
+        function closeAllDropdownsExcept(exceptId) {
+            dropdowns.forEach(function(collapse) {
+                if (collapse.id !== exceptId && collapse.classList.contains('in')) {
+                    $(collapse).collapse('hide');
```
timizuoebideri1, line 144: Is this jQuery?

nautobot/core/templates/inc/nav_menu.html
```diff
@@ -111,3 +112,73 @@
...
+                    localStorage.setItem('lastOpenedDropdown', collapseElement.id);
+                } else {
+                    $(collapseElement).collapse('hide');
+                    localStorage.removeItem('lastOpenedDropdown');
+                }
+            });
+        });
+
+        // Open the last opened dropdown
+        if (lastDropdownId) {
```
timizuoebideri1, line 167: I believe the purpose of this logic is to open the dropdown menu on the current page. 

However, I see some potential downsides. For example, if we open the `DEVICES` dropdown, and click on `Devices` to navigate to the devices list view, and then click the logo to return home, the `DEVICES` dropdown will remain open. Furthermore, if we manually change the URL in the address bar to navigate to a different page, the `DEVICES` dropdown remains.

timizuoebideri1: If I'm right about the intent, we could switch to a URL-based approach for determining the opened dropdown, rather than relying on the last dropdown clicked.

glennmatthews: Is it a bad thing if the menu stays open when you switch to a different page though?

timizuoebideri1: If it has nothing to do with the current page, I don't see why it should be open, and I haven't observed this pattern on any other dashboard I've visited.

I think that gives the user the wrong impression; it kind of appears like that's the current active dropdown.

nautobot/core/templates/inc/nav_menu.html
bradh11: This is a mix of Javascript and Bootstrap integrated Jquery.  I am leveraging the built-in collapse feature within Bootstrap.

nautobot/core/templates/inc/nav_menu.html
bradh11: The intent was _not_ to match the menu to the currently opened page. 

It is a common UX design pattern for a "Side Navbar" to stay open on a current menu context.  For pages that are grouped together it is common to switch between topic areas.  I agree it would be uncommon for a Top mounted navigation to keep the menus collapsed, but sidebar navigation is a different experience and context is important to maintain.

Some easy to imagine examples would be switching between any of these items below:
* Locations -> Location Types
* Circuits -> Circuit Types
* Racks -> Rack Groups
* Power Feeds -> Power Panels
* IP Addresses -> Prefixes -> Namespaces -> etc...
* --- Many more examples here ---

Keeping the current context of the menu state and position of the scrollbar is more natural feeling to the user than click a link and having everything close back up only to have to find where you last were and then navigate all over from scratch.

One case I agree with modifying is when clicking the Nautobot icon (upper left).  When switching back to the homepage / panel view it would be smart to reset all menus to collapsed.

glennmatthews: Thinking of the longer-term UI direction I do think there's an argument for keeping the state of the sidebar nav independent from the state of the page contents.

jmcgill298: I agree with Brad, I have recently been working with Prefixes and IPAddresses and have found it annoying that the side navbar collapses once I click into one of those Models, and that to switching back and forth requires me to reopen the dropdown

timizuoebideri1 (approved):

glennmatthews (approved):

PR #5169, opened 2024-01-24
Added support for request profiling via django-silk

lampwins:
# What's Changed
This adds support for user session profiling via django-silk. This allows a user to enable a setting in their user profile which persists until they log out or disable the setting. Django silk will then log all subsequent requests for that user's session. Administrators can access the logs in the silk UI.

# Screenshots

Screenshot 2024-01-24 at 4 02 01 PM

silk-ui
# TODO

- [x] Unit, Integration Tests


nautobot/users/forms.py
```diff
@@ -38,3 +38,12 @@ class Meta:
         widgets = {
             "expires": DateTimePicker(),
         }
+
+
+class AdvancedProfileSettingsForm(BootstrapMixin, forms.Form):
+    request_profiling = forms.BooleanField(
+        required=False,
+        help_text="Enable request profiling for the durration of the login session. "
```
glennmatthews, line 46:
```suggestion
        help_text="Enable request profiling for the duration of the login session. "
```

nautobot/users/forms.py
```diff
@@ -38,3 +38,12 @@ class Meta:
...
+        "This is for debugging purposes and should only be enabled when "
+        "instructed by an adiministrator.",
```
glennmatthews, line 48:
```suggestion
        "instructed by an administrator.",
```

nautobot/users/tests/test_views.py
```diff
@@ -91,3 +91,29 @@ def test_change_password_disabled(self):
                     "Remotely authenticated user credentials cannot be changed within Nautobot.",
                     str(response.content),
                 )
+
+
+class AdancedProfileSettingsViewTest(TestCase):
+    """
+    Tests for the user's advanced settings profile edit view
+    """
+
```
glennmatthews, line 100: Would be good to also test authn and authz of this view.

nautobot/users/views.py
```diff
@@ -329,3 +329,53 @@ def post(self, request, pk):
...
+        form = AdvancedProfileSettingsForm(request.POST)
+
+        if form.is_valid():
+            silk_record_requests = form.cleaned_data["request_profiling"]
+
+            # Set the value for `silk_record_requests` in the session
+            request.session["silk_record_requests"] = silk_record_requests
+
+            if silk_record_requests:
+                msg = "Enabled request profiling for the durration of the login session."
```
glennmatthews, line 367:
```suggestion
                msg = "Enabled request profiling for the duration of the login session."
```

nautobot/docs/user-guide/administration/guides/request-profiling.md
```diff
@@ -0,0 +1,77 @@
...
+
+Users must have permissions to access the django-silk UI. Used in combination with `SILKY_AUTHENTICATION`.
+
+---
+
+### SILKY_PERMISSIONS
+
+Default: `lambda user: user.is_superuser`
+
+This ensures the users must be a superuser of the system to access the django-silk UI. Used in combination with `SILKY_AUTHENTICATION` and `SILKY_AUTHORISATION`.
```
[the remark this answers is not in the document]
lampwins: I'm sure the data is there, but out of the box, silk is meant as a back of house telemetry tool and does not offer user information or scoped access. What I have built here is more a convenience on the common troubleshooting workflow where an administrator is working with a user that is having an issue, or the superuser is themself the user that wants to do the profiling. In the context of a production system, it's mostly about being able to toggle the profiling on demand and ensure it is scoped.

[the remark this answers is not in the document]
glennmatthews: Sorry, I think I was a bit unclear :-). If I understand you right, user A can enable profiling, then the admin/superuser can view the profiled data from user A's sessions (and any other users that have it enabled as well)? That seems appropriate and desirable to me.

nautobot/users/tests/test_views.py
lampwins: Do we have an example of what you are thinking here? This was mostly copy-pasta from the change password test cases in the same file.

lampwins: Or are you talking about the superuser access of the silk UI?

nautobot/users/urls.py
```diff
@@ -15,4 +15,5 @@
         views.TokenDeleteView.as_view(),
         name="token_delete",
     ),
+    path("advanded-settings/", views.AdvancedProfileSettingsEditView.as_view(), name="advanced_settings_edit"),
```
glennmatthews, line 18:
```suggestion
    path("advanced-settings/", views.AdvancedProfileSettingsEditView.as_view(), name="advanced_settings_edit"),
```

nautobot/docs/user-guide/administration/guides/request-profiling.md
```diff
@@ -0,0 +1,77 @@
...
+
+From there, administrators can view details of individual requests, including timing, SQL queries, and cProfile artifacts.
+
+## Configuration
+
+The Nautobot configuration comes out of the box with django-silk set up to support the above functionality. Those settings are described below, but it is not the intention of the Nautobot docs to describe all django-silk settings. Django-silk provides several other parameters that knowledgeable users may also be able to use, depending on the use case.
+
+### SILKY_PYTHON_PROFILER
+
+Default: `True`
```
itdependsnetworks, line 27: Every profiler i know provides a data load on it, I would have thought we would default to false and add turn on when in "DEBUG" mode.

lampwins: Not sure what you mean by "data load." This setting doesn't actually enable or turn anything on. Requests are only profiled if the user turns the setting on.

itdependsnetworks: Sorry, was multi-tasking... I just meant adds load (CPU/Memory/IO) to the server.

>  Requests are only profiled if the user turns the setting on.

This "Enables use of the built-in Python cProfile profiler." seems to me to imply it would be the thing that turns the setting on.

lampwins: I can see why you might think that :)

It is true that silk is technically always "on", but because of the intercept function, every request is immediately short circuited _unless_ the setting is enabled on the user's session.

`SILKY_PYTHON_PROFILER = True` is just saying, when silk does actually intercept a request, it should use cProfile.

lampwins: I'll add, the point of this setup is to allow quick, reactive profiling, without the need to reconfigure the environment. I.e. I can do it in a production environment. Because it is always scoped to a user, the performance impact is not a concern. Silk even lets you quantify its own impact in overall profile.

This is a pretty common implementation pattern.

gsnider2195:
We want to add a setting to toggle this feature. Also backport to LTM.

nautobot/users/tests/test_views.py
lampwins: `AdvancedProfileSettingsEditView` already has the `LoginRequiredMixin` on it, but I will add a test for the silk UI.

nautobot/core/settings.py
```diff
@@ -569,6 +571,11 @@
 }
 
 CONSTANCE_CONFIG = {
+    "ALLOW_REQUEST_PROFILING": ConstanceConfigItem(
```
glennmatthews, line 574: For completeness/consistency, we should be able to fallback to a `settings` variable by the same name, e.g. as a commented-out item in `nautobot/core/templates/nautobot_config.py.j2` similar to how the other Constance-capable settings are handled presently.

nautobot/docs/user-guide/administration/guides/request-profiling.md
```diff
@@ -0,0 +1,85 @@
...
+
+![silk ui](../../../media/user-guide/administration/guides/request-profiling/silk-ui.png)
+
+From there, administrators can view details of individual requests, including timing, SQL queries, and cProfile artifacts.
+
+## Configuration
+
+The Nautobot configuration comes out of the box with django-silk set up to support the above functionality. Those settings are described below, but it is not the intention of the Nautobot docs to describe all django-silk settings. Django-silk provides several other parameters that knowledgeable users may also be able to use, depending on the use case.
+
+### ALLOW_REQUEST_PROFILING
```
glennmatthews, line 25: This one at least should perhaps also be documented in `optional-settings.md`, perhaps with a cross-reference to this file?

nautobot/docs/user-guide/administration/guides/request-profiling.md
```diff
@@ -0,0 +1,85 @@
+# Request Profiling
```
glennmatthews, line 1: This file probably should be added to  the table of contents in `mkdocs.yml`?

nautobot/users/tests/test_views.py
```diff
@@ -91,3 +91,49 @@ def test_change_password_disabled(self):
                     "Remotely authenticated user credentials cannot be changed within Nautobot.",
                     str(response.content),
                 )
+
+
+class AdancedProfileSettingsViewTest(TestCase):
```
glennmatthews, line 96:
```suggestion
class AdvancedProfileSettingsViewTest(TestCase):
```

nautobot/docs/user-guide/administration/guides/request-profiling.md
lampwins: It is listed under the admin guides section there, did I miss something else?

```diff
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -310,6 +310,7 @@ nav:
         - Permissions: "user-guide/administration/guides/permissions.md"
         - Prometheus Metrics: "user-guide/administration/guides/prometheus-metrics.md"
         - Replicating Nautobot: "user-guide/administration/guides/replicating-nautobot.md"
+        - Request Profiling: "user-guide/administration/guides/request-profiling.md"
         - Django Storages with S3: "user-guide/administration/guides/s3-django-storage.md"
         - Task Queues: "user-guide/administration/guides/celery-queues.md"
     - Feature Guides:
```

glennmatthews: No, but apparently I did :-) Sorry for the noise.

glennmatthews (approved):

nautobot/core/settings.py
```diff
@@ -445,6 +446,7 @@
     "nautobot.core.middleware.ExternalAuthMiddleware",
     "nautobot.core.middleware.ObjectChangeMiddleware",
     "django_prometheus.middleware.PrometheusAfterMiddleware",
+    "silk.middleware.SilkyMiddleware",
```
gsnider2195, line 449: Just want to make sure you checked that the order of the middleware is ok. From https://github.com/jazzband/django-silk?tab=readme-ov-file#installation

> Note: The middleware placement is sensitive. If the middleware before silk.middleware.SilkyMiddleware returns from process_request then SilkyMiddleware will never get the chance to execute. Therefore you must ensure that any middleware placed before never returns anything from process_request. See the [django docs](https://docs.djangoproject.com/en/dev/topics/http/middleware/#process-request) for more information on this.

nautobot/core/settings.py
```diff
@@ -914,3 +922,35 @@ def UI_RACK_VIEW_TRUNCATE_FUNCTION(device_display_name):
...
+SILKY_AUTHENTICATION = True  # User must login
+SILKY_AUTHORISATION = True  # User must have permissions
+
+
+# This makes it so that only superusers can access the silk UI
+def silk_user_permissions(user):
+    return user.is_superuser
+
+
+SILKY_PERMISSIONS = silk_user_permissions
```
gsnider2195, line 944: Nitpick incoming

```suggestion
SILKY_PERMISSIONS = lambda user: user.is_superuser
```

nautobot/core/settings.py
```diff
@@ -569,6 +571,11 @@
 }
 
 CONSTANCE_CONFIG = {
+    "ALLOW_REQUEST_PROFILING": ConstanceConfigItem(
+        default=True,
```
gsnider2195, line 575: Looking at this from an administrator's perspective, I'd probably want this to be disabled by default and only enable it when I want a user to generate data for me. Otherwise users may turn this on accidentally or maliciously and put extra load on the system.

nautobot/users/forms.py
```diff
@@ -38,3 +39,20 @@ class Meta:
         widgets = {
             "expires": DateTimePicker(),
         }
+
+
+class AdvancedProfileSettingsForm(BootstrapMixin, forms.Form):
+    request_profiling = forms.BooleanField(
```
gsnider2195, line 45: Should this be disabled if `ALLOW_REQUEST_PROFILING = False`?

nautobot/core/settings.py
lampwins: Ha, if you can tell me the right way to suppress pylint and ruff here, I am fine with this. I will find several commits from me trying to do it...

nautobot/core/settings.py
lampwins: Good call out. Technically none of our current middleware are problematic, but no reason to not move it towards the top. Side note, we should investigate if the GZip middleware would be useful to us.

nautobot/core/settings.py
lampwins: Yeah I'm fine that, given it is a constance setting.

nautobot/core/settings.py
gsnider2195: PEP8 makes a good case for declaring the function explicitly so we should probably leave it as-is

gsnider2195 (approved):
The only concern I have is there are no links to the `/silk/` URL anywhere in the application.

PR #5176, opened 2024-01-26
feature: implement json and yaml syntax highlighting

### LINKED ISSUE #5098: Syntax highlighting for JSON/YAML
As a Network Automation Engineer, I want nautobot to show properly formatted and syntax highlighted JSON & YAML whenever such data is presented in the UI, so that I can more easily consume and manage this data.

The primary focus of this story is in the presentation of the data, but a stretch goal of offering syntax highlighting in edit forms is a nice to have.

- [ ] **Syntax highlighting for JSON/YAML**
  - Given a Network Automation Engineer
  - When I view JSON or YAML data in the nautobot UI
  - Then the data should be properly formatted and syntax highlighted
  - And I should be able to easily consume and manage the data

norbert-mieczkowski-codilime:
# Closes #5098
# What's Changed

* Introduce client-side JSON and YAML syntax highlighting with [highlight.js](https://github.com/highlightjs/highlight.js) library.
* Embed a subset of highlight.js in the core application which supports only JSON and YAML languages and provide two themes: `github.min.css` and `github-dark.min.css`.
* Implement a main script which on page load highlights all the elements with `language-json` and `language-yaml` class. These elements preferably have the following HTML structure: `<pre><code class="language-json">...</pre></code>`.
* Add `language-json` and `language-yaml` class to all the existing elements which contents utilize `render_json` and/or `render_yaml` Django template filters.
* Update `/theme-preview/` page with syntax highlighting examples.

# Screenshots

| Light theme | Dark theme |
| - | - |
| image | image_1 |

### DIFF
--- a/nautobot/core/templates/inc/media.html
+++ b/nautobot/core/templates/inc/media.html
@@ -20,8 +20,11 @@
           href="{% static 'select2-bootstrap-0.1.0-beta.10/select2-bootstrap.min.css' %}"
           onerror="window.location='{% url 'media_failure' %}?filename=select2-bootstrap-0.1.0-beta.10/select2-bootstrap.min.css'">
     <link rel="stylesheet"
-          href="{% static 'flatpickr-4.6.9/themes/light.min.css' %}"
-          onerror="window.location='{% url 'media_failure' %}?filename=flatpickr-4.6.9/themes/light.min.css'">
+          href="{% static 'highlight.js-11.9.0/github.min.css' %}"
+          onerror="window.location='{% url 'media_failure' %}?filename=highlight.js-11.9.0/github.min.css'"{% if request.COOKIES|get_item:"theme" == 'dark' %} disabled="disabled"{% endif %} >
+    <link rel="stylesheet"
+          href="{% static 'highlight.js-11.9.0/github-dark.min.css' %}"
+          onerror="window.location='{% url 'media_failure' %}?filename=highlight.js-11.9.0/github-dark.min.css'"{% if request.COOKIES|get_item:"theme" != 'dark' %} disabled="disabled"{% endif %} />
     <link rel="stylesheet" id="dark-theme"
           href="{% versioned_static 'css/dark.css' %}"
           onerror="window.location='{% url 'media_failure' %}?filename=css/dark.css'"{% if request.COOKIES|get_item:"theme" != 'dark' %} disabled="disabled"{% endif %} />
--- a/nautobot/extras/templates/extras/graphqlquery.html
+++ b/nautobot/extras/templates/extras/graphqlquery.html
@@ -17,7 +17,7 @@
                 </tr>
                 <tr>
                     <td>Query Variables</td>
-                    <td><pre>{{ object.variables|render_json }}</pre></td>
+                    <td><pre><code class="language-json">{{ object.variables|render_json }}</code></pre></td>
                 </tr>
             </table>
         </div>
--- a/nautobot/core/settings.py
+++ b/nautobot/core/settings.py
@@ -646,10 +646,10 @@ CONSTANCE_CONFIG = {
             "Extend or override default Platform.network_driver translations provided by "
             '<a href="https://netutils.readthedocs.io/en/latest/user/lib_use_cases_lib_mapper/">netutils</a>. '
             "Enter a dictionary in JSON format, for example:\n"
-            "<pre>{\n"
+            '<pre><code class="language-json">{\n'
             '    "netmiko": {"my_network_driver": "cisco_ios"},\n'
             '    "pyats": {"my_network_driver": "iosxe"} \n'
-            "}</pre>",
+            "}</code></pre>",
         ),
         # Use custom field type defined above
         field_type="optional_json_field",
--- a/nautobot/core/templates/inc/custom_fields/panel_data.html
+++ b/nautobot/core/templates/inc/custom_fields/panel_data.html
@@ -44,7 +44,7 @@
                                             Show/Hide
                                         </button>
                                     </p>
-                                    <pre class="collapse" id="cf_{{ field.key }}"><code>{{ value|render_json }}</code></pre>
+                                    <pre class="collapse" id="cf_{{ field.key }}"><code class="language-json">{{ value|render_json }}</code></pre>
                                 {% elif value or value == 0 %}
                                     {{ value }}
                                 {% elif field.required %}
--- a/nautobot/core/templates/inc/javascript.html
+++ b/nautobot/core/templates/inc/javascript.html
@@ -14,6 +14,8 @@
         onerror="window.location='{% url 'media_failure' %}?filename=clipboard.js-2.0.9/clipboard.min.js'"></script>
 <script src="{% static 'flatpickr-4.6.9/flatpickr.min.js' %}"
         onerror="window.location='{% url 'media_failure' %}?filename=flatpickr-4.6.9/flatpickr.min.js'"></script>
+<script src="{% static 'highlight.js-11.9.0/highlight.min.js' %}"
+        onerror="window.location='{% url 'media_failure' %}?filename=highlight.js-11.9.0/highlight.min.js'"></script>
 <script src="{% versioned_static 'js/forms.js' %}"
         onerror="window.location='{% url 'media_failure' %}?filename=js/forms.js'"></script>
 <script src="{% versioned_static 'js/theme.js' %}"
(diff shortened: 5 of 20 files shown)


nautobot/core/templates/inc/media.html
```diff
@@ -20,8 +20,11 @@
           href="{% static 'select2-bootstrap-0.1.0-beta.10/select2-bootstrap.min.css' %}"
           onerror="window.location='{% url 'media_failure' %}?filename=select2-bootstrap-0.1.0-beta.10/select2-bootstrap.min.css'">
     <link rel="stylesheet"
-          href="{% static 'flatpickr-4.6.9/themes/light.min.css' %}"
-          onerror="window.location='{% url 'media_failure' %}?filename=flatpickr-4.6.9/themes/light.min.css'">
```
glennmatthews, line 24: Was this removal intentional?

norbert-mieczkowski-codilime: Absolutely not, thanks for spotting that one!

norbert-mieczkowski-codilime:
@glennmatthews, yeah I definitely see your point and I agree that adding a set of wrapping elements with a certain required class is not ideal. I tried to find a common pattern which could automate the process a bit and here's why I didn't go down that path:
1. `render_json` is used in `<textarea>` element, making it currently impossible to wrap the code in `<code class="...">...</code>` at the template filter level. It's just a single usage, nevertheless it counts.
2. Selecting `pre` or `code` alone on the other hand is too generic because they can contain all sorts of languages or pseudocode, not only JSON or YAML.

glennmatthews:
> @glennmatthews, yeah I definitely see your point and I agree that adding a set of wrapping elements with a certain required class is not ideal. I tried to find a common pattern which could automate the process a bit and here's why I didn't go down that path:
> 
>     1. `render_json` is used in `<textarea>` element, making it currently impossible to wrap the code in `<code class="...">...</code>` at the template filter level. It's just a single usage, nevertheless it counts.

Great thought/catch. I hadn't considered that case.

In that case, let's add this as a highlighted feature under the `Release Overview` in `nautobot/docs/release-notes/version-2.2.md` and use that space to instruct app developers as to what they need to add to take advantage of this feature.

lampwins:
Should we introduce a new template tag that does this though? I want to make sure it is as easy as possible to consume.

HanlinMiao:
> Should we introduce a new template tag that does this though? I want to make sure it is as easy as possible to consume.

For Norbert's reference, do you mean something like this?
```
@register.simple_tag
def json_field(value):
    return <pre><code class="language-json">{{value|render_json }}</code></pre>
```

lampwins:
Let's go with `syntax_highlight` as the argument and default to `True` to make it more explicit and avoid the double negative.

glennmatthews:
Something like (untested):

```python
@library.filter()
@register.filter()
def render_json(value, no_highlight=False):
    """
    Render a dictionary as formatted JSON.

    Unless `no_highlight` is specified, the returned string will be wrapped in a
    `<code class="language-json>` HTML tag to flag it for syntax highlighting by highlight.js.
    """
    rendered_json = json.dumps(value, indent=4, sort_keys=True, ensure_ascii=False)
    if not no_highlight:
        return format_html('<code class="language-json">{}</code>', rendered_json)
    return rendered_json
```

glennmatthews:
How difficult/costly is it to enable additional languages? Might be useful at least to include XML as I think there's some work in several Apps that might benefit from that.

norbert-mieczkowski-codilime:
@glennmatthews, adding more languages boils down to increasing library size which ultimately affects page load times. Here's a list of all languages available in highlight.js: [Download a Custom Build - highlight.js](https://highlightjs.org/download). Supporting one, two or three more (I believe in Nautobot apart from JSON and YAML I also saw code snippets in GraphQL, Python and XML) shouldn't make much of a difference - it's just a couple of kBs (with no gizp) as long as we just don't grab them all blind.

norbert-mieczkowski-codilime:
@glennmatthews @lampwins, now when the `<code class="...">...</code>` wrapper is moved to `render_json`/`render_yaml` template filter, is documenting this feature in `nautobot/docs/release-notes/version-2.2.md` still relevant?

glennmatthews:
IMO it's worth adding a note in the Release Overview still.

glennmatthews:
> @glennmatthews, adding more languages boils down to increasing library size which ultimately affects page load times. Here's a list of all languages available in highlight.js: [Download a Custom Build - highlight.js](https://highlightjs.org/download). Supporting one, two or three more (I believe in Nautobot apart from JSON and YAML I also saw code snippets in GraphQL, Python and XML) shouldn't make much of a difference - it's just a couple of kBs (with no gizp) as long as we just don't grab them all blind.

Let's confirm with @lampwins, but I think it'd be good to add GraphQL and XML now. We could already make use of GraphQL highlighting e.g. in `nautobot/extras/templates/extras/graphqlquery.html` for the `query` field. For that matter, we probably should also (re-)apply JSON highlighting to the `query_output` field when it's populated after running the saved query in question - see for example https://demo.nautobot.com/extras/graphql-queries/87b52a0d-f872-4cc9-961c-e848b9b91b83/.

nautobot/docs/user-guide/platform-functionality/template-filters.md
```diff
@@ -210,6 +210,8 @@ If value is None this renders `<span class="text-muted">&mdash;</span>`
 
 Render a dictionary as formatted JSON.
 
+Unless `syntax_highlight=False` is specified, the returned string will be wrapped in a `<code class="language-json>` HTML tag to flag it for syntax highlighting by highlight.js.
```
glennmatthews, line 213:
```suggestion
+/- 2.2.0
    Unless `syntax_highlight=False` is specified, the returned string will be wrapped in a `<code class="language-json>` HTML tag to flag it for syntax highlighting by highlight.js.
```

nautobot/docs/user-guide/platform-functionality/template-filters.md
```diff
@@ -300,6 +302,8 @@ The set of permitted HTML tags is defined in `nautobot.core.constants.HTML_ALLOW
 
 Render a dictionary as formatted YAML.
 
+Unless `syntax_highlight=False` is specified, the returned string will be wrapped in a `<code class="language-yaml>` HTML tag to flag it for syntax highlighting by highlight.js.
```
glennmatthews, line 305:
```suggestion
+/- 2.2.0
    Unless `syntax_highlight=False` is specified, the returned string will be wrapped in a `<code class="language-yaml>` HTML tag to flag it for syntax highlighting by highlight.js.
```

lampwins:
Yes, let's add GraphQL and XML but draw the line there. Are we creating new template tags for these?

glennmatthews:
> Yes, let's add GraphQL and XML but draw the line there. Are we creating new template tags for these?

I don't think we need to at this time - just maybe document that you need to use `language-graphql` and `language-xml` CSS classes on the containing HTML elements in order to take advantage?

changes/5098.added
```diff
@@ -0,0 +1 @@
+Added client-side JSON and YAML syntax highlighting with highlight.js library.
```
glennmatthews, line 1:
```suggestion
Added client-side GraphQL, JSON, XML, and YAML syntax highlighting with the `highlight.js` library.
```

glennmatthews (approved):

gsnider2195 (changes requested):
When I save and reload a GraphQL query I'm seeing the HTML show up in the Query variables:

image

glennmatthews:
That's going to be in `nautobot/core/templates/graphene/graphiql.html`

nautobot/extras/templates/extras/graphqlquery.html
```diff
@@ -48,9 +48,23 @@
...
+        const code = document.createElement('code');
+        code.classList.add('language-json');
+        code.textContent = JSON.stringify(data, undefined, 2);
+
+        queryOutput.appendChild(code);
+
+        hljs.highlightElement(code);
+    }
+
     function test_query() {
```
gsnider2195, line 65: This is no longer working for me in firefox. When I click the `Execute` button I get this error in the console:

```
Uncaught TypeError: document.querySelector(...) is null
    test_query http://localhost:8080/extras/graphql-queries/46021bfc-3827-4c25-bc9f-2ecc6039d5f7/?tab=main:2985
    onclick http://localhost:8080/extras/graphql-queries/46021bfc-3827-4c25-bc9f-2ecc6039d5f7/?tab=main:1
```

gsnider2195: Looks like this is the issue maybe:

```
document.querySelector('#query_variables')
null 
```

gsnider2195 (approved):

PR #5200, opened 2024-01-31
feature: rearrange job list table row contents

### LINKED ISSUE #5126: Job list view refinements
As a network automation engineer, I want a few refinements to the jobs list view, so that the view is more usable when I need to identify jobs I want to run.

- Reduce the size of the grouping headers
- When a job is disabled, grey the row background to more clearly indicate the status
- Change the name hyperlink to route to the job's run page instead of the detail view of the job.
  - Add the "play" button to the beginning of the name text
  - Replace the existing play button on the right with an mdi-information button that links to the job's detail view
  - Something like this:
Screenshot 2024-01-18 at 10 48 25 AM

Checkboxes are covered in #5126

norbert-mieczkowski-codilime:
# Closes #5126

# What's Changed

- Decrease job group accordion panel `font-size` to 14px and emphasise it with bold `font-weight` instead.
- Move play button to the left of job's name, make the anchor link to job run instead of job detail view.
- Replace play icon linking to job run with information icon linking to job detail view.
- Show disabled jobs with lower opacity on the link in the first column, `cursor: not-allowed;` and grayed out description.

# Screenshots

| Rearranged view | Explanation of changes |
| - | - |
| image | image_explained |

### DIFF
--- a/changes/5126.changed
+++ b/changes/5126.changed
@@ -1,0 +1,1 @@
+Rearranged Job List table row contents.
--- a/nautobot/extras/tables.py
+++ b/nautobot/extras/tables.py
@@ -97,7 +97,7 @@ GITREPOSITORY_BUTTONS = """
 """
 
 JOB_BUTTONS = """
-<a href="{% url 'extras:job_run' pk=record.pk %}" class="btn btn-primary btn-xs" title="Run/Schedule" {% if not perms.extras.run_job or not record.runnable %}disabled="disabled"{% endif %}><i class="mdi mdi-play" aria-hidden="true"></i></a>
+<a href="{% url 'extras:job' pk=record.pk %}" class="btn btn-primary btn-xs" title="Details"><i class="mdi mdi-information" aria-hidden="true"></i></a>
 """
 
 OBJECTCHANGE_OBJECT = """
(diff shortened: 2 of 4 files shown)


itdependsnetworks:
super helpful to have the explanation of change in the screenshot, thanks!

HanlinMiao:
Love the changes and explanation! Just need a changelog fragment and we should be good to go.

gsnider2195:
@lampwins what do you think of the disabled job rows being the same color as the accordion header?

gsnider2195 (approved):
One minor nitpick on the color of the disabled rows. Otherwise this looks great. Thanks!

norbert-mieczkowski-codilime:
@gsnider2195, @lampwins, good point, let me explain. Lighter grays, apart from being outside of our color palette b/c accordion header background gray is already the lightest, have too low contrast and are barely visible. On the other hand - darker grays stand out too much and instead of making disabled rows less visible they actually do the opposite. So my recommendation is to either leave the disabled row background the same as accordion header's or to not change it at all (that is leave it white).

gsnider2195:
> @gsnider2195, @lampwins, good point, let me explain. Lighter grays, apart from being outside of our color palette b/c accordion header background gray is already the lightest, have too low contrast and are barely visible. On the other hand - darker grays stand out too much and instead of making disabled rows less visible they actually do the opposite. So my recommendation is to either leave the disabled row background the same as accordion header's or to not change it at all (that is leave it white).

make background white and description gray

gsnider2195:
Here's how it looks on my job list now
image

PR #5194, opened 2024-01-30
Fixed AssertionError when querying Date type custom fields in GraphQL.

HanlinMiao:
# Closes #3664 
# What's Changed

Fixed AssertionError when querying Date type custom fields in GraphQL.
Will backport this to LTM after it is approved.

# Screenshots

Screen Shot 2024-01-30 at 12 17 08 PM

# TODO

- [x] Explanation of Change(s)
- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [x] Attached Screenshots, Payload Example
- [x] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example Plugin Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #5204, opened 2024-01-31
[LTM] Fixed AssertionError when querying Date type custom fields in GraphQL in LTM

### LINKED ISSUE #3664: Custom field date-type objects AssertionError in GraphQL
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.17b1, also tested in 1.5.8
* Python version: 3.9
* Database platform, version: psql 13.10
* Middleware(s): None

### Steps to Reproduce
1. Create a [custom field object](https://demo.nautobot.com/extras/custom-fields/last_backup/?tab=main) for the Device type template with the type as Date
2. Set a date on a [device](https://demo.nautobot.com/dcim/devices/4d7aecac-addd-4f3c-8efc-e102872b54e5/?tab=main) in YYYY-MM-DD iso8601 format
3. You have to restart nautobot-server; we may need a post_save signal event for this as well, which would be a separate issue if so
4. Using GraphQL attempt to retrieve the value of this field from a device
`
{
  devices (name: "ams01-dist-01") {
    name,
    cf_last_backup
  }
}
`

### Expected Behavior
Query to return the value of the Date

### Observed Behavior
* GrapQL returns `"message": "Received not compatible date \"'2023-04-26'\""`
* stdout from nautobot-server gets an exception

`[27/Apr/2023 17:20:01] "POST /graphql/ HTTP/1.1" 200 4437611
  File "/home/meganerd/.cache/pypoetry/virtualenvs/nautobot-Sgl-QJQc-py3.9/lib/python3.9/site-packages/graphql/execution/executor.py", line 480, in complete_value_catching_error
    completed = complete_value(
  File "/home/meganerd/.cache/pypoetry/virtualenvs/nautobot-Sgl-QJQc-py3.9/lib/python3.9/site-packages/graphql/execution/executor.py", line 563, in complete_value
    return complete_leaf_value(return_type, path, result)
  File "/home/meganerd/.cache/pypoetry/virtualenvs/nautobot-Sgl-QJQc-py3.9/lib/python3.9/site-packages/graphql/execution/executor.py", line 626, in complete_leaf_value
    serialized_result = return_type.serialize(result)
  File "/home/meganerd/.cache/pypoetry/virtualenvs/nautobot-Sgl-QJQc-py3.9/lib/python3.9/site-packages/graphene/types/datetime.py", line 23, in serialize
    assert isinstance(
AssertionError: Received not compatible date "'2023-04-26'"`

HanlinMiao:
# Closes #3664 
# What's Changed

# Screenshots

# TODO

- [x] Unit, Integration Tests


whitej6 (approved):
LGTM, matches #5194

nautobot/core/graphql/types.py
```diff
@@ -34,3 +36,20 @@ class ContentTypeType(OptimizedNautobotObjectType):
...
+
+class DateType(graphene.Date):
+    """
+    Overriding the default serialize method from https://github.com/graphql-python/graphene/blob/master/graphene/types/datetime.py
+    to handle the case where the date object is passed as a str object.
+    """
+
+    @staticmethod
+    def serialize(date):
+        if isinstance(date, datetime.datetime):
```
gsnider2195, line 49: If it's a datetime instance shouldn't we be serializing to a date and time not just the date?

HanlinMiao: The DateType CustomField form selection widget only allows **date** and not **date and time**.  
Screen Shot 2024-01-31 at 4 16 54 PM. This is only an instance check and it should be represented as a `date` object in serialization IMO.

gsnider2195: It should be a datetime.date then, not a datetime.datetime. Probably an unrelated bug somewhere else in the code?

gsnider2195: Maybe the root cause of this issue though?

HanlinMiao: This is a direct lift and shift from graphene https://github.com/graphql-python/graphene/blob/5fb7b543779360c8c3ffbf9d07e5769a14a1a0ed/graphene/types/datetime.py#L20. If you look at the definition of the `serialize` method here. It is checking if the date instance is a `datetime` type even tho it is within the `Date` class.

The root cause of the issue is that the date object we passed into the custom field is converted to a str type somewhere in the code (perhaps in the `_custom_field_data` conversion?) and that is what the elif statement fixed.

gsnider2195: Lets add the date check here to match the parent class. We'll need to update develop too ☹️

HanlinMiao: Is this what you had in mind? https://github.com/nautobot/nautobot/pull/5204/commits/c25daac9832d06a3e456dbec32d24cc559234d34.
Or do you want me to take out the `if isinstance(date, datetime.datetime)` conditional entirely?

nautobot/core/graphql/types.py
```diff
@@ -34,3 +37,20 @@ class ContentTypeType(OptimizedNautobotObjectType):
...
+class DateType(graphene.Date):
+    """
+    Overriding the default serialize method from https://github.com/graphql-python/graphene/blob/master/graphene/types/datetime.py
+    to handle the case where the date object is passed as a str object.
+    """
+
+    @staticmethod
+    def serialize(date):
+        if isinstance(date, (datetime.date, datetime.datetime)):
+            date = date.date()
```
gsnider2195, line 51: This won't work on a `datetime.date` object:

```py
>>> d = datetime.date.today()
>>> d
datetime.date(2024, 2, 1)
>>> d.date()
AttributeError: 'datetime.date' object has no attribute 'date'
```

HanlinMiao: Should be fixed in the latest commit, sorry for the trouble.

gsnider2195: Looks good. We should probably add a test just for this, providing string, date and datetime instances for the serialize method.

gsnider2195 (approved):

PR #5215, opened 2024-02-01
IPAM API queryset tuning and automatic query optimization in list views

### LINKED ISSUE #4898: VRF API Endpoint slow
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.5
* Python version: 3.8.18
* Database platform, version: Postgres 13
* Running on an AWS Workspace

### Steps to Reproduce
1. Create VRFs with some CustomFields, RTs and Prefixes (for us around 200 VRFs)
2. Add Prefixes (for us around 75.000)
3. Open the API Endpoint via Browser $NAUTOBOT/api/ipam/vrfs/

### Expected Behavior
I would expect at least the same loading times like with 1.6 with this data set (1 second).

### Observed Behavior
The API Endpoint with 2.0.5 (and 2.0.3) took about 8 seconds to load.
To open the VRFs within the GUI ($NAUTOBOT/ipam/vrfs) is faster and took about 1,8 seconds ( < 1 Second with 1.6)

This behavior may be related to another performance issue we noticed in the GUI Prefix view. If I try to filter the list of Prefixes with the assigned VRF it took these bespoke 8 seconds to show me the list of VRFs.

I'll add the output of the SQL commands from the Django toolbar, maybe you can see something helpful in there.

[nautobot_2.0.5_vrf_api_issue.txt](https://github.com/nautobot/nautobot/files/13529590/nautobot_2.0.5_vrf_api_issue.txt)

glennmatthews:
# Closes #4898 
# What's Changed

Some of the slowness of the VRF REST API endpoint at scale is attributable to the inclusion of M2M relations in the REST API in 2.x, where they were not included in 1.x. This PR doesn't attempt to make the invasive changes to the REST API to take a different approach; what it *does* do is add some `select_related`/`prefetch_related` calls that were missed in the IPAM model changes in 2.0. These should *generally* improve the performance of the affected REST API endpoints by reducing the number of distinct queries that are evaluated when rendering these endpoints.

Additionally, this addresses a TODO in `nautobot/core/tables.py` by automatically optimizing list-view querysets using `select_related` where appropriate.

# Screenshots

# TODO

- n/a Unit, Integration Tests


nautobot/ipam/views.py
```diff
@@ -80,7 +80,7 @@ class NamespaceUIViewSet(
     form_class = forms.NamespaceForm
     bulk_update_form_class = forms.NamespaceBulkEditForm
     filterset_class = filters.NamespaceFilterSet
-    queryset = Namespace.objects.all()
+    queryset = Namespace.objects.select_related("location")
```
gsnider2195, line 83: Should we prefetch tags here? It's a default column

nautobot/ipam/views.py
```diff
@@ -234,7 +234,7 @@ def get_extra_context(self, request, instance):
 
 
 class VRFListView(generic.ObjectListView):
-    queryset = VRF.objects.all()
+    queryset = VRF.objects.select_related("namespace", "tenant")
```
gsnider2195, line 237: Should we prefetch tags, import targets and export targets here? They're not default columns but they are available columns.

nautobot/ipam/views.py
```diff
@@ -316,7 +316,7 @@ class VRFBulkDeleteView(generic.BulkDeleteView):
 
 
 class RouteTargetListView(generic.ObjectListView):
-    queryset = RouteTarget.objects.all()
+    queryset = RouteTarget.objects.select_related("tenant")
```
gsnider2195, line 319: Prefetch tags here? Tags isn't a default column but can be shown if selected.

changes/4898.fixed
```diff
@@ -0,0 +1,2 @@
+Corrected `prefetch_related`/`select_related` calls to improve performance of `/api/ipam/prefixes/` and `/api/ipam/vrfs/` REST API endpoints.
+Corrected `select_related` calls to improve performance of `/ipam/namespaces/`, `/ipam/route-targets/`, and `/ipam/vrfs/` UI list views.
```
gsnider2195, line 2: Do you think this should just be "Optimized database queries to improve performance of ..." to make this easier to understand for non-Django developers reading the release notes?

nautobot/ipam/views.py
glennmatthews: We actually have code for tables to automatically `prefetch_related` based on the visible table columns - https://github.com/nautobot/nautobot/blob/develop/nautobot/core/tables.py#L102-L131. It may be that the more optimal path here long-term is to address the TODO comment there and add logic to automatically handle `select_related` as well?

changes/4898.fixed
glennmatthews: Yeah, probably so.

```suggestion
Optimized database queries to improve performance of `/api/ipam/prefixes/` and `/api/ipam/vrfs/` REST API endpoints.
Optimized database queries to improve performance of `/ipam/namespaces/`, `/ipam/route-targets/`, and `/ipam/vrfs/` UI list views.
```

nautobot/ipam/views.py
gsnider2195: Yeah that looks like a good idea

glennmatthews: I'll look into it.

nautobot/dcim/views.py
```diff
@@ -360,10 +360,8 @@ class RackGroupBulkDeleteView(generic.BulkDeleteView):
 
 
 class RackListView(generic.ObjectListView):
-    queryset = (
-        Rack.objects.select_related("location", "rack_group", "tenant", "role", "status")
-        .prefetch_related("devices__device_type")
-        .annotate(device_count=count_related(Device, "rack"))
+    queryset = Rack.objects.prefetch_related("devices__device_type").annotate(
```
gsnider2195, line 363: Why are we prefetching device_type here?

nautobot/ipam/api/views.py
```diff
@@ -53,13 +53,8 @@ class NamespaceViewSet(NautobotModelViewSet):
...
 
 class VRFViewSet(NautobotModelViewSet):
-    queryset = (
-        VRF.objects.select_related("tenant").prefetch_related("import_targets", "export_targets", "tags")
-        # FIXME(jathan): See if we need to revise the counts for prefixes/ips, here?
-        # .annotate(
-        #     ipaddress_count=count_related(IPAddress, "vrf"),
-        #     prefix_count=count_related(Prefix, "vrf"),
-        # )
+    queryset = VRF.objects.select_related("namespace", "tenant").prefetch_related(
```
gsnider2195, line 56: I'm guessing we need some more automatic select_related/prefetch_related code to handle UIViewsets?

gsnider2195: Not asking you to do that, just making sure I understand why this wasn't removed.

nautobot/ipam/views.py
```diff
@@ -424,19 +424,7 @@ class PrefixListView(generic.ObjectListView):
...
-        "tenant",
-        "vlan",
-        "rir",
-        "role",
-        "status",
-    ).prefetch_related(
-        "ip_addresses",
-        "children",
-    )
+    queryset = Prefix.objects.select_related("parent").prefetch_related("children")
```
gsnider2195, line 427: Self referencing fields don't work with the automatic table code?

nautobot/ipam/api/views.py
glennmatthews: This is the API views file - we don't yet have automatic code in that case, only the UI ListViews and similar.

nautobot/dcim/views.py
glennmatthews: I'm guessing this was intended to relate to the rack utilization calculation via `Rack.get_available_units()` which uses both `Rack.devices` and `Rack.devices.device_type`. However, I see those queries being evaluated for each rack regardless of whether this `prefetch_related` is present or not, so I **think** it's not needed.

nautobot/ipam/views.py
```diff
@@ -1282,7 +1270,7 @@ class VLANGroupBulkDeleteView(generic.BulkDeleteView):
 
 
 class VLANListView(generic.ObjectListView):
-    queryset = VLAN.objects.select_related("location", "vlan_group", "tenant", "role", "status")
+    queryset = VLAN.objects.prefetch_related("prefixes")
```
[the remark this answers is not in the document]
glennmatthews: Good catch. This is a reverse FK (ManyToOneRel) not a M2M, and that case wasn't being detected. Adding it now to the base code.

gsnider2195 (approved):

PR #5216, opened 2024-02-02
[LTM] Request profiling

lampwins:
LTM port of #5169


nautobot/docs/administration/request-profiling.md
```diff
@@ -0,0 +1,85 @@
...
+
+Enables use of the built-in Python cProfile profiler.
+
+---
+
+### SILKY_PYTHON_PROFILER_BINARY
+
+Default: `True`
+
+Generates a binary `.prof` file for each profiled request, which can be downloaded.
```
glennmatthews, line 45: What if anything does Silk do to clean up these files? Do they expire after a period of time or do we need to worry about disk space consumption over the long run?

gsnider2195 (approved):

nautobot/docs/administration/request-profiling.md
lampwins: Thanks, this made me dig into it and while silk does garbage collection for its model data, it does not actually clean these files. I added a note here and will get it into develop.

glennmatthews (approved):

Issue #5099, opened 2024-01-12
Build settings docs from setting schema


lampwins:

lampwins:
As a nautobot engineer, I want the docs related to nautobot settings/configuration options to be automatically build from the settings JSONSchema, so that I can ensure it is always up to date.

- Build a markdown plugin to parse the settings schema and produce the docs for settings options.

HanlinMiao:
1. Remove optional_settings.md and required_settings.md
2. Create jinja2 templates for those two files.
3. Add a script to a Gen-Files plugin to render the jinja templates using the [JSONSchema](https://app.zenhub.com/workspaces/nautobot-core-engineering-64a875e74842500d9e80e23e/issues/gh/nautobot/nautobot/5097 ) as inputs

gsnider2195:
Look at using https://pypi.org/project/mkdocs-macros-plugin/ for rendering the jinja2

PR #5235, opened 2024-02-05
Create JSON Schema for Core Settings and Expose it via the REST API

### LINKED ISSUE #5097: JSONSchema for core settings
As a Platform Admin, I want a JSONSchema that describes the schema for nautobot settings/configuration so that I can be assured I am producing valid config.

- The schema will describe core settings and mostly target any and settings that would reasonably be configured/manipulated by an end consumer.
- The schema will be versioned with the release of nautobot.
- The schema will be easily accessible as a file, but also available via the REST API.
- The schema is not used for actively validating a nautobot config, but instead for defining what is valid. So no validation tooling is needed related to the schema.

- [ ] **Scenario 1: Accessing JSONSchema file**
  - Given a Platform Admin
  - When they want to access the JSONSchema file
  - Then the file should be easily accessible

- [ ] **Scenario 2: Accessing JSONSchema via REST API**
  - Given a Platform Admin
  - When they want to access the JSONSchema via the REST API
  - Then the JSONSchema should be available through the API

- [ ] **Scenario 3: JSONSchema versioning**
  - Given a Platform Admin
  - When a new release of nautobot is made
  - Then the JSONSchema should be versioned accordingly

- [ ] **Scenario 4: Configuring/manipulating settings**
  - Given an end consumer
  - When they want to configure or manipulate settings
  - Then the JSONSchema should include descriptions for all settings that can be reasonably configured/manipulated

HanlinMiao:
# Closes #5097 
# What's Changed

# Screenshots

Screen Shot 2024-02-05 at 5 22 05 PM

# TODO

- [ ] Unit, Integration Tests


settings.json
```diff
@@ -0,0 +1,600 @@
...
+        "TEST_USE_FACTORIES": {
+            "type": "boolean",
+            "default": false
+        },
+        "TEST_PERFORMANCE_BASELINE_FILE": {
+            "type": "string",
+            "default": "nautobot/core/tests/performance_baselines.yml"
+        },
+        "UI_RACK_VIEW_TRUNCATE_FUNCTION": {},
+        "GIT_SSL_NO_VERIFY": {},
```
HanlinMiao, line 395: How do I express these two settings in a JSON format?

nautobot/core/api/urls.py
```diff
@@ -15,6 +15,7 @@
     GraphQLDRFAPIView,
     NautobotSpectacularRedocView,
     NautobotSpectacularSwaggerView,
+    SettingsJSONFileView,
```
glennmatthews, line 18: Let's call this something like `SettingsJSONSchemaView` for clarity.

nautobot/core/api/urls.py
```diff
@@ -51,6 +52,7 @@
     path("status/", StatusView.as_view(), name="api-status"),
     path("docs/", NautobotSpectacularSwaggerView.as_view(url_name="schema"), name="api_docs"),
     path("redoc/", NautobotSpectacularRedocView.as_view(url_name="schema"), name="api_redocs"),
+    path("settings.json", SettingsJSONFileView.as_view(), name="setting_schema_json"),
```
glennmatthews, line 55: Let's make this something like `settings-schema/` for clarity.

nautobot/core/api/views.py
```diff
@@ -940,3 +941,16 @@ def get(self, request):
...
+
+class SettingsJSONFileView(NautobotAPIVersionMixin, APIView):
+    """View that exposes the JSON Schema of the settings.json file in the REST API"""
+
+    permission_classes = [IsAuthenticated]
+
+    @extend_schema(exclude=True)
+    def get(self, request):
+        file_path = "settings.json"
+        with open(file_path, "r") as jsonfile:
```
glennmatthews, line 954: This should probably be a path relative to `__file__` rather than an un-anchored path.

settings.json
glennmatthews, line 1: This file probably belongs somewhere like `nautobot/core/` rather than in the root of the project.

settings.json
```diff
@@ -0,0 +1,600 @@
+{
+    "type": "object",
+    "properties": {
+        "ALLOWED_HOSTS": {
+            "type": "array",
```
glennmatthews, line 5: It's an array of strings specifically, right? Should be able to specify that in the schema.

settings.json
```diff
@@ -0,0 +1,600 @@
...
+                            "type": "string",
+                            "default": ""
+                        },
+                        "CONN_MAX_AGE": {
+                            "type": "integer",
+                            "default": 300
+                        },
+                        "ENGINE": {
+                            "type": "string",
+                            "default": "django.db.backends.postgresql"
```
glennmatthews, line 40: Wonder if we should specify the list of supported choices here, i.e `[django|django_prometheus].db.backends.[postgresql|mysql]`?

settings.json
```diff
@@ -0,0 +1,600 @@
...
+                        }
+                    }
+                }
+            }
+        },
+        "CACHES": {
```
glennmatthews, line 46: Probably would be good from a maintainability standpoint to alpha-sort the top-level keys?

settings.json
```diff
@@ -0,0 +1,600 @@
+{
+    "type": "object",
+    "properties": {
+        "ALLOWED_HOSTS": {
+            "type": "array",
+            "default": []
+        },
+        "DATABASES": {
```
glennmatthews, line 8: Bearing in mind #5099, at some point we'll need to have an appropriately populated `description` for each of these properties in the schema.

settings.json
```diff
@@ -0,0 +1,600 @@
...
+        },
+        "CACHES": {
+            "type": "object",
+            "properties": {
+                "default": {
+                    "type": "object",
+                    "properties": {
+                        "BACKEND": {
+                            "type": "string",
+                            "default": "django_redis.cache.RedisCache89"
```
glennmatthews, line 54:
```suggestion
                            "default": "django_redis.cache.RedisCache"
```

Here too should we specify the options as either `django_redis.cache.RedisCache` or `django_prometheus.cache.backends.redis.RedisCache`?

settings.json
```diff
@@ -0,0 +1,600 @@
...
+        "BRANDING_URLS": {
+            "type": "object",
+            "properties": {
+                "code": {
+                    "type": "string",
+                    "default": "https://github.com/nautobot/nautobot"
+                },
+                "docs": {
+                    "type": "string",
+                    "default": "<STATIC_URL>docs/index.html"
```
glennmatthews, line 155: Wonder if there's a way for us to make this dynamically correct. Maybe we could treat `settings.json` as a Jinja2 template or something and render it with the appropriate `static_url` as part of the REST endpoint?

settings.json
```diff
@@ -0,0 +1,600 @@
...
+                }
+            }
+        },
+        "CELERY_BROKER_TRANSPORT_OPTIONS": {
+            "type": "object",
+            "default": {}
+        },
+        "CELERY_BROKER_URL": {
+            "type": "string",
+            "default": "redis://localhost:6379/0."
```
glennmatthews, line 169:
```suggestion
            "default": "redis://localhost:6379/0"
```

settings.json
```diff
@@ -0,0 +1,600 @@
...
+        "CELERY_TASK_SOFT_TIME_LIMIT": {
+            "type": "integer",
+            "default": 300
+        },
+        "CELERY_TASK_TIME_LIMIT": {
+            "type": "integer",
+            "default": 600
+        },
+        "CELERY_WORKER_PROMETHEUS_PORTS": {
+            "type": "array",
```
glennmatthews, line 184: array of ints

settings.json
glennmatthews: I suppose similarly we could dynamically sub in in the environment variables as defaults for all of the settings that support environment variable configuration, etc.? Not sure if that's needed or not.

settings.json
glennmatthews: GIT_SSL_NO_VERIFY is an environment variable, not a settings parameter, so it should be fine to omit. Not sure on the function.

settings.json
```diff
@@ -0,0 +1,600 @@
...
+        "INTERNAL_IPS": {
+            "type": "array",
+            "default": [
+                "127.0.0.1",
+                "::1"
+            ]
+        },
+        "LOGGING": {
+            "type": "object",
+            "properties": {
```
glennmatthews, line 449: The LOGGING dict is very configurable so the schema needs to reflect that. At the very least I'd think we need `additioinalProperties: true` at pretty much every level of this object, but it also might be good to move the default values of `formatters`, `handlers`, and `loggers` into just the "default" attribute under each key rather than specifying them as "properties", since `"normal"`, `"verbose"`, `"normal_console"`, `"verbose_console"` are totally arbitrary strings, and under `loggers` potentially any logger name in all of Python/Django/Nautobot/Apps could be used as a key.

settings.json
```diff
@@ -0,0 +1,600 @@
...
+        },
+        "STORAGE_CONFIG": {
+            "type": "object",
+            "default": {}
+        },
+        "STRICT_FILTERING": {
+            "type": "boolean",
+            "default": true
+        },
+        "SUPPORT_MESSAHE": {
```
glennmatthews, line 378:
```suggestion
        "SUPPORT_MESSAGE": {
```

settings.json
```diff
@@ -0,0 +1,600 @@
...
+        },
+        "BANNER_BOTTOM": {
+            "type": "string",
+            "default": ""
+        },
+        "BANNER_LOGIN": {
+            "type": "string",
+            "default": ""
+        },
+        "BRANDING_FILE_PATH": {
```
glennmatthews, line 97:
```suggestion
        "BRANDING_FILEPATHS": {
```

settings.json
HanlinMiao: Just to confirm, something like this?
```
"BACKEND": {
      "type": "string",
      "default": "[django_redis.cache|django_prometheus.cache.backends.redis].RedisCache"
},
```

settings.json
HanlinMiao: Just to confirm, something like this?
```
"ENGINE": {
      "type": "string",
      "default": "[django|django_prometheus].db.backends.[postgresql|mysql]"
},
```

settings.json
glennmatthews: No, the default needs to be a specific value - I was meaning something like https://json-schema.org/understanding-json-schema/reference/enum#enumerated-values

settings.json
glennmatthews: No, something like something like https://json-schema.org/understanding-json-schema/reference/enum#enumerated-values.

nautobot/core/settings.json
```diff
@@ -283,6 +316,69 @@
...
+                        "normal": {
+                            "format": "%(asctime)s.%(msecs)03d %(levelname)-7s %(name)s :\n  %(message)s",
+                            "datefmt": "%H:%M:%S"
+                        },
+                        "verbose": {
+                            "format": "%(asctime)s.%(msecs)03d %(levelname)-7s %(name)-20s %(filename)-15s %(funcName)30s() :\n  %(message)s",
+                            "datefmt": "%H:%M:%S"
+                        }
+                    },
+                    "additioinalProperties": true
```
glennmatthews, line 342:
```suggestion
                    "additionalProperties": true
```

Same throughout.

It would be good for a test to also load this schema into `jsonschema` and ensure that it's a valid schema - a la https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/models/models.py#L276-L279.

nautobot/core/settings.json
```diff
@@ -0,0 +1,544 @@
+{
+    "type": "object",
+    "properties": {
+        "ADMINS": {
+            "type": "array",
```
glennmatthews, line 5: Sorry if I was unclear in the previous comments - IMO every `type: array` in this schema should have a corresponding `items: { type: ... }` to specify what the array contains.

nautobot/core/settings.json
```diff
@@ -0,0 +1,544 @@
...
+        },
+        "RELEASE_CHECK_TIMEOUT": {
+            "type": "integer",
+            "default": 86400
+        },
+        "RELEASE_CHECK_URL": {
+            "type": "string",
+            "default": ""
+        },
+        "SANITIZER_PATTERNS": {},
```
glennmatthews, line 473: Missing `type` here

nautobot/core/tests/test_views.py
```diff
@@ -450,3 +456,31 @@ def test_get_object_with_constrained_permission(self):
...
+            "object": [dict],
+            "integer": [int],
+            "boolean": [bool],
+            "array": [list, tuple],
+            "#/definitions/callable": [types.FunctionType],
+        }
+        for key, value in self.json_data["properties"].items():
+            settings_value = getattr(settings, f"{key}", None)
+            if settings_value:
+                self.assertIn(type(settings_value), TYPE_MAPPING[value.get("type", value.get("$ref", None))])
```
HanlinMiao, line 486: This is currently testing the variables that exist in the `settings.json` file. Some optional settings variables do not exist in `settings.py` and some variables exist in `settings.py` and not in `settings.json`.

nautobot/core/tests/test_views.py
```diff
@@ -450,3 +456,35 @@ def test_get_object_with_constrained_permission(self):
...
+        TYPE_MAPPING = {
+            "string": [str],
+            "object": [dict],
+            "integer": [int],
+            "boolean": [bool],
+            "array": [list, tuple],
+            "#/definitions/callable": [types.FunctionType],
+        }
+        for key, value in self.json_data["properties"].items():
+            settings_value = getattr(settings, f"{key}", None)
```
glennmatthews, line 488:
```suggestion
            settings_value = getattr(settings, key, None)
```

nautobot/core/tests/test_views.py
glennmatthews: I'm surprised that we have variables in the schema that don't exist in some form in `settings.py` - I guess those might include the Constance-capable settings, but what others are missing?

nautobot/core/settings.json
```diff
@@ -0,0 +1,598 @@
...
+            "default": "N j, Y g:i a"
+        },
+        "DEBUG": {
+            "type": "boolean",
+            "default": false
+        },
+        "DEPLOYMENT_ID": {
+            "type": "string"
+        },
+        "DISABLE_PREFIX_LIST_HIERARCHY": {
```
glennmatthews, line 274: No longer supported in v2.x. Looks like we have some cleanup needed:

```
nautobot/core/templates/nautobot_config.py.j2:# DISABLE_PREFIX_LIST_HIERARCHY = False
nautobot/docs/user-guide/administration/configuration/optional-settings.md:## DISABLE_PREFIX_LIST_HIERARCHY
```

nautobot/core/settings.json
```diff
@@ -0,0 +1,598 @@
...
+            "type": "boolean",
+            "default": false
+        },
+        "DYNAMIC_GROUPS_MEMBER_CACHE_TIMEOUT": {
+            "type": "integer",
+            "default": 0
+        },
+        "ENFORCE_GLOBAL_UNIQUE": {
```
glennmatthews, line 282: No longer supported in v2.x. Looks like we have more cleanup needed:

```
nautobot/docs/user-guide/administration/configuration/optional-settings.md:## ENFORCE_GLOBAL_UNIQUE
nautobot/docs/user-guide/administration/configuration/optional-settings.md:Environment Variable: `NAUTOBOT_ENFORCE_GLOBAL_UNIQUE`
nautobot/docs/user-guide/administration/configuration/optional-settings.md:By default, Nautobot will permit users to create duplicate prefixes and IP addresses in the global table (that is, those which are not assigned to any VRF). This behavior can be disabled by setting `ENFORCE_GLOBAL_UNIQUE` to `True`.

```

nautobot/core/settings.json
```diff
@@ -0,0 +1,598 @@
...
+            },
+            "default": []
+        },
+        "CELERY_WORKER_REDIRECT_STDOUTS": {
+            "type": "boolean",
+            "default": true
+        },
+        "CELERY_WORKER_REDIRECT_STDOUTS_LEVEL": {
+            "type": "string",
+            "default": "WARNING"
```
glennmatthews, line 176: Should be possible to provide an enum here since this is a logging level.

nautobot/core/settings.json
```diff
@@ -0,0 +1,598 @@
...
+        },
+        "BANNER_TOP": {
+            "type": "string",
+            "default": ""
+        },
+        "BRANDING_FILEPATHS": {
+            "type": "object",
+            "properties": {
+                "logo": {
+                    "type": "string",
```
glennmatthews, line 42: Wonder if it'd be useful to define a `relative_path` type to use for these properties and similar ones?

nautobot/core/settings.json
```diff
@@ -0,0 +1,598 @@
...
+        "BRANDING_TITLE": {
+            "type": "string",
+            "default": "Nautobot"
+        },
+        "BRANDING_URLS": {
+            "type": "object",
+            "properties": {
+                "code": {
+                    "type": "string",
+                    "default": "https://github.com/nautobot/nautobot"
```
glennmatthews, line 92: Wonder if it'd be useful to define a `url` type for this property and similar ones? Actually, I think jsonschema may have a native type for this already?

nautobot/core/settings.json
```diff
@@ -0,0 +1,598 @@
...
+            "type": "array",
+            "items": {
+                "type": "string"
+            },
+            "default": []
+        },
+        "DATABASES": {
+            "type": "object",
+            "properties": {
+                "default": {
```
glennmatthews, line 218: Possible to have additional databases beyond `default`, so be sure to have `additionalProperties: true` on this one.

More generally, it'd probably be good to make sure we permit the additional configs for the version-control app (https://github.com/nautobot/nautobot-app-version-control/?tab=readme-ov-file#configuring-nautobot-to-use-version-control).

nautobot/core/settings.json
```diff
@@ -0,0 +1,598 @@
...
+        "EXTERNAL_AUTH_DEFAULT_PERMISSIONS": {
+            "type": "object",
+            "default": {}
+        },
+        "FORCE_SCRIPT_NAME": {
+            "type": "string",
+            "default": ""
+        },
+        "GIT_ROOT": {
+            "type": "string",
```
glennmatthews, line 309: Maybe an `absolute_path` type too?

nautobot/core/settings.json
HanlinMiao: Looking through the doc [here](https://json-schema.org/understanding-json-schema/reference/type), I dont see a type for url. I think I will make a new type for that in this schema unless I am missing something

glennmatthews: https://json-schema.org/understanding-json-schema/reference/string#built-in-formats

nautobot/core/settings.json
HanlinMiao: https://github.com/nautobot/nautobot/pull/5235/commits/35537e938779da6d2316a811d4986f6d927487d5

nautobot/core/tests/test_views.py
HanlinMiao: `CELERY_BROKER_TRANSPORT_OPTIONS`
`CONFIG_CONTEXT_DYNAMIC_GROUPS_ENALBED`
`SESSION_ENGINE`

Nautobot Version Control App Settings:

`DATABASE_ROUTERS`
`SESSION_CACHE_ALIAS`

HanlinMiao: Most of the untested setting variables are included in `CONSTANCE_CONFIG` though.

nautobot/core/settings.json
```diff
@@ -137,80 +167,100 @@
...
         "CORS_ALLOWED_ORIGINS": {
             "type": "array",
             "items": {
-                "type": "string"
+                "type": "string",
+                "format": "uri"
             },
-            "default": []
+            "default": [],
+            "description": "If True, all origins will be allowed. Other settings restricting allowed origins will be ignored."
```
glennmatthews, line 247: Looks like this should belong to `CORS_ALLOW_ALL_ORIGINS`, not `CORS_ALLOWED_ORIGINS`?

nautobot/core/settings.json
```diff
@@ -14,92 +20,115 @@
             "items": {
                 "type": "string"
             },
-            "default": []
+            "default": [],
+            "description": "List of administrators' names and emails. e.g.[['Hank Hill', 'hhill@example.com'],['Dale Gribble', 'dgribble@example.com'],]"
```
glennmatthews, line 24: Check the `items` here vs the description - is this an array of strings, or an array of arrays of strings?

nautobot/core/tests/test_views.py
glennmatthews: Might be good to have an explicit allow-list in the test for the above variables? So we can otherwise catch any case where the settings and the schema drift out of sync?

HanlinMiao: Just to clarify, this allow-list is to catch setting variables that exist outside of those specified in the JSON schema right?

nautobot/core/settings.json
HanlinMiao: It is an array of arrays that contains two items username and password

HanlinMiao: https://github.com/nautobot/nautobot/pull/5235/commits/0e0cfb47639628f6bbcfae328af99285ccb9c2a2 this should address the change.

nautobot/core/tests/test_views.py
```diff
@@ -452,6 +459,155 @@ def test_get_object_with_constrained_permission(self):
...
+        try:
+            Draft7Validator.check_schema(self.json_data)
+        except SchemaError as e:
+            raise ValidationError({"data_schema": e.message})
+
+    def test_settings_json_schema_contains_valid_setting_variables(self):
+        """Test the validity of the settings variables from settings.json and their types with those in settings.py"""
+        # This list contains all variables that exist in settings.py but not in the JSON schema in settings.json.
+        # ADMINS does not exist in either settings.json or settings.py
+        ALLOWED_SETTINGS = [
```
glennmatthews, line 482: Oof, that's a longer list than I'd hoped for. Ah well.

nautobot/core/tests/test_views.py
```diff
@@ -452,6 +459,155 @@ def test_get_object_with_constrained_permission(self):
...
+            "ADMINS",
+            "ALLOW_REQUEST_PROFILING",
+            "ALLOWED_URL_SCHEMES",
+            "AUTHENTICATION_BACKENDS",
+            "AUTH_USER_MODEL",
+            "BASE_DIR",
+            "BRANDING_POWERED_BY_URL",
```
glennmatthews, line 489: Should add this to the schema.

nautobot/core/tests/test_views.py
```diff
@@ -452,6 +459,155 @@ def test_get_object_with_constrained_permission(self):
...
+            "CELERY_RESULT_BACKEND",
+            "CELERY_RESULT_EXPIRES",
+            "CELERY_RESULT_EXTENDED",
+            "CELERY_RESULT_SERIALIZER",
+            "CELERY_TASK_SEND_SENT_EVENT",
+            "CELERY_TASK_SERIALIZER",
+            "CELERY_TASK_TRACK_STARTED",
+            "CELERY_WORKER_SEND_TASK_EVENTS",
+            "CONFIG_CONTEXT_DYNAMIC_GROUPS_ENABLED",
+            "CONFIG_CONTEXT_DYNAMIC_GROUPS_ENALBED",
```
glennmatthews, line 503: typo?

nautobot/core/tests/test_views.py
```diff
@@ -452,6 +459,155 @@ def test_get_object_with_constrained_permission(self):
...
+            "DATABASE_ROUTERS",
+            "DATA_UPLOAD_MAX_NUMBER_FIELDS",
+            "DEFAULT_AUTO_FIELD",
+            "DRF_REACT_TEMPLATE_TYPE_MAP",
+            "EXEMPT_EXCLUDE_MODELS",
+            "FILTERS_NULL_CHOICE_LABEL",
+            "FILTERS_NULL_CHOICE_VALUE",
+            "GRAPHENE",
+            "HOSTNAME",
+            "HOSTNAME",
```
glennmatthews, line 521: duplicate?

nautobot/core/tests/test_views.py
```diff
@@ -452,6 +459,155 @@ def test_get_object_with_constrained_permission(self):
...
+            "INSTALLED_APPS",
+            "LANGUAGE_CODE",
+            "LOGIN_REDIRECT_URL",
+            "LOGIN_URL",
+            "LOG_DEPRECATION_WARNINGS",
```
glennmatthews, line 526: Should be in the schema.

nautobot/core/tests/test_views.py
```diff
@@ -452,6 +459,155 @@ def test_get_object_with_constrained_permission(self):
...
+            "REMOTE_AUTH_HEADER",
+            "REST_FRAMEWORK",
+            "REST_FRAMEWORK_ALLOWED_VERSIONS",
+            "REST_FRAMEWORK_VERSION",
+            "ROOT_URLCONF",
+            "SECURE_PROXY_SSL_HEADER",
+            "SESSION_CACHE_ALIAS",
+            "SESSION_ENGINE",
+            "SHELL_PLUS_DONT_LOAD",
+            "SHORT_TIME_FORMAT",
```
glennmatthews, line 542: Add to schema?

nautobot/core/tests/test_views.py
```diff
@@ -452,6 +459,155 @@ def test_get_object_with_constrained_permission(self):
...
+            "STATICFILES_DIRS",
+            "STATIC_URL",
+            "TEMPLATES",
+            "TESTING",
+            "TEST_RUNNER",
+            "USE_I18N",
+            "USE_TZ",
+            "USE_X_FORWARDED_HOST",
+            "VERSION",
+            "VERSION",
```
glennmatthews, line 563: Duplicate?

nautobot/core/tests/test_views.py
```diff
@@ -452,6 +459,155 @@ def test_get_object_with_constrained_permission(self):
...
+            "WEBSERVER_WARMUP",
+            "WSGI_APPLICATION",
+            "X_FRAME_OPTIONS",
+        ]
+
+        # Retrieve all variables from settings.py.
+        # and trim out all noises by retaining variables with only uppercase letters and "_".
+        existing_settings_variables = list(settings.__dict__.keys())
+        existing_settings_variables = [
+            word for word in existing_settings_variables if all([letter in ascii_uppercase + "_" for letter in word])
```
glennmatthews, line 575: Perhaps simpler?

```suggestion
            word for word in existing_settings_variables if word == word.upper() and not word.startswith("_"))
```

nautobot/core/tests/test_views.py
glennmatthews: Maybe something like `UNDOCUMENTED_SETTINGS` might be a better name?

nautobot/core/tests/test_views.py
```diff
@@ -452,6 +459,155 @@ def test_get_object_with_constrained_permission(self):
...
+
+        # Check if there is any unsupported setting variable in settings.py.
+        for variable in existing_settings_variables:
+            if variable not in existing_json_schema_variables + ALLOWED_SETTINGS:
+                self.fail(f"Unsupported settings variable {variable} detected in nautobot/core/settings.py")
+
+        # Check if there is any unsupported setting variable in settings.json.
+        for variable in self.json_data["properties"].keys():
+            if variable not in existing_settings_variables + existing_constance_config_variables + ALLOWED_SETTINGS:
+                self.fail(f"Unsupported settings variable {variable} detected in nautobot/core/settings.json")
```
glennmatthews, line 591:
```suggestion
        # Check if there is any undocumented setting variable in settings.py.
        expected_variables = set(existing_json_schema_variables) & set(ALLOWED_SETTINGS)
        for variable in existing_settings_variables:
            if variable not in expected_variables:
                self.fail(f"Undocumented settings variable {variable} detected in nautobot/core/settings.py")

        # Check if there is any nonexistent settings variable in settings.json.
        expected_variables = set(existing_settings_variables) & set(existing_constance_config_variables) & set(ALLOWED_SETTINGS)
        for variable in self.json_data["properties"].keys():
            if variable not in expected_variables:
                self.fail(f"Nonexistent settings variable {variable} detected in nautobot/core/settings.json")
```

HanlinMiao: The combinator here `set(existing_json_schema_variables) & set(ALLOWED_SETTINGS)` should be "||" right?

glennmatthews: Yeah, sorry :-)  Good catch.

glennmatthews (approved):
Fix ruff and I think this is good to go!

HanlinMiao:
> Fix ruff and I think this is good to go!

Thank you Glenn for the help and guidance!🙏

PR #5236, opened 2024-02-05
Added annotations to document the importance of keeping the TIME_ZONE…

### LINKED ISSUE #3349: Scheduled Jobs > Crontab corresponds to UTC

gsnider2195:
… setting consistent on Nautobot web servers and Celery Beat servers.

# Closes #3349
# What's Changed

# Screenshots

image

image

image

# TODO

- [ ] Unit, Integration Tests


itdependsnetworks:
As mentioned [here](https://github.com/nautobot/nautobot/issues/3349#issuecomment-1915927784) I think we should document beyond just the setting in the nautobot_config.py file and seriously recommend using only UTC on everything.

gsnider2195:
> As mentioned [here](https://github.com/nautobot/nautobot/issues/3349#issuecomment-1915927784) I think we should document beyond just the setting in the nautobot_config.py file and seriously recommend using only UTC on everything.

What's the purpose of setting the system time zone and database time zone?

itdependsnetworks:
> > As mentioned [here](https://github.com/nautobot/nautobot/issues/3349#issuecomment-1915927784) I think we should document beyond just the setting in the nautobot_config.py file and seriously recommend using only UTC on everything.
> 
> What's the purpose of setting the system time zone and database time zone?

To keep all logs in sync, this is common issue we are seeing.

nautobot/docs/user-guide/administration/installation/services.md
```diff
@@ -195,6 +195,9 @@ WantedBy=multi-user.target
 The Celery Beat scheduler enables the periodic execution of and scheduling of background tasks. It is required to take
 advantage of the [job scheduling and approval](../../platform-functionality/jobs/job-scheduling-and-approvals.md) features.
 
+!!! warning
+    It's important that the [`TIME_ZONE`](../configuration/optional-settings.md#time_zone) setting on your Nautobot servers and Celery Beat server match to prevent scheduled jobs from running in the wrong time zone.
```
glennmatthews, line 199: Perhaps:

```suggestion
    It's important that the [`TIME_ZONE`](../configuration/optional-settings.md#time_zone) setting on your Nautobot servers and Celery Beat server match to prevent scheduled jobs from running at the wrong time.
```

nautobot/docs/user-guide/administration/configuration/optional-settings.md
```diff
@@ -1367,6 +1367,9 @@ Default: `"UTC"`
 
 Environment Variable: `NAUTOBOT_TIME_ZONE`
 
+!!! warning
+    Scheduled jobs will run in the time zone configured in this setting. If you change this setting from the default UTC, you must change it on the Celery Beat server and all Nautobot web servers or your scheduled jobs may run in the wrong time zone.
```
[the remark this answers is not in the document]
gsnider2195: That shouldn't be an issue because Django always saves DateTimeFields to the database as UTC. The time zone conversion happens in Python.

glennmatthews (approved):

gsnider2195:
> > > As mentioned [here](https://github.com/nautobot/nautobot/issues/3349#issuecomment-1915927784) I think we should document beyond just the setting in the nautobot_config.py file and seriously recommend using only UTC on everything.
> > 
> > 
> > What's the purpose of setting the system time zone and database time zone?
> 
> To keep all logs in sync, this is common issue we are seeing.

We do recommend using UTC for Django's time zone [here](https://docs.nautobot.com/projects/core/en/stable/user-guide/administration/configuration/optional-settings/#time_zone). With the Celery beat TIME_ZONE setting we know that only one setting will work so we can make the recommendation to use it. With those other settings there's more nuance and would be more suited for a deployment doc where we could explain why you would or wouldn't want to change them. For example if you're using a shared SQL server your DBA wouldn't be happy if you changed the default time zone.

lampwins:
@gsnider2195 in addition to the sections you updated, can we also add a new file in the config section dedicated to time that captures the intent that Ken is describing? The main problem we are running into is that we don't have it plainly spelled out in the docs (in a format that we can easily point a user to) that explains our default recommendation that all of the time related knobs be the same. We don't need to get into the nuance of each, knowledgable administrators will always have their reasons for doing what they do, but we need to make our stance more clear, from a support standpoint.

glennmatthews (approved):

nautobot/docs/user-guide/administration/configuration/time-zones.md
```diff
@@ -0,0 +1,29 @@
+# Time Zones
```
itdependsnetworks, line 1: My thoughts

# Time Zones

Time Zones effect several aspects of the of your Nautobot implementation to include:

* Logging timestamps across all services
* Changelogging timestamps in the database
* The UI time presented
* Scheduling of jobs

It's important to make sure that the time zones match on all of your Nautobot services, including the database server, Nautobot server, Celery Beat and any worker on both the server itself as well as the application settings. This includes setting the system time zone on all of the servers and the configured time zone on all of the service configurations. 

We highly recommend using UTC across all spaces to avoid any issues with daylight savings time changes or misalignment in timestamps. Leverage NTP across all servers and services to ensure the times are synchronized. 

## Nautobot Application Configuration

Nautobot supports the [`TIME_ZONE`](optional-settings.md#time_zone) setting, which will set the default time zone for the Nautobot service, worker, and scheduler.

A common misconfiguration is the Celery Beat and worker servers' `TIME_ZONE` setting. Since these services are running within the Nautobot process they also need this configuration set, either by setting the `NAUTOBOT_TIME_ZONE` environment variable or using the same `nautobot_config.py` file as the Nautobot server. You should also change the Database service's configured time zone to match the Nautobot server's time zone.

## MySQL Time Zone Configuration

The MySQL database's default time zone can be configured through the configuration file usually found in `/etc/mysql/my.cnf` or `/etc/my.cnf`:

```ini
[mysqld]
default-time-zone = "+00:00"
```

This can also be changed with a SQL query:

```sql
SET GLOBAL time_zone = '+00:00';
```

## PostgreSQL Time Zone Configuration

The PostgreSQL database's default time zone can be configured through the configuration file. The location of this file can be found with the SQL query `SHOW config_file;`.

```no-highlight
timezone = UTC
```

!!! info
    The SQL command `SET TIME ZONE` only sets the time zone for the session.


## Server Configuration

While outside the scope of this document to provide the time configuration for every flavor of server, included is an example for Ubuntu, please see your own operating system guide for further information.

```bash
timedatectl set-ntp yes
timedatectl set-timezone UTC
```

gsnider2195: > Changelogging timestamps in the database

It's my understanding that django always stores the time zones in the database as UTC. Have you seen otherwise?

itdependsnetworks: Sorry, not really following the question? I don't think I implied otherwise

glennmatthews (approved):

PR #5237, opened 2024-02-06
feature: implement jobs tile view

### LINKED ISSUE #5129: Jobs tile view
As a network automation engineer I want a tile based view of jobs as an alternative to the table list view, so that I have a more elegant way to interact with my automations.

I will know this is complete when a user has the ability to toggle between two view options in the jobs list view:
Screenshot 2024-01-18 at 11 11 29 AM

Create a card element for jobs like this:
Screenshot 2024-01-18 at 11 10 00 AM

Create a view which shows these cards, grouped by the existing grouping headers.
Screenshot 2024-01-18 at 11 12 52 AM

Out of scope, but in the future, we will do things like allow users to move and reorder the cards in the view to suit their preferences, and save favorited job cards to the top of the page.

norbert-mieczkowski-codilime:
# Closes #5129
# What's Changed

- Create job list `display` query parameter with two variants `tiles` and `list` and save it in user preferences at `extras.job.display`
- Create common CSS classes for `tile` "component"
- Implement jobs tile view next to an existing list view as a skin for `django-tables2`
- Fix buttons with tooltips wrapped in `btn-group` containers
# Screenshots

| Light theme | Dark theme |
| - | - |
| tiles_light | tiles_dark |


nautobot/extras/views.py
```diff
@@ -1253,6 +1253,7 @@ class JobListView(generic.ObjectListView):
     filterset = filters.JobFilterSet
     filterset_form = forms.JobFilterForm
     action_buttons = ()
+    non_filter_params = "view"
```
norbert-mieczkowski-codilime, line 1256: I wanted this value here to be a tuple but ruff refused :man_shrugging:

nautobot/extras/views.py
```diff
@@ -1270,8 +1271,19 @@ def alter_queryset(self, request):
         return queryset
 
     def extra_context(self):
+        # Determine user's preferred view
+        if self.request.GET.get("view") in ["list", "tiles"]:
+            view = self.request.GET.get("view")
+            if self.request.user.is_authenticated:
+                self.request.user.set_config("extras.joblistview.view", view, commit=True)
```
norbert-mieczkowski-codilime, line 1278: Not sure if `extras.joblistview.view` is a good key, looking for confirmation

nautobot/extras/templates/extras/inc/job_tiles.html
```diff
@@ -0,0 +1,72 @@
...
+                {% endcomment %}
+                <div class="btn-group">
+                    <button type="button" class="btn btn-default btn-xs dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" title="Show more">
+                        <i class="mdi mdi-dots-horizontal"></i>
+                    </button>
+                    <ul class="dropdown-menu dropdown-menu-right context-menu">
+                        <li><a href="{% url 'extras:job' pk=row.record.pk %}"><i class="mdi mdi-information-outline text-muted"></i>Details</a></li>
+                        <li><a href="{% url 'extras:job_changelog' pk=row.record.pk %}"><i class="mdi mdi-history text-muted"></i>Change log</a></li>
+                        {% if perms.extras.change_job %}
+                            <li><a href="{% url 'extras:job_edit' pk=row.record.pk %}"><i class="mdi mdi-pencil text-muted"></i>Edit</a></li>
```
glennmatthews, line 37: Should this use `text-warning` to match the coloring of "edit" buttons elsewhere?

nautobot/extras/templates/extras/inc/job_tiles.html
```diff
@@ -0,0 +1,72 @@
+{% load helpers %}
+
+{% block extra_styles %}
+    <style>
+        .panel.panel-default {
```
glennmatthews, line 5: Where is this style being used in this template?

nautobot/extras/views.py
glennmatthews: Were you trying `("view")`? Try `("view",)` instead.

nautobot/extras/views.py
glennmatthews: We don't have a real consistent pattern here. The table config form uses `tables.<ClassName>` to store table column selection, while other views in this file use `extras.<model_name>.format` to store JSON vs YAML rendering preference. I do think `display` might be a better leaf key than `view` though. Given that, I'd suggest either `views.JobListView.display` or `extras.job.display` - no strong preference either way.

nautobot/project-static/css/base.css
```diff
@@ -755,3 +755,57 @@ pre code.hljs {
...
+.context-menu a:hover i,
+.context-menu a:hover span {
+    color: inherit;
+}
+
+/* Tile component styles */
+.tiles {
+    display: grid;
+    grid-gap: 20px;
+    grid-template-columns: repeat(auto-fit, minmax(445px, 1fr));
```
glennmatthews, line 774: What's the significance of 445px here?

nautobot/extras/views.py
norbert-mieczkowski-codilime: That did the trick, thanks!

nautobot/extras/views.py
norbert-mieczkowski-codilime: After all I decided to go with `extras.job.display` and overall changed `view` param name to `display`. Thanks for suggestions

nautobot/extras/templates/extras/inc/job_tiles.html
norbert-mieczkowski-codilime: I'm not sure, I personally like the way it looks now better. I'll defer the decision to you and @lampwins though, here's comparison of the two:

| Default | Warning |
| - | - |
| edit_default | edit_warning |

nautobot/extras/templates/extras/inc/job_tiles.html
norbert-mieczkowski-codilime: Job list content is wrapped in `panel panel-default` (via `panel_table.html`) element which works for the default job table display. In case of tiles, though, we don't want any borders around them and that's essentially what panels do, so I override those styles here. If there is a better way to fix this, I'll be happy to make it right. Here's how tiles display would look like without overriding panel `border` and `box-shadow`:
panel_border

nautobot/project-static/css/base.css
norbert-mieczkowski-codilime: `445px` is the default tile width specified in Figma design. The way template colums attribute is set here for the tile grid, it squeezes in a single row as many `445px` tiles as it can, and then stretches them equally to fill the remaining space.

To give it more context, `auto-fit` has to be paired with a specific value (either absolute or relative, but not `auto`), be it `445px` or whatever else, otherwise it is ignored by the browser. The `minmax(..., 1fr)` part makes tiles stretch, if it wasn't for `minmax`, they would always be of static width like this:
tiles_static

jedelman8: This image looks better given the width of the tiles. If we should change default px, we shouldn't need to stick to what was in Figma.

jedelman8:
It will also be helpful to test this with loading in all the open source Nautobot Apps b/c some ship with Jobs, and some have longer descriptions.  Should do that to test sizing/spacing, etc. 

Two other items (maybe for future):

1. Instead of showing "Last Status: ____" in bottom, right, how about the last 5 runs using green/red rectangles, like this?  For reference, that's in this [App](https://github.com/nautobot/nautobot-app-ssot)

image



2. It looks like there aren't any sections that separate Jobs and we'd have one long tiled-list.  Can we add a filter that would be a dropdown to filter based on source (basically what each section is today in the list view) ..

nautobot/project-static/css/base.css
norbert-mieczkowski-codilime: @jedelman8, the tiles are `445px` wide in the image above with disabled stretching, so the only difference is they do not try to fill the entire space available like they do in the image in the main description. If you like it better, stretching can be disabled. What do you think @lampwins?

glennmatthews:
I'm seeing a rendering bug in the last row of tiles when the jobs don't have long descriptions - the drop-down menu is cut off by the surrounding panel:

image

glennmatthews:
> 2. It looks like there aren't any sections that separate Jobs and we'd have one long tiled-list.  Can we add a filter that would be a dropdown to filter based on source (basically what each section is today in the list view) ..

Filtering by grouping is already supported in the "advanced" filter UI:

image

image

nautobot/project-static/css/base.css
glennmatthews: Can/should we be using Bootstrap's `row`/`col` grid system for the tiles instead?

nautobot/extras/templates/extras/inc/job_tiles.html
```diff
@@ -0,0 +1,72 @@
...
+
+                {% comment "TODO(norbert-mieczkowski-codilime): use the same buttons (actions) in tables and tiles when tables support dropdown menus" %}
+                    {% for item in row.items %}
+                        {% if item.0.name == "actions" %}
+                            {{ item.1 }}
+                        {% endif %}
+                    {% endfor %}
+                {% endcomment %}
+                <div class="btn-group">
+                    <button type="button" class="btn btn-default btn-xs dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" title="Show more">
```
glennmatthews, line 30: Maybe `Additional Links` or `Actions` would be better than `Show more`? Not sure.

nautobot/project-static/css/base.css
norbert-mieczkowski-codilime: That's what I also considered initially but Bootstrap grid system won't keep the same height of tiles in each row and as a result column wrapping may be off in some cases (see attached screenshot). Without custom scripting, current tiles behavior and appearance are only possible with CSS grid as far as I'm aware.

tiles-bootstrap

nautobot/extras/templates/extras/inc/job_tiles.html
norbert-mieczkowski-codilime: @lampwins, which text would you prefer?

lampwins: I like `Actions` and I think we even call these `action_buttons` in other areas of the app.

nautobot/extras/templates/extras/inc/job_tiles.html
lampwins: Yeah I like the original. This is a detail we can iterate on later when we work to expand this to all tables.

nautobot/project-static/css/base.css
lampwins: I prefer the way it is now with the height adjusting to the size of the tile content as you demoed earlier in the week. No doubt we will iterate on this in the not so distant future, hopefully with upgraded UI technology more available to us than bootstrap 3.

As for the width, I get the point Jason is making, but I also want to pick a value which results in a reasonable amount of tiles horizontally, for a given standard window size. For me, that is no more than 4 across, so I feel what we have is sufficient. Again, perhaps in the future, we can iterate to make it more dynamic.

glennmatthews (approved):

changes/5129.added
```diff
@@ -0,0 +1 @@
+Implemented job list tile view.
```
glennmatthews, line 1: Let's call this out in `nautobot/docs/release-notes/version-2.2.md` as a significant new feature.

PR #5242, opened 2024-02-07
Enabled sorting on the API endpoints for tree node models. 

### LINKED ISSUE #4811: Allow sorting of Tree like model list views
### Environment
* Nautobot version (Docker tag too if applicable): 2.0
* Python version: 3.11
* Middleware(s): N/A

### Steps to Reproduce
1. Go to https://demo.nautobot.com/ipam/prefixes/
2. Click on headers
3.

### Expected Behavior

To sort based on the column header you created

### Observed Behavior

No sorting

timizuoebideri1:
# Closes #4811 
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests


nautobot/core/tests/test_api.py
```diff
@@ -805,16 +815,60 @@ def test_descending_sort(self):
...
+    @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
+    def test_sorting_tree_node_models(self):
+        location_type = dcim_models.LocationType.objects.get(name="Campus")
+        locations = dcim_models.Location.objects.filter(location_type=location_type)
+        devices = dcim_models.Device.objects.all()
+
+        # When sorting by the 'parent' field, we may encounter different results from API response and ORM filter due to None values.
+        # For instance, sorting tenants on `parent` = `None` might yield either `<Tenant 1 parent=None>, <Tenant 2 parent=None>`
+        # or `<Tenant 2 parent=None>, <Tenant 1 parent=None>`.
+        # To ensure consistent sorting, we'll delete all None parents except one.
```
timizuoebideri1, line 829: My best guess is this might have something to do with `TreeNode` approach to sorting.

glennmatthews: That's concerning - ordering should be consistent when using sorting. Let's dig into this further if we can.

glennmatthews: This actually makes sense - when our `order_by` criterion is something with the same value across multiple objects, we have no guarantee what relative order those same-value objects will be given.

nautobot/core/tests/test_api.py
```diff
@@ -805,16 +817,64 @@ def test_descending_sort(self):
...
+        ]
+        # Each of the table has at-least two sortable field_names in the field_names
+        fk_fields = ["location", "parent", "location_type", "manufacturer"]
+        model_field_names = ["name", *fk_fields]
+        for model_class in tree_node_models:
+            url = reverse(get_route_for_model(model_class, "list", api=True))
+            serializer = get_serializer_for_model(model_class)
+            serializer_avial_fields = set(model_field_names) & set(serializer().fields.keys())
+            for field_name in serializer_avial_fields:
+                with self.subTest(f'Asset sorting "{model_class.__name__}" using "{field_name}" field name.'):
```
glennmatthews, line 861:
```suggestion
            serializer_avail_fields = set(model_field_names) & set(serializer().fields.keys())
            for field_name in serializer_avail_fields:
                with self.subTest(f'Assert sorting "{model_class.__name__}" using "{field_name}" field name.'):
```

nautobot/core/tests/test_api.py
```diff
@@ -805,16 +817,64 @@ def test_descending_sort(self):
...
+                    response = self.client.get(f"{url}?sort={field_name}&limit=10&depth=1", **self.header)
+                    self._validate_sorted_response(
+                        response=response,
+                        queryset=model_class.objects.extra(order_by=[field_name]),
+                        field_name=field_name,
+                        is_fk_field=field_name in fk_fields,
+                    )
+
+                with self.subTest(f'Asset inverse sorting "{model_class.__name__}" using "{field_name}" field name.'):
```
glennmatthews, line 870:
```suggestion
                with self.subTest(f'Assert inverse sorting "{model_class.__name__}" using "{field_name}" field name.'):
```

nautobot/core/tests/test_api.py
```diff
@@ -780,23 +783,32 @@ def setUpTestData(cls):
...
         }
 
+    def _validate_sorted_response(self, response, queryset, field_name, is_fk_field=False):
+        self.assertHttpStatus(response, 200)
+
+        # If the field is a foreign key field, we cannot guarantee the relative order of objects with the same value across multiple objects.
+        # Therefore, we directly compare the names of the foreign key objects.
+        if is_fk_field:
+            api_data = list(map(lambda p: p[field_name]["name"] if p[field_name] else None, response.data["results"]))
+            queryset_data = list(queryset.values_list(f"{field_name}__name", flat=True)[:10])
```
glennmatthews, line 793: I don't really understand how this part works or the need for it. If the FK has the same value (`id`), surely it will also have the same `name`? It seems like this branch is asserting that the *nested* objects are ordered consistently, instead of asserting that the base objects are ordered consistently - is that the intent, or what am I misunderstanding?

nautobot/core/tests/test_api.py
```diff
@@ -18,13 +18,16 @@
...
+from nautobot.core.api.utils import get_serializer_for_model
 from nautobot.core.api.versioning import NautobotAPIVersioning
 from nautobot.core.constants import COMPOSITE_KEY_SEPARATOR
+from nautobot.core.utils.lookup import get_route_for_model
 from nautobot.dcim import models as dcim_models
 from nautobot.dcim.api import serializers as dcim_serializers
 from nautobot.extras import choices, models as extras_models
 from nautobot.ipam import models as ipam_models
 from nautobot.ipam.api import serializers as ipam_serializers
+from nautobot.tenancy.models import TenantGroup
```
glennmatthews, line 30: Would be good to follow the established pattern here, i.e.:
```suggestion
from nautobot.tenancy import models as tenancy_models
```

and use `tenancy_models.TenantGroup` below.

glennmatthews (approved):

Issue #2668, opened 2022-10-21
Allow `DynamicMultipleChoiceField` to display an arbitrary combination of the models fields


Kircheneer:
### As ...

P.D. - Plugin Developer

### I want ...

`DynamicMultipleChoiceField` to allow me to choose how I want to display each item.

### So that ...

users can clearly see which object they add (e.g. there might be two prefixes with the same prefix but different VRFs, how does the user know which is the correct one?).

### I know this is done when...

I use a `DynamicMultipleChoiceField` in a form and it allows me to customize what exactly I want to display on the dropdown. This might for example be `f"{obj.prefix} ({obj.vrf.name})"` for a `Prefix` field.

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [X] Plugins (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

No

### External Dependencies

No

joaopsys:
This would be very helpful to me as well. We have cases where we have an `alias` field assigned to objects and it would be convenient to show that field together with the name to users

scetron:
## I want ...

MultiObjectVar or ObjectVar to allow me to choose how I want to display each item.

## So I can ...

Better differentiate between objects from other selectors.

For instance, another field selects a number of devices. I can see the interfaces from these devices, but I am not sure which device an interface belongs to.

PR #5244, opened 2024-02-07
Enhance resolvePath() to support nested fields

### LINKED ISSUE #5243: Support nested values for `display_field` in `DynamicModelChoiceField`
### As ...

P.D. - Plugin Developer

### I want ...

Currently, when the UI presents the end-user with a list of dynamically generated choices (for example list of available vlans on the interface), the list is rendered using the `display` attribute of the results returned by the backend API.  The request to the API is made by the machinery included in [select2](https://select2.org) library and [processed on the Nautobot side](https://github.com/nautobot/nautobot/blob/25cef5bb1e4126fcea2801e9b0685d542399e5ab/nautobot/project-static/js/forms.js#L260) to formulate how the list of choices looks like. By default, it uses the `display` value of each record in the response.

This behavior is not always desirable and in some cases, does not provide enough information to the end-user. One particular use case we are currently facing involves selecting a VLAN while editing the Interface, where the list of available choices includes VLAN 200 from VlanGroup "A" and VLAN 200  from VlanGroup "B". With the default behavior, the list of rendered choices just lists "VLAN 200" twice with no way of clarifying which is which.

As a plugin developer, I was able to override a `dcim:interface_edit` view to replace a model form with a custom form. This custom form uses an adjusted `DynamicModelChoiceField` with a custom `display_field`. This is what is possible with the current version of Nautobot at the moment. 
Unfortunately, in its current form `display_field` allows accessing only root-level entries in the API's response. In other words, one can access `id`, `name`, and `display` attributes, but cannot access `vlan_group.name`.

### So that ...

Plugin developers should be able to specify custom display fields, such as `computed_fields.description_with_vg` or `vlan_group.name` instead of just top-level fields.

I would like to propose a small change (PR incoming) in `processResult()` that would allow accessing arbitrary fields in the API response. During our testing, we  used `computedFields.description_with_vg` successfully after modifying's nautobot code and adding the below code to our plugin:

```python
# view_overrides/views.py
from nautobot.core.forms.fields import DynamicModelChoiceField
from nautobot.dcim.forms import InterfaceForm
from nautobot.dcim.views import InterfaceEditView
from nautobot.ipam.models import VLAN


class MyInterfaceForm(InterfaceForm):
    untagged_vlan = DynamicModelChoiceField(
        queryset=VLAN.objects.all(),
        required=False,
        label="Modified Untagged VLAN",
        display_field="computed_fields.description_with_vg",
        query_params={
            "available_on_device": "$device",
            "include": "computed_fields"
        },
    )

class InterfaceEditViewOverride(InterfaceEditView):
    model_form = MyInterfaceForm

override_views = {
    "dcim:interface_edit": InterfaceEditViewOverride.as_view(),
}
```

I would consider this issue "done" when the above code renders a list of choices displaying a value of `computed_fields.description_with_vg` returned by the API.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

No changes are required.

### External Dependencies

No external dependencies.

skrobul:
This commit updates the resolvePath() function to allow accessing deeply nested properties within an object using a string path. This enhancement enables plugin developers to access arbitrary fields in the API response for custom form rendering, addressing limitations in scenarios like selecting VLANs for interfaces.

# Closes #5243 
# What's Changed

This pull request proposes a small enhancement to the `resolvePath()` function to enable accessing arbitrary fields within the API response in the `processResult()` method. The current behavior limits the accessibility to root-level entries, such as `id`, `name`, and `display` attributes, but does not allow accessing nested fields like `vlan_group.name`.

**Issue:**
The current behavior of the UI presenting dynamically generated choices from the backend API, particularly in scenarios like selecting VLANs for interfaces, can be limiting. For instance, when the list includes VLANs with the same name but from different VLAN groups, the default rendering provides insufficient information to differentiate between them.
More details at  https://github.com/nautobot/nautobot/issues/5243

**Solution:** This enhancement facilitates plugin developers to override the default behavior by allowing them to access nested fields within the API response. With this enhancement, developers can utilize custom forms with and specify a custom `display_field`, such as `computed_fields.description_with_vg` or access `vlan_group.name`.

# Screenshots

##### view override in custom plugin
```python
# view_overrides/views.py
from nautobot.core.forms.fields import DynamicModelChoiceField
from nautobot.dcim.forms import InterfaceForm
from nautobot.dcim.views import InterfaceEditView
from nautobot.ipam.models import VLAN


class MyInterfaceForm(InterfaceForm):
    untagged_vlan = DynamicModelChoiceField(
        queryset=VLAN.objects.all(),
        required=False,
        label="Modified Untagged VLAN",
        display_field="computed_fields.description_with_vg",
        query_params={
            "available_on_device": "$device",
            "include": "computed_fields"
        },
    )

class InterfaceEditViewOverride(InterfaceEditView):
    model_form = MyInterfaceForm



override_views = {
    "dcim:interface_edit": InterfaceEditViewOverride.as_view(),
}
```
##### computed field template
computed field used in the example: `description_with_vg`:

```jinja2
{{ obj.name }} {{obj.vid}} VlanGroup: {{ obj.vlan_group }}
```

# TODO

- [x] Unit, Integration Tests - no tests for JS


glennmatthews (commented):
Thanks for the pull request! This looks like a nice step closer to addressing the general problem described in #2668.

changes/5243.added
```diff
@@ -0,0 +1 @@
+Added support for setting display_field on DynamicModelChoiceField to nested values in suggested choices list.
```
glennmatthews, line 1: We really need more visible documentation of the `display_field` property in general, but there's at least a little documentation in `nautobot/docs/development/jobs/index.md` at present that it would be good to update with this feature.

skrobul: Agreed - I initially looked for documentation, but dismissed that file since the feature did not seem relevant to jobs directly. I have now added few examples in the [commit](https://github.com/nautobot/nautobot/pull/5244/commits/e079989971b68b08131a8a5c4d603e336747103c) below.

nautobot/project-static/js/forms.js
```diff
@@ -160,6 +160,53 @@ function initializeColorPicker(context, dropdownParent=null){
...
+
+    if (typeof response !== 'object' || response === null || !response) {
+        console.error('Invalid response object');
+        return null;
+    }
+
+    return fieldPath
+           .replace(/\[|\]\.?/g, '.')
+           .split('.')
+           .filter(value => value)
```
glennmatthews, line 206: What cases is this `.filter()` guarding against? Is it needed?

glennmatthews (approved):

nautobot/docs/development/jobs/index.md
```diff
@@ -403,6 +403,31 @@ device_type = ObjectVar(
...
+    model=VLAN,
+    display_field="vlan_group.name",
+    query_params={
+        "depth": 1
+    },
+)
+```
+
+In the example above, [`"depth": 1`](../../user-guide/platform-functionality/rest-api/overview/#depth-query-parameter) was needed to influence REST API to include details of the associated records.
+Another example of using the nested reference would be to access [computed fields](../../user-guide/platform-functionality/computedfield/) of the model:
```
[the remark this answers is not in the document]
skrobul: thx

nautobot/project-static/js/forms.js
skrobul: It removes 'empty'  fields, for example, if you said `interface..name` it would still work rather than silently failing.

nautobot/project-static/js/forms.js
```diff
@@ -160,6 +160,53 @@ function initializeColorPicker(context, dropdownParent=null){
...
+ * }
+ * // returns "myvlan"
+ * resolvePath(response, "vlan.name")
+ *
+ * // returns "eth0"
+ * resolvePath(response, "interfaces[0].name")
+ *
+ * // returns "eth0"
+ * resolvePath(response, "interfaces.0.name")
+ */
```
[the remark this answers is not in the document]
skrobul: We don't have to - some people might find it much more intuitive to select elements of an array how they are used to at a relatively small cost. If that's a no-go, let me know and I'll remove it.

PR #5245, opened 2024-02-07
[ltm-1.6] Improve location detail view performance

Kircheneer:
# Closes #DNE
# What's Changed

This should significantly speed up

- the location detail view
- any views utilizing `dcim/inc/location_hierarchy.html`

by doing a couple of things:

- Using `without_tree_fields` on the queryset for the view to ensure that `get_object_or_404` in the `get` method of the view doesn't have to calculate the location tree (it doesn't use any of the fields the calculation adds
- Using `without_tree_fields` on the query for the children of the location for the same reasons
- Implement a `LocationQuerySet.ancestors_without_cte` that provides a more optimized implementation of `TreeQuerySet.ancestors` in cases where the tree is shallow but wide
- Use this method in a couple of places

# Screenshots

_With ~40k locations_

ltm-1.6 @ d024edb7060ad665baf62e7aec9dbf66ecb6bdd3:
Screenshot 2024-02-08 at 09 52 33

this branch:
Screenshot 2024-02-08 at 09 51 50

These numbers are pretty consistent when reloading the page. Note that these are number reproduced in a simple dev environment, the view in the actual deployment currently takes ~6-7s.

# TODO
- [ ] Unit, Integration Tests
- Further things to explore:
- [x] Pre-computing `ancestors` for `dcim/inc/location_hierarchy.html` because this still triggers a full tree query whenever used - we can optimize this if we assume our trees are shallow
- [ ] Applicability to 2.0
- [ ] Usage of `ancestors_without_cte` in other places

### DIFF
--- a/nautobot/core/tests/test_tree_queries.py
+++ b/nautobot/core/tests/test_tree_queries.py
@@ -1,0 +1,44 @@
+from nautobot.core.testing import TestCase
+from nautobot.dcim.models import Location
+
+
+class QuerySetAncestorTests(TestCase):
+    """Tests for custom `TreeQuerySet.ancestors` method."""
+
+    def test_empty_ancestors(self):
+        """Test that `TreeQuerySet.ancestors` works even when there are no ancestors."""
+        base_location_with_tree_fields = (
+            Location.objects.with_tree_fields().filter(location_type__name="Campus").first()
+        )
+        base_location_without_tree_fields = (
+            Location.objects.without_tree_fields().filter(location_type__name="Campus").first()
+        )
+
+        self.assertQuerysetEqual(
+            base_location_with_tree_fields.ancestors(),
+            base_location_without_tree_fields.ancestors(),
+            msg="`TreeQuerySet.ancestors()` output doesn't match between custom and original implementation for empty ancestors list",
+        )
+
+    def test_basic_path_comparison(self):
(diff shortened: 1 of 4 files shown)


nautobot/dcim/models/locations.py
```diff
@@ -125,6 +125,22 @@ def get_for_model(self, model):
         content_type = ContentType.objects.get_for_model(model._meta.concrete_model)
         return self.filter(location_type__content_types=content_type)
 
+    @staticmethod
+    def ancestors_without_cte(of, *, include_self=True):
```
Kircheneer, line 129: - Try to merge this with the normal ancestors method
- Put this into the base `TreeQuerySet`

Kircheneer (commented):
Discussion with @glennmatthews

Kircheneer:
Updated to match #5255

nautobot/core/tests/test_tree_queries.py
```diff
@@ -0,0 +1,44 @@
+from nautobot.core.testing import TestCase
```
glennmatthews, line 1:
```suggestion
from nautobot.utilities.testing import TestCase
```

timizuoebideri1 (approved):

HanlinMiao (approved):

PR #5246, opened 2024-02-07
rename SoftwareImage to SoftwareImageFile, add m2m from Device to SoftwareImageFile

### LINKED ISSUE #5101: DLM Migrations for Software models
As a app engineer, I want migrations built in DLM for the porting of the SoftwareLCM and SoftwareImageLCM models into core, so that I can ensure the data is migrated for users.

Create data/schema migrations in support of moving the models into core. Data lives in the models and in custom relationships which will become proper FKs in the new modes. Some features of DLM will need to be refactored to support this move.

Please document this process and lessons learned for future moves of models between apps and core.

gsnider2195:
# Closes #5101
# What's Changed

# Screenshots

## Device edit view

image

## Device detail view

image

# TODO

- [ ] Unit, Integration Tests


changes/1.added
```diff
@@ -1,5 +1,5 @@
-Added new models for software versions and software images.
-Added a many-to-many relationship from `DeviceType` to `SoftwareImage`.
+Added new models for software versions and software image files.
+Added a many-to-many relationship from `DeviceType` to `SoftwareImageFile`.
```
gsnider2195, line 2:
```suggestion
Added a many-to-many relationship from `DeviceType` to `SoftwareImageFile`.
Added a many-to-many relationship from `Device` to `SoftwareImageFile`.
```

nautobot/dcim/templates/dcim/device.html
```diff
@@ -222,6 +222,24 @@
                                             {{ object.software_version|hyperlinked_object }}
                                         </td>
                                     </tr>
+                                    <tr>
+                                        <td>Software Image files</td>
```
glennmatthews, line 226:
```suggestion
                                        <td>Software Image Files</td>
```

nautobot/dcim/templates/dcim/device.html
```diff
@@ -222,6 +222,24 @@
...
+                                                {% else %}
+                                                    {% with software_version_image_file_count=object.software_version.software_image_files.count %}
+                                                        {% if software_version_image_file_count %}
+                                                            <a href="{% url 'dcim:softwareimagefile_list' %}?software_version={{ object.software_version.pk }}">{{ software_version_image_file_count }}</a>
+                                                        {% else %}
+                                                            <span class="text-muted">&mdash;</span>
+                                                        {% endif %}
+                                                    {% endwith %}
+                                                {% endif %}
+                                            {% endwith %}
```
glennmatthews, line 240: We generally expect only a very small (single-digit?) number of image files per device/device-type/software-version, right? Wonder if it'd be better to just list them outright here instead of providing a count and a filtered list link?

nautobot/dcim/templates/dcim/device.html
```diff
@@ -222,6 +222,24 @@
...
                                         </td>
                                     </tr>
+                                    <tr>
+                                        <td>Software Image files</td>
+                                        <td>
+                                            {% with software_image_file_count=object.software_image_files.count %}
+                                                {% if software_image_file_count %}
+                                                <a href="{% url 'dcim:softwareimagefile_list' %}?devices={{ object.pk }}">{{ software_image_file_count }}</a>*
+                                                {% else %}
+                                                    {% with software_version_image_file_count=object.software_version.software_image_files.count %}
```
glennmatthews, line 232: Is this correct, or should we inherit from `object.device_type.software_image_files` instead?

gsnider2195: That gets into the inheritance conversation. I think right now device type <-> software image just determines what images are compatible with a device type. The device -> software version <-> software image is what determines what is (or should be) running on that device, unless you override it with device <-> software image. 😵

nautobot/dcim/templates/dcim/device.html
```diff
@@ -220,6 +220,25 @@
                                         <td>Software Version</td>
                                         <td>
                                             {{ object.software_version|hyperlinked_object }}
+                                                {% with overridden_software_image_files=object.software_image_files.all %}
```
glennmatthews, line 223: Indentation seems off?

nautobot/dcim/templates/dcim/device_edit.html
```diff
@@ -108,3 +121,89 @@
...
     </div>
 {% endblock %}
+
+{% block javascript %}
+<script type="text/javascript">
+
+    const software_version_image_container = document.querySelector('div#software_version_image_container');
+    const software_version_image_description = document.querySelector('span#software_version_image_description');
+    const software_version_image_list = document.querySelector('ul#software_version_image_list');
+    const software_image_url = '/api/dcim/software-image-files/';
```
glennmatthews, line 131:
```suggestion
    const software_image_url = '{% url dcim-api:softwareimagefile-list %}';
```

nautobot/dcim/templates/dcim/device_edit.html
```diff
@@ -24,6 +25,18 @@
         <div class="panel-body">
             {% render_field form.platform %}
             {% render_field form.software_version %}
+            <div class="form-group" id="software_version_image_container" style="display:none">
+                <div class="col-md-3"></div>
+                <div class="col-md-9">
+                    <span class="help-block">
+                        <span id="software_version_image_description"></span>
+                        <ul id="software_version_image_list">
+                            <li> Software images go here </li>
```
glennmatthews, line 34: Is there value in having this `<li>` here or should the list just be empty initially?

nautobot/project-static/css/base.css
```diff
@@ -324,6 +324,10 @@ ul.nav-tabs li a:hover,
 .label a {
     color: inherit;
 }
+ul.software-image-hierarchy {
+    list-style-type: "↳ ";
+    padding-left: 18px;
+}
```
glennmatthews, line 330: If this is specific to the Device detail view, it *might* make sense to do as an inline style there instead of adding to base.css?

nautobot/dcim/templates/dcim/device_edit.html
gsnider2195: I just changed this in my local branch. It now looks like this to help describe what it's there for:

```html
            {% render_field form.software_version %}
            <div class="form-group" id="software_version_image_container" style="display:none">
                <div class="col-md-3"></div>
                <div class="col-md-9">
                    <span class="help-block">
                        <span id="software_version_image_description"><img src="{% static 'img/ajax-loader.gif' %}"></span>
                        <ul id="software_version_image_list">
                            <li> Software images for the selected software version will appear here </li>
                        </ul>
                    </span>
                </div>
            </div>
            {% render_field form.software_image_files %}
```

gsnider2195: BTW, the `<li>` tag is never rendered because of the `style="display:none"` on the parent div and the javascript below.

nautobot/dcim/templates/dcim/softwareimagefile_retrieve.html
```diff
@@ -87,11 +87,25 @@
...
+                                            {% if device_type_count %}
+                                            <a href="{% url 'dcim:devicetype_list' %}?software_image_files={{ object.pk }}">{{ device_type_count }}</a>
+                                            {% else %}
+                                            <span class="text-muted">&mdash;</span>
+                                            {% endif %}
+                                        {% endwith %}
+                                    </td>
+                                </tr>
+                                <tr>
+                                    <td>Devices</td>
```
glennmatthews, line 100: Wonder if something like `Devices overridden to use this file` might be clearer?

glennmatthews (approved):
Let's :shipit: and iterate. :-)

PR #5255, opened 2024-02-08
Improves performance for location detail view.

Kircheneer:
# Closes #DNE
# What's Changed

Related: #5245

# Screenshots

# TODO

- [ ] Unit, Integration Tests


changes/5254.changed
```diff
@@ -0,0 +1,2 @@
+Changed `TreeQuerySet.ancestors` implementation a more efficient approach for shallow trees.
```
glennmatthews, line 1:
```suggestion
Changed `TreeQuerySet.ancestors` implementation to a more efficient approach for shallow trees.
```

nautobot/core/models/tree_queries.py
```diff
@@ -1,6 +1,7 @@
 from functools import cached_property
 
 from django.core.cache import cache
+from tree_queries.compiler import TreeQuery
```
glennmatthews, line 4:
```suggestion
```

nautobot/core/models/tree_queries.py
```diff
@@ -11,6 +12,27 @@ class TreeQuerySet(TreeQuerySet_, querysets.RestrictedQuerySet):
     """
     Combine django-tree-queries' TreeQuerySet with our RestrictedQuerySet for permissions enforcement.
     """
+    def ancestors(self, of, *, include_self=False):
```
glennmatthews, line 15: I wonder how difficult it would be to add a couple of test cases for the two main branches of this function? I.e. make sure that `location.objects.with_tree_fields.last().ancestors() == location.objects.without_tree_fields.last().ancestors()` at least?

Kircheneer: Yes! Note that the former returns a queryset and the latter returns a list. In practice I have not found that this is causing any problems, but I thought it would be good to highlight for you to think about.

Kircheneer: Theoretically, using another query, we could turn it back into a queryset. I'm not sure this is worth it though.

glennmatthews: Thanks for pointing that out. I am a bit concerned that we might have existing code here or in apps that's assuming that `.ancestors()` returns a queryset? So it might be worthwhile to construct a queryset here if it doesn't cause too much performance impact.

glennmatthews: Though as I'm thinking about it, that may be tricky to do in a way that still gives the right ordering...

Kircheneer: I have done this, it is a little complicated to look at but it works.

nautobot/core/tests/test_tree_queries.py
```diff
@@ -0,0 +1,43 @@
...
+            (base_location_without_tree_fields.ancestors()),
+            "`TreeQuerySet.ancestors()` output doesn't match between custom and original implementation",
+        )
+
+    def test_tree_annotations_not_present(self):
+        """Test that using the custom `TreeQuerySet.ancestors` implementation the tree annotations aren't present."""
+        base_location_without_tree_fields = (
+            Location.objects.without_tree_fields().filter(location_type__name="Aisle").first()
+        )
+        self.assertFalse(
```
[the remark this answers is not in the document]
Kircheneer: Thanks, good catch.

glennmatthews (approved):

PR #5278, opened 2024-02-13
feat: ✨ add user token as permission constraints 2.0

### LINKED ISSUE #5104: Evaluating permission constraints referencing the current user
### As ...

Patti - Platform Admin

### I want ...

a constrains mechanism to reference the present User in a abstract manner. For example, I would like to have functionality in order to create a constrain that allows a User to see only his own Changelog and journal entries, barring him from viewing everyone else's. 

### So that ...

we can enforce a granular Data protection policy without having to create a Group for every User. 

- a User can see only his own Changelog and Journal entries
- Administrator can continue to view all User's entries

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation

### Database Changes

- I think not but not sure

### External Dependencies

_No response_

gertzakis:
# Closes #5104
# What's Changed

Added functionality to reference the current User in permission's constraints. That way an administrator can filter models per the active user (eg Change log)
# Screenshots

# TODO

- [x] Unit, Integration Tests


nautobot/core/tests/test_authentication.py
```diff
@@ -453,3 +454,64 @@ def test_delete_object(self):
         url = reverse("ipam-api:prefix-detail", kwargs={"pk": self.prefixes[0].pk})
         response = self.client.delete(url, format="json", **self.header)
         self.assertEqual(response.status_code, 204)
+
+    @override_settings(EXEMPT_VIEW_PERMISSIONS=[])
+    def test_user_token_constraints(self):
+        """
+        Test enabling remote authentication with the default configuration.
```
glennmatthews, line 461: This doesn't seem accurate to me?

gertzakis: You are right, changed it

nautobot/core/tests/test_authentication.py
```diff
@@ -453,3 +454,64 @@ def test_delete_object(self):
...
+        response_user1 = self.client.get(url, **self.header)
+        response_user2 = self.client.get(url, **header_user2)
+
+        # Assert every user has permissions to view Change Logs
+        self.assertTrue(self.user.has_perms(["extras.view_objectchange", "extras.list_objectchange"]))
+        self.assertTrue(obj_user2.has_perms(["extras.view_objectchange", "extras.list_objectchange"]))
+
+        # Check against 1st user's response
+        self.assertEqual(response_user1.status_code, 200)
+        self.assertNotEqual(response_user1.data["count"], ObjectChange.objects.count())
```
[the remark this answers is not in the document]
gertzakis: Sure, changed it

nautobot/core/tests/test_authentication.py
```diff
@@ -453,3 +454,64 @@ def test_delete_object(self):
...
+
+        # Check against 2nd user's response
+        self.assertEqual(response_user2.status_code, 200)
+        self.assertEqual(response_user2.data["count"], 1)
```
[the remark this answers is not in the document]
gertzakis: Sure, added also these assertions

nautobot/users/admin.py
```diff
@@ -241,7 +241,8 @@ def clean(self):
             for ct in object_types:
                 model = ct.model_class()
                 try:
-                    model.objects.filter(*[models.Q(**c) for c in constraints]).exists()
+                    tokens = {"$user": 0}  # setting token to null user ID
```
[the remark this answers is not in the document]
gertzakis: Nice catch, changed it

nautobot/users/admin.py
```diff
@@ -241,7 +241,8 @@ def clean(self):
             for ct in object_types:
                 model = ct.model_class()
                 try:
-                    model.objects.filter(*[models.Q(**c) for c in constraints]).exists()
+                    tokens = {"$user": None}  # setting token to null user ID
+                    model.objects.filter(qs_filter_from_constraints(constraints, tokens)).exists()
```
glennmatthews, line 245: Just confirming - have you tested this out in the admin UI to make sure it works as intended?

nautobot/core/models/querysets.py
```diff
@@ -125,16 +125,11 @@ def restrict(self, user, action="view"):
         # Filter the queryset to include only objects with allowed attributes
         else:
             attrs = Q()
-            for perm_attrs in user._object_perm_cache[permission_required]:
-                if isinstance(perm_attrs, list):
-                    for p in perm_attrs:
-                        attrs |= Q(**p)
```
glennmatthews, line 131: Does this case need to be handled specially by `qs_filter_from_constraints`? This would I believe be for a case where we have an objectpermission with constraints like `[{"user": "$user"}, {"job__name": "MyJob"}]`, i.e OR conditions in the constraints.

Reading the code carefully, this case **might** be dead code, as `user._object_perm_cache` is a list constructed by calling `ObjectPermission.list_constraints()` and appending the result, so it **might** be already exploding such constraints out into multiple separate entries in `_object_perm_cache` - but please test and confirm.

nautobot/users/admin.py
gertzakis: Hey yes! I didn't notice any weird issues

nautobot/core/models/querysets.py
gertzakis: Regarding this, if we have a constraint like `{"user":"$user"}` and not use the `qs_filter_from_constraints` I get a `ValidationError at /extras/object-changes/` with messsage `['“$user” is not a valid UUID.']`. 
If I don't have a constraint like the above, the original code was working as expected.

glennmatthews: I think maybe we're talking past one another :smile:. I definitely get that `qs_filter_from_constraints` is necessary to handle any `$user` tokens in the constraints. My concern was whether the way it's being used here also correctly handles the "list of constraints in a single ObjectPermission" case that I mentioned in my comment above. As I said, I **think** it should work, but wanted to make sure we test that case as well.

gertzakis: Alright, tested it with a list of constraints like `[{"user": "$user"}, {"action": "delete"}]` and I also added one test case for it. 
I noticed that with the the example I just mentioned, the user sees his all of his own logs, and logs of other users only if the action is `delete`.

glennmatthews: Sounds correct to me. Thanks!

PR #5281, opened 2024-02-13
Change Prefix.location to a many-to-many field

### LINKED ISSUE #4334: Allow `Prefixes` to belong to many `Locations`
### As ...

Nelly - Network Engineer

### I want ...

to have a prefix shared across many different locations.

### So that ...

I could model a prefix shared across many different locations (e.g., building), a common use case for Campus Networks where a segment is spread over building on the same Campus.

I can relate a Prefix to many Locations without the current single association limit.

- [ ] Automation
- [ ] Circuits
- [X] DCIM
- [X] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

- Change type of relationship

### External Dependencies

_No response_

### LINKED ISSUE #5283: IP Address list view not correctly sorted by default
### Environment
* Nautobot version (Docker tag too if applicable): 2.1.4, appears broken as far back as 2.0.0

The `/ipam/ip-addresses/` list view lists IPs in a seemingly random order:

image

### Steps to Reproduce
1. https://demo.nautobot.com/ipam/ip-addresses/

### Expected Behavior

IPs to be ordered logically by default

### Observed Behavior

Seemingly random ordering.

Looks to have been broken by #4230; I confirmed locally that reverting 536d75dd802044d5857578748fde58173ca33a26 appears to restore the expected ordering to the table.

glennmatthews:
# Closes #4334 
# What's Changed

Similar to #5052, but for Prefix - replace the `Prefix.location` ForeignKey with a `Prefix.locations` ManyToManyField.

A couple of fixes to VLANs after #5052 that I identified in testing the Prefix behavior
- added missing `locations` field to `VLANFilterForm`
- changed use of `Count` to `count_related()` on VLAN views so as to preserve expected ordering in list views
  - I also corrected this on IPAddress views, which fixes #5283 - will backport this portion of the changes to `develop` as a bug fix.

# Screenshots

image

image

image

# TODO

- [x] Unit, Integration Tests


nautobot/ipam/views.py
```diff
@@ -424,7 +424,7 @@ class PrefixListView(generic.ObjectListView):
     filterset_form = forms.PrefixFilterForm
     table = tables.PrefixDetailTable
     template_name = "ipam/prefix_list.html"
-    queryset = Prefix.objects.all()
+    queryset = Prefix.objects.annotate(location_count=Count("locations"))
```
glennmatthews, line 427: We should be using `count_related` instead here and elsewhere in order to properly preserve queryset order.

nautobot/ipam/api/serializers.py
```diff
@@ -184,6 +184,14 @@ class Meta:
 class PrefixSerializer(NautobotModelSerializer, TaggedModelSerializerMixin):
     prefix = IPFieldSerializer()
     type = ChoiceField(choices=PrefixTypeChoices, default=PrefixTypeChoices.TYPE_NETWORK)
+    # for backward compatibility with 2.0-2.1 where a Prefix had only a single Location
+    location = NautobotHyperlinkedRelatedField(
```
timizuoebideri1, line 188: Why should users be able to utilise a `pre-2.2` feature in `2.2`? I had assumed that these features were reserved for legacy versions.

glennmatthews: Fair question. #5052 did the same thing with a `location` write-only field on the 2.2 `VLANSerializer` - I think the primary motivation was to allow CSV import to still specify a location if desired. (https://github.com/nautobot/nautobot/pull/5052#discussion_r1459014038)

nautobot/ipam/signals.py
```diff
@@ -91,6 +91,20 @@ def ip_address_to_interface_assignment_created(sender, instance, raw=False, **kw
...
+    if action == "pre_add":
+        instance_ct = ContentType.objects.get_for_model(instance)
+        invalid_locations = model.objects.select_related("location_type").filter(
+            Q(pk__in=pk_set), ~Q(location_type__content_types__in=[instance_ct])
+        )
+        if invalid_locations.exists():
+            invalid_location_types = {location.location_type.name for location in invalid_locations}
+            raise ValidationError(
+                {"locations": f"Prefixes may not associate to Locations of types {list(invalid_location_types)}."}
+            )
```
timizuoebideri1, line 105: Since most of the block of code is duplicate of `nautobot.ipam.signals.assert_vlan_locations_content_types`, would it be better to just move it into a standalone function?

timizuoebideri1 (approved):

nautobot/ipam/signals.py
glennmatthews: I'll see what I can do.

glennmatthews: Combined the two signals into one. Also found that we were missing handling for the `reverse` case of the signal, i.e. `Location.prefixes.add(prefix)` or `Location.vlans.add(vlan)` so I've added that and test coverage for it.

PR #5293, opened 2024-02-15
WIP: implements bulk_create with change logging

Kircheneer:
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/core/models/managers.py
+++ b/nautobot/core/models/managers.py
@@ -1,5 +1,8 @@
+import contextvars
 import logging
+from uuid import UUID
 
+from django.db import transaction
 from django.db.models import Manager
 from taggit.managers import _TaggableManager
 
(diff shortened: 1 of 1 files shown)


nautobot/core/models/managers.py
```diff
@@ -13,6 +16,33 @@ class BaseManager(Manager):
...
+                objs=objs,
+                batch_size=batch_size,
+                ignore_conflicts=ignore_conflicts,
+            )
+            object_changes = []
+            for obj in created_objects:
+                if not isinstance(obj, ChangeLoggedModel):
+                    continue
+                object_change = obj.to_objectchange(action=ObjectChangeActionChoices.ACTION_CREATE)
+                object_change.request_id = obj.pk  # Hack, should probably be a common id for all the objects
```
glennmatthews, line 41: Yeah, at minimum this should be a common id within this function call; consider maybe also letting the caller optionally pass a `request_id` and/or `user` as additional kwargs to the function?

Kircheneer: I like all of those ideas!

Kircheneer:
I have opted to go for a separate name for the method. I think this is less intrusive. Open points:
- [ ] Should this be used anywhere here https://github.com/nautobot/nautobot/blob/93b2b0475409d7a3b85d4bd98e7b280411432cb6/nautobot/dcim/models/devices.py#L745? It's the only place I was able to find in nautobot that uses `bulk_create` directly
- [ ] Should there be docs for discoverability?

glennmatthews:
In the new implementation, how is the performance at scale? Have you compared it to a naive for-loop (not using bulk-create) with change-logging enabled? Wondering in particular for the SSOT case if in 2.0 (with the removal of mandatory transaction.atomic wrappers) if the for-loop case is performant enough.

Kircheneer:
> In the new implementation, how is the performance at scale? Have you compared it to a naive for-loop (not using bulk-create) with change-logging enabled? Wondering in particular for the SSOT case if in 2.0 (with the removal of mandatory transaction.atomic wrappers) if the for-loop case is performant enough.

How would I go about this? I see that change logging is normally done through `post_save` but it appears this doesn't fire when I use the ORM directly, neither with `Model.objects.create`, nor with calling `save` or even `validated_save` explicitly.

glennmatthews:
https://docs.nautobot.com/projects/core/en/stable/user-guide/administration/tools/nautobot-shell/?h=web_request_context#change-logging-and-webhooks

glennmatthews:
Hm. I wonder if adding caching to `ChangeLoggedModelsQuery.list_subclasses()` and `FeaturedQueryMixin.as_queryset()` would yield significant performance improvement here.

glennmatthews:
Opened #5329. :-)

Kircheneer:
Going to re-visit once that's in. Thanks!

PR #5322, opened 2024-02-21
Fix #5298 and an unrelated bug

### LINKED ISSUE #5298: Logout throws Validation Error when using `next` branch
### Environment
* Nautobot version (Docker tag too if applicable): v2.2.0a1
* Python version: 3.8
* Middleware(s):

This is applicable when using the `next` branch with version `v2.2.0a1`.
### Steps to Reproduce
1. Login with a User
2. Logout
3.

### Expected Behavior
Logout successfully 

### Observed Behavior
A `ValidationError` is thrown at `/logout/` with message `['“4ci8je1rmskkfei3g8avd9mt2da2bft5” is not a valid UUID.']` when I use the `next` branch.

glennmatthews:
# Closes #5298
# What's Changed

- Fix #5298 by adding an `isinstance(instance, BaseModel)` check so that we don't attempt to delete ContactAssociations belonging to a model that doesn't support them, e.g. a `session` DB record being deleted at logout time.
- Fix an unrelated issue I noticed while troubleshooting some test failures earlier in #5310 development - in some cases (notably the custom-field-related background tasks) we're getting a JobResult created that ends up going to a terminal state without actually ever setting its `date_done` value. There was some mismatched logic in the JobResult detail view and associated JS such that it would continually refresh the page if the JobResult had a terminal status but no date_done. While we probably should chase down and fix the underlying code that's resulting in creation of such a JobResult, making this change at least stops the page from infinitely self-refreshing.

# Screenshots

# TODO

- n/a Unit, Integration Tests


nautobot/extras/templates/extras/inc/jobresult_js.html
```diff
@@ -2,7 +2,7 @@
 {% load static %}
 
 <script type="text/javascript">
-{% if not result.date_done %}
+{% if result.status != 'FAILURE' and result.status != 'REVOKED' and result.status != 'SUCCESS' %}
```
itdependsnetworks, line 5: ??
```suggestion
{% if result.status not in ['FAILURE', 'REVOKED', 'SUCCESS' %}
```

gsnider2195: You could also use [JobResultStatusChoices.READY_STATES](https://github.com/nautobot/nautobot/blob/8738d0b2d666202527728d10267059ce17c6fd32/nautobot/extras/choices.py#L203)

glennmatthews: How would I reference that choices value from a template? Maybe just my brain not working at the moment. :-)

glennmatthews: image

gsnider2195: Django templates don't support that syntax. You could keep your current syntax or pass a template variable for `JobResultStatusChoices.READY_STATES` from the view.

HanlinMiao (approved):
LGTM once we address the conditionals in the template.

glennmatthews:
> LGTM once we address the conditionals in the template.

I'm of the opinion that the way it's done currently is fine. :-)

gsnider2195 (approved):

Issue #2906, opened 2022-11-23
Standardize `max_length` globally across all models


jathanism:
### Proposed Changes

The `max_length` field varies widley. This `grep` catches most of it (with uniqueness counts):

```
$ grep -r " name = .*\models.CharField" nautobot | grep -o "max_length.*\d+" | sort -n | uniq -c
  37 max_length=100
   1 max_length=150
   1 max_length=20
   1 max_length=200
   2 max_length=255
   5 max_length=50
   7 max_length=64
```

### Justification

This inconsistency causes recurring confusion in UX at a minimum, not to mention complicates testing and validation often enough to be a pain.

NB: This may be solved by the introduction of `BasePropertiesModel` or similar in #2822?

# TODO

- [x] #2603 
- [x] #3171

glennmatthews:
Should revisit this aspect of our data models more generally - does `description` need to be limited to 200 characters or should it also increase, etc.

bryanculver:
Should consider `CustomFieldChoice`'s `value` as well.

lampwins:
Will want to doublecheck forms and serializer fields

glennmatthews:
Implemented by #5355.

PR #5325, opened 2024-02-22
Initial work on Controller Models

### LINKED ISSUE #3111: Controllers
### As ...

Nelly - Network Engineer

### I want ...

A way to model and track network controllers

### So that ...

I can track the relationships between controllers, and the devices that they control

- I have a model to define a controller
- Controllers have a Controller-specific role.
- Devices can be optionally related to one or more controllers with a defined priority

- Automation
- DCIM

### Database Changes

A new controller model and relationships with devices

### External Dependencies

None

whitej6:
# Closes #3111 

# What's Changed

- Implemented Controller models.
- Fixed testing RelatedManager fields.
- Fixed bulk forms testing.

# TODO

- [x]  Create Controller model
    - [x] Migrations
    - [x] Default statuses (data migration)
    - [x] Serializer
    - [x] List view / Table
    - [x] Detail view
    - [x] Create/edit form
    - [x] Create/edit view
    - [x] Bulk edit form
    - [x] Bulk edit view
    - [x] Filterset
    - [x] Filter form
    - [x] Factory
    - [x] Unit tests
    - [x] Documentation
    - [x] Update nautobot.apps as needed
    - [x] Validators
    - [x] View all controlled devices in the `Controller` detail view.
- [x]  Create ControllerDeviceGroup model
    - [x] Migrations
    - [x] Serializer
    - [x] List view / Table
        - [x] Replace List View with Tree View
    - [x] Detail view
    - [x] Create/edit form
    - [x] Create/edit view
    - [x] Bulk edit form
    - [x] Bulk edit view
    - [x] Filterset
    - [x] Filter form
    - [x] Factory
    - [x] Unit tests
    - [x] Documentation
    - [x] Update nautobot.apps as needed
    - [x] Create a signal changing the controller for all descendants, when the controller is changed.
    - [x] Validators
    - [x] Allow to select the set of devices that belong to the group when creating/editing a ControllerDeviceGroup.
- [x] Add FK from Device to ControllerDeviceGroup
    - [x] Migration
    - [x] Model validation
    - [x] Update querysets (select_related)
    - [x] Serializer
    - [x] Create/edit form
    - [x] Create/edit view
    - [x] Bulk edit form
    - [x] Bulk edit view
    - [x] Device detail view
    - [x] ControllerDeviceGroup detail view
    - [x] Filterset
    - [x] Filter form
    - [x] Factory
    - [x] Unit tests
    - [x] Documentation
- [ ] Release overview documentation


nautobot/dcim/models/devices.py
```diff
@@ -1209,3 +1210,105 @@ def __str__(self):
...
+    "export_templates",
+    "graphql",
+    "locations",
+    "statuses",
+    "webhooks",
+)
+class Controller(PrimaryModel, ConfigContextModel):
+    """Controller model."""  # TODO: (whitej6) finialize description
+
+    name = models.CharField(max_length=100, unique=True)
```
glennmatthews, line 1233: We're trying to use `max_length=255` for new model charfields to save some future work in #2906

nautobot/dcim/models/devices.py
```diff
@@ -1209,3 +1210,105 @@ def __str__(self):
...
+        blank=True,
+        null=True,
+    )
+    external_integration = models.ForeignKey(
+        to="extras.ExternalIntegration",
+        on_delete=models.PROTECT,
+        related_name="controllers",
+        blank=True,
+        null=True,
+    )
```
glennmatthews, line 1262: What's the thinking behind PROTECT vs SET_NULL on these various fields? I'd naively expect PROTECT on all of them but I see you have SET_NULL on `platform`?

nautobot/dcim/models/devices.py
```diff
@@ -1209,3 +1210,105 @@ def __str__(self):
...
+    deployed_controller_type = models.ForeignKey(
+        to=ContentType,
+        on_delete=models.SET_NULL,
+        related_name="+",
+        blank=True,
+        null=True,
+    )
+    deployed_controller_id = models.UUIDField(blank=True, null=True)
+    deployed_controller = GenericForeignKey(ct_field="deployed_controller_type", fk_field="deployed_controller_id")
```
glennmatthews, line 1271: What kinds of objects does this GFK encompass?

nautobot/dcim/models/devices.py
```diff
@@ -1209,3 +1210,105 @@ def __str__(self):
...
+    "export_templates",
+    "graphql",
+    "locations",
+    "statuses",
+    "webhooks",
+)
+class DeviceGroup(TreeModel, PrimaryModel, ConfigContextModel):
+    """Controller model."""  # TODO: (whitej6) finialize description
+
+    name = models.CharField(max_length=48)
```
glennmatthews, line 1290: See earlier comment, but 48 definitely feels way too short in any case.

nautobot/dcim/models/devices.py
```diff
@@ -1209,3 +1210,105 @@ def __str__(self):
...
+    devices = models.ManyToManyField(
+        to="dcim.Device",
+        related_name="device_groups",
+    )
+
+
+class ControllerDeviceGroupAssociation(BaseModel):
+    """Controller model."""  # TODO: (whitej6) finialize description
+
+    weight = models.PositiveIntegerField(default=1000)
```
glennmatthews, line 1300: Is the idea that we'd have a `Meta.ordering` based on the weight? And just to be clear, is this for weighting devices among a controller, or controllers among a device, or both?

nautobot/dcim/models/devices.py
```diff
@@ -1209,3 +1210,105 @@ def __str__(self):
...
+    controller = models.ForeignKey(
+        to="dcim.Controller",
+        on_delete=models.CASCADE,
+        related_name="+",
+        blank=False,
+        null=False,
+    )
+    device = models.ForeignKey(
```
glennmatthews, line 1308:
```suggestion
    device_group = models.ForeignKey(
```

nautobot/dcim/models/devices.py
whitej6: Ack

nautobot/dcim/models/devices.py
whitej6: Thought process was if you delete the EI then workflows would fail. BUT thinking though it IF the field is nullable I do think `SET_NULL` is the right pattern

nautobot/dcim/models/devices.py
whitej6: Only `Device` and `DeviceRedundancyGroup`

nautobot/dcim/models/devices.py
whitej6: Ack

nautobot/dcim/models/devices.py
whitej6: Yes. Thought is predictive ordering from `Controller` to `DeviceGroup` but technically would handle the same in reverse

nautobot/dcim/models/devices.py
whitej6: I would also think it would be prudent to add `GenericRelation` to those models if we agree on this model/pattern.

nautobot/dcim/models/devices.py
whitej6: If you have 5 top level groups assigned to a controller and you want predictable ordering of the groups returned in the query. E.g. would enable the following example to be a predictable order when you do controller.device_groups

- ControllerA 
    - Group XYZ
    - group 123
    - Another group

nautobot/dcim/models/devices.py
```diff
@@ -1209,3 +1210,105 @@ def __str__(self):
...
+    )
+    role = RoleField(blank=False, null=False)
+    tenant = models.ForeignKey(
+        to="tenancy.Tenant",
+        on_delete=models.PROTECT,
+        related_name="controllers",
+        blank=True,
+        null=True,
+    )
+    external_integration = models.ForeignKey(
```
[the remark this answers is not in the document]
whitej6: Attrs should match the FK

nautobot/dcim/models/devices.py
```diff
@@ -1209,3 +1210,105 @@ def __str__(self):
...
+    "custom_links",
+    "custom_validators",
+    "dynamic_groups",
+    "export_templates",
+    "graphql",
+    "locations",
+    "statuses",
+    "webhooks",
+)
+class DeviceGroup(TreeModel, PrimaryModel, ConfigContextModel):
```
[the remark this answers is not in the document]
whitej6:
```yaml
controllerA:
  device_groups:
    - TopLevelA:
        children:
          - childA
    - TopLevelB:
      children:
        - childB:
            children:
              - childC
```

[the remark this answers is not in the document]
whitej6: Out of scope, that will all be done via a custom through model for the association in the wireless models. This is just the Controller portion.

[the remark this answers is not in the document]
jeffkala: When I see DeviceGroup my mind immediately assumes there will be some kind of attribute assignment to define an ordering/priority/weight of the group(device) members. Thinking of a similar concept to what device redundancy groups offer with the priority, but trying to solve for infrastructure deployments more so then "system redundancy".

Say you have network infrastructure but you want to use DeviceGroups to model failure domains of sorts. 
```yaml
device_groups: "bgp"
  children:
    - device: "rtr1"
       priority: 1
    - device: "rtr2"
       priority: 2
device_groups: "ospf"
  children:
    - device: "rtr2"
       priority: 1
    - device: "rtr1"
       priority: 2
```
This example isn't perfect, but I could see scenarios where a device belongs to multiple groups (device redundancy groups today doesn't support this), where a priority within each group could be vaulable.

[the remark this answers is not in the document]
whitej6: Core has vetoed multi-purpose device groups. That use case can be covered by dynamic groups or to the soon to be released static groups

nautobot/dcim/models/devices.py
```diff
@@ -1209,3 +1217,91 @@ def __str__(self):
...
+    "export_templates",
+    "graphql",
+    "locations",
+    "statuses",
+    "webhooks",
+)
+class ControllerDeviceGroup(TreeModel, PrimaryModel, ConfigContextModel):
+    """Controller model."""  # TODO: (whitej6) finialize description
+
+    name = models.CharField(max_length=255, unique=True)
```
snaselj, line 1299: Wouldn't be better to specify uniqueness based on `["parent", "name"]` fields rather than on `name` only in TreeModel?

nautobot/dcim/models/devices.py
```diff
@@ -1209,3 +1217,91 @@ def __str__(self):
...
+    status = StatusField(blank=False, null=False)
+    description = models.CharField(max_length=200, blank=True)
+    location = models.ForeignKey(
+        to="dcim.Location",
+        on_delete=models.PROTECT,
+        related_name="controllers",
+    )
+    platform = models.ForeignKey(
```
snaselj, line 1248: Should `Controller` be linked to `SoftwareVersion` same as `Device` is?

nautobot/dcim/models/devices.py
glennmatthews: Using `parent, name` adds a lot of complexity to things like natural/composite keys and it's something I regret us doing on Locations. Unless there's a strong use case for non-globally-unique names known already, it's always going to be easier to start with more restrictive uniqueness criteria and loosen them later if necessary, rather than vice versa.

nautobot/dcim/models/devices.py
glennmatthews: I think that's implicit through `Controller.deployed_controller_device.software_version`?

snaselj: Isn't `Controller->Platform` field implicit as well?

nautobot/dcim/models/devices.py
```diff
@@ -1209,3 +1217,91 @@ def __str__(self):
...
+    "custom_links",
+    "custom_validators",
+    "dynamic_groups",
+    "export_templates",
+    "graphql",
+    "locations",
+    "statuses",
+    "webhooks",
+)
+class ControllerDeviceGroup(TreeModel, PrimaryModel, ConfigContextModel):
```
snaselj, line 1296: Does it makes sense to rename model to`ControlledDeviceGroup`?

nautobot/dcim/models/devices.py
```diff
@@ -1209,3 +1217,91 @@ def __str__(self):
...
+    """Controller model."""  # TODO: (whitej6) finialize description
+
+    name = models.CharField(max_length=255, unique=True)
+    weight = models.PositiveIntegerField(default=1000)
+    controller = models.ForeignKey(
+        to="dcim.Controller",
+        on_delete=models.CASCADE,
+        related_name="+",
```
glennmatthews, line 1304: We shouldn't be using `related_name="+"` unless there's a *strong* justification for it.

nautobot/dcim/models/devices.py
```diff
@@ -1209,3 +1217,91 @@ def __str__(self):
...
+    "locations",
+    "statuses",
+    "webhooks",
+)
+class Controller(PrimaryModel, ConfigContextModel):
+    """Controller model."""  # TODO: (whitej6) finialize description
+
+    name = models.CharField(max_length=255, unique=True)
+    status = StatusField(blank=False, null=False)
+    description = models.CharField(max_length=200, blank=True)
```
glennmatthews, line 1242:
```suggestion
    description = models.CharField(max_length=CHARFIELD_MAX_LENGTH, blank=True)
```

nautobot/dcim/models/devices.py
```diff
@@ -1209,3 +1217,91 @@ def __str__(self):
...
+    "export_templates",
+    "graphql",
+    "locations",
+    "statuses",
+    "webhooks",
+)
+class Controller(PrimaryModel, ConfigContextModel):
+    """Controller model."""  # TODO: (whitej6) finialize description
+
+    name = models.CharField(max_length=255, unique=True)
```
glennmatthews, line 1240: Will need to update branch to pull in latest from `next` including #5355, then can update this:

```suggestion
    name = models.CharField(max_length=CHARFIELD_MAX_LENGTH, unique=True)
```

nautobot/dcim/models/devices.py
glennmatthews, line 1299:
```suggestion
    name = models.CharField(max_length=CHARFIELD_MAX_LENGTH, unique=True)
```

nautobot/dcim/models/devices.py
snaselj: Fixed

nautobot/dcim/navigation.py
```diff
@@ -298,6 +298,44 @@
                     ),
                 ),
             ),
+            NavMenuGroup(
+                name="Controllers",
+                weight=250,
```
HanlinMiao, line 303: Should this be weight=300?

glennmatthews: That would require re-weighting of the other groups in this tab - probably worthwhile though.

snaselj:
What should be the default statuses and roles for `Controller` model pls.?

@glennmatthews @whitej6

whitej6:
> What should be the default statuses and roles for `Controller` model pls.?
> 
> 
> 
> @glennmatthews @whitej6 

Statuses should match device defaults. 

Roles we do not provide default values outside of a few ipam models iirc.

snaselj:
What kind of validators / signals should be implemented on Controller models pls?

@glennmatthews @whitej6

glennmatthews:
> What kind of validators / signals should be implemented on Controller models pls?
> 
> @glennmatthews @whitej6

- `Controller.deployed_controller_device` and `Controller.deployed_controller_group` should be mutually exclusive, I believe
- Child `ControllerDeviceGroup` should probably have the same `controller` as their `parent` (if any)? Though might need to think about any edit/update case here, do we need to support moving a ControllerDeviceGroup and its descendants from one Controller to another?

whitej6:
Yes mutually exclusive

Yes a child must have the same as parent (validation), on form we should look at a custom JS script to force this (set field and disable the field)

nautobot/dcim/navigation.py
snaselj: Re-weighted

nautobot/dcim/factory.py
```diff
@@ -669,3 +671,35 @@ class Params:
...
+    name = factory.Faker("word")
+    description = factory.Faker("sentence")
+    status = random_instance(lambda: Status.objects.get_for_model(Controller), allow_null=False)
+    role = random_instance(lambda: Role.objects.get_for_model(Controller), allow_null=False)
+    platform = random_instance(Platform, allow_null=False)
+    location = random_instance(Location, allow_null=False)
+    tenant = random_instance(Tenant, allow_null=True)
+    external_integration = random_instance(ExternalIntegration, allow_null=True)
+    deployed_controller_device = factory.Maybe("has_device", random_instance(Device), None)
+    # deployed_controller_group = factory.Maybe("has_device", random_instance(DeviceRedundancyGroup), None)
```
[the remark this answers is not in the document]
snaselj: Fixed

nautobot/dcim/filters/__init__.py
```diff
@@ -1826,3 +1836,98 @@ class DeviceTypeToSoftwareImageFileFilterSet(BaseFilterSet):
...
+
+    def generate_query__subtree(self, value):
+        """Helper method used by DynamicGroups and by _subtree() method."""
+        if value:
+            max_depth = (
+                ControllerDeviceGroup.objects.with_tree_fields()
+                .extra(order_by=["-__tree.tree_depth"])
+                .first()
+                .tree_depth
+            )
```
glennmatthews, line 1920: Should be able to just use `max_depth = ControllerDeviceGroup.objects.max_depth` property here.

nautobot/dcim/fixtures/controller-data.json
glennmatthews, line 1: Is this file still used for anything, or should it be removed now that we have factories?

nautobot/dcim/tables/devices.py
```diff
@@ -1122,3 +1129,87 @@ class Meta(BaseTable.Meta):
...
+            "status",
+            "location",
+            "platform",
+            "role",
+            "tenant",
+            "external_integration",
+            "deployed_controller_device",
+            "deployed_controller_group",
+            "actions",
+        )
```
glennmatthews, line 1180: That's a lot of default columns - how's the usability here? Should we hide any of these columns by default for a better UX?

nautobot/dcim/views.py
```diff
@@ -2996,3 +3005,69 @@ class SoftwareVersionUIViewSet(NautobotUIViewSet):
...
+    serializer_class = serializers.ControllerSerializer
+    table_class = tables.ControllerTable
+
+    def get_extra_context(self, request, instance):
+        context = super().get_extra_context(request, instance)
+
+        if self.action == "retrieve" and instance:
+            devices = Device.objects.restrict(request.user).filter(controller_device_group__controller=instance)
+            devices_table = tables.DeviceTable(devices)
+            devices_table.columns.show("device_redundancy_group_priority")
```
glennmatthews, line 3030: Why? Same question on the ControllerDeviceGroupUIViewSet as well.

nautobot/dcim/tests/test_api.py
```diff
@@ -2769,3 +2771,72 @@ def setUpTestData(cls):
...
+
+class ControllerTestCase(APIViewTestCases.APIViewTestCase):
+    model = Controller
+
+    @classmethod
+    def setUpTestData(cls):
+        statuses = Status.objects.get_for_model(Controller)
+        roles = Role.objects.get_for_model(Controller)
+        platforms = Platform.objects.all()
+        locations = Location.objects.all()
```
glennmatthews, line 2784: Should be `Location.objects.get_for_model(Controller).all()`

Actually, are we missing validation on `Controller` model to make sure it's only allocated to a LocationType that permits Controller as a content-type?

Additionally, probably need to add Controller content-type to `LocationTypeFactory.content_types`...

nautobot/dcim/tests/test_filters.py
```diff
@@ -3329,3 +3343,107 @@ class DeviceTypeToSoftwareImageFileFilterSetTestCase(FilterTestCases.FilterTestC
...
+        ("deployed_controller_device", "deployed_controller_device__name"),
+        ("deployed_controller_group", "deployed_controller_group__id"),
+        ("deployed_controller_group", "deployed_controller_group__name"),
+    )
+
+    @classmethod
+    def setUpTestData(cls):
+        common_test_data(cls)
+
+        cls.locations = Location.objects.filter(location_type=LocationType.objects.get(name="Campus"))[:3]
```
glennmatthews, line 3368: Perhaps?

```suggestion
        cls.locations = Location.objects.filter(location_type__name="Campus")[:3]
```

nautobot/dcim/tests/test_models.py
```diff
@@ -1882,3 +1884,48 @@ def test_queryset_get_for_object(self):
...
+        )
+        with self.assertRaises(ValidationError) as error:
+            controller.validated_save()
+        self.assertEqual(
+            error.exception.message_dict["deployed_controller_device"][0],
+            "Cannot assign both a device and a device redundancy group to a controller.",
+        )
+
+
+class ControllerDeviceGroupTestCase(ModelTestCases.BaseModelTestCase):
```
glennmatthews, line 1910: Would be good to have a test case here for the custom signal - i.e. changing the `controller` of a parent group and verifying that it cascades to the child groups.

nautobot/dcim/models/devices.py
```diff
@@ -1220,3 +1230,144 @@ def __str__(self):
...
+    tenant = models.ForeignKey(
+        to="tenancy.Tenant",
+        on_delete=models.PROTECT,
+        related_name="controllers",
+        blank=True,
+        null=True,
+    )
+    external_integration = models.ForeignKey(
+        to="extras.ExternalIntegration",
+        on_delete=models.SET_NULL,
```
glennmatthews, line 1281: This "feels" like it should be `PROTECT` - I don't think it makes sense to allow a user to delete an ExternalIntegration if there's a Controller using it.

nautobot/dcim/models/devices.py
```diff
@@ -1220,3 +1230,144 @@ def __str__(self):
...
+    deployed_controller_device = models.ForeignKey(
+        to="dcim.Device",
+        on_delete=models.SET_NULL,
+        related_name="controllers",
+        blank=True,
+        null=True,
+    )
+    deployed_controller_group = models.ForeignKey(
+        to="dcim.DeviceRedundancyGroup",
+        on_delete=models.SET_NULL,
```
glennmatthews, line 1295: Same thoughts here on SET_NULL versus PROTECT.

nautobot/dcim/tests/test_views.py
```diff
@@ -3083,3 +3088,49 @@ def setUpTestData(cls):
...
+class ControllerTestCase(ViewTestCases.PrimaryObjectViewTestCase):
+    model = Controller
+    filterset = ControllerFilterSet
+
+    @classmethod
+    def setUpTestData(cls):
+        statuses = Status.objects.get_for_model(Controller)
+        roles = Role.objects.get_for_model(Controller)
+        platforms = Platform.objects.all()
+        locations = Location.objects.all()
```
glennmatthews, line 3102:
```suggestion
        locations = Location.objects.get_for_model(Controller)
```

nautobot/docs/development/core/controllers.md
```diff
@@ -0,0 +1,164 @@
...
+location: DC-East
+platform: cisco_apic
+```
+
+### Cisco Meraki
+
+```yaml
+name: Cisco Meraki SAAS
+deployed_controller_device: ~
+location: Cloud "Location
```
glennmatthews, line 142: Something incomplete/missing here?

nautobot/dcim/models/devices.py
glennmatthews: Thinking about this more on the latest diffs - see https://github.com/nautobot/nautobot/pull/5325#discussion_r1526772273

glennmatthews:
> > What should be the default statuses and roles for `Controller` model pls.?
> 
> Statuses should match device defaults.
> 
> Roles we do not provide default values outside of a few ipam models iirc.

We're trying to do better at providing default roles just like we do for statuses, so I think it would be good to have a few for Controller.  Not sure offhand what roles would make sense - maybe @mzbroch has thoughts?

mzbroch:
> We're trying to do better at providing default roles just like we do for statuses, so I think it would be good to have a few for Controller. Not sure offhand what roles would make sense - maybe @mzbroch has thoughts?

I don't think role field was included in the original design for this model. There are a few roles, that could also be considered as types: wireless, sdn, dc, fabric etc - however they are not recognized as "standards" so we'd have to propose our opinions.

nautobot/dcim/filters/__init__.py
snaselj: Fixed

nautobot/dcim/fixtures/controller-data.json
snaselj: Removed

nautobot/dcim/tables/devices.py
snaselj: Any proposal pls?

nautobot/dcim/views.py
snaselj: Removed as unnecessary.

nautobot/dcim/tests/test_api.py
snaselj: Fixed, added to factory, added validation and unittest.

nautobot/dcim/tests/test_models.py
snaselj: Added test for the signal.

nautobot/dcim/models/devices.py
snaselj: WDYT @whitej6 ?

nautobot/dcim/models/devices.py
snaselj: WDYT @whitej6 ?

nautobot/docs/development/core/controllers.md
snaselj: Removed apostrophe. Not sure, whether or how to update these examples.

nautobot/dcim/signals.py
```diff
@@ -286,3 +287,27 @@ def prevent_adding_tagged_vlans_with_incorrect_mode_or_site(sender, instance, ac
...
+    logger = logging.getLogger(__name__ + ".ControllerDeviceGroup")
+
+    if raw:
+        logger.debug("Skipping controller update for imported controller device group %s", instance)
+        return
+
+    with transaction.atomic():
+        for group in instance.descendants(include_self=False):
+            group.controller = instance.controller
+            group.save()
```
snaselj, line 312: I would rather keep `save()`, not to block the instance update when some of it's descendants is not valid.

nautobot/dcim/signals.py
```diff
@@ -286,3 +287,27 @@ def prevent_adding_tagged_vlans_with_incorrect_mode_or_site(sender, instance, ac
...
+        return
+
+    logger = logging.getLogger(__name__ + ".ControllerDeviceGroup")
+
+    if raw:
+        logger.debug("Skipping controller update for imported controller device group %s", instance)
+        return
+
+    with transaction.atomic():
+        for group in instance.descendants(include_self=False):
```
snaselj, line 310: Unable to call `instance.descendants(include_self=False).update(controller=instance.controller`, fails with TreeModel related SQL error.

nautobot/dcim/templates/dcim/controllerdevicegroup_create.html
```diff
@@ -0,0 +1,43 @@
+{% extends 'generic/object_create.html' %}
+{% load form_helpers %}
+{% load helpers %}
+{% load static %}
+
+{% block javascript %}
+<script type="text/javascript">
+    const parentInput = $('select#id_parent');
```
snaselj, line 8: Should I refactor this not to be JQuery dependent? @glennmatthews

nautobot/dcim/templates/dcim/controllerdevicegroup_create.html
```diff
@@ -0,0 +1,43 @@
...
+    const updateControllerFromParent = (parentUuid) => {
+        $.ajax({
+            url: `${controllerUrl}${parentUuid}/`,
+            method: "GET",
+            success: (data) => {
+                if (data.controller) {
+                    controllerInput.val(data.controller.id);
+                }
+            },
+            error: (error) => console.log("ERROR", error),
```
snaselj, line 22: Is it necessary to better handle ajax errors? If so, is there some example available pls.? @glennmatthews

nautobot/dcim/templates/dcim/controllerdevicegroup_create.html
glennmatthews: If you *can* - the `select2` library that we use for most select widgets only emits jQuery events so we might have to keep at least some of the jQuery logic here.

nautobot/dcim/tables/devices.py
glennmatthews: Maybe don't show `external_integration`, `deployed_controller_device`, `deployed_controller_group` by default? Not sure.

nautobot/dcim/templates/dcim/controllerdevicegroup_create.html
glennmatthews: I don't think we have an established standard here but maybe should consider doing an `alert()` instead of a `console.log()` for now? No strong opinion.

changes/3111.added
```diff
@@ -0,0 +1 @@
+Initial work on the controller model.
```
glennmatthews, line 1:
```suggestion
Added `Controller` and `ControllerDeviceGroup` models.
```

glennmatthews: We also need a larger section in the `version-2.2.md` release notes highlighting this significant feature.

itdependsnetworks:
> > > We're trying to do better at providing default roles just like we do for statuses, so I think it would be good to have a few for Controller. Not sure offhand what roles would make sense - maybe @mzbroch has thoughts?
> > 
> > 
> > I don't think role field was included in the original design for this model. There are a few roles, that could also be considered as types: wireless, sdn, dc, fabric etc - however they are not recognized as "standards" so we'd have to propose our opinions.
> 
> @lampwins, @whitej6 do you have any opinions on default `Role` records for `Controller`? Or are we thinking that the field may not actually be necessary in this iteration and should be removed?

@mzbroch that is what is being asked for, your opinion.

mzbroch:
> @lampwins, @whitej6 do you have any opinions on default `Role` records for `Controller`? Or are we thinking that the field may not actually be necessary in this iteration and should be removed?

I don't think it's necessary to include role in first iteration. When we had design decisions, this field was not included on the model, however was discussed.

nautobot/dcim/templates/dcim/controllerdevicegroup_create.html
snaselj: Keeping as is

nautobot/dcim/templates/dcim/controllerdevicegroup_create.html
snaselj: Keeping as is

PR #5329, opened 2024-02-22
Add caching of ChangeLoggedModelsQuery().as_queryset() for enqueue_job_hooks()

glennmatthews:
# What's Changed

Some investigation in #5293 (credit to @Kircheneer) showed that when saving many objects in a change-logged context, we are spending a lot of time in `enqueue_job_hooks()`, and specifically in evaluating `ChangeLoggedModelsQuery().as_queryset()`, which returns a queryset of ContentTypes that support change-logging.

Since the contents of this queryset should never change while Django is running, adding an `lru_cache` seemed like a logical addition, and the results seem to speak for themselves. As an example, when creating and then saving 10000 individual Tenant objects in a `web_request_context`, the execution time improved from 336 seconds to 125 seconds.

Setup:

```py
>>> def tenants():
...     return [Tenant(name=f"T{i}") for i in range(10000)]
...
>>> user = User.objects.first()
>>> import time
>>> import contextlib
>>> 
>>> @contextlib.contextmanager
... def time_it():
...     before = time.process_time()
...     yield
...     after = time.process_time()
...     print(f"Took {after - before}s.")
... 
>>> from nautobot.extras.context_managers import web_request_context
>>> with time_it():
...     with web_request_context(user):
...         for tenant in tenants():
...             tenant.save()
... 
>>> Tenant.objects.all().delete()
```

Before:

```
Took 336.556344127s.
```

After:

```
Took 125.292067944s.
```

Although there are other places we're calling `ChangeLoggedModelsQuery().as_queryset()`, the others appear to be one-offs (mostly at declaration time) rather than something that would be called repeatedly, so I've left them alone.

# TODO

- n/a Unit, Integration Tests


nautobot/extras/utils.py
```diff
@@ -103,6 +104,14 @@ def list_subclasses(self):
         return [_class for _class in apps.get_models() if hasattr(_class, "to_objectchange")]
 
 
+@lru_cache(maxsize=5)
```
Kircheneer, line 107: Why 5? Surely there are more change-logged content types than 5? Unless I misunderstand the parameter.

glennmatthews: This is the max number of distinct cache entries to store, which normally would be keyed/indexed by the parameters passed when the function is called (i.e. so we'd have a cache entry for `my_function(1)` calls that's distinct for the cache entry for `my_function(2)`).

Since this function doesn't actually take any parameters, and always should return the same queryset for the lifetime of a Nautobot process, we could *probably* just set `maxsize=1` but for whatever reason `5` is the value we've used in some other places in the code that should also pretty much always be called in a single way, so I was following that pattern.

gsnider2195: I would propose just using [`functools.cache`](https://docs.python.org/3/library/functools.html#functools.cache) here since we don't ever want this to expire.

glennmatthews: I was going that way initially but `functools.cache` is only in python 3.9 and later and we still support 3.8 for now.

gsnider2195: Oh good catch. I think I saw that and didn't register that we still supported 3.8. Can we do `maxsize=None` then?

glennmatthews: If I'm reading the docs right, `maxsize=None` actually means "the cache can grow arbitrarily large", but it does say that it's slightly faster than an lru_cache with a maxsize set. I guess since we're only expecting a single cache entry in any case, it's probably fine?

nautobot/extras/tests/test_jobs.py
```diff
@@ -922,6 +923,12 @@ class JobHookTransactionTest(TransactionTestCase):  # TODO: BaseModelTestCase mi
 
     def setUp(self):
         super().setUp()
+        # Because of TransactionTestCase, and its clearing and repopulation of the database between tests,
+        # the `change_logged_models_queryset` cache of ContentTypes becomes invalid.
+        # We need to explicitly clear it here to have tests pass.
+        # This is not a problem during normal operation of Nautobot because content-types don't normally get deleted
+        # and recreated while Nautobot is running.
```
HanlinMiao, line 930: 👍

HanlinMiao (approved):

nautobot/extras/utils.py
gsnider2195: I don't think it's possible for the cache to be any larger than 1 entry here, since if you try to pass args or kwargs to this method it will raise an exception.

gsnider2195 (approved):

PR #5339, opened 2024-02-23
Interface/VMInterface VRF pre-migration helper

### LINKED ISSUE #4247: Pre-Migration Helper: Validate IPAM Data for Upcoming Restrictions (mixed VRFs on same interface, etc.)

gsnider2195:
# Closes #4247
# What's Changed

# Screenshots

## Test data

image

## Output

```
root@4a5581d8bb9f:/source# nautobot-server pre_migrate
>>> Running check: check_configcontext_uniqueness...
>>> Running check: check_exporttemplate_uniqueness...
>>> Running check: check_virtualchassis_uniqueness...
>>> Running interface/vminterface vrf uniqueness checks...
dcim.Interface Ethernet1/6 (rtr1) is associated to IP Addresses with different VRFs.
dcim.Interface Ethernet1/5 (rtr1) is associated to IP Addresses with different VRFs.
virtualization.VMInterface veth3 (vm1) is associated to IP Addresses with different VRFs.
virtualization.VMInterface veth5 (vm1) is associated to IP Addresses with different VRFs.
virtualization.VMInterface veth4 (vm1) is associated to IP Addresses with different VRFs.
>>> Running permission constraint checks...
CommandError: One or more pre-migration checks failed:
    You cannot migrate Interfaces or VMInterfaces associated to multiple IP Addresses with different VRFs.
```

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/4247.added
+++ b/changes/4247.added
@@ -1,1 +1,1 @@
-Added a pre-migration helper to identify Interfaces and VMInterfaces with multiple VRFs through IPAddress relationships.
+Added a check to the pre_migrate management command to identify Interfaces and VMInterfaces with multiple VRFs through IPAddress relationships.
--- a/changes/4247.added
+++ b/changes/4247.added
@@ -1,0 +1,1 @@
+Added a pre-migration helper to identify Interfaces and VMInterfaces with multiple VRFs through IPAddress relationships.
--- a/changes/4247.added
+++ b/changes/4247.added
@@ -1,1 +1,1 @@
-Added a check to the pre_migrate management command to identify Interfaces and VMInterfaces with multiple VRFs through IPAddress relationships.
+Added a check to the `nautobot-server pre_migrate` command to identify Interfaces and VMInterfaces with multiple VRFs through IPAddress relationships.
--- a/nautobot/core/management/commands/pre_migrate.py
+++ b/nautobot/core/management/commands/pre_migrate.py
@@ -158,7 +158,7 @@ def check_interface_ipaddress_vrf_uniqueness(command):
 
     if failed:
         raise ValidationError(
-            "You cannot migrate Interfaces or VMInterfaces associated to IP Addresses with different VRFs."
+            "You cannot migrate Interfaces or VMInterfaces associated to multiple IP Addresses with different VRFs."
         )
 
 
--- a/nautobot/core/management/commands/pre_migrate.py
+++ b/nautobot/core/management/commands/pre_migrate.py
@@ -443,6 +443,7 @@ class Command(BaseCommand):
             check_interface_ipaddress_vrf_uniqueness(self)
         except ValidationError as err:
             errors.append(err)
+
         check_permissions_constraints(self)
 
         if errors:
--- a/nautobot/core/management/commands/pre_migrate.py
+++ b/nautobot/core/management/commands/pre_migrate.py
@@ -129,6 +129,8 @@ def check_interface_ipaddress_vrf_uniqueness(command):
 
     failed = False
 
+    command.stdout.write(command.style.WARNING(">>> Running interface/vminterface vrf uniqueness checks..."))
+
     for ip_address_assignment in (
         IPAddress.objects.values("assigned_object_type", "assigned_object_id")
         .annotate(
--- a/nautobot/core/management/commands/pre_migrate.py
+++ b/nautobot/core/management/commands/pre_migrate.py
@@ -132,7 +132,8 @@ def check_interface_ipaddress_vrf_uniqueness(command):
     command.stdout.write(command.style.WARNING(">>> Running interface/vminterface vrf uniqueness checks..."))
 
     for ip_address_assignment in (
-        IPAddress.objects.values("assigned_object_type", "assigned_object_id")
+        IPAddress.objects.filter(assigned_object_id__isnull=False, assigned_object_type_id__isnull=False)
+        .values("assigned_object_type", "assigned_object_id")
         .annotate(
             vrf_count=models.Count("vrf_id", distinct=True),
             vrf_null_count=models.Count("pk", filter=models.Q(vrf_id__isnull=True)),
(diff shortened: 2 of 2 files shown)


nautobot/core/management/commands/pre_migrate.py
```diff
@@ -120,6 +120,48 @@ def check_configcontext_uniqueness():
...
+
+    command.stdout.write(command.style.WARNING(">>> Running interface/vminterface vrf uniqueness checks..."))
+
+    for ip_address_assignment in (
+        IPAddress.objects.values("assigned_object_type", "assigned_object_id")
+        .annotate(
+            vrf_count=models.Count("vrf_id", distinct=True),
+            vrf_null_count=models.Count("pk", filter=models.Q(vrf_id__isnull=True)),
+        )
+        .order_by()
```
HanlinMiao, line 140: What does `.order_by()` do here?

HanlinMiao (approved):

nautobot/core/management/commands/pre_migrate.py
gsnider2195: Gets rid of any custom sorting defined on the model so that the annotation doesn't get messed up by it

changes/4247.added
```diff
@@ -0,0 +1 @@
+Added a check to the pre_migrate management command to identify Interfaces and VMInterfaces with multiple VRFs through IPAddress relationships.
```
glennmatthews, line 1:
```suggestion
Added a check to the `nautobot-server pre_migrate` command to identify Interfaces and VMInterfaces with multiple VRFs through IPAddress relationships.
```

nautobot/core/management/commands/pre_migrate.py
```diff
@@ -120,6 +120,48 @@ def check_configcontext_uniqueness():
...
+
+    - All IP Addresses assigned to a specific Interface or VMInterface must share the same VRF or have no VRF.
+    """
+
+    failed = False
+
+    command.stdout.write(command.style.WARNING(">>> Running interface/vminterface vrf uniqueness checks..."))
+
+    for ip_address_assignment in (
+        IPAddress.objects.values("assigned_object_type", "assigned_object_id")
```
glennmatthews, line 135: This "feels" backwards to me, but maybe I'm not understanding it. 
- Why are we iterating over all IPAddresses instead of iterating over all Interfaces and all VMInterfaces?
- Shouldn't we filter out / discard IPs that have no `assigned_object` at all (or (VM)Interfaces that have no `ip_addresses`) for efficiency?

nautobot/core/management/commands/pre_migrate.py
```diff
@@ -120,6 +120,48 @@ def check_configcontext_uniqueness():
...
+    ):
+        has_null_vrf = 1 if ip_address_assignment["vrf_null_count"] > 0 else 0
+        if (
+            ip_address_assignment["vrf_count"] + has_null_vrf > 1
+            and ip_address_assignment["assigned_object_id"] is not None
+        ):
+            failed = True
+            assigned_object_type = ContentType.objects.get_for_id(
+                ip_address_assignment["assigned_object_type"]
+            ).model_class()
```
glennmatthews, line 150: Although highly unlikely to occur, for safety we should still account for the remote possibility that `ContentType.model_class()` returns None (maybe an IPAddress was assigned to some app model that's been uninstalled?)

nautobot/core/management/commands/pre_migrate.py
```diff
@@ -120,6 +120,48 @@ def check_configcontext_uniqueness():
...
+            assigned_object = assigned_object_type.objects.get(pk=ip_address_assignment["assigned_object_id"])
```
glennmatthews, line 151: Also should account for possibility of DoesNotExist exception being thrown here.

nautobot/core/management/commands/pre_migrate.py
```diff
@@ -120,6 +120,48 @@ def check_configcontext_uniqueness():
...
+        )
+        .order_by()
+    ):
+        has_null_vrf = 1 if ip_address_assignment["vrf_null_count"] > 0 else 0
+        if (
+            ip_address_assignment["vrf_count"] + has_null_vrf > 1
+            and ip_address_assignment["assigned_object_id"] is not None
+        ):
+            failed = True
+            assigned_object_type = ContentType.objects.get_for_id(
```
glennmatthews, line 148: Do we need to account for a null assigned_object_type and/or a possible ContentType.DoesNotExist exception here?

nautobot/core/management/commands/pre_migrate.py
```diff
@@ -120,6 +120,48 @@ def check_configcontext_uniqueness():
...
+                ip_address_assignment["assigned_object_type"]
+            ).model_class()
+            assigned_object = assigned_object_type.objects.get(pk=ip_address_assignment["assigned_object_id"])
+            command.stdout.write(
+                command.style.WARNING(
+                    f"{assigned_object_type._meta.label} {assigned_object} ({ip_address_assignment['assigned_object_id']}) "
```
glennmatthews, line 154: Probably would be good to print the Device/VirtualMachine as well since "Ethernet1/6" alone for example isn't very useful.

gsnider2195: This is why I included the PK but I can add the parent too

nautobot/core/management/commands/pre_migrate.py
gsnider2195: This query is providing all of the information that is required for this check directly from the database:

- Aggregation of `assigned_object_type`/`assigned_object_id`
- Count of unique vrfs associated to that `assigned_object_type`/`assigned_object_id` combination
- Count of `IPAddress` objects associated to that `assigned_object_type`/`assigned_object_id` combination with a null vrf (workaround for `null != null` and `COUNT` only returning non-null values)

Using this query:

```sql
SELECT 
  "ipam_ipaddress"."assigned_object_type_id", 
  "ipam_ipaddress"."assigned_object_id", 
  COUNT(
    DISTINCT "ipam_ipaddress"."vrf_id"
  ) AS "vrf_count", 
  COUNT("ipam_ipaddress"."id") FILTER (
    WHERE 
      "ipam_ipaddress"."vrf_id" IS NULL
  ) AS "vrf_null_count" 
FROM 
  "ipam_ipaddress" 
GROUP BY 
  "ipam_ipaddress"."assigned_object_type_id", 
  "ipam_ipaddress"."assigned_object_id";

```

We could probably put a `.filter(assigned_object_id__isnull=False, assigned_object_type_id__isnull=False)` in there to speed it up:

```sql
SELECT 
  "ipam_ipaddress"."assigned_object_type_id", 
  "ipam_ipaddress"."assigned_object_id", 
  COUNT(
    DISTINCT "ipam_ipaddress"."vrf_id"
  ) AS "vrf_count", 
  COUNT("ipam_ipaddress"."id") FILTER (
    WHERE 
      "ipam_ipaddress"."vrf_id" IS NULL
  ) AS "vrf_null_count" 
FROM 
  "ipam_ipaddress" 
WHERE 
  (
    "ipam_ipaddress"."assigned_object_id" IS NOT NULL 
    AND "ipam_ipaddress"."assigned_object_type_id" IS NOT NULL
  ) 
GROUP BY 
  "ipam_ipaddress"."assigned_object_type_id", 
  "ipam_ipaddress"."assigned_object_id";

```

nautobot/core/management/commands/pre_migrate.py
gsnider2195: I didn't think it was likely but it might be a good idea to catch it just in case

glennmatthews (approved):

Issue #5340, opened 2024-02-23
Nautobot Health Checks Documentation


HanlinMiao:
~1. Identify/implement appropriate health check CLI command/commands for the nautobot-server, Celery, Beat.~ done
2. Identify/document health check usage recommendations for single system deployments as well as Kubernetes and docker-compose.

Related issues #1102, #4292

Issue #4292, opened 2023-08-18
`nautobot-server` commands always calling `import_jobs_as_celery_tasks` when called


whitej6:
### Environment
* Nautobot version (Docker tag too if applicable): v2.0.0rc1
* Python version: 3.10
* Database platform, version: psql
* Middleware(s):

### Steps to Reproduce
1. `nautobot-server shell_plus` but appears to be any command 
2.
3.

### Expected Behavior
Command is executed and only the code required is called

### Observed Behavior
output sample
```
root@e65460e688ce:/source# nautobot-server version
20:48:58.015 DEBUG   nautobot.core.celery __init__.py        import_jobs_as_celery_tasks() :
  Importing system Jobs
20:48:58.173 DEBUG   nautobot.core.celery __init__.py                      register_jobs() :
  Registering job nautobot.core.jobs.GitRepositorySync
20:48:58.177 DEBUG   nautobot.core.celery __init__.py                      register_jobs() :
  Registering job nautobot.core.jobs.GitRepositoryDryRun
3.2.20
```

jathanism:
This signal is core to how Jobs operate in v2, because they are all Celery tasks now and additional also imported as real modules. It also fires whenever a worker process starts up.

Unfortunately, without calling this within the main Nautobot application, workers will have the Job modules imported, but Nautobot won't.

This is therefore working as intended.

Can you please elaborate on why you think this is a bug and how we might account for this registration? Is it perhaps the verbose logging that is adding confusion?

whitej6:
This goes back to what should be called for EVERY management command and the tax paid for running a management command.

Another example is `nautobot-server collectstatic` calls this same set of tasks

Every additional piece we add that has to be evaluated slows down the start up process of nautobot which is compounded when running nautobot in containers as the entrypoint for a container wraps several management commands and IMHO nullifies the ability to use effective auto-scaling for nautobot.

jathanism:
> This goes back to what should be called for EVERY management command and the tax paid for running a management command.
> 
> Another example is `nautobot-server collectstatic` calls this same set of tasks
> 
> Every additional piece we add that has to be evaluated slows down the start up process of nautobot which is compounded when running nautobot in containers as the entrypoint for a container wraps several management commands and IMHO nullifies the ability to use effective auto-scaling for nautobot.

[...]

PR #5342, opened 2024-02-23
Switch from curl-based default Docker healthcheck to a CLI-based one

glennmatthews:
# Closes N/A (but relates to #5340)
# What's Changed

- Switch default Docker healthcheck from a `curl` based one (which can fail if all request-processing workers/processes/threads are busy with other requests) to a CLI based one calling `nautobot-server health_check`.
   - Because of #4292 and related startup code, the `nautobot-server health_check` takes about 6 seconds to execute, so I increased the default `interval` and `timeout` for the healthcheck from 5s to 10s. We should be able to bring this back down once we improve the performance of `nautobot-server` startup time.
   - I changed the default `start-period` from 5s to 5m since we know that initial migrations may take several minutes to complete.
   - With the changes to the default health-check, we no longer need to override it with a custom health-check in `docker-compose.yml` and `docker-compose.final.yml`.
- Add a new healthcheck backend based on `health_check.contrib.migrations` implementation. This fails if there are any un-applied migrations detected and passes if all migrations are in effect. This is needed because while the `/health/` URL endpoint won't start responding until the nautobot-server process is serving responses, and therefore implicitly will fail while migrations are in process, the `nautobot-server health_check` CLI spins up its own process to report back ASAP. We don't want the container to report as healthy before migrations are completed, as dependent containers/processes (celery worker, celery beat) are likely to encounter errors if they try to start up while migrations are in-flight.

### QUESTION: should this go into `develop` as a bug fix or `next` as a feature/behavior-change?

# Screenshots

```
# nautobot-server health_check
DatabaseBackend          ... working
DefaultFileStorageHealthCheck ... working
MigrationsBackend        ... unavailable: There are migrations not yet applied
RedisBackend             ... working
```

```
# nautobot-server health_check
DatabaseBackend          ... working
DefaultFileStorageHealthCheck ... working
MigrationsBackend        ... working
RedisBackend             ... working
```

# TODO

- n/a Unit, Integration Tests

### DIFF
--- a/nautobot/extras/apps.py
+++ b/nautobot/extras/apps.py
@@ -43,7 +43,7 @@ class ExtrasConfig(NautobotConfig):
                 "during the execution of the migration command for the first time."
             )
 
-        # Register the DatabaseBackend health check
+        # Register the DatabaseBackend, MigrationsBackend, and RedisBackend health checks
         from nautobot.extras.health_checks import DatabaseBackend, MigrationsBackend, RedisBackend
 
         plugin_dir.register(DatabaseBackend)
--- a/nautobot/extras/apps.py
+++ b/nautobot/extras/apps.py
@@ -44,9 +44,10 @@ class ExtrasConfig(NautobotConfig):
             )
 
         # Register the DatabaseBackend health check
-        from nautobot.extras.health_checks import DatabaseBackend, RedisBackend
+        from nautobot.extras.health_checks import DatabaseBackend, MigrationsBackend, RedisBackend
 
         plugin_dir.register(DatabaseBackend)
+        plugin_dir.register(MigrationsBackend)
         plugin_dir.register(RedisBackend)
 
         # Register built-in SecretsProvider classes
--- a/nautobot/extras/health_checks.py
+++ b/nautobot/extras/health_checks.py
@@ -3,7 +3,8 @@ from typing import Optional
 from urllib.parse import urlparse
 
 from django.conf import settings
-from django.db import connection, DatabaseError, IntegrityError
+from django.db import connection, connections, DatabaseError, DEFAULT_DB_ALIAS, IntegrityError
+from django.db.migrations.executor import MigrationExecutor
 from health_check.backends import BaseHealthCheckBackend
 from health_check.exceptions import ServiceReturnedUnexpectedResult, ServiceUnavailable
 from prometheus_client import Gauge
@@ -65,10 +66,32 @@ class DatabaseBackend(NautobotHealthCheckBackend):
                 obj.title = "newtest"
                 obj.save()
                 obj.delete()
-        except IntegrityError:
-            raise ServiceReturnedUnexpectedResult("Integrity Error")
-        except DatabaseError:
-            raise ServiceUnavailable("Database error")
+        except IntegrityError as e:
+            self.add_error(ServiceReturnedUnexpectedResult("Integrity Error"), e)
+        except DatabaseError as e:
+            self.add_error(ServiceUnavailable("Database error"), e)
+
+
+class MigrationsBackend(NautobotHealthCheckBackend):
+    """Check whether all migrations have been applied."""
+
+    metric = Gauge(
+        "health_check_migrations_info",
+        "State of migrations backend",
+        multiprocess_mode=NautobotHealthCheckBackend.MULTIPROCESS_MODE,
+    )
+
+    def check_status(self):
+        db_alias = getattr(settings, "HEALTHCHECK_MIGRATIONS_DB", DEFAULT_DB_ALIAS)
+        try:
+            executor = MigrationExecutor(connections[db_alias])
+            plan = executor.migration_plan(executor.loader.graph.leaf_nodes())
+            if plan:
+                self.add_error(ServiceUnavailable("There are migrations not yet applied"))
+        except DatabaseError as e:
+            self.add_error(ServiceUnavailable("Database is not ready"), e)
+        except Exception as e:
+            self.add_error(ServiceUnavailable("Unknown error"), e)
 
 
 class RedisHealthCheck(NautobotHealthCheckBackend):
--- a/changes/5342.added
+++ b/changes/5342.added
@@ -1,0 +1,1 @@
+Added `MigrationsBackend` to health-check, which will fail if any unapplied database migrations are present.
--- a/changes/5342.changed
+++ b/changes/5342.changed
@@ -1,0 +1,1 @@
+Changed default Docker HEALTHCHECK to use `nautobot-server health_check` CLI command.
--- a/development/docker-compose.final.yml
+++ b/development/docker-compose.final.yml
@@ -8,12 +8,6 @@ services:
     image: "local/nautobot-final:local-py${PYTHON_VER}"
     ports:
       - 8443:8443
-    healthcheck:
-      test:
-        - "CMD"
-        - "curl"
-        - "-fk"
-        - "https://localhost:8443/health/"
   celery_worker:
     image: "local/nautobot-final:local-py${PYTHON_VER}"
     entrypoint: "nautobot-server celery worker -l INFO --events"
--- a/development/docker-compose.yml
+++ b/development/docker-compose.yml
@@ -25,16 +25,6 @@ services:
     env_file:
       - dev.env
     tty: true
-    healthcheck:
-      interval: 5s
-      timeout: 5s
-      start_period: 5m  # it takes a WHILE to run initial migrations with an empty DB
-      retries: 3
-      test:
-        - "CMD"
-        - "curl"
-        - "-f"
-        - "http://localhost:8080/health/"
   celery_worker:
     image: "local/nautobot-dev:local-py${PYTHON_VER}"
     ports:
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -58,7 +58,9 @@ RUN --mount=type=cache,target="/root/.cache/pip",sharing=locked \
     --mount=type=cache,target="/tmp",sharing=locked \
     pip install --upgrade pip wheel
 
-HEALTHCHECK --interval=5s --timeout=5s --start-period=5s --retries=1 CMD curl --fail http://localhost:8080/health/ || exit 1
+# timeout/interval=10s because `nautobot-server` can be slow to start - https://github.com/nautobot/nautobot/issues/4292
+# start-period=5m because initial migrations can take several minutes to run on a fresh DB
+HEALTHCHECK --interval=10s --timeout=10s --start-period=5m --retries=3 CMD nautobot-server health_check
 
 # Generate nautobot user and its required dirs for later consumption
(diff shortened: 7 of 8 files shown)


nautobot/extras/health_checks.py
```diff
@@ -65,10 +66,32 @@ def check_status(self):
...
+        multiprocess_mode=NautobotHealthCheckBackend.MULTIPROCESS_MODE,
+    )
+
+    def check_status(self):
+        db_alias = getattr(settings, "HEALTHCHECK_MIGRATIONS_DB", DEFAULT_DB_ALIAS)
+        try:
+            executor = MigrationExecutor(connections[db_alias])
+            plan = executor.migration_plan(executor.loader.graph.leaf_nodes())
+            if plan:
+                self.add_error(ServiceUnavailable("There are migrations not yet applied"))
```
gsnider2195, line 90: Does this impact development of new migrations? I'm typically rolling a migration back, starting the web server to add some rest data, then testing my in-flight migration. This won't prevent the server from starting with unapplied migrations right? Biggest impact might be beat and celery workers not starting up?

glennmatthews: I believe you're correct. This shouldn't prevent the server from starting, but it will prevent the docker health check from reporting as healthy until you apply your migration. I think this could be worked around easily enough with a docker-compose.override.yml but not 100% certain.

gsnider2195:
> QUESTION: should this go into develop as a bug fix or next as a feature/behavior-change?

I would say if this was a development environment change, then develop would be ok but since this is changing the Dockerfile, it should go to next.

bryanculver:
Could any of the Celery commands be used to get the status/availability of workers?

glennmatthews:
> Could any of the Celery commands be used to get the status/availability of workers?

Sure, but I don't think we want the server container to report as unhealthy when no workers are running? Would cause a chicken-and-egg problem between the server container and the worker container(s).

gsnider2195:
> > Could any of the Celery commands be used to get the status/availability of workers?
> 
> Sure, but I don't think we want the server container to report as unhealthy when no workers are running? Would cause a chicken-and-egg problem between the server container and the worker container(s).

Also, the celery inspect commands are slow. They send out a ping and wait for a specified timeout for workers to report back.

nautobot/extras/health_checks.py
gsnider2195: As currently implemented, this doesn't actually check the health of the active nautobot server process. If that check was implemented this migrations check wouldn't be required. Are you planning to implement that as a future enhancement? That was one of our items from the design discussion:

> Doing a nautobot-server healthcheck management command - should check both that the “real” server process is running and also that things like DB connectivity are working

nautobot/extras/apps.py
```diff
@@ -44,9 +44,10 @@ def convert_field_to_list_tags(field, registry=None):
             )
 
         # Register the DatabaseBackend health check
```
gsnider2195, line 46:
```suggestion
        # Register the DatabaseBackend, MigrationsBackend, and RedisBackend health checks
```

nautobot/extras/health_checks.py
```diff
@@ -65,10 +66,32 @@ def check_status(self):
...
+        "health_check_migrations_info",
+        "State of migrations backend",
+        multiprocess_mode=NautobotHealthCheckBackend.MULTIPROCESS_MODE,
+    )
+
+    def check_status(self):
+        db_alias = getattr(settings, "HEALTHCHECK_MIGRATIONS_DB", DEFAULT_DB_ALIAS)
+        try:
+            executor = MigrationExecutor(connections[db_alias])
+            plan = executor.migration_plan(executor.loader.graph.leaf_nodes())
```
gsnider2195, line 88: Building the migration plan can be CPU intensive. Is there a way we can query the `django_migrations` table and find out if any of the migrations aren't applied?

nautobot/extras/health_checks.py
glennmatthews: That may be desirable as a separate command - for the Docker case we don't need to check the server process health since the container will die if the server process exits, and I think we don't in general have a way to detect "health" of the process other than "is it running", as "running and busy processing HTTP requests" is mostly indistinguishable externally from "running and stuck in an infinite loop somewhere"

nautobot/extras/health_checks.py
glennmatthews: The `django_migrations` table only records migrations that *have been* applied, so we have to rely on Django machinery to tell us which migrations *should be* applied.

I tried the following alternate implementation:

```diff
@@ -84,10 +85,11 @@ class MigrationsBackend(NautobotHealthCheckBackend):
     def check_status(self):
         db_alias = getattr(settings, "HEALTHCHECK_MIGRATIONS_DB", DEFAULT_DB_ALIAS)
         try:
-            executor = MigrationExecutor(connections[db_alias])
-            plan = executor.migration_plan(executor.loader.graph.leaf_nodes())
-            if plan:
-                self.add_error(ServiceUnavailable("There are migrations not yet applied"))
+            loader = MigrationLoader(connections[db_alias], ignore_no_migrations=True)
+            for app_name, final_migration in loader.graph.leaf_nodes():
+                if (app_name, final_migration) not in loader.applied_migrations:
+                    self.add_error(ServiceUnavailable("There are migrations not yet applied"))
+                    break
         except DatabaseError as e:
             self.add_error(ServiceUnavailable("Database is not ready"), e)
         except Exception as e:
```

but `cProfile` shows little performance difference between the two:

### Original implementation

```
         74090 function calls (74065 primitive calls) in 0.115 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     19/1    0.000    0.000    0.115    0.115 {built-in method builtins.exec}
        1    0.000    0.000    0.115    0.115 <string>:1(<module>)
        1    0.000    0.000    0.114    0.114 health_checks.py:84(check_status)
        1    0.000    0.000    0.113    0.113 executor.py:16(__init__)
        1    0.000    0.000    0.113    0.113 loader.py:43(__init__)
        1    0.000    0.000    0.113    0.113 loader.py:207(build_graph)
        1    0.001    0.001    0.095    0.095 loader.py:68(load_disk)
       18    0.000    0.000    0.045    0.002 loader.py:108(<setcomp>)
      368    0.001    0.000    0.044    0.000 pkgutil.py:110(iter_modules)
      368    0.002    0.000    0.043    0.000 pkgutil.py:144(_iter_file_finder_modules)
       18    0.000    0.000    0.034    0.002 __init__.py:133(reload)
       36    0.026    0.001    0.026    0.001 {built-in method posix.listdir}
      164    0.021    0.000    0.021    0.000 {built-in method posix.stat}
       18    0.000    0.000    0.019    0.001 <frozen importlib._bootstrap>:583(_exec)
       18    0.000    0.000    0.018    0.001 <frozen importlib._bootstrap_external>:837(exec_module)
       18    0.000    0.000    0.018    0.001 <frozen importlib._bootstrap_external>:909(get_code)
       38    0.001    0.000    0.016    0.000 <frozen importlib._bootstrap>:890(_find_spec)
      128    0.000    0.000    0.016    0.000 <frozen importlib._bootstrap_external>:135(_path_stat)
       38    0.000    0.000    0.015    0.000 <frozen importlib._bootstrap_external>:1399(find_spec)
       38    0.000    0.000    0.015    0.000 <frozen importlib._bootstrap_external>:1367(_get_spec)
       38    0.001    0.000    0.015    0.000 <frozen importlib._bootstrap_external>:1498(find_spec)
       18    0.001    0.000    0.015    0.001 <frozen importlib._bootstrap_external>:1029(get_data)
      390    0.001    0.000    0.013    0.000 __init__.py:109(import_module)
      390    0.000    0.000    0.013    0.000 <frozen importlib._bootstrap>:1002(_gcd_import)
      390    0.002    0.000    0.012    0.000 <frozen importlib._bootstrap>:986(_find_and_load)
```


### Alternative implementation

```
         73804 function calls (73779 primitive calls) in 0.107 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     19/1    0.000    0.000    0.107    0.107 {built-in method builtins.exec}
        1    0.000    0.000    0.107    0.107 <string>:1(<module>)
        1    0.000    0.000    0.106    0.106 health_checks.py:85(check_status)
        1    0.000    0.000    0.105    0.105 loader.py:43(__init__)
        1    0.001    0.001    0.105    0.105 loader.py:207(build_graph)
        1    0.001    0.001    0.084    0.084 loader.py:68(load_disk)
       18    0.000    0.000    0.039    0.002 loader.py:108(<setcomp>)
      368    0.000    0.000    0.038    0.000 pkgutil.py:110(iter_modules)
      368    0.001    0.000    0.037    0.000 pkgutil.py:144(_iter_file_finder_modules)
       18    0.000    0.000    0.031    0.002 __init__.py:133(reload)
       37    0.024    0.001    0.024    0.001 {built-in method posix.listdir}
      164    0.018    0.000    0.018    0.000 {built-in method posix.stat}
       18    0.000    0.000    0.017    0.001 <frozen importlib._bootstrap>:583(_exec)
       38    0.001    0.000    0.016    0.000 <frozen importlib._bootstrap>:890(_find_spec)
       18    0.000    0.000    0.016    0.001 <frozen importlib._bootstrap_external>:837(exec_module)
       18    0.000    0.000    0.016    0.001 <frozen importlib._bootstrap_external>:909(get_code)
       38    0.000    0.000    0.015    0.000 <frozen importlib._bootstrap_external>:1399(find_spec)
       38    0.000    0.000    0.015    0.000 <frozen importlib._bootstrap_external>:1367(_get_spec)
       38    0.000    0.000    0.015    0.000 <frozen importlib._bootstrap_external>:1498(find_spec)
      128    0.000    0.000    0.014    0.000 <frozen importlib._bootstrap_external>:135(_path_stat)
       18    0.001    0.000    0.013    0.001 <frozen importlib._bootstrap_external>:1029(get_data)
      390    0.000    0.000    0.012    0.000 __init__.py:109(import_module)
      390    0.000    0.000    0.012    0.000 <frozen importlib._bootstrap>:1002(_gcd_import)
      390    0.002    0.000    0.011    0.000 <frozen importlib._bootstrap>:986(_find_and_load)
```

gsnider2195: It's the graph build that's CPU intensive. The majority of the time is spent in `loader.py:207(build_graph)` but a big percentage of that is loading the migrations from disk. Probably not worth worrying about until we get a few thousand more migration files 😄 but thank you for humoring me!

glennmatthews: Sure thing. It's been helpful in getting my hands dirty with `cProfile` a bit more. Relatedly, it *looks* like the majority of the `nautobot-server` startup time in general is spent purely in discovering and importing modules rather than doing any particularly expensive function calls. So that'll be interesting to try and optimize.

gsnider2195 (approved):

PR #5346, opened 2024-02-26
fix: Fixed LLDP neighbor view where interface names have spaces

abates:
HTML IDs cannot, by standard, have spaces in their names. The LLDP neighbor view tab sets the output table row's ID to the interface name. If a device interface name has a space, this breaks the jquery lookup to find the interface row. This fix changes the lookup to use a `data-` attribute for the interface name and lookup.

# Closes None
# What's Changed

LLDP results from the Napalm getter are mapped to the result table using an HTML ID constructed from the interface name. If an interface name includes a space then the jQuery lookup fails since spaces are not valid characters for HTML IDs. This change alters the behavior so that the rows are identified using the `data-interface-name` attribute and associated jQuery lookup.

# Screenshots

Screen Shot 2024-02-26 at 11 43 24 AM

# TODO

- [ ] ~~Unit, Integration Tests~~
- [ ] ~~Documentation Updates (when adding/changing features)~~
- [ ] ~~Example Plugin Updates (when adding/changing features)~~
- [ ] ~~Outline Remaining Work, Constraints from Design~~


gsnider2195:
This looks ok to me but I don't have the ability to test this. Can you add a screenshot of what this looks like when it's not working? (probably the dev tools console output?)

abates:
Here is a screenshot showing how the IDs have spaces:

Screen Shot 2024-02-27 at 11 32 38 AM

Here is a screenshot showing that the jQuery lookup doesn't find matching elements:

Screen Shot 2024-02-27 at 11 33 47 AM

nautobot/dcim/templates/dcim/device/lldp_neighbors.html
```diff
@@ -59,7 +59,7 @@
         success: function(json) {
             $.each(json['get_lldp_neighbors_detail'], function(iface, neighbors) {
                 var neighbor = neighbors[0];
-                var row = $('#' + iface.split(".")[0].replace(/([\/:])/g, "\\$1"));
+                var row = $('*[data-interface-name="' + iface.split(".")[0].replace(/([\/:])/g, "\\$1") + '"]');
```
glennmatthews, line 62: In the longer term we'll want to replace this with vanilla JS, but this looks OK for now.

glennmatthews (approved):

nautobot/dcim/templates/dcim/device/lldp_neighbors.html
abates: I think this can be done with `document.querySelectorAll`. I'm happy to make that change (or at least attempt to) if you like.

gsnider2195: jQuery's not dead yet! I'm fine with merging this in with jQuery since this is just fixing a bug, it's not implementing any new jQuery code. But if it's something you'd like to do and you have the spare cycles we can keep this open for a bit @abates

gsnider2195 (approved):

Kircheneer:
> This looks ok to me but I don't have the ability to test this. Can you add a screenshot of what this looks like when it's not working? (probably the dev tools console output?)

@gsnider2195 just FYI, the `mock` NAPALM driver theoretically exists, it just loads files from the file system and could be used to test things like these

PR #5365, opened 2024-02-29
Fixed `invalidate_max_depth_cache` itself calculating `max_depth` on …

Kircheneer:
…querysets without tree fields.

# Closes DNE
# What's Changed

`invalidate_max_depth_cache` currently uses `hasattr`, which accesses `sender.objects.max_depth` and therefore causes the value to be populated, and then immediately drops it. This change makes it so that we try deleting the field, and pass the exception that is raised when it doesn't exist.

Note that you can simply revert the `invalidate_max_depth_cache` implementation to the before state and the test should fail at the `assertNumQueries` assertion to validate this change.

# Screenshots
n.a.

- [x] Unit, Integration Tests


nautobot/core/tests/test_tree_queries.py
```diff
@@ -2,6 +2,25 @@
...
+        with self.assertNumQueries(4):
+            Location.objects.without_tree_fields().update_or_create(
+                name=location.name, defaults={"description": "My updated signal description"}
+            )
+
+        try:
+            del Location.objects.max_depth
+            self.fail("`max_depth` should not have been calculated.")
+        except AttributeError:
+            pass
```
joewesch, line 21: Nit-pick, and I like to perpetuate the use of more try/else blocks 😄 
```suggestion
        try:
            del Location.objects.max_depth
        except AttributeError:
            pass
        else:
            self.fail("`max_depth` should not have been calculated.")
```

nautobot/core/tests/test_tree_queries.py
```diff
@@ -2,6 +2,25 @@
...
 
+class TestInvalidateMaxTreeDepthSignal(TestCase):
+    """Tests for the max tree depth cache invalidation signal."""
+
+    def test_invalidate_max_tree_depth_without_tree_fields(self):
+        """Test that max tree depth is not calculated by the invalidate_max_tree_depth signal."""
+        location = Location.objects.first()
+        # Two SAVEPOINT, One SELECT, one UPDATE
+        with self.assertNumQueries(4):
+            Location.objects.without_tree_fields().update_or_create(
```
gsnider2195, line 13: What's the purpose of `update_or_create` here?

Kircheneer: Normal create should be enough, the original issue was using `update_or_create` which is why I started recreating using that. I will change it.

gsnider2195:
@Kircheneer we'd like to get this in the patch release for Monday. Unless you want to work on this over the weekend or Monday morning, I'm going to make the requested updates above and merge this.

Kircheneer:
Go ahead. I wont have time.

nautobot/core/signals.py
```diff
@@ -64,5 +64,11 @@ def invalidate_max_depth_cache(sender, **kwargs):
     """Clear the appropriate TreeManager.max_depth cache as the create/update/delete may have changed the tree."""
     from nautobot.core.models.tree_queries import TreeManager
 
-    if isinstance(sender.objects, TreeManager) and hasattr(sender.objects, "max_depth"):
+    if not isinstance(sender.objects, TreeManager):
+        return
+
+    # Note that we can't use `hasattr`, because this will trigger `max_depth` to be calculated if it doesn't exist
+    try:
```
gsnider2195, line 71: It looks like `__dict__.pop` is actually more performant here and more concise. Thoughts?

```py
# test cached_property invalidation speed
import functools
import timeit


class TestCache:
    @functools.cached_property
    def max_depth(self):
        return 1000


def invalidate_cache_del():
    t = TestCache()
    t.max_depth
    del t.max_depth


def invalidate_cache_pop():
    t = TestCache()
    t.max_depth
    t.__dict__.pop("max_depth", None)


def invalidate_cache_novalue_del():
    t = TestCache()
    try:
        del t.max_depth
    except AttributeError:
        pass


def invalidate_cache_novalue_pop():
    t = TestCache()
    t.__dict__.pop("max_depth", None)


def main():
    print("del:", timeit.timeit(invalidate_cache_del, number=1000000))
    print("pop:", timeit.timeit(invalidate_cache_pop, number=1000000))
    print("del no value:", timeit.timeit(invalidate_cache_novalue_del, number=1000000))
    print("pop no value:", timeit.timeit(invalidate_cache_novalue_pop, number=1000000))


if __name__ == "__main__":
    main()
```

Output:
```
del: 0.5046068249794189
pop: 0.571504608989926
del no value: 0.423761794983875
pop no value: 0.14525465900078416
```

joewesch: LGTM

glennmatthews: Looks small enough to be not a big factor either way, but I'm fine with either.

gsnider2195 (approved):

PR #5390, opened 2024-03-05
Fix pyuwsgi in Dockerfile and add a build-time check for SSL support

### LINKED ISSUE #5387: v2.1.6 and v1.6.13 Docker images - Unknown config directive: https
### Environment
* Nautobot version (Docker tag too if applicable): v2.1.6-py3.11
* Python version: 3.11 (Built into the docker image)
* Database platform, version: Postgres 16.2 inside container
* Middleware(s): Traefik to control ingress routing to Nautobot pod(s)

Upgraded to v2.1.6-py3.11 and receiving the following error in the nautobot-default container logs:

```
[uWSGI] getting INI configuration from /opt/nautobot/uwsgi.ini
[strict-mode] unknown config directive: https
```

The config-map that inserts uwsgi.ini during container creation has the following lines and has worked with no issues in version 2.1.5-py3.11

```
apiVersion: v1
data:
    [uwsgi]
    ; The IP address (typically localhost) and port that the WSGI process should listen on
    http = 0.0.0.0:8080
    https = 0.0.0.0:8443,/opt/nautobot/nautobot.crt,/opt/nautobot/nautobot.key


    ; Fail to start if any parameter in the configuration file isn’t explicitly understood by uWSGI
    strict = true

    ; Enable master process to gracefully re-spawn and pre-fork workers
    master = true

    ; Allow Python app-generated threads to run
    enable-threads = true

    ;Try to remove all of the generated file/sockets during shutdown
    vacuum = true

    ; Do not use multiple interpreters, allowing only Nautobot to run
    single-interpreter = true

    ; Shutdown when receiving SIGTERM (default is respawn)
    die-on-term = true

    ; Prevents uWSGI from starting if it is unable load Nautobot (usually due to errors)
    need-app = true

    ; By default, uWSGI has rather verbose logging that can be noisy
    disable-logging = true

    ; Assert that critical 4xx and 5xx errors are still logged
    log-4xx = true
    log-5xx = true

    ; Enable HTTP 1.1 keepalive support
    http-keepalive = 1

    ;
    ; Advanced settings (disabled by default)
    ; Customize these for your environment if and only if you need them.
    ; Ref: https://uwsgi-docs.readthedocs.io/en/latest/Options.html
    ;

    ; Number of uWSGI workers to spawn. This should typically be 2n+1, where n is the number of CPU cores present. Default 3 as n will be >= 1
    processes = 3

    ; Number of uWSGI threads each worker will be pre-forked into before starting
    threads = 2

    ; set the socket listen queue size, in production the suggested value is 1024, however RHEL based kernels have a max of 128 by default
    ; you may need to increase the somaxconn parameter in your kernel
    listen = 128

    ; If using subdirectory hosting e.g. example.com/nautobot, you must uncomment this line. Otherwise you'll get double paths e.g. example.com/nautobot/nautobot/.
    ; See: https://uwsgi-docs.readthedocs.io/en/latest/Changelog-2.0.11.html#fixpathinfo-routing-action
    ; route-run = fixpathinfo:

    ; If hosted behind a load balancer uncomment these lines, the harakiri timeout should be greater than your load balancer timeout.
    ; Ref: https://uwsgi-docs.readthedocs.io/en/latest/HTTP.html?highlight=keepalive#http-keep-alive
    ; harakiri = 65
    ; add-header = Connection: Keep-Alive
    ; http-keepalive = 1

    ; For larger installations, certain API calls (example: Relationships, GraphQL) can have a length of query parameters that go over uWSGI default limit.
    ; Setting the buffer size to larger than default (4096) can have an impact on memory utilization, but can be set as high as the header limit of 65535.
    buffer-size = 4096
```

### Steps to Reproduce
1. Roll out version 2.1.6 from 2.1.5 via Kubernetes
2. Observe pod state as new pods are spun up to replace the 2.1.5 pods
3. Observe the nautobot-default pod throwing the described error above in the logs and preventing container from fully starting (container keeps restarting)

### Expected behavior
nautobot-default pod should fully start without throwing said error in the logs

### Observed behavior
nautobot-default pod fails to start and constantly restarts

glennmatthews:
# Closes #5387 (will also need fixing in ltm-1.6)
# What's Changed

`poetry config installer.no-binary` is not additive, so `poetry config installer.no-binary pyuwsgi` was being overwritten by `poetry config installer.no-binary lxml`. The correct usage is `poetry config installer.no-binary pyuwsgi,lxml`. My fault for not testing #5359 adequately.

To check for and avoid this specific issue in the future, I've added a build-time check to the Dockerfile that checks `pyuwsgi --cflags` for the presence of the `-DUWSGI_SSL` compiler flag. I verified that without the poetry fix, the build fails due to this check, and with the poetry fix, the build succeeds.

# Screenshots

Before:

```
...
#29 [python-dependencies 7/7] RUN pyuwsgi --cflags | sed 's/ /\n/g' | grep UWSGI_SSL
#29 ERROR: executor failed running [/bin/sh -c pyuwsgi --cflags | sed 's/ /\n/g' | grep UWSGI_SSL]: exit code: 1

------
 > [python-dependencies 7/7] RUN pyuwsgi --cflags | sed 's/ /\n/g' | grep UWSGI_SSL:
------
failed to solve: executor failed running [/bin/sh -c pyuwsgi --cflags | sed 's/ /\n/g' | grep UWSGI_SSL]: exit code: 1
```

After:

```
...
#27 [python-dependencies 7/7] RUN pyuwsgi --cflags | sed 's/ /\n/g' | grep UWSGI_SSL
#27 0.346 -DUWSGI_SSL
#27 DONE 0.4s

#28 [python-dev-dependencies 1/1] RUN --mount=type=cache,target="/root/.cache",sharing=locked     poetry install --no-root --no-ansi --extras all &&     rm -rf /tmp/tmp*
#28 1.033 Skipping virtualenv creation, as specified in config file.
#28 1.759 Installing dependencies from lock file
...
```

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -123,7 +123,7 @@ ENV PATH="${POETRY_HOME}/bin:${PATH}"
 
 RUN poetry config virtualenvs.create ${POETRY_VIRTUALENVS_CREATE} && \
     poetry config installer.parallel "${POETRY_INSTALLER_PARALLEL}" && \
-    poetry config installer.no-binary pyuwsgi,lxml
+    poetry config installer.no-binary lxml,pyuwsgi
 
 ################################ Stage: python-dependencies (intermediate build target)
 
@@ -147,7 +147,8 @@ RUN --mount=type=cache,target="/root/.cache",sharing=locked \
     rm -rf /tmp/tmp*
 
 # Verify that pyuwsgi was installed correctly, i.e. with SSL support
-RUN pyuwsgi --cflags | sed 's/ /\n/g' | grep UWSGI_SSL
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+RUN pyuwsgi --cflags | sed 's/ /\n/g' | grep -e "^-DUWSGI_SSL$"
 
 ################################ Stage: python-dev-dependencies (intermediate build target)
 


docker/Dockerfile
```diff
@@ -147,6 +146,9 @@ RUN --mount=type=cache,target="/root/.cache",sharing=locked \
     poetry install --no-root --only main --no-ansi --extras all && \
     rm -rf /tmp/tmp*
 
+# Verify that pyuwsgi was installed correctly, i.e. with SSL support
+RUN pyuwsgi --cflags | sed 's/ /\n/g' | grep UWSGI_SSL
```
gsnider2195, line 150:
```suggestion
RUN pyuwsgi --cflags | sed 's/ /\n/g' | grep -e "^-DUWSGI_SSL$"
```

gsnider2195: Not blocking

gsnider2195 (approved):
Approved pending hadolint fix 😃

HanlinMiao (approved):

PR #3439, opened 2023-03-17
Remove slug from models

timizuoebideri1:
# Closes: #2907 
# What's Changed

# TODO

- [ ] Explanation of Change(s)
- [ ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [ ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example Plugin Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #5403, opened 2024-03-06
Replace `lru_cache` with django-redis caching

glennmatthews:
# Closes user-reported issue on Slack (no GH issue #)
# What's Changed

### Issue

In a deployment under uwsgi, a user noticed that the UI would give inconsistent behavior which could in some cases be resolved by refreshing the page. The example given was the "edit" form for a Device DynamicGroup, which would sometimes correctly display the defined Device CustomFields as filterable options for the DG, and other times would omit these filter fields from the edit form. 

After some investigation, it appears that the culprit was the `lru_cache` added on the `CustomField.objects.get_for_model()` method in #5024. With multiple uwsgi workers, creating or deleting a CustomField would trigger the `invalidate_lru_cache` signal *on the current worker only*, while any other workers that had also populated this cache locally would not have theirs cleared. Thus the inconsistent behavior - depending on which worker responded to any given HTTP request, the cache would have a different value and so different behavior would be observed.

In investigating and testing this, I found a similar issue with the usage of `@cached_property` on `TreeManager.max_depth` (#5131), with the same root cause.

### Old solution

~The solution proposed here is admittedly a bit of a blunt object. Since the *intended* performance optimization of the `lru_cache` added in #5024 was to reduce the impact of repeatedly running cacheable queries *within the context of a single request*, all I'm doing here is adding a new piece of middleware that clears the relevant caches before processing any new request.~

### New solution

Replace the local in-memory caches for this data with the shared django-redis cache, so that clearing the cache when appropriate automatically applies to all worker processes:
- `TreeManager.max_depth`
- `ComputedField.objects.get_for_model`
- `CustomField.objects.get_for_model`
- `Relationship.objects.get_for_model_source`
- `Relationship.objects.get_for_model_destination`

### Additional changes

- Add `NAUTOBOT_INSTALLATION_METRICS_ENABLED=false` to `development/dev.env` so that installation metrics are not unnecessarily sent when testing with the `final` images (which don't use `development/nautobot_config.py`)
- Add `caching.md` to developer documentation, spelling out best practices and conventions for using caching in Nautobot, and documenting some of the potential pitfalls.
- Remove `ChoiceSetMeta.__getattribute__` as it was making `ChoiceSet` references un-pickleable and was documented in #3439 as being "currently unused" in any case.
- Normalize cache keys as following either of the following two forms:
  - For Manager/Queryset/Model caches, `nautobot.<app>.<model>.<method>[.<variable>...]`
  - For function/method caches, `<module_path>.<function>[.<variable>...]`
- Remove unnecessary `lru_cache` micro-optimization of `remote_auth_enabled`/`sso_auth_enabled`/`ldap_auth_enabled`
- Fix our parallel test runner to ensure that each parallel test worker uses a distinct Redis cache index (`localhost:6379/<index>`) so that caches don't clobber one another during testing.

# Screenshots

Before, with a temporary log message reporting the state of the cache every time `DynamicGroupEditView` was loaded, showing that the initial cache of "no custom fields defined" was not cleared on every worker after a new custom field was created:

```
21:41:16.193 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet []>
21:41:25.946 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet []>
21:41:26.035 INFO    nautobot.extras.views :
  Created dynamic group devices (PK: 616c15a3-0818-4992-abfd-8a3feda312fa)
21:41:28.657 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet []>
21:41:35.369 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet []>
21:41:38.094 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet []>
21:41:55.367 INFO    nautobot.extras.views :
  Created custom field text (PK: 25d1ce4e-3ca2-4344-90f9-1bc72c6ec3ae)
21:42:05.611 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:42:09.636 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:42:13.009 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:42:14.347 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:42:15.477 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:42:16.575 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:42:17.895 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet []>
21:42:30.451 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:42:31.624 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:42:32.720 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:42:33.912 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet []>
21:42:34.988 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
```

After, in the same circumstances, showing that all requests have the correct refreshed cache:

```
21:56:24.777 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet []>
21:56:32.454 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet []>
21:56:32.568 INFO    nautobot.extras.views :
  Created dynamic group devices (PK: 271fd7f9-2b1d-4870-ac5e-4a6314a07813)
21:56:36.697 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet []>
21:56:38.553 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet []>
21:56:40.648 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet []>
21:56:41.974 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet []>
21:56:43.600 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet []>
21:56:46.661 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet []>
21:56:46.742 INFO    nautobot.extras.views :
  Modified dynamic group devices (PK: 271fd7f9-2b1d-4870-ac5e-4a6314a07813)
21:57:05.170 INFO    nautobot.extras.views :
  Created custom field text (PK: 384e3c85-bcad-4f2b-9e66-a421ed3059d7)
21:57:11.353 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:57:15.536 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:57:17.197 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:57:19.012 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:57:20.568 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:57:24.414 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:57:25.744 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:57:27.009 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:57:28.238 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:57:29.526 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:57:30.804 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:57:32.110 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:57:33.277 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:57:34.597 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
21:57:35.800 WARNING nautobot.extras.views :
  CustomField model cache: <RestrictedQuerySet [<CustomField: text>]>
```

# TODO

- [ ] Unit, Integration Tests


gsnider2195:
Would it make sense to move all of the recently added LRU cache to use django caches to prevent any further issues like this?

glennmatthews:
Added a custom `limited_lru_cache` decorator that adds decorated functions to the extras registry so that they can all be cleared iteratively by the middleware. Also I changed it so the middleware clears caches after a request, instead of before a request, as there's no need for cached data to linger around while the server is sitting idle.

I found a similar bug with the `TreeManager.max_depth` cached_property and multiple uwsgi purposes, so that's fixed now too.

nautobot/core/models/tree_queries.py
```diff
@@ -58,7 +57,7 @@ class TreeManager(TreeManager_, BaseManager.from_queryset(TreeQuerySet)):
     _with_tree_fields = True
     use_in_migrations = True
 
-    @cached_property
+    @limited_lru_cache(max_size=16)
```
glennmatthews, line 60: `@lru_cache` is not a drop in replacement for `@cached_property` (method versus property). 🤦 Reworking this, stand by.

glennmatthews:
> Would it make sense to move all of the recently added LRU cache to use django caches to prevent any further issues like this?

Absolutely worth considering! I'll do some investigation.

nautobot/core/models/tree_queries.py
gsnider2195: cached_property should be fine to leave since it's only caching instance attributes?

gsnider2195: Because you should never have an instance shared by multiple threads?

glennmatthews: Issue isn't with sharing by multiple threads, it's the long life of the cached_property and the fact that signals only clear the cache for the process that triggered the signal, so other workers may retain stale cache data. I think you're right that switching to using a shared Django cache is the preferred approach here, working on that now.

nautobot/core/testing/mixins.py
```diff
@@ -63,10 +63,10 @@ def setUpNautobot(self, client=True, populate_status=False):
             self.client.force_login(self.user)
 
     def tearDown(self):
-        """Clear lru_cache data to avoid leakage of information between test cases when running in parallel."""
-        extras_signals.invalidate_lru_cache(extras_models.CustomField)
-        extras_signals.invalidate_lru_cache(extras_models.ComputedField)
-        extras_signals.invalidate_lru_cache(extras_models.Relationship)
+        """Clear cache data to avoid leakage of information between test cases when running in parallel."""
```
glennmatthews, line 66: This "shouldn't" be needed given that we have separate redis caches for parallel runners, but I found that it was still needed. My working hypothesis is that this is due to the use of `TransactionTestCase` for some of our tests resulting in data being removed from the DB but not triggering the expected signals. More investigation might be good?

glennmatthews: https://code.djangoproject.com/ticket/11505

Maybe we should just do `cache.clear()` after each test rather than these specific signals?

HanlinMiao (approved):

nautobot/extras/tests/test_relationships.py
```diff
@@ -428,19 +428,31 @@ def test_check_if_key_is_graphql_safe(self):
...
+            (manager.get_for_model, 2),
+            (manager.get_for_model_source, 1),
+            (manager.get_for_model_destination, 1),
+        ]
 
-        for manager_method in manager_methods:
+        for manager_method, expected_queries in manager_methods:
             with self.subTest(manager_method=manager_method.__name__):
-                qs1 = manager_method(Location)
+                manager_method(Location)
```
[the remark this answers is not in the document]
glennmatthews: Probably? I think I lacked confidence that the first time running the manager method would have only the expected number of queries (perhaps `ContentType.objects.get_for_model` isn't yet cached within Django?), but we can change it and see if it breaks things.

timizuoebideri1 (approved):

nautobot/extras/tests/test_relationships.py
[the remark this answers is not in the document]
glennmatthews: Yeah, as I thought, this first query may or may not already be cached, so I'm going to leave this alone for now.

joewesch:
@glennmatthews it has been reported that this does not fix the custom field option for dynamic groups after all. I have created issue #5545 with reproducible steps to continue tracking this.

PR #5413, opened 2024-03-10
fix: Update Device lldp detail panel for MAC address port-id

abates:
Some LLDP implementations return an interface's MAC address rather than port name. The LLDP detail tab only compares the configured interface name with the lldp interface. This change adds an interface's MAC address as a data attribute and then checks if the lldp interface matches the configured mac address.

# What's Changed

The LLDP detail tab currently compares the configured interface name with the LLDP interface field. This causes the field to be marked with the `danger` class if the return LLDP interface information uses the MAC address as the port identifier. Some LLDP implementations, such as [lldpd](https://github.com/lldpd/lldpd) default to sending the MAC address as the port ID rather than the port name. The updated code uses both the interface name and interface MAC address in the comparison to determine if the `success` class should be added to the table row.

The updates can be easily tested using the Napalm`mock` driver (thanks @Kircheneer!). The following zip file includes a Napalm mock file and Nautobot objects exported using `nautobot-server dumpdata`. Simply unzip the file in the nautobot source directory (the Napalm driver, in the test data, is set to look for mocks in `/source/napalm_mock`) and load the data using `nautobot-server loaddata testdata.yaml`. Then navigate the the `ls1` device LLDP panel.

[napalm_mock.zip](https://github.com/nautobot/nautobot/files/14550700/napalm_mock.zip)

In addition to fixing the MAC port ID issue, this PR also includes updates to the LLDP tab to replace the jQuery code with vanilla javascript. The code is ES6 compliant and should be supported by any browser released since 2015. If jQuery is preferred, I can revert this portion of the PR.

# Screenshots

Current implementation with incorrect `danger` HTML class:

Screen Shot 2024-03-10 at 8 56 59 AM

Updated to handle MAC port id:

Screen Shot 2024-03-10 at 9 36 58 AM

# TODO

- [ ] ~Unit, Integration Tests~
- [ ] ~Example Plugin Updates (when adding/changing features)~
- [ ] ~Outline Remaining Work, Constraints from Design~


changes/5413.fixed
```diff
@@ -0,0 +1 @@
+Update Device lldp detail panel to handle LLDP neighbors with MAC address port-id
```
glennmatthews, line 1:
```suggestion
Updated Device "LLDP Neighbors" detail panel to handle LLDP neighbors with MAC address as port-id.
```

glennmatthews (approved):
Looks pretty solid to me. Thanks for the non-jQuery rewrite!

nautobot/dcim/templates/dcim/device/lldp_neighbors.html
```diff
@@ -27,7 +27,7 @@
                             <td class="configured_device" data="{{ iface.connected_endpoint.device }}" data-chassis="{{ iface.connected_endpoint.device.virtual_chassis.name }}">
                                 {{ iface.connected_endpoint.device|hyperlinked_object }}
                             </td>
-                            <td class="configured_interface" data="{{ iface.connected_endpoint }}">
+                            <td class="configured_interface" data-interface-name="{{ iface.connected_endpoint }}" data-mac-address="{{ iface.connected_endpoint.mac_address }}">
```
glennmatthews, line 30: Actually, I wonder - should we actually render the configured MAC address as an additional column in the table instead?

abates: I'm happy to add this if you like.

glennmatthews: I'm thinking based on the screenshots that it would makes sense just as a way to show how the "match" was determined for a given interface. Up to you.

abates: I like the idea. Just updated with this:

Screen Shot 2024-03-15 at 10 31 05 AM

glennmatthews (approved):

PR #5417, opened 2024-03-12
Remove hard-coded 255-character UI limit on Text-type CustomField

### LINKED ISSUE #5228: Increase text-type Custom Field length to accommodate Azure Resource IDs
### As ...

P.D. - Plugin Developer

### I want ...

to be able to store an Azure Resource IDs whose lengths are over 255 characters in a custom field, in both 1.6 and 2.x.

For example, this is the generic form of a [subnet's ID in Azure's REST API](https://learn.microsoft.com/en-us/rest/api/virtualnetwork/subnets/get):

``` /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/virtualNetworks/{virtualNetworkName}/subnets/{subnetName}
```

In the above, `resourceGroupName`, `virtualNetworkName`, and `subnetName` are arbitrary strings created by an Azure admin, thus can have arbitrary length. 

There isn't a shorter, globally- or subscription-unique identifier I can find in Azure's API, such as an object UUID or something like AWS' IDs (e.g. `subnet-xxxxxxxxxxxxxxxx` for subnets in AWS).

Further reading on Azure resource naming: https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules

### So that ...

my SSOT plugin may use the resource ID stored in the custom field to reference back to the Azure API object to allow DiffSync to locate the associated records.

my plugin can use our Azure resources whose IDs are larger than 255 characters to lookup objects in Nautobot that originated in Azure, and to use the Azure resource ID on a given object in Nautobot to reference back to the canonical object in Azure's API. Currently, that's at least some Azure Virtual Networks' subnets (mapped to IP Prefixes for us), although I'm still investigating what other Azure resource types have similar concerns.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

Based on the [1.6 documentation of Custom Fields](https://docs.nautobot.com/projects/core/en/v1.6.10/user-guides/custom-fields/) and a cursory look at `nautobot/extras/models/customfields.py` on main branch, I believe the 255 character limit is imposed by the existing schema.

### External Dependencies

_No response_

glennmatthews:
# Closes #5228 
# What's Changed

- Remove hard-coded 255-character limit in the Nautobot UI for Text-type CustomField values.
- Extend `validation_minimum` and `validation_maximum` fields on CustomField to apply (as min/max string lengths) to Text, URL, Markdown field values when defined
- Extend `validation_minimum` and `validation_maximum` fields on CustomField to apply (as min/max string lengths of the JSON dump of the value) for JSON field values. (Not sure how useful this is, but it was easy enough to add)
- Extend `validation_minimum` and `validation_maximum` fields on CustomField to apply (as min/max string lengths) to the choices defined for a Select or Multi-Select CustomField.
- Extend `validation_regex` field on CustomField to also apply to Markdown and JSON field values when defined.
- Add test coverage 
- Refactor existing CustomField tests to be slightly less repetitive.
- Update docs

# Screenshots

No default max-length on Text custom fields in the UI:

image

Values can be arbitrarily long:

image

Configured limits apply:

image

image

image

# TODO

- [x] Unit, Integration Tests


nautobot/extras/models/customfields.py
```diff
@@ -525,12 +526,52 @@ def to_form_field(
...
+                        max_length=self.validation_maximum,
+                    )
+                else:
+                    field = JSONField(
+                        encoder=DjangoJSONEncoder,
+                        required=required,
+                        initial=initial,
+                        min_length=self.validation_minimum,
+                        max_length=self.validation_maximum,
+                    )
```
glennmatthews, line 574: Wonder if for the JSONField we shouldn't apply a min_length/max_length to the UI widget since the actual validation is enforced server-side and may differ depending on whitespace present?

nautobot/extras/models/customfields.py
```diff
@@ -525,12 +526,52 @@ def to_form_field(
...
-                field = forms.CharField(max_length=255, required=required, initial=initial)
+                field = forms.CharField(
+                    required=required,
+                    initial=initial,
+                    min_length=self.validation_minimum,
+                    max_length=self.validation_maximum,
+                )
+            elif self.type == CustomFieldTypeChoices.TYPE_MARKDOWN:
+                field = CommentField(
+                    widget=SmallTextarea,
```
glennmatthews, line 552: Missing `required=required, initial=initial`?

timizuoebideri1 (approved):

PR #5172, opened 2024-01-25
Add collapsible navbar functionality

bradh11:
# Closes 
* Issue 5106: https://github.com/nautobot/nautobot/issues/5106
* Issue 5183 https://github.com/nautobot/nautobot/issues/5183
* Issue 5285 https://github.com/nautobot/nautobot/issues/5285
* Issue 5308 https://github.com/nautobot/nautobot/issues/5308
* Discussion: https://github.com/orgs/nautobot/discussions/3872#discussioncomment-8186580
# What's Changed
This PR brings forth a collapse capable navbar behavior. Users will often want to have more screen real estate for their work or want to be able to leverage a responsive mobile view. 

**This PR brings forth the following capabilities:**
Added:
* **Collapse Capable Side Navbar:** Side Navbar is now able to be expanded and collapsed
* **Expandable Main Content:** The Main Content part of the UI grows as the Side Navbar collapses and shrinks as the Side Navbar expands.
* **Better mobile friendly bottom footer:** Bottom footer currently does not handle responsive media queries properly and will overlap as the browser width gets too small. This update will switch to vertically aligned bottom nav menu items once a certain media query is hit.
* **Automatic Collapsing Navbar for mobile:** The Side Navbar will auto-collapse at a certain media width which is a better experience for mobile devices.

Fixes:
* **Brand Icon mouseover Background:** Fix for mouseover effect on the Brand / Icon (was flashing white background vs being transparent) when in dark mode.

# Screenshots
https://github.com/nautobot/nautobot/assets/18483910/2528a394-863a-45c4-83e6-3b6e81e4f875

https://github.com/nautobot/nautobot/assets/18483910/cec88e23-b2e4-4fbc-84bc-d6f2ff770d02

https://github.com/nautobot/nautobot/assets/18483910/cfbe4228-113a-4dad-a085-2d612d9e1fb6

# TODO
There are several items which are still outstanding in this PR. 

- [X] Need to merge styles to `base.css` and `dark.css`
- [X] Ensure responsive breakpoints are properly configured for the bottom navbar.

- [x] Explanation of Change(s)
- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [x] Attached Screenshots, Payload Example
- [x] Unit, Integration Tests
- [x] Documentation Updates (when adding/changing features)
- [x] Outline Remaining Work, Constraints from Design

PR #5423, opened 2024-03-13
Fix navbar graphql

### LINKED ISSUE #5422: Navbar collapse does not work properly on graphql page
### Environment
* Nautobot version (Docker tag too if applicable): v2.1.7
* Python version: 3.11.
* Database platform, version: postgresql
* Middleware(s): n/a

### Steps to Reproduce
1. browse to graphql editor page from link in the footer
2. click side navbar toggle button to collapse the side navbar

### Expected Behavior
The main content area should expand as the navbar is collapsed

### Observed Behavior
The sidebar collapses, but the main content area does not expand when the navbar is collapsed
Other minor styling issues exist which can be improved with footer overflow scroll bar appearing when it is not needed

This is observed on pages which did not leverage the main wrapper div that most other content leveraged. The following pages are affected.
* /api/
* /graphql
* /admin/

bradh11:
# Closes #5422
# What's Changed

* main-content area for the graphql editor is made collapse capable and expandable depending on the side navbar
# Screenshots
Before:
image

After:
image

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/core/templates/graphene/graphiql.html
+++ b/nautobot/core/templates/graphene/graphiql.html
@@ -10,53 +10,34 @@ add "&raw" to the end of the URL within a browser.
 <!DOCTYPE html>
 <html>
 <head>
+  {% include 'inc/media.html' %}
+  {% block extra_styles %}{% endblock %}
   <!-- Nautobot template requirements -->
   <title>{% block title %}GraphiQL{% endblock %} - {{ settings.BRANDING_TITLE }}</title>
-  <link rel="stylesheet"
-        href="{% static 'bootstrap-3.4.1-dist/css/bootstrap.min.css' %}"
-        onerror="window.location='{% url 'media_failure' %}?filename=bootstrap-3.4.1-dist/css/bootstrap.min.css'">
-  <link rel="stylesheet"
-        href="{% static 'materialdesignicons-6.5.95/css/materialdesignicons.min.css' %}"
-        onerror="window.location='{% url 'media_failure' %}?filename=materialdesignicons-6.5.95/css/materialdesignicons.min.css'">
-  <link rel="stylesheet"
-        href="{% static 'jquery-ui-1.13.1/jquery-ui.min.css' %}"
-        onerror="window.location='{% url 'media_failure' %}?filename=jquery-ui-1.13.1/jquery-ui.min.css'">
-  <link rel="stylesheet"
-        href="{% static 'select2-4.0.13/select2.min.css' %}"
-        onerror="window.location='{% url 'media_failure' %}?filename=select2-4.0.13/select2.min.css'">
-  <link rel="stylesheet"
-        href="{% static 'select2-bootstrap-0.1.0-beta.10/select2-bootstrap.min.css' %}"
-        onerror="window.location='{% url 'media_failure' %}?filename=select2-bootstrap-0.1.0-beta.10/select2-bootstrap.min.css'">
-  <link rel="stylesheet"
-        href="{% static 'flatpickr-4.6.9/themes/light.min.css' %}"
-        onerror="window.location='{% url 'media_failure' %}?filename=flatpickr-4.6.9/themes/light.min.css'">
-  <link rel="stylesheet" id="dark-theme"
-        href="{% versioned_static 'css/dark.css' %}"
-        onerror="window.location='{% url 'media_failure' %}?filename=css/dark.css'" disabled="disabled" />
-  <link rel="stylesheet"
-        href="{% versioned_static 'css/base.css' %}"
-        onerror="window.location='{% url 'media_failure' %}?filename=css/base.css'">
-  <link rel="apple-touch-icon" sizes="180x180" href="{% custom_branding_or_static 'icon_180' 'img/nautobot_icon_180x180.png' %}">
-  <link rel="icon" type="image/png" sizes="32x32" href="{% custom_branding_or_static 'icon_32' 'img/nautobot_icon_32x32.png' %}">
-  <link rel="icon" type="image/png" sizes="16x16" href="{% custom_branding_or_static 'icon_16' 'img/nautobot_icon_16x16.png' %}">
-  <link rel="icon" type="image/png" sizes="192x192" href="{% custom_branding_or_static 'icon_192' 'img/nautobot_icon_192x192.png' %}">
-  <link rel="mask-icon" type="image/png" color="#0097ff" href="{% custom_branding_or_static 'icon_mask' 'img/nautobot_icon_monochrome.png' %}">
-  <link rel="shortcut icon" href="{% custom_branding_or_static 'favicon' 'img/favicon.ico' %}">
-  <meta name="msapplication-TileColor" content="#2d89ef">
-  <meta name="theme-color" content="#ffffff">
-  <meta charset="UTF-8">
-  <meta name="viewport" content="initial-scale=1, maximum-scale=1, user-scalable=no, width=device-width">
+
   <!-- GraphiQL template requirements -->
   <style>
-    html, body, #editor {
-      height: 100%;
+    #editor {
       margin: 0;
       width: 100%;
+      height: calc(100vh - 95px - 3rem);
     }
     .toolbar .dropdown-menu > li > a {
         clear: left;
         margin-right: 120px;  /* leave room for the "Save Changes" button if present */
     }
+    body {
+      padding-top: 56px;
+    }
+    .graphiql-container .doc-explorer-title-bar {
+      box-sizing: content-box;
+  }
+    .graphiql-container .history-title-bar {
+      box-sizing: content-box;
+  }
+  .graphiql-container .docExplorerShow {
+    white-space: nowrap;
+  }
   </style>
   <!-- As Nautobot may be run without internet access, we source these files locally rather than from an online CDN -->
   <link rel="stylesheet"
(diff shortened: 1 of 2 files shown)


nautobot/core/templates/graphene/graphiql.html
```diff
@@ -10,53 +10,34 @@
...
-  <link rel="apple-touch-icon" sizes="180x180" href="{% custom_branding_or_static 'icon_180' 'img/nautobot_icon_180x180.png' %}">
-  <link rel="icon" type="image/png" sizes="32x32" href="{% custom_branding_or_static 'icon_32' 'img/nautobot_icon_32x32.png' %}">
-  <link rel="icon" type="image/png" sizes="16x16" href="{% custom_branding_or_static 'icon_16' 'img/nautobot_icon_16x16.png' %}">
-  <link rel="icon" type="image/png" sizes="192x192" href="{% custom_branding_or_static 'icon_192' 'img/nautobot_icon_192x192.png' %}">
-  <link rel="mask-icon" type="image/png" color="#0097ff" href="{% custom_branding_or_static 'icon_mask' 'img/nautobot_icon_monochrome.png' %}">
-  <link rel="shortcut icon" href="{% custom_branding_or_static 'favicon' 'img/favicon.ico' %}">
-  <meta name="msapplication-TileColor" content="#2d89ef">
-  <meta name="theme-color" content="#ffffff">
-  <meta charset="UTF-8">
-  <meta name="viewport" content="initial-scale=1, maximum-scale=1, user-scalable=no, width=device-width">
```
bradh11, line 48: All of these were redundant so I cleaned them up and replace with the `media.html` include.

nautobot/core/templates/graphene/graphiql.html
```diff
@@ -10,53 +10,34 @@
...
       width: 100%;
+      height: calc(100vh - 95px - 3rem);
     }
     .toolbar .dropdown-menu > li > a {
         clear: left;
         margin-right: 120px;  /* leave room for the "Save Changes" button if present */
     }
+    body {
+      padding-top: 56px;
+    }
```
glennmatthews, line 31: Wondering if we can remove this? Since we're not needing to reserve room for the nav above the GraphiQL UI any more?

image

nautobot/project-static/css/base.css
```diff
@@ -42,7 +42,7 @@ body {
     transition-duration: var(--navbar-transition-duration);
     transition-timing-function: ease-in-out;
     transition-property: margin-left, width;
-    overflow: visible;
+    overflow: hidden;
```
timizuoebideri1, line 45: Maybe for now make overflow scroll, till the footer responsiveness is fixed 

```suggestion
    overflow-y: hidden;
    overflow-x: scroll;
```

bradh11: footer responsiveness was fixed in #5172

bradh11: so we should be ok to change to hidden for the footer.

timizuoebideri1: It does not look fixed

Screenshot 2024-03-13 at 4 10 35 PM

bradh11: Referring to the Help button?  If so, I can look into that.  The scrollbar does not fix that however.  It is an issue regardless.  

Having the scrollbar causes a worse experience in my opinion because it causes the screen to jump up and down when scroll the window from left to right.

bradh11: I just adjusted the css on the middle column of the footer to address your comment.  The help button no longer hangs over the right edge on medium width

nautobot/core/templates/graphene/graphiql.html
bradh11: Absolutely think this makes sense.  Just addressed in commit: https://github.com/nautobot/nautobot/pull/5423/commits/06ec225efa2825d2e712b307bbd05bcd36b9e8ba

changes/5423.fixed
```diff
@@ -0,0 +1 @@
+Fixed collapsable navbar for graphql page
```
glennmatthews, line 1:
```suggestion
Fixed collapsable navbar for GraphiQL page.
```

nautobot/project-static/css/base.css
glennmatthews, line 45: This change hides the extra rows when the footer wraps to multiple lines on mobile devices or very narrow windows - I don't think it's desirable.

nautobot/core/templates/inc/footer.html
```diff
@@ -27,7 +27,7 @@
                         </p>
                     {% endif %}
                 </div>
-                <div class="col-xl-4 col-lg-4 col-md-4 col-sm-3 col-xs-12 text-center footer-col-2">
+                <div class="col-xl-4 col-lg-4 col-md-3 col-sm-3 col-xs-12 text-center footer-col-2">
```
glennmatthews, line 30: The three segments of the footer should always sum to 12 columns, so if you're decreasing this one you should increase another one accordingly. I think they all need to be adjusted a bit to avoid overflow. Try this set of classes and see what you think as the window size changes:

```
class="col-xl-4 col-lg-4 col-md-3 col-sm-6 col-xs-12 text-left footer-col-1 no-wrap"
class="col-xl-4 col-lg-3 col-md-3 col-sm-6 col-xs-12 text-center footer-col-2"
class="col-xl-4 col-lg-5 col-md-6 col-sm-12 col-xs-12 text-right footer-col-3 noprint"
```

(If making this change, probably also need some changes to the media queries above so that columns are aligned appropriately at the "sm" breakpoint - currently here they look like this:)

image

bradh11: Ok.  Will fine tune it.  I agree this is the desired state with bootstrap rows/columns.  That said, when I have it as interval of 12 the 3rd item always seems to overflow behind the row length.  I believe it is due to a nowrap that impacts the medium width views.  I don’t want to fix it by removing that however because then the menu buttons on the bottom right begin to wrap at larger width scenarios which is undesirable.  

I’ll get the magic combination for medium width (md) column nailed down as part of this PR.

glennmatthews:
Would you mind addressing the DRF API views as well (https://github.com/nautobot/nautobot/issues/5422#issuecomment-1995766507)?

bradh11:
> Would you mind addressing the DRF API views as well ([#5422 (comment)](https://github.com/nautobot/nautobot/issues/5422#issuecomment-1995766507))?

Yes.  This is the /api/ endpoint.  (Takes you to Django rest framework “drf”. I added it to the TODO section of the PR and will most definitely address it.

changes/5423.fixed
```diff
@@ -0,0 +1,3 @@
+Fixed collapsable navbar for GraphiQL page `/graphql`.
+Fixed collapsable nabar for Admin page `/admin`.
```
glennmatthews, line 2:
```suggestion
Fixed collapsable navbar for Admin page `/admin`.
```

nautobot/core/templates/inc/footer.html
glennmatthews: Did my proposed changes for the `lg`, `md`, and `sm` column widths above end up not working out or being unnecessary?

timizuoebideri1 (approved):

bradh11:
@glennmatthews - This is ready to review again with the footer fixes along with the `/api/`, `/admin`, and `/graphql/` pages.  I was also able to remove a lot of redundant code in some of the templates along with fixing a few errors in the javascript imports that was happening under the hood for the admin page

nautobot/core/templates/admin/base.html
```diff
@@ -95,10 +47,10 @@
...
 </head>
 <body {% if is_popup %}style="padding-top: 0px;"{% endif %}>
-{% if not is_popup %}
-    {% include 'inc/nav_menu.html' %}
-{% endif %}
-<div class="container-fluid wrapper" {% if is_popup %}style="padding-bottom: 0px;"{% endif %}>
+
+{% include 'inc/nav_menu.html' %}
+
+<div class="container-fluid wrapper" id="main-content" {% if is_popup %}style="padding-bottom: 0px;"{% endif %}>
```
glennmatthews, line 53: The admin UI uses pop-up windows for some sub-forms within other forms. Do we still need the `is_popup` conditional here? I wouldn't think we'd want the nav menu to appear in those pop-ups.

nautobot/core/templates/admin/base.html
```diff
@@ -148,47 +100,18 @@ <h1>{% render_with_template_if_exist title|lower|add:"/admin_app_name.html" titl
...
-                        <i class="mdi mdi-graphql text-primary"></i> <a href="{% url 'graphql' %}">GraphQL</a> &middot;
-                        <i class="mdi mdi-xml text-primary"></i> <a href="{{ settings.BRANDING_URLS.code }}">Code</a> &middot;
-                        <i class="mdi mdi-lifebuoy text-primary"></i> <a href="{{ settings.BRANDING_URLS.help }}">Help</a>
-                    </p>
-                </div>
-            </div>
-        </div>
-    </footer>
-{% endif %}
+{% include 'inc/footer.html' %}
```
glennmatthews, line 110: Same question here about `is_popup` - I don't think we need the footer in popups.

nautobot/core/templates/admin/base.html
glennmatthews: Yeah, I think we should add that conditional back in:

image

nautobot/project-static/css/base.css
```diff
@@ -44,6 +45,7 @@ body {
     transition-property: margin-left, width;
     overflow: visible;
     margin: 0;
+    padding-right: 70px;
```
glennmatthews, line 48: This seems odd/unnecessary?

image

glennmatthews: I still think adjusting the sizes of this column (https://github.com/nautobot/nautobot/pull/5423/files#r1523647247) might be the more appropriate fix.

nautobot/core/templates/inc/footer.html
bradh11: I’m pretty sure it helped with right hand buttons overflowing, but had a negative impact to the other two columns crashing into each other.  I’ll try again and validate.

nautobot/project-static/css/base.css
bradh11: See my inline comment.  I’ll try again, but the other options seemed to have different negative outcomes and seemed to move the problem.

nautobot/core/templates/admin/base.html
```diff
@@ -8,55 +8,9 @@
...
-            onerror="window.location='{% url 'media_failure' %}?filename=js/forms.js'"></script>
-
-    <meta name="msapplication-TileColor" content="#2d89ef">
-    <meta name="theme-color" content="#ffffff">
-    <meta charset="UTF-8">
-    <meta name="viewport" content="initial-scale=1, maximum-scale=1, user-scalable=no, width=device-width">
-
+    {% if not is_popup %}
+        {% include 'inc/media.html' %}
+    {% endif %}
```
glennmatthews, line 13: Don't we need at least some of these includes for the popup case? I'll pull down this branch when I get a chance and see what specifically is needed for the popup case and fix it.

nautobot/core/templates/admin/base.html
glennmatthews: Nav menu conditional still needs to be re-added.

nautobot/core/templates/inc/footer.html
```diff
@@ -3,8 +3,8 @@
 {% load plugins %}
 
 <style>
-    /* Styles for viewports narrower than 1007px */
-    @media (max-width: 1007px) {
+    /* Styles for viewports narrower than 1008px */
+    @media (max-width: 1232px) {
```
glennmatthews, line 7:
```suggestion
    /* Styles for viewports narrower than 1232px, i.e. bootstrap 'xs' and 'sm' layouts */
    @media (max-width: 1232px) {
```

glennmatthews (approved):

PR #5424, opened 2024-03-13
Added support for object list view buttons in TemplateExtensions

lampwins:
Adds `list_buttons()` to `TemplateExtension` which operates in the same way as `buttons()` but renders to the model's list view.

# Screenshots
Screenshot 2024-03-13 at 3 12 06 PM
Screenshot 2024-03-13 at 3 12 57 PM

# TODO

- [x] Unit, Integration Tests


glennmatthews (commented):
Needed in 2.1.x or should this be targeted at `next` instead?

nautobot/docs/development/apps/api/ui-extensions/object-views.md
```diff
@@ -2,19 +2,20 @@
...
+* `buttons()` - Add buttons to the top of the object detail page
+* `list_buttons()` - Add buttons to the object list page. This works in the same way as `buttons()` for the object detail page.
+* `detail_tabs()` - Add extra tabs to the end of the list of tabs within the object detail page tabs navigation
 
 Additionally, a `render()` method is available for convenience. This method accepts the name of a template to render, and any additional context data you want to pass. Its use is optional, however.
 
 When a TemplateExtension is instantiated, context data is assigned to `self.context`. Available data include:
 
-* `object` - The object being viewed
+* `object` - The object being viewed (note that this will be `None` when accessed in the context of `buttons()`)
```
glennmatthews, line 18: Should this be:

```suggestion
* `object` - The object being viewed (note that this will be `None` when accessed in the context of `list_buttons()`)
```

nautobot/docs/development/apps/api/ui-extensions/object-views.md
```diff
@@ -2,19 +2,20 @@
...
-* `buttons()` - Add buttons to the top of the page
-* `detail_tabs()` - Add extra tabs to the end of the list of tabs within the page tabs navigation
+* `left_page()` - Inject content on the left side of the object detail page
+* `right_page()` - Inject content on the right side of the object detail page
+* `full_width_page()` - Inject content across the entire bottom of the object detail page
+* `buttons()` - Add buttons to the top of the object detail page
+* `list_buttons()` - Add buttons to the object list page. This works in the same way as `buttons()` for the object detail page.
+* `detail_tabs()` - Add extra tabs to the end of the list of tabs within the object detail page tabs navigation
 
 Additionally, a `render()` method is available for convenience. This method accepts the name of a template to render, and any additional context data you want to pass. Its use is optional, however.
```
glennmatthews, line 14:
```suggestion
* `buttons()` - Add buttons to the top of the object detail page
* `list_buttons()` - Add buttons to the object list page. This works in the same way as `buttons()` for the object detail page.
* `detail_tabs()` - Add extra tabs to the end of the list of tabs within the object detail page tabs navigation

+++ 2.1.8
    Support for the `list_buttons()` method was added.

Additionally, a `render()` method is available for convenience. This method accepts the name of a template to render, and any additional context data you want to pass. Its use is optional, however.
```

nautobot/extras/templatetags/plugins.py
```diff
@@ -56,10 +56,14 @@ def _get_registered_content(obj, method, template_context, return_html=True):
 
 
 @register.simple_tag(takes_context=True)
-def plugin_buttons(context, obj):
+def plugin_buttons(context, obj, view="detail"):
     """
     Render all buttons registered by plugins
     """
+    if view == "list":
+        # The object_list.html template passes content_type.model_class as obj
```
glennmatthews, line 64: Is this true both for old-style ObjectListViews and also NautobotUIViewSet?

glennmatthews: NM, I see you're doing this in the shared generic template not the view code.

glennmatthews: It is a bit irrelevant though since the TemplateExtension only applies to a single model class by design?

changes/5424.added
```diff
@@ -0,0 +1 @@
+Apps may now register button content to be injected into object list views via the new `list_buttons()` method on `TemplateExtension` classes.
```
glennmatthews, line 1:
```suggestion
Added `TemplateExtension.list_buttons()` API, allowing apps to register button content to be injected into object list views.
```

glennmatthews (approved):

nautobot/extras/templatetags/plugins.py
lampwins: Yes because this is driven by the template which is the same in both cases (assuming normal implementation, which is also assumed in all the other TemplateExtension use cases).

lampwins: Somewhat yes. I was mostly trying to not reinvent the wheel. There's a few other things we should add support for in the future for list views, in line with our other UX API improvements. So I think we will be able to refactor TemplateExtensions then.

PR #5427, opened 2024-03-13
DeviceType YAML import: map unknown 'type' values to 'other'

### LINKED ISSUE #5412: DeviceType import handle unknown component enums
Update the import logic to default any device component template CharField with defined choices to an `other` choice if we encounter a choice we don't recognize.

Add a debug log when we encounter such cases.

glennmatthews:
# Closes #5412 
# What's Changed

- On DeviceType YAML import, specifically the import of port templates within a DeviceType, if an unknown value of the `type` enum is encountered, log a debug message and replace the value with the appropriate `TYPE_OTHER` enum.
- Fix a couple of minor bugs I noticed in manual testing of the DeviceTypeImportView.
- Add test case

# Screenshots

image

image

# TODO

- [x] Unit, Integration Tests


nautobot/dcim/forms.py
```diff
@@ -1350,6 +1358,34 @@ def clean_device_type(self):
 
         return data
 
+    def clean_type(self):
+        data = self.cleaned_data["type"]
+
+        try:
+            choices = self.Meta.model._meta.get_field("type").choices
+        except FieldDoesNotExist:
+            # Do we have a `DeviceComponent` subclass that doesn't define a `type` field?
```
timizuoebideri1, line 1367: `InventoryItem`

glennmatthews: Actually this should be referring to a `DeviceComponentTemplate` (we don't yet have an `InventoryItemTemplate` class) but this is also true for `DeviceBayTemplate` not having a `type`. I'll update the comment.

glennmatthews: Actually I'll remove the try/except here as `clean_type()` should only be called automatically if there is a `type` field on the form.

PR #5429, opened 2024-03-14
Add Python 3.12 support

glennmatthews:
# Closes #n/a
# What's Changed

- Add support for python 3.12, make it the default in docker images.
- Remove `mkdocs-include-markdown-plugin` development dependency as it's no longer used and would require a major version bump to reach a version that supports Python 3.12.
- CI and documentation updates.

# Screenshots

image

 Note `"python-version": "3.12.2"`

# TODO

- n/a Unit, Integration Tests


HanlinMiao (approved):

HanlinMiao:
Oops looks like the unit tests are failing?

glennmatthews:
Huh. Wonder why that didn't happen when I ran tests locally with 3.12. Investigating.

glennmatthews:
Looks like it's because CI is currently using latest `poetry` version (1.8.2 - because we're using an older version of https://github.com/networktocode/gh-action-setup-poetry-environment that didn't pin the poetry version) whereas our local Docker build is still explicitly using `poetry 1.5.1`. The newer version of `poetry` doesn't pull in `setuptools` as a dependency. I'll figure out what we need to do here to be consistent, but also we need to pursue 3.12 compatibility with the `netutils` maintainers since that's what's throwing the error.

glennmatthews:
Blocked by networktocode/netutils#280. Working on that now.

glennmatthews:
Also, the tests pass in Docker because the docker build installs `nautobot[all]`, which brings in `napalm`, which brings in `setuptools`, which vendors the `distutils` package.

The CI tests just install `nautobot`, so `napalm` and hence `setuptools` do not get installed.

glennmatthews:
The new version of `gh-action-setup-poetry-environment` has a breaking change in that it now defaults to `poetry install --only dev` which means most of our dependencies don't get installed in the CI after I updated the action version. 😭 I'll look into fixing this tomorrow.

glennmatthews:
Blocked until a new release of `netutils` is available.

PR #5428, opened 2024-03-14
Add celery beat health check

glennmatthews:
# Closes #1102
# What's Changed

- Add logic to our existing custom Celery Beat scheduler subclass such that each time it wakes up (every 5 seconds or less) it will touch a file (configurable via `settings.CELERY_BEAT_HEARTBEAT_FILE` or environment `NAUTOBOT_CELERY_BEAT_HEARTBEAT_FILE`, defaulting to `/tmp/nautobot_celery_beat_heartbeat`) as a proof of life.
- Add a custom healthcheck in `docker-compose.yml` that checks this file's last-modified time.
- Documentation, template config updates, etc.

# Screenshots

Health check successful:

```
❯ docker ps
CONTAINER ID   IMAGE                               COMMAND                  CREATED         STATUS                    PORTS                                             NAMES
c05da4131ed5   local/nautobot-dev:local-py3.8      "nautobot-server cel…"   About a minute ago   Up 59 seconds (healthy)            8080/tcp                                          nautobot-celery_beat-1
```

After patching the `tick()` method so that it hangs and does not return as it should, and restarting the container (since celery beat container doesn't auto-restart on code changes) we can see that the health check begins to fail:

```
❯ docker ps
CONTAINER ID   IMAGE                               COMMAND                  CREATED         STATUS                      PORTS                                             NAMES
b8ac67ab615d   local/nautobot-dev:local-py3.8      "nautobot-server cel…"   6 minutes ago   Up 25 seconds (unhealthy)   8080/tcp                                          nautobot-celery_beat-1
```

# TODO

- [x] Explanation of Change(s)
- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [x] Attached Screenshots, Payload Example
- n/a Unit, Integration Tests
- [x] Documentation Updates (when adding/changing features)
- n/a example Plugin Updates (when adding/changing features)
- [x] Outline Remaining Work, Constraints from Design

PR #5434, opened 2024-03-18
Fixed health check for beat container in `docker-compose.yml`.

Kircheneer:
This gets rid of the error message
`invalid interpolation format for services.celery_beat.healthcheck.test.[]: "[ $(find /tmp/nautobot_celery_beat_heartbeat -mmin -0.1 | wc -l) -eq 1 ] || false". You may need to escape any $ with another $` introduced by f9d84decdaf9c96140a2742082ddee6fc0474691.

### DIFF
--- a/changes/5434.fixed
+++ b/changes/5434.fixed
@@ -1,0 +1,1 @@
+Fixed health check for beat container in `docker-compose.yml`.
--- a/changes/5434.fixed
+++ b/changes/5434.fixed
@@ -1,1 +1,1 @@
-Fixed health check for beat container in `docker-compose.yml`.
+Fixed health check for beat container in `docker-compose.yml` under `docker-compose` v1.x.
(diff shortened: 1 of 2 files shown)


glennmatthews:
Sorry, where does that error message occur?

glennmatthews:
I'll pull it down and try it locally - generally we're using docker compose v2.x at this point so I want to make sure it works with that as well.

```
❯ docker-compose -v 
Docker Compose version v2.13.0
❯ docker compose version  
Docker Compose version v2.13.0
```

If this works for both, we'll want to double-commit to `ltm-1.6` as well. Thanks.

glennmatthews (approved):
Looks valid in docker compose v2.x as well.

changes/5434.fixed
```diff
@@ -0,0 +1 @@
+Fixed health check for beat container in `docker-compose.yml` under `docker-compose` v1.x.
```
gsnider2195, line 1: rename to `5434.housekeeping`?

glennmatthews: Possibly... really the change to add the healthcheck for #5428 should have been under `housekeeping` as well.  Let's merge this and I'll fix it up in the release branch.

PR #5435, opened 2024-03-18
Improve invoke unittest

snaselj:
# Closes Nan

## What's Changed

- Added `--pattern` argument to `invoke unittest`.
- Added `--parallel-workers` argument to `invoke unittest`.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -693,24 +693,27 @@ def check_schema(context, api_version=None):
         "cache_test_fixtures": "Save test database to a json fixture file to re-use on subsequent tests.",
         "keepdb": "Save and re-use test database between test runs for faster re-testing.",
         "label": "Specify a directory or module to test instead of running all Nautobot tests.",
+        "pattern": "Only run tests which match the given substring. Can be used multiple times.",
         "failfast": "Fail as soon as a single test fails don't run the entire test suite.",
         "buffer": "Discard output from passing tests.",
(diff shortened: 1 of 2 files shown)


glennmatthews (commented):
Anything that should be documented in `nautobot/docs/development/core/getting-started.md` as a part of this PR?

Can you provide an example(s) of `invoke unittest --pattern`? Seems useful.

When would you want to use `--parallel-workers` instead of just `--parallel`?

snaselj:
> Anything that should be documented in `nautobot/docs/development/core/getting-started.md` as a part of this PR?
> 
> Can you provide an example(s) of `invoke unittest --pattern`? Seems useful.
> 
> When would you want to use `--parallel-workers` instead of just `--parallel`?

Added relevant examples to getting started.

glennmatthews (approved):

timizuoebideri1 (approved):

tasks.py
```diff
@@ -552,7 +552,7 @@ def build_example_app_docs(context):
         print_command(local_command)
         context.run(local_command, pty=True)
     else:
-        docker_command = f"run --workdir='/source/examples/example_app' --entrypoint '{command}' nautobot"
+        docker_command = f"run --rm --workdir='/source/examples/example_app' --entrypoint '{command}' nautobot"
```
glennmatthews, line 555: Thanks for this! Would be great to backport to `develop` as well, but not urgent.

PR #5464, opened 2024-03-21
View authentication and permission fixes

glennmatthews:
# Closes GHSA-m732-wvh2-7cq4
# What's Changed

## Security fixes

- The endpoint `/extras/job-results/<uuid:pk>/log-table/` now requires user authentication to access, and also enforces appropriate JobResult `view` permissions now.
- The endpoint `/extras/secrets/provider/<str:provider_slug>/form/` now requires user authentication to access.
- Removed the URL endpoints `/api/users/users/my-profile/`, `/api/users/users/session/`, `/api/users/tokens/authenticate/`, and `/api/users/tokens/logout/` as they are unused at this time.
- The `/api/graphql/` REST API endpoint now requires user authentication to execute GraphQL queries, even when `EXEMPT_VIEW_PERMISSIONS` is configured.
- The `nautobot.apps.api.APIRootView` class now enforces user authentication by default. As a consequence, viewing the browsable REST API root endpoints (e.g. `/api/`, `/api/circuits/`, `/api/dcim/`, etc.) now requires user authentication.
- The endpoints `/dcim/racks/<uuid>/dynamic-groups/`, `/dcim/devices/<uuid>/dynamic-groups/`, `/ipam/prefixes/<uuid>/dynamic-groups/`, `/ipam/ip-addresses/<uuid>/dynamic-groups/`, `/virtualization/clusters/<uuid>/dynamic-groups/`, and `/virtualization/virtual-machines/<uuid>/dynamic-groups/` now require user authentication, even when `EXEMPT_VIEW_PERMISSIONS` is configured.

## Additional permissions enforcement

- The `/api/dcim/connected-device/?peer_device=...&?peer_interface=...` REST API endpoint now requires the requesting user to have `view` permission on the specified `peer_interface`.
- The endpoints `/extras/git-repositories/<uuid:pk>/sync/` and `/extras/git-repositories/<uuid:pk>/dry-run/` have corrected permissions now - a user who has `change` permissions for a subset of Git repositories is no longer permitted to sync or dry-run other repositories for which they lack the appropriate permissions.
- Viewing Notes associated to a given model now requires `extras.view_note` permissions for each note to display (this is consistent with how change log views for a model enforce `extras.view_objectchange` permissions).

## Bug fixes, housekeeping, documentation

- Fixed a 500 error when accessing any of the `/dcim/<port-type>/<uuid>/connect/<termination_b_type>/` view endpoints with an invalid/nonexistent `termination_b_type` string.
- The `nautobot.apps.api.OrderedDefaultRouter` class now accepts a `view_name` and/or `view_description` as parameters. Specifying these parameters is to be preferred over defining a custom `APIRootView` subclass when defining App API URLs.
- Updated example views in the App developer documentation to include `ObjectPermissionRequiredMixin` or `LoginRequiredMixin` as appropriate best practices.
- Updated custom views in the `example_plugin` to include `LoginRequiredMixin` as a best practice.
- Added integration test case that enumerates all registered URL patterns and verifies that they return appropriate response codes when accessed by an unauthenticated user.

# Screenshots

# TODO

- [x] Unit, Integration Tests


nautobot/core/tests/integration/test_view_authentication.py
```diff
@@ -0,0 +1,71 @@
...
+                    response = self.client.post(url, follow=True)
+
+                # Is a view that *should* be open to unauthenticated users?
+                if url in [
+                    "/admin/login/",
+                    "/api/plugins/example-plugin/webhook/",
+                    "/api/redoc/",
+                    "/api/swagger/",
+                    "/api/swagger.json",
+                    "/api/swagger.yaml",
```
glennmatthews, line 34: This is at the very least inconsistent - we don't allow access to `/api/docs/` for unauthenticated users, but we do permit `/api/redoc/`? And we let unauthenticated users download the schema? I think we should probably block unauthenticated access to those endpoints - thoughts?

changes/xxxx.added
```diff
@@ -0,0 +1 @@
+Added `nautobot.apps.utils.get_url_for_url_pattern` and `nautobot.apps.utils.get_url_patterns` lookup functions.
```
gsnider2195, line 1: 14 change log entries in 1 PR 😩

examples/example_plugin/example_plugin/views.py
```diff
@@ -46,12 +47,12 @@ class DeviceDetailPluginTabTwoView(views.ObjectView):
     template_name = "example_plugin/tab_device_detail_2.html"
 
 
-class ExamplePluginHomeView(View):
+class ExamplePluginHomeView(LoginRequiredMixin, View):
```
gsnider2195, line 50: Would it make sense to just create a generic nautobot view that inherits from django.views.generic.View and LoginRequiredMixin?

nautobot/core/tests/integration/test_view_authentication.py
HanlinMiao: I feel like disabling all endpoints mentioned above for unauthenticated users should be the way to go. Or could we allow `/api/docs` for unauthenticated users? Is there any sensitive information on that endpoint?

examples/example_plugin/example_plugin/views.py
HanlinMiao: There are a lot of instances of those combinations in this PR alone, I think it makes sense. Should be non-blocking tho.

HanlinMiao (approved):

nautobot/core/utils/lookup.py
```diff
@@ -218,3 +220,122 @@ def get_created_and_last_updated_usernames_for_model(instance):
...
+
+
+def get_url_patterns(urlconf=None, patterns_list=None, base_path="/"):
+    """
+    Recursively yield a list of registered URL patterns.
+
+    Args:
+        urlconf (URLConf): Python module such as `nautobot.core.urls`.
+        patterns_list (list): Used in recursion
+        base_path (str): String to prepend to all URL patterns yielded.
```
gsnider2195, line 232: Add default values for these

changes/xxxx.added
glennmatthews: Yeah 😢 This one scope crept, but I think everything in it is still relevant/necessary.

examples/example_plugin/example_plugin/views.py
glennmatthews: I like the idea, will do.

nautobot/core/tests/integration/test_view_authentication.py
glennmatthews: We made `/api/docs/` require authentication in #4068 and #4865. I think we should add the restriction to these endpoints rather than removing it from `/api/docs/`

glennmatthews: Agreed this is the plan.

nautobot/core/utils/lookup.py
gsnider2195: Can we add the derived defaults here? I'm thinking the default of `settings.ROOT_URLCONF` if the `urlconf` kwarg is `None`

glennmatthews: Ah sure, good thought.

nautobot/core/utils/lookup.py
```diff
@@ -218,3 +220,125 @@ def get_created_and_last_updated_usernames_for_model(instance):
...
+
+    for item in patterns_list:
+        if isinstance(item, URLPattern):
+            yield base_path + str(item.pattern)
+        elif isinstance(item, URLResolver):
+            # Recurse!
+            yield from get_url_patterns(urlconf, item.url_patterns, base_path + str(item.pattern))
+
+
+def get_url_for_url_pattern(url_pattern):
```
gsnider2195, line 300: Should this be `get_valid_url_for_url_pattern` or `get_fake_url_for_url_pattern` maybe?

glennmatthews: I'd be fine with either.

nautobot/core/tests/integration/test_view_authentication.py
```diff
@@ -0,0 +1,67 @@
...
+from nautobot.core.testing import TestCase
+from nautobot.core.utils.lookup import get_url_for_url_pattern, get_url_patterns
+
+
+@tag("integration")
+class AuthenticationEnforcedTestCase(TestCase):
+    r"""
+    Test that all\* registered views require authentication to access.
+
+    \* with a very small number of known exceptions such as login and logout views.
```
gsnider2195, line 12: Why the raw docstring with the escaped asterisks?

nautobot/core/tests/integration/test_view_authentication.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+
+    def test_all_views_require_authentication(self):
```
gsnider2195, line 15: This would be a good test to expose to apps. Not blocking but should we write a generic test for this?

nautobot/users/api/views.py
```diff
@@ -101,21 +53,6 @@ def authentication_classes(self):
...
-        serializer.is_valid(raise_exception=True)
-        user = serializer.validated_data["user"]
-        login(request, user=user)
-        return Response(status=200)
-
-    @action(methods=["GET"], detail=False)
-    def logout(self, request):
-        logout(request)
-        return Response(status=200)
-
```
gsnider2195, line 118: Were these only for the alpha UI?

nautobot/core/tests/integration/test_view_authentication.py
glennmatthews: Thinking that at some point we'll treat docstrings as Markdown (as we already do for stuff in the apps namespace. Didn't want the asterisks to map to bold.

nautobot/core/tests/integration/test_view_authentication.py
glennmatthews: I do plan/want to add a generic test to the cookiecutter that takes in the app's urlconf and iterates/validates its url patterns, yeah. I like the idea of implementing the bulk of it in core rather than in the cookiecutter but that'll take some thinking about how to do it.

nautobot/users/api/views.py
glennmatthews: Yup!

nautobot/core/api/routers.py
```diff
@@ -20,7 +33,18 @@ def get_api_root_view(self, api_urls=None):
...
 
+        if issubclass(self.APIRootView, AuthenticatedAPIRootView):
+            return self.APIRootView.as_view(
+                api_root_dict=api_root_dict, name=self.view_name, description=self.view_description
+            )
+        # Fallback for the established practice of overriding self.APIRootView with a custom class
+        logger.warning(
+            "Something has changed an OrderedDefaultRouter's APIRootView attribute to a custom class. "
+            "Please verify that class %s implements appropriate authentication controls.",
+            self.APIRootView.__name__,
```
gsnider2195, line 48: Should this include the full class path?

nautobot/core/api/utils.py
```diff
@@ -178,6 +178,10 @@ def get_view_name(view, suffix=None):
 
     else:
         # Replicate DRF's built-in behavior.
+        name = getattr(view, "name", None)
+        if name is not None:
+            return view.name
```
gsnider2195, line 183: Is there a reason this is not:

```py
if getattr(view, "name", None) is not None:
    return view.name
```

OR

```py
name = getattr(view, "name", None)
if name is not None:
    return name
```

Just looks weird to me as written?

nautobot/core/tests/integration/test_view_authentication.py
gsnider2195: Is there a follow-up issue for that work?

gsnider2195 (approved):
Just a couple nitpicks left but nothing blocking. This is great!

nautobot/core/api/utils.py
glennmatthews: Should be `return name` - transcription error on my part.

nautobot/core/tests/integration/test_view_authentication.py
glennmatthews: https://github.com/nautobot/cookiecutter-nautobot-app/issues/102

PR #5475, opened 2024-03-25
Remove indentation on filter

### LINKED ISSUE #4811: Allow sorting of Tree like model list views

timizuoebideri1:
# Closes #4811 
# What's Changed

# Screenshots

## Prefix with filter
Screenshot 2024-03-25 at 10 56 17 AM

## Prefix without filter
Screenshot 2024-03-25 at 10 56 46 AM

## Location without filter
Screenshot 2024-03-25 at 10 56 58 AM

## Location with filter
Screenshot 2024-03-25 at 10 57 19 AM

# TODO

- [ ] Unit, Integration Tests


changes/5475.changed
```diff
@@ -0,0 +1 @@
+Changed the behavior of Prefix table nad Tree Model tables: now after filtering is applied, all hierarchy indentations are removed.
```
gsnider2195, line 1:
```suggestion
Changed the behavior of Prefix table and other Tree Model tables: now after filtering is applied, all hierarchy indentations are removed.
```

nautobot/core/tables.py
```diff
@@ -65,6 +65,8 @@ def __init__(self, *args, user=None, **kwargs):
         # Init table
         super().__init__(*args, **kwargs)
 
+        self.is_filter_applied = is_filter_applied
```
gsnider2195, line 68: can we do something like this here?

```py
self.is_filter_applied = is_filter_applied
self.hide_hierarchy_ui = self.is_filter_applied or self.order_by
```

Then we can just use one argument to the filter?

nautobot/core/views/generic.py
```diff
@@ -220,6 +220,7 @@ def get(self, request):
         display_filter_params = []
         dynamic_filter_form = None
         filter_form = None
+        hide_hierarchy_ui = "q" in request.GET
```
glennmatthews, line 223: Why does the `q` filter need to be handled separately from the filterset check below?

nautobot/core/views/generic.py
```diff
@@ -232,6 +233,9 @@ def get(self, request):
                 )
                 self.queryset = self.queryset.none()
 
+            if filterset.is_valid() and filterset.data:
+                hide_hierarchy_ui = True
```
glennmatthews, line 237: Just wondering - what part of this only applies if a filter query parameter is present? Does `filterset.data` not get set if there are no filters?

nautobot/dcim/tests/test_views.py
```diff
@@ -2230,6 +2237,9 @@ def setUpTestData(cls):
             "software_version": software_versions[2].pk,
         }
 
+    def test_table_with_indentation_is_removed_on_filter_or_sort(self):
+        self.skipTest("InventroyItem table has no implementation of indentation.")
```
glennmatthews, line 2241:
```suggestion
        self.skipTest("InventoryItem table has no implementation of indentation.")
```

Though maybe it should??

nautobot/ipam/tests/test_views.py
```diff
@@ -167,6 +167,25 @@ def setUpTestData(cls):
             "remove_locations": [cls.locations[1].pk],
         }
 
+    def test_indentation_is_removed_on_filter_or_sort(self):
```
glennmatthews, line 170: Why a different test function for this view?

timizuoebideri1: `Prefix` is not a `TreeNode` model, hence it is not possible to query for entries with children using `children__isnull=False` in order to determine indentation has been removed. I made a separate test case for Prefix` `rather than changing the Base test case to fit it.

nautobot/core/views/generic.py
timizuoebideri1: An empty filterset query-param is also valid filterset and we dont want to hide hierarchy indentation if no filter query-param is provided.

glennmatthews: No, the goal is understood - I was asking how this logic actually enables the desired behavior.

nautobot/ipam/tests/test_views.py
glennmatthews: It's not a `TreeModel` but it does have a `parent` FK and a `children` reverse relation.

```
>>> Prefix.objects.filter(children__isnull=True).count()
89
>>> Prefix.objects.filter(children__isnull=False).distinct().count()
28
>>> Prefix.objects.count()
117
```

nautobot/core/views/generic.py
timizuoebideri1: I pass the `hide_hierarchy_ui` into the instantiation of the table

```python
table = self.table(self.queryset, user=request.user, order_by=order_by, hide_hierarchy_ui=hide_hierarchy_ui)
```

Then use it here:
```
TREE_LINK = """
 {% load helpers %}
 {% tree_hierarchy_ui_representation record.tree_depth|as_range table.hide_hierarchy_ui %}
 <a href="{{ record.get_absolute_url }}">{{ record.name }}</a>
 """
```

Does this answer the question?

gsnider2195: `filterset.data` is where the filter params are passed in here:

https://github.com/nautobot/nautobot/blob/8d2cf2cd698e31bd69427b4dc88bc02ba5a8cad7/nautobot/core/views/generic.py#L217-L219

`data` is the first arg to `BaseFilterSet.__init__`

nautobot/core/testing/views.py
```diff
@@ -744,16 +745,20 @@ def test_table_with_indentation_is_removed_on_filter_or_sort(self):
             self.user.is_superuser = True
             self.user.save()
 
-            if not issubclass(self.model, TreeModel):
+            if not issubclass(self.model, (TreeModel)) or self.model is not Prefix:
```
gsnider2195, line 748: Should we use duck typing here? 

```py
if not hasattr(self.model, "ancestors"):
```

timizuoebideri1, line 748:
```suggestion
            if not hasattr(self.model, "ancestors"):
```

gsnider2195: That would match

```py
<class 'nautobot.dcim.models.device_components.InventoryItem'>
<class 'nautobot.dcim.models.devices.ControllerManagedDeviceGroup'>
<class 'nautobot.dcim.models.locations.LocationType'>
<class 'nautobot.dcim.models.locations.Location'>
<class 'nautobot.dcim.models.racks.RackGroup'>
<class 'nautobot.ipam.models.Prefix'>
<class 'nautobot.ipam.models.IPAddress'>
<class 'nautobot.tenancy.models.TenantGroup'>
```

nautobot/core/views/renderers.py
```diff
@@ -65,10 +65,14 @@ def construct_table(self, view, **kwargs):
         table_class = view.get_table_class()
         request = kwargs.get("request", view.request)
         queryset = view.alter_queryset(request)
+
         if view.action in ["list", "notes", "changelog"]:
             if view.action == "list":
                 permissions = kwargs.get("permissions", {})
-                table = table_class(queryset, user=request.user)
+                order_by = view.request.GET.getlist("sort")
```
glennmatthews, line 72: I note that the `notes` and `changelog` cases below aren't taking `order_by`/`sort` into account. Is that another bug, or are those tables non-orderable?

nautobot/core/testing/views.py
```diff
@@ -737,6 +739,32 @@ def get_title(self):
         def get_list_view(self):
             return lookup.get_view_for_model(self.model, view_type="List")
 
+        def test_table_with_indentation_is_removed_on_filter_or_sort(self):
```
glennmatthews, line 742: As @whitej6 brought up in standup, let's make sure we document this new generic test case in the changelog and send a heads-up to app developers.

glennmatthews (approved):

nautobot/core/testing/views.py
glennmatthews: Looks like that's causing the latest CI failure?

nautobot/core/views/generic.py
```diff
@@ -294,7 +301,7 @@ def get(self, request):
             # Construct the objects table
             # Order By is needed in the table `__init__` method
             order_by = self.request.GET.getlist("sort")
```
gsnider2195, line 303: Can we do the `order_by` check here too instead of having the `hide_hierarchy` related checks in 2 different places?

```py
if order_by:
    hide_hierarchy_ui=True
```

nautobot/core/views/renderers.py
gsnider2195: It looks like this is redundant because it's happening in the `RequestConfig.configure` [here](https://github.com/jieter/django-tables2/blob/c592b7b9fcff956499507c38d0f80b7e8b76666e/django_tables2/config.py#L42-L44)? (`table.prefixed_order_by_field` defaults to `"sort"`)

nautobot/core/testing/views.py
gsnider2195: What do you think the category should be? Added, changed or housekeeping? Probably added or changed for the best visibility.

glennmatthews: I'd go with `added`

PR #5476, opened 2024-03-25
Fixed Team(s) field not pre-populating when editing a Contact

### LINKED ISSUE #5415: [v2.2] Team(s) field not pre-populating when editing a Contact
### Environment
* Nautobot version (Docker tag too if applicable): 2.2.0a1
* Python version: 3.11.8
* Database platform, version: n/a
* Middleware(s): n/a

### Steps to Reproduce
1. Create a Team
2. Create a Contact and associate it to the Team
3. Edit the Contact

### Expected Behavior
The existing Team would be displayed in the edit form

### Observed Behavior
The `Team(s)` field is empty and if you don't reapply the team it will remove it.

HanlinMiao:
# Closes #5415
# What's Changed

I think the cause of the issue is that `teams` field on Contact is the reverse side of the many-to-many relationship. So that's why the form field is not auto-populating.
# Screenshots

From object detail view:
Screen Shot 2024-03-25 at 12 20 03 PM
From object list view:
Screen Shot 2024-03-25 at 12 20 15 PM

# TODO

- [ ] Unit, Integration Tests


nautobot/core/views/renderers.py
```diff
@@ -184,6 +185,9 @@ def get_context(self, data, accepted_media_type, renderer_context):
                 form = form_class(initial=request.GET)
             elif view.action in ["create", "update"]:
                 initial_data = normalize_querydict(request.GET, form_class=form_class)
+                # Manually populate the reverse side of the manytomany relationship in ContactForm
```
HanlinMiao, line 188: I wonder if there is a more generic way to do this.

nautobot/core/views/renderers.py
```diff
@@ -184,6 +185,9 @@ def get_context(self, data, accepted_media_type, renderer_context):
...
+                if isinstance(instance, Contact):
+                    initial_data["teams"] = instance.teams.all()
```
glennmatthews, line 190: Do we similarly need to load `"contacts"` if `instance` is a `Team`? Or is that handled automatically as the forward side of the M2M?

nautobot/core/views/renderers.py
```diff
@@ -184,6 +185,9 @@ def get_context(self, data, accepted_media_type, renderer_context):
...
                 form = form_class(instance=instance, initial=initial_data)
```
glennmatthews, line 191: Could we instead of implementing this in core, implement it in the `ContactForm` class `__init__`?

gsnider2195: That's what we discussed yesterday in DSU. Not sure if @HanlinMiao has tried it yet

nautobot/core/views/renderers.py
glennmatthews: Handled automatically.

gsnider2195 (approved):
💯

PR #5492, opened 2024-03-26
Fixed Roles API response containing duplicate entries when filtering …

### LINKED ISSUE #5431: Duplicated entries in the Roles dropdown when creating a new Config Context
### Environment
* Nautobot version (Docker tag too if applicable): 2.1.7 https://demo.nautobot.com/
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Create a new Config Context
2. In the Assignment panel, click the Roles field
3. Observe duplicated role entries (distribution, edge, Backbone etc. all appear twice).

### Expected Behavior

Each role should only appear once.

### Observed Behavior

image

### Investigation

This API call sent on click is the one returning the duplicated objects: https://demo.nautobot.com/api/extras/roles/?q=&limit=50&offset=0&depth=0&content_types=dcim.device&content_types=virtualization.virtualmachine

It looks like the duplication comes from `content_types=dcim.device&content_types=virtualization.virtualmachine`, if you remove one of them then roles appear only once in the result.

timizuoebideri1:
…on more than on `content_types`.

# Closes #5431 
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/5431.fixed
+++ b/changes/5431.fixed
@@ -1,0 +1,1 @@
+Fixed Roles API response containing duplicate entries when filtering on more than on `content_types`.
--- a/changes/5431.fixed
+++ b/changes/5431.fixed
@@ -1,1 +1,2 @@
 Fixed Roles API response containing duplicate entries when filtering on more than one `content_types` value.
+Fixed Providers API response containing duplicate entries when filtering on more than one `location` value.
--- a/nautobot/extras/filters/__init__.py
+++ b/nautobot/extras/filters/__init__.py
@@ -387,6 +387,11 @@ class NautobotFilterSet(
     are needed.
     """
 
+    def filter_queryset(self, queryset):
+        # Ensure the filtered result is distinct, particularly when filtering on a
+        # many-to-many field with `conjoined` set to False to avoid duplicate entries.
+        return super().filter_queryset(queryset).distinct()
+
 
 #
 # Custom Links
@@ -1061,9 +1066,6 @@ class RoleFilterSet(NautobotFilterSet):
         # such as 'Device or VirtualMachine' but not 'Device and VirtualMachine'.
         conjoined=False,
     )
-
-    def filter_queryset(self, queryset):
-        return super().filter_queryset(queryset).distinct()
 
     class Meta:
         model = Role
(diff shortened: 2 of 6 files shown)


changes/5431.fixed
```diff
@@ -0,0 +1 @@
+Fixed Roles API response containing duplicate entries when filtering on more than on `content_types`.
```
glennmatthews, line 1:
```suggestion
Fixed Roles API response containing duplicate entries when filtering on more than one `content_types` value.
```

nautobot/extras/filters/__init__.py
```diff
@@ -1062,6 +1062,9 @@ class RoleFilterSet(NautobotFilterSet):
         conjoined=False,
     )
 
+    def filter_queryset(self, queryset):
+        return super().filter_queryset(queryset).distinct()
```
glennmatthews, line 1066: Is this something that should be implemented directly on `NautobotFilterSet` rather than on `RoleFilterSet` specifically? Curious why it's needed here only - is it because of `conjoined=False` on the content_types filter? Do we have any other `conjoned=False` filters elsewhere that would merit a similar fix?

timizuoebideri1: Move logic to `NautobotFilterSet`.  This should fix for all other places where `conjoined=False`

glennmatthews:
Test failure looks legitimate - looks like the test code constructing the queryset to compare the filter against is itself missing a `.distinct()` call?

nautobot/extras/filters/__init__.py
```diff
@@ -387,6 +387,11 @@ class NautobotFilterSet(
     are needed.
     """
 
+    def filter_queryset(self, queryset):
+        # Ensure the filtered result is distinct, particularly when filtering on a
+        # many-to-many field with `conjoined` set to False to avoid duplicate entries.
+        return super().filter_queryset(queryset).distinct()
```
glennmatthews, line 393: On re-reading the django-filter docs, it should be possible on an individual filter to specify `distinct=True` rather than this broad-spectrum approach, but furthermore, a `MultipleChoiceFilter` already defaults to `distinct=True`. I think we actually have a bug in `ContentTypeMultipleChoiceFilter.filter()` not taking `self.distinct` into account.

glennmatthews (approved):

PR #5493, opened 2024-03-27
adds a configuration setting to disable app metrics for specific apps

Kircheneer:
# Closes DNE
# What's Changed

Metrics can be disabled on a per-app basis. This can be used to disable slow app metrics.


gsnider2195:
Would it make sense to put this in `PLUGINS_CONFIG` instead of adding another global setting?

```py
PLUGINS_CONFIG = {
    "nautobot_plugin_xyz": {
        "metrics_enabled": False,
    }
}
```

nautobot/docs/user-guide/administration/configuration/optional-settings.md
```diff
@@ -701,6 +701,14 @@ Toggle requiring authentication to view `/metrics`. See the [Prometheus Metrics]
 
 ---
 
+## METRICS_DISABLED_APPS
+
+Default: `""`
```
glennmatthews, line 706:
```suggestion
Default: `[]` (empty list)
```

nautobot/docs/user-guide/administration/configuration/optional-settings.md
```diff
@@ -701,6 +701,14 @@ Toggle requiring authentication to view `/metrics`. See the [Prometheus Metrics]
...
+
+Environment Variable: `NAUTOBOT_METRICS_DISABLED_APPS`
+
+Disables app metrics by app name. Example value: `nautobot_ssot nautobot_device_device_lifecycle_mgmt`.
```
glennmatthews, line 710:
```suggestion
Disables app metrics by app name. Example value: `["nautobot_ssot", "nautobot_device_lifecycle_mgmt"]`.
```

development/nautobot_config.py
```diff
@@ -73,5 +73,6 @@
 
 METRICS_ENABLED = True
 METRICS_AUTHENTICATED = False
+METRICS_DISABLED_APPS = ""
```
glennmatthews, line 76:
```suggestion
METRICS_DISABLED_APPS = []
```

glennmatthews:
> Would it make sense to put this in `PLUGINS_CONFIG` instead of adding another global setting?
> 
> ```python
> PLUGINS_CONFIG = {
>     "nautobot_plugin_xyz": {
>         "metrics_enabled": False,
>     }
> }
> ```

Not sure we can unilaterally add things to the `PLUGINS_CONFIG` dict, might be impactful to Cloud and its use of JSON schemas for app configurations. Other than that concern, I like the idea!

Kircheneer:
> > Would it make sense to put this in `PLUGINS_CONFIG` instead of adding another global setting?
> > ```python
> > PLUGINS_CONFIG = {
> >     "nautobot_plugin_xyz": {
> >         "metrics_enabled": False,
> >     }
> > }
> > ```
> 
> Not sure we can unilaterally add things to the `PLUGINS_CONFIG` dict, might be impactful to Cloud and its use of JSON schemas for app configurations. Other than that concern, I like the idea!

Yes, this was the exact concern I shared. This would be a new pattern to include in plugin config and it would make this feature far more complicated. If you say I should explore this I can do it, but it would be a far deeper change.

nautobot/docs/user-guide/administration/guides/prometheus-metrics.md
```diff
@@ -25,6 +25,11 @@ CACHES = {
 }
 ```
 
+<!-- TODO: Which version? -->
+++ 2.2.0
```
gsnider2195, line 29: I'm thinking 2.2.1 is fine for this? It's not a breaking change and not significant enough to push out to 2.3.0.

```suggestion
+++ 2.2.1
```

glennmatthews:
Settings documentation needs to be ported over to `settings.yaml` - we can probably handle this part.

nautobot/docs/user-guide/administration/configuration/optional-settings.md
```diff
@@ -701,6 +701,14 @@ Toggle requiring authentication to view `/metrics`. See the [Prometheus Metrics]
 
 ---
 
+## METRICS_DISABLED_APPS
```
gsnider2195, line 704: This will need to be ported to `settings.yaml` for 2.2. Can you model this after `ALLOWED_HOSTS` and add information about how to set the environment variable?

glennmatthews:
Test is failing because as currently implemented the environment variable is space-separated, but the expectation (for every env variable *except* `ALLOWED_HOSTS`) is for it to be comma-separated instead. We probably should change to match the expectation rather than changing the test.

Kircheneer:
Alright, should be good now. Thanks for taking care of porting over to `settings.yaml`.

Kircheneer:
Hm I don't quite understand the CI failure, why should my constant be part of the config?

glennmatthews:
Because it's being imported into nautobot.core.settings, the test sees it and things it's a configurable setting. Probably need to update the test to ignore it.

Kircheneer:
I have instead opted for an `import ... as`, I like that a little better than implementing an explicit exception

nautobot/docs/user-guide/administration/guides/prometheus-metrics.md
```diff
@@ -25,6 +25,9 @@ CACHES = {
 }
 ```
 
++++ 2.2.1
+    In case the `/metrics` endpoint is unacceptably slow to respond, you can disable specific apps that may be causing problems with the `METRICS_DISABLED_APPS` configuration setting.
```
itdependsnetworks, line 29:
```suggestion
    In case the `/metrics` endpoint is not performant or not required slow, you can disable specific apps with the `METRICS_DISABLED_APPS` configuration setting.
```

nautobot/core/templates/nautobot_config.py.j2
```diff
@@ -409,12 +409,17 @@ INSTALLATION_METRICS_ENABLED = is_truthy(os.getenv("NAUTOBOT_INSTALLATION_METRIC
...
+#     METRICS_DISABLED_APPS = os.getenv("NAUTOBOT_METRICS_DISABLED_APPS", "").split(",")
 
 # Require API Authentication to HTTP endpoint '/metrics'
 #
 # METRICS_AUTHENTICATED = is_truthy(os.getenv("NAUTOBOT_METRICS_AUTHENTICATED", "False"))
 
+# Disable app metrics for specific apps
+#
+# METRICS_DISABLED_APPS = os.getenv("NAUTOBOT_METRICS_DISABLED_APPS", "").split(",")
+
```
glennmatthews, line 422:
```suggestion
# METRICS_ENABLED = is_truthy(os.getenv("NAUTOBOT_METRICS_ENABLED", "False"))

# Require API Authentication to HTTP endpoint '/metrics'
#
# METRICS_AUTHENTICATED = is_truthy(os.getenv("NAUTOBOT_METRICS_AUTHENTICATED", "False"))

# Disable app metrics for specific apps
#
# if "NAUTOBOT_METRICS_DISABLED_APPS" in os.environ and os.environ["NAUTOBOT_METRICS_DISABLED_APPS"] != "":
#     METRICS_DISABLED_APPS = os.getenv("NAUTOBOT_METRICS_DISABLED_APPS", "").split(",")

```

nautobot/core/settings.yaml
```diff
@@ -1171,6 +1171,18 @@ properties:
...
+    description: >-
+      A list of app names for which Prometheus metrics should be disabled.
+      (If provided as an environment variable, it should be a comma-separated string, for example
+      `NAUTOBOT_METRICS_DISABLED_APPS="nautobot_ssot, nautobot_device_lifecycle_mgmt"`.)
+    environment_variable: "NAUTOBOT_METRICS_DISABLED_APPS"
+    items:
+      type: "string"
+    see_also:
+      "Guide to Nautobot Prometheus metrics": "../guides/prometheus-metrics.md"
+    type: "array"
```
glennmatthews, line 1185:
```suggestion
    type: "array"
    version_added: "2.2.1"
```

Issue #5518, opened 2024-04-01
Code changes needed to update to drf-spectacular latest version


glennmatthews:
### Environment
* Nautobot version (Docker tag too if applicable): 2.2.0

### Steps to Reproduce
1. `poetry add --extras sidecar drf-spectacular~0.27.2`
2. Build container
3. `invoke check-schema`

### Expected Behavior

Schema generation to pass without errors

### Observed Behavior

```
Warning: enum naming encountered a non-optimally resolvable collision for fields named "type". The same name has been used for multiple choice sets in multiple components. The collision was resolved with "Type6adEnum". add an entry to ENUM_NAME_OVERRIDES to fix the naming.
Warning: enum naming encountered a non-optimally resolvable collision for fields named "type". The same name has been used for multiple choice sets in multiple components. The collision was resolved with "Type09dEnum". add an entry to ENUM_NAME_OVERRIDES to fix the naming.
Warning: enum naming encountered a non-optimally resolvable collision for fields named "type". The same name has been used for multiple choice sets in multiple components. The collision was resolved with "TypeF53Enum". add an entry to ENUM_NAME_OVERRIDES to fix the naming.
Warning: enum naming encountered a non-optimally resolvable collision for fields named "type". The same name has been used for multiple choice sets in multiple components. The collision was resolved with "Type7b8Enum". add an entry to ENUM_NAME_OVERRIDES to fix the naming.
Warning: enum naming encountered a non-optimally resolvable collision for fields named "type". The same name has been used for multiple choice sets in multiple components. The collision was resolved with "Type459Enum". add an entry to ENUM_NAME_OVERRIDES to fix the naming.
Warning: enum naming encountered a non-optimally resolvable collision for fields named "mode". The same name has been used for multiple choice sets in multiple components. The collision was resolved with "Mode308Enum". add an entry to ENUM_NAME_OVERRIDES to fix the naming.
Warning: enum naming encountered a non-optimally resolvable collision for fields named "mode". The same name has been used for multiple choice sets in multiple components. The collision was resolved with "Mode791Enum". add an entry to ENUM_NAME_OVERRIDES to fix the naming.

Schema generation summary:
Warnings: 7 (7 unique)
Errors:   0 (0 unique)

SchemaGenerationError: Failing as requested due to warnings
```

This appears to be due to additional validation logic added by https://github.com/tfranzel/drf-spectacular/pull/1113

lampwins:
Once this is done, ensure the dependency is actually updated.

Issue #5517, opened 2024-04-01
Code changes needed to update to django-rest-framework latest version


glennmatthews:
### Environment
* Nautobot version (Docker tag too if applicable): 2.2.0

### Steps to Reproduce
1. `poetry add django-rest-framework~3.15.0`
2. Build docker container
3. Start Nautobot

### Expected Behavior

Nautobot to start successfully

### Observed Behavior

```
Traceback (most recent call last):
  File "/usr/local/bin/nautobot-server", line 6, in <module>
    sys.exit(main())
  File "/source/nautobot/core/cli/__init__.py", line 286, in main
    execute_from_command_line([sys.argv[0], *unparsed_args])
  File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 395, in execute
    django.setup()
  File "/usr/local/lib/python3.8/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/usr/local/lib/python3.8/site-packages/django/apps/registry.py", line 122, in populate
    app_config.ready()
  File "/source/nautobot/circuits/apps.py", line 14, in ready
    super().ready()
  File "/source/nautobot/core/apps/__init__.py", line 57, in ready
    register_menu_items(menu_items)
  File "/source/nautobot/core/apps/__init__.py", line 174, in register_menu_items
    item.link = reverse(item.link, args=item.args, kwargs=item.kwargs)
  File "/usr/local/lib/python3.8/site-packages/django/urls/base.py", line 54, in reverse
    app_list = resolver.app_dict[ns]
  File "/usr/local/lib/python3.8/site-packages/django/urls/resolvers.py", line 530, in app_dict
    self._populate()
  File "/usr/local/lib/python3.8/site-packages/django/urls/resolvers.py", line 464, in _populate
    for url_pattern in reversed(self.url_patterns):
  File "/usr/local/lib/python3.8/site-packages/django/utils/functional.py", line 48, in __get__
    res = instance.__dict__[self.name] = self.func(instance)
  File "/usr/local/lib/python3.8/site-packages/django/urls/resolvers.py", line 602, in url_patterns
    patterns = getattr(self.urlconf_module, "urlpatterns", self.urlconf_module)
[...]

PR #5521, opened 2024-04-01
Update minor versions of *many* dependencies.

glennmatthews:
# Closes #DNE
# What's Changed

Essentially this is a manual attempt at addressing #4564, albeit targeted to `develop` rather than `next` (Since all dependency bumps in here are minor versions, I'm of the opinion that this should be acceptable in a patch release).

In addition to *not* including any major version bumps (see #4616), this also omits the following updates that had been proposed by #4564:
- `djangorestframework` 3.15+ (#5517)
- `drf-spectacular` 0.27+ (#5518)
- `Markdown` 3.6+ (currently disallowed by latest `mkdocstrings-python` due to unspecified "breaking changes")
- `graphene-django-optimizer` 0.9+ (would require a major version bump of `graphene-django`)

There are minor code changes included in this PR as follows:

- Selenium docker image version updates corresponding to the Selenium dependency bump
- A few whitespace changes resulting from the `ruff` dependency bump
- Docs changes from `storages.backends.s3boto3.S3Boto3Storage` to `storages.backends.s3.S3Storage` and similar resulting from the `django-storages` dependency bump (**note**: the old import is still supported for now but is considered deprecated)
- Integration test update to handle a new URL pattern added by the `django-health-check` bump
- Changes to a couple of `yield` statements to address a new `pylint` warning
- Removal of previously deprecated `ExtendedManager` and `ExtendedQuerySet` base classes from `django-celery-beat` (these were actually removed from the code base in the previous version *before* this dependency bump, but for some reason the code was still present in the published wheel??)

# TODO

- [x] Unit, Integration Tests


changes/5521.dependencies
```diff
@@ -0,0 +1 @@
+Updated most dependencies to the latest versions available (as of 2024-04-01) that could be supported without significant code changes in Nautobot.
```
HanlinMiao, line 1: Should we list the dependencies explicitly here?

HanlinMiao (approved):

changes/5521.dependencies
glennmatthews: We could, I'm just not certain that that's interesting/useful information to most users. Open to persuasion on that point though.

glennmatthews:
```suggestion
Updated most dependencies to the latest versions available as of 2024-04-01.
```

timizuoebideri1 (approved):

PR #5526, opened 2024-04-03
Fixed doc referece to job cprofile file location

itdependsnetworks:
# Closes N/A
# What's Changed

Reflected location of docs: https://github.com/nautobot/nautobot/blob/c6ce7dbd4b98aa776fbf630f733ce44f48f313f7/nautobot/extras/jobs.py#L138C60-L138C78

 
# Screenshots

# TODO

### DIFF
--- a/changes/5526.fixed
+++ b/changes/5526.fixed
@@ -1,0 +1,1 @@
+Fixed doc referece to job cprofile file location.
--- a/changes/5526.fixed
+++ b/changes/5526.fixed
@@ -1,1 +1,1 @@
-Fixed doc referece to job cprofile file location.
+Fixed doc reference to job cprofile file location.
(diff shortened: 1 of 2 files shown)


itdependsnetworks:
This should get back ported to ltm as well.

changes/5526.fixed
```diff
@@ -0,0 +1 @@
+Fixed doc referece to job cprofile file location.
```
itdependsnetworks, line 1:
```suggestion
Fixed doc reference to job cprofile file location.
```

Kircheneer (approved):

glennmatthews (approved):
Thanks! Longer term we want to change the implementation of this feature so that it creates the profile file as a downloadable attachment to the JobResult rather than a file in `/tmp`, but this is good for now. :-)

PR #5529, opened 2024-04-03
Disable automatic ChangeLog creation on signal to enhance performance…

### LINKED ISSUE #2459: Bulk operations do not scale
### Environment
* Nautobot version (Docker tag too if applicable): 1.4.3
* Python version: 3.9
* Database platform, version: All
* Middleware(s): N/A

### Steps to Reproduce
1. Have a large data set with many thousands of devices and/or other objects (>50,000)
2. Attempt to perform a bulk update or bulk delete on a large subset of the inventory (~2000) in the UI. For example, updating a custom field value, changing tags, deleting devices, etc.
3. Wait

### Expected Behavior
The operation should succeed and preferably in a reasonable amount of time and with continual feedback to the user.

### Observed Behavior
The operation is contained within a single HTTP request but this request takes a very long time to process, so the user gets no feedback that Nautobot is actually doing anything. Depending on the size of the operation, the request will eventually time out and the user is presented with an ugly 504. I call out the UI specifically here, but the same fundamental issue exists with the REST API.

timizuoebideri1:
Disable automatic ChangeLog creation on signal to enhance performance and manually manage a bulk ChangeLog creation post-update.

# Closes #2459 
# What's Changed

**Editing 7000 Tags**
- Before bulk change-log create took: `210s`
- After bulk change-log create took: `131s` 

**Deleting  7000 Tags with 2+ Notes per tag**
- Before bulk change-log create took:
- After bulk change-log create took: 

# Screenshots

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/context_managers.py
```diff
@@ -152,3 +158,56 @@ def web_request_context(
...
+            objectchange_entries.append(objectchange)
+        if associations_to_delete:
+            ContactAssociation.objects.filter(pk__in=associations_to_delete).delete()
+        if notes_to_delete:
+            Note.objects.filter(pk__in=notes_to_delete).delete()
+        ObjectChange.objects.bulk_create(objectchange_entries, batch_size=1000)
+    try:
+        yield
+    except Exception as err:
+        change_context_state.reset(prev_state)
```
glennmatthews, line 204: is this needed, since you're also doing it in `finally`?

nautobot/extras/context_managers.py
```diff
@@ -152,3 +158,56 @@ def web_request_context(
...
+        raise err
+    finally:
+        if action == ObjectChangeActionChoices.ACTION_UPDATE:
+            objectchange_entries = []
+            for instance in objs:
+                objectchange = get_objectchange_for_instance(instance, prev_state.old_value, action)
+                objectchange_entries.append(objectchange)
+            ObjectChange.objects.bulk_create(objectchange_entries, batch_size=1000)
```
glennmatthews, line 212: Actually - why is this in the `finally` instead of in the `try`? Do we still want to create these change logs if there's an exception when actually performing the change?

nautobot/extras/context_managers.py
```diff
@@ -152,3 +158,56 @@ def web_request_context(
...
+                notes = instance.notes.values_list("pk", flat=True)
+                notes_to_delete.update(notes)
+
+            objectchange = get_objectchange_for_instance(instance, prev_state.old_value, action)
+            objectchange_entries.append(objectchange)
+        if associations_to_delete:
+            ContactAssociation.objects.filter(pk__in=associations_to_delete).delete()
+        if notes_to_delete:
+            Note.objects.filter(pk__in=notes_to_delete).delete()
+        ObjectChange.objects.bulk_create(objectchange_entries, batch_size=1000)
```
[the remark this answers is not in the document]
timizuoebideri1: I think it should be before, because `ObjectChanges` are created  currently in `pre_delete` signal

changes/2459.changed
```diff
@@ -0,0 +1 @@
+Disable automatic ChangeLog creation on signal to enhance performance and manually manage a bulk ChangeLog creation post-update.
```
glennmatthews, line 1:
```suggestion
Improved performance of UI bulk-edit and bulk-delete operations by refactoring change-log logic for these cases.
```

nautobot/extras/context_managers.py
```diff
@@ -152,3 +159,57 @@ def web_request_context(
...
+        objectchange.change_context_detail = context_state.context_detail[:CHANGELOG_MAX_CHANGE_CONTEXT_DETAIL]
+        return objectchange
+
+    with transaction.atomic():
+        # Disable automatic ChangeLog creation on signal to enhance performance.
+        # Multiple object updates lead to excessive database calls.
+        # Instead, we'll manually manage a bulk ChangeLog creation post-update.
+        prev_state = change_context_state.set(None)
+
+        if action == ObjectChangeActionChoices.ACTION_DELETE:
```
glennmatthews, line 180: It "feels" like this function should, for completeness, handle ACTION_CREATE as well as ACTION_UPDATE and ACTION_DELETE. Maybe it'd be useful in some of the DCIM views where we bulk-create device components?

nautobot/extras/context_managers.py
```diff
@@ -152,3 +159,57 @@ def web_request_context(
         for object_change in ObjectChange.objects.filter(request_id=change_context.change_id).iterator():
             enqueue_job_hooks(object_change)
             enqueue_webhooks(object_change)
+
+
+@contextmanager
+def object_changelogs_bulk_operation(objs, user, action):
```
glennmatthews, line 165: I'm not super happy with this function name but I'm having a hard time thinking of a better one. Maybe something like `deferred_change_logging_for_bulk_operation()`?

glennmatthews, line 165: Please add a detailed docstring to this function. I'm also wondering if we should expose this in `nautobot.apps` so that Apps can take advantage of it?

nautobot/extras/context_managers.py
```diff
@@ -152,3 +159,57 @@ def web_request_context(
...
+            if objectchange_entries:
+                ObjectChange.objects.bulk_create(objectchange_entries, batch_size=1000)
+        try:
+            yield
+            if action == ObjectChangeActionChoices.ACTION_UPDATE and hasattr(objs[0], "to_objectchange"):
+                objectchange_entries = []
+                for instance in objs:
+                    objectchange = get_objectchange_for_instance(instance, prev_state.old_value, action)
+                    objectchange_entries.append(objectchange)
+                ObjectChange.objects.bulk_create(objectchange_entries, batch_size=1000)
```
glennmatthews, line 211: I'm wondering if there's an alternate way we could implement this that would allow for a mixture of actions (CRUD) within a single atomic transaction using this wrapper (for example, for SSOT). Conceptually:

- Initialize `objectchange_entries = []`
- Disconnect the usual `_handle_changed_object` and `_handle_deleted_object` signals
- Connect alternate signals that do basically the same thing, but stick the constructed ObjectChange objects in `objectchange_entries` instead of immediately saving them to the DB
- Yield
- ObjectChange.objects.bulk_create(objectchange_entries)
- Reconnect original signals

Thinking even more outside the box, maybe `atomic` could become an option to the `change_logging()` context manager that would automatically invoke this alternate approach when set?

gsnider2195: As discussed earlier, we may not be able to use this approach because we could become out of sync in this kind of scenario:
```py
with change_logging(change_context):
    with transaction.atomic():
        Namespace.objects.create(name="foo")
        raise Exception("rollback that change!")

    with transaction.atomic():
        Namespace.objects.create(name="bar")
 ```

The above code would create an object change for Namespace `foo` even though it was rolled back

gsnider2195: I wonder if we could just do the `flush` on `transaction.on_commit`?

gsnider2195:
Testing results [here](https://github.com/nautobot/nautobot/issues/2459#issuecomment-2045872093)

nautobot/extras/context_managers.py
gsnider2195: It doesn't look like we're going to be able to go with the generic solution because we don't have a way to clear the queue on transaction rollback.

gsnider2195:
Here's my new [hybrid approach](https://github.com/nautobot/nautobot/commit/bd468230f79df7d534a91168dc28e2aed2eb7a32)

gsnider2195:
> Here's my new [hybrid approach](https://github.com/nautobot/nautobot/commit/bd468230f79df7d534a91168dc28e2aed2eb7a32)

This appears to have similar performance to `develop`

PR #5551, opened 2024-04-09
Fixed contact tab disappearing when accessing dynamic groups tab.

### LINKED ISSUE #5511: Contact tab disappears when accessing Dynamic Groups tab
### Environment
* Nautobot version (Docker tag too if applicable): 2.2
* Python version: 3.11
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Go to a device detail page
2. click on dynamic groups tab
3. Look for contacts tab

### Expected Behavior

Contacts tab showing like this
image

### Observed Behavior

image

HanlinMiao:
# Closes #5511
# What's Changed

Changed the way for models to determine whether they are contact associable models or not. Implemented as a `property` in `BaseModel`.

# Screenshots

Screen Shot 2024-04-09 at 2 39 21 PM
Screen Shot 2024-04-09 at 2 39 05 PM
Screen Shot 2024-04-09 at 2 39 14 PM

# TODO

- [ ] Unit, Integration Tests


nautobot/core/models/__init__.py
```diff
@@ -89,6 +90,13 @@ def present_in_database(self):
         """
         return not self._state.adding
 
+    @property
+    def is_contact_associable_model(self):
+        """
+        True if the record can be associated with a contact/team, False if it cannot.
+        """
+        return f"{self._meta.app_label}.{self._meta.model_name}" not in NON_CONTACT_ASSOCIABLE_MODELS
+
```
glennmatthews, line 99: Rather than a property with a hard-coded list, can we just set a flag on the model class or model Meta to this effect? I can imagine that there may be app models that will similarly want to opt out, for example.

HanlinMiao: Epitome of over-engineering

nautobot/docs/user-guide/core-data-model/extras/contact.md
```diff
@@ -3,3 +3,18 @@
...
+```python
+
+class YourModel(PrimaryModel):
+    name = models.CharField(max_length=CHARFIELD_MAX_LENGTH, db_index=True)
+    is_contact_associable_model = False
+
+    class Meta:
+        abstract = True
+
+```
```
glennmatthews, line 20: Possibly indent this block by four spaces so that it's included inside the 2.2.1 panel?

nautobot/docs/user-guide/core-data-model/extras/contact.md
```diff
@@ -3,3 +3,18 @@
 A contact represents an individual that can be associated with other model instances. It stores the necessary information to uniquely identify and contact the individual. The information that can be stored on a contact includes the individual's name, phone number, email, and address. When a contact is assigned to an object, the status of such assignment needs to be specified.
 
 A contact can be assigned to one or more [teams](team.md)
+
++/- 2.2.1
+
+    A model can opt out of the ability to be asssociated with contacts or teams by specifying `is_contact_associable = False` in the model class definition.
```
glennmatthews, line 9:
```suggestion
    A model can opt out of the ability to be associated with contacts or teams by specifying `is_contact_associable = False` in the model class definition.
```

Might be worth also noting that presently this only restricts the UI - it doesn't actually prevent contacts/teams from associating to this model at the DB level, it just hides the UI for doing so.

nautobot/core/views/renderers.py
```diff
@@ -245,7 +243,7 @@ def get_context(self, data, accepted_media_type, renderer_context):
             context["created_by"] = created_by
             context["last_updated_by"] = last_updated_by
             associated_contacts = instance.associated_contacts.restrict(request.user, "view").order_by("role__name")
```
glennmatthews, line 245: Should this be inside the `if` block below? We don't need to hit the DB here if we aren't actually going to use this data.

glennmatthews: More generally, why is the logic here so different from lines 145-158 above in `generic.py`?

HanlinMiao: You are right, they should mirror each other

glennmatthews (approved):

PR #5567, opened 2024-04-12
Added changelogging for custom field background tasks.

### LINKED ISSUE #1631: Change logging for custom field deletion clean-up task
In addition to the work already done for #1582 on [this PR](https://github.com/nautobot/nautobot/pull/1626), we need to add change logging for each updated object in nautobot.extras.tasks.delete_custom_field_data [code here](https://github.com/nautobot/nautobot/blob/e8030fc86f6d1e8489c644717a42d8bdda3cb151/nautobot/extras/tasks.py#L59)

HanlinMiao:
# Closes #1631 
# What's Changed

# Screenshots

Changelogs populated after deleting a CustomField:
Screenshot 2024-04-12 at 2 09 02 PM
Changelogs populated after provisioning a new CustomField:
Screenshot 2024-04-12 at 2 10 04 PM
Changelogs populated after updating a CustomField choice value:
Screenshot 2024-04-12 at 2 12 11 PM

# TODO

- [x] Unit, Integration Tests


nautobot/extras/models/customfields.py
```diff
@@ -724,7 +724,20 @@ def delete(self, *args, **kwargs):
 
         super().delete(*args, **kwargs)
 
-        delete_custom_field_data.delay(self.key, content_types)
+        # Circular Import
+        from nautobot.extras.signals import _get_user_if_authenticated, change_context_state
```
glennmatthews, line 728: As part of this changeset can we either:

- rename `_get_user_if_authenticated` to something non-private?
- roll it into `change_context.get_user()` as built-in functionality so we don't have to do this check all over the place?

nautobot/extras/models/customfields.py
```diff
@@ -724,7 +724,20 @@ def delete(self, *args, **kwargs):
...
+        change_context = change_context_state.get()
+        if change_context is None:
+            context = {}
+        else:
+            context = {
+                "user": _get_user_if_authenticated(change_context.get_user(), self),
+                "change_id": change_context.change_id,
+                "context_detail": "delete custom field data",
+                "context": change_context.context,
+            }
```
glennmatthews, line 739: How about adding an `.as_dict()` method to `ChangeContext`? (I don't know that we need different context_detail strings for each of these methods, it *might* be preferable to keep whatever context_detail was already set on the ChangeContext itself instead?)

nautobot/extras/tasks.py
```diff
@@ -14,7 +14,7 @@
 
 
 @nautobot_task
-def update_custom_field_choice_data(field_id, old_value, new_value):
+def update_custom_field_choice_data(field_id, old_value, new_value, change_context={}):
```
glennmatthews, line 17: Don't use mutable objects as defaults. :)

```suggestion
def update_custom_field_choice_data(field_id, old_value, new_value, change_context=None):
```

nautobot/extras/tasks.py
```diff
@@ -57,31 +83,47 @@ def update_custom_field_choice_data(field_id, old_value, new_value):
 
 
 @nautobot_task
-def delete_custom_field_data(field_key, content_type_pk_set):
+def delete_custom_field_data(field_key, content_type_pk_set, change_context={}):
```
glennmatthews, line 86:
```suggestion
def delete_custom_field_data(field_key, content_type_pk_set, change_context=None):
```

nautobot/extras/tasks.py
```diff
@@ -57,31 +83,47 @@ def update_custom_field_choice_data(field_id, old_value, new_value):
...
+                        obj.save()
+            else:
+                for obj in model.objects.filter(**{f"_custom_field_data__{field_key}__isnull": False}):
+                    del obj._custom_field_data[field_key]
+                    obj.save()
 
 
 @nautobot_task
-def provision_field(field_id, content_type_pk_set):
+def provision_field(field_id, content_type_pk_set, change_context={}):
```
glennmatthews, line 117:
```suggestion
def provision_field(field_id, content_type_pk_set, change_context=None):
```

nautobot/extras/models/customfields.py
HanlinMiao: The context detail is "" on these ChangeContext. It might be helpful to be more specific?

nautobot/extras/models/customfields.py
HanlinMiao: _get_user_if_authenticated needs the `instance` argument which is not available on the ChangeContext. 
Would it be acceptable to set `instance` to `None` when using it in `get_user()`?

glennmatthews: Nah, the instance is useful debugging info. Can we pass it as an optional parameter to `get_user()`?

nautobot/extras/models/customfields.py
```diff
@@ -724,7 +724,17 @@ def delete(self, *args, **kwargs):
...
-        delete_custom_field_data.delay(self.key, content_types)
+        # Circular Import
+        from nautobot.extras.signals import change_context_state
+
+        change_context = change_context_state.get()
+        if change_context is None:
+            context = None
+        else:
+            context = change_context.as_dict()
+            context["user"] = change_context.get_user(self)
```
glennmatthews, line 735: Can we consolidate this even a bit further? A la:

```suggestion
            context = change_context.as_dict(instance=self)
```

glennmatthews (approved):

PR #5570, opened 2024-04-12
Required relationships not respecting filter when both source and destination content types are the same

### LINKED ISSUE #5569: Required relationships not respecting filter when both source and destination content types are the same
### Environment
* Nautobot version (Docker tag too if applicable): 2.1.5
* Python version: 3.9
* Middleware(s):

### Steps to Reproduce
1. Create three devices, each with a different role (`Device 1` with `Role 1`, `Device 2` with `Role 2`, `Device 3` with `Role 3`).
2. Create a one to many relationship (test) with source type of `dcim | device`, source filter of `{"role": ["Role 1"]}`, destination type of `dcim | device`, destination filter of `{"role": ["Role 2"]}`, and set required on `Destination must implement this relationship`.
3. Attempt to edit `Device 3`.

### Expected Behavior
`Device 3` can be edited and the form submission works with no error.

### Observed Behavior
On form submission, the following error occurs:
`'DeviceForm' has no field named 'cr_test__source'.`

sdoiron0330:
# Closes #5569 
# What's Changed

added logic to determine if a relationship side is required and ensure the given object instance exists in the respective filter.
# Screenshots

# TODO

- [ ] Unit, Integration Tests


HanlinMiao (approved):

nautobot/extras/models/relationships.py
```diff
@@ -244,6 +244,13 @@ def required_related_objects_errors(
             if relation.skip_required(cls, opposite_side):
                 continue
 
+            if getattr(relation, f"{relation.required_on}_filter") and instance:
+                filterset = get_filterset_for_model(cls)
+                if filterset:
+                    filter_params = getattr(relation, f"{relation.required_on}_filter")
+                    if not filterset(filter_params, cls.objects.filter(id=instance.id)).qs.exists():
+                        continue
+
```
glennmatthews, line 253: Should this logic instead be part of the `relation.skip_required` function called above?

sdoiron0330: so i read the `required_related_objects_errors` function and the corresponding call to `relation.skip_required` and interpreted it as "is the opposite side of this relationship required?".  the question i was trying to answer with this code is "is this object even in the filtered queryset", so i added it here.  i may have a bad interpretation of the `skip_required` function.  

that said, i'm sure i can slide that logic into `skip_required` with minimal headache if that's the right place for it.

glennmatthews: Yeah, on a closer re-read `skip_required` is about "should we avoid enforcing this relationship even though it's marked as `required`, due to special circumstances?" So I think this code location is fine. 

Would be good to add a unit test case for this logic - the `source`/`destination` attributes and logic are more than a bit confusing and it'd be good to verify that that this is enforcing things in the right "direction".

sdoiron0330: sounds good -- i should be able to tackle that by end of week

nautobot/extras/tests/test_relationships.py
```diff
@@ -486,6 +488,56 @@ def test_get_for_model_caching_and_cache_invalidation(self):
...
+        role_3.content_types.add(ContentType.objects.get_for_model(Device))
+        device_3 = Device.objects.create(
+            device_type=device_type, role=role_3, name="Device 3", location=self.locations[0], status=status
+        )
+        # Create a one-to-many relationship with destination required, source filter: {"role": ["Role 1"]}
+        # and destination filter {"role": ["Role 2"]}
+        Relationship.objects.create(
+            label="Device to Devices",
+            key="device_to_devices",
+            source_type=ContentType.objects.get_for_model(Device),
```
gsnider2195, line 521:
```suggestion
            source_type=device_ct,
```

nautobot/extras/tests/test_relationships.py
```diff
@@ -486,6 +488,56 @@ def test_get_for_model_caching_and_cache_invalidation(self):
...
+        update_status = Status.objects.get_for_model(Device).last()
+        update_data_for_device_3 = {
+            "location": device_3.location.pk,
+            "device_type": device_3.device_type.pk,
+            "role": device_3.role.pk,
+            "name": device_3.name,
+            "status": update_status.pk,
+        }
+        form = DeviceForm(data=update_data_for_device_3)
+        form.is_valid()
```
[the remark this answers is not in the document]
smk4664: Can we test the form is also valid when:
- [x] Device 1 has no relationship
- [ ] Device 1 has a relationship to Device 2 and Device 3 **(not valid, Device 3 doesn’t match filter)**
- [x] Device 1 has a relationship to Device 2 only
- [ ] Device 2 has a relationship to Device 1 and Device 3 **(not valid, Can’t have multiple source on one-to-many and Device 3 doesn’t match filter)**
And the form is not valid when:
- [x] Device 2 has no relationship
- [x] Device 2 has a relationship to Device 3
- [x] Device 1 has a relationship to Device 3 only 
- [x] And then change the relationship's required_on to source and make sure everything still works as expected?

changes/5569.changed
```diff
@@ -0,0 +1 @@
+Changed logic in determining if a relationship side is required to ensure object instance exists in the filter.
```
glennmatthews, line 1: 1) Should this be `.fixed` since it's addressing a bug?
2) Can we make this any more clear about specifically what it's fixing?

nautobot/extras/models/relationships.py
```diff
@@ -244,6 +244,13 @@ def required_related_objects_errors(
             if relation.skip_required(cls, opposite_side):
                 continue
 
+            if getattr(relation, f"{relation.required_on}_filter") and instance:
+                filterset = get_filterset_for_model(cls)
+                if filterset:
+                    filter_params = getattr(relation, f"{relation.required_on}_filter")
+                    if not filterset(filter_params, cls.objects.filter(id=instance.id)).qs.exists():
+                        continue
```
glennmatthews, line 252: Can we add a comment or two explaining this block for future reference?

glennmatthews (approved):

nautobot/extras/tests/test_relationships.py
```diff
@@ -486,6 +488,223 @@ def test_get_for_model_caching_and_cache_invalidation(self):
...
+            key="device_to_devices",
+            source_type=device_ct,
+            source_filter={"role": ["Role 1"]},
+            destination_type=device_ct,
+            destination_filter={"role": ["Role 2"]},
+            type=RelationshipTypeChoices.TYPE_ONE_TO_MANY,
+            required_on="destination",
+        )
+        # Attempt to update device_3 which will not be in the queryset filtered by the destination filter
+        # Assert that the form is valid and no ValueError is raised.
```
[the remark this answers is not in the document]
smk4664: This test specific assert is validating that an object that is not listed in the filters for a relationship doesn't fail form validation. It's ensuring that you can update a device that is not part or your relationships.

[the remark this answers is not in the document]
smk4664: Re-reading the bug ticket, this test validates that the bug does not occur. I will document that.

PR #5442, opened 2024-03-18
System Job for cleaning up ObjectChange records

glennmatthews:
# Closes #2352, closes #3749
# What's Changed

- Add ~`ObjectChangeCleanup`~ `LogsCleanup` system Job. Fixes #3749.
  - [x] Need to add this to the Release Overview documentation
  - Takes one argument for now, which is the age cutoff in days to apply. If unspecified, defaults to the value of `CHANGELOG_RETENTION`.
  - To improve performance of the bulk delete, it disables the general-purpose deletion signal `_handle_deleted_objects` as it isn't relevant to the deletion of ObjectChange or JobResult records.
  - Additionally, I reworked the `invalidate_models_cache` signal handler so that it no longer triggers on *all* model saves and deletes, but now only on CustomField/ComputedField/Relationship changes as it was intended to.
  - In some quick profiling, I found this set of changes improved the job runtime by roughly an order of magnitude (e.g. deleting 6000 records improved from 13 seconds to roughly 1.3 seconds).
- Added `JobResultFactory`, `JobLogEntryFactory`, and `ObjectChangeFactory` factory classes and added them to `generate_test_data`.
  - [x] Need to verify whether this impacts any unit test behavior.
  - [x] Need to document this in a change fragment
- Removed automatic random inline cleanup of ObjectChanges from the `_handle_changed_object` signal. Fixes #2352.
  - Removed references to `CHANGELOG_RETENTION` in UI templates since it is no longer an automatic given behavior

# Screenshots

Job output:

image

Factory-generated ersatz ObjectChange records:

image

image

# TODO

- [x] Explanation of Change(s)
- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [x] Attached Screenshots, Payload Example
- [x] Unit, Integration Tests
- [x] Documentation Updates (when adding/changing features)
- n/a Example Plugin Updates (when adding/changing features)
- [x] Outline Remaining Work, Constraints from Design

PR #5572, opened 2024-04-14
Change logging performance improvements

### LINKED ISSUE #2459: Bulk operations do not scale

gsnider2195:
# Closes #2459
# What's Changed

## Global change logging behavior changes

- A cache of all change logging operations are stored in the thread local change context, to allow us to skip database hits when looking up related changes when none exist

## Deferred change logging for bulk edit

- `nautobot.extras.context_managers.deferred_change_logging_for_bulk_operation` context manager introduced for use with bulk edit
  - Defers change logging to the end of the context manager and uses `bulk_create` to create the `ObjectChange` instances, slightly improving performance

## Deferred change logging for bulk delete

- `nautobot.extras.utils.bulk_delete_with_bulk_change_logging` function introduced to handle bulk change log creation for bulk delete. This couldn't be a context manager because we need to have the instance so we can create the ObjectChange before the delete happens

# Screenshots

Test results for creating 100 namespaces, then updating the description on those namespaces 50 times and averaging the results:

image

image

image

I also prototyped deferred change logging for bulk delete but it needs a different code path (around 30 LOC):

image

image

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/utils.py
```diff
@@ -610,3 +612,35 @@ def migrate_role_data(
...
+            for obj in qs.iterator():
+                if not hasattr(obj, "to_objectchange"):
+                    break
+                oc = obj.to_objectchange(ObjectChangeActionChoices.ACTION_DELETE)
+                oc.user = change_context.user
+                oc.request_id = change_context.change_id
+                oc.change_context = change_context.context
+                oc.change_context_detail = change_context.context_detail[:CHANGELOG_MAX_CHANGE_CONTEXT_DETAIL]
+                queued_object_changes.append(oc)
+            ObjectChange.objects.bulk_create(queued_object_changes)
```
gsnider2195, line 642: These should be created in batches, like the bulk edit

HanlinMiao (approved):

HanlinMiao (approved):

changes/2459.changed
```diff
@@ -0,0 +1 @@
+Improved performance of bulk-edit operations by refactoring change logging logic.
```
glennmatthews, line 1:
```suggestion
Improved performance of bulk-edit and bulk-delete UI operations by refactoring change logging logic.
```

nautobot/extras/utils.py
```diff
@@ -610,3 +612,38 @@ def migrate_role_data(
...
+                    ObjectChange.objects.bulk_create(queued_object_changes)
+                    queued_object_changes = []
+                oc = obj.to_objectchange(ObjectChangeActionChoices.ACTION_DELETE)
+                oc.user = change_context.user
+                oc.request_id = change_context.change_id
+                oc.change_context = change_context.context
+                oc.change_context_detail = change_context.context_detail[:CHANGELOG_MAX_CHANGE_CONTEXT_DETAIL]
+                queued_object_changes.append(oc)
+            ObjectChange.objects.bulk_create(queued_object_changes)
+            return qs.delete()
```
glennmatthews, line 646: I'll bet you can get better performance here by temporarily unhooking the `_handle_deleted_object` signal altogether, similar to what was done in #5442?

gsnider2195: We would need to manually delete ContactAssociations and Notes and increment metrics here if we disconnect the signal. Is the delete signal making Django do extra checks on delete?

glennmatthews: Good points, out of scope for this PR then. But the short answer is yes, it does.

glennmatthews (approved):

Issue #5407, opened 2024-03-08
Relative path Issue when injecting custom jobs into nautobot 


sslhansen:
Its pretty hard to follow the issue template for this bug. Will try my best :) 
Here is the [slack](https://networktocode.slack.com/archives/C01NWPK6WHL/p1709641144928579) thread with discussion providing more info on this issue
### Environment
* Nautobot version (Docker tag too if applicable): Tested on 2.1.0
* Python version: 3.10
* Database platform, version: Postgres, 13
* Middleware(s):

### Steps to Reproduce
Job is using relative path imports.
1. Sync job to our production nautobot instance using the git repo extensibility feature. Job works fine
2. Clone the same repo to local machine
3. Copy the contents of /jobs to JOBS_ROOT in developer environment instead of using git repo extensibility. This breaks

### Expected Behavior
Nautobot should recognize and be able to run the jobs.

### Observed Behavior
When containers are starting up, nautobot fails with:
![image](https://github.com/nautobot/nautobot/assets/56868370/4df58980-4844-4027-add7-fd7f6e1105d3)

Our tree structure of /jobs is:
![image](https://github.com/nautobot/nautobot/assets/56868370/0ce7516d-8322-49a7-825a-eb76583e943a)

gsnider2195:
If you're running this in `JOBS_ROOT` you can add a `__init__.py` to your sub-directories and switch to absolute module names. So instead of `from ..models.racks.delete_full_rack import delete_full_rack` you would use `from models.racks.delete_full_rack import delete_full_rack`

gsnider2195:
Here's the relevant docs https://docs.nautobot.com/projects/core/en/stable/development/jobs/migration/from-v1/?h=jobs_root#jobs-in-jobs_root

sslhansen:
We did this initially, but that created an error when we tried the git import.
![image](https://github.com/nautobot/nautobot/assets/56868370/67ce4987-49d0-4c78-ad93-ebd2a0bad603) 
No module named 'models'. Then @glennmatthews suggested for us to use relative paths.(I am sorry for the ping, thought it might be relevant to include you here).

Unfortunately i can't reproduce the error that happens with absolute paths right now. The reason for that is also quite "fun", i will add that here as well:
First, i injected the jobs via jobs_root. This works.
[...]

Issue #5094, opened 2024-01-12
Jobs Docs: Document Reserved Keywords Such as "Name"


jvanderaa:
### Proposed Changes

Update docs for Jobs in the Jobs Developer Guides.

### Justification

Was getting an error of:

```
TypeError: object of type 'StringVar' has no len()
```

Such as an example job:

```python
class MyJob(Job):
    """Job to create a specified Rack."""

    name = StringVar(label="Rack Name", description="Name of the rack")
```

jvanderaa:
Unless they should be somewhere else - https://docs.nautobot.com/projects/core/en/stable/development/jobs/?h=stringvar#variables

jvanderaa:
It was found that `name` appears to be a reserved keyword in this, so would like to have the list of reserved keywords that we shouldn't use.

glennmatthews:
Fair request. There are unfortunately a lot of these, some inherent to the base Celery `Task` class and more added by Nautobot's `BaseJob` subclass.

It'd be a breaking change, but I'd prefer to address this long-term by moving the user-input variables into an explicit attribute grouping of their own, rather than having them declared at class-level, which currently requires Nautobot to always introspect every class attribute to see which ones are subclasses of `ScriptVariable` and has this risk of accidental clobbering as you describe above. Something like:

```python
class MyJob(Job):
    class Meta:
        name = "My Job"
        description = "Job to create a specified Rack."
        variables = {
            "name": StringVar(...)
        }
```

glennmatthews:
Related: #4761.

glennmatthews:
Just going by `BaseJob`, without looking upstream to `Task`, we currently have:

```
__init__
__call__
__str__
on_bound
shadow_name
before_start
run
on_success
on_retry
on_failure
after_return
apply
file_path
class_path
class_path_dotted
class_path_js_escaped
grouping
_get_meta_attr_and_assert_type
name
description
description_first_line
dryrun_default
hidden
field_order
read_only
approval_required
soft_time_limit
time_limit
has_sensitive_variables
supports_dryrun
task_queues
properties_dict
registered_name
_get_vars
_get_file_vars
as_form_class
as_form
clear_cache
job_model
job_result
celery_kwargs
user
serialize_data
deserialize_data
validate_data
prepare_job_kwargs
_load_file_from_proxy
_save_file_to_proxy
_delete_file_proxies
load_yaml
load_json
create_file
```

Issue #4498, opened 2023-09-21
multiple functions named register_jobs


gsnider2195:
### Proposed Changes

One of these should be renamed

- `nautobot.extras.plugins.register_jobs`
- `nautobot.core.celery.register_jobs`

### Justification

It's confusing having these named the same

jathanism:
This is my fault and was not meant to persist out of prototype. The former pattern is what inspired the latter. The only thing really differs is `plugins.register_jobs` adds the job model to `plugin_jobs` in the Nautobot registry, and `celery.register_jobs` calls `app.register_task()` to add to the Celery registry.

Ideally new should just unify them if we can, or if we leave them as-is, simply rename them (e.g. `register_plugin_jobs` and `register_celery_jobs` respectively).

gsnider2195:
I think the original idea was to make BaseJob a meta class and make the subclass automatically register in celery but we haven't been able to implement that yet 😦

PR #5586, opened 2024-04-16
Reintroduce run_job task, make Job no longer a Task subclass

### LINKED ISSUE #4889: Misleading warning when nautobot is unable to find the jobs in register_jobs() call during jobs repo sync
### Environment
* Nautobot version 2.0.2:
* Python version: 3.11.2
* Database platform, version: psql (PostgreSQL) 10.23
* Middleware(s): non-docker install on RHEL8

### Steps to Reproduce
Need to make sure the job is not found in order to replicate the error. For example:
1. Create a simple job under jobs/ directory, like a simple helloWorldJob job class instance in the file hello.py.
2. Under jobs/__init__.py, import the helloWorldJob job class instance **incorrectly** to make sure it is not found.. for example, give it an incorrect path to import from. Then call register_jobs() and pass helloWorldJob object.
4. If you sync the jobs repo, you would get a warning that says: "No jobs were registered on loading the jobs submodule. Did you miss a register_jobs() call?"

### Expected Behavior
I was expecting it to give out a different error because I already registered a job on the jobs submodule.  Maybe raise an output like "Unable to register job helloWorldJob . Make sure the import path to the job class is correct." would be more helpful to help users better debug their scripts.

### Observed Behavior
If you sync the jobs repo, you would get a warning that says: "No jobs were registered on loading the jobs submodule. Did you miss a register_jobs() call?"

### LINKED ISSUE #5063: Git detached HEAD in case you have more than one worker - v.2.1.0 in K8S
### Environment
* Nautobot version (Docker tag too if applicable): 2.1.0(Docker image:2.1.0-py3.11)
* Python version: 3.11
* Database platform, version: Postges, 15.4.0-debian-11-r45
* Middleware(s): N/A

### Steps to Reproduce
1. Deploy more than 1 worker
2. Create a Git repository to store the jobs
3. Change the git repository(e.g. add new code)
4. Start Git repository Sync
5. Repository is synced on all workers(based on the hash) but HEAD is detached on all workers except the worker that started the sync job
6. Start a job that is part of the Jobs repository - If job is moved to the worker that is in detached HEAD state, it is executing the previous commit, not the last one, but if the job is moved to the worker that synced correctly, latest code version has been executed

### Expected Behavior
1. Deploy more than 1 worker
2. Create a Git repository to store the jobs
3. Change the git repository(e.g. add new code)
4. Start Git repository Sync
5. Repository is synced on all workers
6. Start a job that is part of the Jobs repository - no matter on which worker, this job will be runned, latest version of the code will be executed

### Observed Behavior
I noticed a really strange behaviour during the testing of version 2.1.0 in K8S environment. 
I used the latest Helm chart from the repo to deploy Nautbot v2.1.0 to K8S.

We have a Jobs repository where all of our jobs are stored. As we have to update the scripts/jobs to be adjusted to version 2(previously we ran them on version 1.6.2), we noticed not all workers updating the repository correctly and thus some workers are not executing the latest version of the code.

We have 4 workers in default queue. I have scaled down the deployment to single worker and issue is not observed. I decide to add one more worker after that(2 deployments) and issue has 50% chance to happen. My findings:
- Update a script in our Jobs repo
- Sync the repository using the standard GitSync job in Nautobot
- On worker 1(logs):

```
12:31:21.717 INFO    nautobot.extras.datasources.git :                                                                                                                                    Refreshed Job record                                                                                                                                                                                   
[2024-01-08 12:31:21,717: INFO/ForkPoolWorker-6] Refreshed Job record                                                                                                                                    
[2024-01-08 12:31:21,736: INFO/ForkPoolWorker-6] nautobot.core.jobs.GitRepositorySync[e1e4bd25-5815-4081-ba64-e16c21dda45e]: Repository synchronization completed in None                                                                                                                                                                                                                                                                                                                                                                                                                                            
12:31:21.847 INFO    nautobot.core.celery.control :                                                                                                                                                      
The current Git repository hash is "15443c1a2ef3dd988d908d9b01f81cec9ecb77df"                                                                                                                          
[2024-01-08 12:31:21,847: INFO/MainProcess] The current Git repository hash is "15443c1a2ef3dd988d908d9b01f81cec9ecb77df"                                                                               
[2024-01-08 12:31:21,893: INFO/ForkPoolWorker-6] Task nautobot.core.jobs.GitRepositorySync[e1e4bd25-5815-4081-ba64-e16c21dda45e] succeeded in 1.0523398811928928s: None                                
[2024-01-08 12:31:21,893: INFO/ForkPoolWorker-6] nautobot.core.jobs.GitRepositorySync[e1e4bd25-5815-4081-ba64-e16c21dda45e]: Job completed

Repository has been updated successfully and HEAD is okay:
nautobot@nautobotv2-celery-default-c456f4d76-2sj2m:~/git/nautobot_jobs_version2$ git status
On branch master
Your branch is up to date with 'origin/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	jobs/L2Interfaces/__pycache__/
	jobs/Onboarding/__pycache__/
	jobs/__pycache__/

nothing added to commit but untracked files present (use "git add" to track)
```

- on worker 2(logs):

```
12:31:22.240 INFO    nautobot.core.utils.git :                                                                                                                                                           
Checking out commit `15443c1a2ef3dd988d908d9b01f81cec9ecb77df` on branch `master`...                                                                                                                   
[2024-01-08 12:31:22,240: INFO/MainProcess] Checking out commit `15443c1a2ef3dd988d908d9b01f81cec9ecb77df` on branch `master`...                                                                         
12:31:22.244 INFO    nautobot.core.celery.control :                                                                                                                                                       Repository successfully refreshed                                                                                                                                                                      
[2024-01-08 12:31:22,244: INFO/MainProcess] Repository successfully refreshed                                                                                                                            
12:31:22.244 INFO    nautobot.core.celery.control :                                                                                                                                                      The current Git repository hash is "15443c1a2ef3dd988d908d9b01f81cec9ecb77df"                                                                                                                          
[2024-01-08 12:31:22,244: INFO/MainProcess] The current Git repository hash is "15443c1a2ef3dd988d908d9b01f81cec9ecb77df"

As we can see, the commit is the same one, but looking at the git HEAD i can see that it is detached now and pointing to the latest commit but in detached head state:

nautobot@nautobotv2-celery-default-c456f4d76-lcnrt:~/git/nautobot_jobs_version2$ git status
HEAD detached at 15443c1
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	jobs/L2Interfaces/__pycache__/
	jobs/Onboarding/__pycache__/
	jobs/__pycache__/

nothing added to commit but untracked files present (use "git add" to track)
```

And looking at the logs of the Git:

```
nautobot@nautobotv2-celery-default-c456f4d76-lcnrt:~/git/nautobot_jobs_version2$ git log --oneline
15443c1 (HEAD, origin/master, origin/HEAD) .
2f071d5 (master) .
63c9119 .
dead8df .
```

As we can see on the second worker, master is pointing to the previous commit. And basically this is causing the issue, as the celery will assign a job to be executed on worker 2, and it will use the previous commit, not the last one, and if job is assigned to worker 1, it will use the latest commit.

Behavior is not strict to worker 1 and worker 2, as it really depends on the celery to which worker the Git repo sync job will be assigned. The first worker will always pull the repo in the correct way, any other worker will always git checkout to the latest commit, which is basically causing the detached HEAD state(or at least this is my opinion, not a Git expert).

### LINKED ISSUE #5498: 2.0 Jobs documentation: celery Task instantiation and storing state on Job instances
We need to document that storing instance variables on Job classes is no longer supported in 2.x due to the way celery only instantiates a single instance of a Task. eg:

```py
class MyJob(Job):
    def run(self, device):
        self.device = device
        self.do_stuff()

    def do_stuff(self):
        device = self.device
```

### LINKED ISSUE #5537: Rollback Jobs as Celery Singleton Classes
Turns out the juice was not worth the squeeze in turning Jobs into Celery native task instances. We will rollback to the 1.x architecture of having a celery task which handles the setup, run, and teardown of job executions.

AC:
- Don't break the existing `JobResult.enqueue_job()` pattern
- django celery results should remain, so that we still properly track state
- Access to results should remain the same
- Fixes these issues:
  - #5063
  - #5498
  - #5407
  - #5094
  - #4889
  - #4498 

Stretch goal:
- Implement sub process wrapper(s) to intercept all logging/stdout/stderr
  - If not achieved in this issue, we will address it as a follow on.

glennmatthews:
# Closes #5537, etc.
# What's Changed

- `nautobot.extras.jobs.Job` is once again a standalone class rather than a subclass of celery's `Task` class. (fixes #5498)
    - We no longer enqueue Jobs directly into Celery; instead the `run_job` Celery task is reintroduced as a wrapper for executing any Job. (part of fixing #5537)
        - Changed `JobResult.enqueue_job()` accordingly
        - Changed how `NautobotScheduleEntry` creates scheduler entries to "hide" the underlying implementation change from the end user. This avoids the need for any data migration of existing `ScheduledJob` records.
        - Likewise, changed how `NautobotDatabaseBackend` creates `JobResult` records to also hide this change and preserve existing behavior.
    - We unfortunately still need to override Celery's base `Task.apply()` function to make it pass through our extra attributes. This only impacts unit tests and the `nautobot-server runjob --local` command, but it'd still be a breaking change if we didn't preserve this.
- Reintroduced 1.x's handling of JOBS_ROOT and Git-sourced jobs as (another) attempt at fixing #5063. Basically we're back to these code sources **not** being added permanently to `sys.path`.
    - Renamed `import_jobs_as_celery_tasks` signal to simply `import_jobs`, but it serves a similar purpose to before.
    - Re-added `get_jobs(reload=True)` function as a way to wrap the dynamic reloading of JOBS_ROOT and Git-sourced jobs
    - Changed behavior of `register_jobs()` API so that, instead of registering tasks with Celery, it now stores system and App job classes (only!) into the Nautobot registry for later retrieval.
        - Removed the duplicate `nautobot.extras.plugins.register_jobs()` API (fixing #4498).

# Screenshots

# TODO

- [x] Update error message on failing to load jobs from a Git repo - fixes #4889.
- [x] Test and verify that this fixes #5063 (most important!)
- [x] Test whether this change does anything for #5407 (I have doubts)
- [x] Docs updates to address #5094
- [x] Unit, Integration Tests


glennmatthews:
Lots of rework in the latest diffs, reintroducing more 1.x behaviors. Sorry for the churn here.

glennmatthews:
Used the test procedure described in https://github.com/nautobot/nautobot/issues/5063#issuecomment-2066550455 to verify Git jobs are correctly updated across multiple celery workers.

Additionally, CLI'd into the running celery worker and deleted its local clone of the git repository, then reran a Git-sourced job to verify the repo is correctly re-cloned and re-imported on the fly:

```
[2024-04-19 13:25:20,006: INFO/MainProcess] Task nautobot.extras.jobs.run_job[1caf51e3-a4f1-41e3-810c-5dae90681fd2] received
13:25:20.036 DEBUG   nautobot.extras.jobs jobs.py                                run_job() :
  Running job test.jobs.data_quality.VerifyHostnames
[2024-04-19 13:25:20,036: DEBUG/ForkPoolWorker-3] Running job test.jobs.data_quality.VerifyHostnames
13:25:20.041 DEBUG   nautobot.core.utils.git git.py                                __init__() :
  Cloning git repository to /opt/nautobot/git/test...
[2024-04-19 13:25:20,041: DEBUG/ForkPoolWorker-3] Cloning git repository to /opt/nautobot/git/test...
13:25:20.930 INFO    nautobot.core.utils.git git.py                                checkout() :
  Checking out commit `2357c2ab21f3cce6aa6043beba4e64b08d8712c0` on branch `nautobot-jobs-2.0`...
[2024-04-19 13:25:20,930: INFO/ForkPoolWorker-3] Checking out commit `2357c2ab21f3cce6aa6043beba4e64b08d8712c0` on branch `nautobot-jobs-2.0`...
13:25:20.945 INFO    nautobot.extras.jobs git.py                   ensure_git_repository() :
  Repository successfully refreshed
[2024-04-19 13:25:20,945: INFO/ForkPoolWorker-3] Repository successfully refreshed
13:25:20.945 INFO    nautobot.extras.jobs git.py                   ensure_git_repository() :
  The current Git repository hash is "2357c2ab21f3cce6aa6043beba4e64b08d8712c0"
[2024-04-19 13:25:20,945: INFO/ForkPoolWorker-3] The current Git repository hash is "2357c2ab21f3cce6aa6043beba4e64b08d8712c0"
[2024-04-19 13:25:20,958: INFO/ForkPoolWorker-3] nautobot.extras.jobs.run_job[1caf51e3-a4f1-41e3-810c-5dae90681fd2]: Running job
[2024-04-19 13:25:20,980: INFO/ForkPoolWorker-3] nautobot.extras.jobs.run_job[1caf51e3-a4f1-41e3-810c-5dae90681fd2]: Using the regular expression: .*
```

nautobot/core/celery/__init__.py
```diff
@@ -53,60 +43,74 @@ def register_task(self, task, **options):
...
+    """
+    import nautobot.core.jobs  # noqa: F401
+
+    _import_jobs_from_jobs_root()
 
-    Note that app-provided Jobs are automatically imported at startup time via NautobotAppConfig.ready()
+    try:
+        _import_jobs_from_git_repositories()
+    except ProgrammingError:  # Database not ready yet, as may be the case on initial startup and migration
+        pass
```
glennmatthews, line 59: This feels gross but it does seem to work and makes it possible for this function to run from the Nautobot main server process even before migrations have been completed.

nautobot/core/celery/task.py
```diff
@@ -1,12 +1,85 @@
...
+from celery.result import EagerResult
+from celery.utils.functional import maybe_list
+from celery.utils.nodenames import gethostname
+from kombu.utils.uuid import uuid
 
 
 class NautobotTask(Task):
     """Nautobot extensions to tasks for integrating with Job machinery."""
 
+    def apply(
```
glennmatthews, line 13: This is just moved from `BaseJob.apply` to `NautobotTask.apply` since BaseJob isn't a task any more.

nautobot/core/testing/api.py
```diff
@@ -13,12 +13,11 @@
 from rest_framework.relations import ManyRelatedField
 from rest_framework.test import APITransactionTestCase as _APITransactionTestCase
 
-from nautobot.core import testing
```
glennmatthews, line 16: I had an intermittent issue where this self-referential import (`nautobot.core.testing.[api|mixins|views]` importing `nautobot.core.testing`) was causing the test startup to abort. All of the relevant imports to these three modules are really coming from `core.testing.utils` in any case, and using the more specific import seems to fix that issue. 🤷

nautobot/extras/models/jobs.py
```diff
@@ -278,13 +280,13 @@ def git_repository(self):
...
-                state=None, repository_pk=self.git_repository.pk, head=self.git_repository.current_head
-            )
-        return app.tasks[f"{self.module_name}.{self.job_class_name}"]
+        """Get an instance of the associated Job class, refreshing it if necessary."""
+        from nautobot.extras.jobs import get_job
+
+        try:
+            return get_job(self.class_path, reload=True)()
+        except TypeError as err:  # keep 2.0-2.2.1 exception behavior
+            raise NotRegistered from err
```
glennmatthews, line 289: Need to double-check this - `get_job()` returns None if if doesn't find the job class, it doesn't normally raise a `TypeError`. Probably good to add a test case exercising this logic.

glennmatthews: Never mind - because we try to instantiate the job class, calling `None()` is what raises the `TypeError`. This is covered by `assert_job_exists()` in `test_datasources`.

nautobot/core/celery/__init__.py
```diff
@@ -53,60 +43,74 @@ def register_task(self, task, **options):
...
-    git_root = settings.GIT_ROOT
-    if git_root and os.path.exists(git_root):
-        if git_root not in sys.path:
-            sys.path.append(git_root)
-
-        # We can't detect which Git directories we're *supposed* to auto-load Jobs from if we can't read GitRepository
-        # records from the DB, unfortunately.
-        # We work around this in JobModel.job_task to try later loading Git jobs on-the-fly if needed.
-        if database_ready:
+    if not (settings.JOBS_ROOT and os.path.isdir(settings.JOBS_ROOT)):
```
gsnider2195, line 66: nitpick
```suggestion
    if not all((settings.JOBS_ROOT, os.path.isdir(settings.JOBS_ROOT))):
```

nautobot/core/celery/__init__.py
```diff
@@ -53,60 +43,74 @@ def register_task(self, task, **options):
...
+        return
+
+    # Flush any previously loaded non-system, non-App Jobs
+    for job_class_path in list(registry["jobs"]):
+        if job_class_path.startswith("nautobot."):
```
gsnider2195, line 71: Since we're storing the jobs in the jobs registry, would it make sense to also store the job source to make this simpler?

nautobot/core/celery/__init__.py
```diff
@@ -53,60 +43,74 @@ def register_task(self, task, **options):
...
+        # Else, it's presumably a JOBS_ROOT job
+        del registry["jobs"][job_class_path]
+
+    # Load all modules in JOBS_ROOT
+    import_modules_privately(path=os.path.realpath(settings.JOBS_ROOT))
+
+
+def _import_jobs_from_git_repositories():
+    git_root = os.path.realpath(settings.GIT_ROOT)
+    if not (git_root and os.path.exists(git_root)):
```
gsnider2195, line 94:
```suggestion
    if not all((git_root, os.path.exists(git_root))):
```

nautobot/core/celery/__init__.py
glennmatthews: It'd just move the logic from here to the `register_jobs()` function, I think - there's no built-in knowledge of "job source" in the register_jobs call.

nautobot/core/celery/__init__.py
```diff
@@ -53,60 +43,74 @@ def register_task(self, task, **options):
...
+            and not GitRepository.objects.filter(slug=filename).exists()
+        ):
+            logger.warning("Deleting unmanaged (leftover?) Git repository clone at %s", filepath)
+            shutil.rmtree(filepath)
 
-            # Make sure all GitRepository records that include Jobs have up-to-date git clones, and load their jobs
-            for repo in GitRepository.objects.all():
-                refresh_git_repository(state=None, repository_pk=repo.pk, head=repo.current_head)
+    # Make sure all GitRepository records that include Jobs have up-to-date git clones, and load their jobs
+    for repo in GitRepository.objects.all():
```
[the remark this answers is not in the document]
glennmatthews: Filtering would definitely make sense, thanks. I'll give it a try.

nautobot/core/celery/backends.py
```diff
@@ -54,14 +55,21 @@ def _get_extended_properties(self, request, traceback):
             if traceback is not None:
                 traceback = sanitize(traceback)
 
+            # Preserve the JobResult data behavior from Nautobot 2.0 through 2.2 (wherein the Job itself was the task)
+            # by manipulating `task_name` and `task_args` to hide the fact that we are now calling
+            # `run_job.apply(args=[JobClass.class_path, ...])` instead of `JobClass.apply(args=[...])`.
+            if task_name == "nautobot.extras.jobs.run_job" and task_args:
+                task_name = task_args[0]
+                task_args = task_args[1:]
```
gsnider2195, line 63:
```suggestion
                task_name = task_args.pop(0)
```

nautobot/extras/datasources/git.py
```diff
@@ -716,56 +715,41 @@ def delete_git_config_context_schemas(repository_record, job_result, preserve=()
...
     """
-    After cloning/updating a GitRepository on disk, call this function to reload and reregister the repo's Python code.
+    After cloning/updating/deleting a GitRepository on disk, call this function to reload and reregister its Python.
 
     Args:
-        repository_slug (str): Repository directory in GIT_ROOT that was refreshed.
-        consumer (celery.worker.Consumer): Celery Consumer to update as well
+        repository_slug (str): Repository directory in GIT_ROOT that was updated or deleted.
         skip_reimport (bool): If True, unload existing code from this repository but do not re-import it.
+        should_raise (bool): If True, any exceptions raised in the import will be re-raised.
```
gsnider2195, line 725: Would it be better to call this kwarg `ignore_import_errors` here and in `import_modules_privately`?

nautobot/core/jobs/__init__.py
```diff
@@ -50,7 +50,8 @@ def run(self, repository):
             # Given that the above succeeded, tell all workers (including ourself) to call ensure_git_repository()
             app.control.broadcast("refresh_git_repository", repository_pk=repository.pk, head=repository.current_head)
```
gsnider2195, line 51: Is this still necessary with the changes you've made or is it just extra noise? If this fails on one of the workers we don't get any feedback anywhere except the worker console logs right? If so this doesn't function well as a warning that your worker wasn't able to clone the git repo and I can't think of another reason to keep this?

gsnider2195: Also noting that this causes the worker that ran this job to reload the git repo code a second time.

nautobot/core/testing/api.py
gsnider2195: Thank you for explaining some of these changes that aren't obvious. Makes my life so much easier as a reviewer!

nautobot/core/celery/__init__.py
gsnider2195: Does the `database_ready` kwarg not work anymore to avoid having to do this?

nautobot/core/celery/__init__.py
gsnider2195: Just to explain myself, I was traumatized by perl's `unless` statement. What does `unless apple is fruit or apple is vegetable` mean? 🤷

nautobot/core/celery/__init__.py
gsnider2195: Do we have any docs on the difference between the jobs registry and the Job model in the database? We use the jobs registry as an in memory cache if I remember our earlier conversation correctly?

nautobot/core/celery/backends.py
gsnider2195: This won't work if `task_args` is a tuple. Would that ever be the case?

nautobot/core/utils/module_loading.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def import_modules_privately(path, module_path=None, should_raise=False):
+    """
+    Import modules from the filesystem without adding the path permanently to `sys.path`.
+
+    This is used for importing Jobs from `JOBS_ROOT` and `GIT_ROOT` in such a way that they remain relatively
+    self-contained and can be easily discarded and reloaded on the fly.
+
+    If you find yourself writing new code that uses this method, please pause and reconsider your life choices.
```
gsnider2195, line 34: 💯

nautobot/core/utils/module_loading.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Args:
+        path (str): Directory path possibly containing Python modules or packages to load.
+        module_path (list): If set to a non-empty list, only modules matching the given chain of modules will be loaded.
+            For example, `["my_git_repo", "jobs"]`.
+        should_raise (bool): Whether any exception raised in importing modules should be re-raised to the caller.
```
gsnider2195, line 40: Same question as earlier, is this more clear if it's called `ignore_import_errors` instead of `should_raise`?

nautobot/core/utils/module_loading.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+    if module_path is None:
+        module_path = []
+        module_prefix = None
+    else:
+        module_prefix = ".".join(module_path)
+    with _temporarily_add_to_sys_path(path):
+        for finder, discovered_module_name, is_package in pkgutil.walk_packages([path], onerror=logger.error):
+            # logger.debug("Discovered module %s", discovered_module_name)
```
gsnider2195, line 49: Remove?

nautobot/core/celery/__init__.py
glennmatthews: IIRC it didn't work exactly to our needs, for similar reasons to why we had to invent the `nautobot_database_ready` signal - if I remember what I encountered correctly, the issue is that during first-time startup of a fresh install, `AppConfig.ready()` gets called before migrations run (so it'd be `database_ready=False`) , but the same function gets called (with the same args) when restarting Nautobot in a steady-state scenario (so it'd be `database_ready=True`). Django doesn't provide any obvious flag that I've ever been able to find to readily distinguish between these two states, so we either need a try/except pattern like here or else we'd need to call into the Django migration code to actually query whether migrations have completed. 

We can't use the `nautobot_database_ready` signal to load jobs into memory because that only gets called specifically in response to `migrate` or `post_upgrade` commands and we need to load jobs into memory on normal server startup.

nautobot/core/celery/__init__.py
glennmatthews: I was hoping the parens were clear enough :) But I don't mind making this change either.

nautobot/core/celery/__init__.py
glennmatthews: I'll see what I can add. The Job model says "A job exists or existed at some point" but doesn't store the executable code of the Job. The registry says "Here's the source code for this Job so you can actually run it".

glennmatthews: But yeah, the registry is being used as a cache so we don't have to re-import the job just to introspect its existence and such like that - we *do* have to always re-import it when we go to actually run it so as to be confident that we're running the right version of the job.

nautobot/core/celery/backends.py
glennmatthews: I think it's always a list (since we're passing things around in JSON) but good question.

nautobot/core/utils/module_loading.py
glennmatthews: Yeah probably. I found this useful when I was developing/troubleshooting this code but it was too noisy to have it output every time we reloaded jobs. Thoughts on removing it versus adding a `debug=False` parameter to the function to gate these logs?

nautobot/core/jobs/__init__.py
gsnider2195: Do we use git repositories to drop arbitrary files on the filesystem of workers? If so, then I think we should keep this.

nautobot/core/celery/__init__.py
gsnider2195: Like I said, it's a nitpick so I'm fine either way

nautobot/core/utils/module_loading.py
glennmatthews: Correct. Just didn't want to make the perceived problem in #4292 any worse. :)

nautobot/core/celery/__init__.py
glennmatthews: Actually the change won't work since `all()` doesn't short-circuit logic:

```python
>>> JOBS_ROOT = None
>>> import os.path
>>> bool(JOBS_ROOT and os.path.isdir(JOBS_ROOT))
False
>>> bool(all((JOBS_ROOT, os.path.isdir(JOBS_ROOT))))
  File "/usr/local/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/genericpath.py", line 42, in isdir
    st = os.stat(s)
TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
```

nautobot/core/utils/module_loading.py
gsnider2195: I'd prefer to just comment it out personally. If developers don't want debug logging they should change their logging levels?

nautobot/core/jobs/__init__.py
gsnider2195: Probably better to keep it for now instead of breaking something. If it becomes an issue we can open another story for it?

nautobot/core/celery/__init__.py
glennmatthews: Added some more doc updates in the latest commits.

nautobot/core/celery/backends.py
glennmatthews: I think I'll leave this alone just to be safe, as I can't tell from a quick read of the Celery code whether there's any particular type assertion here.

nautobot/core/celery/__init__.py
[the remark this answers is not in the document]
gsnider2195: Were you able to try this filter?

nautobot/extras/datasources/git.py
```diff
@@ -716,56 +715,41 @@ def delete_git_config_context_schemas(repository_record, job_result, preserve=()
 #
 
 
-def refresh_code_from_repository(repository_slug, consumer=None, skip_reimport=False):
+def refresh_code_from_repository(repository_slug, skip_reimport=False, should_raise=False):
```
gsnider2195, line 718: Can we rename this function or update the docstring to state that this is only reloading jobs code? There's a lot of code in a git repository that could potentially be refreshed.

nautobot/extras/datasources/git.py
```diff
@@ -716,56 +715,41 @@ def delete_git_config_context_schemas(repository_record, job_result, preserve=()
...
+
+    try:
+        repository = GitRepository.objects.get(slug=repository_slug)
+        if "extras.job" in repository.provided_contents:
+            if not (
+                os.path.isdir(os.path.join(repository.filesystem_path, "jobs"))
+                or os.path.isfile(os.path.join(repository.filesystem_path, "jobs.py"))
+            ):
+                if should_raise:
+                    raise FileNotFoundError(f"No `jobs` subdirectory or file found in Git repository {repository}")
```
gsnider2195, line 743: Perhaps?

```suggestion
                    raise FileNotFoundError(f"No `jobs` module found in Git repository {repository}")
```

nautobot/extras/models/jobs.py
```diff
@@ -278,13 +280,13 @@ def git_repository(self):
...
-            refresh_git_repository(
-                state=None, repository_pk=self.git_repository.pk, head=self.git_repository.current_head
-            )
-        return app.tasks[f"{self.module_name}.{self.job_class_name}"]
+        """Get an instance of the associated Job class, refreshing it if necessary."""
+        from nautobot.extras.jobs import get_job
+
+        try:
+            return get_job(self.class_path, reload=True)()
+        except TypeError as err:  # keep 2.0-2.2.1 exception behavior
```
gsnider2195, line 288: Should this be `2.0-2.2.2`?
```suggestion
        except TypeError as err:  # keep 2.0-2.2.2 exception behavior
```

nautobot/core/celery/__init__.py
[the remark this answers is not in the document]
glennmatthews: Yes, seems to work.

nautobot/extras/signals.py
```diff
@@ -471,15 +471,12 @@ def refresh_job_models(sender, *, apps, **kwargs):
...
+    import_jobs()
 
     job_models = []
-    for task in app.tasks.values():
-        # Skip Celery tasks that aren't Jobs
-        if not isinstance(task, JobClass):
-            continue
 
-        job_model, _ = refresh_job_model_from_job_class(Job, task.__class__)
+    for job_class in get_jobs().values():
```
gsnider2195, line 478: You could remove the above `import_jobs()` if you just pass `reload=True` to this?

```suggestion
    for job_class in get_jobs(reload=True).values():
```

glennmatthews: True. I was trying to keep the diffs ever-so-slightly easier to review. :wink:

PR #5590, opened 2024-04-17
Fixed upstream testing workflow

gsnider2195:
# Closes #DNE
# What's Changed

Fix upstream testing workflows showing successful when one of the steps fail

# Screenshots

## Before

https://github.com/nautobot/nautobot-app-ssot/actions/runs/8610217927

image

## After

https://github.com/gsnider2195/nautobot-app-ssot-gha-test/actions/runs/8724938540

image

https://github.com/gsnider2195/nautobot-app-ssot-gha-test/actions/runs/8725146406

image

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/.github/workflows/plugin_upstream_testing_base.yml
+++ b/.github/workflows/plugin_upstream_testing_base.yml
@@ -16,7 +16,6 @@ env:
 
 jobs:
   unittest:
-    continue-on-error: true
     strategy:
       fail-fast: false
       matrix:
--- a/changes/5590.housekeeping
+++ b/changes/5590.housekeeping
@@ -1,0 +1,1 @@
+Fixed upstream testing workflows showing successful when one of the steps fail.


gsnider2195:
@bryanculver do you see any issues with this? You're the expert on this workflow 😄 

For reference: https://www.edwardthomson.com/blog/github_actions_6_fail_fast_matrix_workflows.html

glennmatthews:
A concern as I understand it was whether failures in repo A would cause runs in-progress in repo B to be auto-cancelled instead of running to completion. Have you been able to test that behavior?

gsnider2195:
> A concern as I understand it was whether failures in repo A would cause runs in-progress in repo B to be auto-cancelled instead of running to completion. Have you been able to test that behavior?

I'm not seeing anywhere in our upstream testing where we could affect the workflows of another repository

bryanculver:
The only side-effect I can forsee (which may have arisen from testing) is that any task trying to be run after the matrix (either here or in the use of this, see: https://github.com/nautobot/nautobot-app-ssot/blob/479d77feb1ca75216f1ab41611d4211778e3e3bc/.github/workflows/upstream_testing.yml#L8-L13) would not be allowed to run.

But my memory is fuzzy on what or why this was put in so 🤷. If it wasn't documented it must not be important.

gsnider2195:
> The only side-effect I can forsee (which may have arisen from testing) is that any task trying to be run after the matrix (either here or in the use of this, see: https://github.com/nautobot/nautobot-app-ssot/blob/479d77feb1ca75216f1ab41611d4211778e3e3bc/.github/workflows/upstream_testing.yml#L8-L13) would not be allowed to run.
> 
> But my memory is fuzzy on what or why this was put in so 🤷. If it wasn't documented it must not be important.

Do we know of any apps that are extending this workflow to run any additional steps?

bryanculver:
> > The only side-effect I can forsee (which may have arisen from testing) is that any task trying to be run after the matrix (either here or in the use of this, see: https://github.com/nautobot/nautobot-app-ssot/blob/479d77feb1ca75216f1ab41611d4211778e3e3bc/.github/workflows/upstream_testing.yml#L8-L13) would not be allowed to run.
> > But my memory is fuzzy on what or why this was put in so 🤷. If it wasn't documented it must not be important.
> 
> Do we know of any apps that are extending this workflow to run any additional steps?

Not that I know of. Could have been something that came up somewhere in this lifecycle:
- https://github.com/nautobot/nautobot/pull/2377
- https://github.com/nautobot/nautobot-plugin-netutils/pull/2

But YAGNI probably at this point.

HanlinMiao (approved):

gsnider2195:
I happened to have a PyGithub script laying around from some research I was doing on the upstream testing workflows earlier this week so I iterated through every github actions workflow with "upstream" in the name and checked the files. The only file that didn't match was https://github.com/nautobot/nautobot-plugin-ssot-servicenow/blob/develop/.github/workflows/upstream_testing.yml and that was just whitespace differences.

PR #5593, opened 2024-04-18
Fixed plugins not loading when using Gunicorn.

### LINKED ISSUE #5592: Plugins not loaded with Gunicorn
### Environment

* Nautobot version (Docker tag too if applicable): 2.2.1
* Python version: 3.11
* Middleware(s):

### Steps to Reproduce
1. Use systemd
2. With gunicorn 21.2.0 or 22.0.0

### Expected Behavior

All applications to show 

### Observed Behavior

We attempted to upgrade our dev environment from 2.1.9 to 2.2.1 but are hitting a weird issue where our plugins are reported as missing.  We are only loading 1 or 2 basic plugins right now while we work on updating all our other plugins for 2.x.  Oddly we are only seeing this issue on 1 out of 3 identical servers with identical Nautobot installs.

This looks very much like this issue from 2021: [Plugin Load Failure · Issue #95 · nautobot/nautobot (github.com)](https://github.com/nautobot/nautobot/issues/95)

gsnider2195:
# Closes #5592
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/5592.fixed
+++ b/changes/5592.fixed
@@ -1,0 +1,1 @@
+Fixed plugins not loading when using Gunicorn.
--- a/nautobot/core/wsgi.py
+++ b/nautobot/core/wsgi.py
@@ -1,11 +1,18 @@
 import logging
-import os
 
 from django.core import cache
 from django.core.wsgi import get_wsgi_application
 from django.db import connections
 
-os.environ["DJANGO_SETTINGS_MODULE"] = "nautobot_config"
+import nautobot
+
+# This is the Django default left here for visibility on how the Nautobot pattern
+# differs.
+# os.environ.setdefault("DJANGO_SETTINGS_MODULE", "nautobot.core.settings")
+
+# Instead of importing `DJANGO_SETTINGS_MODULE` we're using the custom loader
+# pattern from `nautobot.core.runner` to read environment or config path for us.
+nautobot.setup()
 
 # Use try/except because we might not be running uWSGI. If `settings.WEBSERVER_WARMUP` is `True`,
 # will first call `get_internal_wsgi_application` which does not have `uwsgi` module loaded


changes/5592.fixed
```diff
@@ -0,0 +1 @@
+Fixed plugins not loading when using Gunicorn.
```
glennmatthews, line 1: Anything we can/should add from a unit test perspective here?

gsnider2195: What are you thinking? Calling `nautobot.core.uwsgi` and checking for setup being called is the only viable test I can think of

glennmatthews: Was hoping that we could do something similar to the original issue, maybe calling gunicorn via `subprocess` or something? If not, no worries.

nautobot/core/wsgi.py
```diff
@@ -1,11 +1,18 @@
...
 
-os.environ["DJANGO_SETTINGS_MODULE"] = "nautobot_config"
+import nautobot
+
+# This is the Django default left here for visibility on how the Nautobot pattern
+# differs.
+# os.environ.setdefault("DJANGO_SETTINGS_MODULE", "nautobot.core.settings")
+
+# Instead of importing `DJANGO_SETTINGS_MODULE` we're using the custom loader
+# pattern from `nautobot.core.runner` to read environment or config path for us.
```
gsnider2195, line 14: `nautobot.core.runner` doesn't exist anymore so we'll need to update this comment

glennmatthews, line 14:
```suggestion
# Instead of just pointing to `DJANGO_SETTINGS_MODULE` and letting Django run with it,
# we're using the custom Nautobot loader code to read environment or config path for us.
```

Since `nautobot.core.runner` doesn't exist any more :)

glennmatthews (approved):

HanlinMiao (approved):

PR #5606, opened 2024-04-22
Add ability to inject custom HTML head tags (closes #1858)

### LINKED ISSUE #1858: Allow to customize CSS to differentiate multiple instances
### As ...

Patti - Platform Admin

### I want ...

to be able to customize the CSS (maybe the background-color but not just this) of a Nautobot instance

### So that ...

Users (operators, developpers, ..) are able to easily distinguish the different instances (dev, test, staging, prod, ...)

* An option to input CSS is present in the admin section
* This option is overridable using settings.py (to allow copying data from an instance to another while keeping this property)
* The CSS is applied on every pages

Currently, the only option is to set a banner, but having a banner on each instance with the same blue color doesn't help to differentiate without reading the banner text.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation

### Database Changes

* Add an option?

### External Dependencies

None ?

u1735067:
# Closes #1858
# What's Changed

Add the ability to inject custom HTML head tags (to customize the CSS style for example).

Unless a default value must be set for this change, the variable definition is left to the instance admin.

# Config example

`settings.py`
```py
if os.getenv('NAUTOBOT_CUSTOM_HEAD'):
    CUSTOM_HEAD = os.getenv('NAUTOBOT_CUSTOM_HEAD').strip()
elif os.getenv('NAUTOBOT_CUSTOM_HEAD_FILE'):
    with open(
        os.getenv('NAUTOBOT_CUSTOM_HEAD_FILE'),
        mode='r',
        encoding='utf-8'
    ) as fh:
        CUSTOM_HEAD = fh.read()
```

# TODO
- [ ] Unit, Integration Tests

### DIFF
--- a/changes/1858.added
+++ b/changes/1858.added
@@ -1,0 +1,1 @@
+Added ability to inject custom HTML head tags.
--- a/nautobot/core/templates/admin/base.html
+++ b/nautobot/core/templates/admin/base.html
@@ -45,6 +45,7 @@
             width: 100%;
         }
     </style>
+    {% if settings.CUSTOM_HEAD %}{{ settings.CUSTOM_HEAD | safe }}{% endif %}
 </head>
 <body {% if is_popup %}style="padding-left: 0px;"{% endif %}>
 
--- a/nautobot/core/templates/base_django.html
+++ b/nautobot/core/templates/base_django.html
@@ -8,6 +8,7 @@
     <title>{% block title %}Home{% endblock %} - {{ settings.BRANDING_TITLE }}</title>
     {% include 'inc/media.html' %}
     {% block extra_styles %}{% endblock %}
+    {% if settings.CUSTOM_HEAD %}{{ settings.CUSTOM_HEAD | safe }}{% endif %}
 </head>
 <body>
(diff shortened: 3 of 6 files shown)


gsnider2195:
Hi @u1735067 thanks for the submission! Can you provide an example of the html you'd like to inject with this setting?

u1735067:
Of course, for example : 
```py
CUSTOM_HEAD = r"<style>.navbar-default, .footer { background: linear-gradient(to right, #f8f8f8, moccasin); }</style>"
```
The primary goal is to allow some style customisation to differentiate multiple instances, but one might want to do some more generic things.

glennmatthews:
Thanks for the pull request! We're going to move forward at this time with a different approach to #1858, allowing for the provision of a custom CSS file via the `BRANDING_FILEPATHS` config setting rather than allowing injection of arbitrary HTML tags into `<head>`.

PR #5621, opened 2024-04-24
fix dependency on social-auth-core

### LINKED ISSUE #5565: Remove `openidconnect` Extra from Social Auth Core, Removed Upstream
We can remove `openidconnect` extra from Socaial Auth Core as they ditched the dependency on `jose`: https://github.com/python-social-auth/social-core/pull/819

Related: https://github.com/python-social-auth/social-core/issues/895

Our line here: https://github.com/nautobot/nautobot/blob/af15204ee7dedc741593e7f039e0abcc3d8b346b/pyproject.toml#L133

Without doing it, when installing `nautobot[sso]` you will receive the error:

```
social-auth-core 4.5.3 does not provide the extra 'openidconnect'
```

cardoe:
Since 4.5.0 of social-auth-core it has not had an extras of 'openidconnect' so drop it from the dependencies.
see https://github.com/python-social-auth/social-core/issues/895 fixes #5565

- [x] Unit, Integration Tests

### DIFF
--- a/changes/5565.fixed
+++ b/changes/5565.fixed
@@ -1,0 +1,1 @@
+Fixes optional dependency on `social-auth-core` by removing an extras related to openidconnect that no longer exists.
--- a/poetry.lock
+++ b/poetry.lock
@@ -4023,4 +4023,4 @@ sso = ["social-auth-core"]
 [metadata]
 lock-version = "2.0"
 python-versions = ">=3.8,<3.12"
-content-hash = "696a1070c64d48e829d98c3fcf7bbc9c7f3ceb83f1a69ecfb9e8258c47495cd8"
+content-hash = "598fd7b79a8cde691fb4952cf652fb7447144e48f9f496d0c16823f5ef0d410f"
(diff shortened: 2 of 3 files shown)


changes/5565.fixed
```diff
@@ -0,0 +1 @@
+fixes dependency on social-auth-core by removing an extras related to openidconnect that no longer exists
```
glennmatthews, line 1:
```suggestion
Fixed optional dependency on `social-auth-core` by removing an extras related to openidconnect that no longer exists.
```

glennmatthews (approved):
Looks good, thanks! I'm a bit surprised that it didn't result in any change to `poetry.lock` but I tried it locally (`poetry add --optional social-auth-core~4.5.3 --extras saml`) and that seems to be the case.

glennmatthews:
Spoke too soon :) I didn't notice the CI failures. Try `poetry add --optional social-auth-core~4.5.3 --extras saml` and see if that updates the poetry.lock to make CI happy.

changes/5565.fixed
cardoe: Thanks.

cardoe:
@glennmatthews should be rebased and pass cleanly

glennmatthews (approved):

PR #5627, opened 2024-04-25
[LTM] increases performance of `/api/dcim/devices?brief=true` endpoint

Kircheneer:
# Closes DNE
# What's Changed

Removes `select_related` and `prefetch_related` from queryset for API list view for devices when `brief=true` URL parameter is present. This is likely generalizable to other models but since this is LTM and this fix is not portable to 2.0 due to the `depth` changes I decided to not go there and just apply a targeted fix instead. 

This should help speed up `ObjectVar` and `MultiObjectVar` fields for the `Device` model.

# TODO
- [ ] Unit, Integration Tests


glennmatthews:
I think it would be more appropriate to actually make this a general fix, likely here:

https://github.com/nautobot/nautobot/blob/v1.6.19/nautobot/core/api/views.py#L198-L204

Kircheneer:
General fix it is!

glennmatthews (approved):

changes/5626.fixed
```diff
@@ -0,0 +1 @@
+Increased performance of `brief=true` in API endpoints by eliminating unnecessary joins.
```
gsnider2195, line 1:
```suggestion
Increased performance of `brief=true` in API endpoints by eliminating unnecessary database joins.
```

gsnider2195 (approved):

PR #5629, opened 2024-04-25
Add new template tag to resolve #5560

### LINKED ISSUE #5560: Add new template tag which hyperlinks an object, but opens in a new tab when clicked.
### As ...

P.D. - Plugin Developer

### I want ...

I want a template tag similar to `hyperlinked_object` which adds the `target="_blank"` attribute.

https://github.com/nautobot/nautobot/blob/594743bf49ade45899034a4e34c83af956036ca4/nautobot/core/templatetags/helpers.py#L46

### So that ...

The hyperlinked object opens in a new tab.  

I add a hyperlinked object using the new template tag and the object opens in a new tab.  

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation

### Database Changes

No database change required

### External Dependencies

No new dependencies required

tsm1th:
# Closes #5560 
# What's Changed
A new template tag was added that creates a hyperlink which opens the target in a new tab.

In a DRY effort, the original [hyperlinked_object](https://github.com/nautobot/nautobot/blob/594743bf49ade45899034a4e34c83af956036ca4/nautobot/core/templatetags/helpers.py#L46) tag was refactored to be more flexible and is called by the new template tag.

# Screenshots
image
image

# TODO

- [x] Unit, Integration Tests


nautobot/core/templatetags/helpers.py
```diff
@@ -43,7 +43,7 @@
 
 @library.filter()
 @register.filter()
-def hyperlinked_object(value, field="display"):
+def hyperlinked_object(value, field="display", **kwargs):
```
glennmatthews, line 46: I think it would be preferable to specifically enumerate `target` and `rel` as kwargs rather than having a catchall kwargs that could be used to provide any arbitrary HTML attribute, including unsafe ones.

nautobot/core/templatetags/helpers.py
```diff
@@ -73,11 +74,16 @@ def hyperlinked_object(value, field="display"):
...
+        if v is not None:
+            attributes.append(f'{k}="{v}"')
     display = getattr(value, field) if hasattr(value, field) else str(value)
     if hasattr(value, "get_absolute_url"):
+        attributes.insert(0, f'href="{value.get_absolute_url()}"')
         if hasattr(value, "description") and value.description:
-            return format_html('<a href="{}" title="{}">{}</a>', value.get_absolute_url(), value.description, display)
-        return format_html('<a href="{}">{}</a>', value.get_absolute_url(), display)
+            attributes.insert(1, f'title="{value.description}"')
+        return format_html("<a{}>{}</a>", mark_safe(" ".join(["", *attributes])), display)  # noqa: S308 # suspicious-mark-safe-usage Safe attributes
```
glennmatthews, line 86: Definitely let's not use `mark_safe` on an unknown set of attributes! I think it'd be better to use `format_html()` or `format_html_join()` (https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.html.format_html_join) to construct the individual entries in `attributes` and/or concatenate all of the `attributes` into a single safe string.

glennmatthews (commented):
Thanks for the PR! I have concerns about the security risk this poses as currently implemented, but I think they're addressable.

nautobot/core/templatetags/helpers.py
tsm1th: Agreed!  I was trying to accommodate future additions, but that is the better option.

nautobot/core/templatetags/helpers.py
tsm1th: @glennmatthews Refactored to use `format_html_join` which looks cleaner now too.

nautobot/core/templatetags/helpers.py
```diff
@@ -43,7 +43,7 @@
 
 @library.filter()
 @register.filter()
-def hyperlinked_object(value, field="display"):
+def hyperlinked_object(value, field="display", target="", rel=""):
```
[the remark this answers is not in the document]
tsm1th: That's a really good point, it could be misleading.  I'll separate it to avoid that confusion.

glennmatthews (approved):
Nice and clean!

Issue #5472, opened 2024-03-22
Static Groups


lampwins:
As a Network Automation Engineer, I want the ability to create and manage static groups that function much like dynamic groups, except their membership is statically managed.

This is to support grouping use cases in which dynamic membership is problematic or otherwise unwanted. These static groups will also be used to underpin a refactor of the way dynamic groups operate, to solve critical performance issues.

AC:
- Groups are defined per content type
- Membership is based upon being explicitly related to the group
- **Not** implemented as a TreeModel to start

HanlinMiao:
StaticGroup model and StaticGroupAssociation model. 

Model Fields:
StaticGroup: name, contenttype, description, tags
StaticGroupAssociation: FK to StaticGroup, and generic FK to members of the Group.

Design Considerations:
1. It will be very similar to ContactAssociation. 
2. Need the ability to `FeatureQuery` for relevant content types. e.g. exclude ObjectChanges
3. Need a standardized pattern for related object tables in the UI
4. StaticGroup should only contain ONE content type
5. The UI has to mirror that of DynamicGroup (get_table_for_model lookup)
6. From any object List view, it would be nice to have a bulk assign operation button to add objects to a StaticGroup (next to bulk edit and bulk delete)
7. Add to a static group and remove from a static group in all bulk edit forms (mixin to add to the base class)
8. ConfigContext filter (Is this a requirement for 2.3?)
9. `get_groups_for_model()` API is something we should add to the BaseModel class.
10. GraphQL support for `object.static_groups`
11. `BaseFilterSet` enhancement: filter based on StaticGroup membership


Questions to explore:
Any mixins or abstract base classes that can be shared between ContactAssociation and StaticGroupAssociation? 

Stretch goal:
Adding a linked count column to the `BaseTable`.
Retrofit some of these design considerations for the Contact/Team feature.

glennmatthews:
After #5631 and #5668, remaining items are:

- ConfigContext filter (Is this a requirement for 2.3?)
- GraphQL support for object.static_groups
- (stretch) Retrofit some of these design considerations for the Contact/Team feature.

PR #5631, opened 2024-04-25
Static groups: data model and REST API

glennmatthews:
Progress towards #5472.

# What's Changed

- Add `StaticGroup` and `StaticGroupAssociation` data models and migrations
   - Add reverse GenericRelation ~~on BaseModel~~ and a `static_groups` helper method via a `StaticGroupMixin` mixin class that's included by default in `OrganizationalModel` and `PrimaryModel` but not `BaseModel`.
   - Similarly add a `ContactMixin` class and apply it to Organizational/PrimaryModel, plus a few specific `BaseModel` subclasses that feel like they should support contacts.
   - Add `StaticGroup.members` property and `.add_members()` and `.remove_members()` helper methods
- Add factories for these models
- ~Add `StaticGroupModelsQuery` class~ Add `static_groups` feature to `models_features` registry and use it to limit choices for StaticGroup content-types based on the presence of an `is_static_group_associable_model` flag
   - I've also added `contacts` feature to the registry as well, though it's not yet used.
- Add filtersets and basic REST API for both models, including a `/static-groups/<UUID>/members/` convenience endpoint
- Add basic documentation
- Add basic API and model tests 
- Add basic filterset tests

# Screenshots

image

image

image

# TODO

- [x] Unit, Integration Tests


nautobot/core/views/__init__.py
```diff
@@ -177,28 +178,31 @@ def get(self, request):
...
+
+                    if table.page:
+                        results.append(
+                            {
+                                "name": queryset.model._meta.verbose_name_plural,
+                                "table": table,
+                                "url": f"{reverse(url)}?q={form.cleaned_data.get('q')}",
+                            }
+                        )
+                except NoReverseMatch:
```
glennmatthews, line 204: As we have `staticgroup` registered as a searchable model but don't yet have a UI view for the model, this exception is thrown. Handling it gracefully here makes the test `nautobot.core.tests.test_views.HomeViewTestCase.test_search` pass instead of failing for now, but we should probably enhance the test so that it will fail if any "Missing URL" messages are shown in the response?

nautobot/extras/models/groups.py
```diff
@@ -17,16 +18,140 @@
...
+        related_name="static_groups",
+        limit_choices_to=StaticGroupModelsQuery(),
+        help_text="The type of object contained in this Static Group.",
+    )
+
+    clone_fields = ["content_type"]
+    is_static_group_associable_model = False
+
+    class Meta:
+        ordering = ["content_type", "name"]
```
glennmatthews, line 55: Wondering if we should just order by name alone - thoughts?

gsnider2195: I think so. Content type isn't a uniqueness constraint, it's not shown in the string representation of the group and it probably won't be the first column in the list view?

nautobot/core/models/__init__.py
```diff
@@ -53,9 +53,17 @@ class BaseModel(models.Model):
...
+    associated_static_groups = GenericRelation(
+        "extras.StaticGroupAssociation",
+        content_type_field="associated_object_type",
+        object_id_field="associated_object_id",
+        related_query_name="associated_static_groups_%(app_label)s_%(class)s",  # 'associated_static_groups_dcim_device'
+    )
 
     objects = BaseManager.from_queryset(RestrictedQuerySet)()
     is_contact_associable_model = True
+    is_static_group_associable_model = True
```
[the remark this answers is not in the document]
gsnider2195: That might be a good idea so we don't have to opt out a bunch of models

nautobot/docs/user-guide/platform-functionality/staticgroup.md
```diff
@@ -0,0 +1,36 @@
...
+
+* `group.members` - returns a `QuerySet` directly representing the objects (for example, a `QuerySet` of `Prefix` records for a Static Group of Prefixes) that are members of this group
+* `group.static_group_associations` - returns a `QuerySet` of `StaticGroupAssociation` records representing the assignment of objects to this group
+* `group.add_members(list_or_queryset)` - associate the given objects to this group as members by creating the appropriate `StaticGroupAssociation` database records
+* `group.remove_members(list_or_queryset)` - disassociate the given objects from this group by deleting any appropriate `StaticGroupAssociation` database records
+
+From a given record of any object type that can be a member of a Static Group, the following Python APIs are available:
+
+* `object.static_groups` - returns a `QuerySet` directly representing the `StaticGroup` records that contain this object as a member
+* `object.associated_static_groups` - returns a `QuerySet` of `StaticGroupAssociation` records representing the assignment of this object to static groups
```
gsnider2195, line 33: `associated_static_groups` implies this is returning static groups. Should this be named `static_group_associations` instead?

nautobot/extras/filters/__init__.py
```diff
@@ -1104,6 +1112,51 @@ class Meta:
...
+            "static_group__description": "icontains",
+            "associated_object_type__app_label": "icontains",
+            "associated_object_type__model": "icontains",
+        }
+    )
+
+    static_group = NaturalKeyOrPKMultipleChoiceFilter(
+        queryset=StaticGroup.objects.all(),
+        to_field_name="name",
+        label="Static group (name or PK)",
```
gsnider2195, line 1149: We might want to stick with ID to be consistent. We have about 200 filters using ID in the label and only 2 using PK.

```suggestion
        label="Static group (name or ID)",
```

nautobot/extras/models/groups.py
```diff
@@ -17,16 +18,140 @@
...
+    def remove_members(self, objects_to_remove):
+        """Remove the given list or QuerySet of objects from this Static Group."""
+        pks_to_remove = set()
+        for obj in objects_to_remove:
+            self._check_valid_member(obj)
+            pks_to_remove.add(obj.pk)
+
+        StaticGroupAssociation.objects.filter(
+            static_group=self, associated_object_type=self.content_type, associated_object_id__in=pks_to_remove
+        ).delete()
```
gsnider2195, line 118: If a queryset is passed in would it be better to use the queryset values instead of building our own list of pks? Django might be able to make a more optimized query?

nautobot/extras/models/groups.py
```diff
@@ -17,16 +18,140 @@
...
+    associated_object_id = models.UUIDField(db_index=True)
+    associated_object = GenericForeignKey(ct_field="associated_object_type", fk_field="associated_object_id")
+
+    is_static_group_associable_model = False
+
+    class Meta:
+        unique_together = ("static_group", "associated_object_type", "associated_object_id")
+
+    def __str__(self):
+        return f"{self.associated_object} ∈ {self.static_group}"
```
gsnider2195, line 146: What's this `∈` symbol?

nautobot/extras/models/groups.py
```diff
@@ -17,16 +18,140 @@
...
+    associated_object_type = models.ForeignKey(
+        to=ContentType,
+        on_delete=models.CASCADE,
+        related_name="static_group_associations",
+        limit_choices_to=StaticGroupModelsQuery(),
+    )
+    associated_object_id = models.UUIDField(db_index=True)
+    associated_object = GenericForeignKey(ct_field="associated_object_type", fk_field="associated_object_id")
+
+    is_static_group_associable_model = False
```
gsnider2195, line 140: Is this model supposed to be contact associable?

nautobot/extras/models/models.py
```diff
@@ -551,6 +551,7 @@ class FileAttachment(BaseModel):
     filename = models.CharField(max_length=CHARFIELD_MAX_LENGTH)
     mimetype = models.CharField(max_length=CHARFIELD_MAX_LENGTH)
 
+    is_static_group_associable_model = False
```
gsnider2195, line 554: Should this and FileProxy be contact associable?

nautobot/extras/utils.py
```diff
@@ -212,6 +212,24 @@ def list_subclasses(self):
...
+    Helper class to get ContentType models that can be included in a StaticGroup.
+    """
+
+    def list_subclasses(self):
+        return [
+            _class
+            for _class in apps.get_models()
+            if (
+                hasattr(_class, "is_static_group_associable_model")
+                and _class.is_static_group_associable_model
```
gsnider2195, line 227:
```suggestion
                getattr(_class, "is_static_group_associable_model", False)
```

nautobot/docs/user-guide/platform-functionality/staticgroup.md
glennmatthews: I was following the same pattern as `object.associated_contacts`, but I agree it's not super clear.

The problem with using `static_group_associations` is that on `StaticGroup` itself that lookup name becomes ambiguous which side of the M2M it represents.

nautobot/extras/models/groups.py
glennmatthews: Set theory "is a member of" or "belongs to". As I mentioned in dojo yesterday I was expecting this might be too academic/nerdy in general. Do you think `f"{self.associated_object} as a member of {self.static_group}"` would be better?

nautobot/extras/models/groups.py
glennmatthews: Probably shouldn't be.

nautobot/extras/models/models.py
glennmatthews: Probably not.

nautobot/extras/models/groups.py
gsnider2195: Oh, right! Might be better to avoid Unicode in general and yeah it wasn't clear to me, a humble network engineer

nautobot/extras/models/groups.py
```diff
@@ -17,16 +18,152 @@
...
+        limit_choices_to=FeatureQuery("static_groups"),
+    )
+    associated_object_id = models.UUIDField(db_index=True)
+    associated_object = GenericForeignKey(ct_field="associated_object_type", fk_field="associated_object_id")
+
+    is_contact_associable_model = False
+    is_static_group_associable_model = False
+
+    class Meta:
+        unique_together = ("static_group", "associated_object_type", "associated_object_id")
```
glennmatthews, line 155: Should probably at least add `ordering = ["static_group"]` even if we can't define a more specific ordering?

nautobot/core/models/name_color_content_types.py
```diff
@@ -39,12 +39,14 @@ def get_for_models(self, models_):
...
     ChangeLoggedModel,
+    ContactMixin,
     CustomFieldModel,
-    RelationshipModel,
-    NotesMixin,
     DynamicGroupMixin,
+    NotesMixin,
+    RelationshipModel,
+    StaticGroupMixin,
+    BaseModel,
```
HanlinMiao, line 49: Would it be possible to move BaseModel on top so that the list is alpha sorted? or Does it cause a MRO issue?

nautobot/core/testing/views.py
```diff
@@ -265,7 +265,7 @@ def test_get_object_changelog(self):
             response = self.client.get(url)
             self.assertHttpStatus(response, 200)
             response_data = response.content.decode(response.charset)
-            if type(obj) not in [extras_models.Contact, extras_models.Team]:
+            if getattr(obj, "is_contact_associable_model", False):
```
HanlinMiao, line 268: 🥹

nautobot/docs/development/apps/api/models/index.md
```diff
@@ -13,11 +13,22 @@ For more advanced usage, you may want to instead inherit from one of Nautobot's
...
 | [Relationships](../../../../user-guide/platform-functionality/relationship.md) | ❌ | ❌ | ✅ | ✅ |
-| [Note](../../../../user-guide/platform-functionality/note.md) | ❌ | ❌ | ✅ | ✅ |
+| [Static Groups](../../../../user-guide/platform-functionality/staticgroup.md) | ❌ | ❌ | ✅ | ✅ |
 | [Tags](../../../../user-guide/platform-functionality/tag.md) | ❌ | ❌ | ❌ | ✅ |
 
++++ 2.2.0
+    Support for Contact and Team assignment to all model types was added.
+
++/- 2.3.0
+    Support for Contact and Team assignment was removed from `django.db.models.Model` and `BaseModel`. The mixin class `ContactMixin` was added to be usable by `BaseModel` subclasses that want to be assignable to Contacts and Teams.
```
HanlinMiao, line 27: Should we also mention that `ContactMixin` is added to `OrganizationModel` and `PrimaryModel` as well

nautobot/docs/release-notes/version-2.3.md
```diff
@@ -12,6 +12,10 @@ This document describes all new features and changes in Nautobot 2.3.
 
 Added an optional `role` field to Interface and VMInterface models to track common interface configurations. Now the users can create [Role](../user-guide/platform-functionality/role.md) instances that can be assigned to [interfaces](../user-guide/core-data-model/dcim/interface.md) and [vminterfaces](../user-guide/core-data-model/virtualization/vminterface.md).
 
+#### Static Groups
+
+Added the new models `StaticGroup` and `StaticGroupAssociation`, which can be used to define arbitrary groups of objects of any one data type. These serve as a simplified alternative to the existing Dynamic Groups feature.
```
HanlinMiao, line 17: Maybe?
```suggestion
Added the new models `StaticGroup` and `StaticGroupAssociation`, which can be used to define static and arbitrary groups of objects of any one data type. These serve as a simplified alternative to the existing Dynamic Groups feature.
```

nautobot/extras/factory.py
```diff
@@ -298,6 +306,50 @@ def content_types(self, create, extracted, **kwargs):
...
+
+    static_group = random_instance(StaticGroup, allow_null=False)
+    associated_object_type = factory.LazyAttribute(lambda o: o.static_group.content_type)
+
+    @factory.lazy_attribute
+    def associated_object_id(self):
+        queryset = self.associated_object_type.model_class().objects.all()
+        if queryset.exists():
+            return factory.random.randgen.choice(queryset).pk
+        return faker.Faker().uuid4()
```
HanlinMiao, line 350: Hmmm would this be invalid data?

nautobot/core/models/name_color_content_types.py
glennmatthews: Caused an MRO issue. The mixins need to come before the base class.

nautobot/docs/release-notes/version-2.3.md
glennmatthews: I'm on the fence on this one.

nautobot/extras/models/groups.py
```diff
@@ -17,16 +18,153 @@
...
+    content_type = models.ForeignKey(
+        to=ContentType,
+        on_delete=models.CASCADE,
+        related_name="static_groups",
+        limit_choices_to=FeatureQuery("static_groups"),
+        help_text="The type of object contained in this Static Group.",
+    )
+
+    clone_fields = ["content_type"]
+    is_static_group_associable_model = False
```
gsnider2195, line 52: Would it be trivial to make static groups a static group associable model? I could see it being useful for something like this:

- `StaticGroup(content_type = StaticGroup, name="Change Window 1234")`
  - `StaticGroup(content_type = Device, name="Change Window 1234 Devices")`
  - `StaticGroup(content_type = Controller, name="Change Window 1234 Controllers")`

Then an app could be configured to only make changes to the Devices and Controllers in the child groups during the change window.

nautobot/extras/models/groups.py
```diff
@@ -17,16 +18,153 @@
...
+
+        if self.present_in_database:
+            # Check immutable fields
+            database_object = self.__class__.objects.get(pk=self.pk)
+
+            if self.content_type != database_object.content_type:
+                raise ValidationError({"content_type": "ContentType cannot be changed once created"})
+
+    def _check_valid_member(self, member):
+        if not (isinstance(member, self.model) or issubclass(member, self.model)):
```
gsnider2195, line 94: Is `issubclass` valid? Wouldn't that allow more than one ContentType per static group?

nautobot/extras/models/groups.py
```diff
@@ -17,16 +18,153 @@
...
+            raise TypeError(f"{member} is not a {self.model._meta.label_lower}")
+
+    def add_members(self, objects_to_add):
+        """Add the given list or QuerySet of objects to this Static Group."""
+        if isinstance(objects_to_add, models.QuerySet):
+            self._check_valid_member(objects_to_add.model)
```
gsnider2195, line 100: Maybe just check that `objects_to_add.model == self.model` here?

nautobot/extras/models/groups.py
glennmatthews: Interesting idea! I was going by the fact that StaticGroups are not (per #5472) intended to be self-nestable at this time (Group A of Devices containing Group B also of Devices and therefore implicitly including all of Group B's members), but it's an interesting idea to use a static group to relate groups of otherwise unrelated objects. I'm not sure I'm convinced that this is needed at this time though, couldn't the app just point to the "child" groups individually?

nautobot/extras/models/groups.py
glennmatthews: The intent here was to allow either model instances (list of objects case) or a specific model class (queryset case) to be validated. I hadn't thought about the concrete subclass case though, so good catch. I think you're right in your next comment that just checking equality should be sufficient for that case. Will update.

nautobot/extras/models/groups.py
gsnider2195: > couldn't the app just point to the "child" groups individually?

Definitely, I'm just trying to think of why we wouldn't want to support this use case. Are we thinking of making StaticGroups some kind of TreeModel or supporting multiple content types in the future?

nautobot/extras/tests/test_api.py
```diff
@@ -3475,6 +3478,149 @@ def setUpTestData(cls):
...
+    def test_get_members(self):
+        """Test that the `/members/` API endpoint returns what is expected."""
+        self.add_permissions("extras.view_staticgroup")
+        instance = StaticGroup.objects.filter(static_group_associations__isnull=False).distinct().first()
+        self.assertIsNotNone(instance)
+        member_count = instance.members.count()
+        url = reverse("extras-api:staticgroup-members", kwargs={"pk": instance.pk})
+        response = self.client.get(url, **self.header)
+        self.assertHttpStatus(response, status.HTTP_200_OK)
+        self.assertEqual(member_count, len(response.json()["results"]))
```
gsnider2195, line 3544: Isn't there a `count` in the response we could also check?

nautobot/extras/tests/test_api.py
```diff
@@ -3475,6 +3478,149 @@ def setUpTestData(cls):
...
+                "associated_object_type": "dcim.device",
+                "associated_object_id": device_pks[2],
+            },
+            {
+                "static_group": cls.sg2.pk,
+                "associated_object_type": "dcim.device",
+                "associated_object_id": device_pks[3],
+            },
+        ]
+        # TODO: this isn't really valid since we're changing the associated_object_type but not the associated_object_id
```
gsnider2195, line 3606: How do we deal with invalid data passed to bulk update on other classes?

nautobot/extras/models/groups.py
glennmatthews: I think we're at least considering the TreeModel approach at some point yeah. I think I'd be concerned about user expectations of what a static-group-of-static-groups "means" as far as inheritance goes and such, especially given the example we've already got with dynamic groups containing other dynamic groups. I'd prefer to avoid that for now; as you pointed out it *should* be a simple change if we decide later that it's something we need.

nautobot/extras/tests/test_api.py
glennmatthews: If it's truly invalid (would get rejected by bulk-update API) that's one thing - in this case though we're just making the GFK point to a nonexistent object, which is possible in the DB unless we'd have to have a custom clean() or validate() method to catch it. Might be worth adding?

nautobot/extras/tests/test_filters.py
```diff
@@ -1803,6 +1807,44 @@ def test_secret_type(self):
         self.assertEqual(self.filterset(params, self.queryset).qs.count(), 2)
 
 
+class StaticGroupTestCase(FilterTestCases.NameOnlyFilterTestCase):
+    queryset = StaticGroup.objects.all()
+    filterset = StaticGroupFilterSet
+    maxDiff = None
```
gsnider2195, line 1813: Should this be on the base class?

nautobot/extras/tests/test_api.py
gsnider2195: Probably not worth the effort for this issue imo

gsnider2195 (approved):

gsnider2195 (approved):

PR #5639, opened 2024-04-26
[Feature Branch] Implemented SavedView model

### LINKED ISSUE #1758: Saved Views
### As ...

Nelly - Network Engineer

### I want ...

to be able to save multiple list view configurations

### So that ...

I can quickly switch between multiple workflows, where each workflow requires different columns in list view.

- I can save several configuration of columns in list view
- I can name them for easier orientation
- I can quickly load any saved configuration from list view

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation

### Database Changes

None, it could be done with current implementation of user config
But some if cleaner solution would be implemented

### External Dependencies

_No response_

HanlinMiao:
# Closes #1758 
# What's Changed

# Screenshots

Implemented Saved View Model
SavedView List View
Screenshot 2024-05-01 at 5 07 28 PM
SavedView Dropdown List
Screenshot 2024-05-01 at 5 09 28 PM
SavedView modal form
Screenshot 2024-05-01 at 5 09 40 PM

# TODO

- [ ] Unit, Integration Tests


changes/1758.dependencies
```diff
@@ -0,0 +1 @@
+Bumped materialdesignicons from 6.5.95 to 7.4.47.
```
glennmatthews, line 1:
```suggestion
Updated `materialdesignicons` to version 7.4.47.
```

nautobot/core/models/__init__.py
```diff
@@ -56,6 +56,7 @@ class BaseModel(models.Model):
 
     objects = BaseManager.from_queryset(RestrictedQuerySet)()
     is_contact_associable_model = True
+    is_saved_view_model = True
```
glennmatthews, line 59: I'm thinking that (much like Contacts and Static Groups) it might make more sense for `OrganizationalModel` and `PrimaryModel` to default to supporting saved views, and `BaseModel` to default to not supporting them.

Take a look at my latest diffs in #5631 to see how I'm using some mixins to implement that for those other two cases.

nautobot/core/tables.py
```diff
@@ -34,7 +35,9 @@ class Meta:
             "class": "table table-hover table-headings",
         }
 
-    def __init__(self, *args, user=None, hide_hierarchy_ui=False, **kwargs):
+    def __init__(
+        self, *args, table_changes_pending=False, saved_view_pk=None, user=None, hide_hierarchy_ui=False, **kwargs
```
glennmatthews, line 39: Would it be cleaner to pass around the `SavedView` instance instead of just its PK?

nautobot/core/tables.py
```diff
@@ -79,27 +82,53 @@ def __init__(self, *args, user=None, hide_hierarchy_ui=False, **kwargs):
...
+                            self.columns.hide(name)
+                    self.sequence = [c for c in columns if c in self.base_columns]
+
+                    # Always include PK and actions column, if defined on the table
+                    if pk:
+                        self.base_columns["pk"] = pk
+                        self.sequence.insert(0, "pk")
+                    if actions:
+                        self.base_columns["actions"] = actions
+                        self.sequence.append("actions")
```
glennmatthews, line 131: Definitely should be possible to deduplicate some of this logic :-)

nautobot/core/templates/generic/object_list.html
```diff
@@ -46,6 +46,58 @@
...
+        <div class="btn-group">
+            <button type="button" class="btn btn-primary dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                <span class="mdi mdi-view-compact" aria-hidden="true"></span> Saved Views <span class="caret"></span>
+            </button>
+            <ul class="dropdown-menu">
+                {% if perms.users.view_savedview %}
+                    {% for saved_view in saved_views %}
+                        <li>
+                            <a href="{{ saved_view.get_absolute_url }}">
+                                {% if saved_view.pk == current_saved_view.pk %}
```
glennmatthews, line 59:
```suggestion
                                {% if saved_view == current_saved_view %}
```

nautobot/core/templates/generic/object_list.html
```diff
@@ -46,6 +46,58 @@
...
+                                    <span class="mdi mdi-view-compact-outline" aria-hidden="true"></span> 
+                                {% endif %}
+                                {{saved_view.name}}
+                            </a>
+                        </li>
+                    {% endfor %}
+                {% endif %}
+                {% if saved_views %}
+                    <li role="separator" class="divider"></li>
+                {% endif %}
```
glennmatthews, line 71: Maybe?

```suggestion
                    {% if saved_views %}
                        <li role="separator" class="divider"></li>
                    {% endif %}
                {% endif %}
```

nautobot/core/templates/generic/object_list.html
```diff
@@ -60,7 +112,16 @@
         {% endblock export_button %}
     {% endif %}
 </div>
-<h1>{% block title %}{{ title }}{% endblock %}</h1>
+<h1>{% block title %}
+    {{ title }}
+        {% if current_saved_view %}
+            - {{ current_saved_view.name }}
+            {% if new_changes_not_applied %}
+                *
```
glennmatthews, line 120: Possibly?

```suggestion
                <sup>*</sup>
```

nautobot/core/templates/generic/object_list.html
```diff
@@ -60,7 +112,16 @@
         {% endblock export_button %}
     {% endif %}
 </div>
-<h1>{% block title %}{{ title }}{% endblock %}</h1>
+<h1>{% block title %}
+    {{ title }}
+        {% if current_saved_view %}
+            - {{ current_saved_view.name }}
```
glennmatthews, line 118: An em dash might look a bit nicer:

```suggestion
            — {{ current_saved_view.name }}
```

nautobot/users/tables.py
```diff
@@ -0,0 +1,33 @@
...
+        model = SavedView
+        fields = (
+            "pk",
+            "name",
+            "owner",
+            "list_view_name",
+            "table_config",
+            "pagination_count",
+            "filter_params",
+            "sort_order",
```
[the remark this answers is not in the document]
HanlinMiao: Screenshot 2024-04-29 at 12 38 26 PM
It looks okay IMO

nautobot/users/navigation.py
```diff
@@ -0,0 +1,49 @@
...
+                        link="users:savedview_list",
+                        name="Saved Views",
+                        weight=100,
+                        permissions=["users.view_savedviews"],
+                    ),
+                ),
+            ),
+        ),
+    ),
+)
```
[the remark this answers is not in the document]
HanlinMiao: Moved under Extensibility NavMenuTab per standup

nautobot/users/models.py
```diff
@@ -282,3 +286,67 @@ def list_constraints(self):
...
+        # TODO table_config
+        for key, value in self.filter_params.items():
+            for item in value:
+                query_list.append(f"{key}={urllib.parse.quote_plus(item)}")
+
+        query_list.append(f"per_page={self.pagination_count}")
+
+        for key, value in self.sort_order.items():
+            for item in value:
+                query_list.append(f"{key}={item}")
```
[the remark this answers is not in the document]
HanlinMiao: Refactored it to a list

nautobot/users/models.py
```diff
@@ -282,3 +286,67 @@ def list_constraints(self):
...
+
+
+#
+# SavedView
+#
+
+
+@extras_features(
+    "custom_validators",
+    "graphql",
```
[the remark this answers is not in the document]
HanlinMiao: Ahhh removed

nautobot/core/templates/generic/object_list.html
```diff
@@ -60,7 +112,16 @@
         {% endblock export_button %}
     {% endif %}
 </div>
-<h1>{% block title %}{{ title }}{% endblock %}</h1>
+<h1>{% block title %}
```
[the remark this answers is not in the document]
HanlinMiao: Does not seem like we have any in core
```
grep 'block title' nautobot/*/templates/*.html     
nautobot/core/templates/403.html:{% block title %}Access Denied{% endblock %}
nautobot/core/templates/403_csrf_failure.html:{% block title %}Invalid Form Submission{% endblock %}
nautobot/core/templates/404.html:{% block title %}Page Not Found{% endblock %}
nautobot/core/templates/40x.html:                <strong>{% block icon %}{% endblock %} {% block title %}{% endblock %}</strong>
nautobot/core/templates/500.html:{% block title %}Server Error{% endblock %}
nautobot/core/templates/base_django.html:    <title>{% block title %}Home{% endblock %} - {{ settings.BRANDING_TITLE }}</title>
nautobot/core/templates/import_success.html:    <h1>{% block title %}Import Completed{% endblock %}</h1>
nautobot/core/templates/search.html:{% block title %}Search{% endblock %}
nautobot/core/templates/swagger_ui.html:{% block title %}{{ title | default:"Swagger" }}{% endblock title %}
```

HanlinMiao:
Will have tests and documentation included in another PR

nautobot/core/views/utils.py
```diff
@@ -273,3 +273,33 @@ def prepare_cloned_fields(instance):
...
+    for param in view.non_filter_params:
+        if param in query_dict_keys:
+            query_dict_keys.remove(param)
+    filter_param_keys = sorted(list(current_saved_view.filter_params.keys()))
+
+    if query_dict_keys != filter_param_keys:
+        return True
+    for key in filter_param_keys:
+        if sorted(current_saved_view.filter_params.get(key)) != sorted(request.GET.getlist(key)):
+            return True
```
glennmatthews, line 304:
```suggestion
    filter_params = current_saved_view.filter_params
    
    if set(query_dict_keys) != set(filter_params.keys():
        return True
    for key, value in filter_params.items():
        if set(value) != set(request.GET.getlist(key)):
            return True
```

nautobot/project-static/js/forms.js
```diff
@@ -638,7 +638,7 @@ function initializeDynamicFilterForm(context){
         let query_string = location.search.substr(1).split("&");
 
         if (type === "parent") {
-            query_string = query_string.filter(item => item.search(field_value) < 0);
+            query_string = query_string.filter(item => item.split("=")[0].search(field_value) < 0);
```
glennmatthews, line 641: This will need to be reworked after #5646, fyi.

nautobot/project-static/js/forms.js
```diff
@@ -672,6 +696,39 @@ function initializeDynamicFilterForm(context){
...
+        if (per_page_query_string != null) {
+            per_page_query_string = "per_page=" + per_page
+        }
+        const table_changes = urlParams.get('table_changes_pending');
+        var table_changes_query_string = ""
+        if (table_changes_query_string != null) {
+            table_changes_query_string = "&table_changes_pending=" + table_changes
+        }
+        // clear filter params info and retain other information
+        location.replace("?" + per_page_query_string + sort_order_query_string + saved_view_pk_query_string + table_changes_query_string)
```
glennmatthews, line 728: Let's use `location.assign()` instead of `location.replace()` so that back button works as expected.

nautobot/project-static/js/forms.js
```diff
@@ -672,6 +696,39 @@ function initializeDynamicFilterForm(context){
...
+        const per_page = urlParams.get('per_page');
+        var per_page_query_string = ""
+        if (per_page_query_string != null) {
+            per_page_query_string = "per_page=" + per_page
+        }
+        const table_changes = urlParams.get('table_changes_pending');
+        var table_changes_query_string = ""
+        if (table_changes_query_string != null) {
+            table_changes_query_string = "&table_changes_pending=" + table_changes
+        }
```
glennmatthews, line 726: Might be cleaner/simpler to iterate over everything in `urlParams` and just delete from it anything that *isn't* saved_view/sort/per_page/table_changes_pending?

nautobot/project-static/js/tableconfig.js
```diff
@@ -39,7 +39,12 @@ $(document).ready(function() {
             data: JSON.stringify(data),
         }).done(function () {
             // Reload the page
-            window.location.reload(true);
+            if (window.location.href.includes("saved_view=")){
+                var new_url = window.location.href+"&table_changes_pending=true";
+                window.location.replace(new_url);
```
glennmatthews, line 44: Maybe .assign here as well? Not sure.

nautobot/users/models.py
```diff
@@ -282,3 +286,73 @@ def list_constraints(self):
...
+    pagination_count = models.IntegerField(default=50, blank=True, help_text="Pagination Count of this view")
+    filter_params = models.JSONField(
+        encoder=DjangoJSONEncoder, blank=True, default=dict, help_text="filter parameters that are applied to this view"
+    )
+    sort_order = models.JSONField(
+        encoder=DjangoJSONEncoder,
+        blank=True,
+        default=dict,
+        help_text="table column sort orders that are applied to this view",
+    )
```
glennmatthews, line 331: As I mentioned in standup the other day, I wonder if we're making this model too list-view specific - maybe we should combine all four of these fields into a single JSON field that could store a different dict entirely in the future for customizing say a detail view instead of a list view? Not sure - would that change make the code too much more complex at this time?

nautobot/users/views.py
```diff
@@ -213,6 +226,130 @@ def post(self, request):
...
+    queryset = SavedView.objects.all()
+    filterset_class = SavedViewFilterSet
+    table_class = SavedViewTable
+    action_buttons = ("export",)
+
+    def get_table_class_string_from_view_name(self, view_name):
+        app_label, model_name = view_name.split(":")
+        model_name = model_name.replace("_list", "")
+        model = apps.get_model(app_label=app_label, model_name=model_name)
+        return f"{model.__name__}Table"
```
glennmatthews, line 251: can we use `get_table_for_model` here? Maybe we should add a reverse lookup `get_model_for_view_name` in `nautobot.core.utils.lookup` if we don't already have something like that?

nautobot/users/views.py
```diff
@@ -213,6 +226,130 @@ def post(self, request):
...
+        list_view_url = reverse(instance.view) + query_string + f"&saved_view={instance.pk}"
+        return redirect(list_view_url)
+
+    def update(self, request, *args, **kwargs):
+        """
+        request.GET: render the ObjectEditForm which is passed to NautobotHTMLRenderer as Response.
+        request.POST: call perform_update() which validates the form and perform the action of update/partial_update of an existing object.
+        Override to add more variables to Response.
+        """
+        sv = self.get_queryset().get(pk=request.GET.get("saved_view", None))
```
glennmatthews, line 269: do we need to use a `get_object_or_404` here or similar?

nautobot/users/views.py
```diff
@@ -213,6 +226,130 @@ def post(self, request):
...
+        for key, _ in request.GET.items():
+            if key in self.non_filter_params:
+                continue
+            sv.filter_params[key] = request.GET.getlist(key)
+            new_filter_params.append(key)
+
+        # delete filter params that are no longer in use
+        for param in list(sv.filter_params.keys()):
+            if param not in new_filter_params:
+                sv.filter_params.pop(param)
```
glennmatthews, line 287: Maybe?

```suggestion
        sv.filter_params = {}
        for key in request.GET:
            if key in self.non_filter_params:
                continue
            sv.filter_params[key] = request.GET.getlist(key)
```

nautobot/users/views.py
```diff
@@ -213,6 +226,130 @@ def post(self, request):
...
+        if table_changes_pending:
+            table_class = self.get_table_class_string_from_view_name(sv.view)
+            sv.table_config[f"{table_class}"] = request.user.get_config(f"tables.{table_class}")
+        sv.validated_save()
+        query_string = sv.view_config
+        list_view_url = reverse(sv.view) + query_string + f"&saved_view={sv.pk}"
+        messages.success(request, f"Successfully updated current view {sv.name}")
+        return redirect(list_view_url)
```
glennmatthews, line 295: Why do we need to include `query_string` in the redirect? Shouldn't just `?saved_view={sv.pk}` be sufficient to render the updated view?

nautobot/users/views.py
```diff
@@ -213,6 +226,130 @@ def post(self, request):
...
+
+    def create(self, request, *args, **kwargs):
+        """
+        request.GET: render the ObjectForm which is passed to NautobotHTMLRenderer as Response.
+        request.POST: call perform_create() which validates the form and perform the action of create.
+        Override to add more variables to Response.
+        """
+        name = request.POST.get("name")
+        view_name = request.GET.get("view")
```
glennmatthews, line 304: Mixture of `request.POST` and `request.GET` here seems odd... is it intentional/correct?

nautobot/users/views.py
```diff
@@ -213,6 +226,130 @@ def post(self, request):
...
+        derived_view_pk = request.GET.get("saved_view", None)
+        derived_instance = None
+        if derived_view_pk:
+            derived_instance = self.get_queryset().get(pk=derived_view_pk)
+        table_changes_pending = request.GET.get("table_changes_pending", False)
+        try:
+            sv = SavedView.objects.create(name=name, owner=request.user, view=view_name)
+        except IntegrityError:
+            messages.error(
+                request, f"Integrity Error: User {request.user.username} already has a Saved View named {name}"
```
glennmatthews, line 314:
```suggestion
                request, f"You already have a Saved View named {name}"
```

nautobot/users/views.py
```diff
@@ -213,6 +226,130 @@ def post(self, request):
...
+        for key, _ in request.GET.items():
+            if key in [*self.non_filter_params, "view"]:
+                continue
+            sv.filter_params[key] = request.GET.getlist(key)
+            new_filter_params.append(key)
+
+        # delete filter params that are no longer in use
+        for param in list(sv.filter_params.keys()):
+            if param not in new_filter_params:
+                sv.filter_params.pop(param)
```
glennmatthews, line 336: Same here:

```suggestion
        for key in request.GET:
            if key in [*self.non_filter_params, "view"]:
                continue
            sv.filter_params[key] = request.GET.getlist(key)
```

nautobot/users/views.py
```diff
@@ -213,6 +226,130 @@ def post(self, request):
...
+        if table_changes_pending or derived_instance is None:
+            sv.table_config[f"{table_class}"] = request.user.get_config(f"tables.{table_class}")
+        else:
+            sv.table_config[f"{table_class}"] = derived_instance.table_config[f"{table_class}"]
+        try:
+            sv.validated_save()
+            query_string = sv.view_config
+            list_view_url = reverse(sv.view) + query_string + f"&saved_view={sv.pk}"
+            messages.success(request, f"Successfully created new Saved View {sv.name}")
+            return redirect(list_view_url)
```
glennmatthews, line 347: Same question here about necessity of `query_string`

nautobot/core/templates/inc/search_panel.html
```diff
@@ -20,8 +20,8 @@
         <button type="submit" class="btn btn-primary">
             <span class="mdi mdi-magnify" aria-hidden="true"></span> Apply
         </button>
-        <a href="." class="btn btn-default">
+        <button type="reset" class="btn btn-default">
```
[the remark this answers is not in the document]
HanlinMiao: I made changes in `forms.js` to reflect this change as well.

nautobot/core/templates/utilities/templatetags/saved_view_modal.html
```diff
@@ -0,0 +1,38 @@
...
+            </div>
+            <div class="modal-body">
+                <form action="{% url 'users:savedview_add' %}?view={{view}}&{{params}}" method="post">
+                    {% csrf_token %}
+                    <div class="form-group">
+                        {% render_field form.name %}
+                    </div>
+                    <div class="text-right noprint">
+                        <button type="submit" class="btn btn-primary">
+                            <span class="mdi mdi-magnify" aria-hidden="true"></span> Save
```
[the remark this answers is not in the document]
HanlinMiao: Changed it to `mdi-content-save-plus-outline`

nautobot/core/views/generic.py
```diff
@@ -296,11 +301,22 @@ def get(self, request):
...
+        current_saved_view = None
         if self.table:
             # Construct the objects table
             if self.request.GET.getlist("sort"):
                 hide_hierarchy_ui = True  # hide tree hierarchy if custom sort is used
-            table = self.table(self.queryset, user=request.user, hide_hierarchy_ui=hide_hierarchy_ui)
+            current_saved_view_pk = self.request.GET.get("saved_view", None)
+            table_changes_pending = self.request.GET.get("table_changes_pending", False)
+            if current_saved_view_pk:
+                current_saved_view = SavedView.objects.get(pk=current_saved_view_pk)
```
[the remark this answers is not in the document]
HanlinMiao: Quick question here, would `pk=current_saved_view_pk` be sufficient to warrant uniqueness here? Why do we need `view=list_url?`

nautobot/users/views.py
```diff
@@ -213,6 +226,118 @@ def post(self, request):
...
+):
+    queryset = SavedView.objects.all()
+    filterset_class = SavedViewFilterSet
+    table_class = SavedViewTable
+    action_buttons = ("export",)
+
+    def get_table_class_string_from_view_name(self, view_name):
+        model = get_model_for_view_name(view_name)
+        if model:
+            return get_table_for_model(model)
```
glennmatthews, line 250: Not sure if this was an intentional change in the latest commits, but this function is now returning a table class object instead of a string, which at least doesn't match the method name.

HanlinMiao: Oops thank you for catching that!

nautobot/users/views.py
glennmatthews: Ack. Might be good for the docstring to spell out what information it's pulling from where?

nautobot/core/templates/utilities/templatetags/saved_view_modal.html
```diff
@@ -0,0 +1,38 @@
+{% load form_helpers %}
+{% load static %}
```
glennmatthews, line 2: Maybe unnecessary?

nautobot/project-static/js/forms.js
HanlinMiao: Refactored to delete all except non_filter_params.
```
this_context.find("#dynamic-filter-form, #default-filter form").on("reset", function(e){
        e.preventDefault()
        // make two copies of url params
        const urlParams = new URLSearchParams(window.location.search);
        const newUrlParams = new URLSearchParams(window.location.search);
        // every query string that is non-filter-related
        const non_filter_params = ["saved_view", "sort", "per_page", "table_changes_pending"]
        for (const [key, value] of urlParams.entries()) {
            // remove filter params
            if (non_filter_params.includes(key) === false) {
                newUrlParams.delete(key, value)
            }
        }
        location.assign("?" + urlParams.toString())
    })
```

nautobot/users/models.py
HanlinMiao: Collapsed these attributes to one `config` JSONField.
Screenshot 2024-05-01 at 1 52 36 PM

nautobot/core/templates/utilities/templatetags/saved_view_modal.html
HanlinMiao: Deletus

nautobot/core/templates/utilities/templatetags/saved_view_modal.html
```diff
@@ -0,0 +1,37 @@
+{% load form_helpers %}
+
+{% block extra_styles %}
+<style>
+    #saved_view_modal {
+        position: absolute;
+        top: 30%;
+    }
```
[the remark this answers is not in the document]
HanlinMiao: Aesthetics for sure, it will be on the top of the page instead of the middle. It won't look too bad so I will remove it.

nautobot/core/templates/utilities/templatetags/saved_view_modal.html
```diff
@@ -0,0 +1,37 @@
...
+</style>
+{% endblock %}
+
+<div class="modal fade" tabindex="-1" id="saved_view_modal">
+    <div class="modal-dialog modal-lg" style="width: 50% !important;">
```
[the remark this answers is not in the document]
HanlinMiao: Won't look too bad without it so removed.

nautobot/users/templates/users/advanced_settings_edit.html
```diff
@@ -27,7 +27,7 @@
         </div>
         <div class="text-right">
             <button type="submit" name="_update" class="btn btn-primary">Update</button>
-            <a href="{% url 'user:profile' %}" class="btn btn-default">Cancel</a>
+            <a href="{% url 'users:profile' %}" class="btn btn-default">Cancel</a>
```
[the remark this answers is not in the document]
HanlinMiao: [This](https://networktocode-llc.slack.com/archives/C01RMAGBD29/p1713894781593919) is when we discussed this change. Would you recommend leave it as is e.g. `user`

nautobot/core/templates/generic/object_list.html
```diff
@@ -46,6 +46,61 @@
...
+                {% endif %}
+                {% if perms.users.add_savedview %}
+                    <li>
+                        <a href="" data-toggle="modal" data-target="#saved_view_modal" title="Save Current View As">
+                            <i class="mdi mdi-content-save-plus-outline text-muted" aria-hidden="true"></i>
+                            <span>Save As New View</span>
+                        </a>
+                    </li>
+                {% endif %}
+                {% if request.GET.saved_view %}
```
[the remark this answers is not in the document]
HanlinMiao: I would rather pass `current_saved_view` around because we are accessing its `name` attribute in {% block title %} on line 121

nautobot/core/views/generic.py
```diff
@@ -296,11 +301,28 @@ def get(self, request):
...
         if self.table:
             # Construct the objects table
             if self.request.GET.getlist("sort"):
                 hide_hierarchy_ui = True  # hide tree hierarchy if custom sort is used
-            table = self.table(self.queryset, user=request.user, hide_hierarchy_ui=hide_hierarchy_ui)
+            table_changes_pending = self.request.GET.get("table_changes_pending", False)
+            if current_saved_view_pk:
+                current_saved_view = SavedView.objects.restrict(request.user, "view").get(
+                    view=list_url, pk=current_saved_view_pk
+                )
```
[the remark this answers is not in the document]
HanlinMiao: Would this be a reasonable extra layer of protection? Worst case scenario current_saved_view would be `None`
```
try:
     current_saved_view = SavedView.objects.restrict(request.user, "view").get(view=list_url, pk=current_saved_view_pk)
except ObjectDoesNotExist:
      pass
```

nautobot/users/templates/users/advanced_settings_edit.html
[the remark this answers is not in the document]
glennmatthews: yeah, I don't think I understood the implication when we discussed on Slack. If it's possible to revert this I think it'd be safer. If that turns out to be too difficult to do for whatever reason then we can revisit.

nautobot/core/views/generic.py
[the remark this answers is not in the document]
glennmatthews: Worth a try - hard to say for sure without trying it myself. At the very least might be good to do a `messages.error(request, "Saved view <pk> not found")` or similar?

nautobot/core/utils/lookup.py
```diff
@@ -208,6 +213,20 @@ def get_view_for_model(model, view_type=""):
     return result
 
 
+def get_model_for_view_name(view_name):
+    """
+    Return the model class associated with the given view_name e.g. "circuits:circuit_detail", "dcim:device_list" and etc.
+    If the app_label or model_name contained by the given view_name is invalid, this will return `None`.
+    """
+    app_label, model_name = view_name.split(":")  # dcim, device_list
+    model_name = model_name.split("_")[0]  # device
```
glennmatthews, line 222: Does this need the reverse equivalent of the logic in `get_view_for_model`, e.g. `user -> users`, etc.?

nautobot/core/views/renderers.py
```diff
@@ -64,14 +71,28 @@ def construct_table(self, view, **kwargs):
...
+                saved_view = None
+                if saved_view_pk is not None:
+                    try:
+                        self.saved_view = SavedView.objects.restrict(request.user, "view").get(pk=saved_view_pk)
+                    except ObjectDoesNotExist:
+                        pass
+                table = table_class(
+                    queryset,
+                    table_changes_pending=table_changes_pending,
+                    saved_view=saved_view,
```
glennmatthews, line 92:
```suggestion
                self.saved_view = None
                if saved_view_pk is not None:
                    try:
                        self.saved_view = SavedView.objects.restrict(request.user, "view").get(pk=saved_view_pk)
                    except ObjectDoesNotExist:
                        pass
                table = table_class(
                    queryset,
                    table_changes_pending=table_changes_pending,
                    saved_view=self.saved_view,
```

nautobot/core/views/utils.py
```diff
@@ -274,3 +274,33 @@ def prepare_cloned_fields(instance):
...
+    if current_saved_view is None:
+        return False
+    query_dict = request.GET.dict()
+
+    if query_dict.get("table_changes_pending", None):
+        return True
+    if int(query_dict.get("per_page")) != current_saved_view.config.get("pagination_count", 50):
+        return True
+    if (request.GET.getlist("sort", [])) != current_saved_view.config.get("sort_order", []):
+        return True
```
glennmatthews, line 294: Don't we need to handle the case where these params are absent from `request.GET`? In other words, if request.GET has no query params, but the saved view has a `pagination_count` or a `sort_order`, this function should return False, but I think as written here these will return True?

nautobot/core/views/utils.py
```diff
@@ -274,3 +274,33 @@ def prepare_cloned_fields(instance):
...
+    query_dict_keys = sorted(list(query_dict.keys()))
+    for param in view.non_filter_params:
+        if param in query_dict_keys:
+            query_dict_keys.remove(param)
+    filter_params = current_saved_view.config.get("filter_params", {})
+
+    if set(query_dict_keys) != set(filter_params.keys()):
+        return True
```
glennmatthews, line 302: Similar kind of concern here - if there are no query_dict_keys at all that should be OK right?

nautobot/project-static/js/forms.js
```diff
@@ -652,6 +655,30 @@ function initializeDynamicFilterForm(context){
...
+        const per_page = urlParams.get('per_page');
+        var per_page_query_string = ""
+        if (per_page != null) {
+            per_page_query_string = "&per_page=" + per_page
+        }
+        const table_changes = urlParams.get('table_changes_pending');
+        var table_changes_query_string = ""
+        if (table_changes != null) {
+            table_changes_query_string = "&table_changes_pending=" + table_changes
+        }
```
glennmatthews, line 681: This still feels sketchy to me. Can't we just assign these parameters directly into `search_query` rather than manually constructing a bunch of distinct query strings?

nautobot/project-static/js/tableconfig.js
```diff
@@ -39,7 +39,12 @@ $(document).ready(function() {
             data: JSON.stringify(data),
         }).done(function () {
             // Reload the page
-            window.location.reload(true);
+            if (window.location.href.includes("saved_view=")){
+                var new_url = window.location.href+"&table_changes_pending=true";
```
glennmatthews, line 43: This is fragile, right? If the user removes "table_changes_pending=true" from their URL, then it reverts back to the saved-view despite the fact that their user-config has been updated in the DB? I don't immediately have an idea for a better approach but I think it's worth some thought.

nautobot/users/views.py
```diff
@@ -213,6 +226,130 @@ def post(self, request):
...
+        """
+        This method will extract filter_params, pagination and sort_order from request.GET and apply it to the SavedView specified
+        """
+        sv = get_object_or_404(SavedView, pk=request.GET.get("saved_view", None))
+        table_changes_pending = request.GET.get("table_changes_pending", False)
+        pagination_count = request.GET.get("per_page", None)
+        if pagination_count is not None:
+            sv.config["pagination_count"] = int(pagination_count)
+        sort_order = request.GET.getlist("sort", [])
+        sv.config["sort_order"] = sort_order
```
glennmatthews, line 284: Should this also only be updated if the sort order is non-empty?

nautobot/users/views.py
```diff
@@ -213,6 +226,130 @@ def post(self, request):
...
+
+        filter_params = {}
+        for key in request.GET:
+            if key in self.non_filter_params:
+                continue
+            filter_params[key] = request.GET.getlist(key)
+        sv.config["filter_params"] = filter_params
```
glennmatthews, line 291: Should this be only updated if the filter_params are non-empty?

nautobot/users/views.py
```diff
@@ -213,6 +226,130 @@ def post(self, request):
...
+        table_changes_pending = request.GET.get("table_changes_pending", False)
+        try:
+            sv = SavedView.objects.create(name=name, owner=request.user, view=view_name)
+        except IntegrityError:
+            messages.error(request, f"You already have a Saved View named '{name}' for this view '{view_name}'")
+            if derived_view_pk:
+                return redirect(self.get_return_url(request, obj=derived_instance))
+            else:
+                return redirect(reverse(view_name))
+        pagination_count = request.GET.get("per_page", 50)
```
glennmatthews, line 325: Presumably this should default to the user's saved default pagination count, not a hard-coded 50?

nautobot/project-static/js/forms.js
```diff
@@ -635,15 +635,19 @@ function initializeDynamicFilterForm(context){
         let query_params = new URLSearchParams(location.search);
         let type = $(this).attr("data-field-type");
         let field_value = $(this).attr("data-field-value");
+        field_value = field_value.replace("#", "%23")
```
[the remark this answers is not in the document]
HanlinMiao: Yeah to be honest, it should be all similar special characters like this. We can discuss this in standup

nautobot/extras/signals.py
```diff
@@ -137,7 +137,9 @@ def _handle_changed_object(sender, instance, raw=False, **kwargs):
 
         # Generate a unique identifier for this change to stash in the change context
         # This is used for deferred change logging and for looking up related changes without querying the database
-        unique_object_change_id = f"{changed_object_type.pk}__{changed_object_id}__{user.pk}"
+        unique_object_change_id = None
+        if user is not None:
+            unique_object_change_id = f"{changed_object_type.pk}__{changed_object_id}__{user.pk}"
```
glennmatthews, line 142:
```suggestion
        if user is not None:
            unique_object_change_id = f"{changed_object_type.pk}__{changed_object_id}__{user.pk}"
        else:
            unique_object_change_id = f"{changed_object_type.pk}__{changed_object_id}"
```

nautobot/core/models/__init__.py
```diff
@@ -68,6 +68,7 @@ def get_absolute_url(self, api=False):
 
             for action in actions:
                 route = get_route_for_model(self, action, api=api)
+                route = route.replace("users:", "user:")
```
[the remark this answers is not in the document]
HanlinMiao: Resolved in standup

nautobot/core/utils/lookup.py
```diff
@@ -95,6 +96,10 @@ def get_route_for_model(model, action, api=False):
     if apps.get_app_config(app_label).name in settings.PLUGINS:
         viewname = f"plugins{suffix}:{viewname}"
 
+    if model is SavedView:
+        new_viewname = viewname.replace("users:", "user:")
+        return new_viewname
```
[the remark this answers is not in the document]
HanlinMiao: Resolved in standup

nautobot/project-static/js/forms.js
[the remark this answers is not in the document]
HanlinMiao: Deleted and worked as expected :(

nautobot/core/utils/lookup.py
HanlinMiao: No longer needed after resolution in standup

nautobot/project-static/js/tableconfig.js
HanlinMiao: Yeah, the problem is how to represent that intermediate state (e.g. after the user made changes and before the user updated or saved those changes to a saved view) and render the intermediate state table. I agree this is fragile but I think this might be our best bet given how BaseTable is rendered.
Since user_config will be deprecated eventually, it could just be used as an intermediate saved point for table config and such.

nautobot/project-static/js/forms.js
HanlinMiao: Refactored

nautobot/core/templates/generic/object_list.html
```diff
@@ -46,6 +46,61 @@
...
+                            <i class="mdi mdi-content-save-plus-outline text-muted" aria-hidden="true"></i>
+                            <span>Save As New View</span>
+                        </a>
+                    </li>
+                {% endif %}
+                {% if current_saved_view %}
+                    <li>
+                        <a href="{% url 'user:savedview' pk=request.GET.saved_view %}">
+                            <i class="mdi mdi-arrow-u-right-bottom text-danger" aria-hidden="true"></i>
+                            <span class="text-danger">Clear View</span>
```
timizuoebideri1, line 97: This, I assume, is used to undo all unsaved modifications to the view and restore the UI to its initial view state?

timizuoebideri1: If so I don't think `Clear View` might be the right wording here

nautobot/core/views/generic.py
```diff
@@ -184,11 +186,14 @@ class ObjectListView(ObjectPermissionRequiredMixin, View):
...
         "per_page",  # used by get_paginate_count
         "sort",  # table sorting
+        "saved_view",  # saved_view indicator pk or composite keys
+        "table_changes_pending",  # indicator for if there is any table changes not applied to the saved view
     )
 
     def get_filter_params(self, request):
         """Helper function - take request.GET and discard any parameters that are not used for queryset filtering."""
         filter_params = request.GET.copy()
+        print(filter_params)
```
timizuoebideri1, line 196: Remove?

timizuoebideri1 (approved):

nautobot/core/models/generics.py
```diff
@@ -6,17 +6,18 @@
...
 from nautobot.extras.models.mixins import ContactMixin, DynamicGroupMixin, NotesMixin, StaticGroupMixin
 from nautobot.extras.models.relationships import RelationshipModel
+from nautobot.users.models import SavedViewMixin
 
 logger = logging.getLogger(__name__)
 
 
 class OrganizationalModel(
     ChangeLoggedModel,
-    ContactMixin,
```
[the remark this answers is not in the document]
HanlinMiao: Hmm i don't remember removing this, might have been a merge oversight

nautobot/core/views/generic.py
```diff
@@ -147,12 +149,23 @@ class ObjectListView(ObjectPermissionRequiredMixin, View):
...
     def get_filter_params(self, request):
         """Helper function - take request.GET and discard any parameters that are not used for queryset filtering."""
-        filter_params = request.GET.copy()
-        return get_filterable_params_from_filter_params(filter_params, self.non_filter_params, self.filterset())
+        params = request.GET.copy()
+        filter_params = get_filterable_params_from_filter_params(params, self.non_filter_params, self.filterset())
+        if (
+            params.get("saved_view")
+            and not filter_params
+            and not params.get("table_changes_pending")
```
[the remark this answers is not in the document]
HanlinMiao: You might have made this change [here](https://github.com/nautobot/nautobot/pull/5682/files#diff-6d6d85115c2145f877fca192b072690e4d8945d576ca00bff5ad903cce84c2bc). Sorry I didn't catch it when I reviewed it. At the moment it felt like it made sense, I think I should remove it tho.

nautobot/core/views/mixins.py
```diff
@@ -459,8 +460,16 @@ def get_object(self):
...
     def get_filter_params(self, request):
         """Helper function - take request.GET and discard any parameters that are not used for queryset filtering."""
-        filter_params = request.GET.copy()
-        return get_filterable_params_from_filter_params(filter_params, self.non_filter_params, self.filterset_class())
+        params = request.GET.copy()
+        filter_params = get_filterable_params_from_filter_params(params, self.non_filter_params, self.filterset_class())
+        if (
+            params.get("saved_view")
+            and not filter_params
+            and not params.get("table_changes_pending")
```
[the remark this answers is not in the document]
HanlinMiao: Will remove

nautobot/users/models.py
```diff
@@ -282,3 +284,45 @@ def list_constraints(self):
...
+    name = models.CharField(max_length=CHARFIELD_MAX_LENGTH, blank=False, null=False, help_text="The name of this view")
+    view = models.CharField(
+        max_length=CHARFIELD_MAX_LENGTH,
+        blank=False,
+        null=False,
+        help_text="The name of the list view that the saved view is derived from, e.g. dcim:device_list",
+    )
+    config = models.JSONField(
+        encoder=DjangoJSONEncoder, blank=True, default=dict, help_text="Saved Configuration on this view"
+    )
```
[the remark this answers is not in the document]
HanlinMiao: I am not sure this is feasible right now because our factory is using randomized `pydict` data. All those would fail to be created if I assert any structures on the `config` data.

glennmatthews (approved):

nautobot/users/views.py
```diff
@@ -213,6 +231,206 @@ def post(self, request):
...
+                if sv.config.get("table_config", None) is None:
+                    sv.config["table_config"] = {}
+                sv.config["table_config"][f"{table_class}"] = request.user.get_config(f"tables.{table_class}")
+
+        sv.validated_save()
+        list_view_url = sv.get_absolute_url()
+        messages.success(request, f"Successfully updated current view {sv.name}")
+        return redirect(list_view_url)
+
+    def create(self, request, *args, **kwargs):
```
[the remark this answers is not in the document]
HanlinMiao: I agree that there is a lot of repeated logic but they are all subtly different in some ways and need a lot of special case handling.  I can open a follow-up issue for this.

PR #5150, opened 2024-01-22
Updates the Job List to show Job Hook Receiver and Job Button Receiver Jobs

nicholaskuechler:
# Closes #5149
# What's Changed
Updated the Job List to show Job Hook Receiver and Job Button Receiver Jobs, which were previously being hidden from view.

# Screenshots

URL: http://localhost:8080/extras/jobs/

BEFORE:

Screenshot 2024-01-22 at 3 50 48 PM

AFTER:

Screenshot 2024-01-22 at 3 39 06 PM

PR #5642, opened 2024-04-26
Add more cases where it may be necessary to refresh Job class code on the fly

### LINKED ISSUE #5537: Rollback Jobs as Celery Singleton Classes

glennmatthews:
# Closes #5537 (a followup to #5586)

The implementation of `JobModel.job_class` as a `cached_property` was problematic for long running (i.e. production, not runserver) server processes so I've changed it to just a regular `property`. Since its underlying implementation uses `get_job(class_path)` which just pulls directly from the registry, this is effectively removing a layer of no-longer-needed caching that was causing problems.

Secondarily, there are a number of cases where pulling from the registry might still not be up-to-date, again especially in the long-running server process case. I've replaced a number of cases where we were using `job_model.job_class` with `get_job(job_model.class_path, reload=True)` where it's critical to ensure that we have the latest job class code in memory.

Thirdarily, I've made `get_job(class_path, reload=True)` smarter so that it only actually reimports jobs if the specified class_path is not a system or App job (as those aren't actually reloadable).

Some testing in a user environment appears to show that this resolves some pesky issues with Job class loading from Git.

# TODO

- n/a Unit, Integration Tests

### DIFF
--- a/nautobot/extras/api/views.py
+++ b/nautobot/extras/api/views.py
@@ -971,7 +971,7 @@ class ScheduledJobViewSet(ReadOnlyModelViewSet):
 
         # Immediately enqueue the job
         job_class = get_job(job_model.class_path, reload=True)
-        job_kwargs = job_class.prepare_job_kwargs(scheduled_job.kwargs.get("data", {}))
+        job_kwargs = job_class.prepare_job_kwargs(scheduled_job.kwargs or {})
         job_kwargs["dryrun"] = True
         job_result = JobResult.enqueue_job(
             job_model,
--- a/nautobot/extras/api/views.py
+++ b/nautobot/extras/api/views.py
@@ -33,6 +33,7 @@ from nautobot.core.models.querysets import count_related
 from nautobot.extras import filters
 from nautobot.extras.choices import JobExecutionType
 from nautobot.extras.filters import RoleFilterSet
+from nautobot.extras.jobs import get_job
 from nautobot.extras.models import (
     ComputedField,
     ConfigContext,
@@ -520,7 +521,7 @@ class JobViewSetBase(
     def variables(self, request, *args, **kwargs):
         """Get details of the input variables that may/must be specified to run a particular Job."""
         job_model = self.get_object()
-        job_class = job_model.job_class
+        job_class = get_job(job_model.class_path, reload=True)
         if job_class is None:
             raise Http404
         variables_dict = job_class._get_vars()
@@ -602,7 +603,7 @@ class JobViewSetBase(
                     "One of these two flags must be removed before this job can be scheduled or run."
                 )
 
-        job_class = job_model.job_class
+        job_class = get_job(job_model.class_path, reload=True)
         if job_class is None:
             raise MethodNotAllowed(
                 request.method, detail="This job's source code could not be located and cannot be run"
@@ -669,7 +670,7 @@ class JobViewSetBase(
         cleaned_data = None
         try:
             cleaned_data = job_class.validate_data(data, files=files)
-            cleaned_data = job_model.job_class.prepare_job_kwargs(cleaned_data)
+            cleaned_data = job_class.prepare_job_kwargs(cleaned_data)
 
         except FormsValidationError as e:
             # message_dict can only be accessed if ValidationError got a dict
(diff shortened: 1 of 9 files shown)


nautobot/extras/jobs.py
```diff
@@ -1093,7 +1093,18 @@ def get_job(class_path, reload=False):
...
+            then refresh **all** such Jobs before retrieving the job class.
     """
+    if reload:
+        if class_path.startswith("nautobot."):
+            # System job - not reloadable
+            reload = False
+        if any(class_path.startswith(f"{app_name}.") for app_name in settings.PLUGINS):
+            # App provided job - not reloadable
+            reload = False
     jobs = get_jobs(reload=reload)
```
glennmatthews, line 1108: A further optimization would be to have an option to pass through to `get_jobs` to *only* reload the specific data source that contains the requested class_path (i.e. JOBS_ROOT or a single GitRepository). That'd be a bit more invasive though.

nautobot/extras/models/jobs.py
```diff
@@ -290,10 +295,13 @@ def job_task(self):
 
     def clean(self):
         """For any non-overridden fields, make sure they get reset to the actual underlying class value if known."""
-        if self.job_class is not None:
+        from nautobot.extras.jobs import get_job
+
+        job_class = get_job(self.class_path, reload=True)
```
glennmatthews, line 300: In the UI edit case this will result in reloading the jobs twice - once in the form clean(), once in the model clean(). Might be possible to refactor the form clean so that it doesn't need to do this?

nautobot/extras/tests/test_views.py
```diff
@@ -1765,10 +1765,8 @@ class JobTestCase(
     model = Job
 
     def _get_queryset(self):
-        """Don't include hidden Jobs, non-installed Jobs, JobHookReceivers or JobButtonReceivers as they won't appear in the UI by default."""
-        return self.model.objects.filter(
-            installed=True, hidden=False, is_job_hook_receiver=False, is_job_button_receiver=False
-        )
+        """Don't include hidden Jobs or non-installed Jobs, as they won't appear in the UI by default."""
+        return self.model.objects.filter(installed=True, hidden=False)
```
glennmatthews, line 1769: Just a minor test bugfix after #5150.

nautobot/extras/views.py
```diff
@@ -1307,7 +1307,8 @@ def get(self, request, class_path=None, pk=None):
     def post(self, request, class_path=None, pk=None):
         job_model = self._get_job_model_or_404(class_path, pk)
 
-        job_form = job_model.job_class.as_form(request.POST, request.FILES) if job_model.job_class is not None else None
+        job_class = get_job(job_model.class_path, reload=True)
```
glennmatthews, line 1310: Might seem unnecessary as we *usually* do a GET before a POST to this endpoint, but in the job-button case we don't ever hit the GET side, just POST.

nautobot/extras/api/views.py
```diff
@@ -969,13 +970,14 @@ def dry_run(self, request, pk):
             raise PermissionDenied("You do not have permission to run this job.")
 
         # Immediately enqueue the job
-        job_kwargs = job_model.job_class.prepare_job_kwargs(scheduled_job.kwargs.get("data", {}))
+        job_class = get_job(job_model.class_path, reload=True)
+        job_kwargs = job_class.prepare_job_kwargs(scheduled_job.kwargs.get("data", {}))
```
gsnider2195, line 974: Unrelated to this PR but this is a bug and should be:

```suggestion
        job_kwargs = job_class.prepare_job_kwargs(scheduled_job.kwargs or {})
```

glennmatthews: Good catch! Surprised our tests don't pick up on this - I'll fix it and look at adding a quick test case.

PR #5646, opened 2024-04-29
Fix quoting of query parameters in list view

glennmatthews:
# Closes [GHSA-jxgr-gcj5-cqqg](https://github.com/nautobot/nautobot/security/advisories/GHSA-jxgr-gcj5-cqqg)
# What's Changed

- Add proper quoting of query parameters in object-list template to avoid reflected-XSS type attacks.
- Add unit test demonstrating proper behavior.

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/core/tests/test_views.py
+++ b/nautobot/core/tests/test_views.py
@@ -242,6 +242,25 @@ class FilterFormsTestCase(TestCase):
         self.assertInHTML(locations[0].name, response_content)
         self.assertInHTML(locations[1].name, response_content)
 
+    def test_filtering_crafted_query_params(self):
+        """Test for reflected-XSS vulnerability GHSA-jxgr-gcj5-cqqg."""
+        self.add_permissions("dcim.view_location")
+        query_param = "?location_type=1 onmouseover=alert('hi') foo=bar"
+        url = reverse("dcim:location_list") + query_param
+        response = self.client.get(url)
+        self.assertHttpStatus(response, 200)
+        response_content = response.content.decode(response.charset)
+        # The important thing here is that the data-field-parent and data-field-value are correctly quoted
+        self.assertInHTML(
+            """
+<span class="filter-selection-choice-remove remove-filter-param"
+      data-field-type="child"
+      data-field-parent="location_type"
+      data-field-value="1 onmouseover=alert(&#x27;hi&#x27;) foo=bar"
+>×</span>""",  # noqa: RUF001 - ambiguous-unicode-character-string
+            response_content,
+        )
+
 
 class ForceScriptNameTestcase(TestCase):
     """Basic test to assert that `settings.FORCE_SCRIPT_NAME` works as intended."""
--- a/nautobot/project-static/js/forms.js
+++ b/nautobot/project-static/js/forms.js
@@ -644,7 +644,7 @@ function initializeDynamicFilterForm(context){
             let parent = $(this).attr("data-field-parent");
             query_params.delete(parent, field_value);
         }
-        location.replace("?" + query_params);
+        location.assign("?" + query_params);
     })
 
     // On submit of filter form
@@ -660,7 +660,7 @@ function initializeDynamicFilterForm(context){
 
         // Get the serialized data from the forms and:
         // 1) filter out query_params which values are empty e.g ?sam=&dan=2 becomes dan=2
-        // 2) combine the two forms into a single form without duplicate entries
+        // 2) combine the two forms into a single set of data without duplicate entries
         let search_query = new URLSearchParams();
         let dynamic_query = new URLSearchParams(new FormData(document.getElementById("dynamic-filter-form")));
         dynamic_query.forEach((value, key) => { if (value != "") { search_query.append(key, value); }});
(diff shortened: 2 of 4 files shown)


nautobot/core/tests/test_views.py
```diff
@@ -242,6 +242,25 @@ def test_filtering_on_custom_select_filter_field(self):
         self.assertInHTML(locations[0].name, response_content)
         self.assertInHTML(locations[1].name, response_content)
 
+    def test_filtering_crafted_query_params(self):
```
gsnider2195, line 245: Wonder if we could make this more generic without making it ridiculously complex or just grepping the entire codebase

gsnider2195 (approved):

nautobot/project-static/js/forms.js
```diff
@@ -632,18 +632,19 @@ function initializeDynamicFilterForm(context){
...
-        let query_string = location.search.substr(1).split("&");
 
         if (type === "parent") {
-            query_string = query_string.filter(item => item.search(field_value) < 0);
+            // Remove all instances of this query param
+            query_params.delete(field_value);
         } else {
+            // Remove this specific instance of this query param
             let parent = $(this).attr("data-field-parent");
-            query_string = query_string.filter(item => item.search(parent + "=" + field_value) < 0)
```
glennmatthews, line 644: Before the security fix above, with a query of `?name=Hello World`, we'd incorrectly have `data-field-value="Hello" world=""` on the `remove-filter-param` element. This would result in `field_value = "Hello"` and `query_string = ["name=Hello%20World"]`, but since we were (incorrectly IMHO) doing a fuzzy match here (`item.search("name=Hello")`) to find the query param to remove, it still "worked".

After the security fix, with corrected `data-field-value="Hello World"`, we'd correctly get `field_value = "Hello World"` but this then wouldn't match the URL-encoded `query_string` array, which had the net result that clicking the "x" button in the UI wouldn't actually remove such a query parameter from the view. (Yay! Integration tests actually caught a bug!)

The fix here is to not use string-splitting and regex-searching to manipulate the query_params, but instead to use the dedicated `URLSearchParams` interface available in all modern browsers.

nautobot/project-static/js/forms.js
```diff
@@ -657,11 +658,17 @@ function initializeDynamicFilterForm(context){
...
+        // 2) combine the two forms into a single form without duplicate entries
+        let search_query = new URLSearchParams();
+        let dynamic_query = new URLSearchParams(new FormData(document.getElementById("dynamic-filter-form")));
+        dynamic_query.forEach((value, key) => { if (value != "") { search_query.append(key, value); }});
+        let default_query = new URLSearchParams(new FormData(document.getElementById("default-filter").firstElementChild));
+        default_query.forEach((value, key) => {
+            if (value != "" && !search_query.has(key, value)) { search_query.append(key, value); }
+        });
         // Union Operation
-        let search_query = [...new Set([...default_filter_form_query, ...dynamic_filter_form_query])].join("&")
```
glennmatthews, line 664: This set of changes doesn't address a bug so far as I know, but while I was in here, this similarly replaces some string-splitting logic for processing form data as query parameters with the use of URLSearchParams and FormData instead, getting us ever-so-slightly further away from jQuery in the process. I did some manual testing here as well as the integration-tests passing still so I'm reasonably confident in this change.

nautobot/core/tests/test_views.py
glennmatthews: Agreed that would be nice.

We could theoretically add this particular logic to the generic view testing instead of being a single dedicated test, but since the faulty logic was common to all uses of `object_list.html` and involved a single common fix, I didn't feel like that would actually add significant value versus the increased test run time.

I'm always a bit surprised to rediscover that there isn't an `assertNotInHTML` in Django test. It'd be kinda neat to have a set of generic tests that throw various "bad" input into query parameters and assert that that input isn't present in the rendered page, or if present is sanitized appropriately.

nautobot/project-static/js/forms.js
```diff
@@ -657,11 +658,17 @@ function initializeDynamicFilterForm(context){
...
+        // 1) filter out query_params which values are empty e.g ?sam=&dan=2 becomes dan=2
+        // 2) combine the two forms into a single form without duplicate entries
+        let search_query = new URLSearchParams();
+        let dynamic_query = new URLSearchParams(new FormData(document.getElementById("dynamic-filter-form")));
+        dynamic_query.forEach((value, key) => { if (value != "") { search_query.append(key, value); }});
+        let default_query = new URLSearchParams(new FormData(document.getElementById("default-filter").firstElementChild));
+        default_query.forEach((value, key) => {
+            if (value != "" && !search_query.has(key, value)) { search_query.append(key, value); }
+        });
         // Union Operation
```
gsnider2195, line 671: Remove?

nautobot/project-static/js/forms.js
gsnider2195: Did you test these additional fixes on chrome and firefox?

glennmatthews: Just Firefox so far, but will try on Chrome as well. According to https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#browser_compatibility the only one we should need to worry about is Samsung's mobile browser (??)

gsnider2195: Anyone using that browser also has chrome available so it's probably ok?

gsnider2195: Actually, I'll test on samsung internet

glennmatthews: Works fine in Chrome too.

PR #5657, opened 2024-05-01
Add device modules

### LINKED ISSUE #2101: Model Chassis Switches (Module support)
### As ...

Nelly - Network Engineer

### I want ...

### So that ...

- [ ] Automation
- [ ] Circuits
- [X] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

### External Dependencies

N/A

gsnider2195:
# Closes #2101
# What's Changed

This is the feature branch for the device modules feature

- #5667 
- #5724 

# Screenshots

# TODO

- [ ] Unit, Integration Tests
- [ ] Add `name` field to `ModuleBay` and `ModuleBayTemplate` and remove uniqueness constraint on `position` field
- [ ] #5730


nautobot/dcim/api/urls.py
```diff
@@ -81,5 +81,10 @@
 router.register("controllers", views.ControllerViewSet)
 router.register("controller-managed-device-groups", views.ControllerManagedDeviceGroupViewSet)
 
+# Modules
+router.register("modules", views.ModuleViewSet)
+router.register("module-types", views.ModuleTypeViewSet)
+router.register("module-bays", views.ModuleBayViewSet)
```
glennmatthews, line 87: No URL/viewset for ModuleBayTemplate?

nautobot/dcim/factory.py
```diff
@@ -711,3 +714,87 @@ class Params:
...
+    )
+    location = factory.Maybe(
+        "has_parent_module_bay",
+        None,
+        random_instance(
+            lambda: Location.objects.filter(location_type__content_types=ContentType.objects.get_for_model(Module)),
+            allow_null=False,
+        ),
+    )
+    role = random_instance(lambda: Role.objects.get_for_model(Module))
```
glennmatthews, line 795: Should role have a Maybe since it's optional?

nautobot/dcim/factory.py
```diff
@@ -711,3 +714,87 @@ class Params:
...
+    has_asset_tag = NautobotBoolIterator()
+    asset_tag = factory.Maybe("has_asset_tag", UniqueFaker("uuid4"), None)
+    has_serial = NautobotBoolIterator()
+    serial = factory.Maybe("has_serial", factory.Faker("ean", length=8), "")
+    tenant = random_instance(Tenant)
```
glennmatthews, line 800: Similarly - Maybe tenant?

nautobot/dcim/tests/test_api.py
```diff
@@ -2904,3 +2907,47 @@ def setUpTestData(cls):
...
+            manufacturer=Manufacturer.objects.first(),
+            model="Test Module Type",
+        )
+        module_bay = ModuleBay.objects.create(
+            device=Device.objects.first(),
+            position="0",
+        )
+        status = Status.objects.get_for_model(Module).first()
+        location = Location.objects.first()
+        location.location_type.content_types.add(ContentType.objects.get_for_model(Module))
```
glennmatthews, line 2927: Can't you just do `Location.objects.get_for_model(Module).first()`?

nautobot/dcim/factory.py
gsnider2195: Oops! I was getting `random_instance(allow_null=True)` with the logic you implemented for `get_random_instances`:

>    This is not an evenly weighted distribution (all counts equally likely), because in most of our code,
>    the relevant code paths distinguish between 0, 1, or >1 instances - there's not a functional difference between
>    "2 instances" and "10 instances" in most cases. Therefore, this implementation provides:
>        - 1/3 chance of no instances
>        - 1/3 chance of 1 instance
>        - 1/3 chance of (2 to n) instances, where each possibility is equally likely within this range

nautobot/dcim/api/urls.py
gsnider2195: With the complexity of the models I'd like to focus on only the model, factory and model test code in this PR and save the Rest API for the next PR, hence the incomplete implementation here. I can go ahead and move this incomplete API code to another branch to make this review easier.

nautobot/dcim/models/device_component_templates.py
```diff
@@ -79,6 +81,9 @@ def to_objectchange(self, action, **kwargs):
         """
         Return a new ObjectChange with the `related_object` pinned to the `device_type` by default.
         """
+        if "related_object" in kwargs:
+            return super().to_objectchange(action, **kwargs)
+
```
gsnider2195, line 86: Don't clobber a child class's explicitly set kwarg.

nautobot/dcim/models/device_component_templates.py
```diff
@@ -387,3 +447,83 @@ def clean(self):
...
+        blank=False,
+        null=False,
+        help_text="The position of the module bay within the device",
+    )
+    label = models.CharField(max_length=CHARFIELD_MAX_LENGTH, blank=True, help_text="Physical label")
+    description = models.CharField(max_length=CHARFIELD_MAX_LENGTH, blank=True)
+
+    class Meta:
+        ordering = ("device_type", "module_type", "position")
+        unique_together = (("device_type", "position"), ("module_type", "position"))
```
gsnider2195, line 479: Need to change this to a UniqueConstraint like the other new models

nautobot/dcim/models/device_component_templates.py
```diff
@@ -387,3 +447,83 @@ def clean(self):
             raise ValidationError(
                 f'Subdevice role of device type ({self.device_type}) must be set to "parent" to allow device bays.'
             )
+
+
+@extras_features("custom_validators")
+class ModuleBayTemplate(BaseModel, ChangeLoggedModel, CustomFieldModel, RelationshipModel):
```
gsnider2195, line 453: Can this be a subclass of ModularComponentTemplate?

nautobot/dcim/models/device_components.py
```diff
@@ -94,6 +98,59 @@ def to_objectchange(self, action, **kwargs):
...
+        blank=True,
+        null=True,
+    )
+
+    natural_key_field_names = ["name", "device", "module"]
+
+    class Meta:
+        abstract = True
+        constraints = [
+            models.CheckConstraint(
```
gsnider2195, line 120: Probably need a short comment here explaining what this does. Same for the other CheckConstraints added in this PR

nautobot/dcim/models/device_components.py
```diff
@@ -1074,7 +1131,8 @@ class InventoryItem(TreeModel, ComponentModel):
         related_name="inventory_items",
         blank=True,
         null=True,
-        verbose_name="The software version installed on this inventory item",
+        verbose_name="Software Version",
+        help_text="The software version installed on this inventory item",
```
gsnider2195, line 1135: Bug fix unrelated to device modules

nautobot/dcim/models/device_components.py
```diff
@@ -1096,3 +1154,101 @@ def natural_key_field_lookups(cls):  # pylint: disable=no-self-argument
...
+        max_length=CHARFIELD_MAX_LENGTH,
+        help_text="The position of the module bay within the parent device/module",
+    )
+    label = models.CharField(max_length=CHARFIELD_MAX_LENGTH, blank=True, help_text="Physical label")
+    description = models.CharField(max_length=CHARFIELD_MAX_LENGTH, blank=True)
+
+    clone_fields = ["parent_device", "parent_module"]
+
+    class Meta:
+        # TODO: Ordering by parent_module.id is not correct but prevents an infinite loop
```
gsnider2195, line 1197: Django can't resolve the multiple inheritance of `ModuleBay` -> `Module` -> `ModuleBay` -> `Module` without resulting in an infinite loop.

nautobot/dcim/models/devices.py
```diff
@@ -1375,3 +1386,353 @@ def clean(self):
...
+
+    def __str__(self):
+        return self.model
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Save references to the original front/rear images
+        self._original_front_image = self.front_image if self.present_in_database else None
+        self._original_rear_image = self.rear_image if self.present_in_database else None
```
gsnider2195, line 1445:
```suggestion
```

nautobot/dcim/models/devices.py
```diff
@@ -1375,3 +1386,353 @@ def clean(self):
...
+
+    def delete(self, *args, **kwargs):
+        super().delete(*args, **kwargs)
+
+        # Delete any uploaded image files
+        if self.front_image:
+            self.front_image.delete(save=False)
+        if self.rear_image:
+            self.rear_image.delete(save=False)
+
```
gsnider2195, line 1551:
```suggestion
```

nautobot/dcim/factory.py
```diff
@@ -711,3 +715,92 @@ class Params:
...
+        factory.LazyAttribute(lambda o: o.parent_module.module_bays.count() + 1),
+        factory.LazyAttribute(lambda o: o.parent_device.module_bays.count() + 1),
+    )
+    has_label = NautobotBoolIterator()
+    label = factory.Maybe(
+        "has_label",
+        factory.Maybe(
+            "has_parent_module",
+            factory.LazyAttribute(lambda o: f"{o.parent_module!s} Bay {o.position}"),
+            factory.LazyAttribute(lambda o: f"{o.parent_device!s} Bay {o.position}"),
```
HanlinMiao, line 757: Curious as to what syntax this is🤔

nautobot/dcim/models/device_components.py
```diff
@@ -1096,3 +1154,101 @@ def natural_key_field_lookups(cls):  # pylint: disable=no-self-argument
...
+            ),
+            models.UniqueConstraint(
+                fields=["parent_module", "position"],
+                name="dcim_modulebay_parent_module_position_unique",
+            ),
+        ]
+
+    @property
+    def parent(self):
+        return self.parent_device if self.parent_device else self.parent_module
```
HanlinMiao, line 1221: Would this work since they are guaranteed to be mutually exclusive?
```suggestion
        return self.parent_device | self.parent_module
```

nautobot/dcim/models/devices.py
```diff
@@ -1375,3 +1386,353 @@ def clean(self):
...
+      * 2 InterfaceTemplates
+
+    When a new Module of this type is created, the appropriate console, power, and interface
+    objects (as defined by the ModuleType) are automatically created as well.
+    """
+
+    manufacturer = models.ForeignKey(to="dcim.Manufacturer", on_delete=models.PROTECT, related_name="module_types")
+    model = models.CharField(max_length=CHARFIELD_MAX_LENGTH)
+    part_number = models.CharField(
+        max_length=CHARFIELD_MAX_LENGTH, blank=True, help_text="Discrete part number (optional)"
```
HanlinMiao, line 1424: Does `discrete` mean standalone here?

nautobot/dcim/models/devices.py
```diff
@@ -1375,3 +1386,353 @@ def clean(self):
...
+    "export_templates",
+    "graphql",
+    "locations",
+    "statuses",
+    "webhooks",
+)
+class Module(PrimaryModel):
+    """
+    A Module represents a line card, supervisor, or other interchangeable hardware component within a ModuleBay.
+    Each Module is assigned a ModuleType, Role, and (optionally) a Platform.
```
HanlinMiao, line 1569:
```suggestion
    Each Module is assigned a ModuleType, (optionally) Role, and (optionally) a Platform.
```

nautobot/dcim/models/devices.py
```diff
@@ -1375,3 +1386,353 @@ def clean(self):
...
+
+    Each Module must be assigned to either a ModuleBay or a Location.
```
HanlinMiao, line 1571: Nitpick
```suggestion
    Each Module must be assigned to either a ModuleBay or a Location, but not to both.
```

nautobot/dcim/models/devices.py
```diff
@@ -1375,3 +1386,353 @@ def clean(self):
...
+            (ModuleBay, self.module_type.module_bay_templates.all()),
+        ]
+        instantiated_components = []
+        for model, templates in component_models:
+            model.objects.bulk_create([x.instantiate(device=None, module=self) for x in templates])
+        return instantiated_components
+
+    def get_cables(self, pk_list=False):
+        """
+        Return a QuerySet or PK list matching all Cables connected to a component of this Module.
```
HanlinMiao, line 1713: Perhaps?
```suggestion
        Return a QuerySet or PK list matching all Cables connected to all components of this Module.
```

nautobot/dcim/models/device_components.py
jerej: Could there be alternatives like a mechanism to limit the recursion depth? We would need to understand the impact of that recursion on performance, too.

nautobot/dcim/factory.py
gsnider2195: The `!s` in the fstring?

HanlinMiao: oops never mind, didn't see the f-string

nautobot/dcim/models/device_components.py
gsnider2195: This is in the Django SQLCompiler code. We may not want to tread there.

nautobot/dcim/models/devices.py
gsnider2195: In this context, it's more that it's a unique/distinct part number (IE: the manufacturer SKU)

nautobot/dcim/migrations/0060_module_models.py
```diff
@@ -0,0 +1,610 @@
...
+            name="status",
+            field=nautobot.extras.models.statuses.StatusField(
+                on_delete=django.db.models.deletion.PROTECT, related_name="modules", to="extras.status"
+            ),
+        ),
+        migrations.AddField(
+            model_name="module",
+            name="tags",
+            field=nautobot.core.models.fields.TagsField(through="extras.TaggedItem", to="extras.Tag"),
+        ),
```
glennmatthews, line 381: Curious why these aren't included in the `CreateModel` call above?

nautobot/dcim/models/device_component_templates.py
glennmatthews: Yeah, could subclass and override `name` to optional for this specific model, but that might be iffy if there's code assuming that all `ComponentTemplateModels` have a `name`?

nautobot/dcim/models/device_components.py
glennmatthews: Does removing `module_bay` from the Module `ordering` make it possible to order this one correctly? Might be worth considering.

nautobot/dcim/models/devices.py
glennmatthews:
```suggestion
        Return a QuerySet or PK list matching all Cables connected to any component of this Module.
```

nautobot/dcim/models/devices.py
```diff
@@ -1375,3 +1386,329 @@ def clean(self):
...
+            ConsolePort,
+            ConsoleServerPort,
+            PowerPort,
+            PowerOutlet,
+            Interface,
+            FrontPort,
+            RearPort,
+        ]:
+            cable_pks += component_model.objects.filter(module=self, cable__isnull=False).values_list(
+                "cable", flat=True
```
glennmatthews, line 1704: Should that be `"cable_id"` instead?

nautobot/dcim/tests/test_models.py
```diff
@@ -1994,3 +2001,560 @@ def test_parent_controller_change(self):
...
+
+    @classmethod
+    def setUpTestData(cls):
+        cls.device = Device.objects.first()
+        cls._delete_module_bays(cls.device)
+        cls.module = Module.objects.first()
+        cls._delete_module_bays(cls.module)
+
+    # TODO: should this be a helper method (cascade_delete?) on the ModuleBay model?
+    # This works around the relationship between ModuleBay and its installed Module being protected
```
glennmatthews, line 2017: Huh. Don't we have `on_delete=CASCADE` on that field? Would like to understand why this is an issue.

glennmatthews (approved):

nautobot/dcim/migrations/0060_module_models.py
gsnider2195: I'm not sure. When I have Django recreate migrations they don't change.

nautobot/dcim/tests/test_models.py
gsnider2195: I had it set to PROTECT before but when I realized that every other component of device is CASCADE I changed it. I will remove this from the test

nautobot/dcim/models/device_components.py
gsnider2195: If we remove `module_bay` from module ordering then we're sorting based on location, which may be null, and then by module type. The current implementation isn't perfect but it keeps modules that belong to the same module bay parent grouped together:

```py
>>> pprint.pprint(list(Module.objects.all()))
[<Module: A10 ModuleType 3 (Asset Tag: 61083969-d305-4acc-bf90-d5216079134f) installed in rtr1 (1)>,
 <Module: Citrix ModuleType 2 (Serial: 61773776) installed in rtr1 (2)>,
 <Module: F5 ModuleType 4 (Asset Tag: b6e1afdd-0c04-48ed-a21a-7964ddbea574) installed in rtr1 (4)>,
 <Module: Palo Alto ModuleType 1 (Asset Tag: d5e77062-af1f-43c2-b0e3-c67b21d7b4d1) installed in rtr1 (5)>,
 <Module: Citrix ModuleType 2 (Serial: 18084627) installed in rtr2 (1)>,
 <Module: A10 ModuleType 3 installed in rtr2 (2)>,
 <Module: Citrix ModuleType 2 (Serial: 68045364) (Asset Tag: e600f5e8-5383-4454-9823-af9475cca1a6) installed in rtr2 (3)>,
 <Module: F5 ModuleType 4 (Serial: 78941250) installed in rtr2 (5)>]
```

nautobot/dcim/models/devices.py
gsnider2195: This was copy/pasted from Device and it works there:

```py
>>> Device.objects.first().get_cables()
<RestrictedQuerySet [<Cable: #408b097b-4011-40eb-b448-62e5ae9183f2>, <Cable: #2ea6dbac-9936-4207-9d30-61c5a1b3c8a2>]>
```

gsnider2195:
> Any idea about the migrations failures? Looks like the test failure is due to needing a custom graphql type declaration for at least one of the new models, should be fairly straightforward to add?

The migrations should be good now. I needed to re-order the operations to make mysql happy. 

There are a bunch of test failures that won't be resolved until we get at least all of the rest api endpoints added. I'm hoping we can get tests passing in #5667

timizuoebideri1 (approved):

timizuoebideri1 (approved):

nautobot/dcim/api/serializers.py
```diff
@@ -651,38 +692,43 @@ class DeviceNAPALMSerializer(serializers.Serializer):
...
     PathEndpointModelSerializerMixin,
+    NautobotModelSerializer,
 ):
     type = ChoiceField(choices=ConsolePortTypeChoices, allow_blank=True, required=False)
 
     class Meta:
         model = ConsoleServerPort
         fields = "__all__"
         extra_kwargs = {"cable": {"read_only": True}}
+        validators = []
```
glennmatthews, line 707: Not blocking, but it might be nice to move these three lines into `ModularDeviceComponentSerializerMixin.Meta` and have all of these component serializer Metas just inherit from it.

nautobot/dcim/models/device_components.py
```diff
@@ -1096,3 +1130,102 @@ def natural_key_field_lookups(cls):  # pylint: disable=no-self-argument
...
+
+    @property
+    def device(self):
+        """Walk up parent chain to find the Device that this ModuleBay is installed in, if one exists."""
+        if self.parent_device:
+            return self.parent_device
+        return self.parent_module.device
```
glennmatthews, line 1205: I'm not a big fan of `parent` having a different meaning for the `ModularComponentModels` than it does for `Module`. Might it be better to be consistently "wrong" here instead?

nautobot/dcim/models/devices.py
```diff
@@ -1375,3 +1386,329 @@ def clean(self):
...
+        if self.serial == "":
+            self.serial = None
+        if self.asset_tag == "":
+            self.asset_tag = None
+
+        super().save(*args, **kwargs)
+
+        # If this is a new Module, instantiate all related components per the ModuleType definition
+        if is_new:
+            self.create_components()
```
glennmatthews, line 1665: Do I remember correctly that there's supposed to be some logic about letting component names have a token for the module-bay position that gets substituted in when the module gets installed into a specific bay? So the same line card would have `Ethernet1/1` if it's in bay 1 but `Ethernet2/1` if it's in bay 2? Fine with me if that's future work at this point.

gsnider2195: That's correct. I was planning on doing that after the UI is completed

nautobot/dcim/models/device_components.py
gsnider2195: That's a good idea

PR #5667, opened 2024-05-02
Device Modules: REST API and Filters

### LINKED ISSUE #2101: Model Chassis Switches (Module support)

gsnider2195:
# Closes #2101
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests


nautobot/dcim/api/urls.py
```diff
@@ -81,5 +81,10 @@
 router.register("controllers", views.ControllerViewSet)
 router.register("controller-managed-device-groups", views.ControllerManagedDeviceGroupViewSet)
 
+# Modules
+router.register("modules", views.ModuleViewSet)
+router.register("module-types", views.ModuleTypeViewSet)
+router.register("module-bays", views.ModuleBayViewSet)
```
glennmatthews, line 87: Need a REST viewset for module-bay-templates as well

nautobot/dcim/filters/__init__.py
```diff
@@ -961,6 +965,11 @@ class DeviceFilterSet(
         field_name="interfaces",
         label="Has interfaces",
     )
+    interfaces = NaturalKeyOrPKMultipleChoiceFilter(
```
glennmatthews, line 968: I see `interfaces` declared in `Meta.fields` - is this just turning the default model choice filter into a NaturalKeyOrPKMultipleChoiceFilter?

Good to document as a change fragment in any case.

nautobot/dcim/filters/__init__.py
```diff
@@ -1925,3 +1934,239 @@ def _subtree(self, queryset, name, value):
...
+                "lookup_expr": "icontains",
+                "preprocessor": str.strip,
+            },
+            "module_type__model": {
+                "lookup_expr": "icontains",
+                "preprocessor": str.strip,
+            },
+        }
+    )
+    installed = RelatedMembershipBooleanFilter(
```
glennmatthews, line 1966: Perhaps `is_installed`?

nautobot/dcim/tests/test_api.py
```diff
@@ -2904,3 +2907,46 @@ def setUpTestData(cls):
...
+    @classmethod
+    def setUpTestData(cls):
+        mt = ModuleType.objects.create(
+            manufacturer=Manufacturer.objects.first(),
+            model="Test Module Type",
+        )
+        module_bay = ModuleBay.objects.create(
+            parent_device=Device.objects.first(),
+            position="0",
+        )
```
glennmatthews, line 2924: can we factory-ize these?

gsnider2195: There are already factories for these. This setup data will likely be removed when I make it back around to working on the rest api after I finish the filters.

nautobot/dcim/filters/__init__.py
```diff
@@ -688,9 +694,6 @@ def _power_outlets(self, queryset, name, value):
     def _interfaces(self, queryset, name, value):
         return queryset.exclude(interface_templates__isnull=value)
 
-    def _pass_through_ports(self, queryset, name, value):
-        return queryset.exclude(front_port_templates__isnull=value, rear_port_templates__isnull=value)
-
```
gsnider2195, line 693: `FrontPortTemplate.rear_port_template` is not nullable so if you have a `FrontPortTemplate`, you have pass through ports. I omitted this filter entirely on `Module` and `ModuleType` for this reason.

nautobot/core/filters.py
```diff
@@ -395,6 +395,10 @@ class NaturalKeyOrPKMultipleChoiceFilter(django_filters.ModelMultipleChoiceFilte
     Filter that supports filtering on values matching the `pk` field and another
     field of a foreign-key related object. The desired field is set using the `to_field_name`
     keyword argument on filter initialization (defaults to `name`).
+
+    NOTE that the `to_field_name` field does not have to be a "true" natural key (ie. unique), it
+    was just the best name we could come up with for this filter
```
glennmatthews, line 400: :+1:

nautobot/dcim/filters/__init__.py
glennmatthews: Should the change above be using `rear_port_templates` instead? Since a device can have rear-ports without any front-ports, but not the reverse?

nautobot/dcim/tests/test_filters.py
```diff
@@ -701,6 +722,282 @@ def common_test_data(cls):
...
+    )
+    Module.objects.create(
+        parent_module_bay=other_parent_module.module_bays.first(),
+        module_type=module_types[2],
+        status=module_status,
+        tenant=tenants[2],
+    )
+
+
+class ComponentTemplateMixin:
```
glennmatthews, line 942: Should all of these mixins be named `....TestMixin` to make them less ambiguous?

gsnider2195:
> Should the change above be using rear_port_templates instead? Since a device can have rear-ports without any front-ports, but not the reverse?

I understood "has pass-through ports" to mean it has both front ports and rear ports. So if you only have rear ports, you don't have pass-through ports, right? But if you have front ports, you have to have at least one of each. The docstrings for FrontPort and RearPort are "A pass-through port on the front/rear of a Device."

glennmatthews:
> > Should the change above be using rear_port_templates instead? Since a device can have rear-ports without any front-ports, but not the reverse?
> 
> I understood "has pass-through ports" to mean it has both front ports and rear ports. So if you only have rear ports, you don't have pass-through ports, right? But if you have front ports, you have to have at least one of each. The docstrings for FrontPort and RearPort are "A pass-through port on the front/rear of a Device."

Right you are. Sorry for the noise.

nautobot/dcim/api/serializers.py
```diff
@@ -759,22 +813,39 @@ def validate(self, data):
         return super().validate(data)
 
 
-class RearPortSerializer(NautobotModelSerializer, TaggedModelSerializerMixin, CableTerminationModelSerializerMixin):
+class RearPortSerializer(
+    ModularDeviceComponentSerializerMixin,
+    NautobotModelSerializer,
+    TaggedModelSerializerMixin,
+    CableTerminationModelSerializerMixin,
+):
```
glennmatthews, line 821: While you're touching all of these anyway, would it make sense to fix the MRO for the other mixins (i.e. make `NautobotModelSerializer` the last in all cases)?

nautobot/dcim/api/urls.py
```diff
@@ -42,6 +45,8 @@
 router.register("rear-ports", views.RearPortViewSet)
 router.register("device-bays", views.DeviceBayViewSet)
 router.register("inventory-items", views.InventoryItemViewSet)
+router.register("module-bays", views.ModuleBayViewSet)
+router.register("modules", views.ModuleViewSet)
```
glennmatthews, line 49: Duplicate of line 36?

nautobot/dcim/api/views.py
```diff
@@ -837,3 +857,27 @@ class ControllerManagedDeviceGroupViewSet(NautobotModelViewSet):
...
+
+class ModuleViewSet(NautobotModelViewSet):
+    queryset = Module.objects.select_related(
+        "parent_module_bay__parent_device__location",
+        "parent_module_bay__parent_device__tenant",
+        "module_type__manufacturer",
+        "tenant",
+        "role",
+        "location",
+    ).prefetch_related("module_bays", "tags")
```
glennmatthews, line 875: Ugh we need so much to make this logic automatic.

nautobot/dcim/filters/__init__.py
```diff
@@ -1054,7 +1066,7 @@ class PathEndpointFilterSet(PathEndpointModelFilterSetMixin):
 
 class ConsolePortFilterSet(
     BaseFilterSet,
-    DeviceComponentModelFilterSetMixin,
+    ModularDeviceComponentModelFilterSetMixin,
     CableTerminationModelFilterSetMixin,
     PathEndpointModelFilterSetMixin,
```
glennmatthews, line 1071: Should we fix up the MRO on all of these as well?

nautobot/dcim/filters/__init__.py
```diff
@@ -1925,3 +1937,329 @@ def _subtree(self, queryset, name, value):
...
+    )
+    has_console_ports = RelatedMembershipBooleanFilter(
+        field_name="console_ports",
+        label="Has console ports",
+    )
+    console_ports = NaturalKeyOrPKMultipleChoiceFilter(
+        queryset=ConsolePort.objects.all(),
+        to_field_name="name",
+        label="Console Ports (name or ID)",
+    )
```
glennmatthews, line 1996: Wonder if it'd be useful to create another shared filterset mixin for DeviceFilterSet and ModuleFilterSet to avoid the need to duplicate all of these port filters.

nautobot/dcim/filters/__init__.py
```diff
@@ -1925,3 +1937,329 @@ def _subtree(self, queryset, name, value):
...
+    console_port_templates = NaturalKeyOrPKMultipleChoiceFilter(
+        to_field_name="name",
+        queryset=ConsolePortTemplate.objects.all(),
+        label="Console port templates (name or ID)",
+    )
+    has_console_port_templates = RelatedMembershipBooleanFilter(
+        field_name="console_port_templates",
+        label="Has console port templates",
+    )
```
glennmatthews, line 2103: Same question for ModuleTypeFilterSet/DeviceTypeFilterSet.

nautobot/dcim/models/device_component_templates.py
```diff
@@ -372,6 +374,8 @@ class FrontPortTemplate(ModularComponentTemplateModel):
         ],
     )
 
+    natural_key_field_names = ["name", "device_type", "module_type", "rear_port_template", "rear_port_position"]
```
glennmatthews, line 377: Should this be, for consistency?

```suggestion
    natural_key_field_names = ["device_type", "module_type", "name", "rear_port_template", "rear_port_position"]
```

nautobot/dcim/models/device_components.py
```diff
@@ -966,12 +958,16 @@ class FrontPort(ModularComponentModel, CableTermination):
         ],
     )
 
-    class Meta:
-        ordering = ("device", "_name")
-        unique_together = (
-            ("device", "name"),
-            ("rear_port", "rear_port_position"),
-        )
+    natural_key_field_names = ["name", "device", "module", "rear_port", "rear_port_position"]
```
glennmatthews, line 961: Similarly, perhaps?

```suggestion
    natural_key_field_names = ["device", "module", "name", "rear_port", "rear_port_position"]
```

nautobot/dcim/tests/test_filters.py
```diff
@@ -696,6 +722,282 @@ def common_test_data(cls):
     parent_controller_managed_device_group.tags.set(Tag.objects.get_for_model(ControllerManagedDeviceGroup))
     cls.controller_managed_device_groups[1].tags.set(Tag.objects.get_for_model(ControllerManagedDeviceGroup)[:3])
 
+    module_types = ModuleType.objects.all()[:3]
+    ConsolePortTemplate.objects.create(
```
glennmatthews, line 726: All of this still needed even after adding factories for these?

nautobot/dcim/tests/test_filters.py
```diff
@@ -2533,19 +2747,14 @@ def setUpTestData(cls):
         device_bays[1].save()
 
 
-class InventoryItemTestCase(FilterTestCases.FilterTestCase):
+class InventoryItemTestCase(DeviceComponentTestMixin, FilterTestCases.FilterTestCase):
     queryset = InventoryItem.objects.all()
     filterset = InventoryItemFilterSet
     generic_filter_tests = [
         ("asset_tag",),
```
glennmatthews, line 2754: Do we need a `*DeviceComponentTestMixin.generic_filter_tests` here?

nautobot/dcim/filters/__init__.py
gsnider2195: It looks like any filterset mixins that are using `get_filters()` to inject filters into the filterset (`StatusModelFilterSetMixin` and `RoleModelFilterSetMixin`) require this broken MRO so that `BaseFilterSet` can add the extra filter lookups to these filters. We'll need to fix those mixins before we can make this change, unfortunately.

```
Correct MRO:
>>> [k for k in DeviceFilterSet().filters.keys() if "status" in k or "role" in k]
['role', 'status']

Old Incorrect MRO:
>>> [k for k in DeviceFilterSet().filters.keys() if "status" in k or "role" in k]
['role', 'status', 'role__n', 'status__n']
```

gsnider2195: This patch appears to fix it:

```diff
diff --git a/nautobot/core/utils/filtering.py b/nautobot/core/utils/filtering.py
index 568c34a02..ad4ed2485 100644
--- a/nautobot/core/utils/filtering.py
+++ b/nautobot/core/utils/filtering.py
@@ -119,6 +119,8 @@ def get_filterset_parameter_form_field(model, parameter, filterset=None):
             "to_field_name": field.extra.get("to_field_name", "id"),
             "query_params": field.extra.get("query_params", {}),
         }
+        form_attr["query_params"].setdefault("content_type", model._meta.label_lower)
+
         # ConfigContext requires content_type set to Device and VirtualMachine
         if model == ConfigContext:
             form_attr["query_params"] = {"content_types": [Device._meta.label_lower, VirtualMachine._meta.label_lower]}
diff --git a/nautobot/extras/filters/mixins.py b/nautobot/extras/filters/mixins.py
index 56d56ba0e..6bc1fd6f0 100644
--- a/nautobot/extras/filters/mixins.py
+++ b/nautobot/extras/filters/mixins.py
@@ -303,17 +303,7 @@ class RoleModelFilterSetMixin(django_filters.FilterSet):
     Mixin to add a `role` filter field to a FilterSet.
     """
 
-    @classmethod
-    def get_filters(cls):
-        filters = super().get_filters()
-
-        if cls._meta.model is not None:
-            filters["role"] = RoleFilter(
-                field_name="role",
-                query_params={"content_types": [cls._meta.model._meta.label_lower]},
-            )
-
-        return filters
+    role = RoleFilter(field_name="role")
 
 
 class StatusFilter(NaturalKeyOrPKMultipleChoiceFilter):
@@ -332,14 +322,4 @@ class StatusModelFilterSetMixin(django_filters.FilterSet):
     Mixin to add a `status` filter field to a FilterSet.
     """
 
-    @classmethod
-    def get_filters(cls):
-        filters = super().get_filters()
-
-        if cls._meta.model is not None:
-            filters["status"] = StatusFilter(
-                field_name="status",
-                query_params={"content_types": [cls._meta.model._meta.label_lower]},
-            )
-
-        return filters
+    status = StatusFilter(field_name="status")
```

timizuoebideri1 (approved):

nautobot/dcim/filters/__init__.py
glennmatthews: That's reverting a change I made recently... IIRC the downside is that now the filters won't have the correct `query_params` to restrict the queryset in the UI? 

As an alternative, do we just need to split the `get_filters` logic into a couple of separate methods so that mixin code gets called correctly before the base filterset adds the extra lookups to the filters?

glennmatthews: I see you added logic back into get_filterset_parameter_form_field to set the content-types, but it's a step backwards from our goal of moving the brains into FilterSet rather than having a separate API to massage the filterset fields into a more correct state.

nautobot/dcim/api/serializers.py
```diff
@@ -1043,13 +1105,90 @@ class Meta:
...
+            validator = UniqueTogetherValidator(
+                queryset=self.Meta.model.objects.all(), fields=("parent_device", "position")
+            )
+            validator(data, self)
+        if data.get("parent_module"):
+            validator = UniqueTogetherValidator(
+                queryset=self.Meta.model.objects.all(), fields=("parent_module", "position")
+            )
+            validator(data, self)
+        return super().validate(data)
```
glennmatthews, line 1145: Just checking - `super().validate()` covers the case where *neither* `parent_device` nor `parent_module` are provided, right?

nautobot/dcim/filters/__init__.py
```diff
@@ -1925,3 +1809,180 @@ def _subtree(self, queryset, name, value):
...
+        label="Module type (model or ID)",
+    )
+    parent_module_bay = django_filters.ModelMultipleChoiceFilter(
+        queryset=ModuleBay.objects.all(),
+        label="Parent Module Bay",
+    )
+
+    class Meta:
+        model = Module
+        fields = "__all__"
```
glennmatthews, line 1864: Something I noticed yesterday is that when we're using `__all__` on our new FilterSets, this includes a `_custom_field_data` filter, which the GraphQL schema generation complains about because GraphQL filters can't start with an underscore. Not blocking for this PR in any way, but we should fix that, possibly by enhancing `NautobotFilterSet` to auto-exclude `_custom_field_data` as a filter?

nautobot/dcim/filters/__init__.py
```diff
@@ -1925,3 +1809,180 @@ def _subtree(self, queryset, name, value):
...
+        label="Parent module (ID)",
+    )
+    has_parent_module = RelatedMembershipBooleanFilter(
+        field_name="parent_module",
+        label="Has parent module",
+    )
+    has_installed_module = RelatedMembershipBooleanFilter(
+        field_name="installed_module",
+        label="Has installed module",
+    )
```
glennmatthews, line 1984: Should we have an `installed_module` filter as well?

glennmatthews (approved):

nautobot/dcim/filters/__init__.py
gsnider2195: We might want to move the dynamic generation of filterset form fields into the filter field instead of a separate utility function. I opened #5700 to track this and I'll be reverting these changes.

nautobot/dcim/api/serializers.py
gsnider2195: That's correct. The model validation actually covers all of these but drf recommends doing the validation on the serializer in case you decide to stop using `ModelSerializer` 🤷

nautobot/dcim/filters/__init__.py
gsnider2195: > #5703.

Thanks!

PR #5682, opened 2024-05-07
Make 'saved_view=...' query param imply the correct filtering, sorting, config, and pagination values

glennmatthews:
# What's Changed

Adjustment to the work being done in #5639.

Make it so that we don't need `?saved_view=...&filter=...&sort=...&per_page=...` when using a Saved View - just specifying `?saved_view=...` implies the correct filter, sort, per_page, and table config values if not explicitly overridden by query params.

Doesn't manage to get rid of the need for `?table_changes_pending` parameter, maybe in the future we can tackle that too.

# Screenshots

`saved_view` alone implies other parameters:

image

Removing a query parameter by clicking the "x" works and is detected as a change to the saved view:

image

Returning to the saved view, can see that the filter forms are also correctly populated:

image
image

Adding a new filter param gets correctly merged with the saved view params and is detected as a change:

image

# TODO

- n/a Unit, Integration Tests

### DIFF
--- a/nautobot/core/views/utils.py
+++ b/nautobot/core/views/utils.py
@@ -12,7 +12,6 @@ from rest_framework import exceptions, serializers
 from nautobot.core.api.fields import ChoiceField, ContentTypeField, TimeZoneSerializerField
 from nautobot.core.api.parsers import NautobotCSVParser
 from nautobot.core.models.utils import is_taggable
-from nautobot.core.utils.config import get_settings_or_config
 from nautobot.core.utils.data import is_uuid
 from nautobot.core.utils.filtering import get_filter_field_label
 from nautobot.core.utils.lookup import get_form_for_model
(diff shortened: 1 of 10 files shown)


glennmatthews:
Looking into unit test failures now.

nautobot/core/views/utils.py
```diff
@@ -284,26 +283,27 @@ def view_changes_not_saved(request, view, current_saved_view):
     If every configuration is the same, return False
     """
     if current_saved_view is None:
-        return False
+        return True
```
HanlinMiao, line 286: I think this still should be False here because this is the case where the user is making modifications to the original list view.

HanlinMiao: It is also not showing as intended because the if block for this boolean is wrapped in another one.
```
<h1>{% block title %}
    {{ title }}
        {% if current_saved_view %}
            — {{ current_saved_view.name }}
            {% if new_changes_not_applied %}
                <sup>*</sup>
            {% endif %}
        {% endif %}
    {% endblock %}
</h1>
```

HanlinMiao (approved):
Awesome thank you so much, there are some behaviors I noticed during manual testing but we can discuss that in standup

PR #5695, opened 2024-05-09
[LTM] allow a GitRepository to 'reclaim' Jobs from a previously deleted repository

### LINKED ISSUE #2974: [LTM] Deleting/Restoring git repo with jobs results in database error
### Environment
* Nautobot version (Docker tag too if applicable): 1.4.5
* Python version: 3.8
* Database platform, version: PostgreSQL 12

### Steps to Reproduce
1. Add a repository with a job.
2. Delete the repository.
3. Add the repository one more time.

### Expected Behavior
The job should be re-enabled after the repository is re-added.

### Observed Behavior
When the repo is re-added and then sync'd there is a database constraint violation as the sync tries to add the job back (the job already exists in the database, but is marked as not installed).

glennmatthews:
# Closes #2974 
# What's Changed

When deleting a GitRepository, `on_delete=SET_NULL` causes all of its existing Job records to be updated with a null `git_repository` ForeignKey. When creating a new GitRepository record, we try to `get_or_create` Job records with the new GitRepository pk, and this fails since the pk doesn't match null but the other uniqueness constraints of the Job record still apply, resulting in an IntegrityError.

Solution: To handle this specific case, catch the IntegrityError and fall back to looking for an existing Job record that matches all other appropriate attributes but has a null `git_repository`, and update it to point to the new GitRepository.

This is an LTM-specific issue, as Jobs in 2.x do not have a `git_repository` FK but rather just a `@property` that dynamically looks up the appropriate GitRepository record based on the Job's `module_name`.

# Screenshots

Before:

image

After:

image

# TODO

- n/a Unit, Integration Tests

### DIFF
--- a/changes/2974.fixed
+++ b/changes/2974.fixed
@@ -1,0 +1,1 @@
+Fixed an error when deleting and then recreating a GitRepository that provides Jobs.
--- a/nautobot/extras/tests/test_datasources.py
+++ b/nautobot/extras/tests/test_datasources.py
@@ -27,6 +27,7 @@ from nautobot.extras.models import (
     ConfigContextSchema,
     ExportTemplate,
     GitRepository,
+    Job,
     JobLogEntry,
     JobResult,
     Secret,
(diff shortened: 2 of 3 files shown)


timizuoebideri1 (commented):
Can a test be added for this fix?

glennmatthews:
Possibly. It's a bit tricky since all of our GitRepository testing relies heavily on mocks so as to not be dependent on "real" external repositories, but I can look into it.

HanlinMiao (approved):

timizuoebideri1 (approved):

PR #5698, opened 2024-05-09
[LTM] Add 'javascript' and 'css' to BRANDING_FILEPATHS options. Sanitize BANNER_ content

### LINKED ISSUE #1858: Allow to customize CSS to differentiate multiple instances

glennmatthews:
# Closes #1858 
# What's Changed

LTM backport of #5697.

* The BRANDING_FILEPATHS setting can now include javascript and css as keys. The filepaths under MEDIA_ROOT identified by these settings are included as a custom script and a custom stylesheet respectively, if set.
* The BANNER_TOP, BANNER_LOGIN, and BANNER_BOTTOM configs now support Markdown (new feature) and a limited subset of sanitized HTML (new restriction) as a result of replacing the safe templatetag with use of render_markdown. This addresses security concerns about users with Nautobot admin privileges potentially configuring a malicious payload into any of these settings via Constance. (GHSA-r2hr-4v48-fjv3).

# Screenshots

Admin UI shown with custom CSS (overriding nav-bar color) and a Markdown BANNER_TOP:

image

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/core/settings.py
+++ b/nautobot/core/settings.py
@@ -59,6 +59,14 @@ ALLOWED_URL_SCHEMES = (
     "xmpp",
 )
 
+# Banners to display to users. Markdown and limited HTML are allowed.
+if "NAUTOBOT_BANNER_BOTTOM" in os.environ and os.environ["NAUTOBOT_BANNER_BOTTOM"] != "":
+    BANNER_BOTTOM = os.environ["NAUTOBOT_BANNER_BOTTOM"]
+if "NAUTOBOT_BANNER_LOGIN" in os.environ and os.environ["NAUTOBOT_BANNER_LOGIN"] != "":
+    BANNER_LOGIN = os.environ["NAUTOBOT_BANNER_LOGIN"]
+if "NAUTOBOT_BANNER_TOP" in os.environ and os.environ["NAUTOBOT_BANNER_TOP"] != "":
+    BANNER_TOP = os.environ["NAUTOBOT_BANNER_TOP"]
+
 # Base directory wherein all created files (jobs, git repositories, file uploads, static files) will be stored)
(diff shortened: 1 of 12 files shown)


nautobot/core/settings.py
```diff
@@ -586,15 +594,15 @@
     ],
     "BANNER_BOTTOM": [
         "",
-        "Custom HTML to display in a banner at the bottom of all pages.",
+        "Custom Markdown or limited HTML to display in a banner at the bottom of all pages.",
```
timizuoebideri1, line 597: Would it make more sense to write "`pure HTML without css or javascript`" in place of `limited HTML` to be more clear??

glennmatthews: The docs describe exactly which subset of HTML is permitted, but that'd be too verbose to spell out here I think.

timizuoebideri1 (approved):

HanlinMiao (approved):

Issue #5473, opened 2024-03-22
Refactor Dynamic Group mechanics


lampwins:
As a Nautobot Engineer, I want the dynamic group functionality to be refactored such that membership resolution is cached using static groups, so that critic performance issues do not prevent the usage of dynamic groups.

The filering logic and UX for dynamic groups should remain as-is, but the primary goal of this effort is to cache the membership of groups such that the access use case is optimized.

We should allow saved groups to be "hidden" such that these dynamic group static groups are not visible and do not get in the way of "real" static groups.

lampwins:
So I was thinking about this after the meeting and I fear we are starting off to complex. From past experience with cacheops, I think the biggest problem to the user was the lack of understanding/visibility that the cache was outdated and not having an easy way to remediate.

On that hypothesis, I wonder if we would be better off (to start with) by making this a simple time based cache but focusing on the state visibility to the user. Make the time configurable, but default to say 15 minutes. So every 15 minutes, we have a scheduled system job that refreshes the groups. On the dynamic group detail view, clearly show the user the last time the group was updated and give them a button to do an ad-hoc update. Also provide a utility function for SSoT or app authors to manually trigger an update if they so choose.

I think that gives us a solid base to stand on and then expand from there as more well defined use cases arise. If we start too complex, I think we doom ourselves to having to constantly fix cache bugs in a paradigm that folks don't really need or want, or at least one which is prematurely optimized.

bryanculver:
Could this:

- https://github.com/nautobot/nautobot/blob/2e6a144df6fe88a559982eb2506783817235df44/nautobot/extras/signals.py#L416-L442
- https://github.com/nautobot/nautobot/blob/2e6a144df6fe88a559982eb2506783817235df44/nautobot/extras/models/groups.py#L331-L359

be updated to save onto DB instead of Redis? It already leverages a timeout.

And this updated to just be a reverse generic foreign key M2M to said DB-backed cache: https://github.com/nautobot/nautobot/blob/2e6a144df6fe88a559982eb2506783817235df44/nautobot/extras/models/mixins.py#L23-L38

PR #5701, opened 2024-05-10
Add StaticGroup.hidden flag

glennmatthews:
# Progress towards #5473
# What's Changed

- Add `StaticGroup.hidden` flag - not user serviceable via the UI
  - [x] TODO disallow changing this flag via the REST API either
- Add `hidden` filter to StaticGroup and StaticGroupAssociation FilterSets
  - UI views default to hiding hidden groups by default
  - API views default to hiding hidden groups by default
  - [x] TODO hide hidden groups in GraphQL by default?
- Allow `to_objectchange()` to return None in the case of objects opting out of being change-logged
  - StaticGroup and StaticGroupAssociation are not change-logged if hidden
- Update existing tests
  - [x] TODO: add tests explicitly for UI/REST/GraphQL hiding behavior?

# Screenshots

# TODO

- [x] Unit, Integration Tests


nautobot/extras/templates/extras/staticgroup_retrieve.html
```diff
@@ -50,6 +50,12 @@
                 <td>Tenant</td>
                 <td>{{ object.tenant|hyperlinked_object }}</td>
             </tr>
+            {% if object.hidden %}
+                <tr>
+                    <td>Hidden</td>
+                    <td>{{ object.hidden|render_boolean }}</td>
+                </tr>
+            {% endif %}
```
HanlinMiao, line 58: Why do we only display it when `object.hidden` is true?

HanlinMiao (approved):

nautobot/extras/forms/forms.py
```diff
@@ -1597,8 +1598,11 @@ class Meta:
     def clean(self):
         data = super().clean()
 
-        if data["add_to_groups"].filter(pk__in=data["remove_from_groups"].values_list("pk", flat=True)).exists():
-            raise ValidationError("Same group specified for both addition and removal")
+        if "add_to_groups" in data and "remove_from_groups" in data:
+            if data["add_to_groups"].filter(pk__in=data["remove_from_groups"].values_list("pk", flat=True)).exists():
+                raise ValidationError("Same group specified for both addition and removal")
+
+        return data
```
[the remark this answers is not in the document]
glennmatthews: Yeah, looked that way.

nautobot/extras/templates/extras/staticgroup_retrieve.html
glennmatthews: My thinking was that most/all of the groups that the user actually cares about are non-hidden ones, and you have to go out of your way to even see hidden ones, so the presence of this field in the detail view in the default case wouldn't really add any meaning/value to the average user.

PR #5707, opened 2024-05-13
Fix xmlsec/lxml mismatch in Docker image

glennmatthews:
# Closes #N/A
# What's Changed

In order for both of the SSO dependencies `lxml` and `xmlsec` to be usable together, they *both* need to be installed from source rather than from wheels. We already were installing `lxml` from source but missed doing the same for `xmlsec`.

# Screenshots

Before:

```
❯ invoke cli        
root@1d381b1a39e4:/source# python
Python 3.8.19 (default, Apr 24 2024, 13:06:49) 
[GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import xmlsec
xmlsec.InternalError: (-1, 'lxml & xmlsec libxml2 library version mismatch')
>>> 
```

After:

```
❯ invoke cli        
root@915b74fb7e21:/source# python
Python 3.8.19 (default, Apr 24 2024, 13:06:49) 
[GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import xmlsec
>>> 
```

# TODO

- n/a Unit, Integration Tests

### DIFF
--- a/changes/5707.fixed
+++ b/changes/5707.fixed
@@ -1,0 +1,1 @@
+Fixed invalid installation of `xmlsec` library in the Docker images.
--- a/changes/5707.fixed
+++ b/changes/5707.fixed
@@ -1,1 +1,1 @@
-Fixed invalid installation of `xmlsec` library in the Docker images.
+Fixed invalid installation of `xmlsec` library in the Nautobot Docker images.
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -125,7 +125,7 @@ ENV PATH="${POETRY_HOME}/bin:${PATH}"
 
 RUN poetry config virtualenvs.create ${POETRY_VIRTUALENVS_CREATE} && \
     poetry config installer.parallel "${POETRY_INSTALLER_PARALLEL}" && \
-    poetry config installer.no-binary lxml,pyuwsgi
+    poetry config installer.no-binary lxml,pyuwsgi,xmlsec
 
 ################################ Stage: python-dependencies (intermediate build target)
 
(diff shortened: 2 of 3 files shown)


gsnider2195:
I noticed this while testing this behavior:

```
root@4713b1655aed:/# pip3 install --no-binary=lxml xmlsec
DEPRECATION: --no-binary currently disables reading from the cache of locally built wheels. In the future --no-binary will not influence the wheel cache. pip 23.1 will enforce this behaviour change. A possible replacement is to use the --no-cache-dir option. You can use the flag --use-feature=no-binary-enable-wheel-cache to test the upcoming behaviour. Discussion can be found at https://github.com/pypa/pip/issues/11453
```

I don't think this is going to affect us though?

gsnider2195 (approved):

gsnider2195:
Related https://github.com/xmlsec/python-xmlsec/issues/316

glennmatthews:
> I noticed this while testing this behavior:
> 
> > root@4713b1655aed:/# pip3 install --no-binary=lxml xmlsec
> > DEPRECATION: --no-binary currently disables reading from the cache of locally built wheels. In the future --no-binary will not influence the wheel cache. pip 23.1 will enforce this behaviour change. A possible replacement is to use the --no-cache-dir option. You can use the flag --use-feature=no-binary-enable-wheel-cache to test the upcoming behaviour. Discussion can be found at [pypa/pip#11453](https://github.com/pypa/pip/issues/11453)
> 
> I don't think this is going to affect us though?

Shouldn't affect us, but I'll test locally with that `--use-feature` flag just to check.

PR #5711, opened 2024-05-14
Added "Add VRFs" and "Remove VRFs" fields to `PrefixBulkEditForm`.

### LINKED ISSUE #5588: Missing Vrf fields in bulk edit menu
### Environment
* Nautobot version (Docker tag too if applicable): 2.2.1
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce

1.    Select multiple prefixes
2.    Click on Edit selected
3.    Try to adding or removing Vrfs

### Expected Behavior

Fields for adding and removing vrfs should be presented, similar to the "Add tags" and "Remove tags" fields.

### Observed Behavior
The "Add VRFs" and "Remove VRFs" fields are missing from bulk edit.

timizuoebideri1:
# Closes #5588 
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/ipam/views.py
+++ b/nautobot/ipam/views.py
@@ -303,6 +303,12 @@ class VRFBulkEditView(generic.BulkEditView):
     table = tables.VRFTable
     form = forms.VRFBulkEditForm
 
+    def extra_post_save_action(self, obj, form):
+        if form.cleaned_data.get("add_prefixes", None):
+            obj.vrfs.add(*form.cleaned_data["add_prefixes"])
+        if form.cleaned_data.get("remove_prefixes", None):
+            obj.vrfs.remove(*form.cleaned_data["remove_prefixes"])
+
 
 class VRFBulkDeleteView(generic.BulkDeleteView):
     queryset = VRF.objects.select_related("tenant")
--- a/nautobot/ipam/views.py
+++ b/nautobot/ipam/views.py
@@ -713,6 +713,10 @@ class PrefixBulkEditView(generic.BulkEditView):
             obj.locations.add(*form.cleaned_data["add_locations"])
         if form.cleaned_data.get("remove_locations", None):
             obj.locations.remove(*form.cleaned_data["remove_locations"])
+        if form.cleaned_data.get("add_vrfs", None):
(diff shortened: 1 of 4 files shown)


nautobot/ipam/views.py
```diff
@@ -713,6 +713,10 @@ def extra_post_save_action(self, obj, form):
             obj.locations.add(*form.cleaned_data["add_locations"])
         if form.cleaned_data.get("remove_locations", None):
             obj.locations.remove(*form.cleaned_data["remove_locations"])
+        if form.cleaned_data.get("add_vrfs", None):
+            obj.vrfs.add(*form.cleaned_data["add_vrfs"])
+        if form.cleaned_data.get("remove_vrfs", None):
+            obj.vrfs.remove(*form.cleaned_data["remove_vrfs"])
```
glennmatthews, line 719: Do we need to do anything special here to prevent assigning VRFs from the wrong namespace to prefixes in a different namespace? I don't think this implementation will call `VRFPrefixAssignment.clean()` for example.

Also, can we add some vrfs to the Prefix view test case bulk_edit_data to exercise this functionality?

Finally, shouldn't we similarly add `add_prefixes` and `remove_prefixes` fields to the `VRFBulkEditForm` as well?

timizuoebideri1: We don't need to handle, as its been handled by the `vrf_prefix_associated` signal: 
```python
@receiver(m2m_changed, sender=VRFPrefixAssignment)
def vrf_prefix_associated(sender, instance, action, reverse, model, pk_set, **kwargs):
    """
    Disallow adding Prefixes if the namespace doesn't match.
    """

    if action == "pre_add":
        prefixes = model.objects.filter(pk__in=pk_set).exclude(namespace=instance.namespace)
        if prefixes.exists():
            raise ValidationError({"prefixes": "Prefix must match namespace of VRF"})
```

glennmatthews: Ok, thanks for confirming. I see you updated `VRFBulkEditForm` as well now. Can we get some test coverage for these form fields?

glennmatthews (approved):

PR #5724, opened 2024-05-16
Device modules UI

### LINKED ISSUE #2101: Model Chassis Switches (Module support)

gsnider2195:
# Closes #2101
# What's Changed

# Screenshots

### Module

image

### ModuleType

image

### ModuleBay

image

### Modular device component list view

image

### Device bulk add component

image

### Module bulk add component

image

### Module

image

### ModuleType

image

### ModuleBay delete

image

### ModuleBay bulk delete

image

### Module delete

image

### Module bulk delete

image

### Module

image

### Module ModuleBay view tab

image

### ModuleBay

image

### ModuleType

image

### ModuleType detail tab

image

### ModuleType detail "add components" menu

image

### Device Modules tab

image

### DeviceType ModuleBays tab

image

### DeviceType detail "add components" menu

image

# TODO

- [ ] Unit, Integration Tests
- [x] Templated ModuleBay and ModuleBayTemplate `position` similar to other components `name`
- [ ] Bulk move components from Device to Module on Module creation #5730
- [x] Bulk add components to Modules from list view
- [ ] Support for devicetype library import


nautobot/dcim/templates/dcim/module_retrieve.html
```diff
@@ -0,0 +1,68 @@
...
+            </td>
+        </tr>
+        <tr>
+            <td>Installed in</td>
+            <td>
+                {% with parent_module_bay=object.parent_module_bay %}
+                    {% if parent_module_bay.parent_device %}
+                        <a href="{{ parent_module_bay.parent_device.get_absolute_url }}">{{ parent_module_bay }}</a>
+                    {% elif parent_module_bay.parent_module %}
+                        <a href="{{ parent_module_bay.parent_module.get_absolute_url }}">{{ parent_module_bay }}</a>
```
glennmatthews, line 23: Should these be?

```suggestion
                        {{ parent_module_bay.parent_device|hyperlinked_object }}
                    {% elif parent_module_bay.parent_module %}
                        {{ parent_module_bay.parent_module|hyperlinked_object }}
```

nautobot/dcim/views.py
```diff
@@ -1189,6 +1279,30 @@ class DeviceBayTemplateBulkDeleteView(generic.BulkDeleteView):
...
+    bulk_update_form_class = forms.ModuleBayBulkEditForm
+    filterset_class = filters.ModuleBayTemplateFilterSet
+    # filterset_form_class = forms.ModuleBayTemplateFilterForm
+    form_class = forms.ModuleBayTemplateForm
+    serializer_class = serializers.ModuleBayTemplateSerializer
+    # table_class = tables.ModuleBayTemplateTable
+
+    @action(detail=False, url_path="rename", url_name="bulk_rename")
+    def add(self, request, *args, **kwargs):
+        return None
```
glennmatthews, line 1303: TODO I presume?

nautobot/core/filters.py
```diff
@@ -149,11 +147,23 @@ def __init__(
...
+            return qs
+        if self.distinct:
+            qs = qs.distinct()
+        lookup = "%s__%s" % (self.field_name, self.lookup_expr)
+        if bool(value):
+            # return instances with field populated
+            return qs.filter(**{lookup: self.exclude})
+        else:
+            # exclude instances with field populated
+            return qs.exclude(**{lookup: self.exclude})
```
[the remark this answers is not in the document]
gsnider2195: Comments are definitely needed! Thanks for the reminder

nautobot/core/testing/views.py
```diff
@@ -1580,3 +1580,22 @@ def test_bulk_rename(self):
...
                     expected_message = f"No valid {verbose_name_plural} were selected."
                     self.assertIn(expected_message, response.content.decode(response.charset))
+
+    class ModularDeviceComponentViewTestCase(DeviceComponentViewTestCase):
+        def test_edit_object_with_permission(self):
+            instance = self._get_queryset().first()
+            form_data = self.form_data.copy()
+            # device and module are not editable
+            form_data["device"] = getattr(getattr(instance, "device", {}), "pk", None)
+            form_data["module"] = getattr(getattr(instance, "module", {}), "pk", None)
```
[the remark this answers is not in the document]
gsnider2195: Good point. the empty dict would be appropriate if I were doing a `.get()` instead of a `.getattr()`. Will update

nautobot/core/views/renderers.py
```diff
@@ -221,25 +228,30 @@ def get_context(self, data, accepted_media_type, renderer_context):
...
+                "obj_type_plural": queryset.model._meta.verbose_name_plural,  # NOTE: This context key is deprecated in favor of `verbose_name_plural`.
+                "permissions": permissions,
+                "return_url": return_url,
+                "table": table if table is not None else data.get("table", None),
+                "table_config_form": TableConfigForm(table=table) if table else None,
+                "verbose_name": queryset.model._meta.verbose_name,
+                "verbose_name_plural": queryset.model._meta.verbose_name_plural,
+            }
+        )
+        if view.detail:
```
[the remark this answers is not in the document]
gsnider2195: This probably needs further refinement but I made the change to support our documented extra actions [here](https://docs.nautobot.com/projects/core/en/stable/development/apps/api/views/nautobotuiviewset/#adding-custom-views-to-nautobotuiviewset-nautobotuiviewsetrouter).

It seems wrong to only call the viewset's `get_extra_context` for the retrieve view. That should be common to all views in the viewset.

The `common_detail_view_context` does look specific to the retrieve view so perhaps we keep that behavior and rename the function to `common_retrieve_view_context` to avoid confusing DRF's [`detail=True (boolean indicating if the current action is configured for a list or detail view.)`](https://www.django-rest-framework.org/api-guide/viewsets/#introspecting-viewset-actions) for views. I'd actually prefer to move a lot of this logic into the UIViewSet mixins, but not in this goliath of a PR.

nautobot/dcim/filters/mixins.py
```diff
@@ -194,10 +194,31 @@ class DeviceModuleCommonFiltersMixin(django_filters.FilterSet):
         field_name="module_bays",
         label="Has module bays",
     )
+    has_empty_module_bays = django_filters.BooleanFilter(
+        method="filter_has_empty_module_bays",
+        label="Has empty module bays",
+    )
     module_bays = django_filters.ModelMultipleChoiceFilter(
```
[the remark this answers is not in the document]
gsnider2195: Might be appropriate here? I was thinking the module bay "position" wouldn't be suitable for a natural key since in the common case they're going to be integers.

nautobot/dcim/filters/mixins.py
```diff
@@ -194,10 +194,31 @@ class DeviceModuleCommonFiltersMixin(django_filters.FilterSet):
...
+    has_modules = RelatedMembershipBooleanFilter(
+        field_name="module_bays__installed_module",
+        label="Has modules",
+    )
+
+    def generate_query_filter_has_empty_module_bays(self, value):
+        if value is True:
+            query = Q(module_bays__isnull=False, module_bays__installed_module__isnull=True)
+        else:
+            query = Q(module_bays__isnull=True) | Q(module_bays__installed_module__isnull=False)
```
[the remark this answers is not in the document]
gsnider2195: It should match on any single module bay being empty since it's a filter not an exclude but I will add tests for this to be sure.

nautobot/dcim/forms.py
```diff
@@ -886,6 +898,90 @@ class DeviceTypeFilterForm(NautobotFilterForm):
...
+    manufacturer = DynamicModelChoiceField(queryset=Manufacturer.objects.all(), required=False)
+
+
+class ModuleTypeFilterForm(NautobotFilterForm):
+    model = ModuleType
+    q = forms.CharField(required=False, label="Search")
+    manufacturer = DynamicModelMultipleChoiceField(
+        queryset=Manufacturer.objects.all(), to_field_name="name", required=False
+    )
+    has_console_ports = forms.NullBooleanField(
```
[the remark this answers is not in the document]
gsnider2195: I don't feel strongly either way for the filter form. The filters are correctly named so the filter form should probably match.

nautobot/dcim/forms.py
```diff
@@ -896,38 +992,49 @@ class ComponentTemplateCreateForm(ComponentForm):
...
+
+
+class ModularComponentTemplateCreateForm(ComponentTemplateCreateForm):
+    """
+    Base form for the creation of modular device component templates (subclassed from ModularComponentTemplateModel).
+    """
+
     device_type = DynamicModelChoiceField(
         queryset=DeviceType.objects.all(),
-        query_params={"manufacturer": "$manufacturer"},
```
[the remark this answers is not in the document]
gsnider2195: This field is automatically populated and disabled now so we don't need to give the user the option to filter it on manufacturer anymore. This was the one we discussed in DSU where we didn't see a need to be able to create components without first going to the parent device type or module type and clicking on "add components" (which would pre-populate the field)

nautobot/dcim/navigation.py
```diff
@@ -244,6 +244,44 @@
...
+                            NavMenuAddButton(
+                                link="dcim:module_add",
+                                permissions=[
+                                    "dcim.add_module",
+                                ],
+                            ),
+                        ),
+                    ),
+                ),
+            ),
```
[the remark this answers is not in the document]
gsnider2195: I had the same thought about swapping these two. I'll make the change and see if anyone complains 😄 

I think there will be a stronger use case for creating modules through the nav menu add button than for other components so I'd prefer to keep them higher up the list but happy to discuss.

nautobot/dcim/tables/devices.py
```diff
@@ -287,7 +342,7 @@ class Meta(DeviceComponentTable.Meta):
         default_columns = ("pk", "device", "name", "label", "type", "description")
 
 
-class DeviceConsolePortTable(ConsolePortTable):
+class DeviceModuleConsolePortTable(ConsolePortTable):
```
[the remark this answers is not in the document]
gsnider2195: They're not in the apps namespace. Lets discuss if we think making it backwards compatible is necessary

nautobot/dcim/views.py
gsnider2195: Done now but there are still a few custom actions I need to implement for these models.

nautobot/dcim/tables/devices.py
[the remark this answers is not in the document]
glennmatthews: I think this is low-risk, but I've been wrong before. :-)

nautobot/dcim/templates/dcim/device/devicebays.html
```diff
@@ -18,21 +18,21 @@
             {% render_table devicebay_table 'inc/table.html' %}
             <div class="panel-footer noprint">
                 {% if perms.dcim.change_devicebay %}
-                    <button type="submit" name="_rename" formaction="{% url 'dcim:devicebay_bulk_rename' %}?return_url={{ object.get_absolute_url }}%23tab_devicebays" class="btn btn-warning btn-xs">
+                    <button type="submit" name="_rename" formaction="{% url 'dcim:devicebay_bulk_rename' %}?return_url={% url 'dcim:device_devicebays' pk=object.pk %}" class="btn btn-warning btn-xs">
```
gsnider2195, line 21: Unrelated bug fix

nautobot/dcim/forms.py
```diff
@@ -886,48 +898,158 @@ class DeviceTypeFilterForm(NautobotFilterForm):
...
     class Meta:
         model = ConsolePortTemplate
         fields = [
             "device_type",
+            "module_type",
             "name",
             "label",
             "type",
             "description",
         ]
```
glennmatthews, line 1045: Not blocking, but feels like we could cut down on boilerplate a fair bit by declaring `fields` on `ModularComponentTemplateForm.Meta` and `ModularComponentTemplateCreateForm.Meta` and having the concrete subclasses `Meta` inherit from those, or at least do something like `fields = [*ModularComponentTemplateForm.Meta.fields, ...]`?

nautobot/dcim/forms.py
```diff
@@ -1982,6 +2173,191 @@ class DeviceFilterForm(
...
+        queryset=ModuleBay.objects.all(),
+        label="Parent Module Bay",
+        required=False,
+        query_params={"parent_module": "$parent_module_bay_module_filter", "has_installed_module": False},
+        initial_params={"pk": "$parent_module_bay", "parent_module__module_bays": "$parent_module_bay"},
+    )
+    location = DynamicModelChoiceField(
+        queryset=Location.objects.all(),
+        required=False,
+        label="Location",
```
glennmatthews, line 2222: Maybe some sort of help_text here as well like "Only specify Location if this module is not currently installed into a module bay"?

nautobot/dcim/forms.py
```diff
@@ -1982,6 +2173,191 @@ class DeviceFilterForm(
...
+        }
+
+    def clean(self):
+        cleaned_data = self.cleaned_data
+        if cleaned_data["parent_module_bay_device"] and cleaned_data["parent_module_bay_module"]:
+            raise forms.ValidationError("Multiple parent module bays selected.")
+        elif cleaned_data["parent_module_bay_device"]:
+            cleaned_data["parent_module_bay"] = cleaned_data.pop("parent_module_bay_device")
+        elif cleaned_data["parent_module_bay_module"]:
+            cleaned_data["parent_module_bay"] = cleaned_data.pop("parent_module_bay_module")
```
glennmatthews, line 2257: I wonder if it'd make sense, alternately, for `clean()` to throw away any `location` value if a `parent_module_bay` is also specified, just to avoid a likely user error/misunderstanding? Or do you think it's better for this to be a ValidationError that the user needs to address?

nautobot/dcim/forms.py
```diff
@@ -2021,22 +2423,27 @@ class Meta:
...
-class ConsolePortForm(ComponentEditForm):
+class ConsolePortForm(ModularComponentEditForm):
     class Meta:
         model = ConsolePort
         fields = [
             "device",
+            "module",
             "name",
             "label",
             "type",
```
glennmatthews, line 2449: Same thought here as above about inheriting `Meta.fields` from the base class.

nautobot/dcim/forms.py
```diff
@@ -2911,13 +3418,105 @@ class Meta:
...
+        queryset=Device.objects.all(),
+        required=False,
+        label="Parent Device",
+    )
+    parent_module = DynamicModelChoiceField(
+        queryset=Module.objects.all(),
+        required=False,
+        label="Parent Module",
+    )
+    # TODO: Installed module field
```
glennmatthews, line 3454: Would be nice, yeah.

nautobot/dcim/migrations/0060_module_models.py
```diff
@@ -455,6 +456,10 @@ class Migration(migrations.Migration):
             name="rearport",
             options={"ordering": ("device", "module", "_name")},
         ),
+        migrations.AlterModelOptions(
+            name="interface",
+            options={"ordering": ("device", "module", nautobot.core.models.query_functions.CollateAsChar("_name"))},
+        ),
```
glennmatthews, line 462: Out of curiosity, do you know/understand why is `interface` the only model using `nautobot.core.models.query_functions.CollateAsChar()` compared to all of the other components?

nautobot/dcim/models/devices.py
```diff
@@ -1619,11 +1619,16 @@ class Meta:
...
+    @property
+    def display(self):
         if self.location:
-            return f"{display} at location {self.location}"
+            return f"{self!s} at location {self.location}"
+        elif self.parent_module_bay.parent_device is not None:
+            return f"{self.module_type!s} installed in {self.parent_module_bay.parent_device.display}"
         else:
-            return f"{display} installed in {self.parent_module_bay}"
+            return f"{self.module_type!s} installed in {self.parent_module_bay.parent_module.display}"
```
glennmatthews, line 1631: I wonder if these are overly verbose, or whether the clarity is worth the verbosity? some ideas for shorter alterniatives:

- `{self.location} : {self!s}`
- `{self.parent_module_bay.parent_device.display} : {self.module_type!s}`
- `{self.parent_module_bay.parent_module.display} : {self.module_type!s}`

nautobot/dcim/tables/devices.py
```diff
@@ -248,6 +304,13 @@ class Meta(BaseTable.Meta):
         order_by = ("device", "name")
 
 
+class ModularDeviceComponentTable(DeviceComponentTable):
+    module = tables.Column(linkify=True)
+
+    class Meta(DeviceComponentTable.Meta):
+        order_by = ("device", "module", "name")
```
glennmatthews, line 311: Curious why this is needed - doesn't the `order_by` on the model handle this already?

nautobot/dcim/tables/devices.py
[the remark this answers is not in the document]
glennmatthews: Double-check against https://github.com/nautobot/nautobot-app-version-control/blob/develop/nautobot_version_control/diffs.py - at a quick look, I don't *think* it's using these DeviceFooTable classes (and probably could replace all of that code with calls to `get_table_for_model` nowadays) but would be good to confirm.

nautobot/dcim/tables/devices.py
```diff
@@ -267,14 +330,15 @@ class PathEndpointTable(CableTerminationTable):
...
-    class Meta(DeviceComponentTable.Meta):
+    class Meta(ModularDeviceComponentTable.Meta):
         model = ConsolePort
         fields = (
             "pk",
             "device",
+            "module",
             "name",
             "label",
             "type",
```
glennmatthews, line 344: Similarly here, could we define the common fields in `ModularDeviceComponentTable.Meta` and just extend/modify them inline as needed?

nautobot/dcim/tables/devicetypes.py
```diff
@@ -157,7 +195,7 @@ class ConsolePortTemplateTable(ComponentTemplateTable):
     actions = ButtonsColumn(
         model=ConsolePortTemplate,
         buttons=("edit", "delete"),
-        return_url_extra="%23tab_consoleports",
+        return_url_extra=r"%3Ftab=consoleports",
```
glennmatthews, line 198: I'm mildly curious why we have to percent-encode the `?` as `%3F` here rather than providing it as a literal `?`.

nautobot/dcim/templates/dcim/module/base.html
```diff
@@ -0,0 +1,97 @@
+{% extends 'generic/object_retrieve.html' %}
+{% load buttons %}
+{% load helpers %}
+
+{% block extra_buttons %}
+        {% if perms.dcim.change_module %}
```
glennmatthews, line 6: I'm guessing this is following a similar pattern to devices, but it's interesting that we gate on this permission here given that adding components doesn't *actually* change the module itself. I also wonder whether the views actually enforce this additional permission or whether it's only a display feature?

More fuel for the v3.0 permissions rework, I imagine.

nautobot/dcim/templates/dcim/module_list.html
```diff
@@ -0,0 +1,20 @@
+{% extends 'generic/object_list.html' %}
+
+{% block bulk_buttons %}
+    {% if perms.dcim.change_device %}
```
glennmatthews, line 4: Same comment here on permissions -- though in this case, if we do keep it, shouldn't it be `change_module` rather than `change_device`?

nautobot/dcim/templates/dcim/module_poweroutlets.html
```diff
@@ -0,0 +1,54 @@
...
+                <strong>Power Outlets</strong>
+                <div class="pull-right noprint">
+                    {% if request.user.is_authenticated %}
+                        <button type="button" class="btn btn-default btn-xs" data-toggle="modal" data-target="#ModulePowerOutletTable_config" title="Configure table"><i class="mdi mdi-cog"></i> Configure</button>
+                    {% endif %}
+                </div>
+            </div>
+            {% render_table poweroutlet_table 'inc/table.html' %}
+            <div class="panel-footer noprint">
+                {% if perms.dcim.change_powerport %}
```
glennmatthews, line 20: `change_poweroutlet`?

nautobot/dcim/templates/dcim/module_retrieve.html
```diff
@@ -0,0 +1,63 @@
...
+                    {{ None|placeholder }}
+                {% endif %}
+            </td>
+        </tr>
+        <tr>
+            <td>Location</td>
+            <td>
+                {% include 'dcim/inc/location_hierarchy.html' with location=object.location %}
+            </td>
+        </tr>
```
glennmatthews, line 35: Should display of this row be mutually exclusive with the `Installed in` row?

nautobot/dcim/templates/dcim/moduletype_retrieve.html
```diff
@@ -0,0 +1,142 @@
...
+                        <li><a href="{% url 'dcim:modulebaytemplate_add' %}?module_type={{ object.pk }}&return_url={{ object.get_absolute_url }}%3Ftab=modulebays">Module Bays</a></li>
+                    {% endif %}
+                </ul>
+            </div>
+        {% endif %}
+{% endblock extra_buttons %}
+
+{% block extra_nav_tabs %}
+        <li role="presentation" {% if request.GET.tab == 'interfaces' %} class="active"{% endif %}>
+            <a href="{{ object.get_absolute_url }}#interfaces" onclick="switch_tab(this.href)" aria-controls="interfaces" role="tab" data-toggle="tab">
```
glennmatthews, line 47: Should all of these be `?tab=...` instead of `#...`?

nautobot/dcim/views.py
```diff
@@ -1189,6 +1290,202 @@ class DeviceBayTemplateBulkDeleteView(generic.BulkDeleteView):
     filterset = filters.DeviceBayTemplateFilterSet
 
 
+#
+# Module bay templates
+#
+
+
+class ModuleBayCommonViewSetMixin:
+    def create(self, request, *args, **kwargs):
```
glennmatthews, line 1299: Mind adding a docstring or comment explaining how this is different than the usual viewset `create` implementation?

nautobot/dcim/views.py
```diff
@@ -1189,6 +1290,202 @@ class DeviceBayTemplateBulkDeleteView(generic.BulkDeleteView):
...
+            for i, position in enumerate(positions):
+                label = labels[i] if labels else None
+                # Initialize the individual component form
+                data["position"] = position
+                data["label"] = label
+                component_form = self.model_form_class(
+                    data,
+                    initial=normalize_querydict(request.GET, form_class=self.model_form_class),
+                )
+                self.logger.warn(component_form.initial)
```
glennmatthews, line 1341: remove?

nautobot/dcim/views.py
```diff
@@ -1189,6 +1290,202 @@ class DeviceBayTemplateBulkDeleteView(generic.BulkDeleteView):
...
+    def get_selected_objects_parents_name(self, selected_objects):
+        selected_object = selected_objects.first()
+        if selected_object:
+            parent = selected_object.device_type or selected_object.module_type
+            return parent.display
+        return ""
+
+    @action(detail=False, methods=["POST"], url_path="rename", url_name="bulk_rename")
+    def bulk_rename(self, request, *args, **kwargs):
+        return self._bulk_rename(request, *args, **kwargs)
```
glennmatthews, line 1486: Curious why this indirection is necessary - why not just define `bulk_rename` directly on `ModuleBayCommonViewSetMixin`?

nautobot/dcim/views.py
```diff
@@ -1609,6 +1953,364 @@ class DeviceBulkDeleteView(generic.BulkDeleteView):
...
+        if form["parent_module_bay_module"].initial:
+            active_parent_tab = "module"
+        elif form["location"].initial:
+            active_parent_tab = "location"
+
+        return active_parent_tab
+
+    @action(detail=True)
+    def consoleports(self, request, *args, **kwargs):
+        # TODO: permissions enforcement?
```
glennmatthews, line 2105: Were you able to sort this out?

nautobot/dcim/views.py
```diff
@@ -1609,6 +1953,364 @@ class DeviceBulkDeleteView(generic.BulkDeleteView):
...
+        )
+        if form["parent_module_bay_module"].initial:
+            active_parent_tab = "module"
+        elif form["location"].initial:
+            active_parent_tab = "location"
+
+        return active_parent_tab
+
+    @action(detail=True)
+    def consoleports(self, request, *args, **kwargs):
```
glennmatthews, line 2104: Does this need a `url_path="console-ports"` for consistency (etc. below)?

nautobot/extras/templates/extras/role_retrieve.html
```diff
@@ -191,4 +191,20 @@
         </div>
     {% include 'inc/paginator.html' with paginator=controller_table.paginator page=controller_table.page %}
     {% endif %}
+    {% if module_table %}
```
glennmatthews, line 194: Should we similarly update the `location` detail view to include a count of colocated modules?

glennmatthews (commented):
Wow! Great stuff.

nautobot/dcim/views.py
gsnider2195: On day 2 of trying to get this sorted out 😞

nautobot/dcim/views.py
gsnider2195: I don't think the function decorator works properly for subclasses. At least that's what I was experiencing which is why I did it this way.

nautobot/dcim/forms.py
```diff
@@ -886,48 +898,158 @@ class DeviceTypeFilterForm(NautobotFilterForm):
...
+
+
 #
 # Device component templates
 #
 
 
+class ComponentTemplateForm(NautobotModelForm):
+    # TODO: placeholder values shouldn't be form controls, instead use:
+    #       <p class="form-control-static">{{ obj|hyperlinked_object_target_new_tab }}</p>
```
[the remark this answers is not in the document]
gsnider2195: I need to circle back to this because it proved to be complicated to try to do it the "right" way (in the form field) and I might just have to cheat and do it in the html template for now.

nautobot/dcim/models/devices.py
gsnider2195: I find it nice to have it in the display field for more context in the UI:

image

Instead of:

image

nautobot/dcim/migrations/0060_module_models.py
gsnider2195: No idea why but it would be nice to have natural sorting on all component models to avoid this:

image

nautobot/core/views/generic.py
```diff
@@ -1431,6 +1438,7 @@ class BulkComponentCreateView(GetReturnURLMixin, ObjectPermissionRequiredMixin,
 
     parent_model = None
     parent_field = None
+    primary_pattern_field = "name"
```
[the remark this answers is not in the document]
gsnider2195: I must not have committed those changes. I'm still in the middle of getting view tests and permissions working. This was to support ModuleBay using `position` instead of `name`

nautobot/core/testing/views.py
```diff
@@ -1580,3 +1601,22 @@ def test_bulk_rename(self):
...
+            form_data["module"] = getattr(getattr(instance, "module", None), "pk", None)
+            self.form_data = form_data
+            super().test_edit_object_with_permission()
+
+        def test_edit_object_with_constrained_permission(self):
+            instance = self._get_queryset().first()
+            form_data = self.form_data.copy()
+            # device and module are not editable
+            form_data["device"] = getattr(getattr(instance, "device", {}), "pk", None)
+            form_data["module"] = getattr(getattr(instance, "module", {}), "pk", None)
```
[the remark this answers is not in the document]
gsnider2195: Must be a bad habit because this is the second time I've done that!

nautobot/dcim/tables/devicetypes.py
gsnider2195: That would construct a url like this `?return_url=/dcim/module-types/86b9d1de-60f3-46fc-b356-eae2ac2cf25f/?tab=modulebays`

Would the second `?` confuse some older or simpler browsers? (seems to work in firefox fwiw)

glennmatthews: Got it. Makes sense as-is then.

nautobot/dcim/templates/dcim/moduletype_retrieve.html
gsnider2195: The javascript function `switch_tab` is splitting the url on `#` so this is correct

nautobot/dcim/tables/devices.py
gsnider2195: This might actually be a bug. It appears that when you let the table do the sorting it only accounts for the values of the fields, ie: order by `device.pk`, `module.pk`, `name.pk` instead of the ordering of the database fields.

gsnider2195: The number of database queries is the same too. I'm removing these

nautobot/core/management/commands/generate_test_data.py
```diff
@@ -197,7 +197,7 @@ def _generate_factory_data(self, seed, db_name):
         InterfaceTemplateFactory.create_batch(30, using=db_name)
         PowerPortTemplateFactory.create_batch(30, using=db_name)
         PowerOutletTemplateFactory.create_batch(30, using=db_name)
-        ModuleBayTemplateFactory.create_batch(30, using=db_name)
+        ModuleBayTemplateFactory.create_batch(90, using=db_name)
```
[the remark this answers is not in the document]
gsnider2195: I think I needed this for the `DeviceFilterSet.has_empty_module_bays` and `ModuleFilterSet.has_empty_module_bays` tests. I know it was one of the filter tests. All of the components are split between modules and devices and this one is a little different because modules are nestable.

nautobot/dcim/migrations/0060_module_models.py
gsnider2195: The above screenshot was resolved when I fixed the missing `NaturalOrderingField` on `ModuleBay`. I'm not sure what the `CollateAsChar` is doing

nautobot/dcim/views.py
gsnider2195: Today I learned that you should always call `self.get_object()` and `self.get_queryset()` in NautobotUIViewsets if you want permissions to be enforced

nautobot/project-static/css/base.css
```diff
@@ -97,6 +97,7 @@ table.attr-table td:nth-child(1) {
 }
 .table-headings th {
     background-color: #f5f5f5;
+    white-space: nowrap;
```
gsnider2195, line 100: This removes wrapping on all table headings throughout nautobot @lampwins

nautobot/dcim/views.py
glennmatthews: Oh yeah, that's a big footgun. :-(

nautobot/core/management/commands/generate_test_data.py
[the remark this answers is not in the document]
glennmatthews: Might be able to instead call `ModuleBayTemplateFactory.create_batch()` several times with distinct parameters as an alternative (`create_batch(10, has_parent_device=False)` or whatever). Not critical.

nautobot/dcim/tables/devices.py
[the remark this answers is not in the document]
gsnider2195: They're all commented out. I'm going to treat commented out code the same as code that's not present unless you think there's a reason that someone would manually uncomment this code and run it?

https://github.com/nautobot/nautobot-app-version-control/blob/c81c62b54b7a48eb53e1f4dd6c7220432c392409/nautobot_version_control/diffs.py#L162-L171

nautobot/dcim/forms.py
gsnider2195: Do you think that's necessary when the form is rendered only allowing one tab to be selected for device/module/location?

image

nautobot/dcim/tables/devices.py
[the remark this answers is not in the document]
glennmatthews: I think that's fine. Appreciate you taking a look!

nautobot/dcim/forms.py
glennmatthews: Fair enough - probably not.

nautobot/circuits/factory.py
```diff
@@ -83,9 +83,7 @@ class Meta:
         model = ProviderNetwork
         exclude = ("has_description", "has_comments")
 
-    name = factory.LazyAttribute(
-        lambda o: f"{o.provider.name} Network {faker.Faker().word(part_of_speech='noun')}"[:100]
-    )
+    name = factory.LazyAttributeSequence(lambda o, n: f"{o.provider.name} Network {n + 1}"[:100])
```
gsnider2195, line 86: Fixes random failures

nautobot/circuits/tests/test_filters.py
```diff
@@ -108,7 +108,9 @@ def test_location(self):
             has_location=True,
             location=locations[1],
         )
-        expected = self.queryset.filter(circuit_terminations__location__in=[locations[0].pk, locations[1].pk])
+        expected = self.queryset.filter(
```
gsnider2195, line 111: Random flaky test failure fixes

nautobot/core/forms/forms.py
```diff
@@ -121,7 +121,7 @@ class BulkRenameForm(forms.Form):
     """
 
     find = forms.CharField()
-    replace = forms.CharField()
+    replace = forms.CharField(required=False, strip=False)
```
gsnider2195, line 124: This allows you to have a blank replacement string, or to have leading and trailing whitespace on your replacement string. Examples:

`Interface1`

Replace `^Interface` with blank string to change the name to just `1`
Replace `^Interface` with `Interface ` to change the name to `Interface 1`

nautobot/core/testing/views.py
```diff
@@ -780,6 +785,10 @@ def test_list_objects_anonymous(self):
         @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
         def test_list_objects_filtered(self):
             instance1, instance2 = self._get_queryset().all()[:2]
+            if hasattr(self.model, "name") and instance1.name == instance2.name:
```
gsnider2195, line 788: Fixing issues with random data generation sometimes creating two instances with the same name

nautobot/core/testing/views.py
```diff
@@ -1363,6 +1379,7 @@ class BulkRenameObjectsViewTestCase(ModelViewTestCase):
             "replace": "\\1X",  # Append an X to the original value
             "use_regex": True,
         }
+        rename_field = "name"  # The model field to be changed
```
gsnider2195, line 1382: May be able to remove this when we add the `name` field to `ModuleBay`

nautobot/core/testing/views.py
```diff
@@ -1481,6 +1502,38 @@ class DeviceComponentTemplateViewTestCase(
...
+    class ModularDeviceComponentTemplateViewTestCase(DeviceComponentTemplateViewTestCase):
+        def _edit_object_test_setup(self):
+            test_instance = self._get_queryset().first()
+            self.update_data = {
+                "name": test_instance.name,
+                "device_type": getattr(getattr(test_instance, "device_type", None), "pk", None),
+                "module_type": getattr(getattr(test_instance, "module_type", None), "pk", None),
+                "label": "new test label",
+                "description": "new test description",
+            }
```
gsnider2195, line 1535: With the creation of the component factories, this test was setting statically defined data on randomly generated instances. That doesn't work for cases like `FrontPort` where the `RearPort` has to belong to the same device/module. We should update this test to work like the API tests where we can supply specific field values to change instead of having to provide the entire set of form data.

nautobot/dcim/factory.py
```diff
@@ -191,12 +189,7 @@ class Meta:
     has_comments = NautobotBoolIterator()
     comments = factory.Maybe("has_comments", factory.Faker("bs"))
 
-    has_software_version = NautobotBoolIterator()
-    software_version = factory.Maybe(
-        "has_software_version",
-        factory.LazyAttribute(lambda o: get_random_software_version_for_device_type(o.device_type)),
-        None,
-    )
+    software_version = factory.LazyAttribute(lambda o: get_random_software_version_for_device_type(o.device_type))
```
gsnider2195, line 192: DeviceType already has a 50% chance of having a software version

nautobot/dcim/tables/devicetypes.py
```diff
@@ -143,6 +147,40 @@ class Meta(BaseTable.Meta):
...
+        verbose_name="Modules",
+    )
+    tags = TagColumn(url_name="dcim:devicetype_list")
+
+    class Meta(BaseTable.Meta):
+        model = ModuleType
+        fields = (
+            "pk",
+            "model",
+            "manufacturer",
```
glennmatthews, line 170: Might as well linkify the `manufacturer` column too.

nautobot/dcim/templates/dcim/device/devicebays.html
```diff
@@ -18,21 +18,21 @@
...
                         <span class="mdi mdi-pencil" aria-hidden="true"></span> Rename
                     </button>
-                    <button type="submit" name="_edit" formaction="{% url 'dcim:devicebay_bulk_edit' %}?device={{ object.pk }}&return_url={{ object.get_absolute_url }}%23tab_devicebays" class="btn btn-warning btn-xs">
+                    <button type="submit" name="_edit" formaction="{% url 'dcim:devicebay_bulk_edit' %}?device={{ object.pk }}&return_url={% url 'dcim:device_devicebays' pk=object.pk %}" class="btn btn-warning btn-xs">
```
glennmatthews, line 24: Can we use return_url = request.path in this template too?

nautobot/dcim/templates/dcim/device/modulebays.html
```diff
@@ -0,0 +1,50 @@
...
+                <div class="pull-right noprint">
+                    {% if request.user.is_authenticated %}
+                        <button type="button" class="btn btn-default btn-xs" data-toggle="modal" data-target="#DeviceModuleBayTable_config" title="Configure table"><i class="mdi mdi-cog"></i> Configure</button>
+                    {% endif %}
+                </div>
+            </div>
+            {% render_table modulebay_table 'inc/table.html' %}
+            <div class="panel-footer noprint">
+                {% if perms.dcim.change_modulebay %}
+                    <button type="submit" name="_rename" formaction="{% url 'dcim:modulebay_bulk_rename' %}?return_url={% url 'dcim:device_modulebays' pk=object.pk %}" class="btn btn-warning btn-xs">
```
glennmatthews, line 21: Can we use return_url = request.path here and below?

nautobot/dcim/tests/test_models.py
```diff
@@ -2685,34 +2661,14 @@ def test_uniqueness_parent_module(self):
         with self.assertRaises(IntegrityError):
             module_bay.save()
 
-    def test_parent_property(self):
```
glennmatthews, line 2688: No longer relevant/needed?

nautobot/dcim/tests/test_views.py
```diff
@@ -1280,8 +1640,20 @@ def setUpTestData(cls):
             "feed_leg": PowerOutletFeedLegChoices.FEED_LEG_B,
         }
 
+    def _edit_object_test_setup(self):
```
glennmatthews, line 1643: If this is a base method that subclasses are permitted/expected to override, can we remove the `_` prefix? (I know, we have `_get_queryset()` overridden in plenty of places, but that one bugs me too...!)

nautobot/dcim/tests/test_views.py
```diff
@@ -1439,13 +1910,53 @@ def setUpTestData(cls):
...
+
+    def _edit_object_test_setup(self):
+        test_instance = self._get_queryset().first()
+        self.update_data = {
+            "position": test_instance.position,
+            "device_type": getattr(getattr(test_instance, "device_type", None), "pk", None),
+            "module_type": getattr(getattr(test_instance, "module_type", None), "pk", None),
+            "label": "new test label",
+            "description": "new test description",
+        }
```
glennmatthews, line 1957: Why can't we do this as part of `setUpTestData`?

nautobot/dcim/views.py
```diff
@@ -159,6 +176,18 @@ def get_selected_objects_parents_name(self, selected_objects):
         selected_object = selected_objects.first()
         if selected_object and selected_object.device:
             return selected_object.device.name
+        if selected_object and selected_object.module:
+            return selected_object.module.display
```
glennmatthews, line 180: Is this an issue that we use `name` for the device but `display` for the module? Should we use `display` for the device too like we do below?

glennmatthews (approved):

nautobot/core/views/renderers.py
```diff
@@ -263,62 +289,65 @@ def get_context(self, data, accepted_media_type, renderer_context):
...
-                        "fields": get_csv_form_fields_from_serializer_class(view.serializer_class),
-                    }
-                )
-            elif view.action in ["changelog", "notes"]:
-                context.update(
-                    {
-                        "base_template": get_base_template(data.get("base_template"), model),
-                        "active_tab": view.action,
-                    }
-                )
```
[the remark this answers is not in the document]
gsnider2195: I moved it to the mixin which I think is more appropriate instead of having a bunch of view-specific code in the renderer

nautobot/dcim/views.py
gsnider2195: It would be fine to use display for the device too here. That would help in case there is no name on a device

nautobot/dcim/tests/test_models.py
gsnider2195: We removed the different `device` and `parent` properties so that `ModuleBay.parent` would match all of the other components' `parent` property. So I renamed `test_device_property` to `test_parent_property`

gsnider2195: Sorry, to clarify I actually renamed the ModuleBay `device` property to `parent` and removed this old behavior of the `parent` property that was being tested here

nautobot/core/testing/views.py
gsnider2195: I think the issue might actually be where we're creating data that looks realistic like `Interface 1` - `Interface 24` on more than one device.

nautobot/dcim/tests/test_views.py
gsnider2195: It looks like we can

nautobot/dcim/tables/template_code.py
```diff
@@ -270,13 +270,27 @@
...
-        <a href="{% url 'dcim:devicebay_populate' pk=record.pk %}?return_url={% url 'dcim:device_devicebays' pk=object.pk %}" class="btn btn-success btn-xs">
+        <a href="{% url 'dcim:devicebay_populate' pk=record.pk %}?return_url={{ request.path }}" class="btn btn-success btn-xs">
             <i class="mdi mdi-plus-thick" aria-hidden="true" title="Install device"></i>
         </a>
     {% endif %}
 {% endif %}
 """
+
+MODULE_BUTTONS = """
+<a href="{% url 'dcim:module' pk=record.pk %}" class="btn btn-default btn-xs" title="Details"><i class="mdi mdi-information-outline" aria-hidden="true"></i></a>
```
HanlinMiao, line 285: do we need `perms.dcim.view_module` here?

nautobot/dcim/templates/dcim/modulebay_bulk_destroy.html
```diff
@@ -0,0 +1,5 @@
+{% extends "generic/object_bulk_destroy.html" %}
```
HanlinMiao, line 1: nit: can we merge it with `module_bulk_destroy.html`?

HanlinMiao (approved):

nautobot/dcim/tables/template_code.py
gsnider2195: I don't think these would be rendered if you don't have access to view modules (the table would be empty)

PR #5752, opened 2024-05-23
Updates incorrect updated import path for "ordering" in the v1 to v2 docs

qduk:
# What's Changed

- Incorrect import path change going from v1 to v2 in upgrade docs
The `ordering` module actually lives under `nautobot.core.models.ordering` and not `nautobot.core.utils.ordering`


glennmatthews (approved):
Thanks for the PR, though really this document shouldn't be used for app authorship anyway - e.g. in this case the "blessed" import path for apps is `nautobot.apps.models`.

PR #5755, opened 2024-05-24
use pytz timezones instead of zoneinfo

timizuoebideri1:
# Closes #NaN
# What's Changed

Used `pytz.common_timezones` instead of `zoneinfo.available_timezones()` for timezone choices because not all `zoneinfo` options are valid for `TimeZoneField`.
# Screenshots

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/factory.py
+++ b/nautobot/dcim/factory.py
@@ -5,11 +5,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.db.models import Q
 import factory
 from faker import Faker
-
-try:
-    import zoneinfo
-except ImportError:  # python 3.8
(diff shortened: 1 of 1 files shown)


glennmatthews:
I don't think this is the right approach - pytz is deprecated in Django 4 and will be removed entirely in Django 5.

gsnider2195:
It looks like the fix here is to just discard the `Factory` timezone to match django timezone fields: https://github.com/mfogel/django-timezone-field/blob/cb9b81be6f487df3184424f3795adc30dac06056/timezone_field/backends/zoneinfo.py#L15C1-L16C35

PR #5757, opened 2024-05-24
Dynamic group permissions improvements

glennmatthews:
# Closes #n/a
# What's Changed

- Make dynamic groups `/members/` API endpoint only return the members that a user has permission to view
- Make dynamic groups detail UI only show members that a user has permission to view
- Make object dynamic-groups detail tab only show DGs that a user has permission to view.
- Add test coverage for the above.

If accepted, we'll also want to:
- backport this changeset to `ltm-1.6`
- implement a similar set of changes in `next` for the static groups feature.

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/tests/test_views.py
+++ b/nautobot/extras/tests/test_views.py
@@ -775,17 +775,19 @@ class DynamicGroupTestCase(
 
     @classmethod
     def setUpTestData(cls):
-        cls.content_type = ContentType.objects.get_for_model(Device)
+        content_type = ContentType.objects.get_for_model(Device)
 
         # DynamicGroup objects to test.
-        DynamicGroup.objects.create(name="DG 1", content_type=cls.content_type)
-        DynamicGroup.objects.create(name="DG 2", content_type=cls.content_type)
-        DynamicGroup.objects.create(name="DG 3", content_type=cls.content_type)
+        cls.dynamic_groups = [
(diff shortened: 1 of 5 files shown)


nautobot/extras/tests/test_views.py
```diff
@@ -815,7 +850,6 @@ def test_get_object_dynamic_groups_with_permission(self):
         self.assertIn("DG 3", response_body, msg=response_body)
 
     def test_get_object_dynamic_groups_with_constrained_permission(self):
```
gsnider2195, line 852: What is the difference between this and the `test_get_object_with_constrained_permission` above?

gsnider2195 (approved):

nautobot/extras/tests/test_views.py
glennmatthews: This is `/dcim/devices/<pk>/dynamic-groups/`, i.e. the "Dynamic Groups" tab of the Device detail view.

The test above is `/extras/dynamic-groups/<pk>`, i.e. the Dynamic Group detail view (which includes the "Members" tab, listing Devices in this case).

Two sides of the same coin.

gsnider2195: Got it. Thanks!

glennmatthews:
I'll add a change fragment and merge this on Tuesday.

Issue #5471, opened 2024-03-22
Saved Views: User and Global Defaults


lampwins:
As a Nautobot Admin, I want to be able to set a predefined saved view as the default for all users, so that these users do not have to select the view for themselves.

AC:

I know this is complete when a user with adequate permissions is able to assign existing saved views to be the global default for all users for a given model list view.

Individual users can still create their own views and set their own, specific default which then overrides the global default.

- An order of precedence will be defined as:
  - User-specific default saved view
  - Globally defined default saved view
  - No default saved view

As a "shared" boolean to the view model to specify if the view is private to a user or shared for all users.

lampwins:
Per #5774 when this is implemented, the permissions need to then allow for "views I own, or shared views."

lampwins:
For setting the global default, that should be an operation that can happen only from the saved view list view, not for the normal, user facing management of saved views.

When a global has already been set for a given content type, allow the admin user to set a new global and display in a message: "Global default saved view for devices has been changed from A to B"

HanlinMiao:
1. Setting a global default view when a global default view already exists should display a message informing the users that the global default view is changing from `saved view 1` to `saved view 2`.
2. Setting a global default view should only be available to admin users with `users.change_savedview` from the SavedViewListView

PR #5783, opened 2024-06-04
Fixed the bug that required users and administrators to manage additional permission to be able to use saved views.

### LINKED ISSUE #5774: Saved View Permissions
As a user and administrator, I don't to have to manage additional permissions just to be able to use saved views. As a user, if I have access to a list view, I expect to be able to use/create/manage saved views with it.

We will want to consider the other, more administrative duties with saved views, like setting defaults for users. We also don't want user A to be able to randomly delete user B's saved views either.

HanlinMiao:
# Closes #5774 
# What's Changed

- [x] User A should be able to create new views
- [x] User A should be able to edit his own views
- [x] User A should be able to see his own views and shared views
- [x] User A should be able to delete his own views
- [x] User A cannot edit views he does not own
- [x] User A cannot delete views he does not own

# Screenshots

# TODO

- [ ] Unit, Integration Tests


nautobot/core/templatetags/helpers.py
```diff
@@ -756,7 +756,7 @@ def saved_view_modal(
             if current_saved_view_pk:
                 current_saved_view_pk = current_saved_view_pk[0]
                 try:
-                    current_saved_view = SavedView.objects.restrict(request.user, "view").get(pk=current_saved_view_pk)
+                    current_saved_view = SavedView.objects.get(pk=current_saved_view_pk)
```
glennmatthews, line 759: Mind adding a comment here to the effect that we're intentionally *not* calling `restrict()` here and why?

nautobot/core/views/generic.py
```diff
@@ -271,16 +271,12 @@ def get(self, request):
         current_saved_view = None
         current_saved_view_pk = self.request.GET.get("saved_view", None)
         list_url = validated_viewname(model, "list")
-        saved_views = (
-            SavedView.objects.filter(view=list_url).restrict(request.user, "view").order_by("name").only("pk", "name")
-        )
+        saved_views = SavedView.objects.filter(view=list_url).order_by("name").only("pk", "name")
```
glennmatthews, line 274: Same request here and below.

nautobot/users/views.py
```diff
@@ -250,17 +250,29 @@ class SavedViewUIViewSet(
...
+            actions_copy = [self.get_action()]
             for i, item in enumerate(actions):
                 # Enforce users:change_savedview instead of users:clear_savedview
                 if item == "clear":
-                    actions[i] = "change"
+                    actions_copy[i] = "change"
+                # All users should be able to view existing saved views and add new saved views.
+                if item in ["add", "view"]:
+                    actions_copy.pop(i)
+            actions = actions_copy
```
glennmatthews, line 275: Not sure I understand what we're doing here or why `actions_copy` is needed. Mind adding some comments?

nautobot/core/templates/generic/object_list.html
```diff
@@ -53,49 +53,43 @@
...
+                    </a>
+                </li>
+                {% if current_saved_view %}
                     <li>
-                        <a href="" data-toggle="modal" data-target="#clear_view_modal" title="Confirm Clear View">
-                            <i class="mdi mdi-arrow-u-right-bottom text-danger" aria-hidden="true"></i>
-                            <span class="text-danger">Clear View</span>
+                        <a href="{% url 'user:savedview_delete' pk=request.GET.saved_view %}">
+                            <i class="mdi mdi-trash-can-outline text-danger" aria-hidden="true"></i>
+                            <span class="text-danger">Delete current saved view</span>
```
glennmatthews, line 92:
```suggestion
                            <span class="text-danger">Delete Current Saved View</span>
```

nautobot/core/templates/generic/object_list.html
```diff
@@ -53,49 +53,43 @@
...
-                    {% endif %}
-                {% endif %}
-                {% if perms.users.add_savedview %}
                     <li>
-                        <a href="" data-toggle="modal" data-target="#saved_view_modal" title="Save Current View As">
-                            <i class="mdi mdi-content-save-plus-outline text-muted" aria-hidden="true"></i>
-                            <span>Save As New View</span>
+                        <a href="{% url 'user:savedview_edit' pk=request.GET.saved_view %}{% if request.GET %}?{{ request.GET.urlencode }}{% endif %}">
+                            <i class="mdi mdi-content-save-outline text-muted" aria-hidden="true"></i>
+                            <span>Update Current View</span>
```
glennmatthews, line 78:
```suggestion
                            <span>Update Current Saved View</span>
```

Also I wonder if it'd make sense to move this below the "Save As New View" item so that we only need the one `if current_saved_view` block?

nautobot/users/tests/test_views.py
```diff
@@ -235,14 +236,29 @@ def test_get_object_with_constrained_permission(self):
...
         # Try GET to non-permitted object
-        self.assertHttpStatus(self.client.get(instance2.get_absolute_url()), 404)
+        # Should be able to get to any SavedView instance as long as the user has "{app_label}.view_{model_name}" permission
+        self.assertHttpStatus(self.client.get(instance2.get_absolute_url()), 302)
 
     @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
     def test_update_saved_view(self):
         instance = self._get_queryset().first()
-        self.add_permissions("users.change_savedview")
+        # self.add_permissions("users.change_savedview")
```
glennmatthews, line 245: Remove rather than commenting out?

nautobot/core/templates/generic/object_list.html
HanlinMiao: Little pushback on that, I think the first operation you would look for is an update. So putting it first would make sense. The if...endif...block is not too much overhead.

glennmatthews: Fair enough!

changes/5774.fixed
```diff
@@ -0,0 +1 @@
+Fixed the bug that required users and administrators to manage additional permission to be able to use saved views.
```
glennmatthews, line 1: Thinking about this more, let's make sure the saved-view docs clearly explain how permissions do and do not intersect with saved-views...

nautobot/core/templatetags/helpers.py
```diff
@@ -756,7 +756,9 @@ def saved_view_modal(
             if current_saved_view_pk:
                 current_saved_view_pk = current_saved_view_pk[0]
                 try:
-                    current_saved_view = SavedView.objects.restrict(request.user, "view").get(pk=current_saved_view_pk)
+                    # We are not using .restrict(request.user, "view") here
+                    # User should be able to see any saved view that he has the list view access to.
+                    current_saved_view = SavedView.objects.get(pk=current_saved_view_pk)
```
glennmatthews, line 761: This may need to be revisited when we implement #5471 - I'm guessing at that point we should make it so that a non-staff user can view their own saved views, and those marked as "shared", but not any others?

Might make sense to create a helper function specifically for saved-view queryset restriction, as it may need to be used in quite a few places.

nautobot/users/tests/test_views.py
```diff
@@ -169,24 +169,27 @@ def get_view_url_for_saved_view(self, saved_view, action="detail"):
...
-        response = self.client.get(self._get_queryset().first().get_absolute_url())
+        response = self.client.get(self._get_queryset().first().get_absolute_url(), follow=True)
         self.assertHttpStatus(response, 200)
         response_body = response.content.decode(response.charset)
         self.assertIn(
-            "/login/?next=" + self._get_queryset().first().get_absolute_url(), response_body, msg=response_body
+            f'<input type="hidden" name="next" value="{self._get_queryset().first().get_absolute_url()}" />',
+            response_body,
+            msg=response_body,
         )
```
glennmatthews, line 179: what's the story on this change? Retrieving a saved view should still require `view` permissions on the object list view that it's associated with, right? So an unauthenticated user should still get redirected to login?

nautobot/users/tests/test_views.py
```diff
@@ -235,14 +238,28 @@ def test_get_object_with_constrained_permission(self):
         self.assertHttpStatus(self.client.get(instance1.get_absolute_url()), 302)
 
         # Try GET to non-permitted object
-        self.assertHttpStatus(self.client.get(instance2.get_absolute_url()), 404)
+        # Should be able to get to any SavedView instance as long as the user has "{app_label}.view_{model_name}" permission
+        self.assertHttpStatus(self.client.get(instance2.get_absolute_url()), 302)
```
glennmatthews, line 242: Is this test now *assuming* that instance1 and instance2 reference the same list view?

nautobot/users/views.py
```diff
@@ -250,30 +248,45 @@ class SavedViewUIViewSet(
     table_class = SavedViewTable
     action_buttons = ("export",)
 
-    def get_required_permission(self):
+    def get_queryset(self):
         """
-        Obtain the permissions needed to perform certain actions on a model.
+        Get the list of items for this view.
+        All users should be able to see saved views so we do not apply extra permissions.
```
glennmatthews, line 254: This too would need to restrict to views owned by the user, right?

nautobot/users/views.py
```diff
@@ -250,30 +248,45 @@ class SavedViewUIViewSet(
...
+            # which will be handled by self.handle_no_permission() in the UI appropriately in the dispatch() method
+            # Cast permission to a list since has_perms() takes a list type parameter.
+            if "savedview" in permission:
+                continue
+            if not user.has_perms([permission]):
+                self.permission_denied(
+                    request,
+                    message=getattr(permission, "message", None),
+                    code=getattr(permission, "code", None),
+                )
```
glennmatthews, line 277: Not entirely sure I follow this one - you no longer have a custom `get_required_permission` implementation, so won't `permission_required` just always be `<verb>_savedview`? In which case you'll always hit the condition at lines 270-271, and never fall through to the other case?

nautobot/users/views.py
```diff
@@ -250,30 +248,45 @@ class SavedViewUIViewSet(
...
+
+    def list(self, request, *args, **kwargs):
+        if isinstance(request.user, AnonymousUser):
+            return self.handle_no_permission()
```
glennmatthews, line 281: Should be able to use `permission_classes = (IsAuthenticated,)` at the class level so as to not need to specify this in each action method.

HanlinMiao: `permisssion_classes` does not seem to work. I will investigate more in the morning.

nautobot/users/views.py
HanlinMiao: I am not sure this is needed since `<app_label>.view_<model_name>` permissions are enforced at each list view. Also it is difficult to handle the case of `AnonymousUser`. Tried `self.queryset.filter(owner=None)` and it does not work.

nautobot/users/views.py
HanlinMiao: Refactored

nautobot/users/tests/test_views.py
HanlinMiao: This should be 
```
url = reverse(instance.view)
self.assertIn("/login/?next=" + url, response_body, msg=response_body)
```

changes/5774.fixed
HanlinMiao: Will include that with #5471

nautobot/users/tests/test_views.py
HanlinMiao: Sorry it should be `self.assertRedirects(response, f"/login/?next={instance.get_absolute_url()}")`

nautobot/users/tests/test_views.py
HanlinMiao: Refactored

nautobot/users/views.py
glennmatthews: No luck on this?

glennmatthews: OK yeah... we should be following or at least supporting the patterns used in DRF. Follow up issue is fine for now.

glennmatthews (approved):

PR #5779, opened 2024-06-03
Added a feature that replaces {module} with the selected module's parent_module_bay when creating a component template.

timizuoebideri1:
# Closes #5730
# What's Changed

# Screenshots

## SFP ModuleType with templated interface name

image

## SFP created in a module in inventory, then the parent module is installed in a device

image

image

image

image

## SFP installed in another module bay while the line card is still installed in the device

image

image

# TODO

- [ ] Explanation of Change(s)
- [ ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [ ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example App Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #5784, opened 2024-06-04
add name field to modulebay and modulebaytemplate

### LINKED ISSUE #5776: Add `name` field to `ModuleBay`
Follow-up to #2101 

ModuleBay and ModuleBayTemplate should have a name field, similar to all other components. The position field will be used for templating the name of child components (#5730)

gsnider2195:
# Closes #5776
# What's Changed

We identified a good use case for having `name` and `position` on the `ModuleBay` model, so this PR changes the `ModuleBay` and `ModuleBayTemplate` models to be based off of the same base class as the rest of the component models, using the `device`+`name` or `module`+`name` fields as unique constraints instead of the prior implementation that used `device`+`position`/`module`+`position`. The `position` field is now optional and is used to populate child component names but not much else at this time. A lot of the changes here are reverting custom code that was necessary to use a non-standard field name for these components.

# Screenshots

# TODO

- [ ] Unit, Integration Tests


nautobot/dcim/models/device_component_templates.py
```diff
@@ -465,54 +465,25 @@ def clean(self):
...
-        related_name="module_bay_templates",
-        blank=True,
-        null=True,
-    )
-    _position = NaturalOrderingField(target_field="position", max_length=CHARFIELD_MAX_LENGTH, blank=True)
     position = models.CharField(
         max_length=CHARFIELD_MAX_LENGTH,
-        blank=False,
-        null=False,
+        blank=True,
```
glennmatthews, line 473: Does this becoming blankable make #5779 more difficult to implement? Would that be easier if we keep this as a required field?

nautobot/dcim/forms.py
```diff
@@ -1526,37 +1526,55 @@ class Meta:
...
     label_pattern = ExpandableNameField(
         label="Label",
         required=False,
-        help_text="Alphanumeric ranges are supported. (Must match the number of positions being created.)",
+        help_text="Alphanumeric ranges are supported. (Must match the number of names being created.)",
+    )
+    position_pattern = ExpandableNameField(
+        label="Position",
+        required=False,
+        help_text="Numeric ranges are supported. (Must match the number of names being created.)",
```
glennmatthews, line 1546: Numeric only? Not alphanumeric as above?

nautobot/dcim/models/device_component_templates.py
gsnider2195: Doing this for compatibility with devicetype library and a potential use case where you have nested modules but you want to use the position from the grandparent to template out your interface name and not the direct parent module bay

nautobot/dcim/models/device_component_templates.py
```diff
@@ -465,54 +465,25 @@ def clean(self):
 
 
 @extras_features("custom_validators")
```
glennmatthews, line 467: Do our usual other features (custom_links, export_templates, graphql, webhooks) not apply to ModuleBayTemplate? Or is it just that this is the pattern we have for all the other ComponentTemplateModels and something we should fix across the board?

nautobot/dcim/tests/test_views.py
```diff
@@ -1045,9 +1048,12 @@ def test_import_objects_unknown_type_enums(self):
...
 module-bays:
-  - position: Module Bay 1
-  - position: Module Bay 2
-  - position: Module Bay 3
+  - name: Module Bay 1
+    position: 1
+  - name: Module Bay 2
+    position: 2
+  - name: Module Bay 3
+    position: 3
```
glennmatthews, line 1056: No test change corresponding to this data change?

nautobot/dcim/tests/test_views.py
```diff
@@ -1415,9 +1424,12 @@ def test_import_objects_unknown_type_enums(self):
...
```
glennmatthews, line 1432: No test change corresponding to this data change?

nautobot/dcim/tests/test_views.py
```diff
@@ -1961,14 +1972,14 @@ def setUpTestData(cls):
         cls.form_data = {
             "device_type": device_type.pk,
             "module_type": None,
-            "position": "Module Bay Template X",
+            "name": "Module Bay Template X",
```
glennmatthews, line 1975: We should have both `name` and `position` now, right?

nautobot/dcim/tests/test_views.py
```diff
@@ -1961,14 +1972,14 @@ def setUpTestData(cls):
...
             "description": "Test modulebaytemplate description",
             "label": "Test modulebaytemplate label",
         }
 
         cls.bulk_create_data = {
             "module_type": module_type.pk,
-            "position_pattern": "Test Module Bay Template [5-7]",
+            "name_pattern": "Test Module Bay Template [5-7]",
```
glennmatthews, line 1982: We should have both `name_pattern` and `position_pattern` now right?

nautobot/dcim/tests/test_views.py
```diff
@@ -1979,7 +1990,7 @@ def setUpTestData(cls):
 
         test_instance = cls.model.objects.first()
         cls.update_data = {
-            "position": test_instance.position,
+            "name": test_instance.name,
```
glennmatthews, line 1993: Can we update both `name` and `position`?

nautobot/dcim/models/device_component_templates.py
glennmatthews: Ah right - we don't have detail-views for these models, do we? So custom_links at least wouldn't apply. I think the others would though. I'll open an issue.

glennmatthews: #5789

glennmatthews (approved):

PR #5793, opened 2024-06-06
Some `generate_test_data` fixes/improvements

glennmatthews:
# Closes #N/A
# What's Changed

- Fix potentially unpredictable test data from `ControllerManagedDeviceGroupFactory` due to using `order_by("?")` instead of the preferred `factory.random.randgen.choice()`
- Refactor implementation of `generate_test_data` so that:
    - every `_create_batch()` call automatically passes `using=db_name` to the factory; we had a number of cases where this parameter was missing which *may* have caused problems.
    - in CI, or when `GITHUB_ACTIONS` environment variable is set, each `_create_batch()` call automatically prints the hash of object PKs resulting from the factory, rather than printing all model hashes at once in arbitrary (alphabetical) order at the end; this should improve troubleshooting readability of this information.

# Screenshots

```
Seeding the pseudo-random number generator with seed "Nautobot"...
Creating Roles...
  SHA256 hash of Role records: aa41a44e6836afa23c6ea724e759168220e0dcb25f489fe90f284dada30a6505
Creating Statuses...
  SHA256 hash of Status records: b942529ead783565a6e5b963ca9f4dba6aa87e0e46024255928d03eefcf03d13
Creating Tags...
  SHA256 hash of Tag records: 229ffe221f51ffe7d8d474a413b83088e62bcb277802ba9b4dea563bdfeec636
  SHA256 hash of Tag records: 45f5df0201326f019a8ede9191d11bdb6879fed992ccc7c82ee26326145797ef
...
```

# TODO
- n/a Unit, Integration Tests


nautobot/core/management/commands/generate_test_data.py
```diff
@@ -83,7 +83,6 @@ def _generate_factory_data(self, seed, db_name):
             from nautobot.extras.utils import TaggableClassesQuery
             from nautobot.ipam.choices import PrefixTypeChoices
             from nautobot.ipam.factory import (
-                IPAddressFactory,
```
gsnider2195, line 86: The IPAddressFactory gets called as a subfactory of PrefixFactory. Should we output the hash for IPAddressFactory at the same time as PrefixFactory?

nautobot/core/management/commands/generate_test_data.py
```diff
@@ -101,178 +100,134 @@ def _generate_factory_data(self, seed, db_name):
         self.stdout.write(f'Seeding the pseudo-random number generator with seed "{seed}"...')
         factory.random.reseed_random(seed)
 
+        def _create_batch(some_factory, count, **kwargs):
```
gsnider2195, line 103: We could probably move the console output in here too.

nautobot/core/management/commands/generate_test_data.py
```diff
@@ -101,178 +100,134 @@ def _generate_factory_data(self, seed, db_name):
...
-                VLANFactory,
-                VLANGroupFactory,
-                VRFFactory,
-            ]
-        )
-
-    def _output_hash_for_factory_models(self, factories):
-        """Output a hash of the IDs of all objects in the given factories' model.
-
-        Used for identifying factory determinism problems in unit tests. Only prints if GITHUB_ACTIONS environment variable is set to "true".
```
gsnider2195, line 265: Can we keep this docstring to make it more obvious why we're outputting these hashes? Maybe keep this as a separate method so we can have the docstring here?

gsnider2195 (approved):
Thank you!

nautobot/core/management/commands/generate_test_data.py
glennmatthews: Maybe, but would have to be a special case in the `_create_batch` function, not sure it's worth it?

nautobot/core/management/commands/generate_test_data.py
glennmatthews: Added a `--print-hashes` option to the command with appropriate help text; keeping the current behavior of `$GITHUB_ACTIONS=true` automatically enabling hash printing.

nautobot/core/management/commands/generate_test_data.py
```diff
@@ -101,178 +111,135 @@ def _generate_factory_data(self, seed, db_name):
...
-        self.stdout.write("Creating Prefixes and IP Addresses...")
+        _create_batch(LocationFactory, 7, description="as structure", has_parent=True)
+        _create_batch(LocationFactory, 40)
+        _create_batch(LocationFactory, 10, description="without a parent Location", has_parent=False)
+        _create_batch(ControllerFactory, 1, description="without a Device or DeviceRedundancyGroup")
+        _create_batch(ControllerManagedDeviceGroupFactory, 5, description="to contain Devices")
+        _create_batch(RIRFactory, 9)  # only 9 unique RIR names are hard-coded presently
+        _create_batch(RouteTargetFactory, 20)
+        _create_batch(NamespaceFactory, 10)
+        _create_batch(VRFFactory, 20)
```
HanlinMiao, line 152: With or without tenants? Is it important to preserve that distinction?

HanlinMiao (approved):

nautobot/core/management/commands/generate_test_data.py
glennmatthews: Good spot. My thinking was that `has_tenant` would naturally be a 50/50 coin flip, but looking again at the IPAM factories it's actually coded as 75% yes / 25% no (`has_tenant = factory.Faker("boolean", chance_of_getting_true=75)`). I could revert this part, but given that it's passing tests anyway with my change, I'd prefer to keep it since it's one less set of "special" calls to the factories (see also my TODO comment below on the CircuitTerminationFactory calls)

PR #5807, opened 2024-06-12
feat: Add search and sort to ipam.IPAddress.nat_inside

abates:
# Closes N/A
# What's Changed

This change updates the filterset, table and search form for `ipam.IPAddress` to make the nat_inside field sortable and searchable.

# Screenshots
Screenshot 2024-06-12 at 12 41 59 PM
Screenshot 2024-06-12 at 12 42 13 PM

# TODO
- [x] Unit, Integration Tests
- [ ] ~Example App Updates (when adding/changing features)~
- [ ] ~Outline Remaining Work, Constraints from Design~

### DIFF
--- a/nautobot/ipam/tests/test_filters.py
+++ b/nautobot/ipam/tests/test_filters.py
@@ -462,6 +462,7 @@ class IPAddressTestCase(FilterTestCases.FilterTestCase, FilterTestCases.TenancyF
     queryset = IPAddress.objects.all()
     filterset = IPAddressFilterSet
     tenancy_related_name = "ip_addresses"
+    generic_filter_tests = (["nat_inside", "nat_inside__id"],)
 
     @classmethod
     def setUpTestData(cls):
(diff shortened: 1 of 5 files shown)


nautobot/ipam/tests/test_filters.py
```diff
@@ -638,6 +638,13 @@ def setUpTestData(cls):
             status=statuses[0],
             namespace=cls.namespace,
         )
+        IPAddress.objects.create(
+            address="10.1.1.1/32",
+            tenant=None,
+            status=statuses[0],
+            namespace=cls.namespace,
+            nat_inside=ip0,
+        )
```
glennmatthews, line 647: A bit surprised that this test case doesn't already have a `generic_filter_tests` array, but you could add one after line 464 to giver coverage for the `nat_inside` filter:

```py
generic_filter_tests = (
    ["nat_inside", "nat_inside__id"],
)
```

You might also need to add another IPAddress with a non-null `nat_inside` as IIRC the test case wants to be able to find at least two unique values to filter with.

abates: I added the generic filter tests and discovered that the filter was actually implemented wrong. Both the test and filterset have been updated.

glennmatthews (approved):

PR #5808, opened 2024-06-12
User and Global Default and `is_shared` variable

### LINKED ISSUE #5471: Saved Views: User and Global Defaults

HanlinMiao:
# Closes #5471 
# What's Changed

- [x] Added the ability to set Global and User default saved view.
- [x] Added the ability to set public and private saved view via `is_shared` attribute
- [x] Added tests for the new features
- [x] Updated user guide and documentation for the new features

# Screenshots

Admin Saved View Edit: (Could only set is_global_default)
Screenshot 2024-06-17 at 10 22 22 AM
Users can set a saved view as default from the new dropdown button:
Screenshot 2024-06-17 at 10 23 08 AM
New saved view modal update: (is_shared_checkbox)
Screenshot 2024-06-17 at 10 22 56 AM

# TODO

- [x] Unit, Integration Tests


nautobot/core/templates/generic/object_list.html
```diff
@@ -75,11 +75,17 @@
...
                     <li>
-                        <a href="{% url 'user:savedview_edit' pk=request.GET.saved_view %}{% if request.GET %}?{{ request.GET.urlencode }}{% endif %}">
+                        <a href="{% url 'user:savedview_update_config' pk=request.GET.saved_view %}{% if request.GET %}?{{ request.GET.urlencode }}{% endif %}">
                             <i class="mdi mdi-content-save-outline text-muted" aria-hidden="true"></i>
                             <span>Update Current Saved View</span>
                         </a>
                     </li>
+                    <li>
+                        <a href="{% url 'user:savedview_set_default' pk=request.GET.saved_view %}{% if request.GET %}?{{ request.GET.urlencode }}{% endif %}">
+                            <i class="mdi mdi-view-compact text-muted" aria-hidden="true"></i>
```
glennmatthews, line 85: Maybe `mdi-star` or `mdi-heart` as an icon here?

nautobot/core/templates/utilities/templatetags/saved_view_modal.html
```diff
@@ -19,6 +19,8 @@ <h4 class="modal-title">New Saved View</h4>
                                 <pre><code class="language-json" style="background-color: #F2F2F2;">{{ param_dict }}</code></pre>
                             </div>
                         </div>
+                        {% render_field form.is_global_default %}
```
glennmatthews, line 22: Remove?

nautobot/core/views/generic.py
```diff
@@ -197,10 +200,39 @@ def get(self, request):
...
+            if not isinstance(user, AnonymousUser):
+                user_default_saved_view_pk = user.config_data.get("saved_views", {}).get(view_name, None)
+                if user_default_saved_view_pk is not None and not request.GET.get("saved_view"):
+                    try:
+                        SavedView.objects.get(pk=user_default_saved_view_pk)
+                        sv_url = reverse("users:savedview", kwargs={"pk": user_default_saved_view_pk})
+                        return redirect(sv_url)
+                    except ObjectDoesNotExist:
+                        # Saved view was deleted
+                        pass
```
glennmatthews, line 225: Should we clear the config_data entry in this case?

Same question on the equivalent UIViewSet code.

nautobot/core/views/generic.py
```diff
@@ -197,10 +200,39 @@ def get(self, request):
...
+        if not clear_view:
+            # Check if there is a default for this view for this specific user
+            app_label, model_name = model._meta.label.split(".")
+            view_name = f"{app_label}:{model_name.lower()}_list"
+
+            if not isinstance(user, AnonymousUser):
+                user_default_saved_view_pk = user.config_data.get("saved_views", {}).get(view_name, None)
+                if user_default_saved_view_pk is not None and not request.GET.get("saved_view"):
+                    try:
+                        SavedView.objects.get(pk=user_default_saved_view_pk)
```
glennmatthews, line 220: Should we enforce that either `user=user` or `is_shared=True` on this `get()`? So that a user can't have a default saved view that belongs to another user and isn't shared?

Same question on the equivalent UIViewSet code.

nautobot/core/views/generic.py
```diff
@@ -197,10 +200,39 @@ def get(self, request):
...
+                        pass
+
+            # Check if there is a global default for this view
+            global_saved_view = None
+            try:
+                global_saved_view = SavedView.objects.get(view=view_name, is_global_default=True)
+            except ObjectDoesNotExist:
+                pass
+
+            if global_saved_view is not None and not request.GET.get("saved_view"):
```
glennmatthews, line 234: If an explicit `saved_view` was specified, we should be able to skip all of the above lookups - no need to look for a user default view or a global default view...?

Same comment on the UIViewSet code.

nautobot/users/models.py
```diff
@@ -326,3 +328,12 @@ class Meta:
 
     def __str__(self):
         return f"{self.owner.username} - {self.view} - {self.name}"
+
+    def clean(self):
+        super().clean()
+
+        # If this SavedView is set to a global default, all other saved views related to this view name should not be the global default.
+        if self.is_global_default:
+            SavedView.objects.filter(view=self.view).exclude(pk=self.pk).update(is_global_default=False)
```
glennmatthews, line 337: Perhaps:

```suggestion
            SavedView.objects.filter(view=self.view, is_global_default=True).exclude(pk=self.pk).update(is_global_default=False)
```

nautobot/users/models.py
```diff
@@ -326,3 +328,12 @@ class Meta:
...
+            # If the view is set to Global default, is_shared should be true regardless
+            self.is_shared = True
```
glennmatthews, line 339: This feels like it belongs in `save()` rather than `clean()` to me since it has side effects.

nautobot/users/tests/test_models.py
```diff
@@ -14,6 +14,19 @@ def setUp(self):
...
+
+    def setUp(self):
+        self.user = User.objects.create_user(username="Saved View test user")
+        self.sv = SavedView.objects.create(name="Saved View", owner=self.user, view="dcim:location_list")
+
+    def test_is_global_default_saved_view_is_shared_automatically(self):
+        self.sv.is_global_default = True
+        self.sv.save()
+        self.assertEqual(self.sv.is_shared, True)
+
```
glennmatthews, line 28: Would be good to have a test case for:
- setting one saved-view as global default unsets the is_global_default flag on an existing default for the same view
- multiple saved-views for *different* views can all have `is_global_default` set.

nautobot/users/tests/test_views.py
```diff
@@ -354,3 +354,99 @@ def test_create_saved_view(self):
...
+            owner=self.user,
+            view=view_name,
+            is_shared=True,
+            is_global_default=True,
+        )
+        sv = SavedView.objects.create(
+            name="User Location Default View",
+            owner=self.user,
+            view=view_name,
+            is_shared=True,
```
glennmatthews, line 414: unnecessary?

nautobot/users/tests/test_views.py
```diff
@@ -354,3 +354,99 @@ def test_create_saved_view(self):
...
+        )
+        self.user.config_data["saved_views"] = {}
+        self.user.config_data["saved_views"][view_name] = sv.pk
```
glennmatthews, line 417: do you need to call `self.user.save()`?

nautobot/users/tests/test_views.py
```diff
@@ -354,3 +354,99 @@ def test_create_saved_view(self):
...
+        response = self.client.get(reverse(view_name), follow=True)
+        # Assert that Location List View got redirected to Saved View set as user default
+        self.assertHttpStatus(response, 200)
+        response_body = extract_page_body(response.content.decode(response.charset))
+        self.assertIn(
+            "User Location Default View",
+            response_body,
+            msg=response_body,
+        )
```
glennmatthews, line 426: Wonder if this is having a false positive from the presence of the view in the "Saved Views" menu. Might need to use a more specific `assertInHTML` call to check for it in the `<h1>` containing the page title?

Same concern about the two tests above, actually.

nautobot/users/views.py
```diff
@@ -242,17 +243,34 @@ class SavedViewUIViewSet(
...
+    form_class = SavedViewForm
     filterset_class = SavedViewFilterSet
     serializer_class = SavedViewSerializer
     table_class = SavedViewTable
     action_buttons = ("export",)
 
+    def alter_queryset(self, request):
+        """
+        Two scenarios we need to handle here:
+        1. User can view all saved views with users.view_savedview permission.
```
glennmatthews, line 255: Wonder if we should still enforce object permissions on `view_savedview`, i.e. admin A can view saved views belonging to users in group A, but not saved views for users in group B?

nautobot/core/views/mixins.py
```diff
@@ -750,6 +777,18 @@ class ObjectEditViewMixin(NautobotViewSetMixin, mixins.CreateModelMixin, mixins.
...
+        """
+        Context variables for this extra message.
+        """
+        return {}
+
+    def extra_message(self, **kwargs):
+        """
+        Append extra message at the end of create or update success message.
+        """
+        return ""
```
glennmatthews, line 790: Why not just make another separate call to `messages.success()`?

nautobot/users/views.py
```diff
@@ -266,6 +284,25 @@ def dispatch(self, request, *args, **kwargs):
...
+        """
+        return {"new_global_default_view": obj}
+
+    def extra_message(self, **kwargs):
+        new_global_default_view = kwargs.get("new_global_default_view")
+        view_name = new_global_default_view.view
+        message = ""
+        if new_global_default_view.is_global_default:
+            message += f"<br>The global default saved View for '{view_name}' is set to <a href='{new_global_default_view.get_absolute_url()}'>{new_global_default_view.name}</a>."
+        return message
```
glennmatthews, line 299: Is this redundant with lines 466-474 below? If not, why not?

nautobot/core/views/generic.py
HanlinMiao: Sorry if I am not understanding this correctly. The second part of the if statement `not request.GET.get("saved_view")` is basically acting as you said skipping the check if there is a saved view specified.

nautobot/core/views/generic.py
HanlinMiao: It shouldn't really matter but we can clear it just in case.

nautobot/core/views/generic.py
glennmatthews: Sorry - my point was that it's a waste of time to look up the user default and global default saved views if we can know in advance that we don't need them. So something like:

```
if not request.GET.get("saved_view"):
    # look for user default view if any, and apply it if found
    # else look for global default view if any, and apply it if found
```

would be more efficient and possibly cleaner as well.

nautobot/users/tests/test_views.py
HanlinMiao: I changed those tests to check `<strong>User Location Default View</strong>` since the save view name will only be bolded if the saved view is selected.

nautobot/core/views/generic.py
```diff
@@ -197,10 +200,40 @@ def get(self, request):
         content_type = ContentType.objects.get_for_model(model)
 
         filter_params = request.GET
+        user = request.user
         display_filter_params = []
         dynamic_filter_form = None
         filter_form = None
         hide_hierarchy_ui = False
+        clear_view = filter_params.get("clear_view", False)
```
gsnider2195, line 208:
```suggestion
        clear_view = request.GET.get("clear_view", False)
```

nautobot/core/views/mixins.py
```diff
@@ -688,13 +690,43 @@ def list(self, request, *args, **kwargs):
...
+
+            # Check if there is a global default for this view
+            global_saved_view = None
+            try:
+                global_saved_view = SavedView.objects.get(view=view_name, is_global_default=True)
+            except ObjectDoesNotExist:
+                pass
+
+            if global_saved_view is not None:
+                return redirect(reverse("users:savedview", kwargs={"pk": global_saved_view.pk}))
```
gsnider2195, line 728:
```suggestion
            try:
                global_saved_view = SavedView.objects.get(view=view_name, is_global_default=True)
                return redirect(reverse("users:savedview", kwargs={"pk": global_saved_view.pk}))
            except ObjectDoesNotExist:
                pass
```

nautobot/users/views.py
HanlinMiao: This extra message function is for setting `is_global_default` successfully from the SavedView management view.
lines 466-474 were originally for setting `is_global_default` from the create saved view modal. Since we lost the functionality, lines 466-474 are no longer necessary.

nautobot/core/views/generic.py
HanlinMiao: Good catch I think that is reasonable.

nautobot/core/views/mixins.py
HanlinMiao: I am running into formatting issues when I try to call `messages.success()` directly in the function.

nautobot/users/views.py
HanlinMiao: Is this what you are thinking?
```
if user.has_perms(["users.view_savedview"]):
            saved_views = queryset.restrict(user, "view")
else:
            ....
```

glennmatthews: I think so?

nautobot/docs/user-guide/core-data-model/users/savedview.md
```diff
@@ -0,0 +1,46 @@
+# Saved Views
+
++++ 2.3.0
+
+Saved Views give users the ability to save multiple configurations of list views, including table columns, filtering, pagination, and sorting, for ease of later use and reuse. Check out this [user guide](../../feature-guides/saved-views.md) on how to use saved views.
```
glennmatthews, line 5: This is another case where I question why we have separate "model" and "user guide" documentation files. Why not combine them?

nautobot/docs/user-guide/core-data-model/users/savedview.md
```diff
@@ -0,0 +1,46 @@
...
+
+The essential attributes for a Saved View are `owner`, `name`, `view`, `is_shared` and `is_global_default`. The `owner` attribute links to the user account active during the creation of the Saved View. The `view` attribute refers to the specific list view from which the Saved View is derived, such as `dcim:device_list` or `circuits:circuit_list`.
+The `is_shared` attribute dictates whether the saved view will be public or private. The `is_global_default` attribute dictates whether all users trying to access the object list view indicated by the value of the `view` attribute will be redirected to the current saved view or not.
```
glennmatthews, line 8: It occurs to me (somewhat belatedly) that `is_global_default` is a proper database field, but per-user defaults are stored in the user `config_data` JSON, that is to say they are denormalized and are not actual foreign keys. I wonder whether it would be better design wise if we had a proper many-to-many mapping between each user and their per-user default views. It'd be more self-consistent, and would let us use attributes like `on_delete=CASCADE` so that a user would never have "stale" saved view defaults. What do you think?

HanlinMiao: Yeah that does sound like a better design, I will refactor it.

HanlinMiao: Okay, one thing I noticed immediately when I tried to refactor this feature is that we need extra validation to make sure that there are no saved views with the same `view` attribute mapped to the same user, which was not needed in the original implementation.

glennmatthews: Should be doable with a `unique_together` constraint on the through table, but agreed.

HanlinMiao: Hmmm I wonder if `unique_together` supports indirect fields. Something like`unique_together = [("user", "saved_view__name"),]`. Maybe we can store `"saved_view__name"` in a direct field of the through table like `view_name`

nautobot/docs/user-guide/core-data-model/users/savedview.md
```diff
@@ -2,7 +2,7 @@
 
 +++ 2.3.0
```
glennmatthews, line 3: Oh gosh, we had *three* different docs for saved views? :-) Getting rid of the user guide was a good start but IMHO saved views are not a part of the "core data model" and shouldn't exist there either. Can we get rid of this file too?

nautobot/users/api/urls.py
```diff
@@ -10,6 +10,8 @@
 
 # Saved Views
 router.register("saved-views", views.SavedViewViewSet)
+router.register("user-to-savedviews", views.UserToSavedViewViewSet)
```
glennmatthews, line 13:
```suggestion
router.register("user-to-saved-views", views.UserToSavedViewViewSet)
```

Or should we follow other patterns and call this model and endpoint something like `UserSavedViewAssociation` instead?

nautobot/users/filters.py
```diff
@@ -113,6 +114,23 @@ class Meta:
...
+class UserToSavedViewFilterSet(NautobotFilterSet):
+    saved_view = NaturalKeyOrPKMultipleChoiceFilter(
+        queryset=SavedView.objects.all(),
+        to_field_name="name",
+        label="Saved View (ID or name)",
+    )
+    user = NaturalKeyOrPKMultipleChoiceFilter(
+        to_field_name="username",
+        queryset=get_user_model().objects.all(),
+        label="User (ID or name)",
```
glennmatthews, line 126:
```suggestion
        label="User (ID or username)",
```

nautobot/users/migrations/0012_usertosavedview_user_default_saved_views.py
```diff
@@ -0,0 +1,54 @@
+# Generated by Django 4.2.13 on 2024-06-19 22:39
```
glennmatthews, line 1: Can we roll this into the same migration file that creates the SavedView model?

nautobot/users/views.py
```diff
@@ -318,10 +318,12 @@ def set_default(self, request, *args, **kwargs):
...
         sv = SavedView.objects.get(pk=kwargs.get("pk", None))
-        if not user.config_data.get("saved_views", None):
-            user.config_data["saved_views"] = {}
-        user.config_data["saved_views"][f"{sv.view}"] = sv.pk
-        user.save()
+        try:
+            user_to_sv = UserToSavedView.objects.get(user=user, view_name=sv.view)
+            user_to_sv.delete()
+        except ObjectDoesNotExist:
+            pass
```
glennmatthews, line 325: Maybe?

```suggestion
        UserToSavedView.objects.filter(user=user, view_name=sv.view).delete()
```

nautobot/docs/user-guide/core-data-model/users/savedview.md
HanlinMiao: Needed to change the `get_docs_url` function we have. I only added `savedview.md` in `core-data-model` because `get_docs_url` is expecting it there.

glennmatthews:
```
    if hasattr(model, "documentation_static_path"):
        path = model.documentation_static_path
```

You should be able to define a `documentation_static_path` on the model and point it to the platform-functionality documentation.

nautobot/docs/user-guide/feature-guides/images/saved-views/clear-view-button.png
glennmatthews, line 1: I'm not a huge fan of having these in `user-guide/feature-guides/images` now that this isn't a "feature-guide" but I'm not sure `media/models` would be any better - thoughts?

nautobot/docs/user-guide/feature-guides/images/saved-views/navigation-menu.png
glennmatthews, line 1: Would it make sense to crop this one to just focus on the nav menu?

nautobot/docs/user-guide/platform-functionality/savedview.md
```diff
@@ -2,11 +2,12 @@
...
+Saved Views give users the ability to save multiple configurations of list views, including table columns, filtering, pagination, and sorting, for ease of later use and reuse.
 
-The essential attributes for a Saved View are `owner`, `name`, and `view`. The `owner` attribute links to the user account active during the creation of the Saved View. The `view` attribute refers to the specific list view from which the Saved View is derived, such as `dcim:device_list` or `circuits:circuit_list`.
+The essential attributes for a Saved View are `owner`, `name`, `view`, `is_shared` and `is_global_default`. The `owner` attribute links to the user account active during the creation of the Saved View. The `view` attribute refers to the specific list view from which the Saved View is derived, such as `dcim:device_list` or `circuits:circuit_list`.
+The `is_shared` attribute dictates whether the saved view will be public or private. The `is_global_default` attribute dictates whether all users trying to access the object list view indicated by the value of the `view` attribute will be redirected to the current saved view or not.
```
glennmatthews, line 8:
```suggestion
The `is_shared` attribute dictates whether a Saved View will be made available for other users to select. The `is_global_default` attribute (which may be set on a single Saved View per list view) makes this Saved View the default for all users of the system unless they specifically set their own default view (more on this below).
```

nautobot/docs/user-guide/feature-guides/images/saved-views/save-view-modal.png
glennmatthews, line 1: Should we add a checkbox on this modal/form to say "Save this as my default for this view"? So that creating a view and saving it as default can be done in a single operation?

nautobot/docs/user-guide/platform-functionality/savedview.md
```diff
@@ -43,3 +44,123 @@ The `config` attribute holds the configuration of the list view, stored as a `di
     }
 }
 ```
+
+Below is a user guide on how to create, update, save and use saved views in Nautobot.
```
glennmatthews, line 48: Nit - I don't think we have anything written so formal as a "documentation style guide" but personally I like to use Title Case (`Saved Views`) any time we're talking about a data model.

nautobot/docs/user-guide/platform-functionality/savedview.md
```diff
@@ -43,3 +44,123 @@ The `config` attribute holds the configuration of the list view, stored as a `di
...
+
+### How to create a new Saved View
+
+Navigate to any saved-view-supported model's list view. (We will use Location List View for the purpose of this guide)
+
+Note that there is a "Saved Views" dropdown button in the group of buttons on the right hand side.
+
+![Location List View](../feature-guides/images/saved-views/default-location-list-view.png)
+
+Configure the list view by clicking on the "Filter" and "Configure" buttons on the right hand side.
```
glennmatthews, line 64:
```suggestion
Configure the list view to your liking by clicking on the "Filter" and/or "Configure" buttons on the right hand side.
```

Should we also mention sorting the table and/or changing the pagination?

nautobot/docs/user-guide/platform-functionality/savedview.md
```diff
@@ -43,3 +44,123 @@ The `config` attribute holds the configuration of the list view, stored as a `di
...
+
+![Configure Table Columns to Location List View](../feature-guides/images/saved-views/config-table-columns-to-locations.png)
+
+Once you are satisfied with the current list view configurations, click on the "Saved Views" dropdown button. There should be an option in the dropdown menu named "Save As New View".
+
+![Save As New View Dropdown](../feature-guides/images/saved-views/save-as-new-view-drop-down.png)
+
+Click the "Save As New View" option, a modal should appear to prompt you to give a name to your new saved view and decide whether to make this saved view available to other users.
+
+We will name this new saved view "Campus Staging Location List View" and make this view public.
```
glennmatthews, line 80:
```suggestion
In this example, we will name this new saved view "Campus Staging Location List View" and make this view public (shared with other users).
```

nautobot/docs/user-guide/platform-functionality/savedview.md
```diff
@@ -43,3 +44,123 @@ The `config` attribute holds the configuration of the list view, stored as a `di
...
+
+### How to update an existing Saved View
+
+Stay on the same page where we just created the new Saved View and make some modifications to the view.
+
+Note that the title font of the saved view became italicized. This is an indicator that this saved view currently has unsaved changes.
+
+![Unsaved saved view](../feature-guides/images/saved-views/unsaved-saved-view.png)
+
+Once you have made desired changes, click the "Saved Views" dropdown button and you should see the option "Update Current Saved View".
```
glennmatthews, line 100:
```suggestion
Click the "Saved Views" dropdown button and you should now see the option "Update Current Saved View".
```

nautobot/docs/user-guide/platform-functionality/savedview.md
```diff
@@ -43,3 +44,123 @@ The `config` attribute holds the configuration of the list view, stored as a `di
...
+
+Check the box for `is_global_default` and click on the Update button.
+
+You should be redirected to the SavedViewListView with a success message. Now all users without a default view of their own will be directed to "Campus Staing Location List View" when they are trying to access location list view.
+
+![Saved View Edit Success](../feature-guides/images/saved-views/saved-view-admin-edit-success.png)
+
+### How to hide a saved view from all other users but yourself
+
+In order to create a saved view only for yourself to see, you should make sure that you uncheck the box for `is_shared` when you are creating a new saved view. This will make sure that you are the only who is able to see this saved view.
```
glennmatthews, line 138: Thinking about UX, we should make sure that `is_shared` is *off* by default, as most saved views will probably be user-specific?

nautobot/docs/user-guide/platform-functionality/savedview.md
```diff
@@ -43,3 +44,123 @@ The `config` attribute holds the configuration of the list view, stored as a `di
...
+
+There are several ways you can navigate saved views. All saved views created by any user for any list view can be accessed from the navigation menu on the left hand side under "Extensibility -> Saved Views" as an admin or user with "users:view_savedview" permission:
+
+![Navigation Menu](../feature-guides/images/saved-views/navigation-menu.png)
+
+Another way is from the "Saved View" dropdown button on any given model list view. All saved views defined for this specific model list view are populated in the dropdown menu and the saved view the user is currently viewing will be highlighted in bold.
+
+![Location List View with Saved Views](../feature-guides/images/saved-views/location-list-view-with-saved-views.png)
+
+![Current Saved View](../feature-guides/images/saved-views/current-saved-view-drop-down-menu.png)
```
glennmatthews, line 118: Let's mention the more common case first:

```suggestion
There are several ways you can navigate saved views. As we've already seen, any given model list view has a "Saved Views" button, which opens a dropdown showing all Saved Views that you have access to for this view (those that you have created, and those that other users have created and shared). The Saved View that you're currently using will be highlighted in bold.

![Location List View with Saved Views](../feature-guides/images/saved-views/location-list-view-with-saved-views.png)

![Current Saved View](../feature-guides/images/saved-views/current-saved-view-drop-down-menu.png)

Additionally, if you have the `users:view_savedview` object permission (or are a superuser), you can view a table of all saved views created by any user for any list view, accessible from the navigation menu under "Extensibility -> Saved Views":

![Navigation Menu](../feature-guides/images/saved-views/navigation-menu.png)
```

nautobot/docs/user-guide/platform-functionality/savedview.md
```diff
@@ -43,3 +44,123 @@ The `config` attribute holds the configuration of the list view, stored as a `di
...
+
+### How to set a saved view as global default for all users
+
+There can only be one global default for each object list view and the only way to set a saved view as the global default for all users is from the SavedViewTable at SavedViewListView that is only accessible to admins and users with "users:view_savedview" and "users:edit_savedview" permissions.
```
glennmatthews, line 122:
```suggestion
Users with the `users:view_savedview` and `users:change_savedview` object permissions can set a saved view as the default for all users for any given object list view. There can only be one global default for each object list view.
```

nautobot/docs/user-guide/platform-functionality/savedview.md
```diff
@@ -43,3 +44,123 @@ The `config` attribute holds the configuration of the list view, stored as a `di
...
+
+![Saved View Table Edit Button](../feature-guides/images/saved-views/saved-view-admin-edit-buttons.png)
+
+Clicking on the edit button will direct you to a standard edit view where you can edit the `name`, `is_shared` and `is_global_default` attributes of the current saved view.
+
+![Saved View Edit View](../feature-guides/images/saved-views/saved-view-admin-edit-view.png)
+
+Check the box for `is_global_default` and click on the Update button.
+
+You should be redirected to the SavedViewListView with a success message. Now all users without a default view of their own will be directed to "Campus Staing Location List View" when they are trying to access location list view.
```
glennmatthews, line 132:
```suggestion
You should be redirected to the Saved View list view with a success message. Now all users without a default view of their own will be directed to the selected Saved View when they access the given list view.
```

nautobot/docs/user-guide/platform-functionality/savedview.md
```diff
@@ -43,3 +44,123 @@ The `config` attribute holds the configuration of the list view, stored as a `di
...
+
+![Saved View Edit Success](../feature-guides/images/saved-views/saved-view-admin-edit-success.png)
+
+### How to hide a saved view from all other users but yourself
+
+In order to create a saved view only for yourself to see, you should make sure that you uncheck the box for `is_shared` when you are creating a new saved view. This will make sure that you are the only who is able to see this saved view.
+
+![Saved View Modal Unchecked](../feature-guides/images/saved-views/saved-view-modal-unchecked.png)
+
+As an admin or a user with the appropriate permissions, you can also edit the `is_shared` attribute from Saved View Edit View. Make sure to uncheck both the `is_shared` box and the `is_global_default` box as a global default view will be automatically made as public to all users.
```
glennmatthews, line 142: Do we need a little JS so that unchecking `is_shared` automatically unchecks `is_global_default` and checking `is_global_default` automatically checks `is_shared`? Not blocking.

nautobot/docs/user-guide/platform-functionality/savedview.md
```diff
@@ -43,3 +44,123 @@ The `config` attribute holds the configuration of the list view, stored as a `di
...
+
+![Saved View Edit View](../feature-guides/images/saved-views/saved-view-admin-edit-view.png)
+
+Check the box for `is_global_default` and click on the Update button.
+
+You should be redirected to the SavedViewListView with a success message. Now all users without a default view of their own will be directed to "Campus Staing Location List View" when they are trying to access location list view.
+
+![Saved View Edit Success](../feature-guides/images/saved-views/saved-view-admin-edit-success.png)
+
+### How to hide a saved view from all other users but yourself
```
glennmatthews, line 136: I'd like to invert this documentation (see also comment below) and make it so that we default to creating a non-shared saved view and this section describes how to share it with other users. Let's also avoid the term "private" here as non-shared saved views are not truly private - it's a UI feature, not a security feature.

nautobot/docs/user-guide/platform-functionality/savedview.md
```diff
@@ -43,3 +44,123 @@ The `config` attribute holds the configuration of the list view, stored as a `di
...
+
+![Privated Saved View Drop Down](../feature-guides/images/saved-views/saved-view-different-user.png)
+
+### How to set a user-specific default saved view
+
+You can also set a saved view as default for an object list view for yourself, overriding the global default. You can set a view as default by navigate to the view from the saved view dropdown, for example, "Elevator Only Location List View". Clicking on the saved views dropdown again, you should be able to see the option named "Set As My Default".
+
+![Set As My Default Button](../feature-guides/images/saved-views/set-as-my-default-button.png)
+
+Clicking on that option will set the current saved view as your default view for this object list view. The page should refresh with a success message. Now whenever this user is trying to access location list view, he/she will be redirected to "Elevator Only Location List View".
```
glennmatthews, line 156:
```suggestion
Clicking on that option will set the current Saved View as your default view for this object list view. The page should refresh with a success message. Now whenever you access the Location list view, this Saved View will automatically be applied to it.
```

nautobot/docs/user-guide/platform-functionality/savedview.md
```diff
@@ -43,3 +44,123 @@ The `config` attribute holds the configuration of the list view, stored as a `di
...
+
+![Set As My Default Success](../feature-guides/images/saved-views/set-as-my-default-success.png)
+
+### How to clear configurations on a Saved View
+
+If the user is currently viewing a saved view, the current saved view should have a checkmark next to its name in the dropdown menu. Clicking on button with the checkmark (current saved view) will direct you back to the default object list view with no configurations applied.
```
glennmatthews, line 162:
```suggestion
The current Saved View will have a checkmark next to its name in the dropdown menu. Clicking this menu item (deselecting the current Saved View) will direct you back to the default object list view with no Saved View applied.
```

nautobot/users/factory.py
```diff
@@ -46,3 +46,5 @@ class Meta:
     ct = random_instance(lambda: ContentType.objects.filter(FeatureQuery("saved_views").get_query()), allow_null=False)
     view = factory.LazyAttribute(lambda o: f"{o.ct.app_label}:{o.ct.model}_list")
     config = factory.Faker("pydict")
+    is_shared = NautobotBoolIterator()
+    # is_global_default currently just defaults to False for all randomly generated saved views
```
glennmatthews, line 50: yeah... creating some global-default saved views in the factory would probably break a bunch of our view tests :-)

glennmatthews (approved):
Bunch of docs nits but this looks good otherwise!

nautobot/docs/user-guide/platform-functionality/savedview.md
HanlinMiao: Sorry, it might be because of the wording here but `is_shared` is off by default.

nautobot/docs/user-guide/platform-functionality/savedview.md
```diff
@@ -2,11 +2,12 @@
 
 +++ 2.3.0
 
-Saved Views give users the ability to save multiple configurations of list views, including table columns, filtering, pagination, and sorting, for ease of later use and reuse. Check out this [user guide](../feature-guides/saved-views.md) on how to use saved views.
+Saved Views give users the ability to save multiple configurations of list views, including table columns, filtering, pagination, and sorting, for ease of later use and reuse.
 
-The essential attributes for a Saved View are `owner`, `name`, and `view`. The `owner` attribute links to the user account active during the creation of the Saved View. The `view` attribute refers to the specific list view from which the Saved View is derived, such as `dcim:device_list` or `circuits:circuit_list`.
+The essential attributes for a Saved View are `owner`, `name`, `view`, `is_shared` and `is_global_default`. The `owner` attribute links to the user account active during the creation of the Saved View. The `view` attribute refers to the specific list view from which the Saved View is derived, such as `dcim:device_list` or `circuits:circuit_list`.
```
gsnider2195, line 7:
```suggestion
The essential attributes for a Saved View are `owner`, `name`, `view`, `is_shared` and `is_global_default`. The `owner` attribute links to the user account that created the Saved View. The `view` attribute refers to the specific list view from which the Saved View is derived, such as `dcim:device_list` or `circuits:circuit_list`.
```

nautobot/docs/user-guide/feature-guides/images/saved-views/clear-view-button.png
HanlinMiao: Yeah, it would have made sense when the saved view user guide and saved view model documentation were separate. But right now, I think there is no better place to put it. IMHO we should just leave them where they are and figure out a better structure for our doc in general later on.

nautobot/docs/user-guide/platform-functionality/savedview.md
HanlinMiao: Hmmm, I am of the opinion that most users would want to share their saved views with other users. And making a non-shared saved view would be a rare case IMO and it would make sense to come later in the documentation. I do agree I should avoid using the term "private".

nautobot/docs/user-guide/feature-guides/images/saved-views/save-view-modal.png
HanlinMiao: Seeing this is not a required/urgent feature, I am going to open up a follow-up issue. Because this is also gonna require a bunch of screenshots and doc updates.

nautobot/docs/user-guide/platform-functionality/savedview.md
HanlinMiao: Seeing this is not a required/urgent feature, I am going to open up a follow-up issue.

nautobot/users/views.py
```diff
@@ -242,17 +243,34 @@ class SavedViewUIViewSet(
...
     filterset_class = SavedViewFilterSet
     serializer_class = SavedViewSerializer
     table_class = SavedViewTable
     action_buttons = ("export",)
 
+    def alter_queryset(self, request):
+        """
+        Two scenarios we need to handle here:
+        1. User can view all saved views with users.view_savedview permission.
+        2. User without the permission can only view published savedviews and his/her own saved views.
```
gsnider2195, line 256:
```suggestion
        2. User without the permission can only view shared savedviews and his/her own saved views.
```

nautobot/docs/user-guide/feature-guides/images/saved-views/save-view-modal.png
HanlinMiao: https://github.com/nautobot/nautobot/issues/5827

nautobot/docs/user-guide/platform-functionality/savedview.md
HanlinMiao: https://github.com/nautobot/nautobot/issues/5827

nautobot/core/views/mixins.py
HanlinMiao: There could be cases where you would override `extra_message` and not `extra_message_context`. I do not think so since all old models are okay with the default message. And we can reasonably assume that new models implemented in the future will be using `NautobotUIViewSet` and `ObjectEditViewMixin`.

Issue #802, opened 2021-08-11
Multiple errors in template web pages


FloLaco:
### Environment
* Python version: 3.6.14
* Nautobot version: 1.1.2

### Steps to Reproduce
1. Activate debug verbose level for django, nautobot, rq.worker (via `nautobot_config.py`)
2. Go to the home page (http://mydomain.com:8080/) in webui (but I saw same errors on `extra/jobs.html` too)
3. show logs 

### Expected Behavior
No error or traceback log

### Observed Behavior
```
09:26:01.410 DEBUG   nautobot.releases    releases.py                 get_latest_release() :
  Skipping release check; RELEASE_CHECK_URL not defined
09:26:01.672 DEBUG   django.template      base.py                        _resolve_lookup() :
  Exception while resolving variable 'graphiql' in template 'home.html'.
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/template/base.py", line 829, in _resolve_lookup
    current = current[bit]
  File "/usr/local/lib/python3.6/site-packages/django/template/context.py", line 83, in __getitem__
    raise KeyError(key)
KeyError: 'graphiql'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/template/base.py", line 835, in _resolve_lookup
    if isinstance(current, BaseContext) and getattr(type(current), bit):
AttributeError: type object 'RequestContext' has no attribute 'graphiql'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/template/base.py", line 843, in _resolve_lookup
    current = current[int(bit)]
ValueError: invalid literal for int() with base 10: 'graphiql'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/template/base.py", line 850, in _resolve_lookup
    (bit, current))  # missing attribute
[...]

PR #5810, opened 2024-06-13
bugfix: removing object.slug as this does not exist any longer, and g…

jnmms:
When showing platforms, the console throws an error because of a reference to "object.slug" but no models in nautobot uses slug anymore, so it has been removed

# Closes NTC-1734 #<ISSUE NUMBER GOES HERE>
# What's Changed
- Removed an if statement that fails due to missing object, and all the contents within the if

### DIFF
--- a/nautobot/core/templates/inc/object_details_advanced_panel.html
+++ b/nautobot/core/templates/inc/object_details_advanced_panel.html
@@ -31,19 +31,6 @@
                         </td>
                     </tr>
                 {% endif %}
-                {% if object.slug %}
-                    <tr>
-                        <td>Slug</td>
-                        <td>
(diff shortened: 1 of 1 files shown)


glennmatthews:
Thanks for the PR! I know we have various error cases in the templates (see #802) but since they're generally ignored by default and only generate debug logs we haven't prioritized fixing them to date.

My only concern with this change is whether it might represent a loss in functionality for any App models that still have slugs for whatever reason.

glennmatthews:
Thanks again for the PR. We discussed this at a team and decided not to more forward with this change at this time as there may still be Apps that are taking advantage of this functionality, and it isn't causing any functional problems in the meantime for other models. We appreciate your contribution though!

PR #5825, opened 2024-06-18
Auto-disable JobButton/JobHook/ScheduledJob when a Job is uninstalled, and other fixes

### LINKED ISSUE #5550: Jobs from Git Repository not being registered
### Environment
* Nautobot version (Docker tag too if applicable): v2.1.9
* Python version: 3.11.8
* Database platform, version: Postgres 15.5
* Middleware(s): N/A

### Steps to Reproduce
1. Define a job that has been tested in $JOBS_ROOT and move it to a Git Repository under the root folder /jobs/
2. Define __init__.py files and import the relevant job `from . import example_job`
3. Sync Git repository containing jobs and validate job has been modelled.
4. Trigger job hook (from model update)

### Expected Behavior
The job is created/updated in Nautobot and available (installed). When an object is modified that triggers the related Job Hook, the job is executed. 

### Observed Behavior
The job is successfully created and installed and in the UI we can see the Module, Job Class and Class Path are all present, however when the job is executed, a traceback is raised stating the job has not been registered.

image

image

Additionally, running the post_upgrade CLI command removes the job from the list of installed jobs.

glennmatthews:
# Closes #5550 
# What's Changed

- Add `JobButton.enabled` flag, similar to what `JobHook` and `ScheduledJob` already have. When set to False, this JobButton will not be rendered.
- When a Job changes from `installed=True` to `installed=False`, update any associated JobButton, JobHook, and ScheduledJob records to `enabled=False`. (This should address the symptoms reported in #5550)
- As additional safety, when enqueuing JobHooks, in addition to skipping any that are already `enabled=False` (existing logic), additionally do not enqueue any for which the associated Job record is `enabled=False` or `installed=False` or for which the underlying Job class cannot be found.
- Add support in our GitRepository model and helper classes to specify a `branch` that is actually a tag or a commit hash.
- Refactor `test_datasources.py` to create (via new `git_helper.py` utility and files in `git_data/`) and use a temporary but "real" Git repository as the "remote", instead of mocking out all calls to `GitRepo.checkout` to fake Git operations.
    - This SHALL NOT be taken as support for `file://` strings for `remote_url`, _except_ in this narrow testing scenario; form and model validation on `GitRepository` will normally still enforce either `http://` or `https://` only.
- Augment `test_datasources.py` to test the fix for #5550.
- Fix a couple of bugs that I encountered along the way:
    - When resyncing a repository and the new repo contents *completely lack* a `config_contexts` or `export_templates` directory, faulty logic flow meant we wouldn't delete such records previously sourced from the repository
    - When resyncing a repository with *substantial* changes from the last sync, the Git CLI output could include lines with multiple tabs such as `R100        file1.py        file2.py`, which would cause an exception to be thrown by `swap_status_initials` incorrectly assuming that a single tab would be the only one present.

# TODO

- [x] Unit, Integration Tests


nautobot/core/utils/git.py
```diff
@@ -34,8 +35,8 @@
 
 def swap_status_initials(data):
     """Swap Git status initials with its equivalent."""
-    initial, text = data.split("\t")
-    return GitDiffLog(status=GIT_STATUS_MAP.get(initial), text=text)
+    initial, text = data.split("\t", 1)
+    return GitDiffLog(status=GIT_STATUS_MAP.get(initial[0]), text=text)
```
glennmatthews, line 39: GIT_STATUS_MAP uses only single characters (`"R"`) but in some cases `initial` could also include a confidence percentage, e.g. `"R100"`

nautobot/extras/datasources/git.py
```diff
@@ -224,126 +224,127 @@ def refresh_git_config_contexts(repository_record, job_result, delete=False):
 def update_git_config_contexts(repository_record, job_result):
     """Refresh any config contexts provided by this Git repository."""
     config_context_path = os.path.join(repository_record.filesystem_path, "config_contexts")
-    if not os.path.isdir(config_context_path):
-        return
-
     managed_config_contexts = set()
     managed_local_config_contexts = defaultdict(set)
 
-    # First, handle the "flat file" case - data files in the root config_context_path,
```
glennmatthews, line 233: The diffs here are messy but it's mostly just an indentation change, from:

```
if not os.path.isdir(config_context_path):
    return
# add/update contexts
# remove stale contexts - whoops, we need to do this even if it isn't a directory!
```

to

```
if os.path.isdir(config_context_path):
    # add/update contexts
# remove stale contexts regardless
```

nautobot/extras/datasources/git.py
```diff
@@ -599,39 +600,38 @@ def refresh_git_config_context_schemas(repository_record, job_result, delete=Fal
 def update_git_config_context_schemas(repository_record, job_result):
     """Refresh any config context schemas provided by this Git repository."""
     config_context_schema_path = os.path.join(repository_record.filesystem_path, "config_context_schemas")
-    if not os.path.isdir(config_context_schema_path):
-        return
 
     managed_config_context_schemas = set()
 
-    for file_name in os.listdir(config_context_schema_path):
```
glennmatthews, line 607: Same indentation change here resulting in messy diffs.

nautobot/extras/forms/forms.py
```diff
@@ -1013,6 +1013,10 @@ class JobHookForm(BootstrapMixin, forms.ModelForm):
     content_types = MultipleContentTypeField(
         queryset=ChangeLoggedModelsQuery().as_queryset(), required=True, label="Content Type(s)"
     )
+    job = DynamicModelChoiceField(
+        queryset=Job.objects.filter(is_job_hook_receiver=True),
+        query_params={"is_job_hook_receiver": True},
+    )
```
glennmatthews, line 1019: We were missing this (and the similar addition below) which resulted in the form using the default (server-side-rendered) form field instead.

nautobot/extras/tests/test_git.py
glennmatthews, line 1: This was broken by the changes in `diff_remote()`, and IMHO its functionality is covered by the `test_git_dry_run` test case in test_datasources.py.

nautobot/extras/forms/forms.py
```diff
@@ -1013,6 +1013,10 @@ class JobHookForm(BootstrapMixin, forms.ModelForm):
     content_types = MultipleContentTypeField(
         queryset=ChangeLoggedModelsQuery().as_queryset(), required=True, label="Content Type(s)"
     )
+    job = DynamicModelChoiceField(
+        queryset=Job.objects.filter(is_job_hook_receiver=True),
```
HanlinMiao, line 1017: This is the only case I have seen that we needed `.filter` on the `queryset=` argument. I think the `query_params` is sufficient to filter the choices for this field. Do double check me

nautobot/extras/forms/forms.py
```diff
@@ -1180,14 +1184,19 @@ class JobButtonForm(BootstrapMixin, forms.ModelForm):
             api_url="/api/extras/content-types/",
         ),
     )
+    job = DynamicModelChoiceField(
+        queryset=Job.objects.filter(is_job_button_receiver=True),
```
HanlinMiao, line 1188: Same comment as above

nautobot/extras/models/jobs.py
```diff
@@ -363,6 +382,9 @@ def clean(self):
         if not self.type_create and not self.type_delete and not self.type_update:
             raise ValidationError("You must select at least one type: create, update, and/or delete.")
 
+        if self.enabled and not (self.job.installed and self.job.enabled):
+            raise ValidationError({"enabled": "The selected Job is not installed and enabled"})
```
HanlinMiao, line 386: Would this be:
```suggestion
            raise ValidationError({"enabled": "The selected Job is not installed or not enabled"})
```

nautobot/extras/models/jobs.py
```diff
@@ -817,6 +841,12 @@ class Meta:
     def __str__(self):
         return self.name
 
+    def clean(self):
+        super().clean()
+
+        if self.enabled and not (self.job.installed and self.job.enabled):
+            raise ValidationError({"enabled": "The selected Job is not installed and enabled"})
```
HanlinMiao, line 848: Same comment as above

nautobot/extras/models/jobs.py
glennmatthews: I'm fine either way on this one.

HanlinMiao (approved):

nautobot/core/utils/git.py
```diff
@@ -93,69 +94,140 @@ def checkout(self, branch, commit_hexsha=None):
...
+        If `commit_hexsha` is specified and `branch` is either a tag or a commit identifier, they must match.
+        If `commit_hexsha` is specified and `branch` is a branch name, it must contain the specified commit.
+
         Returns:
-            (str, bool): commit_hexsha the repo contains now, whether any change occurred
+            (str, bool): commit_hexsha the repo contains now, and whether any change occurred
         """
         # Short-circuit logic - do we already have this commit checked out?
-        if commit_hexsha and commit_hexsha == self.head:
+        if commit_hexsha and self.head.startswith(commit_hexsha):
```
gsnider2195, line 108: Might be a good idea to write a utility function compare possible git hexsha strings:
```py
def compare_git_hexsha(str1, str2):
    if not set(str1).issubset(string.hexdigits) or not set(str1).issubset(string.hexdigits):
        return False
    if str1 == str2:
        return True
    if len(str1) > 6 and str2.startswith(str1):
        return True
    if len(str2) > 6 and str1.startswith(str2):
        return True
    return False
```

gsnider2195: https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#_short_sha_1

nautobot/core/utils/git.py
```diff
@@ -93,69 +94,140 @@ def checkout(self, branch, commit_hexsha=None):
...
+                        f" If this is a new repo, please add a commit before syncing. {git_error}"
+                    )
+
+            logger.info("Checking out latest commit on branch %s...", branch)
+            branch_head.checkout()
+            # No specific commit hash was given, so make sure we get the latest from origin
+            # We would use repo.remotes.origin.pull() here, but that will fail in the case where someone has
+            # force-pushed to the upstream repo since the last time we did a pull. To be safe, we reset instead.
+            self.repo.head.reset(f"origin/{branch}", index=True, working_tree=True)
+            logger.info(f"Latest commit on branch `{branch}` is `{self.head}`")
```
gsnider2195, line 168:
```suggestion
            logger.info("Latest commit on branch `%s` is `%s`", branch, self.head)
```

nautobot/core/utils/git.py
```diff
@@ -93,69 +94,140 @@ def checkout(self, branch, commit_hexsha=None):
...
-        try:
-            self.repo.remotes.origin.refs[branch]
-        except IndexError as git_error:
-            logger.error(
-                "Branch %s does not exist at %s. %s", branch, next(iter(self.repo.remotes.origin.urls)), git_error
-            )
+        # Is `branch` actually a branch, a tag, or a commit? Heuristics:
+        is_branch = branch in self.repo.remotes.origin.refs
+        is_tag = branch in self.repo.tags
+        maybe_commit = set(branch).issubset(string.hexdigits)
```
gsnider2195, line 208: Good candidate for a helper function?

nautobot/core/utils/git.py
glennmatthews: `self.head` should always be the full non-abbreviated commit hash, so I think the current implementation is fine? Not sure what the more complex utility function would gain us here.

nautobot/extras/tests/test_datasources.py
```diff
@@ -341,8 +234,7 @@ def create_empty_repo(path, url):
                     (job_result.result, list(job_result.job_log_entries.values_list("message", "log_object"))),
                 )
                 self.repo.refresh_from_db()
-                self.assertEqual(self.repo.current_head, self.COMMIT_HEXSHA, job_result.result)
-                MockGitRepo.assert_called_with(os.path.join(tempdir, self.repo.slug), "http://localhost/git.git")
+                # self.assertEqual(self.repo.current_head, self.COMMIT_HEXSHA, job_result.result)
```
gsnider2195, line 237: No longer needed?

glennmatthews: yeah, since we're not mocking the repo creation now, we don't have a well-known sha to compare the head against. I suppose we could grab the sha from the other repo and compare it - think that'd be worthwhile?

gsnider2195: The job log entry lookup below seems to be sufficient for testing this. We can probably just remove this assertion

nautobot/extras/tests/git_helper.py
```diff
@@ -0,0 +1,51 @@
...
+from git import Repo
+
+logger = logging.getLogger(__name__)
+
+
+SOURCE_DIR = os.path.join(os.path.dirname(__file__), "git_data")
+
+
+def create_and_populate_git_repository(target_path):
+    """Main entry point to this script."""
```
gsnider2195, line 18: A short summary of what this is doing would be good here for when someone is trying to troubleshoot failing tests

gsnider2195 (approved):

PR #5829, opened 2024-06-20
MetadataType UI and docs

### LINKED ISSUE #5663: Metadata type model
Create a new model for representing the type of a metadata instance. This will be very similar in nature to custom fields types and in fact will likely share some of the same internals. Core provided types will include:

- DateTime
- Date
- Text
- Selection
- Multi selection
- JSON
- URL
- Integer
- Boolean
- Markdown

Other fields:
- name
- description

Note that datatime does not exist as a custom field type. We may add it as a part of this work.

For metadata specifically, we need to also support Contacts as a type.

glennmatthews:
# Closes #5663 
# What's Changed

Implement MetadataTypeUIViewSet, patterned closely after the (non-UIViewSet) UI for CustomField definition.

# Screenshots

image

image

image

image

# TODO

- [x] Unit, Integration Tests


nautobot/docs/user-guide/platform-functionality/metadata.md
```diff
@@ -1,3 +1,28 @@
...
+
+### The `MetadataChoice` model
+
+If you define a Metadata Type with a `data_type` of "selection" or "multiple selection", you must also define one or more `MetadataChoice` records to define the set of values that can be selected from. Each choice applies to a specific Metadata Type, has a distinct `value`, and has a `weight` that can be set to influence the ordering among all choices within a given type.
+
+### The `ObjectMetadata` model
+
+Once you have defined appropriate `MetadataType` and `MetadataChoice` records, you can then define actual object metadata making use of these.
+
 TODO
```
gsnider2195, line 28: Is there more documentation needed here?

glennmatthews: For the actual ObjectMetadata model once it's implemented, I expect so.

itdependsnetworks:
We already have a `metadata` sub menu under `Organization`, is this going to cause confusion?

glennmatthews:
> We already have a `metadata` sub menu under `Organization`, is this going to cause confusion?

Probably. Would be good to come up with a better name for the organization submenu but I don't think we've been able to agree on one yet.

nautobot/extras/forms/forms.py
```diff
@@ -1235,6 +1241,72 @@ class JobButtonFilterForm(BootstrapMixin, forms.Form):
...
+            "data_type",
+            "content_types",
+            "tags",
+        )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        if self.instance and self.instance.present_in_database:
+            self.fields["data_type"].widget.attrs["readonly"] = True
```
gsnider2195, line 1285: This doesn't seem to be working properly for me. The widget is greyed out but I can still click on it and change the value.

nautobot/extras/templates/extras/metadatatype_create.html
```diff
@@ -0,0 +1,91 @@
...
+    <div class="panel panel-default">
+        <div class="panel-heading"><strong>Assignment</strong></div>
+        <div class="panel-body">
+            {% render_field form.content_types %}
+        </div>
+    </div>
+    <div class="panel panel-default">
+        <div class="panel-heading"><strong>Choices</strong></div>
+        <div class="panel-body">
+            {% if choices.errors %}
```
gsnider2195, line 23: I'm seeing this show up when there is an error in a different field.

image

nautobot/extras/templates/extras/metadatatype_create.html
```diff
@@ -0,0 +1,91 @@
...
+        </div>
+    </div>
+    <div class="panel panel-default">
+        <div class="panel-heading"><strong>Assignment</strong></div>
+        <div class="panel-body">
+            {% render_field form.content_types %}
+        </div>
+    </div>
+    <div class="panel panel-default">
+        <div class="panel-heading"><strong>Choices</strong></div>
```
gsnider2195, line 21: It's not really clear that choices are only relevant for the "select" and "multiselect" data types. Perhaps this would help? I would prefer if this entire panel only showed up when those types are selected but may be out of scope for this issue.

```suggestion
        <div class="panel-heading"><strong>Choices (select and multiselect data types)</strong></div>
```

nautobot/extras/forms/forms.py
glennmatthews: Interesting. This was a copy/paste from the CustomField form, so it's probably an issue there too. I'll look into it.

glennmatthews:
> My vote would be to keep it under organization for now, and can re-think the whole thing in 3.0.

I'd personally prefer to keep this model under Extensibility as that's where it belongs. Tags/statuses/roles really should be considered "data" not "metadata", and grouping this model with them is IMO misleading as to both its nature and purpose.

nautobot/extras/templates/extras/metadatatype_create.html
glennmatthews: Another copy-paste from custom fields, will investigate.

glennmatthews: Fixed:

image

gsnider2195:
I don't know if this is worth chasing down but I'm seeing an incorrect error message under the choices when I try to create a metadata type with choices, then try to go back and resubmit the same form:

image

glennmatthews:
> I don't know if this is worth chasing down but I'm seeing an incorrect error message under the choices when I create a metadata type with choices, then try to go back and resubmit the same form:

I noticed that too. Somehow it's getting confused by the fact that the instance creation is failing due to a uniqueness violation, but then it's trying to attach the choices to the instance that it failed to create. Mind if we treat that as a lower priority followup item?

gsnider2195:
> > I don't know if this is worth chasing down but I'm seeing an incorrect error message under the choices when I create a metadata type with choices, then try to go back and resubmit the same form:
> 
> I noticed that too. Somehow it's getting confused by the fact that the instance creation is failing due to a uniqueness violation, but then it's trying to attach the choices to the instance that it failed to create. Mind if we treat that as a lower priority followup item?

Yeah I don't think that's a blocker

gsnider2195 (approved):

PR #5835, opened 2024-06-21
ObjectMetadata Model and API

### LINKED ISSUE #5664: ObjectMetadata Model
Using a MetadataType allow of the association of the relevant data value to a field or set of fields on an object instance, or to the entire object.

We will implement the ability to view this data in the object detail views and in a new list view for metadata, but there will be no UI to manipulate metadata to begin with. The REST API and ORM will be the means to manage this metadata.

For the detail views, we will begin with a separate tab that shows a table of metadata objects filtered to the given model instance. In the future, we will enhance the UI to do tooltip, or hover, etc icons on the field values that the metadata is attached too.

Fields:
- FK to MetadataType
- JSON field for the actual value
- GFK to the assigned object
- Arrayfield[Charfield] for scoped fields (only direct fields on the model) (empty array implies all fields/entire model instance)
- FK to contact
- FK to team

BaseModel + ChangelogModel for now (more features later if they are warrented)

Unique together is MetadataType + assigned object + scoped fields. Likely will have be implemented in python.

Index on:
- GFK to the assigned object
- GFK to the assigned object + scoped fields
- GFK to the assigned object + contact
- GFK to the assigned object + team

HanlinMiao:
# Closes #5664 
# What's Changed

- [x] ObjectMetadata model
- [x] ObjectMetadata API
- [x] Unit tests

# Screenshots

# TODO

- [x] Unit, Integration Tests


nautobot/extras/filters/__init__.py
```diff
@@ -978,6 +979,19 @@ class Meta:
         fields = "__all__"
 
 
+class ObjectMetadataFilterSet(NautobotFilterSet):
+    q = SearchFilter(
+        filter_predicates={
+            "contact__name": "icontains",
+            "team__name": "icontains",
```
glennmatthews, line 986: maybe should include `value` and/or `metadata_type__name` here?

nautobot/extras/migrations/0109_metadatatype_metadatachoice.py
glennmatthews, line 1: Maybe rename to `0109_metadata.py` now?

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +138,70 @@ def save(self, *args, **kwargs):
...
+        null=True,
+    )
+    team = models.ForeignKey(
+        to=Team,
+        on_delete=models.SET_NULL,
+        related_name="object_metadatas",
+        null=True,
+    )
+    scoped_fields = JSONArrayField(
+        base_field=models.CharField(max_length=50),
```
glennmatthews, line 162: Maybe longer? We want to be able to include custom fields in the scoped fields and what's their max `key` length?

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +138,70 @@ def save(self, *args, **kwargs):
...
+        null=True,
+        help_text="Relevant data value to an object field or a set of object fields",
+    )
+    assigned_object_type = models.ForeignKey(to=ContentType, on_delete=models.SET_NULL, null=True, related_name="+")
+    assigned_object_id = models.UUIDField(db_index=True)
+    assigned_object = GenericForeignKey(ct_field="assigned_object_type", fk_field="assigned_object_id")
+
+    class Meta:
+        ordering = ["metadata_type"]
+        unique_together = ["metadata_type", "assigned_object_type", "assigned_object_id", "scoped_fields"]
```
glennmatthews, line 176: We'll want clean() to check for intersection of `scoped_fields`, i.e. it shouldn't be possible to have two object-metadata for the same type/assigned-object one with fields [a, b, c] and one with fields [c, d, e]. `unique_together` alone can't catch this case.

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +138,70 @@ def save(self, *args, **kwargs):
...
+                fields=["assigned_object_type", "assigned_object_id", "contact"],
+            ),
+            models.Index(
+                name="assigned_object_team",
+                fields=["assigned_object_type", "assigned_object_id", "team"],
+            ),
+        ]
+
+    def __str__(self):
+        return f"{self.metadata_type} - {self.assigned_object}"
```
glennmatthews, line 198: Probably should include value/contact/team in here?

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +138,70 @@ def save(self, *args, **kwargs):
     def delete(self, *args, **kwargs):
         # TODO: should we behave like CustomFieldChoice and raise a ProtectedError if this choice is in use anywhere?
         super().delete(*args, **kwargs)
+
+
+class ObjectMetadata(BaseModel, ChangeLoggedModel):
+    metadata_type = models.ForeignKey(
+        to=MetadataType,
+        on_delete=models.CASCADE,
```
glennmatthews, line 146: Wonder if this should maybe be PROTECT?

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +138,70 @@ def save(self, *args, **kwargs):
...
+    )
+    contact = models.ForeignKey(
+        to=Contact,
+        on_delete=models.SET_NULL,
+        related_name="object_metadatas",
+        null=True,
+    )
+    team = models.ForeignKey(
+        to=Team,
+        on_delete=models.SET_NULL,
```
glennmatthews, line 157: These should either be CASCADE or PROTECT as well, shouldn't they?

HanlinMiao: I think I am fine with either `SET_NULL` or `PROTECT`. I am not sure if deleting a contact should delete the `ObjectMetaData` instances as well.

HanlinMiao:
Should the `contact` and `team` fields be mutually exclusive?

nautobot/extras/models/metadata.py
glennmatthews: SET_NULL means that deleting a contact results in an ObjectMetadata of type Contact/Team that doesn't have a Contact or Team set - that should be invalid data.

glennmatthews:
> Should the `contact` and `team` fields be mutually exclusive?

IMO yes, similar to ContactAssociation.

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +138,108 @@ def save(self, *args, **kwargs):
...
+                name="assigned_object_team",
+                fields=["assigned_object_type", "assigned_object_id", "team"],
+            ),
+        ]
+
+    def __str__(self):
+        if self.contact:
+            return f"{self.contact}: {self.metadata_type} - {self.assigned_object} - {self.value}"
+        else:
+            return f"{self.team}: {self.metadata_type} - {self.assigned_object} - {self.value}"
```
glennmatthews, line 208: Contact and team are both specific types of metadata, so probably this should be something like:

```suggestion
        if self.contact:
            return f"{self.metadata_type} - {self.assigned_object} - {self.contact}"
        elif self.team:
            return f"{self.metadata_type} - {self.assigned_object} - {self.team}"
        else:
            return f"{self.metadata_type} - {self.assigned_object} - {self.value}"
```

I wonder if we should make the `value` database field into a private field (e.g. `_value`) and have `value` instead be an `@property` that returns either the contact, the team, or the `_value` as appropriate?

nautobot/core/management/commands/generate_test_data.py
```diff
@@ -292,6 +293,7 @@ def _create_batch(some_factory, count, description="", **kwargs):
         )
         _create_batch(MetadataChoiceFactory, 20)
         _create_batch(ObjectChangeFactory, 100)
+        _create_batch(ObjectMetadataFactory, 20)
```
glennmatthews, line 296: It'd probably be more "realistic" to create a larger number of these, maybe 100 for now?

nautobot/extras/api/serializers.py
```diff
@@ -766,6 +767,36 @@ class Meta:
...
+
+    class Meta:
+        model = ObjectMetadata
+        fields = "__all__"
+
+    @extend_schema_field(
+        PolymorphicProxySerializer(
+            component_name="ObjectMetadataAssignedObject",
+            resource_type_field_name="object_type",
+            serializers=lambda: nested_serializers_for_models(FeatureQuery("metadata").list_subclasses()),
```
glennmatthews, line 784: That's gotta be a pretty big set of serializers, isn't it? How bad is the impact on the schema?

nautobot/extras/api/serializers.py
```diff
@@ -766,6 +767,36 @@ class Meta:
...
+            allow_null=True,
+        )
+    )
+    def get_assigned_object(self, obj):
+        if obj.assigned_object is None:
+            return None
```
glennmatthews, line 790: Fine as error handling, but data-wise, shouldn't we be doing our best to prevent this from ever happening?

nautobot/extras/factory.py
```diff
@@ -238,6 +239,35 @@ def content_types(self, create, extracted, **kwargs):
...
+    assigned_object_type = random_instance(
+        ContentType.objects.filter(FeatureQuery("metadata").get_query()), allow_null=False
+    )
+
+    @factory.lazy_attribute
+    def assigned_object_id(self):
+        queryset = self.assigned_object_type.model_class().objects.all()
+        if queryset.exists():
+            return factory.random.randgen.choice(queryset).pk
+        return faker.Faker().uuid4()
```
glennmatthews, line 268: Consider changing the implementation here so that we always have a valid `assigned_object` GFK. This is different from something like an ObjectChange that can/should persist after its related object is deleted; imho metadata shouldn't persist after its assigned object is deleted, though maybe @lampwins might disagree?

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +138,108 @@ def save(self, *args, **kwargs):
...
+        help_text="List of scoped fields, only direct fields on the model",
+    )
+    value = models.JSONField(
+        blank=True,
+        null=True,
+        help_text="Relevant data value to an object field or a set of object fields",
+    )
+    assigned_object_type = models.ForeignKey(to=ContentType, on_delete=models.SET_NULL, null=True, related_name="+")
+    assigned_object_id = models.UUIDField(db_index=True)
+    assigned_object = GenericForeignKey(ct_field="assigned_object_type", fk_field="assigned_object_id")
```
glennmatthews, line 181: **if** we agree that deleting an object should also delete its attached metadata (which is my opinion), then we need to add a corresponding `GenericRelation` to `BaseModel` to make that happen automatically.

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +138,108 @@ def save(self, *args, **kwargs):
...
+            return f"{self.contact}: {self.metadata_type} - {self.assigned_object} - {self.value}"
+        else:
+            return f"{self.team}: {self.metadata_type} - {self.assigned_object} - {self.value}"
+
+    def clean(self):
+        super().clean()
+        if self.contact is None and self.team is None:
+            raise ValidationError("Either a contact or a team must be specified")
+        if self.contact is not None and self.team is not None:
+            raise ValidationError("A contact and a team cannot be both specified at once")
```
glennmatthews, line 215: This should depend on the `metadata_type` - if it is `TYPE_CONTACT_TEAM` then either team or contact but not both are mandatory, if it's any other type, then neither should be permitted. Conversely, `TYPE_CONTACT_TEAM` should *not* have a distinct `value`, but any other type must have one.

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +138,108 @@ def save(self, *args, **kwargs):
...
+
```
glennmatthews, line 216: Also, need to validate other types similar to custom fields, for example if `metadata_type` is `TYPE_SELECT` or `TYPE_MULTISELECT`, then we need to compare against the defined MetadataChoices for that type as part of the cleaning. If `TYPE_INTEGER` then we need to make sure `value` is an integer, etc.

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +138,108 @@ def save(self, *args, **kwargs):
...
+            )
+            .exclude(pk=self.pk)
+            .values_list("scoped_fields", flat=True)
+        )
+        duplicate_scoped_fields_list = set([])
+
+        for scoped_fields in object_metadata_scoped_fields:
+            for field in scoped_fields:
+                if field in self.scoped_fields:
+                    duplicate_scoped_fields_list.add(field)
```
glennmatthews, line 238:
```suggestion
            duplicate_scoped_fields_list |= set(self.scoped_fields).intersection(scoped_fields)
```

Or something like that at least.

nautobot/extras/tests/test_models.py
```diff
@@ -1319,6 +1322,55 @@ def get_absolute_url(self):
...
+    model = ObjectMetadata
+
+    def test_immutable_metadata_type(self):
+        instance1 = ObjectMetadata.objects.first()
+        instance2 = ObjectMetadata.objects.exclude(metadata_type=instance1.metadata_type).first()
+        self.assertIsNotNone(instance2)
+        with self.assertRaises(ValidationError):
+            instance1.metadata_type = instance2.metadata_type
+            instance1.validated_save()
+
```
glennmatthews, line 1335: Would want to see `value` validation tests here, similar to `test_customfields.py`.

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +138,108 @@ def save(self, *args, **kwargs):
     def delete(self, *args, **kwargs):
         # TODO: should we behave like CustomFieldChoice and raise a ProtectedError if this choice is in use anywhere?
         super().delete(*args, **kwargs)
+
```
glennmatthews, line 141: Do we need a custom Manager class for ObjectMetadata? It seems like at least we'd want something like `ObjectMetadata.objects.get_for_model(instance)` to be supported.

nautobot/extras/models/metadata.py
```diff
@@ -140,6 +143,23 @@ def delete(self, *args, **kwargs):
...
+    def get_for_model(self, model):
+        """Return all ObjectMetadatas assigned to the given model."""
+        concrete_model = model._meta.concrete_model
+        cache_key = f"{self.get_for_model.cache_key_prefix}.{concrete_model._meta.label_lower}"
+        queryset = cache.get(cache_key)
+        if queryset is None:
+            content_type = ContentType.objects.get_for_model(concrete_model)
+            queryset = self.get_queryset().filter(content_types=content_type)
+            cache.set(cache_key, queryset)
+        return queryset
```
glennmatthews, line 158: This gives us all of the metadata for a given model *class* (e.g. `Device`) - I'm not sure how useful that will be, I was thinking it'd be for a given *instance* (e.g. `device1`).

nautobot/extras/models/metadata.py
```diff
@@ -203,16 +223,121 @@ class Meta:
...
+                except ValueError as exc:
+                    raise ValidationError("Value must be true or false.") from exc
+
+            # Validate date
+            elif metadata_type_data_type == MetadataTypeDataTypeChoices.TYPE_DATE:
+                if not isinstance(value, date):
+                    try:
+                        datetime.strptime(value, "%Y-%m-%d")
+                    except ValueError:
+                        raise ValidationError("Date values must be in the format YYYY-MM-DD.")
```
glennmatthews: Also need to implement TYPE_DATETIME (which may be tricky, given timezone support) and TYPE_FLOAT since those aren't yet supported in CustomFields but are valid MetadataTypes.

nautobot/extras/api/serializers.py
HanlinMiao: Implemented

nautobot/extras/models/metadata.py
HanlinMiao: Implemented `value` helper function

nautobot/extras/models/metadata.py
HanlinMiao: Implemented

nautobot/extras/factory.py
```diff
@@ -238,6 +239,75 @@ def content_types(self, create, extracted, **kwargs):
...
+        elif metadata_type_data_type == MetadataTypeDataTypeChoices.TYPE_DATE:
+            return str(faker.Faker().date())
+        elif metadata_type_data_type == MetadataTypeDataTypeChoices.TYPE_DATETIME:
+            return str(faker.Faker().date_time_between_dates())
+        elif metadata_type_data_type == MetadataTypeDataTypeChoices.TYPE_SELECT:
+            return factory.random.randgen.choice(self.metadata_type.choices.values_list("value", flat=True))
+        elif metadata_type_data_type == MetadataTypeDataTypeChoices.TYPE_MULTISELECT:
+            return [factory.random.randgen.choice(self.metadata_type.choices.values_list("value", flat=True))]
+        elif metadata_type_data_type == MetadataTypeDataTypeChoices.TYPE_CONTACT_TEAM:
+            return None
```
glennmatthews: Maybe a final `else` that raises an exception, so we'll know if we're missing support for any type?

nautobot/extras/filters/__init__.py
```diff
@@ -979,6 +980,21 @@ class Meta:
...
+            "value": "icontains",
+            "metadata_type__name": "icontains",
+            "contact__name": "icontains",
+            "team__name": "icontains",
+        },
+    )
+
+    class Meta:
+        model = ObjectMetadata
+        fields = "__all__"
```
glennmatthews, line 995: Does this result in an undesired `_value` filter? Do we need to implement a custom `value` filter now?

nautobot/extras/factory.py
```diff
@@ -238,6 +239,75 @@ def content_types(self, create, extracted, **kwargs):
                 )
 
 
+class ObjectMetadataFactory(BaseModelFactory):
+    """ObjectMetatdata model factory"""
```
glennmatthews, line 243:
```suggestion
    """ObjectMetadata model factory."""
```

nautobot/extras/api/serializers.py
```diff
@@ -766,6 +767,36 @@ class Meta:
...
 
+class ObjectMetadataSerializer(ValidatedModelSerializer):
+    assigned_object_type = ContentTypeField(
+        queryset=ContentType.objects.filter(FeatureQuery("metadata").get_query()),
+    )
+    assigned_object = serializers.SerializerMethodField()
+
+    class Meta:
+        model = ObjectMetadata
+        fields = "__all__"
```
glennmatthews, line 778: Do we need a special field for `value` now?

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +141,249 @@ def save(self, *args, **kwargs):
...
         # TODO: should we behave like CustomFieldChoice and raise a ProtectedError if this choice is in use anywhere?
         super().delete(*args, **kwargs)
+
+
+class ObjectMetadataManager(BaseManager.from_queryset(RestrictedQuerySet)):
+    use_in_migrations = True
+
+    def get_for_model(self, instance):
+        """Return all ObjectMetadatas assigned to the given instance."""
+        cache_key = f"{self.get_for_model.cache_key_prefix}.{instance}"
```
glennmatthews, line 151: Perhaps?

```suggestion
        cache_key = f"{self.get_for_model.cache_key_prefix}.{instance.pk}"
```

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +141,249 @@ def save(self, *args, **kwargs):
     def delete(self, *args, **kwargs):
         # TODO: should we behave like CustomFieldChoice and raise a ProtectedError if this choice is in use anywhere?
         super().delete(*args, **kwargs)
+
+
+class ObjectMetadataManager(BaseManager.from_queryset(RestrictedQuerySet)):
+    use_in_migrations = True
+
+    def get_for_model(self, instance):
```
glennmatthews, line 149: Maybe `get_for_object` to make it clearer?

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +141,249 @@ def save(self, *args, **kwargs):
...
+        """Return all ObjectMetadatas assigned to the given instance."""
+        cache_key = f"{self.get_for_model.cache_key_prefix}.{instance}"
+        queryset = cache.get(cache_key)
+        if queryset is None:
+            assigned_object_type = ContentType.objects.get_for_model(instance)
+            assigned_object_id = instance.pk
+            queryset = self.get_queryset().filter(
+                assigned_object_type=assigned_object_type, assigned_object_id=assigned_object_id
+            )
+            cache.set(cache_key, queryset)
```
glennmatthews, line 159: If we're going to do caching here (not sure there's as much value given that this is per-instance rather than per-model-class), then you need to make sure the cache gets cleared appropriately when creating, updating, or deleting any ObjectMetadata records. Might not be worth the trouble here?

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +141,249 @@ def save(self, *args, **kwargs):
...
+                fields=["assigned_object_type", "assigned_object_id", "team"],
+            ),
+        ]
+
+    @property
+    def value(self):
+        if self.metadata_type.data_type == MetadataTypeDataTypeChoices.TYPE_CONTACT_TEAM:
+            return self.contact or self.team
+        else:
+            return self._value
```
glennmatthews, line 232: Probably would be good to implement the corresponding property setter as well?

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +141,249 @@ def save(self, *args, **kwargs):
...
+                    if isinstance(value, str):
+                        value = value.split(",")
+                    if not set(value).issubset(self.metadata_type.choices.values_list("value", flat=True)):
+                        raise ValidationError(
+                            f"Invalid choice(s) ({value}). Available choices are: {', '.join(self.metadata_type.choices.values_list('value', flat=True))}"
+                        )
+            else:
+                raise ValidationError(
+                    f"value is a required field that cannot be empty for metadata type {metadata_type_data_type}."
+                )
```
glennmatthews, line 361: This `else` way down here feels lonely to me. Maybe invert the logic, i.e.:

```py
if metadata_type_data_type == MetadataTypeDataTypeChoices.TYPE_CONTACT_TEAM:
    # contact/team validation
elif value in [None, "", []]:  # TODO: aren't some of these valid for TYPE_JSON? Or should they be?
    raise ValidationError
else:
    # type-specific validation
```

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +141,249 @@ def save(self, *args, **kwargs):
...
+        related_name="object_metadatas",
+        blank=True,
+        null=True,
+    )
+    scoped_fields = JSONArrayField(
+        base_field=models.CharField(
+            max_length=CHARFIELD_MAX_LENGTH,
+        ),
+        help_text="List of scoped fields, only direct fields on the model",
+    )
```
glennmatthews, line 195: Do we want/need to enforce that any specified scoped_fields actually exist on the given associated_object?

Need to think about how this will work for custom fields (and *maybe* relationships too?)

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +141,249 @@ def save(self, *args, **kwargs):
...
+    _value = models.JSONField(
+        blank=True,
+        null=True,
+        help_text="Relevant data value to an object field or a set of object fields",
+    )
+    assigned_object_type = models.ForeignKey(to=ContentType, on_delete=models.SET_NULL, null=True, related_name="+")
```
glennmatthews, line 201: Do we need to validate that `assigned_object_type` is a valid content-type for the given `metadata_type`?

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +141,249 @@ def save(self, *args, **kwargs):
...
+                            datetime.strptime(value, "%Y-%m-%d")
+                        except TypeError:
+                            raise ValidationError("Value must be a date or str object.")
+                        except ValueError:
+                            raise ValidationError("Date values must be in the format YYYY-MM-DD.")
+                # Validate datetime
+                elif metadata_type_data_type == MetadataTypeDataTypeChoices.TYPE_DATETIME:
+                    if not isinstance(value, datetime):
+                        try:
+                            datetime.strptime(value, "%Y-%m-%d %H:%M:%S")
```
glennmatthews, line 338: Let's be sure to test the round-trip behavior of this field type, especially when timezone-aware datetimes are involved? Let's also make sure the string format here matches JSON conventions for storing timezone-aware datetimes.

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +141,249 @@ def save(self, *args, **kwargs):
...
+                        raise ValidationError("Value must be an integer.")
+                    # TODO uncomment this when MetaDataType validation_minimum, validation_maximum and validation_regex fields are implemented
+                    # if self.metadata_type.validation_minimum is not None and value < self.metadata_type.validation_minimum:
+                    #     raise ValidationError(f"Value must be at least {self.metadata_type.validation_minimum}")
+                    # if self.metadata_type.validation_maximum is not None and value > self.metadata_type.validation_maximum:
+                    #     raise ValidationError(f"Value must not exceed {self.metadata_type.validation_maximum}")
+                # Validate integer
+                elif metadata_type_data_type == MetadataTypeDataTypeChoices.TYPE_FLOAT:
+                    try:
+                        value = float(value)
```
glennmatthews, line 315: Do we ever update `self._value` with the type-coerced new `value`? I'm not seeing it but it feels like it should be necessary?

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +141,249 @@ def save(self, *args, **kwargs):
...
+
+        for scoped_fields in object_metadata_scoped_fields:
+            duplicate_scoped_fields_list |= set(self.scoped_fields).intersection(scoped_fields)
+
+        if duplicate_scoped_fields_list:
+            raise ValidationError(
+                {
+                    "scoped_fields": f"There are other Object Metadata instances of metadata type {self.metadata_type} scoping the same fields {duplicate_scoped_fields_list} for {self.assigned_object}"
+                }
+            )
```
glennmatthews, line 389: IIRC, per the design, a scoped_fields of `[]` implies "all fields on this record"? If that's the case, then should `[]` be considered to conflict with *any* other metadata of this type on this record?

nautobot/extras/tests/test_models.py
```diff
@@ -1324,6 +1327,252 @@ def get_absolute_url(self):
...
+    def test_contact_team_mutual_exclusive(self):
```
glennmatthews, line 1341: In general I think we need positive tests (valid values) for each datatype as well?

nautobot/extras/tests/test_models.py
```diff
@@ -1324,6 +1327,252 @@ def get_absolute_url(self):
...
+            assigned_object_id=Location.objects.first().pk,
+        )
+        obj_metadata.validated_save()
+
+        invalid_options = ["Not A valid option", "NOT A VALID OPTION"]
+        obj_metadata.value = invalid_options
+        with self.assertRaises(ValidationError) as context:
+            obj_metadata.validated_save()
+        self.assertIn(f"Invalid choice(s) ({invalid_options})", str(context.exception))
+
```
glennmatthews, line 1550: Tests for TYPE_DATETIME and TYPE_FLOAT?

nautobot/extras/models/metadata.py
HanlinMiao: Agreed removed the caching feature from this helper function.

nautobot/extras/filters/__init__.py
HanlinMiao: Sounds like we need to implement a similar mixin like `CustomFieldModelFilterSetMixin`?

nautobot/extras/api/serializers.py
HanlinMiao: Is there something similar with CustomField? I am trying to find it with no avail.

nautobot/extras/filters/__init__.py
glennmatthews: Maybe longer term (so we can filter an object by its metadata) but I'm not sure I follow how that's necessary to address this specific concern?

nautobot/extras/api/serializers.py
glennmatthews: Not that I'm aware of. For CustomField values on an object, it's just a JSON field - might be able to implement something similar for API representation of `value`, though the Contact/Team cases might be easier to represent as distinct `contact` and `team` fields in the API for now rather than trying to have a single `value` that could be a JSON object or a FK reference?

HanlinMiao: I see, I was overthinking it. Declaring `value = serializers.JSONField(read_only=True, required=False)` seems to work just fine in the API.

glennmatthews: Does that work when value is a Contact or Team?

glennmatthews: Also, shouldn't need to be `read_only=True` once the property setter is implemented, maybe? Definitely need the ability to set `value` via REST.

nautobot/extras/tests/test_models.py
```diff
@@ -1444,6 +1446,33 @@ def test_integer_field_value(self):
...
+        # Assign another disallowed value (str) to the first Location
+        with self.assertRaises(ValidationError) as context:
+            obj_metadata.value = "I am not an integer"
+        self.assertIn("Value must be a float", str(context.exception))
+
+        # Assign another disallowed value (int) to the first Location
+        with self.assertRaises(ValidationError) as context:
+            obj_metadata.value = 2
+        self.assertIn("Value must be a float", str(context.exception))
```
glennmatthews, line 1471: Can/should we auto-convert ints to floats?

nautobot/extras/tests/test_models.py
HanlinMiao: Added tests for those types

nautobot/extras/factory.py
```diff
@@ -309,6 +309,9 @@ def assigned_object_type(self):
             content_type = factory.random.randgen.choice(
                 ContentType.objects.filter(FeatureQuery("metadata").get_query(), pk__in=allowed_content_types)
             )
+            # It does not have a get_absolute_url attribute and is causing failure in API unittests
+            if content_type.app_label == "extras" and content_type.model == "taggeditem":
+                continue
```
glennmatthews, line 314: IMO `taggeditem` is an implementation detail and this model shouldn't support metadata as a feature.

nautobot/extras/models/metadata.py
HanlinMiao: https://github.com/nautobot/nautobot/issues/5861

nautobot/extras/tests/test_api.py
```diff
@@ -2567,6 +2578,67 @@ def setUpTestData(cls):
...
+            assigned_object_type=ContentType.objects.get_for_model(Prefix),
+            assigned_object_id=Prefix.objects.first().pk,
+        )
+        ObjectMetadata.objects.create(
+            metadata_type=mdts[1],
+            value="Aloha",
+            scoped_fields=["status"],
+            assigned_object_type=ContentType.objects.get_for_model(Prefix),
+            assigned_object_id=Prefix.objects.last().pk,
+        )
```
glennmatthews, line 2613: Let's make sure we have at least one ObjectMetadata of type TYPE_CONTACT_TEAM. I'm concerned that it may not currently be represented correctly in the API.

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +141,298 @@ def save(self, *args, **kwargs):
...
+        """
+        super().clean()
+        value = self._value
+        metadata_type_data_type = self.metadata_type.data_type
+        if (
+            isinstance(value, list)
+            and len(value) == 1
+            and self.metadata_type.data_type != MetadataTypeDataTypeChoices.TYPE_MULTISELECT
+        ):
+            value = value[0]
```
glennmatthews, line 262: This feels like a kludge to me. Is it really needed?

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +141,298 @@ def save(self, *args, **kwargs):
...
+            if value not in [None, "", []]:
+                raise ValidationError(f"A value cannot be specified for type {metadata_type_data_type}")
+            if self.contact is None and self.team is None:
+                raise ValidationError("Either a contact or a team must be specified")
+            if self.contact is not None and self.team is not None:
+                raise ValidationError("A contact and a team cannot be both specified at once")
+        elif value in [None, "", []]:
+            raise ValidationError(
+                f"value is a required field that cannot be empty for metadata type {metadata_type_data_type}."
+            )
```
glennmatthews, line 279: Do we want to allow TYPE_JSON at least to be an empty string or empty list? What about an empty string for TYPE_TEXT or an empty list for type TYPE_MULTISELECT?

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +141,298 @@ def save(self, *args, **kwargs):
...
+        ):
+            value = value[0]
+        # Validate the assigned_object_type is allowed in metadata_type's content_types
+        if not self.metadata_type.content_types.filter(pk=self.assigned_object_type.id).exists():
+            raise ValidationError(
+                f"Assigned Object Type {self.assigned_object_type} is not allowed by Metadata Type {self.metadata_type}"
+            )
+        # Check for MetadataTypeDataTypeChoices.TYPE_CONTACT_TEAM first
+        if metadata_type_data_type == MetadataTypeDataTypeChoices.TYPE_CONTACT_TEAM:
+            if value not in [None, "", []]:
```
glennmatthews, line 270: I feel like in this case `_value` should be only `None`?

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +141,298 @@ def save(self, *args, **kwargs):
...
+                                f"Datetime values must be in the following formats {acceptable_datetime_formats}"
+                            )
+                    except TypeError:
+                        raise ValidationError("Value must be a datetime or str object.")
+                else:
+                    # if value is a datetime object
+                    # check if datetime object has tzinfo
+                    if value.tzinfo is None:
+                        value = value.replace(tzinfo=timezone.utc)
+                    value = value.replace(microsecond=0).isoformat()
```
glennmatthews, line 385: Why's the timezone conversion needed? Does `isoformat()` not already account for time zones?

nautobot/extras/models/metadata.py
```diff
@@ -135,3 +141,298 @@ def save(self, *args, **kwargs):
...
+        # TODO Make sure that scoped fields are all direct fields on the assigned_object_type model
+
+        for scoped_fields in object_metadata_scoped_fields:
+            duplicate_scoped_fields_list |= set(self.scoped_fields).intersection(scoped_fields)
+            # if self.scoped_fields or scoped_fields are [] which means all fields are scoped
+            # That means any fields in self.scoped_fields or scoped_fields are considered overlapped
+            if not self.scoped_fields or not scoped_fields:
+                raise ValidationError(
+                    {
+                        "scoped_fields": f"There are other Object Metadata instances of metadata type {self.metadata_type} scoping all the fields for {self.assigned_object}"
```
glennmatthews, line 429: Feel like this is not the clearest error message for the case where `self.scoped_fields` is the one covering all fields?

nautobot/extras/tests/test_api.py
```diff
@@ -2567,6 +2578,67 @@ def setUpTestData(cls):
...
+                "assigned_object_type": "dcim.location",
+                "assigned_object_id": Location.objects.first().pk,
+            },
+            {
+                "metadata_type": mdts[0].pk,
+                "scoped_fields": ["device_type"],
+                "value": "Greetings",
+                "assigned_object_type": "dcim.device",
+                "assigned_object_id": Device.objects.first().pk,
+            },
```
glennmatthews, line 2635: Similarly let's make sure create_data includes creating at least one TYPE_CONTACT_TEAM record.

nautobot/extras/tests/test_api.py
```diff
@@ -2567,6 +2578,67 @@ def setUpTestData(cls):
...
+        ]
+        cls.update_data = {
+            "scoped_fields": ["pk"],
```
glennmatthews, line 2638: Isn't this going to trip the scoped_fields uniqueness validation?

nautobot/extras/tests/test_models.py
HanlinMiao: Yeah I think we should, included now in the positive tests.

nautobot/extras/models/metadata.py
HanlinMiao: changed this to `elif value is None:` to allow "" and [].

nautobot/extras/models/metadata.py
HanlinMiao: Separated the two conditions and gave them distinct error messages.

nautobot/extras/tests/test_api.py
glennmatthews: oh, I was thinking of `bulk_update_data`. Sorry about that.

nautobot/extras/models/metadata.py
HanlinMiao: Yeah, I don't like the pattern here either. But this is for the csv create case. `test_recreate_object_csv` caught this edge case. Since the value field is a `JSONField` on the serializer, when you pass a `float`/`str`, etc, the value field will cast them into a list. Declaring it as a `CharField` solves it for the `float/str` case but not the other types. I wonder if a custom `SerializerMethodField` will solve it.
```
['1460.352280751']
float
F
FAIL: test_recreate_object_csv (nautobot.extras.tests.test_api.ObjectMetadataTest)
CSV export an object, delete it, and recreate it via CSV import.
  File "/source/nautobot/core/testing/api.py", line 707, in test_recreate_object_csv
    self.assertHttpStatus(response, status.HTTP_201_CREATED, csv_data)
  File "/source/nautobot/core/testing/mixins.py", line 179, in assertHttpStatus
    self.assertIn(response.status_code, expected_status, err_message)
AssertionError: 500 not found in [201] : display,id,object_type,natural_slug,assigned_object_type,assigned_object,value,scoped_fields,assigned_object_id,metadata_type__name,contact__name,contact__phone,contact__email,team__name,team__phone,team__email,created,last_updated
Accounting technician - Advertising account executive (rachellopez@example.net) - 1460.352280751,86e74add-c680-4d83-86d9-2e684cfc6680,extras.objectmetadata,86e74add-c680-4d83-86d9-2e684cfc6680_86e7,extras.team,25b3f8b0-9fa7-483f-af5b-e3f42e4f0a9a,1460.352280751,"zVLNIBubsHkSgJYJMIxV,dqtBPZXVBeDdQtsIaAGY,bvzwmonNhPOjFdxvRlca,LZQgYNkkTrSifkLZcxxD,uqIdNrochpnVSMgTRVyS,kWvhuuqTsARuOLHJUZkS,tjtdMZFXpCQcjlpFUWeF",25b3f8b0-9fa7-483f-af5b-e3f42e4f0a9a,Accounting technician,NoObject,NoObject,NoObject,NoObject,NoObject,NoObject,2024-06-27T20:39:00.875000Z,2024-06-27T20:39:00.875000Z

Expected HTTP status(es) [201]; received 500:
{"error": "float() argument must be a string or a real number, not 'list'", "exception": "TypeError", "nautobot_version": "2.3.0a1", "python_version": "3.10.14"}
```

nautobot/extras/api/serializers.py
HanlinMiao: Unfortunately, it is a pretty big set of serializers. But it might be a necessary evil since this has always been our pattern for GFKs.

glennmatthews: Yeah, I think it's necessary - was just curious.

nautobot/extras/models/metadata.py
glennmatthews: Thanks for the explanation. I think maybe just adding a comment to this effect may be fine for the moment, and maybe opening a followup issue to investigate whether something needs to be tweaked on the serializer.

HanlinMiao: https://github.com/nautobot/nautobot/issues/5866

nautobot/extras/tests/test_models.py
HanlinMiao: Included type conversion behaviors
1. from int to float
2. from float to int
3. from String to int
4. from String to float

glennmatthews (approved):
Looks like the merge from #5845 wasn't quite correct (sorry about the pain there) but once that's addressed and tests are back to passing, :shipit:

PR #5842, opened 2024-06-24
Fix tests in next and other small fixes

gsnider2195:
fix tests, add mysqldump client to dev requirements, fix `__all__` in ipam models.py

# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/5842.fixed
+++ b/changes/5842.fixed
@@ -1,1 +1,1 @@
-Fixed missing classes when importing * from nautobot.ipam.models.
+Fixed missing classes when importing `*` from nautobot.ipam.models.
--- a/changes/5842.fixed
+++ b/changes/5842.fixed
@@ -1,0 +1,1 @@
+Fixed missing classes when importing * from nautobot.ipam.models.
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -79,7 +79,7 @@ FROM system-dependencies AS system-dev-dependencies
 # Install development/install-time OS dependencies
 # hadolint ignore=DL3008
 RUN apt-get update && \
-    apt-get install --no-install-recommends -y build-essential libssl-dev pkg-config libldap-dev libsasl2-dev libmariadb-dev nodejs npm && \
+    apt-get install --no-install-recommends -y build-essential libssl-dev pkg-config libldap-dev libsasl2-dev libmariadb-dev nodejs npm mariadb-client && \
     apt-get autoremove -y && \
     apt-get clean all && \
     rm -rf /var/lib/apt/lists/*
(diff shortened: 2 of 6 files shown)


changes/5842.fixed
```diff
@@ -0,0 +1 @@
+Fixed missing classes when importing * from nautobot.ipam.models.
```
glennmatthews, line 1:
```suggestion
Fixed missing classes when importing `*` from nautobot.ipam.models.
```

nautobot/ipam/factory.py
```diff
@@ -377,9 +377,7 @@ def children(self, create, extracted, **kwargs):
 
         # Default to maximum of 4 children unless overridden in kwargs
         max_count = int(kwargs.pop("max_count", 4))
-        child_count = faker.Faker().pyint(min_value=0, max_value=min(max_count, self.prefix.size))
-        if child_count == 0:
-            return
+        child_count = faker.Faker().pyint(min_value=1, max_value=min(max_count, self.prefix.size))
```
gsnider2195, line 380: This fixed the current `next` test failures. We weren't creating any IP Addresses for IPv6 because we had multiple random chances to not create children. This `random.pyint` was redundant after the above (line 361) `random.choice` so it should be fine to change.

docker/Dockerfile
```diff
@@ -79,7 +79,7 @@ FROM system-dependencies AS system-dev-dependencies
 # Install development/install-time OS dependencies
 # hadolint ignore=DL3008
 RUN apt-get update && \
-    apt-get install --no-install-recommends -y build-essential libssl-dev pkg-config libldap-dev libsasl2-dev libmariadb-dev nodejs npm && \
+    apt-get install --no-install-recommends -y build-essential libssl-dev pkg-config libldap-dev libsasl2-dev libmariadb-dev nodejs npm mariadb-client && \
```
glennmatthews, line 82: Just confirming that this is a dev dependency only? If it's needed in the prod images it should be added at line 51 instead.

nautobot/ipam/factory.py
glennmatthews: Do we ever have a case where `min(max_count, self.prefix.size)` is 0 or is it always `>= 1`?

docker/Dockerfile
gsnider2195: This should only need to be a dev dependency since it's just required to run the parallel unit tests

nautobot/ipam/factory.py
gsnider2195: That's a good point. If someone explicitly requests `children__max_count=0`, this could break. I'll add a guard rail for that case. Otherwise, `self.prefix.size` should always be at least 1.

HanlinMiao (approved):

glennmatthews (approved):

glennmatthews (approved):

PR #5845, opened 2024-06-24
Collapse `StaticGroup` into `Dynamic Group` feature

### LINKED ISSUE #5786: Collapse StaticGroup model into DynamicGroup
### As ...

### I want ...

Static groups to be a subtype of the existing Dynamic Groups model/feature, rather than a distinct feature and model of its own.

### So that ...

When developing an App or a Job, I can use groups of objects without needing to write double the code or otherwise support two different data models and APIs for the same concept.

The work started in https://github.com/nautobot/nautobot/tree/u/glennmatthews-consolidated-groups is completed.

Note: this branch currently attempts to also change the default for TreeModel managers from `with_tree_fields` to `without_tree_fields`, which is causing a number of test failures and behavior changes. It may be possible/desirable to revert that change, but if so, we need to be sure to test groups of TreeModel records (Location, etc.) as I was initially encountering a number of exceptions due to the presence of tree fields in the group members querysets.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

Yes - ideally (as started in the aforementioned prototype branch) the extras/107 migration that introduced the StaticGroup model should be rewritten or removed, rather than adding a new migration for deleting the model.

### External Dependencies

None

glennmatthews:
# Closes #5786 
# What's Changed

Lots!

- At its most basic, instead of having `StaticGroup` plus two (implicit) types of `DynamicGroup` (filter-defined and set-operation-defined), we now have three explicit types of `DynamicGroup`. 
  - Unfortunately, enhancing the existing `DynamicGroupMixin` class to support the reverse-relation to `StaticGroupAssociation` proved impossible without needing to rewrite a bunch of migration history in any models using this mixin. As a workaround, I've added a new `DynamicGroupsModelMixin` class and deprecated `DynamicGroupMixin`.
- Additionally, due to issues seen with static Groups of TreeModels and similar issues in the past, I've changed `TreeManager` to default to `without_tree_fields` instead of `with_tree_fields`. This should improve performance in many cases but may impact Apps or Jobs that were relying on the old default; such code should be updated to explicitly call `.with_tree_fields()` where appropriate.
- Additionally, I made a significant performance improvement to our OpenAPI schema tests (though they still remain quite slow due to the overhead of generating the schema, it's much less bad than it was). See 0fd142cc2c474d09f573bc41c6cf2803f8c3e384.

Sorry for the size of the diffs! At least for once they're net-negative a few hundred lines of code... :-) 

# Screenshots

image

(Definitely can and probably should revisit the user-facing "group type" labels here!)

# TODO

- [x] Unit, Integration Tests
- [x] Add `tenant` information to DynamicGroup detail view
- [x] Add change fragment and docs about DynamicGroups now supporting tags and tenants


nautobot/core/templates/generic/object_retrieve.html
```diff
@@ -218,21 +207,33 @@ <h1>
...
                                 <div class="panel-heading">
-                                    <strong>Static Groups</strong>
+                                    <strong>Dynamic Groups</strong>
                                     <div class="pull-right noprint">
                                         <!-- Insert table config button here -->
                                     </div>
                                 </div>
                                 <div class="table-responsive">
-                                    {% render_table associated_static_groups_table 'inc/table.html' %}
+                                    {% render_table associated_dynamic_groups_table 'inc/table.html' %}
```
glennmatthews, line 236: This means that the dedicated dynamic group tab views (e.g. `device_dynamicgroups`) are no longer needed. We should probably at least mark those as deprecated if not remove them now.

nautobot/docs/release-notes/version-2.3.md
```diff
@@ -26,6 +26,7 @@ Added the ability for users to save multiple configurations of list views (table
 
 #### Static Groups
 
+TODO update this section.
```
glennmatthews, line 29: TODO

nautobot/extras/filters/mixins.py
```diff
@@ -177,49 +177,32 @@ def __init__(self, side, relationship=None, queryset=None, qs=None, *args, **kwa
         self.side = side
         super().__init__(queryset=queryset, *args, **kwargs)
 
+    def generate_query(self, value, **kwargs):
```
glennmatthews, line 180: This change was needed to fix an issue when defining DynamicGroups with relationship filters. In the "old" code, such filters would work fine for filter-based groups, but would (previously unknown/unreported) cause errors if such a group was included in a set-based parent group. With the different code paths introduced by this refactor, even filter-based groups would demonstrate this issue, which was caught by one of our unit tests.

nautobot/extras/forms/forms.py
```diff
@@ -540,7 +542,10 @@ class DynamicGroupMembershipFormSetForm(forms.ModelForm):
 
     group = DynamicModelChoiceField(
         queryset=DynamicGroup.objects.all(),
```
glennmatthews, line 544: Probably should restrict the queryset by group_type as well?

nautobot/extras/tests/test_views.py
```diff
@@ -857,6 +861,7 @@ def test_get_object_dynamic_groups_with_permission(self):
         self.assertIn("DG 3", response_body, msg=response_body)
 
     def test_get_object_dynamic_groups_with_constrained_permission(self):
+        # TODO this may need re-implementation
```
glennmatthews, line 864:
```suggestion
```

nautobot/extras/tests/test_views.py
```diff
@@ -795,13 +794,18 @@ def setUpTestData(cls):
...
+            "tenant": Tenant.objects.first().pk,
             # Management form fields required for the dynamic formset
             "dynamic_group_memberships-TOTAL_FORMS": "0",
             "dynamic_group_memberships-INITIAL_FORMS": "1",
             "dynamic_group_memberships-MIN_NUM_FORMS": "0",
             "dynamic_group_memberships-MAX_NUM_FORMS": "1000",
         }
 
+    def _get_queryset(self):
+        return super()._get_queryset().filter(group_type=DynamicGroupTypeChoices.TYPE_DYNAMIC_FILTER)  # TODO
```
glennmatthews, line 807: Revisit this, but the test was failing in various cases due to trying to set the filter on a static-type group and this was the quickest fix.

nautobot/tenancy/forms.py
```diff
@@ -12,8 +12,8 @@
     NautobotBulkEditForm,
     NautobotFilterForm,
     NautobotModelForm,
-    TagsBulkEditFormMixin,
 )
+from nautobot.extras.forms.mixins import TagsBulkEditFormMixin
```
glennmatthews, line 16: Working around a circular import issue now that DynamicGroup forms are using Tenancy form mixins.

changes/5786.changed
```diff
@@ -0,0 +1,7 @@
+Changed `TreeManager` class (used for tree-models such as Location, RackGroup, and TenantGroup) default behavior from `with_tree_fields` to `without_tree_fields`. This should improve performance in many cases but may impact Apps or Jobs that were relying on the old default; such code should be updated to explicitly call `.with_tree_fields()` where appropriate.
+Merged the `StaticGroup` model added in #5472 into the existing `DynamicGroup` model as a special type of group.
+Replaced `static_groups` filter added in #5472 with a `dynamic_groups` filter.
+Replaced `static_groups` model form field added in #5472 with a `dynamic_groups` field.
+Replaced `Static Groups` object table column added in #5472 with a `Dynamic Groups` column.
+Replaced `static_groups` GraphQL field added in #5472 with a `dynamic_groups` field.
+Replaced `StaticGroupMixin` model mixin class added in #5631 with a `DynamicGroupsModelMixin` class. Still included by default in `OrganizationalModel` and `PrimaryModel`.
```
gsnider2195, line 7:
```suggestion
Merged the `StaticGroup` model added in [#5472](https://github.com/nautobot/nautobot/issues/5472) into the existing `DynamicGroup` model as a special type of group.
Replaced `static_groups` filter added in [#5472](https://github.com/nautobot/nautobot/issues/5472) with a `dynamic_groups` filter.
Replaced `static_groups` model form field added in [#5472](https://github.com/nautobot/nautobot/issues/5472) with a `dynamic_groups` field.
Replaced `Static Groups` object table column added in [#5472](https://github.com/nautobot/nautobot/issues/5472) with a `Dynamic Groups` column.
Replaced `static_groups` GraphQL field added in [#5472](https://github.com/nautobot/nautobot/issues/5472) with a `dynamic_groups` field.
Replaced `StaticGroupMixin` model mixin class added in [#5631](https://github.com/nautobot/nautobot/pull/5631) with a `DynamicGroupsModelMixin` class. Still included by default in `OrganizationalModel` and `PrimaryModel`.
```

nautobot/core/filters.py
```diff
@@ -760,17 +760,25 @@ def get_filters(cls):
...
-            )
+        if "dynamic_groups" not in filters and getattr(cls._meta.model, "is_dynamic_group_associable_model", False):
+            if not hasattr(cls._meta.model, "static_group_association_set"):
+                logger.warning(
+                    "Model %s has 'is_dynamic_group_associable_model = True' but lacks "
+                    "a 'static_group_association_set' attribute. Perhaps this is due to it inheriting from "
+                    "the deprecated DynamicGroupMixin class instead of the preferred DynamicGroupsModelMixin?",
+                    cls._meta.model,
+                )
+            else:
```
gsnider2195, line 771: Nitpick: Should this be `elif "dynamic_groups" not in filters:` so that this filter can be overridden if desired?

nautobot/core/management/commands/generate_test_data.py
```diff
@@ -278,7 +278,7 @@ def _create_batch(some_factory, count, description="", **kwargs):
         # _create_batch(ClusterFactory, 10)
         # _create_batch(VirtualMachineFactory, 10)
         # We need to remove them from there and enable them here instead, but that will require many test updates.
-        _create_batch(StaticGroupFactory, 20)
+        _create_batch(DynamicGroupFactory, 20, description="and StaticGroupAssociations")
```
gsnider2195, line 281: Is this only creating dynamic groups with type static? Should the description reflect that?

nautobot/core/templates/generic/object_retrieve.html
```diff
@@ -218,21 +207,33 @@ <h1>
...
-        {% if object.is_static_group_associable_model and perms.extras.view_staticgroup %}
-            <div id="static_groups" role="tabpanel" class="tab-pane {% if request.GET.tab == 'static_groups' %}active{% else %}fade{% endif %}">
+        {% if object.is_dynamic_group_associable_model and perms.extras.view_dynamicgroup %}
+            <div id="dynamic_groups" role="tabpanel" class="tab-pane {% if request.GET.tab == 'dynamic_groups' %}active{% else %}fade{% endif %}">
+                <div class="row">
+                    <div class="col-md-12">
+                        <div class="alert alert-warning">
+                            Dynamic group membership is cached for performance reasons,
+                            therefore this table may not always be up-to-date.
+                            <br>You can refresh the membership of any specific group by viewing it from the list below
```
gsnider2195, line 217:
```suggestion
                            <br>You can refresh the membership of any specific group by navigating to it from the list below
```

nautobot/core/testing/models.py
```diff
@@ -44,19 +44,20 @@ def test_get_docs_url(self):
...
-
-                self.assertTrue(hasattr(self.model, "static_groups"))
-                self.assertIsInstance(self.model.objects.first().static_groups, QuerySet)
-                self.assertEqual(self.model.objects.first().static_groups.model, StaticGroup)
-
-                if StaticGroup.objects.get_for_model(self.model).exists():
-                    sg = StaticGroup.objects.get_for_model(self.model).first()
+        def test_dynamic_group_api(self):
+            """For dynamic-group capable models, check that they work as intended."""
+            if getattr(self.model, "is_dynamic_group_associable_model", False):
```
gsnider2195, line 49: should this have an `else: self.skip`?

nautobot/core/filters.py
glennmatthews: Isn't that covered by line 763 above?

nautobot/core/management/commands/generate_test_data.py
glennmatthews: The factory currently makes TYPE_STATIC and TYPE_DYNAMIC_FILTER groups but not yet TYPE_DYNAMIC_SET.

nautobot/core/filters.py
gsnider2195: 🤦‍♂️

nautobot/docs/user-guide/platform-functionality/dynamicgroup.md
```diff
@@ -2,69 +2,107 @@
...
+
+The members of each Dynamic Group are cached and tracked by [Static Group Association](staticgroupassociation.md) database records.
 
 ## Introduction
 
-### Creating Dynamic Groups
+### Creating Dynamic Groups in the UI
 
-Dynamic Groups can be created through the UI under _Organization > Dynamic Groups_ and clicking the "Add" button, or through the REST API.
+Dynamic Groups can be created through the UI under _Organization > Groups > Dynamic Groups_ and clicking the "Add" button, or through the REST API.
```
gsnider2195, line 25: Nitpick: `in the UI` in the heading and `or through the REST API` in the first line feels weird

nautobot/docs/user-guide/platform-functionality/dynamicgroup.md
```diff
@@ -2,69 +2,107 @@
...
+Set-based Dynamic Groups do not directly define a filter for identifying member objects; instead, they define their members based on set operations (AND, OR, and NOT, or Union, Intersection, and Difference if you prefer) involving _other_ Dynamic Groups.
 
-An object is considered a member of an advanced Dynamic Group if it matches the aggregated filter criteria across all descendant groups.
+!!! tip
+    At this time, a set-based Dynamic Group may include filter-based Dynamic Groups and other set-based Dynamic Groups in its members definition, but set-based Dynamic Groups _may not_ currently include static-assignment-based Dynamic Groups. This restriction may be removed in a future Nautobot release.
 
-When editing a Dynamic Group, under the **Filter Options** section, you will find a **Child Groups** tab that allows one to make other Dynamic Groups of the same Content Type children of the parent group.
+An object is considered a member of an set-based Dynamic Group if it matches the aggregated filter criteria derived from the "descendant" groups included in this group definition by way of the "Dynamic Group Membership" through table.
```
gsnider2195, line 84:
```suggestion
An object is considered a member of a set-based Dynamic Group if it matches the aggregated filter criteria derived from the "descendant" groups included in this group definition by way of the "Dynamic Group Membership" through table.
```

gsnider2195: What is the "Dynamic Group Membership" through table?

nautobot/docs/user-guide/platform-functionality/dynamicgroup.md
```diff
@@ -435,48 +480,37 @@ Response:
 }
 ```
 
-#### Updating or Deleting a Dynamic Group
+#### Updating or Deleting a Dynamic Group via REST
 
 !!! important
-    It is not possible to perform a nested update of `children` when updating a new Dynamic Group. You must use the endpoint for creating Dynamic Group Memberships as explained below under [Updating or Deleting Child Groups](#updating-or-deleting-child-groups).
+    It is not possible to perform a nested update of `children` when updating a new Dynamic Group. You must use the endpoint for creating Dynamic Group Memberships as explained below under [Updating or Deleting Child Groups via REST](#updating-or-deleting-child-groups-via-rest).
```
gsnider2195, line 486:
```suggestion
    It is not possible to perform a nested update of `children` when updating a Dynamic Group. You must use the endpoint for creating Dynamic Group Memberships as explained below under [Updating or Deleting Child Groups via REST](#updating-or-deleting-child-groups-via-rest).
```

nautobot/docs/user-guide/platform-functionality/dynamicgroup.md
```diff
@@ -435,48 +480,37 @@ Response:
...
 
 Updating or deleting a Dynamic Group is done by sending a request to the detail endpoint for that object.
 
-A Dynamic Group may be updated using `PUT` or `PATCH` (for a partial update) requests. A `PUT` request requires the entire object to be updated in place. For example if you wanted to update the `name` and the `description` together, leaving every other field with their current values as provided:
+A Dynamic Group may be updated using `PUT` or `PATCH` (for a partial update) requests. As usual for REST, a `PUT` request requires the entire object to be described, while a `PATCH` allows for individual fields to be updated without affecting the other fields. For example if you wanted to update the `name` and the `description` together, leaving every other field with their current values as provided:
```
gsnider2195, line 490:
```suggestion
A Dynamic Group may be updated using `PUT` or `PATCH` requests. As usual for REST, a `PUT` request requires the entire object to be described, while a `PATCH` allows for individual fields to be updated without affecting the other fields. For example if you wanted to update the `name` and the `description` together, leaving every other field with their current values as provided:
```

The following line explains it better IMO

nautobot/docs/user-guide/platform-functionality/dynamicgroup.md
```diff
@@ -565,24 +600,11 @@ GET /api/extras/dynamic-groups/1f825078-b6dc-4b12-9463-be5a9189b03f/
...
-{
-    "group": {"name": "first-child"},
-    "parent_group": {"name": "parent"},
-    "operator": "difference",
-    "weight": 10
-}
-```
-
-Performing a partial update using a `PATCH` request can allow any single field to be updated without affecting the other fields. For example, if we only wanted to update the `weight` for a membership:
+A Dynamic Group Membership may be updated using `PUT` or `PATCH` (for a partial update) requests. Again, a `PUT` request requires the entire object to be described, while performing a partial update using a `PATCH` request can allow any single field to be updated without affecting the other fields. For example, if we only wanted to update the `weight` for a membership:
```
gsnider2195, line 607:
```suggestion
A Dynamic Group Membership may be updated using `PUT` or `PATCH` requests. Again, a `PUT` request requires the entire object to be described, while performing a partial update using a `PATCH` request can allow any single field to be updated without affecting the other fields. For example, if we only wanted to update the `weight` for a membership:
```

Same as above comment

nautobot/docs/user-guide/platform-functionality/dynamicgroup.md
```diff
@@ -598,38 +620,35 @@ To delete a Dynamic Group Membership you would send a `DELETE` request to the de
...
 
-+/- 2.3.0
++/- 2.3.0 "Simpler, always-on caching"
     In prior versions of Nautobot, `.members` did not use a cache (and so would always re-evaluate the Dynamic Group), and a separate `.members_cached` property used a configurable Redis cache to temporarily store the evaluated QuerySet. For performance reasons, the behavior of `.members` has been changed to behave as described above. `.members_cached` should be considered deprecated and will likely be removed in a future major release.
 
     Similarly, `.has_member(obj)` was previously non-cached by default, but the Redis cache could be used by invoking it as `.has_member(obj, use_cache=True)`; this too has been changed in this release, and the `use_cache` flag should be considered deprecated as well.
 
-### Working with a given Object's containing Dynamic Groups
+For static-assignment-based Dynamic groups (only!) you can update its member list by assigning directly to its `.members` or calling its `.add_members()` or `remove_members()` methods as desired. Invoking any of these APIs for a filter-based or set-based Dynamic Group is an error and will raise an exception.
```
gsnider2195, line 641:
```suggestion
For static-assignment-based Dynamic groups (only!) you can update its member list by assigning directly to its `.members` or calling its `.add_members()` or `.remove_members()` methods as desired. Invoking any of these APIs for a filter-based or set-based Dynamic Group is an error and will raise an exception.
```

nautobot/extras/models/groups.py
```diff
@@ -891,83 +731,80 @@ def perform_membership_set_operation(self, operator, query, next_set):
...
-        :param weight:
-            Integer weight used to order filtering
+        Args:
+            child (DynamicGroup): child group to add
+            operator (str): DynamicGroupOperatorChoices choice value used to dictate filtering behavior
+            weight (int): Integer weight used to order filtering
         """
+        if self.group_type != DynamicGroupTypeChoices.TYPE_DYNAMIC_SET:
+            if self.filter or self.group_type != DynamicGroupTypeChoices.TYPE_DYNAMIC_FILTER:
+                raise ValidationError(f"{self} is not a dynamic-set group.")
```
HanlinMiao, line 779: I am a little confused about the code here. Why is the second if statement there?

nautobot/docs/user-guide/platform-functionality/dynamicgroup.md
glennmatthews: Agreed, will reword.

nautobot/docs/user-guide/platform-functionality/dynamicgroup.md
glennmatthews: That's the (poorly named) `DynamicGroupMembership` model. I'll see if I can reword this too.

mkdocs.yml
```diff
@@ -494,7 +493,9 @@ nav:
           - Computed Fields: "user-guide/platform-functionality/computedfield.md"
           - Custom Fields: "user-guide/platform-functionality/customfield.md"
           - Custom Links: "user-guide/platform-functionality/customlink.md"
-          - Dynamic Groups: "user-guide/platform-functionality/dynamicgroup.md"
+          - Dynamic Groups:
+              - "user-guide/platform-functionality/dynamicgroup.md"
+              - Static Group Associations: "user-guide/platform-functionality/staticgroupassociation.md"
```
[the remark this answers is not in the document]
glennmatthews: Personally I'm OK with it - what would you suggest?

nautobot/core/models/__init__.py
```diff
@@ -47,9 +47,9 @@ class BaseModel(models.Model):
 
     objects = BaseManager.from_queryset(RestrictedQuerySet)()
     is_contact_associable_model = False  # ContactMixin overrides this to default True
+    is_dynamic_group_associable_model = False  # DynamicGroupMixin overrides this to default True
```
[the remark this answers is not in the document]
glennmatthews: *Technically* the existing comment is correct - the flag is set by `DynamicGroupMixin` (which `DynamicGroupsModelMixin` is a subclass of). Confusing.

nautobot/docs/user-guide/platform-functionality/dynamicgroup.md
```diff
@@ -2,69 +2,107 @@
...
+2. Set-based assignment from other defined groups (for example, "All devices in Group A or Group B but not Group C")
+3. Static assignment directly to the group (for example, "Device XYZ01, Device XYZ02, and Device XYZ04")
+
++++ 1.4.0 "Set-based group assignment"
+    Support for set-based assignment to groups was added.
+
++++ 2.3.0 "Static group assignment"
+    Support for static assignment to groups was added, and the distinction between filter-based and set-based assignment was formally established.
+
+The members of each Dynamic Group are cached and tracked by [Static Group Association](staticgroupassociation.md) database records.
```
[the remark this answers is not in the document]
glennmatthews: No... *all* DynamicGroups are now cached through StaticGroupAssociation records, that's just a hidden implementation detail for all other group types *except* `static` ones.

nautobot/extras/migrations/0112_dynamic_group_group_type_data_migration.py
```diff
@@ -0,0 +1,28 @@
+from django.db import migrations
```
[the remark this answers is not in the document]
glennmatthews: Yup, sorry!

nautobot/core/testing/models.py
```diff
@@ -44,19 +44,20 @@ def test_get_docs_url(self):
...
+                self.assertTrue(hasattr(self.model, "dynamic_groups"))
+                self.assertIsInstance(self.model.objects.first().dynamic_groups, QuerySet)
+                self.assertEqual(self.model.objects.first().dynamic_groups.model, DynamicGroup)
+
+                if DynamicGroup.objects.get_for_model(self.model).exists():
+                    sg = DynamicGroup.objects.get_for_model(self.model).first()
                     self.assertEqual(sg.members.model, self.model)
```
glennmatthews, line 56:
```suggestion
                    dg = DynamicGroup.objects.get_for_model(self.model).first()
                    self.assertEqual(dg.members.model, self.model)
```

nautobot/dcim/signals.py
```diff
@@ -297,15 +297,15 @@ def prevent_adding_tagged_vlans_with_incorrect_mode_or_site(sender, instance, ac
...
-
     logger = logging.getLogger(__name__ + ".ControllerManagedDeviceGroup")
 
     if raw:
         logger.debug("Skipping controller update for imported controller device group %s", instance)
         return
 
+    if instance.parent or instance._original_controller == instance.controller:
+        return
+
```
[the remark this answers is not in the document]
glennmatthews: because in the `raw` case, `instance.parent` may not yet be valid, and accessing it may throw an exception. Thanks for checking!

nautobot/dcim/views.py
```diff
@@ -199,7 +199,7 @@ def get_selected_objects_parents_name(self, selected_objects):
 
 
 class LocationTypeListView(generic.ObjectListView):
-    queryset = LocationType.objects.with_tree_fields()
+    queryset = LocationType.objects.all()
```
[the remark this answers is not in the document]
glennmatthews: IIRC it wasn't hurting anything, but it's not necessary with the smarter table code (core/tables.py, https://github.com/nautobot/nautobot/pull/5845/files#diff-beaf93ca10d047aa03ff73b83b869568163963f57fb65edb09ce777447a8d476R145-R148)

nautobot/extras/models/groups.py
```diff
@@ -528,45 +289,117 @@ def get_filter_fields(self):
...
+        if isinstance(objects_to_add, models.QuerySet):
+            if objects_to_add.model != self.model:
+                raise TypeError(f"QuerySet does not contain {self.model._meta.label_lower} objects")
+        else:
+            for obj in objects_to_add:
+                if not isinstance(obj, self.model):
+                    raise TypeError(f"{obj} is not a {self.model._meta.label_lower}")
 
-        return qs
+        if self.group_type == DynamicGroupTypeChoices.TYPE_STATIC:
```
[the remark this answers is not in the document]
glennmatthews: See the `else` case below. In the static case, we need change logging, so we have to create each membership individually. In the non-static case, we *don't* want change logging, so we can do a `bulk_create` for more efficiency.

nautobot/extras/models/groups.py
```diff
@@ -528,45 +289,117 @@ def get_filter_fields(self):
...
+                )
+        else:
+            # Cached/hidden static group associations, so we can use bulk-create to bypass change logging.
+            sgas = [
+                StaticGroupAssociation(
+                    dynamic_group=self, associated_object_type=self.content_type, associated_object_id=obj.pk
+                )
+                for obj in objects_to_add
+            ]
+            StaticGroupAssociation.all_objects.bulk_create(sgas)
```
glennmatthews, line 372: Hmm... does this risk IntegrityErrors if it creates duplicate records? Do we need to compare `objects_to_add` against existing records first?

nautobot/extras/models/groups.py
glennmatthews: Before this PR, the only way of distinguishing between a "filter-based" DG and a "set-based" DG was by whether it contained any other groups - if it contained them, it was set-based (even if it had a defined `filter` otherwise) and if it didn't, it was filter-based (even if that filter was empty).  Now that we have explicit `group_type` values, I wanted to remain ORM-backward-compatible with the old pattern where every group started off as filter-based, then upon adding its first child group, became set-based.

So:

- if calling add_child and we're not currently set-based:
    - if we already have an explicit filter (`self.filter` is truthy), we can't become set-based
    - if we're a static group (or anything other than filter-based) we also can't become set-based
    - else, "magically" become set-based at this point.

nautobot/docs/user-guide/platform-functionality/staticgroupassociation.md
```diff
@@ -2,12 +2,19 @@
...
+
+!!! info
+    As an implementation detail, to improve performance, other types of Dynamic Groups (filter-based and set-based) also use Static Group Association records as a cache of their member objects. By default, the Nautobot UI, REST API, and Python ORM do *not* expose such records; see below.
+
+## UI
+
+The `/extras/static-group-associations/` UI views normally only show Static Group Associations corresponding to static-assignment-based Dynamic Groups; those corresponding to other types of Dynamic Groups are not displayed. It is possible to view (but not modify) such records in the UI list view by specifying a Dynamic Group of another type as a query parameter but this should be treated as debugging functionality rather than a feature intended for consumption by end users.
```
gsnider2195, line 12: > by specifying a Dynamic Group of another type as a query parameter

Is this a special query parameter for this view or just adding a filter?

glennmatthews: Just as a filter

gsnider2195: Would it make sense then to reword this

> by filtering on a filter-based or set-based Dynamic Group

nautobot/docs/user-guide/platform-functionality/dynamicgroup.md
glennmatthews: Actually, does the `info` admonition immediately below this line make things clear enough as is?

gsnider2195: Yeah I think that's sufficient. Is that a new model that you added or did DynamicGroup always have that for set-based groups?

glennmatthews: That's how it's already been.

mkdocs.yml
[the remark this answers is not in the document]
HanlinMiao: This is nitpicky, but Static Group Association implies that there is also a Dynamic Group Association. What about just `Group Association`?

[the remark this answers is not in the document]
glennmatthews: Discussed in standup and we decided to leave it alone at this time.

HanlinMiao (approved):

nautobot/docs/user-guide/platform-functionality/dynamicgroup.md
gsnider2195: It might make this easier to understand if you mention that the `extras.DynamicGroupMembership` through table is used for the `DynamicGroup.children` field.

nautobot/extras/models/groups.py
```diff
@@ -1008,19 +850,25 @@ def get_siblings(self, include_self=False):
 
         return siblings.exclude(pk=self.pk)
 
+    # TODO: this is an interesting definition of "root node", as a node with no children has is_root() = False??
```
gsnider2195, line 853: This probably just needs to go away in favor of type=dynamic-set

nautobot/extras/models/groups.py
```diff
@@ -1008,19 +850,25 @@ def get_siblings(self, include_self=False):
...
+    # TODO: unused in core
     def is_root(self):
         """Return whether this is a root node (has children, but no parents)."""
         return self.children.exists() and not self.parents.exists()
 
+    # TODO: this is an interesting definition of "leaf node", as a node with no parents has is_leaf() = False??
```
gsnider2195, line 859: Likewise, this should probably just be `type != "dynamic-set" and not self.children.exists()`

gsnider2195: Actually, `type != "dynamic-set"` should be sufficient

nautobot/extras/querysets.py
```diff
@@ -158,7 +158,15 @@ def _get_config_context_filters(self):
...
 class DynamicGroupQuerySet(RestrictedQuerySet):
-    """Queryset for `DynamicGroup` objects that provides a `get_for_object` method."""
+    """Queryset for `DynamicGroup` objects that provides `get_for_object` and `get_for_model` methods."""
+
+    def get_for_model(self, model):
+        """
+        Return all DynamicGroups assignable to the given model class.
+        """
+        concrete_model = model._meta.concrete_model
+        content_type = ContentType.objects.get_for_model(concrete_model)
```
gsnider2195, line 168: Why is concrete_model required here?

gsnider2195 (approved):

nautobot/extras/models/groups.py
glennmatthews: Agreed, but also not touching this for now.

nautobot/extras/querysets.py
glennmatthews: Not sure it is needed, tbh - was just following the pattern in custom field, computed field, and relationship managers/querysets.

nautobot/docs/user-guide/platform-functionality/dynamicgroup.md
glennmatthews: Updated hopefully for clarity.

nautobot/extras/models/groups.py
gsnider2195: Agree on both of these, not worth touching for this PR :)

PR #5850, opened 2024-06-25
Add Celery Worker Details to Job Result Advanced Tab

### LINKED ISSUE #5817: Include worker details in job results
As a platform admin, I want details of the worker to be included in the advanced output of a job result to aid in troubleshooting.

Notably this would be useful for discovering which worker a job ran on.

smk4664:
This could easily be moved to the Additional Data instead. 

# Closes #5817 
# What's Changed

Adds the Celery Worker details (hostname, meta and task_name) to the Advanced Tab.
# Screenshots

Before:
Screenshot 2024-06-25 at 2 22 05 PM
After:
image

# TODO

- [x] Unit, Integration Tests


glennmatthews:
My vote would be for putting it on the Additional Data tab even though that tab's already pretty crowded. Let's get others' input as well though.

nautobot/extras/templates/extras/jobresult.html
```diff
@@ -160,6 +160,30 @@
...
+                <td>Worker Hostname</td>
+                <td>{{ result.worker }}</td>
+            </tr>
+            <tr>
+                <td>Task Name</td>
+                <td>{{ result.task_name }}</td>
+            </tr>
+            <tr>
+                <td>Meta</td>
+                <td>{% include 'extras/inc/json_data.html' with data=result.meta format="json" %}</td>
```
glennmatthews, line 180: Any idea what sorts of information this field may contain? I haven't ever looked at it.

lampwins:
Always seemed weird to have both the advanced and additional data tabs here. Maybe the answer is to just collaps additional data into advanced to be self consistent?

nautobot/extras/templates/extras/jobresult.html
smk4664: I have no idea... Honestly I wasn't really sure what to put here other than just the Hostname, as we really don't get much data with Celery. I can remove Meta as we don't currently use it in Nautobot.

nautobot/extras/templates/extras/jobresult.html
```diff
@@ -80,8 +77,8 @@
     {% include 'extras/inc/jobresult.html' with result=result log_table=log_table %}
 {% endblock content_full_width_page %}
 
-{% block extra_tab_content %}
-    <div role="tabpanel" class="tab-pane" id="json">
+{% block advanced_content_full_width_page %}
```
glennmatthews, line 80: Seems like the contents of this block are already split into left and right columns, can we convert them to using `advanced_content_left_page` and `advanced_content_right_page` instead of `advanced_content_full_width_page` since they're not actually full-width content?

smk4664: There is no `advanced_content_left_page`. But I could add that.

nautobot/extras/templates/extras/jobresult.html
```diff
@@ -80,86 +77,104 @@
...
+        </div>
+        <div class="panel-body">
+            {% include 'extras/inc/json_data.html' with data=result.celery_kwargs format="json" %}
+        </div>
+    </div>
+{% endblock advanced_content_left_page %}
+{% block advanced_content_right_page %}
+    <div class="panel panel-default">
+        <div class="panel-heading">
+            <strong>Worker</strong>
```
glennmatthews, line 109: Now I'm *really* nitpicking, but I don't know that I like this being the first / most prominent panel in the right column. Logically it makes sense (inputs on the left, then worker info, then outputs on the right) but it just seems like info that's less relevant to most users.

**Maybe** put the Worker panel at the bottom of the left column instead, even though that results in some rather unbalanced columns?

nautobot/core/templates/generic/object_retrieve.html
```diff
@@ -163,6 +163,7 @@ <h1>
             <div class="row">
                 <div class="col-md-6">
                     {% include 'inc/object_details_advanced_panel.html' %}
+                    {% block advanced_content_left_page %}{% endblock advanced_content_left_page %}
```
glennmatthews, line 166: Let's document this in `.added` as a new feature for templates to consume if desired. I just searched and it doesn't look like we document the `advanced_content_*` blocks in our existing dev documentation -- it might be good to add to `nautobot/docs/development/core/templates.md` as well though I won't hold you to it.

glennmatthews (approved):

nautobot/extras/templates/extras/jobresult.html
smk4664: The columns could also be potentialy be unbalanced due to the Custom Fields.

smk4664: After discussing in Parking Lot, we are going to keep Worker Data on the right, but remove File Output, as that is already in the main tab. I will update the screenshot with better examples.

nautobot/core/templates/generic/object_retrieve.html
smk4664: Added.

PR #5857, opened 2024-06-27
Fix Display Issue on Dynamic Groups tags field

### LINKED ISSUE #4237: Dynamic Groups: incorrect tags logic
### Environment
* Nautobot version: 1.6.0
* Python version: 3.11
* Database platform, version: PostgreSQL 11
* Middleware(s): N/A

### Steps to Reproduce
1. Create a dynamic group that includes multiple tags.

### Expected Behavior
The query logic to match the text representation, where tag assignment is OR'ed together.

### Observed Behavior
The query logic matches that of other filters (UI, API), where tag assignment is AND'ed together.

smk4664:
# Closes #4237 
# What's Changed

Correct the display on Dynamic Group Query to show that tags are AND instead of OR. As a followup, it could be nice to create a "Tags OR" filter and form field so that we can also do the OR for tags.

# Screenshots

BEFORE:
image

AFTER:
image

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/changes/4237.fixed
+++ b/changes/4237.fixed
@@ -1,0 +1,1 @@
+Fixed display issue with multiple tags filter on dynamic groups. Multiple Tags are now displayed with an AND.
--- a/changes/4237.fixed
+++ b/changes/4237.fixed
@@ -1,1 +1,1 @@
-Fixed display issue with multiple tags filter on dynamic groups. Multiple Tags are now displayed with an AND.
+Fixed display issue with multiple tags filter on dynamic groups. Multiple Tags are now correctly displayed with an AND.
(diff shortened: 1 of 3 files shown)


smk4664:
This will likely need to be back-ported to LTM 1.6

changes/4237.fixed
```diff
@@ -0,0 +1 @@
+Fixed display issue with multiple tags filter on dynamic groups. Multiple Tags are now displayed with an AND.
```
glennmatthews, line 1:
```suggestion
Fixed display issue with multiple tags filter on dynamic groups. Multiple Tags are now correctly displayed with an AND.
```

glennmatthews (approved):
Not sure the priority of backporting this to LTM since it's only a cosmetic bug, but thanks for thinking of that!

PR #5864, opened 2024-06-27
Update Cable connect form to support modular components

### LINKED ISSUE #5790: Add module components to cable connect form
Follow-up to #2101. Need to be able to connect cables to components that are associated with a module, not just devices and virtual machines

gsnider2195:
# Closes #5790 
# What's Changed

- Updated Cable table to display the parent Device of Cables connected to Modules in a Device.
- Updated device and device_id filters for Cables, Interfaces, and other modular device components to recognize components that are nested in Modules in a Device.
- Updated Cable connect form to allow connecting to Console Ports, Console Server Ports, Interfaces, Power Ports, Power Outlets, Front Ports, and Rear Ports that are nested in Modules in a Device.
- Updated Interface `device` filter to allow filtering on Device `name` or `id`.
- Updated `Cable._termination_a_device` and `Cable._termination_b_device` to cache the Device when cables are connected to an Interface or Port of a Module in a Device.

# Screenshots

## Cable connect form

image

image

image

### Before

image

### After

image

# TODO

- [ ] Unit, Integration Tests


nautobot/dcim/models/devices.py
```diff
@@ -979,56 +981,56 @@ def all_console_ports(self):
...
-        return FrontPort.objects.filter(models.Q(device=self) | models.Q(module__in=self.all_modules))
+        return FrontPort.objects.filter(Q(device=self) | Q(module__in=self.all_modules))
 
     @property
     def all_interfaces(self):
         """
         Return all Interfaces that are installed in the device or in modules that are installed in the device.
         """
-        return self.vc_interfaces | Interface.objects.filter(module__in=self.all_modules)
+        return Interface.objects.filter(Q(device=self) | Q(module__in=self.all_modules))
```
gsnider2195, line 1005: Behavior change here. Since we already had the behavior `'interfaces' are installed in this device. 'vc_interfaces' are installed in any member of the virtual chassis` I made some changes to be consistent with that behavior but adding modular component support.

nautobot/dcim/filters/__init__.py
```diff
@@ -1116,26 +1122,39 @@ class Meta:
...
+        if not hasattr(value, "__iter__") or isinstance(value, str):
+            value = [value]
+
+        device_ids = set(str(item) for item in value if is_uuid(item))
+        device_names = set(str(item) for item in value if not is_uuid(item))
+        devices = Device.objects.filter(Q(name__in=device_names) | Q(pk__in=device_ids))
+        all_interface_ids = []
+        for device in devices:
+            all_interface_ids.extend(device.vc_interfaces.values_list("id", flat=True))
+        return Q(pk__in=all_interface_ids)
```
glennmatthews, line 1135: Does this keep the old behavior where if no `devices` are found (because we specified a nonexistent PK or name), the query fails closed?

nautobot/dcim/filters/__init__.py
```diff
@@ -1116,26 +1122,39 @@ class Meta:
...
     def filter_device(self, queryset, name, value):
-        try:
-            devices = Device.objects.filter(**{f"{name}__in": value})
-            all_interface_ids = []
-            for device in devices:
-                all_interface_ids.extend(device.all_interfaces.values_list("id", flat=True))
-            return queryset.filter(pk__in=all_interface_ids)
-        except Device.DoesNotExist:
-            return queryset.none()
+        # Include interfaces belonging to virtual chassis members
```
glennmatthews, line 1138: Comment seems a bit out of place here, should it be moved up to line 1132/1133?

nautobot/dcim/filters/__init__.py
```diff
@@ -1008,12 +1013,13 @@ class InterfaceFilterSet(
...
     RoleModelFilterSetMixin,
 ):
-    # Override device and device_id filters from DeviceComponentModelFilterSetMixin to
+    # Override device and device_id filters from ModularDeviceComponentModelFilterSetMixin to
     # match against any peer virtual chassis members
-    device = MultiValueCharFilter(
+    device = NaturalKeyOrPKMultipleChoiceFilter(
+        queryset=Device.objects.all(),
+        to_field_name="name",
+        label="Device (name or ID)",
```
glennmatthews, line 1021: Can/should we mark the separate `device_id` filter as deprecated now?

nautobot/dcim/filters/__init__.py
```diff
@@ -1116,26 +1122,39 @@ class Meta:
...
-                all_interface_ids += device.all_interfaces.values_list("id", flat=True)
-            return queryset.filter(pk__in=all_interface_ids)
-        except Device.DoesNotExist:
-            return queryset.none()
+        for device in devices:
+            all_interface_ids.extend(device.vc_interfaces.values_list("id", flat=True))
+        return Q(pk__in=all_interface_ids)
+
+    def filter_device_id(self, queryset, name, value):
+        # Include interfaces belonging to virtual chassis members
```
glennmatthews, line 1155: Same comment/question about the location of this comment :)

nautobot/dcim/forms.py
```diff
@@ -3803,6 +3803,14 @@ class ConnectCableToDeviceForm(ConnectCableExcludeIDMixin, NautobotModelForm):
             "rack": "$termination_b_rack",
         },
     )
+    termination_b_module = DynamicModelChoiceField(
```
glennmatthews, line 3806: Can remove the comment at lines 3777-3779 now :)

nautobot/dcim/forms.py
```diff
@@ -3803,6 +3803,14 @@ class ConnectCableToDeviceForm(ConnectCableExcludeIDMixin, NautobotModelForm):
...
+        queryset=Module.objects.all(),
+        label="Module",
+        required=False,
+        query_params={
+            "device": "$termination_b_device",
```
glennmatthews, line 3811: Can/should we filter by `$termination_b_location` as well? For non-deviced modules?

nautobot/dcim/templates/dcim/cable_connect.html
```diff
@@ -229,5 +232,16 @@ <h3>
             $('#id_termination_b_id').find("option").remove();
         });
     });
+
+    // clear the select2 widget for the module when a new device is selected
+    // jQuery required for select2
+    document.addEventListener("DOMContentLoaded", function() {
+        $('select#id_termination_b_device').select2();
+        $('select#id_termination_b_module').select2();
```
glennmatthews, line 240: Is this necessary? I would have thought it'd already be done automatically. Just an order of operations issue?

nautobot/dcim/filters/__init__.py
gsnider2195: This is where it was at before but I'm ok moving it or removing it completely. I think the comment above the filter definitions is sufficient

nautobot/dcim/forms.py
gsnider2195: I tried that but it didn't work since the the `device` and `location` filters are ANDed together and there's never a case where both of those fields will be populated. I could do a custom filter for this purpose to OR those two fields together but I didn't think there was a use case for cabling up a module in storage (not installed in a device)

nautobot/dcim/templates/dcim/cable_connect.html
gsnider2195: This is for the case where you've already selected a device and module but you change your mind and want to select a different device. This will clear out the module when you select a new device. Is there a way to make that automatic?

nautobot/dcim/forms.py
glennmatthews: Would it make sense to enhance the module `location` filter to include `device__location` as part of its capabilities? Fine if the answer is no.

nautobot/dcim/filters/__init__.py
gsnider2195: We just mark deprecated filters with a todo comment, right?

HanlinMiao (approved):
LGTM!

nautobot/dcim/filters/__init__.py
gsnider2195: No, but it should. I'll fix that.

nautobot/dcim/filters/__init__.py
glennmatthews: TODO plus updating its `label`, I believe.

nautobot/dcim/forms.py
gsnider2195: Another option is to use a custom method filter for `location` and make it filter on either the parent device's  location or the module's `location` field but it might be difficult to write an efficient query for that.

nautobot/dcim/templates/dcim/cable_connect.html
gsnider2195: Sorry, you mean the `select2()` initialization? I ran into race conditions before where `DOMContentLoaded` is initializing the select2 component at the same time as this event listener is firing. That's why I'm doing it explicitly here.

nautobot/dcim/forms.py
gsnider2195: The `location` filter would need to be a method filter so we could look up nested modules' locations (`parent_module_bay__device | parent_module_bay__parent_module__parent_module_bay__device` etc). It's not a problem to create that filter if we need it.

nautobot/dcim/templates/dcim/cable_connect.html
glennmatthews: Yeah, that's what I meant. Sounds reasonable.

nautobot/dcim/filters/__init__.py
gsnider2195: This is actually handled by the `Filter.field` in `NaturalKeyOrPKMultipleChoiceFilter` so we don't need to do it in the method.

timizuoebideri1 (approved):

glennmatthews (approved):

PR #5868, opened 2024-06-28
Fix ObjectMetadata API test failure in next

HanlinMiao:
So the value field is represented as a serialized dict of `contact` or `team` object in the API when the `metadata_type` is `TYPE_CONTACT_OR_TEAM`  because `@property value` helper function. And when we pass that data back again, it makes the validation function unhappy because the `value` is supposed to be `None`.
Quick fixes are to:
1. make the `value` helper function to return `None` and everything else kinda falls into place.
2. Disallowing any value other than `None` in our `setter` function for ObjectMetadata of `TYPE_CONTACT_OR_TEAM`
3. Change the `__str__` function


nautobot/extras/models/metadata.py
```diff
@@ -221,20 +221,14 @@ class Meta:
     @property
     def value(self):
         if self.metadata_type.data_type == MetadataTypeDataTypeChoices.TYPE_CONTACT_TEAM:
-            return self.contact or self.team
+            return None
         else:
             return self._value
```
glennmatthews, line 226: This kinda defeats the purpose of shadowing the `_value` field at all, doesn't it though? Should we just change the db field back to `value` and remove the need for the `@property` at all?

nautobot/extras/filters/__init__.py
```diff
@@ -1019,15 +1019,15 @@ class ObjectMetadataFilterSet(NautobotFilterSet):
...
         model = ObjectMetadata
         fields = "__all__"
 
     def filter_value(self, queryset, name, value):
         value = value.strip()
-        query = Q(_value__icontains=value)
+        query = Q(value__icontains=value)
         if not value:
             return queryset
         return queryset.filter(query)
```
glennmatthews, line 1033: This probably doesn't need to be a method filter now, fwiw.

glennmatthews (approved):

PR #5878, opened 2024-07-01
Make sure that `MetadataType` has objects to delete for delete test cases

HanlinMiao:


nautobot/extras/tests/test_views.py
```diff
@@ -1177,6 +1177,19 @@ def setUp(self):
             "choices-MIN_NUM_FORMS": "0",
             "choices-MAX_NUM_FORMS": "1000",
         }
+
```
gsnider2195, line 1180: Useful vscode config for your new machine:

```json
    "files.trimFinalNewlines": true,
    "files.trimTrailingWhitespace": true,
    "files.insertFinalNewline": true,
```

nautobot/extras/tests/test_views.py
```diff
@@ -1178,6 +1178,19 @@ def setUp(self):
             "choices-MAX_NUM_FORMS": "1000",
         }
 
+    def get_deletable_object(self):
```
gsnider2195, line 1181: Probably need the same for `test_api.py`?

HanlinMiao: Yeah, we do have those functions for `test_api.py`

PR #5885, opened 2024-07-03
Consolidate list view bulk action buttons

### LINKED ISSUE #5869: Consolidate list view bulk action buttons
Use a bootstrap split button dropdown (drop up) to consolidate common bulk action buttons in all list views.

The primary action should be edit, the dropdown should contain "update groups" and a line separated delete (in red).

In case of custom bulk actions like the device list view (add components), leave those alone for now. Specifically for add components, convert to a drop up.

Standardize that this primary/edit group is the first set of buttons and all others come after it. Look through core for other places where this needs to be updated.

timizuoebideri1:
# Closes #5869 

# What's Changed

Combines the bulk edit, delete, and static group assignment buttons at the bottom of a list view into a single split dropdown button when appropriate.

# Screenshots

### extras.add_staticgroupassignment without change_{model} or delete_{model}
image

### extras.add_staticgroupassignment, delete_{model}
image
image

### extras.add_staticgroupassignment, change_{model}, and delete_{model}
image
image

### extras.add_staticgroupassignment, change_{model}
image
image

### change_{model}
image

### delete_{model}
image

# TODO

- [ ] Unit, Integration Tests


lampwins:
The delete option in the drop down should remain red, but the rest of the actions in the downdown should use black text and icons.

nautobot/core/templatetags/buttons.py
```diff
@@ -124,6 +126,141 @@ def job_import_url(content_type):
     return import_url
 
 
+def render_tag_attrs(attrs_dict):
+    """Converts tag attributes from a dictionary to a string format suitable for HTML rendering."""
+    return format_html_join(" ", '{}="{}"', [(key, value) for key, value in attrs_dict.items()])
+
+
+@register.inclusion_tag("buttons/consolidated_bulk_action_buttons.html")
+def consolidate_bulk_action_buttons(request, content_type, perms, bulk_edit_url, bulk_delete_url, permissions):
```
gsnider2195, line 135: Can/should we update the signature to only require the request and content type? We can derive permissions from the request.user and the bulk urls from the content type?

changes/5869.changed
```diff
@@ -0,0 +1 @@
+Replaced hardcoded bulk action buttons in the list view with dynamically generated split button dropdowns using the `consolidate_bulk_action_buttons` inclusion template tag.
```
glennmatthews, line 1: This feels overly technical to me. Perhaps something like:

```Combined bulk-edit/bulk-delete/bulk-group-update buttons into a single button with a pop-up menu when appropriate.```

?

File: nautobot/core/templatetags/buttons.py [hunk 3]
```diff
@@ -124,6 +126,140 @@ def job_import_url(content_type):
...
+                """
+            )
+
+    # Render a generic "Bulk Actions" dropup button if the edit button is not present
+    elif bulk_action_button_count > 1:
+        bulk_action_buttons.append(
+            mark_safe(  # noqa: S308 -- suspicious-mark-safe-usage
+                """
+                <button type="button" class="btn btn-sm btn-default dropdown-toggle" data-toggle="dropdown" aria-haspopup="true">
+                    <span class="mdi mdi-plus-thick" aria-hidden="true"></span> Bulk Actions <span class="caret"></span>
```
glennmatthews, line 213: Feels off to me to use `mdi-plus-thick` for the generic "Bulk Actions" button. Can we omit an icon entirely in this case?

nautobot/core/templatetags/buttons.py
```diff
@@ -124,6 +126,140 @@ def job_import_url(content_type):
...
+    primary_button_fragment = child_button_fragment = """
+        <button {attrs}>
+            <span class="{icon}" aria-hidden="true"></span> {label}
+        </button>
+    """
+
+    edit_button_classes = "btn btn-sm btn-warning"
+    delete_button_classes = "btn btn-sm btn-danger"
+    static_group_button_classes = "btn btn-sm btn-primary"
+    static_group_icon = "mdi mdi-group"
```
glennmatthews, line 174: Why do we have a `static_group_icon` but no `edit_button_icon` or `delete_button_icon`?

nautobot/core/templatetags/buttons.py
```diff
@@ -124,6 +126,140 @@ def job_import_url(content_type):
...
+                        "class": edit_button_classes,
+                        "formaction": reverse(bulk_edit_url) + params,
+                        "type": "submit",
+                    }
+                ),
+                icon="mdi mdi-pencil",
+            ),
+        )
+        if bulk_action_button_count > 1:
+            bulk_action_buttons[0] += mark_safe(  # noqa: S308 -- suspicious-mark-safe-usage
```
glennmatthews, line 198: Use `format_html` instead of `mark_safe` please here and below.

nautobot/core/templates/generic/object_list.html
```diff
@@ -214,28 +214,11 @@ <h1>{% block title %}
...
-                                data-toggle="modal"
-                                data-target="#dynamic_group_assignment_modal"
-                                data-objects="selected"
-                                class="btn btn-primary btn-sm">
-                            <span class="mdi mdi-group" aria-hidden="true"></span> Update Group Assignment
-                        </button>
-                    {% endif %}
-                    {% if bulk_edit_url and permissions.change %}
-                        <button type="submit" name="_edit" formaction="{% url bulk_edit_url %}{% if request.GET %}?{{ request.GET.urlencode }}{% endif %}" class="btn btn-warning btn-sm">
-                            <span class="mdi mdi-pencil" aria-hidden="true"></span> Edit Selected
```
glennmatthews, line 231: Should this PR also address the corresponding edit-all/delete-all buttons in this same file or is that tracked by a separate issue?

nautobot/project-static/css/dark.css
```diff
@@ -190,6 +190,12 @@ html[data-theme="dark"] .navbar-nav>li>a:hover {          /* Navbar top level me
   color: rgb(224, 120, 7) !important;
 }
 
+html[data-theme="dark"] .dropdown-menu button:hover,      /* Navbar button item when hovering */
+html[data-theme="dark"] .dropdown-menu>li>button:hover {
+  color: rgb(51, 51, 51) !important;
+  background-color: rgb(196, 204, 210) !important;
```
glennmatthews, line 196: Curious why a dropdown `button` needs a custom background-color but a dropdown `a` (below) doesn't?

nautobot/core/templatetags/buttons.py
gsnider2195: The edit button is either rendered or it isn't. No conditional icon logic needed there. The delete button icon is red so the `text-muted` class isn't needed to make it less contrasting when rendered in the dropdown

nautobot/project-static/css/dark.css
smk4664: The default behavior is to invert the colors that are on the light theme. The issue with buttons, is the light theme displays them as dark, this is likely to match the nav-bar buttons. Inverting them changes them to white, which is not what we want in a dark theme. This overrides the background-color so that it stays the dark grey color.

As to why it is not needed on the dropdown `a`, the light theme displays the `a` with a white background, so inverting is the correct behavior.

nautobot/core/templatetags/buttons.py
smk4664: I think we should make that change. It would reduce the amount of inputs on the templates. As far as having request be in input, I think this should be changed to using `context`, and change the `inclusion_tag` to include `takes_context=True`.

```
@register.inclusion_tag("buttons/consolidated_bulk_action_buttons.html", takes_context=True)
def consolidate_bulk_action_buttons(context, content_type):
```

nautobot/core/templates/generic/object_list.html
gsnider2195: @lampwins do we want to change these buttons as part of this issue?

image

lampwins: No, out of scope.

smk4664 (approved):
LGTM!

glennmatthews (approved):

PR #5897, opened 2024-07-08
Release v1.6.25

gsnider2195:
- [x] Towncrier
- [x] Version bump

### Security

- [#5891](https://github.com/nautobot/nautobot/issues/5891) - Updated `certifi` to `2024.7.4` to address `CVE-2024-39689`. This is not a direct dependency so it will not auto-update when upgrading. Please be sure to upgrade your local environment.

### Dependencies

- [#5897](https://github.com/nautobot/nautobot/pull/5897) - Pinned dev dependency `faker` to `>=0.7.0,<26.0.0` to work around breaking change in v26.0.0 ([faker/#2070](https://github.com/joke2k/faker/issues/2070)).


pyproject.toml
```diff
@@ -175,6 +175,8 @@ rich = "~13.4.2"
 towncrier = "~23.6.0"
 # Watch files for celery and mkdocs reload
 watchdog = "~3.0.0"
+# Factory Boy dependency. Don't upgrade to v26.x until breaking change is fixed https://github.com/joke2k/faker/issues/2070
+faker = ">=0.7.0,<26.0.0"
```
glennmatthews, line 179: This seems like a really wide range to me. Should we set a higher lower-bound?

Also, can you move this up so the dependencies stay in alphabetical order?

glennmatthews (approved):

pyproject.toml
gsnider2195: `>=0.7.0` was copied from factoryboy, which is our only dependency that needs faker

smk4664 (approved):

HanlinMiao (approved):

nautobot/docs/release-notes/version-1.6.md
```diff
@@ -72,6 +72,16 @@ The default Python version for Nautobot Docker images has been changed from 3.7
 As Python 3.7 has reached end-of-life, Nautobot 1.6 and later do not support installation or operation under Python 3.7.
 
 <!-- towncrier release notes start -->
+## v1.6.25 (2024-07-08)
```
gsnider2195, line 75:
```suggestion
## v1.6.25 (2024-07-09)
```

Issue #5716, opened 2024-05-15
Cloud Network Model


lampwins:
As a cloud and/or network administrator, I want to be able to model cloud network within Nautobot, so that I can represent assets managed in these clouds.

Cloud Network:
- name
- description
- parent (FK to self, only need to support to levels)
- type (FK to cloud type)
- account (FK to cloud account)
- prefixes (m2m)
- Extra Config

PR #5900, opened 2024-07-09
Add CloudService UI

timizuoebideri1:
# Closes #5872 
# What's Changed

# Screenshots

# TODO

- [ ] Explanation of Change(s)
- [ ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [ ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example App Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #5916, opened 2024-07-10
Cloud Network UI

smk4664:
# Partial work towards #5716 
# What's Changed

Add UI for Cloud Networks
# Screenshots

image
image

# TODO

- [ ] Unit, Integration Tests


smk4664:
I need to add more screenshots.

changes/5716.added
```diff
@@ -1,2 +1,4 @@
 Added CloudNetwork model and REST API.
-Added CloudNetwork to CircuitTermination model and REST API.
\ No newline at end of file
+Added CloudNetwork to CircuitTermination model and REST API.
+Add CloudNetwork UI.
```
HanlinMiao, line 3: Should probably combine this with the first line.
Something like:
Added CloudNetwork Model, UI, GraphQL and REST API.

changes/5716.added
```diff
@@ -1,2 +1,4 @@
...
+Add CloudNetwork to Prefix View.
```
HanlinMiao, line 4:
```suggestion
Added CloudNetwork to Prefix View.
```

nautobot/cloud/navigation.py
```diff
@@ -46,6 +46,22 @@
...
+                        ],
+                        buttons=(
+                            NavMenuAddButton(
+                                link="cloud:cloudnetwork_add",
+                                permissions=[
+                                    "cloud.add_cloudnetwork",
+                                ],
+                            ),
+                        ),
+                    ),
```
HanlinMiao, line 64: We should think about the ordering of these cloud models on the navigation menu instead of assigning arbitrary weights.
IMO we should order this as the following, but what do you guys think? @nautobot/core 

1. CloudAccount
2. CloudType
3. CloudNetwork
4. CloudService

nautobot/cloud/tables.py
```diff
@@ -39,6 +40,53 @@ class Meta(BaseTable.Meta):
...
+    class Meta(BaseTable.Meta):
+        model = CloudNetwork
+        fields = (
+            "pk",
+            "name",
+            "description",
+            "cloud_type",
+            "cloud_account",
+            "parent",
+            "prefixes",
```
HanlinMiao, line 60: Have we checked if the prefixes column is rendered correctly?
I think ideally the column should render the number of prefixes with a link that will direct you to the filtered list of prefixes by this cloud network right?

nautobot/cloud/tables.py
```diff
@@ -39,6 +40,53 @@ class Meta(BaseTable.Meta):
...
+            "cloud_account",
+            "parent",
+            "actions",
+        )
+
+
+class CloudNetworkPrefixAssignmentTable(BaseTable):
+    cloud_network = tables.Column(
+        verbose_name="Cloud Network",
+        linkify=lambda record: record.cloud_network.get_absolute_url(),
```
HanlinMiao, line 78: Is this still needed? Could we just do `linkify=True`?

nautobot/cloud/tables.py
```diff
@@ -39,6 +40,53 @@ class Meta(BaseTable.Meta):
...
+        accessor="cloud_network.name",
+    )
+    prefix = tables.Column(linkify=True)
+    rd = tables.Column(accessor="vrf.rd", verbose_name="RD")
+    tenant = TenantColumn(accessor="vrf.tenant")
```
HanlinMiao, line 83: Are these columns necessary?

nautobot/cloud/templates/cloud/cloudnetwork_retrieve.html
```diff
@@ -0,0 +1,56 @@
...
+                            {{ object.parent|hyperlinked_object }}
+                        {% else %}
+                            <span class="text-muted">&mdash;</span>
+                        {% endif %}
+                    </td>
+                </tr>
+                <tr>
+                    <td>Description</td>
+                    <td>{{ object.description|placeholder }}</td>
+                </tr>
```
glennmatthews, line 43: Are we missing `extra_config` here?

nautobot/cloud/templates/cloud/cloudnetwork_retrieve.html
```diff
@@ -0,0 +1,56 @@
...
+                        <a href="{% url 'ipam:prefix_list' %}?cloud_networks={{ object.pk }}">{{ prefix_count }}</a>
+                    </td>
+                </tr>
+            </table>
+        </div>
+{% endblock content_left_page %}
+
+{% block content_right_page %}
+	{% include 'panel_table.html' with table=prefix_table heading="Assigned Prefixes" %}
+{% endblock content_right_page %}
```
HanlinMiao, line 56: Wonder if these are redundant with each other.

nautobot/ipam/filters.py
```diff
@@ -266,6 +267,11 @@ class PrefixFilterSet(
         to_field_name="name",
         label="Locations (name or ID)",
     )
+    cloud_networks = NaturalKeyOrPKMultipleChoiceFilter(
+        queryset=CloudNetwork.objects.all(),
+        to_field_name="name",
+        label="Cloud Network (name or ID)",
+    )
```
HanlinMiao, line 274: Should we add this filter to `generic_filter_tests` of `PrefixTestCase`?

nautobot/cloud/templates/cloud/cloudnetwork_retrieve.html
smk4664: Probably 😅

nautobot/cloud/tables.py
smk4664: These are the same columns that are on the other prefixes table, and seem to be good columns for prefixes.

nautobot/cloud/tables.py
smk4664: Yes, this column renders just fine, here is a screenshot. 
image

nautobot/cloud/tables.py
smk4664: No, this is with `linkify=True`.
image

nautobot/cloud/navigation.py
glennmatthews: IIRC the general rule we go with is "more specific, more commonly accessed things first" rather than "order of creation/database population". For example:

image

So from that standpoint I'd propose:

- CloudNetwork
- CloudService
- CloudType
- CloudAccount

nautobot/cloud/tables.py
glennmatthews: Need to consider how this will work if there's a large number of prefixes, both from a performance and display standpoint.

Our *usual* convention is to use a `LinkedCountColumn` in such cases.

nautobot/cloud/forms.py
```diff
@@ -72,6 +75,95 @@ class CloudAccountFilterForm(NautobotFilterForm):
...
+            "description",
+            "cloud_type",
+            "cloud_account",
+            "parent",
+            "prefixes",
+            "extra_config",
+            "tags",
+        ]
+        EXTRA_CONFIG_HELP_TEXT = """
+            Optional user-defined <a href="https://json.org/">JSON</a> data for this integration. Example:
```
glennmatthews, line 114: Does `extra_config` field need to be marked as `required=False` or is that automatic?

nautobot/cloud/tables.py
glennmatthews: That's weird - looks like it's trying to linkify based on the CloudNetworkPrefixAssignment rather than its `cloud_network` attribute. Maybe because it's the first column in the table?

nautobot/cloud/templates/cloud/cloudnetwork_retrieve.html
```diff
@@ -0,0 +1,86 @@
...
+
+{% block content_full_width_page %}
+    {% if object.children.all %}
```
glennmatthews, line 63:
```suggestion
    {% if object.children.exists %}
```

nautobot/cloud/views.py
```diff
@@ -23,6 +33,58 @@ class CloudAccountUIViewSet(NautobotUIViewSet):
...
+    form_class = CloudNetworkForm
+    bulk_update_form_class = CloudNetworkBulkEditForm
+
+    def get_extra_context(self, request, instance=None):
+        context = super().get_extra_context(request, instance)
+        if self.action == "retrieve":
+            prefixes = instance.prefixes.restrict(request.user, "view")
+            prefix_count = prefixes.count()
+            prefix_table = PrefixTable(prefixes.select_related("namespace"))
+            children_table = CloudNetworkTable(instance.children.all())
```
glennmatthews, line 51:
```suggestion
            children_table = CloudNetworkTable(instance.children.restrict(request.user, "view"))
```

nautobot/cloud/views.py
```diff
@@ -23,6 +33,58 @@ class CloudAccountUIViewSet(NautobotUIViewSet):
...
+
+            circuits = (
+                Circuit.objects.restrict(request.user, "view")
+                .filter(
+                    Q(circuit_termination_a__cloud_network=instance.pk)
+                    | Q(circuit_termination_z__cloud_network=instance.pk)
+                )
+                .select_related("circuit_type", "tenant")
+                .prefetch_related("circuit_terminations__location")
```
glennmatthews, line 60: May not be needed as `BaseTable` now does automatic select_related/prefetch_related determination?

nautobot/cloud/tables.py
HanlinMiao: Right `LinkCountColumn` is what I was describing, did not know that class existed.

HanlinMiao (approved):
LGTM as long as we changed the nav menu ordering to what Glenn proposed.

nautobot/ipam/tests/test_filters.py
```diff
@@ -176,6 +176,7 @@ class PrefixTestCase(FilterTestCases.FilterTestCase, FilterTestCases.TenancyFilt
         ["status", "status__id"],
         ["status", "status__name"],
         ["type"],
+        ["cloud_networks", "cloud_networks__name"],
```
glennmatthews, line 179: Let's keep these alphabetized?

nautobot/cloud/templates/cloud/cloudnetwork_retrieve.html
```diff
@@ -0,0 +1,86 @@
...
+                    <td>Extra Config</td>
+                    <td><pre>{{ object.extra_config | render_json }}</pre></td>
+                </tr>
+            </table>
+        </div>
+{% endblock content_left_page %}
+
+{% block content_right_page %}
+	{% include 'panel_table.html' with table=prefix_table heading="Assigned Prefixes" %}
+{% endblock content_right_page %}
```
glennmatthews, line 60: In #5900, the `extra_config` is presented as part of `content_right_page`. Maybe the `prefix_table` should be moved to `content_full_width_page`? Let's discuss and see if we can come up with a consistent UI pattern for these two models.

nautobot/cloud/forms.py
smk4664: It defaults to `null`

PR #5936, opened 2024-07-15
[LTM] Fixed issue in which a save() could be called unnecessarily on child devices

### LINKED ISSUE #5935: Device update times increase in proportion to the number of child devices it has
### Environment
* Nautobot version (Docker tag too if applicable): 1.6.24,  e48f12217ae9 (v2.3.0a1)
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Create a parent Device with N bays (say 12)
2. Create N child devices
3. Measure the time it takes to save the parent device with zero child devices, 1, 2, ... N

### Expected Behavior
If no changes are made to the parent device, or if updates other than to rack/position are made to the parent device, the update time of the parent device should be O(1) vs the number of associated child devices.

### Observed Behavior
Save time increases linearly in proportion to the number of associated child Devices. This is obviously far more noticeable on an underpowered Nautobot instance (which we have one of, in addition to our more performant production instances).

nrnvgh:
…devices.

# Closes #5935 
# What's Changed

Added conditionals around updated child rack/position information so that they're only updated when needed, and that `child_device.save()` is only called if one or both of the values has changed.
# Screenshots

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/models/devices.py
+++ b/nautobot/dcim/models/devices.py
@@ -842,15 +842,15 @@ class Device(PrimaryModel, ConfigContextModel, StatusModel):
         # Update Site and Rack assignment for any child Devices
         devices = Device.objects.filter(parent_bay__device=self)
         for device in devices:
-            save_child_devices = False
+            save_child_device = False
             if device.site != self.site:
                 device.site = self.site
-                save_child_devices = True
+                save_child_device = True
             if device.rack != self.rack:
                 device.rack = self.rack
-                save_child_devices = True
+                save_child_device = True
 
-            if save_child_devices:
+            if save_child_device:
                 device.save()
 
     def create_components(self):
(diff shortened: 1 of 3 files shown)


HanlinMiao (approved):

nautobot/dcim/models/devices.py
```diff
@@ -842,9 +842,16 @@ def save(self, *args, **kwargs):
...
+            save_child_devices = False
+            if device.site != self.site:
+                device.site = self.site
+                save_child_devices = True
+            if device.rack != self.rack:
+                device.rack = self.rack
+                save_child_devices = True
+
+            if save_child_devices:
+                device.save()
```
gsnider2195, line 854: Minor nitpick

```suggestion
            save_child_device = False
            if device.site != self.site:
                device.site = self.site
                save_child_device = True
            if device.rack != self.rack:
                device.rack = self.rack
                save_child_device = True

            if save_child_device:
                device.save()
```

gsnider2195 (approved):

PR #5942, opened 2024-07-16
Fixed Scheduled Jobs not respecting Job Soft / Hard Time Limit.

### LINKED ISSUE #5785: Job Soft Time Limit not being respected after upgrading from 2.2.0 to 2.2.5
### Environment
* Nautobot version (Docker tag too if applicable): 2.2.5
* Python version: 3.11
* Database platform, version: PG 15
* Middleware(s): None

### Steps to Reproduce
1. Have a job imported via a git repo that has a soft time limit set to, say, 15 mins and within the code let it sleep 10 mins before it completes
2. Try and run the job and see that the soft time limit isn't respected and the job is stopped after 5 minutes (the default)

### Expected Behavior
The Soft Time Limit is respected and the job completes after 10 mins

### Observed Behavior
The job is killed due to the Soft Time Limit not being respected and is killed after 5 minutes

```
  File "/usr/local/lib/python3.11/site-packages/billiard/pool.py", line 228, in soft_timeout_sighandler
    raise SoftTimeLimitExceeded()
billiard.exceptions.SoftTimeLimitExceeded: SoftTimeLimitExceeded()
```

timizuoebideri1:
# Closes #5785 
# What's Changed

Added the Job Soft / Hard Time Limit that was absent from the `ScheduledJob.celery_kwargs} dict to fix the bug.

# Screenshots

# TODO

- [ ] Unit, Integration Tests


glennmatthews (commented):
Looks reasonable to me. Is this a fix that is needed in LTM as well?

glennmatthews:
Also, looks like a similar fix might be needed around here? https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/api/views.py#L474

Maybe it'd be possible/desirable to consolidate the code paths here between UI and API?

timizuoebideri1:
> Looks reasonable to me. Is this a fix that is needed in LTM as well?

Tested on LTM and it works just fine

timizuoebideri1:
> Also, looks like a similar fix might be needed around here? https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/api/views.py#L474
> 
> Maybe it'd be possible/desirable to consolidate the code paths here between UI and API?

I suppose copying and pasting the code would be simpler/easier.
More if/else logic would be needed for consolidation because the serializer and form make use of different attribute names for the same fields.

nautobot/extras/utils.py
```diff
@@ -648,3 +650,62 @@ def bulk_delete_with_bulk_change_logging(qs, batch_size=1000):
...
             change_context.reset_deferred_object_changes()
+
+
+def create_schedule(schedule_data, job_kwargs, job_model, user, approval_required, task_queue=None):
+    """
+    This is an internal function to create a scheduled job from API data.
+    It has to handle both once-offs (i.e. of type TYPE_FUTURE) and interval
+    jobs.
+    """
+    from nautobot.extras.models.jobs import ScheduledJob  # avoid circular import
```
glennmatthews, line 661: Would it make more sense to implement this as a method on the `ScheduledJob` model class, similar to how we have `JobResult.enqueue_job()`, have `ScheduledJob.create_schedule()`?

nautobot/extras/utils.py
```diff
@@ -648,3 +650,62 @@ def bulk_delete_with_bulk_change_logging(qs, batch_size=1000):
         finally:
             change_context.defer_object_changes = False
             change_context.reset_deferred_object_changes()
+
+
+def create_schedule(schedule_data, job_kwargs, job_model, user, approval_required, task_queue=None):
```
glennmatthews, line 655: Can we make this signature a closer match to the `enqueue_job` signature, and get rid of the `schedule_data` parameters dict? Something more like:

```py
def create_schedule(
    job_model,
    user,
    *job_args,
    name=None,
    interval=JobExecutionType.TYPE_IMMEDIATELY,
    start_time=None,
    crontab=None,
    approval_required=True,
    task_queue=None,
    **job_kwargs,
):
```

nautobot/extras/utils.py
```diff
@@ -648,3 +650,62 @@ def bulk_delete_with_bulk_change_logging(qs, batch_size=1000):
...
+    elif type_ == JobExecutionType.TYPE_CUSTOM:
+        time = schedule_data.get("start_time")  # doing .get("key", "default") returns None instead of "default"
+        if time is None:
+            # "start_time" is checked against models.ScheduledJob.earliest_possible_time()
+            # which returns timezone.now() + timedelta(seconds=15)
+            time = timezone.now() + timedelta(seconds=20)
+        name = schedule_data["name"]
+    else:
+        time = schedule_data["start_time"]
+        name = schedule_data["name"]
```
glennmatthews, line 676: Let's add some error checking here - for example if `TYPE_IMMEDIATELY`, specifying any `start_time` should be an error, not silently ignored; a `crontab` should only be specified for `TYPE_CUSTOM`, etc.

timizuoebideri1: The serializer validation handles API checks. For the UI, errors are avoided as the form dynamically displays the appropriate fields based on the selected interval.

nautobot/extras/models/jobs.py
```diff
@@ -1074,6 +1074,84 @@ def get_crontab(cls, crontab):
...
+            job_model (JobModel): The job model instance.
+            user (User): The user who is scheduling the job.
+            name (str, optional): The name of the scheduled job. Defaults to None.
+            start_time (datetime, optional): The start time for the job. Defaults to None.
+            interval (JobExecutionType, optional): The interval type for the job execution.
+                Defaults to JobExecutionType.TYPE_IMMEDIATELY.
+            crontab (str, optional): The crontab string for the schedule. Defaults to "".
+            profile (bool, optional): Flag indicating whether to profile the job. Defaults to False.
+            approval_required (bool, optional): Flag indicating if approval is required. Defaults to False.
+            task_queue (str, optional): The task queue for the job. Defaults to None.
```
gsnider2195, line 1108:
```suggestion
            task_queue (str, optional): The task queue for the job. Defaults to None, which will use the configured default celery queue.
```

nautobot/extras/models/jobs.py
```diff
@@ -1074,6 +1074,84 @@ def get_crontab(cls, crontab):
...
+            **job_kwargs: Additional keyword arguments to pass to the job.
+        """
```
gsnider2195, line 1110:
```suggestion

        Returns:
            ScheduledJob instance
        """
```

gsnider2195 (approved):

Issue #4279, opened 2023-08-17
Rack elevation: rack group ID filter uses slug name instead


david-kn:
### Environment
* Nautobot version (Docker tag too if applicable): 1.6.0
* Python version: -
* Database platform, version: -
* Middleware(s): -

### Steps to Reproduce
1. Create two (sub) rack groups with the same slug (e.g. `server-room`), each in different rack parent rack groups (e.g. A and B) (see image)
![image](https://github.com/nautobot/nautobot/assets/14314517/a1123036-7fab-4925-b7e8-672cc44925ec)
2. Add some rack to both of those sub rack groups
2. Open rack elevation of one of that parent rack group (based on Rack group ID)
![image](https://github.com/nautobot/nautobot/assets/14314517/45f50658-e28a-4125-af0f-7f27b200f8ec)

### Expected Behavior
Rack elevation should show only racks from within that particular rack group ID (in the provided example, just one rack should be displayed).

### Observed Behavior
Rack elevation shows rack(s) from also other rack groups which have identical slug name within their structure.

I assume that it's caused by the facts that at some point a search (filter) is not focused on ID but rather slugs. Well, and because slugs in general are not totally unique, this leads to this incorrect state.

jathanism:
Hi @david-kn thanks for the report. Based on the information you provided I am unable to reproduce this.

I recreated the Rack Group hierarchy that you provided, however, I have some questions on how you arrived at filtering by rack elevations.

- How are you navigating to the Rack Elevations view?
- How did you decide to filter objects in the Rack Elevations view?
- What is in the URL after you've clicked any links that result in a filtered view? Can you please provide the URI without the hostname? (e.g. `/dcim/rack-elevations/?group_id=d3bee724-cc6f-496d-b16f-2f78a35a901b`) 

Thank you.

david-kn:
Hi @jathanism,

OK - hopefully this will be already complete information :-) btw. I've just reproduced this case on `demo.nautobot.com` (but I don't know for how long it can last there?)

1. Navigation:  go via `Organization` -> `Site` and then click on that `Rack elevation` icon next to the `Rack groups` table (I rather attach again a print screen).
![image](https://github.com/nautobot/nautobot/assets/14314517/96648c05-cda6-4a45-8ea8-4ba48463b2c4)

[...]

Issue #4207, opened 2023-08-03
Location filterform uses non-unique field `name`


jmcgill298:
### Environment
* Nautobot version (Docker tag too if applicable): 2.0.0-beta.2
* Python version: 3.10
* Database platform, version: psql
* Middleware(s):

### Steps to Reproduce
1. Create locations with duplicate names: Region1->Site1, Region2->Site1 (Two "sites" with the same name but different regions)
2. Create a device in R1->S1 and another device in R2->S1
4. Navigate to Device ListView
5. Filter based on location named Site1

### Expected Behavior

I expected to be able to filter devices belonging to a particular location named Site1

### Observed Behavior

I saw all devices belonging to any location named Site1

This is probably an issue beyond the Device ListView, that is just where I observed the behavior. Filters should use a `to_field_name` that is unique, but it appears that location filters are using `name` which is not unique.

bryanculver:
From Slack, looks like an oversight in https://github.com/nautobot/nautobot/issues/2662

Fixing:
- https://github.com/nautobot/nautobot/blob/next/nautobot/dcim/form_mixins.py#L60-L65
- https://github.com/nautobot/nautobot/blob/next/nautobot/dcim/filters/mixins.py#L67-L71

_should_ do the trick?

jmcgill298:
Additional unexpected behavior transforms filtering of parent locations into the parent name and the name of all descendants in the tree (as opposed to the actual descendants in the tree). For example, having the following Location Tree:

```
Region 1
  Site 1
Region 2
  Site 1
  Site 2
```

and using the filter `{"location": ["Region 2"]}`

Would result in `location__name__in=["Region 2", "Site 1", "Site 2"]`

I would expect this to be `location__in=[<LocationRegion1>,<LocationSite1>,<LocationSite2>]`

glennmatthews:
Related: #2875

glennmatthews:
Likely the change we will make for both #4207 and #4279 is to change the UI behavior so that all `NaturalKeyOrPKMultipleChoiceFilter` in the UI switch their behavior to preferring the `pk` over the `name` when making a selection and updating the URL query params.

PR #5943, opened 2024-07-16
Allow filters on non-unique fields to instead provide the PK by default

### LINKED ISSUE #2875: NaturalKeyOrPKMultipleChoiceFilter should support filtering using a combination of fields (true natural key)

### LINKED ISSUE #5923: Issues with filtering by VLAN on the Locations list page
### Environment
* Nautobot version (Docker tag too if applicable): 2.2.7
* Python version: 3.9.17
* Database platform, version: MySQL
* Middleware(s):

### Steps to Reproduce
1. Open https://demo.nautobot.com/ipam/vlans/
2. Click on any number from the Locations column

and

1. Open https://demo.nautobot.com/dcim/locations/
2. Click on Filter
3. Click on Advanced
4. Select "Tagged VLANs (VID or ID)", "exact", and pick a specific vlan like "dfw01-103-mgmt"
5. Click Apply

### Expected Behavior
The locations page opens and shows locations for that specific vlan

### Observed Behavior
When coming from the VLANs page:
```
<class 'ValueError'>

Field 'vid' expected a number but got '01acc3fb-d449-4f36-9789-df5503206528'.

Python version: 3.11.9
Nautobot version: 2.2.7
```

When applying the filter in the Locations page:
all the locations for all the VLANs with the same VID are shown

smk4664:
# Closes #5923 #4207 #4279
# Partially Resolves #2875 
# What's Changed

Filtering using NaturalKeyOrPKMultipleChoiceFilter would prefer the to_field_name, which in some cases would be a field that lacked uniqueness. This PR allows us to specify that we prefer the PK when filtering in the UI and providing a specific object. This still allows the user to provide the non-unique name/vid/etc through params.
# Screenshots

image
image

Allows Override:
image

# TODO

- [x] Unit, Integration Tests


nautobot/docs/development/core/best-practices.md
```diff
@@ -270,6 +270,7 @@ class UserFilter(NautobotFilterSet):
     - Otherwise, the field **must** be shadowed with a Nautobot `NaturalKeyOrPKMultipleChoiceFilter` which will automatically try to lookup by UUID or `name` depending on the value of the incoming argument (e.g. UUID string vs. name string).
         - This provides an advantage over the default `django_filters.ModelMultipleChoiceFilter` which only supports a UUID (`pk`) value as an input.
     - Fields that use `slug` or some other natural key field instead of `name` can set the `to_field_name` argument on `NaturalKeyOrPKMultipleChoiceFilter` accordingly.
+    - If you use `NaturalKeyOrPKMultipleChoiceFilter` for a non-unique field, you can pass in `prefers_pk=True` to prefer the PK in the Filter Forms over the `to_field_name`, while still allowing the `to_field_name` to be used as a query parameter. Example: `locations/?vlans=2538` will give all locations that have any VLAN name with the VID of 2538.
```
glennmatthews, line 273: Perhaps something like:

```suggestion
    - If you use `NaturalKeyOrPKMultipleChoiceFilter` for a non-unique field, you should pass in `prefers_pk=True` to have the filter prefer the object ID in the Filter Forms over the `to_field_name`, while still allowing the `to_field_name` to be used as a query parameter. This will for example make the UI prefer `locations/?vlans=<uuid>` over `locations/?vlans=<vid>` (`VLAN.vid` is not a globally unique field) while still understanding the latter option as well.
```

nautobot/core/utils/filtering.py
```diff
@@ -114,9 +114,13 @@ def get_filterset_parameter_form_field(model, parameter, filterset=None):
     elif isinstance(field, NumberFilter):
         form_field = forms.IntegerField()
     elif isinstance(field, ModelMultipleChoiceFilter):
+        if getattr(field, "prefers_pk", False):
```
gsnider2195, line 117: This is a tough one because it's just moving the problem around. In 1.x we had separate filters for `vlan__vid` and `vlan__id` and we had the ability to filter on either one in the UI. This is "fixing" the problem by removing the ability to filter on `vlan__vid` from the UI and making it so you can only filter on `vlan__id`. Perhaps if we encounter a NaturalKeyOrPKMultipleChoiceFilter we need to render two form fields?

gsnider2195: Is there a defined criteria for choosing the filters that should have `prefers_pk` or is it on a case by case basis? If it's not consistent should we just make the behavior change global so all `NaturalKeyOrPKMultipleChoiceFilter`s use the PK?

glennmatthews: My thought was any filter that doesn't have a single-field natural key (globally unique `name` or similar) should have `prefers_pk`. Those that have a single-field natural key can reasonably use that field name.

nautobot/core/tests/test_utils.py
```diff
@@ -561,6 +561,12 @@ def test_get_filterset_parameter_form_field(self):
                 form_field.queryset, extras_utils.ChangeLoggedModelsQuery().as_queryset()
             )
 
+        with self.subTest("Test prefers_pk"):
```
gsnider2195, line 564: Can you add a similar subtest for a filter that has `prefers_pk=False`?

changes/5923.added
```diff
@@ -0,0 +1 @@
+Added `prefers_pk` to allow NaturalKeyOrPKFilterMultipleChoiceFilter to have the filter prefer the object ID in the Filter Forms over the `to_field_name`.
```
gsnider2195, line 1:
```suggestion
Added `prefers_pk` keyword argument to NaturalKeyOrPKMultipleChoiceFilter initialization to use the object ID instead of the `to_field_name` when automatically generating a form field for the filter.
```

nautobot/core/filters.py
```diff
@@ -439,6 +439,11 @@ class NaturalKeyOrPKMultipleChoiceFilter(django_filters.ModelMultipleChoiceFilte
     field_class = forms.MultiMatchModelMultipleChoiceField
 
     def __init__(self, *args, prefers_pk=False, **kwargs):
+        """Initialize the NaturalKeyOrPKMultipleChoiceFilter.
+
+        Args:
+            prefers_pk (bool, optional): Prefer PK (ID) over the 'to_field_name'. Defaults to False.
```
glennmatthews, line 445: Wonder if we should rename the arg to `prefers_id` since that's what it actually uses?

nautobot/core/tests/test_utils.py
```diff
@@ -561,6 +561,14 @@ def test_get_filterset_parameter_form_field(self):
...
+            form_field = filtering.get_filterset_parameter_form_field(dcim_models.Device, "location")
+            self.assertEqual("id", form_field.to_field_name)
+            form_field = filtering.get_filterset_parameter_form_field(dcim_models.Location, "vlans")
+            self.assertEqual("id", form_field.to_field_name)
+            form_field = filtering.get_filterset_parameter_form_field(dcim_models.Location, "racks")
```
glennmatthews, line 569:
```suggestion
            # Test prefers_pk=False (default)
            form_field = filtering.get_filterset_parameter_form_field(dcim_models.Location, "racks")
```

glennmatthews (approved):
CI failure looks like we need to mark `MetadataChoice` as `is_metadata_associable_model = False` if it isn't already.

nautobot/core/filters.py
smk4664: I was wondering the same, time for find/replace.

nautobot/core/tests/test_utils.py
```diff
@@ -561,6 +561,15 @@ def test_get_filterset_parameter_form_field(self):
                 form_field.queryset, extras_utils.ChangeLoggedModelsQuery().as_queryset()
             )
 
+        with self.subTest("Test prefers_id"):
+            form_field = filtering.get_filterset_parameter_form_field(dcim_models.Device, "location")
+            self.assertEqual("id", form_field.to_field_name)
+            form_field = filtering.get_filterset_parameter_form_field(dcim_models.Location, "vlans")
+            self.assertEqual("id", form_field.to_field_name)
+            # Test prefers_pk=False (default)
```
gsnider2195, line 569:
```suggestion
            # Test prefers_id=False (default)
```

gsnider2195 (approved):

PR #5946, opened 2024-07-16
Add sorting indicators to list view table headers

### LINKED ISSUE #5348: List view sorting indicator
### As ...

Ozzie - Operator

### I want ...

to be able to sort a list view and see a visual indication of which column is being sorted, and to be able to remove sorting from within the UI

### So that ...

I have all of the information necessary in the UI, without having to reference the URL (similar to how we show a UI element of current filters) and so that I don't have to modify the URL to remove sorting

I sort on a column in a list view and a visual indicator is added to the UI, possibly put an arrow next to the column name to indicate the sort (up arrow for ascending, down arrow for descending). If I click the column title once it's sorted ascending, twice it's sorted descending, and on the third time the sort is removed.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation

### Database Changes

no

### External Dependencies

_No response_

kvncampos:
# Closes #5348

## What's Changed
- Added UI enhancement for orderable columns.
- UP/DOWN arrows now display sorting order.
- Functionality:
  - 1. First click: UP arrow, ASC sort (e.g., `path/?sort=name`).
  - 2. Second click: DOWN arrow, DESC sort (e.g., `path/?sort=-name`).
  - 3. Third click: Default view, no sort (e.g., `path/`).

    
# Screenshots

default_view

asc_view

desc_view


kvncampos:
Added new js file  to project-static/js
Included new code in templates/inc/javascript.html
Added custom-data-attribute: data-column-name for js code

nautobot/project-static/js/table_sorting_indicator.js
```diff
@@ -0,0 +1,51 @@
...
+        const columnName = header.dataset.columnName;
+        clickCounts[columnName] = 0;
+
+        // Add a click event listener to each header
+        header.addEventListener("click", function(event) {
+            // Prevent the default link behavior
+            event.preventDefault();
+
+            // Increment the click count for the column
+            clickCounts[columnName] += 1;
```
glennmatthews, line 17: Unless I'm missing something, this variable isn't actually used for anything - left over from a previous implementation?

nautobot/project-static/js/table_sorting_indicator.js
```diff
@@ -0,0 +1,51 @@
...
+
+            // Update the URL with the new sort parameter
+            window.location.search = searchParams.toString();
+        });
+
+        // Initialize the header with the appropriate sort state indicator
+        const sortParam = new URLSearchParams(window.location.search).get("sort");
+        if (sortParam === columnName) {
+            // If sorted in ascending order, add an up arrow
+            header.innerHTML += " &#8593;";
```
glennmatthews, line 45: Can we use regular unicode characters in HTML/JS yet these days, or do we still need to use these entity codes?

nautobot/project-static/js/table_sorting_indicator.js
kvncampos: good catch! thank you :D

nautobot/project-static/js/table_sorting_indicator.js
kvncampos: believe we can, just added this and works the same.

gsnider2195:
Does this need to be rebased to the `next` branch?

nautobot/project-static/js/table_sorting_indicator.js
```diff
@@ -0,0 +1,46 @@
...
+            const searchParams = new URLSearchParams(currentUrl.search);
+            let sortParam = searchParams.get("sort");
+
+            if (sortParam === columnName) {
+                // If the column is already sorted in ascending order, switch to descending
+                searchParams.set("sort", "-" + columnName);
+            } else if (sortParam === "-" + columnName) {
+                // If the column is already sorted in descending order, reset to default
+                searchParams.delete("sort");
+                window.location.href = window.location.pathname; // Reload the page with the default view
```
gsnider2195, line 25: Do we want this to be added to the browser history or do we want to use a `location.replace` here? A client-side UI wouldn't require a page load to change the table order so the behavior would match `location.replace`, but our current UI behavior is to initiate a page load.

nautobot/project-static/js/table_sorting_indicator.js
```diff
@@ -0,0 +1,46 @@
...
+        });
+
+        // Initialize the header with the appropriate sort state indicator
+        const sortParam = new URLSearchParams(window.location.search).get("sort");
+        if (sortParam === columnName) {
+            // If sorted in ascending order, add an up arrow
+            header.innerHTML += " ↑";
+        } else if (sortParam === "-" + columnName) {
+            // If sorted in descending order, add a down arrow
+            header.innerHTML += " ↓";
```
gsnider2195, line 43: Should these use mdi icons instead of unicode arrows? Here are some examples using `mdi-arrow-up-thin` `mdi-arrow-down-thin` `mdi-arrow-up` and `mdi-arrow-down`

image

kvncampos: these are awesome, I like the thin ones best! thanks for feedback

nautobot/project-static/js/table_sorting_indicator.js
kvncampos: very interesting, thank you! did not realize that, I can see how this would be cumbersome specially with the back button after several sortings.

gsnider2195:
@kvncampos can you rebase this to `next`? If there are still test failures after the rebase I'll work on getting them fixed

nautobot/project-static/js/table_sorting_indicator.js
```diff
@@ -0,0 +1,46 @@
...
+
+            // Update the URL with the new sort parameter
+            window.location.replace(window.location.pathname + "?" + searchParams.toString());
+        });
+
+        // Initialize the header with the appropriate sort state indicator
+        const sortParam = new URLSearchParams(window.location.search).get("sort");
+        if (sortParam === columnName) {
+            // If sorted in ascending order, add an up arrow icon
+            header.innerHTML += ' <i class="mdi mdi-arrow-up-thin"></i>';
```
glennmatthews, line 40: Just checking - is the use of MDI in this way tested and confirmed to work?

gsnider2195: It's working in Firefox and Edge for me

image
image

kvncampos: I was using chrome/safari as well.

glennmatthews: Cool, thanks for confirming!

glennmatthews (approved):

gsnider2195 (approved):

PR #5956, opened 2024-07-18
Fixed lags on the JobResults UI page when viewing logs with 2000+ rows using pagination.

### LINKED ISSUE #5904: Homepage and Job Result detail view bog down when job result contains a large input file
### Environment
* Nautobot version (Docker tag too if applicable): 2.2.7

### Steps to Reproduce
1. Export the 12303 Interface records on demo.nautobot.com to a CSV file (roughly 7 MB in size)
2. Upload this file as input to the CSV Import job on another Nautobot instance and run it to completion (it will fail, this is expected)

### Expected Behavior

Nautobot UI to remain adequately performant

### Observed Behavior

At least the Nautobot home page and the Job Result detail view take a tremendously long time to load while this JobResult record is present, likely due to the replication of the entire CSV file contents into several of the JobResult JSON fields (result, traceback, input args, etc.). Deleting the JobResult appears to restore normal performance.

timizuoebideri1:
# Closes #5904 
# What's Changed

The lag was caused by {% ajax_table "log_table" "extras:jobresult_log-table" pk=result.pk %} attempting to fetch over 10,000 table rows. This process significantly slows down the Django server and the user's browser because converting such a large number of table rows into HTML and then sending this extensive HTML back via Ajax takes a considerable amount of time and displaying it on the user's page consumes a large amount of resources.
# Screenshots

https://github.com/user-attachments/assets/dfc29646-6bf9-4d50-a5c3-e14da3fc5dbd

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/5904.fixed
+++ b/changes/5904.fixed
@@ -1,1 +1,2 @@
 Fixed lags on the JobResults UI page when viewing logs with more than 2000+ rows using pagination.
+Fixed lags on the Bulk Importing of large csv files.
--- a/changes/5904.fixed
+++ b/changes/5904.fixed
@@ -1,0 +1,1 @@
+Fixed lags on the JobResults UI page when viewing logs with more than 2000+ rows using pagination.
--- a/changes/5904.fixed
+++ b/changes/5904.fixed
@@ -1,2 +1,2 @@
-Fixed lags on the JobResults UI page when viewing logs with more than 2000+ rows using pagination.
-Fixed lags on the Bulk Importing of large csv files.
+Fixed performance of JobResults UI when thousands of JobLogEntries are present.
+Fixed performance when Bulk Importing large csv files.
--- a/nautobot/extras/jobs.py
+++ b/nautobot/extras/jobs.py
@@ -20,7 +20,7 @@ from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ObjectDoesNotExist
 from django.core.files.base import ContentFile
-from django.core.files.uploadedfile import InMemoryUploadedFile
+from django.core.files.uploadedfile import UploadedFile
 from django.core.validators import RegexValidator
 from django.db.models import Model
 from django.db.models.query import QuerySet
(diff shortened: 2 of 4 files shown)


glennmatthews:
Looks like a desirable improvement, but not a complete fix for #5904. We should also include the changes identified in https://github.com/nautobot/nautobot/issues/5904#issuecomment-2218320604, and address the case where rather than a large number of logs we instead have an overly large `result` and or `traceback` and or `job_kwargs` field as a result of file uploads.

nautobot/extras/jobs.py
```diff
@@ -539,7 +539,7 @@ def serialize_data(data):
             elif isinstance(value, Model):
                 return_data[field_name] = value.pk
             # FileVar (Save each FileVar as a FileProxy)
-            elif isinstance(value, InMemoryUploadedFile):
+            elif isinstance(value, UploadedFile):
```
timizuoebideri1, line 542: The bug occurred because we only checked for `InMemoryUploadedFile` types, missing `TemporaryUploadedFile` types for large CSV files.

nautobot/extras/views.py
```diff
@@ -1786,9 +1786,16 @@ def get(self, request, pk=None):
             )
         else:
             queryset = instance.job_log_entries.all()
+        # queryset = instance.job_log_entries.none()
```
gsnider2195, line 1789:
```suggestion
```

nautobot/extras/views.py
```diff
@@ -1786,9 +1786,16 @@ def get(self, request, pk=None):
...
         log_table = tables.JobLogEntryTable(data=queryset, user=request.user)
-        RequestConfig(request).configure(log_table)
-        return HttpResponse(log_table.as_html(request))
+        # RequestConfig(request).configure(log_table)
```
gsnider2195, line 1791:
```suggestion
```

changes/5904.fixed
```diff
@@ -0,0 +1,2 @@
+Fixed lags on the JobResults UI page when viewing logs with more than 2000+ rows using pagination.
+Fixed lags on the Bulk Importing of large csv files.
```
gsnider2195, line 2:
```suggestion
Fixed performance of JobResults UI when thousands of JobLogEntries are present.
Fixed performance when Bulk Importing large csv files.
```

gsnider2195 (approved):

PR #5981, opened 2024-07-23
Change CloudService network_account to M2M

### LINKED ISSUE #5978: CloudService to CloudNetwork should be an M2M rather than a FK
Because a CloudService may have zero, one, or many associated CloudNetworks, the relation between these models should be defined as a ManyToManyField with an appropriate through table, rather than the current implementation as a single ForeignKey.

gsnider2195:
# Closes #5978 
# What's Changed

- Changed `CloudService.network_account` foreign key field to a ManyToMany called `CloudService.network_accounts`
- Updated views, filters, tests and docs

# Screenshots

## CloudService list view

image

## CloudService detail view

image

image

## CloudService edit view

image

## CloudService bulk edit view

image

## CloudNetwork list view

image

## CloudNetwork detail view

image

## CloudNetwork edit view

image

## CloudNetwork bulk edit view

image

## Docs

image

image

# TODO

- [x] Unit, Integration Tests


gsnider2195:
Should Cloud Resource Type be the second field in the edit form?

changes/5978.changed
```diff
@@ -0,0 +1 @@
+Changed CloudService `network_account` field from a ForeignKey to a ManyToMany called `network_accounts`.
```
glennmatthews, line 1:
```suggestion
Changed CloudService `cloud_network` field from a ForeignKey to a ManyToMany called `cloud_networks`.
```

nautobot/cloud/forms.py
```diff
@@ -275,7 +276,7 @@ class CloudServiceBulkEditForm(TagsBulkEditFormMixin, NautobotBulkEditForm):
     pk = forms.ModelMultipleChoiceField(queryset=CloudService.objects.all(), widget=forms.MultipleHiddenInput)
     description = forms.CharField(max_length=CHARFIELD_MAX_LENGTH, required=False)
     cloud_account = DynamicModelChoiceField(queryset=CloudAccount.objects.all(), required=False)
-    cloud_network = DynamicModelChoiceField(queryset=CloudNetwork.objects.all(), required=False)
+    cloud_networks = DynamicModelMultipleChoiceField(queryset=CloudNetwork.objects.all(), required=False)
```
glennmatthews, line 279: Does this work as intended in a bulk-edit form? I thought normally for bulk-edit of M2M fields we had an `add_foo` and `remove_foo` form field instead + special form save logic?

nautobot/cloud/forms.py
```diff
@@ -250,8 +250,9 @@ class CloudServiceForm(NautobotModelForm):
         queryset=CloudAccount.objects.all(),
         required=False,
     )
-    cloud_network = DynamicModelChoiceField(
+    cloud_networks = DynamicModelMultipleChoiceField(
         queryset=CloudNetwork.objects.all(),
+        required=False,
```
glennmatthews, line 255: Since it's an M2M now, should we also add a counterpart `cloud_services` field on `CloudNetworkForm`?

nautobot/cloud/models.py
```diff
@@ -202,10 +202,10 @@ class CloudService(CloudResourceTypeMixin, PrimaryModel):
         blank=True,
         null=True,
     )
-    cloud_network = models.ForeignKey(to=CloudNetwork, on_delete=models.PROTECT, related_name="cloud_services")
+    cloud_networks = models.ManyToManyField(to=CloudNetwork, related_name="cloud_services", blank=True)
```
glennmatthews, line 205: Don't we need a specified `through` table so that we can provide a REST API endpoint for managing the M2M?

nautobot/cloud/templates/cloud/cloudservice_retrieve.html
```diff
@@ -43,3 +39,13 @@
...
+
+{% block content_full_width_page %}
+    <div class="panel panel-default">
+        <div class="panel-heading">
+            <strong>Cloud Networks</strong>
+        </div>
+        {% include 'inc/table.html' with table=networks_table %}
+    </div>
+    {% include 'inc/paginator.html' with paginator=networks_table.paginator page=networks_table.page %}
+{% endblock content_full_width_page %}
```
glennmatthews, line 51: 1) Should this be implemented as a separate tab rather than a panel on the main tab?

2) Should we also add a `services_table` to `cloudnetwork_retrieve.html`?

nautobot/core/management/commands/generate_test_data.py
```diff
@@ -374,6 +375,8 @@ def handle(self, *args, **options):
                 "flush", "--no-input", "--database", options["database"], inhibit_post_migrate=inhibit_post_migrate
             )
 
+        cache.delete("nautobot.extras.utils.change_logged_models_queryset")
```
glennmatthews, line 378: Interesting... if this is necessary/relevant, let's add a comment and change fragment about it. Are there any other content-type querysets that have a similar cache we'd need to clear?

nautobot/docs/user-guide/core-data-model/cloud/cloudservice.md
```diff
@@ -1,8 +1,8 @@
...
+In Nautobot, cloud services enable tracking of resources, applications, and data storage provided to users via remote servers. A cloud service represents a specific instance of a service defined by its [cloud resource type](./cloudresourcetype.md). For example, it could describe an EC2 instance of type "c7gn.xlarge" from Amazon EC2 or a Linux virtual machine instance of type "M416ms_v2" created using Microsoft Azure Virtual Machine Service within the user's virtual private cloud.
 
-Each cloud service must be assigned to a [cloud resource type](./cloudresourcetype.md) and a [cloud network](./cloudnetwork.md) and can be optionally assigned to a [cloud account](./cloudaccount.md). Extra configurations can be stored in JSON format in the `extra_config` attribute.
+Each cloud service must be assigned to a [cloud resource type](./cloudresourcetype.md) and can optionally be assigned to a [cloud account](./cloudaccount.md) and one or more [cloud networks](./cloudnetwork.md). Extra configurations can be stored in JSON format in the `extra_config` attribute.
```
glennmatthews, line 5:
```suggestion
Each cloud service must be assigned to a [cloud resource type](./cloudresourcetype.md) and can optionally be assigned to a [cloud account](./cloudaccount.md) and zero or more [cloud networks](./cloudnetwork.md). Extra configurations can be stored in JSON format in the `extra_config` attribute.
```

nautobot/core/management/commands/generate_test_data.py
gsnider2195: I was seeing this when I was running `generate_test_data --flush` on the default db. But I can't reproduce it now so my database may have been in a bad state. Going to revert it for now

```
Creating 100 object changes...
Got a ValidationError for extras | object metadata  deleted by bryanlam: {'changed_object_type': ['content type instance with id 156 does not exist.']}
```

nautobot/cloud/templates/cloud/cloudservice_retrieve.html
gsnider2195: Cloud network has `prefixes` and `circuits` on the main tab. If we're not creating a separate view, the performance impact is the same either way so it's probably better to have in a separate tab. Shall we move the prefixes and circuits for CloudNetwork to a separate tab too?

gsnider2195: Cloud resource types has `services` and `networks` in separate tabs btw

nautobot/cloud/models.py
gsnider2195: That's only necessary if we're doing custom through table fields, right? It would be good to be prepared for that if we think that's going to be a feature that's desired.

gsnider2195:
TODO:

- [ ] Update screenshots
- [ ] Add docs for through table model

nautobot/cloud/models.py
gsnider2195: I went ahead and made this change

nautobot/cloud/templates/cloud/cloudservice_retrieve.html
gsnider2195: I went ahead and added the tabs to CloudNetwork

mkdocs.yml
```diff
@@ -307,6 +307,7 @@ plugins:
         "models/cloud/cloudnetwork.md": "user-guide/core-data-model/cloud/cloudnetwork.md"
         "models/cloud/cloudnetworkprefixassignment.md": "user-guide/core-data-model/cloud/cloudnetworkprefixassignment.md"
         "models/cloud/cloudresourcetype.md": "user-guide/core-data-model/cloud/cloudresourcetype.md"
+        "models/cloud/cloudservicenetworkassignment.md": "user-guide/core-data-model/cloud/cloudservicenetworkassignment.md"
```
HanlinMiao, line 310: Would this go under `cloudservice.md` similar to the order below?

HanlinMiao (approved):

nautobot/cloud/templates/cloud/cloudnetwork_retrieve.html
```diff
@@ -9,19 +9,38 @@
...
-    {% if object.parent is None %}
+    {% if children_count %}
             <li role="presentation"{% if request.GET.tab == 'children' %} class="active"{% endif %}>
                 <a href="{{ object.get_absolute_url }}#children" onclick="switch_tab(this.href, reload=false)" aria-controls="children" role="tab" data-toggle="tab">
-                    Children {% badge object.children.count %}
+                    Children {% badge children_count %}
                 </a>
             </li>
-    {% else %}
-        {{ block.super }}
```
smk4664, line 19: Why was this removed? Weird things happen if the if statement above evaluates to false.

smk4664: Ah, I see, we can have additional cases below it. I still believe we need a final `block super` in case none of the tabs need to load.

nautobot/cloud/templates/cloud/cloudnetwork_retrieve.html
```diff
@@ -85,22 +120,9 @@
     <div class="panel panel-default">
         <div class="panel-heading">
             <strong>Extra Config</strong>
-            {% include 'extras/inc/json_format.html' %}
         </div>
         <div class="panel-body">
-            {% include 'extras/inc/json_data.html' with data=object.extra_config format=format %}
+            <pre>{{ object.extra_config|render_json }}</pre>
```
smk4664, line 125: Why are we making this change?

glennmatthews: Consistency with other templates perhaps? I'm not familiar with what `json_data.html` provides us, TBH.

nautobot/cloud/templates/cloud/cloudnetwork_retrieve.html
glennmatthews: IIRC the block.super is just an empty block, so it shouldn't be needed here.

nautobot/cloud/filters.py
```diff
@@ -146,4 +150,27 @@ class CloudServiceFilterSet(NautobotFilterSet):
...
+
+
+class CloudServiceNetworkAssignmentFilterSet(BaseFilterSet):
+    q = SearchFilter(
+        filter_predicates={
+            "cloud_network__name": "icontains",
+            "cloud_network__description": "icontains",
+            "cloud_service__name": "icontains",
+            "cloud_service__description": "icontains",
+        }
```
glennmatthews, line 163: Just wondering - should this include all of the nested lookups from `CloudServiceFilterSet.q` and `CloudNetworkFilterSet.q`? It feels slightly wrong to me that a search term that matches a CloudService will not necessarily match each its CloudServiceNetworkAssignments, but I don't feel strongly.

nautobot/cloud/forms.py
```diff
@@ -274,13 +290,14 @@ class Meta:
...
-    cloud_account = DynamicModelChoiceField(queryset=CloudAccount.objects.all(), required=False)
-    cloud_network = DynamicModelChoiceField(queryset=CloudNetwork.objects.all(), required=False)
     cloud_resource_type = DynamicModelChoiceField(
         queryset=CloudResourceType.objects.all(),
         query_params={"content_types": [CloudService._meta.label_lower]},
         required=False,
     )
+    cloud_account = DynamicModelChoiceField(queryset=CloudAccount.objects.all(), required=False)
+    add_cloud_networks = DynamicModelMultipleChoiceField(queryset=CloudNetwork.objects.all(), required=False)
+    remove_cloud_networks = DynamicModelMultipleChoiceField(queryset=CloudNetwork.objects.all(), required=False)
```
glennmatthews, line 300: Should we similarly have `add_cloud_services`/`remove_cloud_services` on `CloudNetworkBulkEditForm`? Not blocking.

nautobot/cloud/models.py
```diff
@@ -215,3 +221,17 @@ class Meta:
 
     def __str__(self):
         return self.name
+
+
+@extras_features("graphql")
```
glennmatthews, line 226: Should this support custom_validators and webhooks at least as well?

nautobot/cloud/templates/cloud/cloudnetwork_retrieve.html
gsnider2195: - The json/yaml buttons didn't work when I tested it
- Large data structures were being rendered as a single long line instead of formatted with line breaks and indents

nautobot/cloud/tests/test_api.py
```diff
@@ -155,63 +155,76 @@ def setUpTestData(cls):
...
             cloud_resource_type=cloud_resource_types[0],
         )
         models.CloudService.objects.create(
             name="Deletable Service 2",
-            cloud_network=cloud_networks[1],
             cloud_resource_type=cloud_resource_types[1],
         )
         models.CloudService.objects.create(
             name="Deletable Service 3",
-            cloud_network=cloud_networks[2],
```
glennmatthews, line 181: Was it necessary to remove these but not add some `.cloud_networks.set(...)` in their places?

nautobot/cloud/tests/test_filters.py
```diff
@@ -75,16 +77,26 @@ class CloudNetworkPrefixAssignmentTestCase(FilterTestCases.FilterTestCase):
...
     generic_filter_tests = [
         ("cloud_account", "cloud_account__id"),
         ("cloud_account", "cloud_account__name"),
-        ("cloud_network", "cloud_network__id"),
-        ("cloud_network", "cloud_network__name"),
+        ("cloud_networks", "cloud_networks__id"),
+        ("cloud_networks", "cloud_networks__name"),
         ("cloud_resource_type", "cloud_resource_type__id"),
         ("cloud_resource_type", "cloud_resource_type__name"),
-        ("description",),
```
glennmatthews, line 88: Why removed?

nautobot/cloud/templates/cloud/cloudnetwork_retrieve.html
```diff
@@ -9,19 +9,38 @@
 
 
 {% block extra_nav_tabs %}
-    {% if object.parent is None %}
+    {% if children_count %}
```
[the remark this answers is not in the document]
gsnider2195: I can revert this but I don't think it's consistent with similar patterns we have elsewhere in core for self-referential parent/child relationships

nautobot/cloud/tests/test_views.py
```diff
@@ -118,36 +118,30 @@ def setUpTestData(cls):
...
+            ),
         )
+        cloud_services[0].cloud_networks.set([cloud_networks[0]])
+        cloud_services[1].cloud_networks.set([cloud_networks[1]])
+        cloud_services[2].cloud_networks.set([cloud_networks[2]])
 
         cls.form_data = {
             "name": "New Cloud Service",
             "description": "It is a new one",
+            "cloud_networks": [cloud_networks[1].pk],
```
glennmatthews, line 137: Should there be a similar addition to `CloudNetworkTestCase`?

nautobot/cloud/tests/test_views.py
```diff
@@ -118,36 +118,30 @@ def setUpTestData(cls):
...
             "cloud_resource_type": cloud_resource_types[1].pk,
-            "cloud_network": cloud_networks[1].pk,
             "cloud_account": cloud_accounts[1].pk,
             "extra_config": '{"role": 1, "status": "greetings"}',
             "tags": [t.pk for t in Tag.objects.get_for_model(CloudService)],
         }
 
         cls.bulk_edit_data = {
-            "cloud_network": cloud_networks[2].pk,
```
glennmatthews, line 150: Should we have `add_cloud_networks`/`remove_cloud_networks` in the bulk_edit_data?

nautobot/cloud/templates/cloud/cloudnetwork_retrieve.html
gsnider2195: From next.demo:

image

When you click on the JSON or YAML buttons:

image

When you click

nautobot/cloud/tests/test_filters.py
gsnider2195: This must have been a bad merge resolution. I'll add it back

nautobot/cloud/tests/test_api.py
gsnider2195: I didn't think it was necessary for the deletable objects since `cloud_networks` isn't a required key anymore. But it's simple enough to add it back

glennmatthews: That's fair. I guess the factory objects will have cloud_networks values so this is fine.

nautobot/cloud/templates/cloud/cloudnetwork_retrieve.html
smk4664: Interesting... We use the pattern of json/yaml in other places that have the same model field. I wonder why this one sometimes fails. I personally use the yaml side more often, and I default my views to show the yaml.

nautobot/cloud/templates/cloud/cloudnetwork_retrieve.html
smk4664: Okay. I had an issue where the tabs were still rendering even if they were child objects that couldn't have children. Calling `block.super` resolved it for me.

nautobot/cloud/templates/cloud/cloudnetwork_retrieve.html
gsnider2195: I think we were missing the handling of the `format` query parameter in the view to make this work. See the implementation for ConfigContext [here](https://github.com/nautobot/nautobot/blob/0f0b2b166629d25554f04640a71216d30d9055bb/nautobot/extras/views.py#L192-L211)

nautobot/cloud/models.py
gsnider2195: Looking at the new model checklist, I think the recommended minimum `extras_features` would all work here: `custom_links, custom_validators, export_templates, graphql, and webhooks`

nautobot/cloud/templates/cloud/cloudnetwork_retrieve.html
smk4664: I reproduced it, changing to the count above allows for the removal of this check.  Thanks for confirming Glenn!

nautobot/cloud/templates/cloud/cloudnetwork_retrieve.html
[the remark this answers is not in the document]
smk4664: That's fine, I don't think this is necessary to have, just something that could be nice to have.

nautobot/cloud/templates/cloud/cloudnetwork_retrieve.html
smk4664: I will create a followup ticket to add this back in with the `format` query parameter.

nautobot/core/management/commands/generate_test_data.py
gsnider2195: I ran into this again and I think we should track this in a follow-up issue. We probably need to invalidate the entire cache when we do a database flush. If the nautobot server is running when you run this command, any cached database data will become invalid when the database is flushed.

glennmatthews: Yeah, probably doing a full-on `cache.clear()` would work well enough for now.

nautobot/cloud/models.py
gsnider2195: Actually, not webhooks since those require the model to be change logged.

glennmatthews: Hmm. *Should* this be a ChangeLoggedModel?

nautobot/cloud/tests/test_views.py
gsnider2195: Does the generic test know how to handle that? I'll give it a shot

nautobot/cloud/models.py
gsnider2195: Possibly. None of our similar through table models in IPAM are so I'd say we should track that in a follow-up to fix all of them at once?

nautobot/cloud/tests/test_views.py
glennmatthews: It may not. If it doesn't, no worries, this is fine.

gsnider2195: It works but the generic test doesn't verify that it does anything since it doesn't match a field on the model.

nautobot/cloud/models.py
```diff
@@ -171,7 +171,12 @@ def clean(self):
         # TODO: should we enforce that self.cloud_resource_type.provider == self.cloud_account.provider?
 
 
-@extras_features("graphql")
+@extras_features(
+    "custom_links",
```
glennmatthews, line 175: I *think* we don't need `custom_links` for the through tables since we don't have a UI viewset for them that would display the custom links?

glennmatthews (approved):

PR #6022, opened 2024-07-31
Update object_list.html

jnmms:
Adding != symbol to the filter label if the filter is negated

# Closes #NTC-1956
# What's Changed

Added a != mdi symbol to the filter label if the filter is negated to visualize that the filter is negated "'__n' in" test
{% if '__n' in field.name %}<i class="mdi mdi-not-equal"></i>{% endif %}
# Screenshots

Screenshot in the NTC-1956 case
image

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/core/templates/generic/object_list.html
+++ b/nautobot/core/templates/generic/object_list.html
@@ -87,7 +87,7 @@
                             data-field-type="child"
                             data-field-parent="{{ field.name }}"
                             data-field-value="{{ value.name }}"
-                    >×</span>{{ value.display }}
+                    >×</span>{% if '__n' in field.name %}<i class="mdi mdi-not-equal"></i>{% endif %}{{ value.display }}
                 </li>
                 {% endfor %}
             </ul>


glennmatthews:
Thanks for the pull request!

This looks like a bug we should fix rather than something we need a new feature for - *some* negated filters are already accurately labeled as "Exclude" in the UI, for example:

image

After doing a bit of digging, I *think* the issue is that whenever we have a filter with an explicitly specified `label`, the derived variants of the filter (`__n`, etc.) are just reusing its label without modification, whereas filters *without* an explicitly specified label get django-filter's [default behavior](https://github.com/carltongibson/django-filter/blob/376d4434b7c2186a566864e654c39a2926814e3f/django_filters/filters.py#L129-L133) that correctly annotates the filter label. *Probably* a desirable fix would be for Nautobot (somewhere around [here](https://github.com/nautobot/nautobot/blob/develop/nautobot/core/filters.py#L684) I think) to borrow some code from [django_filters `label_for_filter()`](https://github.com/carltongibson/django-filter/blob/376d4434b7c2186a566864e654c39a2926814e3f/django_filters/utils.py#L308-L328) to correctly set the label of the new filter.

glennmatthews:
image

glennmatthews (approved):

HanlinMiao (approved):

Issue #5985, opened 2024-07-24
Why is serial empty?


RenJiangZhou2163:
### Environment
* Nautobot version (Docker tag too if applicable):    (v2.3.0a1)
* Python version:    3.11
* Database platform, version:  postgresql14
* Middleware(s):

On the devices page, when importing devices in bulk, the device table includes a serial field. After a successful import, it was found that the serial field in the device is empty.

gsnider2195:
I'm not able to reproduce this behavior on either demo site (next.demo.nautobot.com or demo.nautobot.com) using the device bulk import "CSV File upload" or "CSV Text input". Can you provide an example of the CSV data you're using? This is the data I used:

```csv
name,serial,device_type__manufacturer__name,device_type__model,status__name,role__name,location__id
tst01-dist-01,testserial01,Cisco,Catalyst 6509-E,Active,distribution,9e39051b-e968-4016-b0cf-63a5607375de
tst01-edge-01,testserial02,Arista,DCS-7280CR2-60,Active,edge,9e39051b-e968-4016-b0cf-63a5607375de
```

RenJiangZhou2163:
Importing through "CSV Text input" works correctly; importing through "CSV File upload" has issues.

serial,asset_tag,device_type,location,status,name,role
1021C5389580,CA2112140087,3f756e97-32d3-43e1-ae33-08415779f255,e8337ebc-3420-4f75-b41e-5a4e08b8e622,Active,Test-AC-02,WLAN AC

glennmatthews:
Wondering if this is related to #5812 - can you confirm the file encoding of the uploaded file?

RenJiangZhou2163:
It's not related to https://github.com/nautobot/nautobot/issues/5812. 

The uploaded file is UTF-8 encoded.The import was successful except that the serial field of the imported device was empty.

glennmatthews:
OK, thanks for confirming. I still suspect something in the file encoding handling since 1) you stated that the text input works and only the file input shows this behavior and 2) the serial is the first column in the file. Will continue investigating.

RenJiangZhou2163:
[TMP.csv](https://github.com/user-attachments/files/16415666/TMP.csv)
This is the file I imported.

glennmatthews:
Thanks!

```shell
❯ file -k ~/Downloads/TMP.csv 
~/Downloads/TMP.csv: CSV text
- , Unicode text, UTF-8 (with BOM) text, with CRLF line terminators
```

[...]

PR #6025, opened 2024-07-31
Fixed CSV file upload does not correctly handle 'UTF-8 with BOM' encoding

### LINKED ISSUE #5812: CSV file upload doesn't correctly handle "UTF-8 with BOM" CSVs
### Environment
* Nautobot version (Docker tag too if applicable): v2.2.25
* Python version: 3.11
* Database platform, version: PG15
* Middleware(s): N/A

### Steps to Reproduce
1. Affecting [ipam/prefixes](https://demo.nautobot.com/ipam/prefixes/) CSV Import feature
2. Upload a bulk Prefix CSV file to `Prefix` that contains the columns marked as required: prefix, status
image
4. The job will get an error: `Row <id>: prefix : this field is required.`
image

### Expected Behavior
Upload is parsed without error, job completes

### Observed Behavior
job error: `Row <id>: prefix : this field is required.`

### Mitigation
The unspecified column `display` is included in the CSV export (including the `prefix` column separately)
For the CSV Upload to complete I mirrored this, and include both a `display` and `prefix` colum in my CSV upload

### Headers from Prefix CSV Export
`display,id,object_type,natural_slug,prefix,type,network,broadcast,prefix_length,ip_version,date_allocated,description,status__name,role__name,parent__namespace__name,parent__network,parent__prefix_length,namespace__name,tenant__name,vlan__pk,rir__name,created,last_updated,tags`

HanlinMiao:
# Closes #5812, #5985
# What's Changed

Used a fresh .csv file and the failing .csv file in the issue description to test this fix and they both succeeded.
# Screenshots

N/A
# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/core/jobs/__init__.py
+++ b/nautobot/core/jobs/__init__.py
@@ -1,3 +1,4 @@
+import codecs
 import contextlib
 from io import BytesIO
 
@@ -282,7 +283,7 @@ class ImportObjects(Job):
         if not csv_data and not csv_file:
             raise RunJobTaskFailed("Either csv_data or csv_file must be provided")
         if csv_file:
-            csv_bytes = csv_file
+            csv_bytes = codecs.EncodedFile(csv_file, "utf-8", "utf-8-sig")
         else:
             csv_bytes = BytesIO(csv_data.encode("utf-8"))
 
--- a/nautobot/core/jobs/__init__.py
+++ b/nautobot/core/jobs/__init__.py
@@ -283,6 +283,8 @@ class ImportObjects(Job):
         if not csv_data and not csv_file:
             raise RunJobTaskFailed("Either csv_data or csv_file must be provided")
         if csv_file:
+            # data_encoding is utf-8 and file_encoding is utf-8-sig
+            # Bytes read from the original file are decoded according to file_encoding, and the result is encoded using data_encoding.
             csv_bytes = codecs.EncodedFile(csv_file, "utf-8", "utf-8-sig")
         else:
             csv_bytes = BytesIO(csv_data.encode("utf-8"))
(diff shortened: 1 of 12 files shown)


nautobot/core/jobs/__init__.py
```diff
@@ -282,7 +283,7 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
         if not csv_data and not csv_file:
             raise RunJobTaskFailed("Either csv_data or csv_file must be provided")
         if csv_file:
-            csv_bytes = csv_file
+            csv_bytes = codecs.EncodedFile(csv_file, "utf-8", "utf-8-sig")
```
HanlinMiao, line 286: utf-8 is the data encoding and utf-8-sig is the file encoding. "Bytes read from the original file are decoded according to file_encoding, and the result is encoded using data_encoding." From the python [doc](https://docs.python.org/3/library/codecs.html#codecs.EncodedFile)

glennmatthews (commented):
Can we add a unit test, possibly using one of the simple files provided in the issue?

glennmatthews:
Let's confirm that this also fixes #5985 as well.

nautobot/core/jobs/__init__.py
glennmatthews: Maybe add as a comment in the code?

timizuoebideri1: Do we only want a  fix for `utf-8-sig` are there not other encoding we might want to consider?
or better still only accept `UTF-8` and throw an error when encoding is not in `utf-8`

HanlinMiao: I think the current solution takes utf-8 and utf-8-sig which is more desirable than utf-8 alone. From the issue description, it seems like a lot of the users are using utf-8 with BOM, whether advertently or inadvertently. So IMO, we should support both.

timizuoebideri1 (approved):

glennmatthews (approved):

PR #5929, opened 2024-07-12
Add worker status view

gsnider2195:
# Closes #5873 
# What's Changed

# Screenshots

## Collapsed tasks (default)

image

## No workers running

image

## Expanded tasks

image

# TODO

- [x] Explanation of Change(s)
- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [x] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example App Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design
- [ ] Add entry to nav menu?

PR #6026, opened 2024-08-01
Fixed table overflow.

### LINKED ISSUE #5948: Bottom corners of panels overflow borders on hover
### Environment
* Nautobot version (Docker tag too if applicable): v2.2.7 and probably older versions
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Go to https://demo.nautobot.com/dcim/devices/?per_page=25
2. Hover over the bottom row in the table

### Observed Behavior

<br><br>

See prior discussion [here](https://github.com/nautobot/nautobot/pull/5929#discussion_r1679542993). Need to fix overflowing panels before we fix this if we decide it's worth the effort to fix this minor visual annoyance. It's possible this is fixed by moving to a newer UI framework.

timizuoebideri1:
# Closes #5948 
# What's Changed

# Screenshots

BEFORE: (last row do not respect table curved edges)
Screenshot 2024-08-01 at 10 59 46 AM

AFTER: (last row respects table curve edges)
Screenshot 2024-08-01 at 10 58 58 AM
# TODO

- [ ] Unit, Integration Tests


nautobot/project-static/css/base.css
```diff
@@ -299,6 +299,7 @@ table.report th a {
 }
 .panel table {
     margin-bottom: 0;
+    overflow: hidden;
```
glennmatthews, line 302: Did you refer to the previous discussion on https://github.com/nautobot/nautobot/pull/5929#discussion_r1679542993? This change may impact *a lot* of panel tables, so please be sure to manually test in a lot of different places (object list, related object tables in detail view, home page, etc.) especially with small window sizes to make sure this global change is correct/desirable in all cases.

glennmatthews: Pulled this down locally and it looks like it fixes the square-corner problem in Firefox. It doesn't fix other problems like the table overflow described in https://github.com/nautobot/nautobot/pull/5929#discussion_r1679631833 - it looks like the issue with the BulkDelete views on narrow screens is because they aren't wrapping the table in a `<div class="panel panel-default table-responsive">` but only in a `<div class="panel panel-default">`. We probably need to audit our usage of panels containing object tables in general and make sure they're all correctly flagged as responsive tables? But I'm fine with us doing that in a separate issue/PR.

glennmatthews (approved):

PR #6031, opened 2024-08-01
Move IPAddress save() logic to clean()

### LINKED ISSUE #5912: Improve error messages when import process detects duplicate records
### As ...

Ozzie - Operator

### I want ...

to be able to easily identify which records on the import list already exist in Nautobot.

### So that ...

I can modify the import data and remove duplicate records.

the error message produced by the import job gives me human-friendly identifier or a link to the duplicate object.

The current error message is not easy for operators to understand, see example below with an error message produced when attempting to import data containg duplicate IP Address :

```
{
    "exc_message": [
        "duplicate key value violates unique constraint \"ipam_ipaddress_parent_id_host_89330d7e_uniq\"\nDETAIL:  Key (parent_id, host)=(9a7c5310-8e78-4db5-a036-fe6a06bc68f8, \\x0a00c005) already exists.\n"
    ],
    "exc_module": "django.db.utils",
    "exc_type": "IntegrityError"
}
```

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation

### Database Changes

_No response_

### External Dependencies

_No response_

glennmatthews:
# Closes #5912
# What's Changed

RCA of #5912 indicated that this was actually a specific issue with the `IPAddress` model. In the case where a new IPAddress is being created with a `namespace` rather than a `prefix`, the `parent` is not calculated and set until `save()`, which is too late for `validate_unique()` to correctly detect whether the new IPAddress is violating the `(parent, host)` uniqueness constraint.

The solution is to:
1) Move all of the data manipulation previously being done in `IPAddress.save()` up into `IPAddress.clean()` (including the `self.parent` setting logic)
2) Ensure that `IPAddress.save()` always calls `IPAddress.clean()` (handling cases like `IPAddress.objects.create(...)` that wouldn't normally call `clean()`) to ensure that the data manipulation is properly done.

# Screenshots

Before:

image

After:

image

# TODO

- [x] Unit, Integration Tests


nautobot/ipam/models.py
```diff
@@ -1122,8 +1122,8 @@ def clean(self):
 
         closest_parent = self._get_closest_parent()
         # Validate `parent` can be used as the parent for this ipaddress
-        if self.parent and closest_parent:
-            if self.parent != closest_parent:
+        if closest_parent is not None:
+            if self.parent is not None and self.parent != closest_parent:
```
HanlinMiao, line 1126: Why this change here?

HanlinMiao (approved):

nautobot/ipam/models.py
glennmatthews: Merging in the case formerly at lines 1150-1154 below - if self.parent is None but we can find a closest_parent, then we need to update self.parent = closest_parent before continuing (lines 1135-1136).

PR #6056, opened 2024-08-06
Construct delete success message before `obj.delete()`

HanlinMiao:
Fix for https://github.com/nautobot/nautobot-app-bgp-models/issues/214

### DIFF
--- a/nautobot/core/views/generic.py
+++ b/nautobot/core/views/generic.py
@@ -586,15 +586,16 @@ class ObjectDeleteView(GetReturnURLMixin, ObjectPermissionRequiredMixin, View):
 
         if form.is_valid():
             logger.debug("Form validation was successful")
+            msg = ""
 
             try:
+                msg = f"Deleted {self.queryset.model._meta.verbose_name} {obj}"
                 obj.delete()
             except ProtectedError as e:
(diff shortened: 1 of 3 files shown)


nautobot/core/views/generic.py
```diff
@@ -586,15 +586,16 @@ def post(self, request, **kwargs):
 
         if form.is_valid():
             logger.debug("Form validation was successful")
+            msg = ""
 
             try:
+                msg = f"Deleted {self.queryset.model._meta.verbose_name} {obj}"
```
glennmatthews, line 592: Does this need to be inside the try block for a reason? Or could this just be:

```suggestion
            msg = f"Deleted {self.queryset.model._meta.verbose_name} {obj}"

            try:
```

nautobot/core/views/mixins.py
```diff
@@ -743,10 +743,11 @@ def _process_destroy_form(self, form):
         request = self.request
         obj = self.obj
         queryset = self.get_queryset()
+        msg = ""
```
glennmatthews, line 746: Looks unnecessary here - we only use `msg` inside the `transaction.atomic` block below?

glennmatthews (approved):

PR #6058, opened 2024-08-06
Add migration check to upstream testing workflow

### LINKED ISSUE #6055: Add "check migrations" step to upstream testing base
With the upgrade to Django 4 some no-op migrations were necessary so we need to be checking for missing or out of date migrations in the upstream testing workflow

gsnider2195:
# Closes #6055
# What's Changed

See test workflow [here](https://github.com/gsnider2195/nautobot-app-cookiecutter-test/actions/runs/10272428212/job/28424721271) (screenshot below). The test repo doesn't have an `ltm-1.6` branch and it's also not setup to reproduce the specific issue we're trying to catch here where Django 4 created a no-op migration and the unittest job continues to work. 

# Screenshots

image

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/.github/workflows/plugin_upstream_testing_base.yml
+++ b/.github/workflows/plugin_upstream_testing_base.yml
@@ -18,6 +18,7 @@ jobs:
   build:
     strategy:
       fail-fast: false
+      continue-on-error: true
       matrix:
         nautobot-version: ["develop", "ltm-1.6"]
     runs-on: "ubuntu-20.04"
@@ -62,6 +63,14 @@ jobs:
         nautobot-version: ["develop", "ltm-1.6"]
     runs-on: "ubuntu-20.04"
     steps:
+      # Try to download the artifact first so the job can be skipped if the artifact is not found
+      - name: Download docker image artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: "docker_nautobot_${{ matrix.nautobot-version }}"
+          path: /tmp
+      - name: Load docker image
+        run: "docker load --input /tmp/nautobot_${{ matrix.nautobot-version }}.tar"
       - name: "Check out repository code"
         uses: "actions/checkout@v4"
         with:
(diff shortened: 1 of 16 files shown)


.github/workflows/plugin_upstream_testing_base.yml
```diff
@@ -41,14 +42,77 @@ jobs:
...
+          name: "docker_nautobot_${{ matrix.nautobot-version }}"
+          path: "/tmp/nautobot_${{ matrix.nautobot-version }}.tar"
+  unittest:
+    name: "Run Unit Tests"
+    needs: "build"
+    strategy:
+      fail-fast: false
+      matrix:
+        nautobot-version: ["develop", "ltm-1.6"]
+    runs-on: "ubuntu-20.04"
```
glennmatthews, line 64: Can we use 24.04 at this point, at least for new tasks?

.github/workflows/plugin_upstream_testing_base.yml
```diff
@@ -41,14 +42,77 @@ jobs:
           builder: "${{ steps.buildx.outputs.name }}"
           context: "./"
           push: false
-          load: true
```
glennmatthews, line 44: What's this change do?

glennmatthews (approved):

.github/workflows/plugin_upstream_testing_base.yml
gsnider2195: It will probably work fine for the CI base image

.github/workflows/plugin_upstream_testing_base.yml
gsnider2195: > [Load](https://docs.docker.com/engine/reference/commandline/buildx_build/#load) is a shorthand for --output=type=docker (default false)

We're already setting output explicitly so we don't need this line

.github/workflows/plugin_upstream_testing_base.yml
gsnider2195: Testing 24.04 [here](https://github.com/gsnider2195/nautobot-app-cookiecutter-test/actions/runs/10273784541)

PR #6061, opened 2024-08-07
Enhance IPAddress Model to use `get_or_create` method with namespace attribute.

### LINKED ISSUE #5915: Enhance IPAddress Model to use `get_or_create` method with namespace attribute
### As ...

### I want ...

to be able to use the `IPAddress.objects.get_or_create()` method with specifying a namespace,

### So that ...

I can allow the ORM to use the `IPAddress._get_closest_parent()` method for a namespace other than the default.

I can use the following syntax: IPAddress.objects.get_or_create(host="1.1.1.1", mask_length=24, namespace=<my custom namespace>, status=<status object>)

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [X] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

This would be possible outside of making changes to the database

### External Dependencies

none

timizuoebideri1:
# Closes #5915 
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests


nautobot/users/admin.py
```diff
@@ -96,7 +96,7 @@ def has_change_permission(self, request, obj=None):
 
     def has_delete_permission(self, request, obj=None):
         """Disable delete permission."""
-        return False
+        return request.user.is_superuser
```
timizuoebideri1, line 99:
```suggestion
        return False
```

nautobot/ipam/querysets.py
```diff
@@ -397,6 +397,22 @@ def get_queryset(self):
...
+        namespace = kwargs.pop("namespace", None)
+        if not parent:
+            host = kwargs.get("host")
+            mask_length = kwargs.get("mask_length")
+            if not namespace:
+                namespace = get_default_namespace()
+            parent = Prefix.objects.filter(namespace=namespace).get_closest_parent(
+                cidr=f"{host}/{mask_length}", include_self=True
+            )
+            kwargs["parent"] = parent
```
glennmatthews, line 413: Does this fail in a clean/understandable way if the user omitted `host` or `mask_length`, or if they're invalid in some way?

nautobot/ipam/querysets.py
```diff
@@ -397,6 +397,22 @@ def get_queryset(self):
         """
         return super().order_by("host")
 
+    def get_or_create(self, **kwargs):
+        from nautobot.ipam.models import get_default_namespace, Prefix
+
+        parent = kwargs.get("parent")
+        namespace = kwargs.pop("namespace", None)
+        if not parent:
```
glennmatthews, line 405:
```suggestion
        if parent is None:
```

nautobot/ipam/tests/test_models.py
```diff
@@ -962,6 +962,54 @@ def setUp(self):
         self.status = Status.objects.get(name="Active")
         self.prefix = Prefix.objects.create(prefix="192.0.2.0/24", status=self.status, namespace=self.namespace)
 
+    def test_get_or_create(self):
+        """Test that"""
```
glennmatthews, line 966: that what? :)

nautobot/ipam/tests/test_models.py
```diff
@@ -962,6 +962,54 @@ def setUp(self):
...
+            ipaddress.parent.save()
+            ip_obj, created = IPAddress.objects.get_or_create(
+                host=host,
+                mask_length=mask_length,
+                status=status,
+            )
+            self.assertEqual(ip_obj, ipaddress)
+            self.assertFalse(created)
+
+        with self.subTest("Assert create"):
```
glennmatthews, line 1000: Should we also have a test for create with an explicit non-default namespace?

changes/5915.added
```diff
@@ -0,0 +1 @@
+Enhance IPAddress Model to use `get_or_create` method with namespace attribute.
```
glennmatthews, line 1:
```suggestion
Enhanced `IPAddress.objects.get_or_create` method to permit specifying a namespace as an alternative to a parent prefix.
```

nautobot/ipam/tests/test_models.py
```diff
@@ -962,6 +962,64 @@ def setUp(self):
...
+                namespace=namespace,
+                status=status,
+            )
+            self.assertEqual(ip_obj, ipaddress)
+            self.assertFalse(created)
+
+        with self.subTest(
+            "Assert get_or_create utilizes default namespace when retrieving parent if no namespace is provided"
+        ):
+            ipaddress.parent.namespace = default_namespace
```
glennmatthews, line 990: Does this run the risk of colliding with an existing prefix already in default_namespace?

nautobot/ipam/tests/test_models.py
```diff
@@ -962,6 +962,64 @@ def setUp(self):
...
+            ipaddress.delete()
+
+            ip_obj, created = IPAddress.objects.get_or_create(
+                host=host,
+                mask_length=mask_length,
+                status=status,
+            )
+            self.assertEqual(ip_obj.host, host)
+            self.assertEqual(ip_obj.mask_length, mask_length)
+            self.assertEqual(ip_obj._namespace, default_namespace)
```
glennmatthews, line 1010: `_namespace` is an implementation detail - would be better IMHO to assert `ip_obj.parent` and/or `ip_obj.parent.namespace` instead. Same comment below.

nautobot/ipam/tests/test_models.py
```diff
@@ -962,6 +962,64 @@ def setUp(self):
...
+            self.assertTrue(created)
+
+        with self.subTest("Assert create explicitly defining a non default namespace"):
+            ip_obj.delete()
+            ip_obj, created = IPAddress.objects.get_or_create(
+                host=host, mask_length=mask_length, status=status, namespace=namespace
```
glennmatthews, line 1016: Given that you moved `parent` over to `default_namespace`, what prefix are we expecting this ip_obj to be parented under here?

nautobot/ipam/tests/test_models.py
timizuoebideri1: I don't think so; `PrefixFactory.prefix` utilises `UniqueFaker`, which I believe makes unique prefixes, which means there are no two `Prefix` of the same `network/prefix_length`, therefore there is no way to colide.

nautobot/ipam/tests/test_models.py
```diff
@@ -962,6 +962,78 @@ def setUp(self):
...
+                cidr=f"{host}/{mask_length}", include_self=True
+            )
+            self.assertEqual(ip_obj.host, host)
+            self.assertEqual(ip_obj.mask_length, mask_length)
+            self.assertEqual(ip_obj.parent, expected_parent_for_namespace)
+            self.assertEqual(ip_obj.parent.namespace, namespace)
+            self.assertTrue(created)
+
+        with self.subTest("Assert if host/mask_length is not present"):
+            ip_obj, created = IPAddress.objects.get_or_create(status=status)
```
glennmatthews, line 1033: Kind of a strange test if you ask me... if you're really wanting to test `get` with partial information, I'd think that specifying the `host` without a `mask_length` might be more realistic.

nautobot/ipam/tests/test_models.py
```diff
@@ -962,6 +962,78 @@ def setUp(self):
...
+            self.assertTrue(created)
+
+        with self.subTest("Assert create explicitly defining a non default namespace"):
+            ip_obj.delete()
+            ip_obj, created = IPAddress.objects.get_or_create(
+                host=host, mask_length=mask_length, status=status, namespace=namespace
+            )
+            expected_parent_for_namespace = Prefix.objects.filter(namespace=namespace).get_closest_parent(
+                cidr=f"{host}/{mask_length}", include_self=True
+            )
```
glennmatthews, line 1025: Is it safe to assume that a valid parent prefix exists? What if by chance there was only a single `prefix` in the original `namespace`?

In general I think this test might be cleaner, safer, and clearer if it used bespoke data instead of making assumptions about the data created by the factories.

nautobot/ipam/querysets.py
```diff
@@ -397,6 +398,31 @@ def get_queryset(self):
...
+    def get_or_create(self, **kwargs):
+        from nautobot.ipam.models import get_default_namespace, Prefix
+
+        parent = kwargs.get("parent")
+        namespace = kwargs.pop("namespace", None)
+        host = kwargs.get("host")
+        mask_length = kwargs.get("mask_length")
+        # If `host` or `mask_length` is None skip; then there is no way of getting the closest parent;
+        if parent is None and host is not None and mask_length is not None:
+            if not namespace:
```
glennmatthews, line 410:
```suggestion
            if namespace is None:
```

nautobot/ipam/tests/test_models.py
```diff
@@ -962,6 +962,93 @@ def setUp(self):
...
+            self.assertEqual(ip_obj.host, new_host)
+            self.assertEqual(ip_obj.mask_length, mask_length)
+            self.assertEqual(ip_obj.parent, parent)
+            self.assertEqual(ip_obj.parent.namespace, namespace)
+            self.assertTrue(created)
+
+        with self.subTest("Assert if host is not present"):
+            ip_obj, created = IPAddress.objects.get_or_create(host=ipaddress.host, status=ipaddr_status)
+            self.assertEqual(ip_obj.status, ipaddr_status)
+            self.assertFalse(created)
```
glennmatthews, line 1032: Not sure I understand this test, especially given that host *is* present in contradiction of the subTest text?

glennmatthews (approved):

PR #6072, opened 2024-08-07
Removed indentations for PrefixTable in the child prefixes tab of PrefixDetailView, GlobalSearchView and VLANDetailView

### LINKED ISSUE #5970: Inconsistent indentation behavior for tree like model list view's search results
### As ...

Ozzie - Operator

### I want ...

Prefixes & locations on the global search pages (i.e. `/search/?q=XXXX`) to not show indentations without a parent item.  

Additionally, child prefixes should not show indentation with no parent prefix shown (i.e. on `/ipam/prefixes/XXXXX/prefixes/`).

These items were fixed for other pages in #5475 / #4811

### So that ...

The view of prefixes/locations doesn't look wonky and confusingly indented under the wrong parent items.

- I search for a prefix or location using the All Objects search and there is no indentation for a prefix/location when the parent prefix/location isn't shown in the search results

- I look under the child prefixes tab under a prefix and indentation starts at the root level of indentation for any child prefixes.

- [ ] Automation
- [ ] Circuits
- [X] DCIM
- [X] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation

### Database Changes

I don't believe so

### External Dependencies

_No response_

HanlinMiao:
# Closes #5970 
# What's Changed

# Screenshots

Screenshot 2024-08-07 at 2 21 56 PM
Screenshot 2024-08-07 at 2 22 07 PM
Screenshot 2024-08-07 at 2 22 26 PM

# TODO

- [ ] Unit, Integration Tests


nautobot/core/views/__init__.py
```diff
@@ -292,14 +293,18 @@ def get(self, request):
...
                     # For a UIViewSet, .table_class, for an ObjectListView, .table.
                     table = getattr(view_or_viewset, "table_class", getattr(view_or_viewset, "table", None))
 
                     # Construct the results table for this object type
                     filtered_queryset = filterset({"q": form.cleaned_data["q"]}, queryset=queryset).qs
-                    table = table(filtered_queryset, orderable=False)
+                    if model == Prefix:
+                        table = table(filtered_queryset, hide_hierarchy_ui=True, orderable=False)
+                    else:
+                        table = table(filtered_queryset, orderable=False)
```
glennmatthews, line 307: Why's this need to be conditional? Aren't other models (Location, TenantGroup, etc.) also needing this fix? What breaks if we just always `hide_hierarchy_ui=True` here?

nautobot/ipam/views.py
```diff
@@ -1313,7 +1313,7 @@ def get_extra_context(self, request, instance):
                 "namespace",
             )
         )
-        prefix_table = tables.PrefixTable(list(prefixes))
+        prefix_table = tables.PrefixTable(list(prefixes), hide_hierarchy_ui=True)
```
glennmatthews, line 1316: Similar fix needed in `CloudNetworkUIViewSet`? `RoleUIViewSet`? `VRFVIew`?

changes/5970.removed
```diff
@@ -0,0 +1 @@
+Removed indentations for PrefixTable in the child prefixes tab of PrefixDetailView, GlobalSearchView and VLANDetailView
```
glennmatthews, line 1: I'd vote for `.changed` here, it's not removed functionality, just a change in behavior.

glennmatthews (approved):

cdwchriburg:
Upgraded to 2.3.x finally and it looks like this change doesn't fix the issue from #5970.  In fact it makes the view worse.  The child prefixes are now never showing indentation.  This breaks the entire hierarchical view of networks contained inside a container.

I understand this might be more difficult in the main prefix view when there is sorting or a filter applied.  However in this instance under "child prefix", those complex issues should not apply and it should not be a problem to show the hierarchical view of all the child prefixes under this prefix.

glennmatthews:
Sorry that we had a misunderstanding of what was being asked for in #5970. I think it should be possible/reasonable to add/restore appropriate indentation of the "child prefixes" view (`/ipam/prefixes/<uuid>/prefixes/`) at least, if you'd like to open a new issue for that.

PR #6035, opened 2024-08-02
Next to develop pre 2.3.0

glennmatthews:
Just a few minor features...

PR #6080, opened 2024-08-08
Release/2.3.0

glennmatthews:
Most of this was previously reviewed in #6035. See that PR for comments.


nautobot/docs/release-notes/version-2.3.md
```diff
@@ -0,0 +1,345 @@
+<!-- markdownlint-disable MD024 -->
+
+# Nautobot v2.3
+
+This document describes all new features and changes in Nautobot 2.3.
+
+## Upgrade Actions
+
+### Administrators
```
bryanculver, line 9: We need to add a stanza message around the change of the unadorned Docker image moving to 3.12 and Apps compatibility. We should recommend pulling a specific python version of the Docker image, at least for upgrading but in general moving forward.

bryanculver (changes requested):
Requesting changes to upgrade docs to call out Docker Python version pinning.

nautobot/docs/release-notes/version-2.3.md
glennmatthews: See #6082.

bryanculver (approved):

HanlinMiao (approved):

PR #6082, opened 2024-08-08
Add notes about Python 3.12 and Django 4.2 to release-note

glennmatthews:
image

### DIFF
--- a/nautobot/docs/release-notes/version-2.3.md
+++ b/nautobot/docs/release-notes/version-2.3.md
@@ -22,11 +22,6 @@ Administrators should plan to take these actions during or immediately after upg
 
     If you need to stay with a given Python version for the time being, you must make sure that you're relying on an appropriately specific image tag, such as `2.3-py3.11`, `stable-py3.10`, etc.
 
-- Django 4.2 is now required by Nautobot, replacing the previous Django 3.2 dependency.
-
-!!! warning "Django 4"
-    Django 4 includes a small number of breaking changes compared to Django 3. In our experience, most Apps have required few (or zero) updates to be Django 4 compatible, but please be sure to verify your Apps against this upgrade in a test or staging environment before updating Nautobot in production.
-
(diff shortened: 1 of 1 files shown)


lampwins:
Should the docker notice move down to the developer section? Would also help to make the wall of yellow a little less scary.

nautobot/docs/release-notes/version-2.3.md
```diff
@@ -10,18 +10,34 @@ This document describes all new features and changes in Nautobot 2.3.
 
 Administrators should plan to take these actions during or immediately after upgrade from a previous version. New installations should also take note of these actions where appropriate.
 
+- Python 3.12 is now the default recommended version of Python.
+
+!!! warning "Python 3.12"
+    Because Nautobot prior to 2.3.0 did not declare support for Python 3.12, most Apps similarly needed to previously declare an upper bound of Python 3.11 for their own compatibility. Therefore, older versions of most Apps **will not be installable** under Python 3.12. Before migrating your Nautobot environment to Python 3.12, it is your responsibility to confirm that all relevant Apps in your environment are also compatible and installable.
+
+    There is a minor "chicken-and-egg" problem here in that Apps generally cannot declare support for a new Python version before Nautobot itself publishes a release that does so; therefore, as of the 2.3.0 Nautobot release day, most Apps have not yet been updated to declare support for Python 3.12. We'll be working in the following days to promptly update our supported Apps as needed, so stay tuned.
```
bryanculver, line 18: Does this mean that Apps may need to drop support of Nautobot 2.2 because it's not compatible with 3.12?

I could see Poetry/pip complaining about resolving:
```
python = ">=3.8,<3.13"
nautobot = "^2.0.0"
```

HanlinMiao (approved):

glennmatthews:
> Should the django notice move down to the developer section? Would also help to make the wall of yellow a little less scary.

Could do, my thinking was it's important for admins to be aware that this update brings in potentially breaking changes that may impact their deployed apps. Maybe not critical?

nautobot/docs/release-notes/version-2.3.md
glennmatthews: Appears fine for Poetry at least, it just resolves to nautobot 2.3.0b1 since that supports the full python range.

bryanculver: LGTM

PR #6101, opened 2024-08-13
Updated CloudAccount UI: Set the secrets_group form field to be optio…

### LINKED ISSUE #6096: CloudAccount UI form should not flag secrets_group as required
### Environment

### Steps to Reproduce
1. Create a Manufacturer "AWS"
2. Try to create a CloudAccount "AWS Primary" with no Secrets Group associated

### Expected Behavior

The `CloudAccount.secrets_group` database field is `blank=True, null=True` so the UI should not consider it mandatory.

### Observed Behavior

image

timizuoebideri1:
…nal.

# Closes #6096 
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/cloud/tests/test_forms.py
+++ b/nautobot/cloud/tests/test_forms.py
@@ -1,9 +1,0 @@
-from django.test import TestCase
-
-from nautobot.cloud.forms import CloudAccountForm
-
-
-class CloudAccountTest(TestCase):
-    def test_secrets_group_is_not_required(self):
-        """Assert Secrets Group form field is not required: Fix for https://github.com/nautobot/nautobot/issues/6096"""
-        self.assertFalse(CloudAccountForm().fields["secrets_group"].required)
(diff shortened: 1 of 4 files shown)


nautobot/cloud/tests/test_forms.py
glennmatthews, line 1: I feel like the better test here would be to update the data in `test_views.py` such that it attempts and succeeds to create/update a CloudAccount both with and without with an associated SecretsGroup.

nautobot/cloud/tests/test_views.py
```diff
@@ -36,6 +37,22 @@ def setUpTestData(cls):
             "comments": "New comments",
         }
 
+    def test_post_without_secrets_group(self):
+        """Assert Secrets Group form field is not required: Fix for https://github.com/nautobot/nautobot/issues/6096"""
```
timizuoebideri1, line 41: NB: This test is only for testing without secrets group; the test for with a secret group has already been tested in the `setUpTestData` method.

glennmatthews (approved):

PR #6102, opened 2024-08-13
Updated ContactAssociation API: Set the role field to be required.

### LINKED ISSUE #6097: Contact Association API schema incorrectly labels role as not required
### Environment
* Python version: 3.11.9
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Try and create a contact association via API based on provided schema:

Screenshot 2024-08-12 at 12 14 04 PM

### Expected Behavior
I expected role to be required since it's required in the UI

Screenshot 2024-08-12 at 12 15 56 PM

### Observed Behavior
Role is not required, but responds with error:

```
{
  "role": [
    "This field cannot be null."
  ]
}
```

timizuoebideri1:
# Closes #6097 
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/tests/test_serializer.py
glennmatthews, line 1: I don't think this needs to be a new standalone test file, instead it belongs in `test_api.py` if we need a dedicated test case for this at all?

nautobot/extras/tests/test_serializer.py
```diff
@@ -0,0 +1,10 @@
+from django.test import TestCase
+
+from nautobot.extras.api.serializers import ContactAssociationSerializer
+
+
+class ContactAssociationTest(TestCase):
+    def test_secrets_group_is_not_required(self):
+        """Assert ContactAssociation serializer role field is required: Fix for https://github.com/nautobot/nautobot/issues/6097"""
+        ContactAssociationSerializer
```
glennmatthews, line 9: ?

nautobot/extras/tests/test_serializer.py
timizuoebideri1: I suppose I should delete the file because it has more to do with the inaccurate API schema that was generated than it does with actually submitting an API call.

glennmatthews (approved):

PR #6107, opened 2024-08-14
Fixed `Device` model `clean()` validation logic to allow user to specify a software version on a device without specifying any software image files.

### LINKED ISSUE #5494: [Nautobot 2.2] Unable to add Software to Device without also specifying Software Image
### Environment
* Nautobot version (Docker tag too if applicable): 2.2.0-beta.1
* Python version: 3.11

### Steps to Reproduce
1. Create Software object
2. Try to assign Software object to a Device by editing existing Device object
3. Try saving the object

### Expected Behavior

Save operation should succeed and Software object should be associated with the Device.

### Observed Behavior

Save operation fails with form error stating that no software image files for software version, matching device's device type, are found.

image

HanlinMiao:
# Closes #5494 
# What's Changed

1. Fixed `Device` model `clean()` validation logic to allow user to specify a software version on a device without specifying any software image files.
2. Added validation logic to `DeviceForm` `clean()` method to raise a validation error if there is any invalid software image file specified.

# Screenshots

N/A
# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/forms.py
+++ b/nautobot/dcim/forms.py
@@ -2064,6 +2064,35 @@ class DeviceForm(LocatableModelFormMixin, NautobotModelForm, TenancyForm, LocalC
         if position:
             self.fields["position"].widget.choices = [(position, f"U{position}")]
 
+    def clean(self):
+        super().clean()
+
+        device_type = self.cleaned_data["device_type"]
+        software_version = self.cleaned_data["software_version"]
+        software_image_files = self.cleaned_data["software_image_files"]
+
+        # If software version is specified and any software image file is specified, validate that
+        # each of the software image files belongs to the specified software version's software image files
+        # that match the device's device type or is a default image
+        if software_version and software_image_files:
+            for image_file in software_image_files:
+                if not any(
+                    (
+                        software_version.software_image_files.filter(
+                            device_types=device_type, pk=image_file.pk
+                        ).exists(),
+                        software_version.software_image_files.filter(default_image=True, pk=image_file.pk).exists(),
+                    )
+                ):
+                    raise ValidationError(
+                        {
+                            "software_image_files": (
+                                f"Software image file {image_file} for version '{software_version}' is not "
+                                f"valid for device type {device_type}."
+                            )
+                        }
+                    )
+
     def save(self, *args, **kwargs):
         instance = super().save(*args, **kwargs)
         instance.vrfs.set(self.cleaned_data["vrfs"])
--- a/nautobot/dcim/forms.py
+++ b/nautobot/dcim/forms.py
@@ -2068,26 +2068,17 @@ class DeviceForm(LocatableModelFormMixin, NautobotModelForm, TenancyForm, LocalC
         super().clean()
 
         device_type = self.cleaned_data["device_type"]
-        software_version = self.cleaned_data["software_version"]
         software_image_files = self.cleaned_data["software_image_files"]
 
-        # If software version is specified and any software image file is specified, validate that
-        # each of the software image files belongs to the specified software version's software image files
-        # that match the device's device type or is a default image
-        if software_version and software_image_files:
+        # If any software image file is specified, validate that
+        # each of the software image files belongs to the device's device type or is a default image
+        if software_image_files:
             for image_file in software_image_files:
-                if not any(
-                    (
-                        software_version.software_image_files.filter(
-                            device_types=device_type, pk=image_file.pk
-                        ).exists(),
-                        software_version.software_image_files.filter(default_image=True, pk=image_file.pk).exists(),
-                    )
-                ):
+                if device_type not in image_file.device_types.all() and not image_file.default_image:
                     raise ValidationError(
                         {
                             "software_image_files": (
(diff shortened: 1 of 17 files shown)


nautobot/dcim/forms.py
```diff
@@ -2064,6 +2064,35 @@ def __init__(self, *args, **kwargs):
...
+    def clean(self):
+        super().clean()
+
+        device_type = self.cleaned_data["device_type"]
+        software_version = self.cleaned_data["software_version"]
+        software_image_files = self.cleaned_data["software_image_files"]
+
+        # If software version is specified and any software image file is specified, validate that
+        # each of the software image files belongs to the specified software version's software image files
+        # that match the device's device type or is a default image
```
glennmatthews, line 2076: Why do we need to duplicate the logic from the model clean here in the form? Is the model clean insufficient to provide the desired UI behavior?

nautobot/dcim/models/devices.py
```diff
@@ -818,21 +818,30 @@ def clean(self):
...
+                if not any(
+                    (
+                        self.software_version.software_image_files.filter(
+                            device_types=self.device_type, pk=image_file.pk
+                        ).exists(),
+                        self.software_version.software_image_files.filter(
+                            default_image=True, pk=image_file.pk
+                        ).exists(),
+                    )
+                ):
```
glennmatthews, line 836: This feels backwards/inefficient to me. Shouldn't we check something directly on `image_file`, like maybe `if self.device_type not in image_file.software_version.device_types.all()` or the like?

nautobot/dcim/tests/test_forms.py
```diff
@@ -43,6 +45,15 @@ def setUpTestData(cls):
...
         cls.device_role = Role.objects.get_for_model(Device).first()
+        cls.software_version_contains_no_valid_image_for_device_type = SoftwareVersion.objects.create(
+            platform=cls.platform,
+            version="New version 1.0.0",
+            status=Status.objects.get_for_model(SoftwareVersion).first(),
+        )
+        cls.software_version = SoftwareVersion.objects.first()
+        cls.software_image_files = SoftwareImageFile.objects.exclude(
+            software_version=cls.software_version, default_image=True
+        )
```
glennmatthews, line 56: Is this logic right? Or should it be `.exclude(software_version=...).exclude(default_image=True)`?

nautobot/dcim/forms.py
HanlinMiao: Yeah, I think we have to validate the image file in the `form.clean()` before it makes to `model.clean()` in the editing case.
Because `self.software_image_files` attribute is not updated yet with the new data from the form in `model.clean()`. It will still have outdated but valid image files and it will pass the validation logic in `model.clean()`. The device will have the latest but invalid image files as a result.

glennmatthews: Ah, because `software_image_files` is a many-to-many, I'm guessing? That makes sense.

nautobot/dcim/models/devices.py
```diff
@@ -818,21 +818,20 @@ def clean(self):
...
-            raise ValidationError(
-                {
-                    "software_version": (
-                        f"No software image files for version '{self.software_version}' are "
-                        f"valid for device type {self.device_type}."
+        # If any software image file is specified, validate that
+        # each of the software image files belongs to the device's device type or is a default image
+
+        if self.software_image_files.exists():
+            for image_file in self.software_image_files.all():
```
glennmatthews, line 825: The `if` is probably unnecessary here since the `for` will just be a no-op if it's an empty queryset?

nautobot/dcim/models/devices.py
```diff
@@ -818,21 +818,20 @@ def clean(self):
...
+                if self.device_type not in image_file.device_types.all() and not image_file.default_image:
```
glennmatthews, line 826: Might be marginally more efficient (short-circuit logic) to swap these around so that the inexpensive `image_file.default_image` check comes first?

nautobot/dcim/models/devices.py
```diff
@@ -818,21 +818,20 @@ def clean(self):
...
-        ):
-            raise ValidationError(
-                {
-                    "software_version": (
-                        f"No software image files for version '{self.software_version}' are "
-                        f"valid for device type {self.device_type}."
+        # If any software image file is specified, validate that
+        # each of the software image files belongs to the device's device type or is a default image
+
+        if self.software_image_files.exists():
```
glennmatthews, line 824: It occurs to me that it'd be "cleaner" if we had a proper M2M model for device-to-software-image-file, as really this logic belongs on the `clean()` for *that* model, rather than on Device itself, so that something like `device.software_image_files.set([invalid_file])` would itself raise the validation error independent of needing to call `device.validated_save()`. This is probably a good step in the right direction as-is though.

glennmatthews (approved):

nautobot/dcim/models/devices.py
HanlinMiao: https://github.com/nautobot/nautobot/issues/6110

glennmatthews:
https://github.com/mkdocstrings/python/issues/181

Issue #6120, opened 2024-08-20
Status field for VRFs in the IPAM module


jnmms:
### As ...

Ozzie - Operator

### I want ...

A status field for VRF objects in the IPAM module

![image](https://github.com/user-attachments/assets/9cc52c83-5d25-498a-a788-16ef1cc2dc05)

### So that ...

So that it can reflect reality, where it in the config of for example Nexus 7700 series equipment is possible to issue a shutdown command for the vrf.

### I know this is done when...

You can see, add and edit status on a VRF and also bulk edit/change status for VRF objects

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [X] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Apps (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

VRF model needs to be changed to include a relation to the StatusField datamodel

glennmatthews:
Thanks for the feature request! As described, this might be better implemented as an attribute of the `VRFDeviceAssignment` model rather than the `VRF` model itself, perhaps?

jnmms:
Yea it might be, I’m not so fluent inside the code and design, but I will try and take a look, and will se if I can come up with a PR myself – but not sure if I’m the right one to do this 😉

I will definitely take this a a pointer on where to look..


Venlig hilsen / Best regards

Martin Schyth
Netværks specialist
Network Engineering & Automation

Direct +4563639550
Mobile +4521685952
***@***.******@***.***>

***@***.***
JN Data A/S

|

Havsteensvej 4

|

4000 Roskilde

+4563636363

|

www.jndata.dk<https://www.jndata.dk>

|

***@***.***<https://www.linkedin.com/company/jn-data/>

***@***.***

PR #6130, opened 2024-08-22
Added Status Field to VRF model

jnmms:
# Closes # #6120

# What's Changed
Added a status field to VRFs
- Changed vrf models, and set content_types for useful statuses when migrating - default status set to Active for all VRFs
- Made sure all views, edits and filters work

# Screenshots
I do not have screenshots of this, but the layour of it is exatly the same as under VLAN

# TODO
None

- [ ] Unit, Integration Tests


glennmatthews (commented):
Thanks for the PR, looks like a solid start!

I'm still not entirely convinced from a data modeling perspective that a status attribute makes sense at the network VRF level (what does it mean to say "VRF Red has status DOWN across the entire network/namespace"?) and feel like it'd make more sense at the `VRFDeviceAssignment` model instead ("VRF Red is admin DOWN on site1-router1"). That definitely gets trickier to implement from a UI/UX standpoint though as we'd probably need to add a formset or similar in order to define individual VRFDeviceAssignment records and their statuses when editing a Device (or a VRF?) - something similar to how contact/team associations are handled in the current feature set, for example.

jnmms:
I do not disagree that it could maybe be a good idea to have status on the VRFDeviceAssignment, however
generally, we would use this exact status on VRF to easily indicate if the VRF is in use, in planned state or 
decommissioned. That would give an overview and filters could help finding out if a VRF can be (re)used or if 
it's already active.

More or less, the same argumentation could be given for VLANs, where we have a status on the vlan.
 
You could argue, that if all relations have been removed from a VRF, then it's available, anyway it would be nice 
to reflect this to “status”. This would also enable us to do more generic code, if all ipam objects has this attribute.  
 
However, from a provisioning perspective it would be nice to prepare things and when everything is ready, then activate it.
Like “planned”, “reserved”, “active” etc.

In the ideal world you would also run your enterprise using “Intend based configuration” and all running config would 
originate from this. A large enterprise as we have (4000+ devices) you can’t avoid having some legacy. It will probably 
take years to mature everything. That means that we have a lot of devices, where configs are maintained in the devices 
and we just do ordinary backups. In other words the devices themselves are the SoT. Everything across customers and 
across devices/switchdomains.

HanlinMiao:
We will move forward with this rest of the way and make it ready to go.
We are gonna make the status field to be nullable and blankable instead of having a default value of active.
We will consider the `VRFDeviceAssignment` `status` as future work.

nautobot/ipam/choices.py
```diff
@@ -102,6 +102,23 @@ class IPAddressTypeChoices(ChoiceSet):
     )
 
 
+#
+# VRFs
+#
+
+
+class VRFStatusChoices(ChoiceSet):
```
glennmatthews, line 110: Should this be added to `nautobot.extras.management.STATUS_CHOICESET_MAP` like the other statuses?

Do we need a data migration (similar to `0055_softwareimage_softwareversion_data_migration.py`) to call `populate_status_choices()` so that these statuses get auto-created on upgrade?

nautobot/ipam/models.py
```diff
@@ -114,6 +115,7 @@ class VRF(PrimaryModel):
         verbose_name="Route distinguisher",
         help_text="Unique route distinguisher (as defined in RFC 4364)",
     )
+    status = StatusField(blank=True, null=True)
```
glennmatthews, line 118: I believe this is the first model in core with a nullable Status. Would be good to do some extra testing if so to make sure that this doesn't have unexpected behavior in the UI forms or in the REST API.

Let's also add some `status` values to both `test_views.py` and `test_api.py` to test it as an optional field.

nautobot/ipam/tables.py
```diff
@@ -240,7 +241,7 @@ class Meta(BaseTable.Meta):
             "tags",
         )
         # default_columns = ("pk", "name", "rd", "namespace", "tenant", "description")
-        default_columns = ("pk", "name", "namespace", "tenant", "description")
+        default_columns = ("pk", "status", "name", "namespace", "tenant", "description")
```
glennmatthews, line 244: IMO `status` should come after `name` not before it.

glennmatthews (approved):

PR #6103, opened 2024-08-13
Added JobQueue and JobQueueAssignment models and API.

HanlinMiao:
# Closes #6087 
# What's Changed

Added JobQueue and JobQueueAssignment models and API.

# Screenshots

N/A
# TODO

- [x] Explanation of Change(s)
- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [x] Model and migrations
- [x] Serializers, APIViewSets, and FilterSets
- [x] JobQueue Factory
- [x] Unit, Integration Tests

Issue #5917, opened 2024-07-10
Add grapqhl support for custom field definitions


joewesch:
### As ...

Nelly - Network Engineer

### I want ...

To be able to query custom fields via GraphQL

### So that ...

I can use the GraphQL interface to get a list of all current custom fields as well as the custom field a custom field choice relates to.

### I know this is done when...

- I can query custom fields:
```graphql
query {
  custom_fields {
    key
  }
}
```

- I can query custom field choices and get the related custom field:
```graphql
query {
  custom_field_choices {
    custom_field {
      key
    }
  }
}
```

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [X] Apps (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

### Database Changes

None

### External Dependencies

None

glennmatthews:
Missing `@extras_features("graphql")` on the `CustomField` model declaration. Also missing on the `Relationship` model declaration, would probably make sense to add there as well.

glennmatthews:
Some prototype code in https://github.com/nautobot/nautobot/tree/u/glennmatthews-5917. Needs testing.

lampwins:
Let's cover all the extras features that make sense to also include support for.

PR #6154, opened 2024-08-27
Added JobQueue UI/Refactor Job Code to use JobQueue models

### LINKED ISSUE #6125: Remove all task_queue/task_queues references and refactor Job/ScheduledJob to use JobQueue model instead
### Proposed Changes
Remove all `task_queue/task_queues` references and refactor `Job/ScheduledJob` related code to use the JobQueue model instead.

HanlinMiao:
# Closes #6125 
# What's Changed

- [x] JobQueueUIViewSet
- [x] JobQueueFilterForm
- [x] JobQueueForm
- [x] JobQueueBulkEditForm
- [x] JobQueueTable
- [x]  JobQueue UI View Unit Tests
- [x] JobForm changes to include a `job_queue` field dropdown with assigned job queues populated
- [x] Refactor job code to be able to use either job queue models and `task_queues` legacy attributes

# Screenshots

Progress so far:
Screenshot 2024-08-27 at 1 46 28 PM
Screenshot 2024-08-27 at 1 47 26 PM
Screenshot 2024-08-27 at 1 46 13 PM
Screenshot 2024-08-27 at 1 46 37 PM
Screenshot 2024-08-27 at 1 46 59 PM

# TODO

- [x] Unit, Integration Tests


nautobot/extras/api/serializers.py
```diff
@@ -560,6 +560,8 @@ def validate(self, data):
 
 
 class JobQueueSerializer(NautobotModelSerializer, TaggedModelSerializerMixin):
+    display_with_worker_count = serializers.CharField(read_only=True)
```
glennmatthews, line 563: Seems like a weird property to render as a string in the REST API - what's the intent here?

nautobot/extras/forms/forms.py
```diff
@@ -967,6 +973,13 @@ def __init__(self, *args, **kwargs):
 
 
 class JobEditForm(NautobotModelForm):
+    job_queues = DynamicModelMultipleChoiceField(
+        label="Job Queues",
+        queryset=JobQueue.objects.all(),
+        required=False,
+        widget=APISelectMultiple(api_url="/api/extras/job-queues/"),
```
glennmatthews, line 980: Doesn't/shouldn't the widget work automatically for a `DynamicModelMultipleChoiceField`? Why do we need to explicitly specify it here?

nautobot/extras/jobs.py
```diff
@@ -483,6 +484,15 @@ def as_form(cls, data=None, files=None, initial=None, approval_view=False):
             dryrun_default = cls.dryrun_default
             task_queues = cls.task_queues
 
+        # Initialize job_queue choices
+        form.fields["job_queue"] = DynamicModelChoiceField(
+            display_field="display_with_worker_count",
```
glennmatthews, line 489: Ah, I see.

Given that we expect a relatively small number of distinct queues, it *might* be OK to do a static ModelChoiceField instead of a DynamicModelChoiceField so that we don't need to "pollute" the REST API with this attribute?

Alternately, maybe the `display_with_worker_count` should just become the baseline `display` for the JobQueue model?

nautobot/extras/jobs.py
```diff
@@ -483,6 +484,15 @@ def as_form(cls, data=None, files=None, initial=None, approval_view=False):
...
+            queryset=JobQueue.objects.all(),
```
glennmatthews, line 490: Shouldn't this be populated from / restricted to `job_model.job_queues` instead of all possible queues?

nautobot/extras/templates/extras/job_detail.html
```diff
@@ -260,6 +260,16 @@
                     {% endif %}
                 </td>
             </tr>
+            <tr>
+                <td>Job Queues</td>
+                <td>
+                    <ul>
+                        {% for queue in object.job_queues.all %}
+                            <li><a href="{% url 'extras:jobqueue' pk=queue.pk %}">{{queue}}</a></li>
```
glennmatthews, line 268: Can we use `{{ queue|hyperlinked_object }}` here?

nautobot/extras/utils.py
```diff
@@ -391,6 +391,19 @@ def get_worker_count(request=None, queue=None):
     return celery_queues.get(queue, 0)
 
 
+def get_job_queue_worker_count(request=None, job_queue=None):
+    """
+    Return a count of the active Celery workers in a specified queue. Defaults to the `CELERY_TASK_DEFAULT_QUEUE` setting.
+    Same as get_worker_count() method above, but job_queue is an actual JobQueue model instance.
+    """
+    celery_queues = get_celery_queues()
```
glennmatthews, line 399: This is mildly expensive isn't it? Do we need some caching here so that we're not repeating this lookup N times if we have N defined queues?

nautobot/extras/forms/forms.py
HanlinMiao: Removed

nautobot/extras/jobs.py
HanlinMiao: The `query_params` below actually filters the result to be job queues that are assigned to the job.

nautobot/extras/api/serializers.py
HanlinMiao: Will just replace it with `display`

nautobot/extras/utils.py
HanlinMiao: Understood, can you point me to a place where it is appropriate to clear the cache?

glennmatthews: I think in this case just setting it with a short timeout (say 5 seconds) is probably adequate.

examples/example_app/example_app/jobs.py
```diff
@@ -114,11 +113,6 @@ class ExampleLoggingJob(Job):
     class Meta:
         name = "Example logging job."
         description = "I log stuff to demonstrate how UI logging works."
-        task_queues = [
-            settings.CELERY_TASK_DEFAULT_QUEUE,
-            "priority",
-            "bulk",
-        ]
```
glennmatthews, line 121: Why removed? Feels like it should stay here to verify backwards-compatibility.

nautobot/extras/filters/__init__.py
```diff
@@ -920,6 +920,10 @@ class JobQueueFilterSet(NautobotFilterSet, TenancyModelFilterSetMixin):
         },
     )
     queue_type = django_filters.MultipleChoiceFilter(choices=JobQueueTypeChoices, null_value=None)
+    jobs = NaturalKeyOrPKMultipleChoiceFilter(
+        queryset=Job.objects.all(),
+        label="Job (name or ID)",
+    )
```
glennmatthews, line 926: Possibly also a `has_jobs` RelatedMembershipBooleanFilter?

nautobot/extras/forms/forms.py
```diff
@@ -989,8 +995,7 @@ class Meta:
             "time_limit",
             "has_sensitive_variables_override",
             "has_sensitive_variables",
-            "task_queues_override",
-            "task_queues",
+            "job_queues",
```
glennmatthews, line 998: Have we lost the ability for a Job author to update its defined `task_queues` and have that automatically reflected in Nautobot (if Nautobot didn't override it with a `task_queues_override` value)? Not critical functionality but we should document it if so.

nautobot/extras/forms/forms.py
```diff
@@ -1081,10 +1087,6 @@ class JobBulkEditForm(NautobotBulkEditForm):
         required=False,
         help_text="If checked, time limits will be reverted to the default values defined in each Job's source code",
     )
-    clear_task_queues_override = forms.BooleanField(
-        required=False,
-        help_text="If checked, task queue overrides will be reverted to the default values defined in each Job's source code",
-    )
```
glennmatthews, line 1087: Definitely feels like a loss of functionality.

nautobot/extras/forms/forms.py
```diff
@@ -1201,6 +1203,60 @@ class JobHookFilterForm(BootstrapMixin, forms.Form):
...
+    model = JobQueue
+    q = forms.CharField(required=False, label="Search")
+    name = forms.CharField(required=False)
+    queue_type = forms.MultipleChoiceField(
+        choices=JobQueueTypeChoices,
+        required=False,
+        widget=StaticSelect2Multiple(),
+    )
+    tenant = DynamicModelMultipleChoiceField(queryset=Tenant.objects.all(), to_field_name="name", required=False)
+    tags = TagFilterField(model)
```
glennmatthews, line 1239: Jobs filter field?

nautobot/extras/jobs.py
```diff
@@ -389,11 +389,6 @@ def has_sensitive_variables(cls) -> bool:  # pylint: disable=no-self-argument
     def supports_dryrun(cls) -> bool:  # pylint: disable=no-self-argument
         return isinstance(getattr(cls, "dryrun", None), DryRunVar)
 
-    @final
-    @classproperty
-    def task_queues(cls) -> list:  # pylint: disable=no-self-argument
-        return cls._get_meta_attr_and_assert_type("task_queues", [], expected_type=(list, tuple))
-
```
glennmatthews, line 396: Removed functionality?

nautobot/extras/models/jobs.py
```diff
@@ -301,6 +291,20 @@ def job_task(self):
         except TypeError as err:  # keep 2.0-2.2.2 exception behavior
             raise NotRegistered from err
 
+    @property
+    def task_queues(self):
```
glennmatthews, line 295: Probably need to explicitly add this to the REST serializer (and perhaps GraphQL) for backwards API compatibility.

nautobot/extras/models/jobs.py
```diff
@@ -301,6 +291,20 @@ def job_task(self):
...
+        return self.job_queues.all().values_list("name", flat=True)
```
glennmatthews, line 296: Have we lost the ability to have a specified order to these queues for a given job? What's the default ordering, just by `name`?

nautobot/extras/models/jobs.py
```diff
@@ -301,6 +291,20 @@ def job_task(self):
...
+
+    @task_queues.setter
+    def task_queues(self, value):
+        # value is going to be a comma separated list of queue names
+        # assume all queue types are celery for now
```
glennmatthews, line 301: Feels like something we'll need to rip out soon enough.

nautobot/extras/models/jobs.py
```diff
@@ -301,6 +291,20 @@ def job_task(self):
...
+        if isinstance(value, str):
+            value = value.split(",")
+        for queue in value:
+            job_queue, _ = JobQueue.objects.get_or_create(name=queue, queue_type=JobQueueTypeChoices.TYPE_CELERY)
```
glennmatthews, line 305:
```suggestion
            job_queue, _ = JobQueue.objects.get_or_create(name=queue, defaults={"queue_type": JobQueueTypeChoices.TYPE_CELERY})
```

nautobot/extras/models/jobs.py
```diff
@@ -1127,6 +1129,16 @@ def schedule(self):
 
         return self.to_cron()
 
+    @property
+    def queue(self):
```
glennmatthews, line 1133: Similarly probably need to add this field explicitly to the API serializer and GraphQL.

nautobot/extras/tables.py
```diff
@@ -822,6 +823,31 @@ class Meta(BaseTable.Meta):
         }
 
 
+class JobQueueTable(BaseTable):
+    pk = ToggleColumn()
+    name = tables.Column(linkify=True)
+    queue_type = tables.Column()
+    tenant = tables.Column(linkify=True)
+    description = tables.Column()
```
glennmatthews, line 831:
```suggestion
    name = tables.Column(linkify=True)
    tenant =TenantColumn()
```

nautobot/extras/templates/extras/jobqueue_retrieve.html
```diff
@@ -0,0 +1,47 @@
...
+                <tr>
+                    <td>Description</td>
+                    <td>
+                        <span>{{ object.description | placeholder }}</span>
+                    </td>
+                </tr>
+                <tr>
+                    <td>Tenant</td>
+                    <td>
+                        <span>{{ object.tenant | hyperlinked_object }}</span>
```
glennmatthews, line 32: Common pattern from other models:

```suggestion
                        {% if object.tenant and object.tenant.tenant_group %}
                            {{ object.tenant.tenant_group|hyperlinked_object }} /
                        {% endif %}
                        {{ object.tenant|hyperlinked_object }}
```

nautobot/extras/templates/extras/jobqueue_retrieve.html
```diff
@@ -0,0 +1,47 @@
...
+        <div class="panel panel-default">
+            <div class="panel-heading">
+                <strong>Job Queue</strong>
+            </div>
+
+            <table class="table table-hover panel-body attr-table">
+                <tr>
+                    <td>Name</td>
+                    <td>
+                        <span>{{ object.name }}</span>
```
glennmatthews, line 14: Don't need the `<span>...</span>` wrappings here and below unless I'm missing something?

nautobot/extras/templates/extras/jobqueue_retrieve.html
```diff
@@ -0,0 +1,47 @@
...
+                    <td>
+                        <span>{{ object.tenant | hyperlinked_object }}</span>
+                    </td>
+                </tr>
+                <tr>
+                    <td>Assigned Jobs</td>
+                    <td>
+                        <ul>
+                            {% for job in object.jobs.all %}
+                                <li><a href="{% url 'extras:job' pk=job.pk %}">{{ job }}</a></li>
```
glennmatthews, line 40:
```suggestion
                                <li>{{ job|hyperlinked_object }}</a></li>
```

nautobot/extras/templates/extras/jobqueue_retrieve.html
```diff
@@ -0,0 +1,47 @@
...
+                <tr>
+                    <td>Name</td>
+                    <td>
+                        <span>{{ object.name }}</span>
+                    </td>
+                </tr>
+                <tr>
+                    <td>Queue Type</td>
+                    <td>
+                        <span>{{ object.queue_type }}</span>
```
glennmatthews, line 20:
```suggestion
                        <span>{{ object.get_queue_type_display }}</span>
```

nautobot/extras/templatetags/job_buttons.py
```diff
@@ -108,16 +108,17 @@ def _render_job_button_for_obj(job_button, obj, context, content_type):
     # Disable buttons if the user doesn't have permission to run the underlying Job.
     has_run_perm = Job.objects.check_perms(context["user"], instance=job_button.job, action="run")
     try:
-        _task_queue = job_button.job.task_queues[0]
+        job_queues = list(job_button.job.job_queues.all().values_list("name", flat=True))
+        _job_queue = job_queues[0]
```
glennmatthews, line 112: Behavior change here if we no longer have a user-configurable ordering of `job.job_queues`. Document please!

nautobot/extras/test_jobs/api_test_job.py
```diff
@@ -11,7 +9,6 @@ class APITestJob(Job):
     class Meta:
         name = "Job for API Tests"
         has_sensitive_variables = False
-        task_queues = [settings.CELERY_TASK_DEFAULT_QUEUE, "nonexistent"]
```
glennmatthews, line 14: Why removed?

nautobot/extras/tests/test_api.py
```diff
@@ -1360,8 +1360,7 @@ class JobTest(
         "time_limit": 650,
         "has_sensitive_variables": False,
         "has_sensitive_variables_override": True,
-        "task_queues": ["default", "priority"],
-        "task_queues_override": True,
+        "job_queues": [],
```
glennmatthews, line 1363: Better test would be to have specific queue PKs defined in `setupTestData`.

nautobot/extras/tests/test_jobs.py
```diff
@@ -115,49 +115,6 @@ def test_dryrun_default(self):
         form = job_class().as_form()
         self.assertEqual(form.fields["dryrun"].initial, job_model.dryrun_default)
 
-    @mock.patch("nautobot.extras.utils.get_celery_queues")
-    def test_job_class_task_queues(self, mock_get_celery_queues):
```
glennmatthews, line 119: Should replace these with new tests, not just remove them.

nautobot/extras/views.py
```diff
@@ -1311,9 +1312,9 @@ def get(self, request, class_path=None, pk=None):
                     # for example "?kwargs_from_job_result=<UUID>&integervar=22"
                     explicit_initial = initial
                     initial = job_result.task_kwargs.copy()
-                    task_queue = job_result.celery_kwargs.get("queue", None)
-                    if task_queue is not None:
-                        initial["_task_queue"] = task_queue
+                    job_queue = job_result.celery_kwargs.get("queue", None)
+                    if job_queue is not None:
+                        initial["_job_queue"] = job_queue
```
glennmatthews, line 1317: do we need to map the queue name to a JobQueue record for the form?

nautobot/extras/jobs.py
HanlinMiao: Restored

examples/example_app/example_app/jobs.py
HanlinMiao: Restored

nautobot/extras/filters/__init__.py
HanlinMiao: Added

nautobot/extras/forms/forms.py
HanlinMiao: Added

nautobot/extras/models/jobs.py
HanlinMiao: Added on the API side but it looks like ScheduledJob does not support GraphQL to begin with

glennmatthews: Hm okay. Probably would be desirable to add that but I'm fine with considering that out-of-scope for now. (See also #5917)

nautobot/extras/models/jobs.py
HanlinMiao: Added

nautobot/extras/tests/test_api.py
HanlinMiao: It seems like we can't write directly to `ManyRelatedField` because `read_only` is set to `True`. I looked through the codebase and found no example of updating a many-to-many field from the Rest API.

nautobot/extras/jobs.py
HanlinMiao: Changed it to `queryset=JobQueue.objects.filter(jobs=job_model),`

nautobot/extras/models/jobs.py
HanlinMiao: Just by `name` yeah, will document this if this is an acceptable loss of functionality

nautobot/extras/forms/forms.py
HanlinMiao: Yeah, this one is tricky. I was thinking maybe restore this override field and when it is checked, reset the `job_queues` field with `Job Queues` with names that are defined by the job `Class meta: task_queues`?

nautobot/extras/tests/test_api.py
glennmatthews: Ah good point. Are we going to need a separate model and API endpoint for `job-to-job-queues` M2M in that case? :-(

HanlinMiao: Sorry if I am misunderstanding, but should JobQueueAssignment do the job in this case?

glennmatthews: Yeah, I'd forgotten/not seen that you have such a model already. It may just need a REST viewset/url added in that case.

HanlinMiao: Ahhh, yeah those are already added from https://github.com/nautobot/nautobot/pull/6103. One of the cons of breaking down bigger stories.

mkdocs.yml
```diff
@@ -253,6 +253,8 @@ plugins:
         "models/extras/gitrepository.md": "user-guide/platform-functionality/gitrepository.md"
         "models/extras/jobhook.md": "user-guide/platform-functionality/jobs/jobhook.md"
         "models/extras/joblogentry.md": "user-guide/platform-functionality/jobs/models.md"
+        "models/extras/jobqueue.md": "user-guide/platform-functionality/jobs/jobqueue.md"
+        "models/extras/jobqueueassignment.md": "user-guide/platform-functionality/jobs/jobqueueassignment.md"
```
glennmatthews, line 257: Technically these should be located later on, around line 311-313, since they're new in post-2.x and for whatever reason we decided to divide up the redirects in this file accordingly.

nautobot/docs/user-guide/platform-functionality/jobs/index.md
```diff
@@ -65,6 +65,10 @@ An administrator or user with `extras.change_job` permission can also edit a Job
 
 This is done by setting the corresponding "override" flag (`grouping_override`, `name_override`, etc.) to `True` then providing a new value for the attribute in question. An overridden attribute will remain set to its overridden value even if the underlying Job class definition changes and `nautobot-server <migrate|post_upgrade>` gets run again. Conversely, clearing the "override" flag for an attribute and saving the database record will revert the attribute to the underlying value defined within the Job class source code.
 
++/- 2.4.0
+
+In Nautobot v2.4, users can no longer override the `task_queues` attribute. Instead, they must assign [Job Queues](./jobqueue.md) to Jobs to specify the eligible queues for running those Jobs. Previously, users could define the order of queues by listing queue names in a comma-separated string, but this is no longer possible in v2.4. Now, Job Queues are sorted alphabetically by name and cannot be reordered.
+
```
glennmatthews, line 71: Probably remove `task_queues` from the above bullet list as well?

```suggestion
+/- 2.4.0 "Job queues"
    In Nautobot v2.4 and later, users can no longer override the `task_queues` attribute. Instead, they must assign [Job Queues](./jobqueue.md) to Jobs to specify the eligible queues for running those Jobs. Previously, users could define the order of queues by listing queue names in a comma-separated string, but this is no longer possible in v2.4. Now, Job Queues are sorted alphabetically by name and cannot be reordered.

```

nautobot/docs/user-guide/platform-functionality/jobs/jobqueueassignment.md
```diff
@@ -0,0 +1,5 @@
+# Job Queue Assignments
+
++++ 2.4.0
+
+Job Queue Assignments represent the many-to-many relationships between [Jobs](./index.md) and [Job Queues](./jobqueue.md). It has `job` and `job_queue` as required attributes and each Job Queue Assignment must have a unique combination of `job` and `job_queue`.
```
glennmatthews, line 5: Consider rolling this into `jobqueue.md` instead of as a separate file?

nautobot/docs/user-guide/platform-functionality/jobs/index.md
glennmatthews: Are there other updates needed to this file to reflect the new functionality of Job Queues or was that done in a previous PR?

nautobot/extras/factory.py
```diff
@@ -158,6 +159,22 @@ class Params:
...
+        model = JobQueueAssignment
+
+    job = random_instance(Job, allow_null=True)
+
+    @factory.lazy_attribute
+    def job_queue(self):
+        job_queue = factory.random.randgen.choice(JobQueue.objects.all())
+        while JobQueueAssignment.objects.filter(job=self.job, job_queue=job_queue).exists():
+            job_queue = factory.random.randgen.choice(JobQueue.objects.all())
+        return job_queue
```
glennmatthews, line 175: Possibility of an infinite loop here if a job is already assigned to all queues.

As an alternative, did you consider having `JobQueueFactory` automatically assign jobs to the queues rather than using a separate factory?

nautobot/extras/forms/forms.py
```diff
@@ -951,22 +956,23 @@ class JobForm(BootstrapMixin, forms.Form):
...
-    )
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
         # Move special fields to the end of the form
-        for field in ["_task_queue", "_profile"]:
+        for field in ["_profile"]:
             value = self.fields.pop(field)
             self.fields[field] = value
```
glennmatthews, line 966: Maybe move the `_profile` field over to the `Job.as_form` method as well just to reduce the split-brain logic?

nautobot/extras/forms/forms.py
glennmatthews: I think for parity with the old behavior we'd also need to make sure that `refresh_job_model_from_job_class()` would reset the `job_queues` if the override is specified? That was probably lost by removing `task_queues` from `JOB_OVERRIDABLE_FIELDS`.

nautobot/extras/forms/forms.py
```diff
@@ -1201,6 +1203,61 @@ class JobHookFilterForm(BootstrapMixin, forms.Form):
...
+    tenant = DynamicModelChoiceField(
+        queryset=Tenant.objects.all(),
+        required=False,
+    )
+    description = forms.CharField(required=False, max_length=CHARFIELD_MAX_LENGTH)
+
+    class Meta:
+        model = JobQueue
+        nullable_fields = [
+            "description",
```
glennmatthews, line 1225: Should `tenant` be included here?

nautobot/extras/api/serializers.py
```diff
@@ -535,6 +535,8 @@ def get_parent(self, obj):
 
 
 class JobSerializer(NautobotModelSerializer, TaggedModelSerializerMixin):
+    task_queues = serializers.JSONField(read_only=True, required=False)
```
glennmatthews, line 538: Maybe add some comments to the serializers and graphql types documenting that these fields are for backward compatibility with 2.3 and earlier, just in case we wonder later.

nautobot/extras/jobs.py
```diff
@@ -477,14 +478,17 @@ def as_form(cls, data=None, files=None, initial=None, approval_view=False):
...
+                queryset=JobQueue.objects.filter(jobs=job_model),
+                query_params={"jobs": [job_model.pk]},
+                required=False,
+                help_text="The job queue to route this job to",
+                label="Job queue",
+            )
         except JobModel.DoesNotExist:
             logger.error("No Job instance found in the database corresponding to %s", cls.class_path)
             dryrun_default = cls.dryrun_default
-            task_queues = cls.task_queues
```
glennmatthews, line 484: Do we need to create a stub `_job_queue` form field in this exception case?

nautobot/extras/models/jobs.py
```diff
@@ -301,6 +291,22 @@ def job_task(self):
...
+    @task_queues.setter
+    def task_queues(self, value):
+        # value is going to be a comma separated list of queue names
+        # assume all queue types are celery for now
+        if isinstance(value, str):
+            value = value.split(",")
+        for queue in value:
+            job_queue, _ = JobQueue.objects.get_or_create(
+                name=queue, defaults={"queue_type": JobQueueTypeChoices.TYPE_CELERY}
+            )
```
glennmatthews, line 307: Do we need for backwards compatibility reasons to be able to support *creation* of JobQueue records through assignment to `task_queues = ...`? It feels a bit gross to me here, any chance we can make this require that any specified queues are already defined without it being a breaking change?

nautobot/extras/models/jobs.py
```diff
@@ -1127,6 +1131,17 @@ def schedule(self):
...
         return self.to_cron()
 
+    @property
+    def queue(self):
+        return self.job_queue.name if self.job_queue else ""
+
+    @queue.setter
+    def queue(self, value):
+        if value:
+            self.job_queue, _ = JobQueue.objects.get_or_create(
```
glennmatthews, line 1141: Same question here about create-by-assignment, can/should we change this to just `JobQueue.objects.get(...)`?

nautobot/extras/templates/extras/jobqueue_retrieve.html
```diff
@@ -0,0 +1,50 @@
...
+                    </td>
+                </tr>
+                <tr>
+                    <td>Tenant</td>
+                    <td>
+                        {% if object.tenant and object.tenant.tenant_group %}
+	                        {{ object.tenant.tenant_group|hyperlinked_object }} /
+	                    {% endif %}
+	                    {{ object.tenant|hyperlinked_object }}
+                    </td>
```
glennmatthews, line 36:
```suggestion
                    <td>
                        {% if object.tenant and object.tenant.tenant_group %}
	                    {{ object.tenant.tenant_group|hyperlinked_object }} /
	                {% endif %}
	                {{ object.tenant|hyperlinked_object }}
                    </td>
```

nautobot/extras/tests/test_api.py
```diff
@@ -1564,7 +1570,7 @@ def test_run_job_no_worker(self, mock_get_worker_count):
         response = self.client.post(url, data, format="json", **self.header)
         self.assertHttpStatus(response, status.HTTP_503_SERVICE_UNAVAILABLE)
         self.assertEqual(
-            response.data["detail"], "Unable to process request: No celery workers running on queue default."
+            response.data["detail"].title(), "Unable To Process Request: No Celery Workers Running On Queue Default."
```
glennmatthews, line 1573: Why this change?

nautobot/extras/tests/test_api.py
```diff
@@ -1548,6 +1549,11 @@ def test_run_job_no_worker(self, mock_get_worker_count):
         """Job run cannot be requested if Celery is not running."""
         mock_get_worker_count.return_value = 0
         self.add_permissions("extras.run_job")
+        class_path = "api_test_job.APITestJob"
+        job_model = Job.objects.get_for_class_path(class_path)
+        # Make sure no queues are associated with it so it is using the celery default queue.
+        # And the error message is deterministic on line 1573
```
glennmatthews, line 1555: Since we're alphabetical now, maybe we should define JobQueueFactory such that it only creates queue names that are alphabetically after "default"? Or can/should we define JobQueue.Meta.ordering somehow (how?) so that 'default' comes first regardless of all other names?

nautobot/extras/tests/test_api.py
```diff
@@ -2047,13 +2053,14 @@ def test_run_job_with_valid_task_queue(self, _):
...
     @mock.patch("nautobot.extras.api.views.get_worker_count", return_value=1)
-    def test_run_job_with_default_queue_with_empty_job_model_task_queues(self, _):
+    def test_run_job_with_default_queue_with_empty_job_model_job_queues(self, _):
         self.add_permissions("extras.run_job")
         data = {
             "task_queue": settings.CELERY_TASK_DEFAULT_QUEUE,
         }
 
         job_model = Job.objects.get_for_class_path("pass.TestPass")
+        job_model.job_queues.set([])
```
glennmatthews, line 2063: This looks like the wrong fix to me - shouldn't we instead correct the `data` above to specify a `job_queue`?

nautobot/extras/views.py
```diff
@@ -1311,9 +1319,12 @@ def get(self, request, class_path=None, pk=None):
...
                     initial = job_result.task_kwargs.copy()
-                    task_queue = job_result.celery_kwargs.get("queue", None)
-                    if task_queue is not None:
-                        initial["_task_queue"] = task_queue
+                    job_queue = job_result.celery_kwargs.get("queue", None)
+                    if job_queue is not None:
+                        jq, _ = JobQueue.objects.get_or_create(
+                            name=job_queue, defaults={"queue_type": JobQueueTypeChoices.TYPE_CELERY}
+                        )
+                        initial["_job_queue"] = jq
```
glennmatthews, line 1327: Here too it feels wrong for a `get` to have a side effect of creating a database record.

nautobot/extras/views.py
```diff
@@ -1380,7 +1391,13 @@ def post(self, request, class_path=None, pk=None):
                 "One of these two flags must be removed before this job can be scheduled or run.",
             )
         elif job_form is not None and job_form.is_valid() and schedule_form.is_valid():
-            task_queue = job_form.cleaned_data.pop("_task_queue", None)
+            job_queue = job_form.cleaned_data.pop("_job_queue", None)
+            if job_queue is None:
+                job_queue, _ = JobQueue.objects.get_or_create(
+                    name=settings.CELERY_TASK_DEFAULT_QUEUE, defaults={"queue_type": JobQueueTypeChoices.TYPE_CELERY}
+                )
```
glennmatthews, line 1398: We shouldn't auto-recreate the default queue here either IMO.

nautobot/extras/factory.py
HanlinMiao: Refactored it and remove JobQueueAssignmentFactory

nautobot/extras/jobs.py
HanlinMiao: What would be the choices in the form field in that case? Just default or None

glennmatthews: I don't remember what the code does here in this case - do we still present a form to the user even though it can't be run?

nautobot/extras/tests/test_api.py
HanlinMiao: It was failing in CI but not locally. Reverted it to see what happens

nautobot/extras/jobs.py
HanlinMiao: Hmmm we do still present the same form but we are using the class attribute `cls.task_queues`.
But without the job_model, the queryset for `_job_queue` form field becomes unfilterable. What I am doing rn is something similar to that.
```
         try:
            job_model = JobModel.objects.get_for_class_path(cls.class_path)
            dryrun_default = job_model.dryrun_default if job_model.dryrun_default_override else cls.dryrun_default
            job_queue_queryset = JobQueue.objects.filter(jobs=job_model)
            job_queue_params = {"jobs": [job_model.pk]}
        except JobModel.DoesNotExist:
            logger.error("No Job instance found in the database corresponding to %s", cls.class_path)
            dryrun_default = cls.dryrun_default
            pk_list = []
            for queue in cls.task_queues:
                job_queue, _ = JobQueue.objects.get_or_create(
                            name=queue, defaults={"queue_type": JobQueueTypeChoices.TYPE_CELERY}
                        )
                pk_list.append(job_queue.pk)
            job_queue_queryset = JobQueue.objects.filter(pk__in=pk_list)
            job_queue_params = {"id": pk_list}


        # Initialize job_queue choices
        form.fields["_job_queue"] = DynamicModelChoiceField(
            queryset=job_queue_queryset,
            query_params=job_queue_params,
            required=False,
            help_text="The job queue to route this job to",
            label="Job queue",
        )
```

nautobot/extras/models/jobs.py
HanlinMiao: It is for backwards-compatibility reasons. Require specified queues to be defined before setting `task_queues` as in raise an error if a specified queue does not exist?

nautobot/extras/tests/test_api.py
HanlinMiao: Reverted and it is passing CI

nautobot/extras/tests/test_api.py
HanlinMiao: I think the original test is testing the job without task_queues specified in the class meta and no overrides as well. More specifically, I think it is testing that `valid_queues` will always include the [default](https://github.com/nautobot/nautobot/pull/6154/files/8cd0cdb7575c5bb9bcc53a871b6e00e6210f859d#diff-9a2bdd132c5e3a0e6947134b0ae99c641837ad953eb9f5585633bf33e4a60dc9R602-R603) queue if no queues are specified. And on the `JobInputSerializer`, the field for specifying a queue is still named `task_queue` for backward compatibility reasons.

nautobot/extras/views.py
HanlinMiao: Should we raise an error here JobQueue.DoesNotExist?

nautobot/docs/user-guide/platform-functionality/jobs/index.md
HanlinMiao: I do not think there are other updates needed to this file.

nautobot/extras/tests/test_api.py
HanlinMiao: Should we define a `models.QuerySet` subclass and define a custom ordering method there?
```
class JobQueueQuerySet(models.QuerySet):
       def order_by_queue_names(self):
             # pop "default" and insert it to the front somehow?

class JobQueue:
       ....
       objects = JobQueueQuerySet.as_manager()
```

HanlinMiao: I actually dont know if the above method is quite possible after some experimenting

glennmatthews: Yeah I don't think we want to go down that road, but thanks for investigating.

nautobot/extras/views.py
HanlinMiao: I would also push back a little on that it emulates the previous behavior to the best of our abilities. Because task_queue is only a simple string and not a database record. Requiring the user to manually create a job queue before passing it to celery_kwargs might break some things.

nautobot/extras/jobs.py
glennmatthews: I guess I'm trying to understand why we're presenting a form in this case instead of giving an error message and sending the user back to the jobs list. In this case since we can't find the `job_model`, it's not something we can actually run, right?

nautobot/extras/factory.py
```diff
@@ -158,21 +157,11 @@ class Params:
...
-    def job_queue(self):
-        job_queue = factory.random.randgen.choice(JobQueue.objects.all())
-        while JobQueueAssignment.objects.filter(job=self.job, job_queue=job_queue).exists():
-            job_queue = factory.random.randgen.choice(JobQueue.objects.all())
-        return job_queue
+    @factory.post_generation
+    def jobs(self, create, extracted, **kwargs):
+        jobs = Job.objects.all()[:3]
+        for job in jobs:
+            job.job_queues.add(self)
```
glennmatthews, line 164: Can we instead use `get_random_instances()` here? Something like `self.jobs.set(get_random_instances(Job))`?

nautobot/extras/jobs.py
HanlinMiao: Yeah so the original code is:
```
       except JobModel.DoesNotExist:
            logger.error("No Job instance found in the database corresponding to %s", cls.class_path)
            dryrun_default = cls.dryrun_default
            task_queues = cls.task_queues

        # Update task queue choices
        form.fields["_task_queue"].choices = task_queues_as_choices(task_queues)

        if cls.supports_dryrun and (not initial or "dryrun" not in initial):
            # Set initial "dryrun" checkbox state based on the Meta parameter
            form.fields["dryrun"].initial = dryrun_default
        if not settings.DEBUG:
            form.fields["_profile"].widget = forms.HiddenInput()

        # https://github.com/PyCQA/pylint/issues/3484
        if cls.field_order:  # pylint: disable=using-constant-test
            form.order_fields(cls.field_order)

        if approval_view:
            # Set `disabled=True` on all fields
            for _, field in form.fields.items():
                field.disabled = True

        return form
```
When we do not find the Job in the database, we get those attributes (`dryrun_default`, `task_queues`) using the Job class meta attributes and return the form regardless.

HanlinMiao: Take out the try...except..block and just grab the job_model straight up

nautobot/extras/models/jobs.py
HanlinMiao: Raise a validation error if the task_queue specified does not exist.

nautobot/extras/models/jobs.py
HanlinMiao: If queue does not exist, raise a validation error same as above.

nautobot/extras/jobs.py
glennmatthews: Looks like this broke `test_job_variables` because the jobs that it defines and uses are local to the test and are not registered in the database. I was afraid that might happen. Sorry! Maybe as a fix could explicitly register those jobs as part of the test?

nautobot/extras/migrations/0117_task_queue_to_job_queue_migration.py
```diff
@@ -11,10 +12,17 @@ def migrate_task_queues_to_job_queues(apps, schema):
     ScheduledJob = apps.get_model("extras", "ScheduledJob")
     JobQueue = apps.get_model("extras", "JobQueue")
 
+    default_job_queue, _ = JobQueue.object.get_or_create(
```
glennmatthews, line 15:
```suggestion
    default_job_queue, _ = JobQueue.objects.get_or_create(
```

nautobot/extras/migrations/0117_task_queue_to_job_queue_migration.py
```diff
@@ -11,10 +12,17 @@ def migrate_task_queues_to_job_queues(apps, schema):
...
+        name=settings.CELERY_TASK_DEFAULT_QUEUE, defaults={"queue_type": JobQueueTypeChoices.TYPE_CELERY}
+    )
+
     for job in Job.objects.all():
         task_queues = job.task_queues
         # remove duplicates?
         task_queues = list(set(task_queues))
+        # Add default Job Queue to every job record.
+        job.job_queues.add(default_job_queue)
```
glennmatthews, line 24: I *think* we should only do this if `task_queues` is an empty list? Not 100% sure.

nautobot/extras/migrations/0117_task_queue_to_job_queue_migration.py
HanlinMiao: My vscode syntax highlighting is all messed up for some reason thank you for catching it

nautobot/extras/tests/test_api.py
glennmatthews: We should enhance `JobInputSerializer` to support a `job_queue` argument as an alternative - possibly a versioned API needed.

nautobot/extras/models/jobs.py
```diff
@@ -1134,6 +1139,19 @@ def schedule(self):
...
+        return self.job_queue.name if self.job_queue else ""
+
+    @queue.setter
+    def queue(self, value):
+        if value:
+            try:
+                job_queue = JobQueue.objects.get(name=value)
+                self.job_queue = job_queue
+            except JobQueue.DoesNotExist:
+                raise ValidationError(f"Job Queue {value} does not exist in the database.")
```
glennmatthews, line 1153:
```suggestion
                job_queue = JobQueue.objects.get(name=value, queue_type=JobQueueTypeChoices.TYPE_CELERY)
                self.job_queue = job_queue
            except JobQueue.DoesNotExist:
                raise ValidationError(f"Celery Job Queue {value} does not exist in the database.")
```

nautobot/extras/templates/extras/jobqueue_retrieve.html
```diff
@@ -0,0 +1,50 @@
...
+                </tr>
+                <tr>
+                    <td>Assigned Jobs</td>
+                    <td>
+                        <ul>
+                            {% for job in object.jobs.all %}
+                                <li>{{ job|hyperlinked_object }}</li>
+                            {% endfor %}
+                        </ul>
+                    </td>
```
glennmatthews, line 46: This should probably be a related-object table instead of a list.

nautobot/extras/tests/test_api.py
```diff
@@ -1344,37 +1344,6 @@ class JobTest(
...
-        "hidden_override": True,
-        "hidden": True,
-        "soft_time_limit_override": True,
-        "soft_time_limit": 350.1,
-        "time_limit_override": True,
-        "time_limit": 650,
-        "has_sensitive_variables": False,
-        "has_sensitive_variables_override": True,
-        "task_queues": ["default", "priority"],
-        "task_queues_override": True,
```
glennmatthews, line 1369: Let's see if we can reintroduce this

nautobot/extras/tests/test_job_variables.py
```diff
@@ -30,136 +30,106 @@ class StringVarJob(Job):
             var1 = StringVar(min_length=3, max_length=3, regex=r"[a-z]+")
 
         # Validate min_length enforcement
-        data = {"var1": "xx"}
-        form = StringVarJob().as_form(data)
-        self.assertFalse(form.is_valid())
-        self.assertIn("var1", form.errors)
+        field = StringVarJob().var1.as_field()
```
glennmatthews, line 33: Should be able to slim this test down even a bit further by getting rid of the `Job` subclasses altogether and just instantiating the vars directly.

nautobot/extras/tests/test_jobs.py
```diff
@@ -115,47 +117,35 @@ def test_dryrun_default(self):
...
+    def test_job_task_queues_setter(self):
+        """Test the task_queues property setter on Job."""
+        module = "dry_run"
+        name = "TestDryRun"
+        _, job_model = get_job_class_and_model(module, name)
+        invalid_queue = "Invalid job Queue"
+        with self.assertRaises(ValidationError) as cm:
+            job_model.task_queues = [invalid_queue]
+            job_model.validated_save()
+        self.assertIn(f"Job Queue {invalid_queue} does not exist in the database.", str(cm.exception))
```
glennmatthews, line 129: Let's also add a test that you can't use `task_queues` to set a non-Celery JobQueue (per my above proposed changes).

nautobot/extras/templatetags/job_buttons.py
```diff
@@ -28,7 +28,7 @@
 <input type="hidden" name="object_pk" value="{object_pk}">
 <input type="hidden" name="object_model_name" value="{object_model_name}">
 <input type="hidden" name="_schedule_type" value="immediately">
-<input type="hidden" name="_task_queue" value="{task_queue}">
+<input type="hidden" name="_job_queue" value="{job_queue}">
```
glennmatthews, line 31: Does this work correctly passing a queue name instead of a PK? Because at a glance it looks like the updated view code is expecting a PK.

nautobot/extras/tests/test_jobs.py
```diff
@@ -955,7 +943,7 @@ def test_form_field(self):
         name = "TestJobHookReceiverLog"
         job_class, _job_model = get_job_class_and_model(module, name)
         form = job_class().as_form()
-        self.assertSequenceEqual(list(form.fields.keys()), ["object_change", "_task_queue", "_profile"])
+        self.assertSequenceEqual(list(form.fields.keys()), ["object_change", "_profile", "_job_queue"])
```
glennmatthews, line 946: Can/should we change the form code to restore the previous order of fields?

nautobot/extras/tests/test_models.py
```diff
@@ -1857,6 +1856,14 @@ def setUp(self):
             start_time=datetime(year=2050, month=1, day=22, hour=0, minute=0, tzinfo=ZoneInfo("America/New_York")),
         )
 
+    def test_scheduled_job_queue_setter(self):
+        """Test the queue property setter on ScheduledJob."""
+        invalid_queue = "Invalid job Queue"
+        with self.assertRaises(ValidationError) as cm:
+            self.daily_utc_job.queue = invalid_queue
+            self.daily_utc_job.validated_save()
```
glennmatthews, line 1864: Let's also test with setting a non-Celery queue here.

nautobot/extras/tests/test_views.py
```diff
@@ -2884,19 +2903,22 @@ def test_task_queue_hidden_input_is_present(self):
...
-        self.job.save()
+        job_queues = self.job.job_queues.all().values_list("name", flat=True)
+        self.assertIn(f'<input type="hidden" name="_job_queue" value="{job_queues[0]}">', content, content)
+        self.job.job_queues.set([])
         response = self.client.get(self.location_type.get_absolute_url(), follow=True)
         self.assertEqual(response.status_code, 200)
         content = extract_page_body(response.content.decode(response.charset))
         self.assertIn(
-            f'<input type="hidden" name="_task_queue" value="{settings.CELERY_TASK_DEFAULT_QUEUE}">', content, content
+            f'<input type="hidden" name="_job_queue" value="{settings.CELERY_TASK_DEFAULT_QUEUE}">', content, content
```
glennmatthews, line 2921: Here too, is the use of a queue name correct or should it be a PK?

nautobot/extras/utils.py
```diff
@@ -391,6 +397,19 @@ def get_worker_count(request=None, queue=None):
     return celery_queues.get(queue, 0)
 
 
+def get_job_queue_worker_count(request=None, job_queue=None):
```
glennmatthews, line 400: If this is going to stay celery-specific let's give it a name that indicates such. If it isn't, let's add appropriate TODOs and/or refactor it so that it doesn't call `get_celery_queues` if the given `job_queue` is not a Celery queue.

nautobot/extras/forms/forms.py
glennmatthews: Doing some experimentation around reintroducing `[job|task]_queues_override` flag on the Job model.

glennmatthews: To do this, we may need to go back to your earlier behavior of having the `task_queues` property setter implicitly create JobQueues for any requested queues that don't already exist. Let's discuss tomorrow.

nautobot/extras/templatetags/job_buttons.py
HanlinMiao: Manually tested job buttons and it's working.

nautobot/extras/api/views.py
```diff
@@ -624,7 +628,19 @@ def run(self, request, *args, **kwargs):
...
+                "_task_queue", None
+            ) or input_serializer.validated_data.get("_job_queue", None)
+            if not task_queue:
+                task_queue = default_valid_queue
+
+            # Log a warning if _task_queue and _job_queue fields are both specified out
+            if input_serializer.validated_data.get("_task_queue", None) and input_serializer.validated_data.get(
+                "_job_queue", None
+            ):
+                logger.warning("_task_queue and _job_queue are both specified. Please specifiy only one or another.")
```
glennmatthews, line 643: Should this be a ValidationError?

glennmatthews (approved):

nautobot/extras/models/jobs.py
```diff
@@ -294,20 +298,29 @@ def job_task(self):
...
-            JobQueueAssignment.objects.get_or_create(job_queue=job_queue, job=self)
+                raise ValidationError(f"Job Queue {queue} does not exist in the database.")
+        self.job_queues.set(job_queues)
+
+    @property
+    def task_queues_override(self):
+        return self.job_queues_override
+
+    @task_queues_override.setter
+    def task_queues_override(self, value):
```
HanlinMiao, line 322: Do we need a data validation check here? To make sure that value is boolean and such? Or maybe that's already handled by the booleanfield itself.

nautobot/extras/migrations/0117_task_queue_to_job_queue_migration.py
```diff
@@ -11,10 +12,13 @@ def migrate_task_queues_to_job_queues(apps, schema):
     ScheduledJob = apps.get_model("extras", "ScheduledJob")
     JobQueue = apps.get_model("extras", "JobQueue")
 
+    # should always create default job queue
+    JobQueue.objects.get_or_create(name="default", defaults={"queue_type": JobQueueTypeChoices.TYPE_CELERY})
     for job in Job.objects.all():
-        task_queues = job.task_queues
+        task_queues = job.task_queues or [settings.CELERY_TASK_DEFAULT_QUEUE]
```
HanlinMiao, line 18: For this case here, if we do not have any `task_queues` assigned to the job class meta and we use `[settings.CELERY_TASK_DEFAULT_QUEUE]` for the job model should `job_queues_override` to be set to True at this point?

nautobot/extras/forms/forms.py
```diff
@@ -956,7 +956,7 @@ class JobEditForm(NautobotModelForm):
     job_queues = DynamicModelMultipleChoiceField(
         label="Job Queues",
         queryset=JobQueue.objects.all(),
-        required=False,
+        required=False,  # needed due to job_queues_override checkbox behavior
```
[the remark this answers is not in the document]
HanlinMiao: Screenshot 2024-09-12 at 11 46 56 AM
We want users to assign at least one JobQueue to the Job right? Is this the expected behavior then?

nautobot/extras/models/jobs.py
glennmatthews: In general our ORM API doesn't do these kinds of explicit type checks, though that's not to say it *shouldn't*. Can add it if you think it's a significant risk.

nautobot/extras/migrations/0117_task_queue_to_job_queue_migration.py
glennmatthews: I don't think it's needed, no, an undefined `task_queues` implies `["default"]`, it's not really an override.

HanlinMiao: My issue is if the user goes to the edit page and does not check the `job_queues_override` checkbox or just does not do anything. The default queue will be cleared out which is not ideal imo.

glennmatthews: I *think* I fixed this in fccbd18 but please do double-check me.

HanlinMiao: Yep it's fixed thank you!

PR #6159, opened 2024-08-28
UI Components: Basic layout rendering

### LINKED ISSUE #6133: UI Components: Basic Layout rendering
Basic layout for rendering a details view and tab content.

Must include that Tabs will contain other panel components.

Layout should allow sub-panels to specify left, right, or full width. The order of sub-components, respective of the location, determines the order of the panel rendering. The standard Nautobot layout of a single left, right, and full width sections are supported.

The only tabs included in this layout are those that share the same view (advanced, and defined tabs). Explicitly not tabs which are separate views, such as Notes and Changelog.

glennmatthews:
# Closes #6133

Note that as this is rough work-in-progress code, I'm targeting a feature branch under `next` instead of directly `next` at this time.

# What's Changed

- Move existing UI code for nav and homepage from `nautobot/core/apps/__init__.py` to new `nautobot/core/ui/` submodules, just as a code-organization thing (no functional change in this area).
- Create `nautobot/core/ui/object_detail.py` module
   - Basically, this follows a similar pattern to the nav and homepage definitions - a detail view declares an `ObjectDetailContent`, which contains one or more `Tab` (abstract class), which contains one or more `Layout` (abstract class), which contains one or more `Panel` (abstract class).
   - Declaring an `ObjectDetailContent()` with no arguments produces a functional basic detail view listing all relevant object fields
   - If all that's needed is to customize which fields are listed and/or their order, can instead declare an `ObjectDetailContent(fields={"panel name": ["field1", "field2", ...], ...})`
   - If full control over the tabs and their layouts are needed, can instead declare as specifically and granularly as desired.
   - Concrete subclasses of Tab include `ObjectDetailMainTab` (all that most users will need to use) and skeletons for `ObjectDetailAdvancedTab`, `ObjectDetailContactsTab`, etc.
   - Concrete subclasses of Layout are `LayoutTwoColumn` and `LayoutFullWidth`
   - Concrete subclasses of Panel are currently `TemplatePanel` and `ObjectFieldsPanel`, with more to be added in future PRs.
- Update `generic/object_retrieve.html` to support `ObjectDetailContent` rendering as an alternative code path (old code path still supported; diffs are messy to read due to indentation changes as a result)
- As demonstration of proof of concept, replace Circuit detail view template with an appropriately configured `ObjectDetailContent` instance.
    - Note that `circuit_retrieve.html` is not deleted at this time, as doing so would break the example-app tab extension and other view-based tabs, but its `content_left_page` is removed from the template.

# Screenshots

image

# TODO

- [ ] Unit, Integration Tests


itdependsnetworks:
> Declaring an ObjectDetailContent() with no arguments produces a functional basic detail view listing all relevant object fields

Can you qualify that? Is there inference? I am not trying to scope creep (but clearly scope creeping), can we infer things like:

* model objects to be linkified
* placeholder on nullable fields
* pre block on json/yaml

I am sure there is a dozen more like that. If we do it from the start, makes less customization required and less likely we have to make a breaking change in the future.

Perhaps it does that all already, but just talking out loud.

glennmatthews:
> > Declaring an ObjectDetailContent() with no arguments produces a functional basic detail view listing all relevant object fields
> 
> Can you qualify that? Is there inference? I am not trying to scope creep (but clearly scope creeping), can we infer things like:
> 
>     * model objects to be linkified
> 
>     * placeholder on nullable fields
> 
>     * pre block on json/yaml
> 
> 
> I am sure there is a dozen more like that. If we do it from the start, makes less customization required and less likely we have to make a breaking change in the future.
> 
> Perhaps it does that all already, but just talking out loud.

That's all in the plan, if you look at the `ObjectFieldsPanel` class it handles some of the specific cases above already, but there's absolutely room for more intelligence to be added. Not all of that is in scope for this first story though.

itdependsnetworks:
fair to say this is where that logic is being built?

https://github.com/nautobot/nautobot/pull/6159/files#diff-820d40209d712a3c732a62dc3ba27210bf7549e2a975a4a72406d06028e48a3dR572-R588

nautobot/core/ui/object_detail.py
```diff
@@ -0,0 +1,598 @@
...
+
+        for field_name in fields:
+            field = instance._meta.get_field(field_name)
+            field_label = bettertitle(field.verbose_name)
+            field_value = getattr(instance, field.name)
+
+            if field_name in self.field_transforms:
+                field_display = field_value
+                for transform in self.field_transforms[field_name]:
+                    field_display = transform(field_display)
```
HanlinMiao, line 575: ~What would be a case where we have multiple field transforms for a single field?~

HanlinMiao: Nevermind I just have to look👀

nautobot/core/ui/object_detail.py
```diff
@@ -0,0 +1,598 @@
...
+                                  This addresses the common case where a detail view doesn't care to define custom tabs,
+                                  layouts, or panels, but wants some control over how the default `ObjectDetailMainTab`
+                                  presents the object's fields in an ObjectFieldsPanel(s).
+            field_transforms (dict): Optional dict of {field_name: [list of transform functions], ...}.
+                                     This can be used to provide custom rendering of given field(s) within the automatic
+                                     `ObjectDetailMainTab` class, for example `{"commit_rate": [humanize_speed]}`.
+        """
+        if tabs is None:
+            tabs = [ObjectDetailMainTab()]
+            if detail_fields is not None:
```
HanlinMiao, line 128: I wonder if we should enhance this a little. For example, I can imagine it might be easy to miss certain fields from the model if the user has to declare it manually every time. We should probably compare the declared `detail_fields` to maybe the fields on the serializer and give off a warning in the template `messages.warning`?  But again, in the legacy template, it is also dependent on the user to declare fields in the template. So Im not sure.

nautobot/circuits/views.py
```diff
@@ -127,8 +129,23 @@ class CircuitUIViewSet(NautobotUIViewSet):
...
+    object_detail_content = ObjectDetailContent(
+        detail_fields={
+            "Circuit": [
+                "status",
+                "provider",
+                "cid",
+                "circuit_type",
+                "tenant",
+                "install_date",
+                "commit_rate",
```
itdependsnetworks, line 141: Does it make sense to send directly as a django template variable, e.g. set this as `commit_rate|humanize_speed|placeholder`. I don't know if that is going overboard, but jut thinking out loud.

glennmatthews: The downside of that approach, I think, would be that we'd have to do multiple rendering passes on the template to get this to be interpreted as such, which would probably raise issues with escaping template content and so forth. I'll try and think about it a bit more though.

nautobot/core/ui/object_detail.py
glennmatthews: Yeah this is a tricky one to be sure. In the "default" case we'd just use `detail_fields="__all__"` but I'm not sure how many times that will actually be fit for purpose.

From just attempting the "Circuit" detail view, the `Comments` field is a separate `ObjectFieldsPanel` than the rest of the base fields, and the `circuit_termination_a` and `circuit_termination_z` fields are going to need a custom panel class of their own rather than being presented in an `ObjectFieldsPanel` at all. All of that is intentional and shouldn't result in a warning log that is noise and that the end user can't do anything about in any case. So I'm not sure how we'd automatically detect or warn developers if fields are inadvertently (as opposed to intentionally) omitted...

Hmm... maybe we could/should allow inverting this logic (similar to serializers) and have an `ObjectFieldsPanel` support an explicit `exclude` list of fields? Then we could warn if the ObjectDetailContent as a whole doesn't contain at least one `ObjectFieldsPanel` with `fields=__all__, [exclude=...]` as its attributes?

Definitely bears some further thought.

nautobot/core/ui/object_detail.py
```diff
@@ -0,0 +1,598 @@
...
+
+    def __init__(self, *, left_panels=None, right_panels=None, **kwargs):
+        """Define a two-column layout with the given left column and right column of panels."""
+        super().__init__(**kwargs)
+        if left_panels is None:
+            left_panels = []
+        if right_panels is None:
+            right_panels = []
+        self.left_panels = list(left_panels)
+        self.right_panels = list(right_panels)
```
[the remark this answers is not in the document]
glennmatthews:
```suggestion
        self.left_panels = list(left_panels) if left_panels is not None else []
        self.right_panels = list(right_panels) if right_panels is not None else []
```

nautobot/core/ui/object_detail.py
```diff
@@ -0,0 +1,598 @@
...
+
+class LayoutFullWidth(Layout):
+    """Layout class for full-width content."""
+
+    def __init__(self, *, panels=None, **kwargs):
+        """Define a full-width layout containing the given panels."""
+        super().__init__(**kwargs)
+        if panels is None:
+            panels = []
+        self.panels = list(panels)
```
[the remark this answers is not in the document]
glennmatthews:
```suggestion
        self.panels = list(panels) if panels is not None else []
```

nautobot/core/ui/object_detail.py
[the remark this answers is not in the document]
HanlinMiao: Oops sorry for the wrong suggestions

nautobot/core/ui/object_detail.py
glennmatthews: Added `exclude_fields` option in 7a794fdbc7785f78e71869d3173269374a473c93. We can refine this in a future PR.

glennmatthews:
Updated screenshot:

image

lampwins:
From design discussion:

```python
class UIViewSet():

    detail_view_layout = ObjectDetailView(
        layout="2_column_full_width"
        panels=[
            ObjectDetailModelFieldsPanel(
                name="foo",
                location="left",
                weight=100,
                fields=(
                    FieldSet("name", "status", "tenant", name="Object name"),
                    FieldSet("name", "status", "tenant", name="Object name"),
                )
            ),
            ObjectDetailTemplatePanel(
                name="foo",
                location="right",
                weight=100,
                template_content="<h1>helloo world</h1>"
            ),
        ],
        tabs=[
            ObjectDetailTab(
                name="foo",
                weight=100,
                layout="2_column_full_width"
                panels=[
                    ObjectDetailRelatedObjectTable(
                        name="foo",
                        location="full_width",
                        weight=100,
                        table_class=tables.FooTable,
                        table_data=  # existing patterns from DRF for accessors maybe?
                    ),
                    ObjectDetailTemplatePanel(
                        name="foo",
                        location="right",
                        weight=100,
                        template_content="<h1>helloo world</h1>"
                    ),
                ]
            )
        ]
    )
```

nautobot/circuits/views.py
```diff
@@ -127,8 +136,44 @@ class CircuitUIViewSet(NautobotUIViewSet):
...
+                                field_transforms={"comments": [render_markdown, placeholder]},
+                            ),
+                        ),
+                        right_panels=(
+                            TemplatePanel(
+                                label="Termination - A Side",
+                                # TODO: edit/swap/delete/detail buttons in panel header
+                                content_wrapper=TemplatePanel.ATTR_TABLE_CONTENT_WRAPPER,
+                                template_string="{% include 'circuits/inc/circuit_termination_fragment.html' with termination=circuit_termination_a %}",
+                            ),
```
HanlinMiao, line 163: We have already declared it in the template circuit_retrieve.html right? Do we have to have this here? Or vice versa?

HanlinMiao (approved):

nautobot/core/ui/object_detail.py
```diff
@@ -0,0 +1,567 @@
...
+                ),
+            ),
+        )
+
+    def should_render(self, context):
+        return (
+            context["object"].is_metadata_associable_model
+            and context["request"].user.has_perm("extras.view_objectmetadata")
+            and context["object"].associated_object_metadata.exists()
+        )
```
HanlinMiao, line 331: nitpick: should we swap the orders here?
```suggestion
        return (
            context["object"].associated_object_metadata.exists()
            and context["request"].user.has_perm("extras.view_objectmetadata")
            and context["object"].is_metadata_associable_model
        )
```

HanlinMiao: On second thought, do we need `context["object"].is_metadata_associable_model` here? `context["object"].associated_object_metadata.exists()` should just be enough right?

nautobot/core/ui/object_detail.py
```diff
@@ -0,0 +1,567 @@
...
+                ),
+            ),
+        )
+
+    def should_render(self, context):
+        return (
+            context["object"].is_dynamic_group_associable_model
+            and context["request"].user.has_perm("extras.view_dynamicgroup")
+            and context["object"].dynamic_groups.exists()
+        )
```
HanlinMiao, line 299: Same comment here

nautobot/circuits/views.py
glennmatthews: `circuit_termination.html` isn't used in the detail view any more (it's replaced by this code). I've left it in place in case of the unlikely possibility that some app is relying on it.

nautobot/core/ui/object_detail.py
glennmatthews: The `is_metadata_associable_model` check is cheap since it doesn't require a database hit, unlike the other two, so I think it makes sense as a quick-escape.

nautobot/core/ui/object_detail.py
glennmatthews: Same response. :-)

nautobot/circuits/views.py
HanlinMiao: understood

PR #6162, opened 2024-08-29
[WIP] add support for dirty group identification

lampwins:
This is another fun one from your favorite bug generator. Basically, this uses database triggers to identify when the dynamic group cache for a specific group _might_ be dirty, allowing the `update_cached_members()` method to selectively update groups based on the group's dirty flag. No tests, but it works on my local machine :)


nautobot/extras/signals.py
```diff
@@ -439,6 +440,143 @@ def dynamic_group_update_cached_members(sender, instance, **kwargs):
...
+
+            DROP TRIGGER IF EXISTS tr_d_group_dirty_{table_name} ON {table_name};
+            CREATE TRIGGER tr_d_group_dirty_{table_name}
+            AFTER DELETE ON {table_name}
+            FOR EACH STATEMENT
+            EXECUTE FUNCTION set_group_dirty_flag();
+            """
+
+        # MySQL specific trigger creation
+        elif db_vendor == 'mysql':
```
glennmatthews, line 516: Does this case cover Dolt as well?

glennmatthews (commented):
Interesting solution! Not gonna claim that I fully understand the implications yet but conceptually it makes sense to me.

nautobot/extras/signals.py
lampwins: Good question! I have no idea, I think so though.

PR #6163, opened 2024-08-29
feat: Added `get_http_session` to ExternalIntegration

abates:
# What's Changed

This feature provides a method to create a `requests.Session` initialized with the relevant configuration provided by an `ExternalIntegration`. This includes potential auth tokens, basic auth, ssl verification and custom CA file path. The reasoning behind this feature is to provide a re-usable interface for app developers to get http sessions for API clients defined as ExternalIntegrations. For instance, to interface with AWX a client needs the URL, ssl verification info, token or username/password. These can all be defined in `ExternalIntegration` and `get_http_session` can then be used to initialize an API client.

# Screenshots
This is an internal API available for APP developers, so there aren't any UI components.

# TODO
- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/models/models.py
+++ b/nautobot/extras/models/models.py
@@ -524,7 +524,6 @@ class ExternalIntegration(PrimaryModel):
 
     def get_http_session(
         self,
-        accept="application/json",
         include_auth=True,
         authorization_header="Authorization",
         authorization_prefix="Bearer ",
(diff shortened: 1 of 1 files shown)


nautobot/extras/models/models.py
```diff
@@ -517,6 +522,73 @@ def __str__(self):
...
+        Raises:
+            ValueError: If `include_auth` is set and no appropriate secrets are found.
+
+        Returns:
+            requests.Session: The session object with all the configured attributes.
+        """
+        session = requests.Session()
+        session.verify = self.ca_file_path or self.verify_ssl
+        if session.verify is False:
+            urllib3.disable_warnings(category=urllib3.exceptions.InsecureRequestWarning)
```
glennmatthews, line 563: Feels off to me that we disable this warning globally. Can it be disabled just in the context of the specific session isntead?

abates: I agree, this does feel a little off. I'll investigate if we can restrict it, the way I did it is based on what little documentation I could find. Perhaps we shouldn't do anything, if you've disable SSL verification you probably should get an obnoxious amount of warning messages.

glennmatthews:
Closing this for now as there doesn't seem to be active interest in this feature. Can reopen later if interest is established.

PR #529, opened 2021-06-03
Add VLAN Min and Max VID to VLAN Group

scetron:
### Fixes: #526 

Added min_vid and max_vid to VLANGroup Model.
Add fields to VLANGroup View and Forms.
Added fields to VLANGroup html template.
Added fields to API views (VLANGroup and NestedVLANGroup).
Added validation when manually adding a new VLAN to the group.
Adjusted get_next to pick a free VLAN within the configured min and max vid.

PR #6176, opened 2024-09-03
VLANGroup range field

mzbroch:
Closes #529 
Replaces #526 

# What's Changed
This feature adds a mandatory `Range` field on a `VlanGroup` model with a default value of `1-4094` (which equals to All Vlans in a well known range). `Range` Field allows to constrain `Vlan` members of the `VlanGroup`.

Example ranges are as follows:
```
- 10
- 10-20
- 10-20, 30-50
- 10-20, 30-50, 100
```
The implementation supports non-continuous set of ranges as per screenshot.

Screenshot 2024-09-16 at 11 22 48
Screenshot 2024-09-16 at 11 22 28

# TODO

- [ ] API representation (string type, ie `"1-4094"`)
- [x] API `get_next` - similar to get next for IP
- [x] Update vlan table in vlangroup detail view
- [x] Unit, Integration Tests


nautobot/ipam/models.py
```diff
@@ -1293,6 +1293,20 @@ class VLANGroup(OrganizationalModel):
...
     description = models.CharField(max_length=CHARFIELD_MAX_LENGTH, blank=True)
 
+    range = JSONArrayField(
+        base_field=models.PositiveIntegerField(
+            validators=[
+                MinValueValidator(constants.VLAN_VID_MIN),
+                MaxValueValidator(constants.VLAN_VID_MAX),
+            ]
+        ),
+        verbose_name="Allowed vlan identifiers",
```
glennmatthews, line 1303: So in the default case, this will store the default range 1-4094 as an expanded JSON array, i.e. `[1, 2, 3, 4, 5, ..., 4092, 4093, 4094]`?  How impactful is that on storage size and performance? Wonder if it might be preferable or even necessary to store the collapsed range as a list of `(start, end)` tuples instead of fully expanding it out into a list in the DB?

mzbroch: I did some measurements and they are as follows

```
nautobot=# SELECT name, pg_column_size(range) as rowsize FROM ipam_vlangroup;
    name    | rowsize
------------+---------
 test1      |      17
 test1-10   |     125
 test1-100  |    1208
 test1-1000 |    3410
 test1-4094 |   13873
(5 rows)
```

Each 1 "megabyte" would store 72 vlan groups with full range of 1-4094.

While this is not optimal with regards storage, we could keep the code cleaner and keep this contribution simple. On the implementation side of things, I also don't see this model used a lot.

glennmatthews: Hmm, so even `1-100` is already over 1k size... I'd be concerned about query performance issues with Postgres TOAST (a la https://hakibenita.com/sql-medium-text-performance), and especially with the default case being `1-4094`.

lampwins: Yeah I have to agree, I'm nut sure I understand why we need to expand the array for storage? Especially when the range containment math, and even range optimization, is relatively trivial.

glennmatthews (commented):
Conceptually I like the revised approach a lot better.

I do wonder whether it would be cleaner to implement most of this logic as a custom reusable `CharField` subclass (see examples in `nautobot.core.models.fields`) rather than as methods on the VLANGroup model itself, but that's just a matter of code organization.

mzbroch:
> I do wonder whether it would be cleaner to implement most of this logic as a custom reusable `CharField` subclass (see examples in `nautobot.core.models.fields`) rather than as methods on the VLANGroup model itself, but that's just a matter of code organization.

I implemented `PositiveRangeNumberTextField` now, indeed this looks cleaner.

nautobot/ipam/models.py
```diff
@@ -1293,21 +1302,63 @@ class VLANGroup(OrganizationalModel):
...
+            raise ValidationError(
+                {"range": f"VLAN group range is outside of allowed {constants.VLAN_VID_MIN}-{constants.VLAN_VID_MAX} "
+                          f"range."}
+            )
+
+        # Validate ranges for related VLANs.
+        if self.vlans.all() and any([_vlan.vid not in self.expanded_range for _vlan in self.vlans.all()]):
+            raise ValidationError(
+                {"range": "VLAN group range may not be re-sized due to existing VLANs."}
+            )
```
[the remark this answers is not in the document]
mzbroch: Added

nautobot/ipam/models.py
```diff
@@ -1293,21 +1302,63 @@ class VLANGroup(OrganizationalModel):
...
+        converted_ranges = convert_to_ranges(self.expanded_range)
+        return ",".join([f"{x[0]}" if x[0] == x[1] else f"{x[0]}-{x[1]}" for x in converted_ranges])
+
+    @property
+    def expanded_range(self):
+        """Expand VLAN range.
+
+        '0-3,5' => [0, 1, 2, 3, 5]
+        """
+        return parse_numeric_range(self.range)
```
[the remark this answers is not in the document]
mzbroch: Fixed

nautobot/ipam/models.py
```diff
@@ -1295,33 +1296,61 @@ class VLANGroup(OrganizationalModel):
...
+        blank=False, default="1-4094", help_text="Vlan range, default 1-4094 if left blank."
+    )
+
     class Meta:
         ordering = ("name",)
         verbose_name = "VLAN group"
         verbose_name_plural = "VLAN groups"
 
+    @property
+    def available_vids(self):
```
glennmatthews, line 1309: docstring?

nautobot/ipam/models.py
```diff
@@ -1295,33 +1296,61 @@ class VLANGroup(OrganizationalModel):
...
+        _expanded_range = parse_numeric_range(self.range)
+
+        # Validate range boundaries
+        if _expanded_range[0] < constants.VLAN_VID_MIN or _expanded_range[-1] > constants.VLAN_VID_MAX:
+            raise ValidationError(
+                {
+                    "range": f"VLAN group range is outside of allowed {constants.VLAN_VID_MIN}-{constants.VLAN_VID_MAX} "
+                    f"range."
+                }
+            )
```
glennmatthews, line 1335: This check feels like it could be implemented directly in the `PositiveRangeNumberTextField` class perhaps, or done with an appropriate Django `validator` implementation?

nautobot/ipam/models.py
```diff
@@ -1440,6 +1469,17 @@ def get_vminterfaces(self):
         # Return all VM interfaces assigned to this VLAN
         return VMInterface.objects.filter(Q(untagged_vlan_id=self.pk) | Q(tagged_vlans=self.pk)).distinct()
 
+    def clean(self):
+        super().clean()
+
+        # Validate Vlan Group Range
+        if self.vlan_group:
+            _vlan_group_expanded_range = parse_numeric_range(self.vlan_group.range)
+            if self.vid not in _vlan_group_expanded_range:
```
glennmatthews, line 1478: Might be nice to add a `contains(value)` method to `PositiveRangeNumberTextField` as a convenience helper?

nautobot/ipam/tests/test_utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class ParseNumericRangeTest(TestCase):
+    """Tests for add_available_vlans()."""
+
+    def test_input(self):
+        self.assertEqual(parse_numeric_range(input_string=[1, 2, 3]), [])
+        self.assertEqual(parse_numeric_range(input_string=None), [])
+        self.assertEqual(parse_numeric_range(input_string=1), [])
+        self.assertEqual(parse_numeric_range(input_string=""), [])
```
glennmatthews, line 37: Might be better for these cases to raise a ValidationError or TypeError instead of returning?

glennmatthews (commented):
I like how this is developing!

nautobot/ipam/models.py
mzbroch: Moved to `validate` method under `PositiveRangeNumberTextField`
Implemented min_boundary and max_boundary on `PositiveRangeNumberTextField`

nautobot/ipam/models.py
mzbroch: I am not sure if this can be done. `self.expanded_value` would be actually assigned (computed) during `to_python`

nautobot/ipam/tests/test_utils.py
mzbroch: Added

nautobot/core/models/fields.py
```diff
@@ -11,8 +11,10 @@
 
 from nautobot.core.constants import CHARFIELD_MAX_LENGTH
 from nautobot.core.forms import fields, widgets
+from nautobot.core.forms.utils import parse_numeric_range
 from nautobot.core.models import ordering
 from nautobot.core.models.managers import TagsManager
+from nautobot.core.models.utils import compress_range
```
glennmatthews, line 17: Wondering if `compress_range` should be implemented in `forms.utils` just to be more self consistent?

nautobot/core/models/fields.py
```diff
@@ -415,3 +417,48 @@ def formfield(self, form_class=fields.DynamicModelMultipleChoiceField, **kwargs)
         kwargs.setdefault("required", False)
         kwargs.setdefault("query_params", {"content_types": self.model._meta.label_lower})
         return super().formfield(form_class=form_class, **kwargs)
+
+
+class PositiveRangeNumberTextField(models.TextField):
+    default_error_messages = {
+        "invalid": "Invalid value. Specify a value using positive integers in a range format (i.e. '10-20').",
```
glennmatthews, line 424: Zero is valid for this field in general, right?

```suggestion
        "invalid": "Invalid value. Specify a value using non-negative integers in a range format (i.e. '10-20').",
```

nautobot/core/models/fields.py
```diff
@@ -415,3 +417,48 @@ def formfield(self, form_class=fields.DynamicModelMultipleChoiceField, **kwargs)
...
+    }
+
+    description = "A text based representation of positive number range."
+
+    def __init__(self, min_boundary=None, max_boundary=None, *args, **kwargs):
```
glennmatthews, line 429: Should `min_boundary` default to 0 perhaps?

nautobot/core/models/fields.py
```diff
@@ -415,3 +417,48 @@ def formfield(self, form_class=fields.DynamicModelMultipleChoiceField, **kwargs)
...
+        """
+        super().validate(value, model_instance)
+
+        if self.min_boundary and self.max_boundary:
+            if self.expanded[0] < self.min_boundary or self.expanded[-1] > self.max_boundary:
+                raise exceptions.ValidationError(
+                    message=f"Invalid value. Specify a range value between {self.min_boundary}-{self.max_boundary}",
+                    code="outofrange",
+                    params={"value": value},
+                )
```
glennmatthews, line 464: Should we support specifying one boundary but not the other? At the very least I'd think the default behavior would be to have a `min_boundary` of 0 but no `max_boundary`?

nautobot/core/models/utils.py
```diff
@@ -260,3 +260,20 @@ def construct_natural_slug(values, pk=None):
...
+        [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 100, 101, 102, 103, 104, 105, 1000, 1100, 1101, 1102, 1103, 1104, 1105, 1106]
+        =>
+        iter1: (1, 10)
+        iter2: (100, 105)
+        iter3: (1000, 1000)
+        iter4: (1100, 1106)
+    """
+    iterable = sorted(set(iterable))
+    for _, grp in groupby(enumerate(iterable), lambda t: t[1] - t[0]):
+        grp = list(grp)
```
glennmatthews, line 278: Why's this conversion necessary?

nautobot/ipam/api/views.py
```diff
@@ -391,6 +391,51 @@ class VLANGroupViewSet(NautobotModelViewSet):
     serializer_class = serializers.VLANGroupSerializer
     filterset_class = filters.VLANGroupFilterSet
 
+    @action(
+        detail=True,
+        name="Available VLAN IDs",
+        url_path="available-vids",
+        methods=["get"],
```
glennmatthews, line 398: Do we plan to add POST support here in the future for allocating VLANs from this endpoint? If so, I think this should be `available-vlans` rather than `available-vids`, personally.

nautobot/ipam/forms.py
```diff
@@ -722,11 +722,14 @@ class IPAddressFilterForm(NautobotFilterForm, TenancyFilterForm, StatusModelFilt
 
 
 class VLANGroupForm(LocatableModelFormMixin, NautobotModelForm):
+    range = forms.CharField(required=False, help_text="Vlan range, default 1-4094 if left blank.")
```
glennmatthews, line 725: Is this explicit field definition needed here? Can we make this the default representation for a `PositiveRangeNumberTextField` instead? (https://docs.djangoproject.com/en/4.2/howto/custom-model-fields/#specifying-form-field-for-model-field)

Also, why is this `required=False`?

nautobot/ipam/models.py
```diff
@@ -1299,33 +1300,62 @@ class VLANGroup(OrganizationalModel):
     )
     description = models.CharField(max_length=CHARFIELD_MAX_LENGTH, blank=True)
 
+    range = PositiveRangeNumberTextField(
+        blank=False,
+        default="1-4094",
+        help_text="Vlan range, default 1-4094 if left blank.",
```
glennmatthews, line 1306:
```suggestion
        help_text='Permitted VID range(s) as comma-separated list, default "1-4094" if left blank.',
```

nautobot/ipam/models.py
```diff
@@ -1299,33 +1300,62 @@ class VLANGroup(OrganizationalModel):
...
+        min_boundary=constants.VLAN_VID_MIN,
+        max_boundary=constants.VLAN_VID_MAX,
+    )
+
+    @property
+    def expanded_range(self):
+        """
+        Expand VLAN's range into a list of integers (VLAN IDs).
+        """
+        return sorted(parse_numeric_range(self.range))
```
glennmatthews, line 1316: `parse_numeric_range` is already updated to sort the list.

nautobot/ipam/models.py
```diff
@@ -1299,33 +1300,62 @@ class VLANGroup(OrganizationalModel):
...
-
         # Validate location
         if self.location is not None:
             if ContentType.objects.get_for_model(self) not in self.location.location_type.content_types.all():
                 raise ValidationError(
                     {"location": f'VLAN groups may not associate to locations of type "{self.location.location_type}".'}
                 )
 
+        # Validate ranges for related VLANs.
+        out_of_range_vids = [_vlan.vid for _vlan in self.vlans.all() if _vlan.vid not in self.expanded_range]
```
glennmatthews, line 1343: is this going to re-evaluate `self.expanded_range` N times? Do we need to load it into a local variable first?

nautobot/ipam/models.py
```diff
@@ -1299,33 +1300,62 @@ class VLANGroup(OrganizationalModel):
...
         """
-        Return the first available VLAN ID (1-4094) in the group.
+        Return the first available VLAN ID in the group's range.
         """
-        vlan_ids = VLAN.objects.filter(vlan_group=self).values_list("vid", flat=True)
-        for i in range(1, 4095):
-            if i not in vlan_ids:
-                return i
-        return None
+        return self.available_vids[0] if self.available_vids else None
```
glennmatthews, line 1358: Is this going to evaluate `self.available_vids` twice? Do we need to load it into a local variable?

nautobot/ipam/models.py
```diff
@@ -1444,6 +1474,13 @@ def get_vminterfaces(self):
         # Return all VM interfaces assigned to this VLAN
         return VMInterface.objects.filter(Q(untagged_vlan_id=self.pk) | Q(tagged_vlans=self.pk)).distinct()
 
+    def clean(self):
+        super().clean()
+
+        # Validate Vlan Group Range
+        if self.vlan_group and self.vid not in self.vlan_group.expanded_range:
+            raise ValidationError({"vid": f"Vlan ID is not contained in Vlan Group Range ({self.vlan_group.range})"})
```
glennmatthews, line 1482:
```suggestion
            raise ValidationError({"vid": f"VLAN ID is not contained in VLAN Group range ({self.vlan_group.range})"})
```

nautobot/ipam/templates/ipam/vlangroup.html
```diff
@@ -23,6 +23,10 @@
                         <a href="{% url 'ipam:vlan_list' %}?vlan_group={{ object.pk }}">{{ vlans_count }}</a>
                     </td>
                 </tr>
+                <tr>
+                    <td>Range</td>
+                    <td>{{ object.range }}</td>
+                </tr>
```
glennmatthews, line 29: Possibly move this above `VLANs`?

nautobot/ipam/tests/test_utils.py
```diff
@@ -0,0 +1,59 @@
...
+        ]
+
+        for x in invalid_inputs:
+            with self.assertRaises(TypeError) as exc:
+                parse_numeric_range(input_string=x)
+            self.assertEqual(str(exc.exception), "Input value must be a string using a range format.")
+
+        with self.assertRaises(ValueError) as exc:
+            parse_numeric_range(input_string="3-")
+        self.assertEqual(str(exc.exception), "invalid literal for int() with base 10: ''")
```
glennmatthews, line 59: Not the most helpful error message in this case?

nautobot/core/models/fields.py
mzbroch: Added 👍

nautobot/ipam/models.py
mzbroch: Good catch!

nautobot/ipam/models.py
mzbroch: 👍

nautobot/ipam/forms.py
mzbroch: This is now fixed

nautobot/ipam/tests/test_utils.py
mzbroch: Fixed

nautobot/ipam/api/views.py
```diff
@@ -391,6 +391,51 @@ class VLANGroupViewSet(NautobotModelViewSet):
...
     filterset_class = filters.VLANGroupFilterSet
 
+    @action(
+        detail=True,
+        name="Available VLAN IDs",
+        url_path="available-vlans",
+        methods=["get"],
+        filterset_class=None,
+    )
+    def available_vlans(self, request, pk=None):
```
glennmatthews, line 401: At some point will need to `extend_schema` here to get the correct representation in OpenAPI/Swagger. I'm fine with that coming in a future PR but let's make sure not to forget about it.

nautobot/ipam/models.py
```diff
@@ -1299,33 +1300,64 @@ class VLANGroup(OrganizationalModel):
...
+
     class Meta:
         ordering = ("name",)
         verbose_name = "VLAN group"
         verbose_name_plural = "VLAN groups"
 
+    @property
+    def available_vids(self):
+        """
+        Return all available VLAN IDs within this VlanGroup as a list.
```
glennmatthews, line 1326:
```suggestion
        Return all available VLAN IDs within this VLANGroup as a list.
```

glennmatthews (approved):
Great stuff!

nautobot/core/forms/utils.py
```diff
@@ -18,24 +19,26 @@
...
 
 
-def parse_numeric_range(string, base=10):
+def parse_numeric_range(input_string, base=10):
     """
-    Expand a numeric range (continuous or not) into a decimal or
+    Expand a numeric range (continuous or not) into a sorted decimal or
     hexadecimal list, as specified by the base parameter
       '0-3,5' => [0, 1, 2, 3, 5]
       '2,8-b,d,f' => [2, 8, 9, a, b, d, f]
```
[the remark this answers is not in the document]
mzbroch: Added! Although, we do not use base=16 anywhere in Nautobot.

nautobot/ipam/api/views.py
mzbroch: Thanks, yes indeed, this will be added

nautobot/core/models/utils.py
mzbroch: _The returned group is itself an iterator that shares the underlying iterable with [groupby()](https://docs.python.org/3/library/itertools.html#itertools.groupby). Because the source is shared, when the [groupby()](https://docs.python.org/3/library/itertools.html#itertools.groupby) object is advanced, the previous group is no longer visible. So, if that data is needed later, it should be stored as a list:_

source: https://docs.python.org/3/library/itertools.html#itertools.groupby

glennmatthews (commented):
What content should be added to `nautobot/docs/user-guide/core-data-model/ipam/vlangroup.md`? I think it'd be useful to have some examples of complex or non-contiguous values in the range.

mzbroch:
> What content should be added to `nautobot/docs/user-guide/core-data-model/ipam/vlangroup.md`? I think it'd be useful to have some examples of complex or non-contiguous values in the range.

@glennmatthews Docs added via #6273

PR #6186, opened 2024-09-04
Fix #6184

### LINKED ISSUE #6184: Filter Extensions brake DynamicGroup Edit View when custom filter method is used.
### Environment
* Nautobot version (Docker tag too if applicable): 2.2.4
* Python version: 3:11

Filter extensions implemented under apps `filter_extensions.py` brake Dynamic Groups on rendering edit view.
This happens when a custom filter is implemented and a custom method is passed to a filter class.

For instance such a filter:
```
# filter_extensions.py

def custom_method(queryset, field, value):
    <custom logic>
    return queryset


class DeviceFilterExtension(FilterExtension):
    """Extends Device Filters."""

    model = "dcim.device"

    filterset_fields = {
        "nautobot_my_plugin_name_filter_name": BooleanFilter(
            method=custom_method, label="Extended Filter"
        )
    }


filter_extensions = [InventoryItemFilterExtension]
```
brakes on string concatenation here: https://github.com/nautobot/nautobot/blob/15c8b869bb76eb806e4a5a44302bcc06cebf5679/nautobot/extras/models/groups.py#L254
because in this case `method` attribute is a function.

### Steps to Reproduce
1. Create filter_extensions.py for `dcim.device` under an app like in the example above.
2. Create a Dynamic group for `dcim.device`. 
3. Edit view brakes on the second stage when the form renders available filters.

### Expected Behavior
Filters are correctly rendered in Edit View.

### Observed Behavior
Edit View brakes after creation.

pszulczewski:
# Closes #6184 
# What's Changed
Additional control on string concatenation in extras.models.groups in L255
# TODO
- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/models/groups.py
+++ b/nautobot/extras/models/groups.py
@@ -251,7 +251,10 @@ class DynamicGroup(PrimaryModel):
             # Skip filter fields that have methods defined. They are not reversible.
             if skip_method_filters and filterset_field.method is not None:
                 # Don't skip method fields that also have a "generate_query_" method
-                if hasattr(filterset, "generate_query_" + filterset_field.method):
+                query_attr = (
+                    filterset_field.method.__name__ if callable(filterset_field.method) else filterset_field.method
+                )
+                if hasattr(filterset, f"generate_query_{query_attr}"):
                     logger.debug(
                         "Keeping %s for filterform: has a `generate_query_` filter method", filterset_field_name
                     )
--- a/nautobot/extras/tests/test_dynamicgroups.py
+++ b/nautobot/extras/tests/test_dynamicgroups.py
@@ -494,6 +494,17 @@ class DynamicGroupModelTest(DynamicGroupTestBase):  # TODO: BaseModelTestCase mi
         self.assertIsInstance(fields["cluster"], MultiMatchModelMultipleChoiceField)
         self.assertIsInstance(fields["cluster"].widget, APISelectMultiple)
 
+    def test_map_filter_fields_with_custom_filter_method(self):
+        """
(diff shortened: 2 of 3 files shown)


nautobot/extras/models/groups.py
```diff
@@ -248,11 +248,13 @@ def _map_filter_fields(self):
 
         # Filter out unwanted fields from the filterform
         for filterset_field_name, filterset_field in filterset_fields.items():
-
             # Skip filter fields that have methods defined. They are not reversible.
             if skip_method_filters and filterset_field.method is not None:
                 # Don't skip method fields that also have a "generate_query_" method
-                if hasattr(filterset, "generate_query_" + filterset_field.method):
+                query_attr = (
+                    filterset_field.method if isinstance(filterset_field.method, str) else filterset_field.method.name
```
joewesch, line 255: As many things can be cast to a string, I think you should reverse the check here and do `if callable(filterset_field.method)` instead.

glennmatthews, line 255: Should that be `__name__` rather than `name`?

nautobot/extras/tests/test_dynamicgroups.py
```diff
@@ -494,6 +494,17 @@ def test_map_filter_fields(self):
         self.assertIsInstance(fields["cluster"], MultiMatchModelMultipleChoiceField)
         self.assertIsInstance(fields["cluster"].widget, APISelectMultiple)
 
+    def test_map_filter_fields_with_custom_filter_method(self):
```
pszulczewski, line 497: This test is failing without the fix.

glennmatthews (approved):
`develop` is the appropriate branch for a bug fix in 2.x, thanks!

PR #6213, opened 2024-09-13
Changed `--cache-test-fixtures` and `--keepdb` flags from opt-in to opt-out for `inv unittest` and `inv integration-test` commands.

### LINKED ISSUE #5859: Make unittest `--cache-test-fixtures` and `--keepdb` arguments default-true
### Proposed Changes

Change the `invoke unittest` arguments `--cache-test-fixtures` and `--keepdb` to be enabled by default and opt-out, instead of their current opt-in behavior. #5858 might be a prerequisite.

Maybe also make it so that if not specifying a particular `--label`, we should default-enable the `--parallel` flag since running the full test suite is so time consuming otherwise?

### Justification

Provide well-performing default behavior for the task, reducing the amount of "team knowledge" required for new developers to have a good development experience.

HanlinMiao:
# Closes #5859 
# What's Changed

Changed `--cache-test-fixtures` and `--keepdb` flags from opt-in to opt-out for `inv unittest` and `inv integration-test` commands.
Changed `inv unittest` to automatically include `--parallel` flag when running the entire unit test suite.
# Screenshots

# TODO

- [x] Unit, Integration Tests


tasks.py
```diff
@@ -750,7 +750,7 @@ def unittest(
         command += " --buffer"
     if verbose:
         command += " --verbosity 2"
-    if parallel:
+    if parallel or label == "nautobot":
```
glennmatthews, line 753: Can we do anything to auto-select a `parallel_workers` value that doesn't overly tax the system? I've found when running locally that just `--parallel` defaults to a number of workers (8 on my Mac) that results in 100% CPU consumption while tests are running, which tends to make the rest of the system pretty unhappy, so I usually use `--parallel-workers=6` so as to leave some overhead free for the system. Obviously in CI we'd want to use the default for optimal runtime, but in the local dev case I think it'd be desirable to throttle it somewhat.

tasks.py
```diff
@@ -707,8 +707,8 @@ def check_schema(context, api_version=None):
 )
 def unittest(
     context,
-    cache_test_fixtures=False,
-    keepdb=False,
+    cache_test_fixtures=True,
+    keepdb=True,
```
glennmatthews, line 711: Should CI now do `--no-cache-test-fixtures` and/or `--no-keepdb`? Does it have any measurable performance impact either way?

tasks.py
glennmatthews: In fairness this may be a self-inflicted wound, I need to change my Docker resource allocation so that it's not allowed to use all my CPUs at once. :-)

tasks.py
HanlinMiao: I am a little confused. CI was running with `--cache-test-fixtures` and `--keepdb` in the first place. Why would we want to opt out?

glennmatthews: Huh, that seems counterintuitive to me. What's the point of caching the test fixtures and keeping the DB when each CI test execution runs in isolation?

HanlinMiao: I can explicitly add `--no-cache-test-fixtures` and `--no-keepdb` to the CI

changes/5859.housekeeping
```diff
@@ -0,0 +1,2 @@
+Changed `--cache-test-fixtures` and `--keepdb` flags from opt-in to opt-out for `inv unittest` and `inv integration-test` commands.
```
glennmatthews, line 1: Do any of our development/testing docs need to be updated to describe this and explain when you might want to opt out?

changes/5859.housekeeping
```diff
@@ -0,0 +1,2 @@
...
+Changed `inv unittest` to automatically include `--parallel` flag when running the entire unit test suite.
```
glennmatthews, line 2:
```suggestion
Changed `--cache-test-fixtures` and `--keepdb` flags from opt-in to opt-out for `invoke unittest` and `invoke integration-test` commands.
Changed `invoke unittest` to automatically include `--parallel` flag when running the entire unit test suite.
```

timizuoebideri1 (approved):

nautobot/docs/development/core/getting-started.md
```diff
@@ -636,12 +636,12 @@ Unit tests are run using the `invoke unittest` command (if using the Docker deve
...
 * `--failfast` - Fail as soon as any test failure or error condition is encountered, instead of running to completion.
-* `--keepdb` - Save and reuse the initialized test database between test runs. Can significantly improve the initial startup time of the test suite after your first test run. **Do not use if you're actively making changes to model definitions or migrations.**
+* `--keepdb` - Save and reuse the initialized test database between test runs. Can significantly improve the initial startup time of the test suite after your first test run. This parameter is automatically included when running unit tests. To not use the saved test database, include the `--no-keepdb` parameter. **The `--no-keepdb` option is mandatory if you're actively making changes to model definitions or migrations.**
```
glennmatthews, line 641: Can we flip these and document them as `--no-cache-test-fixtures` and `--no-keepdb` similar to `--no-buffer` below?

nautobot/docs/development/core/getting-started.md
```diff
@@ -636,12 +636,12 @@ Unit tests are run using the `invoke unittest` command (if using the Docker deve
...
 * `--label <module.path>` - Only run the specific subset of tests. Can be broad (`--label nautobot.core.tests`) or specific (`--label nautobot.core.tests.test_graphql.GraphQLQueryTestCase`).
 * `--no-buffer` - Allow stdout/stderr output from the test to be seen in your terminal, instead of being hidden. **If you're debugging code with `breakpoint()`, you should use this option, as otherwise you'll never see the breakpoint happen.**
-* `--parallel` - Split the tests across multiple parallel subprocesses. Can greatly reduce the runtime of the entire test suite when used. Auto-detects the number of workers if not specified with `--parallel-workers`.
+* `--parallel` - Split the tests across multiple parallel subprocesses. Can greatly reduce the runtime of the entire test suite when used. Auto-detects the number of workers if not specified with `--parallel-workers`. This parameter is automatically included when running the entire test suite to reduce unit test run time.
```
glennmatthews, line 644:
```suggestion
* `--parallel` - Split the tests across multiple parallel subprocesses. Can greatly reduce the runtime of the entire test suite when used. Auto-detects the number of workers if not specified with `--parallel-workers`. This parameter is automatically included if no `--label` is specified.
```

glennmatthews (approved):

timizuoebideri1 (approved):

PR #6228, opened 2024-09-17
Fixed SavedView functionality not working in Rack Elevation List View.

### LINKED ISSUE #6217: Saved views not displayed in Saved View drop down for Rack Elevation Table view
### Environment
* Nautobot version 2.3.2
* Python version: 2.11

### Steps to Reproduce
1. Navigate to `/dcim/rack-elevations/`
2. Click Saved Views drop down and click `Save as New View`
3.  Fill out form and recieve the message `Successfully created new Saved View 'Test View'.`
4. Click on Saved Views drop down and created view is not present. 
Screenshot 2024-09-14 at 12 20 08 PM
5. Verify Saved view was created in change log
Screenshot 2024-09-14 at 12 21 47 PM
6. Clicking the URL in the change log redirects to the saved elevation view. I verified that if filters are set they are present, but the name does not appear to the right of Rack Elevation like it does with other saved views.
7. Verified on local docker install and demo.nautobot.com

### Expected Behavior
I expect to be able to create, share and reuse saved views for elevation for ease of visualizing available rack space by location.

### Observed Behavior
Saved views are not saved in the Saved View drop down

HanlinMiao:
# Closes #6217 
# What's Changed

`RackElevationListView` uses `Rack` as its model however its view_name is `dcim:rack_elevation_list` instead of the traditional `{app_label}:{model.verbose_name_lower()}_list`. So fixed that and some minor SavedView logic changes.

# Screenshots

Screenshot 2024-09-17 at 3 03 03 PM
Screenshot 2024-09-17 at 3 03 18 PM

# TODO

- [ ] Unit, Integration Tests


nautobot/dcim/views.py
```diff
@@ -576,6 +576,7 @@ class RackElevationListView(generic.ObjectListView):
     filterset = filters.RackFilterSet
     action_buttons = []
     template_name = "dcim/rack_elevation_list.html"
+    view_name = "dcim:rack_elevation_list"
```
glennmatthews, line 579: Maybe `list_url_name` would be a less generic and more intuitive label?

nautobot/core/views/generic.py
```diff
@@ -205,13 +206,11 @@ def get(self, request):
         filter_form = None
         hide_hierarchy_ui = False
         clear_view = request.GET.get("clear_view", False)
+        view_name = self.view_name or validated_viewname(model, "list")
```
glennmatthews, line 209: Wonder if we could use [`resolve(request.path).url_name`](https://docs.djangoproject.com/en/4.2/ref/urlresolvers/#resolve) as a "smarter" solution in general here that might remove the name for the `view_name` parameter altogether:

nautobot/dcim/views.py
glennmatthews: Is this still even necessary after the above refactor?

nautobot/core/views/generic.py
```diff
@@ -205,13 +205,12 @@ def get(self, request):
         filter_form = None
         hide_hierarchy_ui = False
         clear_view = request.GET.get("clear_view", False)
+        resolved_path = resolve(request.path)
+        view_name = f"{resolved_path.app_name}:{resolved_path.url_name}"
```
glennmatthews, line 209: I'd suggest changing `view_name` to `list_url` for clarity if that doesn't cause any problems.

nautobot/core/views/generic.py
```diff
@@ -302,28 +301,30 @@ def get(self, request):
...
-                SavedView.objects.filter(view=list_url, is_shared=True).order_by("name").only("pk", "name")
+                SavedView.objects.filter(view=view_name, is_shared=True).order_by("name").only("pk", "name")
             )
             user_owned_saved_views = (
-                SavedView.objects.filter(view=list_url, owner=user).order_by("name").only("pk", "name")
+                SavedView.objects.filter(view=view_name, owner=user).order_by("name").only("pk", "name")
             )
             saved_views = shared_saved_views | user_owned_saved_views
+
+        # Construct the objects table
```
glennmatthews, line 317: Comment seems out of place - shouldn't it be inside the `if self.table` block still?

nautobot/core/views/generic.py
```diff
@@ -302,28 +301,30 @@ def get(self, request):
...
+
         if self.table:
-            # Construct the objects table
-            if current_saved_view_pk:
-                try:
-                    # We are not using .restrict(request.user, "view") here
-                    # User should be able to see any saved view that he has the list view access to.
-                    current_saved_view = SavedView.objects.get(view=list_url, pk=current_saved_view_pk)
-                except ObjectDoesNotExist:
-                    messages.error(request, f"Saved view {current_saved_view_pk} not found")
```
glennmatthews, line 326: Is this move fixing a bug, and if so what was the issue?

HanlinMiao: Yeah so current_saved_view is only available if there is a self.table declared. That is not the case in RackElevationListView. However, it should still be able to use SavedView

glennmatthews (approved):

PR #6245, opened 2024-09-19
Enhanced job delete functions to prevent users with the appropriate permissions or super users from deleting system jobs from the UI and the API.

### LINKED ISSUE #6057: Prevent user deletion of Job records corresponding to System jobs
### As ...

Patti - Platform Admin

### I want ...

The Job records corresponding to Nautobot's built-in "system" Jobs to be protected against inadvertent deletion by users.

### So that ...

System jobs are always present and registered in Nautobot.

- Even for a user with `extras.delete_job` permissions, attempting to delete a "system" Job record via the UI or REST API is reported as an error.
- Even for a user with `is_superuser` permissions, attempting to delete a "system" Job record via the UI or REST API is reported as an error.

- [X] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

_No response_

### External Dependencies

_No response_

HanlinMiao:
# Closes #6057 
# What's Changed

Enhanced job delete functions to prevent users with the appropriate permissions or super users from deleting system jobs from the UI and the API.
# Screenshots

After single instance delete:
Screenshot 2024-09-19 at 5 15 55 PM

After bulk delete:
Screenshot 2024-09-19 at 5 16 11 PM

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/changes/6057.changed
+++ b/changes/6057.changed
@@ -1,1 +1,1 @@
-Enhanced job delete functions to prevent users with the appropriate permissions or super users from deleting system jobs from the UI and the API.
+Enhanced job delete functions to prevent users from deleting system jobs from the UI and the API.
--- a/changes/6057.changed
+++ b/changes/6057.changed
@@ -1,0 +1,1 @@
+Enhanced job delete functions to prevent users with the appropriate permissions or super users from deleting system jobs from the UI and the API.
--- a/nautobot/extras/querysets.py
+++ b/nautobot/extras/querysets.py
@@ -1,6 +1,6 @@
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
-from django.db.models import F, Model, OuterRef, Q, Subquery
+from django.db.models import F, Model, OuterRef, ProtectedError, Q, Subquery
 from django.db.models.functions import JSONObject
 
 from nautobot.core.models.query_functions import EmptyGroupByJSONBAgg
@@ -223,6 +223,15 @@ class JobQuerySet(RestrictedQuerySet):
     """
     Extend the standard queryset with a get_for_class_path method.
     """
+
+    def delete(self):
+        for job in self:
+            if job.module_name.startswith("nautobot."):
+                raise ProtectedError(
+                    f"Unable to delete Job {job}. System Job cannot be deleted",
+                    [],
+                )
+        return super().delete()
 
     def get_for_class_path(self, class_path):
         try:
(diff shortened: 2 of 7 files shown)


nautobot/extras/querysets.py
```diff
@@ -224,6 +224,15 @@ class JobQuerySet(RestrictedQuerySet):
     Extend the standard queryset with a get_for_class_path method.
     """
 
+    def delete(self):
```
glennmatthews, line 227: Not entirely sure we want to prevent deletes at the queryset/ORM level - will this cause us problems in the future if we have a migration that renames or obsoletes an existing system job? Not sure.

If we are going to keep this, let's make sure we have some test coverage for this case.

nautobot/extras/tests/test_views.py
```diff
@@ -2366,6 +2366,78 @@ def setUpTestData(cls):
...
+        # Try bulk delete with delete job permission
+        self.add_permissions("extras.delete_job")
+        response = self.client.post(**request, follow=True)
+        self.assertHttpStatus(response, 403)
+        response_body = extract_page_body(response.content.decode(response.charset))
+        self.assertIn(f"Unable to delete Job {instance}. System Job cannot be deleted", response_body)
+        # assert Job still exists
+        self.assertTrue(self._get_queryset().filter(name=job_name).exists())
+
+        # Try bulk delete as a superuser
```
glennmatthews, line 2403:
```suggestion
        # Try delete with delete job permission
        self.add_permissions("extras.delete_job")
        response = self.client.post(**request, follow=True)
        self.assertHttpStatus(response, 403)
        response_body = extract_page_body(response.content.decode(response.charset))
        self.assertIn(f"Unable to delete Job {instance}. System Job cannot be deleted", response_body)
        # assert Job still exists
        self.assertTrue(self._get_queryset().filter(name=job_name).exists())

        # Try delete as a superuser
```

nautobot/extras/querysets.py
HanlinMiao: The reason why I overrode this is that we are using qs.delete() in our BulkDeleteView. https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/utils.py#L716

changes/6057.changed
```diff
@@ -0,0 +1 @@
+Enhanced job delete functions to prevent users with the appropriate permissions or super users from deleting system jobs from the UI and the API.
```
glennmatthews, line 1:
```suggestion
Enhanced job delete functions to prevent users from deleting system jobs from the UI and the API.
```

glennmatthews (approved):

PR #5668, opened 2024-05-02
Static groups UI

glennmatthews:
# Closes #5472
# What's Changed

Continuation of the work from #5631.

- Automatically add a `static_groups` filter to any subclass of BaseFilterSet that is static-group-associable
  - This ended up looking similar to the logic that we had in `get_filterset_parameter_form_field()` for tag, status, and role generic filters, but I liked my implementation better as it's on the filterset itself rather than a separate function. So I refactored how those three filters are set up to similarly move their logic to the filterset or mixin and out of the function.
- Add "Update Static Groups" button to object_list.html generic template
  - This opens a simple modal rather than sending the user to a separate view, similar to the Saved View modal in #5639.
  - Add custom form and custom view for bulk-assigning selected objects to static groups as backing for this modal.
- Add "Static Groups" tab to object_retrieve.html generic template (and device.html template since that overrides it)
  - Like the "Contacts" tab, this is implemented as part of the base view/template rather than a separate standalone view.
- Add `tenant` field to the StaticGroup model, and update factory, filterset, and tests accordingly
  - [x] TODO might need some model documentation update for this
- Add full UI for StaticGroups, including nav menu entry
- Add limited subset of UI for StaticGroupAssociations
- Add static groups count to Tenant detail view and re-sort its contents alphabetically.

# Screenshots

image
image
image
image
image
image
image
image
image

# TODO

- [x] Explanation of Change(s)
- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [x] Attached Screenshots, Payload Example
- [x] Unit, Integration Tests
- [x] Documentation Updates (when adding/changing features)
- n/a Example App Updates (when adding/changing features)
- [x] Outline Remaining Work, Constraints from Design

PR #6248, opened 2024-09-20
Fixed Device Type link and count from Device Family Detail View.

itdependsnetworks:
Fixed Device Type link and count from Device Family Detail View.

Seems to be regression during #5668, I checked a few, and they did not affect anything, but probably better suited for someone that knows the pattern where this may happen a bit better.

### DIFF
--- a/nautobot/dcim/views.py
+++ b/nautobot/dcim/views.py
@@ -4139,9 +4139,12 @@ class DeviceFamilyUIViewSet(NautobotUIViewSet):
             context["device_type_table"] = device_type_table
 
             total_devices = 0
+            device_type_count = 0
             for device_type in device_types:
                 total_devices += device_type.device_count
+                device_type_count += 1
             context["total_devices"] = total_devices
+            context["device_type_count"] = device_type_count
 
         return context
 
--- a/changes/6248.fixed
+++ b/changes/6248.fixed
@@ -1,0 +1,1 @@
+Fixed Device Type link and count from Device Family Detail View.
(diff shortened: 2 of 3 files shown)


nautobot/dcim/views.py
```diff
@@ -4139,9 +4139,12 @@ def get_extra_context(self, request, instance):
             context["device_type_table"] = device_type_table
 
             total_devices = 0
+            device_type_count = 0
             for device_type in device_types:
                 total_devices += device_type.device_count
+                device_type_count += 1
             context["total_devices"] = total_devices
+            context["device_type_count"] = device_type_count
```
glennmatthews, line 4147:
```suggestion
            for device_type in device_types:
                total_devices += device_type.device_count
            context["total_devices"] = total_devices
            context["device_type_count"] = device_types.count()
```

glennmatthews (commented):
More "interesting" UI patterns - not sure why we display the count at all given that we *also* have a full table of the related device-types just below. But thanks for the fix!

nautobot/dcim/views.py
itdependsnetworks: I thought about this, but it seemed like an extra query

itdependsnetworks:
@glennmatthews did my other point make sense to review that PR for similar?

nautobot/dcim/views.py
glennmatthews: Fair point - would be good to test either way. Could also use `len(device_types)` as that shouldn't be an extra query.

glennmatthews:
> @glennmatthews did my other point make sense to review that PR for similar?

Yes - I did a quick grep through the code base and don't immediately see any similar problems.

glennmatthews (approved):

PR #6270, opened 2024-09-24
Fixed device bay populate/depopulate redirect

### LINKED ISSUE #5924: When adding or removing child devices to device bays, clicking "Save" returns user to Device 'Overview' tab rather than to the Device Bays tab
### Environment
* Nautobot version (Docker tag too if applicable): 1.6.24, cff0fbbe8954 (v2.3.0a1)
* Python version:
* Database platform, version:
* Middleware(s): Tested using local 1.6.24 instance and the current running image on next.demo.nautobot.com

### Steps to Reproduce
1. If necessary, create a Device Type which can function as a parent
2. If necessary, create a Device Type which can function as a child
3. Create a parent Device and child Device
4. Create a Device Bay in the Parent Device
5. From the parent Device's Device Bay tab, click the '+' icon on the Bay to add the child device
6. Select the child Device and click 'Save'; you will be redirected to the Overview tab for the parent Device

### Expected Behavior
User should be returned to the Device's 'Device Bays' tab (e.g. `/dcim/devices/UUID/device-bays/`)

### Observed Behavior
User is returned to the Overview page for the Device.

#### return_url query string parameter
For what it's worth: when the '+' sign is clicked to start the assignment to the Device Bay, the `return_url` query string parameter is set to the correct URL; a sample URL at this point is:

`/dcim/device-bays/BAY_UUID/populate/?return_url=/dcim/devices/DEVICE_UUID/device-bays/`

...however when 'Save' is clicked, the POST request doesn't include the `return_url` parameter. Sample URL:

`/dcim/device-bays/BAY_UUID/populate/`

This differs from the behavior when interfaces are connected; when the 'Connect' button is clicked in that dialog, the POST request looks something like this:

`/dcim/interfaces/IFACE_UUID/connect/interface/?return_url=/dcim/devices/DEVICE_UUID/interfaces/`

#### Potential other impact
1. The same problem exists when removing child devices from devices bays. 
1. I haven't done an exhaustive test to see if other areas of the UI exhibit the same behavior, however I did confirm that the problem does /not/ exist for Module Bay operations in `next`.

joewesch:
# Closes #5924 
# What's Changed

Fixed the redirect URL for the Device Bay Populate/Depopulate view to take the user back to the Device Bays tab on the Device page.
# Screenshots

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/5924.fixed
+++ b/changes/5924.fixed
@@ -1,0 +1,1 @@
+Fixed the redirect URL for the Device Bay Populate/Depopulate view to take the user back to the Device Bays tab on the Device page.
--- a/nautobot/dcim/views.py
+++ b/nautobot/dcim/views.py
@@ -3148,7 +3148,7 @@ class DeviceBayPopulateView(generic.ObjectEditView):
                 f"Added {device_bay.installed_device} to {device_bay}.",
             )
 
-            return redirect("dcim:device", pk=device_bay.device.pk)
+            return redirect("dcim:device_devicebays", pk=device_bay.device.pk)
 
         return render(
             request,
@@ -3191,7 +3191,7 @@ class DeviceBayDepopulateView(generic.ObjectEditView):
                 f"Removed {removed_device} from {device_bay}.",
             )
 
-            return redirect("dcim:device", pk=device_bay.device.pk)
+            return redirect("dcim:device_devicebays", pk=device_bay.device.pk)
 
         return render(
             request,


glennmatthews (approved):
Looks like a reasonable change to me. Mildly curious why these specific views (appear to be the only such case in dcim.views) have a hardcoded redirect like this instead of using the more common `return_url` query parameter and `get_return_url()`, but that's not blocking.

timizuoebideri1 (approved):

Issue #6131, opened 2024-08-23
Custom_fields ignored with available-prefixes API in later versions of Nautotbot


kristofferDuctus:
### Environment
* Nautobot version (Docker tag too if applicable): Present in docker images 2.2.7-py3.11 up to nautobot:2.3.1-py3.12
* Python version: Tested both 3.11 and 3.12 in different docker images
* Database platform, version: Postgres 14
* Middleware(s):

### Steps to Reproduce
1. Create a custom field that is possible to use under the prefixes
2. Allocate a new prefix with the custom field parameter (example curl below):

curl --location 'http://localhost:8080/api/ipam/prefixes/ccb2034f-75fe-433f-8c42-a6a1f90f849a/available-prefixes/' \
--header 'accept: application/json' \
--header 'Authorization: Token 0123456789abcdef0123456789abcdef01234567' \
--header 'Content-Type: application/json' \
--data '{
        "prefix_length": 24,
        "status": "Active",
        "custom_fields": {"service_id": "svc123213"}
}'

3. The response nor the prefix contains the custom field parameters.

### Expected Behavior
The same curl command is working fine in docker container 2.2.6-py3.11 and the custom_fields set through the curl command is stored on the actual object.

### Observed Behavior
All custom parameters are empty and not set through the API.

glennmatthews:
Thanks for the report. Probably a regression due to #5856.

PR #6273, opened 2024-09-25
API endpoint to create new VLANs in VLANGroup.

mzbroch:
Complete contribution #6176 
API endpoint to create new VLANs in a VLANGroup

# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/docs/user-guide/core-data-model/ipam/vlangroup.md
+++ b/nautobot/docs/user-guide/core-data-model/ipam/vlangroup.md
@@ -3,3 +3,191 @@
 VLAN groups can be used to organize VLANs within Nautobot. Each group may optionally be assigned to a specific location, but a group cannot belong to multiple locations.
 
 Groups can also be used to enforce uniqueness: Each VLAN within a group must have a unique ID and name. VLANs which are not assigned to a group may have overlapping names and IDs (including VLANs which belong to a common location). For example, you can create two VLANs with ID 123, but they cannot both be assigned to the same group.
+
+## VLAN Group Ranges
+
++++ 2.4.0
+
+VLAN Groups contain a mandatory `range` field with a default value of `1-4094` (permitting all VLANs). This field may be used to constrain the valid member VLANs of the group. VLANs can only be associated with a given VLAN Group if their VLAN identifiers ("VIDs") fall within the specified `range`.
+Range value accepts commas and dashes, with examples as follows:
+
+* 1
+* 1-5
+* 1-5,10
+* 1-5,10-15
+* 1-5,10-15,16,17,18,19,20
+
+Values between dashes will also be expanded into a list of VLANs.
+
+## Creating new VLANs in a VLANGroup programatically
+
++++ 2.4.0
+
+VLAN Groups offer an API endpoint to list available VIDs and create new VLANs programmatically:
+
+`/api/ipam/vlan-groups/<vlan_group_id>/available-vlans/`
+
+A `GET` request to this API endpoint will return a list of all available (unassigned) VIDs available in the given `VLANGroup`.
+
+A `POST` request to the same API endpoint will create `VLAN` record(s) in the given `VLANGroup`. Note that you must specify the `name` and `status` of each VLAN to be created as a part of this request, but the `vlan_group` and `vid` may be omitted as they can be automatically assigned by this API. See below.
+
+### Creating VLANGroup example
+
+To create a new `VLANGroup` with a `range` field of: `5-10` (permit VLANs `5, 6, 7, 8, 9, 10`), issue following POST request:
+
+```no-highlight
+curl -X POST -H "Authorization: Token $TOKEN" \
+-H "Content-Type: application/json" \
+-H "Accept: application/json; version=2.0; indent=4" \
+http://nautobot/api/ipam/vlan-groups/ \
+--data '[{"name": "Enterprise VLANs", "range": "5-10"}]'
+```
+
+`VLANGroup` was created as follows:
+
+```json
+[
+    {
+        "id": "c16d5d1d-40d6-4902-b89d-6294e60cb4bf",
+        "object_type": "ipam.vlangroup",
+        "display": "Enterprise VLANs",
+        "url": "http://nautobot/api/ipam/vlan-groups/c16d5d1d-40d6-4902-b89d-6294e60cb4bf/",
+        "natural_slug": "enterprise-vlans_c16d",
+        "name": "Enterprise VLANs",
+        "description": "",
+        "range": "5-10",
+        "location": null,
+        "created": "2024-09-25T09:27:16.343655Z",
+        "last_updated": "2024-09-25T09:27:16.343670Z",
+        "notes_url": "http://nautobot/api/ipam/vlan-groups/c16d5d1d-40d6-4902-b89d-6294e60cb4bf/notes/",
+        "custom_fields": {},
+        "computed_fields": {},
+        "relationships": {}
+    }
+]
+```
+
+### Listing available VLANs example
+
+Then, to list all available VLAN IDs within the just-created `VLANGroup`, issue a GET request:
+
+```no-highlight
+curl -X GET -H "Authorization: Token $TOKEN" \
+-H "Content-Type: application/json" \
+-H "Accept: application/json; version=2.0; indent=4" \
+http://nautobot/api/ipam/vlan-groups/c16d5d1d-40d6-4902-b89d-6294e60cb4bf/available-vlans/
+```
+
+```json
+{
+    "count": 6,
+    "next": null,
+    "previous": null,
+    "results": [
+        5,
+        6,
+        7,
+        8,
+        9,
+        10
+    ]
+}
+```
+
+### Creating new VLANs example
+
+The `/api/ipam/vlan-groups/<vlan_group_id>/available-vlans/` API endpoint accepts `POST` requests to create a new VLAN record, similar to the `/api/ipam/vlans/` API endpoint. However, there are some characteristic differences in the behavior of `/available-vlans/` with regard to its input and output:
+
+* Accepts a single dictionary as input to create 1 VLAN, in which case it returns a single dictionary describing the created VLAN
+* Accepts a list of dictionaries as input to create one or more VLANs, in which case it returns a list of VLAN objects
+* The fields `name` and `status` of are mandatory inputs for each VLAN to be created
+* The field `vid` is optional and might be specified to override auto-allocation for a given VLAN
+    * Will not accept a request with the same `vid` value specified multiple times
+* Will not accept a `vlan_group` attribute different than that already specified in the request's URL
+* Other `VLAN` model fields are permitted and optional
+* As with the `/vlans/` endpoint, specifying `locations` in the POST request is not presently supported; you can make a subsequent call(s) to the `/api/ipam/vlan-location-assignments/` endpoint to update the many-to-many association between the created VLAN(s) and desired Location(s)
+
+To create two new VLANs, a data payload has to be specified. In the case of the first VLAN, its `VLAN ID` will be automatically determined based on `VLANGroups` availability, in the case of the second VLAN we explicitly request `vid` 8:
+
+```no-highlight
+curl -X POST -H "Authorization: Token $TOKEN" \
+-H "Content-Type: application/json" \
+-H "Accept: application/json; version=2.0; indent=4" \
+http://nautobot/api/ipam/vlan-groups/c16d5d1d-40d6-4902-b89d-6294e60cb4bf/available-vlans/ \
+--data '[
+{"name": "First VLAN", "status": "1e560c4d-07ef-4d49-af6e-c85aa1f295d3"},
+{"name": "Second VLAN", "vid": 8, "status": "1e560c4d-07ef-4d49-af6e-c85aa1f295d3"}
+]'
+```
+
+`VLANs` created:
+
+```json
+{
+    "count": 2,
+    "next": null,
+    "previous": null,
+    "results": [
+        {
+            "id": "58bd2a10-33f7-45e2-93d2-f56be25dfe8c",
+            "object_type": "ipam.vlan",
+            "display": "First VLAN (5)",
+            "url": "http://nautobot/api/ipam/vlans/58bd2a10-33f7-45e2-93d2-f56be25dfe8c/",
+            "natural_slug": "58bd2a10-33f7-45e2-93d2-f56be25dfe8c_58bd",
+            "vid": 5,
+            "name": "First VLAN",
+            "description": "",
+            "vlan_group": {
+                "id": "c16d5d1d-40d6-4902-b89d-6294e60cb4bf",
+                "object_type": "ipam.vlangroup",
+                "url": "http://nautobot/api/ipam/vlan-groups/c16d5d1d-40d6-4902-b89d-6294e60cb4bf/"
+            },
+            "status": {
+                "id": "1e560c4d-07ef-4d49-af6e-c85aa1f295d3",
+                "object_type": "extras.status",
+                "url": "http://nautobot/api/extras/statuses/1e560c4d-07ef-4d49-af6e-c85aa1f295d3/"
+            },
+            "role": null,
+            "tenant": null,
+            "locations": [],
+            "created": "2024-09-25T09:36:07.587950Z",
+            "last_updated": "2024-09-25T09:36:07.587963Z",
+            "tags": [],
+            "notes_url": "http://nautobot/api/ipam/vlans/58bd2a10-33f7-45e2-93d2-f56be25dfe8c/notes/",
+            "custom_fields": {},
+            "computed_fields": {},
+            "relationships": {}
+        },
+        {
+            "id": "9b7c6b03-9140-4b00-940f-5dbcde173de1",
+            "object_type": "ipam.vlan",
+            "display": "Second VLAN (8)",
+            "url": "http://nautobot/api/ipam/vlans/9b7c6b03-9140-4b00-940f-5dbcde173de1/",
+            "natural_slug": "9b7c6b03-9140-4b00-940f-5dbcde173de1_9b7c",
+            "vid": 8,
+            "name": "Second VLAN",
+            "description": "",
+            "vlan_group": {
+                "id": "c16d5d1d-40d6-4902-b89d-6294e60cb4bf",
+                "object_type": "ipam.vlangroup",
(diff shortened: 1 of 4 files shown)


nautobot/docs/user-guide/core-data-model/ipam/vlangroup.md
```diff
@@ -3,3 +3,177 @@
 VLAN groups can be used to organize VLANs within Nautobot. Each group may optionally be assigned to a specific location, but a group cannot belong to multiple locations.
 
 Groups can also be used to enforce uniqueness: Each VLAN within a group must have a unique ID and name. VLANs which are not assigned to a group may have overlapping names and IDs (including VLANs which belong to a common location). For example, you can create two VLANs with ID 123, but they cannot both be assigned to the same group.
+
+## VLAN Group Ranges
+
++/- 2.4.0
+
+VLAN Groups contain a mandatory `range` field with a default value of `1-4094` which equals to all VLANs. This field may be used to constrain member VLANs of the group. Only VLANs with VLAN identifiers ("vids") within a `range` value will be allowed to associated with a `VLANGroup`
```
glennmatthews, line 11:
```suggestion
VLAN Groups contain a mandatory `range` field with a default value of `1-4094` (permitting all VLANs). This field may be used to constrain the valid member VLANs of the group. VLANs can only be associated with a given VLAN Group if their VLAN identifiers ("vids") fall within the specified `range`.
```

nautobot/docs/user-guide/core-data-model/ipam/vlangroup.md
```diff
@@ -3,3 +3,177 @@
...
+* 1-5,10-15
+* 1-5,10-15,16,17,18,19,20
+
+Values between dashes will also be expanded into a list of VLANs.
+
+## Creating new VLANs in a VLANGroup programatically
+
++/- 2.4.0
+
+VLAN Groups offer new API endpoint to list available and create new VLANs programatically. Nautobot accepts API calls to the following endpoint:
```
glennmatthews, line 27:
```suggestion
VLAN Groups offer an API endpoint to list available VIDs and create new VLANs programmatically:
```

nautobot/docs/user-guide/core-data-model/ipam/vlangroup.md
```diff
@@ -3,3 +3,177 @@
...
+
+`/api/ipam/vlan-groups/<vlan_group_id>/available-vlans/`
+
+For the `GET` request, API will return a list of all available VLAN IDs available in the given `VLANGroup`.
```
glennmatthews, line 31:
```suggestion
A `GET` request to this API endpoint will return a list of all available (unassigned) VLAN IDs available in the given `VLANGroup`.
```

nautobot/docs/user-guide/core-data-model/ipam/vlangroup.md
```diff
@@ -3,3 +3,177 @@
...
+
+For the `POST` request, API will create requested number of VLANs in a `VLANGroup`.
```
glennmatthews, line 33:
```suggestion
A `POST` request to the same API endpoint will create `VLAN` record(s) in the given `VLANGroup`. Note that you must specify the `name` and `status` of each VLAN to be created as a part of this request, but the `vlan_group` and `vid` may be omitted as they can be automatically assigned by this API. See below.
```

nautobot/docs/user-guide/core-data-model/ipam/vlangroup.md
```diff
@@ -3,3 +3,177 @@
...
+        "last_updated": "2024-09-25T09:27:16.343670Z",
+        "notes_url": "http://nautobot/api/ipam/vlan-groups/c16d5d1d-40d6-4902-b89d-6294e60cb4bf/notes/",
+        "custom_fields": {},
+        "computed_fields": {},
+        "relationships": {}
+    }
+]
+```
+
+Then, to list all available `VLANs` within an `Enterprise VLANs` `VLANGroup` issue a GET request:
```
glennmatthews, line 71:
```suggestion
Then, to list all available VLAN IDs within the just-created`VLANGroup`, issue a GET request:
```

nautobot/docs/user-guide/core-data-model/ipam/vlangroup.md
```diff
@@ -3,3 +3,177 @@
...
+    "results": [
+        5,
+        6,
+        7,
+        8,
+        9,
+        10
+    ]
+
+```
```
glennmatthews, line 94:
````suggestion
    ]
}
```
````

File: nautobot/docs/user-guide/core-data-model/ipam/vlangroup.md [hunk 7]
```diff
@@ -3,3 +3,177 @@
...
+
+To create two new VLANs, a data payload has to be specified. In the case of first VLAN, its `VLAN ID` will be automatically determined based on `VLANGroups` availability, in the case of second VLAN we manually define `vid` requsting `VLAN` with `id` 8:
```
glennmatthews, line 96: Can we elaborate more on the data payload, i.e. mandatory `name` and `status`, optional `vid`, `tenant`, `role`, etc.?

nautobot/ipam/api/serializers.py
```diff
@@ -481,6 +481,25 @@ def validate(self, data):
...
+
+    class Meta(VLANSerializer.Meta):
+        model = VLAN
+        fields = (
+            # not vid, vlan_group as those are implied by the selected VLANGroup
+            "name",
+            "status",
+            "role",
+            "tenant",
+            "locations",
```
glennmatthews, line 497: Do we actually support setting the `locations` M2M via this API? Normally that'd be done as a separate step via the `vlan-location-assignments` endpoint. If not supported here let's remove it.

nautobot/ipam/api/views.py
```diff
@@ -405,36 +411,86 @@ def available_vlans(self, request, pk=None):
...
-        try:
-            limit = int(request.query_params.get("limit", get_settings_or_config("PAGINATE_COUNT")))
-        except ValueError:
-            limit = get_settings_or_config("PAGINATE_COUNT")
+        if request.method == "POST":
+            with cache.lock(
+                "nautobot.ipam.api.views.available_vlans", blocking_timeout=5, timeout=settings.REDIS_LOCK_TIMEOUT
+            ):
+                # Normalize to a list of objects
+                requested_vlans = request.data if isinstance(request.data, list) else [request.data]
```
glennmatthews, line 421: Would be good to document the single-allocation alternative API input implied here (dict rather than list of dicts) with an example.

nautobot/ipam/api/views.py
```diff
@@ -405,36 +411,86 @@ def available_vlans(self, request, pk=None):
...
+                        },
+                        status=status.HTTP_204_NO_CONTENT,
+                    )
+
+                # Assign VLAN IDs from the list of VLANGroup's available VLAN IDs.
+                _available_vids = iter(available_vids)
+
+                for requested_vlan in requested_vlans:
+                    if "vid" not in requested_vlan:
+                        requested_vlan["vid"] = f"{next(_available_vids)}"
```
glennmatthews, line 441: Perhaps?

```suggestion
                        requested_vlan["vid"] = next(_available_vids)
```

nautobot/ipam/api/views.py
```diff
@@ -405,36 +411,86 @@ def available_vlans(self, request, pk=None):
...
+
+                    requested_vlan["vlan_group"] = vlan_group.pk
```
glennmatthews, line 443: May need special case handling for custom field data here (see #6131), similar to what we do for the available-prefixes and available-ips endpoints:

```py
                    # The serializer usage above has mapped "custom_fields" dict to "_custom_field_data".
                    # We need to convert it back to "custom_fields" as we're going to deserialize it a second time below
                    requested_vlan["custom_fields"] = requested_vlan.pop("_custom_field_data", {})
```

Or is this not necessary because we're not passing `request.data` through a first-round deserialization above (see my other comment)?

nautobot/ipam/api/views.py
glennmatthews, line 421: Is there a reason we're not following the available-prefixes/available-ips pattern of passing `request.data` into a `VLANAllocationSerializer` and doing an initial `serializer.is_valid(raise_exception=True)` before proceeding to the below data manipulation?

nautobot/ipam/api/views.py
```diff
@@ -405,36 +411,86 @@ def available_vlans(self, request, pk=None):
...
+
+                # Initialize the serializer with a list or a single object depending on what was requested
+                context = {"request": request, "depth": 0}
+                serializer = serializers.VLANSerializer(data=requested_vlans, many=True, context=context)
```
glennmatthews, line 447: In `available-prefixes`/`available-ips` we only use `many=True` for the response if the user requested more than one allocation. Is that desirable here as well, or is keeping the API response more predictably structured (always a list/many response) to be preferred, do you think?

nautobot/ipam/tests/test_api.py
```diff
@@ -1006,20 +1015,109 @@ def test_list_available_vlans(self):
...
+        cf = CustomField.objects.create(key="sor", label="Source of Record Field", type="text")
+        cf.content_types.add(ContentType.objects.get_for_model(VLAN))
+        url = reverse("ipam-api:vlangroup-available-vlans", kwargs={"pk": self.vlan_group.pk})
+        self.add_permissions(
+            "ipam.view_vlangroup",
+            "ipam.change_vlangroup",
+            "ipam.add_vlangroup",
+            "ipam.change_vlan",
+            "ipam.view_vlan",
+            "ipam.add_vlan",
```
glennmatthews, line 1040: I'm surprised all of these are necessary - I would have thought maybe just the `view` permissions plus `add_vlan` would be sufficient?

nautobot/ipam/tests/test_api.py
```diff
@@ -1006,20 +1015,109 @@ def test_list_available_vlans(self):
...
+        # Try to create one more VLAN
+        response = self.client.post(url, {"name": "UTILIZED_VLAN_GROUP"}, format="json", **self.header)
+        self.assertHttpStatus(response, status.HTTP_204_NO_CONTENT)
+        self.assertIn("detail", response.data)
+        self.assertIn(
+            f"An insufficient number of VLANs are available within the VLANGroup {self.vlan_group}",
+            response.data["detail"],
+        )
+
+    def test_create_multiple_available_vlans(self):
```
glennmatthews, line 1070: Might be good to add a couple more tests:

- POST to available-vlans and explicitly specifying the vid (as in the docs)
  - POST to available-vlans with multiple entries trying to explicitly specify the same vid (should be rejected)
- POST to available-vlans and explicitly specifying the vlan_group (should probably be rejected?)

nautobot/docs/user-guide/core-data-model/ipam/vlangroup.md
HanlinMiao: Should we keep the language consistent throughout this doc? `VLAN IDs` vs. `VIDs`

nautobot/docs/user-guide/core-data-model/ipam/vlangroup.md
```diff
@@ -3,3 +3,177 @@
...
+### Examples
+
+To create a new `VLANGroup` with a `range` field of: `5-10` (permit VLANs `5, 6, 7, 8, 9, 10`), issue following POST request:
+
+```no-highlight
+curl -X POST -H "Authorization: Token $TOKEN" \
+-H "Content-Type: application/json" \
+-H "Accept: application/json; version=2.0; indent=4" \
+http://nautobot/api/ipam/vlan-groups/ \
+--data '[{"name": "Enterprise VLANs", "range": "5-10"}]'
```
HanlinMiao, line 44: Are we missing VLAN status in this example request?

nautobot/docs/user-guide/core-data-model/ipam/vlangroup.md
HanlinMiao: Just a missing space in Glenn's review comment
```suggestion
Then, to list all available VLAN IDs within the just-created `VLANGroup`, issue a GET request:
```

nautobot/ipam/api/views.py
mzbroch: 🙄

nautobot/ipam/api/views.py
mzbroch: It is not needed as we return `serializer.data` and not `serializer_validated_data`.
We do not run the first deserialization as in case of `VLAN` it is not possible to fully validate it without a VID.

nautobot/ipam/api/views.py
mzbroch: It is not possible to validate `VLAN` instance in the same way because of missing `vid` attribute.

I think, functionally our used second serialization does the same?

nautobot/ipam/api/views.py
mzbroch: Today endpoint accepts a single dict or list of dicts, but in each cases responds with a list.
I think this is more predictable, especially in context of count / next / previous / results keys we added to the Response

nautobot/docs/user-guide/core-data-model/ipam/vlangroup.md
mzbroch: This is a request to the `VLANGroup` and that model does not have a status

nautobot/ipam/api/views.py
mzbroch: OK, now I think it makes it more consistent to respond in the request format and similar to existing endpoints. 👍

nautobot/ipam/tests/test_api.py
mzbroch: Good point - Actually , the only one needed is `add_vlangroup` (tested).

glennmatthews: That sounds like a bug in waiting to me - it should be `add_vlan` even though this is a `vlangroup` endpoint.

nautobot/ipam/api/views.py
glennmatthews: I think you're right. Wonder if we should change the approach used in the other two endpoints to follow this pattern?

HanlinMiao:
Hmmmm I dont know why markdownlint is failing, it is not for me locally

nautobot/ipam/tests/test_api.py
mzbroch: Done 👍

nautobot/ipam/api/views.py
mzbroch: I revisited this and fixed 👌

nautobot/ipam/api/serializers.py
mzbroch: Removed

nautobot/ipam/tests/test_api.py
mzbroch: Fixed:
get - view_vlangroup
post - view_vlangroup, add_vlan 
permissions are required

nautobot/docs/user-guide/core-data-model/ipam/vlangroup.md
```diff
@@ -3,3 +3,196 @@
...
+    * accepts a list of dictionaries to create multiple VLANs (single dictionary is also permitted), returns a list of VLAN objects
+    * accepts a single dictionary to create 1 VLAN, returns a single dictionary describing a VLAN
+    * fields `name` and `status` of `VLAN` object are mandatory
+    * field `vid` is optional and might be specified to override auto-allocation for given VLAN
+    * field `locations` is not supported
+    * other `VLAN` model fields are permitted and optional
+    * will not accept an optional `vlan_group` attribute different than specified in request's URL
+    * will not accept a request with the same VLAN IDs specified multiple times
+
+To create two new VLANs, a data payload has to be specified. In the case of the first VLAN, its `VLAN ID` will be automatically determined based on `VLANGroups` availability, in the case of second VLAN we manually define `vid` requsting `VLAN` with `id` 8:
```
glennmatthews, line 115:
```suggestion
The `/api/ipam/vlan-groups/<vlan_group_id>/available-vlans/` API endpoint accepts `POST` requests to create a new VLAN record, similar to the `/api/ipam/vlans/` API endpoint. However, there are some characteristic differences in the behavior of `/available-vlans/` with regard to its input and output:

* Accepts a single dictionary as input to create 1 VLAN, in which case it returns a single dictionary describing the created VLAN
* Accepts a list of dictionaries as input to create one or more VLANs, in which case it returns a list of VLAN objects
* The fields `name` and `status` of are mandatory inputs for each VLAN to be created
* The field `vid` is optional and might be specified to override auto-allocation for a given VLAN
    * Will not accept a request with the same `vid` value specified multiple times
* Will not accept a `vlan_group` attribute different than that already specified in the request's URL
* Other `VLAN` model fields are permitted and optional
* As with the `/vlans/` endpoint, specifying `locations` in the POST request is not presently supported; you can make a subsequent call(s) to the `/api/ipam/vlan-location-assignments/` endpoint to update the many-to-many association between the created VLAN(s) and desired Location(s)

To create two new VLANs, a data payload has to be specified. In the case of the first VLAN, its `VLAN ID` will be automatically determined based on `VLANGroups` availability, in the case of the second VLAN we explicitly request `vid` 8:
```

nautobot/ipam/api/views.py
```diff
@@ -391,11 +392,35 @@ class VLANGroupViewSet(NautobotModelViewSet):
...
+    def restrict_queryset(self, request, *args, **kwargs):
+        """
+        Apply "view" permissions on the POST /available-vlans/ endpoint, otherwise as ModelViewSetMixin.
+        """
+        if request.user.is_authenticated and self.action == "available_vlans":
+            self.queryset = self.queryset.restrict(request.user, "view")
+        else:
+            super().restrict_queryset(request, *args, **kwargs)
+
+    class CreateVlanPermissions(TokenPermissions):
```
glennmatthews, line 404:
```suggestion
    class AvailableVLANPermissions(TokenPermissions):
```

nautobot/ipam/api/views.py
```diff
@@ -391,11 +392,35 @@ class VLANGroupViewSet(NautobotModelViewSet):
...
+        responses={201: serializers.VLANSerializer(many=True)},
+        request=serializers.VLANAllocationSerializer(many=True),
+    )
     @action(
         detail=True,
         name="Available VLAN IDs",
         url_path="available-vlans",
-        methods=["get"],
+        methods=["get", "post"],
+        permission_classes=[CreateVlanPermissions],
```
glennmatthews, line 423:
```suggestion
        permission_classes=[AvailableVLANPermissions],
```

glennmatthews (approved):

PR #6277, opened 2024-09-25
Added `default_job_queue` field to the Job Model

### LINKED ISSUE #6259: Specify Job Queue default for a Job
As a Network Automation Engineer, I want to specify a default Job Queue for a Job, when multiple Job Queues are available for a Job, so that end users are presented a default selection in the Job Run form.

I will know this is complete when:
- Consistent with the "override" checkbox functionality in the model/UI
- A new field is added to Job model to specify a default Job Queue
- The data migration populates this from the first entry in the exist task_queue list
- When a Job is imported/updated, the default is populated from the first in the list
- In the case of no Job Queues defined on the Job, we should look for the presence of the "default" record and use it a the default queue. If the "default" record does not exist, the Job Run form will have no default option selected.

HanlinMiao:
# Closes #6259 
# What's Changed

- [x] Added `default_job_queue` field to the Job Model
- [x] Added `default_job_queue` and `default_job_queue_override` JobEditForm fields
- [x] Added automatic population of job_queue with default job queue in Job Run form
- [x] Modified data migrations to populate `default_job_queue` field with the first element of the `task_queues` or `default`
- [x] When a Job is imported/updated, the default is populated from the first in the list
- [x] Doc Changes

# Screenshots

Screenshot 2024-09-25 at 3 53 35 PM
Screenshot 2024-09-25 at 3 47 25 PM
Screenshot 2024-09-25 at 3 26 15 PM
Screenshot 2024-09-25 at 3 47 02 PM

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/forms/forms.py
```diff
@@ -982,9 +986,34 @@ class Meta:
...
+            except JobQueue.DoesNotExist:
+                pass
+        self.fields["default_job_queue"] = DynamicModelChoiceField(
+            queryset=job_queue_queryset,
+            query_params=job_queue_params,
+            required=False,
+            help_text="The default job queue to route this job to",
+            label="Default Job queue",
+        )
+        self.fields["default_job_queue"].initial = default_job_queue
```
glennmatthews, line 1015: This feels awry to me - since you can edit the `job_queues` and the `default_job_queue` at the same time, it doesn't make sense for the `default_job_queue` to constrain the queryset or query_params based on the preexisting job-queue assignments?

Hmm... this is similar actually to the recent discussion around setting a `VDC.primary_ip4` while also changing the `VDC.interfaces`. What if we changed things around so that the selected `default_job_queue` gets implicitly/automatically included into `job_queues` even if it wasn't explicitly specified as such by the user?

nautobot/extras/api/views.py
```diff
@@ -599,11 +599,19 @@ def run(self, request, *args, **kwargs):
...
         # Get a default queue from either the job model's specified task queue or
         # the system default to fall back on if request doesn't provide one
-        default_valid_queue = valid_queues[0]
+        if job_model.default_job_queue:
+            default_valid_queue = job_model.default_job_queue.name
+        else:
+            default_valid_queue, _ = JobQueue.objects.get_or_create(
+                name=settings.CELERY_TASK_DEFAULT_QUEUE, defaults={"queue_type": JobQueueTypeChoices.TYPE_CELERY}
+            )
+            default_valid_queue = default_valid_queue.name
```
glennmatthews, line 614: I don't think we should be creating a new JobQueue as a side effect of calling the API. If a job has no default queue, and CELERY_TASK_DEFAULT_QUEUE doesn't exist, and the user didn't specify a queue, it feels to me like we should reject the request.

nautobot/extras/forms/forms.py
```diff
@@ -982,9 +986,31 @@ class Meta:
...
+                default_job_queue = default_job_queue.pk
+            except JobQueue.DoesNotExist:
+                pass
+        self.fields["default_job_queue"] = DynamicModelChoiceField(
+            queryset=JobQueue.objects.all(),
+            required=False,
+            help_text="The default job queue to route this job to",
+            label="Default Job queue",
+        )
+        self.fields["default_job_queue"].initial = default_job_queue
```
glennmatthews, line 1012: This feels over-complex to me. Can we maybe instead design things so that `Job.default_job_queue` is non-nullable (with a on_delete=PROTECT) so that we always have one available, instead of needing this kind of fallback logic?

nautobot/extras/jobs.py
```diff
@@ -498,6 +498,20 @@ def as_form(cls, data=None, files=None, initial=None, approval_view=False):
...
+        if job_model.default_job_queue:
+            default_job_queue = job_model.default_job_queue.pk
+        else:
+            # Populate the field with settings.CELERY_TASK_DEFAULT_QUEUE JobQueue if it exists
+            try:
+                default_job_queue = JobQueue.objects.get(name=settings.CELERY_TASK_DEFAULT_QUEUE)
+                default_job_queue = default_job_queue.pk
+            except JobQueue.DoesNotExist:
+                pass
+        form.fields["_job_queue"].initial = default_job_queue
```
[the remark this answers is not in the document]
glennmatthews: Here too this feels like it'd be a lot simpler if `job.default_job_queue` is non-nullable.

nautobot/extras/models/jobs.py
```diff
@@ -359,6 +369,11 @@ def clean(self):
     def save(self, *args, **kwargs):
         """When a Job is uninstalled, auto-disable all associated JobButtons, JobHooks, and ScheduledJobs."""
         super().save(*args, **kwargs)
+        # Add specified default_job_queue to job.job_queues if it is not included.
+        if self.default_job_queue:
+            if not self.job_queues.filter(pk=self.default_job_queue.pk).exists():
+                self.job_queues.add(self.default_job_queue)
+
```
glennmatthews, line 376: Not sure this works as intended given the usual order of updates (`job.save()` then `job.job_queues.set(...)`) as the actual M2M update might blow this addition away. Maybe something to be done via an m2m signal handler instead?

nautobot/extras/tests/test_api.py
```diff
@@ -1755,6 +1755,36 @@ def test_run_job_with_both_task_queue_and_job_queue_specified(self, mock_get_wor
             "task_queue and job_queue are both specified. Please specifiy only one or another.", str(response.content)
         )
 
+    @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
+    @mock.patch("nautobot.extras.api.views.get_worker_count")
+    def test_run_job_with_no_queue_specified(self, mock_get_worker_count):
+        """Test job run response contains nested job result."""
```
glennmatthews, line 1761: Wrong docstring?

nautobot/extras/tests/test_api.py
```diff
@@ -1755,6 +1755,36 @@ def test_run_job_with_both_task_queue_and_job_queue_specified(self, mock_get_wor
...
+        url = self.get_run_url()
+        response = self.client.post(
+            url,
+            {
+                "data": job_data,
+            },
+            format="json",
+            **self.header,
+        )
+        # TODO Hmmm how do I determine which queue did the job use to run?
```
glennmatthews, line 1785: I think it's in `JobResult.celery_kwargs`

nautobot/extras/tests/test_models.py
```diff
@@ -1234,6 +1235,17 @@ def test_clean_input_validation(self):
...
         )
 
+    def test_default_job_queue_always_included_in_job_queues(self):
+        default_job_queue = JobQueue.objects.first()
+        job_queues = list(JobQueue.objects.exclude(pk=default_job_queue.pk))[:3]
+
+        job = JobModel.objects.first()
+        job.default_job_queue = default_job_queue
+        job.job_queues.set(job_queues)
+        job.save()
```
glennmatthews, line 1245: Yeah normally the order of operations would be the reverse here.

nautobot/extras/jobs.py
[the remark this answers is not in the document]
HanlinMiao: Should I set the `settings.CELERY_TASK_DEFAULT_QUEUE` as its `default_job_queue` if the job's meta `task_queues` is empty?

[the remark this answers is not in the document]
glennmatthews: I think so? That seems like our expected behavior in 2.3.

nautobot/extras/models/jobs.py
HanlinMiao: It seems like it is working as intended. See [test_default_job_queue_always_included_in_job_queues](https://github.com/nautobot/nautobot/blob/u/hanlin-%236259/nautobot/extras/tests/test_models.py#L1238)

HanlinMiao: Ohhhh I see what you mean

nautobot/extras/api/views.py
```diff
@@ -599,11 +599,10 @@ def run(self, request, *args, **kwargs):
                 request.method, detail="This job's source code could not be located and cannot be run"
             )
 
-        job_queues = job_model.job_queues.all().values_list("name", flat=True)
+        job_queues = list(job_model.job_queues.all().values_list("name", flat=True))
         valid_queues = job_queues if job_queues else [settings.CELERY_TASK_DEFAULT_QUEUE]
```
glennmatthews, line 603: Could this just be `valid_queues = job_model.task_queues`?

Actually, is `valid_queues` even needed/used at all now?

nautobot/extras/forms/forms.py
```diff
@@ -982,9 +986,22 @@ class Meta:
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        job_model = self.instance
+        self.fields["default_job_queue"] = DynamicModelChoiceField(
+            queryset=JobQueue.objects.all(),
+            required=False,
+            help_text="The default job queue to route this job to",
+            label="Default Job queue",
+        )
+        self.fields["default_job_queue"].initial = job_model.default_job_queue.pk
```
glennmatthews, line 1003: Still needed here? I'd really hope that the declaration-time field should "just work" at this point.

nautobot/extras/migrations/0116_create_job_queue_model.py
```diff
@@ -94,6 +94,18 @@ class Migration(migrations.Migration):
...
+            name="default_job_queue_override",
+            field=models.BooleanField(default=False),
+        ),
+        migrations.AddField(
+            model_name="job",
+            name="default_job_queue",
+            field=models.ForeignKey(
+                blank=False, null=False, on_delete=django.db.models.deletion.PROTECT, to="extras.jobqueue"
+            ),
+        ),
```
glennmatthews, line 108: Does this actually work? I'd have thought we'd need to create the field as `blank=True, null=True`, populate it in the data migration, and only then change it to `blank=False, null=False`?

nautobot/extras/migrations/0117_task_queue_to_job_queue_migration.py
```diff
@@ -16,16 +16,20 @@ def migrate_task_queues_to_job_queues(apps, schema):
...
-
+        default_job_queue, _ = JobQueue.objects.get_or_create(
+            name=task_queues[0], defaults={"queue_type": JobQueueTypeChoices.TYPE_CELERY}
+        )
+        job.default_job_queue = default_job_queue
+        job.save()
         # Go through each task_queue
         # make or get the corresponding job_queue object and assign it to the job
+        job_queues = [job.default_job_queue]
         for task_queue in task_queues:
```
glennmatthews, line 27: Possibly `task_queues[1:]` to be slightly more efficient?

nautobot/extras/templates/extras/job_detail.html
```diff
@@ -284,6 +284,32 @@
...
                 </td>
             </tr>
+            <tr>
+                <td>Default Job Queue</td>
+                <td>
+                    {% if object.default_job_queue %}
+                        {{ object.default_job_queue|hyperlinked_object }}
+                    {% else %}
+                        {{ None|placeholder }}
+                    {% endif %}
```
glennmatthews, line 294: Pretty sure this can just be reduced to `{{ object.default_job_queue|hyperlinked_object }}` without the if/else.

nautobot/extras/templates/extras/job_detail.html
```diff
@@ -284,6 +284,32 @@
...
+                </td>
+                <td>
+                    {% if object.default_job_queue_override %}
+                        <span class="text-muted">
+                            overridden; default is:&nbsp;
```
glennmatthews, line 299: The non-breaking space (`&nbsp;`) here isn't doing much since it's immediately followed by other whitespace.

nautobot/extras/tests/test_api.py
```diff
@@ -1755,6 +1755,38 @@ def test_run_job_with_both_task_queue_and_job_queue_specified(self, mock_get_wor
...
+            url,
+            {
+                "data": job_data,
+            },
+            format="json",
+            **self.header,
+        )
+        self.assertHttpStatus(response, self.run_success_response_status)
+        latest_job_result = JobResult.objects.latest()
+        self.assertIn(job_model.default_job_queue.name, latest_job_result.celery_kwargs)
```
glennmatthews, line 1788:
```suggestion
        self.assertEqual(job_model.default_job_queue.name, latest_job_result.celery_kwargs["queue"])
```

glennmatthews: Also, might be good to test this with a job whose default queue *isn't* `CELERY_TASK_DEFAULT_QUEUE`?

nautobot/extras/tests/test_views.py
```diff
@@ -2752,6 +2752,28 @@ def test_run_job_with_invalid_task_queue(self, _):
...
+        self.test_pass.save()
+        data = {
+            "_schedule_type": "immediately",
+        }
+
+        for run_url in self.run_urls:
+            response = self.client.post(run_url, data)
+            result = JobResult.objects.latest()
+            self.assertIsNotNone(result, msg=run_url)
+            self.assertIn(self.test_pass.default_job_queue.name, result.celery_kwargs)
```
glennmatthews, line 2774:
```suggestion
            self.assertEqual(self.test_pass.default_job_queue.name, result.celery_kwargs["queue"])
```

nautobot/extras/views.py
```diff
@@ -1523,6 +1523,18 @@ def extra_post_save_action(self, obj, form):
...
             obj.job_queues.set(cleaned_data["job_queues"])
 
+        # Handle default job queue
+        clear_override_field = "clear_default_job_queue_override"
+        reset_override = cleaned_data.get(clear_override_field, False)
+        if reset_override:
+            meta_task_queues = obj.job_class.task_queues
+            job_queues = []
+            obj.default_job_queue_override = False
+            obj.default_job_queue = JobQueue.objects.get(name=meta_task_queues[0]) if meta_task_queues else None
```
glennmatthews, line 1533:
```suggestion
            obj.default_job_queue = JobQueue.objects.get(name=meta_task_queues[0] if meta_task_queues else settings.CELERY_TASK_DEFAULT_QUEUE)
```

glennmatthews: Might actually need `get_or_create` here to create the default queue if it doesn't exist?

nautobot/extras/migrations/0116_create_job_queue_model.py
HanlinMiao: Yep that's what Im bumping into in the CI I think, thank you!

nautobot/extras/api/views.py
HanlinMiao: Yeah valid_queues is used after this. If a user specified a task queue,  valid_queues is used to determine whether the user-specified queue is valid or not.

nautobot/extras/models/jobs.py
HanlinMiao: But would this create an "infinite loop"? Since u have to modify the m2m field within the signal handler?

glennmatthews: That's a good question. :) I'm not sure.

nautobot/extras/signals.py
```diff
@@ -455,6 +456,13 @@ def job_result_delete_associated_files(instance, **kwargs):
             file_proxy.file.delete()
 
 
+@receiver(post_save, sender=JobModel)
+def add_default_job_queue_to_job_queues(instance, **kwargs):
+    # Add specified default_job_queue to job.job_queues if it is not included.
+    if not instance.job_queues.filter(pk=instance.default_job_queue.pk).exists():
+        instance.job_queues.add(instance.default_job_queue)
```
glennmatthews, line 463: This still feels like the wrong place to do this. In response to your question about whether doing this in the `m2m_changed` signal will be problematic, StackOverflow at least [suggests](https://stackoverflow.com/questions/26493254/using-djangos-m2m-changed-to-modify-what-is-being-saved-pre-add) that updating `pk_set` in the signal handler is a valid way to influence the set of M2M relations.

nautobot/extras/utils.py
```diff
@@ -514,6 +514,10 @@ def refresh_job_model_from_job_class(job_model_class, job_class, job_queue_class
 
     try:
         with transaction.atomic():
+            default_job_queue, _ = job_queue_class.objects.get_or_create(
+                name=job_class.task_queues[0] if job_class.task_queues else settings.CELERY_TASK_DEFAULT_QUEUE,
+                defaults={"queue_type": JobQueueTypeChoices.TYPE_CELERY},
```
glennmatthews, line 519: Thinking toward the future, at some point we might want a `NAUTOBOT_DEFAULT_QUEUE_TYPE` config variable for all of these auto-creation cases. Not blocking at this time.

nautobot/extras/tests/test_api.py
```diff
@@ -1537,6 +1537,7 @@ def test_run_job_not_installed(self, mock_get_worker_count):
             installed=False,
             enabled=True,
         )
+        job_model.default_job_queue = JobQueue.objects.get(name="default", queue_type=JobQueueTypeChoices.TYPE_CELERY)
```
glennmatthews, line 1540: Why do this as a separate step rather than as a part of lines 1532-1539 above?

nautobot/extras/tests/test_api.py
```diff
@@ -1716,7 +1718,7 @@ def test_run_job_response_job_result(self, mock_get_worker_count):
         response = self.client.post(url, {"data": job_data}, format="json", **self.header)
         self.assertHttpStatus(response, self.run_success_response_status)
 
-        job_result = JobResult.objects.get(name=self.job_model.name)
+        job_result = JobResult.objects.latest()
```
glennmatthews, line 1721: Why this change? Should we at least include a `filter(name=self.job_model.name)` before the `.latest()`?

nautobot/extras/forms/forms.py
```diff
@@ -997,10 +1005,19 @@ def clean(self):
             for field_name in JOB_OVERRIDABLE_FIELDS:
                 if not cleaned_data.get(f"{field_name}_override", False):
                     cleaned_data[field_name] = getattr(job_class, field_name)
+            # Get default Job Queue first
+            if not cleaned_data.get("default_job_queue_override", False):
+                meta_task_queues = getattr(job_class, "task_queues", []) or [settings.CELERY_TASK_DEFAULT_QUEUE]
+                cleaned_data["default_job_queue"] = JobQueue.objects.get(name=meta_task_queues[0])
```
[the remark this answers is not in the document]
HanlinMiao: Should be get_or_create since this is a POST request? I am not sure honestly.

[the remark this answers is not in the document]
glennmatthews: Me neither, but also leaning toward get_or_create.

nautobot/docs/development/jobs/index.md
```diff
@@ -309,10 +309,10 @@ class ExampleJobWithSoftTimeLimit(Job):
...
+A list of Job Queue names that the job can be routed to. An empty list will default to only allowing the user to select the [default Celery queue](../../user-guide/administration/configuration/settings.md#celery_task_default_queue) (`default` unless changed by an administrator). The queue specified in the job's `default_job_queue` will be used if a queue is not specified in a job run API call.
 
 +/- 2.4.0 "Changed default queue selection"
-    As a result of the addition of Job Queues, the default queue when running a Job without explicitly selecting a queue is now the alphabetically first queue of those listed, rather than the first queue specified in this list.
+    As a result of the addition of Job Queues, the default queue when running a Job without explicitly selecting a queue is now the job queue specified in the `default_job_queue` field on the Job model. When that value is empty, the job will be run on the job queue with the name provided by `settings.CELERY_TASK_DEFAULT_QUEUE`.
```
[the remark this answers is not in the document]
glennmatthews: Still wondering on this point. :)

nautobot/extras/tests/test_api.py
```diff
@@ -1784,8 +1786,8 @@ def test_run_job_with_no_queue_specified(self, mock_get_worker_count):
             **self.header,
         )
         self.assertHttpStatus(response, self.run_success_response_status)
-        latest_job_result = JobResult.objects.latest()
-        self.assertEqual(job_model.default_job_queue.name, latest_job_result.celery_kwargs["queue"])
+        # latest_job_result = JobResult.objects.latest()
+        # self.assertEqual(job_model.default_job_queue.name, latest_job_result.celery_kwargs["queue"])
```
HanlinMiao, line 1826: These fields are commented out because the "queue" key is not available for some reason. In `JobResult.enqueue_job` if the synchronous argument is specified as `False`, JobResult will have an empty `celery_kwargs`

glennmatthews: Huh. Maybe it only gets populated when the job actually gets picked up for execution?

nautobot/docs/development/jobs/index.md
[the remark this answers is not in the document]
HanlinMiao: The latest commit should address some of the behavior changes that users should be made aware of.

nautobot/extras/models/jobs.py
```diff
@@ -816,6 +816,8 @@ def enqueue_job(
         if celery_kwargs is not None:
             job_celery_kwargs.update(celery_kwargs)
 
+        print("hey")
+        print(synchronous)
```
glennmatthews, line 820: revert

nautobot/extras/models/jobs.py
```diff
@@ -852,6 +854,7 @@ def enqueue_job(
             job_result.save()
         else:
             # Jobs queued inside of a transaction need to run after the transaction completes and the JobResult is saved to the database
+            print("hmmm")
```
glennmatthews, line 857: revert

nautobot/extras/signals.py
```diff
@@ -455,6 +456,22 @@ def job_result_delete_associated_files(instance, **kwargs):
             file_proxy.file.delete()
 
 
+@receiver(m2m_changed, sender=JobModel)
+def add_default_job_queue_to_job_queues(instance, **kwargs):
+    # Add specified default_job_queue to job.job_queues if it is not included.
+    default_job_queue = instance.default_job_queue
+    action = kwargs.pop("action", None)
+    if action in ["post_remove"]:
+        instance.job_queues.add(default_job_queue)
```
glennmatthews, line 465: Perhaps this needs to be something like:

```suggestion
def add_default_job_queue_to_job_queues(instance, action, model, pk_set, **kwargs):
    if action == "pre_remove":
        if model is JobModel:  # job_model.job_queues.remove()
            # Don't allow removing the default job queue
            pk_set.discard(instance.default_job_queue.pk)
        elif model is JobQueue:  # job_queue.jobs.remove()
            # Don't allow removing jobs that this queue is default for
            for job_model in instance.default_for_jobs.all():
                pk_set.discard(job_model.pk)
    elif action == "post_clear":
        if model is JobModel:  # job_model.job_queues.clear()
            # Re-add the default job queue
            instance.job_queues.add(instance.default_job_queue)
        elif model is JobQueue:  # job_queue.jobs.clear()
            # Re-add the jobs this queue is default for
            for job_model in instance.default_for_jobs.all():
                job_model.job_queues.add(instance.pk)
```

nautobot/extras/forms/forms.py
```diff
@@ -997,10 +1005,21 @@ def clean(self):
...
                     cleaned_data[field_name] = getattr(job_class, field_name)
-            if not cleaned_data.get("job_queues_override", False):
-                cleaned_data["job_queues"] = JobQueue.objects.filter(
-                    name__in=getattr(job_class, "task_queues", []) or [settings.CELERY_TASK_DEFAULT_QUEUE]
+            # Get default Job Queue first
+            if not cleaned_data.get("default_job_queue_override", False):
+                meta_task_queues = getattr(job_class, "task_queues", []) or [settings.CELERY_TASK_DEFAULT_QUEUE]
+                cleaned_data["default_job_queue"], _ = JobQueue.objects.get_or_create(
+                    name=meta_task_queues[0], defaults={"queue_type": JobQueueTypeChoices.TYPE_CELERY}
                 )
```
glennmatthews, line 1013: Maybe just a nitpick but `form.clean()` ideally shouldn't be creating a DB record, this should be happening in `form.save()`. Not sure how important this is though.

nautobot/extras/models/jobs.py
```diff
@@ -212,14 +212,25 @@ class Job(PrimaryModel):
...
     )
-
+    default_job_queue = models.ForeignKey(
+        to="extras.JobQueue",
+        related_name="default_for_jobs",
+        on_delete=models.PROTECT,
+        verbose_name="Default Job Queue",
+        null=False,
+        blank=False,
+    )
```
glennmatthews, line 233: nit: can we move `job_queues` and `default_job_queue` definition up to around line 173 so that they're kept with the other model fields rather than down here after all the `override` fields?

nautobot/extras/models/jobs.py
```diff
@@ -784,7 +795,9 @@ def enqueue_job(
         )
 
         if task_queue is None:
-            task_queue = settings.CELERY_TASK_DEFAULT_QUEUE
+            task_queue = (
+                job_model.default_job_queue.name if job_model.default_job_queue else settings.CELERY_TASK_DEFAULT_QUEUE
+            )
```
glennmatthews, line 800: default_job_queue is non-nullable now, so:

```suggestion
            task_queue = job_model.default_job_queue.name
```

nautobot/extras/tests/test_api.py
```diff
@@ -1791,6 +1793,49 @@ def test_run_job_with_both_task_queue_and_job_queue_specified(self, mock_get_wor
...
+        url = self.get_run_url()
+        response = self.client.post(
+            url,
+            {
+                "data": job_data,
+            },
+            format="json",
+            **self.header,
+        )
+        self.assertHttpStatus(response, self.run_success_response_status)
```
glennmatthews, line 1837: Without being able to check the celery_kwargs on the job result, this test case seems like it's not really doing much.

nautobot/extras/tests/test_models.py
glennmatthews: Swap it around?

nautobot/extras/tests/test_views.py
```diff
@@ -2683,6 +2692,7 @@ def test_rerun_job(self):
             "queue": job_queue.name,
         }
         self.test_required_args.job_queues.set([job_queue])
+        print(self.test_required_args.job_queues)
```
glennmatthews, line 2695: revert?

nautobot/extras/tests/test_views.py
```diff
@@ -2693,6 +2703,7 @@ def test_rerun_job(self):
         run_url = reverse("extras:job_run", kwargs={"pk": self.test_required_args.pk})
         response = self.client.get(f"{run_url}?kwargs_from_job_result={previous_result.pk!s}")
         content = extract_page_body(response.content.decode(response.charset))
+        print(content)
```
glennmatthews, line 2706: revert?

nautobot/extras/tests/test_views.py
```diff
@@ -2827,6 +2838,28 @@ def test_run_job_with_invalid_task_queue(self, _):
...
+        self.test_pass.save()
+        data = {
+            "_schedule_type": "immediately",
+        }
+
+        for run_url in self.run_urls:
+            response = self.client.post(run_url, data)
+            result = JobResult.objects.latest()
+            self.assertIsNotNone(result, msg=run_url)
+            # self.assertEqual(self.test_pass.default_job_queue.name, result.celery_kwargs["queue"])
```
glennmatthews, line 2860: Add a note explaining why we can't check this?

Does this test provide significant value even without being able to check this?

HanlinMiao: Removed

nautobot/extras/tests/test_api.py
HanlinMiao: Removed

glennmatthews (approved):

PR #6278, opened 2024-09-26
Fix Prefix not displaying the correct utilization for IP Addresses

### LINKED ISSUE #6170: Network Prefix does not display the correct utilization for IP Addresses, and may present in-use IP Addresses
### Environment
* Nautobot version (Docker tag too if applicable): v2.3.1
* Python version: 3.11
* Middleware(s): N/A

There is a bug in the UI view where we can have a Network Prefix object that displays the full host subnet (n-2) as available, when there are IP Addresses in use from that subnet (that belong to a child Prefix object of that parent Network Prefix).

/23 Subnet with n-2 (510-hosts) should instead display 509-hosts. (I have a 5.5.2.1/23 IP Address object under a Child Prefix)
image

When you press the button to add an IP Address to the parent Prefix, it can also auto-fill an unavailable and in-use IP Address, which is prevented from being created on form submit.
image

I also noticed, when you add 1 (or more) child IP Address objects, that directly belong to the parent Network Prefix it will then account for both the used IP Addresses from Child Prefix objects and the ones that directly belong to the Network Prefix. And if you delete the IP Address object it will return to the previous view.

### Steps to Reproduce
1. Create Prefix object, eg: 5.5.10.0/23, type=Network
2. Create a Child Prefix object inside of that Subnet, eg: 5.5.10.1/30 (Pool or Network type result is the same but Network will give warning)
3. Add an IP Address object from the Child Prefix object, eg: 5.5.10.1/32
4. Observe that the Network Prefix 'IP Addresses' tab does not update IP Addresses available to 509-hosts
5. Optionally, now create an IP Address object inside of 5.5.10.0/23, but outside of 5.5.10.1/30 and the UI will account for the nested Prefix used IP Addresses

### Expected Behavior
In-use hosts are represented in the UI

### Observed Behavior
In-use hosts not not represented under this condition

timizuoebideri1:
# Closes #6170 

# What's Changed

# Screenshots

**Child Prefix and its IP**

Screenshot 2024-09-25 at 10 28 17 PM

**Parent Prefix including Childs IPs**
Screenshot 2024-09-27 at 3 01 18 PM

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/ipam/models.py
+++ b/nautobot/ipam/models.py
@@ -898,6 +898,7 @@ class Prefix(PrimaryModel):
         Returns:
             IPAddress QuerySet
         """
+        # Question: This is a possible fix for this PR; should this be uncommented?
         # 3.0 TODO: uncomment this to enable this logic
         # if self.type == choices.PrefixTypeChoices.TYPE_POOL:
         #     return IPAddress.objects.filter(
--- a/nautobot/ipam/models.py
+++ b/nautobot/ipam/models.py
@@ -889,23 +889,16 @@ class Prefix(PrimaryModel):
         )
         return available_ips
 
-    def get_child_ips(self):
+    def get_all_ips(self):
         """
-        Return IP addresses with this prefix as parent.
-
-        In a future release, if this prefix is a pool, it will return IP addresses within the pool's address space.
+        Return IP addresses with this prefix as parent including child prefixes IP addresses.
 
         Returns:
             IPAddress QuerySet
         """
-        # Question: This is a possible fix for this PR; should this be uncommented?
-        # 3.0 TODO: uncomment this to enable this logic
-        # if self.type == choices.PrefixTypeChoices.TYPE_POOL:
-        #     return IPAddress.objects.filter(
-        #         parent__namespace=self.namespace, host__gte=self.network, host__lte=self.broadcast
-        #     )
-        # else:
-        return self.ip_addresses.all()
+        return IPAddress.objects.filter(
+            parent__namespace=self.namespace, host__gte=self.network, host__lte=self.broadcast
+        )
 
     def get_first_available_prefix(self):
         """
--- a/nautobot/ipam/models.py
+++ b/nautobot/ipam/models.py
@@ -891,7 +891,7 @@ class Prefix(PrimaryModel):
 
     def get_all_ips(self):
         """
-        Return IP addresses with this prefix as parent including child prefixes IP addresses.
+        Return all IP addresses contained within this prefix, including child prefixes' IP addresses.
 
         Returns:
             IPAddress QuerySet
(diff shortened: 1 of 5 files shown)


nautobot/ipam/models.py
```diff
@@ -898,6 +898,7 @@ def get_child_ips(self):
         Returns:
             IPAddress QuerySet
         """
+        # Question: This is a possible fix for this PR; should this be uncommented?
```
timizuoebideri1, line 901: ?

glennmatthews: As discussed in standup, yeah, let's go with changing the behavior of `get_child_ips` since it's not used anywhere else that we're aware of.

nautobot/ipam/models.py
```diff
@@ -889,22 +889,16 @@ def get_available_ips(self):
         )
         return available_ips
 
-    def get_child_ips(self):
+    def get_all_ips(self):
         """
-        Return IP addresses with this prefix as parent.
-
-        In a future release, if this prefix is a pool, it will return IP addresses within the pool's address space.
+        Return IP addresses with this prefix as parent including child prefixes IP addresses.
```
glennmatthews, line 894:
```suggestion
        Return all IP addresses contained within this prefix, including child prefixes' IP addresses.
```

glennmatthews (commented):
Mind providing a screenshot or two showing the fixed UI behavior?

nautobot/ipam/tests/test_views.py
```diff
@@ -328,6 +328,44 @@ def test_create_object_warnings(self):
...
+        IPAddress.objects.create(
+            address="5.5.10.1/32",
+            status=ip_status,
+            namespace=self.namespace,
+        )
+        IPAddress.objects.create(
+            address="5.5.10.4/30",
+            status=ip_status,
+            namespace=self.namespace,
+        )
```
glennmatthews, line 355:
```suggestion
        Prefix.objects.create(
            prefix="5.5.10.0/30",
            namespace=self.namespace,
            type=PrefixTypeChoices.TYPE_POOL,
            status=self.statuses[1],
        )
        IPAddress.objects.create(
            address="5.5.10.1/23",
            status=ip_status,
            namespace=self.namespace,
        )
        IPAddress.objects.create(
            address="5.5.10.4/23",
            status=ip_status,
            namespace=self.namespace,
        )
```

Just making the parent/child relationships a bit clearer.

nautobot/ipam/tests/test_views.py
```diff
@@ -328,6 +328,44 @@ def test_create_object_warnings(self):
...
+        url = reverse("ipam:prefix_ipaddresses", args=(instance.pk,))
+        response = self.client.get(url)
+        self.assertHttpStatus(response, 200)
+        content = response.content.decode(response.charset)
+        # This validates that both parent prefix and child prefix IPAddresses are present in parent prefix IPAddresses list
+        self.assertIn("5.5.10.1/32", strip_tags(content))
+        self.assertIn("5.5.10.4/30", strip_tags(content))
```
glennmatthews, line 362:
```suggestion
        self.assertIn("5.5.10.1/23", strip_tags(content))
        self.assertIn("5.5.10.4/23", strip_tags(content))
```

glennmatthews (approved):

PR #6290, opened 2024-09-29
Implement Service Add button.

### LINKED ISSUE #6289: IPAM Navigation Services Add button is missing
### Environment
* Nautobot version (Docker tag too if applicable): 2.3.X
* Python version: 3.11
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Go to demo.nautobot.com
2. Expand IPAM on the left hand navbar.
3. Services does not have an add button

### Expected Behavior

I can add a service from the Navbar, like other models.

### Observed Behavior

The Services does not have an Add button on the Navbar.

smk4664:
# Closes #6289 
# What's Changed

Implement IPAM Service Add button in the navigation. It seems this was never in the codebase, so maybe not a bug, but having one navigation item different from others for seemingly no specific reason felt like a bug to me.

# Screenshots

Before
image

After
image

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/6289.added
+++ b/changes/6289.added
@@ -1,0 +1,1 @@
+Added the add button to IPAM Services.
--- a/nautobot/ipam/navigation.py
+++ b/nautobot/ipam/navigation.py
@@ -191,7 +191,14 @@ menu_items = (
                         permissions=[
                             "ipam.view_service",
                         ],
-                        buttons=(),
+                        buttons=(
+                            NavMenuAddButton(
+                                link="ipam:service_add",
+                                permissions=[
+                                    "ipam.add_service",
+                                ],
+                            ),
+                        ),
                     ),
                 ),
             ),


glennmatthews:
How's the UX around this? IIRC the existing functionality of the edit form pretty much assumes you're explicitly specifying either a Device or a VirtualMachine to associate the Service to as a URL query parameter - in particular I see that it looks like the `ip_addresses` form field is prepopulated with static options depending on the Device or VM, and it looks like if neither one is specified, that form field currently gets an empty list of choices, which is probably not ideal.

In other words, I agree from a consistency standpoint that having an Add button in the nav makes sense, but I'm suspicious that this may not go far enough to actually make that link usable.

smk4664:
> How's the UX around this? IIRC the existing functionality of the edit form pretty much assumes you're explicitly specifying either a Device or a VirtualMachine to associate the Service to as a URL query parameter - in particular I see that it looks like the `ip_addresses` form field is prepopulated with static options depending on the Device or VM, and it looks like if neither one is specified, that form field currently gets an empty list of choices, which is probably not ideal.
> 
> In other words, I agree from a consistency standpoint that having an Add button in the nav makes sense, but I'm suspicious that this may not go far enough to actually make that link usable.

I was wondering why the IP Address field was blank when creating.  The form is semi-usable then. You can select a Device or Virtual Machine when adding the Service if you use this button. This is also the same functionality you get when you go to the Services List endpoint, you can click Add Service. I think if this is decided not to be implemented, we should probably remove the Add button on the Services List page, but that is my opinion.

changes/6289.fixed
```diff
@@ -0,0 +1 @@
+Service Form can now dynamically lookup IPAddresses.
```
glennmatthews, line 1:
```suggestion
Fixed lookup of IP Addresses in the Service form.
```

glennmatthews (approved):

HanlinMiao (approved):

PR #6306, opened 2024-10-02
Enhance ObjectsTablePanel

### LINKED ISSUE #6136: UI Components: BaseTable backed table panels
Tables backed by BaseTable (Nautobot base table class).

Simplify table config:
- Include/exclude columns (fields)
- Record limit (max)
  - Records over the max result in a "x more objects" link to the list view
- Optional add button (takes a view reference as input)
  - Account for permissions

timizuoebideri1:
# Closes #6136 
# What's Changed

# Screenshots

Screenshot 2024-10-02 at 1 14 36 PM

# TODO

- [ ] Unit, Integration Tests


nautobot/circuits/views.py
```diff
@@ -102,6 +103,31 @@ class ProviderUIViewSet(NautobotUIViewSet):
     queryset = Provider.objects.all()
     serializer_class = serializers.ProviderSerializer
     table_class = tables.ProviderTable
+    object_detail_content = ObjectDetailContent(
+        panels=(
+            ObjectFieldsPanel(
+                section=SectionChoices.LEFT_HALF,
+                weight=100,
+                fields="__all__",
+                exclude_fields=["name", "comments"],
```
timizuoebideri1, line 112: TODO: 
 Missing Fields: `Circuits` - circuits_count
 Missing Feature: `Customer Portal` is not hyperlinked

nautobot/circuits/views.py
```diff
@@ -102,6 +103,27 @@ class ProviderUIViewSet(NautobotUIViewSet):
...
+            ),
+            ObjectFieldsPanel(
+                label="Comments",
+                weight=200,
+                section=SectionChoices.RIGHT_HALF,
+                fields=["comments"],
+                value_transforms={"comments": [render_markdown, placeholder]},
+            ),
```
glennmatthews, line 120: Let's look at making `comments` an automatic exclusion and automatic separate panel much like `tags` is at present. Maybe out of scope for this PR; if you'd prefer to open as a separate issue that's fine by me.

nautobot/core/templates/components/panel/header_table.html
```diff
@@ -0,0 +1,3 @@
+{% load helpers %}
+
+<strong>{{ body_content_table_verbose_name_plural|bettertitle }}</strong> <span class="badge badge-default">{{ body_content_table.rows|length }}</span>
```
glennmatthews, line 3: Sooner or later we'll probably want an option to specify a custom title for one of these panel tables. Maybe not essential at this moment, but maybe have the `verbose_name_plural|bettertitle` be done in code instead of in the template?

nautobot/core/templates/components/panel/footer_table.html
```diff
@@ -0,0 +1,17 @@
...
+        {% endif %}
+
+        {% if body_content_table_list_url and more_queryset_count > 0 %}
+        <a href="{{ body_content_table_list_url }}" class="btn btn-default btn-xs">
+            <span class="mdi mdi-database-search-outline" aria-hidden="true"></span> View {{ more_queryset_count }} more object(s)
+        </a>
+        {% endif %}
+    {% endblock %}
+</div>
+<div class="clearfix"></div>
```
glennmatthews, line 17: Is the `clearfix` div actually needed here? I've never been real clear what it actually does, especially within a panel.

nautobot/core/tests/test_views.py
```diff
@@ -611,3 +612,42 @@ def test_permission_classes_attribute_is_enforced(self):
...
+            for x in range(10)
+        ]
+        Circuit.objects.bulk_create(circuits)
+
+        self.add_permissions("circuits.view_provider", "circuits.view_circuit")
+        url = reverse("circuits:provider", args=(provider.pk,))
+        response = self.client.get(f"{url}?tab=main")
+        self.assertHttpStatus(response, 200)
+        response_data = response.content.decode(response.charset)
+        print(response_data)
```
glennmatthews, line 642: remove?

nautobot/core/ui/object_detail.py
```diff
@@ -378,25 +384,111 @@ def __init__(
...
+        header_extra_content_template_path="components/panel/header_table.html",
+        footer_content_template_path="components/panel/footer_table.html",
         **kwargs,
     ):
         """Instantiate an ObjectsTable panel.
 
         Args:
             table_key (str): The render context key string that contains the BaseTable instance of interest.
+            max_display_count (int, optional):  Maximum number of items to display in the table.
+                If None, defaults to the user's preference or a global setting.
```
glennmatthews, line 404: Can you clarify/elaborate on this? Maybe just reference `get_paginate_count()` if that's what makes the decision?

nautobot/core/ui/object_detail.py
```diff
@@ -378,25 +384,111 @@ def __init__(
...
             **kwargs,
         )
+        self.max_display_count = max_display_count
+        self.add_button_route = add_button_route
+        if exclude_fields and include_fields:
+            raise ValueError("You can only specify either `exclude_fields` or `include_fields`")
+        self.exclude_fields = exclude_fields
+        self.include_fields = include_fields
+        self.related_field_name = related_field_name
+        self.add_permissions = add_permissions
```
glennmatthews, line 431: Just self-consistency in ordering:

```suggestion
        if exclude_fields and include_fields:
            raise ValueError("You can only specify either `exclude_fields` or `include_fields`")
        self.include_fields = include_fields
        self.exclude_fields = exclude_fields
        self.add_button_route = add_button_route
        self.add_permissions = add_permissions
        self.related_field_name = related_field_name
```

Actually, is there a reason these are being done after the `super().__init__()` instead of before it like `table_key` is?

nautobot/core/ui/object_detail.py
```diff
@@ -378,25 +384,111 @@ def __init__(
...
+            body_content_table = context[self.table_key]
+
+            body_content_table_model = body_content_table.Meta.model
+            permission_name = get_permission_for_model(body_content_table_model, "add")
+            if request.user.has_perms([permission_name]):
+                with contextlib.suppress(NoReverseMatch):
+                    add_route = reverse(get_route_for_model(body_content_table_model, "add"))
+                    body_content_table_add_url = f"{add_route}?{related_field_name}={obj.pk}&return_url={return_url}"
+
+        elif self.add_button_route not in [None, "default"]:
```
glennmatthews, line 450:
```suggestion
        elif self.add_button_route is not None:
```

nautobot/core/ui/object_detail.py
```diff
@@ -378,25 +384,111 @@ def __init__(
...
+        return_url = context["return_url"]
+
+        if self.add_button_route == "default":
+            body_content_table = context[self.table_key]
+
+            body_content_table_model = body_content_table.Meta.model
+            permission_name = get_permission_for_model(body_content_table_model, "add")
+            if request.user.has_perms([permission_name]):
+                with contextlib.suppress(NoReverseMatch):
+                    add_route = reverse(get_route_for_model(body_content_table_model, "add"))
```
glennmatthews, line 447: Feels like maybe this should be a try/except that logs a message in the NoReverseMatch case, since it indicates a misconfiguration?

nautobot/core/ui/object_detail.py
```diff
@@ -378,25 +384,111 @@ def __init__(
...
+        self.include_fields = include_fields
+        self.related_field_name = related_field_name
+        self.add_permissions = add_permissions
+
+    def _get_table_add_url(self, context):
+        obj = context["obj"]
+        body_content_table_add_url = None
+        request = context["request"]
+        related_field_name = self.related_field_name or obj._meta.model_name
+        return_url = context["return_url"]
```
glennmatthews, line 438: Is this context variable guaranteed to be present and set? Or should it default to `obj.get_absolute_url()` if absent?

nautobot/core/ui/object_detail.py
```diff
@@ -378,25 +384,111 @@ def __init__(
...
+        self.max_display_count = max_display_count
+        self.add_button_route = add_button_route
+        if exclude_fields and include_fields:
+            raise ValueError("You can only specify either `exclude_fields` or `include_fields`")
+        self.exclude_fields = exclude_fields
+        self.include_fields = include_fields
+        self.related_field_name = related_field_name
+        self.add_permissions = add_permissions
+
+    def _get_table_add_url(self, context):
```
glennmatthews, line 433: Docstring please.

nautobot/core/ui/object_detail.py
```diff
@@ -378,25 +384,111 @@ def __init__(
...
+
+        elif self.add_button_route not in [None, "default"]:
+            if request.user.has_perms(self.add_permissions or []):
+                add_route = reverse(self.add_button_route)
+                body_content_table_add_url = f"{add_route}?{related_field_name}={obj.pk}&return_url={return_url}"
+
+        return body_content_table_add_url
 
     def get_extra_context(self, context):
         """Take the context value under `self.table_key` and add it to the context under key `"body_content_table"`."""
```
glennmatthews, line 458: Mind updating this since it does a lot more now?

nautobot/core/ui/object_detail.py
```diff
@@ -378,25 +384,111 @@ def __init__(
...
-        return {"body_content_table": context[self.table_key]}
+        body_content_table = context[self.table_key]
+        request = context["request"]
+
+        if self.exclude_fields is not None:
+            body_content_table.exclude = self.exclude_fields
+        elif self.include_fields is not None:
+            body_content_table.Meta.default_columns = self.include_fields
```
glennmatthews, line 465: Is this the best set of APIs to use, or should it instead be something like:

```py
if self.exclude_fields or self.include_fields:
    for column in body_content_table.columns:
        if (exclude_fields and column.name in exclude_fields) or (include_fields and column.name not in include_fields):
            body_content_table.columns.hide(column.name)
        else:
            body_content_table.columns.show(column.name)
```

Alternately, maybe we should change up how things work in the **view** `get_extra_context` and have it provide just the table class and the queryset in its context, and let this actually initialize the table instance?

nautobot/core/ui/object_detail.py
```diff
@@ -378,25 +384,111 @@ def __init__(
...
+
+        body_content_table_add_url = self._get_table_add_url(context)
+
+        return {
+            "body_content_table": body_content_table,
+            "body_content_table_add_url": body_content_table_add_url,
+            "body_content_table_list_url": body_content_table_list_url,
+            "body_content_table_verbose_name": body_content_table_model._meta.verbose_name,
+            "body_content_table_verbose_name_plural": body_content_table_model._meta.verbose_name_plural,
+            "more_queryset_count": max(more_queryset_count, 0),
```
glennmatthews, line 490: Maybe move the `max()` call up to line 470 where `more_queryset_count` is set in the first place?

nautobot/core/ui/object_detail.py
```diff
@@ -502,7 +594,10 @@ def render_value(self, key, value, context) -> str:
         - Etc.
         """
         display = value
-
+        try:
+            field_instance = context["obj"]._meta.get_field(key)
+        except FieldDoesNotExist:
+            field_instance = None
```
glennmatthews, line 600: I was trying to keep `KeyValueTablePanel` agnostic of model objects - if so it might be cleaner to implement this case in `ObjectFieldsPanel.render_value` instead, similar to what it's already doing for `render_key`?

nautobot/circuits/views.py
timizuoebideri1: https://github.com/nautobot/nautobot/issues/6321

nautobot/core/ui/object_detail.py
timizuoebideri1: Ive moved this to part of the acceptance criteria for `UI Components: Key/value model fields`, hence illd resolve it in that PR and remove this changes from this PR

> Key value pairs of model fields - some built in smarts for commonly implemented features
> - ...
> - URL fields should be automatically hyperlinked

nautobot/circuits/views.py
```diff
@@ -102,6 +103,27 @@ class ProviderUIViewSet(NautobotUIViewSet):
     queryset = Provider.objects.all()
     serializer_class = serializers.ProviderSerializer
     table_class = tables.ProviderTable
+    object_detail_content = ObjectDetailContent(
+        panels=(
+            ObjectFieldsPanel(
+                section=SectionChoices.LEFT_HALF,
+                weight=100,
+                fields="__all__",
+                exclude_fields=["name", "comments"],
```
glennmatthews, line 112: Is excluding `name` correct/desirable? Even if it's a bit redundant with the page title/header I feel like we should include it for consistency.

nautobot/circuits/views.py
```diff
@@ -102,6 +103,27 @@ class ProviderUIViewSet(NautobotUIViewSet):
...
+            ),
+            ObjectFieldsPanel(
+                label="Comments",
+                weight=200,
+                section=SectionChoices.RIGHT_HALF,
+                fields=["comments"],
+                value_transforms={"comments": [render_markdown, placeholder]},
+            ),
+            ObjectsTablePanel(
+                weight=100,
```
glennmatthews, line 122: - same `weight` as the `ObjectFieldsPanel` above?
- might be good for clarity to explicitly specify the `section`?

nautobot/core/templates/components/panel/footer_table.html
```diff
@@ -0,0 +1,17 @@
+{% block footer_left %}
+    {% if body_content_table_list_url and more_queryset_count > 0 %}
+        <a href="{{ body_content_table_list_url }}" class="btn btn-default btn-xs">
+            <span class="mdi mdi-database-search-outline" aria-hidden="true"></span> View {{ more_queryset_count }} more object(s)
```
glennmatthews, line 4: Maybe `mdi-dots-vertical` or `mdi-dots-horizontal` instead? Or no icon at all? This isn't a search operation per se so I find `mdi-database-search-outline` a bit unintuitive.

glennmatthews: Also wondering if it really should be a button, or whether it should just be a normal hyperlink?

nautobot/core/templates/components/panel/footer_table.html
```diff
@@ -0,0 +1,17 @@
+{% block footer_left %}
```
glennmatthews, line 1: Let's for now not add the `footer_left` and `footer_right` blocks until/unless we find we need them.

nautobot/core/templates/components/panel/header_table.html
```diff
@@ -0,0 +1,3 @@
+{% load helpers %}
+
+<strong>{{ body_content_table_verbose_name_plural }}</strong>  <a href="{{ body_content_table_list_url }}" class="badge badge-primary">{{ body_content_table.rows|length }}</a>
```
glennmatthews, line 3: For consistency, does this also need a `if body_content_table_list_url` check?

nautobot/core/templatetags/helpers.py
```diff
@@ -96,6 +96,15 @@ def hyperlinked_phone_number(value):
...
 
+@library.filter()
+@register.filter()
+def hyperlinked_field(value, hyperlink):
+    """Render a value as a hyperlink."""
+    if not value:
+        return placeholder(value)
+    return format_html('<a href="{}">{}</a>', value, hyperlink)
+
+
```
glennmatthews, line 107: Remove this since we're deferring that feature to another issue?

nautobot/core/templates/components/panel/footer_table.html
glennmatthews: Also, can we have this use `body_content_table_verbose_name`/`body_content_table_verbose_name_plural` in place of the generic `object(s)`?

nautobot/core/ui/object_detail.py
```diff
@@ -378,25 +385,132 @@ def __init__(
...
+        max_display_count=None,
+        include_fields=None,
+        exclude_fields=None,
+        add_button_route="default",
+        add_permissions=None,
+        related_field_name=None,
         body_wrapper_template_path="components/panel/body_wrapper_table.html",
         body_content_template_path="components/panel/body_content_table.html",
+        header_extra_content_template_path="components/panel/header_table.html",
+        footer_content_template_path="components/panel/footer_table.html",
```
glennmatthews, line 398: For self-consistency, should these maybe be named `header_extra_content_table.html` and `footer_content_table.html`?

nautobot/core/ui/object_detail.py
```diff
@@ -378,25 +385,132 @@ def __init__(
...
+
+        body_content_table_add_url = self._get_table_add_url(context)
+        body_content_table_verbose_name_plural = self.table_title or body_content_table_model._meta.verbose_name_plural
+
+        return {
+            "body_content_table": body_content_table,
+            "body_content_table_add_url": body_content_table_add_url,
+            "body_content_table_list_url": body_content_table_list_url,
+            "body_content_table_verbose_name": body_content_table_model._meta.verbose_name,
+            "body_content_table_verbose_name_plural": bettertitle(body_content_table_verbose_name_plural),
```
glennmatthews, line 511: Feels self-inconsistent that `verbose_name` is as-is but `verbose_name_plural` is title-cased. Let's remove the bettertitle here and do that in the template as needed instead.

nautobot/circuits/views.py
timizuoebideri1: Yup, `name` was not part of the previous location type details view

timizuoebideri1: This is correct since `name` was not part of the Provider detail view before.

glennmatthews: My vote is to add it now.

nautobot/core/ui/object_detail.py
glennmatthews: > Alternately, maybe we should change up how things work in the view get_extra_context and have it provide just the table class and the queryset in its context, and let this actually initialize the table instance?

Thoughts on this idea? The current pattern feels like duplicate code to me since we have to create and configure the table in the view code, then this code here overrides that with its own configuration.

glennmatthews:
Needs a change fragment for #6136.

nautobot/core/ui/object_detail.py
timizuoebideri1: This would require me to also to modify `associated_contacts_table` in `nautobot.core.views.utils.common_detail_view_context` to supply both the `AssociatedContactsTable ` and its `queryset`. This would likely involve updates in other areas, such as the `NautobotHTMLRenderer`. 
Additionally, we can’t specify fields to include at table initialization—only fields to exclude—so we might end up using a similar approach to the current solution i.e `body_content_table.columns.show(column.name)`.

Given this, I believe the current implementation is simple and sufficient for now.

glennmatthews: Fair. I still feel the current pattern involves more redundancy than is desirable, but I'll open a follow up issue rather than holding this one up any further.

nautobot/circuits/views.py
```diff
@@ -102,6 +103,28 @@ class ProviderUIViewSet(NautobotUIViewSet):
     queryset = Provider.objects.all()
     serializer_class = serializers.ProviderSerializer
     table_class = tables.ProviderTable
+    object_detail_content = ObjectDetailContent(
```
glennmatthews, line 106: Should be possible to rip most content out of the `provider_retrieve.html` template now. (Leave the skeleton of the template in case an App is providing a custom tab and still using the base template)

glennmatthews (approved):

HanlinMiao (approved):

Issue #6321, opened 2024-10-04
UI Components: Comments automatic panel just like tags


timizuoebideri1:
### Proposed Changes
`comments` should automatically  be excluded from Panels and automatic have separate panel much like `tags` is at present

Issue #6139, opened 2024-08-23
UI Components: text based content


lampwins:
Text content, markdown, or a template fragment (this is explicitly separate from the base component that points to a template file).
- Data passed into the render context, or as an accessor at declaration time

PR #6328, opened 2024-10-07
Enhance ObjectFieldPanel

### LINKED ISSUE #6135: UI Components: Key/value model fields
Key value pairs of model fields - some built in smarts for commonly implemented features
  - Linkifying FKs
  - Rendering proper field types (JSON, Boolean, Blankable/nullable, etc)
  - Hierarchy data (e.g. Locations)
  - custom rendering: django_tables2 `render_<field>()` pattern
  - Copy button at row (value) level
  - URL fields should be automatically hyperlinked

timizuoebideri1:
# Closes #6135, #6321
# What's Changed

# Screenshots

Screenshot 2024-10-07 at 1 29 54 PM

# TODO

- [ ] Unit, Integration Tests


nautobot/core/templatetags/helpers.py
```diff
@@ -616,6 +616,48 @@ def render_uptime(seconds):
     )
 
 
+@library.filter()
+@register.filter()
+def hyperlinked_field(value, hyperlink):
+    """Render a value as a hyperlink."""
+    if not value:
+        return placeholder(value)
+    return format_html('<a href="{}">{}</a>', value, hyperlink)
```
glennmatthews, line 625: Might make sense for `hyperlink` to be optional and just default to using `value` as both the text and the hyperlink?

nautobot/core/templatetags/helpers.py
```diff
@@ -616,6 +616,48 @@ def render_uptime(seconds):
...
+
+
+@library.filter()
+@register.filter()
+def render_m2m(value):
```
glennmatthews, line 630: Is this specific to content-types or is it actually applicable to M2M in general?

nautobot/core/templatetags/helpers.py
```diff
@@ -616,6 +616,48 @@ def render_uptime(seconds):
...
+    """Render sorted by model and app_label ManyToManyField value"""
+    if value.exists() is False:
+        return HTML_NONE
+    sorted_content_types = sorted(value, key=lambda ct: (ct.model, ct.app_label))
+
+    output = "<ul>"
+    for content_type in sorted_content_types:
+        output += f"<li>{content_type}</li>"
+    output += "</ul>"
```
glennmatthews, line 639: For content-types specifically it might make sense to render *all* objects, but for M2M in general we'll want to impose some limit.

glennmatthews, line 639: This needs to use `format_html`.

nautobot/core/templatetags/helpers.py
```diff
@@ -616,6 +616,48 @@ def render_uptime(seconds):
...
+    append_to_result = ""
+    for ancestor in value.ancestors():
+        nestable_tag = '<span title="nestable">↺</span>' if ancestor.nestable else ""
+        result += f'<li>{hyperlinked_object(ancestor, "name")} {nestable_tag}<ul>'
+        append_to_result += "</ul></li>"
+    nestable_tag = '<span title="nestable">↺</span>' if value.nestable else ""
+    result += f"<li><strong>{value}{nestable_tag}</strong></li>"
+    result += f"{append_to_result}</ul>"
+
+    return format_html(result)
```
glennmatthews, line 658: Need to use `format_html` on each individual step above - using it at the end here on the entire string is effectively the same as `mark_safe()` would be, and opens us up to potential XSS.

nautobot/core/templatetags/helpers.py
```diff
@@ -616,6 +616,48 @@ def render_uptime(seconds):
...
+
+
+@library.filter()
+@register.filter()
+def render_tree_queryset(value):
+    """Renders a tree queryset as an HTML nested list."""
+    result = '<ul class="tree-hierarchy">'
+    append_to_result = ""
+    for ancestor in value.ancestors():
+        nestable_tag = '<span title="nestable">↺</span>' if ancestor.nestable else ""
```
glennmatthews, line 651: Most `TreeModel` do not have a `nestable` field (only `LocationType` does) so this will need to handle that.

nautobot/core/ui/object_detail.py
```diff
@@ -732,6 +738,26 @@ def render_label(self, context):
...
+            return render_tree_queryset(value)
+
+        if isinstance(field_instance, URLField):
+            return hyperlinked_field(value, value)
+
+        if isinstance(field_instance, BooleanField):
+            return render_boolean(value)
+
+        if isinstance(field_instance, ManyToManyField):
+            return render_m2m(value.all())
```
glennmatthews, line 757: In *general* I don't think we should render M2M in `ObjectFieldsPanel` - that's what `ObjectTablePanel` is for. Content types *might* be an exception to that rule.

nautobot/core/ui/object_detail.py
```diff
@@ -754,13 +780,18 @@ def get_data(self, context):
...
                     continue
                 fields.append(field.name)
             # TODO: apply a default ordering "smarter" than declaration order? Alphabetical? By field type?
             # TODO: allow model to specify an alternative field ordering?
 
         data = {}
+
+        if isinstance(instance, TreeModel):
+            # using `_hierarchy` with the prepended `_` to try to archive a unique name, in cases where a model might have hierarchy field.
+            data["_hierarchy"] = instance
```
glennmatthews, line 793: Why does this need to be done here instead of handled in `render_value`?

nautobot/dcim/views.py
```diff
@@ -199,15 +201,42 @@ def get_selected_objects_parents_name(self, selected_objects):
...
-class LocationTypeView(generic.ObjectView):
+class LocationTypeUIViewSet(
+    view_mixins.ObjectDetailViewMixin,
+    view_mixins.ObjectListViewMixin,
+    view_mixins.ObjectEditViewMixin,
+    view_mixins.ObjectDestroyViewMixin,
+    view_mixins.ObjectBulkDestroyViewMixin,
+    view_mixins.ObjectBulkCreateViewMixin,  # 3.0 TODO: remove this mixin as it's no longer used
+    view_mixins.ObjectChangeLogViewMixin,
+    view_mixins.ObjectNotesViewMixin,
```
glennmatthews, line 212: Can we not just use `NautobotUIViewSet`? Why all the individual mixins?

nautobot/dcim/views.py
```diff
@@ -199,15 +201,42 @@ def get_selected_objects_parents_name(self, selected_objects):
...
+                fields="__all__",
+            ),
+            ObjectsTablePanel(
+                weight=100,
+                table_key="children_table",
+                table_title="Child Location Type(s)",
+            ),
+            ObjectsTablePanel(
+                weight=200,
+                table_key="locations_table",
```
glennmatthews, line 235: Presumably `get_extra_context` no longer needs to paginate `children_table` and `locations_table` since that's now handled by ObjectsTablePanel itself?

nautobot/dcim/views.py
```diff
@@ -199,15 +201,42 @@ def get_selected_objects_parents_name(self, selected_objects):
...
+                table_title="Location(s) of this Type",
```
glennmatthews, line 236: Need to specify `exclude_columns=["location_type"]` or not?

nautobot/project-static/css/base.css
```diff
@@ -347,6 +347,15 @@ ul.software-image-hierarchy {
...
 }
 
+ul.tree-hierarchy {
+    list-style-type: none;
+    padding-left: 0px;
+}
+ul.tree-hierarchy ul {
+    list-style-type: "↳ ";
+    padding-left: 18px;
+}
```
glennmatthews, line 357: Is this coming from somewhere else? Can it be removed from there?

nautobot/dcim/views.py
timizuoebideri1: Because of `BulkEdit`, which `LocationType` do not support

nautobot/core/ui/object_detail.py
timizuoebideri1: Because it needs to be part of the `data`, if not it would not be seen by `render_value` because it is not a model field

nautobot/core/ui/object_detail.py
timizuoebideri1: Oooh I see, then illd limit it to only ContentType

nautobot/core/templatetags/helpers.py
timizuoebideri1: ContentTypes

nautobot/project-static/css/base.css
timizuoebideri1: Not really, as we currently use `location-hierarchy` and `software-image-hierarchy`. this is a more generic `hierarchy` class

nautobot/dcim/views.py
glennmatthews: Let's get an issue opened for adding LocationType bulk-edit support for its `content-types` at least.

nautobot/core/templatetags/helpers.py
```diff
@@ -616,6 +616,51 @@ def render_uptime(seconds):
...
 
 
+@library.filter()
+@register.filter()
+def hyperlinked_field(value, hyperlink=None):
+    """Render a value as a hyperlink."""
+    if not value:
+        return placeholder(value)
+    hyperlink = hyperlink or value
+    return format_html('<a href="{}">{}</a>', value, hyperlink)
```
glennmatthews, line 626:
```suggestion
    return format_html('<a href="{}">{}</a>', hyperlink, value)
```

nautobot/core/templatetags/helpers.py
```diff
@@ -616,6 +616,51 @@ def render_uptime(seconds):
...
+        nestable_tag = format_html('<span title="nestable">↺</span>' if getattr(ancestor, "nestable", False) else "")
+        result += format_html(
+            "<li>{value} {nestable_tag}<ul>", value=hyperlinked_object(ancestor, "name"), nestable_tag=nestable_tag
+        )
+        append_to_result += format_html("</ul></li>")
+    nestable_tag = format_html('<span title="nestable">↺</span>' if getattr(ancestor, "nestable", False) else "")
+    result += format_html("<li><strong>{value}{nestable_tag}</strong></li>", value=value, nestable_tag=nestable_tag)
+    result += format_html("{value}</ul>", value=append_to_result)
+
+    return format_html(result)
```
glennmatthews, line 661:
```suggestion
    result = format_html('<ul class="tree-hierarchy">')
    append_to_result = format_html("</ul>")
    for ancestor in value.ancestors():
        nestable_tag = format_html('<span title="nestable">↺</span>' if getattr(ancestor, "nestable", False) else "")
        result += format_html(
            "<li>{value} {nestable_tag}<ul>", value=hyperlinked_object(ancestor, "name"), nestable_tag=nestable_tag
        )
        append_to_result += format_html("</ul></li>")
    nestable_tag = format_html('<span title="nestable">↺</span>' if getattr(value, "nestable", False) else "")
    result += format_html("<li><strong>{value} {nestable_tag}</strong></li>", value=value, nestable_tag=nestable_tag)
    result += append_to_result

    return result
```

nautobot/core/ui/object_detail.py
```diff
@@ -685,10 +692,23 @@ def render_body_content(self, context):
...
-                # TODO: add a copy button on hover to all display items
-                result += format_html("<tr><td>{key}</td><td>{value}</td></tr>", key=key_display, value=value_display)
+            if value_display := self.render_value(key, value, context):
+                if value in [None, ""]:
+                    value_tag = value_display
+                else:
+                    value_tag = format_html(
+                        """
+                            <span class="hover_copy">
+                                <span id="{key}_copy">{value}</span>
```
glennmatthews, line 702: Possibly something like `value_{key}` might be clearer as an `id` string?

But also need to consider that keys may not be globally unique in the page, for example the "termination A" and "termination Z" panels in the Circuit detail view. Might need to *also* include the panel's `id` as a part of this string?

Do we need to furthermore `slugify` the `key` or anything like that to have correct handling of whitespace in keys, or is that handled just fine as-is?

nautobot/core/ui/object_detail.py
```diff
@@ -732,6 +752,26 @@ def render_label(self, context):
...
+            field_instance = None
+
+        if key == "_hierarchy":
+            return render_tree_queryset(value)
+
+        if isinstance(field_instance, URLField):
+            return hyperlinked_field(value)
+
+        if isinstance(field_instance, BooleanField):
+            return render_boolean(value)
```
glennmatthews, line 768: Might make sense to move this to `KeyValueTablePanel` and just do a `if isinstance(value, bool)` instead?

nautobot/core/ui/object_detail.py
```diff
@@ -732,6 +752,26 @@ def render_label(self, context):
...
+
+        if isinstance(field_instance, ManyToManyField) and field_instance.related_model == ContentType:
+            return render_content_types(value)
```
glennmatthews, line 771: Do we need an `else` here for the case of M2M that *aren't* ContentTypes?

nautobot/dcim/views.py
timizuoebideri1: https://github.com/nautobot/nautobot/issues/6347

nautobot/core/ui/object_detail.py
glennmatthews: I think there was at least one case in the circuits view where I was passing a display label as the `key` because the automatic rendering didn't handle capitalization correctly. Not sure this is a safe assumption to make.

nautobot/core/ui/object_detail.py
glennmatthews: No - M2M fields *in general* shouldn't be rendered in an ObjectFieldsPanel because there may be arbitrarily many related objects. Content-type is at least bounded to a maximum of low three-digits. M2M should be rendered as an ObjectsTablePanel instead so that we get pagination and so forth.

nautobot/circuits/views.py
```diff
@@ -111,15 +111,8 @@ class ProviderUIViewSet(NautobotUIViewSet):
                 fields="__all__",
                 exclude_fields=["comments"],
```
glennmatthews, line 112: This can be removed too now right?

nautobot/core/templatetags/helpers.py
```diff
@@ -616,6 +616,51 @@ def render_uptime(seconds):
...
+    return format_html('<a href="{}">{}</a>', hyperlink, value)
+
+
+@library.filter()
+@register.filter()
+def render_content_types(value):
+    """Render sorted by model and app_label ContentTypes value"""
+    if value.exists() is False:
+        return HTML_NONE
+    sorted_content_types = sorted(value.all(), key=lambda ct: (ct.model, ct.app_label))
```
glennmatthews, line 635: Huh. Is it intentional to sort by model first rather than by app label? That seems counterintuitive to me.

Also, would be more efficient I think to do `value.order_by()` instead of sorting the queryset in code?

nautobot/core/templatetags/helpers.py
```diff
@@ -616,6 +616,51 @@ def render_uptime(seconds):
...
+
+    output = format_html("<ul>")
+    for content_type in sorted_content_types:
+        output += format_html("<li>{content_type}</li>", content_type=content_type)
+    output += format_html("</ul>")
+
+    return format_html(output)
```
glennmatthews, line 642:
```suggestion
    return output
```

nautobot/core/templatetags/helpers.py
```diff
@@ -616,6 +616,51 @@ def render_uptime(seconds):
...
+        return placeholder(value)
+    hyperlink = hyperlink or value
+    return format_html('<a href="{}">{}</a>', hyperlink, value)
+
+
+@library.filter()
+@register.filter()
+def render_content_types(value):
+    """Render sorted by model and app_label ContentTypes value"""
+    if value.exists() is False:
```
glennmatthews, line 633:
```suggestion
    if not value.exists():
```

nautobot/core/ui/object_detail.py
```diff
@@ -683,12 +695,29 @@ def render_body_content(self, context):
...
         result = format_html("")
+        panel_label = slugify(self.label or "")
         for key, value in data.items():
             key_display = self.render_key(key, value, context)
-            value_display = self.render_value(key, value, context)
-            if value_display:
-                # TODO: add a copy button on hover to all display items
-                result += format_html("<tr><td>{key}</td><td>{value}</td></tr>", key=key_display, value=value_display)
+            if value_display := self.render_value(key, value, context):
+                if value in [None, ""]:
```
glennmatthews, line 702: What about the case where `value` is an empty queryset?

Maybe it might be simpler/cleaner to compare `value_display` against the `HTML_NONE` constant instead?

nautobot/core/ui/object_detail.py
```diff
@@ -1053,6 +1104,33 @@ def get_extra_context(self, context):
         }
 
 
+class _ObjectCommentPanel(ObjectFieldsPanel):
```
glennmatthews, line 1107: This should be changed to a TextPanel (or MarkdownPanel, or equivalent) subclass as part of @mateusz-lachowski-codilime work on #6139.

nautobot/core/ui/object_detail.py
```diff
@@ -1053,6 +1104,33 @@ def get_extra_context(self, context):
...
+    """Panel displaying an object's comments as a space-separated panel."""
+
+    def __init__(
+        self,
+        *,
+        label="Comments",
+        section=SectionChoices.LEFT_HALF,
+        weight=Panel.WEIGHT_COMMENTS_PANEL,
+        value_transforms={"comments": [render_markdown, placeholder]},
```
glennmatthews, line 1116: pylint is rightly complaining about using a dict as a default arg value. This needs to be `None` and set in the function body if otherwise unspecified.

nautobot/dcim/views.py
```diff
@@ -235,15 +257,6 @@ def get_extra_context(self, request, instance):
...
-class LocationTypeEditView(generic.ObjectEditView):
-    queryset = LocationType.objects.all()
-    model_form = forms.LocationTypeForm
-
-
-class LocationTypeDeleteView(generic.ObjectDeleteView):
-    queryset = LocationType.objects.all()
-
-
 class LocationTypeBulkImportView(generic.BulkImportView):  # 3.0 TODO: remove, unused
```
glennmatthews, line 260: Shouldn't `LocationTypeBulkImportView` and `LocationTypeBulkDeleteView` be removed as well?

nautobot/core/ui/object_detail.py
timizuoebideri1: I think I should open an issue for auto M2M fields as `ObjectFieldsPanel`. Alternatively, perhaps we should display M2M fields only in cases where the `ObjectFieldsPanel` is supplied; in the event that it is not, the M2M field is ignored.

nautobot/core/ui/object_detail.py
timizuoebideri1: Maybe if this PR gets merge in before https://github.com/nautobot/nautobot/issues/6139. is complete then we can add it as part of its acceptance criteria

nautobot/dcim/urls.py
```diff
@@ -46,28 +45,9 @@
 router.register("module-bay-templates", views.ModuleBayTemplateUIViewSet)
 router.register("module-types", views.ModuleTypeUIViewSet)
 router.register("virtual-device-contexts", views.VirtualDeviceContextUIViewSet)
+router.register("location-types", views.LocationTypeUIViewSet)
```
glennmatthews, line 48: Can we alphabetize this list? I don't see any logical order to the current code.

glennmatthews (approved):

PR #6333, opened 2024-10-07
Fix DynamicGroup.group_type automatic setting

### LINKED ISSUE #6329: Issue with Dynamic Groups when upgrading from Nautobot v2.2 to v2.3
### Environment
* Nautobot version (Docker tag too if applicable): 2.3.4
* Python version: 3.11
* Database platform, version: Postgres 13
* Middleware(s): N/A

### Steps to Reproduce
1. While still in Nautobot v2.2, create the following Devices.

    - DeviceA with Role `router` and Status `Active`
    - DeviceB with Role `router` and Status `Active`
    - DeviceC with Role `switch` and Status `Active`

2. While still in Nautobot v2.2, create the following Dynamic Groups:

    - `Active All` with filter `{status: 'Active'}` >>> Resulting members: DeviceA, DeviceB, DeviceC
    - `Active Routers` with filter `{status: 'Active', role: 'router'}`  >>> Resulting members: DeviceA, DeviceB
    - `Active Non-Routers` with children `INCLUDE Active All` & `EXCLUDE Active Routers` >>> Resulting members: DeviceC

3. Upgrade to Nautobot v2.3.

### Expected Behavior
- Dynamic Group `Active All` gets `group_type: filter_defined`
- Dynamic Group `Active Routers` gets `group_type: filter_defined`
- Dynamic Group `Active Non-Routers` gets `group_type: group_of_groups`

### Observed Behavior
After the upgrade the Dynamic Groups are assigned group type values which are the inverse of the expected ones.
- Dynamic Group `Active All` gets `group_type: group_of_groups`
- Dynamic Group `Active Routers` gets `group_type: filter_defined`
- Dynamic Group `Active Non-Routers` gets `group_type: filter_defined`

glennmatthews:
# Closes #6329
# What's Changed

I still don't understand the cause, but confirmed that, as reported in #6329, extras migration 0112 is updating a queryset different than what it should - while it's calling `DynamicGroup.objects.filter(children__isnull=False)`, it's actually getting the set of DynamicGroups whose `parents` are non-null, resulting in some groups being incorrectly set to TYPE_DYNAMIC_SET that should remain TYPE_DYNAMIC_FILTER, and some remaining as TYPE_DYNAMIC_FILTER that should be set to TYPE_DYNAMIC_SET.

(FWIW, I verified that this query gives the expected results in `nbshell` - it seems to only be in the migration context that it behaves incorrectly. My best *guess* at the moment is that this might be because the migration doesn't explicitly specify the `through_fields` on the `ManyToManyField`, while the actual final model does do so. But that's only a guess.)

Because I don't fully understand the root cause, and because the above migration has been in the wild for several releases already, rather than try and fix that migration, I've instead added a new migration with new logic to fix up the incorrect data.

```
  Applying extras.0116_fix_dynamic_group_group_type_data_migration...
    Found and fixed 1 DynamicGroup(s) that should be typed as "Group of groups".

    Found and fixed 2 DynamicGroup(s) that should be typed as "Filter-defined".
```

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/utils.py
+++ b/nautobot/extras/utils.py
@@ -597,7 +597,7 @@ def fixup_null_statuses(*, model, model_contenttype, status_model):
         print(f"    Found and fixed {updated_count} instances of {model.__name__} that had null 'status' fields.")
 
 
-def fixup_dynamic_group_group_types(apps, *args, **kwargs):
+def fixup_dynamic_group_group_types(apps, *args, **kwargs):  # pylint: disable=redefined-outer-name
     """Set dynamic group group_type values correctly."""
     DynamicGroup = apps.get_model("extras", "DynamicGroup")
     DynamicGroupMembership = apps.get_model("extras", "DynamicGroupMembership")
--- a/nautobot/extras/utils.py
+++ b/nautobot/extras/utils.py
@@ -21,7 +21,7 @@ from nautobot.core.choices import ColorChoices
 from nautobot.core.constants import CHARFIELD_MAX_LENGTH
 from nautobot.core.models.managers import TagsManager
 from nautobot.core.models.utils import find_models_with_matching_fields
-from nautobot.extras.choices import ObjectChangeActionChoices
+from nautobot.extras.choices import DynamicGroupTypeChoices, ObjectChangeActionChoices
 from nautobot.extras.constants import (
     CHANGELOG_MAX_CHANGE_CONTEXT_DETAIL,
     EXTRAS_FEATURES,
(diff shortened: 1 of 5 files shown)


itdependsnetworks:
I understand that the migration is in the wild already, but in theory an ounce of prevention is worth a pound of cure, meaning those that get the updated migration would be better off to not have the issue at all.

glennmatthews:
> I understand that the migration is in the wild already, but in theory an ounce of prevention is worth a pound of cure, meaning those that get the updated migration would be better off to not have the issue at all.

It's a bit of a moot point I think, because if we update the existing migration, any deployment that would get the updates to that migration would *also* get the newly added migration in any case.

nautobot/extras/utils.py
```diff
@@ -597,6 +597,32 @@ def fixup_null_statuses(*, model, model_contenttype, status_model):
...
+    parent_groups_with_wrong_type = DynamicGroup.objects.filter(name__in=parent_group_names).exclude(
+        group_type=DynamicGroupTypeChoices.TYPE_DYNAMIC_SET
+    )
+    if parent_groups_with_wrong_type.exists():
+        count_1 = parent_groups_with_wrong_type.update(group_type=DynamicGroupTypeChoices.TYPE_DYNAMIC_SET)
+        print(f'\n    Found and fixed {count_1} DynamicGroup(s) that should be typed as "Group of groups".')
+
+    filter_groups_with_wrong_type = DynamicGroup.objects.exclude(filter__exact={}).exclude(
+        group_type=DynamicGroupTypeChoices.TYPE_DYNAMIC_FILTER
+    )
```
HanlinMiao, line 618: Is this query missing the exclusion of existing statically assigned groups as well or is `.exclude(filter__exact={})` already doing that?

itdependsnetworks:
> > I understand that the migration is in the wild already, but in theory an ounce of prevention is worth a pound of cure, meaning those that get the updated migration would be better off to not have the issue at all.
> 
> It's a bit of a moot point I think, because if we update the existing migration, any deployment that would get the updates to that migration would _also_ get the newly added migration in any case.

Perhaps I am taking this too literal `I've instead added a new migration with new logic to fix up the incorrect data`, but if that is accurate, it would seem that fixing the data would only be done as a backup to making the migration correct the first time.

lampwins:
While I can’t point to any documentation, Glenn’s approach is usually the recommended one in Django, because it ensures all installs (both old and new) take a single code path to reach the end state.

nautobot/extras/utils.py
glennmatthews: Thanks for the sanity check! Static groups should always have an empty filter so should be excluded.

HanlinMiao (approved):

PR #6340, opened 2024-10-08
allow id to be provided at object creation time

### LINKED ISSUE #6166: support idempotent / upsert objects via the API
### As ...

Tammy - Tools Integrator

### I want ...

To develop integrations with Nautobot which need to create objects (like devices, prefixes, etc) in an idempotent way

### So that ...

So that I know the object I am creating beforehand

I can create objects idempotently or via an upsert (update or insert).

So via REST it is common to have `POST /myobject/` create a new record and return a reference to that record via a 201 CREATED and Nautobot does this today. It's also common to support `PATCH /myobject/$REF` and include a subset or even all the fields to update. Which again is something that Nautobot supports. The last pattern is the idempotent create or upsert which is commonly `PUT /myobject/$REF` which requires the entire object and will either create it or update it. Nautobot returns a 404 if it doesn't exist and updates it if it does. This is something the Django REST framework is doing. Searching shows its a common-ish compliant and worked around via the serializer.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation
- [X] Other (not directly a platform feature)

### Database Changes

No

### External Dependencies

None.

cardoe:
# Closes #6166 
# What's Changed
When creating an object via the API, allow for the id to be defined to allow for idempotent object creation/updating, which is commonly referred to as upsert. This allows for bulk upload of records where some entries might already exist without creating duplicate entries.

# Screenshots

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/changes/6166.changed
+++ b/changes/6166.changed
@@ -1,0 +1,1 @@
+Enhanced the REST API to generally make it possible to create objects with known ids on request.
--- a/nautobot/core/api/serializers.py
+++ b/nautobot/core/api/serializers.py
@@ -134,6 +134,7 @@ class BaseModelSerializer(OptInFieldsMixin, serializers.HyperlinkedModelSerializ
 
     serializer_related_field = NautobotHyperlinkedRelatedField
 
+    id = serializers.UUIDField(read_only=False, default=serializers.CreateOnlyDefault(uuid.uuid4))
     display = serializers.SerializerMethodField(read_only=True, help_text="Human friendly display value")
     object_type = ObjectTypeField()
     # composite_key = serializers.SerializerMethodField()  # TODO: Revisit if we reintroduce composite keys
--- a/nautobot/core/testing/api.py
+++ b/nautobot/core/testing/api.py
@@ -707,7 +707,11 @@ class APIViewTestCases:
             self.assertHttpStatus(response, status.HTTP_201_CREATED, csv_data)
             # Note that create via CSV is always treated as a bulk-create, and so the response is always a list of dicts
             new_instance = self._get_queryset().get(pk=response.data[0]["id"])
-            self.assertNotEqual(new_instance.pk, orig_pk)
+            if isinstance(orig_pk, int):
+                self.assertNotEqual(new_instance.pk, orig_pk)
+            else:
+                # for our non-integer PKs, we're expecting the creation to respect the requested PK
+                self.assertEqual(new_instance.pk, orig_pk)
 
             new_serializer = serializer_class(new_instance, context={"request": None})
             new_data = new_serializer.data
--- a/nautobot/core/tests/test_api.py
+++ b/nautobot/core/tests/test_api.py
@@ -3,6 +3,7 @@ from io import BytesIO, StringIO
 import json
 import os
 from unittest import skip
+import uuid
 
 from constance import config
 from constance.test import override_config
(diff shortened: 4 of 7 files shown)


cardoe:
Rebased onto #6455 but I'm at loss here how to make the CSV path operate as an update in addition to importing new entries. So I'm at a loss and looking for help here. @glennmatthews do you have any suggestions?

nautobot/core/api/serializers.py
```diff
@@ -134,6 +134,7 @@ class BaseModelSerializer(OptInFieldsMixin, serializers.HyperlinkedModelSerializ
 
     serializer_related_field = NautobotHyperlinkedRelatedField
 
+    id = serializers.UUIDField(read_only=False, default=serializers.CreateOnlyDefault(uuid.uuid4))
```
glennmatthews, line 137: I believe the test failure is because Django's built-in Group model has an integer PK instead of a UUID PK. Might just need to override this field on `GroupSerializer` (and probably also `ContentTypeSerializer` for similar reasons).

glennmatthews (commented):
Sorry for the delay on this!

I think supporting user-specified PK values in REST is a useful feature step on its own even without yet addressing the CSV update goal. Do you think it'd be possible to get the current CSV test back to passing so that we could get this merged more-or-less as-is? We could then look at what's needed to get the API serializers to do a `create_or_update` in place of a `create` as a next step.

nautobot/core/api/serializers.py
cardoe: We have success! Thanks for the help here @glennmatthews.

changes/6166.changed
```diff
@@ -0,0 +1 @@
+Made it possible to create objects with known ids.
```
glennmatthews, line 1: Perhaps:

```suggestion
Enhanced the REST API to generally make it possible to create objects with known ids on request.
```

nautobot/core/tests/test_csv.py
```diff
@@ -269,27 +269,29 @@ def test_round_trip_export_import(self):
         self.assertEqual(response.status_code, 200)
         # uploading the CSV always returns a 200 code with a page with an error message on it
         # ensure we don't have that error message
-        self.assertNotIn("FORM-ERROR", response.content.decode(response.charset))
-        self.assertEqual(Device.objects.count(), 4)
+        # nautobot/nautobot#1543 - assert entry did not get updated
+        self.assertIn("FORM-ERROR", response.content.decode(response.charset))
+        # since the same IDs are used the existing records were updated instead of created new
```
glennmatthews, line 274: Taking my first look at this in a while, I wonder if it'd be more in the spirit of what this test was originally checking for if we instead (up at line 265) were to replace the `id` columns in the data with new values as well and keep the below steps the same as they were? What do you think?

cardoe: Yeah I think you're right.

cardoe:
So locally I've run `invoke start` and then connected to the Nautobot and went to the API UI and tried to create a manufacturer:

```json
{
  "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "name": "manufac",
  "description": "describe"
}
```

Initially I got a 201 and it created this. Running this again gives:

image

But then changing the name results in:

image

But if I delete it and reapply the original once again I get:

image

cardoe:
The first push after my screenshots is the code I was running. The 2nd push is just a rebase to the latest develop branch.

glennmatthews:
Based on the screenshots, I'd say that's more evidence that we will want the longer-term feature as mentioned above `to get the API serializers to do a create_or_update in place of a create as a next step`.

nautobot/core/testing/api.py
```diff
@@ -707,7 +707,11 @@ def test_recreate_object_csv(self):
             self.assertHttpStatus(response, status.HTTP_201_CREATED, csv_data)
             # Note that create via CSV is always treated as a bulk-create, and so the response is always a list of dicts
             new_instance = self._get_queryset().get(pk=response.data[0]["id"])
-            self.assertNotEqual(new_instance.pk, orig_pk)
+            if isinstance(orig_pk, int):
+                self.assertNotEqual(new_instance.pk, orig_pk)
+            else:
+                # for our non-integer PKs, we're expecting the creation to be idempotent
```
glennmatthews, line 713: Perhaps:

```suggestion
                # for our non-integer PKs, we're expecting the creation to respect the requested PK
```

nautobot/core/tests/test_api.py
```diff
@@ -754,6 +755,25 @@ def test_related_by_invalid(self):
         self.assertHttpStatus(response, status.HTTP_400_BAD_REQUEST)
         self.assertEqual(ipam_models.VLAN.objects.filter(name="Test VLAN 100").count(), 0)
 
+    def test_idempotent_id_create(self):
```
glennmatthews, line 758: Perhaps:

```suggestion
    def test_create_with_specified_id(self):
```

nautobot/core/tests/test_csv.py
```diff
@@ -261,15 +264,35 @@ def test_round_trip_export_import(self):
...
+                filtered_row = {key: row[key] for key in fieldnames}
+                writer.writerow(filtered_row)
+            data = {"csv_data": import_csv.getvalue()}
         url = reverse("dcim:device_import")
         response = self.client.post(url, data)
         self.assertEqual(response.status_code, 200)
         # uploading the CSV always returns a 200 code with a page with an error message on it
         # ensure we don't have that error message
         self.assertNotIn("FORM-ERROR", response.content.decode(response.charset))
+        # since the same IDs are used the existing records were updated instead of created new
```
glennmatthews, line 295: remove this comment for now?

glennmatthews (approved):
Nearly there! Thanks for your persistence.

PR #6345, opened 2024-10-08
Added `StatsPanel` UI component Python class.

### LINKED ISSUE #6140: UI Components: Stats panel
Stats panel (e.g. Location detail view)

User can:
- Pass a list of model fields (accessors at declaration, or data into the render context), which:
  - Attempt to link a related object list view or fall back to just showing the count
- Pass a dict of numbers, optional links, and labels/tooltips

HanlinMiao:
# Closes #6140 
# What's Changed

Added `StatsPanel` UI component Python class.
Refactored Location detail view and tenant detail view to use UI component Python class `StatsPanel`.

# Screenshots

Screenshot 2024-10-08 at 3 58 39 PM
Screenshot 2024-10-08 at 3 58 53 PM

# TODO

- [ ] Unit, Integration Tests


nautobot/core/templates/components/panel/stats_panel_body.html
```diff
@@ -0,0 +1,8 @@
+<div class="panel-body">
+    {% for key, value in stats.items %}
```
glennmatthews, line 2: is `key` unused here? Could maybe do `for url, value, title in stats.values` instead?

nautobot/core/ui/object_detail.py
```diff
@@ -729,6 +729,66 @@ def render_body_content(self, context):
         return result
 
 
+class StatsPanel(Panel):
+    def __init__(
+        self,
+        *,
+        stats=None,
+        filter_name=None,
```
glennmatthews, line 737: From the template above, it seems like `filter_name` should be a required argument?

nautobot/core/ui/object_detail.py
```diff
@@ -729,6 +729,66 @@ def render_body_content(self, context):
...
+        if self.body_content_template_path:
+            stats = {}
+            for related_object_model_class, related_object_count in self.stats.items():
+                related_object_model_class_meta = related_object_model_class._meta
+                app_label, model_name = (
+                    related_object_model_class_meta.app_label,
+                    related_object_model_class_meta.model_name,
+                )
+
+                related_object_list_url = f"{app_label}:{model_name}_list"
```
glennmatthews, line 776: maybe use `get_route_for_model(related_object_model_class, "list")` instead?

nautobot/dcim/views.py
```diff
@@ -281,24 +283,23 @@ def get_extra_context(self, request, instance):
...
             .filter(circuit_terminations__location__in=related_locations)
             .count(),
-            "vm_count": VirtualMachine.objects.restrict(request.user, "view")
+            VirtualMachine: VirtualMachine.objects.restrict(request.user, "view")
             .filter(cluster__location__in=related_locations)
             .count(),
         }
+        stats_panel = StatsPanel(
+            label="Stats", section=SectionChoices.RIGHT_HALF, weight=100, filter_name="location", stats=stats
+        )
```
glennmatthews, line 302: Let's see if we can come up with a pattern for including the panel definition in the view's `object_detail_content` alongside the view's other panels, rather than generating it on-the-fly in `get_extra_context()` and passing it as a context variable.

HanlinMiao:
With the latest commit, the LocationDetailView looks like this.
Screenshot 2024-10-09 at 3 32 22 PM

Still need the image and rack_group table panels and the rendering of hierarchy field is proven to be difficult.

- [ ] ImageTablePanel
- [ ] RackGroupPanel
- [ ] Hierarchy table cell (include tag?)

glennmatthews:
Isn't the pattern for rendering of the hierarchy implemented by #6328?

nautobot/core/templates/components/panel/contact_panel_footer.html
glennmatthews, line 1: Since this isn't a generic component, but specific to the Location view, *maybe* it should be located in a new `dcim/templates/dcim/components/panel/` directory and named something like `location_contact_panel_footer.html`? What do you think?

nautobot/core/templatetags/helpers.py
```diff
@@ -368,6 +368,36 @@ def tzoffset(value):
...
+    Render the count of the location's child locations
+    """
+    if not value.all().exists():
+        return HTML_NONE
+
+    parent = value.first().parent
+    child_count = value.filter(parent=parent).count()
+    url = reverse("dcim:location_list")
+    html_template = '<a href="{}?parent={}">{}</a>'
+    return format_html(html_template, url, parent.pk, child_count)
```
glennmatthews, line 398: Honestly not sure this is needed now - I recognize it was there in the previous implementation but now we have this same count in the related-object table for children, right?

nautobot/core/ui/object_detail.py
```diff
@@ -876,9 +881,24 @@ def render_body_content(self, context):
             ...
         }
         """
+        instance = context["obj"]
+        request = context["request"]
+        if isinstance(instance, Location):
```
glennmatthews, line 939: Maybe `TreeModel` in general?

nautobot/dcim/views.py
```diff
@@ -278,28 +287,99 @@ class LocationView(generic.ObjectView):
...
+                value_transforms={
+                    "physical_address": [linebreaksbr, placeholder],
+                    "shipping_address": [linebreaksbr, placeholder],
+                },
+            ),
+            ContactPanel(
+                label="Contact Info",
+                section=SectionChoices.LEFT_HALF,
+                weight=120,
+                footer_content_template_path="components/panel/contact_panel_footer.html",
```
glennmatthews, line 346: This probably makes sense to embed in the `ContactPanel.__init__` definition?

nautobot/dcim/views.py
```diff
@@ -278,28 +287,99 @@ class LocationView(generic.ObjectView):
...
+            ),
+            ObjectFieldsPanel(
+                label="Comments",
+                weight=130,
+                section=SectionChoices.LEFT_HALF,
+                fields=["comments"],
+                value_transforms={"comments": [render_markdown, placeholder]},
+            ),
```
glennmatthews, line 354: Shouldn't be needed after #6328

nautobot/dcim/views.py
```diff
@@ -278,28 +287,99 @@ class LocationView(generic.ObjectView):
...
+                    Rack: "location__in",
+                    Device: "location__in",
+                    Prefix: "location__in",
+                    VLAN: "location__in",
+                    Circuit: "circuit_terminations__location__in",
+                    VirtualMachine: "cluster__location__in",
+                },
+                section=SectionChoices.RIGHT_HALF,
+                weight=100,
+            ),
```
glennmatthews, line 369: Much nicer!

nautobot/dcim/views.py
```diff
@@ -278,28 +287,99 @@ class LocationView(generic.ObjectView):
...
+            ObjectsTablePanel(
+                label="Child",
+                weight=300,
```
glennmatthews, line 372: why 300?

nautobot/core/ui/object_detail.py
```diff
@@ -452,7 +452,7 @@ def _get_table_add_url(self, context):
         This method determines the URL for adding a new object to the table. It checks if the user has
         the necessary permissions and creates the appropriate URL based on the specified add button route.
         """
-        obj = context["obj"]
+        obj = context.get("obj") or context.get("object")
```
glennmatthews, line 455: Thought we always had `obj` as a context variable :(

nautobot/core/templates/components/panel/contact_panel_footer.html
HanlinMiao: Agreed

HanlinMiao:
Screenshot 2024-10-09 at 4 03 20 PM
I am seeing hierarchy rendered in all of my panels, am I using it wrong?

glennmatthews:
Looks like a bug/oversight to me - might need to add logic to the ObjectFieldsPanel to make `_hierarchy` an opt-in key instead of one that gets automatically added whenever the panel references a TreeModel?

nautobot/core/ui/object_detail.py
```diff
@@ -811,7 +809,7 @@ def get_data(self, context):
 
         data = {}
 
-        if isinstance(instance, TreeModel):
+        if isinstance(instance, TreeModel) and self.label is None:
```
glennmatthews, line 812: I get what this is doing but it's not super-obvious. Would it be better to do something like `and (self.fields == "__all__" or "_hierarchy" in self.fields)`?

nautobot/core/templatetags/helpers.py
```diff
@@ -368,6 +368,20 @@ def tzoffset(value):
...
+@library.filter()
+@register.filter()
+def format_time_zone(value):
+    """
+    Returns the timezone and the hour offset of the time zone using the current time.
+    """
+    timezone = value
+    offset = tzoffset(value)
+    # TODO: This should be but it is not working for some reason
+    # {% timezone "{}" %}{% now "SHORT_DATETIME_FORMAT" %}{% endtimezone %}
```
glennmatthews, line 380: Any guesses?

nautobot/core/ui/object_detail.py
```diff
@@ -894,6 +894,79 @@ def render_body_content(self, context):
...
+        related_field_names=None,
+        body_content_template_path="components/panel/stats_panel_body.html",
+        **kwargs,
+    ):
+        """Instantiate a `StatsPanel`.
+
+        Args:
+            related_field_names (dict): key/value pair of { <model_class>: <query_strings> }.
+            filter_pks (list of uuids): the list of pks work in tandem with the query_strings.
+            filter_name (str): a valid query filter append to the anchor tag for each stat button.
```
glennmatthews, line 912: Might be good to provide an example?

nautobot/core/ui/object_detail.py
```diff
@@ -894,6 +894,79 @@ def render_body_content(self, context):
...
+        """
+
+        self.filter_name = filter_name
+        self.related_field_names = related_field_names
+        self.filter_pks = filter_pks
+        self.body_content_template_path = body_content_template_path
```
glennmatthews, line 918: Redundant?

nautobot/core/ui/object_detail.py
```diff
@@ -894,6 +894,79 @@ def render_body_content(self, context):
...
+
+        Args:
+            related_field_names (dict): key/value pair of { <model_class>: <query_strings> }.
+            filter_pks (list of uuids): the list of pks work in tandem with the query_strings.
+            filter_name (str): a valid query filter append to the anchor tag for each stat button.
+        """
+
+        self.filter_name = filter_name
+        self.related_field_names = related_field_names
+        self.filter_pks = filter_pks
```
glennmatthews, line 917: Is this always overwritten by lines 937-942 below? Not sure I immediately see the application of providing this at declaration time.

nautobot/tenancy/views.py
```diff
@@ -78,29 +80,40 @@ class TenantListView(generic.ObjectListView):
...
+                    "tenant_group",
+                    "description",
+                ],
+            ),
+            StatsPanel(
+                label="Stats",
+                filter_name="tenant",
+                filter_pks=[],
+                related_field_names={
+                    Circuit: "tenant__in",
```
glennmatthews, line 98: Might be nice if we could do either or both of:

- automatically append "__in" to the filter string in all cases?
- use either the model or the `filter_name` to derive a default filter string?

Just trying to think if we could do something even simpler on the declaration side, something like `related_models = [Circuit, Cluster, Device, IPAddress, Location, ....]`

nautobot/core/ui/object_detail.py
```diff
@@ -894,6 +894,79 @@ def render_body_content(self, context):
...
+                related_object_list_url = get_route_for_model(related_object_model_class, "list")
+                related_object_title = bettertitle(related_object_model_class_meta.verbose_name_plural)
+                value = [related_object_list_url, related_object_count, related_object_title]
+                stats[related_object_model_class] = value
+
+            return get_template(self.body_content_template_path).render(
+                {
+                    **context,
+                    "stats": stats,
+                    # TODO validate the filter
```
glennmatthews, line 963: Any ideas what would be needed here?

nautobot/dcim/templates/dcim/inc/location_contact_panel_footer.html
glennmatthews, line 1: Revert/remove?

nautobot/tenancy/views.py
```diff
@@ -78,29 +80,40 @@ class TenantListView(generic.ObjectListView):
...
-        return {"stats": stats, **super().get_extra_context(request, instance)}
+    object_detail_content = ObjectDetailContent(
+        panels=(
+            ObjectFieldsPanel(
+                section=SectionChoices.LEFT_HALF,
+                weight=100,
+                fields=[
+                    "tenant_group",
+                    "description",
+                ],
```
glennmatthews, line 91: what would we gain/lose by using `__all__` here instead?

nautobot/dcim/templates/dcim/inc/location_contact_panel_footer.html
HanlinMiao: I think it does not hurt to keep this in here for when we eventually move on to the component-based views. What do you think?

nautobot/tenancy/views.py
HanlinMiao: I like the pattern you suggested too. But I think when it comes to more complex queries in LocationDetailView, this current pattern might be more desirable.
For example:
```
"circuit_count": Circuit.objects.restrict(request.user, "view")
.filter(circuit_terminations__location__in=related_locations)
.count(),
"vm_count": VirtualMachine.objects.restrict(request.user, "view")
.filter(cluster__location__in=related_locations)
.count(),
```

nautobot/core/templatetags/helpers.py
HanlinMiao: Should be fixed in the latest commit

nautobot/core/ui/object_detail.py
HanlinMiao: Yeah still a number of differences between the context variables in legacy view patterns and those in NautobotUIViewSet.

nautobot/core/templatetags/helpers.py
```diff
@@ -376,10 +376,9 @@ def format_time_zone(value):
     """
     timezone = value
     offset = tzoffset(value)
-    # TODO: This should be but it is not working for some reason
-    # {% timezone "{}" %}{% now "SHORT_DATETIME_FORMAT" %}{% endtimezone %}
+    current_time = datetime.datetime.now(timezone).strftime("%Y-%m-%d %I:%M %p")
```
glennmatthews, line 379: Downside of this approach is that it doesn't respect the configured `setttings.SHORT_DATETIME_FORMAT`, right?

nautobot/dcim/templates/dcim/inc/location_contact_panel_footer.html
glennmatthews: Who knows if the pattern will evolve between now and then - since it's presently unused, my vote would be to yank it out for now.

nautobot/tenancy/views.py
glennmatthews: Yeah, I think we'd need to support both - maybe something like `related_models = [Device, Prefix, (Circuit, "circuit_terminations__location"), (VirtualMachine, "cluster__location"), ...]` might work and still be reasonably comprehensible to other developers?

nautobot/core/ui/object_detail.py
glennmatthews: Let's create a `get_obj_from_context`  helper function that we can use throughout then.

nautobot/core/ui/object_detail.py
```diff
@@ -894,6 +894,82 @@ def render_body_content(self, context):
         return result
 
 
+class StatsPanel(Panel):
+    def __init__(
+        self,
+        *,
+        filter_name,
+        filter_pks=None,
```
glennmatthews, line 902:
```suggestion
```

nautobot/core/ui/object_detail.py
```diff
@@ -894,6 +894,82 @@ def render_body_content(self, context):
...
+        related_field_names=None,
+        body_content_template_path="components/panel/stats_panel_body.html",
+        **kwargs,
+    ):
+        """Instantiate a `StatsPanel`.
+
+        Args:
+            related_field_names (dict): key/value pair of { <model_class>: <query_strings> }.
+            filter_pks (list of uuids): the list of pks work in tandem with the query_strings.
```
glennmatthews, line 911:
```suggestion
```

nautobot/core/templatetags/helpers.py
glennmatthews: Could revert this for now, given it's specific to Location?

nautobot/tenancy/views.py
```diff
@@ -78,29 +80,36 @@ class TenantListView(generic.ObjectListView):
...
+        panels=(
+            ObjectFieldsPanel(
+                section=SectionChoices.LEFT_HALF,
+                weight=100,
+                fields="__all__",
+            ),
+            StatsPanel(
+                label="Stats",
+                filter_name="tenant",
+                related_field_names=[
```
glennmatthews, line 93: Rename to `related_models` perhaps?

nautobot/core/ui/object_detail.py
```diff
@@ -894,6 +894,84 @@ def render_body_content(self, context):
...
+        *,
+        filter_name,
+        related_field_names=None,
+        body_content_template_path="components/panel/stats_panel_body.html",
+        **kwargs,
+    ):
+        """Instantiate a `StatsPanel`.
+
+        Args:
+            related_field_names (dict): key/value pair of { <model_class>: <query_strings> }.
```
glennmatthews, line 909: Update docstring and provide an example?

nautobot/core/ui/object_detail.py
```diff
@@ -894,6 +894,84 @@ def render_body_content(self, context):
         return result
 
 
+class StatsPanel(Panel):
```
glennmatthews, line 897: Add this class to `nautobot.apps.ui` namespace

nautobot/core/ui/object_detail.py
```diff
@@ -894,6 +894,84 @@ def render_body_content(self, context):
...
+                if isinstance(related_field, tuple):
+                    related_object_model_class, query = related_field
+                else:
+                    related_object_model_class, query = related_field, f"{self.filter_name}__in"
+                filter_dict = {query: self.filter_pks}
+                related_object_count = (
+                    related_object_model_class.objects.restrict(request.user, "view").filter(**filter_dict).count()
+                )
+                related_object_model_class_meta = related_object_model_class._meta
+                related_object_list_url = get_route_for_model(related_object_model_class, "list")
```
glennmatthews, line 955: Possibly use `validated_viewname` here so that we don't get an error if the related model doesn't actually have a list view?

nautobot/dcim/templates/dcim/location.html
```diff
@@ -52,12 +52,7 @@
             <tr>
                 <td>Time Zone</td>
                  <td>
-                    {% if object.time_zone %}
-                        {{ object.time_zone }} (UTC {{ object.time_zone|tzoffset }})<br />
-                        <small class="text-muted">Local time: {% timezone object.time_zone %}{% now "SHORT_DATETIME_FORMAT" %}{% endtimezone %}</small>
-                    {% else %}
-                        <span class="text-muted">&mdash;</span>
-                    {% endif %}
+                    {{ object.time_zone | format_time_zone}}
```
glennmatthews, line 55: Revert?

glennmatthews (approved):

PR #6356, opened 2024-10-10
Added Event Broker Config

### LINKED ISSUE #6324: Event Publishing: Config
#6312 implemented dynamic registration of event brokers

We should follow existing django config patterns here. Something like:

```
NAUTOBOT_EVENT_PUBLISHERS = {
    "<name>": {
        "CLASS": "nautobot.something.MyPublisherClass",
        "OPTIONS": {
            "url": "redis://example.com",
        },
        "TOPICS": {
            "INCLUDE": ["foo.*", "bar.*"],
            "EXCLUDE": ["bar.bazz", "bar.fuzz"]
        }
    }
}
```

timizuoebideri1:
# Closes #6324 
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests


nautobot/core/tests/test_events.py
```diff
@@ -83,37 +82,42 @@ def test_publish_events_to_broker(self):
...
+    @override_settings(
+        NAUTOBOT_EVENT_BROKERS={
+            "SyslogEventBroker": {
+                "CLASS": "nautobot.core.events.SyslogEventBroker",
+                "TOPICS": {
+                    "INCLUDE": ["nautobot.test.*"],
+                },
+            }
+        }
+    )
```
timizuoebideri1, line 94: Modifying the settings here does not seem to work

nautobot/core/events/__init__.py
```diff
@@ -12,6 +16,32 @@
 logger = logging.getLogger(__name__)
 
 
+def load_event_brokers(settings):
+    """Process plugins and log errors if they can't be loaded."""
+    if not getattr(settings, "NAUTOBOT_EVENT_BROKERS", None):
```
glennmatthews, line 21: We should define an appropriate default (empty?) value in `settings.py` and assert that it's always present.

nautobot/core/events/__init__.py
```diff
@@ -12,6 +16,32 @@
...
+        return
+
+    for publisher_name, publisher in settings.NAUTOBOT_EVENT_BROKERS.items():
+        options = publisher.get("OPTIONS", {})
+        include_topics = None
+        exclude_topics = None
+        if topics := publisher.get("TOPICS"):
+            include_topics = topics.get("INCLUDE")
+            exclude_topics = topics.get("EXCLUDE")
```
glennmatthews, line 30:
```suggestion
        topics = publisher.get("TOPICS", {})
        include_topics = topics.get("INCLUDE", [])
        exclude_topics = topics.get("EXCLUDE", [])
```

Might actually make sense for `include_topics` to default to `["*"]` if unset, actually?

nautobot/core/events/__init__.py
```diff
@@ -12,6 +16,32 @@
...
+        try:
+            event_publisher_path = publisher["CLASS"]
+            module_path, class_name = event_publisher_path.rsplit(".", 1)
+            module = importlib.import_module(module_path)
+            event_publisher_class = getattr(module, class_name)
+            event_broker = event_publisher_class(**options)
+            register_event_broker(event_broker)
+        except ModuleNotFoundError as err:
+            raise EventPublisherNotFound(
+                f"Unable to import event publisher {publisher_name}: Module not found"
```
glennmatthews, line 41:
```suggestion
                f"Unable to import event publisher {publisher_name}: Module {module_path} not found"
```

nautobot/core/events/__init__.py
```diff
@@ -12,6 +16,32 @@
...
+            include_topics = topics.get("INCLUDE")
+            exclude_topics = topics.get("EXCLUDE")
+        options.update({"include_topics": include_topics, "exclude_topics": exclude_topics})
+        try:
+            event_publisher_path = publisher["CLASS"]
+            module_path, class_name = event_publisher_path.rsplit(".", 1)
+            module = importlib.import_module(module_path)
+            event_publisher_class = getattr(module, class_name)
+            event_broker = event_publisher_class(**options)
+            register_event_broker(event_broker)
```
glennmatthews, line 38: More error handling please - malformed event_publisher_path, event_publisher_class not found in the module, etc.

nautobot/core/events/__init__.py
```diff
@@ -55,7 +89,12 @@ def publish_event(*, topic, payload):
             lowest common denominator for serializability.
     """
     for event_broker in _EVENT_BROKERS:
-        event_broker.publish(topic=topic, payload=payload)
+        exclude_topics = event_broker.exclude_topics
+        include_topics = event_broker.include_topics
+        if (include_topics and is_topic_match(topic, include_topics)) or (
+            exclude_topics and not is_topic_match(topic, exclude_topics)
```
glennmatthews, line 95: I think the preferred logic here would be:
`if not is_topic_match(topic, exclude_topics) and is_topic_match(include_topics)`

nautobot/core/events/base.py
```diff
@@ -2,10 +2,19 @@
...
 
+from nautobot.core.events.exceptions import EventPublisherImproperlyConfigured
+
 
 class EventBroker(ABC):
     """Abstract base class for concrete implementations of event brokers such as syslog, Redis, Kafka, etc."""
 
+    def __init__(self, *args, include_topics=None, exclude_topics=None, **kwargs) -> None:
+        if include_topics and exclude_topics:
+            raise EventPublisherImproperlyConfigured("You can only provide either include or exclude_topics not both.")
```
glennmatthews, line 13: We definitely want the option to specify both, e.g.:

```
INCLUDE: ["nautobot.events.create.*", "nautobot.events.update.*"]
EXCLUDE: ["*.ipam.*"]
```

nautobot/core/tests/test_events.py
glennmatthews: Given the current pattern, might need to add some internal/test APIs like `unregister_all_event_brokers` or `load_event_brokers(reload=True)`?

nautobot/core/settings.py
```diff
@@ -1082,3 +1082,5 @@ def silk_request_logging_intercept_logic(request):
 
 
 SILKY_INTERCEPT_FUNC = silk_request_logging_intercept_logic
+
+NAUTOBOT_EVENT_BROKERS = {}
```
glennmatthews, line 1086: I'd probably just call this `EVENT_BROKERS` and move it up to around line 98 of this file to be alphabetically inline. Don't forget to also:

- Add as a commented-out and documented entry in `nautobot/core/templates/nautobot_config.py.j2` around line 362
- Add documentation in `nautobot/core/settings.yaml`

nautobot/core/events/__init__.py
```diff
@@ -12,6 +16,43 @@
...
+        if not isinstance(exclude_topics, list):
+            raise EventPublisherImproperlyConfigured(
+                f"Malformed Event Publisher Settings: Expected `EXCLUDE` to be a 'list', instead a '{type(exclude_topics).__name__}' was provided"
+            )
+        options.update({"include_topics": include_topics, "exclude_topics": exclude_topics})
+        try:
+            event_publisher_path = publisher["CLASS"]
+            module_path, class_name = event_publisher_path.rsplit(".", 1)
+            module = importlib.import_module(module_path)
+            event_publisher_class = getattr(module, class_name)
```
glennmatthews, line 43: Can probably simplify this by using `django.utils.module_loading.import_string` instead?

nautobot/core/events/__init__.py
```diff
@@ -12,6 +16,43 @@
 logger = logging.getLogger(__name__)
 
 
+def load_event_brokers(nautobt_event_brokers):
+    """Process plugins and log errors if they can't be loaded."""
+    for publisher_name, publisher in nautobt_event_brokers.items():
```
glennmatthews, line 21:
```suggestion
def load_event_brokers(event_broker_configs):
    """Process plugins and log errors if they can't be loaded."""
    for publisher_name, publisher in event_broker_configs.items():
```

Also, is there a reason you're using `publisher` verbiage throughout instead of `broker`?

nautobot/core/events/__init__.py
```diff
@@ -55,7 +100,10 @@ def publish_event(*, topic, payload):
             lowest common denominator for serializability.
     """
     for event_broker in _EVENT_BROKERS:
-        event_broker.publish(topic=topic, payload=payload)
+        exclude_topics = event_broker.exclude_topics
+        include_topics = event_broker.include_topics
+        if not is_topic_match(topic, exclude_topics) and is_topic_match(topic, include_topics):
```
glennmatthews, line 105: If logically equivalent, this might be more intuitive:

```suggestion
        if is_topic_match(topic, include_topics) and not is_topic_match(topic, exclude_topics):
```

nautobot/core/events/__init__.py
```diff
@@ -12,6 +16,43 @@
...
+def load_event_brokers(nautobt_event_brokers):
+    """Process plugins and log errors if they can't be loaded."""
+    for publisher_name, publisher in nautobt_event_brokers.items():
+        options = publisher.get("OPTIONS", {})
+        topics = publisher.get("TOPICS", {})
+        if not isinstance(topics, dict):
+            raise EventPublisherImproperlyConfigured(
+                f"Malformed Event Publisher Settings: Expected `TOPICS` to be a 'dict', instead a '{type(topics).__name__}' was provided"
+            )
+        include_topics = topics.get("INCLUDE", ["*"])
```
glennmatthews, line 28: Probably makes sense to default these to `None` here since the EventBroker class itself knows how to handle `None` as inputs, just to avoid duplication of defaults definitions? I know it'd make the type checking below a bit more complex but I think it's worthwhile.

nautobot/core/tests/test_events.py
```diff
@@ -83,37 +85,108 @@ def test_publish_events_to_broker(self):
...
-            with self.assertLogs("nautobot.events.nautobot.test.event") as cm:
-                publish_event(topic="nautobot.test.event", payload={"a": 1})
-            self.assertEqual(cm.output, ["INFO:nautobot.events.nautobot.test.event:" + json.dumps({"a": 1}, indent=4)])
-        finally:
-            deregister_event_broker(event_broker)
-
+        with self.assertLogs("nautobot.events.nautobot.test.event") as cm:
+            publish_event(topic="nautobot.test.event", payload={"a": 1})
+            publish_event(topic="nautobot.test.no-publish.event", payload={"a": 1})
+        # This test assets that only `nautobot.test.event` topic was published
```
glennmatthews, line 103: I don't think this is correct - think you'd want to do a separate test with `assertNoLogs("nautobot.events.nautobot.test.no-publish.event")`

nautobot/core/tests/test_events.py
```diff
@@ -83,37 +85,108 @@ def test_publish_events_to_broker(self):
...
-            publish_event(topic="nautobot.test.event", payload={"b": 2})
-            self.assertEqual(
-                sub.get_message(timeout=5.0),
-                {"type": "pmessage", "pattern": "nautobot.*", "channel": "nautobot.test.event", "data": '{"b": 2}'},
-            )
-        finally:
-            if sub is not None:
-                sub.close()
-            deregister_event_broker(event_broker)
+        connection = redis.StrictRedis.from_url(url, decode_responses=True)
```
glennmatthews, line 116: Why'd you remove the try/finally block?

nautobot/core/tests/test_events.py
```diff
@@ -83,37 +85,108 @@ def test_publish_events_to_broker(self):
...
+        sub.psubscribe("nautobot.*")
+        self.assertEqual(
+            sub.get_message(timeout=5.0),
+            {"type": "psubscribe", "pattern": None, "channel": "nautobot.*", "data": 1},
+        )
+        publish_event(topic="nautobot.test.event", payload={"b": 2})
+        self.assertEqual(
+            sub.get_message(timeout=5.0),
+            {"type": "pmessage", "pattern": "nautobot.*", "channel": "nautobot.test.event", "data": '{"b": 2}'},
+        )
```
glennmatthews, line 127: Make sense to add a negative/non-publish test here?

nautobot/core/events/__init__.py
```diff
@@ -12,6 +16,43 @@
...
+            event_publisher_class = getattr(module, class_name)
+            event_broker = event_publisher_class(**options)
+            register_event_broker(event_broker)
+        except ModuleNotFoundError as err:
+            raise EventPublisherNotFound(
+                f"Unable to import event publisher {publisher_name}: Module {module_path} not found"
+            ) from err
+        except AttributeError as err:
+            raise EventPublisherNotFound(
+                f"Unable to import event publisher {publisher_name}: {publisher_name} not found in module {module_path}"
```
glennmatthews, line 52:
```suggestion
                f"Unable to import event publisher {publisher_name}: {class_name} not found in module {module_path}"
```

nautobot/core/tests/test_events.py
```diff
@@ -83,37 +85,108 @@ def test_publish_events_to_broker(self):
...
+        with self.assertRaises(EventPublisherNotFound) as err:
+            broker_config = {
+                "TestEventBroker": {
+                    "CLASS": "nautobot.core.tests.test_events.InvalidBroker",
+                }
+            }
+            load_event_brokers(broker_config)
+        self.assertEqual(
+            str(err.exception),
+            "Unable to import event publisher TestEventBroker: TestEventBroker not found in module nautobot.core.tests.test_events",
```
glennmatthews, line 151:
```suggestion
            "Unable to import event publisher TestEventBroker: InvalidBroker not found in module nautobot.core.tests.test_events",
```

nautobot/core/events/__init__.py
```diff
@@ -12,6 +16,43 @@
...
+            raise EventPublisherImproperlyConfigured(
+                f"Malformed Event Publisher Settings: Expected `EXCLUDE` to be a 'list', instead a '{type(exclude_topics).__name__}' was provided"
+            )
+        options.update({"include_topics": include_topics, "exclude_topics": exclude_topics})
+        try:
+            event_publisher_path = publisher["CLASS"]
+            module_path, class_name = event_publisher_path.rsplit(".", 1)
+            module = importlib.import_module(module_path)
+            event_publisher_class = getattr(module, class_name)
+            event_broker = event_publisher_class(**options)
```
glennmatthews, line 44: While you're doing type-checking, should we assert that `issubclass(event_publisher_class, EventBroker)`?

nautobot/core/tests/test_events.py
timizuoebideri1: `cm.output` is a list of all logs collected, Here im asserting only 1 and correct entire exists.

glennmatthews: Doesn't the parameter to `assertLogs` apply as a filter on the logs collected though? I wouldn't think this would capture *all* logs from Python.

timizuoebideri1: Not all logs from python but all logs inside its context

```python
 with self.assertLogs("nautobot.events.nautobot.test.event") as cm:
            publish_event(topic="nautobot.test.event", payload={"a": 1})
            publish_event(topic="nautobot.test.event.another", payload={"b": 1})
```

In this case: `cm.output` would be:
```json
[
     'INFO:nautobot.events.nautobot.test.event:{\n    "a": 1\n}', 
     'INFO:nautobot.events.nautobot.test.event.another:{\n    "b": 1\n}'
]
```

Since both topics are not excluded

nautobot/core/templates/nautobot_config.py.j2
```diff
@@ -360,6 +360,17 @@ SECRET_KEY = os.getenv("NAUTOBOT_SECRET_KEY", "{{ secret_key }}")
...
+# Event Brokers
+# EVENT_BROKERS = {
+#     "SyslogEventBroker": {
+#         "CLASS": "nautobot.core.events.SyslogEventBroker",
+#         "TOPICS": {
+#             "INCLUDE": ["nautobot.test.*"],
+#             "EXCLUDE": ["*.no-publish*"],
+#         },
+#     }
+# }
```
glennmatthews, line 372: Generally in this file the commented-out case should be the same as the default value for the config, so this would just be:

```suggestion
# EVENT_BROKERS = {}
```

nautobot/core/tests/test_events.py
glennmatthews: Right... my point was that `(nautobot.events.)nautobot.test.no-publish.event` isn't included in the `nautobot.events.nautobot.test.event` context, so it wouldn't be captured even if logged?

nautobot/core/tests/test_events.py
```diff
@@ -113,7 +140,78 @@ def test_publish_events_to_redis(self):
                 sub.get_message(timeout=5.0),
                 {"type": "pmessage", "pattern": "nautobot.*", "channel": "nautobot.test.event", "data": '{"b": 2}'},
             )
+
+            # Assert exclude topics are not published
+            publish_event(topic="nautobot.test.event.no-publish", payload={"c": 3})
+            self.assertNotEqual(
+                sub.get_message(timeout=5.0),
```
glennmatthews, line 147: Should `sub.get_message(timeout=5.0)` just return None in the case where there's no published message in that time? Might be better to just assert that instead?

nautobot/core/tests/test_events.py
timizuoebideri1: oooh

glennmatthews (approved):

PR #6365, opened 2024-10-11
Refactored `ObjectsTablePanel` to accept either a table class and a queryset or an already initialized table.

### LINKED ISSUE #6330: Revisit ObjectsTablePanel data pattern
### Proposed Changes

Revisit how `ObjectsTablePanel` gets its content provided to it. Currently the base view/viewset needs to provide a `BaseTable` instance already initialized with appropriate data, but then `ObjectsTablePanel` modifies that instance to show/hide columns, paginate the queryset, etc. In the best case this is somewhat repetitive, in the worst case there's outright duplication of code and wasted effort as the view applies one set of table configs that are then undone or overwritten by the panel.

### Justification

It might be cleaner for the base view to provide a table class and a basic queryset, and let `ObjectsTablePanel` handle the instantiation of the table?

HanlinMiao:
# Closes #6330 
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests


nautobot/core/ui/object_detail.py
```diff
@@ -483,17 +485,26 @@ def get_extra_context(self, context):
...
-        body_content_table = context[self.table_key]
+        if isinstance(context[self.table_key], tuple):
+            body_content_table_class, body_content_table_queryset = context[self.table_key]
+            body_content_table = body_content_table_class(body_content_table_queryset)
+        else:
+            body_content_table = context[self.table_key]
         request = context["request"]
 
         if self.exclude_fields or self.include_fields:
-            for column in body_content_table.columns:
```
HanlinMiao, line 490: The "pk" column is missing from `body_content_table.columns`, so the toggle was not showing up on the `AssociatedContactsTable` in the contacts tab. This change fixes it.

nautobot/core/ui/object_detail.py
```diff
@@ -483,17 +485,26 @@ def get_extra_context(self, context):
...
+                    self.include_fields and column not in self.include_fields
                 ):
-                    body_content_table.columns.hide(column.name)
+                    body_content_table.columns.hide(column)
                 else:
-                    body_content_table.columns.show(column.name)
+                    body_content_table.columns.show(column)
+        else:
+            # If no exclude and include fields are specified
+            # show all columns
```
HanlinMiao, line 505: Let me know if this makes sense

glennmatthews: Is this not the automatic/default behavior?

nautobot/core/ui/object_detail.py
glennmatthews: For most related object tables I think we *don't* want the `pk` column, as we don't usually support bulk operations from this view? Might need to be a special case for the AssociatedContactsTable?

nautobot/dcim/views.py
```diff
@@ -247,8 +247,8 @@ def get_extra_context(self, request, instance):
             .select_related("parent", "location_type")
```
glennmatthews, line 247: Would it make sense to move the `.restrict()` call into the ObjectsTablePanel class logic so that we don't need to repeat it throughout?

nautobot/dcim/views.py
```diff
@@ -247,8 +247,8 @@ def get_extra_context(self, request, instance):
...
         )
 
-        children_table = tables.LocationTypeTable(children)
-        locations_table = tables.LocationTable(locations)
+        children_table = (tables.LocationTypeTable, children)
+        locations_table = (tables.LocationTable, locations)
```
glennmatthews, line 251: I was hoping we could take this a step further and have a pattern similar what you've got with StatsPanel now, something like:

```
object_detail_content = ...
   ObjectsTablePanel(table_class=tables.LocationTypeTable, filter_field="parent")
   ObjectsTablePanel(table_class=tables.LocationTable, filter_field="location_type"),
```

nautobot/core/ui/object_detail.py
```diff
@@ -399,8 +399,6 @@ def __init__(
         table_class,
         table_filter=None,
         table_attribute=None,
-        select_related_fields=None,
-        prefetch_related_fields=None,
```
glennmatthews: To be clear, I think there's still cases where specifying these may be necessary (automatic optimization can only be so smart), so I wouldn't remove the option entirely.

nautobot/core/ui/object_detail.py
```diff
@@ -412,7 +416,12 @@ def __init__(
         """Instantiate an ObjectsTable panel.
 
         Args:
-            table_key (str): The render context key string that contains the BaseTable instance of interest.
+            table_class (obj): The table class object used for render the table e.g. CircuitTable, DeviceTable.
+            table_filter (str, optional): The name of the filter to filter the queryset to initialize the table class.
+                Cannot be used together with `table_attribute`
+            table_attribute (str, optional): The attribute of the detail view instance that contains the queryset to initialize the table class. e.g. `instance.dynamic_groups`
```
glennmatthews, line 422:
```suggestion
            table_attribute (str, optional): The attribute of the detail view instance that contains the queryset to initialize the table class. e.g. `"dynamic_groups"`
```

nautobot/core/ui/object_detail.py
```diff
@@ -484,17 +499,34 @@ def get_extra_context(self, context):
...
+        instance = get_obj_from_context(context)
+        body_content_table_queryset = body_content_table_model
+        if self.table_attribute:
+            object_manager = getattr(instance, self.table_attribute)
+        if self.table_filter:
+            object_manager = body_content_table_model.objects.filter(**{self.table_filter: instance})
+        body_content_table_queryset = (
+            object_manager.restrict(request.user, "view")
+            .order_by(*self.order_by_fields or [])
+        )
```
glennmatthews, line 514: I think the fallback here might result in removing the default ordering if no order_by_fields are set? Not sure, but maybe safer:

```suggestion
        body_content_table_queryset = object_manager.restrict(request.user, "view")
        if self.order_by_fields:
            body_content_table_queryset = body_content_table_queryset.order_by(*self.order_by_fields)
```

nautobot/core/ui/object_detail.py
```diff
@@ -484,17 +499,34 @@ def get_extra_context(self, context):
         for listing and adding objects. It also handles field inclusion/exclusion and
         displays the appropriate table title if provided.
         """
-        body_content_table = context[self.table_key]
+        body_content_table_class = self.table_class
+        body_content_table_model = body_content_table_class.Meta.model
         request = context["request"]
+        instance = get_obj_from_context(context)
+        body_content_table_queryset = body_content_table_model
```
glennmatthews, line 506:
```suggestion
```

nautobot/core/ui/object_detail.py
glennmatthews: (bump)

nautobot/core/ui/object_detail.py
glennmatthews: (bump)

nautobot/core/ui/object_detail.py
HanlinMiao: Sorry I missed the comment but yeah the pk column is not showing previously

nautobot/core/ui/object_detail.py
```diff
@@ -503,30 +512,36 @@ def get_extra_context(self, context):
...
-            for column in body_content_table._sequence:
-                body_content_table.columns.show(column)
+                    body_content_table.columns.show(column.name)
+        # Enable bulk action toggle if the user has appropriate permissions
+        app_label = body_content_table_model._meta.app_label
+        model_name = body_content_table_model._meta.model_name
+        user = request.user
+        if self.enable_bulk_actions and (
+            user.has_perm(f"{app_label}.delete_{model_name}") or user.has_perm(f"{app_label}.change_{model_name}")
+        ):
```
glennmatthews, line 543: Could use `get_permission_for_model()` here, might be a bit cleaner.

nautobot/circuits/views.py
```diff
@@ -112,33 +112,14 @@ class ProviderUIViewSet(NautobotUIViewSet):
             ),
             ObjectsTablePanel(
                 weight=200,
-                table_key="circuits_table",
+                table_class=tables.CircuitTable,
+                table_filter="provider",
                 section=SectionChoices.FULL_WIDTH,
+                exclude_fields=["pk", "provider"],
```
glennmatthews, line 118: Don't need to explicitly exclude pk now, right?

nautobot/core/ui/object_detail.py
```diff
@@ -395,14 +396,20 @@ class ObjectsTablePanel(Panel):
...
+        table_class,
+        table_filter=None,
+        table_attribute=None,
+        select_related_fields=None,
+        prefetch_related_fields=None,
+        order_by_fields=None,
         table_title=None,
         max_display_count=None,
         include_fields=None,
         exclude_fields=None,
```
glennmatthews, line 408: Not critical for this PR but we should consider renaming these to `include_columns` and `exclude_columns` instead?

nautobot/core/ui/object_detail.py
```diff
@@ -426,8 +440,18 @@ def __init__(
...
                 permissions are determined by default based on the model.
             related_field_name (str, optional): The name of the field used to filter related objects (typically for query string parameters).
+            enable_bulk_actions (bool, optional): Show the pk toggle columns on the table if the user has the appropriate permissions.
         """
-        self.table_key = table_key
+        self.table_class = table_class
+        if table_filter and table_attribute:
+            raise ValueError("You can only specify either `table_filter` or `table_attribute`")
+        if not table_filter and not table_attribute:
+            raise ValueError("You cannot specify both `table_filter` and `table_attribute`")
```
glennmatthews, line 449:
```suggestion
            raise ValueError("You must specify either `table_filter` or `table_attribute`")
```

nautobot/core/ui/object_detail.py
```diff
@@ -484,8 +508,23 @@ def get_extra_context(self, context):
...
+        body_content_table_model = body_content_table_class.Meta.model
         request = context["request"]
+        instance = get_obj_from_context(context)
+        object_manager = None  # This is to make pylint happy
+        if self.table_attribute:
+            object_manager = getattr(instance, self.table_attribute)
+        if self.table_filter:
+            object_manager = body_content_table_model.objects.filter(**{self.table_filter: instance})
+        body_content_table_queryset = (
+            object_manager.restrict(request.user, "view")
```
glennmatthews, line 521:
```suggestion
        if self.table_attribute:
            body_content_table_queryset = getattr(instance, self.table_attribute)
        else:
            body_content_table_queryset = body_content_table_model.objects.filter(**{self.table_filter: instance})
        body_content_table_queryset = (
            body_content_table_queryset.restrict(request.user, "view")
```

nautobot/core/ui/object_detail.py
```diff
@@ -484,8 +508,23 @@ def get_extra_context(self, context):
...
+            .select_related(*self.select_related_fields or [])
+            .prefetch_related(*self.prefetch_related_fields or [])
```
glennmatthews, line 523: Per the docs, `select_related()` with no args actually does a select related on *all* foreign keys, which is probably not what we want here! So this might need to be:

```suggestion
        if self.select_related_fields:
            body_content_table_queryset = body_content_table_queryset.select_related(*self.select_related_fields)
        if self.prefetch_related_fields:
            body_content_table_queryset = body_content_table_queryset.prefetch_related(*self.prefetch_related_fields)
```

nautobot/core/ui/object_detail.py
```diff
@@ -1362,7 +1412,14 @@ def __init__(
         **kwargs,
     ):
         if panels is None:
-            panels = (ObjectsTablePanel(weight=100, table_key="associated_dynamic_groups_table"),)
+            panels = (
+                ObjectsTablePanel(
+                    weight=100,
+                    table_class=DynamicGroupTable,
+                    table_attribute="dynamic_groups",
+                    exclude_fields=["pk", "content_type"],
```
glennmatthews, line 1420: No need to explicitly exclude pk now right?

nautobot/core/ui/object_detail.py
```diff
@@ -1394,7 +1451,15 @@ def __init__(
...
     ):
         if panels is None:
-            panels = (ObjectsTablePanel(weight=100, table_key="associated_object_metadata_table"),)
+            panels = (
+                ObjectsTablePanel(
+                    weight=100,
+                    table_class=ObjectMetadataTable,
+                    table_attribute="associated_object_metadata",
+                    order_by_fields=["metadata_type", "scoped_fields"],
+                    exclude_fields=["pk", "assigned_object"],
```
glennmatthews, line 1460: No need to explicitly exclude pk now right?

nautobot/dcim/views.py
```diff
@@ -227,35 +227,19 @@ class LocationTypeUIViewSet(
...
                 weight=100,
-                table_key="children_table",
+                table_class=tables.LocationTypeTable,
+                table_filter="parent",
                 table_title="Child Location Type(s)",
             ),
             ObjectsTablePanel(
                 weight=200,
-                table_key="locations_table",
-                table_title="Location(s) of this Type",
```
glennmatthews, line 236: Don't we still want the table_title?

nautobot/circuits/views.py
```diff
@@ -248,7 +229,7 @@ def queryset_list_url_filter(self, key, value, context):
                 section=SectionChoices.LEFT_HALF,
                 weight=100,
                 fields="__all__",
-                exclude_fields=["comments", "circuit_termination_a", "circuit_termination_z"],
+                exclude_columns=["comments", "circuit_termination_a", "circuit_termination_z"],
```
glennmatthews, line 232: Sorry, too far - for an ObjectFieldsPanel, keeping it as `exclude_fields` would be correct/preferable. Only ObjectsTablePanel should be changed to `exclude_columns`.

nautobot/core/ui/object_detail.py
```diff
@@ -731,7 +777,7 @@ def __init__(
         self,
         *,
         fields="__all__",
-        exclude_fields=(),
+        exclude_columns=(),
         context_object_key="object",
```
glennmatthews, line 781: revert

nautobot/core/ui/object_detail.py
```diff
@@ -744,15 +790,15 @@ def __init__(
             fields (str, list): The ordered list of fields to display, or `"__all__"` to display fields automatically.
                 Note that ManyToMany fields and reverse relations are **not** included in `"__all__"` at this time, nor
                 are any hidden fields, nor the specially handled `id`, `created`, `last_updated` fields on most models.
-            exclude_fields (list): Only relevant if `fields == "__all__"`, in which case it excludes the given fields.
+            exclude_columns (list): Only relevant if `fields == "__all__"`, in which case it excludes the given fields.
```
glennmatthews, line 793: revert

nautobot/core/ui/object_detail.py
```diff
@@ -744,15 +790,15 @@ def __init__(
...
             context_object_key (str): The key in the render context that will contain the object to derive fields from.
             ignore_nonexistent_fields (bool): If True, `fields` is permitted to include field names that don't actually
                 exist on the provided object; otherwise an exception will be raised at render time.
             label (str): If omitted, the provided object's `verbose_name` will be rendered as the label
                 (see `render_label()`).
         """
         self.fields = fields
-        self.exclude_fields = exclude_fields
+        self.exclude_columns = exclude_columns
```
glennmatthews, line 801: revert

nautobot/core/ui/object_detail.py
```diff
@@ -815,7 +861,7 @@ def get_data(self, context):
             data["_hierarchy"] = instance
 
         for field_name in fields:
-            if field_name in self.exclude_fields:
+            if field_name in self.exclude_columns:
```
glennmatthews, line 864: revert

glennmatthews: missed this one?

glennmatthews (approved):

PR #6378, opened 2024-10-16
Develop to next post 2.3.7

HanlinMiao:


glennmatthews:
Need to renumber some migrations again it looks like.

nautobot/extras/migrations/0119_fix_dynamic_group_group_type_data_migration.py
glennmatthews, line 1: This is the wrong migration to renumber isn't it? Since it's in v2.3.7 already.

glennmatthews (changes requested):

glennmatthews (approved):

timizuoebideri1 (approved):

PR #6389, opened 2024-10-17
K8s/Kubectl connectivity/auth

### LINKED ISSUE #6255: k8s/kubectl connectivity/auth
For the k8s native jobs integration, we need to support specifying all the connectivity and auth parameters for talking the k8s API (kubectl equivalent commands) to enqueue jobs and check state.

It likely make sense to track this via ExternalIntration instances linked to JobQueue instances with appropriate secrets support.

The idea for now is to have a k8s scoped service account token that is pre-provisioned in the pod spec. We will need to build a built-in secrets provider that can read the token.

HanlinMiao:
# Closes #6255
# What's Changed

- [x] Added `external_integration` FK to Job Queue model
- [x] Added validation logic to enforce that only Kubernetes job queues can have external integrations assigned
- [ ] Added code to call K8S jobs via external integration linked to the job queue

# Screenshots

Screenshot 2024-10-17 at 12 25 14 PM
Screenshot 2024-10-17 at 12 25 56 PM

# TODO

- [x] Unit, Integration Tests


nautobot/extras/factory.py
```diff
@@ -151,11 +151,20 @@ class Meta:
...
+        has_external_integration = NautobotBoolIterator()
 
     name = factory.LazyAttributeSequence(lambda o, n: f"{o.queue_type} Job Queue - {n}")
     description = factory.Maybe("has_description", factory.Faker("text", max_nb_chars=CHARFIELD_MAX_LENGTH), "")
     queue_type = factory.Iterator(JobQueueTypeChoices.CHOICES, getter=lambda choice: choice[0])
     tenant = factory.Maybe("has_tenant", random_instance(Tenant))
+    external_integration = factory.Maybe(
+        "has_external_integration",
+        factory.LazyAttribute(
+            lambda obj: get_random_instances(ExternalIntegration, minimum=1)[0]
```
glennmatthews, line 163: Can we not use `random_instance(ExternalIntegration)` here?

nautobot/extras/filters/__init__.py
```diff
@@ -916,11 +916,20 @@ class JobQueueFilterSet(NautobotFilterSet, TenancyModelFilterSetMixin):
...
     )
     queue_type = django_filters.MultipleChoiceFilter(choices=JobQueueTypeChoices, null_value=None)
+    external_integration = NaturalKeyOrPKMultipleChoiceFilter(
+        queryset=ExternalIntegration.objects.all(),
+        label="External Integration (name or ID)",
+    )
+    has_external_integration = RelatedMembershipBooleanFilter(
+        field_name="external_integration",
+        label="Has external integration",
+    )
```
glennmatthews, line 932: I thought we normally only have `has_` filters for reverse relations and many-to-many, not regular foreignkey?

nautobot/extras/models/jobs.py
```diff
@@ -571,6 +578,16 @@ def display(self):
         workers = "worker" if worker_count == 1 else "workers"
         return f"{self.queue_type}: {self.name} ({worker_count} {workers})"
 
+    def clean(self):
+        super().clean()
+        # Validate that only kubernetes job queue can be linked to an external integration object.
+        if self.external_integration and self.queue_type != JobQueueTypeChoices.TYPE_KUBERNETES:
```
glennmatthews, line 584: I'm wondering if we want the converse as well - if TYPE_KUBERNETES, should an external_integration be required?

nautobot/extras/api/serializers.py
```diff
@@ -569,6 +570,21 @@ class Meta:
...
+    def validate(self, data):
+        if self.instance:
+            job_queue_type = data.get("queue_type", self.instance.queue_type)
+            external_integration = data.get("external_integration", self.instance.external_integration)
+
+            if external_integration and job_queue_type != JobQueueTypeChoices.TYPE_KUBERNETES:
+                error_message = f"Only job queues with type {JobQueueTypeChoices.TYPE_KUBERNETES} can be linked to external integration objects."
+                errors = {}
+                if "external_integration" in data:
+                    errors["external_integration"] = [error_message]
```
mateusz-lachowski-codilime, line 582: I wonder if this error shouldn't be assigned to the `job_queue_type` as well?

HanlinMiao: Good call I had the error assigned to `job_queue_type` as well before but I deleted that part because I thought it was redundant. I can add it back

HanlinMiao:
Context is an optional `CharField` on `JobQueue`, and we can have a settings variable for the kube-config file path.

nautobot/extras/models/jobs.py
```diff
@@ -799,87 +815,150 @@ def enqueue_job(
...
-                # Set alarm_handler to be called on a SIGALRM, and schedule a SIGALRM based on the soft time limit
-                signal.signal(signal.SIGALRM, alarm_handler)
-                signal.alarm(int(job_model.soft_time_limit) or settings.CELERY_TASK_SOFT_TIME_LIMIT)
-
-                try:
-                    eager_result = run_job.apply(
+        job_queue = JobQueue.objects.get(name=task_queue)
+        # Kubernetes Job Queue logic
+        if job_queue.queue_type == JobQueueTypeChoices.TYPE_KUBERNETES:
+            config_file_path = "./development/kind-kube-config"
```
glennmatthews, line 821: This will become a settings variable right?

nautobot/extras/models/jobs.py
```diff
@@ -799,87 +815,150 @@ def enqueue_job(
...
+            config.load_kube_config(config_file_path)
+            core_v1 = core_v1_api.CoreV1Api()
+            api_instance = core_v1
+            name = "nautobot-pod-demo"
```
glennmatthews, line 825: Does this need to become a settings variable as well? Or does it need to be another attribute of the JobQueue?

nautobot/extras/models/jobs.py
```diff
@@ -799,87 +815,150 @@ def enqueue_job(
...
+            resp = None
+            try:
+                resp = api_instance.read_namespaced_pod(name=name, namespace="default")
```
glennmatthews, line 828: Do we need a settings variable or JobQueue field for the `namespace`?

nautobot/extras/models/jobs.py
```diff
@@ -799,87 +815,150 @@ def enqueue_job(
...
+                print("Found old pod")
```
glennmatthews, line 829: Change prints to logger calls throughout please

nautobot/extras/models/jobs.py
```diff
@@ -799,87 +815,150 @@ def enqueue_job(
...
+            except ApiException as e:
+                if e.status != 404:
+                    print(f"Unknown error: {e}")
+                    sys.exit(1)
```
glennmatthews, line 833: Killing Django in case of an API failure is definitely not the right behavior here.

nautobot/extras/models/jobs.py
```diff
@@ -799,87 +815,150 @@ def enqueue_job(
...
+            if not resp:
+                print(f"Pod {name} does not exist. Creating it...")
+                pod_manifest = {
+                    "apiVersion": "v1",
+                    "kind": "Pod",
+                    "metadata": {"name": name},
+                    "spec": {
+                        "containers": [
+                            {
+                                "image": "networktocode/nautobot:latest",
```
glennmatthews, line 844: Settings variable?

nautobot/extras/models/jobs.py
```diff
@@ -799,87 +815,150 @@ def enqueue_job(
...
+                resp = api_instance.read_namespaced_pod(name=name, namespace="default")
+                print("Found old pod")
+            except ApiException as e:
+                if e.status != 404:
+                    print(f"Unknown error: {e}")
+                    sys.exit(1)
+
+            if not resp:
+                print(f"Pod {name} does not exist. Creating it...")
+                pod_manifest = {
```
glennmatthews, line 837: Should we instead assume that the pod manifest is part of the environment setup, rather than trying to guess the correct manifest here?

nautobot/extras/models/jobs.py
```diff
@@ -799,87 +815,150 @@ def enqueue_job(
...
+                        ]
+                    },
+                }
+                resp = api_instance.create_namespaced_pod(body=pod_manifest, namespace="default")
+                print("create pod")
+                while True:
+                    resp = api_instance.read_namespaced_pod(name=name, namespace="default")
+                    if resp.status.phase != "Pending":
+                        break
+                    time.sleep(1)
```
glennmatthews, line 857: Do we need this block? Blocking like this feels bad. Can we just fire-and-forget instead?

nautobot/extras/models/jobs.py
```diff
@@ -799,87 +815,150 @@ def enqueue_job(
...
+                stdout=True,
+                tty=False,
+                _preload_content=False,
+            )
+            resp.write_stdin("nautobot-server runjob --local -u admin nautobot.core.jobs.ExportObjectList\n")
+            sresult = resp.read_stdout()
+            serror = resp.read_stderr()
+            print(f"Job Result is: {sresult}")
+            print(f"Error is: {serror}")
+            resp.close()
```
glennmatthews, line 878: We definitely shouldn't block until the job runs to completion like this, *unless* the `synchronous` flag is true.

nautobot/extras/models/jobs.py
```diff
@@ -799,87 +815,150 @@ def enqueue_job(
...
+            return job_result
+        else:
```
glennmatthews, line 880: Since we're `return`-ing above, the `else` here and corresponding indentation change isn't *strictly* necessary.

nautobot/core/settings.py
```diff
@@ -1092,3 +1092,22 @@ def silk_request_logging_intercept_logic(request):
 
 
 SILKY_INTERCEPT_FUNC = silk_request_logging_intercept_logic
+
+#
+# Kubernetes settings variables
+#
+
+# File path to the config file used to load and configure the kubernetes cluster
+KUBERNETES_CONFIG_FILE = os.getenv("NAUTOBOT_KUBERNETES_CONFIG_FILE", "./development/kind-kube-config")
```
glennmatthews, line 1101: Wonder if this should just be `settings.KUBECONFIG` since the standard environment variable is just `$KUBECONFIG`.

nautobot/core/settings.py
```diff
@@ -1092,3 +1092,22 @@ def silk_request_logging_intercept_logic(request):
...
+KUBERNETES_JOB_POD_NAME = os.getenv("NAUTOBOT_KUBERNETES_JOB_POD_NAME", "nautobot-job-pod")
+
+# Namespace of the kubernetes pod created in the kubernetes cluster
+KUBERNETES_JOB_POD_NAMESPACE = os.getenv("NAUTOBOT_KUBERNETES_JOB_POD_NAMESPACE", "default")
+
+# Nautobot container name in the kubernetes pod
+KUBERNETES_JOB_CONTAINER_NAME = os.getenv("NAUTOBOT_KUBERNETES_JOB_CONTAINER_NAME", "nautobot")
+
+# Nautobot Image name used to build a container in the kubernetes pod
+KUBERNETES_JOB_IMAGE_NAME = os.getenv("NAUTOBOT_KUBERNETES_JOB_IMAGE_NAME", "networktocode/nautobot:latest")
```
glennmatthews, line 1113: Not sure if any/all of these should have global defaults other than `""` or `None`. Can set them to development-specific values in `development/nautobot_config.py` or `nautobot/core/tests/nautobot_config.py` as appropriate.

nautobot/extras/models/jobs.py
```diff
@@ -558,6 +563,16 @@ class JobQueue(PrimaryModel):
         blank=True,
         null=True,
     )
+    context = models.CharField(
+        max_length=CHARFIELD_MAX_LENGTH, help_text="Kubernetes Configuration Context", blank=True
+    )
+    secrets_group = models.ForeignKey(
```
glennmatthews, line 569: Maybe need an appropriate `help_text` here as well?

nautobot/extras/forms/forms.py
```diff
@@ -1246,11 +1246,18 @@ class JobQueueBulkEditForm(TagsBulkEditFormMixin, NautobotBulkEditForm):
...
+        queryset=SecretsGroup.objects.all(),
+        required=False,
+    )
+    context = forms.CharField(required=False, max_length=CHARFIELD_MAX_LENGTH)
     description = forms.CharField(required=False, max_length=CHARFIELD_MAX_LENGTH)
 
     class Meta:
         model = JobQueue
         nullable_fields = [
+            "context",
```
glennmatthews, line 1259: It doesn't look like `context` is `null=True` (nor should it be) so is this wrong?

nautobot/extras/models/jobs.py
```diff
@@ -799,6 +814,80 @@ def enqueue_job(
         if task_queue is None:
             task_queue = job_model.default_job_queue.name
 
+        job_queue = JobQueue.objects.get(name=task_queue)
+        # Kubernetes Job Queue logic
+        if job_queue.queue_type == JobQueueTypeChoices.TYPE_KUBERNETES:
+            kube_config_context = job_queue.context
```
glennmatthews, line 820: Maybe split all of this out into a private helper function? `enqueue_job()` is already beastly long.

nautobot/extras/models/jobs.py
```diff
@@ -799,6 +814,80 @@ def enqueue_job(
...
+                pod_manifest = {
+                    "apiVersion": "v1",
+                    "kind": "Pod",
+                    "metadata": {"name": pod_name},
+                    "spec": {
+                        "containers": [
+                            {
+                                "image": nautobot_image_name,
+                                "name": nautobot_container_name,
+                                "ports": [{"containerPort": nautobot_container_port_number}],
```
glennmatthews, line 857: What purpose does port this serve - is it even necessary?

nautobot/extras/models/jobs.py
```diff
@@ -799,6 +814,80 @@ def enqueue_job(
...
+            resp = None
+
+            # Try to read existing pod
+            try:
+                logger.info(f"Reading existing pod {pod_name} in namespace {pod_namespace}")
+                resp = api_instance.read_namespaced_pod(name=pod_name, namespace=pod_namespace)
+                logger.info(f"Found existing pod {pod_name} in namespace {pod_namespace}")
+            except ApiException as e:
+                if e.status != 404:
+                    logger.info(f"Unknown error: {e}")
```
glennmatthews, line 843: Perhaps `logger.error`? Should we return early in this case rather than continuing?

nautobot/extras/models/jobs.py
```diff
@@ -799,6 +814,80 @@ def enqueue_job(
...
+                }
+                logger.info(f"Creating pod {pod_name} in namespace {pod_namespace}")
+                resp = api_instance.create_namespaced_pod(body=pod_manifest, namespace=pod_namespace)
+                logger.info(f"Reading pod {pod_name} in namespace {pod_namespace}")
+                resp = api_instance.read_namespaced_pod(name=pod_name, namespace=pod_namespace)
+                logger.info("Done.")
+
+            # Calling exec interactively
+            logger.info(f"Waiting for container {nautobot_container_name} to be fully set up")
+            time.sleep(10)
```
glennmatthews, line 870: Can we get rid of this? That's a long time to wait.

nautobot/extras/tests/test_filters.py
```diff
@@ -929,15 +929,36 @@ class JobQueueFilterSetTestCase(FilterTestCases.FilterTestCase, FilterTestCases.
...
         # create some job queues that do not have jobs attached to them
         # for has_jobs boolean filter
-        JobQueue.objects.create(name="Empty Job Queue 1", queue_type=JobQueueTypeChoices.TYPE_CELERY)
+        secrets_groups = (
+            SecretsGroup.objects.create(name="Secrets Group 1"),
+            SecretsGroup.objects.create(name="Secrets Group 2"),
+            SecretsGroup.objects.create(name="Secrets Group 3"),
+        )
+        JobQueue.objects.create(
+            name="Empty Job Queue 1", queue_type=JobQueueTypeChoices.TYPE_CELERY, secrets_group=secrets_groups[0]
```
glennmatthews, line 947: Should we disallow setting a secrets_group (or a context) on a TYPE_CELERY JobQueue?

nautobot/extras/models/jobs.py
HanlinMiao: Excerpt from https://faun.pub/should-i-configure-the-ports-in-kubernetes-deployment-c6b3817e495:
"containerPortList of ports to expose from the container. Exposing a port here gives the system additional information about the network connections a container uses, but is primarily informational." I think it is not necessary.

nautobot/extras/models/jobs.py
HanlinMiao: I am playing around with a pod.yaml file that should replace the hardcoded `pod_manifest` here in this [story](https://app.zenhub.com/workspaces/nautobot-core-engineering-64a875e74842500d9e80e23e/issues/gh/nautobot/nautobot/6088)

nautobot/extras/api/serializers.py
```diff
@@ -569,6 +570,30 @@ class Meta:
...
+
+                if "secrets_group" in data:
+                    errors["secrets_group"] = [error_message]
+
+            if context and queue_type == JobQueueTypeChoices.TYPE_CELERY:
+                error_message = f"Context is not allowed on job queues of type {JobQueueTypeChoices.TYPE_CELERY}"
+                if "context" in data:
+                    errors["context"] = [error_message]
+
+                raise serializers.ValidationError(errors)
```
glennmatthews, line 593: This feels off to me - so if the instance has a secrets group or context, but `secrets_group` or `context` isn't in `data`, we will raise an empty ValidationError?

Also the indentation on that `raise` seems off to me.

glennmatthews (approved):
Let's merge and iterate.

nautobot/extras/api/serializers.py
HanlinMiao: Yeah it should raise the errors regardless. Thanks for catching it.

PR #6402, opened 2024-10-20
U/grelleum 4899 app provided table column extensions

### LINKED ISSUE #4899: App provided table column extensions
### As ...

P.D. - Plugin Developer

### I want ...

The ability to contribute extra columns to existing tables in core.

### So that ...

I can add additional data/context to a core table that is relevant to my app.

I have a similar app API interface to filter extensions that allows me to register new columns to existing tables.

I expect to specify my columns as django_tables2/nautobot core Column classes.

Because columns are typically tightly coupled to data on the instance, it would be a good idea to provide a mechanism that allows me to optionally override/augment the queryset being used by the table. This way I could add a prefetch for my app's related model, etc.

- Plugins (and other Extensibility)

### Database Changes

_No response_

### External Dependencies

_No response_

grelleum:
# Closes #4899

# What's Changed
Added TableExtension class that allows app developers to add columns to core model tables.
- App can add one or more new columns to a table.
- App can add one or more, new or core columns, as default visible columns.
- App can remove one or more core default columns from the default visible columns.

# Screenshots
Unmodified Circuits table:

image

Circuits table modified by Example App.
- Added column `Provider ASN` as a default column.
- Removed column `Tenant` from the default columns.

image

# TODO
- [X] Unit, Integration Tests


examples/example_app/example_app/table_extensions.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # Add extended columns here only if the columns should be visible by default.
+    add_to_default_columns = ["example_app_manufacturer"]
+
+    @classmethod
+    def alter_queryset(cls, queryset):
+        """Return an optimized queryset.
+
+        Implementing this method is optional but is recommended to optimize
+        the database queries to account for the additional columns.
```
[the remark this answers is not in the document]
grelleum: I'll let the core team decide if this is useful or if it should be pulled out.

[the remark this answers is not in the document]
lampwins: There are some examples that come to mind where this is useful, like when a custom column is derived from a specific custom relationship and thus the automatic logic is unable to generate a query optimization for that case. I have seen that a few times now.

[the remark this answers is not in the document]
glennmatthews: Yeah - again, was trying to be clear that I totally understand the value of the feature in general. I was just curious about *this specific* queryset optimization for *this specific table*. :-)

nautobot/docs/development/apps/api/platform-features/table-extensions.md
```diff
@@ -0,0 +1,64 @@
...
+
+* The file must be named `table_extensions.py`
+* The variable `table_extensions` must be declared in that file, and contain a list of `TableExtension` subclasses
+* The `model` attribute of each `TableExtension` subclass must be set to a valid model name in the dotted pair format (`{app_label}.{model}`, e.g. `tenant.tenant` or `dcim.device`)
+
+There are four ways that `TableExtension` can be used to modify a table.
+
+* Custom columns can be added to a table.
+* Core and custom columns can be added to the default visible columns.
+* Core columns can be removed from the default visible columns when they are not used or applicable.
```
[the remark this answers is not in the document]
grelleum: Interesting questions, I'll start with the last.

Because of the way this is implemented in the `CustomApp.NautobotAppConfig` class, it would be difficult to implement a weight system, because it would require knowledge of all modifications.  So it would need to be redesigned to accommodate that.   I can see the desire to have deterministic results.

There should be no issue if multiple apps remove columns from the same table. The `default_columns` is rebuilt using a comprehension that checks filters out any column names that are in the `remove_from_default_columns` list.  For example: When App 'A' is initialized it will remove column 'a' from the XYZTable, then when App 'B' is initialized column 'a' will no longer be in `default_columns`.  If App 'B' want to remove the same or different columns, it should not encounter any issues.

Last case: If one app wants to remove a column and another app want to add that same column.  The outcome would be determined based on the order the apps are initialized, with the last initialized app winning the battle.  I believe this scenario would not be common because:
1. If an app is adding a column to the table's `default_columns`, it is either a new column added by that app, and would be prefixed with the app_name.
2. If an app is adding a core column to the table's `default_columns`, it is likely not in the defaults and so it would not need to be removed.
3. If an app is removing a core column from the table's `default_columns`, it is likely already in the defaults and would not need to be added.

nautobot/extras/tests/test_plugins.py
```diff
@@ -647,6 +650,96 @@ def test_filter_form(self):
         self.assertIn("example_app_description", form.fields.keys())
 
 
+class TableExtensionTest(TestCase):
+    """Tests for adding table extensions."""
+
+    def test_alter_queryset(self):
+        """Test the 'alter_queryset' method of the TableExtension class."""
```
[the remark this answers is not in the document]
grelleum: Yes, that was the idea there - that it returns what is passed to it.  Really it is there for coverage.

nautobot/docs/development/apps/api/platform-features/table-extensions.md
```diff
@@ -0,0 +1,64 @@
...
+
++++ 2.9.99-alpha
+
+Apps can extend any model-based `Table` classes that are provided by the Nautobot core.
+
+The requirements to extend a table are:
+
+* The file must be named `table_extensions.py`
+* The variable `table_extensions` must be declared in that file, and contain a list of `TableExtension` subclasses
+* The `model` attribute of each `TableExtension` subclass must be set to a valid model name in the dotted pair format (`{app_label}.{model}`, e.g. `tenant.tenant` or `dcim.device`)
```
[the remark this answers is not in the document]
lampwins: For now, let's just focus on the models, and thus the list views. I'd be concerned with tightly coupling this to specific tables, and issues with this breaking as updates occur. Perhaps it can be implemented as an advanced option in the future, but not needed now.

nautobot/docs/development/apps/api/platform-features/table-extensions.md
```diff
@@ -0,0 +1,64 @@
+# Extending Tables
+
++++ 2.9.99-alpha
```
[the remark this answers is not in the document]
lampwins: `develop`/2.3

examples/example_app/example_app/table_extensions.py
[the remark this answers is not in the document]
grelleum: It is in the example_app and it's not really an optimization, just a demo how to use it.  I get that the examples should be meaningful, but also they should be simple, so I think I errored on the side of simplicity.

nautobot/extras/tests/test_plugins.py
```diff
@@ -647,6 +650,96 @@ def test_filter_form(self):
...
+                    context.exception, "Custom column `test_column` is not an instance of django_tables2.Column."
+                )
+
+        with self.subTest("raises AttributeError"):
+            column = tables.Column()
+            with self.assertRaises(AttributeError) as context:
+                plugins._add_column_to_table_base_columns(table, "pk", column)
+                self.assertEqual(
+                    context.exception, "There was a conflict with table column `pk`, the custom column was ignored."
+                )
```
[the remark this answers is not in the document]
grelleum: I can modify this to change the behavior.  It is based on the `FilterExtension` implementation, which will fail in a similar way.

nautobot/docs/development/apps/api/platform-features/table-extensions.md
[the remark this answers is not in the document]
grelleum: I've updated this to 2.3.9.

nautobot/extras/tests/test_plugins.py
```diff
@@ -647,6 +650,96 @@ def test_filter_form(self):
...
+    """Tests for adding table extensions."""
+
+    def test_alter_queryset(self):
+        """Test the 'alter_queryset' method of the TableExtension class."""
+        extension = plugins.TableExtension()
+        queryset = object()
+        result = extension.alter_queryset(queryset)
+        self.assertEqual(result, queryset)
+
+    @mock.patch("nautobot.extras.plugins.logger.error")
```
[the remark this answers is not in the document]
grelleum: today I learned ... `self.assertLogs`

nautobot/extras/plugins/__init__.py
```diff
@@ -262,6 +266,29 @@ def validate(cls, user_config, nautobot_version):
...
+    def _register_table_extensions(self):
+        """Register tables extensions (if any)."""
+        table_extensions = import_object(f"{self.__module__}.{self.table_extensions}")
+        if table_extensions is not None:
+            register_table_extensions(table_extensions, self.name)
+            self.features["table_extensions"] = {
+                "columns": [
+                    f"{table_extension.model} -> {column_name}"
+                    for table_extension in table_extensions
+                    for column_name in table_extension.table_columns
```
glennmatthews, line 278: Add `if table_extension.table_columns` check here?

nautobot/extras/plugins/__init__.py
```diff
@@ -491,6 +518,106 @@ def register_filter_extensions(filter_extensions, plugin_name):
...
+
+def _add_column_to_table_base_columns(table, column_name, column):
+    """Attach a column to an existing table."""
+    import django_tables2
+
+    if not isinstance(column, django_tables2.Column):
+        raise TypeError(f"Custom column `{column_name}` is not an instance of django_tables2.Column.")
+
+    if column_name in table.base_columns:
+        raise AttributeError(f"There was a conflict with table column `{column_name}`, the custom column was ignored.")
```
glennmatthews, line 582: Not opposed to extra error handling, just curious - with the requirement that the column name include the app name as a prefix, this should pretty much "never" occur, no?

nautobot/extras/plugins/__init__.py
```diff
@@ -491,6 +518,106 @@ def register_filter_extensions(filter_extensions, plugin_name):
...
+        _validate_table_column_name_is_prefixed_with_app_name(name, app_name)
+        try:
+            _add_column_to_table_base_columns(table, name, column)
+        except AttributeError as error:
+            logger.error(*error.args)
+
+
+def _add_column_to_table_base_columns(table, column_name, column):
+    """Attach a column to an existing table."""
+    import django_tables2
```
glennmatthews, line 576: Does this need to be a local import here, or could it be moved to the top of the file?

nautobot/extras/plugins/__init__.py
```diff
@@ -491,6 +518,106 @@ def register_filter_extensions(filter_extensions, plugin_name):
...
+
+    table.base_columns[column_name] = column
+
+
+def _alter_table_view_queryset(table_extension, app_name):
+    """Replace the model view queryset with an optimized queryset from the app."""
+    from nautobot.core.utils.lookup import get_view_for_model
+
+    view = get_view_for_model(table_extension.model, view_type="List")
+    view.queryset = table_extension.alter_queryset(view.queryset)
```
glennmatthews, line 592: Hm. Does this work for both old-style `ObjectListView` as well as `NautobotUIViewSet`? I'm assuming so since the example_app is extending one of each - but I'm a bit concerned in the NautobotUIViewSet case since the class-level `queryset` attribute is shared across all sub-views under the ViewSet. I wonder if in that case it **might** be safer to actually wrap/monkey-patch the `view.alter_queryset()` functions to just call `table_extension.alter_queryset()` when specifically handling a list request?

nautobot/extras/plugins/__init__.py
```diff
@@ -491,6 +518,106 @@ def register_filter_extensions(filter_extensions, plugin_name):
...
+    table = get_table_for_model(table_extension.model)
+    add_registered_columns = [name for name in table_extension.add_to_default_columns if name in table.base_columns]
+
+    if add_registered_columns:
+        table.Meta.default_columns = (*table.Meta.default_columns, *add_registered_columns)
+
+    if table_extension.remove_from_default_columns:
+        table.Meta.default_columns = tuple(
+            column for column in table.Meta.default_columns if column not in table_extension.remove_from_default_columns
+        )
```
glennmatthews, line 608: It *might* be nice to log a debug message if `remove_from_default_columns` includes any items that aren't actually in the table's default_columns? Definitely no higher than `debug` though.

nautobot/extras/plugins/__init__.py
grelleum: I was concerned about the `NautobotUIViewSet`, which is why I added the code to extend the Circuit table.  This is a good point and I'll look into modifying this behavior.

glennmatthews: Much appreciate your thoroughness! :-)

nautobot/extras/plugins/__init__.py
grelleum: This needs to be a local import.  Otherwise we get this when trying to start the server or workers.
```
celery_worker-1  | Traceback (most recent call last):
celery_worker-1  |   File "/usr/local/bin/nautobot-server", line 3, in <module>
celery_worker-1  |     from nautobot.core.cli import main
celery_worker-1  |   File "/source/nautobot/core/cli/__init__.py", line 17, in <module>
celery_worker-1  |     from nautobot.extras.plugins.utils import load_plugins
celery_worker-1  |   File "/source/nautobot/extras/plugins/__init__.py", line 7, in <module>
celery_worker-1  |     import django_tables2
celery_worker-1  |   File "/usr/local/lib/python3.12/site-packages/django_tables2/__init__.py", line 19, in <module>
celery_worker-1  |     from .tables import Table, table_factory
celery_worker-1  |   File "/usr/local/lib/python3.12/site-packages/django_tables2/tables.py", line 186, in <module>
celery_worker-1  |     class Table(metaclass=DeclarativeColumnsMetaclass):
celery_worker-1  |   File "/usr/local/lib/python3.12/site-packages/django_tables2/tables.py", line 26, in __new__
celery_worker-1  |     attrs["_meta"] = opts = TableOptions(attrs.get("Meta", None), name)
celery_worker-1  |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
celery_worker-1  |   File "/usr/local/lib/python3.12/site-packages/django_tables2/tables.py", line 121, in __init__
celery_worker-1  |     DJANGO_TABLES2_TEMPLATE = getattr(
celery_worker-1  |                               ^^^^^^^^
celery_worker-1  |   File "/usr/local/lib/python3.12/site-packages/django/conf/__init__.py", line 102, in __getattr__
celery_worker-1  |     self._setup(name)
celery_worker-1  |   File "/usr/local/lib/python3.12/site-packages/django/conf/__init__.py", line 89, in _setup
celery_worker-1  |     self._wrapped = Settings(settings_module)
celery_worker-1  |                     ^^^^^^^^^^^^^^^^^^^^^^^^^
celery_worker-1  |   File "/usr/local/lib/python3.12/site-packages/django/conf/__init__.py", line 217, in __init__
celery_worker-1  |     mod = importlib.import_module(self.SETTINGS_MODULE)
celery_worker-1  |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
celery_worker-1  |   File "/usr/local/lib/python3.12/importlib/__init__.py", line 90, in import_module
celery_worker-1  |     return _bootstrap._gcd_import(name[level:], package, level)
celery_worker-1  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
celery_worker-1  | ModuleNotFoundError: No module named 'nautobot_config'
```

itdependsnetworks:
Can you update docs here as well? https://github.com/nautobot/nautobot/blob/044490de8c4e88a09be11cdd0c31ccc7c19140bf/nautobot/docs/development/apps/api/setup.md?plain=1#L11

nautobot/extras/plugins/__init__.py
glennmatthews: Ack. Didn't know that importing tables2 forced evaluation of the Django settings.

nautobot/extras/plugins/__init__.py
grelleum: Pretty much it should not occur - the app name prefix should provide a namespace to avoid that. Really the only conceivable way that could happen would be if the app name was the same as a base model.

nautobot/extras/plugins/__init__.py
```diff
@@ -491,6 +516,106 @@ def register_filter_extensions(filter_extensions, plugin_name):
...
+    view = get_view_for_model(table_extension.model, view_type="List")
+    view.queryset = table_extension.alter_queryset(view.queryset)
+
+
+def _modify_default_table_columns(table_extension):
+    """Add or remove columns from the table default columns."""
+    from nautobot.core.utils.lookup import get_table_for_model
+
+    table = get_table_for_model(table_extension.model)
+    add_registered_columns = [name for name in table_extension.add_to_default_columns if name in table.base_columns]
```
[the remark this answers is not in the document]
grelleum: I've refactored this code and added debug logging for when a column name is missing.

nautobot/extras/plugins/__init__.py
grelleum: I've refactored this code and added debug logging for when a column name is missing.

grelleum:
> Can you update docs here as well?
> 
> https://github.com/nautobot/nautobot/blob/044490de8c4e88a09be11cdd0c31ccc7c19140bf/nautobot/docs/development/apps/api/setup.md?plain=1#L11

Added!

nautobot/extras/tests/test_plugins.py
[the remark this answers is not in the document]
grelleum: I have changed this to log an error rather than raise an exception as recommended.

nautobot/extras/plugins/__init__.py
grelleum: I am still pending on this.

glennmatthews:
> @glennmatthews Do you think there would be a benefit if I moved all the TableExtension code to a `table_extensions.py` module, exposed the two functions needed by `__init__` and imported that?

I don't have a strong opinion but moving stuff out of `__init__.py` in the long run would be nice. Maybe move to the existing `tables.py` rather than creating a new module?

nautobot/extras/plugins/__init__.py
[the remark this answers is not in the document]
glennmatthews: New approach looks good.

grelleum:
>  Maybe move to the existing `tables.py` rather than creating a new module?
Importing `tables` into `__init__` prevents Nautobot from starting.  `ModuleNotFoundError: No module named 'nautobot_config'`.
I'll skip it and focus on isolating the queryset for the list view.

glennmatthews:
> > Maybe move to the existing `tables.py` rather than creating a new module?

> Importing `tables` into `__init__` prevents Nautobot from starting.  `ModuleNotFoundError: No module named 'nautobot_config'`.

Ack, same problem as before. thanks for trying it.

nautobot/extras/plugins/__init__.py
grelleum: I am unsure of a good way to patch this in the UIViewSet that targets just the list view queryset.  I think I would need to patch either the `.as_view` or the `.get_queryset` method, however, I would sill need to identify how the method is being called to know if it is for a list view or not.  I'm not saying it can't be done, but I'm not sure what the effort would be and if it is worth mucking that deep into the internals.

glennmatthews: Fair enough. I think maybe for now, we can leave it as is with a TODO comment and see if it causes any issues?

grelleum: Good deal - I've added a TODO on this.

glennmatthews (approved):

PR #6406, opened 2024-10-21
DataTablePanel UI Component

### LINKED ISSUE #6137: UI Components: Tables not backed by BaseTable
Tables not backed by BaseTable, i.e. generic data, lists of dicts, etc.
- ~Copy button at row (value) level~
- Column headers are passed as input to the render context
- Data provided as input to the render context
  - List of dicts is the primary use case
  - Options to pass into the render context method, or as an accessor at declaration time (data in a model's jsonfield)

glennmatthews:
# Closes #6137
# What's Changed

- Add `DataTablePanel` UI component for rendering raw data without using a django-tables2 `Table` class.
- As I'm not aware of use cases for this component in core itself, add some examples in the example app "examplemodel" detail view (which I've now converted to use the components framework)
- Enhance `ObjectsTablePanel` to support a `context_table_key` parameter for cases where you still want to dynamically construct a `Table` instance in your view context instead of letting the panel generate one for you
- Fix missing `related_field_name` on the panels in the LocationType detail view.

# Screenshots

image

# TODO

- [x] Unit, Integration Tests


changes/6137.fixed
```diff
@@ -0,0 +1 @@
+Fixed `related_field_name` on object tables in LocationType detail view.
```
HanlinMiao, line 1: Is there a difference between `related_field_name` and `table_filter`? Can we combine them?

HanlinMiao (approved):

changes/6137.fixed
glennmatthews: I was wondering a bit about that myself. As I understand it:

- `related_field_name` is used as a URL query parameter both for the `add` button link (to pre-populate the related model form with a selection) and the "show all" link (as a filter parameter on the related model list view)
- `table_filter` is a queryset filter parameter used to get the correct queryset of related objects to render into the table.

In many (most?) cases they will be the same, but I have no doubt there will be exceptions at some point. Maybe it would make sense to default an unset `related_field_name` to fall through to the `table_filter` instead of defaulting to `obj._meta.model_name` though?

HanlinMiao: I like that, not blocking for this PR. Feel free to merge

glennmatthews:
Went down a couple of rabbit holes and ended up finding and fixing a couple more issues in a582187:

- Add fallback from related_field_name to table_filter
- Convert ExternalIntegration detail view to use component framework (since it had multiple valid ObjectMetadata records resulting from our factories)
- Add handling of JSONField in ObjectFieldsPanel (since ExternalIntegration has two JSONFields)
- Fix configuration of Contacts, Dynamic Groups, and Metadata tab tables after some manual testing with paginated data in each of these.

examples/example_app/example_app/views.py
```diff
@@ -77,6 +79,72 @@ class ExampleModelUIViewSet(views.NautobotUIViewSet):
...
+                {"col_1": "value_1a", "col_2": "value_2", "col_3": "value_3", "col_4": "not shown"},
+                {"col_1": "value_1b", "col_2": None},
+            ]
+            context["data_2"] = [
+                {
+                    "a": format_html('<a href="https://en.wikipedia.org/wiki/{val}">{val}</a>', val=1),
+                    "e": format_html('<a href="https://en.wikipedia.org/wiki/{val}">{val}</a>', val=5),
+                    "i": format_html('<a href="https://en.wikipedia.org/wiki/{val}">{val}</a>', val=9),
+                    "o": format_html('<a href="https://en.wikipedia.org/wiki/{val}">{val}</a>', val=15),
+                    "u": format_html('<a href="https://en.wikipedia.org/wiki/{val}">{val}</a>', val=21),
```
mateusz-lachowski-codilime, line 133: I'm not sure of the purpose of those urls but I would at least suggest to take the wikipedia address into some variable. Or maybe even go with:
```
{letter: format_html('<a href="https://en.wikipedia.org/wiki/{val}">{val}</a>', val=number) for letter, number in zip('aeiou', [1,5,9,15,21])}
# or if numbers are not needed just iterate over aeiou and put it in the url?
```

examples/example_app/example_app/views.py
```diff
@@ -77,6 +79,72 @@ class ExampleModelUIViewSet(views.NautobotUIViewSet):
...
+        context = super().get_extra_context(request, instance)
+        if self.action == "retrieve":
+            # Add dynamic table of objects for custom panel
+            context["dynamic_table"] = CircuitTable(Circuit.objects.restrict(request.user, "view"))
+            # Add non-object data for object detail view custom tables
+            context["data_1"] = [
+                {"col_1": "value_1a", "col_2": "value_2", "col_3": "value_3", "col_4": "not shown"},
+                {"col_1": "value_1b", "col_2": None},
+            ]
+            context["data_2"] = [
```
mateusz-lachowski-codilime, line 127: Another good idea might be to move those table definitions to a separate file and then import as dict or import and call some function to generate the proper data. Then looking at `get_extra_context` would be much simpler with just `context["data_2"], context["columns_2"], context["column_headers_2"] = example_table_data()`.

examples/example_app/example_app/views.py
glennmatthews: Purpose was purely just to demonstrate that the data table can contain HTML content if desired/appropriate. I can add some comments to that effect - efficiency of implementation was not a major concern to me since this is just example code.

HanlinMiao (approved):

PR #6410, opened 2024-10-22
Add No-Index Robots.txt

bryanculver:
# Closes #DNE
# What's Changed

If there are cases where you will deploy Nautobot to a public-facing connection, you may want to discourage it from being picked up by web crawlers. Likely you would want this for the entire site and not specific sections (like docs).

- Added Django route to return a "disallow entire site" `robots.txt` file when enabled
- Added configuration option `NO_INDEX_ROBOTS` to control the publishing of the above route, default enabled. When disabled, no route `robots.txt` file is made available (leaving up to the user to control via other means like Nginx).

**TODO:?**
- [ ] Add docs about `robots.txt` to deployment instructions and how to configure something else via Nginx?
- [ ] Should the `robots.txt` always be emitted (not configurable) and/or change just the disallow statement?
- [ ] ?
# Screenshots

Tested via ngrok locally:
Screenshot 2024-10-21 at 16 32 31

# TODO

- [ ] Unit, Integration Tests


nautobot/core/settings.yaml
```diff
@@ -1405,6 +1405,12 @@ properties:
     is_constance_config: true
     type: "object"
     version_added: "1.6.0"
+  NO_INDEX_ROBOTS:
+    default: true
+    description: "Setting this to true causes Nautobot to publish a robot.txt that discourages indexing by web crawlers. Irrelevant if deployed behind a firewall."
```
glennmatthews, line 1410:
```suggestion
    description: "Setting this to true causes Nautobot to publish a `robots.txt` that discourages indexing by web crawlers. Irrelevant if deployed behind a firewall."
```

nautobot/core/settings.py
```diff
@@ -430,6 +430,7 @@
 DATE_FORMAT = os.getenv("NAUTOBOT_DATE_FORMAT", "N j, Y")
 DATETIME_FORMAT = os.getenv("NAUTOBOT_DATETIME_FORMAT", "N j, Y g:i a")
 DEBUG = is_truthy(os.getenv("NAUTOBOT_DEBUG", "False"))
+NO_INDEX_ROBOTS = is_truthy(os.getenv("NAUTOBOT_NO_INDEX_ROBOTS", "True"))
```
glennmatthews, line 433: 1. This isn't a standard Django setting so this is the wrong "section" of the settings file - this should be somewhere around line 158 instead if keeping the name `NO_INDEX_ROBOTS`
2. `NO_INDEX_ROBOTS` is an unintuitive name to me. How about something like `PUBLISH_ROBOTS_TXT` instead?
3. Need to also add the new setting to `nautobot/core/templates/nautobot_config.py.j2` as well.
4. Are we confident that enabled-by-default is the correct behavior for this new feature?

mateusz-lachowski-codilime:
It might be worth to add `<meta name='robots' content='noindex,follow' />` as well to block all cases of possible indexing.

bryanculver:
> It might be worth to add `<meta name='robots' content='noindex,follow' />` as well to block all cases of possible indexing.

Thanks for that feedback! I'll take a look at dusting off this PR this week.

nautobot/core/settings.yaml
```diff
@@ -1473,6 +1473,14 @@ properties:
...
     is_constance_config: true
     type: "boolean"
+  PUBLISH_ROBOTS_TXT:
+    default: true
+    description: >-
+        Setting this to true causes Nautobot to publish a `robots.txt` that discourages indexing by web crawlers.
+        Irrelevant if deployed behind a firewall.
+    environment_variable: "NAUTOBOT_PUBLISH_ROBOTS_TXT"
+    type: "boolean"
+    version_added: "2.3.13"
```
glennmatthews, line 1483:
```suggestion
    version_added: "2.3.15"
```

nautobot/core/templates/inc/media.html
```diff
@@ -47,3 +47,4 @@
     <meta name="theme-color" content="#ffffff">
     <meta charset="UTF-8">
     <meta name="viewport" content="initial-scale=1, maximum-scale=1, user-scalable=no, width=device-width">
+    <meta name="robots" content="noindex, nofollow">
```
glennmatthews, line 50: Should this be conditional on the `PUBLISH_ROBOTS_TXT` setting?

bryanculver: Done :)

glennmatthews (approved):

PR #6419, opened 2024-10-23
Added logic to ModuleBay model to ensure that if the `position` field is empty, its value will be automatically populated from the `name` of the Module Bay instance.

### LINKED ISSUE #6336: Module Interface Template substitution not working
### Environment
* Nautobot version (Docker tag too if applicable): 2.3.4
* Python version: 3.11
* Database platform, version: PG15
* Middleware(s): N/A

### Steps to Reproduce
1. Create Device Type, Parent/Child status: Parent, Create Module Bays xe-0/0/[0-31]
2. Create Device and associate to Device Type created in Step 1
3. Create Module Type
4. Edit Module Type from Step4, Add Interfaces, Use template: Name: `{module}`
5. Create Module, associate with Module Type from step 4 and then install in to device created in Step2 and any bay

### Expected Behavior
An interface is created with a name that mirrors the device's module bay name, e.g. xe-0/0/1

### Observed Behavior
An interface is created with a name of `{module}`

HanlinMiao:
# Closes #6336 
# What's Changed

~~- [x] Added info banner on ModuleBayCreate/UpdateView to remind users that if a value for `position` field is empty, its value will be automatically populated from the `name` of the Module Bay instance.~~
- [x] Added logic to ModuleBay model to ensure that if the `position` field is empty, its value will be automatically populated from the `name` of the Module Bay instance.
- [x] Enhanced `position` fields on ModuleBayCreate/UpdateForms to auto-populate their values from `name` fields.

# Screenshots

Screenshot 2024-10-23 at 3 15 14 PM
Screenshot 2024-10-23 at 3 15 35 PM
Using a new feature
# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/forms.py
+++ b/nautobot/dcim/forms.py
@@ -13,6 +13,8 @@ from nautobot.core.forms import (
     add_blank_choice,
     APISelect,
     APISelectMultiple,
+    AutoPositionField,
+    AutoPositionPatternField,
     BootstrapMixin,
     BulkEditNullBooleanSelect,
     ColorSelect,
@@ -1569,8 +1571,7 @@ class ModuleBayBaseCreateForm(BootstrapMixin, forms.Form):
         required=False,
         help_text="Alphanumeric ranges are supported. (Must match the number of names being created.)",
     )
-    position_pattern = ExpandableNameField(
-        label="Position",
+    position_pattern = AutoPositionPatternField(
         required=False,
         help_text="Alphanumeric ranges are supported. (Must match the number of names being created.)"
         " Default to the names of the module bays unless manually supplied by the user.",
@@ -3562,8 +3563,7 @@ class ModuleBayFilterForm(NautobotFilterForm):
 
 
 class ModuleBayForm(NautobotModelForm):
-    position = forms.CharField(
-        label="Position",
+    position = AutoPositionField(
         max_length=CHARFIELD_MAX_LENGTH,
         help_text="The position of the module bay within the parent device/module. "
         "Defaults to the name of the module bay unless overridden.",
--- a/nautobot/dcim/forms.py
+++ b/nautobot/dcim/forms.py
@@ -1573,7 +1573,8 @@ class ModuleBayBaseCreateForm(BootstrapMixin, forms.Form):
     position_pattern = ExpandableNameField(
         label="Position",
         required=False,
-        help_text="Alphanumeric ranges are supported. (Must match the number of names being created.)",
+        help_text="Alphanumeric ranges are supported. (Must match the number of names being created.)"
+        " Default to the names of the module bays unless manually supplied by the user.",
     )
     description = forms.CharField(max_length=CHARFIELD_MAX_LENGTH, required=False)
 
@@ -3566,6 +3567,13 @@ class ModuleBayFilterForm(NautobotFilterForm):
 
 
 class ModuleBayForm(NautobotModelForm):
+    position = forms.CharField(
+        label="Position",
+        max_length=CHARFIELD_MAX_LENGTH,
+        help_text="""The position of the module bay within the parent device/module.
+        Default to the name of the module bay unless manually supplied by the user.""",
+        required=False,
+    )
     parent_device = DynamicModelChoiceField(
         queryset=Device.objects.all(),
         required=False,
--- a/nautobot/dcim/forms.py
+++ b/nautobot/dcim/forms.py
@@ -3565,8 +3565,8 @@ class ModuleBayForm(NautobotModelForm):
     position = forms.CharField(
         label="Position",
         max_length=CHARFIELD_MAX_LENGTH,
-        help_text="""The position of the module bay within the parent device/module.
-        Default to the name of the module bay unless manually supplied by the user.""",
+        help_text="The position of the module bay within the parent device/module. "
+        "Defaults to the name of the module bay unless overridden.",
         required=False,
     )
     parent_device = DynamicModelChoiceField(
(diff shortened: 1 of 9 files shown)


nautobot/dcim/forms.py
```diff
@@ -3561,6 +3562,13 @@ class ModuleBayFilterForm(NautobotFilterForm):
 
 
 class ModuleBayForm(NautobotModelForm):
+    position = forms.CharField(
+        label="Position",
+        max_length=CHARFIELD_MAX_LENGTH,
+        help_text="""The position of the module bay within the parent device/module.
+        Default to the name of the module bay unless manually supplied by the user.""",
```
glennmatthews, line 3569: should be broken into two strings

```suggestion
        help_text="The position of the module bay within the parent device/module. "
        "Defaults to the name of the module bay unless overridden.",
```

nautobot/dcim/views.py
```diff
@@ -1486,13 +1486,28 @@ class DeviceBayTemplateBulkDeleteView(generic.BulkDeleteView):
 class ModuleBayCommonViewSetMixin:
     """NautobotUIViewSet for ModuleBay views to handle templated create and bulk rename views."""
 
+    def update(self, request, *args, **kwargs):
+        message = format_html(
+            "<div>The value of the <strong>Position</strong> field will default to"
+            " the <strong>Name</strong> of the module bay unless supplied by the user.</div>"
+        )
+        messages.warning(request, message)
```
glennmatthews, line 1494: If these are always present, it might be cleaner to just use add them directly to the edit view template file(s)?

HanlinMiao: There is no specific ModuleBay edit view template currently. Seems like we are using the default edit view template.
Would it be worth the effort to create a new one just for this warning message?

glennmatthews: Might be - should be a real short/simple template inheriting from the default one for the most part.

HanlinMiao:
Screen-recording for the latest commit:
Added the enhancement so that when the `position`/`position_pattern` field is empty, whatever the user enters in the `name` field will be used to populate those fields. This feature is not present in SlugField previously, as slug/key fields are usually marked as required.

https://github.com/user-attachments/assets/61673332-820d-4c34-9714-507d125393d9

development/kind-kube-config
glennmatthews, line 1: wrong branch?

nautobot/project-static/js/forms.js
```diff
@@ -127,6 +129,88 @@ function initializeSlugField(context){
     }
 }
 
+function initializePositionField(context){
+    this_context = $(context);
```
glennmatthews, line 133: Can we do this new code with vanilla JS instead of jquery?

nautobot/project-static/js/forms.js
```diff
@@ -127,6 +129,88 @@ function initializeSlugField(context){
...
+                }
+            });
+        }
+        this_context.find('button.reslugify').click(repopulate);
+    }
+}
+
+function initializePositionPatternField(context){
+    this_context = $(context);
+    var position_field = this_context.find('#id_position_pattern');
```
glennmatthews, line 175: Is this the only difference between these two functions? Can we consolidate logic into a common helper function if so?

development/kind-kube-config
HanlinMiao: `git add .` did me dirty

nautobot/project-static/js/forms.js
```diff
@@ -127,6 +129,51 @@ function initializeSlugField(context){
     }
 }
 
+function initializePositionField(context){
+    var position_field = document.getElementById('id_position');
+    if (!position_field) {
+        var position_field = document.getElementById('id_position_pattern');
+    }
+    if (position_field.length != 0) {
```
glennmatthews, line 137: Is `length` meaningful here or should this just be a `if (position_field)`?

nautobot/project-static/js/forms.js
```diff
@@ -127,6 +129,51 @@ function initializeSlugField(context){
...
+            let str = "";
+            for (source_str of source_arr) {
+                if (str != "") {
+                    str += " ";
+                }
+                let source_id = 'id_' + source_str;
+                let source = document.getElementById(source_id)
+                str += source.value;
+            }
+            position_field.value = str.slice(0, length ? length : 100);
```
glennmatthews, line 162: Might make sense to default to 255 rather than 100?

nautobot/project-static/js/forms.js
```diff
@@ -127,6 +129,51 @@ function initializeSlugField(context){
...
+        var position_field = document.getElementById('id_position_pattern');
+    }
+    if (position_field.length != 0) {
+        var source_arr = position_field.getAttribute('source').split(" ");
+        var length = position_field.getAttribute('maxlength');
+        if (position_field.value) {
+            position_field.setAttribute('_changed', true);
+        } else {
+            position_field.setAttribute('_changed', false);
+        }
```
glennmatthews, line 144: Perhaps `position_field.setAttribute('_changed', Boolean(position_field.value))`

nautobot/project-static/js/forms.js
```diff
@@ -127,6 +129,51 @@ function initializeSlugField(context){
...
+        position_field.addEventListener('change', function() {
+            if (!position_field.value) {
+                position_field.setAttribute('_changed', false);
+            } else {
+                position_field.setAttribute('_changed', true);
+            }
```
glennmatthews, line 150: similar suggestion here

nautobot/project-static/js/forms.js
```diff
@@ -127,6 +129,51 @@ function initializeSlugField(context){
...
+        for (source_str of source_arr) {
+            let source_id = 'id_' + source_str;
+            let source = document.getElementById(source_id);
+            source.addEventListener('keyup', function() {
+                if (position_field && position_field.getAttribute('_changed')=="false") {
+                    repopulate();
+                }
+            });
+        }
+        this_context.find('button.reslugify').click(repopulate);
```
glennmatthews, line 173: Should that be `context` rather than `this_context`? Note that this line is still using jquery either way :)

glennmatthews (approved):

PR #6420, opened 2024-10-24
Delete all just shows confirmation  dialog

### LINKED ISSUE #5321: Delete all OOM
### Environment
* Nautobot version (Docker tag too if applicable): ghcr.io/nautobot/nautobot:2.1.4-py3.9
* Python version: 3.9.18
* Database platform, version: postgres:13
* Middleware(s): uWSGI/nginx - whatever comes with the container + custome uWSGI config

### Steps to Reproduce
1. Select # of (IP's) in this case 56k:
image
2. Press Delete all
3. Get error:
image

### Expected Behavior
I would just get a list viewing all the IP's with the option to delete them all. Similar to deleting any other object.

### Observed Behavior
<class 'MemoryError'>

timizuoebideri1:
# Closes #5321 
# What's Changed

# Screenshots

Screenshot 2024-10-24 at 1 40 38 PM

# TODO

- [ ] Unit, Integration Tests


nautobot/core/views/generic.py
```diff
@@ -1278,18 +1278,53 @@ def get_required_permission(self):
...
 
+    def handle_bulk_delete_all(self, request, **kwargs):
+        model = self.queryset.model
+        self.template_name = "generic/object_bulk_destroy_all.html"
+
+        if self.filterset is not None:
+            # We take this approach because filterset.qs has already applied .distinct(),
+            # and performing a .delete directly on a queryset with .distinct applied is not allowed.
+            pk_list = self.filterset(request.GET, model.objects.only("pk")).qs.values_list("pk")
+            queryset = self.queryset.filter(pk__in=pk_list)
```
glennmatthews, line 1289: This approach is likely to still have issues at scale (e.g. if `pk_list` evaluates to 100000 records). We should be able to pass around the filter params (`request.GET` in the base view code) and pass those to the filterset/queryset instead.

nautobot/core/views/generic.py
```diff
@@ -1278,18 +1278,53 @@ def get_required_permission(self):
     def get(self, request):
         return redirect(self.get_return_url(request))
 
+    def handle_bulk_delete_all(self, request, **kwargs):
+        model = self.queryset.model
+        self.template_name = "generic/object_bulk_destroy_all.html"
```
glennmatthews, line 1283: Because we have views that currently override the base `object_bulk_delete.html`, I'd prefer that we avoid creating a separate template for this code branch - would be better to have conditional logic in the template itself, possibly passing in a new context variable as needed.

nautobot/core/testing/views.py
```diff
@@ -1293,40 +1289,12 @@ def test_bulk_delete_form_contains_all_pks(self):
             response = self.client.post(self._get_url("bulk_delete"), selected_data)
             self.assertHttpStatus(response, 200)
             response_body = utils.extract_page_body(response.content.decode(response.charset))
-            # Check if all the pks are passed into the BulkDeleteForm/BulkDestroyForm
-            for pk in pk_list:
-                self.assertIn(f'<input type="hidden" name="pk" value="{pk}"', response_body)
-
-        @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
-        def test_bulk_delete_form_contains_all_filtered(self):
```
glennmatthews, line 1301: Might need to keep this test as a no-op in case Apps have tests that are calling `super().test_bulk_delete_form_contains_all_filtered()`?

nautobot/core/views/generic.py
```diff
@@ -1278,18 +1278,53 @@ def get_required_permission(self):
     def get(self, request):
         return redirect(self.get_return_url(request))
 
+    def handle_bulk_delete_all(self, request, **kwargs):
+        model = self.queryset.model
+
+        if self.filterset is not None:
+            queryset = self.filterset(request.GET, model.objects.only("pk")).qs
```
glennmatthews, line 1285: Does calling a filterset on a `model.objects.only("pk")` require a bunch of extra DB lookups to resolve the other model fields relevant to the filters? I'd think so.

timizuoebideri1: Yea, that seems plausible

glennmatthews (commented):
Can you please provide some context on manual/scale testing you've done for this fix, before/after load times for the delete view, etc.?

I'd suggest using a model such as Manufacturer, Status, or Role that are trivial to bulk-create. Please test with 100000 records (filtered or unfiltered both) if possible.

nautobot/core/testing/views.py
```diff
@@ -1321,12 +1320,11 @@ def test_bulk_delete_form_contains_all_filtered(self):
...
-            # Check if the first and second pk is passed into the form.
-            self.assertIn(f'<input type="hidden" name="pk" value="{first_pk}"', response_body)
-            self.assertIn(f'<input type="hidden" name="pk" value="{second_pk}"', response_body)
-            self.assertIn("<strong>Warning:</strong> The following operation will delete 2 ", response_body)
-            # Check if the third pk is not passed into the form.
+            # Check if all pks is not part of the html.
+            self.assertNotIn(f'<input type="hidden" name="pk" value="{first_pk}"', response_body)
+            self.assertNotIn(f'<input type="hidden" name="pk" value="{second_pk}"', response_body)
             self.assertNotIn(f'<input type="hidden" name="pk" value="{third_pk}"', response_body)
+            self.assertIn("<strong>Warning:</strong> The following operation will delete 2 ", response_body)
```
mateusz-lachowski-codilime, line 1327: Might be worth to use `assertNotInHtml` and `assertInHtml` just in case that formatting change in template don't break the tests.

mateusz-lachowski-codilime, line 1327: `self.assertInHTML('<input type="hidden" name="_all" value="true" />', response_body)` could be added to this test as well.

timizuoebideri1: There is no `assertNotInHtml`, also the full html tag is a bit unpredictable `<input type="hidden" name="pk" value="<uuid>" class="form-control" required="required" placeholder="None" id="id_pk_0">` as I dont really know which  might be `id__pk__1` or `id__pk__2` etc.

timizuoebideri1:
> Can you please provide some context on manual/scale testing you've done for this fix, before/after load times for the delete view, etc.?
> 
> I'd suggest using a model such as Manufacturer, Status, or Role that are trivial to bulk-create. Please test with 100000 records (filtered or unfiltered both) if possible.

TIME TAKEN to delete 10_000 records: approx 320s
 -> deleting the objects took: approx 30s
 -> redirecting back to the list page after deleting: approx 290s

TIME TAKEN to delete 1_111 filtered records: approx 5s
 -> deleting the objects took: approx 3.5s
 -> redirecting back to the list page after deleting: approx 1.5s

nautobot/core/testing/views.py
glennmatthews: Can possibly use `assertNotContains(..., html=True)`.

nautobot/core/testing/views.py
```diff
@@ -1293,13 +1289,16 @@ def test_bulk_delete_form_contains_all_pks(self):
...
             self.assertHttpStatus(response, 200)
             response_body = utils.extract_page_body(response.content.decode(response.charset))
-            # Check if all the pks are passed into the BulkDeleteForm/BulkDestroyForm
-            for pk in pk_list:
-                self.assertIn(f'<input type="hidden" name="pk" value="{pk}"', response_body)
+            # Assert the table which shows all the selected objects is not part of the html body in delete all case
+            self.assertNotIn('<table class="table table-hover table-headings">', response_body)
+            # Assert none of the hidden input fields for each of the pks that would be deleted is part of the html body
+            for pk in self._get_queryset().values_list("pk", flat=True):
+                self.assertNotIn(f'<input type="hidden" name="pk" value="{pk}"', response_body)
```
glennmatthews, line 1296: Maybe just `assertNotIn(str(pk), response_body)`?

nautobot/core/testing/views.py
```diff
@@ -1321,12 +1320,12 @@ def test_bulk_delete_form_contains_all_filtered(self):
...
             response_body = utils.extract_page_body(response.content.decode(response.charset))
-            # Check if the first and second pk is passed into the form.
-            self.assertIn(f'<input type="hidden" name="pk" value="{first_pk}"', response_body)
-            self.assertIn(f'<input type="hidden" name="pk" value="{second_pk}"', response_body)
-            self.assertIn("<strong>Warning:</strong> The following operation will delete 2 ", response_body)
-            # Check if the third pk is not passed into the form.
+            # Check if all pks is not part of the html.
+            self.assertNotIn(f'<input type="hidden" name="pk" value="{first_pk}"', response_body)
+            self.assertNotIn(f'<input type="hidden" name="pk" value="{second_pk}"', response_body)
             self.assertNotIn(f'<input type="hidden" name="pk" value="{third_pk}"', response_body)
```
glennmatthews, line 1326: Similarly, maybe just `assertNotIn(str(first_pk), response_body)`, etc.?

nautobot/core/views/mixins.py
```diff
@@ -909,7 +912,41 @@ def perform_update(self, request, *args, **kwargs):  # pylint: disable=arguments
...
+    def _get_bulk_delete_all_queryset(self, request):
+        model = self.queryset.model
+
+        # This Mixin is currently been used by both NautobotViewSet ObjectBulkDestroyViewMixin
+        # and BulkDeleteView which uses different keys for accessing filterset
+        filterset = getattr(self, "filterset", None)
+        if filterset is None:
+            filterset = getattr(self, "filterset_class", None)
+
+        if filterset is not None:
```
glennmatthews, line 929: Feels like maybe this should also include something like `and request.GET` so that the delete-all-unfiltered case doesn't need to evaluate the filterset and generate the PK list, but instead can fall through to the `else` case?

nautobot/core/views/mixins.py
```diff
@@ -909,7 +912,41 @@ def perform_update(self, request, *args, **kwargs):  # pylint: disable=arguments
...
+    """
+    UI mixin to bulk destroy all model instances.
+    """
+
+    def _get_bulk_delete_all_queryset(self, request):
+        model = self.queryset.model
+
+        # This Mixin is currently been used by both NautobotViewSet ObjectBulkDestroyViewMixin
+        # and BulkDeleteView which uses different keys for accessing filterset
+        filterset = getattr(self, "filterset", None)
```
glennmatthews, line 925: Nit: I'd prefer `filterset_class` as the variable name in this function rather than `filterset`.

nautobot/core/views/mixins.py
```diff
@@ -414,7 +414,7 @@ def form_valid(self, form):
         else:
             # render the form with the error message.
             data = {}
-            if self.action in ["bulk_update", "bulk_destroy"]:
+            if not request.POST.get("_all") and self.action in ["bulk_update", "bulk_destroy"]:
```
glennmatthews, line 417: Aren't we going to need similar code changes for bulk-update views/templates? IIRC they have the same set of problems with rendering the full table of impacted objects.

nautobot/core/views/mixins.py
```diff
@@ -909,7 +912,41 @@ def perform_update(self, request, *args, **kwargs):  # pylint: disable=arguments
             return self.form_invalid(form)
 
 
-class ObjectBulkDestroyViewMixin(NautobotViewSetMixin, BulkDestroyModelMixin):
+class DeleteAllModelMixin:
+    """
+    UI mixin to bulk destroy all model instances.
+    """
+
+    def _get_bulk_delete_all_queryset(self, request):
```
glennmatthews, line 920: Docstrings on these functions please - yes they're "hidden" but they're an important part of how these views work.

nautobot/core/views/mixins.py
```diff
@@ -951,35 +991,32 @@ def perform_bulk_destroy(self, request, **kwargs):
...
-                )
-            else:
-                self.pk_list = list(model.objects.all().values_list("pk", flat=True))
-        else:
-            self.pk_list = list(request.POST.getlist("pk"))
+        if delete_all:
+            queryset = self._get_bulk_delete_all_queryset(self.request)
+            data = self._bulk_delete_all_context(request, queryset)
+
+        self.pk_list = list(request.POST.getlist("pk"))
```
glennmatthews, line 1001: Should this be in an `else` block to the above `if delete_all` case?

nautobot/core/views/mixins.py
timizuoebideri1: Implementing `BulkEdit` is more complex than `BulkDelete` since each view relies on a custom form rather than a generic one. Additionally, forms currently accept only a `pk_list` rather than a full `queryset`, which would require additional changes to accommodate. Given the scope, I’ll defer this implementation for a future PR.

timizuoebideri1: https://github.com/nautobot/nautobot/issues/6465

glennmatthews (approved):

PR #1081, opened 2021-11-17
[WIP] Implementation of singleton pattern for Jobs

jathanism:
#### Done
- Adds `celery-singleton` as a dependency
- The value `singleton = True` can be set on the inner `Meta` class of a `Job` class to enable one and only one copy of that job being ran at any given time.
- When calling `JobResult.enqueue_job()` now, the `task_id` from the Celery task if it is successfuly published is saved as the `JobResult.job_id` to creating a 1-to-1 mapping of Nautobot `JobResult` and Celery `AsyncResult`.
- In the event that a singleton job is executed twice, subsequent attempts to run it will result in a `DuplicateTaskError` being raised from `JobResult.enqueue_job()`. This is handled in the `JobView` handling code to present a user-friendly error message back to the UI
- New Celery tasks decorated with `@nautobot_task` may now optionally pass any extra arguments supported by `celery-singleton`, the most important of which will be `on_unique=["foo", "bar"]` where the argument are the names of arguments used to identify task uniqueness for the purpose of generating a idempotent hash key for the unique lock.
- Added a new `SingletonJob` to `dummy_plugin.jobs` that can be used to test and demonstrate singleton jobs.
- A custom subclass of `celery_singleton.Singleton` base task class has been added at `nautobot.core.celery.task.NautobotTask` that can conditionally behave as a base Celery `Task` or as a `Singleton` when `apply_async` is called using the `really_singleton=True` argument.
- The Celery application now publishes its internal `task_cls` as `NautobotTask` making it the default Celery task class within Nautobot.
- The specialized `_dumps()` method that gets registered with Kombu for the `nautobot_json` serializer can now take optional `kwargs` that get passed along to `json.dumps()` with the encoder.
- `settings.CELERY_SINGLETON_RAISE_ON_DUPLICATE = True` has been set to always result in singleton tasks raising an exception as the default.

#### Not Done
- API/CLI changes (if any)
- Unit tests
- Integration tests
- Documentation

### Screenshots

When you try to execute a singleton job multiple times while the initial job is running, you get an error like this in the UI
Cursor_and_Singleton_Job_-_Nautobot

PR #6431, opened 2024-10-28
implements 'is_singleton' property for jobs

### LINKED ISSUE #1004: Enforce number of concurrent Jobs running at the same time
### Environment
* Python version: 3.7
* Nautobot version:  1.1.3

### Proposed Functionality
As Austin the Network automation Engineer I want to be able to create a Nautobot Job that can be executed only once at a time. 
If a second execution of the same Job is requested while the Job is already running, the new Job should be queued until the first one finishes.

> there are few other options that could be interesting to explore but not sure what is possible within the existing Framework: terminate the existing job early and start the new one, allow N numbers of concurrent execution ...

### Use Case
Some Jobs are accessing shared resources and if multiple Jobs are running at the same time it could result in some inconsistency.

### Database Changes
Probably not

### External Dependencies
No

Kircheneer:
# Closes #1004

Contrary to the discussion in #1081 I have opted to error out duplicate jobs in favor of queueing them. This is because the main use case I have in mind is SSoT jobs where the desired behaviour is definitely to fail rather than queue, and also because queueing seems a lot more difficult and invasive to implement.

# What's Changed

Adds `is_singleton` meta flag to jobs including overrides.

# Screenshots

_Trying to run a singleton job when the lock is present (i.e. ideally another instance is already running):_

Screenshot 2024-11-11 at 08 05 28

_Showing functionality to override faulty locks:_

Screenshot 2024-11-11 at 08 06 26

# TODO
- [x] UI
- [x] Override functionality (?)
- [x] Unit, Integration Tests


nautobot/extras/jobs.py
```diff
@@ -187,6 +189,23 @@ def before_start(self, task_id, args, kwargs):
...
+                    extra={"object": self.job_model, "grouping": "initialization"},
+                )
+                raise RunJobTaskFailed(f"Job '{self.job_model}' is a singleton and already running.")
+            cache_parameters = {
+                "key": self.singleton_cache_key,
+                "value": 1,
+            }
+            if self.job_model.time_limit:
+                cache_parameters["timeout"] = self.job_model.time_limit
+            cache.set(**cache_parameters)
```
HanlinMiao, line 207: Sorry this might be a stupid question, but should this be:
```suggestion
            cache_parameters = {
                self.singleton_cache_key: {
                     "key": self.singleton_cache_key,
                     "value": 1,
                }
            }
            if self.job_model.time_limit:
                cache_parameters[self.singleton_cache_key]["timeout"] = self.job_model.time_limit
            cache.set(**cache_parameters)
```
Since we are trying to check if `self.singleton_cache_key` is in the cache or not on line 193?

HanlinMiao: Never mind, checked manually in `nbshell` and it works as expected.

nautobot/docs/development/jobs/index.md
```diff
@@ -148,6 +148,7 @@ As of Nautobot 2.2.3, the current list of reserved names (not including low-leve
 | `grouping`                | [module metadata property](#module-metadata-attributes) |
 | `has_sensitive_variables` | [metadata property](#has_sensitive_variables)           |
 | `hidden`                  | [metadata property](#hidden)                            |
+| `is_singleton`            | [metadata property](#is_singleton)                      |
```
glennmatthews, line 151: mind updating "As of Nautobot 2.2.3" above (line 128)?

nautobot/extras/forms/forms.py
```diff
@@ -1089,6 +1091,11 @@ class JobBulkEditForm(NautobotBulkEditForm):
         required=False,
         help_text="Default Job Queue the job runs on if no Job Queue is specified",
     )
+    is_singleton = forms.NullBooleanField(
+        required=False,
+        widget=BulkEditNullBooleanSelect,
+        help_text="Whether this job should be able to run only once",
```
glennmatthews, line 1097:
```suggestion
        help_text="Whether this job should fail to run if another instance of this job is already running",
```

nautobot/extras/forms/forms.py
```diff
@@ -1132,6 +1139,10 @@ class JobBulkEditForm(NautobotBulkEditForm):
         required=False,
         help_text="If checked, the values of has sensitive variables will be reverted to the default values defined in each Job's source code",
     )
+    is_singleton_override = forms.BooleanField(
+        required=False,
+        help_text="If checked, the values of has sensitive variables will be reverted to the default values defined in each Job's source code",
```
glennmatthews, line 1144:
```suggestion
        help_text="If checked, the values of is singleton will be reverted to the default values defined in each Job's source code",
```

There's got to be a better way to phrase these though :-)

nautobot/extras/jobs.py
```diff
@@ -189,6 +191,25 @@ def before_start(self, task_id, args, kwargs):
             )
             raise RunJobTaskFailed(f"Job {self.job_model} is not enabled to be run!")
 
+        if self.job_model.is_singleton:
+            if kwargs.get("_force_singleton_lock"):
+                cache.delete(self.singleton_cache_key)
```
glennmatthews, line 196: Feels off to me that this clears/resets the cache key. Why not just bypass the `is_running` check in this case instead?

nautobot/extras/jobs.py
```diff
@@ -486,6 +521,15 @@ def as_form(cls, data=None, files=None, initial=None, approval_view=False):
...
             help_text="Profiles the job execution using cProfile and outputs a report to /tmp/",
         )
+        # It is not clear to me why pylint thinks that this is a constant, the comparison and form rendering
+        # works correctly.
+        if cls.is_singleton:  # pylint: disable=using-constant-test
+            form.fields["_force_singleton"] = forms.BooleanField(
+                required=False,
+                initial=False,
+                label="Force singleton lock",
+                help_text="Force a singleton job to run even when its lock is currently placed",
```
glennmatthews, line 531: Perhaps:

```suggestion
                label="Ignore singleton lock",
                help_text="Allow this singleton job to run even when another instance is already running",
```

nautobot/extras/models/jobs.py
```diff
@@ -140,6 +140,8 @@ class Job(PrimaryModel):
         default=True, help_text="Whether this job contains sensitive variables"
     )
 
+    is_singleton = models.BooleanField(default=False, help_text="Whether this job should be able to run only once")
```
glennmatthews, line 143:
```suggestion
    is_singleton = models.BooleanField(default=False, help_text="Whether this job should fail to run if another instance of this job is already running")
```

nautobot/extras/tests/test_jobs.py
```diff
@@ -606,6 +606,20 @@ def test_job_profiling(self):
...
 
+    def test_job_singleton(self):
+        module = "singleton"
+        name = "TestSingletonJob"
+
+        job_class, _ = get_job_class_and_model(module, name, "local")
+        self.assertTrue(job_class.is_singleton)
+        create_job_result_and_run_job(module, name)
+        self.assertEqual(cache.get(job_class.singleton_cache_key), 1)
+        failed_job_result = create_job_result_and_run_job(module, name)
```
glennmatthews, line 617: I don't think this will work as a test, as `create_job_result_and_run_job` runs the requested job to completion before returning. Probably it would be sufficient as a test to simply set the cache key and then try to run the job once?

nautobot/extras/utils.py
```diff
@@ -579,6 +579,7 @@ def refresh_job_model_from_job_class(job_model_class, job_class, job_queue_class
                     "installed": True,
                     "enabled": False,
                     "default_job_queue": default_job_queue,
+                    "is_singleton": getattr(job_class.Meta, "is_singleton", False),
```
glennmatthews, line 582: Can't this just use `job_class.is_singleton`?

Kircheneer: Yes

nautobot/extras/tests/test_jobs.py
Kircheneer: Thank you, that explains why I was struggling to make it work.

Kircheneer:
There is one more problem I see. What if I start a job without a hard time limit (i.e. no timeout on the redis cache) and the worker crashes (i.e. the cache is never cleared). Should we default the time out to the global timeout?

glennmatthews:
Yes, there's always a global hard time limit in Celery so if the job doesn't have its own specific timeout we should use that.

Kircheneer:
Done!

changes/1004.added
```diff
@@ -0,0 +1 @@
+Added singleton job functionality to limit jobs to one concurrent execution.
```
glennmatthews, line 1:
```suggestion
Added singleton job functionality to limit specified jobs to one concurrent execution.
```

nautobot/extras/jobs.py
```diff
@@ -486,6 +520,15 @@ def as_form(cls, data=None, files=None, initial=None, approval_view=False):
             label="Profile job execution",
             help_text="Profiles the job execution using cProfile and outputs a report to /tmp/",
         )
+        # It is not clear to me why pylint thinks that this is a constant, the comparison and form rendering
+        # works correctly.
+        if cls.is_singleton:  # pylint: disable=using-constant-test
+            form.fields["_ignore_singleton"] = forms.BooleanField(
```
glennmatthews, line 526: Just checking - I see a mix of `_ignore_singleton` and `_ignore_singleton_lock` in this code, and it's not immediately obvious to me where one gets translated to the other. Should it be `ignore_singleton_lock` throughout?

nautobot/extras/tests/test_jobs.py
```diff
@@ -606,6 +606,19 @@ def test_job_profiling(self):
...
+        name = "TestSingletonJob"
+
+        job_class, _ = get_job_class_and_model(module, name, "local")
+        self.assertTrue(job_class.is_singleton)
+        cache.set(job_class.singleton_cache_key, 1)
+        failed_job_result = create_job_result_and_run_job(module, name)
+
+        self.assertEqual(
+            failed_job_result.status, JobResultStatusChoices.STATUS_FAILURE, msg="Duplicate singleton job didn't error."
+        )
```
glennmatthews, line 620: Would be good if possible to also add a test for running the job with `_ignore_singleton[_lock]` and confirm that it then runs and clears the flag.

glennmatthews (commented):
Looking good!

nautobot/extras/tests/test_jobs.py
```diff
@@ -606,6 +606,32 @@ def test_job_profiling(self):
...
+        name = "TestSingletonJob"
+
+        job_class, _ = get_job_class_and_model(module, name, "local")
+        self.assertTrue(job_class.is_singleton)
+        cache.set(job_class.singleton_cache_key, 1)
+        failed_job_result = create_job_result_and_run_job(module, name, _ignore_singleton_lock=True)
+
+        self.assertEqual(
+            failed_job_result.status, JobResultStatusChoices.STATUS_SUCCESS, msg="Duplicate singleton job didn't error."
+        )
```
glennmatthews, line 633: Can/should we also assert that the cache lock has been cleared?

nautobot/extras/test_jobs/singleton.py
```diff
@@ -0,0 +1,16 @@
...
+from nautobot.extras.jobs import Job
+
+
+class TestSingletonJob(Job):
+    class Meta:
+        name = "Example Singleton Job"
+        is_singleton = True
+
+    def run(self):
+        time.sleep(60)
```
glennmatthews, line 13: Maybe reduce this number somewhat, or make it configurable as a job input, so that running the added tests don't take a minute each?

nautobot/extras/tests/test_jobs.py
```diff
@@ -606,6 +606,32 @@ def test_job_profiling(self):
...
+        module = "singleton"
+        name = "TestSingletonJob"
+
+        job_class, _ = get_job_class_and_model(module, name, "local")
+        self.assertTrue(job_class.is_singleton)
+        cache.set(job_class.singleton_cache_key, 1)
+        failed_job_result = create_job_result_and_run_job(module, name, _ignore_singleton_lock=True)
+
+        self.assertEqual(
+            failed_job_result.status, JobResultStatusChoices.STATUS_SUCCESS, msg="Duplicate singleton job didn't error."
```
glennmatthews, line 632:
```suggestion
        passed_job_result = create_job_result_and_run_job(module, name, _ignore_singleton_lock=True)

        self.assertEqual(
            passed_job_result.status, JobResultStatusChoices.STATUS_SUCCESS, msg="Duplicate singleton job didn't succeed with _ignore_singleton_lock=True."
```

nautobot/extras/models/jobs.py
```diff
@@ -864,6 +877,7 @@ def enqueue_job(
             "nautobot_job_job_model_id": job_model.id,
             "nautobot_job_profile": profile,
             "nautobot_job_user_id": user.id,
+            "nautobot_job_ignore_singleton_lock": ignore_singleton_lock,
```
glennmatthews, line 880: Many of the other `nautobot_job_*` celery kwargs get handled in `nautobot/core/celery/backends.py` and `nautobot/core/celery/schedulers.py`. Do we need similar code added for this parameter?

nautobot/extras/constants.py
```diff
@@ -30,6 +30,7 @@
     "soft_time_limit",
     "time_limit",
     "has_sensitive_variables",
+    "is_singleton",
```
glennmatthews, line 33: Do we need to add `nautobot_job_ignore_singleton_lock` to `JOB_RESULT_CUSTOM_CELERY_KWARGS` below?

nautobot/extras/models/jobs.py
```diff
@@ -807,6 +816,7 @@ def enqueue_job(
         task_queue=None,
         job_result=None,
         synchronous=False,
+        ignore_singleton_lock=False,
```
glennmatthews, line 819: Do we need to similarly enhance the `create_schedule` method to allow creating scheduled jobs that ignore the singleton?

nautobot/extras/tests/test_jobs.py
```diff
@@ -606,6 +606,32 @@ def test_job_profiling(self):
...
+        )
+
+    def test_job_ignore_singleton(self):
+        module = "singleton"
+        name = "TestSingletonJob"
+
+        job_class, _ = get_job_class_and_model(module, name, "local")
+        self.assertTrue(job_class.is_singleton)
+        cache.set(job_class.singleton_cache_key, 1)
+        failed_job_result = create_job_result_and_run_job(module, name, _ignore_singleton_lock=True)
```
glennmatthews, line 629: Would be good to also enhance the `test_rerun_job` test in `test_views.py` to make sure that a job rerun correctly inherits the `_ignore_singleton_lock` setting from the original job run (i.e. the code you added below in `views.py`).

nautobot/extras/tests/test_jobs.py
Kircheneer: We had discussed some time earlier in this MR that we shouldn't clear the lock, but just ignore it, when we have the flag set. This is the current behaviour. Do you think we should change this again?

glennmatthews: Ah sorry, wasn't meaning to be inconsistent :-) Doesn't it clear the lock at the end of the job though in any case?

Kircheneer: That's of course correct. Will add this.

nautobot/extras/models/jobs.py
Kircheneer: I'm afraid I don't quite understand the consequences here.

- Why would a scheduled job behave differently in this regard?
- What is the purpose of `backends.py`?

nautobot/extras/models/jobs.py
Kircheneer: I just tested it manually and it works both way, i.e. the job fails as expected when one is already running and correctly overrides the lock if the checkbox is set. This appears to happen because this option is passed along via the `kwargs` field on `ScheduledJob`. I am happy to change it such that it uses `celery_kwargs` instead if desired, I just wanted to check in with you because as of now it already work.

```
>>> sj = ScheduledJob.objects.get(name="TEST")
>>> sj.celery_kwargs
{'queue': None, 'nautobot_job_profile': False}
>>> sj.kwargs
{'_ignore_singleton_lock': True}
```

nautobot/extras/constants.py
Kircheneer: See [this](https://github.com/nautobot/nautobot/pull/6431/files/1ff1a2ea39568fb057923c2b2f2e9d76df88b071#r1849663502) comment.

nautobot/extras/models/jobs.py
Kircheneer: Oh, I guess `kwargs` is supposed to be the job arguments that get passed to `run`, is that accurate?

nautobot/extras/tests/test_jobs.py
Kircheneer: I will tackle this once the other conversation is resolved.

nautobot/extras/models/jobs.py
glennmatthews: Yeah... it looks like keys in `kwargs` that don't match the Job's input variables get discarded by `deserialize_data()` so I think that's why it's working presently. But it might be cleaner to follow the pattern of `nautobot_job_profile` and store it in the `celery_kwargs` instead?

nautobot/extras/models/jobs.py
glennmatthews: `NautobotDatabaseBackend` handles the automatic update of JobResult records. I think if we move `ignore_singleton_lock` into `celery_kwargs` it should be fine there since we're already passing around `celery_kwargs` in that code.

nautobot/extras/constants.py
Kircheneer: Added as per discussion.

nautobot/extras/models/jobs.py
Kircheneer: Added as per discussion.

nautobot/extras/tests/test_jobs.py
Kircheneer: Done!

Kircheneer:
Alright all comments addressed, had to introduce some new logic in `BaseJob.as_form` to make the form work for jobs that are singletons only by overriding.

nautobot/extras/jobs.py
```diff
@@ -189,6 +191,24 @@ def before_start(self, task_id, args, kwargs):
             )
             raise RunJobTaskFailed(f"Job {self.job_model} is not enabled to be run!")
 
+        if self.job_model.is_singleton and not self.celery_kwargs.get("nautobot_job_ignore_singleton_lock", False):
```
glennmatthews, line 194: In the case where the job *is* a singleton *and* the `is_running` key is set, *but* `ignore_singleton_lock` is set, I wonder if it would be good to log a log entry to the effect of "Ignoring singleton lock because ignore_singleton_lock is set"?

nautobot/extras/jobs.py
```diff
@@ -189,6 +191,24 @@ def before_start(self, task_id, args, kwargs):
...
+                    extra={"object": self.job_model, "grouping": "initialization"},
+                )
+                raise RunJobTaskFailed(f"Job '{self.job_model}' is a singleton and already running.")
+            cache_parameters = {
+                "key": self.singleton_cache_key,
+                "value": 1,
+                "timeout": settings.CELERY_TASK_TIME_LIMIT,
+            }
+            if self.job_model.time_limit:
+                cache_parameters["timeout"] = self.job_model.time_limit
```
glennmatthews, line 209: Maybe?

```suggestion
                "timeout": self.job_model.time_limit or settings.CELERY_TASK_TIME_LIMIT,
            }
```

nautobot/extras/tests/test_jobs.py
```diff
@@ -686,6 +686,38 @@ def test_job_profiling(self):
...
+        self.assertTrue(job_class.is_singleton)
+        cache.set(job_class.singleton_cache_key, 1)
+        passed_job_result = create_job_result_and_run_job(
+            module, name, celery_kwargs={"nautobot_job_ignore_singleton_lock": True}
+        )
+
+        self.assertEqual(
+            passed_job_result.status,
+            JobResultStatusChoices.STATUS_SUCCESS,
+            msg="Duplicate singleton job didn't succeed with _ignore_singleton_lock=True.",
```
glennmatthews, line 717:
```suggestion
            msg="Duplicate singleton job didn't succeed with nautobot_job_ignore_singleton_lock=True.",
```

nautobot/extras/test_jobs/singleton.py
```diff
@@ -0,0 +1,16 @@
...
+from nautobot.extras.jobs import Job
+
+
+class TestSingletonJob(Job):
+    class Meta:
+        name = "Example Singleton Job"
+        is_singleton = True
+
+    def run(self):
+        time.sleep(10)
```
glennmatthews, line 13: Can we change this to a `1` to save a bit more test run time in `test_job_ignore_singleton`?

nautobot/extras/tests/test_jobs.py
```diff
@@ -686,6 +686,38 @@ def test_job_profiling(self):
...
+
+        job_class, _ = get_job_class_and_model(module, name, "local")
+        self.assertTrue(job_class.is_singleton)
+        cache.set(job_class.singleton_cache_key, 1)
+        failed_job_result = create_job_result_and_run_job(module, name)
+
+        self.assertEqual(
+            failed_job_result.status, JobResultStatusChoices.STATUS_FAILURE, msg="Duplicate singleton job didn't error."
+        )
+        self.assertEqual(cache.get(job_class.singleton_cache_key, None), None)
```
glennmatthews, line 701:
```suggestion
        self.assertIsNone(cache.get(job_class.singleton_cache_key, None))
```

nautobot/extras/tests/test_jobs.py
```diff
@@ -686,6 +686,38 @@ def test_job_profiling(self):
...
+        passed_job_result = create_job_result_and_run_job(
+            module, name, celery_kwargs={"nautobot_job_ignore_singleton_lock": True}
+        )
+
+        self.assertEqual(
+            passed_job_result.status,
+            JobResultStatusChoices.STATUS_SUCCESS,
+            msg="Duplicate singleton job didn't succeed with _ignore_singleton_lock=True.",
+        )
+        self.assertEqual(cache.get(job_class.singleton_cache_key, None), None)
```
glennmatthews, line 719:
```suggestion
        self.assertIsNone(cache.get(job_class.singleton_cache_key, None))
```

glennmatthews (approved):

nautobot/extras/jobs.py
Kircheneer: Good idea, done

nautobot/extras/test_jobs/singleton.py
Kircheneer: Works, done!

glennmatthews (approved):
Sorry, one last request - can you add this feature to the release overview in `docs/release-notes/index.md` and `docs/release-notes/version-2.4.md`? It's significant and deserves attention. :-)

Kircheneer:
> can you add this feature to the release overview

Sure! I am glad to. 👍🏻

PR #6440, opened 2024-10-28
Performance improvements for custom-field provisioning and web_request_context

### LINKED ISSUE #5834: A long running custom field cleanup task will fail
### Environment
* Nautobot version: 2.2.4
* Python version: 3.8.10
* Database platform, version: PostgreSQL

### Steps to Reproduce
1. Create several custom fields that apply to a large number of objects (10s of thousands)
2. Populate the custom fields with data
3. Delete the custom fields

The custom fields and data should be deleted and removed from forms and views.

The cleanup task fails and not all custom fields are removed. This is possibly due to the task timing out due to the large number of objects that need to be updated. Even after running `fix_custom_fields` to correct the problem, the custom fields remain in the list view configure panel options.

### LINKED ISSUE #6421: Custom Field lock's out model instances
### Environment
* Python version: 3.11
* Middleware(s): N/A

### Steps to Reproduce

1. Add a custom field (this will run a sytem job) 
2. Before that system job completes, update one of the model instances that is being affected by the system job
3. Ensure system job is set to be longer than the uwsgi timeout. 

### Expected Behavior

Everything to work without dlay

### Observed Behavior

You get a gateway timeout

glennmatthews:
# Closes #6421
# Closes #5834 
# What's Changed

- Increased soft/hard time limits on CustomField provisioning background tasks to 1800 and 2000 seconds respectively. (fixes #5834)
- Fixed long-running-at-scale transaction lock on records while adding/removing a CustomField definition. (fixes #6421)
    - Add `JSONSet` and `JSONRemove` database helpers based on a currently-in-flight upstream Django PR, and use these to efficiently provision/deprovision specific keys on `_custom_field_data` on all impacted objects in a single query.
        - I don't currently have an equivalent function for the specific case of "Change the defined choices for a TYPE_MULTI_SELECT custom field" as that involves changing members of a list rather than just setting/deleting a single key/value pair, so for now that one will remain relatively inefficient/slow.
    - Instead of one combined `transaction.atomic()`/`web_request_context()` block that covers both updating the `_custom_field_data` on every affected record (quick) and generating the corresponding ObjectChange records (slow), we now use the efficient single query to update the `_custom_field_data` very quickly, then a separate transaction (via `deferred_change_logging_for_bulk_operation()`) to generate the ObjectChanges, avoiding the long-running lock on the object records that was reported in #6421.
- Added logging to JobResults for CustomField provisioning background tasks.

Additionally, before settling on the above approach, I went down a few rabbit holes investigating the *large* number of database queries being executed in the original implementation of these tasks, which led to the following fixes:

- Added cacheable `CustomField.objects.keys_for_model(model)` API. This makes the API serializer calls involved in creating the ObjectChange records more efficient at scale, because we can hit the database once to get the CF keys, then all subsequent calls for other instances of the same model will just be a cache hit instead of another query.
- Added queryset caching in `web_request_context` for more efficient JobHook and Webhook dispatching on bulk requests.If we're processing, say, a web_request_context that updated 1000 Manufacturers, we can look up the relevant JobHook and Webhook records for the dcim.manufacturer content-type once instead of repeating 1000 times.
- Added appropriate `select_related` to the ObjectChange queryset used in `web_request_context`.

# Screenshots

Tested manually by repeatedly creating, updating, and deleting CustomFields that applied to Manufacturer and Status models while having 100000 Manufacturer records and several hundred Status records. Even in the Manufacturer case, the UI is responsive, and the background task runs to completion in about 5 minutes (5 seconds for the efficient `_custom_field_data` update, and the remaining time spent generating 100000 ObjectChange records atomically).

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/jobs.py
+++ b/nautobot/extras/jobs.py
@@ -1154,7 +1154,7 @@ def enqueue_job_hooks(object_change, may_reload_jobs=True, jobhook_queryset=None
         jobhook_queryset (QuerySet): Previously retrieved set of JobHooks to potentially enqueue
 
     Returns:
-        result (tuple): whether Jobs were reloaded here, and the jobhooks that were considered
+        result (tuple[bool, QuerySet]): whether Jobs were reloaded here, and the jobhooks that were considered
     """
     jobs_reloaded = False
 
--- a/nautobot/extras/jobs.py
+++ b/nautobot/extras/jobs.py
@@ -1144,41 +1144,47 @@ def run_job(self, job_class_path, *args, **kwargs):
         raise
 
 
-def enqueue_job_hooks(object_change, may_reload_jobs=True):
+def enqueue_job_hooks(object_change, may_reload_jobs=True, jobhook_queryset=None):
     """
     Find job hook(s) assigned to this changed object type + action and enqueue them to be processed.
 
+    Args:
+        object_change (ObjectChange): The change that may trigger JobHooks to execute.
+        may_reload_jobs (bool): Whether to reload JobHook source code from disk to guarantee up-to-date code.
+        jobhook_queryset (QuerySet): Previously retrieved set of JobHooks to potentially enqueue
+
     Returns:
-        jobs_reloaded (bool): whether Jobs were reloaded to make this happen
+        result (tuple): whether Jobs were reloaded here, and the jobhooks that were considered
     """
     jobs_reloaded = False
 
     # Job hooks cannot trigger other job hooks
     if object_change.change_context == ObjectChangeEventContextChoices.CONTEXT_JOB_HOOK:
-        return jobs_reloaded
+        return jobs_reloaded, jobhook_queryset
 
     # Determine whether this type of object supports job hooks
     content_type = object_change.changed_object_type
     if content_type not in change_logged_models_queryset():
-        return jobs_reloaded
+        return jobs_reloaded, jobhook_queryset
 
     # Retrieve any applicable job hooks
-    action_flag = {
-        ObjectChangeActionChoices.ACTION_CREATE: "type_create",
-        ObjectChangeActionChoices.ACTION_UPDATE: "type_update",
-        ObjectChangeActionChoices.ACTION_DELETE: "type_delete",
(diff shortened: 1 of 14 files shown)


nautobot/extras/jobs.py
```diff
@@ -1144,41 +1144,47 @@ def run_job(self, job_class_path, *args, **kwargs):
...
     Find job hook(s) assigned to this changed object type + action and enqueue them to be processed.
 
+    Args:
+        object_change (ObjectChange): The change that may trigger JobHooks to execute.
+        may_reload_jobs (bool): Whether to reload JobHook source code from disk to guarantee up-to-date code.
+        jobhook_queryset (QuerySet): Previously retrieved set of JobHooks to potentially enqueue
+
     Returns:
-        jobs_reloaded (bool): whether Jobs were reloaded to make this happen
+        result (tuple): whether Jobs were reloaded here, and the jobhooks that were considered
```
mateusz-lachowski-codilime, line 1157:
```suggestion
        result (tuple[bool, QuerySet]): whether Jobs were reloaded here, and the jobhooks that were considered
```

mateusz-lachowski-codilime:
Is it worth to save your manual testing path and steps as some jupyter notebook or just list of python commands for future reference?

mateusz-lachowski-codilime (approved):

glennmatthews:
> Is it worth to save your manual testing path and steps as some jupyter notebook or just list of python commands for future reference?

There's not a whole lot more to say about it - the only Python command was bulk creating the records through nbshell, a la:

```py
for i in range(100000):
    Manufacturer.objects.create(name=f"Manufacturer {i}")
```

and the actual create/update/delete of various arbitrary CustomField records was done through the UI. I generally had three tabs open:

- Manufacturer list view, so as to confirm that custom field data was correctly updated, and also to edit individual Manufacturers to confirm that they were not blocked while the background task was running.
- Custom field create/list/detail/delete view, to take various test actions.
- JobResult list/detail view to inspect the status and progress of the background task(s).

PR #6446, opened 2024-10-29
Add Controller Capabilities field

smk4664:
# What's Changed

Add a capabilities field onto Controller and ControllerManagedDeviceGroup. This is to support the collapse of AccessPointGroup into the ControllerManagedDeviceGroup.

This also adds Controller and ControllerManagedDeviceGroup to the Wireless Navigation Menu Group. These needed to be filtered lists, so this PR includes a new feature for `query_params` on the NavMenuItem and NavMenuButton.
The nav_menu.js was updated to strip the search parameter when comparing the current page to the clicked link. This allows the nav_bar to keep the Wireless Controllers or Access Point Group highlighted when navigating tabs on the Controller or ControllerManagedDeviceGroup object.

In addition, this pr changes array_to_string to support all JSONArrayFields, not just the Numeric Arrays.

# Screenshots

Controllers (not filtered)
image

Wireless Controllers (Controllers filtered on capabilities='wireless')
image

Access Point Group (Controller Managed Device Group filtered on capabilities='wireless')
image

Controller Detail
image

Controller Managed Device Group Detail
image

# TODO

A followup-PR is to remove the existing AccessPointGroup and add their fields to the ControllerManagedDeviceGroup.

- [x] Unit, Integration Tests


nautobot/core/ui/nav.py
```diff
@@ -153,7 +155,9 @@ def fixed_fields(self) -> tuple:
     args = []
     kwargs = {}
 
-    def __init__(self, link, name, args=None, kwargs=None, permissions=None, buttons=(), weight=1000):
+    def __init__(
+        self, link, name, args=None, kwargs=None, query_params=None, permissions=None, buttons=(), weight=1000
```
glennmatthews, line 159: Please add `query_params` to the docstring.

nautobot/core/ui/nav.py
```diff
@@ -215,6 +221,7 @@ def __init__(
         button_class=ButtonActionColorChoices.DEFAULT,
         permissions=None,
         weight=1000,
+        query_params=None,
```
glennmatthews, line 224: Please add to the docstring.

nautobot/dcim/choices.py
```diff
@@ -1578,3 +1578,14 @@ class VirtualDeviceContextStatusChoices(ChoiceSet):
...
+
+#
+# Controllers
+#
+
+
+class ControllerCapabilitiesChoices(ChoiceSet):
+    WIRELESS = "wireless"
+
+    CHOICES = ((WIRELESS, "Wireless"),)
```
glennmatthews, line 1591: Do we have a clear picture at this time what other capabilities might logically be added to this choiceset?

nautobot/dcim/models/devices.py
```diff
@@ -1425,6 +1432,10 @@ def clean(self):
                     {"location": f'Devices may not associate to locations of type "{self.location.location_type}".'}
                 )
 
+    @property
+    def capabilities_list(self):
+        return array_to_string(self.capabilities)
```
glennmatthews, line 1437: Feels weird to me to have a property called `_list` that returns a string :) Would `capabilities_string` be a better name?

nautobot/dcim/choices.py
smk4664: We do not at this time, though John would like this to be a Choice field for future implementations. But this is a good field to discuss as Cristian would like for an App to be able to extend the choices.

nautobot/dcim/models/devices.py
smk4664: Yes it does, though this is a pattern that RackReservation.unit_list and Service.port_list have.

smk4664:
I pulled in Next to try to fix the InterfaceTest case that was failing for me, which did not fix the issue.

nautobot/core/ui/nav.py
```diff
@@ -195,6 +201,7 @@ def initial_dict(self) -> dict:
             "weight": self.weight,
             "buttons": {},
             "permissions": self.permissions,
+            "querystring": self.querystring,
```
gsnider2195, line 204: Does `initial_dict` need to be updated for `NavMenuItem` too?

nautobot/dcim/filters/__init__.py
```diff
@@ -1881,6 +1888,12 @@ class ControllerManagedDeviceGroupFilterSet(NautobotFilterSet):
             "name": "icontains",
         }
     )
+    capabilities = django_filters.MultipleChoiceFilter(
+        choices=ControllerCapabilitiesChoices,
+        null_value=None,
+        lookup_expr="icontains",
```
gsnider2195, line 1894: I believe MultipleChoiceFilters get the automatically created `__ic` filter. If so we should remove the `lookup_expr` from the default filter?

nautobot/dcim/models/devices.py
```diff
@@ -1456,6 +1467,11 @@ class ControllerManagedDeviceGroup(TreeModel, PrimaryModel):
         null=False,
         help_text="Controller that manages the devices in this group",
     )
+    capabilities = JSONArrayField(
+        base_field=models.CharField(choices=ControllerCapabilitiesChoices),
+        blank=True,
+        null=True,
+    )
```
gsnider2195, line 1474: Was there a reason this wasn't being inherited from the controller? Was it because we wanted to support a use case where a controller supports "X and Y" but the managed group only supports "X"? Have we identified any controllers that would have multiple capabilities?

nautobot/dcim/models/devices.py
```diff
@@ -1425,6 +1432,10 @@ def clean(self):
                     {"location": f'Devices may not associate to locations of type "{self.location.location_type}".'}
                 )
 
+    @property
+    def capabilities_string(self):
+        return array_to_string(self.capabilities)
```
gsnider2195, line 1437: This seems like it's duplicating the django pattern of [`get_FOO_display`](https://docs.djangoproject.com/en/4.2/ref/models/instances/#django.db.models.Model.get_FOO_display)? I'm not sure if this would be relevant to the `JSONArrayField` with choices but it's worth looking into.

nautobot/dcim/tests/test_api.py
```diff
@@ -688,9 +688,10 @@ def setUpTestData(cls):
             status=statuses[0],
         )
         # Place a device in Rack 4
-        device = Device.objects.filter(
-            location=populated_rack.location, rack__isnull=True, device_type__u_height__gt=0
-        ).first()
+        device = Device.objects.filter(location=populated_rack.location, rack__isnull=True).first()
+        # Ensure the device height is 1U
```
gsnider2195, line 692: What problem is this trying to prevent? Does it have to be 1 or does it just need to be non-zero? If that's the case we should say that in this comment.

examples/example_app/example_app/navigation.py
```diff
@@ -26,6 +26,12 @@
                             ),
                         ),
                     ),
+                    NavMenuItem(
+                        link="plugins:example_app:examplemodel_list",
+                        name="Example Models filtered",
+                        permissions=["example_app.view_examplemodel"],
+                        query_params={"number": 100},
```
gsnider2195, line 33:
```suggestion
                        query_params={"number": "100"},
```

nautobot/docs/user-guide/core-data-model/dcim/controller.md
```diff
@@ -54,7 +54,7 @@ devices.count()
 
 ## Base Fields
 
-A controller is identified by a unique `name` and can include a `description`.
+A controller is identified by a unique `name` and can include a `description` and `capabilities`.
```
gsnider2195, line 57:
```suggestion
A controller is identified by a unique `name` and can include a `description` and optionally one or more `capabilities`.
```

nautobot/extras/factory.py
```diff
@@ -167,6 +167,15 @@ def jobs(self, create, extracted, **kwargs):
             job.validated_save()
 
 
+class JobQueueFactoryWithTenancy(JobQueueFactory):
+    class Meta:
+        model = JobQueue
+
+    class Params:
+        has_description = NautobotBoolIterator()
+        has_tenant = NautobotBoolIterator(chance_of_getting_true=100)
```
gsnider2195, line 176: Bad merge conflict? This isn't relevant to your PR is it?

nautobot/project-static/js/nav_menu.js
```diff
@@ -237,8 +237,9 @@ document.addEventListener('DOMContentLoaded', function() {
     if (activeLink) {
         let previouslyClickedLink = document.querySelector('.navbar-fixed-left .navbar-nav > .dropdown > .nav-dropdown-menu > li > a[href="' + activeLink + '"]');
         let currentLocation = window.location.pathname + window.location.search;
+        let previouslyClickedLinkNoSearch = previouslyClickedLink.getAttribute('href').split('?')[0];
 
-        if (previouslyClickedLink && currentLocation.includes(previouslyClickedLink.getAttribute('href'))) {
+        if (previouslyClickedLink && (currentLocation.includes(previouslyClickedLink.getAttribute('href')) || currentLocation.includes(previouslyClickedLinkNoSearch))) {
```
gsnider2195, line 242: Is this a bad merge conflict resolution or is this required for your PR?

nautobot/core/models/utils.py
```diff
@@ -16,10 +16,15 @@
...
-    Generate an efficient, human-friendly string from a set of integers. Intended for use with ArrayField.
+    Generate an efficient, human-friendly string from a set of integers or strings. Intended for use with ArrayField.
     For example:
         [0, 1, 2, 10, 14, 15, 16] => "0-2, 10, 14-16"
+        ["foo", "bar", "baz"] => "foo, bar, baz"
     """
+    if not array:
+        return ""
+    if isinstance(array[0], str):
+        return ", ".join(array)
```
gsnider2195, line 27: I don't think this is the right place for this change. Even though the name of this function suggests it's intended for converting arrays to strings it contains more specific logic for converting consecutive integers. Also, unless we have a use case for this to be used on arrays that could either be integers OR strings and we don't know ahead of time which one will be used, lets go ahead and just do the `", ".join(array)` on the fields where it's appropriate.

nautobot/extras/factory.py
smk4664: This either was me trying to fix an issue with JobQueue tests that often failed due to missing Tenant filter, or a bad merge. I don't remember so I will remove.

nautobot/dcim/tests/test_api.py
smk4664: Just non-zero, but in all the factories, it only set it to 0, 1 or 2.  I thought I would just pick one.

smk4664: Just non-zero, but in all the factories, it only set it to 1 or 2.  I thought I would just pick one.

nautobot/dcim/models/devices.py
smk4664: Right now we only have 1 targeted capability. @lampwins Would you like to weigh in?

nautobot/dcim/filters/__init__.py
smk4664: They are automatically created and don't work as MultipleChoiceFilter doesn't support Arrays. Out of all the automatically created filters, the only one that worked was the `icontains`.

nautobot/core/ui/nav.py
smk4664: It wouldn't hurt. Only the NavMenuButton was utilizing the Initial dict during rendering. The logic for NavMenuItem was handled by different code.

nautobot/project-static/js/nav_menu.js
smk4664: Required.  This keeps the NavMenuItem focused (in orange) in the navbar when navigating to a detail view from the list view. This is because the Detail view will drop the `search` string.

nautobot/wireless/tables.py
```diff
@@ -104,7 +105,7 @@ def render_channel_width(self, value):
         return ", ".join(f"{v}MHz" for v in value)
 
     def render_allowed_channel_list(self, value):
-        return ", ".join(f"{v}" for v in value)
+        return array_to_string(value)
```
smk4664, line 108:
```suggestion
        return ", ".join(f"{v}" for v in value)
```

nautobot/wireless/tables.py
```diff
@@ -278,7 +279,7 @@ def render_channel_width(self, value):
...
```
smk4664, line 282:
```suggestion
        return ", ".join(f"{v}" for v in value)
```

nautobot/dcim/tables/devices.py
```diff
@@ -1421,9 +1424,14 @@ class Meta(BaseTable.Meta):
             "platform",
             "role",
             "tenant",
+            "capabilities",
             "actions",
         )
 
+    def render_capabilities(self, value):
+        """Render capabilities."""
+        return array_to_string(value)
```
smk4664, line 1433:
```suggestion
        return ", ".join(f"{v}" for v in value)
```

nautobot/dcim/tables/devices.py
```diff
@@ -1459,10 +1469,15 @@ class Meta(BaseTable.Meta):
...
             "controller",
             "weight",
+            "capabilities",
             "tags",
             "actions",
         )
 
+    def render_capabilities(self, value):
+        """Render capabilities."""
+        return array_to_string(value)
```
smk4664, line 1479:
```suggestion
        return ", ".join(f"{v}" for v in value)
```

nautobot/dcim/filters/__init__.py
gsnider2195: If all of the automatically created filters are broken we either need to fix or remove them

nautobot/core/models/utils.py
```diff
@@ -16,9 +16,10 @@
 
 def array_to_string(array):
     """
-    Generate an efficient, human-friendly string from a set of integers. Intended for use with ArrayField.
+    Generate an efficient, human-friendly string from a set of integers or strings. Intended for use with ArrayField.
     For example:
         [0, 1, 2, 10, 14, 15, 16] => "0-2, 10, 14-16"
+        ["foo", "bar", "baz"] => "foo, bar, baz"
```
gsnider2195, line 22: revert docstring change

nautobot/dcim/models/devices.py
```diff
@@ -1425,6 +1431,9 @@ def clean(self):
                     {"location": f'Devices may not associate to locations of type "{self.location.location_type}".'}
                 )
 
+    def get_capabilities_display(self):
+        return ", ".join(f"{v}" for v in self.capabilities)
```
gsnider2195, line 1435: Shouldn't this just be
```suggestion
        return ", ".join(self.capabilities)
```

nautobot/dcim/models/devices.py
```diff
@@ -1480,6 +1494,9 @@ def clean(self):
                 {"controller": "Controller device group must have the same controller as the parent group."}
             )
 
+    def get_capabilities_display(self):
+        return ", ".join(f"{v}" for v in self.capabilities)
```
gsnider2195, line 1498:
```suggestion
        return ", ".join(self.capabilities)
```

nautobot/dcim/tables/devices.py
```diff
@@ -1421,9 +1423,14 @@ class Meta(BaseTable.Meta):
             "platform",
             "role",
             "tenant",
+            "capabilities",
             "actions",
         )
 
+    def render_capabilities(self, value):
+        """Render capabilities."""
+        return ", ".join(f"{v}" for v in value)
```
gsnider2195, line 1432:
```suggestion
        return ", ".join(value)
```

nautobot/dcim/tables/devices.py
```diff
@@ -1459,10 +1468,15 @@ class Meta(BaseTable.Meta):
...
             "controller",
             "weight",
+            "capabilities",
             "tags",
             "actions",
         )
 
+    def render_capabilities(self, value):
+        """Render capabilities."""
+        return ", ".join(f"{v}" for v in value)
```
gsnider2195, line 1478:
```suggestion
        return ", ".join(value)
```

nautobot/dcim/tables/devices.py
```diff
@@ -1421,9 +1423,14 @@ class Meta(BaseTable.Meta):
             "platform",
             "role",
             "tenant",
+            "capabilities",
             "actions",
         )
 
+    def render_capabilities(self, value):
+        """Render capabilities."""
+        return ", ".join(value)
```
glennmatthews, line 1432: Since these are fixed choices, it *might* be nice to render them similar to how we render multiselect custom fields (CustomFieldColumn in nautobot/core/tables.py), i.e. as `<span class="label label-default">{value}</span>`.

glennmatthews (approved):

Issue #6441, opened 2024-10-29
Position field UI handling incorrect


lampwins:
### Environment
* Nautobot version (Docker tag too if applicable): 2.3.9
* Python version: 3.11
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Navigate to the device edit form
2.
3.

### Expected Behavior
The select2 drop downs should function normally

### Observed Behavior
Most are broken, with console errors pointing to unsafe handling of any field referencing "position," in this case the rack position. This was added for module position handing in #6419

PR #6449, opened 2024-10-29
Add integration test for creation of a device and all prerequisite models

gsnider2195:
# Closes #DNE
# What's Changed

This is an integration test to hopefully prevent issues like #6441 from being introduced again.

This also fixes a bug in the existing navbar integration tests where we're doing an incorrect `assertTrue(bool("true"))` which would also pass for the negative case of `assertTrue(bool("false"))`

# Screenshots

https://github.com/user-attachments/assets/8a283cab-eaf8-4e35-bd4b-2882f927d4fb

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/tests/integration/test_create_device.py
+++ b/nautobot/dcim/tests/integration/test_create_device.py
@@ -79,6 +79,7 @@ class CreateDeviceTestCase(SeleniumTestCase):
         - Device Type
         - LocationType
         - Location
+        - Role
         - Device
 
         """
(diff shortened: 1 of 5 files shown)


nautobot/dcim/tests/integration/test_create_device.py
```diff
@@ -0,0 +1,140 @@
...
+
+from nautobot.core.testing.integration import SeleniumTestCase
+
+
+class CreateDeviceTestCase(SeleniumTestCase):
+    """
+    Create a device and all pre-requisite objects through the UI.
+    """
+
+    def _click_nav_bar(self, parent_menu_name, child_menu_name):
```
glennmatthews, line 12: Might make sense for this (and maybe some of the other helpers below) to be added to `SeleniumTestCase` so that we can later refactor some of the other integration tests to use them?

glennmatthews (commented):
Thanks for adding this!

nautobot/dcim/tests/integration/test_create_device.py
gsnider2195: That sounds good to me. I'm happy with how these helper functions turned out

nautobot/dcim/tests/integration/test_create_device.py
```diff
@@ -87,52 +24,61 @@ def test_create_device(self):
...
-        self._click_add_button()
+        self.click_list_view_add_button()
         self.assertEqual(self.browser.url, self.live_server_url + reverse("dcim:device_add"))
         self.browser.fill("name", "Test Device Integration Test 1")
-        self._fill_select2_field("role", "")  # pick first role
-        self._fill_select2_field("device_type", "Test Device Type 1")
-        self._fill_select2_field("location", "Test Location 1")
-        self._fill_select2_field("status", "")  # pick first status
-        self._click_create_button()
+        self.fill_select2_field("role", "")  # pick first role
```
glennmatthews, line 77: Might as well use the `Test Role 1`?

glennmatthews (approved):

PR #6450, opened 2024-10-30
Smarter display of LinkedCountColumn

### LINKED ISSUE #6231: Prefix summary table shows locations count in table. Previous 1:1 behaviour displayed actual location name
### As ...

Ozzie - Operator

### I want ...

An easy way of seeing at a glance which prefixes are assigned to which location in a summarised table. New behaviour now that prefixes can be assigned to many locations is to display the count in the "locations" column. 
https://demo.nautobot.com/ipam/prefixes/ for example.
This change was made to make the table cope with prefixes assigned to very many locations but has made the table less useful for the majority of cases where prefixes are assigned to a single location. Suggest a compromise:
* if the prefix is assigned to one location display the location name
* if the prefix is assigned to multiple locations display the count

### So that ...

The current summary table displays a more useful view showing which prefixes are assigned to which location

The prefix summary table as per https://demo.nautobot.com/ipam/prefixes/ behaves like the previous behaviour unless a prefix is assigned to more than one location

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [X] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

UI change only - no database changes required

### External Dependencies

_No response_

glennmatthews:
# Closes #6231
# What's Changed

- Add additional optional parameter `lookup` to `LinkedCountColumn` API.
    - If present on a given column, the queryset optimizer in BaseTable will add an appropriate `Prefetch()` call to the queryset that will fetch at most one instance of the related objects for each record (row) in the table.
    - On render, if the related instance is present, and the related object count is exactly 1 (meaning the related instance is the *only* related object for this column/row), render that related object directly (via standard `hyperlinked_object()` helper) rather than rendering a `1` with a link to the filtered related-objects table.
    - Otherwise, keep the current behavior (rendering the count of related objects with a hyperlink to the filtered related-objects table)
    - I did change the behavior in case of related count of 0 to display our standard em-dash placeholder instead of an unlinked `0` but I'm perfectly willing to change it back if that gives any heartburn.

# Screenshots

image

Note the `Prefixes` and `Locations` columns now display the related Prefix or Location if there is only a single such related object, while in cases where there are multiple Locations for a VLAN, we still display the count of related locations instead of a single value.

# TODO

- [x] Unit, Integration Tests


glennmatthews:
Changed to render linked-count values as a badge to make a clearer distinction:

image

itdependsnetworks:
This looks great!!

digital-fortress:
I've been waiting for this for so long! Thank you, mate, for listening.

nautobot/core/testing/views.py
```diff
@@ -873,6 +873,8 @@ def test_list_objects_with_permission(self):
                 response_body,
             )
 
+            return response
```
HanlinMiao, line 876: Delete here?

HanlinMiao (approved):

PR #6454, opened 2024-10-30
Fixes issue with ip removal/deletion from interfaces in modules

### LINKED ISSUE #6453: Unable to remove IP from Interfaces in Modules
### Environment
* Nautobot version (Docker tag too if applicable): 2.3.8
* Python version: 3.12
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Create a Module
2. Add Interface
3. Add IP
4. Remove IP

### Expected Behavior
Successful Removal of the IP from the Interface

### Observed Behavior
"AttributeError: 'NoneType' object has no attribute 'primary_ip4'"

image

kvncampos:
# Closes #6453 
# What's Changed
- Added the module to be checked in signals.py

# Screenshots
Before
image

After
image

# TODO
- [ x] Explanation of Change(s)
- [x ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [x ] Attached Screenshots, Payload Example


changes/6442.fixed
glennmatthews, line 1: Unintended deletion?

nautobot/ipam/signals.py
```diff
@@ -65,16 +67,33 @@ def ip_address_to_interface_pre_delete(instance, raw=False, **kwargs):
...
 
     if getattr(instance, "interface"):
-        host = instance.interface.device
-        other_assignments_exist = (
-            IPAddressToInterface.objects.filter(interface__device=host, ip_address=instance.ip_address)
-            .exclude(id=instance.id)
-            .exists()
-        )
+        interface = instance.interface
+        host = interface.device or interface.module.device
```
glennmatthews, line 69: Could use `interface.parent` here as that has the built-in logic to return either `.module.device` or `.device`.

nautobot/ipam/signals.py
```diff
@@ -65,12 +65,23 @@ def ip_address_to_interface_pre_delete(instance, raw=False, **kwargs):
...
+                .exclude(id=instance.id)
+                .exists()
+            )
+        elif interface.module:
+            # Check assignments within the module
+            other_assignments_exist = (
+                IPAddressToInterface.objects.filter(interface__module=interface.module, ip_address=instance.ip_address)
+                .exclude(id=instance.id)
+                .exists()
+            )
```
glennmatthews, line 84: I'd think we'd want this to actually be inclusive (check *both* possibilities, as for example I think it should be OK to remove an IP from an interface directly on the Device if that IP is also assigned to an interface on a Module on that same Device. Also need to consider other modules (potentially nested) on the same device. I think this should maybe be something like:

```suggestion
        other_assignments_exist = (
            IPAddressToInterface.objects.filter(interface__in=host.all_interfaces(), ip_address=instance.ip_address)
            .exclude(id=instance.id)
            .exists()
        )
```

glennmatthews (commented):
Thanks!

nautobot/ipam/signals.py
kvncampos: seems to work after testing with nested modules and module on device.

changes/6442.fixed
kvncampos: @glennmatthews did not know if I was supposed to keep both of these.
6442 was not me. if so, I can add it back!

glennmatthews: Yeah, it's in the branch already so leave it be :-).

kvncampos: done!

nautobot/ipam/signals.py
kvncampos: awesome, did not know that, added/changed it!

glennmatthews (approved):

HanlinMiao (commented):
Thank you for the PR! Can we add a unit test or two to ensure this is the correct behavior?

HanlinMiao (approved):

kvncampos:
oops, I made a new test file, let me move to test_models.py

PR #6455, opened 2024-10-30
fix tests that always passed due to errors

cardoe:
These two tests always passed due to errors in how the tests were implemented. This ensures that the tests provide value by checking against the correct results.


glennmatthews (commented):
Looks like good fixes to me! Mind adding a change fragment?

glennmatthews (approved):

PR #6471, opened 2024-11-04
Kubernetes Job Execution

### LINKED ISSUE #6088: k8s Jobs - Enqueue to k8s

HanlinMiao:
# Closes #6088 
# What's Changed

# Screenshots

1. `kubectl apply -f <filename>` to set up the deployments and services
Screenshot 2024-11-04 at 4 08 53 PM
2. `kubectl exec --stdin --tty nautobot-679bdc765-w7p8h -- /bin/bash` command to shell into the nautobot pod and make an api request. See a JobResult with status PENDING returned.
Screenshot 2024-11-04 at 4 10 01 PM
3. Execute command `nautobot-server nbshell` and you can check the latest `JobResult` from there. The status should be SUCCESS.

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/utils.py
```diff
@@ -615,82 +613,49 @@ def refresh_job_model_from_job_class(job_model_class, job_class, job_queue_class
...
+    with kubernetes.client.ApiClient(configuration) as api_client:
+        api_instance = kubernetes.client.BatchV1Api(api_client)
+
+    with open("./development/nautobot-job-job.yaml", "r") as f:
+        pod_manifest = yaml.safe_load(f)
+
+    user = job_result.user.username
+    job_model = job_result.job_model
+    job_class_path = f"{job_model.module_name}.{job_model.job_class_name}"
+    pod_manifest["metadata"]["name"] = "nautobot-job"
```
glennmatthews, line 640: Is this where we'd need to insert the job_result.pk to allow for reruns and multiple jobs without collision?

nautobot/extras/utils.py
```diff
@@ -615,82 +613,49 @@ def refresh_job_model_from_job_class(job_model_class, job_class, job_queue_class
...
+    configuration.api_key["authorization"] = token
+    with kubernetes.client.ApiClient(configuration) as api_client:
+        api_instance = kubernetes.client.BatchV1Api(api_client)
+
+    with open("./development/nautobot-job-job.yaml", "r") as f:
+        pod_manifest = yaml.safe_load(f)
+
+    user = job_result.user.username
+    job_model = job_result.job_model
+    job_class_path = f"{job_model.module_name}.{job_model.job_class_name}"
```
glennmatthews, line 639: Can't we use `job_model.class_path`?

nautobot/extras/utils.py
```diff
@@ -615,82 +613,49 @@ def refresh_job_model_from_job_class(job_model_class, job_class, job_queue_class
...
+        "--local",
+        "-u",
+        f"{user}",  # TODO replace these arguments with actual
+        f"{job_class_path}",
+        "-d",
+        f"{job_kwargs}",
+        "-r",
+        f"{job_result.pk}",
+    ]
+    logger.info(f"Creating job pod {pod_name} in namespace {pod_namespace}")
```
glennmatthews, line 653: nit - use %s-style formatting in log messages.

glennmatthews (commented):
Nice progress!

nautobot/extras/utils.py
```diff
@@ -615,82 +613,49 @@ def refresh_job_model_from_job_class(job_model_class, job_class, job_queue_class
...
-    sresult = resp.read_stdout()
-    serror = resp.read_stderr()
-    logger.info(f"Job Result is: {sresult}")
-    logger.info(f"Error is: {serror}")
-    resp.close()
+    configuration = kubernetes.client.Configuration()
+    # configure API Key authorization: BearerToken
+    configuration.host = settings.KUBERNETES_JOB_POD_HOST
+    configuration.ssl_ca_cert = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+    with open("/var/run/secrets/kubernetes.io/serviceaccount/token", "r") as token_file:
```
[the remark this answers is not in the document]
HanlinMiao: So currently these credentials are automatically generated in these paths in a Kubernetes pod. I don't see a reason for the user to move them around. But we can make them configurable just in case.

nautobot/extras/utils.py
```diff
@@ -615,82 +613,49 @@ def refresh_job_model_from_job_class(job_model_class, job_class, job_queue_class
...
+        token = token_file.read().strip()
+    configuration.api_key_prefix["authorization"] = "Bearer"
+    configuration.api_key["authorization"] = token
+    with kubernetes.client.ApiClient(configuration) as api_client:
+        api_instance = kubernetes.client.BatchV1Api(api_client)
+
+    with open("./development/nautobot-job-job.yaml", "r") as f:
```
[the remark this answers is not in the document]
lampwins: I think for now the entire spec can/should be a settings variable. The spec and specifically, things like image name and environment setup will vary.

nautobot/extras/utils.py
lampwins: Yes, we should also add a sane default for `ttlSecondsAfterFinished` which will purge these entities after jobs finish. https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/

nautobot/extras/management/commands/runjob.py
```diff
@@ -44,6 +50,28 @@ def handle(self, *args, **options):
...
+                    f"There is a mismatch between the user specified in the command {user} and the user associated with the job result {job_result.user}"
+                )
+
+            job_model = job_result.job_model
+            job_class_path = f"{job_model.module_name}.{job_model.job_class_name}"
+            if job_class_path != options["job"]:
+                job = options["job"]
+                raise CommandError(
+                    f"There is a mismatch between the job specified in the command {job} and the job associated with the job result {job_class_path}"
+                )
```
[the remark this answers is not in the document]
HanlinMiao: Took this approach here and created `runjob_with_job_result.py`

nautobot/extras/api/views.py
```diff
@@ -700,6 +700,7 @@ def run(self, request, *args, **kwargs):
             # of errors under messages
             return Response({"errors": e.message_dict if hasattr(e, "error_dict") else e.messages}, status=400)
 
+        # TODO: refactor this to take into account of Kubernetes queue
         if not get_worker_count(queue=task_queue):
```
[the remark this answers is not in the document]
HanlinMiao: Did some refactoring on `get_worker_count`: Basically if the JobQueue is type `Kubernetes`, always return 1. That should achieve the same effect here.

nautobot/extras/utils.py
```diff
@@ -615,82 +613,49 @@ def refresh_job_model_from_job_class(job_model_class, job_class, job_queue_class
...
+        "runjob",
+        "--local",
+        "-u",
+        f"{user}",  # TODO replace these arguments with actual
+        f"{job_class_path}",
+        "-d",
+        f"{job_kwargs}",
+        "-r",
+        f"{job_result.pk}",
+    ]
```
lampwins: If we are passing the job result, we shouldn't need to pass anything else, no?

HanlinMiao: [f1109ec](https://github.com/nautobot/nautobot/pull/6471/commits/f1109ec424f5b1077d2ee298b5bb83a08db1d728) In this commit here, with the new management command we are only passing in the `job_kwargs` and job result.  Correct me if I am wrong but `job_kwargs` seems to be never stored on the job result, it is passed into `apply_async()` by itself.

nautobot/core/settings.py
```diff
@@ -1097,17 +1097,14 @@ def silk_request_logging_intercept_logic(request):
 # Kubernetes settings variables
 #
 
-# File path to the config file used to load and configure the kubernetes cluster
-KUBECONFIG = os.getenv("NAUTOBOT_KUBECONFIG", "")
+# Filepath of the kubernetes pod manifest used to create a job pod in the kubernetes cluster
+KUBERNETES_JOB_POD_MANIFEST = os.getenv("NAUTOBOT_KUBERNETES_JOB_POD_MANIFEST", {})
```
glennmatthews, line 1101: Update comment appropriately; also this probably isn't doable as an environment variable.

nautobot/extras/api/views.py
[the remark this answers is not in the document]
glennmatthews: I don't quite like that approach - better to not call that API rather than making it return a mock value.

nautobot/extras/utils.py
```diff
@@ -615,82 +627,39 @@ def refresh_job_model_from_job_class(job_model_class, job_class, job_queue_class
...
-        tty=False,
-        _preload_content=False,
-    )
-    resp.write_stdin("nautobot-server runjob --local -u admin nautobot.core.jobs.ExportObjectList\n")
-    sresult = resp.read_stdout()
-    serror = resp.read_stderr()
-    logger.info(f"Job Result is: {sresult}")
-    logger.info(f"Error is: {serror}")
-    resp.close()
+    pod_manifest = settings.KUBERNETES_JOB_POD_MANIFEST
```
glennmatthews, line 636: Since we're modifying the manifest below, probably need a `copy()` or even a `deepcopy()` here?

nautobot/extras/utils.py
```diff
@@ -615,82 +627,39 @@ def refresh_job_model_from_job_class(job_model_class, job_class, job_queue_class
...
+    configuration.api_key["authorization"] = token
+    with kubernetes.client.ApiClient(configuration) as api_client:
+        api_instance = kubernetes.client.BatchV1Api(api_client)
+
+    pod_manifest["metadata"]["name"] = "nautobot-job-" + str(job_result.pk)
+    pod_manifest["spec"]["template"]["spec"]["containers"][0]["command"] = [
+        "nautobot-server",
+        "runjob_with_job_result",
+        "-d",
+        f"{job_kwargs}",
```
glennmatthews, line 654: This "feels" off to me still, as the job_result definitely stores the job_kwargs *after* execution, so it feels weird that we need to (re?)specify the job_kwargs here. Maybe there's something that Celery does automatically for us that we just need to add manually when creating a k8s job_result?

nautobot/extras/utils.py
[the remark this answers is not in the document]
lampwins: Yes these should be configurable, with defaults. The CA cert specifically, will vary, depending on how one has k8s setup.

development/docker-compose-min.yml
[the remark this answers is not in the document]
HanlinMiao: This file was used to generate the new yaml files with `Kompose`.  We can delete it for now imo.

development/nautobot-cm1-configmap.yaml
```diff
@@ -0,0 +1,111 @@
+apiVersion: v1
+data:
+  nautobot_config.py: |
```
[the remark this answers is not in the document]
HanlinMiao: Yes unfortunately I have been manually updating it.

nautobot/core/settings.py
HanlinMiao: Instead of os.getenv, just set it to a dictionary.

nautobot/core/settings.yaml
```diff
@@ -1058,23 +1058,16 @@ properties:
...
-    type: "string"
-    version_added: "2.4.0"
-  KUBERNETES_JOB_IMAGE_NAME:
-    default: ""
-    description: "The name of the docker image used to build the nautobot container running on the Kubernetes pod."
-    environment_variable: "NAUTOBOT_KUBERNETES_JOB_IMAGE_NAME"
-    type: "string"
+  KUBERNETES_JOB_POD_MANIFEST:
+    default: {}
+    description: "A dictionary that contains the pod manifest used to create a job pod in the kubernetes cluster."
```
glennmatthews, line 1069: We should *probably* plan to add a new section in the Job docs with details of how to set this up. Not replicating the k8s documentation or anything to that level of detail, but more guidance on how, given you have a k8s setup, to configure Nautobot to make use of it.

nautobot/extras/management/__init__.py
```diff
@@ -389,3 +395,82 @@ def _clear_custom_role_or_status_instances(
 
     if verbosity >= 2:
         print(f"    Deleted {deleted_total}, unlinked {unlinked_total} {metadatamodel} records")
+
+
+def validate_job_and_job_data(command, user, job_class_path, data=None):
+    job_data = {}
+    try:
+        if data:
+            job_data = json.loads(data)
```
glennmatthews, line 404: Do we need to use `NautobotKombuJSONEncoder` here by any chance? Thinking about things like datetime and UUID handling.

nautobot/extras/management/__init__.py
```diff
@@ -389,3 +395,82 @@ def _clear_custom_role_or_status_instances(
...
+    except Job.DoesNotExist as error:
+        raise CommandError(f"Job {job_class_path} does not exist.") from error
+
+    try:
+        job_model = Job.objects.restrict(user, "run").get_for_class_path(job_class_path)
+    except Job.DoesNotExist:
+        raise CommandError(f"User {user.username} does not have permission to run this Job") from None
+
+    if not job_model.installed or job_model.job_class is None:
+        raise CommandError("Job is not presently installed")
```
glennmatthews, line 422: This *might* be partially redundant with lines 408-409?

nautobot/extras/models/jobs.py
```diff
@@ -820,20 +831,30 @@ def enqueue_job(
...
+                )
 
         if task_queue is None:
             task_queue = job_model.default_job_queue.name
 
         job_queue = JobQueue.objects.get(name=task_queue)
         # Kubernetes Job Queue logic
-        if job_queue.queue_type == JobQueueTypeChoices.TYPE_KUBERNETES:
-            return run_kubernetes_job_and_return_job_result(job_queue, job_result)
+        if job_queue.queue_type == JobQueueTypeChoices.TYPE_KUBERNETES and not synchronous:
```
glennmatthews, line 856: Add a comment explaining the reason for the `and not synchronous` logic?

nautobot/extras/utils.py
```diff
@@ -402,6 +400,8 @@ def get_celery_queues():
 def get_worker_count(request=None, queue=None):
     """
     Return a count of the active Celery workers in a specified queue (Could be a JobQueue instance, instance pk or instance name).
+    However if the queue is of type `Kubernetes`, always return 1
+    because the number of workers is irrelevant in this case and we should not let that block the job from running.
```
glennmatthews, line 404: revert?

nautobot/extras/utils.py
```diff
@@ -411,10 +411,17 @@ def get_worker_count(request=None, queue=None):
...
+                queue = queue.name
             except JobQueue.DoesNotExist:
                 return 0
         else:
+            try:
+                # check if the string passed in is a valid name
+                queue = JobQueue.objects.get(name=queue)
+                queue = queue.name
+            except JobQueue.DoesNotExist:
+                return 0
```
glennmatthews, line 424: Not sure I see the need for this addition?

nautobot/extras/utils.py
```diff
@@ -615,82 +622,41 @@ def refresh_job_model_from_job_class(job_model_class, job_class, job_queue_class
...
+    pod_manifest["spec"]["template"]["spec"]["containers"][0]["command"] = [
+        "nautobot-server",
+        "runjob_with_job_result",
+        f"{job_result.pk}",
+    ]
+    logger.info("Creating job pod %s in namespace %s", pod_name, pod_namespace)
+    api_instance.create_namespaced_job(body=pod_manifest, namespace=pod_namespace)
+    logger.info("Reading job pod %s in namespace %s", pod_name, pod_namespace)
+    api_instance.read_namespaced_job(name="nautobot-job-" + str(job_result.pk), namespace=pod_namespace)
+    logger.info("Done.")
```
glennmatthews, line 658: Maybe remove this message - the job isn't actually done at this point, it's just been successfully enqueued right?

nautobot/extras/utils.py
```diff
@@ -615,82 +622,41 @@ def refresh_job_model_from_job_class(job_model_class, job_class, job_queue_class
...
+    job_result.refresh_from_db()
```
glennmatthews, line 659: Is this needed?

nautobot/extras/management/__init__.py
HanlinMiao: We did not have NautobotKombuJSONEncoder originally in the `runjob` code. But I can definitely add it.

nautobot/extras/management/__init__.py
HanlinMiao: I am a little hesitant to modify it since it was in the original `runjob.py`. Since these lines are checking if the job is installed and lines 408-409 are checking whether the job can be imported?

nautobot/extras/management/__init__.py
glennmatthews: Should be fine as-is then. Thanks.

nautobot/extras/management/__init__.py
glennmatthews: Fine to leave as-is.

nautobot/extras/utils.py
```diff
@@ -446,6 +437,33 @@ def get_job_queue_worker_count(request=None, job_queue=None):
...
+
+    queue = None
+    if is_uuid(job_queue):
+        try:
+            # check if the string passed in is a valid UUID
+            queue = JobQueue.objects.get(pk=job_queue)
+        except JobQueue.DoesNotExist:
+            queue, _ = JobQueue.objects.get_or_create(
+                name=settings.CELERY_TASK_DEFAULT_QUEUE, defaults={"queue_type": JobQueueTypeChoices.TYPE_CELERY}
+            )
```
glennmatthews, line 455: This addition feels weird to me - so if I call `get_job_queue("some_nonexistent_name")`, I get the default Celery queue as a response instead of `None`?

glennmatthews: Feels like it'd be better to just check for `queue is not None`?

development/nautobot_config.py
```diff
@@ -288,3 +288,9 @@
 
 # Host of the kubernetes pod created in the kubernetes cluster
 KUBERNETES_JOB_POD_HOST = "https://kubernetes.default.svc"
+
+# File path to the SSL CA CERT used for authentication to create the job and job pod
+KUBERNETES_JOB_POD_SSL_CA_CERT = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+
+# File path to the Bearer token used for authentication to create the job and job pod
+KUBERNETES_JOB_POD_TOKEN = "/var/run/secrets/kubernetes.io/serviceaccount/token"  # noqa: S105
```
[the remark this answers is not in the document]
HanlinMiao: Set them as default in `settings.py` and removed them from `nautobot_config.py`

glennmatthews (approved):

nautobot/core/settings.yaml
```diff
@@ -1058,23 +1058,16 @@ properties:
       +/- 2.0.0
           This directory no longer requires an `__init__.py` file.
     environment_variable: "NAUTOBOT_JOBS_ROOT"
-  KUBECONFIG:
+  KUBERNETES_JOB_POD_HOST:
```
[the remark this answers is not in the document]
HanlinMiao: Changed it to `KUBERNETES_DEFAULT_SERVICE_ADDRESS`

nautobot/extras/management/__init__.py
```diff
@@ -389,3 +396,82 @@ def _clear_custom_role_or_status_instances(
 
     if verbosity >= 2:
         print(f"    Deleted {deleted_total}, unlinked {unlinked_total} {metadatamodel} records")
+
+
+def validate_job_and_job_data(command, user, job_class_path, data=None):
+    job_data = {}
+    try:
+        if data:
+            job_data = json.loads(data, cls=NautobotKombuJSONEncoder)
```
HanlinMiao, line 405: We should have a decoder here instead of an encoder.

PR #5313, opened 2024-02-21
Adjusted Edit / Create panels to accommodate larger sizes on  medium …

timizuoebideri1:
…and large screens by updating classes to `col-lg-8 col-lg-offset-2 col-md-10 col-md-offset-1`

# Closes #5062 
# What's Changed

# Screenshots

### Large Screens
Screenshot 2024-02-21 at 8 45 36 AM

### Medium Screens
Screenshot 2024-02-21 at 8 45 49 AM

# TODO

- [ ] Explanation of Change(s)
- [ ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [ ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example Plugin Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #6473, opened 2024-11-05
Fix incorrect placement of buttons in create and edit views

### LINKED ISSUE #6472: Edit form buttons don't align properly in large viewport
### Environment
* Nautobot version (Docker tag too if applicable): v2.3.10
* Python version: N/A
* Database platform, version: N/A
* Middleware(s): N/A

### Steps to Reproduce
1. Navigate to any create/edit view and scroll down to the buttons
2. Resize window until it gets to the "large" bootstrap viewport size

### Expected Behavior

image

### Observed Behavior

image

gsnider2195:
# Closes #6472
# What's Changed

Fixed incorrect placement of buttons in create and edit views. Screenshots below

Related to #5313

# Screenshots

## Before :grimacing:

image

## After

image

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/core/templates/generic/object_create.html
+++ b/nautobot/core/templates/generic/object_create.html
@@ -50,7 +50,7 @@
             </div>
         </div>
         <div class="row">
-            <div class="col-md-6 col-md-offset-3 text-right">
+            <div class="col-lg-8 col-lg-offset-2 col-md-10 col-md-offset-1 text-right">
                 {% block buttons %}
                     {% if editing %}
                         <button type="submit" name="_update" class="btn btn-primary">Update</button>
--- a/changes/6472.fixed
+++ b/changes/6472.fixed
@@ -1,1 +1,2 @@
 Fixed incorrect placement of buttons in create and edit views.
+Fixed the panel width in multiple create and edit views.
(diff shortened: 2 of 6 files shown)


nautobot/core/templates/generic/object_create.html
```diff
@@ -50,7 +50,7 @@ <h3>
             </div>
         </div>
         <div class="row">
-            <div class="col-md-6 col-md-offset-3 text-right">
+            <div class="col-lg-8 col-lg-offset-2 col-md-10 col-md-offset-1 text-right">
```
glennmatthews, line 53: Are there model-specific templates that may also need updates? Grepping for `col-md-offset-3 text-right` shows a few candidates such as `ipam/ipaddress_assign.html` and `virtualization/cluster_add_devices.html`.

glennmatthews (approved):

nautobot/core/templates/generic/object_create.html
gsnider2195: Looks like those were missed in #5313 so the bug wouldn't be present here. Do we want to make the #5313 changes to those two templates?

glennmatthews: It would be good to update them for consistency's sake but I'm also fine if you want to consider that out of scope for this PR.

glennmatthews (approved):

PR #6474, opened 2024-11-05
skip rendering the table of related objects in the confirmation page.

### LINKED ISSUE #6465: "Edit All" view should skip rendering the table of related objects in the confirmation page
### Proposed Changes
"Edit All" view should skip rendering the table of related objects in the confirmation page, just like Delete all in https://github.com/nautobot/nautobot/pull/6420

timizuoebideri1:
# Closes #6465 
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/core/templates/generic/object_bulk_update.html
+++ b/nautobot/core/templates/generic/object_bulk_update.html
@@ -36,7 +36,7 @@
             </div>
         </div>
         {% endif %}
-        <div class="{% if edit_all %} col-lg-8 col-lg-offset-2 col-md-10 col-md-offset-1 {% else %} col-md-4 {% endif %}">
+        <div class="{% if table %} col-md-4 {% else %} col-lg-8 col-lg-offset-2 col-md-10 col-md-offset-1 {% endif %}">
             <div class="panel panel-default">
                 <div class="panel-heading"><strong>{% block form_title %}Attributes{% endblock %}</strong></div>
                 <div class="panel-body">
--- a/nautobot/core/templates/generic/object_bulk_update.html
+++ b/nautobot/core/templates/generic/object_bulk_update.html
@@ -4,7 +4,7 @@
 {% load render_table from django_tables2 %}
 
 {% block content %}
-<h1>{% block title %}Editing {{ obj_count }} {{ obj_type_plural|bettertitle }}{% endblock %}</h1>
+<h1>{% block title %}Editing {{ objs_count }} {{ obj_type_plural|bettertitle }}{% endblock %}</h1>
 {% if form.errors %}
     <div class="panel panel-danger">
         <div class="panel-heading"><strong>Errors</strong></div>
--- a/nautobot/core/templates/generic/object_bulk_update.html
+++ b/nautobot/core/templates/generic/object_bulk_update.html
@@ -4,7 +4,7 @@
 {% load render_table from django_tables2 %}
 
 {% block content %}
-<h1>{% block title %}Editing {{ table.rows|length }} {{ obj_type_plural|bettertitle }}{% endblock %}</h1>
+<h1>{% block title %}Editing {{ obj_count }} {{ obj_type_plural|bettertitle }}{% endblock %}</h1>
 {% if form.errors %}
     <div class="panel panel-danger">
         <div class="panel-heading"><strong>Errors</strong></div>
@@ -27,6 +27,7 @@
         {{ field }}
     {% endfor %}
     <div class="row">
+        {% if table %}
         <div class="col-md-8">
             <div class="panel panel-default">
                 <div class="table-responsive">
@@ -34,7 +35,8 @@
                 </div>
             </div>
         </div>
-        <div class="col-md-4">
+        {% endif %}
+        <div class="{% if edit_all %} col-lg-8 col-lg-offset-2 col-md-10 col-md-offset-1 {% else %} col-md-4 {% endif %}">
             <div class="panel panel-default">
                 <div class="panel-heading"><strong>{% block form_title %}Attributes{% endblock %}</strong></div>
                 <div class="panel-body">
--- a/nautobot/core/testing/views.py
+++ b/nautobot/core/testing/views.py
@@ -1144,7 +1144,7 @@ class ViewTestCases:
 
         @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
         def test_bulk_edit_form_contains_all_filtered(self):
-            # We are testing the intermediary step of all bulk_edit with additional filter.
+            # We are testing the intermediary step of bulk editing all filtered objects.
             # i.e. "_all" passed in the form and filter using query params.
             self.add_permissions(f"{self.model._meta.app_label}.change_{self.model._meta.model_name}")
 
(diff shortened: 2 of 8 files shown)


nautobot/core/views/mixins.py
```diff
@@ -1178,38 +1184,42 @@ def perform_bulk_update(self, request, **kwargs):  # pylint: disable=arguments-d
...
-            messages.warning(
-                request,
-                f"No {queryset.model._meta.verbose_name_plural} were selected to update.",
-            )
-            return redirect(self.get_return_url(request))
-        data.update({"table": table})
+        table = None
+        if not edit_all:
+            table_class = self.get_table_class()
+            table = table_class(queryset.filter(pk__in=self.pk_list), orderable=False)
```
glennmatthews, line 1209: Is the filter() call here redundant with the filter() call at line 1194 above?

nautobot/extras/forms/base.py
```diff
@@ -62,3 +64,9 @@ class NautobotBulkEditForm(
     RelationshipModelBulkEditFormMixin,
 ):
     """Base class for bulk-edit forms for models that support relationships, custom fields and notes."""
```
glennmatthews, line 66: Are there any models that have a BulkEditForm that isn't a a NautobotBulkEditForm subclass?

nautobot/core/views/generic.py
```diff
@@ -1041,7 +1041,8 @@ def post(self, request, **kwargs):
                 try:
                     with deferred_change_logging_for_bulk_operation():
                         updated_objects = []
-                        for obj in self.queryset.filter(pk__in=form.cleaned_data["pk"]):
+                        queryset = queryset if edit_all else queryset.filter(pk__in=form.cleaned_data["pk"])
```
glennmatthews, line 1044: Is this filter() redundant with the filter() at line 1024 above?

itdependsnetworks:
Can you take a look here: https://github.com/nautobot/nautobot-app-golden-config/blob/develop/nautobot_golden_config/views.py#L280 and check if there is any impact from this change?

nautobot/core/views/mixins.py
```diff
@@ -912,23 +912,23 @@ def perform_update(self, request, *args, **kwargs):  # pylint: disable=arguments
...
 
         This method handles the retrieval of a queryset of model instances that match the specified
         filter criteria in the request parameters, allowing a bulk delete operation to be performed
         on all matching instances.
         """
         model = self.queryset.model
 
-        # This Mixin is currently been used by both NautobotViewSet ObjectBulkDestroyViewMixin
-        # and BulkDeleteView which uses different keys for accessing filterset
+        # This Mixin is currently been used by both NautobotViewSet ObjectBulkDestroyViewMixin, ObjectBulkUpdateViewMixin
```
HanlinMiao, line 930: Perhaps?
```suggestion
        # This Mixin is currently been used by both NautobotUIViewSet ObjectBulkDestroyViewMixin, ObjectBulkUpdateViewMixin
```

nautobot/core/views/renderers.py
```diff
@@ -249,19 +249,22 @@ def get_context(self, data, accepted_media_type, renderer_context):
...
+                    table = self.construct_table(view, pk_list=pk_list)
             elif view.action == "bulk_create":  # 3.0 TODO: remove, replaced by ImportObjects system Job
                 form = view.get_form()
                 if request.data:
                     table = data.get("table")
             elif view.action == "bulk_update":
+                edit_all = request.POST.get("_all")
                 pk_list = getattr(view, "pk_list", [])
-                if pk_list:
+                if pk_list or edit_all:
```
HanlinMiao, line 262: Can we delete the if statement here?

glennmatthews:
> Can you take a look here: https://github.com/nautobot/nautobot-app-golden-config/blob/develop/nautobot_golden_config/views.py#L280 and check if there is any impact from this change?

Yes, most likely GC will need to be updated to account for the changes from #6420.

nautobot/core/templates/generic/object_bulk_update.html
```diff
@@ -27,14 +27,16 @@ <h1>{% block title %}Editing {{ table.rows|length }} {{ obj_type_plural|betterti
...
         <div class="col-md-8">
             <div class="panel panel-default">
                 <div class="table-responsive">
                     {% render_table table 'inc/table.html' %}
                 </div>
             </div>
         </div>
-        <div class="col-md-4">
+        {% endif %}
+        <div class="{% if edit_all %} col-lg-8 col-lg-offset-2 col-md-10 col-md-offset-1 {% else %} col-md-4 {% endif %}">
```
glennmatthews, line 39: Do we really need a separate `edit_all` flag in this template or could this just be changed to `if not table`?

nautobot/core/testing/views.py
```diff
@@ -1113,14 +1113,11 @@ def test_bulk_edit_objects_with_permission(self):
             self.validate_object_data_after_bulk_edit(pk_list)
 
         @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
-        def test_bulk_edit_form_contains_all_pks(self):
-            # We are testing the intermediary step of bulk_edit with pagination applied.
+        def test_bulk_edit_form_contains_all_objects(self):
```
glennmatthews, line 1116: Renaming this test case *might* be impactful to Apps, we should consider keeping the name unchanged.

nautobot/core/views/generic.py
```diff
@@ -1041,7 +1041,7 @@ def post(self, request, **kwargs):
                 try:
                     with deferred_change_logging_for_bulk_operation():
                         updated_objects = []
-                        for obj in self.queryset.filter(pk__in=form.cleaned_data["pk"]):
+                        for obj in queryset:
```
glennmatthews, line 1044: Any significance to the fact that we were doing `pk__in=form.cleaned_data["pk"]` before but now (lines 1023-1024 above) we're doing `pk__in=request.POST.getlist("pk")`? Any subtle behavior change that results from this?

nautobot/core/views/renderers.py
glennmatthews: Not sure - don't we need to account for the possibility of an empty `pk_list`?

HanlinMiao: I see we are accounting for the scenario where no objects are selected at all. That makes sense.

nautobot/core/views/generic.py
glennmatthews: My concern was about the non-`edit_all` case - just trying to understand the significance of the differing logic between line 1044 (before) and line 1024 (after).

nautobot/core/views/generic.py
```diff
@@ -1041,7 +1041,8 @@ def post(self, request, **kwargs):
                 try:
                     with deferred_change_logging_for_bulk_operation():
                         updated_objects = []
-                        for obj in self.queryset.filter(pk__in=form.cleaned_data["pk"]):
+                        queryset = queryset if edit_all else self.queryset.filter(pk__in=form.cleaned_data["pk"])
```
glennmatthews, line 1044: In the non-edit-all case, doesn't this overwrite the assignment at line 1024? Should it perhaps be:

```suggestion
                        queryset = queryset if edit_all else queryset.filter(pk__in=form.cleaned_data["pk"])
```

Or alternately, is the assignment at line 1024 unnecessary/irrelevant?

timizuoebideri1: `queryset` is utilised on line 1140

glennmatthews (approved):

PR #6485, opened 2024-11-07
Update request-profiling.md

gskjelstad:
DOCUMENTATION UPDATE ONLY.
_Additional wordsmithing may be desired._

Updating to ensure clarity about globally enabling silk profiling prior to being able to enable per user, and then a note to disable when not in use.

### DIFF
--- a/nautobot/docs/user-guide/administration/guides/request-profiling.md
+++ b/nautobot/docs/user-guide/administration/guides/request-profiling.md
@@ -10,8 +10,10 @@ Request profiling may be enabled by individual users in their profile within the
 
 Once a user enables request profiling, all subsequent HTTP requests made by that specific user to the system will be logged by django-silk. This setting will persist until the user either logs out or disables the setting in their profile.
 
-#### User Setting Notes:
+### User Setting Notes
+
 - This option will only be usable once the global configuration option `ALLOW_REQUEST_PROFILING` has been enabled (described below).
+
 - *Warning!* This will impact the performance of Nautobot greatly! This should be disabled when not in use.
 
 ## Silk UI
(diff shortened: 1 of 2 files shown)


nautobot/docs/user-guide/administration/guides/request-profiling.md
```diff
@@ -10,6 +10,12 @@ Request profiling may be enabled by individual users in their profile within the
 
 Once a user enables request profiling, all subsequent HTTP requests made by that specific user to the system will be logged by django-silk. This setting will persist until the user either logs out or disables the setting in their profile.
 
+### User Setting Notes
+
+- This option will only be usable once the global configuration option `ALLOW_REQUEST_PROFILING` has been enabled (described below).
+
+- *Warning!* This will impact the performance of Nautobot greatly! This should be disabled when not in use.
```
glennmatthews, line 17: Would be good to clarify `This` here - IIRC it's just the user-enabled request profiling, globally enabling it shouldn't have any significant impact if no user has it enabled?

HanlinMiao (commented):
Thank you for the PR and we need a changelog fragment titled `6485.documentation` as well.

nautobot/docs/user-guide/administration/guides/request-profiling.md
gskjelstad: Updated with additional clarification that global is not an issue.

glennmatthews: Did you miss a `git add`? I don't see any change to the text.

gskjelstad: no, you're right. i used the github direct inline editor for all of this, and attempted two changes via it... herp derp on my part.

glennmatthews (approved):

PR #6488, opened 2024-11-07
Add render_jinja2 rest api endpoint

gsnider2195:
# Closes #DNE (zenhub issue in apps workspace)
# What's Changed

I have a story to create a rest api and view to implement a jinja2 live renderer in core similar to https://j2live.ttl255.com/

~I need some assistance from @nautobot/core to determine where this should end up (currently in extras) and what permissions it needs, if any.~

This PR implements the REST API at `/api/core/render-jinja-template`. The view will be implemented in a separate PR.

Example payload:

```json
{
  "template_code": "Hello {{ name }}",
  "context": {
    "name": "World"
  }
}
```

Example response:

```json
{
  "rendered_template": "Hello World",
  "rendered_template_lines": [
    "Hello World"
  ],
  "template_code": "Hello {{ name }}",
  "context": {
    "name": "World"
  }
}
```
# Screenshots

## Swagger docs

image

## Nautobot docs

These docs will be expanded when the follow-up story is implemented that brings in a UI view for this.

image

# TODO

- [ ] Unit, Integration Tests


glennmatthews:
My 2 cents:
- core rather than extras is probably appropriate
- I think just `IsAuthenticated` is probably a sufficient permission since (if I understand this right) using this endpoint doesn't actually access any of the data stored in Nautobot?

glennmatthews:
> It only takes user input for the context data but it does use our jinja2 environment so insecure jinja filters could expose data. I've looked through core and apps and couldn't find any jinja filters that would leak data if used maliciously.

Good point and thanks for the due diligence. Could/should we make it use a locked-down/"vanilla" Jinja2 environment instead or would that run counter to the purpose of this feature?

gsnider2195:
> > It only takes user input for the context data but it does use our jinja2 environment so insecure jinja filters could expose data. I've looked through core and apps and couldn't find any jinja filters that would leak data if used maliciously.
> 
> Good point and thanks for the due diligence. Could/should we make it use a locked-down/"vanilla" Jinja2 environment instead or would that run counter to the purpose of this feature?

I think users will want to be able to use our standard Jinja2 environment so they have access to all the same filters that they would have access to in CustomLinks, ExportTemplates, etc.

gsnider2195:
Marking this as ready for review because I think the CI failures are coming from upstream and not from my branch

nautobot/core/tests/test_api.py
```diff
@@ -838,3 +839,68 @@ def test_get(self):
...
+            format="json",
+            **self.header,
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertSequenceEqual(
+            list(response.data.keys()),
+            ["rendered_template", "rendered_template_lines", "template_code", "context"],
+        )
+        self.assertEqual(response.data["rendered_template"], "bar")
+        self.assertEqual(response.data["rendered_template_lines"], ["bar"])
```
glennmatthews, line 876: Might be good to also have:

- test for multi-line template (distinguishing behavior of `rendered_template` and `rendered_template_lines` a bit further)
- test for a template that includes usage of some of Nautobot's built-in template filters, since that's a desired piece of the feature.

nautobot/docs/user-guide/platform-functionality/rendering-jinja-templates.md
```diff
@@ -0,0 +1,18 @@
+# Rendering Jinja Templates
```
glennmatthews, line 1:
```suggestion
# Rendering Jinja Templates

+++ 2.4.0
```

nautobot/extras/tests/test_api.py
```diff
@@ -4189,7 +4189,7 @@ def test_create_tags_with_invalid_content_types(self):
         data = {**self.create_data[0], "content_types": [Manufacturer._meta.label_lower]}
         response = self.client.post(self._get_list_url(), data, format="json", **self.header)
 
-        tag = Tag.objects.filter(name=data["name"])
+        tag = Tag.objects.filter(name=data["name"])  # pylint: disable=redefined-outer-name
```
glennmatthews, line 4192: revert here and below?

nautobot/core/tests/test_api.py
```diff
@@ -838,3 +839,68 @@ def test_get(self):
         }
         for view_name, view_kwarg in view_kwargs.items():
             self.assertEqual(view_name, get_view_name(viewset(**view_kwarg)))
+
+
+@tag("api")
+class RenderJinjaViewTest(testing.TestCase):
```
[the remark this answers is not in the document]
gsnider2195: I guess I made an assumption that since `APITestCase` was inheriting from `ModelTestCase` that it wasn't appropriate here. Seems to work though

nautobot/core/tests/test_api.py
gsnider2195: It looks like we're only exposing the netutils filters to the django template engine. We're probably going to want those exposed to the jinja engine for this feature?
https://github.com/nautobot/nautobot/blob/develop/nautobot/core/templatetags/netutils.py

glennmatthews: They *should* be exposed to both already. Are you seeing behavior that demonstrates otherwise?

glennmatthews: https://github.com/nautobot/nautobot/blob/develop/nautobot/core/apps/__init__.py#L886-L888

gsnider2195: Thanks! I didn't expect them to be in two different places but it makes sense now that I've looked at #3296

glennmatthews (approved):

PR #6501, opened 2024-11-11
U/igpaec 6500 add invoke showmigrations

### LINKED ISSUE #6500: Add support for invoke showmigrations
### Proposed Changes
Add the`invoke showmigrations` command.

### Justification
Migrations are an integral part of building an app, especially when we start out a new app. It is pretty common to create some new models. Usually, that involves the use of `nautobot-server makemigrations`, `nautobot-server showmigrations`, and `nautobot-server migrate`. I would also expect these commands to be available in the Docker-based workflow. That means I want to use `invoke makemigrations`, `invoke showmigrations`, and `invoke migrate`. Currently, `invoke makemigrations `and `invoke migrate` are available while `invoke showmigrations` is missing. We should add the option to show migrations for consistency and to make the migrations process as smooth as possible.

igpaec:
# Closes #6500 
# What's Changed

Add support  for `invoke showmigrations` command.

# Screenshots

# TODO

- [ ] Unit, Integration Tests


changes/6500.housekeeping
```diff
@@ -0,0 +1 @@
+Add support for `invoke showmigrations` command
```
gsnider2195, line 1:
```suggestion
Added support for `invoke showmigrations` command.
```

tasks.py
```diff
@@ -476,6 +476,14 @@ def makemigrations(context, name=""):
     run_command(context, command)
 
 
+@task
+def showmigrations(context, name=""):
```
gsnider2195, line 480:
```suggestion
def showmigrations(context):
```

timizuoebideri1 (approved):

PR #6511, opened 2024-11-13
Create Bulk Edit job

### LINKED ISSUE #6354: Bulk edit as a system job
### As ...

Ozzie - Operator

### I want ...

Bulk operations like bulk edit to function as a system job

### So that ...

No matter the requested update size, I can trust the request will succeed, and I can get feedback as to the current progress of the operation.

The bulk edit UX workflow progresses in the current form, but the action of performing the actual update is executed as a system job, so that is can run in the background for larger request sizes.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

_No response_

### External Dependencies

_No response_

timizuoebideri1:
# Closes #6354 

Move Bulk Edit operation to System Job
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests


nautobot/core/jobs/__init__.py
```diff
@@ -347,5 +366,168 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
...
+
+            form_custom_fields = getattr(form, "custom_fields", [])
+            form_relationships = getattr(form, "relationships", [])
+            standard_fields = [
+                field
+                for field in form.fields
+                if field not in form_custom_fields + form_relationships + ["pk"] + ["object_note"]
+            ]
+
+            for obj in queryset:
```
glennmatthews, line 409: Just a general note - let's make sure the job has sufficient logging, especially of progress for long-running operations.

nautobot/core/jobs/__init__.py
```diff
@@ -347,5 +366,161 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
             raise RunJobTaskFailed("CSV import not fully successful, see logs")
 
 
-jobs = [ExportObjectList, GitRepositorySync, GitRepositoryDryRun, ImportObjects, LogsCleanup, RefreshDynamicGroupCaches]
+class BulkEditObjects(Job):
```
glennmatthews, line 369: Let's add a new `nautobot.core.jobs.bulk_actions` submodule for this Job and the upcoming bulk-delete job, just to keep things a bit more organized.

nautobot/core/jobs/__init__.py
```diff
@@ -347,5 +366,161 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
...
+    """System Job to bulk Edit objects."""
+
+    content_type = ObjectVar(
+        model=ContentType,
+        description="Type of objects to import",
```
glennmatthews, line 374:
```suggestion
        description="Type of objects to update",
```

nautobot/core/jobs/__init__.py
```diff
@@ -347,5 +366,161 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
...
+        query_params={"has_serializer": True},
```
glennmatthews, line 375: Is this correct/relevant?

nautobot/core/jobs/__init__.py
```diff
@@ -347,5 +366,161 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
...
+            )
+            raise
+        form = form_cls(model, post_data, edit_all=edit_all)
+        restrict_form_fields(form, self.user)
+
+        if form.is_valid():
+            self.logger.debug("Form validation was successful")
+            nullified_fields = post_data.get("_nullify")
+            self._process_valid_form(model, form, filter_query_params, edit_all, nullified_fields)
+            return
```
glennmatthews, line 510:
```suggestion
            return self._process_valid_form(model, form, filter_query_params, edit_all, nullified_fields)
```

nautobot/core/jobs/__init__.py
```diff
@@ -347,5 +366,161 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
...
+                return updated_objects
+        except ValidationError as e:
+            raise ValidationError(f"{obj} failed validation: {e}")
+
+    def _process_valid_form(self, model, form, filter_query_params, edit_all, nullified_fields):
+        try:
+            if updated_objects := self._update_objects(model, form, filter_query_params, edit_all, nullified_fields):
+                msg = f"Updated {len(updated_objects)} {model._meta.verbose_name_plural}"
+                self.logger.info(msg)
+            return
```
glennmatthews, line 473: Any need for an `else` case here?

Also, perhaps:

```suggestion
            if updated_objects := self._update_objects(model, form, filter_query_params, edit_all, nullified_fields):
                msg = f"Updated {len(updated_objects)} {model._meta.verbose_name_plural}"
                self.logger.info(msg)
                return msg
```

nautobot/core/jobs/__init__.py
```diff
@@ -347,5 +366,161 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
...
+                if queryset.filter(pk__in=[obj.pk for obj in updated_objects]).count() != len(updated_objects):
+                    raise ObjectDoesNotExist
+                return updated_objects
+        except ValidationError as e:
+            raise ValidationError(f"{obj} failed validation: {e}")
+
+    def _process_valid_form(self, model, form, filter_query_params, edit_all, nullified_fields):
+        try:
+            if updated_objects := self._update_objects(model, form, filter_query_params, edit_all, nullified_fields):
+                msg = f"Updated {len(updated_objects)} {model._meta.verbose_name_plural}"
```
glennmatthews, line 471: I'm concerned about the scalability of having `_update_objects` return a list of all updated objects. Won't that be problematic if we do an "all" update with 10k or 100k records?

nautobot/core/jobs/__init__.py
```diff
@@ -347,5 +366,161 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
...
+
+                    if hasattr(form, "save_note") and callable(form.save_note):
+                        form.save_note(instance=obj, user=self.user)
+
+                # Enforce object-level permissions
+                if queryset.filter(pk__in=[obj.pk for obj in updated_objects]).count() != len(updated_objects):
+                    raise ObjectDoesNotExist
+                return updated_objects
+        except ValidationError as e:
+            raise ValidationError(f"{obj} failed validation: {e}")
```
glennmatthews, line 466: Why catch one validationerror to just raise a different one?

nautobot/core/jobs/__init__.py
```diff
@@ -347,5 +366,161 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
...
+                form_custom_fields = getattr(form, "custom_fields", [])
+                form_relationships = getattr(form, "relationships", [])
+                standard_fields = [
+                    field
+                    for field in form.fields
+                    if field not in form_custom_fields + form_relationships + ["pk"] + ["object_note"]
+                ]
+
+                for obj in queryset:
+                    self.logger.debug(f"Performing update on {obj} (PK: {obj.pk})")
```
glennmatthews, line 411: This may be a bit overly verbose at scale?

nautobot/core/jobs/__init__.py
```diff
@@ -347,5 +366,161 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
...
+                    # Update custom fields
+                    for field_name in form_custom_fields:
+                        if field_name in form.nullable_fields and field_name in nullified_fields:
+                            obj.cf[remove_prefix_from_cf_key(field_name)] = None
+                        elif form.cleaned_data.get(field_name) not in (None, "", []):
+                            obj.cf[remove_prefix_from_cf_key(field_name)] = form.cleaned_data[field_name]
+
+                    obj.full_clean()
+                    obj.save()
+                    updated_objects.append(obj)
```
glennmatthews, line 445: Same concern here about scalability

nautobot/core/jobs/__init__.py
```diff
@@ -347,5 +366,161 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
...
+                    self.logger.debug(f"Saved {obj} (PK: {obj.pk})")
```
glennmatthews, line 446: Same concern here about verbosity at scale

nautobot/core/views/generic.py
```diff
@@ -1032,93 +1030,7 @@ def post(self, request, **kwargs):
...
-                                obj.tags.remove(*form.cleaned_data["remove_tags"])
-
-                            if hasattr(form, "save_relationships") and callable(form.save_relationships):
-                                # Add/remove relationship associations
-                                form.save_relationships(instance=obj, nullified_fields=nullified_fields)
-
-                            if hasattr(form, "save_note") and callable(form.save_note):
-                                form.save_note(instance=obj, user=request.user)
-
-                            self.extra_post_save_action(obj, form)
```
glennmatthews, line 1099: I don't see where `extra_post_save_action` is called in the new code - how will it be handled, e.g for `PrefixBulkEditView`?

One idea that occurs to me would to make it a new method on `BulkEditForm` base class and make `BulkEditView.extra_post_save_action(obj, form) just redirect to `form.post_save(obj)`?

nautobot/core/jobs/__init__.py
```diff
@@ -347,5 +366,161 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
...
+                    if hasattr(form, "save_relationships") and callable(form.save_relationships):
+                        # Add/remove relationship associations
+                        form.save_relationships(instance=obj, nullified_fields=nullified_fields)
+
+                    if hasattr(form, "save_note") and callable(form.save_note):
+                        form.save_note(instance=obj, user=self.user)
+
+                # Enforce object-level permissions
+                if queryset.filter(pk__in=[obj.pk for obj in updated_objects]).count() != len(updated_objects):
+                    raise ObjectDoesNotExist
```
glennmatthews, line 463: Might be able to address scale concerns if `updated_objects` just stored the PKs rather than the full objects?

nautobot/core/tests/test_jobs.py
```diff
@@ -606,3 +606,90 @@ def test_cleanup_object_changes(self):
...
+        self.add_permissions("extras.change_status", "extras.view_status")
+        # By default Active and Available are some of the example of Status that starts with A
+        statuses = Status.objects.filter(name__istartswith="A")
+        self.assertNotEqual(statuses.count(), 0)
+        job_result = create_job_result_and_run_job(
+            "nautobot.core.jobs",
+            "BulkEditObjects",
+            content_type=self.status_ct.id,
+            edit_all=True,
+            filter_query_params={"name__istartswith": "A"},
```
glennmatthews, line 669: This looks wrong to me - since it's a filterset param, shouldn't it be `name__isw`?

nautobot/core/tests/test_jobs.py
```diff
@@ -606,3 +606,90 @@ def test_cleanup_object_changes(self):
...
+    def test_bulk_edit_with_pk(self):
+        self.add_permissions("extras.change_role", "extras.view_role")
+        roles_to_update = [Role.objects.create(name=f"Example Role {x}") for x in range(3)]
+        roles_pks = [str(role.pk) for role in roles_to_update]
+        job_result = create_job_result_and_run_job(
+            "nautobot.core.jobs",
+            "BulkEditObjects",
+            content_type=self.role_ct.id,
+            edit_all=False,
+            filter_query_params={"name__istartswith": "Example Role"},
```
glennmatthews, line 691: Same question here

nautobot/core/utils/form.py
```diff
@@ -0,0 +1,23 @@
+from nautobot.core.utils.requests import is_single_choice_field
+
+
+def serialize_query_dict_for_form(query_dict, form):
```
glennmatthews, line 4: Docstrings for these functions please

nautobot/core/views/mixins.py
```diff
@@ -912,9 +912,9 @@ def perform_update(self, request, *args, **kwargs):  # pylint: disable=arguments
             return self.form_invalid(form)
 
 
-class EditAndDeleteAllModelMixin:
+class EditAndDeleteModelMixin:
```
glennmatthews, line 915: maybe `BulkEditAndBulkDeleteModelMixin`?

nautobot/core/views/mixins.py
```diff
@@ -952,8 +952,35 @@ def _bulk_delete_all_context(self, request, queryset):
...
+        post_data = serialize_query_dict_for_form(request.POST, form)
+        if filterset_class := lookup.get_filterset_for_model(model):
+            filter_query_params = serialize_query_dict_for_filterset(request.GET, filterset_class())
+        else:
+            filter_query_params = None
+        job_form = BulkEditObjects.as_form(
+            data={
+                "post_data": post_data,
+                "content_type": ContentType.objects.get_for_model(model),
+                # This feature is not yet available on the next branch
```
glennmatthews, line 967: ?

nautobot/core/views/mixins.py
```diff
@@ -1082,94 +1109,14 @@ def perform_bulk_create(self, request):
...
-class ObjectBulkUpdateViewMixin(NautobotViewSetMixin, BulkUpdateModelMixin, EditAndDeleteAllModelMixin):
+class ObjectBulkUpdateViewMixin(NautobotViewSetMixin, BulkUpdateModelMixin, EditAndDeleteModelMixin):
     """
     UI mixin to bulk update model instances.
     """
 
     filterset_class = None
     bulk_update_form_class = None
 
-    def _process_bulk_update_form(self, form):
```
glennmatthews, line 1093: You've removed this method but I don't see where you've removed any code that was previously calling it?

timizuoebideri1: This is the part of UIViewSet that was moved to Bulk Edit System Job

glennmatthews: My point is that presumably *something* was calling `_process_bulk_update_form()` previously, but I don't see where that "something" was removed from the code base. You've removed the definition of the method but not the code that was calling that method?

changes/6354.added
```diff
@@ -0,0 +1 @@
+Bulk edit operation now a system job.
```
glennmatthews, line 1:
```suggestion
Added "Bulk Edit Objects" system job.
```

I'd suggest also adding a .changed fragment for the change in behavior of the UI.

nautobot/core/jobs/bulk_actions.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+class BulkEditObjects(Job):
+    """System Job to bulk Edit objects."""
+
+    content_type = ObjectVar(
+        model=ContentType,
+        description="Type of objects to update",
+    )
+    post_data = JSONVar(description="Data to update with i.e request.POST")
```
glennmatthews, line 30: I kind of prefer the clarity that the delete job has with having a distinct `pk_list` parameter. Would that be difficult to implement in this case?

nautobot/core/jobs/bulk_actions.py
```diff
@@ -0,0 +1,156 @@
...
+        soft_time_limit = 1800
+        time_limit = 2000
+
+    def _update_objects(self, model, form, filter_query_params, edit_all, nullified_fields):
+        with deferred_change_logging_for_bulk_operation():
+            updated_objects_pk = []
+            filterset_cls = get_filterset_for_model(model)
+            if edit_all:
+                if filterset_cls:
+                    queryset = filterset_cls(filter_query_params).qs.restrict(self.user, "change")
```
glennmatthews, line 47: Should we log and/or raise an error if there are `filter_query_params` but we can't find the `filterset_cls`?

nautobot/core/jobs/bulk_actions.py
```diff
@@ -0,0 +1,156 @@
...
+            form_custom_fields = getattr(form, "custom_fields", [])
+            form_relationships = getattr(form, "relationships", [])
+            standard_fields = [
+                field
+                for field in form.fields
+                if field not in form_custom_fields + form_relationships + ["pk"] + ["object_note"]
+            ]
+
+            self.logger.debug(f"Performing update on {queryset.count()} {model._meta.verbose_name_plural}")
+            for obj in queryset:
```
glennmatthews, line 62: Might make for better performance at scale to use `queryset.iterator()`?

nautobot/core/testing/views.py
```diff
@@ -1015,9 +1016,16 @@ class BulkEditObjectsViewTestCase(ModelViewTestCase):
 
         bulk_edit_data = {}
 
-        def validate_object_data_after_bulk_edit(self, pk_list):
```
glennmatthews, line 1018: Might be a breaking change to remove this method - may need to leave it around for now?

nautobot/core/tests/test_jobs.py
```diff
@@ -606,3 +608,177 @@ def test_cleanup_object_changes(self):
...
+            "BulkEditObjects",
+            content_type=self.status_ct.id,
+            edit_all=False,
+            filter_query_params={},
+            post_data={"pk": statuses_to_update, "color": "aa1409"},
+            username=self.user.username,
+        )
+        self.assertEqual(job_result.status, JobResultStatusChoices.STATUS_FAILURE)
+        job_log = JobLogEntry.objects.get(job_result=job_result, log_level=LogLevelChoices.LOG_ERROR)
+        self.assertEqual(job_log.message, f'User "{self.user}" does not have permission to update status objects')
```
glennmatthews, line 646: Would be good to assert the Status objects aren't updated.

nautobot/core/tests/test_jobs.py
```diff
@@ -606,3 +608,177 @@ def test_cleanup_object_changes(self):
...
+            filter_query_params={},
+            post_data={"color": "aa1409"},
+            username=self.user.username,
+        )
+        self._common_no_error_test_assertion(Role, job_result, Role.objects.all().count(), color="aa1409")
+
+    def test_bulk_edit_filter_all(self):
+        self.add_permissions("extras.change_status", "extras.view_status")
+        # By default Active and Available are some of the example of Status that starts with A
+        statuses = Status.objects.filter(name__istartswith="A")
```
glennmatthews, line 664: For safety would be good to assert this queryset is non-empty?

nautobot/core/tests/test_jobs.py
```diff
@@ -606,3 +608,177 @@ def test_cleanup_object_changes(self):
...
+            post_data={"pk": [str(role_to_update.pk)], "color": "aa1409"},
+            username=self.user.username,
+        )
+        error_log = JobLogEntry.objects.get(job_result=job_result, log_level=LogLevelChoices.LOG_ERROR)
+        self.assertIn("Form validation unsuccessful", error_log.message)
+        self.assertIn(f"{role_to_update.pk} is not one of the available choices.", error_log.message)
+        role_to_update.refresh_from_db()
+        self.assertNotEqual(role_to_update.color, "aa1409")
+
+        obj_perm.constraints = {"pk__isnull": False}  # Match a non-existent pk (i.e., allow all)
```
glennmatthews, line 772: Would be good to also have a test with legitimately constrained permissions, i.e. allowing update of *some* objects but not others?

nautobot/core/utils/requests.py
```diff
@@ -171,6 +183,8 @@ def normalize_querydict(querydict, form_class=None):
             ):
                 # Even though there's only a single value in the querydict for this key, the form wants it as a list
                 result[key] = value_list
+            elif filterset is not None and filterset.filters.get(key) and not is_single_choice_field(filterset, key):
+                result[key] = value_list
```
glennmatthews, line 187: Would be good to enhance nautobot.core.tests.test_utils test_normailze_querydict() to cover this case. (same comment also left on Bulk Delete Job PR)

nautobot/dcim/forms.py
```diff
@@ -588,7 +588,7 @@ class RackBulkEditForm(
         required=False,
         widget=StaticSelect2(),
     )
-    u_height = forms.IntegerField(required=False, label="Height (U)")
+    u_height = forms.IntegerField(required=False, label="Height (U)", min_value=1, max_value=100)
```
glennmatthews, line 591: Why this change?

nautobot/dcim/forms.py
```diff
@@ -5324,6 +5324,16 @@ class Meta:
             "tenant",
         ]
 
+    def post_save(self, obj):
```
glennmatthews, line 5327: Why this addition?

nautobot/dcim/tests/test_views.py
```diff
@@ -1136,40 +1136,6 @@ def test_devicetype_export(self):
         self.assertEqual(data[0]["manufacturer"], device_type.manufacturer.name)
         self.assertEqual(data[0]["model"], device_type.model)
 
-    @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
-    def test_rack_height_bulk_edit_set_zero(self):
```
glennmatthews, line 1140: Feels like this logic should be reimplemented in a Job-based test case rather than removed entirely?

nautobot/extras/forms/forms.py
```diff
@@ -1136,6 +1136,57 @@ class JobBulkEditForm(NautobotBulkEditForm):
     class Meta:
         model = Job
 
+    def post_save(self, obj):
+        cleaned_data = self.cleaned_data
+
+        if self.cleaned_data.get("add_tags", None):
+            obj.tags.add(*self.cleaned_data["add_tags"])
+        if self.cleaned_data.get("remove_tags", None):
+            obj.tags.remove(*self.cleaned_data["remove_tags"])
```
glennmatthews, line 1145: Shouldn't this be automatic now?

nautobot/dcim/forms.py
timizuoebideri1: Here, the form field name of the model field name are different, while the super().post expects both name and field name to be same

nautobot/core/jobs/bulk_actions.py
timizuoebideri1: I could change it to `form_data`, as `pk_list` is part of the `BulkEdit` form fields.

nautobot/core/tests/test_jobs.py
timizuoebideri1: on Line 666

nautobot/dcim/forms.py
timizuoebideri1: Rack.u_height has a min and max validation, adding it here so that we can catch it early

nautobot/extras/forms/forms.py
timizuoebideri1: if I call super().post_save

nautobot/core/jobs/bulk_actions.py
glennmatthews: Understood - I was wondering if it'd make sense to split it out from the rest of the form data into a separate Job input. Fine if not.

nautobot/extras/forms/forms.py
glennmatthews: Shouldn't we?

glennmatthews (commented):
Aside from fixing the integration tests, this looks solid.

glennmatthews (approved):

Issue #6509, opened 2024-11-12
Intermittent test failures in `next` branch


gsnider2195:
This test has been failing occasionally in `next` on the `mysql-python3.12` unittest step.

https://github.com/nautobot/nautobot/actions/runs/11806093120/job/32890279349

```
======================================================================
FAIL: test_list_objects_unknown_filter_no_strict_filtering (nautobot.extras.tests.test_views.ObjectMetadataTestCase.test_list_objects_unknown_filter_no_strict_filtering)
Verify that without STRICT_FILTERING, an unknown filter is ignored.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/unittest/case.py", line 58, in testPartExecutor
    yield
  File "/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/unittest/case.py", line 634, in run
    self._callTestMethod(testMethod)
    ^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/unittest/case.py", line 589, in _callTestMethod
    if method() is not None:
    ^^^^^^^^^^^^^^^^^
  File "/home/runner/.cache/pypoetry/virtualenvs/nautobot-zO_ZO7c3-py3.12/lib/python3.12/site-packages/django/test/utils.py", line 461, in inner
    return func(*args, **kwargs)
    ^^^^^^^^^^^^^^^^^
  File "/home/runner/work/nautobot/nautobot/nautobot/core/testing/views.py", line 799, in test_list_objects_unknown_filter_no_strict_filtering
    self.assertIn(instance2.get_absolute_url(), content, msg=content)
    ^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/unittest/case.py", line 1152, in assertIn
    self.fail(self._formatMessage(msg, standardMsg))
    ^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/unittest/case.py", line 715, in fail
    raise self.failureException(msg)
    ^^^^^^^^^^^^^^^^^
AssertionError: '/extras/object-metadata/69aa3c8c-8605-4cf1-8242-1a7652df21f4/' not found in  ...
```

glennmatthews:
[...]

PR #6517, opened 2024-11-13
[Kubernetes] Scheduled Jobs

### LINKED ISSUE #6323: k8s Jobs - Scheduled Jobs
Need to account for scheduled jobs that run in the k8s context.

Needs design, but an option might be further enhance our already custom celery beat scheduler class to account for this.

Also need to account for any changes to Job Approvals.

HanlinMiao:
# Closes #6323 
# What's Changed

- [x] Overriden `NautobotDataBaseScheduler.apply_async` method to distinguish between Celery and Kubernetes Jobs
- [x] Added `celery-beat-deployment.yaml`
- [x] Restored `docker-compose.min.yaml`  in case that re-generations of Kubernetes deployment files are needed in the future

# Screenshots

1. Run `kubectl apply -f development/kubernetes/celery-beat-deployment.yaml`
2. See the Kubernetes pod setup
Screenshot 2024-11-13 at 11 53 02 AM

3. Run `kubectl exec --stdin --tty nautobot-679bdc765-snfps -- /bin/bash`  and then `nautobot-server nbshell` to create a Scheduled Job in `nbshell`
Screenshot 2024-11-13 at 11 53 45 AM

4. See that a Kubernetes Job and Kubernetes Job Pod is created
Screenshot 2024-11-13 at 11 53 11 AM
Screenshot 2024-11-13 at 11 56 14 AM

5. See that a new Job Result with the status `SUCCESS` is created.
Screenshot 2024-11-13 at 11 53 56 AM

# TODO

- [ ] Unit, Integration Tests


nautobot/core/celery/schedulers.py
```diff
@@ -111,8 +117,44 @@ def apply_async(self, entry, producer=None, advance=True, **kwargs):
...
+                        name=scheduled_job.job_model.name,
+                        job_model=scheduled_job.job_model,
+                        scheduled_job=scheduled_job,
+                        user=scheduled_job.user,
+                    )
+                    job_result = run_kubernetes_job_and_return_job_result(
+                        job_queue, job_result, json.dumps(entry_kwargs)
+                    )
+                    # Return an AsyncResult object to mimic the behavior of Celery tasks after the job is finished by Kubernetes Job Pod.
+                    resp = AsyncResult(job_result.id)
```
glennmatthews, line 142: Can we add a unit test for this new code (probably mocking out `run_kubernetes_job_and_return_job_result()` calls for the time being)? I don't think we normally have beat running during unit tests, but it should be possible to at least directly call the `apply_async()` method and verify that it works?

nautobot/extras/management/utils.py
```diff
@@ -13,7 +12,7 @@ def validate_job_and_job_data(command, user, job_class_path, data=None):
     job_data = {}
     try:
         if data:
-            job_data = json.loads(data, cls=NautobotKombuJSONEncoder)
+            job_data = json.loads(data)
```
glennmatthews, line 15: Anything to add testing-wise about this logic?

nautobot/core/celery/schedulers.py
```diff
@@ -111,8 +117,44 @@ def apply_async(self, entry, producer=None, advance=True, **kwargs):
...
+                else:
+                    resp = task.apply_async(entry_args, entry_kwargs, producer=producer, **entry.options)
```
glennmatthews, line 144: Is there anything we need to do with the `entry.options` in the Kubernetes case?

glennmatthews (commented):
Can you compare the resulting JobResult from a k8s scheduled job versus a Celery scheduled job to make sure they're both fully/correctly populated?

glennmatthews:
unittest failure is #6509. I'm going to pick that one up and look into it.

nautobot/core/celery/schedulers.py
HanlinMiao: This is one of the example `entry.options` it seems like we encapsulated all the information in the job_result we created above.
```
{'nautobot_job_scheduled_job_id': UUID('320bc2c1-6e98-48db-abd7-7c58c26fdd26'), 'headers': {}, 'nautobot_job_user_id': UUID('970ef2cc-d8e8-456f-90f2-1c45bac440ec'), 'nautobot_job_job_model_id': UUID('80f97ef5-5dfe-4bd7-8c2a-dafdb0b12870')}
```

nautobot/extras/tests/test_management.py
```diff
@@ -55,3 +59,29 @@ def test_populate_role_choices_idempotent(self):
...
         populate_role_choices(apps=apps, schema_editor=None)
         self.assertEqual(Role.objects.count(), initial_roles_count)
+
+
+class JobManagementTestCase(TestCase):
+    """Test for the nautobot-server runjob helper function."""
+
+    def test_validate_job_and_job_data(self):
+        """
+        Test the validate_job_and_job_data function.
```
glennmatthews, line 69: You're not actually calling that function anywhere that I see though?

HanlinMiao: The only thing we do to `job_data` in `validate_job_and_job_data()` is calling `json.loads()`. I am not sure comment on the next line is clear or not. Let me know if should be clearer.

HanlinMiao:
> Can you compare the resulting JobResult from a k8s scheduled job versus a Celery scheduled job to make sure they're both fully/correctly populated?

Here is a comparison of the fields populated. `Kubernetes` is on the left and `Celery` is on the right. The main differences lie within the `celery_kwargs` field. I wonder if it is acceptable since we are not using celery workers to execute the job anyway.
Screenshot 2024-11-15 at 2 48 09 PM

glennmatthews:
Regarding the difference in `celery_kwargs`, do the `nautobot_job_profile` and time-limit parameters get set appropriately for the executing job in some other fashion?

Can we fix up the difference in `name` (Job.name versus Job.class_path)?

nautobot/extras/tests/test_management.py
glennmatthews: I feel like the test as written isn't doing anything particularly useful/valuable (we should be able to assume that json.dumps/json.load work as expected). Probably don't need to add/keep this test.

glennmatthews (approved):

PR #6504, opened 2024-11-12
feature: implement app marketplace view

norbert-mieczkowski-codilime:
# Closes #6432
# What's Changed

Implement new Marketplace page under Apps -> General -> Marketplace (`/apps/marketplace/`). The view includes all available apps with already installed apps filtered out.

**Please note that already existing Installed Apps page is not affected by this PR.**
# Screenshots

| Marketplace | App Details |
| - | - |
| marketplace | marketplace_app |

PR #6539, opened 2024-11-18
feature: implement installed apps page tile view

### LINKED ISSUE #6432: App Marketplace View
As a Nautobot user, I want a view a marketplace of available apps in the Nautobot ecosystem so I can learn about solutions that are already out there and that I can install in my environment.

- This should be a new view in conjunction with the existing Installed Apps view, that shows details of apps in the marketplace.
- The UI should mimic the card/tile based view for Jobs, where each tile is an app
- The existing Installed Apps view, should also be converted to this style, however the user still needs to easily differentiate between their installed apps, and those available in the marketplace. Plus, we must maintain the existing installed apps view/url such that it is not a breaking change.
  - For Installed Apps with no logo, we will have a default logo to show
- App data will be sourced from a local repository that contains all the information needed, like app name, description, author, icon, etc.
  - Out of scope of this story, but still relevant: this "repository file" will be generated every so often from an NTC internal repo containing all of the same information. This means the marketplace data for the new view will be sourced local to each instance.

norbert-mieczkowski-codilime:
# Closes #6432
# What's Changed

**This PR is continuation of #6504**

Implement Installed Apps page tile view. This includes:
1. Tiles, similar to previously implemented Apps Marketplace tiles (see attached screenshot).
2. List/Tiles view switch. Selected view is persisted in user preferences the same way as views on Jobs page.
3. Link to Apps Marketplace page in the upper right corner of the page (next to Configure table button).
4. Disabled Configure table button when `tiles` display is on.
5. Breadcrumbs at the top of the page.
# Screenshots

installed_apps_tiles

### DIFF
--- a/nautobot/extras/templates/extras/plugins_list.html
+++ b/nautobot/extras/templates/extras/plugins_list.html
@@ -15,10 +15,10 @@
 
     <div class="pull-right noprint">
         <div class="btn-group">
-            <a class="btn btn-default set-plugins_list-display" data-display="tiles" href="" title="Tiles">
+            <a class="btn btn-default set-apps-display" data-display="tiles" href="" title="Tiles">
                 <i class="mdi mdi-view-grid{% if display == "list" %}-outline{% endif %}"></i>
             </a>
-            <a class="btn btn-default set-plugins_list-display" data-display="list" href="" title="List">
+            <a class="btn btn-default set-apps-display" data-display="list" href="" title="List">
                 <i class="mdi mdi-view-sequential{% if display == "tiles" %}-outline{% endif %}"></i>
             </a>
         </div>
@@ -58,7 +58,7 @@
 
 <script>
     // Set display anchor links without losing other potentially active query params such as applied filters.
-    [...document.querySelectorAll('.set-plugins_list-display')].forEach(button => {
+    [...document.querySelectorAll('.set-apps-display')].forEach(button => {
         const queryParams = new URLSearchParams(window.location.search);
         queryParams.set('display', button.dataset.display);
         button.setAttribute('href', `?${queryParams.toString()}`);
--- a/nautobot/extras/templates/extras/plugins_list.html
+++ b/nautobot/extras/templates/extras/plugins_list.html
@@ -24,7 +24,11 @@
         </div>
 
         {% if request.user.is_authenticated and table_config_form %}
-            {% table_config_button table table_name="ObjectTable" %}
+            {% if display == "tiles" %}
+                {% table_config_button table table_name="ObjectTable" disabled=True %}
+            {% else %}
+                {% table_config_button table table_name="ObjectTable" %}
+            {% endif %}
         {% endif %}
 
         <a class="btn btn-primary" href="{% url 'apps:apps_marketplace' %}">Visit Apps Marketplace</a>
--- a/nautobot/extras/templates/extras/plugins_list.html
+++ b/nautobot/extras/templates/extras/plugins_list.html
@@ -3,10 +3,31 @@
 {% load static %}
 
 {% block content %}
+    <div class="row noprint">
+        <div class="col-md-12">
+            <ol class="breadcrumb">
+                {% block breadcrumbs %}
+                    <li><a href="{% url 'apps:apps_list' %}">Installed Apps</a></li>
+                {% endblock breadcrumbs %}
+            </ol>
+        </div>
+    </div>
+
     <div class="pull-right noprint">
+        <div class="btn-group">
+            <a class="btn btn-default set-plugins_list-display" data-display="tiles" href="" title="Tiles">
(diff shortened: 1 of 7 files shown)


nautobot/extras/templates/extras/plugins_list.html
```diff
@@ -3,10 +3,35 @@
...
+            </a>
+        </div>
+
         {% if request.user.is_authenticated and table_config_form %}
-            {% table_config_button table table_name="ObjectTable" %}
+            {% if display == "tiles" %}
+                {% table_config_button table table_name="ObjectTable" disabled=True %}
+            {% else %}
+                {% table_config_button table table_name="ObjectTable" %}
+            {% endif %}
```
norbert-mieczkowski-codilime, line 31: I was banging my head against the wall how to make `disabled` a boolean rather than to duplicate the same `table_config_button` block with and without `disabled` arg

nautobot/extras/templates/extras/plugins_list.html
```diff
@@ -3,10 +3,35 @@
...
+
     <div class="pull-right noprint">
+        <div class="btn-group">
+            <a class="btn btn-default set-plugins_list-display" data-display="tiles" href="" title="Tiles">
+                <i class="mdi mdi-view-grid{% if display == "list" %}-outline{% endif %}"></i>
+            </a>
+            <a class="btn btn-default set-plugins_list-display" data-display="list" href="" title="List">
+                <i class="mdi mdi-view-sequential{% if display == "tiles" %}-outline{% endif %}"></i>
+            </a>
+        </div>
```
norbert-mieczkowski-codilime, line 24: copied from jobs: https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/templates/extras/job_list.html#L31:L38

nautobot/extras/templates/extras/plugins_list.html
```diff
@@ -30,4 +55,13 @@ <h1>{% block title %}Installed Apps{% endblock %}</h1>
...
 {{ block.super }}
 <script src="{% static 'js/tableconfig.js' %}"></script>
+
+<script>
+    // Set display anchor links without losing other potentially active query params such as applied filters.
+    [...document.querySelectorAll('.set-plugins_list-display')].forEach(button => {
+        const queryParams = new URLSearchParams(window.location.search);
+        queryParams.set('display', button.dataset.display);
+        button.setAttribute('href', `?${queryParams.toString()}`);
+    });
```
norbert-mieczkowski-codilime, line 65: copied from jobs: https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/templates/extras/job_list.html#L81:L86

nautobot/project-static/css/base.css
```diff
@@ -934,7 +934,7 @@ code.hljs {
 .tiles {
     display: grid;
     grid-gap: 20px;
-    grid-template-columns: repeat(auto-fit, minmax(445px, 1fr));
+    grid-template-columns: repeat(auto-fill, minmax(445px, 1fr));
 }
```
norbert-mieczkowski-codilime, line 938: I missed that before when I worked on pages with many elements but `fill` feels more reasonable for tiles rather than `fit`. The former respects the number of grid columns regardless of number of elements inside, while the latter simply spans elements across the whole container width if there are more columns than chlidren. Here's comparison:

| `auto-fill` | `auto-fit` |
| - | - |
| fill | fit |

nautobot/extras/templates/extras/plugins_list.html
glennmatthews: can you do `disabled=(display == "tiles")`?

Otherwise I think this is fine.

nautobot/extras/templates/extras/plugins_list.html
```diff
@@ -3,10 +3,35 @@
...
+                {% block breadcrumbs %}
+                    <li><a href="{% url 'apps:apps_list' %}">Installed Apps</a></li>
+                {% endblock breadcrumbs %}
+            </ol>
+        </div>
+    </div>
+
     <div class="pull-right noprint">
+        <div class="btn-group">
+            <a class="btn btn-default set-plugins_list-display" data-display="tiles" href="" title="Tiles">
```
glennmatthews, line 18: nit: change these to `set-apps-display` perhaps?

nautobot/extras/templates/extras/plugins_list.html
norbert-mieczkowski-codilime: unfortunately not, this does not work either :cry:

nautobot/extras/templates/extras/plugins_list.html
norbert-mieczkowski-codilime: sure, not a problem 8be0cbd9d9f69b37dee06bb7c91b78eeecf09e9e

glennmatthews (approved):

PR #6540, opened 2024-11-18
UI Component Framework: Buttons

### LINKED ISSUE #6142: UI Components: Buttons
Support detail view buttons.

glennmatthews:
# Closes #6142 
# What's Changed

- Add `Button` and `DropdownButton` classes to UI Component Framework
- Allow `ObjectDetailContent` to declare `extra_buttons` consisting of the above
   - Convert `Secret` views to use NautobotUIViewSet + Component Framework
   - Convert `Device` "Add Components" dropdown button to Component Framework (converting the full view is out of scope at this time)
- Allow `TemplateExtension` to declare `object_detail_buttons` and deprecate existing `buttons()` API.
   - Partially convert `example_app` to use the new pattern.

# Screenshots

image

image

# TODO

- [ ] Unit, Integration Tests


nautobot/core/ui/object_detail.py
```diff
@@ -713,7 +821,10 @@ def __init__(
                 - `[render_markdown, placeholder]` - render the given text as Markdown, or render a placeholder if blank
                 - `[humanize_speed, placeholder]` - convert the given kbps value to Mbps or Gbps for display
         """
+        if data and context_data_key:
+            raise ValueError("The data and context_data_key parameters are mutually exclusive")
```
HanlinMiao, line 825: Should we also check if one of them is specified as well? What would a KeyValueTablePanel look like if no data is provided?

EDIT: seems like all the other subclasses can be benefited from this change as well.

nautobot/core/ui/object_detail.py
```diff
@@ -922,12 +1033,13 @@ def __init__(
...
         if self.label is None:
-            return bettertitle(get_obj_from_context(context)._meta.verbose_name)
+            return bettertitle(get_obj_from_context(context, self.context_object_key)._meta.verbose_name)
         return super().render_label(context)
 
     def render_value(self, key, value, context: Context):
+        obj = get_obj_from_context(context, self.context_object_key)
         try:
-            field_instance = get_obj_from_context(context)._meta.get_field(key)
+            field_instance = obj._meta.get_field(key)
```
HanlinMiao, line 1042: nitpick:
Should we set `self.obj = get_obj_from_context(context, self.context_object_key)` so we can reference it in these functions?

nautobot/core/ui/object_detail.py
```diff
@@ -953,7 +1070,7 @@ def get_data(self, context: Context):
             (dict): Key-value pairs corresponding to the object's fields, or `{}` if no object is present.
         """
         fields = self.fields
-        instance = context.get(self.context_object_key, None)
+        instance = get_obj_from_context(context, self.context_object_key)
```
HanlinMiao, line 1073: Same comment here as above regarding `self.obj`

nautobot/core/views/utils.py
```diff
@@ -98,8 +98,14 @@ def csv_format(data):
...
-    """From the given context, return the object that is in the context"""
+def get_obj_from_context(context, key=None):
+    """From the given context, return the `object` that is in the context.
+
+    If a key is specified, return the value for that key.
+    Otherwise return the value for either of the keys `"obj"` or `"object"` as default behavior.
+    """
+    if key is not None:
+        return context.get(key)
     return context.get("obj") or context.get("object")
```
HanlinMiao, line 109: Coud this be `context.get(key) or context.get("obj") or context.get("object")` or will it throw an error when `key` is None?

nautobot/dcim/urls.py
```diff
@@ -1319,6 +1373,11 @@
         views.DeviceBulkAddInventoryItemView.as_view(),
         name="device_bulk_add_inventoryitem",
     ),
+    path(
+        "devices/<uuid:pk>/inventory-items/add/",
+        RedirectView.as_view(url="/dcim/inventory-items/add/?device=%(pk)s&return_url=/dcim/devices/%(pk)s/inventory/"),
```
HanlinMiao, line 1378: Should this be?
```suggestion
        RedirectView.as_view(url="/dcim/inventory-items/add/?device=%(pk)s&return_url=/dcim/devices/%(pk)s/inventory-items/"),
```

nautobot/docs/development/apps/api/ui-extensions/object-views.md
```diff
@@ -2,22 +2,25 @@
...
 
 * `object_detail_tabs` - List of `Tab` instances to add to the detail view as additional tabs.
+* `object_detail_buttons` - List of `Button` instances to add to the detail view.
 * `object_detail_panels` - List of `Panel` instances to add to the main tab of the detail view.
 * `left_page()` - Inject content on the left side of the object detail page (deprecated since Nautobot 2.4.0; `object_detail_panels` is preferred)
 * `right_page()` - Inject content on the right side of the object detail page (deprecated since Nautobot 2.4.0; `object_detail_panels` is preferred)
 * `full_width_page()` - Inject content across the entire bottom of the object detail page (deprecated since Nautobot 2.4.0; `object_detail_panels` is preferred)
-* `buttons()` - Add buttons to the top of the object detail page
+* `buttons()` - Add buttons to the top of the object detail page (deprecrated since Nautobot 2.4.0; `object_detail_buttons` is preferred)
```
HanlinMiao, line 13:
```suggestion
* `buttons()` - Add buttons to the top of the object detail page (deprecated since Nautobot 2.4.0; `object_detail_buttons` is preferred)
```

nautobot/docs/development/core/templates.md
```diff
@@ -32,8 +35,12 @@ The most customizable template is `generic/object_detail.html`, as object detail
     * `nav_tabs`: are the navigation tabs. If overloaded, custom tabs can be
       rendered instead of the default.
         * `extra_nav_tabs`: this block allows to add new tabs without having to
-          override the default ones.
-* `content`: is the entire content of the page below the `header`.
+          override the default ones. Note that since v2.4.0 you can also define
+          `extra_tabs` in your view's [`object_detail_content`](ui-component-framework.md)
+          rather than overriding and extending the template.
+* `content`: is the entire content of the page below the `header`. Note that since v2.4.0,
+  if your view defines [`object_detail_content`](ui-component-framework.md), that content will
```
HanlinMiao, line 42: Should we redirect them to the relevant section of the documentation? e.g. ObjectDetailContent Definition

nautobot/extras/models/secrets.py
```diff
@@ -46,6 +46,11 @@ class Meta:
     def __str__(self):
         return self.name
 
+    def get_provider_display(self):
+        if provider := registry["secrets_providers"].get(self.provider):
```
HanlinMiao, line 50: For my own edification, What does `:=` do in this context?

nautobot/extras/views.py
```diff
@@ -2589,38 +2594,56 @@ def get_extra_context(self, request, instance):
 #
 
 
-class SecretListView(generic.ObjectListView):
+class SecretUIViewSet(
```
HanlinMiao, line 2597: I noticed that there is no way to add a `secret` to a `secret group` from the secret side. I wonder if it is a miss or an intentional thing.

nautobot/dcim/urls.py
glennmatthews: Nope! Inventory-item URLs are weird.

nautobot/extras/models/secrets.py
glennmatthews: Walrus operator - https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions

Basically a shorter way to say:

```py
provider = registry["secrets_providers"].get(self.provider)
if provider:
```

nautobot/extras/views.py
glennmatthews: Since assigning a secret to a group requires specifying additional attributes on the many-to-many through table we'd have to figure out what the UI would need to look like.

examples/example_app/example_app/template_content.py
```diff
@@ -173,6 +175,16 @@ class ExampleModelContent(TemplateExtension):
     model = "example_app.examplemodel"
     template_name = "example_app/panel.html"
 
+    object_detail_buttons = (
+        Button(
+            weight=100,
+            label="Example App Button",
+            color=ButtonColorChoices.BLUE,
+            icon="mdi-plus-thick",
```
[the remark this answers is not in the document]
glennmatthews: It's only on the example-app page but I see your point. Any preferences for an arbitrary icon?

nautobot/core/ui/object_detail.py
glennmatthews: Subclasses might choose to specify neither `data` nor `context_data_key` and instead implement a custom `get_data(context)` function instead. Agreed it's a bit messy to have so many ways to do things though.

nautobot/core/ui/object_detail.py
glennmatthews: Potentially, though we'd have to define which function gets called "first" to do this assignment. Can't do it in `__init__()` since we don't have context at that point.

examples/example_app/example_app/template_content.py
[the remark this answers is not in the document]
HanlinMiao: Perhaps the information icon? https://pictogrammers.com/library/mdi/icon/information/

nautobot/dcim/urls.py
```diff
@@ -598,46 +599,99 @@
         views.DeviceConsolePortsView.as_view(),
         name="device_consoleports",
     ),
+    path(
+        "devices/<uuid:pk>/console-ports/add/",
+        RedirectView.as_view(
+            url="/dcim/console-ports/add/?device=%(pk)s&return_url=/dcim/devices/%(pk)s/console-ports/"
+        ),
```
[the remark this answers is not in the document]
mateusz-lachowski-codilime: Those url's are pretty generic - can't we just generate separate list just and merge it with `urlpatterns` ?

For example:
```
redirection_urls = [path(f"path generation", RedirectView(...), name="") for url in ['console-ports', 'console-server-ports']]
```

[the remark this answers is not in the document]
glennmatthews: It'd be a different code pattern than we currently have in this file, but maybe? I do remember there being a couple of exceptions to the generic pattern though, like `inventory`/`inventory-items`.

examples/example_app/example_app/views.py
```diff
@@ -38,6 +39,7 @@ class DeviceDetailAppTabOneView(views.ObjectView):
 
     queryset = Device.objects.all()
     template_name = "example_app/tab_device_detail_1.html"
+    object_detail_content = ui.ObjectDetailContent(extra_buttons=DeviceView.object_detail_content.extra_buttons)
```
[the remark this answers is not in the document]
mateusz-lachowski-codilime: Maybe just buttons are a different type than `object_detail_content` - there are tied more with detail page as a whole rather than individual "tab" - custom or not. I can think of two possible solutions:
1. Create separate class field specifically for `page_buttons` and then all of the devices tab's should inherit that.  To still allow users to define custom buttons per tab we can add some logic into main view to combine `page_buttons` with those `extra_buttons` defined in give `object_detail_content`.
2. Create something like `ExtraButtonsMixin` and then `DevicePageButtonsMixin` and add it to all of the device related tabs. In this mixin you can add a logic, again, to merge `page_buttons` with `extra_buttons`.

nautobot/core/ui/object_detail.py
```diff
@@ -153,6 +166,98 @@ def get_extra_context(self, context: Context):
...
+    ):
+        """
+        Initialize a Button component.
+
+        Args:
+            label (str): The text of this button, not including any icon.
+            color (ButtonColorChoices): The color (class) of this button.
+            link_name (str, optional): View name to link to, for example "dcim:locationtype_retrieve".
+                This link will be reversed and will automatically include the current object's PK as a parameter to the
+                `reverse()` call when the button is rendered.
```
mateusz-lachowski-codilime, line 193: Don't we need and option for "external links" as well, to just pass the `https://...` url?

glennmatthews: We undoubtedly will at some point, yeah.

examples/example_app/example_app/views.py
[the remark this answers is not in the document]
glennmatthews: In theory `object_detail_content` defines the entire page rather than just the main tab, we just provide some shortcuts to make it easier to directly manipulate the main tab since that's the 80% use case. That's probably not clearly documented as a design goal though.

A couple other possibilities that occur to me:

- Require/recommend custom tabs for a detail view to inherit from the detail view class itself (e.g., `dcim.views.DeviceView`) rather than from `generic.ObjectView`. This wouldn't work for UIViewSet views though so we'd need a different pattern for those. Plus this makes app tabs more tightly coupled to the base view implementation such that converting a base view from an ObjectView to a UIViewSet would become a breaking change. So maybe not.
- Add another template filter used in `object_retrieve.html` that takes the `obj` from the context, does a `get_view_for_model(obj)` to get the base detail view, and injects the `extra_buttons` from that view automatically? I think this has some potential, I'll try and explore it today.

nautobot/extras/templates/extras/secret.html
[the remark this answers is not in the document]
glennmatthews: Done!

nautobot/core/ui/object_detail.py
glennmatthews: Added `get_link()` method for ease of extension by subclasses.

HanlinMiao (approved):

nautobot/dcim/urls.py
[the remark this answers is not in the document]
mateusz-lachowski-codilime: After I scrolled through this file I definitely start thinking about some new code pattern in this file anyway 😱

mateusz-lachowski-codilime (approved):

PR #6541, opened 2024-11-18
Add Success log level.

### LINKED ISSUE #6387: Colored Job Log Results
### As ...

### I want ...

To be able to color the background of my info log messages any color. Implementation could be: `self.logger.info("Repository successfully refreshed", extras={"color": "green"})` It would be nice if it would support Hex etc. Might need to include `color-hover` as well.

Bonus point: If I could change the "level" as well but that's nit-picking. It's fine that it says info. The bigger human eye catcher is the color.

### So that ...

If I have a successful log then I can indicate the log by color.
image

I create a job that uses the built-in logger.

- [X] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation

### Database Changes

I am not sure - probably add a column for color to JobResult 🤷🏻 

### External Dependencies

_No response_

smk4664:
# Closes #6387 
# What's Changed

Adds the custom "SUCCESS" log level and the logger.success log function to allow job developers to add a a log for success events. This adds back the highlighting we had from log_success.

# Screenshots
Screenshot 2024-11-18 at 10 33 55 AM

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/__init__.py
+++ b/nautobot/__init__.py
@@ -22,9 +22,6 @@ def add_success_logger():
             self._log(SUCCESS, message, args, **kws)
 
     logging.Logger.success = success
-
-    logger = logging.getLogger(__name__)
-    logger.success("Custom log level 'SUCCESS' added.")
     return success
 
 
--- a/nautobot/__init__.py
+++ b/nautobot/__init__.py
@@ -11,6 +11,24 @@ __version__ = metadata.version(__name__)
 # Sentinel to make sure we only initialize once.
 __initialized = False
 
+
+def add_success_logger():
+    """Add a custom log level for success messages."""
+    SUCCESS = 25
+    logging.addLevelName(SUCCESS, "SUCCESS")
+
+    def success(self, message, *args, **kws):
+        if self.isEnabledFor(SUCCESS):
+            self._log(SUCCESS, message, args, **kws)
+
+    logging.Logger.success = success
+
+    logger = logging.getLogger(__name__)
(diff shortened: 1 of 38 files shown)


nautobot/__init__.py
```diff
@@ -11,6 +11,21 @@
 # Sentinel to make sure we only initialize once.
 __initialized = False
 
+
+def add_success_logger():
+    """Add a custom log level for success messages."""
+    SUCCESS = 25
```
HanlinMiao, line 17: Should 25 be an environment variable or a settings variable?

smk4664: This happens before settings are loaded, it would have to be an environment variable, but care would need to be taken to ensure that it does not use the same level as the Logging standard levels.

HanlinMiao (approved):

nautobot/__init__.py
smk4664: Usually you don't change the levels, you are not going to change any of the Python standard levels. I believe 25 is a good level number, because if you have INFO, you probably want SUCCESS, but if you only want WARNING and ERRORS, your not going to see the Successes in your logging handler (like if you forward your logs to splunk). This also allows the LoggingHandlers that are already configured for INFO to also pick up the new SUCCESS logs.

glennmatthews: Agreed - this doesn't need to be user definable.

PR #6545, opened 2024-11-18
Fixed unexpected failure when adding an interface component to a device from the Device detail view

### LINKED ISSUE #6537: Adding an Interface component to a Device fails
### Environment
* Nautobot version (Docker tag too if applicable): `next` branch - as of 2024-Nov-13
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Create a Device instance in UI
2. Attempt to add Interface component through Device detail view (GUI)
3.

### Observed Behavior

Error is returned during validation:
```
Virtual Device Contexts
Ethernet1: "" is not a valid UUID.
```

This might potentially be related, but as a part of testing tried attaching an Interface to a VDC and this resulted in a different error message:
```
ValidationError at /dcim/interfaces/add/
{'virtual_device_contexts': ["Virtual Device Context with names ['test'] must all belong to the same device as the interface's device."]}
```

HanlinMiao:
# Closes #6537 
# What's Changed

Fixed unexpected failure when adding an interface component to a device from the Device detail view.
More specifically fixed `virtual_device_contexts` field on `InterfaceCreateForm` to `DynamicMultipleModelChoiceField` from the incorrect `DynamicModelChoiceField`.
# Screenshots

Screenshot 2024-11-18 at 4 16 11 PM

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/core/testing/views.py
+++ b/nautobot/core/testing/views.py
@@ -470,7 +470,6 @@ class ViewTestCases:
                 "path": self._get_url("edit", instance),
                 "data": utils.post_data(update_data),
             }
-            print(self.client.post(**request).content)
             self.assertHttpStatus(self.client.post(**request), 302)
             self.assertInstanceEqual(self._get_queryset().get(pk=instance.pk), update_data)
 
--- a/nautobot/core/testing/views.py
+++ b/nautobot/core/testing/views.py
@@ -470,6 +470,7 @@ class ViewTestCases:
                 "path": self._get_url("edit", instance),
                 "data": utils.post_data(update_data),
             }
+            print(self.client.post(**request).content)
             self.assertHttpStatus(self.client.post(**request), 302)
             self.assertInstanceEqual(self._get_queryset().get(pk=instance.pk), update_data)
 
--- a/nautobot/dcim/tests/test_api.py
+++ b/nautobot/dcim/tests/test_api.py
@@ -3488,6 +3488,21 @@ class VirtualDeviceContextTestCase(APIViewTestCases.APIViewTestCase):
             vdc.refresh_from_db()
             self.assertEqual(vdc.primary_ip6, ip_v6)
 
+    def test_changing_device_on_vdc_raise_validation_error(self):
+        """
+        Validate that changing device on the virutal device context is not allowed.
+        """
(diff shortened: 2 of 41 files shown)


glennmatthews:
Does this address both of the symptoms described in #6537 or only the first one?

HanlinMiao:
> Does this address both of the symptoms described in #6537 or only the first one?

The first one is fixed. The second is working as intended but I think it's incomplete. When the `device` on the virtual device context is changed, we do not check if the changed device is the same as that of the assigned interfaces. Im going to try and add that.

HanlinMiao:
Screenshot 2024-11-19 at 11 36 57 AM

nautobot/dcim/tests/test_api.py
```diff
@@ -3488,6 +3488,21 @@ def test_patching_primary_ip_success(self):
             vdc.refresh_from_db()
             self.assertEqual(vdc.primary_ip6, ip_v6)
 
+    def test_changing_device_on_vdc_raise_validation_error(self):
```
glennmatthews, line 3491: Similar test needed in `test_views.py`?

Also, should we add a check in the model `clean()` itself?

HanlinMiao: Hmmm, how do I compare the previous `device` to the new `device` assigned in model `clean()`, any hints?

HanlinMiao: Also, I am not sure if a test is needed in test_views.py. I am just disabling the `device` field and marking it as `non-required`. No validation error should be raised.

glennmatthews: https://github.com/nautobot/nautobot/blob/develop/nautobot/ipam/models.py#L1142-L1146

nautobot/core/testing/views.py
```diff
@@ -470,6 +470,7 @@ def test_edit_object_with_permission(self):
                 "path": self._get_url("edit", instance),
                 "data": utils.post_data(update_data),
             }
+            print(self.client.post(**request).content)
```
glennmatthews, line 473: remove :)

glennmatthews (approved):

PR #6555, opened 2024-11-19
add jinja template renderer view

gsnider2195:
# Closes #DNE (apps zenhub issue 172)
# What's Changed

Add a view to expose the jinja template renderer feature from #6488 through the UI. ~Let me know if the link needs to move to the nav menu or if it should remain in the footer. Currently the `navigation.py` will not work in the core app so we'll have to discuss how we work around that if we decide to move this to the nav menu.~
# Screenshots

## View

image

image

### Small viewport

This is the reason for the `button-container` css class. Otherwise the render button ends up touching the panel below

image

## Footer

image

# TODO

- [ ] Unit, Integration Tests


bryanculver:
@lampwins Can you review the placement of the link?

lampwins:
This has the same "tool" feeling that GraphiQL and Swagger to, so I think it is fine to put it in the footer. We can revisit in v3.

nautobot/core/templates/utilities/render_jinja2.html
```diff
@@ -0,0 +1,117 @@
...
+                        <div class="col-md-9">
+                            <textarea cols="40" rows="10" class="form-control" placeholder="{}" id="id_context">{}</textarea>
+                            <span class="help-block">Enter data in <a href="https://json.org/">JSON</a> format.</span>
+                        </div>
+                    </div>
+                </div>
+            </div>
+            <div class="button-container text-right">
+                <button type="submit" class="btn btn-primary">Render</button>
+                <a href="{{ return_url }}" class="btn btn-default">Cancel</a>
```
glennmatthews, line 49: Where's the `return_url` coming from? Honestly it feels a bit weird to me to have a Cancel button on this page at all?

nautobot/core/templates/utilities/render_jinja2.html
```diff
@@ -0,0 +1,117 @@
...
+</form>
+{% endblock content %}
+
+{% block javascript %}
+    {{ block.super }}
+    <script>
+        new ClipboardJS('.copy-rendered-template');
+        const sanitize = function(string) {
+            return string.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;");
+        };
```
glennmatthews, line 75: Is there something in the JS "standard library" that we could use here that would be more comprehensive?

nautobot/core/views/__init__.py
```diff
@@ -478,3 +478,9 @@ def get_file_with_authorization(request, *args, **kwargs):
     get_object_or_404(queryset, file=request.GET.get("name"))
 
     return get_file(request, *args, **kwargs)
+
+
+class RenderJinjaView(TemplateView):
```
glennmatthews, line 483: Does this give appropriate permission constraints (at least `is_authenticated`)?

nautobot/core/templates/utilities/render_jinja2.html
gsnider2195: I'm ok with removing it completely

nautobot/core/templates/utilities/render_jinja2.html
gsnider2195: Nothing built in to javascript and I don't see anything in our project-static/js. There are complex libraries for preventing xss like https://github.com/cure53/DOMPurify but since we don't want any HTML tags to render here I think this small function is sufficient.

nautobot/core/views/__init__.py
gsnider2195: Nope I totally forgot about auth for this view! Will fix it. I think `is_authenticated` will be ok for this one since this is just calling the REST API and we can do more granular permissions there if necessary?

nautobot/core/templates/utilities/render_jinja2.html
```diff
@@ -0,0 +1,117 @@
...
+<form class="form form-horizontal" onsubmit="handleFormSubmit(event)" id="render_jinja_template_form">
+    <div class="row">
+        <div class="col-lg-6 col-md-6">
+            <div class="panel panel-default">
+                <div class="panel-heading"><strong>Template</strong></div>
+                <div class="panel-body">
+                    <div class="form-group">
+                        <label class="col-md-3 control-label required" for="id_template_code">Jinja Template</label>
+                        <div class="col-md-9">
+                            <textarea cols="40" rows="10" class="form-control" id="id_template_code" required=""></textarea>
```
glennmatthews, line 36: With hard-coded initial `cols` and `rows` on these textareas, are they still reasonably responsive to different window sizes? Or is there a better way to size them?

glennmatthews (approved):
Thanks for updating the release-notes and docs as well!

nautobot/core/templates/utilities/render_jinja2.html
gsnider2195: They are. I actually don't think the cols here do anything because bootstrap sets these to 100% width anyway.

Issue #6344, opened 2024-10-08
ValidationError when removing invalid software image files from device


gsnider2195:
### Environment
* Nautobot version (Docker tag too if applicable): v2.3.6
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Override a device's software image file in the device edit form:

![{37946D1E-CED2-4F01-9CB2-4DC0EC69732F}](https://github.com/user-attachments/assets/ca9480f0-2bf5-4712-81cc-7e4356b2caa0)

2. Remove all device types from the software image edit form
3. Try to edit the device and remove the software image file

### Expected Behavior

The no longer valid software image file is removed

### Observed Behavior

The device clean fails which causes the form to fail validation

![{BFC696CB-1927-4E96-8B6F-21C648C714E8}](https://github.com/user-attachments/assets/891e8c37-5ddb-4735-82fe-83ce607446cb)

**M2M relationships can't be validated in a model's clean method since the M2M is typically modified after the model is already saved.**

glennmatthews:
This issue also makes it possible to *add* invalid software image files in the first place, at least via the REST API.

PR #6581, opened 2024-11-27
Add and use `exclude_m2m` REST API parameter

### LINKED ISSUE #6556: Allow REST API exclude M2M fields
To support cases where many to many fields degrade performance of API responses (even in depth 0 cases), introduce support to ignore/exclude M2M fields in the API response.

So an API consumer can specify `?exclude_m2m=true` and not get any M2M fields at all in the response. This is particularly needed in problem areas like VRF where there are several M2M relationships and frequent opportunity for data scale. The API select2 widget will generally benefit from this across the board.

glennmatthews:
# Closes #6556 
# What's Changed

- Add support for `exclude_m2m=true` as a REST API query parameter. When present, the API serializer will automatically exclude many-to-many fields that would otherwise be automatically included in the serialized data. When used, this can greatly reduce the overhead of calling certain m2m-heavy endpoints such as `/api/ipam/vrfs/` that could otherwise be slow to completely nonresponsive with scaled data.
    - Add `exclude_m2m` field to the OpenAPI schema for appropriate endpoints.
- Add support for basic queryset auto-optimization in REST API viewsets, similar to what was recently added for `BaseTable` in the UI. If the serializer fields include foreign keys or the equivalent, said fields will be automatically `select_related()`, and similarly many-to-many fields and the like will be automatically `prefetch_related()`. This works in combination with `exclude_m2m` so that when `exclude_m2m` is specified, the m2m fields aren't present, and so the corresponding `prefetch_related` calls are not made as they wouldn't be necessary.
    - In support of this, update all REST API viewset `queryset` declarations to remove the select/prefetch fields that are now automatically included.
        - I've been intentionally conservative here, *only* removing the fields that I can verify to now be automatically selected/prefetched. In a number of the DCIM API endpoints, I think we're selecting/prefetching some nested fields that are holdovers from v1.x's custom NestedSerializer classes and don't necessarily apply in the default `depth=0` case any more, but I've left those in for now just to minimize the scope of behavior being changed by this PR.
    - To take full advantage of `exclude_m2m=true`, Apps will need to similarly remove any hardcoded `prefetch_related` calls from their API querysets. It will still work without making such a change, but the database queries will be unnecessarily made in such a case.
- Change `DynamicModelChoiceMixin` (base class for `DynamicModelChoiceField`, `DynamicModelMultipleChoiceField`, etc.) to default to including `exclude_m2m=true` in its `query_params` as 90+% of the time, these select fields do *not* need to know or care about the m2m fields, improving their performance.
- Enhance generic API ListObjectsViewTestCase to test with `exclude_m2m=True` and verify that:
   - M2M fields are correctly excluded from the response
   - Calls with `exclude_m2m=true` decrease the number of database queries (if any m2m fields exist on the model being called) or at least do not increase the number of database queries

The generic test enhancements uncovered a few latent issues:
- The `get_depth_fields()` test helper method was coded incorrectly (using `model._meta.fields` instead of `model._meta.get_fields()`) such that it generally wasn't actually finding any fields to inspect for proper `depth` support. Fixing this uncovered the fact that Contact.teams and Team.contacts REST API fields weren't properly supporting `depth`, so I've fixed that now.
- Testing for M2M field presence/absence found that quite a few REST API endpoints were missing the `tags` field due to not correctly including `TaggedModelSerializerMixin`. I've fixed those.
- Testing for database query counts showed that the `Interface.ip_address_count` REST API field was using an instance property accessor instead of a `count_related` annotation, making it inefficient when using `exclude_m2m=true`. I've fixed that.

**The enhanced generic API tests may introduce new test failures in Apps. These will likely fall into one of several categories:**

- Related field (FK or M2M) is not present in the API but should be - App should correct the relevant serializer definition, e.g. by adding `TaggedModelSerializerMixin` where appropriate
- Related field (FK or M2M) is intentionally *not* present in the API but test now thinks it should be - App can override `get_depth_fields()` and/or `get_m2m_fields()` on the relevant App test case to explicitly discard the relevant field name(s).
- Number of database queries increases when using `exclude_m2m=True` - App will need to investigate the view and serializer implementation to determine the root cause.

# Screenshots

(TODO)

# TODO

- [x] Unit, Integration Tests


glennmatthews:
Possible TODO here: do we need PUT/POST/PATCH endpoints to also support `?exclude_m2m` so that their responses also are smaller in scale?

nautobot/core/api/serializers.py
```diff
@@ -83,8 +83,8 @@ def fields(self):
                 # No available request?
                 return fields
 
-            # opt-in fields only applies on GET requests, for other methods we support these fields regardless
-            if request is not None and request.method != "GET":
+            # opt-in fields only applies on GET / OPTIONS requests, for other methods we support these fields regardless
+            if request is not None and request.method not in ["GET", "OPTIONS"]:
```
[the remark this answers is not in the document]
glennmatthews: Gets info about what the endpoint supports (e.g. encodings, choices-field values, that sort of thing). Mostly this was because I noticed that rendering the browsable API (e.g. accessing `/api/ipam/vrfs/` with a browser) actually instantiates the serializer for both GET and OPTIONS and it seemed weird to have OPTIONS still load the M2M fields and the like.

I'll revisit this and see if it's still needed/useful with my latest approach.

nautobot/core/api/serializers.py
```diff
@@ -591,7 +601,7 @@ def to_internal_value(self, data):
         pk = None
 
         if isinstance(self.Meta.model._meta.pk, models.AutoField):
-            # PK is an int for this model. This is usually the User model
+            # PK is an int for this model. This is usually the Group or Content-Type model
```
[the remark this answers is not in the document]
glennmatthews: We have a custom User model that has a UUID PK.

nautobot/core/testing/api.py
```diff
@@ -231,9 +232,9 @@ def get_filterset(self):
         def get_depth_fields(self):
             """Get a list of model fields that could be tested with the ?depth query parameter"""
             depth_fields = []
-            for field in self.model._meta.fields:
+            for field in self.model._meta.get_fields():
                 if not field.name.startswith("_"):
-                    if isinstance(field, (ForeignKey, GenericForeignKey, ManyToManyField, core_fields.TagsField)) and (
+                    if isinstance(field, (ForeignKey, ManyToManyField, core_fields.TagsField)) and (
```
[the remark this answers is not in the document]
glennmatthews: I had some test failures around GFK fields when I fixed the logic so that we were actually correctly detecting depth fields. I'll revisit and see whether it should be reintroduced.

nautobot/core/testing/api.py
```diff
@@ -280,38 +292,91 @@ def test_list_objects_depth_0(self):
...
+                        len(cqc), base_num_queries, "Number of queries did not decrease with ?exclude_m2m=true"
+                    )
+                else:
+                    # Less strict check for non-core APIs
+                    self.assertLessEqual(
+                        len(cqc), base_num_queries, "Number of queries increased with ?exclude_m2m=true"
+                    )
+            else:
+                # No M2M fields to exclude
+                self.assertLessEqual(len(cqc), base_num_queries, "Number of queries increased with ?exclude_m2m=true")
```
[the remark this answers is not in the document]
glennmatthews: Shouldn't ever, but also I wouldn't personally consider it a test failure if it did decrease for some weird reason?

nautobot/docs/release-notes/version-2.4.md
```diff
@@ -16,6 +16,7 @@ This document describes all new features and changes in Nautobot 2.4.
 
 - App developers should begin to adopt the [UI Component Framework](#ui-component-framework) introduced in Nautobot 2.4, as this will reduce the amount of boilerplate HTML/CSS content that they need to develop and maintain, and will insulate Apps from future CSS changes planned for Nautobot v3.
 - Additionally, App developers should familiarize themselves with the new [Event Publication Framework](#event-publication-framework) and the possibilities it enables for Apps to publish their own relevant events.
+- As a side benefit of adding [REST API `exclude_m2m` support](#rest-api-exclude_m2m-support), the Nautobot REST API `ViewSet` classes now attempt to intelligently apply `select_related()` and/or `prefetch_related()` optimizations to the `queryset` associated to a given REST API viewset. Apps defining their own REST API viewsets (and requiring Nautobot 2.4.0 or later) can typically remove most explicit calls to `select_related()` and `prefetch_related()`; furthermore, in order to benefit most from the `exclude_m2m=true` query paremeter, apps in Nautobot 2.4.0 and later should **not** explicitly `prefetch_related()` their many-to-many related fields any longer.
```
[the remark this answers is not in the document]
glennmatthews: Good thought, I'll need to put some thought into the wording.

nautobot/users/tests/test_api.py
```diff
@@ -88,6 +88,16 @@ def get_deletable_object(self):
...
 
+    def get_depth_fields(self):
+        depth_fields = super().get_depth_fields()
+        depth_fields.remove("user_permissions")
+        return depth_fields
+
+    def get_m2m_fields(self):
+        m2m_fields = super().get_m2m_fields()
+        m2m_fields.remove("user_permissions")
+        return m2m_fields
```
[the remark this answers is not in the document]
glennmatthews: Removed in latest rework, expect new diffs shortly.

nautobot/dcim/tests/test_api.py
```diff
@@ -1469,11 +1469,24 @@ def setUpTestData(cls):
...
+                image_file_name="test_software_image_file_1.bin",
+            ),
+            SoftwareImageFile.objects.create(
+                status=Status.objects.get_for_model(SoftwareImageFile).last(),
+                software_version=software_version,
+                image_file_name="test_software_image_file_2.bin",
+            ),
+        )
+        for software_image_file in software_image_files:
+            software_image_file.device_types.add(device_type)
```
glennmatthews, line 1489: This set of changes is because the set of software_image_files selected were actually invalid for the given `device_type`, but this wasn't being detected due to #6344. When I enhanced `test_update_object` with the additional idempotence step, the invalid data started being detected on the "idempotent" update.

nautobot/ipam/tests/test_api.py
```diff
@@ -348,6 +348,8 @@ def setUpTestData(cls):
         cls.create_data = [
             {
                 "prefix": "192.168.4.0/24",
+                "network": "192.168.4.0",  # not required, but possible to specify
+                "broadcast": "192.168.4.255",  # not required, but possible to specify
```
glennmatthews, line 352: Adding these here was easier than updating the enhanced `test_update_object` test case to understand that setting the `prefix` field in a PATCH would automatically change the `network` and `broadcast` fields correspondingly.

changes/6556.added
```diff
@@ -0,0 +1,3 @@
+Added REST API support for an `?exclude_m2m=true` query parameter. Specifying this parameter prevents the API from retrieving and serializing many-to-many relations on the requested object(s) (for example, the list of all Prefixes associated with a given VRF), which can in some cases greatly improve the performance of the API and reduce memory and network overhead substantially.
+Added automatic optimization of REST API querysets via appropriate `select_related` and `prefetch_related` calls.
+Enhanced generic REST API test methods `ListObjectsViewTestCase.test_list_objects_depth_0` and `ListObjectsViewTestCase.test_list_objects_depth_1` to test the `?exclude_m2m=true` query parameter. This enhancement includes checks for missing related fields in the serialized data, which may result in new test failures being seen in Apps using this test class.
```
glennmatthews, line 3:
```suggestion
Enhanced generic REST API test methods `ListObjectsViewTestCase.test_list_objects_depth_0` and `ListObjectsViewTestCase.test_list_objects_depth_1` to test the `?exclude_m2m=true` query parameter. This enhancement includes checks for missing related fields in the serialized data, which may result in new test failures being seen in Apps using this test class.
Enhanced generic REST API test method `UpdateObjectViewTestCase.test_update_object` to include an idempotency test in combination with the `?exclude_m2m=true` query parameter. This may result in new test failures being seen in Apps using this test class.
```

nautobot/core/testing/api.py
```diff
@@ -798,10 +923,16 @@ def strip_serialized_object(this_object):
             # Verify that a PATCH with some data updates that data correctly.
             response = self.client.patch(url, update_data, format="json", **self.header)
             self.assertHttpStatus(response, status.HTTP_200_OK)
+            serialized_object = response.json()
+            strip_serialized_object(serialized_object)
```
glennmatthews, line 927: This logic missing before meant that the `serialized_object` we were testing below is the one from the previous subsection of this test ("Send empty PATCH request" above), which we already asserted was unchanged, rather than the response to this update request as it should be.

nautobot/core/testing/forms.py
```diff
@@ -18,6 +19,7 @@ def test_form_dynamic_model_choice_fields_query_params(self):
                     continue
                 with self.subTest(f"Assert {self.form_class.__name__}.{field_name} query_params are valid."):
                     query_params_fields = set(field_class.query_params.keys())
+                    query_params_fields = query_params_fields - set(NON_FILTER_QUERY_PARAMS)
```
glennmatthews, line 22: This addition is to keep the test from failing due to the presence of `exclude_m2m=True` as a query parameter.

mateusz-lachowski-codilime (approved):
It would be good to have at least some unit tests for `get_queryset`, `maybe_select_related` and `maybe_prefetch_related`. Other than that it looks like solid piece of work :)

glennmatthews:
Good points - those functions are *exercised* quite extensively by the view code but that's not the same as being *unit tested*. :-) I'll see what I can add.

nautobot/core/api/views.py
```diff
@@ -264,7 +264,7 @@ def get_queryset(self):
                     model_field = model._meta.get_field(field_instance.source)
                 except FieldDoesNotExist:
                     continue
-                if isinstance(model_field, (ManyToManyField, RelatedField, ManyToOneRel, TagsField)):
+                if isinstance(model_field, (ManyToManyField, ManyToManyRel, RelatedField, ManyToOneRel, TagsField)):
```
glennmatthews, line 267: I noticed when adding new tests using the `IPAddress` view that we weren't optimizing the `interfaces` and `vm_interfaces` reverse-m2m lookups. Whoops!

nautobot/extras/api/serializers.py
```diff
@@ -211,10 +211,10 @@ class Meta:
...
 
-    def get_default_field_names(self, declared_fields, model_info):
-        """Add reverse M2M for teams to the "all" fields for this serializer."""
-        field_names = super().get_default_field_names(declared_fields, model_info)
-        field_names.append("teams")
+    def get_field_names(self, declared_fields, info):
+        """Add reverse M2M for teams to the fields for this serializer."""
+        field_names = list(super().get_default_field_names(declared_fields, info))
+        self.extend_field_names(field_names, "teams")
         return field_names
```
glennmatthews: Updated this to follow the same pattern we're already using in other serializers.

nautobot/ipam/api/views.py
```diff
@@ -356,7 +356,7 @@ class PrefixLocationAssignmentViewSet(NautobotModelViewSet):
 
 
 class IPAddressViewSet(NautobotModelViewSet):
-    queryset = IPAddress.objects.all()
+    queryset = IPAddress.objects.select_related("parent__namespace")
```
glennmatthews, line 359: We don't auto-optimize nested lookups yet, and we weren't correctly selecting this one before. Another separate query I noticed testing this endpoint.

PR #6584, opened 2024-11-28
Move rack unit validation to save()

### LINKED ISSUE #6585: Able to create devices with conflicting Rack U designators.
### Environment
* Nautobot version (Docker tag too if applicable): v2.2.0+
* Python version: v3.{8,11}
* Middleware(s):NA

### Issue description:
Improper rendering of rack after a device with a conflicting U# is created.

### Steps to Reproduce
1. Create a 4U device at U2
2. Notice Rack is rendered successfully.
3. Recreate a second 4U device at U3 (conflicting, but is permitted to happen)
4. Notice Rack is now _not_ rendered successfully/fully.

### Expected Behavior
I would have expected either a notice that there is a conflicting rack unit and not allowed to create, or there to be some side-by-side rendering.

### Noticed Behavior:
Nautobot allows me to create 2 devices with conflicting rack U designations, these then cause the rack to not be rendered correctly. After this the render of the racks fails, and no front image is shown - the rear image is shown - but does not have the details of the devices - only the marking that the slots are taken.

icovada:
# Closes #6585 
# What's Changed

Move rack unit conflict checks to Device.save() so they run even when calling Device.objects.create()

As of today, these checks are only run when interacting via GUI or API
# TODO

- [] Unit, Integration Tests

### DIFF
--- a/changes/6585.fixed
+++ b/changes/6585.fixed
@@ -1,0 +1,1 @@
+Added checks to prevent being able to put devices in the same rack unit
--- a/nautobot/dcim/models/devices.py
+++ b/nautobot/dcim/models/devices.py
@@ -725,22 +725,6 @@ class Device(PrimaryModel, ConfigContextModel):
                         }
                     )
 
-                # Validate rack space
-                rack_face = self.face if not self.device_type.is_full_depth else None
-                exclude_list = [self.pk] if self.present_in_database else []
-                available_units = self.rack.get_available_units(
-                    u_height=self.device_type.u_height,
-                    rack_face=rack_face,
-                    exclude=exclude_list,
-                )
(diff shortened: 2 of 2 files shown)


icovada:
I'm against duplicating code, so we might just suggest using ´validated_save()´ as best practice

timizuoebideri1:
> I'm against duplicating code, so we might just suggest using ´validated_save()´ as best practice

Maybe add a check in save method, checking if `Validate rack space` has been validated, if not validate it?

glennmatthews:
@icovada Do you still want to move forward with this PR?

glennmatthews:
Closing this for now, can reopen later if desired.

PR #6595, opened 2024-12-03
Fixed `table_config` field not showing up correctly in the saved view modal

### LINKED ISSUE #6466: Table_config always NULL in New Saved View
Nautobot version: 2.3.10 Ubuntu
Django version: 4.2.16
Configuration file: /opt/nautobot/nautobot_config.py
* Database platform, version: mysql
* Middleware(s):

Save view is not saving the columns i set in table view configure

### Steps to Reproduce
1. Press configure on any table view
2. Change columns in view
3. Press save as view
4. See that table_config is null

### Expected Behavior
Table_config is filled and save to a view

### Observed Behavior
table_view is allways NULL

HanlinMiao:
# Closes #6466 
# What's Changed

Fixed `saved_view_modal` template tag helper method and `get_table_class_string_from_view_name` helper method to introspect the view class to obtain the table/table_class value instead of making a best guess.

# Screenshots

PrefixListView:
Screenshot 2024-12-03 at 3 01 40 PM

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/core/templatetags/helpers.py
+++ b/nautobot/core/templatetags/helpers.py
@@ -795,7 +795,13 @@ def saved_view_modal(
         "clear_view",
     ]
 
-    table_name = lookup.get_table_for_model(model).__name__
+    view_class = lookup.get_view_for_model(model, "List")
+    table_name = None
+    if hasattr(view_class, "table"):
+        table_name = view_class.table.__name__
+    if hasattr(view_class, "table_class"):
+        table_name = view_class.table_class.__name__
+
     for param in non_filter_params:
         if param == "saved_view":
             current_saved_view_pk = filters_applied.pop(param, None)
--- a/nautobot/core/utils/lookup.py
+++ b/nautobot/core/utils/lookup.py
@@ -8,7 +8,7 @@ from django.conf import settings
 from django.contrib.auth.models import Group
 from django.contrib.contenttypes.models import ContentType
 from django.db.models import Model
-from django.urls import get_resolver, URLPattern, URLResolver
+from django.urls import get_resolver, resolve, reverse, URLPattern, URLResolver
 from django.utils.module_loading import import_string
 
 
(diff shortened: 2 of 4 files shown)


HanlinMiao (commented):
Probably need a unit test to test `get_table_class_string_from_view_name`.

nautobot/core/templatetags/helpers.py
```diff
@@ -795,7 +795,13 @@ def saved_view_modal(
         "clear_view",
     ]
 
-    table_name = lookup.get_table_for_model(model).__name__
+    view_class = lookup.get_view_for_model(model, "List")
```
glennmatthews, line 798: Just confirming - this works for both legacy views and NautobotUIViewSet views?

nautobot/core/utils/lookup.py
```diff
@@ -276,11 +276,16 @@ def get_table_class_string_from_view_name(view_name):
...
-    if model:
-        table_class = get_table_for_model(model)
-        if table_class:
-            return table_class.__name__
+    app_label, model_name = view_name.split(":")
+    model_name = model_name.split("_")[0]
+    model_name = f"{app_label}.{model_name}"
+    # model name should be in the form of "dcim.location".
+    model = get_model_from_name(model_name)
+    view_class = get_view_for_model(model, "List")
```
glennmatthews, line 284: This feels overly convoluted to me. Given we have `view_name`, can't we resolve that directly to the `view_class` without needing to go by way of the `model`?

nautobot/core/templatetags/helpers.py
HanlinMiao: Yeah, get_view_for_model covers both. In the case of NautobotUIViewSet, the "List" parameter is unused.

nautobot/core/utils/lookup.py
```diff
@@ -276,12 +276,9 @@ def get_table_class_string_from_view_name(view_name):
...
     """
-    app_label, model_name = view_name.split(":")
-    model_name = model_name.split("_")[0]
-    model_name = f"{app_label}.{model_name}"
-    # model name should be in the form of "dcim.location".
-    model = get_model_from_name(model_name)
-    view_class = get_view_for_model(model, "List")
+    url = reverse(view_name)
+    match = resolve(url)
+    view_class = match.func.view_class
```
HanlinMiao, line 281: After adding some unit tests, this pattern does not seem to work for UIViewSets based views.

glennmatthews: In other code we've used this construct:

```
                    # For UIViewSet, view_func.cls gets what we need; for an ObjectListView, view_func.view_class is it.
                    view_or_viewset = getattr(view_func, "cls", getattr(view_func, "view_class", None))
````

glennmatthews (approved):

PR #6616, opened 2024-12-06
Added logic to set the parent in the clean() method of the Prefix model, ensuring correct assignment during validation.

### LINKED ISSUE #6548: `.parent` is not accessible for a Prefix object during custom validation, but `.ancestors()` is
### Environment
* Nautobot version (Docker tag too if applicable): 2.3.11
* Python version: 3.11
* Middleware(s): N/A

### Steps to Reproduce
1. Add a custom validator to a Prefix and just raise a validation error with `f"Parent is {self.context['object'].parent} and Ancestors is {self.context['object'].ancestors()} and Descendants is {self.context['object'].descendants()}"`
2. Verify that `.parent` is always `None`
3. Verify that `.ancestors()` and `.descendants()` work as expected

### Expected Behavior
I expected `.parent` to resolve to the closest ancestor. This works if we build a custom validator for `IPAddress`, it will return the parent prefix correctly

### Observed Behavior
`.parent` is always `None`, while `.ancestors()` and `.descendants()` work as expected

image

timizuoebideri1:
# Closes #6548 
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests


nautobot/ipam/models.py
```diff
@@ -612,18 +612,24 @@ def delete(self, *args, **kwargs):
...
             # This also has the subtle side effect of calling self._deconstruct_prefix(),
             # which will (re)set the broadcast and ip_version values of this instance to their correct values.
             self.prefix = self.prefix.cidr
 
-        # Determine if a parent exists and set it to the closest ancestor by `prefix_length`.
-        supernets = self.supernets()
-        if supernets:
-            parent = max(supernets, key=operator.attrgetter("prefix_length"))
-            self.parent = parent
+        self.parent = self.get_parent()
```
HanlinMiao, line 632: Do we need this logic here in save() after we already have it in clean()?

nautobot/ipam/models.py
```diff
@@ -612,18 +612,24 @@ def delete(self, *args, **kwargs):
...
             return super().delete(*args, **kwargs)
 
+    def get_parent(self):
+        # Determine if a parent exists and set it to the closest ancestor by `prefix_length`.
+        if supernets := self.supernets():
+            return max(supernets, key=operator.attrgetter("prefix_length"))
+        return None
+
+    def clean(self):
+        self.parent = self.get_parent()
```
glennmatthews, line 622: Can we possibly add some caching and optimization so that (on an update to an existing Prefix) we don't need to recalculate the parent if the network/prefix_length/namespace hasn't changed?

Same idea on the `reparent_subnets` and `reparent_ips` calls in `save()`.

nautobot/ipam/models.py
timizuoebideri1: Yes, we still need it, incase user uses `.save` instead of `.validated_save`

nautobot/ipam/models.py
timizuoebideri1: How do we add caching for `reparent_subnets` and `reparent_ips`

timizuoebideri1: Would this be the correct way to add caching to `get_parent`?

```python
def get_parent(self):
        # Determine if a parent exists and set it to the closest ancestor by `prefix_length`.
        cache_key = f"nautobot.ipam.prefix.parent.{self.network}-{self.prefix_length}-{self.namespace}"
        if cahched_parent := cache.get(cache_key):
            return cahched_parent

        if supernets := self.supernets():
            parent = max(supernets, key=operator.attrgetter("prefix_length"))
            cache.set(cache_key, parent)
            return parent
        return None

```

glennmatthews: Sorry, I was a bit vague. I wasn't thinking full-blown Redis caching, more just caching of initial values on the instance itself as a part of `__init__`.

glennmatthews: Something like:

- at init time, if present_in_database, save self.network, self.prefix_length, self.namespace as private instance variables
- at clean/save time, if present_in_database, skip the reparenting calls if comparison shows that those attributes are unchanged from init.

nautobot/ipam/models.py
```diff
@@ -660,10 +680,17 @@ def save(self, *args, **kwargs):
...
-        # Determine the subnets and reparent them to this prefix.
-        self.reparent_subnets()
-        # Determine the child IPs and reparent them to this prefix.
-        self.reparent_ips()
+        # Only reparent subnets and ips if any of these fields has been updated.
+        if (
+            not self.present_in_database
+            or self._network is not self.network
+            or self._namespace is not self.namespace
+            or self._prefix_length is not self.prefix_length
```
glennmatthews, line 688:
```suggestion
            or self._network != self.network
            or self._namespace != self.namespace
            or self._prefix_length != self.prefix_length
```

nautobot/ipam/models.py
```diff
@@ -612,18 +618,32 @@ def delete(self, *args, **kwargs):
...
+            self._parent = parent
+            return parent
+        return None
+
+    def clean(self):
+        super().clean()
+
+        if self.prefix:
+            # self.parent depends on self.prefix having a value
+            self.parent = self.get_parent()
```
glennmatthews, line 637: Is the order here correct/desirable or do we need to set self.parent before calling super().clean()?

nautobot/ipam/models.py
```diff
@@ -660,10 +680,17 @@ def save(self, *args, **kwargs):
             if self._location is not None:
                 self.location = self._location
 
-        # Determine the subnets and reparent them to this prefix.
-        self.reparent_subnets()
-        # Determine the child IPs and reparent them to this prefix.
-        self.reparent_ips()
+        # Only reparent subnets and ips if any of these fields has been updated.
```
glennmatthews, line 683: Can we add some unit testing around this? I know it's feature creep (sorry!) but I think it's important to test.

glennmatthews (commented):
Caching logic looks good

nautobot/ipam/models.py
```diff
@@ -531,8 +531,19 @@ def __init__(self, *args, **kwargs):
...
         super().__init__(*args, **kwargs)
+
         self._deconstruct_prefix(prefix)
 
+        self._parent = None
+        self._network = None
+        self._prefix_length = None
+        self._namespace = None
+
+        if self.pk:
```
timizuoebideri1, line 542: using `self.pk` at the `__init__` level instead of `self.present_in_database` because currently thats the only way that works after testing several times.
`self.present_in_database` always returns `False` at `__init__`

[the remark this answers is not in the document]
timizuoebideri1: https://github.com/nautobot/nautobot/pull/6616#discussion_r1880232542

using self.pk at the __init__ level instead of self.present_in_database because currently thats the only way that works after testing several times.
self.present_in_database always returns False at __init__

glennmatthews: That's weird... it *should* work when initing from an existing DB record. I'm concerned that checking `self.pk` may lead to false positives, especially after #6340.

glennmatthews: present_in_database evaluates to false even after calling super().__init__()?

timizuoebideri1: Yes

glennmatthews: Ok, looks like after some reading we might need to instead overload `from_db()` (https://docs.djangoproject.com/en/4.2/ref/models/instances/#customizing-model-loading) and set the cached attributes there instead of in `__init__()`?

nautobot/ipam/tests/test_models.py
```diff
@@ -413,6 +414,39 @@ def setUp(self):
...
+        """Assert reparent_subnets and reparent_ips are only called if there is an update to either of network, namespace or prefix_length"""
+        prefix_ip = "101.102.0.0/28"
+        with self.subTest("Assert reparent_subnets"):
+            with patch.object(Prefix, "reparent_subnets", return_value=None) as mock_reparent_subnets:
+                Prefix.objects.create(prefix=prefix_ip, status=self.status, namespace=self.namespace)
+                mock_reparent_subnets.assert_called_once()
+
+            with patch.object(Prefix, "reparent_subnets", return_value=None) as mock_reparent_subnets:
+                prefix = Prefix.objects.get(prefix=prefix_ip)
+                prefix.description = "Sample Description"
```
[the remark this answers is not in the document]
timizuoebideri1: Yup it exists; `nautobot.ipam.tests.test_models.TestPrefix.test_reparenting`

nautobot/ipam/models.py
```diff
@@ -655,15 +692,26 @@ def save(self, *args, **kwargs):
...
         with transaction.atomic():
             super().save(*args, **kwargs)
             if self._location is not None:
                 self.location = self._location
 
-        # Determine the subnets and reparent them to this prefix.
-        self.reparent_subnets()
-        # Determine the child IPs and reparent them to this prefix.
-        self.reparent_ips()
+        # Only reparent subnets and ips if any of these fields has been updated.
```
glennmatthews, line 704: Let's add a `.changed` fragment for this enhancement.

glennmatthews (approved):

Issue #6614, opened 2024-12-05
Prefix IP address table shows counts for interfaces and devices rather than actual values.


nrnvgh:
### As ...

Ozzie - Operator

### I want ...

A way to see at a glance the device and interface of the IPs in a prefix. This request is also designed to bring the behavior of this table in line with the main Prefixes table, as updated in #6231 (see also: #6613).

### So that ...

The IP address table displays a more useful view showing which IPs are associated with which device+interface.

### I know this is done when...

The prefix IP address table (e.g. https://demo.nautobot.com/ipam/prefixes/d265dacf-ddc5-4287-ba2a-23e923f6e750/ip-addresses/) behaves like the previous behavior unless a IP is assigned to more than one device/interface. Basically, if an IP is assigned to a single interface, the device name and interface name should be displayed. If the IP is assigned to more than one interface on the same device, the device name should be displayed and a count of interfaces can be displayed. If an IP is assigned to interfaces on multiple devices, a count should be displayed for both fields.

### Optional - Feature groups this request pertains to.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [X] IPAM
- [ ] Misc (including Data Sources)
- [ ] Organization
- [ ] Apps (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation
- [ ] Other (not directly a platform feature)

glennmatthews:
The requested logic is going to take some time to implement. In an upcoming PR (mainly focused on addressing #6605) I'll at least be able to make the "Interfaces" and "VM Interfaces" columns render a single interface name/label if present, similar to what's been done elsewhere recently in the UI. What this PR *won't* do (at least yet) is render the name of a single device in the Devices column (this is trickier to do efficiently because it's a nested database lookup, e.g. IPAddress.interfaces[0].device.name instead of just IPAddress.interfaces[0].name)

nrnvgh:
It's been a minute, so I wanted to bring this up again as it's still a thing in 3.2:

https://next.demo.nautobot.com/ipam/prefixes/0617a2ff-cfac-58d4-bc2a-433a83c2f998/ip-addresses/

itdependsnetworks:
@joewesch can you take a look?

PR #6626, opened 2024-12-06
Improve performance and usability of IPAddress list view

### LINKED ISSUE #6605: Improve Load Times on IPAddress List View
### As ...

Nelly - Network Engineer

### I want ...

I want the IPAM - IP Address list view to load more efficiently

### So that ...

the view is more usable by network engineers who need to frequently access the IP Addresses table view for planning purposes.

- Load times are reduced from 20-30 seconds to 5-6 seconds

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [X] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

_No response_

### External Dependencies

_No response_

### LINKED ISSUE #6613: Child prefix table shows location counts rather than location name.
### As ...

Ozzie - Operator

### I want ...

A way to see at a glance the location of the child prefixes of a given prefix. This request is also designed to bring the behavior of this table in line with the main Prefixes table, as updated in #6231 (see also: #6614).

### So that ...

The child prefix table displays a more useful view showing which child prefixes are assigned to which location.

The child prefix table (e.g. https://demo.nautobot.com/ipam/prefixes/d265dacf-ddc5-4287-ba2a-23e923f6e750/prefixes/) behaves like the previous behavior unless a prefix is assigned to more than one location.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [X] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation

### Database Changes

_No response_

### External Dependencies

_No response_

glennmatthews:
- Closes #6605 
- Closes #6613 
- Partial progress towards #6614

# What's Changed

- Added `ip_addresses` and `has_ip_addresses` filter support to Device, Interface, and VirtualMachine FilterSets.
- Enhanced `LinkedCountColumn` to add a `distinct=True` option.

- The above enhancements make it possible to replace the bespoke annotations of the queryset in `IPAddressListView` and generic text columns in `IPAddressTable` with proper `LinkedCountColumn` instances and automatic queryset optimization. 
- This in turn makes it so that hiding the "Interfaces", "VM Interfaces", "Devices", "Virtual Machines", and/or "Assigned Count" columns in this view (4 of these 5 are hidden by default) correspondingly reduces the query complexity and therefore rendering time of the view. (#6605)

- Enhanced `BaseTable` to add a `data_transform_callback` option.
- The above makes it possible to refactor related-object tables (in PrefixPrefixesView, PrefixIPAddressesView, VLANGroupView) that transform their queryset data into a list (adding "available" pseudo-records in the process) to still take advantage of LinkedCountColumn and the automatic queryset optimizations associated.

# Screenshots

Locations column in PrefixPrefixesView (fixing #6613):
image

Interfaces and Devices columns in PrefixIPAddressesView (progress toward #6614):
image

IPAddressListView with all columns displayed:
image

Same data set with count columns hidden:
image

Same data set after hiding the Dynamic Groups and Tags columns as well:
image

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/ipam/views.py
+++ b/nautobot/ipam/views.py
@@ -2,7 +2,9 @@ import logging
 
 from django.conf import settings
 from django.contrib import messages
+from django.contrib.auth.models import AnonymousUser
 from django.core.cache import cache
+from django.core.exceptions import ObjectDoesNotExist
 from django.db import models, transaction
 from django.db.models import Prefetch, ProtectedError, Q
 from django.forms.models import model_to_dict
@@ -23,7 +25,7 @@ from nautobot.core.views import generic, mixins as view_mixins
 from nautobot.core.views.paginator import EnhancedPaginator, get_paginate_count
 from nautobot.core.views.utils import handle_protectederror
 from nautobot.dcim.models import Device, Interface, Location
-from nautobot.extras.models import Role, Status, Tag
+from nautobot.extras.models import Role, SavedView, Status, Tag
 from nautobot.ipam import choices, constants
 from nautobot.ipam.api import serializers
 from nautobot.tenancy.models import Tenant
@@ -101,22 +103,11 @@ class NamespaceIPAddressesView(generic.ObjectView):
 
     def get_extra_context(self, request, instance):
         # Find all IPAddresses belonging to this Namespace
-        ip_addresses = (
-            instance.ip_addresses.restrict(request.user, "view")
-            .select_related("role", "status", "tenant")
-            .annotate(
-                interface_count=count_related(Interface, "ip_addresses"),
-                interface_parent_count=count_related(Device, "interfaces__ip_addresses", distinct=True),
-                vm_interface_count=count_related(VMInterface, "ip_addresses"),
-                vm_interface_parent_count=count_related(VirtualMachine, "interfaces__ip_addresses", distinct=True),
-            )
-        )
+        ip_addresses = instance.ip_addresses.restrict(request.user, "view").select_related("role", "status", "tenant")
 
-        ip_address_table = tables.IPAddressTable(ip_addresses)
+        ip_address_table = tables.IPAddressTable(ip_addresses, exclude=["namespace"])
         if request.user.has_perm("ipam.change_ipaddress") or request.user.has_perm("ipam.delete_ipaddress"):
             ip_address_table.columns.show("pk")
-
-        ip_address_table.exclude = ("namespace",)
 
         paginate = {
             "paginator_class": EnhancedPaginator,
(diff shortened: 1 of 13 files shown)


nautobot/ipam/views.py
```diff
@@ -498,19 +483,22 @@ class PrefixPrefixesView(generic.ObjectView):
...
+        child_prefixes = instance.descendants().restrict(request.user, "view")
 
         # Add available prefixes to the table if requested
         if child_prefixes and request.GET.get("show_available", "true") == "true":
-            child_prefixes = add_available_prefixes(instance.prefix, child_prefixes)
 
-        prefix_table = tables.PrefixDetailTable(child_prefixes, hide_hierarchy_ui=True)
-        prefix_table.exclude = ("namespace",)
+            def data_transform_callback(prefixes):
+                return add_available_prefixes(instance.prefix, prefixes)
```
mateusz-lachowski-codilime, line 492: Just wondering, what looks better: current function def or: `data_transform_callback = lambda prefixes: add_available_prefixes(instance.prefix, prefixes)`

glennmatthews: I actually went with the lambda initially but pylint yelled at me. :-)

nautobot/ipam/views.py
```diff
@@ -550,21 +538,21 @@ def get_extra_context(self, request, instance):
...
-                instance.prefix, ipaddresses, instance.type == choices.PrefixTypeChoices.TYPE_POOL
-            )
 
-        ip_table = tables.IPAddressTable(ipaddresses)
+            def data_transform_callback(ip_addresses):
+                return add_available_ipaddresses(
+                    instance.prefix, ip_addresses, instance.type == choices.PrefixTypeChoices.TYPE_POOL
+                )
+        else:
+            data_transform_callback = None
```
mateusz-lachowski-codilime, line 551: Or since you already refactoring this "pattern" maybe go even further and prepare more robust functions like:
```
def add_available_ipaddresses_callback(instance, show_available):
    ...

data_transform_callback = add_available_ipaddresses_callback(instance, request.GET.get("show_available", "true")
```
I can see multiple benefits: we get rid off weird `def` from the view, remove this `if request.`, move callback logic into one place, allow simpler unittest for each callback function. And for the `PrefixPrefixes` also remove additional queryset for `child_prefixes` from the get extra context.

glennmatthews:
Exploring one more optimization - as the default `ordering` on Interface is a fairly expensive one, and we only care about the prefetched object instance if it's the only such instance in the queryset, adding a `.order_by()` in this case. Will update if it proves out.

mateusz-lachowski-codilime (approved):

Issue #1150, opened 2021-12-13
Nautobot 12 factor adherence


nniehoff:
### Environment
* Python version: NA
* Nautobot version: 1.2.0b1

With the implementation of #1079 we have removed the ability to set several settings from environment variables when those settings were migrated to the DB.  I like the DB configuration and we need it but it seems some functionality was lost by removing the default `os.getenv` calls from the default `nautobot_config.py.j2`.  This seems to drift a bit from the [12-factor app ](https://12factor.net/config) philosophy.  This probably isn't really a huge deal (especially with the set of variables that were moved to the DB) but it does impact the docker deployments out there as we can no longer specify env vars for these settings.  

I'd like to come up with a plan for all settings/config variables, not just this small subset, so we can be consistent and if possible meet the 12-factor guidance.  

Today the docker build uses the `nautobot_config.py.j2` template to build its default config and appends some logging configuration to it.  We can come up with some logic for the docker entrypoint to pull env vars and inject them into a custom nautobot_config, build a completely custom nautobot_config.py for the containers, or we could move the env var parsing to the defaults in settings.py which would allow the DB to override them.  

I'd also like to come up with something that would allow for the configuration of other settings vars like LDAP or SSO for the docker container.  This may require a hybrid solution.  Looking for some ideas/feedback on how we can maintain this long term.

glennmatthews:
Just a quick note that it's still quite possible to manage the settings as environment variables, just not with a completely vanilla `nautobot_config.py` freshly generated from the built-in template since it no longer includes those by default. Probably that's what you're alluding to when you mention a custom nautobot_config for the container build, but I just wanted to make that explicit. Thanks! :-)

glennmatthews:
Not sure if #1088 is also related here.

nniehoff:
> Just a quick note that it's still quite possible to manage the settings as environment variables, just not with a completely vanilla `nautobot_config.py` freshly generated from the built-in template since it no longer includes those by default. Probably that's what you're alluding to when you mention a custom nautobot_config for the container build, but I just wanted to make that explicit. Thanks! :-)

[...]

PR #6627, opened 2024-12-07
WIP: yaml loading functionality for app config

cardoe:
# What's Changed

Allow apps (plugins) to be enabled and configured via a yaml file to help further the goal of settings not being part of the code. ref #1150

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/core/settings.py
+++ b/nautobot/core/settings.py
@@ -11,7 +11,7 @@ from django.utils.safestring import mark_safe
 
 from nautobot import __version__
 from nautobot.core.constants import CONFIG_SETTING_SEPARATOR as _CONFIG_SETTING_SEPARATOR
-from nautobot.core.settings_funcs import ConstanceConfigItem, is_truthy, parse_redis_connection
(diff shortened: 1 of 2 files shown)


cardoe:
An example config could be something like:

```yaml
nautobot_ssot:
    hide_example_jobs: true
nautobot_plugin_nornir:
    username: !env_var NAUTOBOT_NORNIR_USERNAME
    password: !env_var NAUTOBOT_NORNIR_PASSWORD
```

Which would enable `nautobot_ssot` and `nautobot_plugin_nornir` and set that configuration if that file was saved and the path supplied in `NAUTOBOT_APP_CONFIG`.

This is very much a work in progress change and I'm looking for feedback. But it would address @gsnider2195 comment on #1150.

lampwins:
@cardoe I know this spawned off of #1150 but let's open a specific issue to this idea to discuss it further. We might have some other ideas on how we can achieve the same end goal.

cardoe:
@lampwins I've created nautobot/helm-charts#558 with a specific ask around that deployment mechanism.

PR #6672, opened 2024-12-17
Requesting media files requires authentication

timizuoebideri1:
# Closes #DNE
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests


glennmatthews (commented):
1) Please target to `develop` rather than `next`
2) Seems like there probably should be some update to `nautobot.core.tests.integration.test_view_authentication` corresponding to this change - there must be a gap in the current test given that this change didn't make any part of it fail?

nautobot/core/views/__init__.py
```diff
@@ -132,6 +134,15 @@ def get(self, request, *args, **kwargs):
         return self.render_to_response(context)
 
 
+class MediaView(GenericView):
+    """
+    Wrap Django's serve() view to enforce LOGIN_REQUIRED for static media.
```
glennmatthews, line 139: This may be problematic actually, as the custom-branding files (`settings.BRANDING_FILEPATHS`) currently assume files stored in `settings.MEDIA_ROOT`, but those specific files **must** be accessible unauthenticated, for example to provide custom branding on the login page.

In the short term, we might have to carve out in this view an explicit exception for those files/paths in particular? In the longer term, I wonder if hosting them in `MEDIA_ROOT` was an incorrect design decision, and maybe (as a breaking change) we might move them to being expected in `STATIC_ROOT` instead?

glennmatthews (changes requested):

nautobot/core/views/__init__.py
```diff
@@ -132,6 +133,34 @@ def get(self, request, *args, **kwargs):
...
+        with contextlib.suppress(FileNotFoundError):
+            for branding_filepath in settings.BRANDING_FILEPATHS.values():
+                if not branding_filepath:
+                    continue
+                branding_filepath = os.path.join(settings.MEDIA_ROOT, branding_filepath)
+                requested_path = os.path.join(settings.MEDIA_ROOT, path)
+                same_path = os.path.samefile(branding_filepath, requested_path)
+                if same_path:
+                    can_view = True
+                    break
```
glennmatthews, line 157: Would be more efficient to skip this block if `can_view` is already `True`.

glennmatthews (approved):

bryanculver (commented):
I'm fine with these changes as-is and shouldn't break anything for S3-backed assets as well but I am still verifying if it will be an incomplete fix in those cases (confirming if assets are served from s3 directly or if they all go through this view).

PR #6675, opened 2024-12-17
Fixed untagged VLAN dropdown options mismatch in InterfaceEditForm and in InterfaceBulkEditForm.

### LINKED ISSUE #6470: Untagged VLAN drop-down widget for Bulk Interface Edit populates different than available_on_device from Interface Edit
### Environment
* Nautobot version (Docker tag too if applicable): v2.3.10, v2.3.11b
* Python version: 3.11
* Middleware(s): N/A

### Issue
I am seeing unexpected differences between populated results for the `Untagged VLAN` drop-down widget between `Bulk Interface Edit` vs `Interface Edit `. The drop-down is populating on Interface Edit, but on Bulk Interface Edit it is empty. Even if I am only doing a "bulk edit" of a single interface, which I would expect to populate the same data as `Interface Edit`?

It appears to be related to inhereted VLAN, or is my suspicion (see digging notes)

### Steps to Reproduce
1. 
2.
3.

### Expected Behavior
The same options should populate under `Untagged VLAN` between the 2-forms

### Observed Behavior
The drop-down populates differently for the same interface between the 2-forms

### Digging
The drop-downs populate based on response from 2-different queries to /api/ipam/vlans endpoint
 - Bulk Interface Edit HTTP GET -> locations=null&locations={uuid}
 - Interface Edit HTTP GET -> available_on_device={uuid}

Looking in to available_on_device, led me to the following candidate I was considering:

> I suspect the difference of not using available_on_device on the Bulk Interface Edit API query is going to be the issue
>
> I found public issue (#5050) and patch (#6091) which indicated to me that available_on_device was created to address a possible scenario
>
> PR Notes: updated the available_on_device/get_for_device filter in the VLANFilterSet to include not just the device(s)' locations, but also those location(s)' ancestors

#### showing the widgets in question, but not showing data to demo the issue
image
image

HanlinMiao:
# Closes #6470 
# What's Changed

In the case of InterfaceBulkEdit:
1. In the case of a single location, use the `available_on_device` query param to limit untagged VLAN choices to those available on the devices in that location and in the ancestors of the location.
2. In the case of multiple locations, we use the old query params with locations=null and locations=[list of location pks].

# Screenshots

N/A
# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/forms.py
+++ b/nautobot/dcim/forms.py
@@ -3239,6 +3239,8 @@ class InterfaceBulkEditForm(
                 self.fields["tagged_vlans"].widget.add_query_param("locations", location.pk)
             else:
                 self.fields["untagged_vlan"].widget.add_query_param("locations", "null")
+                if locations.count() > 1:
+                    self.fields["untagged_vlans"].widget.add_query_param("locations", list(locations.values_list("pk", flat=True)))
 
         # Restrict parent/bridge/LAG interface assignment by device (or VC master)
         if device_count == 1:
--- a/nautobot/dcim/forms.py
+++ b/nautobot/dcim/forms.py
@@ -3183,6 +3183,9 @@ class InterfaceBulkEditForm(
     untagged_vlan = DynamicModelChoiceField(
         queryset=VLAN.objects.all(),
         required=False,
+        query_params={
+            "locations": "null",
+        },
     )
     tagged_vlans = DynamicModelMultipleChoiceField(
         queryset=VLAN.objects.all(),
@@ -3234,15 +3237,8 @@ class InterfaceBulkEditForm(
                 location = locations.first()
                 # In the case of a single location, use the available_on_device query param to limit untagged VLAN choices
                 # to those available on the devices in that location and in the ancestors of the location.
-                device_pks = [device.pk for device in devices]
-                self.fields["untagged_vlan"].widget.add_query_param("available_on_device", device_pks)
+                self.fields["untagged_vlan"].widget.add_query_param("available_on_device", device.pk)
                 self.fields["tagged_vlans"].widget.add_query_param("locations", location.pk)
-            else:
-                self.fields["untagged_vlan"].widget.add_query_param("locations", "null")
-                if locations.count() > 1:
-                    self.fields["untagged_vlans"].widget.add_query_param(
-                        "locations", list(locations.values_list("pk", flat=True))
-                    )
 
         # Restrict parent/bridge/LAG interface assignment by device (or VC master)
         if device_count == 1:
--- a/nautobot/dcim/forms.py
+++ b/nautobot/dcim/forms.py
@@ -3240,7 +3240,9 @@ class InterfaceBulkEditForm(
             else:
                 self.fields["untagged_vlan"].widget.add_query_param("locations", "null")
                 if locations.count() > 1:
-                    self.fields["untagged_vlans"].widget.add_query_param("locations", list(locations.values_list("pk", flat=True)))
+                    self.fields["untagged_vlans"].widget.add_query_param(
+                        "locations", list(locations.values_list("pk", flat=True))
+                    )
 
         # Restrict parent/bridge/LAG interface assignment by device (or VC master)
         if device_count == 1:
--- a/nautobot/dcim/forms.py
+++ b/nautobot/dcim/forms.py
@@ -3183,9 +3183,6 @@ class InterfaceBulkEditForm(
     untagged_vlan = DynamicModelChoiceField(
         queryset=VLAN.objects.all(),
         required=False,
-        query_params={
-            "locations": "null",
-        },
     )
     tagged_vlans = DynamicModelMultipleChoiceField(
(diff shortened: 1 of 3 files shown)


nautobot/dcim/forms.py
```diff
@@ -3235,8 +3232,15 @@ def __init__(self, *args, **kwargs):
...
-                self.fields["untagged_vlan"].widget.add_query_param("locations", location.pk)
+                # In the case of a single location, use the available_on_device query param to limit untagged VLAN choices
+                # to those available on the devices in that location and in the ancestors of the location.
+                device_pks = [device.pk for device in devices]
+                self.fields["untagged_vlan"].widget.add_query_param("available_on_device", device_pks)
                 self.fields["tagged_vlans"].widget.add_query_param("locations", location.pk)
+            else:
+                self.fields["untagged_vlan"].widget.add_query_param("locations", "null")
+                if locations.count() > 1:
+                    self.fields["untagged_vlans"].widget.add_query_param("locations", list(locations.values_list("pk", flat=True)))
```
HanlinMiao, line 3243: Should I account for the ancestors of these locations as well?

nautobot/dcim/forms.py
```diff
@@ -3235,8 +3232,17 @@ def __init__(self, *args, **kwargs):
             # Limit VLAN choices by Location
             if locations.count() == 1:
                 location = locations.first()
-                self.fields["untagged_vlan"].widget.add_query_param("locations", location.pk)
+                # In the case of a single location, use the available_on_device query param to limit untagged VLAN choices
+                # to those available on the devices in that location and in the ancestors of the location.
+                device_pks = [device.pk for device in devices]
+                self.fields["untagged_vlan"].widget.add_query_param("available_on_device", device_pks)
```
glennmatthews, line 3238: Is this the right logical check to add this query parameter, or should it be in the `if device_count == 1` case below?

HanlinMiao: I believe so as I understand it from this comment https://github.com/nautobot/nautobot/issues/6470#issuecomment-2546380761

glennmatthews: On review, I think this is OK - `available_on_device` basically translates to `(locations__isnull=True | locations__in=[device.location, device.location.ancestors for device in filter])`, so if the selected devices share the same location then `available_on_device` will evaluate the same as if a single device were selected.

glennmatthews: As such, it might not even be necessary to pass `device_pks` to the filter in this specific case, it might actually result in a simpler/more efficient query if you just pass devices.first().pk instead? Worth testing.

nautobot/dcim/forms.py
```diff
@@ -3235,8 +3232,17 @@ def __init__(self, *args, **kwargs):
...
                 self.fields["tagged_vlans"].widget.add_query_param("locations", location.pk)
+            else:
+                self.fields["untagged_vlan"].widget.add_query_param("locations", "null")
+                if locations.count() > 1:
```
glennmatthews, line 3242: Isn't this implied by the above `else`? Or is this just specifically covering the case where `locations` might be empty?

nautobot/dcim/forms.py
glennmatthews: Probably should... does InterfaceEditForm do so now though?

nautobot/dcim/forms.py
HanlinMiao: I see cause we already checked that we only have one location. That makes sense

nautobot/dcim/forms.py
HanlinMiao: Yeah think about it more, this might not be necessary after all. I wrote it specifically for the case we might have multiple interfaces from multiple locations. But I think the location.count() can only be 0 or 1.

nautobot/dcim/forms.py
HanlinMiao: `available_on_device` is the filter that we use on InterfaceEditForm and that includes the ancestors. But I do not think this code block matters because I think the `locations.count()` can only be 0 or 1

nautobot/dcim/forms.py
```diff
@@ -3235,7 +3235,9 @@ def __init__(self, *args, **kwargs):
             # Limit VLAN choices by Location
             if locations.count() == 1:
                 location = locations.first()
-                self.fields["untagged_vlan"].widget.add_query_param("locations", location.pk)
+                # In the case of a single location, use the available_on_device query param to limit untagged VLAN choices
+                # to those available on the devices in that location and in the ancestors of the location.
+                self.fields["untagged_vlan"].widget.add_query_param("available_on_device", device.pk)
```
glennmatthews, line 3240: I'm confused now, don't we still need to remove `locations: null` from the base field definition (line 3187) for this to work?

nautobot/dcim/forms.py
glennmatthews: I'm confused then - if you're editing interfaces from N different devices, can't there be N locations?

PR #6683, opened 2024-12-19
6583 Fixed permissions constraints on available ips/prefixes/vlans.

### LINKED ISSUE #6583: Prefix permission constraints are bypassed when using the `available-prefixes` API
### Environment
* Nautobot version (Docker tag too if applicable): 2.3.12
* Python version: 3.11
* Database platform, version: PG16
* Middleware(s):

### Steps to Reproduce
1. Create a Prefix, e.g `10.10.0.0/24` with description `ThisDescriptionIsMandatory`.
2. Create a new user and set his permissions to view all prefixes
3. Create a custom add/edit/delete Prefix permissions for this user with the constraint `{"description": "ThisDescriptionIsMandatory"}`
4. As this new user, query the `available-prefixes` API endpoint on the newly created prefix `10.10.0.0/24` and request a new child prefix with any random description, e.g. `test` (e. John: need to verify that this exists on both GET and write versions of this endpoint)
5. Verify the Prefix is created successfully

### Expected Behavior

The new user should have gotten an object-level permission error, since the newly created prefix (child of `10.10.0.0/24`) does not respect the Prefix add/edit/delete permission constraint assigned to the user - i.e. the description of the newly created prefix is NOT `ThisDescriptionIsMandatory`.

### Observed Behavior

The new child prefix was created not respecting the permission constraints.

This same user is NOT allowed to create that same prefix in the GUI, but the `available-prefixes` endpoint allows it.

Besides, by browsing this new child prefix in the GUI, the user can't delete or edit it - which is expected since the description doesn't match the constraints. But if this is the case, the user shouldn't have been able to create it.

mateusz-lachowski-codilime:
# Closes #6583 
# What's Changed
- Object-level constrained permissions are now enforced on `available-ips/prefixes/vlans` API Endpoints
- Added handy `NautobotTestCaseMixin.add_permission` method for easier creating a constrained permissions.
- Changed `available-vlans` required permissions: now it's the same as for other `available-*` endpoint and require both "parent" permission and "object" permission: `view_vlangroup` and `view_vlan` - the same as for `available-ips`: `view_prefix` and `view_ipaddress`.
# Screenshots

# TODO
By the occasion, in some other places might be worth to use the `add_permission` method and remove some code from other unittests.

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/core/testing/mixins.py
+++ b/nautobot/core/testing/mixins.py
@@ -144,28 +144,21 @@ class NautobotTestCaseMixin:
     # Permissions management
     #
 
-    def add_permissions(self, *names):
+    def add_permissions(self, *names, **kwargs):
         """
         Assign a set of permissions to the test user. Accepts permission names in the form <app>.<action>_<model>.
+        Additional keyword arguments will be passed to the ObjectPermission constructor to allow creating more detailed permissions.
+
+        Examples:
+            >>> add_permissions("ipam.add_vlangroup", "ipam.view_vlangroup")
+            >>> add_permissions("ipam.add_vlangroup", "ipam.view_vlangroup", constraints={"pk": "uuid-1234"})
         """
         for name in names:
             ct, action = permissions.resolve_permission_ct(name)
-            obj_perm = users_models.ObjectPermission(name=name, actions=[action])
+            obj_perm = users_models.ObjectPermission(name=name, actions=[action], **kwargs)
             obj_perm.save()
             obj_perm.users.add(self.user)
             obj_perm.object_types.add(ct)
-
-    def add_permission(self, name, **kwargs):
-        """
-        Create and assign permissions to the test user. Accepts permission names in the form <app>.<action>_<model>.
-        Additional arguments will be passed to the ObjectPermission constructor to allow creating more detailed permissions.
-        """
-
-        ct, action = permissions.resolve_permission_ct(name)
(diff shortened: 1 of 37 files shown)


nautobot/core/testing/mixins.py
```diff
@@ -155,6 +155,18 @@ def add_permissions(self, *names):
             obj_perm.users.add(self.user)
             obj_perm.object_types.add(ct)
 
+    def add_permission(self, name, **kwargs):
```
glennmatthews, line 158: Maybe just add `**kwargs` support to `add_permissions()` above? This seems functionally equivalent for the case where there's only a single `name`.

nautobot/ipam/api/views.py
```diff
@@ -396,21 +398,20 @@ class VLANGroupViewSet(NautobotModelViewSet):
...
+    @staticmethod
+    def vlan_group_queryset():
+        return (
+            VLANGroup.objects.select_related("location")
+            .prefetch_related("tags")
+            .annotate(vlan_count=count_related(VLAN, "vlan_group"))
+        )
 
     class AvailableVLANPermissions(TokenPermissions):
         """As nautobot.core.api.authentication.TokenPermissions, but enforcing add_vlan permission."""
```
glennmatthews, line 410: Could you update this?

glennmatthews (approved):
Looks really solid testing wise!

HanlinMiao (approved):
Thank you for the tests!

PR #6684, opened 2024-12-19
Updates to theme and marketplace/installed-apps views

glennmatthews:
# What's Changed

- ~Re-color "info" buttons and labels as dark grey instead of the not-quite-the-same-as-"primary blue" they were previously~ _reverted based on feedback_

- Change sidebar "add" buttons to "primary blue" rather than "info blue"

image

- ~Re-color "default" labels as~ _Add new `label-transparent` class that renders as_ transparent with primary-blue outline and text

image

- Update "author" and "availability" text in marketplace data
- Unify code path for installed-apps UI and API extraction of information from AppConfig + marketplace data
- Various layout adjustments to Marketplace view

image

image

image

image

- Various layout adjustments to Installed-Apps list and detail views

image

image

image

image

- Add summary count of "success" results to JobResult table view:

image

# TODO

- [x] Unit, Integration Tests


nautobot/core/choices.py
```diff
@@ -182,14 +182,14 @@ class ButtonActionColorChoices(ChoiceSet):
...
-    ADD = "info"
+    ADD = "primary"
     CANCEL = "default"
     CLONE = "success"
     CONFIGURE = "default"
     CONNECT = "success"
     DEFAULT = "default"
     DELETE = "danger"
-    DISCONNECT = "info"
+    DISCONNECT = "danger"
```
glennmatthews, line 192: AFAIK we're not using this constant anywhere in core, but all of the actual "disconnect" buttons (Interface, ConsolePort, etc.) were already hard-coded to `btn-danger` so this is just making it consistent.

nautobot/core/choices.py
```diff
@@ -182,14 +182,14 @@ class ButtonActionColorChoices(ChoiceSet):
     Map standard button actions to Bootstrap color classes.
     """
 
-    ADD = "info"
+    ADD = "primary"
```
glennmatthews, line 185: Needed to make sidebar "add" buttons not turn gray. Also now consistent with `IMPORT = "primary"` below.

nautobot/core/templates/utilities/templatetags/tag.html
```diff
@@ -1,3 +1,3 @@
 {% load helpers %}
 
-{% if url_name %}<a href="{% url url_name %}?tags={{ tag.name }}">{% endif %}<span class="label label-default" style="color: {{ tag.color|fgcolor }}; background-color: #{{ tag.color }}">{{ tag }}</span>{% if url_name %}</a>{% endif %}
+{% if url_name %}<a href="{% url url_name %}?tags={{ tag.name }}">{% endif %}<span class="label" style="color: {{ tag.color|fgcolor }}; background-color: #{{ tag.color }}">{{ tag }}</span>{% if url_name %}</a>{% endif %}
```
glennmatthews, line 3: Needed because `label-default` now has a primary-blue border on it, and safe to remove as we were overriding all other `label-default` attributes anyway with the `style=` declaration.

nautobot/extras/plugins/views.py
```diff
@@ -132,8 +143,8 @@ def get(self, request, app=None, plugin=None):
             request,
             "extras/plugin_detail.html",
             {
+                "app_data": extract_app_data(app_config, load_marketplace_data()),
                 "object": app_config,
```
HanlinMiao, line 147: Would it make sense to include min_version and max_version in "app_data" as well? So we only need to pass in one context variable and we can leave "object" out?

HanlinMiao (commented):
The .css and .css.map files are a little hard to review, but I assume it is auto-generated?

nautobot/extras/plugins/views.py
glennmatthews: `object` is still needed because it contains all of the app "features" for the right panel of the detail view. Since that info is *only* needed for the detail view, I don't think it makes sense to populate all of it into `app_data`.

glennmatthews:
> The .css and .css.map files are a little hard to review, but I assume it is auto-generated?

Yes, those are generated from the corresponding `nautobot-bootstrap` PR, no manual changes to them.

nautobot/extras/plugins/views.py
```diff
@@ -29,12 +29,53 @@ def load_marketplace_data():
...
+    try:
+        reverse(app_config.config_view_name)
+        config_url = app_config.config_view_name
+    except NoReverseMatch:
+        config_url = None
+    try:
+        reverse(app_config.docs_view_name)
+        docs_url = app_config.docs_view_name
+    except NoReverseMatch:
+        docs_url = None
```
mateusz-lachowski-codilime, line 61: Hmm, what about something like this? I'm not sure if there will be more than 3 urls in future but three try/except one after one is on the edge between move to loop or leave as it is :D 
```
def _get_app_urls(app_config):
    app_urls = {"home_url": app_config.home_view_name, "config_url": app_config.config_view_name, "docs_url": app_config.docs_view_name}
    
    for key, url in app_urls.items():
        try:
            reverse(url)
        except NoReverseMatch:
            app_urls[key] = None
     return app_urls

def extract_app_data(app_config, marketplace_data):
    ...
    {
        ...
        **_get_app_urls(app_config)
    }
    
```

mateusz-lachowski-codilime (approved):

nautobot/extras/plugins/views.py
glennmatthews: I think we can refactor this in future if we need to add additional URLs at that point. Good thought though.

PR #6659, opened 2024-12-13
Nautobot-version-aware docker compose environment and invoke tasks

glennmatthews:
# What's Changed

- Add capability in `tasks.py` to detect the current ~Git base branch (from `main`, `develop`, `next`, `ltm-1.6`)~ _Nautobot major/minor version number_ and update the project name accordingly
- Enhance docker-compose-using tasks to pass the ~Git base branch~ _Nautobot major/minor version number_ as an environment variable to docker compose
- Enhance docker compose YAML files to use different image labels depending on the provided ~Git base branch~ `$NAUTOBOT_VER`
- Add `invoke branch` task for switching between and/or creating new branches, ensuring that it automatically stops the current Docker containers before doing so.

The net result here **should** be that once we have this enhancement in all of the major base branches, switching between release trains (or child branches under them) should be able to automatically use a different Docker image and different db volume depending on the relevant major.minor version. This should eliminate the time-consuming need to `invoke destroy build` whenever switching between a bugfix branch and a feature branch or vice versa.

Please pull this down and try it out.

# Screenshots

```
❯ invoke build
...
#39 naming to docker.io/local/nautobot-dev:local-2.3-py3.12 done
```

```
❯ invoke start
Starting Nautobot in detached mode...
Running docker compose command "up --detach"
PYTHON_VER=3.12 \
NAUTOBOT_VERSION=2.3 \
docker compose \
    --project-name "nautobot-2-3" \
    --project-directory "/Users/glennmatthews/Documents/git-stuff/nautobot/development/" \
    -f "/Users/glennmatthews/Documents/git-stuff/nautobot/development/docker-compose.yml" \
    -f "/Users/glennmatthews/Documents/git-stuff/nautobot/development/docker-compose.postgres.yml" \
    -f "/Users/glennmatthews/Documents/git-stuff/nautobot/development/docker-compose.dev.yml" up \
    --detach
 Network nautobot-2-3_default  Creating
 Network nautobot-2-3_default  Created
 Volume "nautobot-2-3_pgdata_nautobot"  Creating
 Volume "nautobot-2-3_pgdata_nautobot"  Created
 Volume "nautobot-2-3_media_root"  Creating
 Volume "nautobot-2-3_media_root"  Created
 Container nautobot-2-3-db-1  Creating
 Container nautobot-2-3-redis-1  Creating
 Container nautobot-2-3-selenium-1  Creating
 Container nautobot-2-3-redis-1  Created
 Container nautobot-2-3-selenium-1  Created
[...]

PR #6688, opened 2024-12-19
VSCode DevContainers bootstrapping cleanup

msheiny:
# Closes: None
# What's Changed

* Removes some documentation that was no longer relevant related to VSCode DevContainers. Specifically, there is no need  to create a `docker-compose.override.yml` to pass in the ARCH on Mac Silicon machines. This was fixed on the `Dockerfile` itself using [a global TARGETARCH variable at some point](https://github.com/nautobot/nautobot/pull/4906).
* Few couple helper `invoke` lines to generate an `.env` docker compose file for building. Not critical need but negates the need to keep this static default file up-to-date.
* Updates the `type` in the vscode workspace file from `python` -> `debugpy` (at some point they'll deprecate the `python` type completely according to their documentation. I just noticed VSCode was flagging it while I was inspecting the file.
* Adds the `development/docker-compose.vscode-rdb.yml` file to the DevContainer build file.

# TODO

- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments)) - Couldn't grok whether this was needed without a previous issue attached?
- [ ] ~Attached Screenshots, Payload Example~ N/A
- [ ] ~Unit, Integration Tests~ Not needed, purely dev environment changes
- [ ] ~Example App Updates (when adding/changing features)~ N/A
- [ ] ~Outline Remaining Work, Constraints from Design~ N/A

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -465,6 +465,12 @@ def vscode(context):
     """Launch Visual Studio Code with the appropriate Environment variables to run in a container."""
     command = "code nautobot.code-workspace"
 
+    # Setup PYTHON
+    env_file_path = os.path.join(BASE_DIR, "development/.env")
+    if not os.path.exists(env_file_path):
+        with open(env_file_path, "w") as env_file_obj:
+            env_file_obj.write(f"PYTHON_VER={context.nautobot.python_ver}")
+
     context.run(command, env={"PYTHON_VER": context.nautobot.python_ver})
 
 
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,7 +3,8 @@
     "dockerComposeFile": [
         "../development/docker-compose.yml",
         "../development/docker-compose.postgres.yml",
-        "../development/docker-compose.dev.yml"
+        "../development/docker-compose.dev.yml",
+        "../development/docker-compose.vscode-rdb.yml"
     ],
     "features": {
         "ghcr.io/devcontainers/features/docker-outside-of-docker:1": { }
--- a/changes/6688.housekeeping
+++ b/changes/6688.housekeeping
@@ -1,0 +1,1 @@
+Cleaned-up documentation, misc fixes for VSCode DevContainer workflow.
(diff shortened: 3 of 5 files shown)


glennmatthews (approved):
Thank you! I don't use the VSCode dev container myself but these changes look reasonable to me so I'll take your word for it :-)

glennmatthews:
> Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments)) - Couldn't grok whether this was needed without a previous issue attached?

The CI tooling does enforce such a file regardless, but for a PR with no prior issue, you can use the PR number instead. Thanks!

msheiny:
> > Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments)) - Couldn't grok whether this was needed without a previous issue attached?
> 
> The CI tooling does enforce such a file regardless, but for a PR with no prior issue, you can use the PR number instead (so in this case, probably `changes/6688.housekeeping` would be appropriate). Thanks!

okay, added to the latest revision 👍

tasks.py
```diff
@@ -465,6 +465,12 @@ def vscode(context):
     """Launch Visual Studio Code with the appropriate Environment variables to run in a container."""
     command = "code nautobot.code-workspace"
 
+    # Setup PYTHON
+    env_file_path = os.path.join(BASE_DIR, "development/.env")
+    if not os.path.exists(env_file_path):
+        with open(env_file_path, "w") as env_file_obj:
+            env_file_obj.write(f"PYTHON_VER={context.nautobot.python_ver}")
+
     context.run(command, env={"PYTHON_VER": context.nautobot.python_ver})
```
glennmatthews, line 474: I'm curious - after #6659, do we need to also pass `NAUTOBOT_VER` to either of these calls, or is it not needed?

msheiny: Ahh this call is just running `code nautobot.code-workspace` to reopen the current session using the workspace file so I dont really think we need `PYTHON_VER` passed into that at all. It's not like it keeps that environment around in my VSCode session (it doesn't seem to be visible in my terminal session at least)

PR #6691, opened 2024-12-20
documents different approach for initial population of data

### LINKED ISSUE #6671: Documentation on population of initial data
### As ...

P.D. - Nautobot App Developer

### I want ...

clarity on which mechanisms I can use to populate initial objects for my apps and jobs

### So that ...

I can make a factual decision on which mechanism I use

the documentation provides an overview over which of the following mechanisms are applicable when:

- Signal handler to `nautobot_database_ready`
- Creating objects in-place when they are needed
- Fixtures
- Manual jobs / design builder jobs
- Data migrations

[Relevant Django Docs](https://docs.djangoproject.com/en/5.1/howto/initial-data/)

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Organization
- [X] Apps (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [X] Documentation

### Database Changes

None

### External Dependencies

None

Kircheneer:
# Closes #6671
# What's Changed

See attached issue.

# TODO

- [x] I would appreciate some more input as I don't feel like my overview is comprehensive yet


nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
```diff
@@ -59,3 +63,76 @@ config = AnimalSoundsConfig
...
+**Pro**:
+
+- Have good support for usage in unit tests
+- Have good performance
+
+**Con**:
+
+- Execution: Manual
+- Implementation: Medium
+- Creating/updating them is not straight-forward, especially for big data sets
```
glennmatthews, line 114: - The possibility of later user modifications to data must be accounted for and handled in code.

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
```diff
@@ -59,3 +63,76 @@ config = AnimalSoundsConfig
...
+- Prevent Nautobot from starting when there are errors
+- Immutable—if you add new objects, you need to add a new migration
+- Can't easily be feature-toggled (if you include settings lookups in your migration and later change those settings, the migrations will _not_ run again)
+
+### Django Fixtures
+
+**Pro**:
+
+- Have good support for usage in unit tests
+- Have good performance
```
glennmatthews, line 108: - Users can later modify the created data like any other record, if a need arises.

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
```diff
@@ -59,3 +63,76 @@ config = AnimalSoundsConfig
...
+
+### Design Builder Jobs
+
+**Pro**:
+
+- Implementation: Easy
+
+**Con**:
+
+- Execution: Manual
```
glennmatthews, line 124: - The possibility of later user modifications to data must be accounted for and handled in code.

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
```diff
@@ -59,3 +63,76 @@ config = AnimalSoundsConfig
...
+## Advanced
+
+While the signal handler approach works for most use cases, it does have its downsides. Other options for the population of base state are available, each with their own trade-offs:
+
+- Signal Handlers on `nautobot_database_ready` (see above)
+- [Django Onboard Tooling](https://docs.djangoproject.com/en/4.2/howto/initial-data/)
+    - [Django Data Migrations](https://docs.djangoproject.com/en/4.2/topics/migrations/#data-migrations)
+    - [Django Fixtures](https://docs.djangoproject.com/en/4.2/topics/db/fixtures/#fixtures)
+- [Design Builder Jobs](https://docs.nautobot.com/projects/design-builder/en/latest/)
+- Creating data in-place where needed
```
glennmatthews, line 76: Given the repetition of pros/cons below, I wonder if it'd be possible and reasonable to create a table here of approaches versus their use cases/pros/cons?

glennmatthews (commented):
Thanks for taking a stab at this!

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
```diff
@@ -59,3 +63,76 @@ config = AnimalSoundsConfig
...
+- Provide a clean way to delete any associated objects when uninstalling the app
+
+**Con**:
+
+- Implementation: Medium
+- Prevent Nautobot from starting when there are errors
+- Immutable—if you add new objects, you need to add a new migration
+- Can't easily be feature-toggled (if you include settings lookups in your migration and later change those settings, the migrations will _not_ run again)
+
+### Django Fixtures
```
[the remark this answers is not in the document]
Kircheneer: Good question. I think the answer is "it depends". What do you think of scrapping pro/con altogether and just listing facts?

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
```diff
@@ -59,3 +63,76 @@ config = AnimalSoundsConfig
...
+
+- Execution: Automatic
+- Provide a clean way to delete any associated objects when uninstalling the app
+
+**Con**:
+
+- Implementation: Medium
+- Prevent Nautobot from starting when there are errors
+- Immutable—if you add new objects, you need to add a new migration
+- Can't easily be feature-toggled (if you include settings lookups in your migration and later change those settings, the migrations will _not_ run again)
```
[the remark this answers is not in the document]
Kircheneer: The startup performance, is it really significant? For first startups (i.e. when migrations run) and upgrades I agree, but consequent startups should have a very minor performance impact just to check that the migration has already been applied, right?

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
Kircheneer: I don't quite follow this one, I thought that with fixtures there would be no code involved? Or are you referencing fixtures generated through code?

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
Kircheneer: Are you thinking something that groups the common categories (e.g. execution type, implementation complexity, support for later modification, performance implications) into a table? That sounds like a reasonable approach and would also resolve my earlier question around stripping pro/con entirely. We could, e.g., add ➕ and ➖ into the table cells if needed to indicate whether something is a pro or a con.

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
glennmatthews: Sorry, I meant that since the data is created at a single point in time by the fixture, any runtime code that relies on the created data must account for the possibility that the user has modified the data in the interim.

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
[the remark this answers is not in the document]
glennmatthews: Agreed, it's mostly only a one-time cost to performance. Maybe doesn't need to be mentioned then.

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
glennmatthews: Yeah, I think that would be good.

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
Kircheneer: This pretty much applies to all options except the one that creates data in place, and even that option has to account for the possibility of that data already existing/being modified, right?

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
```diff
@@ -1,10 +1,14 @@
...
+Furthermore, sometimes apps might want to create a baseline of available data. This could, for example, be a baseline of circuit providers for a [Single Source of Truth](https://docs.nautobot.com/projects/ssot/en/latest/) integration that synchronizes circuit data or creating new default [statuses](https://docs.nautobot.com/projects/core/en/stable/user-guide/platform-functionality/status/) and/or [roles](https://docs.nautobot.com/projects/core/en/stable/user-guide/platform-functionality/role/), or adding to the allowed content types for existing statuses and roles.
+
+## Using A Signal Handler
+
++++ 1.2.0
+
+In general, it is recommended to use custom [signal](https://docs.djangoproject.com/en/stable/topics/signals/), `nautobot_database_ready` that apps can register to listen for. This signal is triggered when `nautobot-server migrate` or `nautobot-server post_upgrade` is run after installing an app, and provides an opportunity for the app to make any desired additions to the database at this time.
```
glennmatthews, line 11: Having just gone through updating a *bunch* of app signals (autocreating statuses, roles, custom fields, relationships, etc.) from Nautobot v1 to v2 for a customer, I'm no longer certain this is the best practice we want to recommend. If the data population had been implemented via data migrations instead, the existing data would have been automatically and correctly migrated to v2 data models as a part of any upgrade; because the signals run every time the app starts, they all had to be modified/rewritten to account for the v2 data model changes. Just food for thought.

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
```diff
@@ -1,10 +1,14 @@
-# Populating Extensibility Features
-
-+++ 1.2.0
+# Populating Extensibility Feature And Base Data
```
glennmatthews, line 1: Maybe update the entry in `mkdocs.yml` as well, perhaps just as `Prepopulating Data`?

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
```diff
@@ -59,3 +63,61 @@ config = AnimalSoundsConfig
...
+- Signal Handlers on `nautobot_database_ready` (see above)
+- [Django Onboard Tooling](https://docs.djangoproject.com/en/4.2/howto/initial-data/)
+    - [Django Data Migrations](https://docs.djangoproject.com/en/4.2/topics/migrations/#data-migrations)
+    - [Django Fixtures](https://docs.djangoproject.com/en/4.2/topics/db/fixtures/#fixtures)
+- [Design Builder Jobs](https://docs.nautobot.com/projects/design-builder/en/latest/)
+- Creating data in-place where needed
+
+The following table details points that are somewhat easily comparable between the different options.
+
+| Option                                       | Startup Impact                       | Execution | Performance | Re-running | Idempotence | Modifications Possible      | Custom App Needed |
```
glennmatthews, line 80: *might* be worth flipping the axes of this table (so the "options" are across the top and the "features" are rows) as I think this will get hard to read as the number of features to consider increases but we don't expect to add many more "options" over time? Would also make it possible to elaborate on the "features" a bit more verbosely.

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
```diff
@@ -59,3 +63,61 @@ config = AnimalSoundsConfig
...
+|----------------------------------------------|--------------------------------------|-----------|-------------|------------|-------------|-----------------------------|-------------------|
+| Signal Handlers on `nautobot_database_ready` | Exceptions prevent container startup | Automatic | Average     | Possible   | Enforced    | Overwrite/crash on next run | Yes               |
```
glennmatthews, line 82: Seems like the `Re-running` and `Modifications Possible` columns always align - should they be combined into one column?

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
```diff
@@ -59,3 +63,61 @@ config = AnimalSoundsConfig
...
+
+The following sections detail additional points that weren't possible to fit.
+
+### Signal Handlers On `nautobot_database_ready`
+
+- ➕ Can be updated in-place in later App releases
+- Re-runs every time `nautobot-server migrate` or `nautobot-server post_upgrade` is run
+    - ➖ Must be idempotent, as such it mustn't introduce duplicate records or errors if run repeatedly
+    - ➖ May recreate or revert records that a user intentionally deleted or altered - if this is not desired it has to be accounted for and handled in the code
+- ➖ Impacts Nautobot startup and upgrade performance, care needs to be taken
```
glennmatthews, line 96:
```suggestion
- ➖ Impacts Nautobot startup and upgrade performance, care needs to be taken
- ➖ May require updates over time to remain operational as Nautobot's data model evolves
```

(Alternately, this probably should be another item in the table above as it's a factor for many of these approaches)

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
Kircheneer: I like the general idea. There is another point about unit testing there, I've found some weird edge cases where sometimes with `--keepdb` the signal-handler-created objects just disappear and you end up troubleshooting weird issues. Do we have real world experience using data migrations over signal handlers? I want to make sure that we don't recommend this now and then wind up finding out it is equally or more annoying to deal with these.

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
Kircheneer: I've gone and changed everything to this, I like the title.

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
Kircheneer: Sounds good, done!

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
Kircheneer: I went with the alternative approach.

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
Kircheneer: Done!

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
glennmatthews: That's a fair question. We know that the `Status` and `Role` objects created by data migrations in core don't stick around when using `TransactionTestCase` (https://docs.djangoproject.com/en/4.2/topics/testing/overview/#rollback-emulation) (it would be interesting to experiment with `serialized_rollback = True` but I don't believe that's something we've explored...) but otherwise I'd expect them to be present - but as you say, I'm not sure how much hands-on evidence we have about that specific concern.

mkdocs.yml
```diff
@@ -595,7 +595,7 @@ nav:
               - Git Repository Content: "development/apps/api/platform-features/git-repository-content.md"
               - Jinja2 Filters: "development/apps/api/platform-features/jinja2-filters.md"
               - Jobs: "development/apps/api/platform-features/jobs.md"
-              - Populating Extensibility Features: "development/apps/api/platform-features/populating-extensibility-features.md"
+              - Prepopulating Data: "development/apps/api/platform-features/prepopulating-data.md"
```
glennmatthews, line 598: Might be good to add a redirect from the old file name (around line 301 of this file)

nautobot/docs/development/apps/api/platform-features/prepopulating-data.md
```diff
@@ -0,0 +1,125 @@
...
+|-------------------|--------------------------------------------------------------------------------------|--------------------------------------|--------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------|
+| Startup Impact    | Exceptions prevent container startup                                                 | Exceptions prevent container startup | N/A                                                                                  | N/A                                                                                  | N/A                                                                                  |
```
glennmatthews, line 82:
```suggestion
| Startup Impact    | Exceptions occur after database migrations                                                 | Exceptions occur during database migrations | N/A                                                                                  | N/A                                                                                  | N/A                                                                                  |
```

nautobot/docs/development/apps/api/platform-features/prepopulating-data.md
```diff
@@ -0,0 +1,125 @@
...
+| Execution         | Automatic                                                                            | Automatic                            | Manual                                                                               | Manual                                                                               | Automatic                                                                            |
+| Performance       | Average                                                                              | Average                              | Good                                                                                 | Average                                                                              | Average                                                                              |
```
glennmatthews, line 84: Four "Average" and one "Good" doesn't provide a lot of differentiation. :) Perhaps:

```suggestion
| Performance Impact | Each time migrations are rerun | Once when migrations are first run | Once when fixture is loaded | Each time design job is run | Each time the data is needed |
```

nautobot/docs/development/apps/api/platform-features/prepopulating-data.md
```diff
@@ -0,0 +1,125 @@
...
+| Re-running        | Possible, modified data may be overwritten/cause crashes                             | Impossible                           | Possible, modified data may be overwritten/cause crashes                             | Possible, modified data may be overwritten/cause crashes                             | Possible, modified data may be overwritten/cause crashes                             |
+| Idempotence       | Enforced                                                                             | N/A                                  | TBD                                                                                  | Possible                                                                             | Possible                                                                             |
```
glennmatthews, line 86: Still feel like maybe these two rows are redundant with one another?

nautobot/docs/development/apps/api/platform-features/prepopulating-data.md
```diff
@@ -0,0 +1,125 @@
...
+| Custom App Needed | Yes                                                                                  | Yes                                  | No                                                                                   | No                                                                                   | No                                                                                   |
+| Maintenance       | May require updates over time to remain operational as Nautobot's data model evolves | No maintenance required              | May require updates over time to remain operational as Nautobot's data model evolves | May require updates over time to remain operational as Nautobot's data model evolves | May require updates over time to remain operational as Nautobot's data model evolves |
```
glennmatthews, line 88:
```suggestion
| Maintenance required over time | Moderate | None | Moderate | Moderate | High |
```

nautobot/docs/development/apps/api/platform-features/populating-extensibility-features.md
Kircheneer: Okay, since we have scrapped having a recommendation now I think I would be fine resolving this, WDYT?

nautobot/docs/development/apps/api/platform-features/prepopulating-data.md
```diff
@@ -0,0 +1,183 @@
...
+- Creating data in-place where needed
+
+Some of these methods are explained in further detail at the bottom of this page.
+
+The following table details points that are somewhat easily comparable between the different options.
+
+| Feature                        | Signal Handlers on `nautobot_database_ready`             | Django Data Migrations                      | Django Fixtures                                          | Design Builder Jobs                                      | Creating Data In-place Where Needed                      |
+|--------------------------------|----------------------------------------------------------|---------------------------------------------|----------------------------------------------------------|----------------------------------------------------------|----------------------------------------------------------|
+| Startup Impact                 | Exceptions occur after database migrations               | Exceptions occur during database migrations | N/A                                                      | N/A                                                      | N/A                                                      |
```
glennmatthews, line 24: Might suggest bolding the left column here (`**Startup Impact**`, `**Execution**`, etc.) to improve readability.

nautobot/docs/development/apps/api/platform-features/prepopulating-data.md
```diff
@@ -0,0 +1,183 @@
...
+| Performance Impact             | Each time migrations are rerun                           | Once when migrations are first run          | Once when fixture is loaded                              | Each time design job is run                              | Each time the data is needed                             |
+| Re-running / Idempotence       | Possible, modified data may be overwritten/cause crashes | Impossible                                  | Possible, modified data may be overwritten/cause crashes | Possible, modified data may be overwritten/cause crashes | Possible, modified data may be overwritten/cause crashes |
+| Custom App Needed              | Yes                                                      | Yes                                         | No                                                       | No                                                       | No                                                       |
```
glennmatthews, line 28: Should `Custom App Needed` be `Yes` under `Creating Data In-place where Needed`?

nautobot/docs/development/apps/api/platform-features/prepopulating-data.md
```diff
@@ -0,0 +1,183 @@
...
+
+    def ready(self):
+        super().ready()
+        nautobot_database_ready.connect(create_location_to_animal_relationship, sender=self)
+
+config = AnimalSoundsConfig
+```
+
+!!! warning
+It is crucial that you add the `sender=self` parameter to the `connect` method call - otherwise your signal handler will run as many times as the `NautobotAppConfig.ready` method is called, which may be a lot of times.
```
glennmatthews, line 128:
```suggestion
!!! warning
    It is crucial that you add the `sender=self` parameter to the `connect` method call - otherwise your signal handler will run as many times as the `NautobotAppConfig.ready` method is called, which may be a lot of times.
```

nautobot/docs/development/apps/api/platform-features/prepopulating-data.md
```diff
@@ -0,0 +1,183 @@
...
+}
+
+def add_content_types_to_default_statuses(apps, schema_editor):
+    """Adds additional content types to default statuses."""
+    Status = apps.get_model("extras.Status")
+    ContentType = apps.get_model("contenttypes.ContentType")
+    for model, statuses in status_content_type_mapping.items():
+        model_class = apps.get_model(model)
+        for status in statuses:
+            Status.objects.get(name=status).content_types.add(
```
glennmatthews, line 157: Should we use `Status.objects.get_or_create()` here and below as a best practice in case those named statuses don't exist?

glennmatthews (commented):
I like this!

dsoftait:
Hi @Kircheneer. Will you continue these changes ?

PR #6693, opened 2024-12-20
Updated Dockerfile, prerelease.yml and release.yml to address release workflow failure

HanlinMiao:
1. Changed container builds from fail-fast: true to fail-fast: false
2. Added serial retry to poetry install if parallel install fails.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -151,6 +151,9 @@ RUN --mount=type=cache,target="/root/.cache",sharing=locked \
         echo "Parallel installation failed, retrying serially"; \
         poetry config installer.parallel false; \
         poetry install --no-root --only main --no-ansi --extras all; \
+    else \
+        echo "Serial installation failed, exiting"; \
+        exit 1; \
     fi
 
 RUN rm -rf /tmp/tmp*
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -114,7 +114,7 @@ FROM python:${PYTHON_VER}-slim AS poetry
 # https://python-poetry.org/docs/master/#installing-with-the-official-installer
 # This also makes it so that Poetry will *not* be included in the "final" image since it's not installed to /usr/local/
 ARG POETRY_HOME=/opt/poetry
-ARG POETRY_INSTALLER_PARALLEL=false
+ARG POETRY_INSTALLER_PARALLEL=true
 ARG POETRY_VERSION=1.8.2
 ARG POETRY_VIRTUALENVS_CREATE=false
 ADD https://install.python-poetry.org /tmp/install-poetry.py
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -145,17 +145,15 @@ WORKDIR /source
 
 # Install (non-development) Python dependencies of Nautobot
 RUN --mount=type=cache,target="/root/.cache",sharing=locked \
-    if poetry install --no-root --only main --no-ansi --extras all && \
-        rm -rf /tmp/tmp*; then \
+    if poetry install --no-root --only main --no-ansi --extras all; then \
         echo "Parallel installation succeeded"; \
     else \
         echo "Parallel installation failed, retrying serially"; \
-        poetry config virtualenvs.create ${POETRY_VIRTUALENVS_CREATE} && \
-        poetry config installer.parallel false && \
-        poetry config installer.no-binary lxml,pyuwsgi,xmlsec && \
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -144,9 +144,19 @@ COPY --chown=nautobot:nautobot examples /source/examples
 WORKDIR /source
 
 # Install (non-development) Python dependencies of Nautobot
-RUN --mount=type=cache,target="/root/.cache",sharing=locked \
+RUN if --mount=type=cache,target="/root/.cache",sharing=locked \
```
glennmatthews, line 147: Have you tested this locally? I'm pretty sure that `--mount,`etc. is an argument to `RUN` and not actually part of the command being run.

docker/Dockerfile
```diff
@@ -144,9 +144,19 @@ COPY --chown=nautobot:nautobot examples /source/examples
...
+    rm -rf /tmp/tmp*, then \
+        echo "Parallel installation succeeded"; \
+    else \
+        echo "Parallel installation failed, retrying serially"; \
+        poetry config virtualenvs.create ${POETRY_VIRTUALENVS_CREATE} && \
+        poetry config installer.parallel false && \
+        poetry config installer.no-binary lxml,pyuwsgi,xmlsec && \
+        --mount=type=cache,target="/root/.cache",sharing=locked \
+        poetry install --no-root --only main --no-ansi --extras all && \
+        rm -rf /tmp/tmp*
```
glennmatthews, line 158: The `rm -rf /tmp/tmp*` should probably be a separate statement rather than part of the conditional logic.

docker/Dockerfile
HanlinMiao: Yep just tested it locally and --mount should be only on the RUN command.

docker/Dockerfile
```diff
@@ -145,8 +145,15 @@ WORKDIR /source
 
 # Install (non-development) Python dependencies of Nautobot
 RUN --mount=type=cache,target="/root/.cache",sharing=locked \
-    poetry install --no-root --only main --no-ansi --extras all && \
-    rm -rf /tmp/tmp*
+    if poetry install --no-root --only main --no-ansi --extras all; then \
+        echo "Parallel installation succeeded"; \
+    else \
```
glennmatthews, line 150: Note that `POETRY_INSTALLER_PARALLEL` is a build arg, so maybe for most correctness this should be something like:

```suggestion
        echo "Installation succeeded"; \
    elif [ ${POETRY_INSTALLER_PARALLEL} == "true" ]; then \
```

glennmatthews (approved):
Looks reasonable to me. @bryanculver if you have the bandwidth, I'd appreciate another set of eyes on this, particularly the workflow change.

docker/Dockerfile
HanlinMiao: hadolint was complaining lmk if this change looks reasonable https://github.com/nautobot/nautobot/pull/6693/commits/6cfad49b812f2577241af86c07297d03f13f80b8

docker/Dockerfile
```diff
@@ -145,8 +145,15 @@ WORKDIR /source
...
 RUN --mount=type=cache,target="/root/.cache",sharing=locked \
-    poetry install --no-root --only main --no-ansi --extras all && \
-    rm -rf /tmp/tmp*
+    if poetry install --no-root --only main --no-ansi --extras all; then \
+        echo "Installation succeeded"; \
+    elif [ "${POETRY_INSTALLER_PARALLEL}" = "true" ]; then \
+        echo "Parallel installation failed, retrying serially"; \
+        poetry config installer.parallel false; \
+        poetry install --no-root --only main --no-ansi --extras all; \
+    fi
```
glennmatthews, line 154: Presumably we should now have an `else` that fails out?

docker/Dockerfile
```diff
@@ -114,7 +114,7 @@ FROM python:${PYTHON_VER}-slim AS poetry
 # https://python-poetry.org/docs/master/#installing-with-the-official-installer
 # This also makes it so that Poetry will *not* be included in the "final" image since it's not installed to /usr/local/
 ARG POETRY_HOME=/opt/poetry
-ARG POETRY_INSTALLER_PARALLEL=true
+ARG POETRY_INSTALLER_PARALLEL=false
```
glennmatthews, line 117: I'd prefer to leave this defaulted to true for local development purposes. Can we override it in the release and prerelease workflows instead?

glennmatthews (approved):

PR #6704, opened 2024-12-30
Resolve issue with TagsBulkEditFormMixin not having access to self.model

### LINKED ISSUE #6702: TagsBulkEditFormMixin should inherit from BulkEditForm instead of forms.Form
Now that TagsBulkEditForm accesses self.model, it needs BulkEditForm instead of forms.Form to enforce proper MRO, otherwise you can get an AttributeError if specifying TagsBulkEditFormMixin after NautobotBulkEditForm.  See https://github.com/nautobot/nautobot-app-data-validation-engine/issues/191

### Environment
* Nautobot version (Docker tag too if applicable): 2.3.14
* Python version: 3.11
* Database platform, version: N/A
* Middleware(s):

### Steps to Reproduce
1. Create a  BulkEditForm for a model.
2. Change the Inheritance for that BulkEditForm to be `(NautobotBulkEditForm, TagsBulkEditFormMixin)`
3. Try to Bulk Edit that model.

### Expected Behavior

The Bulk Edit Form would load, and it doesn't matter what order I define my class inheritance in. 

### Observed Behavior

AttributeError at /plugins/nautobot-data-validation-engine/min-max-rules/edit/
'MinMaxValidationRuleBulkEditForm' object has no attribute 'model'

smk4664:
# Closes #6702 
# What's Changed

TagsBulkEditFormMixin now inherits from BulkEditForm which gives it access to self.model. This allows a Form to inherit TagsBulkEditFormMixin in any order.

# Screenshots

# TODO

- [x] Unit, Integration Tests


nautobot/extras/forms/mixins.py
```diff
@@ -837,7 +837,7 @@ def __init__(self, *args, **kwargs):
         self.order_fields(self.field_order)  # Reorder fields again
 
 
-class TagsBulkEditFormMixin(forms.Form):
+class TagsBulkEditFormMixin(BulkEditForm):
```
glennmatthews, line 840: Should we make similar changes to the other classes above (e.g. `RoleModelBulkEditFormMixin`, `StatusModelBulkEditFormMixin`) in the name of completeness? Or does that risk unintended side effects?

smk4664: Ah yes, both of these utilize self.model, but inherit from forms.Form which does not have the model attribute.

glennmatthews (approved):

PR #6713, opened 2025-01-02
Removed the need for available-prefixes, available-ips, and available-vlans API endpoints to run validation multiple times

### LINKED ISSUE #6669: `available-prefixes` deserialization causes validation to run multiple times
### Environment
* Nautobot version (Docker tag too if applicable): 2.3
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Create a prefix via `available-prefixes`
2. Have a custom validator setup on prefix, that relies on the `.prefix` or `.ancestors()` methods
3.

### Expected Behavior
Validation to run only once, with all properties and data populated

### Observed Behavior
Validation is run multiple times, with the object in various partial states.

HanlinMiao:
# Closes #6669
# What's Changed

1. Removed the need for `available-prefixes`, `available-ips`, and `available-vlans` API endpoints to run validation multiple times
2. Removed the `serializer.is_valid()` method calls prior to object creations using those API endpoints
3. Replaced the `serializer.is_valid()` method calls with manual validation

# Screenshots

N/A
# TODO

- [x] Unit, Integration Tests


nautobot/ipam/api/views.py
```diff
@@ -487,20 +452,24 @@ def available_vlans(self, request, pk=None):
...
-                        return Response(
-                            {
-                                "detail": f"Invalid VLAN Group requested: {requested_vlan['vlan_group']}. "
-                                f"Only VLAN Group {vlan_group} is permitted."
-                            },
-                            status=status.HTTP_204_NO_CONTENT,
-                        )
+                    if "vlan_group" in requested_vlan:
+                        requested_vlan_group = None
+                        if "vlan_group" in requested_vlan:
```
glennmatthews, line 457: Nested identical `if` conditions seems wrong to me?

nautobot/ipam/api/views.py
```diff
@@ -487,20 +452,24 @@ def available_vlans(self, request, pk=None):
...
+                            try:
+                                requested_vlan_group = VLANGroup.objects.get(pk=requested_vlan["vlan_group"])
+                            except VLANGroup.DoesNotExist:
+                                pass
```
glennmatthews, line 461: Should this be a failure case?

nautobot/ipam/tests/test_api.py
```diff
@@ -506,7 +506,7 @@ def test_create_single_available_prefix(self):
             url, {"prefix_length": "hello", "status": self.status.pk}, format="json", **self.header
         )
         self.assertHttpStatus(response, status.HTTP_400_BAD_REQUEST)
-        self.assertIn("prefix_length", response.data[0])
+        self.assertIn("prefix", response.data)
```
glennmatthews, line 509: Change here seems wrong to me - shouldn't our 400 response still indicate that `prefix_length` was the invalid input parameter?

HanlinMiao: I think it is because `request_prefixes` no longer contains the "prefix" field because we are not using serializer.validated_data. Let me see what I can come up with.

glennmatthews (approved):

PR #6721, opened 2025-01-03
Enabled extended filter lookup expressions of the serial filter for Device, Rack, and InventoryItem

### LINKED ISSUE #5805: Serial number of a device does not support extended filters
### Environment
* Nautobot version (Docker tag too if applicable): v2.2.5
* Python version: 3.11
* Middleware(s): None

### Steps to Reproduce
Execute graphql query to get a list of all devices that have no serial number
``
{
  devices (serial:"") {
    name
    serial
  }
}
``

For most string filters there’s a __nre (not regex) variant, which could do something like name__nre=".+", but looks like serial doesn’t have any of the variants enabled.

### Expected Behavior
To get a list of all devices with no serial number.

### Observed Behavior
I got a list of all devices

HanlinMiao:
# Closes #5805 
# What's Changed

1. Enabled extended filter lookup expressions of the serial filter for Device, Rack, and InventoryItem
# Screenshots

Screenshot 2025-01-03 at 11 58 24 AM
Screenshot 2025-01-03 at 11 58 34 AM

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/tests/test_graphql.py
+++ b/nautobot/dcim/tests/test_graphql.py
@@ -86,30 +86,30 @@ class GraphQLTestCase(TestCase):
 
         with self.subTest("device serial number query"):
             non_empty_serial_device = Device.objects.first()
-            non_empty_serial_device.serial = "1234567890"
+            non_empty_serial_device.serial = "1234567890abceFGHIJKL"
             non_empty_serial_device.save()
 
             # Test device serial query default behavior: serial__ie
-            query = 'query { devices (serial:" ") { name serial } }'
+            query = 'query { devices (serial: " ") { name serial } }'
             resp = execute_query(query, user=self.user).to_dict()
             self.assertFalse(resp["data"].get("error"))
             for device in resp["data"]["devices"]:
                 self.assertEqual(device["serial"], "")
 
             # Test device serial default filter with non-empty serial number
-            query = 'query { devices (serial:"' + non_empty_serial_device.serial + '") { name serial } }'
+            query = 'query { devices (serial:"' + non_empty_serial_device.serial.lower() + '") { name serial } }'
             resp = execute_query(query, user=self.user).to_dict()
             self.assertFalse(resp["data"].get("error"))
             self.assertEqual(resp["data"]["devices"][0]["serial"], non_empty_serial_device.serial)
 
             # Test device serial iexact filter with non-empty serial number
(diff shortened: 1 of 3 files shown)


nautobot/dcim/tests/test_graphql.py
```diff
@@ -83,3 +83,55 @@ def test_execute_query(self):
             self.assertEqual(
                 resp["data"]["devices"][0]["platform"]["network_driver_mappings"]["scrapli"], "cisco_iosxe"
             )
+
+        with self.subTest("device serial number query"):
+            non_empty_serial_device = Device.objects.first()
+            non_empty_serial_device.serial = "1234567890"
```
glennmatthews, line 89: Can we use a serial that contains letters so that we can assert that the default filtering is still case-insensitive?

nautobot/dcim/tests/test_graphql.py
```diff
@@ -83,3 +83,55 @@ def test_execute_query(self):
...
+            non_empty_serial_device.save()
+
+            # Test device serial query default behavior: serial__ie
+            query = 'query { devices (serial:" ") { name serial } }'
```
glennmatthews, line 93: that seems weird - we're specifying a space character as the serial value and expecting it to match devices with an empty serial? Do we or django-filters auto-strip whitespace or something?

nautobot/dcim/tests/test_graphql.py
```diff
@@ -83,3 +83,55 @@ def test_execute_query(self):
...
+            resp = execute_query(query, user=self.user).to_dict()
+            self.assertFalse(resp["data"].get("error"))
+            for device in resp["data"]["devices"]:
+                self.assertEqual(device["serial"], "")
+
+            # Test device serial default filter with non-empty serial number
+            query = 'query { devices (serial:"' + non_empty_serial_device.serial + '") { name serial } }'
```
glennmatthews, line 100: If we had letters in the serial string, we should use either `.upper()` or `.lower()` here to assert case insensitivity.

nautobot/dcim/tests/test_graphql.py
HanlinMiao: I think we are only appending non-empty values [here](https://github.com/nautobot/nautobot/blob/develop/nautobot/core/forms/fields.py#L280-L285). Honestly not sure if this is the correct way to approach this.

glennmatthews (approved):
Not sure this is the "right" fix long-term but it seems reasonable for now.

PR #6722, opened 2025-01-03
Enable remaining pylint "error" checkers

glennmatthews:
# What's Changed

- Enable pylint `not-callable` checker and address issues it raised
   - In particular, a lot of our generic views implicitly require definition of certain class attributes when subclassing them, but were defaulting them to None, which meant that a subclass that omitted those attributes would load perfectly well but throw an exception when actually accessed (due to the unexpected None). I changed these to abstract/type-only attributes, which makes Pylint much happier, but may have a downstream impact to any App that implements such an invalid view as it will (I think) change from a runtime error to an import-time error. We'll want to watch for impacts in upstream CI when this is merged.
- Enable pylint `no-member` checker and address issues it raised
  - Unfortunately there are a lot of false positives to this one (in part due to pylint-django not being a very actively maintained extension) that I had to add `pylint: disable` comments for, but it did catch probably a half dozen legitimate bugs as well (which I've fixed), so it's well worth enabling.

With this PR, we no longer have any "Error" severity Pylint checkers disabled, which I think is a good step forward in code quality assurance. :-)

# TODO

- [x] Unit, Integration Tests


nautobot/extras/health_checks.py
```diff
@@ -160,5 +160,4 @@ def check_status(self):
         elif client_class == "django_redis.client.DefaultClient":
             self.check_redis(redis_url=location, **options.get("CONNECTION_POOL_KWARGS", {}))
         else:
-            if self.redis_url is None:
-                self.add_error(ServiceUnavailable(f"{client_class} is an unsupported CLIENT_CLASS!"))
+            self.add_error(ServiceUnavailable(f"{client_class} is an unsupported CLIENT_CLASS!"))
```
glennmatthews, line 163: Seems like an actual bug to me - I couldn't see anywhere `self.redis_url` was defined otherwise.

nautobot/extras/migrations/0018_joblog_data_migration.py
```diff
@@ -123,14 +122,13 @@ def reverse_migrate_params(apps, schema_editor):
                 ]
             )
 
-            if entry.log_level != LogLevelChoices.LOG_DEFAULT:
```
glennmatthews, line 126: Actual bug caught by `no-member`.

nautobot/extras/models/groups.py
```diff
@@ -147,19 +148,19 @@ def _map_filter_fields(self):
         """Return all FilterForm fields in a dictionary."""
 
         # Fail gracefully with an empty dict if nothing is working yet.
-        if not self.form_class:
+        if self.form_class is None or self.filterform_class is None or self.filterset_class is None:
```
glennmatthews, line 151: I tried to add enough type annotations and checks here that `not-callable` wouldn't trip below, but ultimately gave up.

nautobot/extras/models/jobs.py
```diff
@@ -610,7 +610,7 @@ def set_status(self, status):
             # Only add metrics if we have a related job model. If we are moving to a terminal state we should always
             # have a related job model, so this shouldn't be too tight of a restriction.
             if self.job_model:
-                duration = self.date_done - self.created
+                duration = self.date_done - self.date_created
```
glennmatthews, line 613: Actual bug caught by `no-member`.

nautobot/extras/test_jobs/fail.py
```diff
@@ -7,15 +7,15 @@
 logger = get_task_logger(__name__)
 
 
-class TestFail(Job):
+class TestFailJob(Job):
```
glennmatthews, line 10: I renamed `TestFail` and `TestPass` to `TestFailJob` and `TestPassJob` so that the `generated-members` config added in pyproject.toml for `.*Job.request` would stop `no-member` from complaining about their multiple references to `self.request`.

nautobot/ipam/factory.py
```diff
@@ -238,9 +238,9 @@ def locations(self, create, extracted, **kwargs):
                         lambda: Location.objects.filter(location_type__content_types__in=[vlan_ct]), minimum=0
                     )
                 )
-                if self.vlan_group and self.vlan_group.location:
+                if self.vlan_group and self.vlan_group.location:  # pylint: disable=no-member
```
glennmatthews, line 241: The way factory classes work seems to frequently confuse `no-member` - it might be better to add a blanket `generated-members` rule in pyproject.toml for something like `.*Factory\..*`?

nautobot/ipam/tests/test_forms.py
```diff
@@ -9,62 +10,63 @@
...
-            prefix="192.168.1.0/24", namespace=self.namespace, status=self.prefix_status
-        )
-        self.parent2 = Prefix.objects.create(
-            prefix="192.168.0.0/16", namespace=self.namespace, status=self.prefix_status
-        )
-        self.parent6 = Prefix.objects.create(
-            prefix="2001:0db8::/40", namespace=self.namespace, status=self.prefix_status
-        )
+class NetworkFormTestCases:
+    class BaseNetworkFormTest(TestCase):
```
glennmatthews, line 14: Changed this to a TestCase subclass so that pylint wouldn't complain about it trying to call methods like `self.assertTrue`. Moved it to an inner class (similar to other generic test base classes) so that unittest won't discover and run this base class directly.

glennmatthews:
Bah, python 3.8 support strikes again (`type[Foo]` is only supported in 3.9 and later, in 3.8 you have to use `typing.Type[Foo]` but that's deprecated in all later versions.) I may see about rebasing and retargeting this PR to `next` next week if I can find time.

smk4664:
> Bah, python 3.8 support strikes again (`type[Foo]` is only supported in 3.9 and later, in 3.8 you have to use `typing.Type[Foo]` but that's deprecated in all later versions.) I may see about rebasing and retargeting this PR to `next` next week if I can find time.

Does `from __future__ import annotations` work?

HanlinMiao (approved):

PR #6736, opened 2025-01-07
Fixed inconsistent rendering of DynamicModelChoiceField throughout the UI.

### LINKED ISSUE #6461: Fix inconsistency in DynamicModelChoiceField use cases
Fix the inconsistent default/null option behavior in the `DynamicModelChoiceField` across its implementation use cases.

In this example there are concrete model fields and non-concrete filter fields in the form and there is inconsistency in the usage of `----` vs `None`.

Additionally, the drop down arrow on the right seems to be inconsistently present, depending on the the field.

In both cases, standardize the usage of these fields/widgets and ideally never leave it up to the app developer to have to make a choice.

HanlinMiao:
# Closes #6461
# What's Changed

Fixed inconsistent rendering of DynamicModelChoiceField throughout the UI by modifying forms.js
# Screenshots

Fix tested In next as well where the issue is originated from:
Screenshot 2025-01-07 at 11 19 19 AM
Screenshot 2025-01-07 at 11 19 27 AM

# TODO

- [ ] Unit, Integration Tests


HanlinMiao:
> From the screenshots, it looks like the down-arrow is still missing in the `VLAN Locations (filter)` widget, any idea why?

Yeah I think `VLAN Locations (filter)` is a `DynamicModelMultipleChoiceField` instead of a `DynamicModelChoiceField`.
For example, `TenantGroup and Tenant` --> `DynamicModelChoiceField` and `DynamicGroups and tags` --> `DynamicModelMultipleChoiceField`. They are of diffe
rent classes with select2 as well. `DynamicModelChoiceField` is of class "select2-selection__rendered" (with arrow) and `DynamicModelMultipleChoiceField` is of class "select2-search__field" (without arrow). I will see if I can unify them in some way.
Screenshot 2025-01-07 at 12 45 55 PM

HanlinMiao:
> > From the screenshots, it looks like the down-arrow is still missing in the `VLAN Locations (filter)` widget, any idea why?
> 
> Yeah I think `VLAN Locations (filter)` is a `DynamicModelMultipleChoiceField` instead of a `DynamicModelChoiceField`. For example, `TenantGroup and Tenant` --> `DynamicModelChoiceField` and `DynamicGroups and tags` --> `DynamicModelMultipleChoiceField`. They are of different classes with select2 as well. `DynamicModelChoiceField` is of class "select2-selection__rendered" (with arrow) and `DynamicModelMultipleChoiceField` is of class "select2-search__field" (without arrow). I will see if I can unify them in some way. Screenshot 2025-01-07 at 12 45 55 PM

With some investigation, it seems this line gives the arrow to "select2-selection__rendered" in `select2.min.js`.
`e.html('<span class="select2-selection__rendered"></span><span class="select2-selection__arrow" role="presentation"><b role="presentation"></b></span>')`. In short, it is not our code that is responsible for this.

glennmatthews:
FWIW, the current logic was introduced in #3315 to address #3187.

HanlinMiao:
> FWIW, the current logic was introduced in #3315 to address #3187.

Idk we should probably introduce a new parameter like `default_option` or something for that fix. Auto-filling the form field with the `null_option` in the case of default/required fields does not sit right with me and probably is confusing UI behavior.

glennmatthews:
> > FWIW, the current logic was introduced in #3315 to address #3187.
> 
> Idk we should probably introduce a new parameter like `default_option` or something to fix #3187. Auto-filling the form field with the `null_option` in the case of default/required fields does not sit right with me and probably is confusing UI behavior.

I agree. This seems similar to Django's `empty_label` attribute but is something we can revisit in future.

glennmatthews (approved):

PR #6746, opened 2025-01-08
Added the ability to perform a shallow copy of a GitRepository instance and optionally checkout a different branch and/or a specific commit hash.

### LINKED ISSUE #6487: Clone a GitRepository to a temporary location
### As ...

P.D. - Nautobot App Developer

### I want ...

To be able to clone a `GitRepository` to a temporary location ([`tempfile.mkdtemp`](https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp)) and optionally checkout a branch that is not the `GitRepository.branch`.

### So that ...

My app can support development and testing of various `GitRepository` provided features directly within Nautobot.

I can use a Python API in `nautobot.apps` or on the `GitRepository` instance to clone a repository and a reference to the file path is returned where the repository has been cloned/checked out.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

_No response_

### External Dependencies

_No response_

HanlinMiao:
# Closes #6487 
# What's Changed

Added `clone()` method on `GitRepository` model class to add the ability to perform a shallow copy of a `GitRepository` instance and to optionally checkout a different branch and/or a specific commit hash.

# Screenshots

Calling the clone method:
Screenshot 2025-01-08 at 2 08 56 PM

Temp directories created as a result of calling clone():
Screenshot 2025-01-08 at 2 12 02 PM

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/models/datasources.py
```diff
@@ -167,3 +167,56 @@ def sync(self, user, dry_run=False):
...
+            GitRepository
+        """
+        import copy
+        import uuid
+
+        # import tempfile
+        from nautobot.core.utils.git import GitRepo
+        from nautobot.extras.datasources.git import get_repo_from_url_to_path_and_from_branch
+
+        # tempfile.mkdtemp(dir=self.filesystem_path + "_clone")
```
HanlinMiao, line 189: I am confused about where this method call would come in and play a role.

changes/6487.added
```diff
@@ -0,0 +1 @@
+Added the ability to perform a shallow copy of a GitRepository instance and optionally checkout a different branch and/or a specific commit hash.
```
HanlinMiao, line 1:
```suggestion
Added the ability to perform a shallow copy of a GitRepository instance and to optionally checkout a different branch and/or a specific commit hash.
```

nautobot/extras/models/datasources.py
```diff
@@ -167,3 +167,56 @@ def sync(self, user, dry_run=False):
...
+        elif branch is not None:
+            cloned_git_repo.branch = branch
+            if head is not None:
+                cloned_git_repo.current_head = head
+        else:
+            cloned_git_repo.save()
+            cloned_git_repo.delete()
+            raise ValueError("You must provide a branch to checkout a specific commit hash.")
+
+        cloned_git_repo.save()
```
glennmatthews, line 211: I'm not sure I see the need for creating a new GitRepository DB record, can't it all be done via appropriate params to the GitRepo helper class?

nautobot/extras/models/datasources.py
glennmatthews: Looks like right now you're having it clone into GIT_ROOT, rather than /tmp. Per the comment on #6487, I think we'd decided we wanted the caller to be able to specify which directory to clone to?

nautobot/extras/models/datasources.py
```diff
@@ -167,3 +173,51 @@ def sync(self, user, dry_run=False):
         if dry_run:
             return enqueue_git_repository_diff_origin_and_local(self, user)
         return enqueue_pull_git_repository_and_refresh_data(self, user)
+
+    def clone(self, path=None, branch=None, head=None):
```
glennmatthews, line 177: Were we going to provide an option to use this as a context manager? I.e.:

```py
with my_repo.clone(head="...") as directory:
   # ... operate on files in directory
# directory is automatically deleted on block end
```

nautobot/extras/models/datasources.py
```diff
@@ -167,3 +173,51 @@ def sync(self, user, dry_run=False):
...
+
+        if not branch:
+            branch = self.branch
+
+        try:
+            repo_helper = GitRepo(absolute_path_name, self.remote_url)
+            head, _ = repo_helper.checkout(branch, head)
+        except Exception as e:
+            # Log the exception and raise
+            shutil.rmtree(absolute_path_name)
```
glennmatthews, line 200: Probably should try/except on this too?

nautobot/extras/tests/test_models.py
```diff
@@ -1072,6 +1073,36 @@ def test_remote_url_hostname(self):
         self.repo.remote_url = "http://some-private-host/example.git"
         self.repo.validated_save()
 
+    def test_clone_and_discard_clone_functions(self):
+        """Confirm that the clone() and discard_clone() methods work as expected."""
+        self.repo.remote_url = "https://github.com/nautobot/nautobot.git"
```
glennmatthews, line 1078: That's gonna be slow :-) Would be better to use the `create_and_populate_git_repository` helper in `nautobot/extras/tests/git_helper.py`

nautobot/extras/models/datasources.py
```diff
@@ -167,3 +173,51 @@ def sync(self, user, dry_run=False):
         if dry_run:
             return enqueue_git_repository_diff_origin_and_local(self, user)
         return enqueue_pull_git_repository_and_refresh_data(self, user)
+
+    def clone(self, path=None, branch=None, head=None):
+        """
+        Perform a shallow clone of the Git repository in a temporary directory.
+
+        Args:
+            path (str): Optional The absolute path to clone the repository to. If not specified, the directory will be created in the /tmp directory.
```
glennmatthews, line 182:
```suggestion
            path (str, optional): The absolute directory path to clone into. If not specified, `tempfile.gettempdir()` will be used.
```

nautobot/extras/tests/test_models.py
```diff
@@ -1072,6 +1073,36 @@ def test_remote_url_hostname(self):
...
+
+        with self.subTest("Clone a repository with no path argument provided"):
+            path = self.repo.clone()
+            # assert that the temporary directory was created in the expected location i.e. /tmp/
+            default_dir = "tmp"
+            self.assertIn(default_dir, path)
```
glennmatthews, line 1084: Feels overly non-specific. We should assert that `path.startswith(tempfile.gettempdir()`.

nautobot/extras/tests/test_models.py
```diff
@@ -1072,6 +1073,36 @@ def test_remote_url_hostname(self):
...
+            self.assertTrue(os.path.exists(path))
+            # assert that the temporary directory was discarded
+            self.assertTrue(self.repo.discard_clone(path))
+            self.assertFalse(os.path.exists(path))
+
+        with self.subTest("Clone a repository with a path argument provided"):
+            specified_path = "/source/nautobot/"
```
glennmatthews, line 1091: This is specific to the docker-compose dev environment, and runs the risk of polluting your local source tree if the test errors out. Would be better to use `tempfile.mkdtemp()` to create a temp directory to use as the `specified_path` and clean it up afterwards.

nautobot/extras/tests/test_models.py
```diff
@@ -1072,6 +1073,36 @@ def test_remote_url_hostname(self):
...
+            path = self.repo.clone(path=specified_path)
+            # assert that the temporary directory was created in the expected location i.e. /tmp/
+            self.assertIn(specified_path, path)
```
glennmatthews, line 1094: Same here, assert that `path.startswith(specified_path)`.

nautobot/extras/models/datasources.py
```diff
@@ -167,3 +173,51 @@ def sync(self, user, dry_run=False):
...
+            branch (str): Optional The branch to checkout. If not set, the GitRepository.branch will be used.
+            head (str): Optional Git commit hash to check out instead of pulling branch latest.
+
+        Returns:
+            Return the absolute path of the cloned repo if clone was successful
+        """
+
+        absolute_path_name = tempfile.mkdtemp(dir=path)
```
glennmatthews, line 190: Probably need a try/except here in case the user tries to create in a path they don't have write access to.

nautobot/extras/tests/test_models.py
```diff
@@ -1072,6 +1073,36 @@ def test_remote_url_hostname(self):
...
+            self.assertTrue(os.path.exists(path))
+            # pass in a path that doesn't exist
+            self.assertFalse(self.repo.discard_clone(path + "/nonexistent"))
+            # pass in a valid path and assert that the temporary directory was discarded
+            self.assertTrue(self.repo.discard_clone(path))
+            self.assertFalse(os.path.exists(path))
+
```
glennmatthews, line 1101: Need some success-path tests with `head` and `branch` - using the `git_helper` should make those easier to implement.

nautobot/extras/models/datasources.py
```diff
@@ -167,3 +173,51 @@ def sync(self, user, dry_run=False):
...
         return enqueue_pull_git_repository_and_refresh_data(self, user)
+
+    def clone(self, path=None, branch=None, head=None):
+        """
+        Perform a shallow clone of the Git repository in a temporary directory.
+
+        Args:
+            path (str): Optional The absolute path to clone the repository to. If not specified, the directory will be created in the /tmp directory.
+            branch (str): Optional The branch to checkout. If not set, the GitRepository.branch will be used.
+            head (str): Optional Git commit hash to check out instead of pulling branch latest.
```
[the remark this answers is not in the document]
HanlinMiao: It would not be "incorrect" to specify both branch and head right? But I guess in that case we need to handle a branch and head mismatch. I will make it mutually exclusive.

gsnider2195:
Is it ok if you give me some time to review this before merging? We opened the FR issue as it was blocking another feature that the apps team is working on and I want to make sure this PR addresses all of the requirements.

nautobot/extras/models/datasources.py
```diff
@@ -167,3 +174,55 @@ def sync(self, user, dry_run=False):
...
+    @contextmanager
+    def clone(self, path=None, branch=None, head=None, shallow=True):
+        """
+        Perform a (shallow) clone of the Git repository in a temporary directory.
+
+        Args:
+            path (str, optional): The absolute directory path to clone into. If not specified, `tempfile.gettempdir()` will be used.
+            branch (str, optional): The branch to checkout. If not set, the GitRepository.branch will be used.
+            head (str, optional): Git commit hash to check out instead of pulling branch latest.
+            shallow (bool, optional): If set, perform a shallow clone.
```
gsnider2195, line 187:
```suggestion
            shallow (bool, optional): If True, perform a shallow clone. Defaults to True.
```

nautobot/extras/models/datasources.py
```diff
@@ -167,3 +174,55 @@ def sync(self, user, dry_run=False):
...
+
+            logger.info(f"Cloned repository {self.name} to {path_name}")
+            yield path_name
+        finally:
+            # Cleanup the temporary directory if the clone fails
+            try:
+                shutil.rmtree(path_name)
+            except OSError as os_error:
+                # log error if the cleanup fails
+                logger.error(f"Failed to cleanup temporary directory at {path_name}: {os_error}")
```
gsnider2195, line 228: There's going to be a use case for an app calling `GitRepository().clone()` manually and managing the directory cleanup themself. In its current form the cloned repo is cleaned up after the current web request is finished. Golden config will likely have a use case where we need to clone a copy of the repo for a specific user to do some work against and leave it on the filesystem for multiple requests. Then when the user is done (or an arbitrary timeout passes or the user's session times out, etc) we clean up the cloned repo. For that use case I think it makes sense to create a separate context manager and move this `finally` logic there and change this `clone` method to a regular instance method.

nautobot/extras/models/datasources.py
```diff
@@ -167,3 +174,55 @@ def sync(self, user, dry_run=False):
...
+        Perform a (shallow) clone of the Git repository in a temporary directory.
+
+        Args:
+            path (str, optional): The absolute directory path to clone into. If not specified, `tempfile.gettempdir()` will be used.
+            branch (str, optional): The branch to checkout. If not set, the GitRepository.branch will be used.
+            head (str, optional): Git commit hash to check out instead of pulling branch latest.
+            shallow (bool, optional): If set, perform a shallow clone.
+
+        Returns:
+            Return the absolute path of the cloned repo if clone was successful
```
gsnider2195, line 190:
```suggestion
            Returns the absolute path of the cloned repo if clone was successful, otherwise returns None.
```

nautobot/extras/models/datasources.py
```diff
@@ -167,3 +174,55 @@ def sync(self, user, dry_run=False):
         if dry_run:
             return enqueue_git_repository_diff_origin_and_local(self, user)
         return enqueue_pull_git_repository_and_refresh_data(self, user)
+
+    @contextmanager
+    def clone(self, path=None, branch=None, head=None, shallow=True):
```
gsnider2195, line 179: Is `clone` too ambiguous of a name? This isn't part of the normal `GitRepository` flow so it should probably have a special name like `clone_to_directory` or something?

nautobot/extras/models/datasources.py
```diff
@@ -167,3 +174,55 @@ def sync(self, user, dry_run=False):
...
+        Returns:
+            Return the absolute path of the cloned repo if clone was successful
+        """
+
+        try:
+            if branch and head:
+                raise ValueError("Cannot specify both branch and head")
+
+            try:
+                path_name = tempfile.mkdtemp(dir=path)
```
gsnider2195, line 198:
```suggestion
                path_name = tempfile.mkdtemp(dir=path, prefix=self.slug)
```

So that the directory name isn't completely random.

nautobot/core/utils/git.py
```diff
@@ -57,7 +57,7 @@ class BranchDoesNotExist(Exception):
 
 
 class GitRepo:
-    def __init__(self, path, url, clone_initially=True):
+    def __init__(self, path, url, clone_initially=True, shallow=False):
```
glennmatthews, line 60: Add new param to docstring please.

nautobot/extras/tests/test_models.py
```diff
@@ -1072,6 +1074,58 @@ def test_remote_url_hostname(self):
         self.repo.remote_url = "http://some-private-host/example.git"
         self.repo.validated_save()
 
+    def test_clone_context_manager(self):
+        """Confirm that the clone() context manager methods work as expected."""
+        self.tempdir = tempfile.TemporaryDirectory()  # pylint: disable=consider-using-with
```
glennmatthews, line 1079: I don't see where you're cleaning up the tempdir when done - might be better to change it to context-manager style for the whole test case so that it cleans up automatically even if the test fails?

nautobot/extras/tests/test_models.py
```diff
@@ -1072,6 +1074,58 @@ def test_remote_url_hostname(self):
...
+        self.repo = GitRepository(
+            name="New Git Repository",
+            slug=self.repo_slug,
+            remote_url="file://" + self.tempdir.name,  # file:// URLs aren't permitted normally, but very useful here!
+            branch="main",
+            # Provide everything we know we can provide
+            provided_contents=[entry.content_identifier for entry in get_datasource_contents("extras.gitrepository")],
+        )
+        self.repo.save()
+        specified_path = tempfile.mkdtemp()
```
glennmatthews, line 1092: I don't see where you're cleaning up this directory either.

nautobot/extras/models/datasources.py
```diff
@@ -167,3 +174,55 @@ def sync(self, user, dry_run=False):
...
+            try:
+                repo_helper = GitRepo(path_name, self.remote_url, shallow=shallow)
+                head, _ = repo_helper.checkout(branch, head)
+            except Exception as e:
+                # Cleanup the temporary directory if the clone fails
+                try:
+                    shutil.rmtree(path_name)
+                except OSError as os_error:
+                    # log error if the cleanup fails
+                    logger.error(f"Failed to cleanup temporary directory at {path_name}: {os_error}")
```
glennmatthews, line 215: Isn't this `except` redundant with the `finally` block below?

nautobot/extras/tests/test_models.py
```diff
@@ -1072,6 +1074,58 @@ def test_remote_url_hostname(self):
...
+        with self.subTest("Clone a repository with no path argument provided"):
+            with self.repo.clone_to_directory() as path:
+                # assert that the temporary directory was created in the expected location i.e. /tmp/
+                self.assertTrue(path.startswith(tempfile.gettempdir()))
+                self.assertTrue(os.path.exists(path))
```
glennmatthews, line 1097: Would be good to add a spot-check or two to assert that the files in the repo are present in `path` and that the correct HEAD is checked out.

nautobot/core/utils/git.py
```diff
@@ -73,7 +73,10 @@ def __init__(self, path, url, clone_initially=True):
         elif clone_initially:
             # Don't log `url` as it may include authentication details.
             logger.debug("Cloning git repository to %s...", path)
-            self.repo = Repo.clone_from(url, to_path=path, env=GIT_ENVIRONMENT)
+            if not shallow:
+                self.repo = Repo.clone_from(url, to_path=path, env=GIT_ENVIRONMENT)
+            else:
+                self.repo = Repo.clone_from(url, to_path=path, env=GIT_ENVIRONMENT, depth=1)
```
glennmatthews, line 79: Wonder if it'd be any cleaner from an API perspective to implement a `depth` parameter instead of a `shallow` parameter? Alternatively:

```suggestion
            kwargs = {}
            if shallow:
                kwargs["depth"] = 1
            self.repo = Repo.clone_from(url, to_path=path, env=GIT_ENVIRONMENT, **kwargs)
```

nautobot/extras/models/datasources.py
```diff
@@ -167,3 +174,55 @@ def sync(self, user, dry_run=False):
...
+            except PermissionError as e:
+                logger.error(f"Failed to create temporary directory at {path}: {e}")
+                raise e
+
+            if not branch:
+                branch = self.branch
+
+            try:
+                repo_helper = GitRepo(path_name, self.remote_url, shallow=shallow)
+                head, _ = repo_helper.checkout(branch, head)
```
glennmatthews, line 208: https://github.blog/open-source/git/get-up-to-speed-with-partial-clone-and-shallow-clone/ seems like a useful reference. 

If I'm understanding it right, we should *not* generally support `clone_to_directory(head=...., shallow=True)` since by definition `shallow=True` implies "latest commits only".

Additionally, we should probably enhance `GitRepo.__init__()` to also take `branch` as an optional parameter so that we can do `git clone ... --depth=1 --branch=<branch>` rather than cloning the default branch and then trying to switch to the specified branch.

nautobot/extras/tests/test_models.py
```diff
@@ -1072,6 +1074,58 @@ def test_remote_url_hostname(self):
...
+        with self.subTest("Clone a repository with a path argument provided"):
+            with self.repo.clone_to_directory(path=specified_path) as path:
+                # assert that the temporary directory was created in the expected location i.e. /tmp/
+                self.assertTrue(path.startswith(specified_path))
+                self.assertTrue(os.path.exists(path))
+            # Temp directory is cleaned up after the context manager exits
+            self.assertFalse(os.path.exists(path))
+
+        with self.subTest("Clone a repository with the branch argument provided"):
+            with self.repo.clone_to_directory(path=specified_path, branch="main") as path:
```
glennmatthews, line 1109: Need to test with a non-default branch as well

nautobot/core/utils/git.py
HanlinMiao: Implementing a `depth` parameter instead of `shallow` makes sense

nautobot/extras/tests/test_models.py
```diff
@@ -1074,57 +1075,179 @@ def test_remote_url_hostname(self):
...
+                    self.assertTrue(os.path.exists(path + "/config_contexts/badcontext2.json"))
+                # Temp directory is cleaned up after the context manager exits
+                self.assertFalse(os.path.exists(path))
+
+            with self.subTest("Assert an Exception is raised when an invalid commit hash is provided"):
+                with self.assertRaises(Exception):
+                    with self.repo.clone_to_directory_context(path=specified_path, head="non-existent") as path:
+                        pass
+        finally:
+            shutil.rmtree(specified_path, ignore_errors=True)
```
glennmatthews, line 1162: Possible error case here if the test bombs out before defining `specified_path`?

nautobot/core/utils/git.py
```diff
@@ -73,10 +75,10 @@ def __init__(self, path, url, clone_initially=True, shallow=False):
         elif clone_initially:
             # Don't log `url` as it may include authentication details.
             logger.debug("Cloning git repository to %s...", path)
-            if not shallow:
-                self.repo = Repo.clone_from(url, to_path=path, env=GIT_ENVIRONMENT)
+            if not depth:
+                self.repo = Repo.clone_from(url, to_path=path, env=GIT_ENVIRONMENT, branch=branch)
             else:
-                self.repo = Repo.clone_from(url, to_path=path, env=GIT_ENVIRONMENT, depth=1)
+                self.repo = Repo.clone_from(url, to_path=path, env=GIT_ENVIRONMENT, branch=branch, depth=1)
```
glennmatthews, line 81:
```suggestion
                self.repo = Repo.clone_from(url, to_path=path, env=GIT_ENVIRONMENT, branch=branch, depth=depth)
```

nautobot/extras/tests/test_models.py
HanlinMiao: I am going to define `specified_path` and `self.tempdir` at the very beginning of the `try` block. So the error case shouldn't happen

nautobot/extras/tests/test_models.py
```diff
@@ -1072,6 +1075,179 @@ def test_remote_url_hostname(self):
...
+                with self.repo.clone_to_directory_context(path=specified_path, depth=1) as path:
+                    # assert that the temporary directory was created in the expected location i.e. /tmp/
+                    self.assertTrue(path.startswith(specified_path))
+                    self.assertTrue(os.path.exists(path))
+                    self.assertTrue(os.path.exists(path + "/config_context_schemas/badschema1.json"))
+                    self.assertTrue(os.path.exists(path + "/config_contexts/badcontext2.json"))
+                # Temp directory is cleaned up after the context manager exits
+                self.assertFalse(os.path.exists(path))
+
+            with self.subTest("Assert an Exception is raised when an invalid commit hash is provided"):
```
gsnider2195, line 1157: Add a test case for `branch` and `head` supplied at the same time? I suspect this will fail when you try to access `path_name` before it's been defined:

```py
        finally:
            # Cleanup the temporary directory if the clone fails
            try:
                shutil.rmtree(path_name)
```

nautobot/extras/models/datasources.py
```diff
@@ -167,3 +174,98 @@ def sync(self, user, dry_run=False):
...
+            head (str, optional): Git commit hash to check out instead of pulling branch latest.
+            depth (int, optional): The depth of the clone. If set to 0, a full clone will be performed.
+
+        Returns:
+            Returns the absolute path of the cloned repo if clone was successful, otherwise returns None.
+        """
+
+        try:
+            if branch and head:
+                raise ValueError("Cannot specify both branch and head")
```
gsnider2195, line 195: This should probably be moved outside of the `try/finally`?

nautobot/extras/models/datasources.py
```diff
@@ -167,3 +174,98 @@ def sync(self, user, dry_run=False):
...
+                repo_helper = GitRepo(path_name, self.remote_url, depth=depth)
+                head, _ = repo_helper.checkout(branch, head)
+            except Exception as e:
+                logger.error(f"Failed to clone repository {self.name} to {path_name}: {e}")
+                raise e
+
+            logger.info(f"Cloned repository {self.name} to {path_name}")
+            yield path_name
+        finally:
+            # Cleanup the temporary directory if the clone fails
```
gsnider2195, line 216:
```suggestion
            # Cleanup the temporary directory
```

nautobot/extras/models/datasources.py
```diff
@@ -167,3 +174,98 @@ def sync(self, user, dry_run=False):
...
+            repo_helper = GitRepo(path_name, self.remote_url, depth=depth)
+            head, _ = repo_helper.checkout(branch, head)
+        except Exception as e:
+            logger.error(f"Failed to clone repository {self.name} to {path_name}: {e}")
+            raise e
+
+        logger.info(f"Cloned repository {self.name} to {path_name}")
+        return path_name
+
+    def cleanup_cloned_directory(self, path):
```
gsnider2195, line 259: Currently not being used anywhere?

nautobot/extras/models/datasources.py
```diff
@@ -167,3 +174,78 @@ def sync(self, user, dry_run=False):
...
+            path_name = tempfile.mkdtemp(dir=path, prefix=self.slug)
+        except PermissionError as e:
+            logger.error(f"Failed to create temporary directory at {path}: {e}")
+            raise e
+
+        if not branch:
+            branch = self.branch
+
+        try:
+            repo_helper = GitRepo(path_name, self.remote_url, depth=depth)
```
glennmatthews, line 230: Not passing `branch` as an input here?

nautobot/extras/tests/test_models.py
```diff
@@ -1072,6 +1075,187 @@ def test_remote_url_hostname(self):
...
+                    # assert that the temporary directory was created in the expected location i.e. /tmp/
+                    self.assertTrue(path.startswith(specified_path))
+                    self.assertTrue(os.path.exists(path))
+                    self.assertTrue(os.path.exists(path + "/config_context_schemas/schema-1.yaml"))
+                    self.assertTrue(os.path.exists(path + "/config_contexts/context.yaml"))
+                # Temp directory is cleaned up after the context manager exits
+                self.assertFalse(os.path.exists(path))
+
+            with self.subTest("Clone a repository with depth argument provided"):
+                with self.repo.clone_to_directory_context(path=specified_path, depth=1) as path:
```
glennmatthews, line 1148: Also need tests for `depth=1` in combination with `branch` or `head`?

nautobot/extras/models/datasources.py
HanlinMiao: I might be wrong here but I thought we were checking out the branch on the next line.

glennmatthews: See my earlier comment - https://github.com/nautobot/nautobot/pull/6746#discussion_r1914848390 - I *think* it's more correct to clone the desired branch directly rather than cloning the default branch and then trying to switch to the desired branch.

nautobot/extras/tests/test_models.py
```diff
@@ -1072,6 +1075,240 @@ def test_remote_url_hostname(self):
...
+                    self.assertTrue(os.path.exists(path + "/config_contexts/badcontext2.json"))
+                # Temp directory is cleaned up after the context manager exits
+                self.assertFalse(os.path.exists(path))
+
+            with self.subTest("Assert an Exception is raised when an invalid commit hash is provided"):
+                with self.assertRaises(Exception):
+                    with self.repo.clone_to_directory_context(path=specified_path, head="non-existent") as path:
+                        pass
+            with self.subTest("Assert a value error is raised when branch and head are both provided"):
+                with self.assertRaises(ValueError):
```
gsnider2195, line 1188: `assertRaises` for generic exceptions like `Exception` and `ValueError` is too broad. Can we switch to `assertRaisesRegex` instead so that we know the `ValueError` that's being raised is the one we expected?

nautobot/extras/models/datasources.py
```diff
@@ -167,3 +174,79 @@ def sync(self, user, dry_run=False):
...
+        except PermissionError as e:
+            logger.error(f"Failed to create temporary directory at {path}: {e}")
+            raise e
+
+        if not branch:
+            branch = self.branch
+
+        try:
+            repo_helper = GitRepo(path_name, self.remote_url, depth=depth, branch=branch)
+            head, _ = repo_helper.checkout(branch, head)
```
glennmatthews, line 232: Shouldn't be needed if we're already cloning the specified branch?

nautobot/extras/tests/test_models.py
```diff
@@ -1072,6 +1075,240 @@ def test_remote_url_hostname(self):
...
+                self.assertFalse(os.path.exists(path))
+
+            with self.subTest("Clone a shallow repository with depth and invalid head arguments provided"):
+                with self.assertRaises(Exception):
+                    # Shallow copy a repo should only have the latest commit
+                    with self.repo.clone_to_directory_context(path=specified_path, depth=1, head="valid-files") as path:
+                        pass
+
+            with self.subTest("Clone a shallow repository with depth and valid branch arguments provided"):
+                with self.repo.clone_to_directory_context(path=specified_path, depth=1, branch="main") as path:
```
glennmatthews, line 1174: Should test with a non-default branch if possible. May require some enhancement of `create_and_populate_git_repository` to create a divergent branch?

nautobot/extras/tests/test_models.py
```diff
@@ -1072,6 +1075,240 @@ def test_remote_url_hostname(self):
...
+                    # assert that the temporary directory was created in the expected location i.e. /tmp/
+                    self.assertTrue(path.startswith(specified_path))
+                    self.assertTrue(os.path.exists(path))
+                    self.assertTrue(os.path.exists(path + "/config_context_schemas/badschema1.json"))
+                    self.assertTrue(os.path.exists(path + "/config_contexts/badcontext2.json"))
+                # Temp directory is cleaned up after the context manager exits
+                self.assertFalse(os.path.exists(path))
+
+            with self.subTest("Clone a shallow repository with depth and invalid head arguments provided"):
+                with self.assertRaises(Exception):
```
glennmatthews, line 1168: Same as Gary's comment below - too broad/generic.

nautobot/extras/models/datasources.py
HanlinMiao: Is this what you mean?
```
repo_helper = GitRepo(path_name, self.remote_url, depth=depth, branch=branch)
if head:
     head, _ = repo_helper.checkout(branch, head)
```

glennmatthews: think so, yeah. Thanks.

nautobot/extras/tests/git_helper.py
```diff
@@ -69,6 +69,12 @@ def create_and_populate_git_repository(target_path):
         # Directory "01-valid-files" --> tag "valid-files" so that we won't break tests if we renumber the directories
         repo.create_tag(dirname[3:], message=f"Tag based on {dirname} files")
 
+    if divergent_branch:
+        repo.create_head(divergent_branch)
+        repo.head.reference = repo.heads[divergent_branch]
+        repo.head.reset(index=True, working_tree=True)
+        repo.create_tag(f"{divergent_branch}", message=f"Tag for divergent branch {divergent_branch}")
```
glennmatthews, line 76: Not sure I entirely understand what this block is actually doing, but is the resulting divergent branch any way different in content from the main branch (i.e., at least a different commit hash)?

glennmatthews (approved):

PR #6749, opened 2025-01-09
Reintroduce indentation of child prefixes and improve generation of "available" prefixes in Prefix detail view

### LINKED ISSUE #6745: Hierarchical view of IP Prefixes broken in 2.3 - Child Prefixes no longer show hierarchical indentation
### Environment
* Nautobot version: 2.3.1+ (specifically tested 2.3.14)
* Python version: 3.12
* Database platform, version: Postgres 16
* Middleware(s):

After changes in #6072 which attempted to fix indentation issues in #5970, the result is worse than it was before.  

The child prefixes are now never showing indentation. This breaks the entire hierarchical view of networks contained inside a container.
There is now no way to view an IP prefix hierarchy at all in the program from what I can see.  I believe a hierarchical view of IP prefixes is a core feature of any IPAM.  

I understand this might be more difficult to solve in the main prefix view `/ipam/prefixes` when there is sorting or a filter applied.

However in this instance under "child prefix", those complex issues should not apply and it should not be a problem to show the hierarchical view of all the child prefixes under this prefix.

I think maybe this is handled as two issues 
 1. Bug for the child prefix view `/ipam/prefixes/xxxxx/prefixes` since that should be straightforward to fix
 2. Feature request for prefixes screen `/ipam/prefixes` to show indentation when no filter is applied to restore that functionality.

This issue is opened primarily to fix number 1 above - Child Prefix view no longer shows indentation.  I can open a separate issue for number 2 above if that is deemed the right approach.

I'm opening this as a new issue as requested in #6072 comments. 

### Steps to Reproduce
1. Create a prefix (parent prefix)
3. Create a child prefix underneath parent prefix
4. Create a child prefix underneath 1st child prefix (Parent -> Child 1 -> Child 2)
5. Navigate to parent prefix
6. Click on Child Prefixes tab
7. The child prefixes should be shown in a hierarchy

e.g. view https://demo.nautobot.com/ipam/prefixes/39a951f0-a96a-4ef3-9283-13cb00fb3018/prefixes/ 

### Expected Behavior
The child prefixes should be shown in a hierarchy similar to this:

```
10.0.0.0/18
  * 10.0.0.0/23
  *  * 10.0.1.0/28
```

### Observed Behavior

The child prefixes are shown with no hierarchy information as a flat list.  e.g.

```
10.0.0.0/18
10.0.0.0/23
10.0.1.0/28
```

glennmatthews:
# Closes #6745 
# What's Changed

- Remove `hide_hierarchy_ui=True` mistakenly added to Child Prefixes table in #6072.
- Add `base_depth` optional parameter to Prefix table rendering so that a deeply nested prefix's child prefixes can show indentation relative to the base prefix instead of their global nesting.
- Fix `add_available_prefixes` function logic so that "available" prefixes are assigned to the correct namespace and therefore get the correct nesting depth
- Enhance `add_available_prefixes` to add prefix space available under nested descendant Container prefixes rather than only space directly under this prefix itself.

# Screenshots

## Before #6072 

image

image

## Current `develop`

image

image

## After this set of changes

image

image

@cdwchriburg If you have a chance to pull this down and try it out (note that it's based on latest `develop`, which will be the 2.4.0 release shortly, so you'll want to use a test environment and run migrations) I'd really appreciate it!

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/ipam/views.py
+++ b/nautobot/ipam/views.py
@@ -437,7 +437,6 @@ class PrefixPrefixesView(generic.ObjectView):
 
         prefix_table = tables.PrefixDetailTable(
             child_prefixes,
-            hide_hierarchy_ui=True,
             exclude=["namespace"],
             data_transform_callback=data_transform_callback,
         )
@@ -461,6 +460,7 @@ class PrefixPrefixesView(generic.ObjectView):
 
         return {
             "first_available_prefix": instance.get_first_available_prefix(),
+            "base_tree_depth": instance.ancestors().count(),
             "prefix_table": prefix_table,
             "permissions": permissions,
             "bulk_querystring": bulk_querystring,
--- a/changes/6745.fixed
+++ b/changes/6745.fixed
@@ -1,0 +1,2 @@
+Restored indentation of child prefixes in Prefix detail view.
+Improved logic for generation of "available" child prefixes in Prefix detail view.
--- a/nautobot/core/templatetags/helpers.py
+++ b/nautobot/core/templatetags/helpers.py
@@ -1050,12 +1050,13 @@ def versioned_static(file_path):
 
 
 @register.simple_tag
-def tree_hierarchy_ui_representation(tree_depth, hide_hierarchy_ui):
+def tree_hierarchy_ui_representation(tree_depth, hide_hierarchy_ui, base_depth=0):
     """Generates a visual representation of a tree record hierarchy using dots.
 
     Args:
         tree_depth (range): A range representing the depth of the tree nodes.
         hide_hierarchy_ui (bool): Indicates whether to hide the hierarchy UI.
+        base_depth (int, optional): Starting depth (number of dots to skip rendering).
 
     Returns:
         str: A string containing dots (representing hierarchy levels) if `hide_hierarchy_ui` is False,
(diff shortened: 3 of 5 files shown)


HanlinMiao (approved):

cdwchriburg:
Thanks, I will setup a test box based on this branch and give it a whirl.

nautobot/ipam/views.py
```diff
@@ -461,6 +460,7 @@ def get_extra_context(self, request, instance):
 
         return {
             "first_available_prefix": instance.get_first_available_prefix(),
+            "base_tree_depth": instance.ancestors().count(),
```
cdwchriburg, line 463: Indent starting at 0 indenting? i.e. 

`"base_tree_depth": instance.ancestors().count() - 1`

cdwchriburg:
Got test system setup on this branch with data mirrored from prod.  I do see child prefixes are now indented again.  First question I have on that is should the first prefix be at indent level 0 rather than indent level 1?

i.e. `"base_tree_depth": instance.ancestors().count() - 1` 

So instead of:
image

It would look like:
image

Otherwise every child prefix is indented even if its just a flat structure underneath the parent prefix.


I think there are some other pages to review as well.  I'll check out a few more locations and update any findings here.

glennmatthews:
Yeah, I went back and forth on that one. My thinking was that the indentation of the child prefixes serves as a useful reminder that they *are* child prefixes at all, which might be less obvious if we decrease it by one.

Feel free to give feedback on other pages but this PR is pretty narrowly targeted to the Prefix - Child Prefixes view at this time. :-)

cdwchriburg:
Ok gotcha sorry - I'll post those other items in issues and not this PR. 

I was also going back and forth on the 1st level indentation thing.  I think with lots of hierarchy that can make sense.  

Just looks maybe a little weird with only one level of child prefixes.  
image


But both options are solving the original issue either way is fine.

For Prefix-Child Prefix view I think this looks good to me.  Fixes the original issue with root level indenting and the accidental removal of indenting altogether on the child prefix page.

glennmatthews:
Thanks for the quick feedback @cdwchriburg!

PR #6766, opened 2025-01-13
Add default admin and read only roles

### LINKED ISSUE #6765: Add Nautobot default Permissions
### As ...

Patti - Platform Admin

### I want ...

to have pre-configured basic roles for read only and read/write

### So that ...

I can give immediate access to data to users that do not have the right to edit data

I find a read/write and a read only pre-cofnigured role after install

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

no change

### External Dependencies

none

icovada:
# Closes #6765
# What's Changed
Added two default Permission roles and relative Groups for read/write and read only access to all content types
# Screenshots
Not relevant
# TODO
- [X] Unit, Integration Tests
- [] Outline Remaining Work, Constraints from Design

### DIFF
--- a/nautobot/users/migrations/0011_create_default_groups_and_permissions.py
+++ b/nautobot/users/migrations/0011_create_default_groups_and_permissions.py
@@ -1,0 +1,51 @@
+from django.contrib.auth.models import Group, Permission
+from django.contrib.contenttypes.models import ContentType
+from django.db import migrations
+
+from nautobot.users.models import ObjectPermission
+
+
+def add_default_groups(apps, schema_editor):
+    read_only, _ = Group.objects.get_or_create(name="Read Only")
+    admin, _ = Group.objects.get_or_create(name="Admins")
+
+    ro_obj_permission, created = ObjectPermission.objects.get_or_create(
+        name="Default: Global Read Only",
+        actions=["view"],
+    )
+
+    if not created:
+        # Only exit if it didn't exist before
+        return
+
+    rw_obj_permission, created = ObjectPermission.objects.get_or_create(
+        name="Default: Global Read/Write",
+        actions=["read", "view", "modify", "delete"],
+    )
+
+    if not created:
+        return
+
+    ro_obj_permission.groups.add(read_only)
+    rw_obj_permission.groups.add(admin)
+
+    # Get all ContentTypes and add them to Permissions
+    for x in ContentType.objects.all():
+        ro_obj_permission.object_types.add(x)
+        rw_obj_permission.object_types.add(x)
+
+
+def remove_default_groups(apps, schema_editor):
+    read_only, _ = Group.objects.get_or_create(name="Default: Global Read Only")
+    admin, _ = Group.objects.get_or_create(name="Default: Global Read/Write")
+
+    read_only.delete()
+    admin.delete()
+
+
+class Migration(migrations.Migration):
+    dependencies = [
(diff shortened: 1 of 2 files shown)


nautobot/users/migrations/0011_create_default_groups_and_permissions.py
```diff
@@ -0,0 +1,51 @@
...
+        rw_obj_permission.object_types.add(x)
+
+
+def remove_default_groups(apps, schema_editor):
+    read_only, _ = Group.objects.get_or_create(name="Default: Global Read Only")
+    admin, _ = Group.objects.get_or_create(name="Default: Global Read/Write")
+
+    read_only.delete()
+    admin.delete()
+
```
HanlinMiao, line 44: We should probably remove the `ObjectPermission` instances created as well.

icovada: Good catch!

nautobot/users/migrations/0011_create_default_groups_and_permissions.py
```diff
@@ -0,0 +1,51 @@
...
+    admin, _ = Group.objects.get_or_create(name="Admins")
+
+    ro_obj_permission, created = ObjectPermission.objects.get_or_create(
+        name="Default: Global Read Only",
+        actions=["view"],
+    )
+
+    if not created:
+        # Only exit if it didn't exist before
+        return
```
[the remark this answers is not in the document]
icovada: Yeah I'm checking it a couple lines later after I get_or_create the rw permission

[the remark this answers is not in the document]
icovada: You're right, this way we fail if either of those objects exist, but that's what I was going for.
In my idea, should any of those objects already exist we shouldn't add any of the default ones, because then one would be standard and the other wouldn't be.

Let me know if you agree with my approach

nautobot/users/migrations/0011_create_default_groups_and_permissions.py
```diff
@@ -0,0 +1,70 @@
...
+    if not created:
+        return
+
+    ro_obj_permission.groups.add(read_only)
+    rw_obj_permission.groups.add(admin)
+
+    # Get all ContentTypes and add them to Permissions
+    for x in ContentType.objects.all():
+        ro_obj_permission.object_types.add(x)
+        rw_obj_permission.object_types.add(x)
```
glennmatthews, line 36: This is going to be a one-off migration, so any models added in later migrations will *not* be included in the constructed permissions. Is that an acceptable limitation or should a different approach be developed?

icovada: A good discussion point indeed!

Are we looking for install-time permission setting, or is this meant to be an immutable standard allowing the right access to all object types that will ever be?

If we're going for the second route we should also find a way to make these permissions immutable.

HanlinMiao:
Thanks for the proposal, we definitely agree with the use case but we are a bit concerned with the maintainability of this approach as new models are created. So we decided as a team to close this for now and investigate it more in the future. Feel free to add more context and/or justification to the issue.

PR #6775, opened 2025-01-14
Fixed the issue where the Wireless Network detail view

### LINKED ISSUE #6763: Wireless Network tab fails to render under Controller View
### Environment
* Nautobot version (Docker tag too if applicable): 2.4.0
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Create and populate wireless network data (including network, prefix and a vlan associated with a network)
2. Navigate to Controller View -> Wireless Network

### Observed Behavior
Error occurs:
```
<class 'TypeError'>

Cannot encode None for key 'vlan_id' in a query string. Did you mean to pass an empty string or omit the value?

Python version: 3.10.16
Nautobot version: 2.4.0
```

timizuoebideri1:
# Closes #6763 
# What's Changed

Attempting to `urlencode` a `None` value results in an error, so I replaced the `None` value with an empty string `""`.

# Screenshots

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/core/tables.py
+++ b/nautobot/core/tables.py
@@ -1,6 +1,7 @@
 import contextlib
 import logging
 
+from django.conf import settings
 from django.contrib.auth.models import AnonymousUser
 from django.contrib.contenttypes.fields import GenericForeignKey
 from django.core.exceptions import FieldDoesNotExist, FieldError
@@ -557,7 +558,10 @@ class LinkedCountColumn(django_tables2.Column):
                 related_record = related_records[0]
         url = reverse(self.viewname, kwargs=self.view_kwargs)
         if self.url_params:
-            url += "?" + urlencode({k: (getattr(record, v) or "") for k, v in self.url_params.items()})
+            url += "?" + urlencode(
+                # Replace None values with `FILTERS_NULL_CHOICE_VALUE` to handle URL parameters correctly
+                {k: (getattr(record, v) or settings.FILTERS_NULL_CHOICE_VALUE) for k, v in self.url_params.items()}
+            )
         if value > 1:
             return format_html('<a href="{}" class="badge">{}</a>', url, value)
         if related_record is not None:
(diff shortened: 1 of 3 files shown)


nautobot/wireless/tests/test_views.py
```diff
@@ -72,6 +73,26 @@ def setUpTestData(cls):
             "description": "New description",
         }
 
+    @override_settings(EXEMPT_VIEW_PERMISSIONS=[])
+    def test_get_with_controled_manged_device_group_vlan_set_to_none(self):
+        """Assert bug report #6763 (Wireless Network tab fails to render under Controller View) has been resolved"""
+        instance = self.wirless_network
```
HanlinMiao, line 79:
```suggestion
        instance = self.wireless_network
```

nautobot/core/tables.py
```diff
@@ -557,7 +557,7 @@ def render(self, *, bound_column, record, value):  # pylint: disable=arguments-d
                 related_record = related_records[0]
         url = reverse(self.viewname, kwargs=self.view_kwargs)
         if self.url_params:
-            url += "?" + urlencode({k: getattr(record, v) for k, v in self.url_params.items()})
+            url += "?" + urlencode({k: (getattr(record, v) or "") for k, v in self.url_params.items()})
```
HanlinMiao, line 560:
```suggestion
            url += "?" + urlencode({k: (getattr(record, v, "")) for k, v in self.url_params.items()})
```

glennmatthews, line 560: This won't have the expected UX in this specific case, will it? It'll lead to an unfiltered list of "all" records, which doesn't seem like it'd be that useful as a list of "related" Prefixes to a given Wireless Network.

timizuoebideri1: This would not work in this case as `record` has the attribute `v` but the attributes value is `None`, this would only work when `record` do not have the attribute `v`

glennmatthews (approved):
Seems fine for now. Initial feedback from Architecture appears to be that we can potentially update the data model to make `vlan` non-nullable here, but the change to tables.py looks useful regardless.

PR #6789, opened 2025-01-15
Export rack visualization to csv (third time's the charm)

### LINKED ISSUE #4141: Ability to export rack visualization to text (excel)
### As ...

Dan - Data Owner

### I want ...

to export rack visualization to text format (excel). Current SVG export is nice (for printing) but we don't find it that useful :-)

### So that ...

I can draft changes / manipulation / re-arrange / re-design racks before actual data updates in Nautobot.

- I can export a rack visualization to text format, incl. empty slots, names etc.

- [ ] Automation
- [ ] Circuits
- [X] DCIM
- [ ] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [X] UI/UX
- [ ] Documentation

### Database Changes

_No response_

### External Dependencies

_No response_

icovada:
# Closes #4141 
# What's Changed

New "Save CSV" file under the graphical representation of racks to get a quasi-graphical representation of a rack to be used in spreadsheet editors.

Output is as follows:

```csv
unit,name
FrontU1,coresw-01
FrontU2,
RearU1,coresw-01 (Rear)
RearU2,
```

# Screenshots

# TODO

- [X] Unit, Integration Tests


nautobot/dcim/models/racks.py
```diff
@@ -363,8 +365,9 @@ def get_reserved_units(self):
                 reserved_units[u] = r
         return reserved_units
 
-    def get_elevation_svg(
+    def get_elevation(
```
glennmatthews, line 368: Renaming this is arguably a breaking API change - can you add back in a `get_elevation_csv` method that has the same args/kwargs and wraps this "new" function?

nautobot/dcim/elevations.py
```diff
@@ -9,9 +13,9 @@
 from .constants import RACK_ELEVATION_BORDER_WIDTH
 
 
-class RackElevationSVG:
+class RackElevationGraphicalOutput:
```
glennmatthews, line 16: This isn't a "public" API (not exposed in `nautobot.apps` namespace) but I still feel a bit nervous renaming it without providing some sort of backwards-compatibility shim.

nautobot/dcim/elevations.py
```diff
@@ -217,7 +221,93 @@ def merge_elevations(self, face):
 
         return elevation
 
-    def render(self, face, unit_width, unit_height, legend_width):
+    def render(
+        self,
+        fileformat: Literal["svg", "csv"],
```
glennmatthews, line 226: Similarly, can we preserve the existing API signature by making `fileformat` an optional kwarg (defaulting to `"svg"`) instead of a new mandatory arg?

icovada:
I have re-created RackElevationSVG as subclass of RackElevationGraphicalOutput with a single function render() calling the svg format only

icovada:
Before I fix tests, let me know if this is a good direction to keep working on

nautobot/dcim/api/views.py
```diff
@@ -240,6 +227,24 @@ def elevation(self, request, pk=None):
                 rack_units = serializers.RackUnitSerializer(page, many=True, context={"request": request})
                 return self.get_paginated_response(rack_units.data)
 
+        # Render and return the elevation as an SVG drawing with the correct content type
```
glennmatthews, line 230:
```suggestion
        # Render and return the elevation as an SVG drawing or CSV file with the correct content type
```

glennmatthews (commented):
New approach appears to address my concerns, thanks.

nautobot/dcim/elevations.py
```diff
@@ -217,7 +222,78 @@ def merge_elevations(self, face):
 
         return elevation
 
-    def render(self, face, unit_width, unit_height, legend_width):
+    def render_csv(self, face: Literal["front", "rear"]) -> str:
+        @dataclass
+        class DeviceCSVRepresentation:
+            name: Optional[int]
+            unit: str
+            is_full_depth: Optional[int]
```
HanlinMiao, line 230: Perhaps?
```suggestion
            is_full_depth: Optional[bool]
```

nautobot/dcim/elevations.py
```diff
@@ -217,7 +222,78 @@ def merge_elevations(self, face):
...
+                    )
+
+        front_faces = _get_face_data("front")
+        rear_faces = _get_face_data("rear")
+
+        for front, rear in zip(front_faces, rear_faces):
+            if front.is_full_depth:
+                rear.name = f"{front.name} (Rear)"
+            elif rear.is_full_depth:
+                front.name = f"{rear.name} (Rear)"
```
HanlinMiao, line 283: Sorry I am not super familiar or sure what this code block is doing. Can someone double-check me on this?

HanlinMiao, line 283: Should this be
```suggestion
                front.name = f"{rear.name} (Front)"
```

nautobot/dcim/models/racks.py
```diff
@@ -393,12 +397,27 @@ def get_elevation_svg(
...
         if unit_height is None:
-            unit_height = get_settings_or_config("RACK_ELEVATION_DEFAULT_UNIT_HEIGHT", fallback=22)
-        elevation = RackElevationSVG(
+            unit_height = get_settings_or_config("RACK_ELEVATION_DEFAULT_UNIT_HEIGHT")
+        elevation = RackElevationGraphicalOutput(
             self, user=user, include_images=include_images, base_url=base_url, display_fullname=display_fullname
         )
 
-        return elevation.render(face, unit_width, unit_height, legend_width)
+        return elevation.render(fileformat, face, unit_width, unit_height, legend_width)
```
HanlinMiao, line 405: Perhaps?
```suggestion
        return elevation.render_svg(fileformat, face, unit_width, unit_height, legend_width)
```

HanlinMiao: This should address the pylint and unit test failures as well

nautobot/dcim/elevations.py
icovada: No, because this is meant to show that you're looking at the rear of the device, regardless of which rack face you're looking at

icovada: To quote a ticket I received this morning,
"I was typing a long email to explain my point but then I realised I was wrong".

Good catch. This is a tad more complex than I thought

nautobot/dcim/choices.py
```diff
@@ -105,8 +105,9 @@ class RackDimensionUnitChoices(ChoiceSet):
 class RackElevationDetailRenderChoices(ChoiceSet):
     RENDER_JSON = "json"
     RENDER_SVG = "svg"
+    RENDER_CSV = "csv"
 
-    CHOICES = ((RENDER_JSON, "json"), (RENDER_SVG, "svg"))
+    CHOICES = ((RENDER_JSON, "json"), (RENDER_SVG, "svg"), (RENDER_CSV, "csv"))
```
glennmatthews, line 110: Nit: Since the second element in each tuple is the "human-facing" label, I wonder if this should really be:

```suggestion
    CHOICES = ((RENDER_JSON, "JSON"), (RENDER_SVG, "SVG"), (RENDER_CSV, "CSV"))
```

nautobot/dcim/elevations.py
```diff
@@ -217,7 +223,77 @@ def merge_elevations(self, face):
...
+            for unit in self.rack.get_rack_units(face=face, expand_devices=True):
+                if unit["device"] is None:
+                    rack_positions[unit["id"] - 1] = DeviceCSVRepresentation(name="", unit=unit["name"], device=None)
+
+                else:
+                    rack_positions[unit["id"] - 1] = DeviceCSVRepresentation(
+                        name=unit["device"].name,
+                        unit=unit["name"],
+                        device=unit["device"],
+                    )
```
glennmatthews, line 247: Would something like the below work just as well?

```suggestion
            rack_positions = []

            # Fill with used rack units
            for unit in self.rack.get_rack_units(face=face, expand_devices=True):
                if unit["device"] is None:
                    rack_positions.append(DeviceCSVRepresentation(name="", unit=unit["name"], device=None))

                else:
                    rack_positions.append(DeviceCSVRepresentation(
                        name=unit["device"].name,
                        unit=unit["name"],
                        device=unit["device"],
                    ))
```

I think this would also preserve the `desc_units` ordering of the given rack, simplifying `_output_face_to_csv` below as well?

nautobot/dcim/elevations.py
```diff
@@ -217,7 +223,77 @@ def merge_elevations(self, face):
...
+        def _output_face_to_csv(
+            buffer: io.StringIO, rack_positions: list[DeviceCSVRepresentation], ru_prefix: str = ""
+        ):
+            if self.rack.desc_units:
+                position_iterator = range(self.rack.u_height)
+            else:
+                position_iterator = range(self.rack.u_height - 1, 0, -1)
+
+            for position in position_iterator:
+                this_position = rack_positions[position]
```
glennmatthews, line 260: If `rack_positions` were already ordered by `self.rack.desc_units`, this simplifies, right? To something like:

```suggestion
            for this_position in rack_positions:
```

nautobot/dcim/elevations.py
```diff
@@ -217,7 +223,77 @@ def merge_elevations(self, face):
...
+                if this_position.name is not None:
+                    writer.writerow(
+                        {
+                            "unit": ru_prefix + this_position.unit,
```
glennmatthews, line 264: Is this format desirable or would something like `"face": ru_prefix, "unit": this_position.unit` be better and clearer?

nautobot/dcim/elevations.py
```diff
@@ -217,7 +223,77 @@ def merge_elevations(self, face):
 
         return elevation
 
-    def render(self, face, unit_width, unit_height, legend_width):
+    def render_csv(self, face: Literal["front", "rear"]) -> str:
```
glennmatthews, line 226: Not immediately seeing what code `render_csv` has in common with the existing class to justify merging them, versus just creating a separate standalone `RackElevationCSV` class with just this one method. What am I missing?

nautobot/dcim/models/racks.py
```diff
@@ -363,8 +371,9 @@ def get_reserved_units(self):
                 reserved_units[u] = r
         return reserved_units
 
-    def get_elevation_svg(
+    def get_elevation(
         self,
+        fileformat: Literal[RackElevationDetailRenderChoices.RENDER_CSV, RackElevationDetailRenderChoices.RENDER_SVG],
```
glennmatthews, line 376: Similarly, seems like a separate `get_elevation_csv` method would make more sense given that most of the parameters in the existing method are unused/irrelevant when exporting to CSV.

nautobot/dcim/elevations.py
```diff
@@ -217,7 +223,77 @@ def merge_elevations(self, face):
...
+            name: Optional[int]
+            unit: str
+            device: "Device"
+
+        def _get_face_data(face: Literal["front", "rear"]) -> list[DeviceCSVRepresentation]:
+            # Prepare an empty rack representation
+            rack_positions = [None for x in range(0, self.rack.u_height)]
+
+            # Fill with used rack units
+            for unit in self.rack.get_rack_units(face=face, expand_devices=True):
```
glennmatthews, line 238: Shouldn't we be passing `self.user`, if any, to the `get_rack_units` call?

nautobot/dcim/elevations.py
icovada: Yes... and no.
In case we have to export multiple racks side-by-side (this is a requested feature) and those racks are of different sizes, we can simply find the tallest rack in the row and start requesting directly by unit.

But then again, as I type, I realise we can just index by position and skip that rack should it not be tall enough.
What do you think?

glennmatthews: I'm not sure how this code would need to change to export multiple racks in a single CSV, to be honest.

icovada: Another option could be to return two different rows, one for the front and one for the back view.

This way, should we produce a "row" view, we could just ´itertools.zip_longest()´over the entire row for both front and back and we can "ground" the racks on the same line.

What do you think?

nautobot/dcim/elevations.py
icovada: I see no mention of rack positions being outputted depending on desc_units from .get_rack_units(), maybe we should add that as well?

nautobot/dcim/elevations.py
icovada: I thought I'd keep all the exports in one class but I can split them if you think this is the way to go

lampwins:
@icovada perhaps we can hash out the UX flow of this in #4141 instead of in the PR back and forth?

PR #6791, opened 2025-01-16
fix: passing pk for JobButton job_queue variable

### LINKED ISSUE #6770: JobButtons not triggering Jobs on click
### Environment
* Nautobot version: v2.4.0
* Python version: 3.11

### Steps to Reproduce
1. Create a Job Button
2. Click Job Button
3. Nothing happens

### Expected Behavior
Job triggers

### Observed Behavior
When the Job Button is clicked, inspecting the network packets reveals that the _job_queues value being passed is 'default', when inspecting the same network packets on running a regular job, the _job_queues value looks to be the pk of the default job queue instead of the name 'default'.

I believe that the required change is modifying the _job_queues hidden value in the form created for the job buttons to instead query the pk of the JobQueue object instead of the name value

(my guess) Bad Payload:

no_work_payload

(my guess) Good Payload:

work_payload

my guess for the problem location is within file https://github.com/nautobot/nautobot/blob/v2.4.0/nautobot/extras/templatetags/job_buttons.py under the function _render_job_button_for_obj() and how the _job_queue variable is gathered.

LochieP:
Original implemenation was passing the string name of the job_queue to be used. This was creating a payload that would not run the job on click as what is required is the UUID (pk) of the JobQueue object instead.

# Closes #6770 
# What's Changed

Fixed populating the hidden job queue variable associated to JobButtons, changing from the 'name' field of the JobQueue object being stored to the UUID primary key.

I have just made modifications to the job_buttons.py file under the function _render_job_button_for_obj().

# Screenshots
Before:
image

Intended After:
image

# TODO
- [x] Unit, Integration Tests

### DIFF
--- a/changes/6770.fixed
+++ b/changes/6770.fixed
@@ -1,0 +1,1 @@
+Fixed JobButtons to now trigger jobs on click
--- a/nautobot/extras/templatetags/job_buttons.py
+++ b/nautobot/extras/templatetags/job_buttons.py
@@ -8,7 +8,7 @@ from django.utils.html import format_html
 from django.utils.safestring import mark_safe
 
 from nautobot.core.utils.data import render_jinja2
-from nautobot.extras.models import Job, JobButton
+from nautobot.extras.models import Job, JobButton, JobQueue
 
 register = template.Library()
 
--- a/nautobot/extras/templatetags/job_buttons.py
+++ b/nautobot/extras/templatetags/job_buttons.py
@@ -108,17 +108,17 @@ def _render_job_button_for_obj(job_button, obj, context, content_type):
     # Disable buttons if the user doesn't have permission to run the underlying Job.
     has_run_perm = Job.objects.check_perms(context["user"], instance=job_button.job, action="run")
     try:
-        job_queues = list(job_button.job.job_queues.all().values_list("name", flat=True))
+        job_queues = job_button.job.job_queues.all()
         _job_queue = job_queues[0]
     except IndexError:
-        _job_queue = settings.CELERY_TASK_DEFAULT_QUEUE
+        _job_queue = JobQueue.objects.get(name=settings.CELERY_TASK_DEFAULT_QUEUE)
     hidden_inputs = format_html(
         HIDDEN_INPUTS,
         csrf_token=context["csrf_token"],
         object_pk=obj.pk,
         object_model_name=f"{content_type.app_label}.{content_type.model}",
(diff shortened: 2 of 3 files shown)


glennmatthews (commented):
Code change looks appropriate to me - thank you! Would you like to take a try at adding a unit test for this? I think it'd be an enhancement to `nautobot.extras.tests.test_views.JobButtonRenderingTestCase`.

LochieP:
> Code change looks appropriate to me - thank you! Would you like to take a try at adding a unit test for this? I think it'd be an enhancement to `nautobot.extras.tests.test_views.JobButtonRenderingTestCase`.

Thanks for the review, I have modified the test "test_task_queue_hidden_input_is_present()" to reflect the changes made by this PR.

HanlinMiao (approved):

glennmatthews:
Is it because we get the `queue` directly with `get_job_queue()` but *also* later in the same function do `job_form.is_valid()` and then look up a `job_queue` from the job form? Seems redundant to me.

HanlinMiao:
> Is it because we get the `queue` directly with `get_job_queue()` but _also_ later in the same function do `job_form.is_valid()` and then look up a `job_queue` from the job form? Seems redundant to me.

`job_form` gets its data from `request.POST` [here](https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/views.py#L1373) and `job_form.is_valid()` raises the [error](https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/views.py#L1405) because it is expecting a job queue `pk` instead of a `name`. We need to pass the `job_queue.pk` into the original request data to fix this issue. So I think this fix is necessary. Let me know if that makes sense

glennmatthews:
> > Is it because we get the `queue` directly with `get_job_queue()` but _also_ later in the same function do `job_form.is_valid()` and then look up a `job_queue` from the job form? Seems redundant to me.
> 
> `job_form` gets its data from `request.POST` [here](https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/views.py#L1373) and `job_form.is_valid()` raises the [error](https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/views.py#L1405) because it is expecting a job queue `pk` instead of a `name`. We need to pass the `job_queue.pk` into the original request data to fix this issue. So I think this fix is necessary. Let me know if that makes sense

OK - I think there's some redundant code in here (`queue` vs `job_queue`) that we should revisit but this is fine for now.

PR #6799, opened 2025-01-20
6779 integration tests followup

mateusz-lachowski-codilime:
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests


nautobot/core/testing/integration.py
```diff
@@ -220,3 +253,328 @@ def fill_select2_multiselect_field(self, field_name, value):
...
+    - set edit field and value
+    """
+
+    class BaseTestCase(SeleniumTestCase):
+        model_menu_path = (None, None)
+        model_base_viewname = ""
+        model_edit_data = {}
+        model_filter_by = {}
+        model_class = None
+        model_plural = None
```
glennmatthews, line 287: If these are mandatory for a subclass to define, and don't work with the given defaults, let's use typing to declare them as such with no defaults, perhaps:

```suggestion
        model_menu_path: tuple[str, str]
        model_base_viewname: str
        model_edit_data: dict[str, Any]
        model_filter_by: dict[str, Any]
        model_class: type[Model]
        model_plural: str
```

nautobot/core/testing/integration.py
```diff
@@ -28,23 +28,39 @@ class ObjectsListMixin:
     """
 
     def select_all_items(self):
-        self.browser.find_by_xpath('//*[@id="object_list_form"]//input[@class="toggle"]').click()
+        self.browser.find_by_css("#object_list_form input.toggle").click()
```
glennmatthews, line 31: TIL `find_by_css`. Definitely seems more user-friendly than `find_by_xpath` in these cases!

nautobot/core/testing/integration.py
```diff
@@ -28,23 +28,39 @@ class ObjectsListMixin:
     """
 
     def select_all_items(self):
```
glennmatthews, line 30: Mind adding some docstrings to clarify the distinction between `select_all_items`/`select_all_items_from_all_pages`, `click_bulk_delete`/`click_bulk_delete_all`, etc.?

nautobot/dcim/tests/integration/test_device_bulk_operations.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    Test devices bulk edit / delete operations.
+    """
+
+    model_menu_path = ("Devices", "Devices")
+    model_base_viewname = "dcim:device"
+    model_edit_data = {"serial": "Test serial"}
+    model_filter_by = {"location": "Test Location 2"}
+    model_class = Device
+    model_plural = Device._meta.verbose_name_plural
```
glennmatthews, line 20: Can we add logic in the base class to use `self.model_class._meta.verbose_name_plural` if `model_plural` isn't explicitly set?

glennmatthews (approved):

PR #6824, opened 2025-01-27
Fixed Bulk Edit Objects job when adding/removing a single tag to multiple objects.

### LINKED ISSUE #6810: Bulk Jobs Failing When Editing Multiple Devices
### Environment
* Nautobot version (Docker tag too if applicable): demo.nautobot.com / 2.4.1
* Python version: demo.nautobot.com
* Database platform, version: demo.nautobot.com
* Middleware(s): 

### Steps to Reproduce
1. Log into demo.nautobot.com
2. Create a new tag, specifying dcim.device as a compatible type
3. Under devices, select multiple devices, then "Edit Selected"
4. Attempt to add your new tag and save changes

### Expected Behavior
The selected devices would have the new tag added

### Observed Behavior
Job fails with:
```
2025-01-22 21:31:16.157043 | run | Error | — | Form validation unsuccessful: {"add_tags": [{"message": "Enter a list of values.", "code": "invalid_list"}]}
```

If you check before the data is refreshed, you can see me attempt here:
https://demo.nautobot.com/extras/job-results/678a6c96-df59-41b4-9f8b-683dac2ef83d/?tab=main

HanlinMiao:
# Closes #6810 
# What's Changed

1. Fixed Bulk Edit Objects job when adding/removing a single tag to multiple objects.
2. The problem lies within the `normalize_query_dict` method. It extracts the first value of the value_list when len(values_list) == 1, thus converting a list of str objects to a single str object.
3. It was also checking whether `add_tags`/`remove_tags` field exist in `form_class.base_fields` where it should be looking for them in `form_class.fields`.
4. Will need to add a unit test

# Screenshots

https://github.com/user-attachments/assets/e25c4f6e-ee84-434f-9443-26b049b519e6

# TODO

- [x] Unit, Integration Tests


nautobot/core/views/mixins.py
```diff
@@ -995,7 +995,7 @@ def send_bulk_delete_objects_to_job(self, request, pk_list, model, delete_all):
     def send_bulk_edit_objects_to_job(self, request, form, model):
         """Prepare and enqueue a bulk edit job."""
         job_model = Job.objects.get_for_class_path(BulkEditObjects.class_path)
-        form_data = normalize_querydict(request.POST, form)
+        form_data = normalize_querydict(request.POST, form_class=form)
```
glennmatthews, line 998: Looking at this again, I'm wondering if this call is even needed - normally the whole point of `normalize_querydict` is for use with `request.GET` since there's no easy way to distinguish between single- and multi-value query params. For a form-encoded POST request, the form itself should already have the correct single- versus multi-value representation, shouldn't it?

HanlinMiao: It seems like passing `request.POST` directly to the job does not work, it still makes the job fail with the same error. we still need the `normalize_query_dict` to convert the Querydict object to a dict. I am starting to feel like maybe the problem lies within `TagsBulkEditFormMixin`

HanlinMiao: Okay, I think the right fix is to instantiate the `form_class` when it is not null, calling the `__init__` method and checking the `fields` attribute. https://github.com/nautobot/nautobot/pull/6824/commits/b286940e228a1f670393fc274874454cf482d0c7

glennmatthews: Latest CI failure suggests that `form` is already an instance rather than a class?

HanlinMiao: https://github.com/nautobot/nautobot/pull/6824/commits/ff4205507992b0a81d58c23d7aad8fc65f8e50c2 Yeah added some parameter type enforcement for `normalize_querydict` method. Hopefully it is not a breaking change for apps.

nautobot/core/utils/requests.py
```diff
@@ -164,23 +165,38 @@ def normalize_querydict(querydict, form_class=None, filterset=None):
 
     Raises:
         AttributeError: If both `form_class` and `filterset` are provided.
+        AttributeError: Provided `form_class` is a subclass of `BulkEditForm` but `model` is not provided.
```
glennmatthews, line 168: Any chance of this being a breaking change for Apps?

nautobot/core/views/generic.py
```diff
@@ -1051,7 +1051,7 @@ def post(self, request, **kwargs):
 
             if form.is_valid():
                 logger.debug("Form validation was successful")
-                return self.send_bulk_edit_objects_to_job(request, form, model)
+                return self.send_bulk_edit_objects_to_job(request, self.form, model)
```
glennmatthews, line 1054: Still feels weird/off to me - should we be passing `form.cleaned_data` as a parameter here instead of passing the form class and calling `normalize_querydict()` on the request?

nautobot/core/utils/requests.py
HanlinMiao: There is a chance but I think it is highly unlikely. The most likely scenario is that `send_bulk_edit_objects_to_job` is overridden for some reason, but that is a relatively new method that I do not think apps had the time to yet.

HanlinMiao: Well, I guess unit test failure suggests that it might be a breaking change still.

nautobot/core/tests/test_jobs.py
```diff
@@ -760,6 +760,38 @@ def test_bulk_edit_select_some(self):
...
+            3,
+            description="Example description for bulk edit",
+        )
+
+        # Assert Namespaces withing pk_list updated tags
+        for namespace in namespaces[:3]:
+            self.assertTrue(namespace.tags.filter(pk__in=[self.tags[0].pk]).exists())
+            self.assertFalse(namespace.tags.filter(pk__in=[self.tags[-1].pk]).exists())
+
+        # Assert Namespaces not withing pk_list tags did not get updated
```
glennmatthews, line 790:
```suggestion
        # Assert Namespaces within pk_list get updated tags
        for namespace in namespaces[:3]:
            self.assertTrue(namespace.tags.filter(pk__in=[self.tags[0].pk]).exists())
            self.assertFalse(namespace.tags.filter(pk__in=[self.tags[-1].pk]).exists())

        # Assert Namespaces not within pk_list did not get updated tags
```

nautobot/core/views/mixins.py
```diff
@@ -992,10 +992,9 @@ def send_bulk_delete_objects_to_job(self, request, pk_list, model, delete_all):
         )
         return redirect("extras:jobresult", pk=job_result.pk)
 
-    def send_bulk_edit_objects_to_job(self, request, form, model):
+    def send_bulk_edit_objects_to_job(self, request, form_data, model):
```
glennmatthews, line 995: This is *technically* a breaking change to a public API, though from a quick github search I don't see where any apps have adopted/overridden it yet since it's new in 2.4.0.

changes/6810.fixed
```diff
@@ -0,0 +1 @@
+Fixed Bulk Edit Objects job when adding/removing a single tag to multiple objects.
```
glennmatthews, line 1: I'm *guessing* this isn't specific to tags, but any m2m with `add_` and `remove_` bulk edit fields? Agreed tags are the most likely scenario though, just wondering if the changelog should call out other cases as well?

glennmatthews (approved):

nautobot/core/views/mixins.py
HanlinMiao: Thank you for the research

changes/6810.fixed
HanlinMiao: There are a lot of fields to call out lmk if this works. https://github.com/nautobot/nautobot/pull/6824/commits/55cf8d8f43f4a7d796a47329bc8ca00650301afd

changes/6810.fixed
```diff
@@ -0,0 +1 @@
+Fixed Bulk Edit Objects job failure when passing a single value to add_*/remove_* fields.
```
glennmatthews, line 1:
```suggestion
Fixed Bulk Edit Objects job failure when passing a single value to `add_*`/`remove_*` fields (such as `add_tags`).
```

PR #6825, opened 2025-01-28
Fixed missing Linkify for manufacturer on platform table.

itdependsnetworks:
Didn't test locally, but seems to make sense to me :)

### DIFF
--- a/changes/6825.fixed
+++ b/changes/6825.fixed
@@ -1,1 +1,1 @@
-Fixed missing Linkify for manufacturer on platform table.
+Added links to the `manufacturer` column of the Platform table.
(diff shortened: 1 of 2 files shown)


changes/6825.fixed
```diff
@@ -0,0 +1 @@
+Fixed missing Linkify for manufacturer on platform table.
```
HanlinMiao, line 1: Added links to the `manufacturer` column of the Platform table.

HanlinMiao (approved):

changes/6825.fixed
HanlinMiao: Should probably change the name of the changelog fragment from `.fixed` to `added`.

PR #6832, opened 2025-01-29
Improved returned data when syncing a Git repo via the REST API (#6808)

### LINKED ISSUE #6808: Improve returned data when syncing a Git repository via the REST API
### As ...

### I want ...

to be able to see more useful return data when syncing a Git repository via the REST API. Today, e.g.:

```
curl -s -X POST \
-H "Authorization: Token $TOKEN" \
-H "Accept: application/json; version=2.4" \
http://nautobot/api/extras/git-repositories/2ecb8556-db58-466d-8278-860b8fd74627/sync/
```

returns

```
{
  "message": "Repository demo-git-datasource sync job added to queue."
}
```

I would want to see more data in that return data, including, for example:
- job-result ID
- selected job queue
- worker ID

### So that ...

I can make better programmatic decisions based on the return data.

A more detailed JSON structure is returned, perhaps e.g.:
```
{
  "message": "Repository demo-git-datasource sync job added to queue.",
  "job-id": "XXXXXX",
  "job-queue": "XXXXXX"
  etc.
}

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [x] Misc (including Data Sources)
- [ ] Organization
- [x] Apps (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

Already exists.

### External Dependencies

No

### PR DESCRIPTION
**Pull request by cdtomkins**
# Closes #6808
# What's Changed
- Added extra detail to returned data when syncing a Git repo via the REST API, to resolve #6808.
- Instead of including the proposed `selected job queue` and `worker ID`, I included the serialised `JobResult`, from which the above and more can be easily derived.
- I've left `message` untouched to avoid breaking any existing integrations.
- Also left the HTTP response code at `200` for now to avoid breaking changes. `201` would be more appropriate in future.

# Screenshots
**BEFORE (REST)**
```
curl -s -X POST -H "Authorization: Token <SNIPPED>" -H "Content-Type: application/json" -H "Accept: application/json; version=2.4" http://nautobot:8080/api/extras/git-repositories/35c54754-3a8e-49f5-ae65-2c16202b17e4/sync/ | jq '.'
{
  "message": "Repository demo-git-datasource sync job added to queue."
}
```
**BEFORE (Swagger)**
Screenshot 2025-01-30 095658

**AFTER (REST)**
```
curl -s -X POST -H "Authorization: Token <SNIPPED>" -H "Content-Type: application/json" -H "Accept: application/json; version=2.4" http://nautobot:8080/api/extras/git-repositories/35c54754-3a8e-49f5-ae65-2c16202b17e4/sync/ | jq '.'
{
  "message": "Repository demo-git-datasource sync job added to queue.",
  "job_result": {
    "id": "68500ca5-27c3-488c-a24a-e98858bf52a1",
    "object_type": "extras.jobresult",
    "display": "Git Repository: Sync started at 2025-01-30 19:35:14.120625+00:00 (PENDING)",
    "url": "http://nautobot/api/extras/job-results/68500ca5-27c3-488c-a24a-e98858bf52a1/",
    "natural_slug": "68500ca5-27c3-488c-a24a-e98858bf52a1_6850",
    "status": {
      "value": "PENDING",
      "label": "PENDING"
    },
    "name": "Git Repository: Sync",
    "task_name": null,
    "date_created": "2025-01-30T19:35:14.120625Z",
    "date_done": null,
    "result": null,
    "worker": null,
    "task_args": [],
    "task_kwargs": {},
    "celery_kwargs": {},
    "traceback": null,
    "meta": null,
    "job_model": {
      "id": "ad2b27c8-adf0-4e23-a4d3-a37fe3c42abd",
      "object_type": "extras.job",
      "url": "http://nautobot/api/extras/jobs/ad2b27c8-adf0-4e23-a4d3-a37fe3c42abd/"
    },
    "user": {
      "id": "569138fe-f0b9-4abf-9812-c85a7ec73bbd",
      "object_type": "users.user",
      "url": "http://nautobot/api/users/users/569138fe-f0b9-4abf-9812-c85a7ec73bbd/"
    },
    "scheduled_job": null,
    "custom_fields": {},
    "computed_fields": {},
    "files": []
  }
}
```
**AFTER (Swagger)**
Screenshot 2025-01-31 090015

# TODO
- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/api/serializers.py
+++ b/nautobot/extras/api/serializers.py
@@ -1169,3 +1169,15 @@ class WebhookSerializer(ValidatedModelSerializer, NotesSerializerMixin):
             raise serializers.ValidationError(conflicts)
 
         return validated_attrs
+
+
+#
+# More Git repositories
+#
+
+
+class GitRepositorySyncResponseSerializer(serializers.Serializer):
+    """Serializer representing responses from the GitRepository.sync() POST endpoint."""
+
+    message = serializers.CharField(read_only=True)
+    job_result = JobResultSerializer(read_only=True)
--- a/nautobot/extras/api/views.py
+++ b/nautobot/extras/api/views.py
@@ -425,7 +425,7 @@ class GitRepositoryViewSet(NautobotModelViewSet):
     serializer_class = serializers.GitRepositorySerializer
     filterset_class = filters.GitRepositoryFilterSet
 
-    @extend_schema(methods=["post"], request=serializers.GitRepositorySerializer)
+    @extend_schema(methods=["post"], responses={"200": serializers.GitRepositorySyncResponseSerializer}, request=None)
     # Since we are explicitly checking for `extras:change_gitrepository` in the API sync() method
     # We explicitly set the permission_classes to IsAuthenticated in the @action decorator
     # bypassing the default DRF permission check for `extras:add_gitrepository` and the permission check fall through to the function itself.
@@ -441,8 +441,16 @@ class GitRepositoryViewSet(NautobotModelViewSet):
             raise CeleryWorkerNotRunningException()
 
         repository = get_object_or_404(GitRepository, id=pk)
-        repository.sync(user=request.user)
-        return Response({"message": f"Repository {repository} sync job added to queue."})
+        job_result = repository.sync(user=request.user)
+
+        data = {
+            # Kept message for backward compatibility for now
+            "message": f"Repository {repository} sync job added to queue.",
+            "job_result": job_result,
+        }
+
+        serializer = serializers.GitRepositorySyncResponseSerializer(data, context={"request": request})
+        return Response(serializer.data, status=status.HTTP_200_OK)
 
 
 #
--- a/changes/6808.changed
+++ b/changes/6808.changed
@@ -1,0 +1,1 @@
+Improved returned data when syncing a Git repository via the REST API.
--- a/nautobot/docs/user-guide/platform-functionality/gitrepository.md
+++ b/nautobot/docs/user-guide/platform-functionality/gitrepository.md
@@ -437,14 +437,50 @@ curl -s -X POST \
 -H "Authorization: Token $TOKEN" \
 -H "Content-Type: application/json" \
 -H "Accept: application/json; version=2.4" \
-http://nautobot/api/extras/git-repositories/2ecb8556-db58-466d-8278-860b8fd74627/sync/
+http://nautobot/api/extras/git-repositories/2ecb8556-db58-466d-8278-860b8fd74627/sync/ | jq '.'
 ```
 
 Which returns, for example:
 
 ```no-highlight
 {
-  "message": "Repository demo-git-datasource sync job added to queue."
+  "message": "Repository demo-git-datasource sync job added to queue.",
+  "job_result": {
+    "id": "68500ca5-27c3-488c-a24a-e98858bf52a1",
+    "object_type": "extras.jobresult",
+    "display": "Git Repository: Sync started at 2025-01-30 19:35:14.120625+00:00 (PENDING)",
+    "url": "http://nautobot/api/extras/job-results/68500ca5-27c3-488c-a24a-e98858bf52a1/",
+    "natural_slug": "68500ca5-27c3-488c-a24a-e98858bf52a1_6850",
+    "status": {
+      "value": "PENDING",
+      "label": "PENDING"
+    },
+    "name": "Git Repository: Sync",
+    "task_name": null,
+    "date_created": "2025-01-30T19:35:14.120625Z",
+    "date_done": null,
+    "result": null,
+    "worker": null,
+    "task_args": [],
+    "task_kwargs": {},
+    "celery_kwargs": {},
+    "traceback": null,
+    "meta": null,
+    "job_model": {
+      "id": "ad2b27c8-adf0-4e23-a4d3-a37fe3c42abd",
+      "object_type": "extras.job",
+      "url": "http://nautobot/api/extras/jobs/ad2b27c8-adf0-4e23-a4d3-a37fe3c42abd/"
+    },
+    "user": {
+      "id": "569138fe-f0b9-4abf-9812-c85a7ec73bbd",
+      "object_type": "users.user",
+      "url": "http://nautobot/api/users/users/569138fe-f0b9-4abf-9812-c85a7ec73bbd/"
+    },
+    "scheduled_job": null,
+    "custom_fields": {},
+    "computed_fields": {},
+    "files": []
+  }
 }
 ```
 
@@ -457,7 +493,7 @@ curl -s -X GET \
 -H "Authorization: Token $TOKEN" \
 -H "Content-Type: application/json" \
 -H "Accept: application/json; version=2.4" \
-http://nautobot/api/extras/jobs/?module_name__isw=demo_git_datasource. | jq .
+http://nautobot/api/extras/jobs/?module_name__isw=demo_git_datasource. | jq '.'
 ```
 
 Here are the first 20 lines of JSON returned:
@@ -527,7 +563,7 @@ curl -s -X GET \
 -H "Authorization: Token $TOKEN" \
 -H "Content-Type: application/json" \
 -H "Accept: application/json; version=2.4" \
-http://nautobot/api/extras/config-contexts/?owner_object_id=2ecb8556-db58-466d-8278-860b8fd74627 | jq .
+http://nautobot/api/extras/config-contexts/?owner_object_id=2ecb8556-db58-466d-8278-860b8fd74627 | jq '.'
 ```
 
 Here's the first part of the output:
--- a/nautobot/extras/tests/test_api.py
+++ b/nautobot/extras/tests/test_api.py
@@ -1207,6 +1207,10 @@ class GitRepositoryTest(APIViewTestCases.APIViewTestCase):
         url = reverse("extras-api:gitrepository-sync", kwargs={"pk": self.repos[0].id})
         response = self.client.post(url, format="json", **self.header)
         self.assertHttpStatus(response, status.HTTP_200_OK)
+        self.assertIn("message", response.data)
+        self.assertIn("job_result", response.data)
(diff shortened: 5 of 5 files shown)

### CODE REVIEW

**Comment by cdtomkins** [comment]
@glennmatthews Any comments on the formatting of the extra data I've returned would be appreciated. If there are any standards on the naming, I want to follow them.

File: nautobot/extras/api/views.py [hunk 1]
```diff
@@ -441,9 +441,18 @@ def sync(self, request, pk):
...
-        repository.sync(user=request.user)
-        return Response({"message": f"Repository {repository} sync job added to queue."})
 
+        job_result = repository.sync(user=request.user)
+
+        return Response({
+                "message": f"Repository {repository} sync job added to queue.",
+                "repository_id": repository.id,
+                "repository_name": repository.name,
+                "sync_jobmodel_id": job_result.job_model_id,
```
glennmatthews, line 451: The Job ID probably isn't needed here since it will be the same for all GitRepository syncs.

nautobot/extras/api/views.py
```diff
@@ -441,9 +441,18 @@ def sync(self, request, pk):
...
+                "sync_jobresult_id": job_result.id,
+                "sync_status": job_result.status
```
glennmatthews, line 453: Since the sync runs asynchronously (😄) the `sync_status` here will almost always just be `pending` at the time we send this response, so I'd suggest omitting it.

nautobot/extras/api/views.py
```diff
@@ -441,9 +441,18 @@ def sync(self, request, pk):
             raise CeleryWorkerNotRunningException()
 
         repository = get_object_or_404(GitRepository, id=pk)
-        repository.sync(user=request.user)
-        return Response({"message": f"Repository {repository} sync job added to queue."})
 
+        job_result = repository.sync(user=request.user)
+
+        return Response({
```
glennmatthews, line 447: Ideally we'd define a proper Serializer class for this response (and add it as the `response` type in the `extend_schema` decorator call above so that it's properly documented in Swagger). Compare for example to the `JobViewSetBase.run` view.

nautobot/extras/api/views.py
```diff
@@ -441,9 +441,18 @@ def sync(self, request, pk):
...
+                "message": f"Repository {repository} sync job added to queue.",
+                "repository_id": repository.id,
+                "repository_name": repository.name,
```
glennmatthews, line 450: Probably not needed since you have to know the `repository.id` in order to call this endpoint in the first place?

nautobot/extras/api/views.py
```diff
@@ -441,9 +441,18 @@ def sync(self, request, pk):
...
+                "sync_jobmodel_id": job_result.job_model_id,
+                "sync_jobresult_id": job_result.id,
```
glennmatthews, line 452: It'd probably be better to have a `job_result` key and get its value as a dict from a `serializers.JobResultSerializer(job_result, context={"request": request}).data` call - similar to `JobViewSetBase.run()`.

glennmatthews (commented):
Thanks!

nautobot/extras/api/views.py
cdtomkins: True! I was aware of the fact it's always `pending` but my thinking was that there may be scenarios, now (not sure?) or future, where it's something other than that.

It's not something I feel strongly about though, so I'm happy to remove it if you don't see merit in that.

nautobot/extras/api/views.py
cdtomkins: My thinking here was to allow an easy lookup of the `queue` as per the original enhancement request (which I did write, to be fair!). If we drop this, it's two queries to find out what queue we're scheduled on, rather than just one.

Happy to drop it though!

nautobot/extras/api/views.py
cdtomkins: Noted; I'll do this. I absolutely want to do it the "right way".

nautobot/extras/api/views.py
cdtomkins: True; but when I update e.g. a `Prefix` via the API I get a lot back that I already sent in the request...

nautobot/extras/api/views.py
cdtomkins: Noted; I'll do this. I absolutely want to do it the "right way".

nautobot/extras/api/views.py
cdtomkins: Regarding the Swagger comment, when looking into this I notice that /`extras/git-repositories/{id}/sync/` also shows a `Request body` as `required` - which is not the case if I understand correctly - indeed I didn't think this API endpoint would do anything with a `Request body` at all.

If I'm right about that, I could try to fix that too?

nautobot/extras/api/views.py
cdtomkins: Dropped it as suggested after more thought.

nautobot/extras/api/views.py
cdtomkins: I think I'm on the right track - will follow up.

nautobot/extras/api/views.py
cdtomkins: Done, I think!

nautobot/extras/api/views.py
cdtomkins: Done; I hope I've understood correctly. See what you think.

nautobot/extras/api/views.py
cdtomkins: Done; I hope I've understood correctly. See what you think.

nautobot/extras/api/serializers.py
```diff
@@ -473,6 +473,13 @@ class Meta:
         fields = "__all__"
 
 
+class GitRepositorySyncResponseSerializer(serializers.Serializer):
+    """Serializer representing responses from the GitRepository.sync() POST endpoint."""
+
+    message = serializers.CharField()
+    scheduled_job_id = serializers.UUIDField()
```
glennmatthews, line 480: Can you try:

```suggestion
    message = serializers.CharField(read_only=True)
    job_result = JobResultSerializer(read_only=True)
```

nautobot/extras/api/views.py
```diff
@@ -425,7 +425,7 @@ class GitRepositoryViewSet(NautobotModelViewSet):
     serializer_class = serializers.GitRepositorySerializer
     filterset_class = filters.GitRepositoryFilterSet
 
-    @extend_schema(methods=["post"], request=serializers.GitRepositorySerializer)
+    @extend_schema(methods=["post"], responses={"201": serializers.GitRepositorySyncResponseSerializer}, request=None)
```
glennmatthews, line 428: Good call on `request=None`

nautobot/extras/api/views.py
```diff
@@ -441,8 +441,17 @@ def sync(self, request, pk):
...
-        repository.sync(user=request.user)
-        return Response({"message": f"Repository {repository} sync job added to queue."})
+        job_result = repository.sync(user=request.user)
+
+        serialized_job_result = serializers.JobResultSerializer(job_result, context={"request": request})
+        data = {
+            # Kept message for backward compatibility for now
+            "message": f"Repository {repository} sync job added to queue.",
+            "scheduled_job_id": serialized_job_result.data["id"],
+        }
```
glennmatthews, line 451: Try:

```suggestion
        serialized_job_result = serializers.JobResultSerializer(job_result, context={"request": request})
        data = {
            # Kept message for backward compatibility for now
            "message": f"Repository {repository} sync job added to queue.",
            "job_result": serialized_job_result.data,
        }
```

glennmatthews (commented):
Getting there!

nautobot/extras/api/serializers.py
cdtomkins: nautobot-1       |   File "/source/nautobot/extras/api/serializers.py", line 480, in GitRepositorySyncResponseSerializer
nautobot-1       |     job_result = JobResultSerializer(read_only=True)
nautobot-1       |                  ^^^^^^^^^^^^^^^^^^^
nautobot-1       | NameError: name 'JobResultSerializer' is not defined

^ trying to figure out why, since nearly the same thing is defined in JobRunResponseSerializer without issue

nautobot/extras/api/views.py
cdtomkins: Ah got it. I misunderstood your intent.

nautobot/extras/api/serializers.py
cdtomkins: The problem is the file order... But the serializers are in alphabetical order. Not sure of the right fix, here.

glennmatthews: Sorry - it occurred to me too late that that might be an issue. You might have to move this specific serializer to the end of the file.

cdtomkins: No problem, I guessed you'd say that. It's ugly but I don't see a better way so I'll do that, thanks

cdtomkins: That's dying with `HyperlinkedIdentityField requires the request in the serializer context`. I'll come back to this after I've done a little more learning.

cdtomkins: Fixed

nautobot/extras/api/views.py
```diff
@@ -441,8 +441,16 @@ def sync(self, request, pk):
...
+        job_result = repository.sync(user=request.user)
+
+        data = {
+            # Kept message for backward compatibility for now
+            "message": f"Repository {repository} sync job added to queue.",
+            "job_result": job_result,
+        }
+
+        serializer = serializers.GitRepositorySyncResponseSerializer(data, context={"request": request})
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
```
glennmatthews, line 453: Just to avoid a "breaking" change, let's keep it at a 200 rather than 201 status for now.

glennmatthews (approved):

cdtomkins:
Good to go, I believe.

PR #6843, opened 2025-01-31
6842 Fix invoke integration-tests

### LINKED ISSUE #6842: Run integration tests without running whole nautobot
### Observed issue
- Make sure that nautobot is not running, run `invoke stop`
- Run integration tests: `invoke integration-test`
- Tests tries to navigate to `host.docker.internal:<ephemeral-port>`
- Networking issue can be seen

### Proposed Changes
In `tasks.run_command` we have small `if` that checks if there is nautobot service running. If it is, integration tests will be working because by default `development/dev.env` will set the `SELENIUM_HOST=nautobot` which is a docker compose service name accessible by Selenium instance.

However, if nautobot is not started by running `invoke start`, `tasks.run_command` will run `docker compose run ...` command but without `development/dev.env` loaded. Because it's a `run --rm` container it will have name like `nautobot-2-4-nautobot-run-82c9baf27452`.

Simplest solution for me would be to set `--name` in `tasks.run_command` and then use it in integrations test either by loading `development/dev.env` for this command or changing the default value of `SELENIUM_HOST` in `testing.integration`.

Image

```
# in this case we need to load default development/dev.env to run command
SELENIUM_HOST = os.getenv("NAUTOBOT_SELENIUM_HOST", "host.docker.internal")

# or just change the default
SELENIUM_HOST = os.getenv("NAUTOBOT_SELENIUM_HOST", "nautobot")
```

mateusz-lachowski-codilime:
# Closes #6842 
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -200,7 +200,7 @@ def run_command(context, command, service="nautobot", **kwargs):
         else:
             # Explicitly set the container name to allow network access by calling "nautobot:<port>"
             compose_command = (
-                f"run {'--user=root ' if root else ''}--rm --name 'nautobot' --entrypoint '{command}' {service}"
+                f"run {'--user=root ' if root else ''}--rm --name '{service}' --entrypoint '{command}' {service}"
             )
 
         return docker_compose(context, compose_command, pty=True, **kwargs)
--- a/tasks.py
+++ b/tasks.py
@@ -198,7 +198,10 @@ def run_command(context, command, service="nautobot", **kwargs):
         if service in results.stdout:
             compose_command = f"exec {'--user=root ' if root else ''}{service} {command}"
         else:
-            compose_command = f"run {'--user=root ' if root else ''}--rm --entrypoint '{command}' {service}"
+            # Explicitly set the container name to allow network access by calling "nautobot:<port>"
+            compose_command = (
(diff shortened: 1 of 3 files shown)


tasks.py
```diff
@@ -198,7 +198,10 @@ def run_command(context, command, service="nautobot", **kwargs):
         if service in results.stdout:
             compose_command = f"exec {'--user=root ' if root else ''}{service} {command}"
         else:
-            compose_command = f"run {'--user=root ' if root else ''}--rm --entrypoint '{command}' {service}"
+            # Explicitly set the container name to allow network access by calling "nautobot:<port>"
+            compose_command = (
+                f"run {'--user=root ' if root else ''}--rm --name 'nautobot' --entrypoint '{command}' {service}"
```
glennmatthews, line 203: Should this be `--name '{service}'` instead for cases where the target service isn't `nautobot`?

glennmatthews (commented):
`invoke stop; invoke integration-test` now works locally and looks like it's passing CI as well.

glennmatthews (approved):

tasks.py
mateusz-lachowski-codilime: Fixed.

PR #6849, opened 2025-02-03
6113 Fix menu inside of tables not showing properly

### LINKED ISSUE #6113: Menu inside of tables is rendering behind table footer element (wrong z-index?)
### Environment
* Nautobot version (Docker tag too if applicable): v2.3.0
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Go to https://demo.nautobot.com/dcim/devices/37a938b0-bd5a-4c25-9d98-a51c75d15ba9/interfaces/ or any device interfaces tab of a device that has interfaces
2. Click on the `Connect Cable` button on the last interface in the table

### Expected Behavior

Menu appears above the table

### Observed Behavior

image

mateusz-lachowski-codilime:
# Closes #6113 
# What's Changed
Added custom js code to dynamically attach current dropdown to the body, to override browser behavior when custom overflow is set for `.table-responsive`. Then absolute position is calculated to match the clicked `.dropdown-toggle` button and match right side edges.

# Screenshots
Before:
image
image

After:
image
image

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/project-static/js/dropdown.js
+++ b/nautobot/project-static/js/dropdown.js
@@ -15,17 +15,14 @@ $(document).on('show.bs.dropdown', '.table-responsive .dropdown', function() {
         display: 'table', // required, because we're outside any container
     });
 
-    const hideMenu = function () {
+    // move dropdown back and hide
+    $dropdown.one('hidden.bs.dropdown', function () {
         $menu.appendTo($dropdown);
         $menu.removeAttr('style');
-    }
-
-    // move dropdown back and hide
-    $dropdown.one('hidden.bs.dropdown', hideMenu);
+    });
 
     // hide dropdown when scrolling vertically to avoid recalculating position
     $('.table-responsive').one('scroll', function() {
-        hideMenu();
         $toggle.trigger("click");
     });
 });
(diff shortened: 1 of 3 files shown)


nautobot/project-static/js/dropdown.js
```diff
@@ -0,0 +1,24 @@
+$(document).on('show.bs.dropdown', '.table-responsive .dropdown', function() {
```
glennmatthews, line 1: This is probably going to go away when we move to Bootstrap v5 anyway, but just in case it will need to live on, can we implement it with vanilla JS instead of jQuery? Or is jQuery necessary to work with the bootstrap 3 event implementation?

mateusz-lachowski-codilime: It seems like those events are not standard JS events so I would need to either change them and re-fire or just stick with current approach. In theory I can write offset calculating in vanilla js but I think it's more clean this way.

mateusz-lachowski-codilime:
[Screencast from 2025-02-03 16-49-44.webm](https://github.com/user-attachments/assets/96f007e4-294c-44e1-9316-5214ae904d82)

timizuoebideri1 (approved):

PR #6851, opened 2025-02-03
Fix #6650 - change fallback behavior of get_base_template()

### LINKED ISSUE #6650: NautobotUIViewSet notes and changelog views should work without an html template
### Environment
* Nautobot version (Docker tag too if applicable): v2.4.0-beta1
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Create a NautobotUIViewSet using the new component framework
2. Navigate to a detail view and click on the Notes or Changelog tab

### Expected Behavior

The tabs are rendered correctly because the `base_template` of the viewset defaults to `generic/object_retrieve.html`

### Observed Behavior

The tabs are not rendered because the `base_template` of the viewset defaults to `base.html`

glennmatthews:
# Closes #6650
# What's Changed

- Change default fallback case of `get_base_template()` to be `generic/object_retrieve.html` instead of `base.html` (fix #6650)
    - Manually tested by deleting `circuits/circuit.html` and `circuits/circuit_retrieve.html` locally and confirming that the circuit detail view, and its associated "Notes" and "Changelog" tabs all still rendered successfully.
    - This is *technically* a breaking change to a `nautobot.apps.utils` API but I think we're in agreement that the old behavior was wrong (a bug). **I do note that both `nautobot-app-bgp-models` and `nautobot-app-data-validation` are calling this API, but based on a quick read of their usage (in both cases, adding additional generic detail tabs, similar to Notes/Changelog tabs) this change should be a net improvement/fix to those apps rather than causing any breakage to current use.**
    - Improved the docstring for this function and code that was using it in various generic views and mixins.
    - Added some basic unit tests for this function
- Because `generic/object_detail.html` has been a mere redirect to `generic/object_retrieve.html` since NautobotUIViewSet was introduced in v1.4, I went on a search-and-replace spree throughout our code and docs to replace lingering references to `generic/object_detail.html`. (I did **not** remove the actual file itself, as that would be a breaking change that would potentially impact Apps, but we should strongly consider doing so as part of the upcoming UI changes in v3.0)

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/extras/utils.py
+++ b/nautobot/extras/utils.py
@@ -6,6 +6,7 @@ import hmac
 import logging
 import re
 import sys
+from typing import Optional
 
 from django.apps import apps
 from django.conf import settings
@@ -13,7 +14,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.core.cache import cache
 from django.core.validators import ValidationError
 from django.db import transaction
-from django.db.models import Q
+from django.db.models import Model, Q
 from django.template.loader import get_template, TemplateDoesNotExist
 from django.utils.deconstruct import deconstructible
 import kubernetes.client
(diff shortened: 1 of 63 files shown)


HanlinMiao (approved):
LGTM!

HanlinMiao (commented):
Maybe we can get this one in 2.4.2

nautobot/extras/utils.py
```diff
@@ -36,24 +37,32 @@
...
+    Args:
+        base_template (str, optional): If not None, this explicitly specified template will be preferred.
+        model (Model): The model to identify a base template for, if base_template is None.
+
+    Returns the specified `base_template`, if not `None`.
+    Otherwise, if `"<app>/<model_name>.html"` exists (legacy ObjectView pattern), returns that string.
+    Otherwise, if `"<app>/<model_name>_retrieve.html"` exists (as used in `NautobotUIViewSet`), returns that string.
+    If all else fails, returns `"generic/object_retrieve.html"`.
+
+    Note: before Nautobot 2.4.3, this API would default to "base.html" rather than "generic/object_retrieve.html".
```
glennmatthews, line 53:
```suggestion
    Note: before Nautobot 2.4.2, this API would default to "base.html" rather than "generic/object_retrieve.html".
```

PR #6856, opened 2025-02-04
6841 Fixed missing termination side

### LINKED ISSUE #6841: [2.4] Circuits: Adding a termination A or Z fails
BUG:  Unable to add circuit termination locations on either A or Z side of the circuit.

* Nautobot version: v2.4.1
* Python version: Python 3.12.3
* Database platform, version: PostgreSQL 16.6
* Middleware(s):

### Steps to Reproduce
1. go to Circuits, Add a new circuit
2. Once a circuit is created, add a location to either A or Z side
3. Return to the circuit, should see the location... (doesn't work) 

### Expected Behavior
Add the location to A or Z side to then be able to attache the circuit to a port. 

### Observed Behavior
Fails, doesn't insert into the database either, table circuits_circuittermination shows no new terminations. 
Tested with an older system where it works. 

Please help :)

mateusz-lachowski-codilime:
# Closes #6841 
# What's Changed
- Added side context variable to the `CircuitTerminationPanel`

# Screenshots

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/core/testing/integration.py
+++ b/nautobot/core/testing/integration.py
@@ -316,24 +316,27 @@ class SeleniumTestCase(StaticLiveServerTestCase, testing.NautobotTestCaseMixin):
         self.browser.is_element_not_present_by_css(".loading-results", wait_time=5)
         return search_box
 
+    def _select_select2_result(self):
+        found_results = self.browser.find_by_css(".select2-results li.select2-results__option")
+        # click the first found item if it's not `None`: special value to nullify field
+        if found_results.first.text != "None":
+            found_results.first.click()
+        else:
+            found_results[1].click()
+
     def fill_select2_field(self, field_name, value):
         """
         Helper function to fill a Select2 single selection field on add/edit forms.
         """
-        search_box = self._fill_select2_field(field_name, value)
-        if value:
-            self.browser.find_by_xpath(
-                f"//li[contains(@class, 'select2-results__option') and text()='{value}']"
-            ).click()
-        else:
-            search_box.first.type(Keys.ENTER)
+        self._fill_select2_field(field_name, value)
+        self._select_select2_result()
 
     def fill_filters_select2_field(self, field_name, value):
         """
         Helper function to fill a Select2 single selection field on filters modals.
         """
         self._fill_select2_field(field_name, value, search_box_class="select2-search select2-search--inline")
-        self.browser.find_by_xpath(f"//li[contains(@class, 'select2-results__option') and text()='{value}']").click()
+        self._select_select2_result()
 
     def fill_select2_multiselect_field(self, field_name, value):
         """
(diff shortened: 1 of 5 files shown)


glennmatthews (approved):
Fix looks good. Think it's worthwhile to add an integration test for this?

mateusz-lachowski-codilime:
> Fix looks good. Think it's worthwhile to add an integration test for this?

Actually I'm currently writing some integration tests. I was just wondering whether we should wait with this for tests or not.

nautobot/core/testing/integration.py
```diff
@@ -108,6 +112,22 @@ def apply_filter(self, field, value):
...
 
+class ObjectDetailsMixin:
+    def assertPanelValue(self, panel_label, field_label, expected_value, exact_match=False):
+        """
+        Find the proper panel and asserts if given value match rendered field value.
+        By default, it's not using the exact match, because on the UI we're often adding
+        additional tags, relationships or units.
+        """
+        panel_xpath = f'//*[@id="main"]//div[@class="panel-heading"][contains(normalize-space(), "{panel_label}")]/following-sibling::table'
+        value = self.browser.find_by_xpath(f'{panel_xpath}//td[text()="{field_label}"]/following-sibling::td[1]').text
```
mateusz-lachowski-codilime, line 123: This could be done in more intuitive way by chaining found elements:
```
panel = self.browser.find_by_xpath(panel_xpath)
value = panel.find_by_xpath(...).text
```

But apparently, there is a bug in splinter itself... https://github.com/cobrateam/splinter/issues/1325

nautobot/core/testing/integration.py
```diff
@@ -296,7 +321,12 @@ def fill_select2_field(self, field_name, value):
         Helper function to fill a Select2 single selection field on add/edit forms.
         """
         search_box = self._fill_select2_field(field_name, value)
-        search_box.first.type(Keys.ENTER)
+        if value:
+            self.browser.find_by_xpath(
+                f"//li[contains(@class, 'select2-results__option') and text()='{value}']"
+            ).click()
+        else:
+            search_box.first.type(Keys.ENTER)
```
mateusz-lachowski-codilime, line 329: When searching in select with `None` listed, we can't just click Enter because it will select None instead of value. Maybe this is a separate issue to remove `None` when search is applied in select2?

nautobot/core/testing/integration.py
```diff
@@ -296,7 +321,12 @@ def fill_select2_field(self, field_name, value):
         Helper function to fill a Select2 single selection field on add/edit forms.
         """
         search_box = self._fill_select2_field(field_name, value)
-        search_box.first.type(Keys.ENTER)
+        if value:
+            self.browser.find_by_xpath(
+                f"//li[contains(@class, 'select2-results__option') and text()='{value}']"
```
glennmatthews, line 326: Looks like this change is failing one of the other test cases - maybe we need to do a partial match on `value` since it's a search string rather than an exact match? Or just pick the first entry that isn't a `None`?

mateusz-lachowski-codilime: Fixed by selecting second item if first is "None".

timizuoebideri1 (approved):

PR #6863, opened 2025-02-05
Fixed global and user default saved views incorrectly overriding filtered views.

### LINKED ISSUE #6764: Default saved view overrides incoming links with different filters
### Environment
* Nautobot version (Docker tag too if applicable): 2.4.1b1 (latest develop branch)
* Python version: 3.12.8
* Middleware(s): N/A

### Steps to Reproduce
1. Go to Device list view
2. Filter the view using a filter
3. Create a new Saved View based on the filter
4. Set that new Saved View as Default
5. Go to Platforms and click one of the links in the Devices column (or any list view that provides filtered links to Devices)

### Expected Behavior
I expect to be taken to the Devices list view with the platform filter applied

### Observed Behavior
I get the default saved view instead

HanlinMiao:
# Closes #6764 
# What's Changed

Fixed global and user default saved views incorrectly overriding filtered views.
# Screenshots

N/A
# TODO

~~- [ ] Attached Screenshots, Payload Example~~
- [x] Unit, Integration Tests
~~- [ ] Documentation Updates (when adding/changing features)~~
~~- [ ] Example App Updates (when adding/changing features)~~
~~- [ ] Outline Remaining Work, Constraints from Design~~

### DIFF
--- a/nautobot/core/views/generic.py
+++ b/nautobot/core/views/generic.py
@@ -26,7 +26,7 @@ from django_filters import FilterSet
 from django_tables2 import RequestConfig, Table
 
 from nautobot.core.api.utils import get_serializer_for_model
-from nautobot.core.constants import MAX_PAGE_SIZE_DEFAULT
+from nautobot.core.constants import MAX_PAGE_SIZE_DEFAULT, NON_FILTER_PARAMS
 from nautobot.core.exceptions import AbortTransaction
 from nautobot.core.forms import (
     BootstrapMixin,
@@ -163,7 +163,7 @@ class ObjectListView(ObjectPermissionRequiredMixin, View):
         params = request.GET.copy()
         filter_params = get_filterable_params_from_filter_params(
             params,
-            self.non_filter_params,
+            NON_FILTER_PARAMS,
             self.filterset(),  # pylint: disable=not-callable  # this fn is only called if filterset is not None
         )
         if params.get("saved_view") and not filter_params and not params.get("all_filters_removed"):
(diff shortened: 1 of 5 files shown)


nautobot/core/views/generic.py
```diff
@@ -163,7 +163,7 @@ def get_filter_params(self, request):
         params = request.GET.copy()
         filter_params = get_filterable_params_from_filter_params(
             params,
-            self.non_filter_params,
+            NON_FILTER_PARAMS,
```
glennmatthews, line 166: Not sure I understand this change. Is `self.non_filter_params` unused now? Have we lost the ability for a specific view to extend this list with its own particular special case parameters?

nautobot/core/views/generic.py
```diff
@@ -214,8 +214,14 @@ def get(self, request):
         resolved_path = resolve(request.path)
         list_url = f"{resolved_path.app_name}:{resolved_path.url_name}"
 
+        skip_user_and_global_default_saved_view = False
+        for key in request.GET.keys():
+            if key not in NON_FILTER_PARAMS:
```
glennmatthews, line 219: Can/should we use `get_filterable_params_from_filter_params()` here?

nautobot/extras/tests/test_views.py
```diff
@@ -1543,6 +1543,39 @@ def test_user_default_precedes_global_default(self):
...
+            view=view_name,
+            is_global_default=True,
+        )
+        response = self.client.get(reverse(view_name) + "?location_type=Campus", follow=True)
+        # Assert that the user is not redirected to the global default view
+        # But instead redirected to the filtered view
+        self.assertNotIn(
+            "<strong>Global Location Default View</strong>",
+            extract_page_body(response.content.decode(response.charset)),
+        )
```
glennmatthews, line 1561: Wonder if instead/in addition we should make the saved view apply a different filter (e.g. `location_type=Floor`) and assert that any `Campus` location is shown in the response and no `Floor` location is shown in the response?

nautobot/core/views/generic.py
HanlinMiao: Yeah I did not think about the fact there might be views extending this attribute, I will revert this change

nautobot/core/views/mixins.py
```diff
@@ -723,8 +723,14 @@ def list(self, request, *args, **kwargs):
             if response is not None:
                 return response
 
+        skip_user_and_global_default_saved_view = get_filterable_params_from_filter_params(
+            request.GET.copy(),
+            self.non_filter_params,
+            self.filterset_class(),  # pylint: disable=not-callable  # this fn is only called if filterset is not None
```
timizuoebideri1, line 729: if `self.filterset_class` could be `None`, what about?

```python
filterset = self.filterset_class() if self.filterset_class() else None
```

glennmatthews (approved):
I *think* this looks OK. :-)

Issue #6218, opened 2024-09-15
Saved View: Devices -> Has Primary IP throws Exception


kingfetty:
### Environment
* Nautobot version (Docker tag too if applicable): 2.3.2
* Python version: 3.11.9
* Database platform, version: Nautobot Cloud
* Middleware(s): Nautobot Cloud

### Steps to Reproduce
1. Created a filter of devices with has_primary_ip == True
2. Save this filter as a view

### Expected Behavior

View is accessible with no exceptions thrown.

### Observed Behavior
Exception is thrown, but view is saved.  When attempting to access view exception is regenerated.
```
<class 'TypeError'>

unhashable type: 'list'
```

View Contents:
```
{
    "filter_params": {
        "has_primary_ip": [
            "True"
        ]
    },
    "per_page": 50,
    "sort_order": [],
    "table_config": [
        "name",
        "status",
        "device_type",
        "location",
        "rack",
        "primary_ip"
    ]
}

```

HanlinMiao:
Thank you for the submission! It seems like the origin of the problem is from this [line](https://github.com/nautobot/nautobot/blob/develop/nautobot/core/views/generic.py#L240). Not entirely sure why this is only failing when applying the `has_primary_ip` filter. Seems to be fine with other filters. Maybe because `has_primary_ip` is a method filter?

HanlinMiao:
After further testing, `local_config_context_data` which is another BooleanFilter with method also failed as well.

HanlinMiao:
So the issue is from [NullBooleanSelect](https://github.com/django/django/blob/main/django/forms/widgets.py#L838-L850) which is the default widget for django's built-in BooleanFilter. Changing the filter field to RelatedMembershipBooleanFilter seems to fix the issue.

qduk:
Also fails with `platforms__isnull` which again is a BooleanFilter. Just an FYI.

HanlinMiao:
Closed by #6864

PR #6864, opened 2025-02-05
Fixed Saved Views throwing an unexpected error when they contain boolean filter parameters.

### LINKED ISSUE #6785: [V2.4.0]Jobs saved view results in error when filtered to "Hidden"
### Environment
* Nautobot version (Docker tag too if applicable): Docker nautobot:2.3.16
* Python version: as provided with image
* Database platform, version: as provided with image
* Middleware(s): n/a

### Steps to Reproduce
1.  Set a job's attribute Hidden to True, successful
2. Add filter view to Hidden=True, able to see hidden job
3.  Saved Views > Save as New View as "Hidden", able to see saved view in "Saved View" drop down
4. Click Saved View > "Hidden", results in nautobot error page

### Expected Behavior
We expect to see view that shows only Hidden=True as seen in step2

### Observed Behavior
Nautobot Page with error show as follow:
```
<class 'TypeError'>

unhashable type: 'list'

Python version: 3.12.8
Nautobot version: 2.3.16
```

HanlinMiao:
# Closes #6785 
# What's Changed

Fixed Saved Views throwing an unexpected error when they contain boolean filter parameters.
# Screenshots

Screenshot 2025-02-05 at 1 30 13 PM

# TODO

- Unit, Integration Tests


nautobot/extras/views.py
```diff
@@ -1265,9 +1265,10 @@ class JobListView(generic.ObjectListView):
     def alter_queryset(self, request):
         queryset = super().alter_queryset(request)
         # Default to hiding "hidden" and non-installed jobs
-        if "hidden" not in request.GET:
+        filter_params = self.get_filter_params(request)
```
HanlinMiao, line 1268: SavedView `filter_params` attribute could contain `hidden` or `installed` filters as well

glennmatthews:
I'm guessing this likely also fixes #6218?

Are we lucky enough to also have this fix nautobot/nautobot-app-device-lifecycle-mgmt#413?

nautobot/core/filters.py
```diff
@@ -593,6 +593,7 @@ class BaseFilterSet(django_filters.FilterSet):
     FILTER_DEFAULTS.update(
         {
             models.AutoField: {"filter_class": MultiValueNumberFilter},
+            models.BooleanField: {"filter_class": MultiValueCharFilter},
```
glennmatthews, line 596: Feels wrong to me. Do we have (or need) a `MultiValueBooleanFilter` instead?

nautobot/core/templatetags/helpers.py
```diff
@@ -870,9 +870,7 @@ def saved_view_modal(
 ):
     from nautobot.extras.forms import SavedViewModalForm
     from nautobot.extras.models import SavedView
-
-    param_dict = {}
-    filters_applied = parse_qs(params)
```
HanlinMiao, line 875: Sorry for changing the direction, but I think this will be a more correct and generic fix. So the problem originates from parse_qs which returns every query value as a list for example `{"hidden": [True]}`, which caused the error we saw in the issue. The more generic fix should be using `is_single_choice_field` to determine whether to pass the value as a list or a single value into the Saved View config.

nautobot/extras/views.py
```diff
@@ -1839,14 +1841,14 @@ def create(self, request, *args, **kwargs):
...
         is_shared = request.POST.get("is_shared", False)
         if is_shared:
             is_shared = True
         params = request.POST.get("params", "")
+        param_dict = format_query_param_dict(parse_qs(params), view_name, self.non_filter_params)
 
-        param_dict = parse_qs(params)
-
-        single_value_params = ["saved_view", "table_changes_pending", "all_filters_removed", "q", "per_page"]
+        single_value_params = ["saved_view", "table_changes_pending", "all_filters_removed", "per_page"]
```
HanlinMiao, line 1851: "q"'s single value nature is already taken care of by format_query_param_dict.

nautobot/core/templatetags/helpers.py
glennmatthews: Never apologize for making things more correct! :-)

nautobot/extras/utils.py
```diff
@@ -880,3 +882,22 @@ def bulk_delete_with_bulk_change_logging(qs, batch_size=1000):
...
+    Args:
+        param_dict (dict): key-value pairs of query parameters.
+        view_name (str): The name of the view that the saved view is associated with. "dcim:location_list" for example.
+        non_filter_params (list): List of non-query parameters that should not be formatted.
+    """
+    model = get_model_for_view_name(view_name)
+    filterset = get_filterset_for_model(model)
+
+    for filter_field, value in param_dict.items():
+        if filter_field not in non_filter_params and is_single_choice_field(filterset(), filter_field):
```
glennmatthews, line 901: Probably more efficient to not re-instantiate the filterset on each iteration of the for loop?

glennmatthews, line 901: From the test failure in CI, looks like probably need a try/except here for filter_field names that aren't actually in the filterset

nautobot/extras/tests/test_views.py
```diff
@@ -1625,31 +1624,117 @@ def test_is_shared(self):
...
+            self.assertEqual(instance.config["sort_order"], ["name"])
+            response = self.client.get(reverse(view_name) + "?saved_view=" + str(instance.pk), follow=True)
+            # Assert that Job List View rendered with the boolean filter parameter without error
+            self.assertHttpStatus(response, 200)
+            response_body = extract_page_body(response.content.decode(response.charset))
+            self.assertIn(str(instance.pk), response_body, msg=response_body)
+            self.assertBodyContains(response, f"<strong>{sv_name}</strong>", html=True)
+
+        @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
+        def test_update_saved_view_contain_boolean_filter_params(self):
```
glennmatthews, line 1689: wrong indentation here and below?

nautobot/extras/utils.py
```diff
@@ -880,3 +883,30 @@ def bulk_delete_with_bulk_change_logging(qs, batch_size=1000):
         finally:
             change_context.defer_object_changes = False
             change_context.reset_deferred_object_changes()
+
+
+def format_query_param_dict(param_dict, view_name, non_filter_params):
```
glennmatthews, line 888: I feel like this is a really generic name - can we come up with something clearer, maybe like `fixup_filterset_query_params`? I dunno.

glennmatthews (approved):

glennmatthews:
> I'm guessing this likely also fixes #6218?
> 
> Are we lucky enough to also have this fix [nautobot/nautobot-app-device-lifecycle-mgmt#413](https://github.com/nautobot/nautobot-app-device-lifecycle-mgmt/issues/413)?

@HanlinMiao Any comment on this?

HanlinMiao:
> > I'm guessing this likely also fixes #6218?
> > Are we lucky enough to also have this fix [nautobot/nautobot-app-device-lifecycle-mgmt#413](https://github.com/nautobot/nautobot-app-device-lifecycle-mgmt/issues/413)?
> 
> @HanlinMiao Any comment on this?

Sorry, I missed the comment. #6218 is fixed, I confirmed manually and added unit tests. I just confirmed that Saved View by dates, which is https://github.com/nautobot/nautobot-app-device-lifecycle-mgmt/issues/413, should also work as well
Screenshot 2025-02-07 at 10 00 07 AM

PR #6865, opened 2025-02-05
Refactored tenancy app to use NautobotUIViewSet and UI component framework

HanlinMiao:
# Closes #N/A
# What's Changed

Refactored tenancy app to use `NautobotUIViewSet` and UI component framework.
Added Bulk Edit functionality for the Tenant Group model.

# Screenshots

TenantGroup Detail Before/After:
Screenshot 2025-02-05 at 5 41 24 PM
Screenshot 2025-02-05 at 5 41 54 PM

Tenant is already using UI component framework; but I did notice this weird wrapping/layout on the StatsPanel:
Screenshot 2025-02-05 at 5 43 28 PM

# TODO

- Unit, Integration Tests


HanlinMiao:
There are some subtle differences between the old and new TenantGroup detail view:

+++:
`Hierarchy` table cell that was not there before but I think it would be helpful.

---:
Missing `tenant` table cell that is a counter of how many tenants belong to this tenant group and contains a link to the filtered list of tenants.

nautobot/tenancy/views.py
```diff
@@ -18,68 +20,72 @@
...
+class TenantGroupUIViewSet(
+    mixins.ObjectDetailViewMixin,
+    mixins.ObjectListViewMixin,
+    mixins.ObjectEditViewMixin,
+    mixins.ObjectDestroyViewMixin,
+    mixins.ObjectBulkDestroyViewMixin,
+    mixins.ObjectBulkCreateViewMixin,
+    mixins.ObjectChangeLogViewMixin,
+    mixins.ObjectNotesViewMixin,
+):
```
HanlinMiao, line 32: Only need `BulkUpdateViewMixin` to replace these lines with `NautobotUiViewSet`

HanlinMiao: https://github.com/nautobot/nautobot/pull/6865/commits/92388203b4235a21498651cd1b9b7b13fc0a066b

HanlinMiao:
Screenshot 2025-02-05 at 10 02 16 PM
Did not think I changed the template at all, but I think it's somehow fixed.

timizuoebideri1 (approved):

nautobot/tenancy/templates/tenancy/tenant_edit.html
```diff
@@ -1,21 +1,2 @@
...
-    {% include 'inc/extras_features_edit_form_fields.html' %}
-    <div class="panel panel-default">
-        <div class="panel-heading"><strong>Comments</strong></div>
-        <div class="panel-body">
-            {% render_field form.comments %}
-        </div>
-    </div>
-{% endblock %}
+{% extends 'tenancy/tenant_create.html' %}
+{% comment %}2.0 TODO: remove this template, which only exists for backward compatibility with 1.3 and earlier{% endcomment %}
```
glennmatthews, line 2: update version numbers mentioned here?

nautobot/tenancy/templates/tenancy/tenantgroup.html
```diff
@@ -1,44 +1,2 @@
...
-                        <span class="mdi mdi-plus-thick" aria-hidden="true"></span> Add tenant
-                    </a>
-                </div>
-            {% endif %}
-        </div>
-        {% include 'inc/paginator.html' with paginator=tenant_table.paginator page=tenant_table.page %}
-        <div class="row"></div>
-{% endblock content_right_page %}
+{% extends 'tenancy/tenantgroup_retrieve.html' %}
+{% comment %}2.0 TODO: remove this template, which only exists for backward compatibility with 1.3 and earlier{% endcomment %}
```
glennmatthews, line 2: and here

nautobot/tenancy/templates/tenancy/tenantgroup_retrieve.html
```diff
@@ -0,0 +1,2 @@
+{% extends 'generic/object_retrieve.html' %}
+{% load helpers %}
```
glennmatthews, line 2: remove as unneeded?

nautobot/tenancy/tests/test_views.py
```diff
@@ -14,6 +14,10 @@ def setUpTestData(cls):
             "description": "A new tenant group",
         }
 
+        cls.bulk_edit_data = {
```
glennmatthews, line 17: Do we need to add a new test case mixin as well?

nautobot/tenancy/views.py
```diff
@@ -18,68 +19,64 @@
...
-        tenant_table.columns.hide("tenant_group")
-
-        paginate = {
-            "paginator_class": EnhancedPaginator,
-            "per_page": get_paginate_count(request),
-        }
-        RequestConfig(request, paginate).configure(tenant_table)
+        context = super().get_extra_context(request, instance)
+        if self.action == "retrieve":
+            tenants = Tenant.objects.restrict(request.user, "view").filter(
```
glennmatthews, line 51: Is this chunk of code even needed still? I thought using `ObjectsTablePanel` could let it be autogenerated.

Is it just needed because we're having to use `instance.descendants` as the filter parameter? If so, a comment to that effect would be good.

nautobot/tenancy/views.py
```diff
@@ -18,68 +19,64 @@
...
+    table_class = tables.TenantGroupTable
+    object_detail_content = ObjectDetailContent(
+        panels=(
+            ObjectFieldsPanel(
+                section=SectionChoices.LEFT_HALF,
+                weight=100,
+                fields="__all__",
+            ),
+            ObjectsTablePanel(
+                weight=200,
```
glennmatthews, line 38:
```suggestion
                weight=100,
```

nautobot/tenancy/templates/tenancy/tenant_edit.html
HanlinMiao: I changed this and other templates in core to
```
{% comment %}3.0 TODO: remove this template, which only exists for backward compatibility with 2.4 and earlier{% endcomment %}
```
Let me know if these version numbers are correct

glennmatthews:
> ---: Missing `tenant` table cell that is a counter of how many tenants belong to this tenant group and contains a link to the filtered list of tenants. I don't think any of the `value_transforms` can create such value yet??

This is now automatically provided in the Tenants table header:

image

nautobot/circuits/templates/circuits/circuit.html
```diff
@@ -1,2 +1,2 @@
 {% extends 'circuits/circuit_retrieve.html' %}
-{% comment %}2.0 TODO: remove this template, which only exists for backward compatibility with 1.3 and earlier{% endcomment %}
+{% comment %}3.0 TODO: remove this template, which only exists for backward compatibility with 2.4 and earlier{% endcomment %}
```
glennmatthews, line 2: I don't think there's a need to update the comment in the `circuits` templates - the current comment is correct that the circuit templates were changed in 1.4 and the old files really should have been removed in 2.0.

nautobot/core/templates/generic/object_bulk_delete.html
```diff
@@ -1,2 +1,2 @@
 {% extends 'generic/object_bulk_destroy.html' %}
-{% comment %}2.0 TODO: remove this template, which only exists for backward compatibility with 1.3 and earlier{% endcomment %}
+{% comment %}3.0 TODO: remove this template, which only exists for backward compatibility with 2.4 and earlier{% endcomment %}
```
glennmatthews, line 2: Probably also fine not to update the comments on the core/generic templates either - though it may be a long time before we can "safely" remove them, unlike the circuits ones.

nautobot/tenancy/views.py
```diff
@@ -18,68 +19,67 @@
...
+            ObjectFieldsPanel(
+                section=SectionChoices.LEFT_HALF,
+                weight=100,
+                fields="__all__",
+            ),
+            ObjectsTablePanel(
+                weight=100,
+                section=SectionChoices.RIGHT_HALF,
+                table_class=tables.TenantTable,
+                table_filter="tenant_group",
```
glennmatthews, line 41: Something still seems off here to me - `table_class` and `table_filter` are kwargs to use to auto-create and auto-populate the table. Given you're explicitly providing a table instance in `get_extra_context`, shouldn't this be using `context_table_key="tenant-table"` instead?

glennmatthews (approved):

PR #6879, opened 2025-02-09
Feature: Download button added to download cable trace diagram as SVG

Dharshini1810:
download_svg
downloaded_svg

# What's Changed

I've added Download SVG button in cable_trace.html that downloads the SVG format of the path generated. The SVG image contains clickable links of every segment of the path just like the generated path

The above feature involves minimal code changes, just a small script written in JS, which is triggered when the Download SVG button is clicked
# Screenshots

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/dcim/templates/dcim/cable_trace.html
+++ b/nautobot/dcim/templates/dcim/cable_trace.html
@@ -1,130 +1,221 @@
 {% extends 'base.html' %}
 {% load helpers %}
-
 {% block header %}
-    <h1>{% block title %}Cable Trace for {{ object|meta:"verbose_name"|bettertitle }} {{ object }}{% endblock %}</h1>
+<h1>{% block title %}Cable Trace for {{ object|meta:"verbose_name"|bettertitle }} {{ object }}{% endblock %}</h1>
 {% endblock %}
 
 {% block content %}
-    <div class="row">
-        <div class="col-md-5 col-sm-12">
-            <div class="cable-trace">
-                {% with traced_path=path.origin.trace %}
-                    {% for near_end, cable, far_end in traced_path %}
+<div class="row">
+    <div class="col-md-5 col-sm-12">
+        <div class="cable-trace" id="cable-trace">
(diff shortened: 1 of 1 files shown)


glennmatthews (commented):
Could you please revert the unnecessary whitespace changes in this file? It makes it difficult to review and identify what has actually been changed. Thanks!

Dharshini1810:
> > Could you please revert the unnecessary whitespace changes in this file? It makes it difficult to review and identify what has actually been changed. Thanks!

I'll update and push the changes again. Thanks!

Dharshini1810 (commented):
Kindly do review the change. Thanks!

glennmatthews (commented):
I didn't anticipate that this PR would be adding net-new code for rendering an SVG client-side; I'd assumed that this would be more like the rack-elevation functionality where the SVG is rendered server-side for both the "display" and "download" features. Here it seems that the "display" of a cable trace is constructed from HTML elements and so the "download" needs to be a completely different rendering pipeline. I'm hesitant to move forward with this PR for that reason alone; it might be worth revisiting this as part of the v3.0 UI refresh instead.

Dharshini1810:
> I didn't anticipate that this PR would be adding net-new code for rendering an SVG client-side; I'd assumed that this would be more like the rack-elevation functionality where the SVG is rendered server-side for both the "display" and "download" features. Here it seems that the "display" of a cable trace is constructed from HTML elements and so the "download" needs to be a completely different rendering pipeline. I'm hesitant to move forward with this PR for that reason alone; it might be worth revisiting this as part of the v3.0 UI refresh instead.

> Would it be advisable for my new code that I'll be working on to generate the SVG server-side so that it can be used for both display and download, avoiding issues with maintaining separate rendering pipelines?

glennmatthews:
> Would it be advisable for my new code that I'll be working on to generate the SVG server-side so that it can be used for both display and download, avoiding issues with maintaining separate rendering pipelines?

Personally I think that'd be desirable, but I'd welcome community feedback on such a proposed change.

PR #6898, opened 2025-02-14
3041 Assign parent bay during creation of Device

### LINKED ISSUE #3041: API creates device without feedback on inability to add parent device or device bay
### Environment
* Nautobot version (Docker tag too if applicable): 1.5.5
* Python version: 3.9.7
* Database platform, version: Postgresql 13.9
* Middleware(s):

### Steps to Reproduce
1. Create device in nautobot API using POST method, including valid parent device and device bay.
 
```
curl -X 'POST' \
  'https://nautobot.com/api/dcim/devices/' \
  -H 'X-CSRFTOKEN: $TOKEN' \
  -d
{
  "asset_tag": "my_tag",
  "cluster": {
    "name": "cluster_name
  },
  "custom_fields": {
    "field_1": "my_field",
    "field_2": "my_field",
    "field_3": "my_field"
  },
  "device_role": {
    "name": "my_role"
  },
  "device_type": {
    "manufacturer": {
      "name": "my_manufacturer"
    },
    "model": "my_model"
  },
  "name": "device_name",
  "parent_device": {
    "device_bay": {
      "name": "/23"
    },
    "name": "parent_device_name"
  },
  "site": {
    "name": "my_site"
  },
  "status": "active",
  "tenant": {
    "name": "tenant_name"
  }
}
```

2. View API response in which device is created, but parent device field is "null".

Response: 
```
{ 
...
"parent_device": null,
...
}
```

### Expected Behavior
Either device with device bay and parent is created, or feedback from the API that it could not be created.

### Observed Behavior
Device was created without the parent device and device bay.

mateusz-lachowski-codilime:
# Closes #3041
# What's Changed
Parent bay are now assigned during the creation of Device via API or CSV import.

However some cases might still need to be taken care of:
- while creating a device to a parent bay that is already taken, it will be overwritten and reassigned
- during the device update you can't reassign device to a new parent bay: you need to uninstall it first
# Screenshots

# TODO
- [ ] Unit, Integration Tests


nautobot/dcim/tests/test_api.py
```diff
@@ -1716,6 +1716,126 @@ def test_patching_device_redundancy_group(self):
...
+
+        response = self.client.post(url, data, format="json", **self.header)
+        self.assertHttpStatus(response, status.HTTP_201_CREATED)
+
+        # Assert that on the #1 device, parent bay stayed the same
+        created_device = Device.objects.get(name="Device parent bay test #3")
+        self.assertEqual(created_device.parent_bay.pk, device_bay_1.pk)
+
+        old_device = Device.objects.get(name="Device parent bay test #1")
+        self.assertNone(old_device.parent_bay)
```
HanlinMiao, line 1807: I do not know if automatically unassigning parent bay from the old device and assigning it to the new device is desirable here. What do you think? @glennmatthews

glennmatthews: I agree, it would be less surprising if this raised an exception instead IMO.

nautobot/dcim/api/serializers.py
```diff
@@ -563,8 +563,29 @@ def validate(self, attrs):
         # Enforce model validation
         super().validate(attrs)
 
+        # Validate parent bay
+        if self.instance and attrs.get("parent_bay"):
+            parent_bay = attrs.get("parent_bay", None)
```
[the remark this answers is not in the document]
glennmatthews: Or even:

```suggestion
        if self.instance and parent_bay := attrs.get("parent_bay", None):
```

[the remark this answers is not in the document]
mateusz-lachowski-codilime: TIL ` if self.instance and parent_bay := attrs.get("parent_bay", None):` it's incorrect :D

It throws: 'SyntaxError: cannot use assignment expressions with expression`

But anyway, I needed to apply additional validation to prevent auto unassigning so I rebuilt this fragment a bit.

nautobot/dcim/tests/test_api.py
```diff
@@ -1716,6 +1716,123 @@ def test_patching_device_redundancy_group(self):
...
+            "device_type": device_type_child.pk,
+            "role": parent_device.role.pk,
+            "location": parent_device.location.pk,
+            "name": "Device parent bay test #3",
+            "status": parent_device.status.pk,
+            "parent_bay": device_bay_1.pk,
+        }
+
+        response = self.client.post(url, data, format="json", **self.header)
+        self.assertHttpStatus(response, status.HTTP_400_BAD_REQUEST)
```
glennmatthews, line 1800: Would probably be good to check the response content and make sure it has the expected error message and isn't some unrelated error.

nautobot/dcim/tests/test_api.py
```diff
@@ -1716,6 +1716,123 @@ def test_patching_device_redundancy_group(self):
...
+        response = self.client.patch(self._get_detail_url(child_device), patch_data, format="json", **self.header)
+        self.assertHttpStatus(response, status.HTTP_200_OK)
+
+        updated_device = Device.objects.get(name="Device parent bay test #4")
+        self.assertEqual(updated_device.parent_bay.pk, device_bay_1.pk)
+
+        # Changing the parent bay is not allowed without removing it first
+        patch_data = {"parent_bay": device_bay_2.pk}
+        response = self.client.patch(self._get_detail_url(child_device), patch_data, format="json", **self.header)
+        self.assertHttpStatus(response, status.HTTP_400_BAD_REQUEST)
```
glennmatthews, line 1830: Same comment here.

glennmatthews (approved):

nautobot/dcim/tests/test_api.py
HanlinMiao: https://github.com/nautobot/nautobot/pull/6898/commits/f4032c72079d90e47a96286552a1849339e4d49a

nautobot/dcim/tests/test_api.py
HanlinMiao: https://github.com/nautobot/nautobot/pull/6898/commits/f4032c72079d90e47a96286552a1849339e4d49a

nautobot/dcim/tests/test_api.py
```diff
@@ -1716,6 +1716,128 @@ def test_patching_device_redundancy_group(self):
...
+        self.assertIn(
+            f"Cannot install the specified device; device is already installed in {device_bay_1.name}",
+            response.content.decode(response.charset),
+        )
+
+        # Assert that parent bay stayed the same
+        updated_device = Device.objects.get(name="Device parent bay test #4")
+        self.assertEqual(updated_device.parent_bay.pk, device_bay_1.pk)
```
glennmatthews, line 1839: I'm a bit surprised by this part of the behavior actually - is the intent that the user needs to explicitly remove the device from the current parent bay as a separate prerequisite step? That feels non-obvious to me, but if it's what we agree on, let's at least make sure to have a test or two that covers that successful workflow.

mateusz-lachowski-codilime: This whole behavior strictly comes from `DeviceBay.clean()` method. It looks like we never had a way to update `DeviceBay` through `Device` directly and because of that we have not covered such cases.

For me it's also non-obvious so maybe I will just update this.

nautobot/dcim/api/serializers.py
```diff
@@ -560,11 +561,47 @@ def validate(self, attrs):
...
+        update_parent_bay = "parent_bay" in validated_data.keys()
+        parent_bay = validated_data.get("parent_bay")
+        if update_parent_bay:
+            if parent_bay:
+                parent_bay.installed_device = instance
+                parent_bay.save()
+            elif instance.parent_bay:
+                parent_bay = instance.parent_bay
+                parent_bay.installed_device = None
+                # instance.parent_bay.installed_device = None
```
glennmatthews, line 602: remove?

nautobot/dcim/api/serializers.py
```diff
@@ -560,11 +561,47 @@ def validate(self, attrs):
...
+            if parent_bay.installed_device and parent_bay.installed_device != self.instance:
+                raise ValidationError(
+                    {
+                        "installed_device": f"Cannot install device; parent bay is already taken ({parent_bay.installed_device})"
+                    }
+                )
+
+            if self.instance:
+                parent_bay.installed_device = self.instance
+                parent_bay.full_clean()
```
glennmatthews, line 575: Is this right to do in `validate()`? And is it redundant with `update_parent_bay()` below?

nautobot/dcim/api/serializers.py
```diff
@@ -560,11 +561,47 @@ def validate(self, attrs):
...
+        parent_bay = validated_data.get("parent_bay")
+        if update_parent_bay:
+            if parent_bay:
+                parent_bay.installed_device = instance
+                parent_bay.save()
+            elif instance.parent_bay:
+                parent_bay = instance.parent_bay
+                parent_bay.installed_device = None
+                # instance.parent_bay.installed_device = None
+                parent_bay.save()
```
glennmatthews, line 603: Perhaps `validated_save()`?

nautobot/dcim/tests/test_api.py
```diff
@@ -1716,6 +1716,181 @@ def test_patching_device_redundancy_group(self):
...
+        self.assertHttpStatus(response, status.HTTP_200_OK)
+        self.assertEqual(response.json()["parent_bay"]["id"], str(device_bay_1.pk))
+
+        # Test unassigning parent bay
+        patch_data = {"parent_bay": None}
+        response = self.client.patch(child_device_detail_url, patch_data, format="json", **self.header)
+        self.assertHttpStatus(response, status.HTTP_200_OK)
+
+        updated_device = Device.objects.get(name=device_name)
+        self.assertFalse(hasattr(updated_device, "parent_bay"))
```
glennmatthews, line 1868: Is this valid, or would it be better to do something like:

```py
with self.assertRaises(DeviceBay.DoesNotExist):
    updated_device.parent_bay
```

nautobot/dcim/tests/test_api.py
```diff
@@ -1716,6 +1716,181 @@ def test_patching_device_redundancy_group(self):
...
+        response = self.client.get(child_device_detail_url, **self.header)
+        self.assertHttpStatus(response, status.HTTP_200_OK)
+        self.assertEqual(response.json()["parent_bay"]["id"], str(device_bay_1.pk))
+
+        # Test unassigning parent bay
+        patch_data = {"parent_bay": None}
+        response = self.client.patch(child_device_detail_url, patch_data, format="json", **self.header)
+        self.assertHttpStatus(response, status.HTTP_200_OK)
+
+        updated_device = Device.objects.get(name=device_name)
```
glennmatthews, line 1867: Could also do `child_device.refresh_from_db()` if preferred.

glennmatthews:
I filed https://github.com/mkdocstrings/autorefs/issues/68 for the mkdocs-autorefs error on RTD. We should probably pin that package in `docs/requirements.txt` as well in the meantime.

nautobot/dcim/api/serializers.py
mateusz-lachowski-codilime: I need to run `full_clean` in `validate` to not cause 500 by `ValidationError`. And from the other hand `full_clean` logic require `installed_device`... This case is intended as extra check for `update` scenarios. Maybe I should add refresh from db after this step to remove assigned instance and rely only on logic from `update` method?

glennmatthews: Modifying `parent_bay.installed_device` in a validate() method was my concern but since we're not calling `parent_bay.save()` it's *maybe* okay as-is?

mateusz-lachowski-codilime: I have couple more ideas how to tackle this:
- add reverse foreign key on Device side and then some validation logic there and set parent_bay on Device directly
- move logic outside of `clean()` method, call it there and here in `validate()` method just call it like `parent_bay.validate_can_install_device(self.instance)`
- leave it as it is, but add some todo here, to fix it in future?

glennmatthews: Let's go with that last option for now. Thanks

PR #6919, opened 2025-02-20
Closes #6848: Fixed DoesNotExist in GUI at /extras/git-repositories/X/result/ if git repository created via API and not yet synced

### LINKED ISSUE #6848: DoesNotExist in GUI at /extras/git-repositories/X/result/ if git repository created via API and not yet synced
### Environment
* Nautobot version (Docker tag too if applicable): 2.4.1
* Python version: 3.12.3
* Database platform, version: PostgreSQL 14.15
* Middleware(s): -

### Steps to Reproduce
1. Create a new git repo via the API, e.g.:
```
curl -s -X POST \
-H "Authorization: Token <SNIPPED>" \
-H "Accept: application/json; version=2.4" \
http://localhost:8080/api/extras/git-repositories/ \
--data '{"name": "minimal-git-datasource", "provided_contents": ["extras.job"], "remote_url": "https://github.com/cdtomkins/nautobot-git-demo.git"}'
```
2. In the GUI, head to `Extensibility/Data Sources/Git Repositories`.
3. Click the name of your new repo.
4. Click `Synchronization Status`.

### Expected Behavior
I would expect to see something similar to the normal `Summary of Results/Logs` that are returned at `Synchronization Status`, except obviously more minimal, perhaps explicitly informing the user that a sync has not yet occurred.

### Observed Behavior
If you have debugging on, you'll see:
```
DoesNotExist at /extras/git-repositories/X/result/

JobResult matching query does not exist.
```
The cause here is that the GUI code assumes that any new repo will have had at least one sync attempt, because that's what happens when a repo is created in the GUI. The same assumption is not valid for a repo created via the API, though.

cdtomkins:
# Closes #6848 
# What's Changed
- If a git repository is created via the API and has not yet synced but someone tries to view the `Synchronization Status`, a `DoesNotExist` error no longer occurs.
- Also as flagged by Glenn, given that JobResults are generally deletable, this fix would apply to the case of all results associated with a given GitRepository being deleted as well.
- Fixed cosmetic typo `"specifiy"` elsewhere in code and a UT.

# Screenshots
# Successful view of the page that would previously have errored:
Screenshot 2025-02-21 145913

# Normal view still unaffected after a sync:
Screenshot 2025-02-20 143039

-->
# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/changes/6848.fixed
+++ b/changes/6848.fixed
@@ -1,0 +1,1 @@
+Fixed a `DoesNotExist` error in the GUI at `/extras/git-repositories/X/result/` when a Git repository was created via API and not yet synced.
--- a/nautobot/extras/templates/extras/inc/jobresult.html
+++ b/nautobot/extras/templates/extras/inc/jobresult.html
@@ -12,45 +12,43 @@
         <strong>Summary of Results</strong>
     </div>
     <table class="table table-hover panel-body">
-        {% if result.job_model is not None %}
             <tr>
                 <td>Job Description</td>
-                <td>{{ result.job_model.description | render_markdown }}</td>
+                <td>{{ result.job_model.description | render_markdown | placeholder }}</td>
             </tr>
-        {% endif %}
         <tr>
             <td>Status</td>
             <td><span id="pending-result-label">{% include 'extras/inc/job_label.html' with result=result %}</span></td>
         </tr>
         <tr>
             <td>Started at</td>
-            <td>{{ result.date_created }}</td>
+            <td>{{ result.date_created | placeholder }}</td>
         </tr>
         <tr>
             <td>User</td>
-            <td>{{ result.user }}</td>
+            <td>{{ result.user | placeholder }}</td>
         </tr>
         <tr>
             <td>Duration</td>
             <td>
-            {% if result.date_done %}
-                {{ result.duration }}
-            {% else %}
+            {% if result.date_created and not result.date_done %}
                 <img src="{% static 'img/ajax-loader.gif' %}">
+            {% else %}
+                {{ result.duration | placeholder}}
             {% endif %}
             </td>
         </tr>
         <tr>
             <td>Return Value</td>
             <td>
-            {% if result.date_done %}
+            {% if result.date_created and not result.date_done %}
+                <img src="{% static 'img/ajax-loader.gif' %}">
+            {% else %}
                 {% if result.result %}
                     <pre>{{ result.result | render_json }}</pre>
                 {% else %}
                     {{ result.result | placeholder }}
                 {% endif %}
-            {% else %}
-                <img src="{% static 'img/ajax-loader.gif' %}">
             {% endif %}
             </td>
         </tr>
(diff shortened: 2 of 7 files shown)


cdtomkins:
This is ready for review.

changes/6848.fixed
```diff
@@ -0,0 +1 @@
+Fixed DoesNotExist in GUI at /extras/git-repositories/X/result/ if git repository created via API and not yet synced.
```
glennmatthews, line 1:
```suggestion
Fixed a `DoesNotExist` error in the GUI at `/extras/git-repositories/X/result/` when a Git repository was created via API and not yet synced.
```

nautobot/extras/views.py
```diff
@@ -1165,6 +1165,17 @@ def get_required_permission(self):
...
+        if job_result is None:
+            job_result = {
+                "completed": "—",
+                "date_created": "—",
+                "date_done": "—",
+                "duration": "—",
+                "name": "—",
+                "status": "N/A",
+                "user": "—",
+            }
```
glennmatthews, line 1177: Can we add a view unit test case covering the originally described issue with this view?

HanlinMiao, line 1177: Probably should be an empty dictionary here and use the `placeholder` template tag in the template for these fields? For example `{{ result.duration | placeholder }}`

```suggestion
            job_result = {}
```

HanlinMiao (approved):
Thank you for the PR!

nautobot/extras/views.py
cdtomkins: Done. This worked fine, thanks.

nautobot/extras/templates/extras/inc/jobresult.html
```diff
@@ -12,45 +12,43 @@
...
         <tr>
             <td>Return Value</td>
             <td>
-            {% if result.date_done %}
+            {% if result.date_created and not result.date_done %}
+                <img src="{% static 'img/ajax-loader.gif' %}">
+            {% else %}
                 {% if result.result %}
                     <pre>{{ result.result | render_json }}</pre>
                 {% else %}
```
cdtomkins, line 49: This conditional was here before my changes. In my testing, it seems unnecessary, but I have left it in case there's some scenario I'm not expecting.

nautobot/extras/templates/extras/inc/jobresult.html
```diff
@@ -12,45 +12,43 @@
...
                     {{ result.result | placeholder }}
                 {% endif %}
```
cdtomkins, line 51: This conditional was here before my changes. In my testing, it seems unnecessary, but I have left it in case there's some scenario I'm not expecting.

glennmatthews: Looking at it I **think** it's just so that the placeholder doesn't get unnecessarily wrapped in a `<pre>` tag, but that may not be an issue in practice.

cdtomkins: Yeah my experience when I tested it was that it was not an issue when it did. But I'll leave it nonetheless.

nautobot/extras/views.py
cdtomkins: Done! Let me know what you think.

glennmatthews (approved):

PR #6921, opened 2025-02-20
Fix some UI framework issues (#6920)

### LINKED ISSUE #6920: Filters have to be explicitly defined to be used in UI framework
### Environment
* Nautobot version (Docker tag too if applicable): 2.4.2
* Python version: 3.11
* Middleware(s):

### Steps to Reproduce
1. Create a new model with necessary items
2. Create the filterset inheriting from `NautobotFilterSet` with `fields = "__all__"` under `Meta`
      ```
      class ExampleModelFilterSet(NautobotFilterSet):
      
          class Meta:
              """Meta attributes for filter."""
      
              model = models.ExampleModel
              fields = "__all__"
      ```
4. Create a StatsPanel in a ViewSet using `ObjectDetailContent` from UI framework that is showing stats using the above model and filterset.
      ```
      object_detail_content = ObjectDetailContent(
              panels=[
                 StatsPanel(
                      weight=800,
                      section=SectionChoices.LEFT_HALF,
                      label="Record Statistics",
                      filter_name="name",
                      related_models=[ ExampleModel],
                  ),
      ]
      ```

### Expected Behavior
I was expecting this to work and show statistics about the ExampleModel in the viewset that has the `StatsPanel`. 

### Observed Behavior
Thrown an error that any field is not a valid filter field, even though it is a valid option.
`name key is not a valid filter field for Example model` 

Noticed that if you explicitly specify the field in the `FilterSet` everything works as expected.

glennmatthews:
# Closes #6920
# What's Changed

- When rendering "extra_buttons" in the UI framework:
  - Use the context `object_detail_content` if provided, only fail back to looking up the detail view by name if needed
  - If getting the context `obj`/`object` fails, or if looking up the detail view by name fails, log helpful messages to aid in troubleshooting
- Use `FilterSet.get_filters()` instead of accessing `FilterSet.declared_filters` in a couple of cases, as the latter only includes explicitly declared filter fields, while the former also includes those implicitly generated by `fields = "__all__"`.

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/6920.fixed
+++ b/changes/6920.fixed
@@ -1,0 +1,3 @@
+Added log messages to help troubleshoot failures when rendering UI Component Framework `extra_buttons`.
+Fixed logic in UI Component Framework `StatsPanel` that incorrectly disallowed use of FilterSet filters implicitly defined through `fields = "__all__"`.
+Fixed logic in generic `FormTestCase` class that would incorrectly fail if form fields used FilterSet filters implicitly defined through `fields = "__all__"`.
--- a/nautobot/core/templatetags/ui_framework.py
+++ b/nautobot/core/templatetags/ui_framework.py
@@ -1,8 +1,12 @@
+import logging
+
 from django import template
 from django.utils.html import format_html_join
 
 from nautobot.core.utils.lookup import get_view_for_model
 from nautobot.core.views.utils import get_obj_from_context
+
+logger = logging.getLogger(__name__)
 
 register = template.Library()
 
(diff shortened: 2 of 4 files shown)


changes/6920.fixed
```diff
@@ -0,0 +1,3 @@
+Added log messages to help troubleshoot failures when rendering UI Component Framework `extra_buttons`.
```
HanlinMiao, line 1: Should this be a separate changelog fragment?

HanlinMiao (approved):

changes/6920.fixed
glennmatthews: Were you thinking `.added` for this? I consider it a "fix" rather than a "feature", but I'm persuadable.

HanlinMiao: Yeah, I was thinking `.added` for this. It is not technically something that was `missing` but an enhancement to the previous iteration. Definitely not blocking.

PR #6934, opened 2025-02-24
Fixes bug #5193

### LINKED ISSUE #5193: Unable to associate tags to config context via API
### Environment
* Nautobot version v2.1.2 (demo.nautobot.com), but also reproducible in v2.0.4

### Steps to Reproduce

1. Create a new config context
2. Issue the following curl command (POST, PUT, PATCH have the same error)
```
curl -X 'POST' \
  'https://demo.nautobot.com/api/extras/config-contexts/' \
  -H 'X-CSRFTOKEN: Y9fyq3moe0OLDAAnm8GQtNj7j9vbPax6FR5n3snvjxCBTysbqAZk4anbdJ7EajRn' \
  -d '{
  "name": "test_context",
  "data": {
    "test": "string"
  },
  "tags": [
    {
      "id": "91283c9c-e2ba-4cfe-8f9f-ed6dd6e18ba1"
    }
  ]
}'
```

### Expected Behavior
Tags specified should be associated with the config context. This does not happen for any other relation (locations, roles, etc) only tags.
**Note that this works correctly if done via the GUI**

### Observed Behavior
Error is returned from the API:
```
{
  "error": "['“sup-720” is not a valid UUID.']",
  "exception": "ValidationError",
  "nautobot_version": "2.1.2",
  "python_version": "3.11.7"
}
```

jpvrenen:
# Closes #5193
# What's Changed
Fixed an erroneous `ValidationError` when attempting to apply Tags to an object via the REST API.

# TODO

- [x] Unit, Integration Tests


glennmatthews (commented):
Change looks correct, can we add a unit test covering this fix?

jpvrenen:
> Change looks correct, can we add a unit test covering this fix?

@glennmatthews Hi Glenn, I am not really sure how or where to add the test. Does that go into the 'create_data' variable of the class ConfigContextTest in test_api.py?

glennmatthews:
Sorry for the terseness of my previous response. Yes, you have it exactly right. Most likely to have it work correctly, you'll need to move the definition of `create_data` inside the `setUpTestData()` method so that it can reference existing Tag objects at runtime rather than at class declaration time, something like:

```py
    @classmethod
    def setUpTestData(cls):
        ConfigContext.objects.create(name="Config Context 1", weight=100, data={"foo": 123})
        ConfigContext.objects.create(name="Config Context 2", weight=200, data={"bar": 456})
        ConfigContext.objects.create(name="Config Context 3", weight=300, data={"baz": 789})
        cls.create_data = [
            {
                "name": "Config Context 4",
                "data": {"more_foo": True},
                "tags": Tag.objects.get_for_model(Device),   # <----------<<
            },
            {
                "name": "Config Context 5",
                "data": {"more_bar": False},
            },
            {
                "name": "Config Context 6",
                "data": {"more_baz": None},
            },
        ]
```

(as an aside, that existing `create_data` is woefully sparse - we should definitely add to it at some point!)

HanlinMiao (commented):
we probably need a changelog for this as well named `6934.fixed`. You can look at the fragments that exist in the `/changes` directory for some examples.

jpvrenen:
Added the 5193.fixed changelog to the /changes directory.

jpvrenen:
@glennmatthews  I have added a unittest for creating ConfigContext using 'tag' but had to make a change to nautobot/core/testing/mixins.py. Not sure if that fix has to be handled differently. Basically the validation 'assertInstanceEqual' of the api input with 'tags' as 'str' and the actual ConfigContext instance with 'tags' as UUID.

Before the fix, test outcome:
ConfigContextTest_create_objects_pre-fix_fail_2025-02-27 134825

After the fix, test outcome:
ConfigContextTest_create_objects_post-fix_success_2025-02-27 105134

nautobot/extras/tests/test_api.py
```diff
@@ -220,6 +206,21 @@ def setUpTestData(cls):
         ConfigContext.objects.create(name="Config Context 1", weight=100, data={"foo": 123})
         ConfigContext.objects.create(name="Config Context 2", weight=200, data={"bar": 456})
         ConfigContext.objects.create(name="Config Context 3", weight=300, data={"baz": 789})
+        cls.create_data = [
+            {
+                "name": "Config Context 4",
+                "data": {"more_foo": True},
+                "tags": [str(tag.pk) for tag in Tag.objects.get_for_model(Device)]
```
glennmatthews, line 213: If you change this to `[tag.pk` instead of `[str(tag.pk)` does that avoid the need to update the base mixin class? I see that pattern in a few other existing tests in this file.

changes/5193.fixed
```diff
@@ -0,0 +1 @@
+The _save_tags method now uses the 'tags' object list as it is already a list of Tag objects.
```
glennmatthews, line 1: Perhaps:

```suggestion
Fixed an erroneous `ValidationError` when attempting to apply Tags to an object via the REST API.
```

nautobot/extras/tests/test_api.py
jpvrenen: Yes you are absolutely correct, thank you for pointing that out! As you pointed out this does avoid the need to update the base mixin class!

glennmatthews (approved):

PR #6936, opened 2025-02-24
Added optional VRF relationship to Virtual Device Context.

### LINKED ISSUE #6924: Optional VRF relationship to VDC
### As ...

### I want ...

A VDC to support an optional relationship to an assigned VRF 

### So that ...

I can model more realistic multi-tenancy use cases where tenant customer traffic is segregated by both the usage of VDC and VRF topology.

- A new relation is added between VDC and VRF such that a VDC may interact with zero or more VRFs, in the same way that VRFs may related to Devices and Virtual Machines today (see `VRFDeviceAssignment` model)
- No change to the way that VRFs already relate to devices and VM. i.e. the same VRF may relate to a device and VDCs on that device, or it may not.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [x] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

- New field on `VRFDeviceAssignment` to allow for relation to VDC. Update constraints and clean logic to account for new assignment option.

### External Dependencies

_No response_

HanlinMiao:
# Closes #6924 
# What's Changed

Added optional VRF relationship to Virtual Device Context.
# Screenshots

VDC Detail View:
Screenshot 2025-02-24 at 4 41 18 PM
VDC Update View:
Screenshot 2025-02-24 at 4 41 07 PM
VRF Detail View:
Screenshot 2025-02-24 at 4 25 27 PM
VRF Update View:
Screenshot 2025-02-24 at 4 43 34 PM

# TODO

- [x] Unit, Integration Tests


nautobot/dcim/forms.py
```diff
@@ -5298,6 +5298,11 @@ class VirtualDeviceContextForm(NautobotModelForm):
         required=True,
         query_params={"content_types": VirtualDeviceContext._meta.label_lower},
     )
+    vrfs = DynamicModelMultipleChoiceField(
```
glennmatthews, line 5301: Do we want an add/remove vrfs option on the corresponding bulk-edit form?

nautobot/dcim/forms.py
```diff
@@ -5323,11 +5329,15 @@ def __init__(self, *args, **kwargs):
             self.fields["device"].disabled = True
             self.fields["device"].required = False
 
+        self.initial["vrfs"] = self.instance.vrfs.values_list("id", flat=True)
```
glennmatthews, line 5332: Do we need something similar for `initial["interfaces"]`? If not, why not?

nautobot/dcim/templates/dcim/virtualdevicecontext_retrieve.html
```diff
@@ -65,4 +65,5 @@
 
 {% block content_full_width_page %}
     {% include 'panel_table.html' with table=interfaces_table heading="Interfaces" %}
+    {% include 'panel_table.html' with table=vrf_table heading="Assigned VRFs" %}
```
glennmatthews, line 68: How much work would it be to convert this view to the UI framework while you're in here?

nautobot/ipam/forms.py
```diff
@@ -111,6 +111,9 @@ class VRFForm(NautobotModelForm, TenancyForm):
     namespace = DynamicModelChoiceField(queryset=Namespace.objects.all())
     devices = DynamicModelMultipleChoiceField(queryset=Device.objects.all(), required=False)
     virtual_machines = DynamicModelMultipleChoiceField(queryset=VirtualMachine.objects.all(), required=False)
+    virtual_device_contexts = DynamicModelMultipleChoiceField(
```
glennmatthews, line 114: Should we have add/remove equivalents to this in the bulk-edit form?

nautobot/ipam/migrations/0051_added_optional_vrf_relationship_to_vdc.py
```diff
@@ -0,0 +1,41 @@
...
+        ("virtualization", "0030_alter_virtualmachine_local_config_context_data_owner_content_type_and_more"),
+        ("dcim", "0067_controllermanageddevicegroup_tenant"),
+        ("ipam", "0050_vlangroup_range"),
+    ]
+
+    operations = [
+        migrations.AlterUniqueTogether(
+            name="vrfdeviceassignment",
+            unique_together={("vrf", "device"), ("vrf", "virtual_machine")},
+        ),
```
glennmatthews, line 18: Why this?

nautobot/ipam/models.py
```diff
@@ -289,14 +333,15 @@ class Meta:
         unique_together = [
             ["vrf", "device"],
             ["vrf", "virtual_machine"],
+            ["vrf", "virtual_device_context"],
             # TODO: desirable in the future, but too strict for 1.x-to-2.0 data migrations,
             #       as multiple "cleanup" VRFs in different cleanup namespaces might be assigned to a single device/VM.
             # ["device", "rd", "name"],
             # ["virtual_machine", "rd", "name"],
```
glennmatthews, line 340: Wonder if we should go ahead and add `virtual_device_context, rd, name` here, or even go ahead and enable that constraint proactively?

nautobot/ipam/models.py
```diff
@@ -135,6 +135,12 @@ class VRF(PrimaryModel):
         through="ipam.VRFDeviceAssignment",
         through_fields=("vrf", "virtual_machine"),
     )
+    virtual_device_contexts = models.ManyToManyField(
```
glennmatthews, line 138: Can we augment the relevant factory(s) to populate this data?

nautobot/dcim/forms.py
HanlinMiao: Device/VM BulkEditForm are missing an add/remove vrfs option as well. Do you want me to add those in this PR or follow up issue is fine?

nautobot/dcim/forms.py
HanlinMiao: Seems like the initial["interfaces"] is populated correctly. Is it because `interfaces` on VDC is a forward m2m relationship?

glennmatthews: Could be, yeah - forward vs reverse relationship.

nautobot/dcim/forms.py
glennmatthews: Let's get a followup issue for those just to keep the scope manageable here.

nautobot/ipam/migrations/0051_added_optional_vrf_relationship_to_vdc.py
HanlinMiao: Just did a `nautobot-server makemigrations` here so it is autogenerated. I will delete it and see how it goes.

nautobot/dcim/templates/dcim/virtualdevicecontext_retrieve.html
HanlinMiao: I refactored it here. But I wonder if we should leave this for the QA team in the future.

nautobot/ipam/models.py
HanlinMiao: Going to add a followup issue to augment VRF factory to include `devices` and `virtual_machines` @post_generation as well

nautobot/ipam/models.py
HanlinMiao: Added the constraint and added a test case

HanlinMiao: https://github.com/nautobot/nautobot/pull/6936/commits/d0782f498af14906fcf2c02303b001d274e98c95

nautobot/dcim/forms.py
HanlinMiao: https://github.com/nautobot/nautobot/issues/6938

nautobot/ipam/models.py
HanlinMiao: https://github.com/nautobot/nautobot/issues/6938

nautobot/ipam/forms.py
HanlinMiao: https://github.com/nautobot/nautobot/issues/6938

nautobot/ipam/models.py
```diff
@@ -310,11 +356,19 @@ def clean(self):
...
+        if all([self.device, self.virtual_device_context]):
+            raise ValidationError("A VRF cannot be associated with both a device and a virtual device context.")
+        if all([self.virtual_machine, self.virtual_device_context]):
+            raise ValidationError(
+                "A VRF cannot be associated with both a virtual machine and a virtual device context."
+            )
+        if not any([self.device, self.virtual_machine, self.virtual_device_context]):
+            raise ValidationError(
+                "A VRF must be associated with a device, a virtual machine, or a virtual device context."
+            )
```
glennmatthews, line 371: I feel like these error messages are a bit unclear/misleading - e.g. a VRF **can** be associated with both a device and a virtual-device-context, it just requires two separate entries in the through table. Similarly a VRF does not **have** to be associated with a device, VM, or VDC, it's just that if you're adding an entry in the through table, it needs one of those.

nautobot/ipam/tests/test_api.py
```diff
@@ -154,25 +163,45 @@ def setUpTestData(cls):
...
         )
+        VRFDeviceAssignment.objects.create(
+            vrf=cls.vrfs[0],
+            virtual_device_context=cls.vdcs[0],
+            rd="65000:5",
+        )
+        VRFDeviceAssignment.objects.create(
+            vrf=cls.vrfs[0],
+            virtual_device_context=cls.vdcs[1],
+            rd="65000:6",
```
glennmatthews, line 174: Let's make sure we have diversity in the test data - one of these should probably set both `name` and `rd`, and the other should omit both?

nautobot/ipam/tests/test_api.py
```diff
@@ -154,25 +163,45 @@ def setUpTestData(cls):
...
+        )
 
         cls.create_data = [
             {
                 "vrf": cls.vrfs[2].pk,
                 "device": cls.devices[4].pk,
                 "virtual_machine": None,
-                "rd": "65000:4",
+                "virtual_device_context": None,
```
glennmatthews, line 182: I would hope that in the "create" API data we can just omit the null values, rather than needing to explicitly set them to None?

nautobot/ipam/tests/test_api.py
```diff
@@ -154,25 +163,45 @@ def setUpTestData(cls):
...
-                "rd": "65000:6",
+                "virtual_device_context": None,
+                "rd": "65000:9",
+            },
+            {
+                "vrf": cls.vrfs[4].pk,
+                "device": None,
+                "virtual_machine": None,
+                "virtual_device_context": cls.vdcs[0].pk,
+                "rd": "65000:10",
```
glennmatthews, line 204: Here too let's make sure the test covers both providing and omitting the fields (name, rd) that should be optional.

nautobot/ipam/tests/test_views.py
```diff
@@ -94,6 +103,8 @@ def setUpTestData(cls):
             "namespace": prefixes[0].namespace.id,
             "add_prefixes": [prefixes[0].id],
             "remove_prefixes": [prefixes[1].id],
+            "add_virtual_device_contexts": [vdcs[0].id, vdcs[1].id],
```
glennmatthews, line 106: Shouldn't we also be augmenting the `form_data` with a VDC as well for this test?

nautobot/ipam/models.py
```diff
@@ -238,6 +244,41 @@ def remove_virtual_machine(self, virtual_machine):
         instance = self.virtual_machines.through.objects.get(vrf=self, virtual_machine=virtual_machine)
         return instance.delete()
 
+    def add_virtual_device_context(self, virtual_device_context, rd="", name=""):
```
glennmatthews, line 247: This seems wrong to me - shouldn't `rd` and `name` default to `None`, not `""`? The latter will cause problems with the `(vdc, rd, name)` uniqueness constraint.

(edit) Actually it's worse than that, since `rd` is `null=True` but `name` is not. What gives? 🤢

nautobot/ipam/models.py
```diff
@@ -289,14 +333,16 @@ class Meta:
         unique_together = [
             ["vrf", "device"],
             ["vrf", "virtual_machine"],
+            ["vrf", "virtual_device_context"],
+            ["virtual_device_context", "rd", "name"],
```
glennmatthews, line 337: Changing my mind here due to the previous comment - let's leave this constraint as a TODO for now. I think if anything it should probably be two separate constraints, i.e `[vdc, rd]` and `[vdc, name]` but let's not try to solve that at the moment.

nautobot/ipam/tables.py
```diff
@@ -255,13 +255,18 @@ class VRFDeviceAssignmentTable(BaseTable):
         accessor="virtual_machine.name",
         verbose_name="Virtual Machine",
     )
+    virtual_device_context = tables.Column(
```
glennmatthews, line 258: Not critical, but I wonder if it'd be better UX to have two columns `related object type` and `related object name` instead of three columns `device`/`virtual_machine`/`virtual_device_context` that are mutually exclusive in their data?

nautobot/dcim/views.py
```diff
@@ -4513,15 +4513,25 @@ class VirtualDeviceContextUIViewSet(NautobotUIViewSet):
...
+            object_detail.ObjectsTablePanel(
+                weight=200,
+                table_class=tables.InterfaceTable,
+                table_attribute="interfaces",
+                section=SectionChoices.FULL_WIDTH,
+                exclude_columns=["device"],
+            ),
+            object_detail.ObjectsTablePanel(
+                weight=300,
+                table_class=VRFTable,
```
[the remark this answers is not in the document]
HanlinMiao: VRF detail view has a VRFDeviceAssignmentTable and that includes the VDC/VM/Device
Screenshot 2025-02-25 at 2 14 21 PM

nautobot/ipam/tables.py
HanlinMiao: Added in https://github.com/nautobot/nautobot/pull/6936/commits/d5efbd8ff23602a85a79b73ae0edbe57e30860f1
Screenshot 2025-02-25 at 2 25 27 PM

nautobot/ipam/filters.py
```diff
@@ -124,10 +129,15 @@ class VRFDeviceAssignmentFilterSet(NautobotFilterSet):
         to_field_name="name",
         label="Virtual Machine (ID or name)",
     )
+    virtual_device_context = NaturalKeyOrPKMultipleChoiceFilter(
```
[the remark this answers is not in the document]
HanlinMiao: Looks like we are missing VRFDeviceAssignment test case. Will add

nautobot/dcim/views.py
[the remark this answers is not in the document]
glennmatthews: Makes sense. Mea culpa.

nautobot/ipam/tables.py
```diff
@@ -247,26 +247,35 @@ class VRFDeviceAssignmentTable(BaseTable):
...
-        verbose_name="Virtual Device Context",
+    related_object_name = tables.TemplateColumn(
+        template_code="""
+        {% if record.device %}
+            <a href="{{ record.device.get_absolute_url }}">{{ record.device.name }}</a>
+        {% elif record.virtual_machine %}
+            <a href="{{ record.virtual_machine.get_absolute_url }}">{{ record.virtual_machine.name }}</a>
+        {% else %}
+            <a href="{{ record.virtual_device_context.get_absolute_url }}">{{ record.virtual_device_context.name }}</a>
+        {% endif %}
```
glennmatthews, line 269: Would something like this work?

```
{% load helpers %}
{{ record.device or record.virtual_machine or record.virtual_device_context | hyperlinked_object }}
```

nautobot/ipam/filters.py
```diff
@@ -134,6 +144,8 @@ class VRFDeviceAssignmentFilterSet(NautobotFilterSet):
         to_field_name="name",
         label="Virtual Device Context (ID or name)",
     )
+    name = MultiValueCharFilter(label="Name")
+    rd = MultiValueCharFilter(label="Route Distinguisher")
```
glennmatthews, line 148: Maybe add these to `fields` instead of explicitly defining them?

nautobot/ipam/tables.py
HanlinMiao: It did not work here. I also tried using `with` tag, it didn't work either
```
{% load helpers %}
        {% with related_object=record.device or record.virtual_machine or record.virtual_device_context %}
            {{ related_object | hyperlinked_object }}
        {% endwith %}
```

nautobot/ipam/models.py
glennmatthews: I guess let's leave this as-is for now since it matches the device and VM APIs already existing, but maybe open a followup item?

HanlinMiao: https://github.com/nautobot/nautobot/issues/6945

PR #6896, opened 2025-02-14
Optional development setting to utilize dynamic docker ports

msheiny:
# Closes #6893
# What's Changed

* Adds optional setting to utilize ephemeral docker ports for all nautobot docker containers. This is controlled via invoke configuration so can either be overwritten in `invoke.yml` file locally, or an env var *(`INVOKE_NAUTOBOT_EPHEMERAL_PORTS=True`)
* Updates the VSCode workspace launch configs to be able to hit the ephemeral `debugpy` ports. This requires a new VSCode extension [rioj7.command-variable](https://marketplace.visualstudio.com/items?itemName=rioj7.command-variable)
* Adds a task that launches upon first entering the VSCode workspace to make sure that these users have the `docker-compose.vscode-rdb.yml` over-ride configured.
* Adds invoke commands to drive a web-browser to the `mkdocs` and `nautobot` web interfaces

# Discussion points
Notice that this feature is disabled by default to not ruffle feathers of users whom prefer static ports. The only potential controversial change is that if you are a VSCode user of the workspace file, this new extension `rioj7.command-variable` is required in order to run the launch configurations because of the potential for dynamic ports on the debugpy process. Since I added it to the recommended extensions, it should prompt users to install it AND it's ONLY required for VSCode users that want to utilize the remote debug launch commands. I didn't see a great way around this but willing to do some more research if that's going to be a tough pill for us to require this new extension in this scenario.

# TODO

- [x] Explanation of Change(s)
- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/core/#creating-changelog-fragments))
- [x] Documentation Updates (when adding/changing features)
- [x] Outline Remaining Work, Constraints from Design

PR #6943, opened 2025-02-26
Revert debug task behavior, modify ephemeral port to be async

msheiny:
# What's Changed

As a follow-up to [https://github.com/nautobot/nautobot/pull/6896](https://github.com/nautobot/nautobot/pull/6896) I realized I changed the behavior of the original `debug` task to make it different in a way developers might not appreciate/agree with/enjoy.

So I've reverted back to the old behavior and modifying the port detection logic in a way that allows it to run concurrently in parallel.

Previous behavior:
1. Launch `inv debug`
2. Containers start in foreground with logs streaming from all containers
3. Ctrl+C to exit stops logs streaming and kills containers

Current behavior:
1. launch `inv debug`
2. Containers start in background
3. Port mappings wrote to disk
4. Container logs followed again
5. Ctrl+C will stop following logs BUT containers still remain active

New behavior:
1. launch `inv debug`
2. Port mapping detection starts in background thread, looping until the required services are detected
3. Containers start in foreground as previous behavior
2. Ctrl+C will stop following logs and kill active containers 

# Screenshots

# TODO

### DIFF
--- a/changes/6943.housekeeping
+++ b/changes/6943.housekeeping
@@ -1,0 +1,1 @@
+Mimic `debug` invoke task to previous behavior.
--- a/tasks.py
+++ b/tasks.py
@@ -12,10 +12,12 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+import concurrent.futures
 import json
 import os
 import platform
 import re
+import time
 
 from invoke import Collection, task as invoke_task
 from invoke.exceptions import Exit, Failure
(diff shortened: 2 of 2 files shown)


glennmatthews:
I'm realizing now another potential impact of the previous change - if container startup fails during the `docker-compose up` portion of `invoke debug`, do we no longer get the logs of the failed startup at all?

msheiny:
> I'm realizing now another potential impact of the previous change - if container startup fails during the `docker-compose up` portion of `invoke debug`, do we no longer get the logs of the failed startup at all?

You are absolutely right. Please take a look at my latest commit I force pushed, I changed the behavior to revert to how the OG `debug` was working and moved the new port detection logic to the background to run in parallel.

glennmatthews (approved):
"Feels" better to me, though I'm not in a position to test it at the moment, I much appreciate the provided screen recording. :-)

PR #6637, opened 2024-12-10
6518 Add VRf's column to Prefix and Child Prefixes

mateusz-lachowski-codilime:
# Closes #6518 
# What's Changed

# Screenshots
## Before
image
image

## After
image
image
# TODO

- [ ] Explanation of Change(s)
- [ ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [ ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example App Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #6958, opened 2025-02-27
Use short location name and VRF name in prefix tables

### LINKED ISSUE #6957: Show location.name in prefix tables rather than prefix.display
### As ...

Nelly - Network Engineer

### I want ...

Prefix tables to include the name of a location as they worked in v2.1.  This used the `location.name` field rather than `location.display`.

v2.2 changed this to a count with prefix-location M2M relationship change
v2.3.11 re-added names for 1:1 relationships but used `display` field

### So that ...

Prefix tables are more readable and concise.  

The problem with the change to `location.display` is that it includes the full location hierarchy which can be quite long.  This often requires line wrapping in the table which makes it harder to read.

The prefix list and child prefix lists show locations using only the `location.name` field and not the full `location.display` field/heirarchy.

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [x] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

No database changes needed to implement this.

### External Dependencies

No external dependencies required.

cdwchriburg:
# Closes #6957 
# What's Changed

- Add an additional optional parameter `display_field` to `LinkedCountColumn`
  - If present on a given column, this field name will be passed to the `field` parameter of `hyperlinked_object()` helper
  - Otherwise "display" will be the field name used (i.e. same as current/default behavior)
- Change PrefixTable to use this additional parameter for the `Locations` column and pass "name" as the field to use for displaying location hyperlinks.
- Change PrefixTable to use "name" field for VRF column as well (mentioned in #6637)
- Note that this does *not* change the text used for Location or VRF links hyperlinks in other areas - only in prefix tables.
# Screenshots

### Locations
Prefixes Before
Child Prefixes Before
Parent Prefixes Before

### VRFs
VRF Prefixes Before
VRF Parent Prefixes Before

### Locations
Prefixes Before
Child Prefixes After
Parent Prefixes After

### VRFs
VRF Prefixes After
VRF Parent Prefixes After

# TODO

- [ ] Unit, Integration Tests


changes/6957.changed
```diff
@@ -0,0 +1 @@
+Changed "Locations" column in Prefix table to show `location.name` instead of `location.display` to avoid verbose location hierarchy in tables
```
glennmatthews, line 1:
```suggestion
Changed "Locations" column in Prefix table to show `location.name` instead of `location.display` to avoid verbose location hierarchy in tables.
```

glennmatthews:
> I ran the tests locally but tbh its pretty unclear if the integration tests passed from the output. It just gave like code coverage percentages but no real "yes/no" passing output.

If you get a code coverage report, it passed - agree that's not necessarily obvious. We do have an open issue where integration tests are failing somewhat frequently in CI so if that happens it isn't necessarily your code at fault. We can try a rerun as a starting point.

HanlinMiao (commented):
Changes look good, we probably have to modify the unit test here. https://github.com/nautobot/nautobot/blob/develop/nautobot/ipam/tests/test_views.py#L211 to something like 
`self.assertBodyContains(response, hyperlinked_object(prefix.locations.first().name))`

cdwchriburg:
Should we introduce another test to validate the case where `display_field` is default value/not overridden as well now?

nautobot/ipam/tests/test_views.py
```diff
@@ -208,7 +208,7 @@ def test_list_objects_with_permission(self):
                         response, f'<a href="{locations_list_url}?prefixes={prefix.pk}" class="badge">{count}</a>'
                     )
                 elif count == 1:
-                    self.assertBodyContains(response, hyperlinked_object(prefix.locations.first()))
+                    self.assertBodyContains(response, hyperlinked_object(prefix.locations.first().name))
```
glennmatthews, line 211: I think this'd be more correct?

```suggestion
                    self.assertBodyContains(response, hyperlinked_object(prefix.locations.first(), "name"))
```

glennmatthews:
> Should we introduce another test to validate the case where `display_field` is default value/not overridden as well now?

If you don't mind, that would definitely be good.

nautobot/ipam/tests/test_views.py
phobozad: Yep good catch.  Otherwise it would incorrectly pass if the name was anywhere in the field along with other text rather than just the pure `location.name`

cdwchriburg: Yep good catch. Otherwise it would incorrectly pass if the name was anywhere in the field along with other text rather than just the pure `location.name`

cdwchriburg:
Added the additional test case.  I think this should be fully ready for review now.

glennmatthews (approved):

PR #6962, opened 2025-02-28
Migrated ipam:service views to NautobotViewset

Sridevi-NTC:
# What's Changed
Migrated Ipam service views to use NautobotViewset
# TODO

Migrated Ipam service views to use NautobotViewset
- [X] Attached Screenshots

| List View       | image | image |
| Detail View     | image|image image|
| Update View     | image |image|
| Delete View     | image image|  image image|
| Bulk Edit View  | image image |image image|
| Bulk Delete View| image image| image image|

- [x] Unit, Integration Tests


nautobot/core/forms/fields.py
```diff
@@ -801,6 +801,11 @@ def to_python(self, value):
         try:
             if not value:
                 value = ""
+
+            elif isinstance(value, list):
```
glennmatthews, line 805: Would you mind adding a test case or two to `nautobot.core.tests.test_forms.NumericArrayFieldTest` that cover the scenario(s) that required this addition?

nautobot/ipam/templates/ipam/service_retrieve.html
```diff
@@ -0,0 +1 @@
+{% extends 'ipam/service.html' %}
```
glennmatthews, line 1: Do we need to create this file? What fails if it isn't created?

nautobot/ipam/views.py
```diff
@@ -1373,18 +1373,6 @@ class VLANBulkDeleteView(generic.BulkDeleteView):
...
-    filterset_form = forms.ServiceFilterForm
-    table = tables.ServiceTable
-    action_buttons = ("add", "import", "export")
-
-
-class ServiceView(generic.ObjectView):
-    queryset = Service.objects.prefetch_related("ip_addresses")
-
-
 class ServiceEditView(generic.ObjectEditView):
```
glennmatthews, line 1376: Remove as well?

nautobot/core/forms/fields.py
Sridevi-NTC: sure @glennmatthews

nautobot/ipam/templates/ipam/service_retrieve.html
Sridevi-NTC: I'm unsure about deleting ipam/service.html, I took ref from this
https://github.com/nautobot/nautobot/blob/develop/nautobot/tenancy/templates/tenancy/tenantgroup.html#L2

I'm only extending ipam/service.html because ports are a many-to-many field.
https://github.com/nautobot/nautobot/blob/develop/nautobot/ipam/templates/ipam/service.html#L33

nautobot/ipam/views.py
Sridevi-NTC: This is being used in devices
https://github.com/nautobot/nautobot/blob/develop/nautobot/dcim/urls.py#L740

HanlinMiao: Can we mark this as deprecated, at least? Or should we make a completely separate view for the one that is used in devices?

Sridevi-NTC: The same view is also used in devices and virtualization:
https://github.com/nautobot/nautobot/blob/develop/nautobot/virtualization/urls.py#L214.

Would you suggest creating separate views for each?

Sridevi-NTC: @HanlinMiao Any suggestion on the above plz?

glennmatthews: I'd vote to leave it as-is for now but add comments here explaining that the view is used in those locations presently. Longer-term it *should* be possible to use the UIViewSet `create` action in those places instead.

nautobot/core/jobs/bulk_actions.py
```diff
@@ -108,7 +108,7 @@ def _update_objects(self, model, form, filter_query_params, edit_all, nullified_
                         if form.cleaned_data[field_name]:
                             getattr(obj, field_name).set(form.cleaned_data[field_name])
                     # Normal fields
-                    elif form.cleaned_data[field_name] not in (None, ""):
+                    elif form.cleaned_data[field_name] not in (None, "", []):
```
glennmatthews, line 111: If this change fixes a bug that end users could encounter, please document it in a `changes/6962.fixed` file.

nautobot/ipam/views.py
```diff
@@ -1401,23 +1389,19 @@ def alter_obj(self, obj, request, url_args, url_kwargs):
...
-    table = tables.ServiceTable
+    serializer_class = serializers.ServiceSerializer
+    table_class = tables.ServiceTable
+
+    def get_queryset(self):
+        if self.action in ["bulk_update", "bulk_destroy"]:
+            self.queryset = Service.objects.select_related("device", "virtual_machine")
+        elif self.action == "retrieve":
+            self.queryset = Service.objects.prefetch_related("ip_addresses")
+        return super().get_queryset()
```
glennmatthews, line 1407: Not sure the conditional logic is needed here - both the retrieve view and the ServiceTable display all three of the specified fields, so I'd think just declaring `queryset = Service.objects.select_related("device", "virtual_machine").prefetch_related("ip_addresses")` above would be sufficient.

nautobot/ipam/tables.py
```diff
@@ -833,7 +833,7 @@ class ServiceTable(BaseTable):
     pk = ToggleColumn()
     name = tables.Column(linkify=True)
     parent = tables.LinkColumn(order_by=("device", "virtual_machine"))
-    ports = tables.TemplateColumn(template_code="{{ record.port_list }}", verbose_name="Ports")
+    ports = tables.TemplateColumn(template_code="{{record.ports|join:','}}", verbose_name="Ports")
```
glennmatthews, line 836: This change seems undesirable to me - what prompted it?

nautobot/ipam/templates/ipam/service_retrieve.html
```diff
@@ -0,0 +1,48 @@
...
+                    <td>{{ object.parent|hyperlinked_object }}</td>
+                </tr>
+                <tr>
+                    <td>Protocol</td>
+                    <td>{{ object.get_protocol_display }}</td>
+                </tr>
+                <tr>
+                    <td>Ports</td>
+                    <td>
+                        {{ object.ports|join:',' }}
```
glennmatthews, line 29: Why the change away from `object.port_list` here as well?

nautobot/core/forms/utils.py
```diff
@@ -41,6 +41,10 @@ def parse_numeric_range(input_string, base=10):
                 raise TypeError("Input value must be a string using a range format.")
         except ValueError:
             begin, end = dash_range, dash_range
+
+        if any(c in "abcdefABCDEF" for c in begin + end):
+            base = 16
```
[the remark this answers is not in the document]
itdependsnetworks: Yea, we spoke, I don't think this is the correct approach, in general `base` is a parameter, and should not automatically be adjusted. It's not clear to me what they appropriate approach is, but something is off.

nautobot/ipam/templates/ipam/service_retrieve.html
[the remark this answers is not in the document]
itdependsnetworks: Had it as future work, but can do it before this PR is merged, but will have @Sridevi-NTC concentrate on the NautobotUIViewSet work for now.

nautobot/ipam/views.py
itdependsnetworks: okay, so there is no place where we are not getting this data, so this would actually optimize those other queries.

nautobot/ipam/views.py
```diff
@@ -1398,3 +1408,21 @@ class ServiceUIViewSet(NautobotUIViewSet):  # 3.0 TODO: remove, unused BulkImpor
...
+                weight=100,
+                fields=[
+                    "name",
+                    "parent",
+                    "protocol",
+                    "port_list",
+                    "ip_addresses",
+                    "description",
+                ],
+                value_transforms={"ip_addresses": [to_list]},
```
glennmatthews, line 1425: Think it'd make more sense to use an `ObjectsTablePanel` to actually display the IP addresses in detail, rather than fitting them into the ObjectFieldsPanel like this.

itdependsnetworks: Yes, that was how she had it originally, did not realize this was preferred path. Should there be the add button on that `ObjectsTablePanel`?

Sridevi-NTC: Sure will do that

nautobot/ipam/views.py
```diff
@@ -1414,15 +1404,28 @@ class ServiceUIViewSet(NautobotUIViewSet):  # 3.0 TODO: remove, unused BulkImpor
...
-                    "ip_addresses",
-                    "description",
-                ],
-                value_transforms={"ip_addresses": [to_list]},
+                fields=["name", "parent", "protocol", "port_list", "description"],
+            ),
+            object_detail.ObjectsTablePanel(
+                weight=200,
+                section=SectionChoices.RIGHT_HALF,
+                context_table_key="ip_addresses_table",
```
glennmatthews, line 1412: Instead of manually constructing the table in `get_extra_context` and using a `context_table_key`, it should be possible (and simpler) to use a `table_class` and `table_filter` here, probably:

```suggestion
                table_class=tables.IPAddressTable,
                table_filter="services",
                add_button_route=None,
```

changes/6962.fixed
```diff
@@ -0,0 +1 @@
+Fixed a `Validation` error with empty ports in bulk editing.
```
HanlinMiao, line 1:
```suggestion
Fixed a Validation error with empty ports field in Service Bulk Edit Form.
```

changes/6962.housekeeping
```diff
@@ -0,0 +1 @@
+Refactored `service views` to use `NautobotUIViewSet` and `UI component framework`.
```
HanlinMiao, line 1:
```suggestion
Refactored Service model related UI views to use `NautobotUIViewSet` and `UI component framework`.
```

changes/6962.fixed
HanlinMiao: RadioProfile also uses NumericArrayField in its forms. This PR likely fixes this issue for it as well. Would we want to capture that in the changelog as well?

`RadioProfileBulkEditForm/RadioProfileForm` -> `allowed_channel_list`

nautobot/core/forms/fields.py
```diff
@@ -800,9 +800,13 @@ class NumericArrayField(SimpleArrayField):
     def to_python(self, value):
         try:
             if not value:
-                value = ""
-            else:
-                value = ",".join([str(n) for n in forms.parse_numeric_range(value)])
+                return []
```
HanlinMiao, line 803: Would this return an empty list if `value` is "0"?

nautobot/core/tests/test_forms.py
```diff
@@ -486,20 +486,23 @@ def setUp(self):
...
+        #  List of (input, expected output) tuples
+        tests = [
+            (None, []),
+            ("", []),
+            ("80,443-444", [80, 443, 444]),
+            ("1024-1028,31337", [1024, 1025, 1026, 1027, 1028, 31337]),
+            (["47-49", "103"], [47, 48, 49, 103]),
+            ([231, 432, 313], [231, 313, 432]),
+        ]
+        for test, expected in tests:
```
HanlinMiao, line 498: Very nice!

HanlinMiao: Curious to see what's the output would be for a test case like ("0", [0])

HanlinMiao (commented):
Awesome work! We are very close to be ready to merge this.

nautobot/core/forms/fields.py
Sridevi-NTC: I will check this scenario

Sridevi-NTC: No this will return [0] but we have SERVICE_PORT_MIN = 1 and SERVICE_PORT_MAX = 65535 defined in the form. Should this be handled??

HanlinMiao: I was wondering that `if not value` will evaluate to True when `value` is equal to`"0"` for which it returns an empty list. Seems like it's working as expected though. Thanks for the investigation.

HanlinMiao (approved):

PR #6965, opened 2025-02-28
Add Job.logger.failure() and Job.fail() APIs

### LINKED ISSUE #6384: Adding additional Celery Exception Handling.
### As ...

### I want ...

_We use Sentry for monitoring uncaught exceptions. As such the currently documented solution to fail jobs by throwing an uncaught exception is missing some detail as it will (often) create potentially unnecessary Sentry issues for any job that **needs** to be marked failed to alert the user of a failure condition the job encounters. We also make extensive use of Celery’s Canvas features such as sub-task groups and chains and need to generate new tasks based on retry-able failure modes so we want to avoid impacting that functionality._

### So that ...

Allowing the Celery mechanism to update the Job Result Status could provide additional details that may be useful to operators and troubleshooters of failure events (both expected and unexpected). This would allow to explicitly mark a failure condition rather than implicitly trigger a failure based on the log level such as logger.error() or logger.exception(). 

I am able to write a job to capture these failure events and can customize the alerting based on the different types of failures of the celery worker. 
```
    def after_return(self, status, retval, task_id, args, kwargs, einfo):
        self.celery_task.update_state(
            state=JobResultStatusChoices.STATUS_FAILURE,
            meta={
                "exc_type": type(Ignore).__name__,
                "exc_message": "",
                "custom": "...",
            },
        )
        raise Ignore()
```

I have attached a patch with the suggested modifications.
[celery-result-handling-patch.txt](https://github.com/user-attachments/files/17412604/celery-result-handling-patch.txt)

- [ ] Automation
- [ ] Circuits
- [ ] DCIM
- [ ] IPAM
- [ ] Organization
- [X] Apps (and other Extensibility)
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

I am unsure of database changes.

### External Dependencies

to my knowledge, an additional import would be required for this:
`from celery.exceptions import Ignore, Reject
`
into:
`nautobot/extras/jobs.py`

glennmatthews:
# Closes #6384 
# What's Changed

- Add `Job.logger.failure()` support and corresponding `FAILURE` log level - similar to what was done in #6541.
   - Enhanced JobResult table, view, and JobLogEntry table to handle the new log level appropriately.
- Add `Job.fail(message)` API.
   - This allows a Job author to mark the job as failed without raising an uncaught exception and Celery reporting and recording a traceback.
   - This API does not itself automatically abort the Job execution in progress, but that can be done trivially if desired (e.g. `self.fail("an error occurred"); return None`) or the execution can continue normally if preferred with only a final terminal failure being reported.
- Update Job development docs
- Add example Job to the example App that demonstrates the two different ways of failing a Job (exception vs self.fail()) and their impacts.

Additionally:

- Add `init: true` to the development `docker-compose.yml` to address an issue I was encountering locally where timed-out healthchecks on a busy system were accumulating endlessly as zombie processes instead of being cleanly reaped.
- Moved code **out** of the default `Job.before_start()` and `Job.after_return()` method implementations to private functions, eliminating a "foot-gun" where a Job author might override these methods but forget to call `super()` (which we did **not** document as a requirement anywhere)

# Screenshots

image

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot/__init__.py
+++ b/nautobot/__init__.py
@@ -18,6 +18,7 @@ def add_success_logger():
     logging.addLevelName(SUCCESS, "SUCCESS")
 
     def success(self, message, *args, **kwargs):
+        kwargs["stacklevel"] = kwargs.get("stacklevel", 1) + 1  # so that funcName is the caller function, not "success"
         if self.isEnabledFor(SUCCESS):
             self._log(SUCCESS, message, args, **kwargs)
 
@@ -31,6 +32,7 @@ def add_failure_logger():
     logging.addLevelName(FAILURE, "FAILURE")
 
     def failure(self, message, *args, **kwargs):
+        kwargs["stacklevel"] = kwargs.get("stacklevel", 1) + 1  # so that funcName is the caller function, not "failure"
         if self.isEnabledFor(FAILURE):
             self._log(FAILURE, message, args, **kwargs)
 
--- a/nautobot/__init__.py
+++ b/nautobot/__init__.py
@@ -14,18 +14,32 @@ __initialized = False
 
 def add_success_logger():
     """Add a custom log level for success messages."""
-    SUCCESS = 25
+    SUCCESS = 25  # between INFO and WARNING
     logging.addLevelName(SUCCESS, "SUCCESS")
 
-    def success(self, message, *args, **kws):
+    def success(self, message, *args, **kwargs):
         if self.isEnabledFor(SUCCESS):
-            self._log(SUCCESS, message, args, **kws)
+            self._log(SUCCESS, message, args, **kwargs)
 
     logging.Logger.success = success
     return success
 
 
+def add_failure_logger():
+    """Add a custom log level for failure messages less severe than an ERROR."""
+    FAILURE = 35  # between WARNING and ERROR
+    logging.addLevelName(FAILURE, "FAILURE")
+
+    def failure(self, message, *args, **kwargs):
+        if self.isEnabledFor(FAILURE):
+            self._log(FAILURE, message, args, **kwargs)
+
+    logging.Logger.failure = failure
+    return failure
+
+
 add_success_logger()
+add_failure_logger()
 logger = logging.getLogger(__name__)
 
 
--- a/nautobot/core/tests/test_jobs.py
+++ b/nautobot/core/tests/test_jobs.py
@@ -70,7 +70,7 @@ class ExportObjectListTest(TransactionTestCase):
             username=self.user.username,  # otherwise run_job_for_testing defaults to a superuser account
             content_type=ContentType.objects.get_for_model(Status).pk,
         )
-        self.assertJobResultStatus(job_result, JobResultStatusChoices.STATUS_SUCCESS)
+        self.assertJobResultStatus(job_result)
         self.assertTrue(job_result.files.exists())
         self.assertEqual(Path(job_result.files.first().file.name).name, "nautobot_statuses.csv")
         csv_bytes = job_result.files.first().file.read()
@@ -86,7 +86,7 @@ class ExportObjectListTest(TransactionTestCase):
             "ExportObjectList",
             content_type=ContentType.objects.get_for_model(Status).pk,
         )
-        self.assertJobResultStatus(job_result, JobResultStatusChoices.STATUS_SUCCESS)
+        self.assertJobResultStatus(job_result)
         self.assertTrue(job_result.files.exists())
         self.assertEqual(Path(job_result.files.first().file.name).name, "nautobot_statuses.csv")
         csv_data = job_result.files.first().file.read().decode("utf-8")
@@ -107,7 +107,7 @@ class ExportObjectListTest(TransactionTestCase):
             content_type=ContentType.objects.get_for_model(Status).pk,
             export_template=et.pk,
         )
-        self.assertJobResultStatus(job_result, JobResultStatusChoices.STATUS_SUCCESS)
+        self.assertJobResultStatus(job_result)
         self.assertTrue(job_result.files.exists())
         self.assertEqual(Path(job_result.files.first().file.name).name, "nautobot_statuses.txt")
         text_data = job_result.files.first().file.read().decode("utf-8")
@@ -129,7 +129,7 @@ class ExportObjectListTest(TransactionTestCase):
             content_type=ContentType.objects.get_for_model(DeviceType).pk,
             export_format="yaml",
         )
-        self.assertJobResultStatus(job_result, JobResultStatusChoices.STATUS_SUCCESS)
+        self.assertJobResultStatus(job_result)
         self.assertTrue(job_result.files.exists())
         self.assertEqual(Path(job_result.files.first().file.name).name, "nautobot_device_types.yaml")
         yaml_data = job_result.files.first().file.read().decode("utf-8")
@@ -220,7 +220,7 @@ class ImportObjectsTestCase(TransactionTestCase):
             content_type=ContentType.objects.get_for_model(Status).pk,
             csv_data=self.csv_data,
         )
-        self.assertJobResultStatus(job_result, JobResultStatusChoices.STATUS_SUCCESS)
+        self.assertJobResultStatus(job_result)
         self.assertFalse(
             JobLogEntry.objects.filter(job_result=job_result, log_level=LogLevelChoices.LOG_WARNING).exists()
         )
@@ -246,7 +246,7 @@ class ImportObjectsTestCase(TransactionTestCase):
             content_type=ContentType.objects.get_for_model(Prefix).pk,
             csv_file=csv_file.id,
         )
-        self.assertJobResultStatus(job_result, JobResultStatusChoices.STATUS_SUCCESS)
+        self.assertJobResultStatus(job_result)
         self.assertFalse(
             JobLogEntry.objects.filter(job_result=job_result, log_level=LogLevelChoices.LOG_WARNING).exists()
         )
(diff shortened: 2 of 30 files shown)


nautobot/core/tests/test_jobs.py
```diff
@@ -70,7 +70,7 @@ def test_export_with_constrained_permission(self):
             username=self.user.username,  # otherwise run_job_for_testing defaults to a superuser account
             content_type=ContentType.objects.get_for_model(Status).pk,
         )
-        self.assertEqual(job_result.status, JobResultStatusChoices.STATUS_SUCCESS)
+        self.assertJobResultStatus(job_result, JobResultStatusChoices.STATUS_SUCCESS)
```
HanlinMiao, line 73: Probably only need for this and below right? Since the default parameter is already `JobResultStatusChoices.STATUS_SUCCESS`
```suggestion
        self.assertJobResultStatus(job_result)
```

nautobot/docs/development/jobs/index.md
```diff
@@ -619,19 +620,19 @@ Again, defining user variables is totally optional; you may create a Job with a
...
 
 #### The `after_return()` Method
 
-Regardless of the overall Job execution success or failure, the `after_return()` method will be called after `on_success()` or `on_failure()`. It has the signature `after_return(self, status, retval, task_id, args, kwargs, einfo)`; the `status` will indicate success or failure (using the `JobResultStatusChoices` enum), `retval` is *either* the return value from `run()` (on success) or the exception raised (on failure), and once again `kwargs` contains the user variables.
+Regardless of the overall Job execution success or failure, the `after_return()` method will be called after `on_success()` or `on_failure()`. It has the signature `after_return(self, status, retval, task_id, args, kwargs, einfo)`; the `status` will indicate success or failure (using the `JobResultStatusChoices` enum), `retval` is *either* the return value from `run()` or the exception raised, and once again `kwargs` contains the user variables.
```
HanlinMiao, line 635: I would love to unify the `retval` and `exc` in the future. The mismatch is the method signature is a bit confusing imo.

nautobot/docs/development/jobs/index.md
```diff
@@ -682,8 +683,11 @@ Markdown rendering is supported for log messages, as well as [a limited subset o
 +/- 2.0.0
     The `AbortTransaction` class was moved from the `nautobot.utilities.exceptions` module to `nautobot.core.exceptions`. Jobs should generally import it from `nautobot.apps.exceptions` if needed.
 
-+++ 2.4.0
-    You can now use `self.logger.success()` to set the log level to `SUCCESS`.
++++ 2.4.0 "`logger.success()` added"
+    You can now use `self.logger.success()` to log a message at the level `SUCCESS`, which is located between the standard `INFO` and `WARNING` log levels.
+
++++ 2.4.5 "`logger.failure()` added"
+    You can now use `self.logger.failure()` to log a message at the level `FAILURE`, which is located betwen the standard `WARNING` and `ERROR` log levels.
```
HanlinMiao, line 690:
```suggestion
    You can now use `self.logger.failure()` to log a message at the level `FAILURE`, which is located between the standard `WARNING` and `ERROR` log levels.
```

nautobot/docs/development/jobs/index.md
```diff
@@ -706,20 +710,26 @@ The maximum size of any single created file (or in other words, the maximum numb
...
-        if var1 != "Taco":
-            self.logger.error("var1 must be 'Taco'")
+    def run(self, occasion):
+        if not occasion.startswith("Taco"):
+            self.logger.failure("The occasion must begin with 'Taco'")
             raise Exception("Argument input validation failed.")
+        if not occasion.endswith(" Tuesday"):
+            self.fail("It's supposed to be Tuesday")
+        self.logger.info("Today is %s", occasion)
+        return occasion
```
HanlinMiao, line 732: Nice

nautobot/extras/jobs.py
```diff
@@ -163,7 +165,11 @@ def __call__(self, *args, **kwargs):
     def __str__(self):
         return str(self.name)
 
-    # See https://github.com/PyCQA/pylint-django/issues/240 for why we have a pylint disable on each classproperty below
+    def fail(self, msg, *args, **kwargs):
+        """Mark this job as failed without immediately raising an exception and aborting."""
+        # Instead of failing in "fail" grouping, fail in the parent function's grouping by default
```
HanlinMiao, line 170: Does this have to do with the stacklevel = 2 declaration below?

nautobot/extras/jobs.py
```diff
@@ -1205,56 +1148,191 @@ def get_job(class_path, reload=False):
...
+            time_limit,
+            soft_time_limit,
+            time_limit,
+            extra={"grouping": "initialization", "object": job.job_model},
+        )
+
+    # Send notice that the job is running
+    event_payload = {
+        "job_result_id": request.id,
+        "job_name": job.name,  # TODO: should this be job.job_model.name instead? Possible breaking change
```
HanlinMiao, line 1231: Why would this be a breaking change? The old code is essentially the same

```
payload = {
        "job_result_id": self.request.id,
        "job_name": job.name,
        "user_name": job_result.user.username,
    }
```

nautobot/extras/jobs.py
```diff
@@ -1205,56 +1148,191 @@ def get_job(class_path, reload=False):
...
+        # We don't want to overwrite the manual state update that we did above, so:
+        raise Ignore()
+
+    except Reject:
+        status = status or JobResultStatusChoices.STATUS_REJECTED
+        raise
+
+    except Ignore:
+        status = status or JobResultStatusChoices.STATUS_IGNORED
+        raise
```
HanlinMiao, line 1320: In which case would status be `None`? If job.run() raise Ignore or Reject right?

nautobot/extras/tables.py
```diff
@@ -877,8 +877,10 @@ class JobResultTable(BaseTable):
         linkify=True,
         verbose_name="Scheduled Job",
     )
-    actions = tables.TemplateColumn(
-        template_code="""
+    actions = ButtonsColumn(
+        JobResult,
+        buttons=("delete",),
+        prepend_template="""
```
HanlinMiao, line 883: Can we put this block into a CONSTANT as well ,something like JOB_RESULT_BUTTON?

nautobot/extras/tests/test_relationships.py
```diff
@@ -1813,11 +1813,11 @@ def test_required_relationships(self):
 
         # Try editing 6 VLANs and adding the required device (succeeds):
         job_result = self.create_job(pk_list, add_cr_vlans_devices_m2m__source=[str(device_for_association.id)])
-        self.assertEqual(job_result.status, JobResultStatusChoices.STATUS_SUCCESS)
+        self.assertJobResultStatus(job_result, JobResultStatusChoices.STATUS_SUCCESS)
```
HanlinMiao, line 1816:
```suggestion
        self.assertJobResultStatus(job_result)
```

nautobot/extras/tests/test_jobs.py
```diff
@@ -762,7 +769,7 @@ def test_job_fires_webhooks(self, mock_enqueue_webhooks):
         webhook.content_types.set([status_ct])
 
         job_result = create_job_result_and_run_job(module, name)
-        self.assertEqual(job_result.status, JobResultStatusChoices.STATUS_SUCCESS)
+        self.assertJobResultStatus(job_result, JobResultStatusChoices.STATUS_SUCCESS)
```
HanlinMiao, line 772: A lot of cases like this in test_*.py
```suggestion
        self.assertJobResultStatus(job_result)
```

nautobot/__init__.py
```diff
@@ -14,18 +14,34 @@
 
 def add_success_logger():
     """Add a custom log level for success messages."""
-    SUCCESS = 25
+    SUCCESS = 25  # between INFO and WARNING
     logging.addLevelName(SUCCESS, "SUCCESS")
 
-    def success(self, message, *args, **kws):
+    def success(self, message, *args, **kwargs):
+        kwargs["stacklevel"] = kwargs.get("stacklevel", 1) + 1  # so that funcName is the caller function, not "success"
```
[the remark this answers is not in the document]
glennmatthews: https://docs.python.org/3/library/logging.html#logging.Logger.debug

> The third optional keyword argument is stacklevel, which defaults to 1. If greater than 1, the corresponding number of stack frames are skipped when computing the line number and function name set in the [LogRecord](https://docs.python.org/3/library/logging.html#logging.LogRecord) created for the logging event. This can be used in logging helpers so that the function name, filename and line number recorded are not the information for the helper function/method, but rather its caller. The name of this parameter mirrors the equivalent one in the [warnings](https://docs.python.org/3/library/warnings.html#module-warnings) module.

In short, it lets you walk up the call stack to control which function gets reported in the log message as the "location" (`funcName`) of the logger call. Without this change, the log gets reported as originating from within the `success()` method itself; with this change, the log gets reported as originating from the method that's *calling* `success()`, which is generally more useful and is consistent with how the built-in `logger.info()`, etc. operate.

nautobot/core/tests/test_jobs.py
glennmatthews: Yeah, I got lazy here. It's harmless but redundant; I can remove it if you think it's worth the effort.

nautobot/extras/jobs.py
glennmatthews: Yes, same idea as the other comment above. Without it, the failure gets reported as occurring within `fail()` instead of within the function that *called* `fail()`.

nautobot/extras/jobs.py
glennmatthews: Right, if `before_start` or `run` raises either of these exceptions directly (which I understand to be part of the user use case driving this whole body of work) then `status` will not have been set.

nautobot/extras/tables.py
glennmatthews: Sure, can do.

nautobot/extras/jobs.py
glennmatthews: before:
image

after:
image

nautobot/__init__.py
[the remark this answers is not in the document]
HanlinMiao: Awesome thank you

HanlinMiao (approved):
LGTM!

PR #6966, opened 2025-02-28
Add user object access to custom validators

lampwins:
# Closes #
# What's Changed

This adds support for accessing the current user object via `self.context["user"]` from a custom validator, in most situations (when inside of a `web_request_context` (yes this includes Jobs). Otherwise, an `AnonymousUser` is returned so developers can rely on a sane default user that does not have any permissions.

# Screenshots

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/changes/6966.added
+++ b/changes/6966.added
@@ -1,0 +1,1 @@
+Added support for accessing the current user within Custom Validators via `self.context["user"]`.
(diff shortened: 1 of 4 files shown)


glennmatthews (approved):
Nice work!

Issue #5819, opened 2024-06-14
GitRepository URL Validator does not allow for local DNS URLs


MeganerdDev:
### Environment
* Nautobot version (Docker tag too if applicable): v2.2.25
* Python version: 3.11
* Database platform, version: PG15
* Middleware(s): N/A

### Steps to Reproduce
1. Navigate to `extras/git-repositories` in the Nautobot UI
2. Press the `Add` button to begin creating a new GIT Repository object
3. On the `Remote URL` charfield entry provide a local DNS name eg: `https://local-dns-example/golden-config-backup.git`
4. Fill in remaining required entries and press `Update & Sync` button

### Expected Behavior
Object created and syncs as normal operation

### Observed Behavior
URL validator throws an error preventing creation of the object
![image](https://github.com/nautobot/nautobot/assets/114895043/fe48edd2-e6d4-4afb-897a-2f5af08d6408)

### Additional Details

- I wasn't sure if this would be a feature request or a bug report. I ended up with bug report.

- [URLValidator](https://github.com/nautobot/nautobot/blob/eb1a1796fce4be45d7086137f551e5f46e64e1a3/nautobot/extras/models/datasources.py#L9) is upstream and provided by Django
`from django.core.validators import URLValidator`

- URLValidator in [Django](https://github.com/django/django/blob/fa7848146738a9fe1d415ee4808664e54739eeb7/django/core/validators.py#L127-L209)

MeganerdDev:
I was able to mitigate this over the ORM for now
```python
from nautobot.extras.models import GitRepository

GitRepository.objects.create(
    name="test",
    slug="test",
    remote_url="https://local-dns-example/golden-config-backup.git",
    branch="main"
)
```

Also unrelated to the URL Validator for local DNS names, but local pathfiles might also be considered during any refactoring change that is done to this entry? I had a separate case where I worked around this by creating the object over ORM

glennmatthews:
Nautobot has a custom `nautobot.core.models.validators.EnhancedURLValidator` that supports hostnames, it looks like it might just be as simple as fixing the `GitRepository` definition to use that instead of Django's default `URLValidator`.

glennmatthews:
Similarly the form itself needs to use `nautobot.core.forms.fields.LaxURLField` instead of Django's `forms.URLField`.

PR #6991, opened 2025-03-06
Improve URL schema control for SoftwareFileImage model  #6739

### LINKED ISSUE #6739: Update Software Image File Download URL to use LaxURLField
### Environment
* Nautobot version (Docker tag too if applicable): 2.3
* Python version: 3.11
* Middleware(s): n/a

### Steps to Reproduce
1. Create software version
2. Navigate to Software Image Files, and add a new sw file
3. Use download file and try sftp or another schema `sftp://1.1.1.1`.

### Expected Behavior

Allow a more flexible value for schemes that the URL validator accepts.  SFTP and a few others would be valid for a software image file within an enterprise

### Observed Behavior
Error that its an invalid URL
Screenshot 2025-01-07 at 12 50 58 PM

Currently it’s using Django’s built-in URLField (which only allows http/https/ftp/ftps by default) instead of Nautobot’s LaxURLField (which allows everything in ALLOWED_URL_SCHEMES)

garymccann:
# Closes #6739
# What's Changed

`download_url` within the `SoftwareFileImage` model only permits a limited set of URL Schemas, as defined by Django URLField ("https", "http", "ftp", "ftps" - source [Django 4.2 - URLValidator](https://docs.djangoproject.com/en/4.2/ref/validators/#django.core.validators.URLValidator.schemes))

Moving to `LaxURLField`, which is derived from Django URLField, allows for an extended list of URL Schemas. The allowed list is a controlled property in `nautobot_config.py` - using the property `ALLOWED_URL_SCHEMES` [docs](https://docs.nautobot.com/projects/core/en/stable/user-guide/administration/configuration/settings/#allowed_url_schemes)

This closely relates to #5819, similar intent to relax the constraints associated with URLField
# Screenshots

### Issue described in #6739
before_invalid

### PR addresses this use case 
after_now_valid

### While allowing more URL Schemas
after_valid

### However the allowed Schemas are still limited to those under administrative control using `ALLOWED_URL_SCHEMES`
after_invalid

# TODO

- [x] Unit, Integration Tests


garymccann:
@glennmatthews - Unit Test added, but it's currently of limited usefulness. My intent was to set `ALLOWED_URL_SCHEME` and validate entries against that, but i don't see a way to dynamically set it, nor for the laxURLField Validator to get updated. Are there other means to achieve this? Any alternative tests you would recommend?

glennmatthews:
> Thought we were considering a FK to external integration instead ?

I think in the longer term (#7003) that absolutely makes sense, however I think we will *also* need to keep the local `download_url` field in 2.x to avoid any breaking change to existing users, and I think this is an appropriate fix to that field. Would you agree?

nautobot/core/models/validators.py
```diff
@@ -27,8 +27,15 @@ class EnhancedURLValidator(URLValidator):
         r"\Z",
         re.IGNORECASE,
     )
+
     schemes = settings.ALLOWED_URL_SCHEMES
 
+    def __getattribute__(self, name):
+        """Dynamically fetch schemes each time it's accessed."""
+        if name == "schemes":
+            self.schemes = settings.ALLOWED_URL_SCHEMES  # Always return the latest list
```
glennmatthews, line 36: Is this just for unit testing purposes? Would something like the following work as well and be a bit cleaner?

```py
@property
def schemes(self):
    return settings.ALLOWED_URL_SCHEMES
```

garymccann:
> > Thought we were considering a FK to external integration instead ?
> 
> I think in the longer term (#7003) that absolutely makes sense, however I think we will _also_ need to keep the local `download_url` field in 2.x to avoid any breaking change to existing users, and I think this is an appropriate fix to that field. Would you agree?

Yes, i believe keeping download_url, but relaxing it's validation (this PR) is the best way to avoid a breaking change, while adding the external integrations as an optional field is acceptable for the use case ( PR: [link)](https://github.com/nautobot/nautobot/pull/7011)

nautobot/core/models/validators.py
garymccann: No - i considered this first of all, see below. The best way i discovered was to intercept the access to the attribute.

The parent class (django validators) requires an iterable , so a Type error is raised:
```
  File "/usr/local/lib/python3.12/site-packages/django/core/validators.py", line 121, in __call__
    if scheme not in self.schemes:
    ^^^^^^^^^^^^^^^^^
TypeError: argument of type 'method' is not iterable
```

nautobot/dcim/tests/test_models.py
```diff
@@ -2610,6 +2611,70 @@ def test_remove_ip_addresses(self):
...
+        platform = Platform.objects.first()
+        software_version_status = Status.objects.get_for_model(SoftwareVersion).first()
+        software_image_file_status = Status.objects.get_for_model(SoftwareImageFile).first()
+        software_version = SoftwareVersion.objects.create(
+            platform=platform, version="Test version 1.0.0", status=software_version_status
+        )
+
+        # Test that the download_url field is correctly validated with the default ALLOWED_URL_SCHEMES setting
+        for scheme in settings.ALLOWED_URL_SCHEMES:
+            software_image = SoftwareImageFile.objects.create(
```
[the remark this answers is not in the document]
garymccann: OK, implemented

nautobot/dcim/filters/__init__.py
```diff
@@ -1791,6 +1791,11 @@ class SoftwareImageFileFilterSet(NautobotFilterSet, StatusModelFilterSetMixin):
     default_image = django_filters.BooleanFilter(
         label="Is default image for associated software version",
     )
+    external_integration = NaturalKeyOrPKMultipleChoiceFilter(
```
glennmatthews, line 1794: Wrong branch? :-)

nautobot/core/models/validators.py
glennmatthews: Interesting, I would have thought `@property` would avoid that issue, but OK.

glennmatthews (approved):

nautobot/dcim/filters/__init__.py
garymccann: Yikes - where did that slip in from. Fixing

PR #7002, opened 2025-03-07
Enhance Notes display for objects

### LINKED ISSUE #4181: Enhancement of “Notes” feautre
### As ...

### I want ...

the notes overview of Devices to be more informative

### So that ...

one can gather facts easily. For that,

* the `Notes` tab in the (device) menu should be enhanced by the number of notes – just like `Interfaces` or `Power Ports`
* Notes should be sortable by creation time in the view. For that, the creation time of a note would have to be stored in the database – not only the creation date
* Updated timestamp should be able to be activated to view in the GUI … and also be sortable

### I know, this would ...

enhance the usage of notes, especially when used on automated actions in the network which have to be documented at a device / interface / IP level. Currently, automations create notes which hold the information on what was done – but are not showed in a chronological on the same date.

Let's say in a server healing procedure, the automation did a reboot which did not fix the problem and created a note „reboot initiated“ and afterwards started the automated re-initiation from scratch by running an Ansible automation which is noted by „Ansible playbook xyz started“, it's not clearly visible what the execution order was ... especially to view timings is not possible and even the creation order of notes is not respected in the view at all.

- [ ] Automation
- [ ] Circuits
- [X] DCIM
- [X] IPAM
- [ ] Organization
- [ ] Security (Secrets, etc)
- [ ] Image Management
- [ ] UI/UX
- [ ] Documentation

### Database Changes

* Creation date in database would have to be improved to a timestamp

### External Dependencies

_No response_

matt852:
# Closes #4181 
# What's Changed

Minor display enhancements requested in #4181:

- I added a counter badge to the Notes tab for objects detail view.
- I also slightly updated the Notes List table.  The "Notes" field (text contents of the note) is now the LinkColumn that opens the specific Note.  By making the "Created" field a non-Linked Column, the timestamp now displays in a more human-readable format.
- I added a "Last Updated" column.

With new Notes model changes in 2.0, no database changes were needed.

## Screenshots
1. Added count in tab name if existing notes
2. Rearranged table column ordering.  Made "Note" field linked and unlinked "Created" field (which changed the timestamp displayed to be more human readable)
3. Added "Last update" sortable column to table

### Individual object Notes tab before/after:
notes-before-screenshot
notes-after-screenshot

### Notes pages for all notes before/after:
notes-before-screenshot2
notes-after-screenshot2

- [ ] Unit, Integration Tests


changes/4181.changed
```diff
@@ -0,0 +1,2 @@
+Changed how existing Notes are displayed in table.
+Add Note count to tab for each object.
```
HanlinMiao, line 2:
```suggestion
Changed the default column ordering of Note List View table.
Added display for number of notes attached to each object on the Notes tab header in Object Detail View.
```

HanlinMiao (approved):
I am surprised that we do not have a `_ObjectDetailNotesTab` in our UI component framework.

changes/4181.changed
HanlinMiao: Probably could break out the last two entries to 4181.added?

nautobot/core/templates/generic/object_retrieve.html
```diff
@@ -117,7 +117,7 @@ <h1>
         {% if perms.extras.view_note %}
             {% if active_tab != 'notes' and object.get_notes_url or active_tab == 'notes' %}
                 <li role="presentation"{% if active_tab == 'notes' %} class="active"{% endif %}>
-                    <a href="{{ object.get_notes_url }}">Notes</a>
+                    <a href="{{ object.get_notes_url }}">Notes {% badge object.notes|length %}</a>
```
glennmatthews, line 120: Maybe?

```suggestion
                    <a href="{{ object.get_notes_url }}">Notes {% badge object.notes.count %}</a>
```

nautobot/extras/tables.py
```diff
@@ -1063,14 +1063,15 @@ def render_value(self, record):
 
 class NoteTable(BaseTable):
     actions = ButtonsColumn(Note)
-    created = tables.LinkColumn()
-    note = tables.Column(
+    created = tables.Column()
+    last_updated = tables.Column()
```
glennmatthews, line 1067: **Maybe** could use a `tables.DateTimeColumn()`?

nautobot/extras/tables.py
```diff
@@ -1063,14 +1063,15 @@ def render_value(self, record):
...
+    note = tables.LinkColumn(
```
glennmatthews, line 1068: `LinkColumn` is deprecated, so perhaps:

```suggestion
    note = tables.Column(linkify=True)
```

However, given that the content of the note is rendered Markdown, I'm concerned about how this will look/behave given a more "realistic" set of test data, e.g a multi-paragraph note with inline Markdown formatting and maybe even hyperlinks of its own??

glennmatthews (changes requested):
See concerns about the rendering of the `note` column.

nautobot/extras/tables.py
matt852: Oh, that's an interesting point I hadn't considered.  It might be best to keep the "Created Date" column as the object link, but linkify it instead like you mentioned.  I'll test it out.

glennmatthews (approved):

PR #7005, opened 2025-03-10
Fix #6902 - wrong date format in Home Page

### LINKED ISSUE #6902: Django default for SHORT_DATETIME_FORMAT used instead of Nautobot default on the HomePage.
On the Home Page Job History and Change Log panes, Nautobot uses Django default setting for `SHORT_DATETIME_FORMAT`, while Nautobot default setting has different format.
### Environment
* Nautobot version all 2.X
* Python version: 3.11

### Steps to Reproduce
Visit Home Page, check documentation for [default setting](https://docs.nautobot.com/projects/core/en/stable/user-guide/administration/configuration/settings/#short_datetime_format).

Change default setting in nautobot_config.py or set `NAUTOBOT_SHORT_DATETIME_FORMAT` env var.

### Expected Behavior
Format is applied according to SHORT_DATETIME_FORMAT

### Observed Behavior
Default Django setting is used, instead of Nautobot setting.
The issue is only on the Home Page, /extras/object-changes/ and /extras/job-results/ first column are formatted according to `SHORT_DATETIME_FORMAT` setting.

dsoftait:
# Closes #6902 
# What's Changed
Fixed formatting date in templates. Changed access to the settings variables, because in [context_processors.py](https://github.com/nautobot/nautobot/blob/develop/nautobot/core/context_processors.py) we have only defined `settings` variable.

# Screenshots
Before - different date format in Home Page vs. List View. In Home Page is format from default Django settings what is wrong.
before-homepage-job-history-changelog
before-changelog-detail-view
before-job-detail-view

After - the same correct date format in Home Page and in List View
homepage-job-history-changelog
changelog-detali-view
jobresult-detail-view

# TODO
- [ ] Unit, Integration Tests


glennmatthews (commented):
Is this the correct fix, or should we instead move away from supporting/recommending `SHORT_DATETIME_FORMAT` and its equivalents in favor of instead recommending setting the Django locale correctly?

dsoftait:
> Is this the correct fix, or should we instead move away from supporting/recommending `SHORT_DATETIME_FORMAT` and its equivalents in favor of instead recommending setting the Django locale correctly?

In my opinion we had a bug in templates where we wanted to read a variable from settings the wrong way, as a result we had a different format when we used templates and a different one when we used Django view, so this PR fixes that. But if we want to change the behavior, maybe we should open another task for this purpose?

glennmatthews:
That's fair, let's get a follow-on issue opened to deprecate these settings in the documentation.

HanlinMiao (approved):
LGTM if the follow-up issue is opened!

PR #7017, opened 2025-03-12
Backport test fixes from next

glennmatthews:
# What's Changed

Backport of test fixes that were merged to `next` in #6978.

- Because `VRFFactory` creates many-to-many relations to VDC objects now, update a couple of tests so that they no longer try to additionally (unnecessarily) create bespoke VRF-to-VDC relations of their own, as depending on the factory state, this could result in data conflicts.
- Fix state leakage between tests introduced inadvertently in #6966 that resulted in a test-specific custom validator sometimes impacting data population in other tests.

# TODO

- [x] Unit, Integration Tests


mateusz-lachowski-codilime (approved):

nautobot/extras/tests/test_plugins.py
```diff
@@ -527,22 +527,28 @@ def test_relationship_association_validator_raises_exception(self):
...
 
     def test_custom_validator_web_request_uses_real_user(self):
         location_type = LocationType.objects.get(name="Campus")
-        registry["plugin_custom_validators"]["dcim.locationtype"] = [TestUserContextCustomValidator]
-
-        with self.assertRaises(ValidationError) as context:
-            with web_request_context(user=self.user):
-                location_type.clean()
-        self.assertEqual(context.exception.message, self.user)
+        before = registry["plugin_custom_validators"]["dcim.locationtype"]
```
mateusz-lachowski-codilime, line 542: Can we possibly create some checks to avoid such things in future? At first I thought about updating `Registry` itself with some additional checks. But maybe some simple context manager would be better? Something like:

```
with registry_override('plugin_custom_validators', 'dcim.locationtype', new_value=[TestUser...]):
    # pass
```

glennmatthews: I don't think we have that many places where we're manipulating the registry contents within a test, but agreed it might be a useful longer-term enhancement if we find it's something we do need to do often.

PR #7019, opened 2025-03-12
Update cryptography to latest and drop python 3.9.0/3.9.1 support

glennmatthews:
# Closes (Dependabot alert)
# What's Changed

`cryptography` package latest releases have a fix for a CVE (https://github.com/advisories/GHSA-79v4-65xg-pq4g). However, the latest releases, while otherwise compatible back to Python 3.7, explicitly disallow installation with specific versions of Python 3.9.0 and 3.9.1 (https://github.com/pyca/cryptography/pull/12045). So if we want that CVE fix in Nautobot’s poetry.lock (note that for us it’s a transitive dependency via `paramiko` and `social-auth-core` and others, not a direct dependency) we also need to drop support for Python 3.9.0 and 3.9.1.

# TODO

- n/a Unit, Integration Tests


glennmatthews:
Not yet ready to merge - discussing downstream impacts of the Python change internally.

PR #7031, opened 2025-03-17
Added object permisison enforcement to related objects when modifying records through the REST API.

### LINKED ISSUE #5780: Permission constraints to endpoints are not applied over API when modifying record
### Environment
* Nautobot version (Docker tag too if applicable): v1.6.23
* Python version: 3.11
* Database platform, version: pgsql 15
* Middleware(s): N/A

### Steps to Reproduce
1. Create permission for /ipam/prefixes endpoint with VIEW/ADD/CHANGE [actions](https://docs.nautobot.com/projects/core/en/v1.6.19/administration/permissions/#actions), assign to test user
2. Create permission for /extras/tags endpoint with VIEW actions and a [constraint](https://docs.nautobot.com/projects/core/en/v1.6.17/administration/permissions/#constraints) applied, assign to test user
 - example constraint: `{"slug__icontains": "test-whitelisted-tag"}` - should only allow the user to object `test-whitelisted-tag` and not any other tag. Which this does work in API and UI as expected.
3. Send a POST request to update a /ipam/prefixes object with the tag by UUID

### Expected Behavior
Expecting HTTP-403 error, same as if we removed the /extras/tags permission altogether and tried to do the same record update

### Observed Behavior
The record is updated with a HTTP-200 response

### Other notes

* From the UI on object, when you go to edit and open the drop-down widget for Tags does not populate. This obeys the permission constraints as expected.

HanlinMiao:
# Closes #5780 
# What's Changed

1. Added object permisison enforcement to related objects when modifying records through the REST API.
2. Added @override_settings(EXEMPT_VIEW_PERMISSION=["*"]) to relevant tests
# Screenshots

n/a
# TODO

- [x] Unit, Integration Tests


glennmatthews:
> Looks good :) I'm just a little bit worried why you needed to add `EXEMPT_VIEW_PERMISSIONS` in some many places in tests. Was it a bug in tests because of this bug or maybe this is some kind of breaking change? 🤔

The generic API view tests generally only add permissions for the model under test; with this change we're now also enforcing permissions for other models related by FK and M2M relations. The correct thing to do (both here and in the UI View tests) would be to stop using EXEMPT_VIEW_PERMISSIONS and actually set all needed object permissions for each test case, but that would have downstream effects on all Apps using these test classes as well.

Maybe something we could add to the TODO list for v3.x?

changes/5780.added
```diff
@@ -0,0 +1 @@
+Added object permisison enforcement to related objects when modifying records through the REST API.
```
glennmatthews, line 1:
```suggestion
Added object permission enforcement to related objects when modifying records through the REST API.
```

nautobot/core/testing/api.py
```diff
@@ -702,6 +702,7 @@ def check_expected_slug(self, obj):
             else:
                 self.assertEqual(obj.key, expected_slug)
 
+        @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
```
glennmatthews, line 705: Maybe add a `# TODO` comment on each of these indicating the plan for a long-term fix to the tests to use proper permissions instead?

nautobot/core/tests/test_authentication.py
```diff
@@ -383,7 +383,16 @@ def test_create_object(self):
         )
         obj_perm.users.add(self.user)
         obj_perm.object_types.add(ContentType.objects.get_for_model(Prefix))
-
+        related_obj_perm = ObjectPermission.objects.create(
+            name="Related object permission",
+            actions=["view", "add", "change"],
```
glennmatthews, line 388: Are all three actions actually needed here (and if so, why?) or can we narrow this in scope? Same question below.

nautobot/core/tests/test_authentication.py
```diff
@@ -383,7 +383,16 @@ def test_create_object(self):
...
+        )
+        related_obj_perm.users.add(self.user)
+        related_obj_perm.object_types.add(
+            ContentType.objects.get_for_model(Namespace),
+            ContentType.objects.get_for_model(Status),
+            ContentType.objects.get_for_model(Location),
+        )
```
glennmatthews, line 395: Feels like we should also *add* some test steps here and below to verify the issue being fixed - i.e. user with `obj_perm` but without `related_obj_perm` should also be attempted and fail as well.

nautobot/core/tests/test_api.py
```diff
@@ -631,6 +631,7 @@ def setUp(self):
         self.vlan_group2 = ipam_models.VLANGroup.objects.create(name="Test VLANGroup 2", location=self.location2)
         self.vlan_group3 = ipam_models.VLANGroup.objects.create(name="Test VLANGroup 3", location=self.location3)
 
+    @override_settings(EXEMPT_VIEW_PERMISSIONS=["*"])
```
glennmatthews, line 634: I'd prefer (even though it's more work) to only add EXEMPT_VIEW_PERMISSIONS to the generic tests in `testing/api.py` (since those are the ones with downstream impact to Apps). All of the core-exclusive tests like this one should instead be fixed with correct object permissions instead.

nautobot/core/tests/test_authentication.py
HanlinMiao: Seems like "view" action is sufficient here. I will remove the rest.

nautobot/core/tests/test_authentication.py
glennmatthews: It occurs to me also that we need to test the impact of this change specifically on GET and PATCH requests, for example:

- If I have view permissions for Location but not for Status, can I GET `/api/dcim/locations/` or does that now reject since I can't view the location `status`?
- If I have view and edit permission for Location but not for Status, can I PATCH `/api/dcim/locations/<UUID>/` with a new `name` value (not changing its `status`) or does that now reject?

HanlinMiao: Added some tests to confirm these expected behaviors

glennmatthews (approved):

PR #7033, opened 2025-03-17
HTMX demo for nautobot list views

### LINKED ISSUE #6816: Spike: HTMX two column layout investigation
Investigate using HTMX to implement a basic two column layout, where the nav bar is not rerendered on every request, only the actual list/detail view content on the right.

The outcome of this story is to understand the feasibility and what will be impacted in terms of how views are written and maintained, how the UI component framework in impacted, etc.

The goal of this spike is not to have a final working solution, but rather to have a general understanding of what the pattern will look like, focusing on the backend, so we can get a feel for what the level of effort and impact will be to go forward with HTMX as a fundamental technology.

mateusz-lachowski-codilime:
# Closes #6816 
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests


nautobot/core/templates/generic/object_list.html
```diff
@@ -194,8 +194,8 @@ <h1>{% block title %}
 </div>
 <hr>
 {% endif %}
-
-<div class="row">
+<script src="https://unpkg.com/htmx.org@1.9.2"></script>
```
glennmatthews, line 197: Note: In a "finished" version of this code, we'd want to add `htmx.org` to our `package.json` (currently only in the bootstrap 5 feature branch) and have it compiled in with the rest of the Nautobot distribution.

nautobot/core/views/generic.py
```diff
@@ -402,6 +403,9 @@ def get(self, request):
         setattr(self, "request", request)
         context.update(self.extra_context())
 
+        if request.headers.get('HX-Request'):
```
glennmatthews, line 406: Did you investigate the possibility of using django-htmx? Any arguments against it?

nautobot/core/templates/generic/object_list.html
glennmatthews, line 197: Any particular reason for not using the current 2.0.x release train of htmx?

nautobot/core/views/generic.py
mateusz-lachowski-codilime: Yes,  I did investigation and analyze using it. It's great helper package wrapping different HTMX functionalities into python-django world. Here I just wanted to show what this package does in the background. The same code with using it would be:
```
if request.htmx:
```

nautobot/core/templates/generic/object_list.html
mateusz-lachowski-codilime: Yes, definitely!

PR #7034, opened 2025-03-17
feature: implement base v3 ui sidenav

### LINKED ISSUE #6947: Navbar phase 1
Implement the basic scripture of the navbar void of the actual nav elements until we decide on which direction to go. So this story covers the nav bar placement, logo, and collapsibility to the minimized version, so we can unblock over layout work.

norbert-mieczkowski-codilime:
# Closes #6947
# What's Changed
Implement base v3 UI sidenav (a.k.a. navbar, nav menu, sidebar). It doesn't include accordions or fly-outs.

I initially considered three solutions to implement navigation:
1. Bootstrap navbar component - sadly Bootstrap 5 navbar still doesn't support placement on the sides and would require plenty of overrides.
2. Building the component on top of utility classes (https://getbootstrap.com/docs/5.2/examples/sidebars/) - while Bootstrap 5 utility classes could likely account for the majority of navbar looks, given the complexity of Nautobot navbar they would not suffice to get everything right.
3. **Create a set of separate in-house CSS classes (https://getbootstrap.com/docs/5.2/examples/dashboard/) - perhaps the most verbose and most unwanted of all listed here, but at the same time the cleanest, because the whole navbar style ends up in a single place and is manageable in its entirety rather than being either existing classes overrides or extensions (along with class definitions too anyway).**
# Screenshots
| Expanded | Collapsed |
| - | - |
| image | image_2 |


nautobot/core/templates/inc/nav_menu.html
```diff
@@ -3,6 +3,73 @@
 {% load helpers %}
 {% registry %}
 
+<nav class="sidenav" id="sidenav">
```
norbert-mieczkowski-codilime, line 6: @glennmatthews, I renamed the side navigation to `sidenav` because to me it feels like the most accurate name and at the same time there is no requirement to keep backward compatibility with naming, structure or anything. Let me know if that works for you or if you'd prefer to have it called `sidebar`. I'm just opposed to `navbar` because it's an already existing Bootstrap 5 component looking way different than what we're having here (as mentioned in PR description).

nautobot/core/templates/inc/nav_menu.html
```diff
@@ -3,6 +3,73 @@
...
+    ></button>
+
+    <a class="sidenav-brand" href="{% url 'home' %}">
+        <img src="{% custom_branding_or_static 'icon' 'img/nautobot_icon.svg' %}" />
+    </a>
+    <a class="position-absolute sidenav-brand" href="{% url 'home' %}">
+        <img src="{% custom_branding_or_static 'logo' 'img/nautobot_logo.svg' %}" />
+    </a>
+
+    <ul class="list-unstyled my-0 overflow-auto pt-10 w-100">
```
glennmatthews, line 22: Should there be a single custom class for this element that combines all of these individual styles? Or is that not the preferred idiom in bootstrap 5?

nautobot/project-static/src/scss/nautobot.scss
```diff
@@ -542,3 +544,227 @@ $btn-variants: (
...
+  #main-content {
+    grid-area: main-content;
+  }
+
+  #footer {
+    grid-area: footer;
+  }
+}
+
+.sidenav {
```
glennmatthews, line 571: Do we anticipate having multiple `sidenav` class elements (maybe as part of the megamenu design) or should this be `#sidenav` as a unique id instead?

nautobot/project-static/src/scss/nautobot.scss
```diff
@@ -542,3 +544,227 @@ $btn-variants: (
...
+  align-items: center;
+  background-color: $navy-0;
+  color: $nav-white-0;
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  position: relative;
+  width: 15rem; /* 240px */
+
+  &#{&}-collapsed {
```
glennmatthews, line 618: Mind adding a comment explaining this syntax?

glennmatthews (commented):
Couple of quick comments - I'll try to pull it down locally and give it a spin later today or early tomorrow.

nautobot/core/templates/inc/nav_menu.html
```diff
@@ -3,6 +3,73 @@
...
+                        </button>
+
+                        {% for group_name, group_details in tab_details.groups.items %}
+                            {% if not group_details.permissions or request.user|has_one_or_more_perms:group_details.permissions %}
+                                <ul class="list-unstyled">
+                                    <li class="sidenav-link-group" data-group-weight="{{ group_details.weight }}">{{ group_name }}</li>
+                                    {% for item_link, item_details in group_details.items.items %}
+                                        {% if not item_details.permissions or request.user|has_one_or_more_perms:item_details.permissions %}
+                                            <li {% if item_details.permissions and not request.user|has_perms:item_details.permissions %} class="disabled"{% endif %}>
+                                                <a class="sidenav-link{% if item_link in request.path %} sidenav-link-active{% endif %}" data-item-weight="{{ item_details.weight }}" href="{{ item_link }}">
```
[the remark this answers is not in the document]
glennmatthews: If we're OK with rerendering the sidenav as part of every page load, the 'active link' could just be a part of the render context that each view sets (would be a new requirement for custom views, but this is for 3.0 so that's OK as a "breaking" change). If we're wanting to do something non-rerendering (e.g. HTMX), we might be able to make the 'active link' a part of the HTMX response and use some simple JS to deactivate the old and activate the new?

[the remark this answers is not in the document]
glennmatthews: One flaw I noticed with this current implementation is that there can be cases where multiple links get highlighted, for example "Example Menu -> Example Models" shares a URL with "Apps -> Example Models" (they're actually the same view just placed more than once in the nav) but as a result clicking either one results in multiple distinct nav menu entries being highlighted simultaneously. Not sure what our best approach is in the case of truly duplicate links like this, but at least we should consider what happens if there's a more-specific and less-specific form of the same link both present (e.g. `/apps/ssot/` and `apps/ssot/some-specific-data-source/`)

nautobot/project-static/js/nav_menu.js
```diff
@@ -1,250 +1,13 @@
...
-            navbar.classList.remove('collapsed');
-            toggler.style.left = '225px';
-            expandedByHover = true; // Set flag when expanded by hover
-        } else if (expandedByHover && e.clientX > 240) {
-            navbar.classList.add('collapsed');
-            toggler.style.left = '-5px';
-            expandedByHover = false; // Reset flag after auto-collapse
-        }
+        const sidenav = document.getElementById(controls);
+        sidenav.classList.toggle('sidenav-collapsed', expanded)
```
glennmatthews, line 11: One thing I noticed with this toggler implementation (at least in Firefox) is that the Nautobot/N logos move around oddly as the sidebar is collapsing/expanding, "hopping upward" for a fraction of a second as `Nautobot` converts to `N` or vice versa.

nautobot/project-static/src/scss/nautobot.scss
```diff
@@ -542,3 +544,227 @@ $btn-variants: (
...
+
+  @mixin fade-slide-out {
+    opacity: 0;
+    visibility: hidden;
+    transform: translateY(-100%);
+  }
+
+  @mixin filter-nav-orange-0 {
+    /*
+     * CSS workaround to change background-image from white to any given color.
```
glennmatthews, line 593: Should that say `background-color` or even `color` rather than `background-image`?

nautobot/project-static/src/scss/nautobot.scss
glennmatthews, line 618: Is it currently "expected" that collapsing the sidenav doesn't reduce it to an "icons only" view but instead contains the same content, just truncated?

image

nautobot/project-static/src/scss/nautobot.scss
```diff
@@ -542,3 +544,227 @@ $btn-variants: (
...
+        @include fade-slide-out;
+      }
+    }
+
+    .sidenav-toggler::before {
+      transform: rotate(180deg);
+    }
+  }
+
+  &, * {
```
glennmatthews, line 642: maybe a comment explaining the meaning of this too? :-)

nautobot/project-static/src/scss/nautobot.scss
```diff
@@ -542,3 +544,227 @@ $btn-variants: (
...
+
+    .sidenav-brand {
+      &:first-of-type {
+        @include fade-slide-in;
+      }
+
+      &:nth-of-type(2) {
+        @include fade-slide-out;
+      }
+    }
```
glennmatthews, line 629: Can you add some comments explaining what this part does?

nautobot/core/templates/inc/nav_menu.html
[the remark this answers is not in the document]
norbert-mieczkowski-codilime: I think for the case when two links have the exact same URL, regardless of being in different menus, both should be highlighted. Alternatively we could make one of them canonical and skip the others in these kinds of checks. Figuring out which one out of two (or more) should be active will always be prone to errors and false-positives or -negatives.

nautobot/core/templates/inc/nav_menu.html
norbert-mieczkowski-codilime: Since I could handle this element style with just Bootstrap 5 utility classes, I didn't see a reason to create its own class name. I don't have a strong preference though, moving its styles to `nautobot.scss` is a trivial task anyway

nautobot/project-static/src/scss/nautobot.scss
norbert-mieczkowski-codilime: right, that sounds like good reasoning, I removed the class name in favor of id

nautobot/project-static/src/scss/nautobot.scss
norbert-mieczkowski-codilime: since I removed the `.sidenav` class name, the selector doesn't use this obscure syntax anymore

nautobot/project-static/js/nav_menu.js
norbert-mieczkowski-codilime: it's an intended slide behavior I am generally used to implementing but it's something that wasn't discussed before. if you prefer the items to just fade out with no slide transition, it's just adding or removing two lines of SCSS anyway. here's comparison:

| fade-slide | fade |
| - | - |
| [fade-slide.webm](https://github.com/user-attachments/assets/a919cf11-16ef-41d0-90cc-c4f5e7aea24b) | [fade.webm](https://github.com/user-attachments/assets/66ea20af-0441-463f-bf04-b04e9b4fbb36) |

nautobot/project-static/src/scss/nautobot.scss
norbert-mieczkowski-codilime: cI elaborated a bit more on the comment, but to get into even more detail - `background-image` with and url-encoded svg icon (as Bootstrap 5 does and as I did here as well, following the pattern) embeds an icon with hardcoded color which is not changeable by any other CSS property other than `filter`, that's why I specifically referred to `background-image` in the comment. but you're right that `filter` will affect all element colors

nautobot/project-static/src/scss/nautobot.scss
norbert-mieczkowski-codilime: good point, I was trying not to get in a class name hell with each element having its own, and then also variations and states. anyway from what you're saying it sounds like ultimately using `-child` and `-of-type` selectors got me anyway. I commented each one of them to improve their readability

nautobot/project-static/js/nav_menu.js
mateusz-lachowski-codilime: Fade version is better :)

nautobot/core/templates/inc/nav_menu.html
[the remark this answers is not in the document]
mateusz-lachowski-codilime: I think that anyway we need to support both cases: rendering 100% from backend and just updating the highlighted item. First page load will always relay purely on backend. Same goes to some reload cases. 

Updating on FE it's pretty simple: on HTMX event just remove and add active-link class in different place. On the backend we will need some logic to maybe mark sections as active depending on some sort identifiers? We also need to update navbar generation system to support favorites or different structure in new UI so we can plan those things together :)

mateusz-lachowski-codilime:
I think that we can safely merge this without having a passing integration or unit tests. It should be handled in separate task.

nautobot/project-static/js/nav_menu.js
norbert-mieczkowski-codilime: np, I removed slide from the transition 77aab4fd100d4e13e67e0d7b665e2e9ffcc8298a

glennmatthews (approved):

PR #7037, opened 2025-03-17
Lock GHA actions to hashes, add enforcement

msheiny:
# What's Changed

In [response to recent security news](https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised) in the Github Actions ecosystem. This PR takes some precautionary measures to try and  protect nautobot runner environments.

Particularly I've:

* Converted all Github actions to use actions from tags --> hashes on repositories we don't "trust". Open to debate on which repos we trust but first pass I assumed github `actions/*` and anything nautobot adjacent.
* Added a `security-checks` action to the repo, and added a job in most of the files to run this action. This in particular added `zgosalvez/github-actions-ensure-sha-pinned` which will throw build errors if we are using github actions without a hash. This serves as a linter in the future to try to and prevent tags getting reintroduced inadvertently. It looks like it scans all the github actions in the repo based on the output.

# Motivation

The recent issue with [tj-actions/changed-files](https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised) shows that using tags in github actions is a recipe for a security incident - particularly on repos we don't have a lot of trust with. Note that I also took this guidance directly from the [Github action hardening guide](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions):

> Pin actions to a tag only if you trust the creator
> Although pinning to a commit SHA is the most secure option, specifying a tag is more convenient and is widely used. If you’d like to specify a tag, then be sure that you trust the action's creators. The ‘Verified creator’ badge on GitHub Marketplace is a useful signal, as it indicates that the action was written by a team whose identity has been verified by GitHub. Note that there is risk to this approach even if you trust the author, because a tag can be moved or deleted if a bad actor gains access to the repository storing the action.

and

> You can help mitigate this risk by following these good practices:

> Pin actions to a full length commit SHA

> Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload. When selecting a SHA, you should verify it is from the action's repository and not a repository fork.

# Screenshots

This is an example of what `zgosalvez/github-actions-ensure-sha-pinned-actions` looks like if it fails in CI:

image

# TODO


.github/actions/build-nautobot-image/action.yaml
```diff
@@ -57,7 +57,7 @@ runs:
   steps:
     - name: "Setup Metadata"
       id: "metadata"
-      uses: "docker/metadata-action@v4"
+      uses: "docker/metadata-action@818d4b7b91585d195f67373fd9cb0332e31a7175" # v4
```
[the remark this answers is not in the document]
msheiny: Yeah, looks like [renovate does](https://docs.renovatebot.com/modules/manager/github-actions/).

Looks like [dependabot does as well](https://github.com/dependabot/dependabot-core/pull/5951).

.github/workflows/ci_integration.yml
```diff
@@ -65,6 +73,8 @@ jobs:
     steps:
       - name: "Check out repository code"
         uses: "actions/checkout@v4"
+      - name: "Security checks"
+        uses: "./.github/actions/security-checks"
```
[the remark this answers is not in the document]
msheiny: Ahh yeah good eye - ill remove it here

.github/workflows/ci_integration.yml
```diff
@@ -11,6 +11,14 @@ on:  # yamllint disable
   workflow_dispatch:
 
 jobs:
+  security-tooling:
+    name: "Security Tooling"
+    runs-on: "ubuntu-22.04"
```
glennmatthews, line 16: Can we use 24.04 instead here and below?

.github/workflows/plugin_upstream_testing_base.yml
```diff
@@ -27,6 +27,8 @@ jobs:
         uses: "actions/checkout@v4"
         with:
           ref: "${{ matrix.nautobot-version == 'develop' && 'develop' || matrix.nautobot-version == 'ltm-1.6' && 'ltm-1.6' || env.GITHUB_REF_NAME || 'develop' }}"
+      - name: "Security checks"
+        uses: "./.github/actions/security-checks"
```
glennmatthews, line 31: Is this going to make all Apps start failing the upstream testing until we make a similar change to the App cookiecutter and push it out with driftmanager?

.github/workflows/prerelease.yml
```diff
@@ -18,6 +18,8 @@ jobs:
     steps:
       - name: "Check out repository code"
         uses: "actions/checkout@v4"
+      - name: "Security checks"
+        uses: "./.github/actions/security-checks"
```
glennmatthews, line 22: I'm curious as to whether this actually adds value in the `prerelease` and `release` workflows. I'd like to reduce the risk of actions failing when we're trying to publish a release or prerelease.

.github/workflows/plugin_upstream_testing_base.yml
msheiny: Yes - I didn't realize how this workflow was used. I can also remove this.

.github/workflows/prerelease.yml
msheiny: it makes more sense in my mind that the prerelease and release workflows are protected since they utilize the most sensitive NTC credentials. I don't see a way to limit that check to just this file though. I see your point of why fail this release/prerelease workflow if another workflow file we don't care as much about is failing. I could fork that upstream repo to add an explicit file input?


my longer reply before i re-read your question and better understood your POV being wary of a check that fails for a non-relevant workflow file.
----
Well the hash locks make the most sense in three scenarios in my mind:

* wherever we use any kind of sensitive credentials in a workflow step - especially anything with write permissions to push containers, pypi packages, and make commits to a repo.
* anywhere we are dealing with non open-source repo code we want to limit release
* a workflow where immutability is an important trait - a guard against bugs being introduced that break the workflow. where if we re-run a workflow multiple times we have a better guarantee that the GHA action logic running is the same. leaving some stuff like `actions/` un-hashed breaks this immutability but my intent with this PR was to hit the first two points.


I did a quick search of the github action files for `secrets.` and that came up with:
* `build_dependency_image.yml`
* `prerelease.yml`
* `release.yml`
* `ci_integration.yml`
* `ci_pullrequest.yml`

Though I'm looking closer at `ci_integration.yml` and `ci_integration.yml` it looks like those creds are only for pulling github container registry images. I have no idea if those are read-only with my limited access.

msheiny: @glennmatthews and i spoke out of band - I am removing this check on release/prerelease since the integration source branches will be covered by the ci_integration/pullrequest workflows.

glennmatthews (approved):
I'm trusting you that the provided hashes are correct. :-)

Issue #446, opened 2021-05-13
Restart required to view new Relationship in GraphQL


pke11y:
### Environment
* Python version: 3.8.5
* Nautobot version: 1.0.1

### Steps to Reproduce
1. Create a new Relationship e.g. between dcim.device and ipam.vlan called `device_vlan`
2. Go to the GraphiQL interface
3. Query for all devices and include the relationship attribute created in step 1 e.g. 

```
{
  devices{
    rel_device_vlan {
      id
    }
  }
}
```

### Expected Behavior
The relationship attribute (e.g. `rel_device_vlan`) is available for selection the GraphIQL interface and is processed and values returned as part of the query.

### Observed Behavior
The relationship attribute (e.g. `rel_device_vlan`) is not available on the models as per definition. The nautobot service needs to be restarted for the relationship to be loaded and be available for querying.

```
{
  "errors": [
    {
      "message": "Cannot query field \"rel_device_vlan\" on type \"DeviceType\".",
      "locations": [
        {
          "line": 37,
          "column": 5
        }
      ]
    }
  ]
}
```

jedelman8:
Would like to make this broader in scope in that we support the ability to restart Nautobot services from the UI. Will determine if that needs to be a seperate issue.

pke11y:
Yeah makes sense. Any change for this would be quite significant.

glennmatthews:
Probably also an issue with other "dynamic" schema content - custom fields, computed fields in particular.

hellerve:
I worked on this for a while now, sadly none of the approaches that we could come up with proved viable. For posterity, let me document them:

- The simplest way to make this work is to generate the schema everytime it is needed (i.e. on every API call). Unfortunately, this is prohibitively slow (on the order of seconds).
- It would also be possible to only recalculate it when it changes, based on Django signals. This proved not to be viable, either, since it would break if multiple instances of Nautobot are deployed behind a load balancer; only the instance saving the model (Relationship or CustomField) would update the API and queries to the GraphQL API would become non-deterministic.
- Finally, we could cache the schema in the database as a singleton, pickled. This does not work, either, since the classes are generated on-the-fly, and pickle does not work well with that. The class definition itself is not pickled, so any instance of NautoBot other than the one that generated the Schema would not be able to unpickle the definition (furthermore, local types are generated, and pickle does not work at all with non-global types).

[...]

PR #7046, opened 2025-03-20
6998 Added caches for openapi schema endpoint

### LINKED ISSUE #6998: OpenAPI/Swagger Docs Large Spec Loading Slow
Wasn't sure if this would be better reported as a feature or a bug, but adding as a bug because I'm not sure the best way to handle options for shrinking the size of the generated spec and whether that constitutes a bug or a feature (e.g. using common schemas, splitting the spec, optionally excluding some data based on an env var or something) but wanted to get it on y'alls radar and see if there's ways to make it a little more user-friendly. 

### Environment
* Nautobot version (Docker tag too if applicable): 2.4.4
* Python version: 3.11
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Visit the API documentation page in Nautobot (https://demo.nautobot.com/api/docs/) 
2. AND/OR Attempt to download the 9MB spec YAML file (https://demo.nautobot.com/api/swagger/?api_version=2.4)  

### Expected Behavior
The API page to render quickly.

### Observed Behavior
Depending on the amount of plugins installed, the API page can take 15-30 seconds to load in my experience.

mateusz-lachowski-codilime:
# Closes #6998 
# What's Changed
- Added browser caching and server side caching for OpenApi schema endpoint
- Added gzip for OpenApi schema endpoint. There was no speed gain in my tests but I bet there will be on slower network connections.

With combined browser cache and backend cache, we have three options how cache will be used depending on the cache status:
- cache expired on browser, but Etag is the same (no changes in nautobot) -> 70-100ms response
- cache expired on browser, Etag is different, cache present on backend -> 400-600ms response
- cache expired on browser, Etag is different, no cache on backend -> 3-4s response 

# Screenshots

Before:
image

After:
image

# TODO

- Unit, Integration Tests


nautobot/core/api/views.py
```diff
@@ -608,6 +611,41 @@ class NautobotSpectacularRedocView(APIVersioningGetSchemaURLMixin, SpectacularRe
...
+        if not schema:
+            generator = self.generator_class(urlconf=self.urlconf, api_version=version, patterns=self.patterns)
+            schema = generator.get_schema(request=request, public=self.serve_public)
+            cache.set(cache_key, schema, 60 * 60 * 24 * 7)
+
+        return Response(
+            data=schema,
+            headers={
+                "Content-Disposition": f'inline; filename="{self._get_filename(request, version)}"',
+                "Cache-Control": f"max-age={3 * 24 * 60 * 60}, public",
```
mateusz-lachowski-codilime, line 642: There is no reasoning behind those expiration cache times - just random thoughts. Ideally we should match those times to the Nautobot release date and keep browser cache shorter. It's the fastest cache but with almost no control on our side. After that period browser will be requesting backend to validate ETag and requests will be around 100ms - just network/server latency, response is just 304 Not modified in such cases.

nautobot/core/api/urls.py
```diff
@@ -55,7 +55,7 @@
     path("docs/", NautobotSpectacularSwaggerView.as_view(url_name="schema"), name="api_docs"),
     path("redoc/", NautobotSpectacularRedocView.as_view(url_name="schema"), name="api_redocs"),
     path("settings-schema/", SettingsJSONSchemaView.as_view(), name="setting_schema_json"),
-    path("swagger/", SpectacularAPIView.as_view(), name="schema"),
+    path("swagger/", NautobotSpectacularAPIView.as_view(), name="schema"),
     path("swagger.json", SpectacularJSONAPIView.as_view(), name="schema_json"),
     path("swagger.yaml", SpectacularYAMLAPIView.as_view(), name="schema_yaml"),
```
mateusz-lachowski-codilime, line 60: I wasn't sure if we want to update those two endpoints as well. Only the `swagger/` endpoint is involved in `/api/docs/` generation so I've updated only this endpoint.

glennmatthews: Not sure if anyone's using swagger.yaml but I'd imagine some dynamic client generators may make use of swagger.json. Can always do that later though.

nautobot/core/api/views.py
```diff
@@ -608,6 +611,41 @@ class NautobotSpectacularRedocView(APIVersioningGetSchemaURLMixin, SpectacularRe
...
+        # With combined browser cache and backend cache, we have three options:
+        # - cache expired on browser, but Etag is the same (no changes in nautobot) -> 70-100ms response
+        # - cache expired on browser, Etag is different, cache present on backend -> 400-600ms response
+        # - cache expired on browser, Etag is different, no cache on backend -> 3-4s response
+
+        if_none_match = request.META.get("HTTP_IF_NONE_MATCH", "")
+        if if_none_match == etag:
+            return Response(status=304)
+
+        schema = cache.get(cache_key)
```
glennmatthews, line 632: To avoid introducing issues like #446, we'll need to make sure to invalidate this cache when creating/updating/deleting CustomField and Relationship records. Should be able to use some of the existing extras signals for this purpose.

nautobot/extras/signals.py
```diff
@@ -125,6 +125,17 @@ def invalidate_relationship_models_cache(sender, **kwargs):
             cache.delete_pattern(f"{method.cache_key_prefix}.*")
 
 
+@receiver(post_save, sender=CustomField)
+@receiver(post_delete, sender=CustomField)
```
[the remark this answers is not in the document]
mateusz-lachowski-codilime: `post_save` should be enough here. I'm more worried about `Relationship` dependent signals that there might more signals needed.

PR #7047, opened 2025-03-20
Refactored VLANGroups

Pavan-183:
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/ipam/forms.py
+++ b/nautobot/ipam/forms.py
@@ -744,6 +744,21 @@ class IPAddressFilterForm(NautobotFilterForm, TenancyFilterForm, StatusModelFilt
 #
 
 
+class VLANGroupBulkEditForm(NautobotBulkEditForm):
+    """Bulk edit form for VLANGroup objects."""
+
+    pk = forms.ModelMultipleChoiceField(queryset=VLANGroup.objects.all(), widget=forms.MultipleHiddenInput())
+    description = forms.CharField(max_length=VLANGroup._meta.get_field("description").max_length, required=False)
+    name = forms.CharField(max_length=VLANGroup._meta.get_field("name").max_length, required=False)
+    location = forms.ModelChoiceField(queryset=Location.objects.all(), required=False)
+    range = forms.CharField(max_length=VLANGroup._meta.get_field("range").max_length, required=False)
+
+    class Meta:
+        model = VLANGroup
+        fields = ["description", "location", "range"]
+        nullable_fields = ["location", "description"]
+
+
 class VLANGroupForm(LocatableModelFormMixin, NautobotModelForm):
     class Meta:
         model = VLANGroup
--- a/nautobot/ipam/urls.py
+++ b/nautobot/ipam/urls.py
@@ -8,7 +8,6 @@ from .models import (
     IPAddress,
     Prefix,
     VLAN,
-    VLANGroup,
 )
 
 app_name = "ipam"
@@ -20,6 +19,7 @@ router.register("rirs", views.RIRUIViewSet)
 router.register("route-targets", views.RouteTargetUIViewSet)
 router.register("services", views.ServiceUIViewSet)
 router.register("vrfs", views.VRFUIViewSet)
+router.register("vlan-groups", views.VLANGroupUIViewSet)
 
 urlpatterns = [
     # Namespaces
@@ -154,42 +154,6 @@ urlpatterns = [
         "ip-addresses/<uuid:pk>/delete/",
         views.IPAddressDeleteView.as_view(),
         name="ipaddress_delete",
-    ),
-    # VLAN groups
-    path("vlan-groups/", views.VLANGroupListView.as_view(), name="vlangroup_list"),
-    path("vlan-groups/add/", views.VLANGroupEditView.as_view(), name="vlangroup_add"),
-    path(
-        "vlan-groups/import/",
-        views.VLANGroupBulkImportView.as_view(),  # 3.0 TODO: remove, unused
-        name="vlangroup_import",
-    ),
-    path(
-        "vlan-groups/delete/",
-        views.VLANGroupBulkDeleteView.as_view(),
-        name="vlangroup_bulk_delete",
-    ),
-    path("vlan-groups/<uuid:pk>/", views.VLANGroupView.as_view(), name="vlangroup"),
-    path(
-        "vlan-groups/<uuid:pk>/edit/",
-        views.VLANGroupEditView.as_view(),
-        name="vlangroup_edit",
-    ),
-    path(
-        "vlan-groups/<uuid:pk>/delete/",
-        views.VLANGroupDeleteView.as_view(),
-        name="vlangroup_delete",
-    ),
-    path(
-        "vlan-groups/<uuid:pk>/changelog/",
-        ObjectChangeLogView.as_view(),
-        name="vlangroup_changelog",
-        kwargs={"model": VLANGroup},
-    ),
-    path(
-        "vlan-groups/<uuid:pk>/notes/",
-        ObjectNotesView.as_view(),
-        name="vlangroup_notes",
-        kwargs={"model": VLANGroup},
     ),
     # VLANs
     path("vlans/", views.VLANListView.as_view(), name="vlan_list"),
--- a/nautobot/ipam/views.py
+++ b/nautobot/ipam/views.py
@@ -48,10 +48,10 @@ from .models import (
     VLANGroup,
     VRF,
 )
+from .tables import VLANDetailTable
 from .utils import (
     get_add_available_ipaddresses_callback,
     get_add_available_prefixes_callback,
-    get_add_available_vlans_callback,
     handle_relationship_changes_when_merging_ips,
     retrieve_interface_or_vminterface_from_request,
 )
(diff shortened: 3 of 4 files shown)


nautobot/ipam/views.py
```diff
@@ -48,10 +48,10 @@
     VLANGroup,
     VRF,
 )
+from .tables import VLANDetailTable
```
glennmatthews, line 51: We already have `from . import tables`, please use `tables.VLANDetailTable`

nautobot/ipam/views.py
```diff
@@ -1161,75 +1161,32 @@ def list(self, request, *args, **kwargs):
...
 
-    def get_extra_context(self, request, instance):
-        vlans = (
-            VLAN.objects.restrict(request.user, "view")
-            .filter(vlan_group=instance)
-            .prefetch_related(Prefetch("prefixes", queryset=Prefix.objects.restrict(request.user)))
-        )
-        vlans_count = vlans.count()
-
-        data_transform_callback = get_add_available_vlans_callback(show_available=True, vlan_group=instance)
```
glennmatthews, line 1184: Need to preserve this functionality in the vlans table (augmenting the `vlan_group.vlans` queryset with "available" not-yet-allocated vlans).

glennmatthews (changes requested):

dsoftait:
Hi @Pavan-183 any updates on this issue? Do you need clarification or assistance?

Pavan-183:
> Hi @Pavan-183 any updates on this issue? Do you need clarification or assistance?

Hi @dsoftait, Actually Hanlin made a few changes to the code, and I initially assumed he would be continuing with this task. I was uncertain whether I should proceed with it or not, which is why I’ve been focusing on another view.

nautobot/ipam/views.py
```diff
@@ -1147,13 +1148,14 @@ def get_extra_context(self, request, instance):
...
             }
             RequestConfig(request, paginate).configure(vlan_table)
-
             context.update(
                 {
                     "first_available_vlan": instance.get_next_available_vid(),
                     "bulk_querystring": f"vlan_group={instance.pk}",
                     "vlan_table": vlan_table,
                     "vlans_count": vlans_count,
+                    "filtered_vlans": excluded_vlans,
```
HanlinMiao, line 1157: Remove

nautobot/ipam/forms.py
```diff
@@ -744,6 +744,20 @@ class IPAddressFilterForm(NautobotFilterForm, TenancyFilterForm, StatusModelFilt
 #
 
 
+class VLANGroupBulkEditForm(NautobotBulkEditForm):
+    """Bulk edit form for VLANGroup objects."""
+
+    pk = forms.ModelMultipleChoiceField(queryset=VLANGroup.objects.all(), widget=forms.MultipleHiddenInput())
+    description = forms.CharField(max_length=CHARFIELD_MAX_LENGTH, required=False)
+    name = forms.CharField(max_length=CHARFIELD_MAX_LENGTH, required=False)
```
glennmatthews, line 752: `name` is globally unique, so it doesn't make sense to bulk-edit.

nautobot/ipam/forms.py
```diff
@@ -744,6 +744,20 @@ class IPAddressFilterForm(NautobotFilterForm, TenancyFilterForm, StatusModelFilt
...
+    location = forms.ModelChoiceField(queryset=Location.objects.all(), required=False)
```
glennmatthews, line 753: Instead of this declaration, the form should use `LocatableModelBulkEditFormMixin` as a mixin class.

nautobot/ipam/tables.py
```diff
@@ -737,6 +737,7 @@ class VLANDetailTable(VLANTable):
     tags = TagColumn(url_name="ipam:vlan_list")
 
     class Meta(VLANTable.Meta):
+        model = VLAN
```
glennmatthews, line 740: This addition shouldn't be needed given it's inheriting from `VLANTable.Meta` which already sets this same attribute value?

nautobot/ipam/urls.py
```diff
@@ -20,6 +19,7 @@
 router.register("route-targets", views.RouteTargetUIViewSet)
 router.register("services", views.ServiceUIViewSet)
 router.register("vrfs", views.VRFUIViewSet)
+router.register("vlan-groups", views.VLANGroupUIViewSet)
```
glennmatthews, line 22:
```suggestion
router.register("vlan-groups", views.VLANGroupUIViewSet)
router.register("vrfs", views.VRFUIViewSet)
```

glennmatthews (commented):
Shouldn't this PR be also removing some/all content from `nautobot/ipam/templates/ipam/vlangroup.html`?

nautobot/ipam/forms.py
itdependsnetworks: a search of docs.nautobot.com does not show this. Can you advise where we can find a list of these? is it simply https://github.com/nautobot/nautobot/blob/8e17b8e5e0bdf949c5fc297441bd168128983afa/nautobot/dcim/form_mixins.py

Where should we add docs to docs.nautobot.com for something like this?

nautobot/ipam/tables.py
Pavan-183: Okay @glennmatthews

nautobot/ipam/urls.py
Pavan-183: Okay @glennmatthews

nautobot/ipam/forms.py
Pavan-183: Okay @glennmatthews will remove it.

nautobot/ipam/forms.py
glennmatthews: Should be added somewhere under `docs/development/core`, either in `best-practices.md` or as a specific new page for working with location-related models.

mateusz-lachowski-codilime: Are the docs updated or should we create some task to not lost this update?

mateusz-lachowski-codilime (approved):

nautobot/ipam/forms.py
mateusz-lachowski-codilime: Found this as #7125 :)

nautobot/core/ui/object_detail.py
```diff
@@ -785,6 +785,8 @@ def get_extra_context(self, context: Context):
         request = context["request"]
         if self.context_table_key:
             body_content_table = context.get(self.context_table_key)
+            body_content_table_class = type(body_content_table)
+            body_content_table_model = body_content_table_class.Meta.model
```
glennmatthews, line 789: Is this two-liner necessary or can we just do `body_content_table_model = body_content_table.Meta.model`?

nautobot/ipam/views.py
```diff
@@ -1104,75 +1104,60 @@ def list(self, request, *args, **kwargs):
...
-    table = tables.VLANGroupTable
+    def get_extra_context(self, request, instance):
+        context = super().get_extra_context(request, instance)
+        if self.action == "retrieve":
+            vlans = (
+                VLAN.objects.restrict(request.user, "view")
+                .filter(vlan_group=instance)
+                .prefetch_related(Prefetch("prefixes", queryset=Prefix.objects.restrict(request.user)))
+            )
+            vlans_count = vlans.count()
```
glennmatthews, line 1140: Is this still used in the new view setup?

nautobot/ipam/views.py
```diff
@@ -1104,75 +1104,60 @@ def list(self, request, *args, **kwargs):
...
+            excluded_vlans = vlans.exclude(prefixes__isnull=True)
```
glennmatthews, line 1141: Not sure I understand why this is the right calculation to use for `badge_count_override` - can you add a comment explaining it?

nautobot/ipam/views.py
```diff
@@ -1104,75 +1104,60 @@ def list(self, request, *args, **kwargs):
...
+                vlans, exclude=["vlan_group"], data_transform_callback=data_transform_callback
+            )
+            paginate = {
+                "paginator_class": EnhancedPaginator,
+                "per_page": get_paginate_count(request),
+            }
+            RequestConfig(request, paginate).configure(vlan_table)
+            context.update(
+                {
+                    "first_available_vlan": instance.get_next_available_vid(),
```
glennmatthews, line 1153: Is this still used/needed in the new view setup?

nautobot/ipam/views.py
Pavan-183: Hi @glennmatthews, if you don’t mind, could you please clarify what the new view setup means?

itdependsnetworks: Was the count converted to a table, and thus the count is not needed.

Pavan-183: Okay @itdependsnetworks

nautobot/ipam/views.py
Pavan-183: image
Hi @glennmatthews,
In the first screenshot, the count is showing as 6.
image
However, in the detailed view (second screenshot), the count shows as 9.
This discrepancy is because the green button is being counted as a row in the detailed view. To account for that, we are using this logic in the calculation.

nautobot/core/ui/object_detail.py
```diff
@@ -303,6 +303,36 @@ def get_extra_context(self, context: Context):
         }
 
 
+class FormButtonNoPK(Button):
+    """
+    A button used to submit a form without including the object's primary key in the URL.
```
glennmatthews, line 308: Feels like we should just add an optional `link_includes_pk` kwarg to `Button.__init__()` rather than creating another new subclass?

nautobot/ipam/views.py
```diff
@@ -1106,73 +1107,81 @@ def list(self, request, *args, **kwargs):
...
+                    object_detail.FormButtonNoPK(
+                        link_name="ipam:vlan_bulk_delete",
+                        label="Delete Selected",
+                        color=ButtonActionColorChoices.DELETE,
+                        icon="mdi-trash-can-outline",
+                        size="xs",
+                        form_id="vlan_form",
+                        weight=100,
+                    ),
+                ],
```
glennmatthews, line 1153: Feels like there should be a pattern for automatically including the relevant bulk action buttons on an ObjectsTablePanel, without needing to explicitly describe them like this. Is this the first ObjectsTablePanel we've encountered that had a need for these buttons?

nautobot/ipam/views.py
glennmatthews: Why doesn't `vlans_count = vlans.count()` (commented out above) work for this purpose?

nautobot/core/ui/object_detail.py
Pavan-183: yeah sure @glennmatthews

nautobot/ipam/views.py
HanlinMiao: I feel like this is the time I have seen it personally, as far as BulkEdit and BulkDelete buttons in a related object table in a detail view.

HanlinMiao (approved):
https://github.com/nautobot/nautobot/pull/7047/commits/4c45fdf58d5308ac2acb18cb76f7b7c1a3752f2e should address all the PR feedback here.

nautobot/core/ui/object_detail.py
```diff
@@ -275,7 +275,14 @@ def get_extra_context(self, context: Context):
 
 
 class FormButton(Button):
-    def __init__(self, form_id: str, link_name: str, template_path="components/button/formbutton.html", **kwargs):
+    def __init__(
+        self,
+        form_id: str,
+        link_name: str,
+        template_path="components/button/formbutton.html",
+        link_includes_pk=True,
```
glennmatthews, line 283: Should this move "up" one level further and be an attribute of `Button` rather than `FormButton`? I'm not seeing why the need for this flag is specific to FormButton.

glennmatthews (approved):

Issue #7050, opened 2025-03-20
Bulk Edit and Bulk Delete functionalities do not work in the Contacts Tab of Object Detail View.


HanlinMiao:
### Environment
* Nautobot version (Docker tag too if applicable): v2.4.0
* Python version:
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Go to any objects detail view with contacts or teams associate with it
2. Go to its Contact tab
3. Select some contacts or teams to bulk edit or bulk delete

### Expected Behavior
Nautobot takes me to Contact Bulk Edit/Delete View.

### Observed Behavior
Nothing happens

### Triage
We are using `_ObjectDetailContactsTab` to render the contacts tab and its table. However, its table is not wrapped around a `<form>` element like it does in obj_table.html https://github.com/nautobot/nautobot/blob/develop/nautobot/core/templates/utilities/obj_table.html#L3. That causes the no-op.

architekture:
This bug appears to still exist in v3.0. Able to replicate this behavior in both the demo deployment and in a live environment.

PR #7056, opened 2025-03-21
ClusterUIViewset

Sridevi-NTC:
# What's Changed

# Screenshots
| Update View     |imageimage|image|
| Delete View     |imageimage|image image|
| Cluster Add Device View |image image image|imageimage|
| Cluster Delete Device View |imageimageimage|imageimageimage|

# TODO

- [ ] Unit, Integration Tests


Sridevi-NTC:
Remove devices and Virtual Machines counts are work in progress

cluster


Can you please suggest on following 

1. Is it okay to create `ClusterAddDeviceObjectsTablePanel`, or should this be handled within the existing [ObjectsTablePanel._get_table_add_url](https://github.com/nautobot/nautobot/blob/develop/nautobot/core/ui/object_detail.py#L738) method by checking the exception message for the presence of a pk?
2. Do we need separate Virtual Machines table or just counts?
3. Can i override footer_content_template_path to add Remove devices or do we have any better way?

glennmatthews:
>     1. Is it okay to create `ClusterAddDeviceObjectsTablePanel`, or should this be handled within the existing [ObjectsTablePanel._get_table_add_url](https://github.com/nautobot/nautobot/blob/develop/nautobot/core/ui/object_detail.py#L738) method by checking the exception message for the presence of a pk?

So for clarity, the issue here is that the "add" URL in this case is not a standard CRUD URL (e.g. `/virtualization/virtual-machines/add/` but is instead a special sub-action under the detail route (`/virtualization/clusters/<pk>/devices/add/`), which IIRC is **not** for creating devices, it's for adding existing devices to this cluster. So using `add_button_route` is actually a bit "wrong" in this case. I think it'd be better to add a custom footer to this specific case (see also (3) below)

> 
>     2. Do we need separate Virtual Machines table or just counts?

Let's add a table and remove the special count.

> 
>     3. Can i override footer_content_template_path to add Remove devices or do we have any better way?

Since this is another specialized action (it's removing devices from the cluster, **not** deleting the devices), I think a custom footer is the appropriate thing to do.

Sridevi-NTC:
Thanks @glennmatthews I will work on your suggestions

Sridevi-NTC:
hostdevices| image

I have selected items in the checklist, but it seems like `request.POST.getlist('pk')`  is returning an empty list in the form. Not sure what might be causing this issue.

Sridevi-NTC:
I've added `{% include 'responsive_table.html' with table=body_content_table %}` in the form, and bulk edit is working, but it's creating a duplicate device table due to the `body_content_template_path`. Would it be advisable to set this to None?

glennmatthews:
Not clear to me what exactly your approach is but that *sounds* like the wrong approach to me.

Sridevi-NTC:
> Not clear to me what exactly your approach is but that _sounds_ like the wrong approach to me.

@glennmatthews  I apologize for any confusion, I have committed the code could you please check now?? I would appreciate your guidance on the right direction

glennmatthews:
That approach looks like it's moving the entire table from the `body` of the panel down into the `footer`, which seems undesirable. Fixing this the right way (allowing a button in the footer to submit the form described in the body) may require adding an `id` to the form/body template and adding a `form="<id>"` to the button. Likely a similar fix is needed to address #7050 in general.

Sridevi-NTC:
We don't have <form> tag associated with the body content template, i have created custom body template and add form tag with action url and linked form reference id to the remove device button. Please review the code and kindly provide your suggestions

nautobot/virtualization/templates/virtualization/cluster_devices_body_content.html
```diff
@@ -0,0 +1,7 @@
+{% load render_table from django_tables2 %}
+<form id="object-table-form" method="post" action="{% url 'virtualization:cluster_remove_devices' pk=object.pk %}">
```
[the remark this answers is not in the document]
Sridevi-NTC: If we have both `bulkedit ` and  `bulkdelete ` actions for an `ObjectsTablePanel` , how can we differentiate the `form_id `and `form_action_url ` for each of these actions?

Ex: vlangroups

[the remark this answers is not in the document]
glennmatthews: Should be a shared `id` for the form, and a different `formaction` attached to each button. Sounds like maybe some design work is needed here though.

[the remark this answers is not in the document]
glennmatthews: Maybe we could let the code actually pass `Button` instances into the `ObjectsTablePanel` initializer in some fashion (`footer_buttons=...`)?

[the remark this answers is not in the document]
Sridevi-NTC: @glennmatthews Are these changes considered blockers for this PR?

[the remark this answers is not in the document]
glennmatthews: I'd like to see us get this "right" for this PR + #7047 + #7050 and whatever other cases there may be in core and in apps. If you don't feel comfortable tackling that yet, we can see if it makes sense to pull the task into the core team's sprint next week - let us know.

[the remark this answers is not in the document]
Sridevi-NTC: In yesterday's QA meeting, we agreed to close this PR and track it as a follow-up. That's the reason I asked the question. I will certainly give it a try and will reach out if I encounter any challenges.

nautobot/core/templates/components/panel/body_content_objects_table.html
```diff
@@ -1,4 +1,7 @@
 {% load render_table from django_tables2 %}
-{% with table=body_content_table %}
-    {% render_table table 'inc/table.html' %}
-{% endwith %}
+<form id="object-table-form" method="post">
```
glennmatthews, line 2: Probably need a way to specify/override the `id` here as some views may have multiple ObjectsTablePanel in them?

nautobot/core/templates/components/panel/footer_content_table.html
```diff
@@ -4,11 +4,34 @@
     </a>
 {% endif %}
 
+<div class="pull-left">
+    {% for button_context in buttons_context %}
+        <button form="object-table-form" type="button" class="btn btn-{{button_context.color}} primary btn-xs" action-url="{{button_context.link}}">
+            {{button_context.label}}
+        </button>
```
glennmatthews, line 11: This feels weird to me:
1) contexts are not specific to individual buttons?
2) why aren't we just using `button.render()`?

nautobot/virtualization/templates/virtualization/cluster_add_devices_table_footer.html
glennmatthews, line 1: Can/should this be removed now?

nautobot/virtualization/views.py
```diff
@@ -89,52 +90,56 @@ class ClusterGroupUIViewSet(NautobotUIViewSet):
...
+                table_filter="cluster",
+                table_title="Host Devices",
+                enable_bulk_actions=True,
+                add_button_route=None,
+                form_id="device_form",
+                footer_buttons=[
+                    object_detail.Button(
+                        link_name="virtualization:cluster_remove_devices",
+                        label="Remove Device",
+                        weight=200,
```
glennmatthews, line 122: Sorry for any confusion - weights don't need to be globaly unique, just within a grouping - so these two buttons can use 100 and 200, and the full-width panel below can use 100 as well.

nautobot/virtualization/templates/virtualization/cluster_add_devices_table_footer.html
Sridevi-NTC: sure @glennmatthews

Sridevi-NTC:
After clicking on cluster edit dropdown in detail page i'v page not found issue , after click on back button then i can see delete cluster/clone cluster 

**Drop down on Edit cluster** 
image

**Page not found issue**
Screenshot 2025-04-04 171543

**Image after clicking on back button** 
image

nautobot/core/templates/components/panel/footer_content_table.html
```diff
@@ -4,11 +4,30 @@
...
     </div>
+
     <div class="clearfix"></div>
 {% endif %}
+
+<script>
+    const buttons = document.querySelectorAll('button');
+    buttons.forEach(button => {
+        button.addEventListener('click', function() {
+            const actionUrl = button.getAttribute('action-url');
```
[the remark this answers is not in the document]
Sridevi-NTC: sure @glennmatthews

nautobot/core/ui/object_detail.py
```diff
@@ -180,6 +180,7 @@ def __init__(
         required_permissions=None,
         javascript_template_path=None,
         attributes=None,
+        render_as_button=False,
```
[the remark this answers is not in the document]
Sridevi-NTC: sure @glennmatthews

nautobot/core/ui/object_detail.py
```diff
@@ -697,6 +731,10 @@ def __init__(
                 to the base model. Defaults to the same as `table_filter` if unset. Used to populate URLs.
             enable_bulk_actions (bool, optional): Show the pk toggle columns on the table if the user has the
                 appropriate permissions.
+            footer_buttons (list, optional): A list of Button or FormButton components to render in the panel footer.
+                These buttons typically perform actions like bulk delete, edit, or custom form submission.
+            form_id (str, optional): A unique identifier used to set the `data-form-id` attribute on each `FormButton`.
```
glennmatthews, line 736:
```suggestion
            form_id (str, optional): A unique ID for this table's form; used to set the `data-form-id` attribute on each `FormButton`.
```

nautobot/virtualization/views.py
```diff
@@ -89,52 +90,60 @@ class ClusterGroupUIViewSet(NautobotUIViewSet):
...
+                enable_bulk_actions=True,
+                add_button_route=None,
+                form_id="device_form",
+                footer_buttons=[
+                    object_detail.FormButton(
+                        link_name="virtualization:cluster_remove_devices",
+                        label="Remove Devices",
+                        weight=100,
+                        color=ButtonActionColorChoices.DELETE,
+                        icon="mdi mdi-trash-can-outline",
```
glennmatthews, line 124: `mdi ` class is implied and doesn't need to be explicitly specified as part of `icon`.

nautobot/core/templates/components/panel/body_content_objects_table.html
```diff
@@ -1,4 +1,7 @@
 {% load render_table from django_tables2 %}
-{% with table=body_content_table %}
-    {% render_table table 'inc/table.html' %}
-{% endwith %}
+<form id="{{form_id}}" method="post">
```
glennmatthews, line 2: Probably need an `if form_id` conditional here so that we don't add an `id="None"` in the default case?

nautobot/core/templates/components/panel/footer_content_table.html
```diff
@@ -1,14 +1,38 @@
...
 {% if body_content_table_list_url and more_queryset_count > 0 %}
     <a href="{{ body_content_table_list_url }}">
         <span class="mdi mdi-dots-horizontal" aria-hidden="true"></span> View {{ more_queryset_count }} more {{ body_content_table_verbose_name_plural }}
     </a>
 {% endif %}
 
+<div class="pull-left">
+    {% render_components footer_buttons %}
+</div>
+<div class="clearfix"></div>
```
glennmatthews, line 12: Wrap in `if footer_buttons` perhaps?

nautobot/core/templates/components/panel/footer_content_table.html
```diff
@@ -1,14 +1,38 @@
...
+                form.action = actionUrl;
+                form.submit();
+            }
+            else {
+                console.warn('Missing form_id or action_url:', button);
+                return;
+            }
+        });
+    });
+</script>
```
glennmatthews, line 38: Wrap in `if footer_buttons` perhaps?

nautobot/core/ui/object_detail.py
Sridevi-NTC: Updated as requested, should be resolve, let me know if I missed anything

nautobot/core/ui/object_detail.py
```diff
@@ -270,6 +273,35 @@ def get_extra_context(self, context: Context):
...
+    def __init__(self, render_as_button: bool, template_path="components/button/formbutton.html", **kwargs):
+        """
+        Initialize a FormButton instance.
+
+        Args:
+            render_as_button (bool): Determines whether to render as a button
+        """
+
+        if not render_as_button:
+            raise ValueError("FormButton requires 'render_as_button' to be enabled.")
```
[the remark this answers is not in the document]
Sridevi-NTC: Updated as requested, should be resolve, let me know if I missed anything

glennmatthews (approved):

PR #5746, opened 2024-05-21
DRAFT: Upgrade Graphene to 3.2.x

jamesharr:
# Closes #5745 
# What's Changed

# Screenshots

# TODO

- [ ] Explanation of Change(s)
- [ ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [ ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example App Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #7070, opened 2025-03-25
Graphene upgrade to v3

### LINKED ISSUE #4769: Update graphiql to latest
### Proposed Changes

We're packaging graphiql 1.5.16 at present but graphiql is up to at least version 3.0.9 now.

### Justification

Bug fixes, native dark-mode support (added in 2.0.0), etc.

### LINKED ISSUE #5745: Upgrade to graphene to latest
### Proposed Changes
The graphene library is up to at least 3.3.x, but Nautobot is currently pinned to ~2.16.0 because 3.x includes some breaking changes.

### Justification
- Upgrading will help support #4769 
- Graphene v3 may bring some performance improvements, specifically around N+1 issues

### References
1. #4769 - upgrade GraphIQL
2. [graphene-python issue 1365](https://github.com/graphql-python/graphene/issues/1365) -  issue regarding N+1 queries
3. [Chat in #nautobot on NTC Slack about Graphene and N+1 query performance](https://networktocode.slack.com/archives/C01NWPK6WHL/p1716306578055549)
4. [Graphene v3 release notes](https://github.com/graphql-python/graphene/wiki/v3-release-notes)

glennmatthews:
# Closes #5745 
# Closes #4769 
# What's Changed

Building heavily on the terrific work by @jamesharr in #5746, this updates Nautobot 3.0 to use the latest and greatest versions of `graphene-django` and `graphene-django-optimizer`.

The impacts to the end user are surprisingly minimal, consisting of the following:

- `execute_query()` and `execute_saved_query()` now return a different class of response object, which will impact Jobs and Apps using those APIs.
- We no longer have to rename the filterset field `type` to `_type` to avoid a conflict with GraphQL (yay!) but we now DO have to rename the filterset field `description` to `_description` to avoid a conflict with GraphQL (boo!)

This brings along with it an upgrade to the GraphiQL UI from v1.x to v2.4.7, which is not the latest-and-greatest by any means but it is the version supported presently by `graphene-django` and makes for a nice UI refresh in any case. I've used its native theming functionality via CSS variables to make it integrate a bit more cleanly with Nautobot's own theme but this may need additional fine-tuning in the future.

# Screenshots

image

# TODO

- [x] Unit, Integration Tests


nautobot/core/api/views.py
```diff
@@ -683,18 +689,15 @@ def init_graphql(self):
         if not self.schema:
             self.schema = graphene_settings.SCHEMA
 
-        if self.backend is None:
-            self.backend = get_default_backend()
-
-        self.graphql_schema = self.graphql_schema or self.schema
+        self.graphql_schema = self.graphql_schema or self.schema.graphql_schema
```
glennmatthews, line 692: I don't think we actually are using `self.graphql_schema` anywhere except locally within this function, I'll see if we can remove it.

nautobot/core/api/views.py
```diff
@@ -773,38 +776,44 @@ def execute_graphql_request(self, request, data, query, variables, operation_nam
...
 
-        try:
-            extra_options = {}
-            if self.executor:
-                # We only include it optionally since
-                # executor is not a valid argument in all backends
-                extra_options["executor"] = self.executor
+        validation_errors = validate(schema, document, self.validation_rules, graphene_settings.MAX_VALIDATION_ERRORS)
+        if validation_errors:
+            return ExecutionResult(data=None, errors=validation_errors)
```
glennmatthews, line 801: Possibly could omit this addition (copied in from the updated graphene-django base view) as we aren't currently setting any `validation_rules`.

nautobot/core/graphql/generators.py
```diff
@@ -305,6 +306,11 @@ def generate_list_resolver(schema_type, resolver_name):
 
     def list_resolver(self, info, limit=None, offset=None, **kwargs):
         filterset_class = schema_type._meta.filterset_class
+
+        # Inverse of substitution logic from get_filtering_args_from_filterset() - transform "_description" back to "description"
+        if "_description" in kwargs:
+            kwargs["description"] = kwargs.pop("_description")
```
glennmatthews, line 312: 2.x probably has a bug with `_type` filters due to the lack of an equivalent to this code.

HanlinMiao (approved):
I tried to nit-pick and could not find anything. Awesome stuff

nautobot/core/graphql/generators.py
```diff
@@ -354,7 +360,7 @@ def generate_attrs_for_schema_type(schema_type):
     # Define Attributes for single item and list with their search parameters
     search_params = generate_list_search_parameters(schema_type)
     attrs[single_item_name] = graphene.Field(schema_type, id=graphene.ID())
-    attrs[list_name] = graphene.List(schema_type, **search_params)
+    attrs[list_name] = DjangoListField(schema_type, **search_params)
```
HanlinMiao, line 363: [DjangoListField](https://github.com/graphql-python/graphene-django/blob/main/graphene_django/fields.py#L21-L22) has a positional argument called `_type` as well. That is why we are running into this error.

PR #7071, opened 2025-03-25
statusuiviewset

Sridevi-NTC:
# What's Changed

# Screenshots
| Update View     |image|image|
| Delete View     |image image| imageimage|
| Bulk Edit View  |image |image image|
| Bulk Delete View | image|image image|

# TODO

- [ ] Unit, Integration Tests


nautobot/extras/models/statuses.py
```diff
@@ -33,6 +34,13 @@ class Meta:
         ordering = ["name"]
         verbose_name_plural = "statuses"
 
+    def get_color_display(self):
+        if self.color:
+            return format_html(
+                '<span class="label color-block" style="background-color: #{}">&nbsp;</span>', self.color
+            )
+        return format_html('<span class="text-muted">&mdash;</span>')
```
glennmatthews, line 42: Would be good to use the templatetag `placeholder()` here just to reduce the number of places we have to repeat this HTML fragment.

Sridevi-NTC: Is this what you are suggestion?
```
def get_color_display(self):
        if self.color:
            return format_html(
                '<span class="label color-block" style="background-color: #{}">&nbsp;</span>', self.color
            )
        return helpers.placeholder(self.color)
```

glennmatthews: Yes

Sridevi-NTC:
BulkEdit is pending because of ContentType is not JSON serializable

HanlinMiao (approved):

nautobot/extras/models/statuses.py
```diff
@@ -33,6 +35,13 @@ class Meta:
         ordering = ["name"]
         verbose_name_plural = "statuses"
 
+    def get_color_display(self):
```
glennmatthews, line 38: Would it make sense to add this to the base `NameColorContentTypesModel` class instead of only to `Status`?

glennmatthews (approved):

PR #6866, opened 2025-02-06
Move DVE into core

lampwins:
This PR is the initial lift to move DVE into core. I have also taken the liberty of exposing `DataComplianceRule` in the apps API, and I have migrated the rules models to the UI component framework, however the data compliance UI remains as-is with templates. There are also some misc things that I cleaned up because they were no longer needed, like url fields in the serializers, etc.

# Screenshots

Screenshot 2025-02-05 at 11 49 46 PM

# TODO

- [x] Explanation of Change(s)
- [ ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/core/#creating-changelog-fragments))
- [x] Attached Screenshots, Payload Example
- [x] Unit, Integration Tests
- [x] Documentation Updates (when adding/changing features)
- [ ] Example App Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #7074, opened 2025-03-25
DVE Code Refactor

### LINKED ISSUE #6874: DVE code cleanup

HanlinMiao:
# Closes #6874
# What's Changed

Address PR comments from https://github.com/nautobot/nautobot/pull/6866.

Addressed the majority of the comments but there are still some leftovers that need some resolution.

# Screenshots

N/A
# TODO

- [x] `Data Compliance vs Data Validation Engine vs Data Validation` https://github.com/nautobot/nautobot/pull/6866#discussion_r1944875406
https://github.com/nautobot/nautobot/pull/6866#discussion_r1944877929
- [x] https://github.com/nautobot/nautobot/pull/6866#discussion_r1944879281
- [x] Data Validation Engine Icon https://github.com/nautobot/nautobot/pull/6866#discussion_r1944882773
- [x] DataComplianceRule workflow Diagram https://github.com/nautobot/nautobot/pull/6866#discussion_r1944887396
- [x] https://github.com/nautobot/nautobot/pull/6866#discussion_r1944888301
- [x] https://github.com/nautobot/nautobot/pull/6866#discussion_r1944891171
- [x] https://github.com/nautobot/nautobot/pull/6866#discussion_r1944896186
- [x] https://github.com/nautobot/nautobot/pull/6866#discussion_r1944898591
- [x] Data Compliance Data Cascade Delete https://github.com/nautobot/nautobot/pull/6866#discussion_r1944901594
- [x] https://github.com/nautobot/nautobot/pull/6866#discussion_r1944923878
- [x] https://github.com/nautobot/nautobot/pull/6866#discussion_r1944931330
- [x] https://github.com/nautobot/nautobot/pull/6866#discussion_r1944934142
- [x] https://github.com/nautobot/nautobot/pull/6866#discussion_r1944936119
- [x] Not really sure what is the repeated logic here? https://github.com/nautobot/nautobot/pull/6866#discussion_r1944962566
- [x] Data Compliance Data Cascade Delete solution https://github.com/nautobot/nautobot/pull/6866#discussion_r1944981701
- [x] Use `generic_filter_tests` https://github.com/nautobot/nautobot/pull/6866#discussion_r1946655722
- [x] https://github.com/nautobot/nautobot/pull/6866#discussion_r1946658248


nautobot/core/jobs/__init__.py
```diff
@@ -361,6 +371,131 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
...
 
+def get_data_compliance_rules():
+    """Generate a list of Audit Ruleset classes that exist from the registry."""
+    validators = []
+    for rule_sets in get_data_compliance_rules_map().values():
+        validators.extend(rule_sets)
+    return validators
+
+
+def get_choices():
```
glennmatthews, line 382: Can we give this a less generic name? `get_data_compliance_choices` or something?

nautobot/core/jobs/__init__.py
```diff
@@ -361,6 +371,131 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
...
+
+    selected_data_compliance_rules = MultiChoiceVar(
+        choices=get_choices,
+        label="Select Data Compliance Rules",
+        required=False,
+        description="Not selecting any rules will run all rules listed.",
+    )
+
+    run_builtin_rules_in_report = BooleanVar(
+        label="Run built-in validation rules?", description="Include created, built-in data validation rules in report."
```
glennmatthews, line 420: The description is confusing to me - are these created rules, or built-in rules? Seems wrong to be both.

nautobot/core/jobs/__init__.py
```diff
@@ -361,6 +371,131 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
...
+    def run(self, *args, **kwargs):
+        """Run the validate function on all given DataComplianceRule classes."""
+        selected_data_compliance_rules = kwargs.get("selected_data_compliance_rules", None)
+
+        compliance_classes = []
+        compliance_classes.extend(get_data_compliance_rules())
+
+        for repo in GitRepository.objects.all():
+            if "nautobot_data_validation_engine.data_compliance_rules" in repo.provided_contents:
+                compliance_classes.extend(get_classes_from_git_repo(repo))
```
glennmatthews, line 432: Feels weird to me that `get_data_compliance_rules()` doesn't automatically return the rules provided by git repositories, but instead we need to add this extra logic in each case.

nautobot/core/jobs/__init__.py
```diff
@@ -361,6 +371,131 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
...
+
+        for compliance_class in compliance_classes:
+            if selected_data_compliance_rules and compliance_class.__name__ not in selected_data_compliance_rules:
+                continue
+            self.logger.info(f"Running {compliance_class.__name__}")
+            app_label, model = compliance_class.model.split(".")
+            for obj in global_apps.get_model(app_label, model).objects.all():
```
glennmatthews, line 439: Possibly use `.iterator()` here to reduce overhead when there's a lot of objects?

nautobot/core/jobs/__init__.py
```diff
@@ -361,6 +371,131 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
...
+        for model_class in model_classes:
+            model_custom_validators = registry["plugin_custom_validators"][model_class._meta.label_lower]
+            # Get only DataValidationCustomValidators
+            # otherwise, we would get all validators (more than those dynamically created)
+            validator_dicts.extend(
+                [
+                    {cv: model_class}
+                    for cv in model_custom_validators
+                    if cv.__name__
+                    == f"{model_class._meta.app_label.capitalize()}{model_class._meta.model_name.capitalize()}CustomValidator"
```
glennmatthews, line 474: This seems really fragile to me, maybe there's a better way we can do this, like a class attribute that specifies it's dynamically created?

nautobot/core/jobs/__init__.py
```diff
@@ -361,6 +371,131 @@ def run(self, *, content_type, csv_data=None, csv_file=None, roll_back_if_error=
...
+                ]
+            )
+
+        # Run validation on existing objects and add to report
+        for validator_dict in validator_dicts:
+            for validator, class_name in validator_dict.items():
+                if getattr(validator, "clean") == getattr(CustomValidator, "clean"):
```
glennmatthews, line 481: not sure I see why getattr is needed here, perhaps just?

```suggestion
                if validator.clean == CustomValidator.clean:
```

nautobot/docs/user-guide/feature-guides/data-compliance.md
```diff
@@ -8,15 +8,16 @@ This is ideal for implementing some kind of business logic or standardization re
...
->
-> For example, if a user fixes an object attribute that was incompliant with a built-in rule and then navigates to its `Data Compliance` tab, the object will still show as invalid for that built-in rule. This will remain the case until the job is ran again with the `Run built-in validation rules?` option checked.
+!!! note
+    Rules created using the `DataComplianceRule` class automatically re-run and re-validate whenever the target object is modified and saved. They can also be run manually using the `RunRegisteredDataComplianceRules` job.
+    However, `DataCompliance` objects generated by the `Run built-in validation rules?` option in the job do not update or re-validate automatically. They will only refresh if the job is explicitly run again.
+    For example, if a user fixes a non-compliant attribute on an object and checks the Data Compliance tab, the object will still appear as invalid for that built-in rule. This invalid status won't update until the job is rerun with the `Run built-in validation rules?` option enabled.
```
glennmatthews, line 20: This is confusing to me even on a re-read. :-) Any chance we can fix this behavior now that this is in core, or at least elaborate on the example with a concrete example of an `object`, `non-compliant attribute`, and `built-in rule` that would demonstrate this behavior?

nautobot/extras/datasources/git.py
```diff
@@ -1050,6 +1050,15 @@ def delete_git_graphql_queries(repository_record, job_result, preserve=None):
                 job_result.log(error_msg, level_choice=LogLevelChoices.LOG_ERROR, grouping="graphql queries")
 
 
+def refresh_git_data_compliance_rules(repository_record, job_result, delete=False):  # pylint: disable=W0613
+    """Callback for repo refresh."""
+    from nautobot.nautobot_data_validation_engine.custom_validators import get_classes_from_git_repo
```
glennmatthews, line 1055: Can we rename this method to `get_data_compliance_classes_from_git_repo` to be less generic?

nautobot/nautobot_data_validation_engine/models.py
```diff
@@ -42,13 +38,19 @@ def get_for_model(self, content_type):
 class ValidationRule(PrimaryModel):
     """Base model for all validation engine rule models."""
 
-    name = models.CharField(max_length=100, unique=True)
+    name = models.CharField(max_length=CHARFIELD_MAX_LENGTH, unique=True)
+    field = models.CharField(
+        max_length=50,
```
glennmatthews, line 43: Any reason not to make this use CHARFIELD_MAX_LENGTH as well?

nautobot/nautobot_data_validation_engine/tests/test_custom_validators.py
```diff
@@ -182,13 +188,16 @@ def test_field_value_type_raise_validation_error(self):
         )
 
         location = Location(
-            name="does not match the regex",
+            name="Location with a invalid latitude",
```
glennmatthews, line 191:
```suggestion
            name="Location with an invalid latitude",
```

nautobot/nautobot_data_validation_engine/tests/test_filters.py
```diff
@@ -21,32 +22,45 @@
...
         """
         Create test data
         """
-        tag_1 = Tag.objects.create(
-            name="Tag 1",
-            color="00bcd4",
-            description="Test tag 1",
-        )
+        tag_1 = Tag.objects.first()
         tag_1.content_types.set([ContentType.objects.get_for_model(RegularExpressionValidationRule)])
```
glennmatthews, line 62: Possibly `.add` rather than `.set` here and below?

nautobot/nautobot_data_validation_engine/views.py
```diff
@@ -4,18 +4,20 @@
...
 )
 from nautobot.core.views.generic import ObjectView
+from nautobot.core.views.mixins import (
+    ObjectBulkDestroyViewMixin,
+    ObjectChangeLogViewMixin,
+    ObjectDestroyViewMixin,
+    ObjectDetailViewMixin,
+    ObjectListViewMixin,
+    ObjectNotesViewMixin,
+)
```
glennmatthews, line 20: What keeps us from using UIViewSet? Lack of bulk-edit support?

nautobot/nautobot_data_validation_engine/views.py
```diff
@@ -139,7 +141,12 @@ class UniqueValidationRuleUIViewSet(NautobotUIViewSet):
 
 
 class DataComplianceListView(  # pylint: disable=W0223
```
glennmatthews, line 143: Should this be `DataComplianceUIViewSet`?

nautobot/core/jobs/__init__.py
HanlinMiao: I feel like user-created data validation rules might be a little more accurate.

nautobot/nautobot_data_validation_engine/models.py
HanlinMiao: It is from this [comment](https://github.com/nautobot/nautobot/pull/6866#discussion_r1944974196) right here. We definitely could change it to CHARFIELD_MAX_LENGTH if you think this is necessary.

nautobot/core/jobs/__init__.py
HanlinMiao: I think this is the corresponding [code](https://github.com/nautobot/nautobot/blob/next/nautobot/extras/plugins/validators.py#L31-L35) that wraps the dynamically generated custom validators `clean` method to the `model.clean` method. I'm not sure why we are doing it this way.

nautobot/docs/user-guide/feature-guides/data-compliance.md
HanlinMiao: Rewrote it with a concrete example

HanlinMiao:
Marking this as for review because UTs are passing. There are still some TODOs

- [ ] https://github.com/nautobot/nautobot/pull/6866#discussion_r1944887396
- [ ] https://github.com/nautobot/nautobot/pull/6866#discussion_r1944896186
- [ ] Question? https://github.com/nautobot/nautobot/pull/7074#discussion_r2015084581

HanlinMiao:
Data Compliance Workflow Diagram
Screenshot 2025-03-27 at 11 42 54 AM

nautobot/nautobot_data_validation_engine/models.py
glennmatthews: I think it would be a good safety to make the change yeah.

nautobot/nautobot_data_validation_engine/views.py
glennmatthews: Ah okay, I guess DataCompliance isn't user-editable, which makes sense as I think about it. No issue here then.

nautobot/core/jobs/__init__.py
```diff
@@ -410,14 +411,14 @@ class RunRegisteredDataComplianceRules(Job):
...
+        choices=get_data_compliance_choices,
         label="Select Data Compliance Rules",
         required=False,
         description="Not selecting any rules will run all rules listed.",
     )
 
-    run_builtin_rules_in_report = BooleanVar(
-        label="Run built-in validation rules?", description="Include created, built-in data validation rules in report."
+    run_existing_rules_in_report = BooleanVar(
+        label="Run existing validation rules?", description="Include existing data validation rules in report."
```
glennmatthews, line 421: "Existing" seems at least as confusing to me - we wouldn't ever run **non**-existing rules...? Might need to read the code to understand what this is actually affecting about the operation of the job.

nautobot/docs/user-guide/feature-guides/data-compliance.md
```diff
@@ -10,14 +10,30 @@ This is ideal for implementing some kind of business logic or standardization re
...
 
+```mermaid
+---
+title: Data Compliance WorkFlow Diagram
+---
+flowchart LR
+    A["Object"] -- full_clean() invoked --> B("Data Compliance Rules")
+    B -- clean() methods invoked --> C["Object"]
+    C -- audit() method invoked --> D["audit() return value"]
```
glennmatthews, line 20: Is `audit` a method on the `object`, or a method on the `DataComplianceRule`?

nautobot/extras/datasources/git.py
glennmatthews: Sorry, I think I was unclear. I was referring to renaming `get_classes_from_git_repo`, not renaming `refresh_git_data_compliance_rules`!

nautobot/nautobot_data_validation_engine/custom_validators.py
```diff
@@ -193,12 +198,19 @@ def get_classes_from_git_repo(repo: GitRepository):
...
-    for importer, discovered_module_name, _ in pkgutil.iter_modules([f"{repo.filesystem_path}/custom_validators"]):
-        if discovered_module_name in sys.modules:
-            del sys.modules[discovered_module_name]
-        module = importer.find_module(discovered_module_name).load_module(discovered_module_name)
-        for _, compliance_class in inspect.getmembers(module, is_data_compliance_rule):
-            class_list.append(compliance_class)
+    custom_validators_path = pathlib.Path(f"{repo.filesystem_path}/custom_validators")
+
+    for module_path in custom_validators_path.glob("*.py"):  # Only load Python files
+        module_name = module_path.stem  # Extract the filename without ".py" extension
```
glennmatthews, line 204: Let's compare this to how Job loading is done from a Git repo. In particular I'm concerned that this approach doesn't *look* like it namespaces the loaded modules under the Git repo slug, and as such may potentially clobber system modules (`nautobot.py -> sys.modules["nautobot"]`)? It also may not support relative imports amongst files in the same directory which is something we know users will want to do.

nautobot/nautobot_data_validation_engine/filters.py
```diff
@@ -15,6 +15,30 @@
 )
 
 
+class CustomContentTypeFilter(filters.MultipleChoiceFilter):
+    """Filter for ContentType that doesn't rely on the model's plural name to be in the registry."""
```
glennmatthews, line 19: Can you elaborate on what this is doing, why it's needed, and why it's replacing `ContentTypeMultipleChoiceFilter` below?

nautobot/nautobot_data_validation_engine/forms.py
```diff
@@ -93,10 +92,10 @@ class RegularExpressionValidationRuleFilterForm(NautobotFilterForm):
     q = forms.CharField(required=False, label="Search")
     # "CSV" field is being used here because it is using the slug-form input for
     # content-types, which improves UX.
-    content_type = CSVMultipleContentTypeField(
-        queryset=ContentType.objects.filter(FeatureQuery("custom_validators").get_query()).order_by(
-            "app_label", "model"
-        ),
+    content_type = MultipleContentTypeField(
+        feature=None,
```
glennmatthews, line 96: Why not `feature="custom_validators"`?

nautobot/core/jobs/__init__.py
HanlinMiao: https://github.com/nautobot/nautobot/pull/7074/commits/c9e43ec8ad8b8b5d842315d5f946f20d9a018db5

nautobot/core/jobs/__init__.py
HanlinMiao: It seems like enabling this option will validate relevant objects against all the user created validation rules, which are `RegularExpressionValidationRule`, `MinMaxValidationRule`, `RequiredValidationRule`, and `UniqueValidationRule`.
I was confused because the data compliance objects were not being generated even tho I had those rules and non-compliant objects in place. So I thought it was something else. However, that is due to what I mentioned here https://networktocode-llc.slack.com/archives/C01RMAGBD29/p1743106959423749

HanlinMiao: TLDR, `user-created` seems to be the most explicit and appropriate here

nautobot/nautobot_data_validation_engine/filters.py
HanlinMiao: So we have logic in [get_filterset_parameter_form_field](https://github.com/nautobot/nautobot/blob/next/nautobot/core/utils/filtering.py#L131-L156) to look up the plural names of the content-type related models in the `registry` when we use `ContentTypeMultipleChoiceFilter`. 

However, these validation rules' plural names ("data_compliance",  "unique_validation_rule", and "min_max_validation_rules") are not in our `registry`, therefore the function raises a key error at line [144](https://github.com/nautobot/nautobot/blob/next/nautobot/core/utils/filtering.py#L144) and at line [154](https://github.com/nautobot/nautobot/blob/next/nautobot/core/utils/filtering.py#L154).

so I added an if statement to redirect them to look up plural_name "custom_validators" instead and we can remove `CustomContentTypeFilter` now.

nautobot/nautobot_data_validation_engine/custom_validators.py
HanlinMiao: The job loading code is quite a beast to understand, in the doc string it literally says "If you find yourself writing new code that uses this method, please pause and reconsider your life choices." 

I think this [commit](https://github.com/nautobot/nautobot/pull/7074/commits/59bebd441521423b958e9d896e1fba74fe930efd) addresses the potential clobbering of system modules concern by adding namespacing the loaded modules with the Git repo slug.

I am a little stuck on adding support for relative imports.

nautobot/extras/datasources/git.py
```diff
@@ -1050,6 +1050,16 @@ def delete_git_graphql_queries(repository_record, job_result, preserve=None):
                 job_result.log(error_msg, level_choice=LogLevelChoices.LOG_ERROR, grouping="graphql queries")
 
 
+def refresh_git_data_compliance_rules(repository_record, job_result, delete=False):  # pylint: disable=W0613
```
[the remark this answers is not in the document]
glennmatthews: Can you test this and confirm? I'd assume that just by the fact they're in `sys.modules` they might stick around, but maybe our lookup pattern is such that that isn't the case.

nautobot/nautobot_data_validation_engine/migrations/0009_alter_char_fields_length.py
```diff
@@ -0,0 +1,52 @@
...
+            model_name="requiredvalidationrule",
+            name="field",
+            field=models.CharField(max_length=255),
+        ),
+        migrations.AlterField(
+            model_name="uniquevalidationrule",
+            name="field",
+            field=models.CharField(max_length=255),
+        ),
+    ]
```
dsoftait, line 52: From what I understood this code is to integrate the DVE application with the nautobot code, so that it is not a separate application. For this reason, don't we need some data migration ?

dsoftait (commented):
I added one comment. From a programmatic point of view it looks good, unfortunately I can't verify it from a substantive point of view yet.

nautobot/nautobot_data_validation_engine/migrations/0009_alter_char_fields_length.py
HanlinMiao: I believe we did a necessary data migration in the previous migration file `0008`. We do not need one for this change, because we are relaxing the constraints on these `CharFields`. 
`max_length=100 -> 255` and `max_length=50 -> 255`

dsoftait (approved):
LGTM. Looks like all review remarks was addressed.

glennmatthews (approved):

PR #7079, opened 2025-03-26
feature: implement sidenav flyouts

norbert-mieczkowski-codilime:
# What's Changed
Implement sidenav flyouts which expand on sidenav item click and contain grouped links to specific pages.

# Screenshots
| Sidenav expanded | Sidenav collapsed |
| - | - |
| image | image_2 |


nautobot/core/templates/inc/nav_menu.html
```diff
@@ -22,39 +22,63 @@
...
             </li>
 
             <li class="sidenav-list-divider"></li>
 
             {% for tab_name, tab_details in registry.nav_menu.tabs.items %}
                 {% if not tab_details.permissions or request.user|has_one_or_more_perms:tab_details.permissions %}
                     <li>
-                        <button class="sidenav-list-item" type="button">
+                        <button
+                            aria-controls="sidenav-flyout-{{ tab_name|slugify }}"
```
glennmatthews, line 49: I don't think we have any enforcement at present that the tab names be unique or slugify to unique strings (though obviously naming them in violation of this assumption would not be a best practice!). Wonder if instead we should use `forloop.counter` to get unique IDs?

nautobot/core/templates/inc/nav_menu.html
```diff
@@ -22,39 +22,63 @@
...
+                                            {% if not item_details.permissions or request.user|has_one_or_more_perms:item_details.permissions %}
+                                                <li class="sidenav-link-wrapper{% if item_details.permissions and not request.user|has_perms:item_details.permissions %} disabled{% endif %}">
+                                                    <a class="sidenav-link{% if item_link in request.path %} sidenav-link-active{% endif %}" data-item-weight="{{ item_details.weight }}" href="{{ item_link }}">
+                                                        {{ item_details.name }}
+                                                    </a>
+                                                    <button class="sidenav-favorite" type="button">
+                                                            <span class="visually-hidden">
+                                                                Add to favorites
+                                                            </span>
+                                                    </button>
```
glennmatthews, line 74: wrong indentation?

nautobot/project-static/js/nav_menu.js
```diff
@@ -10,4 +10,30 @@ document.addEventListener('DOMContentLoaded', function() {
         const sidenav = document.getElementById(controls);
         sidenav.classList.toggle('sidenav-collapsed', expanded)
     });
+
+    [...document.querySelectorAll('.sidenav-list-item')].forEach(sidenavListItem => {
```
glennmatthews, line 14: How difficult would it be to add this to `project-static/src/js` instead of here, since ultimately this file will need to be migrated anyway?

nautobot/project-static/src/scss/nautobot.scss
```diff
@@ -568,11 +568,24 @@ body {
...
...9804 13.5897 18.7025 13.8477 18.5419 14.1604C18.4 14.4368 18.3244 14.7424 18.3211 15.0531C18.3174 15.4046 18.4446 15.7671 18.699 16.4921V16.4921C19.5294 18.8587 19.9446 20.042 19.673 20.7338C19.4374 21.3339 18.927 21.7836 18.3021 21.9418C17.5816 22.1241 16.4767 21.5717 14.267 20.4668L13.4311 20.0489C12.9064 19.7865 12.644 19.6553 12.3688 19.6037C12.1251 19.558 11.8749 19.558 11.6312 19.6037C11.356 19.6553 11.0937 19.7865 10.5689 20.0489L9.73299 20.4668C7.52325 21.5717 6.41838 22.1241 5.69791 21.9418C5.07298 21.7836 4.56261 21.3339 4.32703 20.7338C4.05543 20.042 4.47062 18.8587 5.30101 16.4921V16.4921C5.55541 15.7671 5.68261 15.4046 5.67891 15.0531C5.67565 14.7424 5.60005 14.4368 5.45811 14.1604C5.29751 13.8477 5.01964 13.5897 4.46388 13.0736L4.11954 12.7539C2.49141 11.242 1.67734 10.4861 1.57953 9.78001C1.49453 9.16639 1.69867 8.54802 2.13229 8.10561C2.63127 7.59653 3.73538 7.47385 5.9436 7.22849L6.26976 7.19225C6.90189 7.12201 7.21795 7.08689 7.49423 6.96694C7.73879 6.86077 7.95965 6.70676 8.14383 6.51399C8.35189 6.29621 8.49411 6.01178 8.77855 5.44291L9.13784 4.72433Z' fill='none' stroke='#{$nav-white-0}' stroke-width='1.5' stroke-linecap='round' stroke-linejoin='round'/></svg>";
```
glennmatthews, line 571: Is this  (SVG-in-CSS) going to be our pattern for *all* of the custom icons that we eventually plan to add, or is there a better/cleaner pattern that we should begin to establish instead?

nautobot/core/templates/inc/nav_menu.html
norbert-mieczkowski-codilime: good point, updated

nautobot/project-static/js/nav_menu.js
norbert-mieczkowski-codilime: just a matter of copy paste, I somehow didn't realize that's proper place for JS code now, thanks for pointing it out

nautobot/project-static/src/scss/nautobot.scss
norbert-mieczkowski-codilime: yeah, this way of embedding icons has probably more cons than it has pros. I personally don't like it but that's what bootstrap does and for this limited sidenav scope I'm fine doing it that way, at least for now, but definitely not long term.

there is one substantial question that needs to be asked again - are we planning to use off-the-shelf icon library (e.g. tabler icons) or go with our own? if the former, just grab whatever it provides, for the latter however I would like to experiment with fontforge and generating font out of svg files as you suggested at the very beginning

glennmatthews (approved):

PR #7090, opened 2025-03-28
6923 Fix auto populating Module Bay position

### LINKED ISSUE #6923: Position field for ModuleBay doesn't autopopulate when adding to DeviceType
### Environment
* Nautobot version (Docker tag too if applicable): 2.3.13, 2.4.3, 3.0.0a1
* Python version: 
* Database platform, version:
* Middleware(s):

### Steps to Reproduce
1. Select a `DeviceType` record to update
2. Add Components -> Module Bays
3. Type something (e.g. [1-3]) in the name field

### Expected Behavior
Position field should autopopulate like it does when `ModuleBay` records are added to `Device` records

### Observed Behavior
Position field remains blank. This can lead to `Device` records with unpopulated `Position` fields on their bays, which in turn leads to interface name expansion failing when using interfaces named like `foo-0/{module}/1`. Tested locally on 2.3.13 and on 2.4.3(demo) and 3.0.0a1(next.demo).

See related issue: #6336

mateusz-lachowski-codilime:
# Closes #6923 
# What's Changed
Fixed auto population Module Bay position field on:
- Device Type details page
- bulk create from Device list
- bulk create from Modules list

Added integration tests.

# Screenshots

# TODO

- Attached Screenshots, Payload Example
- [x] Unit, Integration Tests


nautobot/dcim/views.py
```diff
@@ -1769,6 +1769,7 @@ class DeviceView(generic.ObjectView):
                 weight=100,
                 color=ButtonColorChoices.BLUE,
                 label="Add Components",
+                attributes={"id": "device-add-components-button"},
```
HanlinMiao, line 1772: ~Is this change used anywhere?~

HanlinMiao: NVM I saw it in the test.

nautobot/dcim/tests/integration/test_module_bay_position.py
```diff
@@ -0,0 +1,125 @@
...
+        details_url = self.live_server_url + reverse("dcim:device", kwargs={"pk": device.pk})
+        self.browser.visit(details_url)
+
+        # Navigate to module bay create page
+        self.browser.find_by_css("#device-add-components-button").click()
+        self.browser.find_by_xpath(
+            "//*[@id='device-add-components-button']/following-sibling::*//a[normalize-space()='Module Bays']"
+        ).click()
+
+        self._validate_position_field()
```
HanlinMiao, line 71: So you fixed this issue for Device Detail page as well correct? Can you make sure you update the changelog fragment to reflect that as well?

changes/6923.fixed
```diff
@@ -0,0 +1 @@
+Fixed auto populating position field when creating Module Bay from Device Type details and bulk creating from Devices or Modules list.
```
HanlinMiao, line 1: Probably should include the Device Detail here, since you did fix it for that page as well.

nautobot/dcim/templates/dcim/modulebay_bulk_create.html
```diff
@@ -0,0 +1,39 @@
+{% extends 'generic/object_bulk_add_component.html' %}
+
+{% block javascript %}
+{{ block.super }}
+<script>
+document.addEventListener('DOMContentLoaded', function() {
+    var position_field = document.getElementById('id_position_pattern');
```
glennmatthews, line 7: So this is basically a de-JQuery-ed version of `initializeSlugField` in `forms.js`, minus the slugify logic. Should we invest any time into making this share code with the slug-field functionality, rather than duplicating it?

nautobot/dcim/tests/integration/test_module_bay_position.py
```diff
@@ -0,0 +1,125 @@
...
+
+        # Change pattern manually and name to verify if it's not updating then
+        position_field.fill("")
+        for _ in position_field.type("new pattern", slowly=True):
+            pass
+
+        for _ in name_pattern_field.type("v2", slowly=True):
+            pass
+
+        self.assertEqual(position_field.value, "new pattern", "Position field value has changed")
```
glennmatthews, line 33:
```suggestion
        self.assertEqual(position_field.value, "new pattern", "Position field value has unexpectedly changed")
```

nautobot/dcim/views.py
```diff
@@ -1651,7 +1651,7 @@ class ModuleBayTemplateUIViewSet(
     model_form_class = forms.ModuleBayTemplateForm
     serializer_class = serializers.ModuleBayTemplateSerializer
     table_class = tables.ModuleBayTemplateTable
-    create_template_name = "dcim/device_component_add.html"
+    create_template_name = "dcim/modulebay_create.html"
```
glennmatthews, line 1654: Why this change? Seems wrong to me that we're using `modulebay_create.html` for creation of a `modulebaytemplate` object.

nautobot/dcim/tests/integration/test_module_bay_position.py
mateusz-lachowski-codilime: I've just created a test for that as well to have each form using that logic covered.

changes/6923.fixed
mateusz-lachowski-codilime: It's working on Device Detail currently. Just covered this on tests.

nautobot/dcim/views.py
mateusz-lachowski-codilime: `modulebay_create.html` anyway just extending the `device_component_add` but with addition of this JS fragment to support position auto-populate logic. I've found that on this form this logic were also missing and it seems that it should be there as well. 

I can try to rename this template to something different.

nautobot/dcim/templates/dcim/modulebay_bulk_create.html
mateusz-lachowski-codilime: It seems that you're right. I've just copied this fragment from original working solution for Device module bay create form, haven't done anything more with it. But since I know that this is a copy of `initializeSlugField` I will definitely try to do some refactor here.

nautobot/dcim/views.py
glennmatthews: Might make sense to create a `modulebaytemplate_create.html` even if it's nothing but `{% extends modulebay_create.html %}`

nautobot/project-static/js/forms.js
```diff
@@ -86,43 +86,90 @@ function initializeCheckboxes(context){
...
+function initializeSlugField(){
+    // Function to support slug fields auto-populate and slugify logic
+    const slugField = getSlugField()
+    if(!slugField){
+        return
     }
+    initializeAutoField(slugField, 'slug-source', 100, slugify);
+}
+
+function initializeAutoPopulateField(){
```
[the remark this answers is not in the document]
mateusz-lachowski-codilime: We're anyway passing this context as `document` so it will be the same behavior, just with vanilla JS document instead of jQuery one.

The most important is to run this when document is ready and `jsify_form` function provides us such functionality.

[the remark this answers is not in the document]
mateusz-lachowski-codilime: I could possibly update `initializeInputs` function to pass just `document` not the `jQuery object` to those 2 functions but then I will break up the "common" interface of this functions.

PR #7091, opened 2025-03-28
Add DNS Models app to the marketplace

glennmatthews:
# What's Changed

Add marketplace entry for https://docs.nautobot.com/projects/dns-models/en/latest/.

# TODO

- n/a Unit, Integration Tests


nautobot/extras/plugins/marketplace_manifest.yml
```diff
@@ -94,6 +94,24 @@
...
+- "name": "DNS Models"
```
HanlinMiao, line 97: I didn't see the whole file but are these alpha sorted?

HanlinMiao (approved):

nautobot/extras/plugins/marketplace_manifest.yml
glennmatthews: Yes! The tooling handles that for us, and this is between `Circuit Maintenance Tracker` and `Data Validation Engine` (I guess capitals sort before lowercase letters).

PR #7092, opened 2025-03-28
Fix route lookup error in DynamicModelChoiceMixin

glennmatthews:
# What's Changed

Fix for an error encountered in a customer's environment after an upgrade. An installed App wasn't correctly registering its REST API routes, which made it so that accessing any form that had a `DynamicModel[Multiple]ChoiceField` referencing that App's model would throw a `NoReverseMatch` exception and break the view.

This PR just adds graceful handling for that exception case. Note that the form field will not function properly (due to having no `data-url` to retrieve data from to populate the Select2 widget), which is why I made the log message a `logger.error()`, but at least the rest of the form can render now.

Example log messages (accessing the Devices list view after temporarily disabling the Rack REST API):

```
nautobot-1  | 19:13:10.440 ERROR   nautobot.core.forms.fields fields.py                      get_bound_field() :
nautobot-1  |   API route lookup "dcim-api:rack-list" failed for model Rack, form field "rack" will not work properly
nautobot-1  | 19:13:10.443 ERROR   nautobot.core.forms.fields fields.py                      get_bound_field() :
nautobot-1  |   API route lookup "dcim-api:rack-list" failed for model Rack, form field "rack__n" will not work properly
nautobot-1  | 19:13:10.469 DEBUG   nautobot.core.utils.querysets querysets.py              maybe_select_related() :
nautobot-1  |   Applying .select_related(['status', 'tenant', 'role', 'device_type', 'location', 'rack', 'device_type__manufacturer']) to Device QuerySet
nautobot-1  | 19:13:10.583 ERROR   nautobot.core.forms.fields fields.py                      get_bound_field() :
nautobot-1  |   API route lookup "dcim-api:rack-list" failed for model Rack, form field "rack" will not work properly
```

# TODO

- n/a Unit, Integration Tests


nautobot/core/forms/fields.py
```diff
@@ -591,8 +595,16 @@ def get_bound_field(self, form, field_name):
...
-            widget.attrs["data-url"] = data_url
+            try:
+                data_url = reverse(route)
+                widget.attrs["data-url"] = data_url
+            except NoReverseMatch:
+                logger.error(
+                    'API route lookup "%s" failed for model %s, form field "%s" will not work properly',
+                    route,
+                    self.queryset.model.__name__,
+                    bound_field.name,
```
HanlinMiao, line 606: Should we use `field_name` on line 567 which was passed into the method here or `bound_field.name`? In the screenshot of the PR description, `rack__n` could be a little confusing to the end user.

Although, the `route` and `self.queryset.model.__name__` contain the most important information here so nvm.

changes/7092.fixed
```diff
@@ -0,0 +1 @@
+Added graceful handling for route-lookup errors when constructing a DynamicModelChoiceField.
```
HanlinMiao, line 1: Perhaps?
```suggestion
Added graceful handling for route-lookup errors when constructing a DynamicModelChoiceField or a DynamicModelMultipleChoiceField.
```

HanlinMiao (approved):

nautobot/core/forms/fields.py
glennmatthews: It's a logger message, not a UI-facing message, so I'm OK with it being a bit technical and precise.

PR #7098, opened 2025-04-01
Develop to next post 2.4.6

HanlinMiao:
Merge conflicts resolved

- [x] poetry.lock file supported `Python versions`
- [x] nautobot.core.test_graphql.py `imports`


nautobot/ipam/tests/test_api.py
```diff
@@ -197,7 +197,13 @@ def setUpTestData(cls):
```
glennmatthews, line 197: To fix the test, we should probably create *new* VRFs to assign to, instead of relying on the assumption that the existing VRFs don't already have an assignment to specific existing Devices/VDCs/VMs.

Issue #7073, opened 2025-03-25
Browser tab showing raw html when editing Saved View


joewesch:
### Environment
* Nautobot version (Docker tag too if applicable): 2.4.4
* Python version: 3.11.11
* Database platform, version: Postgres
* Middleware(s):

### Steps to Reproduce
1. Create a Saved View
2. Add a filter

### Observed Behavior
The browser tab title reads as follows:
```
Devices - <i title="Pending changes not saved>test</i> - ...
```

#### Chrome
![Image](https://github.com/user-attachments/assets/f533aae3-0edb-4b04-92c1-d8f9bbcc80ca)

#### Firefox
![Image](https://github.com/user-attachments/assets/93a3100f-f809-4a2b-8760-f0fde8dbfed5)

### Expected Behavior
I'm not exactly sure what the tab title is _supposed_ to look like, I just noticed the tab name change.

HanlinMiao:
This is the problematic code in the template.
https://github.com/nautobot/nautobot/blob/4e9d0e233dbd982c2da6703449d59011ed8990f8/nautobot/core/templates/generic/object_list.html#L149-L160

PR #7100, opened 2025-04-01
feature: implement v3 ui global header

### LINKED ISSUE #6946: Header and footer layout
Implement the base structure of the header and footer, which will generally be true on any page.

Footer:
- Links become icons only (use current material icons for now)
- hight and color adjustments

Header:
- H1 and breadcrumb placement
- search centered
- User on the top right (note the bell icon in the figma will be added at a later date)
  - Clicking on any part of the user, will show the dropdown that exists today
  - Show initials of users
- Support variances in some pages
  - Home page has now H1 or breadcrumbs
  - List views show the model name in H1 with no breadcrumbs

norbert-mieczkowski-codilime:
# Closes #6946
# What's Changed
Implement v3 UI global header including page title, breadcrumbs, user dropdown menu and optional tabs.

**Search input is a placeholder for now, it will be implemented in scope of another story.**

# Screenshots
image
image


nautobot/core/templates/generic/object_list.html
```diff
@@ -5,35 +5,9 @@
...
-                    </span>
-                </div>
-            </form>
-        </div>
-        {% endif %}
-    </div>
-{% endblock header %}
-
+{% block breadcrumbs %}
+    {% include 'inc/breadcrumbs.html' with title=title %}
```
norbert-mieczkowski-codilime, line 9: using breadcrumbs just for page title like this is nice and easy but real pain starts when it comes to actual `crumbs`. django doesn't support multiline template tags which boils down to super lengthy HTML lines containing all the breadcrumbs as argument. I'm genuintely not happy with this but I wasn't able to figure out anything else. perhaps UI framework component could be more usable here but I didn't dive into that so far. thoughts?

nautobot/core/templates/base_django.html
```diff
@@ -11,32 +11,42 @@
 </head>
 <body>
     {% include 'inc/nav_menu.html' %}
-    <div class="container-fluid wrapper" id="main-content">
+    <header id="header">
```
norbert-mieczkowski-codilime, line 14: I'm fully aware this global header, just like sidenav and footer, will need to present in more templates than just `base_django` but just for now, until it has its final structure (see my other comment), I didn't bother with updating the others

nautobot/core/templates/base_django.html
```diff
@@ -11,32 +11,42 @@
...
+                    <div class="alert alert-info text-center" role="alert">
+                        {{ "BANNER_TOP"|settings_or_config|render_markdown }}
+                    </div>
+                {% endif %}
+            {% endif %}
+            {% if settings.MAINTENANCE_MODE %}
+                <div class="alert alert-warning text-center" role="alert">
+                    <h4><i class="mdi mdi-alert"></i> Maintenance Mode</h4>
+                    <p>Nautobot is currently in maintenance mode. Functionality may be limited.</p>
                 </div>
```
glennmatthews, line 27: Does this render OK in the new header?

nautobot/core/templates/base_django.html
```diff
@@ -11,32 +11,42 @@
...
+                    {{ message }}
+                </div>
+            {% endfor %}
+        </div>
+        <div class="container-fluid mb-12 mt-16">
+            <div class="row row-cols-3">
+                <div class="col">
+                    {% block breadcrumbs %}{% endblock breadcrumbs %}
+                </div>
+                {% include 'inc/header.html' %}
```
glennmatthews, line 44: Should all of the `<header>...</header>` be in `inc/header.html` instead of this file?

nautobot/core/templates/generic/object_list.html
```diff
@@ -147,7 +121,6 @@
     {% endif %}
 </div>
 <h1>{% block title %}
```
glennmatthews, line 123: Should this block declaration be moved into `inc/breadcrumbs.html`?

nautobot/core/templates/generic/object_list.html
```diff
@@ -147,7 +121,6 @@
...
-    {{ title }}
         {% if current_saved_view %}
             —
```
glennmatthews, line 125: How are we planning to indicate the application of a saved view in the new UI?

nautobot/core/templates/generic/object_list.html
glennmatthews: I think we'll need to add UI component support for defining breadcrumbs sooner than later, agreed.

nautobot/core/templates/base_django.html
```diff
@@ -11,32 +11,42 @@
...
         <div class="banner-alert-area">
-        {% if request.user.is_authenticated %}
-            {% if "BANNER_TOP"|settings_or_config %}
-                <div class="alert alert-info text-center" role="alert">
-                    {{ "BANNER_TOP"|settings_or_config|render_markdown }}
+            {% if request.user.is_authenticated %}
+                {% if "BANNER_TOP"|settings_or_config %}
+                    <div class="alert alert-info text-center" role="alert">
+                        {{ "BANNER_TOP"|settings_or_config|render_markdown }}
+                    </div>
```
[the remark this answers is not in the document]
norbert-mieczkowski-codilime: yes, here's an example
image

nautobot/core/templates/base_django.html
norbert-mieczkowski-codilime: As much as I wanted to do exactly that, breadcrumbs due to their configurable nature couldn't be a `block` in such case because to my understanding Django doesn't support blocks inside `{% include %}` template tag. Perhaps that's a problem UI framework could solve

nautobot/core/templates/generic/object_list.html
norbert-mieczkowski-codilime: apparently in django I can't use include with a block tag, that's what I don't like about much of the proposal in this PR

nautobot/core/templates/generic/object_list.html
norbert-mieczkowski-codilime: good point, need to talk about it because I don't think global header can host this information the way it is designed today

nautobot/core/templates/inc/header.html
```diff
@@ -0,0 +1,42 @@
...
+                {% firstof request.user.get_full_name request.user.username %}
+            </button>
+            <ul class="dropdown-menu dropdown-menu-end">
+                <li><a class="dropdown-item" href="{% url 'about' %}"><i class="mdi mdi-information"></i>About</a></li>
+                <li><a class="dropdown-item" href="{% url 'user:profile' %}"><i class="mdi mdi-account-box"></i>Profile</a></li>
+                {% if request.user.is_staff %}
+                    <li><a class="dropdown-item" href="{% url 'worker_status' %}"><i class="mdi mdi-server-network"></i>Worker Status</a></li>
+                    <li><a class="dropdown-item" href="{% url 'admin:index' %}"><i class="mdi mdi-cogs"></i>Admin</a></li>
+                {% endif %}
+                <li class="dropdown-divider"></li>
```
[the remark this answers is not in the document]
norbert-mieczkowski-codilime: I suppose you *could* use `hr`, but Bootstrap already exports a divider component which has a specific color and margin so it is generally recommended to use it

nautobot/core/templates/generic/object_list.html
```diff
@@ -147,7 +121,6 @@
     {% endif %}
 </div>
 <h1>{% block title %}
-    {{ title }}
```
HanlinMiao, line 150: Just a note. Removing the title from here also causes the title tag to be incomplete in the browser tab. It is rendered as "- Nautobot" right now
Screenshot 2025-04-02 at 10 14 00 AM

nautobot/core/templatetags/helpers.py
```diff
@@ -1153,3 +1172,36 @@ def _build_hyperlink(value, field="", target="", rel=""):
...
+
+    if not current_saved_view:
+        return title
+
+    if new_changes_not_applied:
+        new_title = f"{title} — <i title='Pending changes not saved'>{current_saved_view.name}</i>"
+    else:
+        new_title = f"{title} — {current_saved_view.name}"
+
+    new_title = mark_safe(new_title)
```
glennmatthews, line 1194: As Ruff correctly flagged, we should use `format_html` rather than `mark_safe`, probably something like:

```suggestion
    if new_changes_not_applied:
        new_title = format_html("{} — <i title='Pending changes not saved'>{}</i>", title, current_saved_view.name)
    else:
        new_title = format_html("{} — {}", title, current_saved_view.name)
```

nautobot/core/templates/inc/breadcrumbs.html
```diff
@@ -0,0 +1,33 @@
...
+        />
+        {% if copy_title %}
+        <span class="hover_copy">
+            <span id="copy_title">{{ title }}</span>
+            <button type="button" class="btn btn-inline btn-default hover_copy_button" data-clipboard-text="{{ title }}">
+                <span class="mdi mdi-content-copy"></span>
+            </button>
+        </span>
+        {% else %}
+            {{ title }}
```
glennmatthews, line 20: Feels unintuitive to me that the the displayed title is part of `breadcrumbs.html`. Why isn't it part of `header.html` instead?

nautobot/core/templates/generic/object_list.html
```diff
@@ -5,35 +5,15 @@
...
-    </div>
-{% endblock header %}
+{% block breadcrumbs %}
+    {% format_title_with_saved_view title as formatted_title %}
+    {% include 'inc/breadcrumbs.html' with title=formatted_title %}
+{% endblock breadcrumbs %}
 
+{% block title %}
+    {% format_title_with_saved_view title as formatted_title %}
+    {{ formatted_title }}
```
glennmatthews, line 15: **Should** the page title (as displayed in the browser window/tab header) include the saved view annotation? Or is that just noise?

nautobot/core/templates/generic/object_retrieve.html
```diff
@@ -10,38 +10,18 @@
 {% load tz %}
 {% load ui_framework %}
 
+{% block breadcrumbs %}
+    {% format_title_with_saved_view object.display|default:object as formatted_title %}
```
glennmatthews, line 14: Detail views don't support saved views yet, right?

nautobot/core/templatetags/helpers.py
```diff
@@ -1123,6 +1123,25 @@ def hyperlinked_object_target_new_tab(value, field="display"):
     return _build_hyperlink(value, field, target="_blank", rel="noreferrer")
 
 
+@register.filter()
+def get_object_link(value):
```
glennmatthews, line 1127: Not quite sure I understand why we need this when we already have the `hyperlinked_object` templatetag?

mateusz-lachowski-codilime: Previous breadcrumbs was building in place with different logic for different point in path. However this might be a good idea to refactor usage of breadcrumbs and pass list of name/url's and let "breadcrumb" component render it properly. That's why I needed just a link and not a full <a href html fragment.

nautobot/core/templates/generic/object_list.html
mateusz-lachowski-codilime: Browsers seems to be skipping HTML within a `<title>` tag. And this is mostly the old logic just moved as helper to be easier to reuse.

nautobot/core/templates/inc/breadcrumbs.html
mateusz-lachowski-codilime: For me as well, but since I was just contributing to this PR I wanted to wait for Norbert to agree some refactor here. Currently `breadcrumb` is a new `header` - probably at the end of work on this task we might be able to refactor this.

nautobot/core/templates/base_django.html
```diff
@@ -11,31 +11,43 @@
...
+                    {% block breadcrumbs %}{% endblock breadcrumbs %}
+                </div>
+                {% include 'inc/header.html' %}
             </div>
-        {% endfor %}
         </div>
         {% block header %}{% endblock header %}
+    </header>
+    <div class="container-fluid wrapper" id="main-content">
+        {% block page_top %}{% endblock %}
```
glennmatthews, line 50: Unused?

nautobot/core/templates/generic/object_list.html
glennmatthews: #7073 suggests otherwise?

glennmatthews (approved):
Let's merge and iterate.

nautobot/core/templates/generic/object_list.html
norbert-mieczkowski-codilime: @mateusz-lachowski-codilime :point_up_2:

nautobot/core/templates/base_django.html
norbert-mieczkowski-codilime: impressive catch :eyes: removed in 120d3d41683cd4231c8e081049e6981fbdda24d2

nautobot/core/templates/generic/object_list.html
mateusz-lachowski-codilime: Hmm, that's weird but maybe I just didn't test it properly. So this might be a second place that we need to move this logic from templates into view. Maybe having something like: `title` and `rich_title` will work? Or it will be enough to just strip html tags here?

nautobot/core/templates/base_django.html
mateusz-lachowski-codilime: I left that on purpose to maybe put there the top content with created date and additional buttons. Because I can imagine some custom implementations that would like to put something there. So having special place at top, but below tabs had sense in my mind.

PR #7110, opened 2025-04-03
Avoid unnecessarily enqueueing CustomField data background tasks

### LINKED ISSUE #7108: `Infoblox => Nautobot` sync creates hundreds of thousands of subtask results
### Environment
* Python version:  3.11
* Nautobot version:  2.4.2
* nautobot-ssot version: 3.3.0

### Expected Behavior

When running a job, the amount of subtask job result entries shouldn't be in the hundreds of thousands.

### Observed Behavior

When running Infoblox to Nautobot sync, the job calls the `nautobot.extras.tasks.provision_field` subtask and create hundreds of thousands of subtask `job result` entries.

Image

Image

### Steps to Reproduce
1. Connect SSoT app to Infoblox
2. Attempt a sync from Infoblox to Nautobot
3. Excess entries created in job results

glennmatthews:
# Closes #7108 
# What's Changed

- Fixed various cases where changes to CustomField definitions resulted in unnecessary enqueueing and execution of background Celery tasks `delete_custom_field_data` and/or `provision_field`:
   - Bulk delete via UI of CustomFields that have no currently assigned `content_types`
   - Deletion via ORM, REST, or UI of CustomFields that have no currently assigned `content_types`
   - `m2m_changed` signal resulting from:
       - Calling `CustomField.content_types.remove(some_content_type_that_is_already_absent)`
       - Calling `CustomField.content_types.add(some_content_type_that_is_already_present)` (this was the case in the Infoblox SSOT specifically that was resulting in hundreds/thousands of unnecessary tasks and their associated JobResults)
- Additionally, the `m2m_changed` signal wasn't currently doing anything in the case of `CustomField.content_types.clear()`, so I added some logic to account for that possibility.

# TODO

- [x] Unit, Integration Tests


nautobot/extras/tests/test_customfields.py
```diff
@@ -2083,6 +2083,30 @@ def test_delete_custom_field_data_task(self):
         self.assertEqual(oc_list[0].change_context_detail, "delete custom field data")
         self.assertEqual(oc_list[0].user, self.user)
 
+    def test_clear_custom_field_data_task(self):
```
HanlinMiao, line 2086: Should we add tests for situations where `content_types.add()/.delete()/.clear()` did not actually change the custom fields and assert that no object changes are generated?

HanlinMiao (approved):

nautobot/extras/tests/test_customfields.py
glennmatthews: Really we should have tests to assert that the tasks aren't triggered at all in those cases, but I already spent more time on this issue today than I'd meant to spend given it was an emergent bug report rather than planned work.

PR #7117, opened 2025-04-04
Branch aware job running

glennmatthews:
# What's Changed

- Add `maybe_with_branch` context manager which optionally checks out a VC branch, optionally using a specific DB connection, and optionally enables VC autocommit on exit. (see TODO below)
- When encoding/decoding Nautobot data records as Job inputs, use the appropriate branch
- Wrap the entire `run_job` task in `maybe_with_branch` (see TODO below)
- Add `maybe_with_branch` to our custom JobResult manager class so that when Celery calls `store_result()` to create a task JobResult record, it uses the appropriate branch
- Enhance `JobResult.enqueue_job()` to include the requested branch as a Celery kwarg.
- Add `maybe_with_branch` to `JobResult.log()` so that log entries are written to the correct branch even when using the separate JOB_LOGS DB connection.
- Make `JobResultTable` not throw an exception on rendering if the provided queryset doesn't have the `*_log_count` annotations included (as is the case when rendering JobResult diffs in VC).

# Screenshots

image

image

# TODO

- [ ] Can we enhance `query_on_branch` and `AutoDoltCommit` to support the `using` DB parameter, avoiding the need for a `transaction.atomic` block in this case?
- [ ] Can we enhance our Celery task runner to be branch-aware itself so that `run_job` specifically doesn't need to be and so that other Celery tasks will also run in the correct branches?
- [ ] Make ScheduledJob execution branch aware **or** disable creation of ScheduledJobs in a branch (if branch-aware, how will this need to work when a branch with ScheduledJobs gets merged down into main??)
- [ ] Make k8s job execution branch aware (might be automatic if we enhance the Celery runner itself above?)

- [ ] Unit, Integration Tests


nautobot/core/celery/task.py
```diff
@@ -4,12 +4,47 @@
...
     """Nautobot extensions to tasks for integrating with Job machinery."""
 
+    def before_start(self, task_id, args, kwargs):
+        User = get_user_model()
+        user_id = self.request.properties.get("nautobot_job_user_id", None)
+        self._nautobot_branch_context = BranchContext(
+            branch_name=self.request.properties.get("nautobot_job_branch_name", None),
+            user=User.objects.get(id=user_id) if user_id else None,
+        )
+        self._nautobot_branch_context.__enter__()
```
HanlinMiao, line 24: Should we call `super().before_start()` here and `super().after_return()` below?

nautobot/core/celery/task.py
```diff
@@ -4,12 +4,47 @@
...
+
+    def after_return(self, status, retval, task_id, args, kwargs, einfo):
+        self._nautobot_branch_context.__exit__(None, None, None)  # TODO: good enough?
+        delattr(self, "_nautobot_branch_context")
```
HanlinMiao, line 28: Why is this necessary?

nautobot/core/branching.py
```diff
@@ -0,0 +1,57 @@
...
+
+        from nautobot_version_control.utils import active_branch, checkout_branch  # pylint: disable=import-error
+
+        self.cursor = connections[self.using].cursor()
+        self.cursor.__enter__()
+        self.original_branch = active_branch(using=self.using)
+
+        if self.branch_name != self.original_branch:
+            if self.using == "default":
+                LOGGER.debug("Switching to branch %s", self.branch_name)
```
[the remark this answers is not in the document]
HanlinMiao: I see, so `self.using `would either be "default" or "JOB_LOGS". We do not want this message to be logged as actual job log entries. LMK if I'm understanding this correctly

nautobot/core/tests/test_branching.py
```diff
@@ -0,0 +1,154 @@
...
+from nautobot.core.branching import BranchContext
+from nautobot.core.testing import TestCase
+from nautobot.extras.models import Status
+from nautobot.extras.models.jobs import JOB_LOGS
+
+
+class BranchContextTest(TestCase):
+    """Test cases for the BranchContext context manager."""
+
+    databases = ["default", JOB_LOGS]
```
HanlinMiao, line 14: Is this used anywhere?

nautobot/extras/tables.py
```diff
@@ -809,14 +810,16 @@ class Meta(BaseTable.Meta):
         )
 
     def render_last_sync_time(self, record):
-        if record.name in self.context["job_results"]:  # pylint: disable=no-member
-            return self.context["job_results"][record.name].date_done  # pylint: disable=no-member
+        if hasattr(self, "context"):
+            if record.name in self.context.get("job_results", {}):  # pylint: disable=no-member
+                return self.context["job_results"][record.name].date_done  # pylint: disable=no-member
```
HanlinMiao, line 815: Wonder if we can delete the `pylint` comments now.

nautobot/extras/tables.py
```diff
@@ -809,14 +810,16 @@ class Meta(BaseTable.Meta):
...
         return self.default
 
     def render_last_sync_user(self, record):
-        if record.name in self.context["job_results"]:  # pylint: disable=no-member
-            user = self.context["job_results"][record.name].user  # pylint: disable=no-member
-            return user
+        if hasattr(self, "context"):
+            if record.name in self.context.get("job_results", {}):  # pylint: disable=no-member
+                user = self.context["job_results"][record.name].user  # pylint: disable=no-member
+                return user
```
HanlinMiao, line 822: Same comment here.

HanlinMiao (commented):
I am still digesting the code changes, sorry for asking so many questions.

nautobot/core/branching.py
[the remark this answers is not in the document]
glennmatthews: IIRC mostly this was an anti-noise thing more than anything else. We don't switch the default connection between branches via this function very frequently (when starting/completing a branched Job, when updating the JobResult state automatically to STARTED/SUCCESS/FAILURE, maybe one or two other cases I'm forgetting) but every call to the Job logger via JOB_LOGS results in switching that connection to the branch and back again, so a job that logged 100 separate messages would result in 100 "Switching to branch" and 100 "Returned to branch" debug logs on the celery worker console.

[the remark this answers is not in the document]
glennmatthews: Thinking about that again, I wonder if there might be a way around that for efficiency, maybe we could switch **both** connections to the branch when starting the Job and keep them that way until it completes? I can investigate.

nautobot/core/celery/task.py
glennmatthews: IIRC the base celery Task class has no-ops for those, but it wouldn't hurt. Will do.

nautobot/core/celery/task.py
glennmatthews: The delattr shouldn't be strictly necessary, I was just cleaning up after myself.

I should check the TODO still to see whether there's any inherent value to translating `retval`/`einfo` to the equivalent `exc_type`, `exc_value`, `traceback` that context manager `__exit__` can receive. Right now the branch context doesn't do anything with those params but it might be useful as future-proofing?

HanlinMiao (approved):
I think it is good to go as long as K8S jobs are confirmed to work.

PR #7123, opened 2025-04-06
Develop a POC for an error code system for nautobot core and apps.

itdependsnetworks:
This is just a PoC, so no review needed, and not a priority. 

I did want to put together the PoC, to show what it could look like.

Error output:
image

Navigation doc

image

An error code doc

image

Some features

* Ability for any application can add an error code
* Automatic rendering of docs and navigation
* Centralized and coordinated error codes


glennmatthews:
At a glance, this seems to me like an argument for improving the error message itself, rather than creating a separate out-of-band documentation that the user needs to refer to in order to understand what the error "really" means?

itdependsnetworks:
> At a glance, this seems to me like an argument for improving the error message itself, rather than creating a separate out-of-band documentation that the user needs to refer to in order to understand what the error "really" means?

I think we need to start with being empathetic to our users. It should be non-controversial that errors with context for how to fix them, common reasons they happened, links to other resources, etc provide clarity to the average user who has no understanding of how to read a stack trace. 

Making error's easy to find within the code has tremendous benefit by itself, as it clearly defines the interesting data to look for in the code, versus a guess and check for the error. 

Ensuring that a human has reviewed not only the error, but the use cases likely to beget a better understanding of the issues and potential re-work of how to properly address. 

Seeing the same issues happen over and over from users, to me clearly states something needs to improve. I am not sure this is the answer, but have not heard potential alternatives nor do am I seeing something fundamentally wrong with the approach.

PR #7141, opened 2025-04-09
Refactored power panel uiviewset

Sridevi-NTC:
# What's Changed

# Screenshots
|Create View|imageimage|imageimage|
| Detail View   |image|imageimage|
| Delete View|imageimage|imageimage|
# TODO

- [ ] Unit, Integration Tests


Sridevi-NTC:
@HanlinMiao, I believe I created this branch based on the contactuiviewset branch. Should I close this PR and create a new branch from develop to commit the code fresh?

HanlinMiao:
> @HanlinMiao, I believe I created this branch based on the contactuiviewset branch. Should I close this PR and create a new branch from develop to commit the code fresh?

No that's okay I am gonna merge the contact PR next so that the branch should be up to date.

Sridevi-NTC:
> No that's okay I am gonna merge the contact PR next so that the branch should be up to date.

Okay, thank you

nautobot/dcim/views.py
```diff
@@ -4138,50 +4139,37 @@ class VirtualChassisBulkDeleteView(generic.BulkDeleteView):
...
-    table = tables.PowerPanelTable
+    object_detail_content = object_detail.ObjectDetailContent(
+        panels=(
+            object_detail.ObjectFieldsPanel(
+                section=SectionChoices.LEFT_HALF,
+                weight=100,
+                fields="__all__",
+                value_transforms={
+                    "location": [helpers.render_ancestor_hierarchy],
+                    "rack_group": [helpers.hyperlinked_object],
```
HanlinMiao, line 4159: Should `rack_group` be using `render_ancestor_hierarchy` as well? I think it is a `TreeModel` like `Location`.

nautobot/core/templatetags/helpers.py
```diff
@@ -705,7 +705,11 @@ def render_ancestor_hierarchy(value):
         )
         append_to_result += format_html("</ul></li>")
     nestable_tag = format_html('<span title="nestable">↺</span>' if getattr(value, "nestable", False) else "")
-    result += format_html("<li><strong>{value} {nestable_tag}</strong></li>", value=value, nestable_tag=nestable_tag)
+    result += format_html(
+        "<li><strong>{value} {nestable_tag}</strong></li>",
+        value=hyperlinked_object(value, "name"),
+        nestable_tag=nestable_tag,
+    )
```
HanlinMiao, line 712: `value` is the `TreeModel` object itself right? Would it make sense to `hyperlink` it, since you are already at its detail view?

HanlinMiao (commented):
Looks good! Need a changelog fragment

nautobot/core/templatetags/helpers.py
Sridevi-NTC: https://github.com/nautobot/nautobot/blob/develop/nautobot/core/ui/object_detail.py#L1116

According to the reference above, when the key is `_hierarchy`, it is rendered as ancestor_hierarchy. Is there a specific condition, such as `isinstance(obj, TreeModel)` or similar one, that triggers this rendering as ancestor_hierarchy?

nautobot/core/templatetags/helpers.py
```diff
@@ -692,20 +692,62 @@ def render_content_types(value):
...
+@register.filter()
+def location_hierarchy_with_location_type(value):
+    return render_ancestor_hierarchy(value, include_location_type=True)
+
+
 @library.filter()
 @register.filter()
-def render_ancestor_hierarchy(value):
-    """Renders a nested HTML list representing the hierarchy of ancestors for a given object."""
+def render_ancestor_hierarchy(value, include_location_type=False):
```
glennmatthews, line 702: Instead of adding a new parameter and a new template tag, can we just make this tag do the right thing automatically? Something like an initial `include_location_type = hasattr(value, "location_type")` would seem like it should work.

nautobot/dcim/templates/dcim/powerpanel_retrieve.html
glennmatthews, line 1: Why does this new file need to be added?

Sridevi-NTC: For `extra_breadcrumbs ` I have added this

nautobot/core/templatetags/helpers.py
Sridevi-NTC: sure @glennmatthews

nautobot/dcim/urls.py
```diff
@@ -44,6 +43,7 @@
 router.register("module-bay-templates", views.ModuleBayTemplateUIViewSet)
 router.register("modules", views.ModuleUIViewSet)
 router.register("module-types", views.ModuleTypeUIViewSet)
+router.register("power-panels", views.PowerPanelUIViewSet)
```
HanlinMiao, line 46: `power-feeds` should come first

nautobot/dcim/templates/dcim/powerpanel_retrieve.html
```diff
@@ -0,0 +1,9 @@
+{% extends 'generic/object_retrieve.html' %}
+{% load helpers %}
+{% block extra_breadcrumbs %}
+    <li>{{ object.location|hyperlinked_object }}</li>
```
glennmatthews, line 4: We should fix the issue that `NautobotUIViewSet` isn't correctly detecting and rendering `powerpanel.html`, rather than introducing a new, redundant template as a part of this fix. See discussion on our Slack from Monday.

nautobot/dcim/views.py
```diff
@@ -4106,50 +4107,37 @@ class VirtualChassisBulkDeleteView(generic.BulkDeleteView):
...
-    table = tables.PowerPanelTable
+    object_detail_content = object_detail.ObjectDetailContent(
+        panels=(
+            object_detail.ObjectFieldsPanel(
+                section=SectionChoices.LEFT_HALF,
+                weight=100,
+                fields="__all__",
+                value_transforms={
+                    "location": [helpers.render_ancestor_hierarchy],
+                    "rack_group": [helpers.render_ancestor_hierarchy],
```
glennmatthews, line 4127: Isn't this being handled automatically already by logic in `ObjectFieldsPanel`?

Sridevi-NTC: I guess location and rackgroups are instance of models.Model hence they are displaying as hyperlinked_object
https://github.com/nautobot/nautobot/blob/develop/nautobot/core/ui/object_detail.py#L986

glennmatthews: I was thinking of https://github.com/nautobot/nautobot/blob/develop/nautobot/core/ui/object_detail.py#L1170-L1172 but I guess that's only if the object itself is a TreeModel, not if any of its related objects are. We should fix this at or around the line 986 you linked above.

nautobot/dcim/templates/dcim/powerpanel_retrieve.html
HanlinMiao: Yeah, we need to refactor the logic in `get_base_template()` to detect both `object.html` and `object_retrieve.html`. Let me take a stab at this.

itdependsnetworks: Was speaking to Hanlin about this, imho we don't make it work for both, if I am looking for the template, to me it would be better to have the single standard vs any actual or perceived magic that future folks need to find.

glennmatthews: The alternative is this mess of duplicate/redundant files though, where we have both `foo.html` (for backward compatibility with apps) and `foo_retrieve.html` (for any custom templating logic) and it's even more confusing to anyone looking at the code base.

itdependsnetworks: These are only until 3.0. I don't think we should have reasonable expectations that these templates should be considered in the [public interface](https://docs.nautobot.com/projects/core/en/stable/development/core/#public-interface).

glennmatthews: The problem is that IIRC if there are Apps that are adding custom tabs to an existing object detail view, in at least *some* cases (thinking of Device in particular here) there has been a requirement for the custom tab's template to inherit from the base view template.

Personally, I consider the introduction of the secondary `foo_retrieve.html` pattern in NautobotUIViewSet to have been a misstep on our part and would prefer not to propagate it as a requirement.

nautobot/core/templatetags/helpers.py
```diff
@@ -695,17 +695,52 @@ def render_content_types(value):
...
-            "<li>{value} {nestable_tag}<ul>", value=hyperlinked_object(ancestor, "name"), nestable_tag=nestable_tag
+            "<li><strong>{value} {location_type} {nestable_tag}</strong></li>",
+            value=hyperlinked_object(value, "name"),
+            location_type=location_type,
+            nestable_tag=nestable_tag,
+        )
+
+    else:
+        result += format_html(
+            "<li><strong>{value}{nestable_tag}</strong></li>",
```
glennmatthews, line 740:
```suggestion
            "<li><strong>{value} {nestable_tag}</strong></li>",
```

nautobot/dcim/views.py
```diff
@@ -4102,50 +4102,37 @@ class VirtualChassisBulkDeleteView(generic.BulkDeleteView):
...
+    object_detail_content = object_detail.ObjectDetailContent(
+        panels=(
+            object_detail.ObjectFieldsPanel(
+                section=SectionChoices.LEFT_HALF,
+                weight=100,
+                fields=[
+                    "name",
+                    "location",
+                    "rack_group",
+                ],
```
glennmatthews, line 4123: Is there any problem with using `__all__` instead of an explicit list of fields here?

Sridevi-NTC: Yes @glennmatthews, power panel name is at the bottom of the table
image

glennmatthews (approved):

PR #7148, opened 2025-04-10
Refactored RackGroup

Pavan-183:
# What's Changed

# Screenshots

| View             | Before                                                                                                                                         | After                                                                                                                                          |
|------------------|-------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------|
| List View        | image                                                      | image                                                       |
| Detail View      | image                                                      | image                                                       |
| Update View      | image image | image image |
| Delete View      | image image | image image |
| Bulk Edit View   |                                                                                                                                                 | image image |
| Bulk Delete View |                                                                                                                                                 | image image |

# TODO

- [ ] Unit, Integration Tests


nautobot/dcim/tests/test_views.py
```diff
@@ -451,6 +451,10 @@ def setUpTestData(cls):
             "location": location.pk,
             "description": "A new rack group",
         }
+        cls.bulk_edit_data = {
+            "description": "Updated description",
+            "location": location.pk,
```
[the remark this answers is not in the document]
Pavan-183: Sure @HanlinMiao

Pavan-183:
> Can you provide the screenshots for before and after?

Updated now @HanlinMiao .

nautobot/dcim/views.py
```diff
@@ -489,54 +489,54 @@ def post(self, request, *args, **kwargs):
...
-
-class RackGroupView(generic.ObjectView):
-    queryset = RackGroup.objects.all()
+    object_detail_content = object_detail.ObjectDetailContent(
+        panels=[
+            object_detail.ObjectFieldsPanel(
+                label="Rack Group",
+                section=SectionChoices.LEFT_HALF,
+                weight=100,
+                fields=["description", "parent", "location"],
```
HanlinMiao, line 507: location should use `render_ancestor_hierarchy` `value_transform` here https://github.com/nautobot/nautobot/pull/7141/files#diff-292c4e491fa8d2a6b999f52af45f55576703202fa9bd731438c815c0af769af5R4125-R4127

glennmatthews: I may be mistaken, given that's what we currently have in the `before` case though.

nautobot/dcim/tests/test_views.py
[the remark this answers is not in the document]
glennmatthews: Not sure I understand the concern with regard to bulk-edit specifically though - don't we already allow changing a rack-group's location through the individual edit form?

nautobot/dcim/views.py
HanlinMiao: Yeah, I think let's keep the before and after consistent. So use `location_hierarchy_with_location_type` here. Thanks for the investigation @Pavan-183

nautobot/dcim/tests/test_views.py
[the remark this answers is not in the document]
HanlinMiao: I see that the validation logic for the `rack's` location and the corresponding `rackgroup's` location is in the Rack's `clean()` method, and no validation logic on the `RackGroup's` side.
Hmm, I wonder if we are allowing mismatching locations between Rack and RackGroup by allowing changing a rack-group's location. Definitely out of scope for this PR tho.

nautobot/dcim/views.py
Pavan-183: Hi, @HanlinMiao I think @Sridevi-NTC has updated the code and removed location_hierarchy_with_location_type, so I will now use render_ancestor_hierarchy instead.

glennmatthews (approved):

PR #7150, opened 2025-04-10
7131 fix need for  description filter in graphene v

### LINKED ISSUE #7131: Fix need for `_description` filter in Graphene v3
Following #7070, GraphQL query filters no longer support a `description` filter, but instead require filtering by description to reference the filter as `_description` due to `description` being a reserved keyword argument on Graphene's `Field` class (https://docs.graphene-python.org/en/latest/api/#fields-mounted-types). 

It looks like we should be able to work around this, and once again permit a `description` filter, using something like `args={"description": String()}`, see https://docs.graphene-python.org/en/latest/types/objecttypes/#resolverparamgraphqlarguments for example code.

dsoftait:
# Closes #7131
# What's Changed
- Fixed graphene filter `description`
- Fixed test `test_search_parameters`
- Changed docs
# Screenshots
Screenshot 2025-04-10 at 15 53 38
Screenshot 2025-04-10 at 16 00 09
Screenshot 2025-04-10 at 16 01 49

# TODO
- [ ] Unit, Integration Tests


nautobot/core/graphql/utils.py
```diff
@@ -81,9 +81,9 @@ def get_filtering_args_from_filterset(filterset_class):
             required=False,
         )
 
-    # Hack to swap `description` fields to `_description` since they will conflict with `graphene.types.field.Field.description`.
+    # Hack to avoid conflict with `graphene.types.field.Field.description`.
     if "description" in args:
-        args["_description"] = args.pop("description")
+        args["args"].update({"description": args.pop("description")})
```
glennmatthews, line 86: Maybe as simple as:

```py
if "type" in args and "_type" not in args:
    # for backwards compatibility with our filters in graphene v2 where `type` was a reserved keyword
    args["_type"] = args["type"]
```

itdependsnetworks: Wouldn't it be

```python
if "_type" in args and "type" not in args: # for backwards compatibility with our filters in graphene v2 where `type` was a reserved keyword
    args["type"] = args["_type"]
```

We mark deprecated now, log it, then fail migrations on it say 3.2 / 3.3?

glennmatthews: this is defining the schema, not processing user input.

itdependsnetworks: thanks! That is what I was missing.

dsoftait: Adding this code causes this error
```
nautobot-1       |   File "/source/nautobot/core/graphql/schema_init.py", line 5, in <module>
nautobot-1       |     DynamicGraphQL = generate_query_mixin()
nautobot-1       |                      ^^^^^^^^^^^^^^^^^^^^^^
nautobot-1       |   File "/source/nautobot/core/graphql/schema.py", line 545, in generate_query_mixin
nautobot-1       |     class_attrs.update(generate_attrs_for_schema_type(schema_type))
nautobot-1       |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
nautobot-1       |   File "/source/nautobot/core/graphql/generators.py", line 355, in generate_attrs_for_schema_type
nautobot-1       |     attrs[list_name] = DjangoListField(schema_type, **search_params)
nautobot-1       |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
nautobot-1       | TypeError: DjangoListField.__init__() got multiple values for argument '_type'
```

HanlinMiao: Here is the reason why we are running into this error. Maybe we should switch back to `graphene.List()`? Since it does not have a position argument `_type`?
https://github.com/nautobot/nautobot/pull/7070/files#r2047821177

HanlinMiao: Switching back to `graphene.List()` does fix the issue.
Screenshot 2025-04-16 at 5 56 59 PM

glennmatthews: Fix using `args["args"]` incoming. Still not sure we *need* DjangoListField versus graphene.List, but it's at least self-consistent.

glennmatthews (approved):

PR #7152, opened 2025-04-10
Location.display now honors LOCATION_NAME_AS_NATURAL_KEY

lampwins:
# What's Changed

When `LOCATION_NAME_AS_NATURAL_KEY=True`, `Location.display()` will honor the setting and only return `Location.name` otherwise it will fallback to `TreeModel.display()` which renders the full tree, joined on `→`.

# Screenshots

With `LOCATION_NAME_AS_NATURAL_KEY=False`
Screenshot 2025-04-10 at 1 58 36 PM

With `LOCATION_NAME_AS_NATURAL_KEY=True`
Screenshot 2025-04-10 at 1 59 21 PM

# TODO

- [x] Unit, Integration Tests

### DIFF
--- a/changes/7152.fixed
+++ b/changes/7152.fixed
@@ -1,0 +1,1 @@
+Location.display now honors LOCATION_NAME_AS_NATURAL_KEY - meaning the display of location names will not render the full hierarchy, in places where only the name is relevant.


glennmatthews (approved):
Seems reasonable as an enhancement to me, though we might *also* want the Location breadcrumbs (at least) to always show only the `name` even when this setting is False?

We'll need to keep an eye out for any impact to apps that are making assumptions about the `display` behavior (a la https://github.com/nautobot/pynautobot/pull/236 when we changed `Prefix.display`)

PR #7155, opened 2025-04-11
Refactored RackReservationUIviewset

Pavan-183:
# What's Changed

# Screenshots

| View             | Before                                                                                                                                         | After                                                                                                                                          |
|------------------|-------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------|
| List View        | image                                                      | image                                                       |
| Detail View      | image                                                      | image                                                       |
| Update View      | image image | image image |
| Delete View      | image image | image image |
| Bulk Edit View   |                                                                                                                                                 | image image |
| Bulk Delete View |                                                                                                                                                 | image image |
| ADD Reservation(USER Autofills)  | image                                                      | image                                                       |
# TODO

- [ ] Unit, Integration Tests


nautobot/dcim/views.py
```diff
@@ -672,50 +679,13 @@ class RackBulkDeleteView(generic.BulkDeleteView):
...
-class RackReservationEditView(generic.ObjectEditView):
+class RackReservationUIViewSet(NautobotUIViewSet):
     queryset = RackReservation.objects.all()
-    model_form = forms.RackReservationForm
-    template_name = "dcim/rackreservation_edit.html"
-
-    def alter_obj(self, obj, request, url_args, url_kwargs):
-        if not obj.present_in_database:
-            if "rack" in request.GET:
-                obj.rack = get_object_or_404(Rack, pk=request.GET.get("rack"))
```
itdependsnetworks, line 694: How is this taken into consideration.

Pavan-183: Hi @itdependsnetworks , I now understand the issue more clearly.
Previously, the user field was being automatically prefilled with the logged-in user during object creation. After switching to NautobotUIViewSet, that behavior seems to be missing because the necessary logic wasn’t carried over into the new form/viewset setup.

I’m currently working on bringing back that functionality by updating the form to prefill the field and making sure the user gets set correctly before saving.

Correct me if I’m wrong @itdependsnetworks

itdependsnetworks: Seems reasonable to me, just doing a stare and compare so making sure we carry everything over, but don’t fully understand all of the details here

nautobot/dcim/templates/dcim/rackreservation_edit.html
```diff
@@ -1,24 +1,2 @@
...
-            {% render_field form.description %}
-        </div>
-    </div>
-    <div class="panel panel-default">
-        <div class="panel-heading"><strong>Tenant Assignment</strong></div>
-        <div class="panel-body">
-            {% render_field form.tenant_group %}
-            {% render_field form.tenant %}
-        </div>
-    </div>
```
glennmatthews, line 22: Do we still have this logical grouping and ordering of the form fields after removing this? Otherwise this seems over-aggressive removal of code.

nautobot/dcim/templates/dcim/rackreservation_retrieve.html
glennmatthews, line 1: What's the need for moving everything from `rackreservation.html` into this new file?

changes/7154.added
```diff
@@ -0,0 +1 @@
+Added bulk-edit support for the RackReservation model.
```
glennmatthews, line 1: This functionality appears to already exist - remove this note?

nautobot/dcim/views.py
```diff
@@ -55,7 +55,9 @@
 from nautobot.core.views.paginator import EnhancedPaginator, get_paginate_count
 from nautobot.core.views.viewsets import NautobotUIViewSet
 from nautobot.dcim.choices import LocationDataToContactActionChoices
-from nautobot.dcim.forms import LocationMigrateDataToContactForm
+from nautobot.dcim.forms import (
+    LocationMigrateDataToContactForm,
+)
```
glennmatthews, line 60: Revert?

nautobot/dcim/views.py
```diff
@@ -672,50 +672,26 @@ class RackBulkDeleteView(generic.BulkDeleteView):
...
         if not obj.present_in_database:
-            if "rack" in request.GET:
-                obj.rack = get_object_or_404(Rack, pk=request.GET.get("rack"))
-            obj.user = request.user
-        return obj
+            obj.user = self.request.user
 
+            rack_id = self.request.GET.get("rack")
+            if rack_id:
+                obj.rack_id = rack_id
```
HanlinMiao, line 692: Can you confirm this works? In the original code, we are assigning to `obj.rack` instead of `obj.rack_id`.
And should we do an obj.save() after this assignment?

Pavan-183: Yes, this works correctly @HanlinMiao 
I'm just setting the initial value for the form using obj.rack_id = rack_id.
We don't need to call obj.save() here because the object isn’t being saved yet , it will be saved later after the form is submitted.

glennmatthews: I *think* this change is not entirely desirable, because `rack_id` could be any UUID, whereas before we would throw a 404 error if it was not an actual existing rack?

Pavan-183: Updated as requested, should be resolved, let me know if I missed anything.

glennmatthews (approved):

PR #7158, opened 2025-04-13
Refactored cloudnetwork uicomponent

Sridevi-NTC:
# What's Changed

# Screenshots

# TODO

- [ ] Unit, Integration Tests


Sridevi-NTC:
We have extra tabs for children, circuits, cloud sources, and prefixes under cloudnetwork. I've implemented them using `DistinctViewTab `and `action decorators`. Is there a way to use `add_child `in the table footer section within DistinctViewTab?

Addchild

HanlinMiao:
> We have extra tabs for children, circuits, cloud sources, and prefixes under cloudnetwork. I've implemented them using `DistinctViewTab `and `action decorators`. Is there a way to use `add_child `in the table footer section within DistinctViewTab?
> 
> Addchild

Checkout [panel_table.html](https://github.com/nautobot/nautobot/blob/develop/nautobot/core/templates/panel_table.html#L17) and if you pass in an `add_obj_url` context variable, a button will be rendered.

Sridevi-NTC:
Please Note that Instead Add child button we have `Add cloud network` now
image

Sridevi-NTC:
Restored edit, delete actions as per https://github.com/nautobot/nautobot/pull/7163#issuecomment-2804960637
image

nautobot/cloud/forms.py
```diff
@@ -155,9 +155,7 @@ class CloudNetworkBulkEditForm(TagsBulkEditFormMixin, NautobotBulkEditForm):
     add_prefixes = DynamicModelMultipleChoiceField(
         queryset=Prefix.objects.all(), required=False, query_params={"namespace": "$namespace"}
     )
-    remove_prefixes = DynamicModelMultipleChoiceField(
-        queryset=Prefix.objects.all(), required=False, query_params={"namespace": "$namespace"}
-    )
+    remove_prefixes = DynamicModelMultipleChoiceField(queryset=Prefix.objects.all(), required=False)
```
HanlinMiao, line 158: Can you state the reason why you removed the `query_params` here and not on `add_prefixes`? Should we remove both?

Sridevi-NTC: If wanted to remove `10.31.2.0/24: Global ` and  i selected `branch` namespace in the edit form, in remove prefixes drop down i have only branch prefixes.

image

HanlinMiao: Sorry, I am a little confused. The namespace field is serving as a `filter` field to filter the `prefixes` dropdown based on the value of the `namespace`.
So this is intended behavior. If you want to remove `10.31.2.0/24: Global`, you should select the `Global` namespace in the Bulk Edit form.

HanlinMiao: If you want to select IPAddresses from different namespaces, I would say:
1. Select the namespace `branch`, pick `10.32.0.0/11: branch`
2. Then select the namespace `global`, and pick `10.31.2.0/24: Global`

Let me know if that works.

nautobot/cloud/views.py
```diff
@@ -67,50 +71,130 @@ class CloudNetworkUIViewSet(NautobotUIViewSet):
...
+        )
+        children_table.columns.hide("parent")
+        RequestConfig(
+            request, paginate={"paginator_class": EnhancedPaginator, "per_page": get_paginate_count(request)}
+        ).configure(children_table)
+        child_url = reverse("cloud:cloudnetwork_add")
+        child_table_add_url = f"{child_url}?parent={instance.id}"
+        return Response(
+            {"children_table": children_table, "active_tab": "children", "child_table_add_url": child_table_add_url}
+        )
```
glennmatthews, line 142: This *feels* like a step backwards to me. Ideally each of these distinct detail action views would be able to have their own `object_detail_content` that just contains a single `ObjectsTablePanel`, instead of needing all of this custom data logic in each case. Can we investigate that option?

HanlinMiao: https://github.com/nautobot/nautobot/pull/7158/commits/175c238282cd6bf1cd91dd04884dc390599d2b0f

nautobot/core/ui/object_detail.py
```diff
@@ -388,18 +389,17 @@ def __init__(
...
         label_wrapper_template_path="components/tab/label_wrapper_distinct_view.html",
         related_object_attribute="",
+        panels=(),
         **kwargs,
     ):
         self.url_name = url_name
         self.related_object_attribute = related_object_attribute
-        super().__init__(label_wrapper_template_path=label_wrapper_template_path, **kwargs)
+        self.panels = panels
+        super().__init__(label_wrapper_template_path=label_wrapper_template_path, panels=panels, **kwargs)
```
glennmatthews, line 398: Similarly, aren't these two lines redundant?

nautobot/core/ui/object_detail.py
```diff
@@ -814,9 +820,25 @@ def get_extra_context(self, context: Context):
...
+                    body_content_table_queryset,
+                    hide_hierarchy_ui=self.hide_hierarchy_ui,
+                    extra_columns=[
+                        (
+                            "actions",
+                            ButtonsColumn(
+                                model=body_content_table_queryset.model, return_url_extra=f"?tab={self.tab_id}"
+                            ),
+                        )
+                    ],
```
glennmatthews, line 836: This feels wrong to me... can we instead allocate the table as before and just modify `body_content_table.columns["actions"]` directly instead?

HanlinMiao: Apparently, we cannot modify `body_content_table.columns`, but we can with `body_content_table.base_columns`. That achieved the same effect.

nautobot/core/views/renderers.py
```diff
@@ -293,7 +293,17 @@ def get_context(self, data, accepted_media_type, renderer_context):
...
+        if view.action not in [
+            "list",
+            "create",
+            "update",
+            "destroy",
+            "bulk_create",
+            "bulk_update",
+            "bulk_destroy",
+            "bulk_rename",
+        ] and not view.action.startswith("bulk_add"):
```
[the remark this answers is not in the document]
HanlinMiao: Gonna have to revert this because of https://github.com/nautobot/nautobot/actions/runs/14539230972/job/40793943634

nautobot/core/ui/object_detail.py
```diff
@@ -820,24 +818,15 @@ def get_extra_context(self, context: Context):
...
-                            ),
-                        )
-                    ],
-                )
-            else:
-                body_content_table = body_content_table_class(
-                    body_content_table_queryset, hide_hierarchy_ui=self.hide_hierarchy_ui
+                body_content_table.base_columns["actions"] = ButtonsColumn(
+                    model=body_content_table_queryset.model,
+                    return_url_extra=f"?tab={self.tab_id}",
```
glennmatthews, line 829: This still feels off to me - lots of tables have a custom `actions` column (e.g. with nonstandard `buttons` tuple, custom `prepend_template`, etc.) that we're just overwriting here with something generic. Can we modify the existing object with a `return_url_extra` rather than replacing it entirely? Or something like that?

HanlinMiao: Should be fixed in this commit https://github.com/nautobot/nautobot/pull/7158/commits/be37636e0a6d9b2c21a6ce778452641d44321a0f

nautobot/core/views/renderers.py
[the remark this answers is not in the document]
glennmatthews: I'd like to understand this further - this feels overly fragile as implemented to me.

nautobot/cloud/templates/cloud/cloudnetwork_children.html
```diff
@@ -0,0 +1,3 @@
+{% extends 'cloud/cloudnetwork_retrieve.html' %}
+
+{% block title %}{{ block.super }} - Children {% endblock %}
```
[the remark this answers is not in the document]
HanlinMiao: I am honestly not sure if this pattern is prevalent or is something we want to enforce in core. I feel like it is a little unnecessary.
Screenshot 2025-04-21 at 12 19 10 PM

Updated `get_template_name` helper function to handle custom action cases so we no longer need for those custom templates to exist.

HanlinMiao (approved):

HanlinMiao (commented):
Should be qa-approved, just need to resolve a merge conflict

glennmatthews (approved):

PR #7159, opened 2025-04-14
Navbar favorites

### LINKED ISSUE #7068: Navbar Favorites
As a Nautobot user, I want the ability to "favorite" items that are listed within the navbar, so that those favorited items show up in a consolidated list of "Favorites" at the top of the navbar, and I can quickly access them there.

Requirements:
- Each user may select their own favorites
- A user's favorites persist
- Users may favorite any item present within the navbar
- Items that have been favorited will still render in their normal position within the navbar, but will also be added to the favorites list
- Users favorite an item by clicking the star icon to the right of the item, turning the star a solid color.
- Users may unfavorite an item at any time by clicking the solid star icon
- The favorites list is sorted alphabetically.

Future facing but out of scope for now:
- We will likely expand favorites to include links to Saved Views, other than the default list view and to specific object detail views.
  - With this, users will likely want the ability to group items together (e.g. bookmark folders), but we can likely limit that to one level of grouping

mateusz-lachowski-codilime:
# Closes #7068 
# What's Changed

- Added HTMX to the Nautobot Core
- Added navbar favorites HTMX endpoints
- Navbar favorites feature is now working and saving items dynamically per user

# Screenshots
image
image
image
[Screencast from 2025-04-14 13-59-17.webm](https://github.com/user-attachments/assets/37849650-592f-4126-b2d5-bb117611ffef)

# TODO
- Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/core/#creating-changelog-fragments))
- Unit, Integration Tests


nautobot/core/templates/base_django.html
```diff
@@ -1,15 +1,17 @@
...
 {% load plugins %}
+{% load django_htmx %}
 
 <!DOCTYPE html>
 <html lang="en"{% if request.COOKIES|get_item:"theme" == 'dark' %} data-theme="dark"{% endif %}>
 <head>
     <title>{% block title %}Home{% endblock %} - {{ settings.BRANDING_TITLE }}</title>
     {% include 'inc/media.html' %}
     {% block extra_styles %}{% endblock %}
+    {% htmx_script %}
```
mateusz-lachowski-codilime, line 12: We can adjust this later to have htmx sources separately from `django-htmx` package and load it via webpack but this is convenient for now.

nautobot/core/templates/inc/nav_favorites.html
```diff
@@ -0,0 +1,31 @@
...
+                    {% endif %}
+                </a>
+                <button class="sidenav-favorite favorite-item"
+                        type="button"
+                        hx-post="{%  url "user:navbar_favorites" %}"
+                        hx-vals='{"link": "{{ item.link }}", "remove": true}'
+                        hx-target="#sidenav-favorite-link-{{ forloop.counter }}"
+                        hx-swap="delete"
+                        hx-select-oob="#{{ item.tab_name }}"
+                        hx-on::before-cleanup-element="showEmpty()"
```
mateusz-lachowski-codilime, line 18: I will left some explanation here :)
- `/navbar-favorites` endpoint is returning the whole navbar instead of just "small" fragment of html
- `hx-post` means: whenever user click this, send a post to this url
- `hx-vals` post data that we want to attach to the payload
- `hx-swap` method of swapping: delete here, because this is a "favorite" tab and we want only to "remove" items
- `hx-target` item that we will delete
- `hx-select-oob` this is tricky at first point, because I'm using OOB - Out Of Band swap. The whole navbar is marked as oob ready, but here I'm selecting only the "part" of it to swap only one "tab". Swapping more / at higher html level cause the main JS to fail and then navbar sliding is not working.

nautobot/core/templates/inc/nav_menu.html
```diff
@@ -32,56 +32,64 @@
...
-                        </div>
-                    </li>
-                {% endif %}
-            {% endfor %}
+                            <div class="sidenav-flyout" id="sidenav-flyout-{{ forloop.counter }}" hx-vals='{"tab_name": "{{ tab_name }}"}'
+                                 id="{{ tab_name }}" hx-swap-oob="true"
+                            >
+                                {% for group_name, group_details in tab_details.groups.items %}
+                                    {% if not group_details.permissions or request.user|has_one_or_more_perms:group_details.permissions %}
+                                        <ul class="list-unstyled" hx-vals='{"group_name": "{{ group_name }}"}'>
```
mateusz-lachowski-codilime, line 61: `hx-vals` again and because it's inherited I can put it here and will be passed to each button click.

nautobot/core/templates/inc/nav_favorites.html
```diff
@@ -0,0 +1,31 @@
...
+                    <span class="sidenav-link-context">{{ item.tab_name }} / {{ item.group_name }}</span>
+                    {% endif %}
+                </a>
+                <button class="sidenav-favorite favorite-item"
+                        type="button"
+                        hx-post="{%  url "user:navbar_favorites" %}"
+                        hx-vals='{"link": "{{ item.link }}", "remove": true}'
+                        hx-target="#sidenav-favorite-link-{{ forloop.counter }}"
+                        hx-swap="delete"
+                        hx-select-oob="#{{ item.tab_name }}"
```
mateusz-lachowski-codilime, line 17: It would be great to update the registry structure to have also a "slugified" tab name and a link name, not the already resolved url. Then it would be safer here to swap proper tabs. But I wanted to discuss if it's desired change or not.

nautobot/users/views.py
```diff
@@ -195,6 +205,33 @@ def post(self, request):
             return redirect("user:preferences")
 
 
+class UserNavbarFavoritesView(GenericView):
+
+    def post(self, request):
```
mateusz-lachowski-codilime, line 210: I was struggling with this method, because at first I started with post and delete methods to be somehow "compliant" with standard HTTP. But I couldn't properly choose between those methods during the requests being made at HTMX level. I could use header like `HTMX-Method` but then still, the HTTP method was POST and only header could be set as `DELETE`. Also dynamic swapping of attributes from `hx-post` to `hx-delete` was not working properly.

nautobot/core/templates/inc/nav_favorites.html
```diff
@@ -0,0 +1,31 @@
...
+                       Remove from favorites
+                    </span>
+                </button>
+            </li>
+        {% endif %}
+    {% endfor %}
+
+    <li class="sidenav-link-group text-center{% if request.user.navbar_favorites %} d-none{% endif %}" style="text-transform: none;" id="favorites-empty">
+        You don't have any favorite pages yet.
+    </li>
```
glennmatthews, line 30: Could this be conditionally rendered by adding an `{% empty %}` clause inside the `for` loop?

nautobot/core/templates/inc/nav_favorites.html
glennmatthews: We'll definitely need a slugified tab name so that tabs like `Golden Config` don't break. Still not sure I understand why we need to use a link name - do we just need to url-encode the link here for safety instead?

nautobot/core/templates/inc/nav_menu.html
```diff
@@ -32,56 +32,64 @@
...
-                            type="button"
-                        >
-                            <!-- TODO(norbert-mieczkowski-codilime): This is just a placeholder icon, replace it when new icons are ready to use. -->
-                            <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><circle cx="12" cy="12" fill="none" r="10" stroke="currentcolor" stroke-width="1.5" /></svg>
-                            <span>{{ tab_name }}</span>
-                        </button>
+            {% with request.user.navbar_favorites_link_list as navbar_favorites_link_list %}
+                {% for tab_name, tab_details in registry.nav_menu.tabs.items %}
+                    {% if not tab_details.permissions or request.user|has_one_or_more_perms:tab_details.permissions %}
+                        <li class="navbar-tabs">
```
norbert-mieczkowski-codilime, line 44: unused class, there are no `tabs` in navbar anyway so :shrug:

nautobot/core/templates/inc/nav_favorites.html
glennmatthews: I'd like to understand that last point in more depth. Is there just something in the bootstrap JS that we need to call after the swap to rehydrate the replaced sidenav?

nautobot/core/templates/inc/nav_menu.html
```diff
@@ -32,56 +32,64 @@
...
+                                                        </a>
+                                                        <button class="sidenav-favorite{% if item_link in navbar_favorites_link_list %} favorite-item{% endif %}"
+                                                                type="button"
+                                                                hx-post="{%  url "user:navbar_favorites" %}"
+                                                                hx-vals='js:{"name": "{{ item_details.name }}", "link": "{{ item_link }}", "remove": event.target.classList.contains("favorite-item")}'
+                                                                hx-target="#sidenav-flyout-favorites"
+                                                                hx-select="#sidenav-flyout-favorites"
+                                                                hx-swap="outerHTML"
+                                                                hx-on::after-request="toggleFavorite(this, event)"
+                                                        >
```
norbert-mieczkowski-codilime, line 77: this style of html formatting is very outdated, AFAIK no modern code formatter recognizes it. now, I know we don't employ formatters for UI code (*yet*, hopefully) but the least we can do is to apply standards manually:

```suggestion
                                                        <button
                                                            class="sidenav-favorite{% if item_link in navbar_favorites_link_list %} favorite-item{% endif %}"
                                                            type="button"
                                                            hx-post="{%  url "user:navbar_favorites" %}"
                                                            hx-vals='js:{"name": "{{ item_details.name }}", "link": "{{ item_link }}", "remove": event.target.classList.contains("favorite-item")}'
                                                            hx-target="#sidenav-flyout-favorites"
                                                            hx-select="#sidenav-flyout-favorites"
                                                            hx-swap="outerHTML"
                                                            hx-on::after-request="toggleFavorite(this, event)"
                                                        >
```

nautobot/core/templates/inc/nav_menu.html
```diff
@@ -32,56 +32,64 @@
...
+                                            <li class="sidenav-link-group" data-group-weight="{{ group_details.weight }}">{{ group_name }}</li>
+                                            {% for item_link, item_details in group_details.items.items %}
+                                                {% if not item_details.permissions or request.user|has_one_or_more_perms:item_details.permissions %}
+                                                    <li class="sidenav-link-wrapper{% if item_details.permissions and not request.user|has_perms:item_details.permissions %} disabled{% endif %}">
+                                                        <a class="sidenav-link{% if item_link in request.path %} sidenav-link-active{% endif %}" data-item-weight="{{ item_details.weight }}" href="{{ item_link }}">
+                                                            {{ item_details.name }}
+                                                        </a>
+                                                        <button class="sidenav-favorite{% if item_link in navbar_favorites_link_list %} favorite-item{% endif %}"
```
norbert-mieczkowski-codilime, line 69: according to bootstrap standard, `active`, rather than `favorite-item`, should be used as class name here

nautobot/core/templates/inc/nav_menu.html
```diff
@@ -32,56 +32,64 @@
...
+                                                                type="button"
+                                                                hx-post="{%  url "user:navbar_favorites" %}"
```
norbert-mieczkowski-codilime, line 71: double whitespace after `%`

nautobot/users/forms.py
```diff
@@ -75,6 +75,14 @@ class PreferenceProfileSettingsForm(BootstrapMixin, forms.Form):
     timezone = TimeZoneFormField(required=False, help_text="Set your default timezone", widget=StaticSelect2)
 
 
+class NavbarFavoritesForm(forms.Form):
+    link = forms.CharField()
+    name = forms.CharField(required=False)
+    tab_name = forms.CharField(required=False)
+    group_name = forms.CharField(required=False)
```
glennmatthews, line 82: Why are these three `required=False`?

pyproject.toml
```diff
@@ -139,6 +139,8 @@ social-auth-core = {version = "~4.5.6", optional = true, extras = ["saml"]}
 social-auth-app-django = "~5.4.3"
 # Rendering of SVG images (for rack elevations, etc.)
 svgwrite = "~1.4.2"
+# Utility tools to support HTMX requests within a Django framework.
+django-htmx = "^1.23.0"
```
glennmatthews, line 143: Please move this up to keep these alphabetized.

nautobot/core/templates/inc/nav_menu.html
```diff
@@ -32,56 +32,64 @@
...
+                                                                hx-vals='js:{"name": "{{ item_details.name }}", "link": "{{ item_link }}", "remove": event.target.classList.contains("favorite-item")}'
```
norbert-mieczkowski-codilime, line 72: JSON should have a space after an opening bracket and before closing bracket. again, UI code formatting standards

nautobot/project-static/src/js/nautobot.js
```diff
@@ -70,3 +70,20 @@ document.addEventListener('DOMContentLoaded', function() {
...
+const showEmpty = () => {
+    const favoritesEmpty = document.getElementById('favorites-empty');
+    if(document.querySelectorAll('#sidenav-flyout-favorites > ul > li').length - 1 === 1){
+        favoritesEmpty.classList.remove('d-none')
+    }
+}
+
+window.toggleFavorite = toggleFavorite
+window.showEmpty = showEmpty
+
```
norbert-mieczkowski-codilime, line 89: 1. thank god semicolons and spaces are free
2. should be part of `// Sidenav` logic in `DOMContentLoaded` handler

nautobot/project-static/src/js/nautobot.js
```diff
@@ -70,3 +70,20 @@ document.addEventListener('DOMContentLoaded', function() {
...
+const toggleFavorite = (element, event) => {
+    if(event.detail.successful){
+        element.classList.toggle('favorite-item')
+    }
+}
+
+const showEmpty = () => {
+    const favoritesEmpty = document.getElementById('favorites-empty');
+    if(document.querySelectorAll('#sidenav-flyout-favorites > ul > li').length - 1 === 1){
```
norbert-mieczkowski-codilime, line 82: why `length - 1 === 1`? move condition to a more descriptive variable

nautobot/project-static/src/scss/nautobot.scss
```diff
@@ -690,6 +690,7 @@ body {
...
...13.5897 18.7025 13.8477 18.5419 14.1604C18.4 14.4368 18.3244 14.7424 18.3211 15.0531C18.3174 15.4046 18.4446 15.7671 18.699 16.4921V16.4921C19.5294 18.8587 19.9446 20.042 19.673 20.7338C19.4374 21.3339 18.927 21.7836 18.3021 21.9418C17.5816 22.1241 16.4767 21.5717 14.267 20.4668L13.4311 20.0489C12.9064 19.7865 12.644 19.6553 12.3688 19.6037C12.1251 19.558 11.8749 19.558 11.6312 19.6037C11.356 19.6553 11.0937 19.7865 10.5689 20.0489L9.73299 20.4668C7.52325 21.5717 6.41838 22.1241 5.69791 21.9418C5.07298 21.7836 4.56261 21.3339 4.32703 20.7338C4.05543 20.042 4.47062 18.8587 5.30101 16.4921V16.4921C5.55541 15.7671 5.68261 15.4046 5.67891 15.0531C5.67565 14.7424 5.60005 14.4368 5.45811 14.1604C5.29751 13.8477 5.01964 13.5897 4.46388 13.0736L4.11954 12.7539C2.49141 11.242 1.67734 10.4861 1.57953 9.78001C1.49453 9.16639 1.69867 8.54802 2.13229 8.10561C2.63127 7.59653 3.73538 7.47385 5.9436 7.22849L6.26976 7.19225C6.90189 7.12201 7.21795 7.08689 7.49423 6.96694C7.73879 6.86077 7.95965 6.70676 8.14383 6.51399C8.35189 6.29621 8.49411 6.01178 8.77855 5.44291L9.13784 4.72433Z' fill='#{$orange-0}' stroke='#{$orange-0}' stroke-width='1.5' stroke-linecap='round' stroke-linejoin='round'/></svg>";
```
norbert-mieczkowski-codilime, line 693: remove this icon. use `filter-nav-orange-0` mixin on `$sidenav-favorite-icon-svg-filled` icon instead

nautobot/project-static/src/scss/nautobot.scss
```diff
@@ -882,6 +888,37 @@ body {
...
+
+    .sidenav-link-context {
+      color: $nav-gray-3;
+      font-size: $font-size-sm;
+    }
+  }
+
+  .sidenav-favorite-link-breadcrumbs {
+    color: $nav-gray-3;
+  }
```
norbert-mieczkowski-codilime, line 920: not needed

nautobot/core/templates/inc/nav_favorites.html
```diff
@@ -0,0 +1,31 @@
+<ul class="list-unstyled sidenav-favorite-list">
```
norbert-mieczkowski-codilime, line 1: get rid of `sidenav-favorite-list` class, in terms of styles there should be no special treatment for any particular flyout

nautobot/core/templates/inc/nav_favorites.html
norbert-mieczkowski-codilime, line 30: this item should be the *first* in the list, not *last*, because now it breaks segmenting:
image

nautobot/core/templates/inc/nav_favorites.html
```diff
@@ -0,0 +1,31 @@
...
+    {% for item in request.user.navbar_favorites %}
+        {% if item.link and item.name %}
+            <li class="sidenav-link-wrapper" id="sidenav-favorite-link-{{ forloop.counter }}">
+                <a class="sidenav-link" href="{{ item.link }}">
+                    <span class="sidenav-link-text">{{ item.name }}</span>
+                    {% if item.tab_name and item.group_name %}
+                    <span class="sidenav-link-context">{{ item.tab_name }} / {{ item.group_name }}</span>
+                    {% endif %}
+                </a>
```
norbert-mieczkowski-codilime, line 10: less custom css and custom classes = less noise. prefer proper html elements and bootstrap utils over custom styles when possible
```suggestion
                    <span>
                        {{ item.name }}
                        {% if item.tab_name and item.group_name %}
                            <br />
                            <small class="text-secondary">{{ item.tab_name }} / {{ item.group_name }}</small>
                        {% endif %}
                    </span>
```

norbert-mieczkowski-codilime (changes requested):

nautobot/users/views.py
```diff
@@ -195,6 +205,33 @@ def post(self, request):
...
+                    navbar_favorites = sorted(navbar_favorites, key=lambda d: d.get("name", ""))
+                request.user.set_config("navbar_favorites", navbar_favorites, commit=True)
+
+                return render(
+                    request,
+                    "inc/nav_menu.html",
+                    status=HTTPStatus.OK if remove else HTTPStatus.CREATED,
+                )
+
+        referer = request.META.get("HTTP_REFERER")
```
glennmatthews, line 231: What's this for?

nautobot/core/templates/inc/nav_favorites.html
mateusz-lachowski-codilime: Could be, but still I need to show this dynamically when last element is deleted. From my perspective it's better to show/hide instead of define this in JS as well to dynamically add to DOM when list is empty.

nautobot/users/forms.py
mateusz-lachowski-codilime: To simplify deleting from favorite list I'm simply just not passing this there. But from other perspective I can as well send this and fix this form.

nautobot/users/views.py
mateusz-lachowski-codilime: Simply to redirect user back to the previous page if for example HTMX won't load into browser and user click this link.

nautobot/project-static/src/scss/nautobot.scss
mateusz-lachowski-codilime: Why? :) Like in the `&::before` I'm removing the "sgmenting" / "paths" lines from the left side of the links. How I should remove this then?

nautobot/core/templates/inc/nav_favorites.html
mateusz-lachowski-codilime: Not sure why but I wanted to avoid `<br />`. Maybe `vstack` will be better? What do you think? 

In your suggestion even the `<span>` is not needed, right?

nautobot/core/templates/inc/nav_favorites.html
mateusz-lachowski-codilime: Something like that, but even not in the bootstrap itself but in our custom js code that controls the navbar. I didn't want to spend too much time on it, because "partial" swapping is working just fine. The only thing I would like to change would be to use different HTTP methods to add/delete items to "improve" code on backend side.

nautobot/core/templates/inc/nav_favorites.html
mateusz-lachowski-codilime: In the current setup link is not needed. But when I was using the `DELETE` method, the link was passing via query parameter and then having slugified link would be helpful as well.

nautobot/core/templates/inc/nav_favorites.html
glennmatthews: Ah, I thought that we'd be handling that case via the HTMX re-render as well?

mateusz-lachowski-codilime: Actually we do, but when we're clicking from "standard" list. But as second thought I need to see if I can use the `hx-select` here to maybe dynamically swap the whole list during the delete process.

nautobot/core/templates/inc/nav_menu.html
mateusz-lachowski-codilime: Hnm, I can't confirm that. I checked couple of online linters and they all remove such spaces. Also on the JSON documentations they don't have such spaces. They have even less... https://www.w3schools.com/JS/js_json.asp

nautobot/core/templates/inc/nav_favorites.html
mateusz-lachowski-codilime: As agreed on PM I will create `.flat` class to apply for favorite links in favorite flyout to treat "flyoat" as the same but only style links inside properly according to the designs.

nautobot/core/templates/inc/nav_menu.html
mateusz-lachowski-codilime: I've applied Pycharm Code formatting but I bet it's not enough. I would suggest to skip formatting for now until proper linters will be applied to whole code.

nautobot/core/templates/inc/nav_menu.html
norbert-mieczkowski-codilime: it's compliant with JSON RFC and you can leave it like this if you prefer, but mind that just the fact that something is available publicly doesn't make it a standard, in response I could link formatters and guidelines that suggest the opposite but that would be an unnecessary waste of time. it certainly doesn't block my approval anyway

nautobot/project-static/src/scss/nautobot.scss
```diff
@@ -834,6 +833,14 @@ body {
       position: absolute;
       width: $sidenav-link-segment-icon-width;
     }
+
+    &.flat {
```
glennmatthews, line 837: Wondering if we should establish a prefixed naming convention for Nautobot-specific CSS classes (as opposed to those that have meaning in Bootstrap, etc.) both for clarity and for avoiding collisions down the road?

At a minimum, feels like this should maybe be something like `sidenav-flat`?

nautobot/core/templates/inc/nav_favorites.html
```diff
@@ -0,0 +1,31 @@
...
+            <li class="sidenav-link-wrapper" id="sidenav-favorite-link-{{ forloop.counter }}">
+                <a class="sidenav-link flat vstack align-items-start" href="{{ item.link }}">
+                    {{ item.name }}
+                    {% if item.tab_name and item.group_name %}
+                        <small class="text-secondary">{{ item.tab_name }} / {{ item.group_name }}</small>
+                    {% endif %}
+                </a>
+                <button class="sidenav-favorite active"
+                        type="button"
+                        hx-post="{% url "user:navbar_favorites" %}"
```
glennmatthews, line 18:
```suggestion
                        hx-post="{% url 'user:navbar_favorites' %}"
```

nautobot/core/templates/inc/nav_favorites.html
```diff
@@ -0,0 +1,31 @@
+<ul class="list-unstyled">
+    <li class="sidenav-link-group text-center{% if request.user.navbar_favorites %} d-none{% endif %}"
+        style="text-transform: none;" id="favorites-empty">
+        You don't have any favorite pages yet.
+    </li>
+
+    {% for item in request.user.navbar_favorites %}
+        {% if item.link and item.name %}
```
glennmatthews, line 8: What would be the case where this isn't true? Should we have an `else` that renders a warning or error or anything?

nautobot/core/templates/inc/nav_favorites.html
```diff
@@ -0,0 +1,31 @@
...
+            <li class="sidenav-link-wrapper" id="sidenav-favorite-link-{{ forloop.counter }}">
+                <a class="sidenav-link flat vstack align-items-start" href="{{ item.link }}">
+                    {{ item.name }}
+                    {% if item.tab_name and item.group_name %}
```
glennmatthews, line 12: Similarly, does the page rendering look at all weird/wrong if we have items without tab/group names? Do we need an `else` case to render a spacer or placeholder or anything?

nautobot/core/templates/inc/nav_favorites.html
```diff
@@ -0,0 +1,31 @@
...
+                        <small class="text-secondary">{{ item.tab_name }} / {{ item.group_name }}</small>
+                    {% endif %}
+                </a>
+                <button class="sidenav-favorite active"
+                        type="button"
+                        hx-post="{% url "user:navbar_favorites" %}"
+                        hx-vals='{"link": "{{ item.link }}", "name": "{{ item.name }}", "tab_name": "{{ item.tab_name }}", "group_name": "{{ item.group_name }}", "remove": true}'
```
glennmatthews, line 19: Are all of these necessary? I would have thought just `link` and `remove: true` would suffice.

nautobot/users/views.py
glennmatthews: Can/should we use `GetReturnURLMixin` and `self.get_return_url` like many other Nautobot views use?

nautobot/core/templates/inc/nav_favorites.html
mateusz-lachowski-codilime: image

It's more like a future-proof if some other things than favorite from navbar will be added. But I can as well delete it.

norbert-mieczkowski-codilime (approved):
UI looks good to me :+1:

nautobot/core/templates/inc/nav_favorites.html
mateusz-lachowski-codilime: It came from pycharm formatting so I'm not sure if I should change this.

PR #7161, opened 2025-04-14
feature: implement tabs collapsing behavior

### LINKED ISSUE #7142: Detail view collapse tabs to menu on overflow
The tabs on a detail view should be responsive to the width of the window, moving tabs that don't fit into a dropdown on the right.

In the initial work we just need the basic layout and the dynamic size based mechanics of the dropdown. Ultimately, in the future, we want items within the dropdown to be dynamic based on user interaction such that when a user clicks on a tab in the dropdown, that tab comes back up to the main tab bar, and replaces another item.

The "main" (the primary view) and the advanced tab should always be the first two tabs.

norbert-mieczkowski-codilime:
# Closes #7142
# What's Changed

Implement tabs collapsing behavior when they overflow (i.e. wrap). This is achieved with (unfortunately) custom JavaScript code which I tried to make as understandable and readable as possible with comments describing certain steps in the flow.

# Screenshots

image


nautobot/core/templates/generic/object_retrieve.html
```diff
@@ -267,59 +267,46 @@
...
-            var tabLink = document.querySelector(`#tabs a[href="${location.pathname}#${event.state.id}"]`);
-            // Check if the tab link exists
-            if (tabLink) {
-                // click on the tab link
-                tabLink.click();
-            }
-        }
-    }
+    window.addEventListener('popstate', (event) => show_tab(event.state.id));
+
```
norbert-mieczkowski-codilime, line 309: this is mostly refactoring of existing tabs code that 1. makes use of modern web APIs and 2. fixes bug with `main` tab not being active when `tab` wasn't explicitly given in query params

glennmatthews: This still auto-adds a `?tab=main` to the window right? How difficult would it be to get rid of that behavior (i.e. navigating to `/dcim/devices/<uuid>/'` correctly shows the `main` tab, but does *not* auto-rewrite the URL to include the query param)?

nautobot/project-static/src/js/tabs.js
```diff
@@ -0,0 +1,138 @@
...
+
+  // Excessive tabs should be collapsed if tabs clone is wider than the original. Remember - clone does not wrap.
+  const shouldCollapseTabs = () => tabsClone.getBoundingClientRect().width > tabs.getBoundingClientRect().width;
+
+  // If there is no need to collapse tabs, show original tabs (if previously hidden) clean up tabs clone and `return` early.
+  if (!shouldCollapseTabs()) {
+    tabs.classList.remove(...TABS_HIDDEN_CLASSES);
+    tabsParent.removeChild(tabsClone);
+    return;
+  }
```
glennmatthews, line 47: Shouldn't we be able to do this check and early return much earlier in the function? Such as by doing this same check on `tabs` before even creating `tabsClone`?

nautobot/project-static/src/js/tabs.js
```diff
@@ -0,0 +1,138 @@
...
+    const lastTab = dropdown.previousElementSibling; // With dropdown appended to DOM, its previous siblings are tabs.
+    collapsedTabs.unshift(lastTab); // Use `unshift` instead of `push` because this loop is iterating backward in a sense.
+    tabsClone.removeChild(lastTab);
+  }
+
+  /*
+   * Properly convert collapsed tabs to dropdown items:
+   *   1. Remove all `<li>` `nav-item` element classes.
+   *   2. Remove all `<a>` `nav-link` element classes other than `active` and `disabled` to maintain valid tab state.
+   *   3. Add `dropdown-item` class to `<a>` element (formerly `nav-link`).
```
glennmatthews, line 91: Just wondering - is it simpler to manipulate the `collapsedTab` element like this than it would be to create a new element and just copy over the text/contents that are relevant?

glennmatthews (commented):
Really appreciate the readable code and thorough commenting!

nautobot/project-static/src/js/tabs.js
norbert-mieczkowski-codilime: it's true that for default use case it's not. but for example if there was a tab with, let's say, custom text color, I consider manipulating known attributes and their values simpler than the unknown, somewhat like getting from point A to B vs. starting from scratch and getting to B. and I like the idea of recycling :grinning:

nautobot/project-static/src/js/tabs.js
norbert-mieczkowski-codilime: I debated whether it can be simplified further but ultimately stuck to this logic just because it works for any scenarios while other solutions would be heuristic. and when it comes to heuristics the best I can think of is comparing single tab height against its parent height because when parent wraps it should be more or less twice the height of a single tab - that would work but only assuming that *all* tabs are equal height. there are of course many other ways this can be done but I'm always thinking simplicity over performance in the first place

nautobot/project-static/src/js/tabs.js
```diff
@@ -0,0 +1,138 @@
...
+  [...tabsParent.querySelectorAll('.nav.nav-tabs[data-clone="true"]')].forEach((tabsClone) => tabsClone.remove());
+
+  // For safety, do not operate on original tabs element, deep clone tabs to operate on its clone instead.
+  const tabsClone = tabs.cloneNode(true);
+
+  // Remove `id` element attribute from cloned tabs element to avoid potential `id` conflicts.
+  tabsClone.removeAttribute('id');
+
+  // Set `data-clone` attribute to `"true"` on tabs clone element to be able to easily distinguish it from the original.
+  tabsClone.dataset.clone = 'true';
```
mateusz-lachowski-codilime, line 24: Could be just `= true`?

nautobot/project-static/src/js/tabs.js
```diff
@@ -0,0 +1,138 @@
...
+    dropdownToggleIcon,
+    dropdownToggleLabel
+  );
+  const dropdown = createElement(
+    'li',
+    { className: 'dropdown flex-grow-1 mb-n1 text-end' },
+    dropdownToggle,
+    dropdownMenu,
+  );
+  tabsClone.appendChild(dropdown);
```
mateusz-lachowski-codilime, line 74: I'm wondering if having this always in html but hidden wouldn't be a simpler and maybe more readable idea?

nautobot/project-static/src/js/tabs.js
norbert-mieczkowski-codilime: HTML attributes are strings hence `'true'` is closer to the actual representation than `true`

nautobot/project-static/src/js/tabs.js
mateusz-lachowski-codilime: I try to think from perspective of developer who wants to adjust this code and then having this in HTML would be more readable. But yea, no hard feelings in this case :)

mateusz-lachowski-codilime (approved):

nautobot/core/templates/generic/object_retrieve.html
norbert-mieczkowski-codilime: done in 869192c5110bc4cbfb94d7f5a1dc2cdbd9b932f9

glennmatthews (approved):

PR #7164, opened 2025-04-14
Added state transition logic for Approval Workflow related models.

### LINKED ISSUE #7135: State transition logic
With the basic models implemented, we need to add the state transition logic, for processing stages in an approval workflow.

Need logic:
- To identify the current active stage
- Capturing the user input for the stage
- Capturing the final state of the stage when the minimum number of approvals or a rejection is collected
- Transitioning the workflow state to the next stage
- Ensuring exactly one stage is active per workflow instance at a time
- Capturing the final workflow state upon termination of the final stage, or capturing the rejection of any previous stage along the way

---

We do not need views to create or edit a workflow instance, that should all happen internally (Probably makes sense to leave them in for dev testing).

HanlinMiao:
# Closes #7135 
# What's Changed

- [x] To identify the current active stage
- [x] Capturing the user input for the stage (?)
- [x] Capturing the final state of the stage when the minimum number of approvals or a rejection is collected
- [x] Transitioning the workflow state to the next stage
- [x] Ensuring exactly one stage is active per workflow instance at a time
- [x] Capturing the final workflow state upon termination of the final stage, or capturing the rejection of any previous stage along the way
- [x] Unit tests for approval workflow models
- [x] Unit tests for state transition logics

# Screenshots

N/A
# TODO

- [ ] Unit, Integration Tests


nautobot/extras/models/approvals.py
```diff
@@ -138,6 +141,48 @@ def __str__(self):
         """Stringify instance."""
         return f"{self.approval_workflow.name}: {self.object_under_review} ({self.current_state})"
 
+    @property
+    def current_stage(self):
+        """
+        Return the current stage of the workflow.
+        1. Return the stage that is denied if there is a stage that has been denied.
+        2. Return the first stage that is pending, if there is no stage that has been approved or denied.
+        3. Return the stage that is pending that is immediately after the last stage that has been approved.
```
glennmatthews, line 150: These two cases appear to be equivalent and handled by a single check below - might simplify the docstring?

nautobot/extras/models/approvals.py
```diff
@@ -138,6 +141,48 @@ def __str__(self):
...
+        # Check if there is a stage that is denied
+        denied_stage = self.approval_workflow_stage_instances.filter(state=ApprovalWorkflowStateChoices.DENIED)
+        if denied_stage.exists():
+            return denied_stage.first()
+        # Get the pending stage with the lowest weight
+        pending_stages = self.approval_workflow_stage_instances.filter(
+            state=ApprovalWorkflowStateChoices.PENDING
+        ).order_by("approval_workflow_stage__weight")
+        if pending_stages.exists():
+            return pending_stages.first()
```
glennmatthews, line 162: Actually, could this be simplified to just be "first stage that isn't Approved"? Something like:

```py
first_nonapproved_stage = self.approval_workflow_stage_instances.filter(
    state__n=ApprovalWorkflowStateChoices.APPROVED
).first()
return first_nonapproved_stage
```

nautobot/extras/models/approvals.py
```diff
@@ -138,6 +141,48 @@ def __str__(self):
...
+            **kwargs: keyword arguments
+        """
+        # Check if there is one or more denied response for this stage instance
+        denied_stages = self.approval_workflow_stage_instances.filter(state=ApprovalWorkflowStateChoices.DENIED)
+        if denied_stages.exists():
+            self.current_state = ApprovalWorkflowStateChoices.DENIED
+        # Check if all stages are approved
+        approved_stages = self.approval_workflow_stage_instances.filter(state=ApprovalWorkflowStateChoices.APPROVED)
+        if approved_stages.count() == self.approval_workflow.approval_workflow_stages.count():
+            self.current_state = ApprovalWorkflowStateChoices.APPROVED
```
glennmatthews, line 183: Feels wrong to me for this to be in the `save()` on `ApprovalWorkflowInstance` since just saving the workflow instance wouldn't normally result in a state change. I'd think that saving an `ApprovalWorkflowStageInstance` should potentially update the corresponding `ApprovalWorkflowInstance` `current_state` at that time? (I see now that you're accomplishing that by having that model's `save` call `self.approval_workflow_instance.save()` but I wonder if this logic should just be moved over there in the first place?)

(Aside, `current_stage` versus `current_state` feels like it's going to be a fun source of typo-related bugs...)

nautobot/extras/models/approvals.py
```diff
@@ -184,6 +229,40 @@ def __str__(self):
         """Stringify instance."""
         return f"{self.approval_workflow_stage}: {self.state}"
 
+    def save(self, *args, **kwargs):
+        """
+        Override save method:
+        1. set the stage as denied if one user denies it.
+        2. set the stage as approved if min number of eligibl users approve it.
```
glennmatthews, line 236:
```suggestion
        2. set the stage as approved if min number of eligible users approve it.
```

nautobot/extras/models/approvals.py
```diff
@@ -184,6 +229,40 @@ def __str__(self):
...
+        approved_responses = self.approval_workflow_stage_instance_responses.filter(
+            state=ApprovalWorkflowStateChoices.APPROVED
+        )
+        if approved_responses.count() >= self.approval_workflow_stage.min_approvers:
+            self.state = ApprovalWorkflowStateChoices.APPROVED
+
+        # Set the decision date if the state is changed to APPROVED or DENIED
+        decision_made = self.state in [ApprovalWorkflowStateChoices.APPROVED, ApprovalWorkflowStateChoices.DENIED]
+        if decision_made:
+            self.decision_date = date.today()
```
glennmatthews, line 261: This needs a check to make sure that the state actually *changed* to one of the above states, not if it already was there?

nautobot/extras/models/approvals.py
HanlinMiao: Nice!

nautobot/extras/models/approvals.py
lampwins: While you are in here, can we make `decision_date` date and time, instead of just date, please?

nautobot/extras/models/approvals.py
```diff
@@ -128,16 +134,54 @@ class ApprovalWorkflowInstance(PrimaryModel):
...
+    @property
+    def active_stage(self):
+        """
+        Return the current stage of the workflow. The current stage is the first stage that is not yet approved.
+        If all stages are approved, return None.
+        Returns:
+            ApprovalWorkflowStageInstance: The current stage of the workflow.
+        """
+        first_nonapproved_stage = (
+            self.approval_workflow_stage_instances.exclude(state=ApprovalWorkflowStateChoices.APPROVED)
```
mateusz-lachowski-codilime, line 159: I'm not sure if understood the purpose of this method properly so please don't be mad about my questions :) Do we want to return denied stages as well? From `save` method it looks like we gonna denied whole workflow if one stage is denied. So maybe it would more safe to check this here as well? Return None if whole workflow is denied? Then renaming this method to `next_stage` would be more readable for me :)

nautobot/extras/models/approvals.py
```diff
@@ -169,21 +213,79 @@ class ApprovalWorkflowStageInstance(PrimaryModel):
...
+        )
+        if denied_responses.exists():
+            self.state = ApprovalWorkflowStateChoices.DENIED
+
+        # Check if the number of approvers is met
+        approved_responses = self.approval_workflow_stage_instance_responses.filter(
+            state=ApprovalWorkflowStateChoices.APPROVED
+        )
+        if approved_responses.count() >= self.approval_workflow_stage.min_approvers:
+            self.state = ApprovalWorkflowStateChoices.APPROVED
```
mateusz-lachowski-codilime, line 259: I wonder if we should worry about race condition like: user click `Deny` and other click `Approve`. Then here we end up with `Approved` status even if `Deny` should be stronger.

nautobot/extras/models/approvals.py
```diff
@@ -169,21 +213,79 @@ class ApprovalWorkflowStageInstance(PrimaryModel):
...
+            )
+            if denied_stages.exists():
+                approval_workflow_instance.current_state = ApprovalWorkflowStateChoices.DENIED
+            # Check if all stages are approved
+            approved_stages = approval_workflow_instance.approval_workflow_stage_instances.filter(
+                state=ApprovalWorkflowStateChoices.APPROVED
+            )
+            if approved_stages.count() == approval_workflow_instance.approval_workflow.approval_workflow_stages.count():
+                approval_workflow_instance.current_state = ApprovalWorkflowStateChoices.APPROVED
+            approval_workflow_instance.save()
```
mateusz-lachowski-codilime, line 287: Shouldn't we have this logic in some `ApprovalWorkflowInstance` method, for example `refresh_state` / `recheck_state` / `calculate_status` or other - just to keep things separated properly and keep model logic in its own class?

nautobot/extras/models/approvals.py
```diff
@@ -226,3 +328,19 @@ class Meta:
...
+            **kwargs: keyword arguments
+        """
+        super().save(*args, **kwargs)
+        # Call save() on the parent ApprovalWorkflowStageInstance to potentially update its state as well
+        # If this stage is approved or denied and the state has just been updated.
+        if self.state in [
+            ApprovalWorkflowStateChoices.APPROVED,
+            ApprovalWorkflowStateChoices.DENIED,
+        ]:
+            self.approval_workflow_stage_instance.save()
```
mateusz-lachowski-codilime, line 346: I'm not sure if I like approach with empty "safe". It's nothing bad, but maybe, according to the previous comment if you're going to create some method just to check current state it will be standard for approval workflow to call such methods? But if it's standard for nautobot to call `save` for dependent objects just ignore this comment :)

nautobot/extras/tests/test_models.py
```diff
@@ -101,6 +108,203 @@
...
+
+        # Create an Approval Workflow Instance
+        cls.approval_workflow_instance = ApprovalWorkflowInstance.objects.create(
+            approval_workflow=cls.approval_workflow,
+            object_under_review_content_type=job_ct,
+            object_under_review_object_id=job.pk,
+            current_state=ApprovalWorkflowStateChoices.PENDING,
+        )
+        # Create three Approval Workflow Stage Instances
+        cls.approval_workflow_stage_1_instance = ApprovalWorkflowStageInstance.objects.create(
```
mateusz-lachowski-codilime, line 176: How we're gonna create all of this stage instances and responses in target solution?

nautobot/extras/models/approvals.py
HanlinMiao: https://github.com/nautobot/nautobot/pull/7164#discussion_r2042867429 This is the comment where I moved the logic here.

nautobot/extras/models/approvals.py
HanlinMiao: Yes good call, we should at least swap the order of these two

nautobot/extras/tests/test_models.py
HanlinMiao: I believe it is with trigger functions, for example, if a scheduled job is set to have an `ApprovalWorkflow` attached to it. If a user wants to run the scheduled job, an approval workflow instance with the appropriate approval workflow stage instances will be created automatically. This is the [story](https://app.zenhub.com/workspaces/nautobot-core-engineering-64a875e74842500d9e80e23e/issues/gh/nautobot/nautobot/7136) where we will work on that.

nautobot/extras/models/approvals.py
HanlinMiao: Yes, I think the purpose of the function, at least in my mind, is to return the current stage awaiting approval(PENDING) or the stage where the approval workflow instance has been DENIED. So the user is always looking at the most "interesting"/relevant stage, per se

nautobot/extras/models/approvals.py
HanlinMiao: I think the logic here is not overly complex as to warrant another helper method. And yeah, I think the propagation that one `save()` method calling another `save()` is needed here to achieve the desired effect of automatically updating the status of approval workflow stages and approval workflow instances.

mateusz-lachowski-codilime (approved):

nautobot/extras/models/approvals.py
glennmatthews: Do need to at least consider the standard args/kwargs that `save` normally accepts - in particular I think at minimum we'd want to pass through `kwargs["using"]` to the chained `save()` call(s).

glennmatthews (commented):
Haven't looked at the tests in detail yet but have some comments anyway :)

nautobot/extras/filters/__init__.py
```diff
@@ -284,10 +303,19 @@ class ApprovalWorkflowStageInstanceFilterSet(BaseFilterSet):
...
     )
-    decision_date = MultiValueDateFilter()
+    approval_workflow_instance = NaturalKeyOrPKMultipleChoiceFilter(
+        queryset=ApprovalWorkflowInstance.objects.all(),
+        to_field_name="pk",
+    )
+    approval_workflow_stage = NaturalKeyOrPKMultipleChoiceFilter(
+        queryset=ApprovalWorkflowStage.objects.all(),
+        to_field_name="pk",
+    )
```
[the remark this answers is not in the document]
HanlinMiao: I can refactored them to `ModelMultipleChoiceFilters` for now

nautobot/extras/filters/__init__.py
```diff
@@ -305,6 +333,14 @@ class ApprovalWorkflowStageInstanceResponseFilterSet(BaseFilterSet):
             "state": "icontains",
         }
     )
+    approval_workflow_stage_instance = NaturalKeyOrPKMultipleChoiceFilter(
+        queryset=ApprovalWorkflowStageInstance.objects.all(),
+        to_field_name="pk",
```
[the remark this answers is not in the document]
HanlinMiao: I can refactor it to `ModelMultipleChoiceFilters` for now

nautobot/extras/migrations/0123_approval_workflow_models.py
```diff
@@ -95,6 +96,10 @@ class Migration(migrations.Migration):
             ],
             options={
                 "verbose_name": "Approval Workflow Instance",
+                "unique_together": {
+                    ("approval_workflow", "object_under_review_content_type", "object_under_review_object_id")
```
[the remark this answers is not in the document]
HanlinMiao: The second way (reusing the existing one) is what I originally pictured. But it would be interesting to implement the mechanism to clear out the `ApprovalWorkflowInstance`

nautobot/extras/models/approvals.py
glennmatthews: Be careful about passing them all through - I'm not sure that e.g. `force_insert=True` or `force_update=True` would necessarily be appropriate to pass through in all cases. Needs care.

glennmatthews (approved):

PR #7171, opened 2025-04-15
Data Validation Engine rename

glennmatthews:
# What's Changed

The initial motivation here was to work around the fact that DVE's default database table names were dangerously close to MySQL/Dolt's 64-character limit on identifiers. Specifically, we were hitting https://code.djangoproject.com/ticket/33169 when trying to migrate to `next` on a Dolt database. This PR has the side benefit of making the insourced app more closely follow the conventions of other core apps.

- Rename `nautobot.nautobot_data_validation_engine` app to `nautobot.data_validation`
- Set explicit `db_table` strings on the DVE models to shorten the table names even further. (before: `nautobot_data_validation_engine_regularexpressionvalidationrule` etc., now: `data_validation_regexrule` etc.)
- Since we're no longer directly reusing existing the `nautobot_data_validation_engine` app's database tables:
   - Blow away the schema migrations that we'd copied over from the app, and instead regenerate a simple `0001_initial` migration.
   - Add a data migration that copies any existing data from the app tables (if they exist at all) to the new `data_validation` tables.
   - The data migration also uses the (previously added but unused in core) `migrate_content_type_references_to_new_model` helper method to migrate over things like tags, notes, etc. that previously had references to the app data.
      - I discovered that this helper method was missing a number of such cases (metadata, contacts, roles) so I've added those.
- Add missing logic to inject "Data Compliance" tabs into object detail views.

# TODO

- [ ] Unit, Integration Tests


mateusz-lachowski-codilime (approved):
Looking good for now.

mateusz-lachowski-codilime:
I did run tests on my local machine via `inv tests -s -r 8` previously building image without cache and removing pgdata volumes. It took 32 minutes - reported from test runner.

PR #7183, opened 2025-04-17
feature: migrate generic object list view to bootstrap 5

norbert-mieczkowski-codilime:
# What's Changed
Migrate generic object list view to Bootstrap 5:
* change old structures and class names with new
* make the general page layout resemble figma design file
* customize Bootstrap 5 pagination component, it is no our go-to but at least will do for now
* re-style table sortable column headers

*There are still some glitches related to table (most notably border-bottom and border-radius) but they will be addressed in scope of another ticket.*

# Screenshots
image


nautobot/project-static/src/scss/nautobot.scss
```diff
@@ -440,10 +456,43 @@ button {
...
+    position: relative;
+
+    &.orderable {
+      a,
+      button {
+        &:empty {
+          display: none; /* Fix for focusable actions column header. */
+        }
+
+        &:not(.disabled):not(:disabled):not(:empty) {
```
mateusz-lachowski-codilime, line 473: This is scary :D Is there any option to just style disabled/empty items without need to create such "not" notation?

mateusz-lachowski-codilime:
image
Two thoughts just to discuss:
1. Edit selected button looks weird in small version, especially when the table is empty.
2. It would be nice to have some sort of empty view. But as discussed, this is not designed right now.

nautobot/project-static/src/scss/nautobot.scss
norbert-mieczkowski-codilime: changed in 2ecf6c430f01e6029c4d83510fd33537ea1d1127

nautobot/core/templates/inc/paginator.html
```diff
@@ -1,21 +1,46 @@
...
+    {% endif %}
+    <form class="align-items-center d-flex gap-4" method="get">
+        {% for key, value_list in request.GET.lists %}
+            {% if key != 'per_page' %}
+                {% for value in value_list %}
+                    <input type="hidden" name="{{ key }}" value="{{ value }}" />
+                {% endfor %}
+            {% endif %}
+        {% endfor %}
+        <select name="per_page" id="per_page" class="form-select d-inline-block" style="padding-inline-start: 0.4375rem; padding-inline-end: 1.9375rem; width: 4.5rem;"<!-- 7px; 31px; 72px; -->>
```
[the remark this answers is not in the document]
norbert-mieczkowski-codilime: I wasn't quite fond of an idea with creating a new custom class since this is a very specific and in fact unique case of select element. I resued element ID instead cc54d409057625d76f1781ba22474e89f9e81f05

nautobot/core/templates/generic/object_list.html
```diff
@@ -16,74 +16,10 @@
 {% endblock %}
 
 {% block content %}
-<div class="float-end noprint">
+<div class="align-items-center d-flex gap-8 justify-content-end mb-16 noprint">
     {% block buttons %}{% endblock %}
     {% plugin_buttons content_type.model_class 'list' %}
-    {% if table and request.user.is_authenticated and table_config_form %}
```
HanlinMiao, line 22: Why did we move this block?

nautobot/core/templates/inc/paginator.html
```diff
@@ -1,21 +1,46 @@
 {% load helpers %}
 
-<div class="paginator text-end">
+<div class="paginator align-items-center d-flex fs-5 gap-10 justify-content-end mt-16 text-secondary">
```
HanlinMiao, line 3: Seems like the paginator is not padded on the bottom, and it looks a little awkward. Would we be able to fix that?
Screenshot 2025-04-17 at 11 50 48 PM

nautobot/project-static/css/base.css
```diff
@@ -60,18 +60,6 @@ td.min-width {
     width: 1%;
 }
 
-/* Paginator */
-div.paginator {
-    margin-bottom: 20px;
-}
-div.paginator form {
-    margin-bottom: 6px;
-}
```
HanlinMiao, line 69: The previous comment might be because we removed these lines?

HanlinMiao (commented):
Overall it looks great, comparing it with the Figma here, I don't see where the table configuration button is supposed to be.
Screenshot 2025-04-17 at 11 56 29 PM

nautobot/core/templates/inc/paginator.html
norbert-mieczkowski-codilime: good catch! it looks like I missed it from the very beginning of re-styling the app with bootstrap 5. fixed in a08624d30c91a104d959b574d3393a2199d524e0

nautobot/core/templates/generic/object_list.html
norbert-mieczkowski-codilime: no good reason to be honest, I just felt it wasn't the right place for it after I moved saved views and filters buttons to the right-hand side of the viewport. anyway, also to answer your other comment, table config button is designed to be in the table's rightmost column header now, so its placement here is only temporary

mateusz-lachowski-codilime (approved):

HanlinMiao (approved):

glennmatthews (approved):

PR #7184, opened 2025-04-17
Refactored RadioProfileUIViewSet to use UI Components

sirisha-bode:
# What's Changed

# Screenshots
| Create View     |imageimage|imageimage|
| Delete View     |imageimage|imageimage|

# TODO

- [ ] Unit, Integration Tests


nautobot/wireless/views.py
```diff
@@ -37,6 +38,27 @@
...
+                )
+            return helpers.placeholder(value)
+
+        # Custom display for allowed_channel_list
+        if key == "allowed_channel_list":
+            if value:
+                return format_html_join(
+                    " ", '<span class="label label-default">{0}</span>', ((channel,) for channel in value)
+                )
+            return helpers.placeholder(value)
```
Sridevi-NTC, line 57: The two blocks for `channel_width` and `allowed_channel_list` are very similar. Could you simplify the code to avoid repetition?

sirisha-bode: Updated as requested, should be resolve, let me know if I missed anything

Sridevi-NTC: Since these fields are `JSONArrayField`, the isinstance(field_instance, JSONField) check at this [line](https://github.com/nautobot/nautobot/blob/develop/nautobot/core/ui/object_detail.py#L1122) will be triggered. As a result, the value transformations don't work as expected, which is why you added the custom rendering code above is that correct?

sirisha-bode: yes @Sridevi-NTC

itdependsnetworks:
Can you add a overview of accomodations made, I think you would put something like this:

* Converted `supported_data_rates_table` to a proper `ObjectsTablePanel`
* Added transform_value for attributes `tx_power_min`, `tx_power_max`, and `rx_power_min`
* Added custom render_key for `channel_width` and `allowed_channel_list`

nautobot/wireless/views.py
```diff
@@ -37,6 +38,28 @@
...
+    """
+
+    def render_value(self, key, value, context):
+        """
+        Custom render for 'channel_width' and 'allowed_channel_list'.
+        """
+        if key in ["channel_width", "allowed_channel_list"]:
+            if value:
+                # Add 'MHz' only for 'channel_width'
+                suffix = "MHz" if key == "channel_width" else ""
```
itdependsnetworks, line 53: You should be able to do a lambda instead, something like

`"channel_width": lambda v: f"{v.channel_width}MHz"`

sirisha-bode: Updated as requested, should be resolve, let me know if I missed anything

sirisha-bode: Since channel_width and allowed_channel_list are JSONArrayFields, the default rendering falls under the isinstance(field_instance, JSONField) check, which uses render_json() to display the raw JSON. This bypasses any value_transforms, which are not applied to JSON fields by default.

To handle this, we added custom rendering in RadioProfileFieldsPanel.render_value() to display these fields as labeled tags, ensuring they're user-friendly and consistent with the UI design, instead of showing raw JSON.

glennmatthews: We should support value_transforms for all field types, including JSON fields. Seems like a bug in how `ObjectFieldsPanel.render_value` is presently implemented that it doesn't check for the key in `self.value_transforms` before it checks the field type.

sirisha-bode: Updated as requested, should be resolved, let me know if I missed anything

nautobot/wireless/views.py
sirisha-bode: Updated as requested, should be resolved, let me know if I missed anything

nautobot/core/templatetags/helpers.py
```diff
@@ -798,6 +798,19 @@ def render_job_run_link(value):
     return str(value)
 
 
+@library.filter()
+@register.filter()
+def label_list(value, suffix=""):
```
itdependsnetworks, line 803: For core team, just making sure this makes sense as a filter and with this name.

nautobot/core/templatetags/helpers.py
```diff
@@ -798,6 +798,19 @@ def render_job_run_link(value):
...
+    """Render a list of values with optional suffix (like 'MHz') as span labels."""
+    if not value:
+        return ""
+    return format_html_join(
+        " ",
+        '<span class="label label-default">{0}{1}</span>',
+        ((item, f" {suffix}" if suffix else "") for item in value),
```
glennmatthews, line 810: Seems wrong to me to automatically prepend `" "` to the provided suffix - shouldn't the suffix itself have control over that?

nautobot/core/templatetags/helpers.py
```diff
@@ -798,6 +798,19 @@ def render_job_run_link(value):
     return str(value)
 
 
+@library.filter()
+@register.filter()
+def label_list(value, suffix=""):
+    """Render a list of values with optional suffix (like 'MHz') as span labels."""
+    if not value:
+        return ""
```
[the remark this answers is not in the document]
sirisha-bode: Updated as requested, should be resolved, let me know if I missed anything

nautobot/wireless/views.py
```diff
@@ -46,42 +48,42 @@ class RadioProfileUIViewSet(NautobotUIViewSet):
...
 
+class SupportedDataRateUIViewSet(NautobotUIViewSet):
+    queryset = SupportedDataRate.objects.all()
+    filterset_class = SupportedDataRateFilterSet
+    filterset_form_class = SupportedDataRateFilterForm
+    serializer_class = SupportedDataRateSerializer
+    table_class = SupportedDataRateTable
+    form_class = SupportedDataRateForm
+    bulk_update_form_class = SupportedDataRateBulkEditForm
+
```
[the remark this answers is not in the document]
sirisha-bode: Updated as requested, should be resolved, let me know if I missed anything

nautobot/core/ui/object_detail.py
```diff
@@ -1179,6 +1179,12 @@ def render_value(self, key, value, context: Context):
         except FieldDoesNotExist:
             field_instance = None
 
+        if key in self.value_transforms:
+            display = value
+            for transform in self.value_transforms[key]:
+                display = transform(display)
+            return display
```
glennmatthews, line 1186: Possibly add a `.fixed` change fragment, something like:

```
Fixed missing support for `value_transforms` with certain model field types in `ObjectFieldsPanel`.
```

glennmatthews (approved):

PR #7185, opened 2025-04-17
config setting to disable summary field in jobresult list view

### LINKED ISSUE #7178: JobResult table view `summary` field causing extreme load times on later pages
### Environment
* Nautobot version (Docker tag too if applicable): 2.4.5
* Python version: 3.11

### Steps to Reproduce
1. Populate the database with JobResult objects that have a large number of log messages, enough to fill a hundred or so 25 item pages
2. Observe load times on page 1, 20, 40, 80 and 100
  -  In our environment, the load times were 3, 15, 29, 58, and 73 seconds, respectively

### Expected Behavior
Each page in the list view should load in relatively equal time

### Observed Behavior
Near-linear growth in load times the higher the page number that is loaded in the list view, especially when jobs contain a significant number of logs. The query that is run to fill the `summary` field scans the log table and generates a count of each log type per job, which does not scale well when there are a large number of log messages per job result or a large number of job results. In our environment, this is causing gateway timeouts to occur as a result of the long query time to calculate all of the summary field.

See below for the query profile:

Image

brandomando:
# Closes #7178
# What's Changed
* Constance config setting to allow the `summary` field of the `JobResultTable` to be disabled, preventing the expensive database query in the case that many jobs with large numbers of log entries exist.

# Screenshots

image
image
image
image

# TODO

- [x ] Explanation of Change(s)
- [x ] Attached Screenshots, Payload Example
- [x ] Documentation Updates (when adding/changing features)

### DIFF
--- a/nautobot/core/settings.py
+++ b/nautobot/core/settings.py
@@ -128,6 +128,14 @@ INSTALLATION_METRICS_ENABLED = is_truthy(os.getenv("NAUTOBOT_INSTALLATION_METRIC
 if "NAUTOBOT_JOB_CREATE_FILE_MAX_SIZE" in os.environ and os.environ["NAUTOBOT_JOB_CREATE_FILE_MAX_SIZE"] != "":
     JOB_CREATE_FILE_MAX_SIZE = int(os.environ["NAUTOBOT_JOB_CREATE_FILE_MAX_SIZE"])
 
+# Display summary field in JobResult objects in list view? Default is True, but needs to be configurable for performance
+# purposes if there are a lot of JobResult objects with large numbers of log messages.
+if (
+    "NAUTOBOT_JOB_RESULTS_SUMMARY_FIELD_ENABLED" in os.environ
+    and os.environ["NAUTOBOT_JOB_RESULTS_SUMMARY_FIELD_ENABLED"] != ""
+):
+    JOB_RESULTS_SUMMARY_FIELD_ENABLED = is_truthy(os.environ["NAUTOBOT_JOB_RESULTS_SUMMARY_FIELD_ENABLED"])
+
 # The storage backend to use for Job input files and Job output files
 JOB_FILE_IO_STORAGE = os.getenv("NAUTOBOT_JOB_FILE_IO_STORAGE", "db_file_storage.storage.DatabaseFileStorage")
 
@@ -770,6 +778,12 @@ CONSTANCE_CONFIG = {
         ),
         field_type=int,
     ),
+    "JOB_RESULTS_SUMMARY_FIELD_ENABLED": ConstanceConfigItem(
+        default=True,
+        help_text="Enable the 'Summary' column in Job Result list view.\n"
+        "This can be disabled if performance issues are encountered due to job results with large numbers of log messages.",
+        field_type=bool,
+    ),
     "LOCATION_NAME_AS_NATURAL_KEY": ConstanceConfigItem(
         default=False,
         help_text="Location names are not guaranteed globally-unique by Nautobot but in practice they often are. "
@@ -862,7 +876,7 @@ CONSTANCE_CONFIG_FIELDSETS = {
     "Installation Metrics": ["DEPLOYMENT_ID"],
     "Natural Keys": ["DEVICE_NAME_AS_NATURAL_KEY", "LOCATION_NAME_AS_NATURAL_KEY"],
     "Pagination": ["PAGINATE_COUNT", "MAX_PAGE_SIZE", "PER_PAGE_DEFAULTS"],
-    "Performance": ["JOB_CREATE_FILE_MAX_SIZE"],
+    "Performance": ["JOB_CREATE_FILE_MAX_SIZE", "JOB_RESULTS_SUMMARY_FIELD_ENABLED"],
     "Rack Elevation Rendering": [
         "RACK_ELEVATION_DEFAULT_UNIT_HEIGHT",
         "RACK_ELEVATION_DEFAULT_UNIT_WIDTH",
--- a/nautobot/core/settings.yaml
+++ b/nautobot/core/settings.yaml
@@ -1058,6 +1058,7 @@ properties:
       If `True`, the Job Result "summary" column will be available in the Job Results list UI.
       If `False`, the column will not be available.
     environment_variable: "NAUTOBOT_JOB_RESULTS_SUMMARY_FIELD_ENABLED"
+    is_constance_config: true
     type: "boolean"
     version_added: "2.4.8"
   JOBS_ROOT:
--- a/nautobot/core/settings.yaml
+++ b/nautobot/core/settings.yaml
@@ -1052,6 +1052,14 @@ properties:
       "`JOB_CREATE_FILE_MAX_SIZE`": "#job_create_file_max_size"
     type: "string"
     version_added: "2.1.0"
+  JOB_RESULTS_SUMMARY_FIELD_ENABLED:
+    default: true
+    description: >-
+      If `True`, the Job Result "summary" column will be available in the Job Results list UI.
+      If `False`, the column will not be available.
+    environment_variable: "NAUTOBOT_JOB_RESULTS_SUMMARY_FIELD_ENABLED"
+    type: "boolean"
+    version_added: "2.4.8"
   JOBS_ROOT:
     "$ref": "#/definitions/absolute_path"
     default: "~/.nautobot/jobs"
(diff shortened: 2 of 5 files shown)


nautobot/core/settings.py
```diff
@@ -869,7 +883,7 @@
         "RACK_ELEVATION_UNIT_TWO_DIGIT_FORMAT",
     ],
     "Release Checking": ["RELEASE_CHECK_URL", "RELEASE_CHECK_TIMEOUT"],
-    "User Interface": ["SUPPORT_MESSAGE", "NTC_SUPPORT_CONTRACT_EXPIRATION_DATE"],
+    "User Interface": ["SUPPORT_MESSAGE", "NTC_SUPPORT_CONTRACT_EXPIRATION_DATE", "JOB_RESULTS_SUMMARY_FIELD_ENABLED"],
```
glennmatthews, line 886: Maybe move this to the `Performance` section?

nautobot/core/settings.yaml
```diff
@@ -1052,6 +1052,14 @@ properties:
       "`JOB_CREATE_FILE_MAX_SIZE`": "#job_create_file_max_size"
     type: "string"
     version_added: "2.1.0"
+  JOB_RESULTS_SUMMARY_FIELD_ENABLED:
+    default: true
+    description: >-
+      If `True`, the Job results summary field will be displayed in the UI.
+      If `False`, the Job results summary field will not be displayed in the UI.
```
glennmatthews, line 1059:
```suggestion
      If `True`, the Job Result "summary" column will be available in the Job Results list UI.
      If `False`, the column will not be available.
```

nautobot/core/settings.py
```diff
@@ -770,6 +778,12 @@
         ),
         field_type=int,
     ),
+    "JOB_RESULTS_SUMMARY_FIELD_ENABLED": ConstanceConfigItem(
+        default=True,
+        help_text="Enable the 'Summary' field in Job Results.\n"
```
glennmatthews, line 783:
```suggestion
        help_text="Enable the 'Summary' column in Job Result list view.\n"
```

nautobot/extras/tables.py
```diff
@@ -919,28 +920,36 @@ def render_summary(self, record):
...
+                "pk",
+                "date_created",
+                "name",
+                "job_model",
+                "scheduled_job",
+                "duration",
+                "date_done",
+                "user",
+                "status",
+                "summary" if get_settings_or_config("JOB_RESULTS_SUMMARY_FIELD_ENABLED") else None,
```
glennmatthews, line 935: As you mentioned on Slack, this probably won't be dynamic as written since this code is evaluated at import time. Probably disabling this column should be done instead in the `JobResultListView.extra_context()` method (by taking the `"table"` context entry provided by the base view, and calling `table.columns.hide("summary")` if necessary) since that gets evaluated every time the view is rendered.

nautobot/extras/tables.py
```diff
@@ -904,18 +911,20 @@ def render_summary(self, record):
...
-            <label class="label label-info">{}</label>
-            <label class="label label-warning">{}</label>
-            <label class="label label-danger">{}</label>""",
-            record.debug_log_count,
-            record.success_log_count,
-            record.info_log_count,
-            record.warning_log_count,
-            record.error_log_count,
-        )
+        if get_settings_or_config("JOB_RESULTS_SUMMARY_FIELD_ENABLED"):
```
glennmatthews, line 914: Is this check needed? This column should be hidden (and therefore this function shouldn't even be called) if that setting is disabled, right?

(Edit: I don't see that logic on the view diffs at this time, but I think it would be appropriate to add?)

nautobot/extras/views.py
```diff
@@ -2065,48 +2065,90 @@ class JobHookUIViewSet(NautobotUIViewSet):
...
+    else:
+        return JobResult.objects.defer("result").select_related("job_model", "user")
 
 
 class JobResultListView(generic.ObjectListView):
     """
     List JobResults
     """
 
-    queryset = get_annotated_jobresult_queryset()
```
glennmatthews, line 2104: Same fix needed on JobResultBulkDeleteView, perhaps?

nautobot/extras/views.py
```diff
@@ -2065,48 +2065,90 @@ class JobHookUIViewSet(NautobotUIViewSet):
...
     table = tables.JobResultTable
     action_buttons = ()
 
+    def alter_queryset(self, request):
+        """
+        Override to use the annotated queryset
+        """
+        queryset = super().alter_queryset(request)
+
+        if get_settings_or_config("JOB_RESULTS_SUMMARY_FIELD_ENABLED"):
```
glennmatthews, line 2119: Maybe change the signature of `get_annotated_job_result_queryset` so that it (optionally?) takes a `queryset` as input, and just call `queryset = get_annotated_job_result_queryset(queryset)` here, so that we don't need to duplicate the below annotation logic?

glennmatthews:
@brandomando Anything we can do to help get this across the finish line? Seeing some similar issues in other environments.

nautobot/extras/tables.py
brandomando: Just debugged this and recognize it is not being called.

nautobot/core/settings.yaml
```diff
@@ -1052,6 +1052,14 @@ properties:
       "`JOB_CREATE_FILE_MAX_SIZE`": "#job_create_file_max_size"
     type: "string"
     version_added: "2.1.0"
+  JOB_RESULTS_SUMMARY_FIELD_ENABLED:
+    default: true
+    description: >-
+      If `True`, the Job Result "summary" column will be available in the Job Results list UI.
+      If `False`, the column will not be available.
+    environment_variable: "NAUTOBOT_JOB_RESULTS_SUMMARY_FIELD_ENABLED"
```
glennmatthews, line 1060:
```suggestion
    environment_variable: "NAUTOBOT_JOB_RESULTS_SUMMARY_FIELD_ENABLED"
    is_constance_config: true
```

nautobot/extras/tables.py
```diff
@@ -16,7 +16,12 @@
     TagColumn,
     ToggleColumn,
 )
-from nautobot.core.templatetags.helpers import render_boolean, render_json, render_markdown
+from nautobot.core.templatetags.helpers import (
+    render_boolean,
+    render_json,
+    render_markdown,
+)
+from nautobot.core.utils.config import get_settings_or_config
```
glennmatthews, line 24:
```suggestion
```

nautobot/extras/tables.py
```diff
@@ -724,7 +729,9 @@ class JobTable(BaseTable):
     time_limit = tables.Column()
     default_job_queue = tables.Column(linkify=True)
     job_queues_count = LinkedCountColumn(
-        viewname="extras:jobqueue_list", url_params={"jobs": "pk"}, verbose_name="Job Queues"
+        viewname="extras:jobqueue_list",
+        url_params={"jobs": "pk"},
+        verbose_name="Job Queues",
```
glennmatthews, line 734: revert?

nautobot/extras/tables.py
```diff
@@ -1050,7 +1057,11 @@ def render_value(self, record):
             return render_boolean(record.value)
         elif record.metadata_type.data_type == MetadataTypeDataTypeChoices.TYPE_CONTACT_TEAM:
             if record.contact:
-                return format_html('<a href="{}">{}</a>', record.contact.get_absolute_url(), record.contact)
+                return format_html(
+                    '<a href="{}">{}</a>',
+                    record.contact.get_absolute_url(),
+                    record.contact,
+                )
```
glennmatthews, line 1064: revert?

nautobot/extras/views.py
```diff
@@ -2064,56 +2064,67 @@ class JobHookUIViewSet(NautobotUIViewSet):
...
     action_buttons = ()
 
+    def alter_queryset(self, request):
+        """
+        Override to use the annotated queryset
+        """
+        queryset = super().alter_queryset(request)
+
+        return get_annotated_jobresult_queryset(queryset)
+
```
glennmatthews, line 2120: Shouldn't there be something either in the view or in the table to hide the `summary` column altogether if the setting is disabled? I'm thinking maybe in the view's `get_extra_context()`?

glennmatthews: Doesn't that have the problem that it's defined at declaration time, and so won't change if the Constance setting gets changed?

glennmatthews:
Closing this in favor of an alternate approach. Thanks!

PR #7209, opened 2025-04-22
feature: move filter form modal to flyout

norbert-mieczkowski-codilime:
# What's Changed

Move filter form modal to a flyout. I didn't use Bootstrap's [Offcanvas](https://getbootstrap.com/docs/5.3/components/offcanvas/) component because it locks the interaction scope exclusively to the flyout and doesn't allow interacting with the page while it's open. The UX we're targeting here is to be able to interact with filters and the table at once without having to switch between them.

# Screenshots

image


nautobot/core/templatetags/helpers.py
```diff
@@ -844,8 +844,8 @@ def table_config_form(table, table_name=None):
     }
 
 
-@register.inclusion_tag("utilities/templatetags/filter_form_modal.html")
-def filter_form_modal(
+@register.inclusion_tag("utilities/templatetags/filter_form_drawer.html")
+def filter_form_drawer(
```
norbert-mieczkowski-codilime, line 848: I called the component `drawer` since this is a common convetion across other UI component libraries.

nautobot/project-static/src/scss/nautobot.scss
```diff
@@ -1049,6 +1068,49 @@ body {
...
+  position: relative;
+  width: 0;
+
+  &:has(.nb-drawer-open) {
+    @include box-shadow($nb-drawer-open-box-shadow);
+    width: $nb-drawer-open-width;
+  }
+}
+
+.nb-drawer {
```
norbert-mieczkowski-codilime, line 1089: I used `nb` prefix for class name a data attributes as agreed previously

nautobot/core/templates/utilities/templatetags/filter_form_drawer.html
norbert-mieczkowski-codilime, line 1: This file is mostly a copy of what was formerly called `filter_form_drawer.html`. I just changed HTML wrapper from modal to drawer

dsoftait (approved):
LGTM. I noticed there are some errors in tests, but I see that we have a separate task for addressing them in the sprint backlog.

nautobot/project-static/src/js/drawer.js
```diff
@@ -0,0 +1,66 @@
...
+const DRAWER_OPEN_CLASS = 'nb-drawer-open';
+
+/**
+ * Toggle drawer element with an option to force open/close.
+ * @param {HTMLElement} drawer - Drawer HTML element to be toggled.
+ * @param {boolean} [force] - Optionally force opening (`true`) or closing (`false`) regardless of current state.
+ * @returns {void} Do not return any value, modify existing HTML element in-place.
+ */
+const toggleDrawer = (drawer, force) => {
+  drawer?.classList.toggle(DRAWER_OPEN_CLASS, force);
```
glennmatthews, line 11: Should we return early if `drawer` is null? Or are all of the `?` below necessary/appropriate even in such a case?

glennmatthews (approved):

nautobot/project-static/src/js/drawer.js
norbert-mieczkowski-codilime: yeah, I think you're right. I mean, handling nulls and undefineds is certainly not a must here and ideally this should never happen but in general there are no guarantees with arguments as dynamic as DOM nodes. I agree early return may express the intention clearer

glennmatthews: yeah, to be clear - I appreciate you taking the time to put in appropriate logic to handle that case. :-)

norbert-mieczkowski-codilime: done in b3f2dd30617cbbc645d1b771d4ad7d5814ddc501

PR #7210, opened 2025-04-22
Issue 7102

### LINKED ISSUE #7102: Sparse event payloads on deletes.
### Environment
* Nautobot version: 2.4.6
* Python version: 3.11
* Database platform, version: PostgreSQL 15.1
* Middleware(s): N/A

### Steps to Reproduce
1. Install a trivial event producer:
```
import logging

from nautobot.core.events import EventBroker

log = logging.getLogger("nautobot.broker.foo")

class Foo(EventBroker):
    def publish(self, *, topic: str, payload: str) -> None:
        log.info("topic=%s, payload=%s" % (topic, payload))
        
return None
```

2. Create a device, and then an interface on that device.
3. Delete the interface.

### Expected Behavior
The log output to include a comprehensive payload with the entire interface described in `prechange`.

### Observed Behavior
The log output includes a payload that is missing change data:
```
topic=nautobot.delete.dcim.interface, payload={"prechange": null, "postchange": null, "differences": {"removed": null, "added": null}, "context": {"change_context": "Web", "change_context_detail": "dcim:interface_delete", "request_id": "61712b6b-bfb8-4205-9d7c-27d6425a6393", "user_name": "ebusto", "timestamp": "2025-04-01 18:57:18.763477+00:00"}}
```

Note that this but isn't quite deterministic. If you edit an interface, and then immediately delete it, the payload is correct.

ebusto:
# Closes #7102 
# What's Changed
This commit adds a check that ensures at least one change log entry is present for every object updated or deleted, creating change log entries as necessary. This allows the code generating the event payload to have a previous change log entry. Without it, the payload is essentially empty and thus useless.

# TODO
- [X ] Explanation of Change(s)
- [ ] Unit, Integration Tests


HanlinMiao:
> Thanks for the PR! We were actually aiming to tackle this issue in the current sprint but you've beaten us to it. :)
> 
> That said, I'm not _entirely_ sure I understand the approach you've taken or how it addresses the issue. `_handle_changed_object()` and `_handle_deleted_object()` are already responsible for creating ObjectChange records as appropriate, so:
> 
> 1. why does `_handle_saved_object()` need to be added? (I see that it's using `pre_save` versus `_handle_changed_object()`'s `post_save`, but isn't it still redundant?)
> 2. why does `_handle_deleted_object()` need to call `_ensure_changelog_exists()` given that it's already going to create an ObjectChange of its own later in the function?

Currently working on this issue, and I was wondering if we **do** need `_ensure_changelog_exists()` in a `pre_save` signal to create an object change that captures the object's previous state and use that to populate the prechange field that will otherwise be null if a previous objectchange object does not exist?

ebusto:
> Thanks for the PR! We were actually aiming to tackle this issue in the current sprint but you've beaten us to it. :)
> 
> That said, I'm not _entirely_ sure I understand the approach you've taken or how it addresses the issue. `_handle_changed_object()` and `_handle_deleted_object()` are already responsible for creating ObjectChange records as appropriate, so:
> 
> 1. why does `_handle_saved_object()` need to be added? (I see that it's using `pre_save` versus `_handle_changed_object()`'s `post_save`, but isn't it still redundant?)
> 2. why does `_handle_deleted_object()` need to call `_ensure_changelog_exists()` given that it's already going to create an ObjectChange of its own later in the function?

Good questions.

Creating the "previous" object change in `pre_save` allows us to generate the object representation using its current values. If this happened in `post_save`, the updated object would already be saved to the database.

Strictly speaking, a previous object change doesn't need to exist for an object being deleted, but this approach allowed me to scope the proposed fix to one file using a consistent method. As you point out, you could handle the case of a deleted object not having a previous change log entry, but the current implementation doesn't do that correctly.

By the way, my heart would not be broken if you toss this PR and took another approach to fixing this. We're upgrading our internal Nautobot instance from v1 to v2 in a few weeks, so I rather desperately needed a workaround to hold us over.

glennmatthews:
Thanks Eric. We're proceeding with #7213 for this issue, which is definitely ~~stealing~~ _being inspired by_ your approach here. Much appreciated!

PR #7221, opened 2025-04-24
Add ARM64 Docker build to CI for next + 3.13

### LINKED ISSUE #7186: Add ARM images for `nautobot-dev` in the `next` branch
Let's add ARM images for nautobot-dev for next because until we cut a pre-release, if you need to work on an App’s compatibility for 3.0, you’re going to be AMD on ARM locally, which will be less than ideal.
We should NOT do the same for `develop` as it’ll be rarer that you need that.

glennmatthews:
# Closes #7186
# What's Changed

Update CI to build and publish `nautobot-dev` images for ARM64 + Python 3.13 to facilitate App development against `next`. Untested.

# TODO

- n/a Unit, Integration Tests


.github/workflows/ci_integration.yml
```diff
@@ -292,3 +292,18 @@ jobs:
...
+          steps.config.outputs.branch == 'next'
+        uses: "./.github/actions/build-nautobot-image"
+        with:
+          branch: "${{ steps.config.outputs.branch }}"
+          image: "ghcr.io/nautobot/nautobot-dev"
+          platforms: "linux/arm64"
+          push: "true"
+          python-version: "${{ matrix.python-version }}"
+          tag-latest: "${{ steps.config.outputs.tag-latest }}"
+          tag-latest-for-branch: "${{ steps.config.outputs.tag-latest-for-branch }}"
```
glennmatthews, line 307: I wonder actually if this is going to cause any conflicts with the `latest` tag for amd64 builds above?

Maybe we need more of an if/else case instead, e.g.:

- build arm64 alone if *not* latest + next
- else build arm64 *and* amd64 in the same step

HanlinMiao (approved):

PR #7226, opened 2025-04-25
feature: move table config button to table header and form to drawer

norbert-mieczkowski-codilime:
# What's Changed

Move table config button from top buttons row to table header and table config form from modal to drawer.

# Screenshots
image


nautobot/core/templates/generic/object_list.html
```diff
@@ -241,14 +236,18 @@
...
     {% saved_view_modal request.GET.urlencode list_url model request %}
 {% endif %}
 {% dynamic_group_assignment_modal request=request content_type=content_type %}
 {% endblock %}
 
-{% block drawer %}{% filter_form_drawer filter_form dynamic_filter_form model_plural_name=title %}{% endblock drawer %}
+{% block drawer %}
+    {% filter_form_drawer filter_form dynamic_filter_form model_plural_name=title %}
+    {% if table %}
+        {% table_config_form table %}
```
norbert-mieczkowski-codilime, line 248: previously in many cases `table_config_form` would also have an explicit `table_name="ObjectTable"` argument but since it defaults to object python class name I found it unnecessary to make it explicit in any case. I left the function signature unchanged in case there is some app using it but looks like at least the core fares quite well without it

nautobot/core/templates/inc/table.html
```diff
@@ -1,41 +1,66 @@
...
-                        <th {{ column.attrs.th.as_html }}><a href="{% querystring table.prefixed_order_by_field=column.order_by_alias.next %}" data-column-name="{{ column.name }}">{{ column.header }}</a></th>
-                    {% else %}
-                        <th {{ column.attrs.th.as_html }}>{{ column.header }}</th>
+{% with last_column_name=table.columns.items|last|first %}
+    <table{% if table.attrs %} {{ table.attrs.as_html }}{% endif %}>
+        {% if table.show_header %}
+            <thead>
+                <tr>
+                    {% for column in table.columns %}
+                        {% if not forloop.last or last_column_name != 'actions' %}
```
norbert-mieczkowski-codilime, line 10: I wasn't sure if looking for the last column with name `'actions'` wouldn't bee too naive and break some more advanved use cases but I noticed that this behavior is imposed by `core/tables.py` so I don't see a reason why shouldn't I make use of this fact here

nautobot/core/templates/inc/table.html
```diff
@@ -1,41 +1,66 @@
...
+                            <th {{ column.attrs.th.as_html }}>
+                                {% table_config_button table %}
+                            </th>
+                        {% endif %}
+                    {% endfor %}
+                    {% if last_column_name != 'actions' %}
+                        <th class="nb-actionable w-0">
+                            {% table_config_button table %}
+                        </th>
+                    {% endif %}
```
norbert-mieczkowski-codilime, line 26: preferably table config button should be in the actions column but in case table doesn't have it, an additional column to host the config button is required. likewise, when there's no actions column, I'm also adding an empty `td` in each row below

nautobot/core/templatetags/helpers.py
```diff
@@ -744,16 +744,21 @@ def table_config_button(table, table_name=None, extra_classes="", disabled=False
...
+        '    <span class="mdi mdi-cog" aria-hidden="true"></span>'
+        '    <span class="visually-hidden">Configure</span>'
+        "</button>"
     )
-    return format_html(html_template, extra_classes, table_name, 'disabled="disabled"' if disabled else "")
-
-
-@register.simple_tag()
-def table_config_button_small(table, table_name=None):
-    return table_config_button(table, table_name, "btn-xs")
```
norbert-mieczkowski-codilime, line 756: removed this guy because there's only one table config button and it's in table header

nautobot/core/templates/inc/table.html
norbert-mieczkowski-codilime: I was also trying to find a table that wouldn't have a config button but I didn't, so I assumed it's okay to add config button like this to *every* table. in case this reasoning is incorrect, let me know and let's figure out how to handle tables without config buttons

norbert-mieczkowski-codilime: @glennmatthews, I see. I identified that for example Contacts table in detail views is in fact not configurable, as you suggested. I'm thinking about two different approaches now:
1. Imposing all tables configurability with no opt-out mechanism. That would make the mechanism simple and consistent, I'm just not sure if it would be morally right to do :sweat_smile:
2. Utilizing `table.configurable_columns` to check whether table is configurable or not. For now most (if not all) tables define them and it would require an additional effort to make sure `configurable_columns` are empty when config button should not be rendered.

Or would you suggest something else? What do you think?

glennmatthews: It's tricky because a given table class might be used in multiple ways (e.g. a `DeviceTable` could be used in the Device list view, but it could also be used within a SoftwareVersion detail view to list the subset of Devices that have that version attached - I think in the current 2.x UI paradigm the former is configurable and the latter is not.)  It seems like a valid design question as to whether the tables in a detail view *should* be configurable, but if so, that poses its own set of challenges:
- should the same configuration apply to all uses of a given table?
- if there are multiple tables in a single view, can the UI support configurability of each table independently?

nautobot/core/templatetags/helpers.py
glennmatthews: For now I think just adding a `.removed` change fragment so that we don't forget about this specific change will suffice. Let's get a story open to begin building that upgrade guide.

norbert-mieczkowski-codilime: That makes sense, I added `.removed` change fragment 4472e3bb716b2415f8676d954f9a61cc23cddead and added an issue to create a migrate guide to team backlog.

nautobot/core/templates/inc/table.html
norbert-mieczkowski-codilime: Agreed, let's discuss it further in a meeting, right now I could only guess how to proceed with a concrete implementation

lampwins: For now let's stick with the end user UX we have today, where only the list view tables are configurable, and detail view tables are statically configured.

From conversation, this likely means we need a new context var provided by list view to identify those tables.

norbert-mieczkowski-codilime: I think in the end I was able to keep already existing behavior without introducing any additional variables. @glennmatthews could you take a look? 406a88a2f3cf9f0fa398618ac17f3682e50c9858

nautobot/core/templates/utilities/templatetags/table_config_form.html
```diff
@@ -1,28 +1,25 @@
...
-                    <br />
-                    <div class="text-right">
-                        <input type="submit" class="btn btn-primary" value="Save" />
-                        <input type="submit" class="btn btn-danger reset" value="Reset" />
-                    </div>
-                </form>
+            <br />
+            <div class="text-right">
+                <input type="submit" class="btn btn-primary" value="Save" />
+                <input type="submit" class="btn btn-danger reset" value="Reset" />
```
glennmatthews, line 21: Nit - are we actually using XHTML anywhere (I don't think so?) or are the trailing `/` on these tags unnecessary?

nautobot/core/templatetags/helpers.py
```diff
@@ -744,16 +744,21 @@ def table_config_button(table, table_name=None, extra_classes="", disabled=False
...
+        '    data-nb-toggle="drawer"'
+        '    data-nb-target="#{}_config"'
+        "    {}"
+        '    title="Configure table"'
+        '    aria-controls="{}_config"'
+        '    aria-expanded="false"'
+        ">"
+        '    <span class="mdi mdi-cog" aria-hidden="true"></span>'
+        '    <span class="visually-hidden">Configure</span>'
+        "</button>"
```
glennmatthews, line 759: Nit: consider using a single triple-quoted string instead of concatenating a bunch of differently-quoted strings?

nautobot/dcim/templates/dcim/device/wireless.html
```diff
@@ -2,6 +2,11 @@
 {% load helpers %}
 {% load static %}
 
+{% block drawer %}
+    {% table_config_form wireless_networks_table %}
+    {% table_config_form radio_profiles_table %}
```
glennmatthews, line 7: Does this work as expected/intended with two tables?

glennmatthews, line 7: Does this actually work as expected/intended with two tables? Shouldn't be affected by your changes either way, I'm just honestly curious.

glennmatthews (approved):

nautobot/core/templates/utilities/templatetags/table_config_form.html
norbert-mieczkowski-codilime: right, they're not necessary, it's a bias I must've got from writing JSX for most of the time

nautobot/core/templatetags/helpers.py
norbert-mieczkowski-codilime: yeah, that looks cleaner, thanks

nautobot/dcim/templates/dcim/device/wireless.html
norbert-mieczkowski-codilime: oh, it actually **did** work until I removed configure buttons from tables other than those in list views. makes me wonder whether there's a point in purusing a "smart" solution rather than creating a simple `configurable` parameter and then reviewing each table individually... thoughts?

glennmatthews: Explicit is better than implicit in many cases, sure. :-)

changes/7226.added
```diff
@@ -0,0 +1 @@
+Added `configurable` table property to toggle table config button visibility.
```
glennmatthews, line 1: This will be something we want to include in the upgrade guide for apps that don't use the generic view code, I think?

nautobot/core/views/generic.py
```diff
@@ -351,6 +351,7 @@ def get(self, request):
                 saved_view=current_saved_view,
                 user=request.user,
                 hide_hierarchy_ui=hide_hierarchy_ui,
+                configurable=True,
```
glennmatthews, line 354: Probably need a similar addition around line 98 of `nautobot/core/views/renderers.py` for the equivalent NautobotUIViewSet functionality.

glennmatthews (approved):

nautobot/core/views/generic.py
norbert-mieczkowski-codilime: thank you for pointing this out 366274109445e1777e5f2e99edeaa5ce0f750b98

PR #7228, opened 2025-04-25
Fixed broken advanced filters in Nautobot v2.4.7

### LINKED ISSUE #7188: [2.4.7] Advanced Filters broken for some items
### Environment
* Nautobot version: 2.4.7
* Python version: 3.12
* Database platform, version: Postgres 1.6
* Middleware(s):

### Steps to Reproduce
1. Connect to demo.nautobot.com
2. Go to locations -> Filter -> Advanced
3. Select "Tags" ->  Exact 

Value field should allow data entry

Value field grows to double size when "exact" is selected, and as far as I can tell you can't enter data, so it is not possible to create advanced filters.

HanlinMiao:
# Closes #7188 
# What's Changed

We used `context.find("id_slug")` before, and now we are using `context.getElementById`, where context is not guaranteed to be a `document` object every time. In the reported issue, context is actually a `<td></td>` element, thus it does not have an `getElementById` function
# Screenshots

https://github.com/user-attachments/assets/40fa00ed-e41d-4b9e-8e83-f74242cab443

# TODO

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot/core/tests/integration/test_filters.py
+++ b/nautobot/core/tests/integration/test_filters.py
@@ -226,6 +226,9 @@ class ListViewFilterTestCase(SeleniumTestCase):
         """Assert that filters are applied successfully when using the advanced filter."""
         # Go to the location list view
         self.browser.visit(f'{self.live_server_url}{reverse("dcim:location_list")}')
+        # create a new tag
+        tag = Tag.objects.create(name="Tag1")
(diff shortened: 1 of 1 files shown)


glennmatthews:
How much work would it be to add a single integration test case exercising the advanced filter UI?

PR #7237, opened 2025-04-29
Refactored DeviceRedundancyGroup views to use UI component

Sridevi-NTC:
# What's Changed

- Display all model fields on the left side of the detail view using ObjectFieldsPanel
- Display devices across the full width using ObjectTablePanel

# Screenshots

|View |Before|After|
|--|--|--|
|ListView|image|image|
|DetailView|image|image|
|CreateView|imageimage|imageimage|
|Add Controller|Not Available|imageimage|
|Add Device|Not Available|imageimage|
|UpdateView|imageimage|imageimage|
|BulkEditView|imageimage|imageimage|
|DeleteView|imageimage|imageimage|
|BulkDeleteView|imageimage|imageimage|

# TODO

- [ ] Unit, Integration Tests


Sridevi-NTC:
I've added [device_redundancy_group_priority](https://github.com/nautobot/nautobot/pull/7237/files#diff-464408ed1a7c636b182cc6ede2ca06739afc47f5dbcf427f688ad0ddd21c2c4eR225) to the default_columns of DeviceTable, so it will now appear in all models that use DeviceTable (**Manufacturers in image below**). Is it okay to include this in default_columns or should i handle in ObjectsTablePanel?


manufacturers

glennmatthews:
As the column takes up a lot of space, and is irrelevant in many cases, I do *not* think it should be included in `default_columns`.

Sridevi-NTC:
> Isn't that what `include_columns` is for already?

No @glennmatthews It is not working because of we are hiding columns which are not in default_columns in BaseTable
https://github.com/nautobot/nautobot/blob/develop/nautobot/core/tables.py#L134-L138

https://github.com/nautobot/nautobot/blob/develop/nautobot/core/ui/object_detail.py#L882

glennmatthews:
That seems like a bug. Perhaps any `include_columns` on the panel should automatically be passed as `extra_columns` into the table init?

Sridevi-NTC:
> That seems like a bug. Perhaps any `include_columns` on the panel should automatically be passed as `extra_columns` into the table init?

extra_columns must be a list of tuples [example](https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/views.py#L295-L301), We already have device_redundancy_group_priority column in DeviceTable. Do we need to use extra_columns or am i misunderstood the approach?

glennmatthews:
Maybe I'm misunderstanding then. Taking a fresh look at the two code references you provided, I'm not sure the `tables.py` `default_columns` logic is relevant after all as the `object_detail.py` logic is explicitly calling `.show()` on every column either included in `include_columns` or not excluded by `exclude_columns` - so maybe I'm not clear why just providing an explicit `include_columns = [..., "device_redundancy_group_priority", ...]` to the ObjectsTablePanel declaration shouldn't already meet requirements?

itdependsnetworks:
@glennmatthews In regards to `providing an explicit include_columns = [..., "device_redundancy_group_priority", ...] to the ObjectsTablePanel declaration shouldn't already meet requirements?` in theory it should, but seems to be a bug. @Sridevi-NTC is going to take a crack at it, but there it seems like include_columns is either overwritten or comes in too late into the code.

Sridevi-NTC:
Refactored include_columns logic, Kindly review and let me know if any further adjustments are needed

itdependsnetworks:
@Sridevi-NTC can you add a recap of the changes in the original message?

HanlinMiao (approved):

nautobot/core/ui/object_detail.py
```diff
@@ -878,14 +878,18 @@ def get_extra_context(self, context: Context):
                 # to redirect the user back to the correct tab after editing/deleteing an object
                 body_content_table.columns["actions"].column.extra_context["return_url_extra"] = f"?tab={self.tab_id}"
 
-        if self.exclude_columns or self.include_columns:
+        if self.exclude_columns:
             for column in body_content_table.columns:
-                if (self.exclude_columns and column.name in self.exclude_columns) or (
-                    self.include_columns and column.name not in self.include_columns
-                ):
+                if self.exclude_columns and column.name in self.exclude_columns:
```
glennmatthews, line 883:
```suggestion
                if column.name in self.exclude_columns:
```

nautobot/core/ui/object_detail.py
```diff
@@ -878,14 +878,18 @@ def get_extra_context(self, context: Context):
...
                     body_content_table.columns.hide(column.name)
                 else:
                     body_content_table.columns.show(column.name)
+
+        if self.include_columns:
+            for column in self.include_columns:
+                if column in body_content_table.base_columns:
```
glennmatthews, line 890: Why this check? Shouldn't everything in `include_columns` be shown whether it's in `base_columns` or not?

nautobot/core/ui/object_detail.py
```diff
@@ -878,14 +878,18 @@ def get_extra_context(self, context: Context):
...
 
-        if self.exclude_columns or self.include_columns:
+        if self.exclude_columns:
             for column in body_content_table.columns:
-                if (self.exclude_columns and column.name in self.exclude_columns) or (
-                    self.include_columns and column.name not in self.include_columns
-                ):
+                if self.exclude_columns and column.name in self.exclude_columns:
                     body_content_table.columns.hide(column.name)
                 else:
```
glennmatthews, line 885: Should we remove the `else` case, not explicitly showing all other columns?

nautobot/dcim/views.py
```diff
@@ -4135,18 +4135,32 @@ class DeviceRedundancyGroupUIViewSet(NautobotUIViewSet):
...
+            object_detail.ObjectsTablePanel(
+                section=SectionChoices.FULL_WIDTH,
+                weight=100,
+                table_class=tables.ControllerTable,
+                table_attribute="controllers_sorted",
+                related_field_name="controller_device_redundancy_group",
+            ),
+            object_detail.ObjectsTablePanel(
+                section=SectionChoices.FULL_WIDTH,
+                weight=100,
```
glennmatthews, line 4154:
```suggestion
                weight=200,
```

nautobot/core/ui/object_detail.py
Sridevi-NTC: I gave 'test' in the include_columns list and got KeyError, so I added this check

glennmatthews: That seems like something that *should* raise an error - specifying a nonexistent column?

glennmatthews:
> @glennmatthews / @HanlinMiao
> 
> Do we need this [check ](https://github.com/nautobot/nautobot/blob/develop/nautobot/core/ui/object_detail.py#L780-L781)now?

If it works correctly when specifying both (add a test perhaps?), I'm fine with removing it.

nautobot/core/ui/object_detail.py
Sridevi-NTC: Updated as requested, should be resolved, let me know if I missed anything

Sridevi-NTC:
> If it works correctly when specifying both (add a test perhaps?), I'm fine with removing it.

Could you point me to a reference or example for adding this test case?

glennmatthews:
Probably belongs in `nautobot/core/tests/test_ui.py` - just create an ObjectsTablePanel that specifies both `include_columns` and `exclude_columns` and verify that the resulting table's `columns` are as expected?

nautobot/core/ui/object_detail.py
```diff
@@ -878,14 +874,17 @@ def get_extra_context(self, context: Context):
...
-                ):
+                if column.name in self.exclude_columns:
                     body_content_table.columns.hide(column.name)
-                else:
-                    body_content_table.columns.show(column.name)
+
+        if self.include_columns:
+            for column in self.include_columns:
+                if column not in body_content_table.base_columns:
+                    raise ValueError(f"You are specifying a non-existent column `{column}`")
```
HanlinMiao, line 885: Can we add a test for this code path as well?

Sridevi-NTC: Updated as requested, should be resolved, let me know if I missed anything

PR #7239, opened 2025-04-29
build: replace yarn with npm

norbert-mieczkowski-codilime:
# What's Changed

Replace `yarn` with `npm` and introduce simple JavaScript formatting with `prettier`.

* Packages are now installed with `npm ci` command instead of `yarn install --immutable`
* Packages now reside in standard `node_modules` directory instead of yarn exclusive `pnp` files
* Frontend lockfile has been regenerated and `package-lock.json` replaced `yarn.lock`. This means packages with higher compatible versions have been upgraded, a side effect which can't be avoided in such case
* Webpack build is now triggered with `npm run build` instead of `yarn build` command
* Prettier can now format JavaScript source code on demand with `npm run prettier` command


mateusz-lachowski-codilime (approved):
Good work. Only thing I can suggest is to add README file next to package.json just in case somebody is not familiar with standard npm commands.

.gitignore
```diff
@@ -35,5 +35,4 @@ development/factory_dump*.json
 development/kind-kube-config
 development/pod.yaml
 node_modules
-.yarn
```
itdependsnetworks, line 38: I would leave this in, as people are flipping between older and newer versions, it will cause issues.

norbert-mieczkowski-codilime: makes sense, I reverted this change e0ab0014ba3de91756b156f4d083c04a1d08f7b5

HanlinMiao (approved):

nautobot/project-static/.gitignore
```diff
@@ -1,13 +1,138 @@
...
+.cache
+.parcel-cache
+
+# Next.js build output
+.next
+out
+
+# Nuxt.js build / generate output
+.nuxt
+dist
```
glennmatthews, line 94: Don't we want/need to commit `dist` files to the repo for our current distribution model?

glennmatthews (approved):
Agreed that we need a good `README` in the `nautobot/project-static/src/` but that's not necessarily blocking for this PR.

nautobot/project-static/.gitignore
norbert-mieczkowski-codilime: good catch, there's something quirky about ignoring files which have already been commited hence I (wrongfully) didn't bother to take a closer look at this. anyway, respect for reviewing gitignore files, both of you! @glennmatthews @itdependsnetworks

PR #7243, opened 2025-04-30
Refactored DeviceTypeUIViewSet

Pavan-183:
# What's Changed
Overview of accommodations made:

1. Migrated DeviceType views to use DeviceTypeUIViewSet, replacing legacy views with the new Nautobot UI-based 
  implementation.
2. Kept the import view for DeviceType unchanged, as NautobotUIViewSet does not support import functionality by                          
 default.
3. Updated the **test_add_device_page** integration test to align with the new UI rendered by the DeviceTypeUIViewSet. 
 This included changing the page_loaded_confirmation value from **Device Type** to the new heading text **Add a new 
 device type** to ensure the Selenium test waits for the correct element when verifying page load.

# Screenshots

| View               | Before                                                                                                                                         | After                                                                                                                                          |
|--------------------|-------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------|
| **List View**        | image                                                      | image                                                       |
| **Detail View**      | image                                                      | image                                                       |
| **Update View**      | image image image | image image image |
| **Delete View**      | image image | image image |
| **Bulk Edit View**   | image image | image image |
| **Bulk Delete View** | image image | image image |

# TODO

- [ ] Unit, Integration Tests


glennmatthews:
`DeviceTypeImportView` or equivalent functionality is still needed. It's used for importing YAML-formatted data from the device-types-library Git repo.

Pavan-183:
> `DeviceTypeImportView` or equivalent functionality is still needed. It's used for importing YAML-formatted data from the device-types-library Git repo.

Thanks for the clarification @glennmatthews,
In that case, would you recommend restoring the existing DeviceTypeImportView using ObjectImportView, or is there a preferred alternative implementation you’d suggest that better aligns with current best practices?

glennmatthews:
IIRC there isn't a built-in equivalent to `generic.ObjectImportView` in the viewset implementation, so it might be simplest to keep it as-is for that purpose.

Pavan-183:
> IIRC there isn't a built-in equivalent to `generic.ObjectImportView` in the viewset implementation, so it might be simplest to keep it as-is for that purpose.

Sure @glennmatthews , thanks for confirming.

nautobot/dcim/tests/integration/test_fileinputpicker.py
```diff
@@ -50,9 +50,11 @@ def test_add_device_page(self):
         """
         Confirm device type add page input is working correctly.
         """
+        print(self.browser.html)
```
Sridevi-NTC, line 53: remove this

nautobot/dcim/views.py
```diff
@@ -741,20 +741,18 @@ class ManufacturerUIViewSet(NautobotUIViewSet):
...
+    table_class = tables.DeviceTypeTable
     queryset = DeviceType.objects.all()
-    filterset = filters.DeviceTypeFilterSet
-    filterset_form = forms.DeviceTypeFilterForm
-    table = tables.DeviceTypeTable
-    template_name = "dcim/devicetype_list.html"
-
-
-class DeviceTypeView(generic.ObjectView):
-    queryset = DeviceType.objects.select_related("manufacturer").prefetch_related("software_image_files")
```
Sridevi-NTC, line 755: Restore this

itdependsnetworks:
The bulk delete seems to fail in the provided screenshots

nautobot/dcim/views.py
```diff
@@ -829,16 +827,6 @@ def get_extra_context(self, request, instance):
         }
 
 
-class DeviceTypeEditView(generic.ObjectEditView):
-    queryset = DeviceType.objects.all()
-    model_form = forms.DeviceTypeForm
-    template_name = "dcim/devicetype_edit.html"
```
itdependsnetworks, line 835: Shouldn't this either still be referenced, copied to devicetype_update.html, or the file deleted as the viewset verb it would look for would be `update` and not `edit`. I presume the correct action is to reference this template.

Pavan-183:
> The bulk delete seems to fail in the provided screenshots

Yes @itdependsnetworks, the bulk delete failed because the selected DeviceTypes are still linked to existing Devices through foreign key relationships. This prevents them from being deleted. The second screenshot shows more linked devices, so the error message is longer.

nautobot/dcim/views.py
Pavan-183: Updated as requested, should be resolved, let me know if I missed anything

Pavan-183:
and @itdependsnetworks , I have observed that while updating a record instead of showing Update it is showing like create as shown in the screenshot below

image

HanlinMiao (approved):

PR #7246, opened 2025-05-02
Refactored wireless network views to use UI component

Sridevi-NTC:

# What's Changed
- Display all model fields on the left side of the detail view using ObjectFieldsPanel
- Display controller-managed device groups across the full width using ObjectTablePanel

# Screenshots
| Detail View   |image|image|
|CreateView|imageimage|imageimage|
|Delete View |imageimage|imageimage|
| Bulk Delete View|imageimage|imageimage|

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/7248.housekeeping
+++ b/changes/7248.housekeeping
@@ -1,0 +1,1 @@
+Refactored ModuleBayUIViewSet model related UI views to use `UI component framework`.
--- a/nautobot/dcim/templates/dcim/modulebay_retrieve.html
+++ b/nautobot/dcim/templates/dcim/modulebay_retrieve.html
@@ -1,6 +1,5 @@
 {% extends 'generic/object_retrieve.html' %}
 {% load helpers %}
-
 {% block breadcrumbs %}
                 {% if object.parent_device %}
                     <li><a href="{% url 'dcim:device_list' %}">Devices</a></li>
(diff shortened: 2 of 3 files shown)


Sridevi-NTC:
Wireless Network is part of the wireless app (`wireless:wirelessnetwork_list`), and Controller Managed Device Groups are in the dcim app (`dcim:controllermanageddevicegroup_list`)

In ObjectsTablePanel, when trying to get the list URL for **Controller Managed Device Groups**, it's looking for `wireless:controllermanageddevicegroup_list`, which is going to exception clause because a NoReverseMatch error ([Reference](https://github.com/nautobot/nautobot/blob/develop/nautobot/core/ui/object_detail.py#L903-L911))

Can handle this with pagination, or do we need a solution like how `add_button_route` is implemented?

itdependsnetworks:
@HanlinMiao @glennmatthews per the [Reference](https://github.com/nautobot/nautobot/blob/develop/nautobot/core/ui/object_detail.py#L903-L911), there seems to be an implied "app" within object_detail, meaning, if the app is wireless, it expects the reverse url to also be wireless, in this case it is dcim. I see it as 2 options

1. Extend component ui to support explicitly setting the full reverse url path
2. Always use pagination on these

I think 1 is desirable, but let me know.

glennmatthews:
Ah - the table in question is a `ControllerManagedDeviceGroupWirelessNetworkAssignmentTable`, which is a table of `wireless` `ControllerManagedDeviceGroupWirelessNetworkAssignment` records. I don't think we have a list URL for these as it's really just a M2M through table.

glennmatthews:
So probably option 1 is desirable and appropriate here, as I guess we'd need to set the list URL to the ControllerManagedDeviceGroup list URL. Design-wise it might be worth considering whether it'd be best to actually support defining the list URL on the Table definition itself rather than on the ObjectsTablePanel definition.

glennmatthews:
I guess so, yeah - have a default of `None` on `BaseTable` that subclasses can override if they need to.

Sridevi-NTC:
Refactored include_columns logic, Kindly review and let me know if any further adjustments are needed

Sridevi-NTC:
Added support for List URL in BaseTable class, kindly please review and let me know if any changes are required?
image

itdependsnetworks:
@Sridevi-NTC can you capture the change to object_detail in the original comment that was made?

nautobot/core/ui/object_detail.py
```diff
@@ -906,8 +906,18 @@ def get_extra_context(self, context: Context):
...
-            body_content_table_list_url = f"{list_route}?{related_field_name}={obj.pk}"
         except NoReverseMatch:
+            list_url_attr = getattr(self.table_class, "list_url", None)
+            try:
+                if list_url_attr:
+                    list_route = reverse(list_url_attr)
+                else:
+                    list_route = None
+            except NoReverseMatch:
+                list_route = None
```
HanlinMiao, line 917:
```suggestion
            list_url_attr = getattr(self.table_class, "list_url", "")
            try:
                list_route = reverse(list_url_attr)
            except NoReverseMatch:
                list_route = None
```

nautobot/wireless/views.py
```diff
@@ -105,28 +103,34 @@ class WirelessNetworkUIViewSet(NautobotUIViewSet):
     form_class = WirelessNetworkForm
     bulk_update_form_class = WirelessNetworkBulkEditForm
 
+    object_detail_content = object_detail.ObjectDetailContent(
+        panels=(
+            object_detail.ObjectFieldsPanel(
+                section=SectionChoices.LEFT_HALF,
+                weight=100,
```
HanlinMiao, line 110: Do we need to specify `fields="__all"`  here?

Sridevi-NTC: We have default argument in ObjectFieldsPanel https://github.com/nautobot/nautobot/blob/develop/nautobot/core/ui/object_detail.py#L1144 

 I'll make the changes based on your suggestion

Sridevi-NTC: Updated as requested, should be resolved, let me know if I missed anything

HanlinMiao (approved):

nautobot/core/ui/object_detail.py
```diff
@@ -908,8 +908,15 @@ def get_extra_context(self, context: Context):
 
         try:
             list_route = reverse(get_route_for_model(body_content_table_model, "list"))
-            body_content_table_list_url = f"{list_route}?{related_field_name}={obj.pk}"
         except NoReverseMatch:
+            list_url_attr = getattr(self.table_class, "list_url", "")
+            try:
+                list_route = reverse(list_url_attr)
+            except NoReverseMatch:
+                list_route = None
```
glennmatthews, line 916: This feels backwards to me - if a `list_url` is explicitly specified, shouldn't that take precedence over the `get_route_for_model()` case? Should it be something like:

```py
if not list_url := getattr(self.table_class, "list_url", None):
    list_url = get_route_for_model(body_content_table_model, "list")
try:
    list_route = reverse(list_url)
except NoReverseMatch:
    list_route = None
```

nautobot/core/tables.py
```diff
@@ -49,6 +49,7 @@ def __init__(
         hide_hierarchy_ui=False,
         order_by=None,
         data_transform_callback=None,
+        list_url=None,
```
glennmatthews, line 52: Unused?

PR #7255, opened 2025-05-05
Refactored ScheduledJobUIViewSet (approval_queue view incomplete – he…

Pavan-183:
…lp needed)

# What's Changed

Overview of accommodations made:
1. Migrated ScheduledJob views to a single ScheduledJobUIViewSet by composing specific ObjectView mixins for list, detail, delete, bulk delete, changelog, and notes views.
2. Retained the separate ScheduledJobApprovalQueueListView with a custom template override for specialized approval queue rendering.
# Screenshots

| View               | Before                                                                                                                                         | After                                                                                                                                          |
|--------------------|-------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------|
| **List View**        | image                                                      | image                                                       |
| **Detail View**      | image                                                      | image                                                       |
| **Update View**      | Not available                                                                                                                                  | Not available                                                                                                                                   |
| **Delete View**      | image image | image image |
| **Bulk Edit View**   | Not available                                                                                                                                  | Not available                                                                                                                                   |
| **Bulk Delete View** | image image | image image |

# TODO

- [ ] Unit, Integration Tests


Pavan-183:
image

Hi team, I'm facing this issue while accessing the **approval_queue**, I'm not entirely sure if the approach I've taken is correct. Could someone please review the logic and suggest improvements?

nautobot/extras/templates/extras/job_approval_request.html
```diff
@@ -122,7 +122,7 @@ <h1>{{ object.name }}</h1>
                             >
                                 <i class="mdi mdi-close"></i> Deny
                             </button>
-                            <a href="{% url 'extras:scheduledjob_approval_queue_list' %}" class="btn btn-default">
+                            <a href="extras/jobs/approval-queue" class="btn btn-default">
```
glennmatthews, line 125: This change is a bad code smell. We should still be able to use `{% url ... %}` here.

glennmatthews:
Hard to tell what the issue is in your screenshot without the traceback for context.

Pavan-183:
> Hard to tell what the issue is in your screenshot without the traceback for context.

Hi @glennmatthews , I have added the following logic for the `approval_queue` custom view inside `ScheduledJobUIViewSet`:
However, I’m facing a rendering issue ('str' object has no attribute '_meta'), which seems related to the context or how the template is using the view. I’d appreciate your thoughts on whether this usage aligns with the Nautobot UI framework, or if there's a better way to structure this.

    @action(detail=False, url_path="approval-queue")
    def approval_queue(self, request, *args, **kwargs):
        queryset = ScheduledJob.objects.needs_approved()
        table = tables.ScheduledJobApprovalQueueTable(queryset)
        context = {
            "view": self,
            "queryset": queryset,
            "table": table,
            "filter_form": self.filterset_form_class(request.GET or None),
        }
        return render(request, "extras/scheduled_jobs_approval_queue_list.html", context)
        
I have added this logic

nautobot/extras/views.py
```diff
@@ -1983,52 +1983,40 @@ def destroy(self, request, *args, **kwargs):
...
             **super().get_extra_context(request, instance),
         }
 
-
-class ScheduledJobDeleteView(generic.ObjectDeleteView):
-    queryset = ScheduledJob.objects.all()
+    @action(detail=False, url_path="approval-queue")
+    def approval_queue(self, request, *args, **kwargs):
+        queryset = ScheduledJob.objects.needs_approved()
+        table = tables.ScheduledJobApprovalQueueTable(queryset)
```
itdependsnetworks, line 2012: Just taking a guess, but perhaps remove the callable to the function, as logic down the road may be doing that.

HanlinMiao:
The more I look at this, the more I think this might be the correct approach https://github.com/nautobot/nautobot/pull/7255/commits/13e778f34a72f75c3c176359b30311bae852701a

Pavan-183:
> The more I look at this, the more I think this might be the correct approach [13e778f](https://github.com/nautobot/nautobot/commit/13e778f34a72f75c3c176359b30311bae852701a)

Okay @HanlinMiao .

nautobot/extras/urls.py
```diff
@@ -319,17 +320,9 @@
...
-    path("jobs/scheduled-jobs/<uuid:pk>/delete/", views.ScheduledJobDeleteView.as_view(), name="scheduledjob_delete"),
-    path(
-        "jobs/scheduled-jobs/delete/",
-        views.ScheduledJobBulkDeleteView.as_view(),
-        name="scheduledjob_bulk_delete",
-    ),
     path(
         "jobs/scheduled-jobs/approval-queue/",
-        views.ScheduledJobApprovalQueueListView.as_view(),
+        views.ScheduledJobApprovalQueueListView.as_view({"get": "list"}),
```
itdependsnetworks, line 325: I don't fully get this, but will defer to core to understand a bit better.

nautobot/extras/views.py
```diff
@@ -1984,42 +1984,32 @@ def destroy(self, request, *args, **kwargs):
         return super().destroy(request, *args, **kwargs)
 
 
-class ScheduledJobListView(generic.ObjectListView):
+class ScheduledJobUIViewSet(
```
itdependsnetworks, line 1987: Can you mark up which is missing as a comment and why it's not there. Makes it easier to understand down the road.

nautobot/extras/urls.py
glennmatthews: Not sure I understand the need for this either :-)

nautobot/extras/views.py
```diff
@@ -2028,8 +2018,17 @@ def get_extra_context(self, request, instance):
...
+    queryset = ScheduledJob.objects.needs_approved()
+    table_class = tables.ScheduledJobApprovalQueueTable
+    filterset_class = filters.ScheduledJobFilterSet
+    filterset_form_class = forms.ScheduledJobFilterForm
+    action_buttons = ()
+
+    def get_template_name(self):
+        if self.action == "list":
+            return "extras/scheduled_jobs_approval_queue_list.html"
+        return super().get_template_name()
```
glennmatthews, line 2031: When (if ever) will the action *not* be `list`?

nautobot/extras/views.py
glennmatthews: At a glance I *think* the only missing one is `ObjectCreateViewMixin`, which makes sense, but agreed, comments would be helpful.

nautobot/extras/views.py
```diff
@@ -1984,42 +1984,32 @@ def destroy(self, request, *args, **kwargs):
...
-class ScheduledJobListView(generic.ObjectListView):
+class ScheduledJobUIViewSet(
+    mixins.ObjectDetailViewMixin,
+    mixins.ObjectListViewMixin,
+    mixins.ObjectEditViewMixin,
+    mixins.ObjectDestroyViewMixin,
+    mixins.ObjectBulkDestroyViewMixin,
+    mixins.ObjectBulkCreateViewMixin,
+    mixins.ObjectChangeLogViewMixin,
+    mixins.ObjectNotesViewMixin,
```
glennmatthews, line 1995: Actually, I don't think `Edit`, `BulkCreate` apply to ScheduledJobs at present, and as `ScheduledJob` is a `BaseModel` only, I don't think `ChangeLog` or `Notes` apply either?

nautobot/extras/views.py
itdependsnetworks: If that is the case I think it would make sense to auto raise an error, something like 

```python
if self.action != "list":
    raise Something("some descriptive message")
```

nautobot/extras/urls.py
HanlinMiao: So if you use a standalone `ObjectListViewMixin` and not register it with `router.register()`. 
The error below would appear, as we are missing [this](https://github.com/nautobot/nautobot/blob/5f30141bcf28cd43b9a95723b499a7f8394e54c1/nautobot/core/views/routers.py#L9-L16) in the `NautobotUIViewSetRouter`. Similar patterns can be found [here](https://github.com/nautobot/nautobot/blob/5f30141bcf28cd43b9a95723b499a7f8394e54c1/examples/example_app/example_app/urls.py#L40-L44) and [here](https://github.com/nautobot/nautobot/blob/5f30141bcf28cd43b9a95723b499a7f8394e54c1/nautobot/circuits/urls.py#L23-L27).

```
celery_beat-1    |     views.ScheduledJobApprovalQueueListView.as_view(),
celery_beat-1    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
celery_beat-1    |   File "/usr/local/lib/python3.11/site-packages/rest_framework/viewsets.py", line 83, in as_view
celery_beat-1    |     raise TypeError("The `actions` argument must be provided when "
celery_beat-1    | TypeError: The `actions` argument must be provided when calling `.as_view()` on a ViewSet. For example `.as_view({'get': 'list'})`
```

HanlinMiao:
Close due to potential conflict with remaining work with `ApprovalWorkflow` feature related to `ScheduledJob` in `next`

PR #7267, opened 2025-05-07
Hide contacts tab if user does not have permission to view contacts

### LINKED ISSUE #7266: Contacts tab appears even without Contacts view permission
### Environment
* Nautobot version (Docker tag too if applicable): 2.4.7
* Python version: 3.11

### Steps to Reproduce
1. Create a new user
2. Assign enough permissions to view a Location but not Contacts
3. Open the Location with the new limited user profile

### Expected Behavior
The Contacts tab does not appear

### Observed Behavior
The Contacts tab appears. Clicking it, it shows no associated contacts despite the Location having associated contacts

icovada:
# Closes #7266 
# What's Changed

Hide the Contacts tab from overview page of contact-associable models if the user has no Contact View permission.
Before this change, the tab would show but appear empty

### Limitations
This fix only checks of the user has the permission to view Contacts. It will not keep track of additional permission constraints.

If the user is unable to view contacts associated to a specific object due to constraints, the tab will appear but once clicked it will show an empty table, as always.

# Screenshots

View from admin user:
immagine

View from user with the sole Location View permission (no Contact View, nothing else)
immagine

# TODO

- [ ] Unit, Integration Tests


nautobot/core/templates/generic/object_retrieve.html
```diff
@@ -85,11 +85,13 @@ <h1>
             </a>
         </li>
         {% if object.is_contact_associable_model %}
-            <li role="presentation"{% if request.GET.tab == 'contacts' %} class="active"{% endif %}>
-                <a href="{{ object.get_absolute_url }}#contacts" onclick="switch_tab(this.href)" aria-controls="contacts" role="tab" data-toggle="tab">
-                    Contacts
-                </a>
-            </li>
+            {% if perms.extras.view_contact %}
```
glennmatthews, line 88: *Probably* this should either:
- instead check `view_contactassociation` as that's what the tab/table actually contains?
- or else check both `view_contact` and `view_team`?

I'm leaning towards the former but please try it out both ways and see what feels best.

icovada: Good thinking, I didn't check if it was a second relationship.
I just tried and it allows the user to see contacts and their details despite not having permission to actually do so, which would be a nice way to allow users to only see contacts associated to exactly the objects they have access to

nautobot/core/templates/generic/object_retrieve.html
```diff
@@ -85,11 +85,13 @@ <h1>
             </a>
         </li>
         {% if object.is_contact_associable_model %}
-            <li role="presentation"{% if request.GET.tab == 'contacts' %} class="active"{% endif %}>
-                <a href="{{ object.get_absolute_url }}#contacts" onclick="switch_tab(this.href)" aria-controls="contacts" role="tab" data-toggle="tab">
-                    Contacts
-                </a>
-            </li>
+            {% if perms.extras.view_contactassociation %}
```
glennmatthews, line 88: Possibly combine the two conditionals in a single line, i.e. `{% if object.is_contact_associable_model and perms.extras.view_contactassociation %}` similar to the dynamic group case below?

changes/7267.changed
```diff
@@ -0,0 +1 @@
+Hidden contacts tab if user does not have permission to view contacts
```
HanlinMiao, line 1:
```suggestion
Hide the `contacts` tab in object detail view if users do not have permission to view contacts.
```

glennmatthews (approved):

PR #7276, opened 2025-05-08
feature: move table action buttons to dropdown menus

norbert-mieczkowski-codilime:
# What's Changed

Move table action buttons to dropdown menus. This required updating all table action button templates to render as dropdown items rather than flat structure buttons.
# Screenshots

image

- [ ] Unit, Integration Tests


nautobot/core/tables.py
```diff
@@ -402,42 +403,82 @@ class ButtonsColumn(django_tables2.TemplateColumn):
...
-        template_code = self.template_code.format(
+        buttons_to_render = buttons or self.buttons
+        template_code_to_render = (
+            self.dropdown_template_code
+            if (
+                buttons_to_render
+                and ((len(buttons_to_render) > 1) or (len(buttons_to_render) == 1 and len(prepend_template) > 0))
+            )
+            else self.buttons_template_code
+        )
```
norbert-mieczkowski-codilime, line 465: The idea here is to render a flat button when there's only one and otherwise put all buttons in a dropdown menu. I'm not entirely sure about this logic though and I'd like to get some feedback on it. Does it even make sense to do that and not just always render a dropdown menu? Am I not overengineering it? As usual, I'm feeling that "better" here is an enemy of simpler and less prone to errors.

nautobot/core/tables.py
```diff
@@ -402,42 +403,82 @@ class ButtonsColumn(django_tables2.TemplateColumn):
...
+        template_code = template_code_to_render.format(
             app_label=app_label,
             model_name=model._meta.model_name,
             changelog_route=changelog_route,
             edit_route=edit_route,
             delete_route=delete_route,
             pk_field=pk_field,
             buttons=buttons,
+            prepend_template=prepend_template,
```
norbert-mieczkowski-codilime, line 474: in case of dropdown menus, `prepend_template` is not as simple as glueing it to the beginning of template. I put `prepend_template` as standard `format` parameter to accomodate it like this but let me know if there's anything wrong in my reasoning

nautobot/project-static/src/scss/nautobot.scss
```diff
@@ -104,6 +104,7 @@ $dark-border-subtle-dark:      $gray-05-dark;
 
 /* Options */
 $enable-shadows:          true;
+$enable-caret:            false;
```
norbert-mieczkowski-codilime, line 107: firstly, bootstrap dropdown toggle caret is not configurable and also not especially visually pleasing. secondly, there's no simple way to remove it from specific dropdown toggles which looks weird when toggle is an icon, such as the new table actions menu. it gives more control to opt out of it, using icons from app icon library looks much better too

nautobot/core/tables.py
HanlinMiao: IMHO, we should always render a dropdown menu to keep the UX across all tables consistent. The user might have different permissions, which will cause the different table actions columns to be rendered differently. That might be a little confusing IMO.

nautobot/project-static/src/scss/nautobot.scss
```diff
@@ -524,11 +525,45 @@ button {
...
+  .nb-actions {
+    .dropdown {
+      font-size: 0; // Use `font-size: 0;` to remove any visible whitespace.
+
+      .dropdown-toggle {
+        margin-block: -#{$table-cell-padding-y};
+        margin-inline: -#{$table-cell-padding-x};
+        padding-block: $table-cell-padding-y - $border-width;
+        padding-inline: $table-cell-padding-x - $border-width;
+      }
```
HanlinMiao, line 538: Would you mind adding some comments here for my own edification?

HanlinMiao (approved):

nautobot/project-static/src/scss/nautobot.scss
norbert-mieczkowski-codilime: sure 19b83ca590fc614f7d8b86fb1be422cca2934b30

nautobot/core/tables.py
norbert-mieczkowski-codilime: right,  when I looked at it from a broader persperspective, I cannot agree more 84225716ca66bea0927941cf06d72426e9502319

mateusz-lachowski-codilime (approved):

PR #7279, opened 2025-05-09
Fix ExportObjectList job

### LINKED ISSUE #7245: Export Object List ignores saved_view parameter
### Environment
* Nautobot version (Docker tag too if applicable): 2.4.5
* Python version: 3.11
* Database platform, version: Nautobot Cloud
* Middleware(s): Nautobot Cloud

### Steps to Reproduce
1. Create a saved view of an object with filters applied
2. Export the object list while the saved view is active
3.  Note that the filter parameters in the saved view are not observed

### Expected Behavior
I expected the filters from the view to be applied to the export as they are normally when filtering an object list and running the export action.

### Observed Behavior
Only filters applied outside the view are passed to the export job.  The saved_view id is passed to the job but is not using the filters inside the view.

dsoftait:
# Closes #7245 
# What's Changed

When the user tried to export a saved view, the exported list contained all items without any filters applied. This was happening because the `filter_params` calculated in `ExportObjectList` job is based on `query_params`, which in the case of saved view does not contain filter_params. This has now been fixed by the use of `filter_params`, which are stored in the saved view instance, and exporting the saved view works as expected.

# Screenshots
## BEFORE - Filterset params is empty
export object list job before
## AFTER - Filterset params are set
export object list job after

# TODO
- [x] Unit, Integration Tests


nautobot/extras/views.py
```diff
@@ -1809,7 +1809,17 @@ def retrieve(self, request, *args, **kwargs):
...
+
+        config = getattr(instance, "config", {})
+        filter_params = config.get("filter_params")
+
+        if filter_params:
+            for key, value in filter_params.items():
+                query_params[key] = value
+
+        base_url = reverse(instance.view)
+        list_view_url = f"{base_url}?{urlencode(query_params, doseq=True)}"
```
HanlinMiao, line 1822: I am not 100% sure if this fix is correct here. I believe when we first created the saved view feature, there was a strong need for us to keep the Saved view URL consistent in the format of 
`<list_url>?saved_view={pk}` and **not** `<list_url>?saved_view={pk}&{filter_params}`. As the latter can change as the user modifies the saved view configurations, whereas we want the user to be able to bookmark a saved view URL that would not change like the former. What do you think? @glennmatthews

glennmatthews: Haven't looked at the diff in context, but `<list_url>?saved_view=<pk>` definitely should be a sufficient URL for a user to bookmark, share, and access to view the desired content.

dsoftait: I added this here because of how the Export Object List job [works](https://github.com/nautobot/nautobot/blob/develop/nautobot/core/jobs/__init__.py#L182). It uses query_params to compute filter_params. So, if we don't have a saved view, all filters are passed as query parameters, like:

`https://demo.nautobot.com/extras/statuses/?content_types=circuits.circuit&content_types=dcim.cable`
However, if this is a special case where we don't want to apply these filters to the saved view, I can add custom logic to the Export Object List job to read filter_params from the config stored in the saved view instance—if the job is run with a saved view.

Let me know if there's a third option I'm missing.

dsoftait: Discussed during the stand-up and it was decided to modify the job to use `filter_params` from config if it is executed on a saved view.

nautobot/core/jobs/__init__.py
```diff
@@ -178,9 +178,12 @@ def run(self, *, content_type, query_string="", export_format="csv", export_temp
...
         )
-        filter_params = get_filterable_params_from_filter_params(
-            query_params, default_non_filter_params, filterset_class()
-        )
+        if "saved_view" in query_params:
+            filter_params = SavedView.objects.get(pk=query_params["saved_view"]).config.get("filter_params", {})
+        else:
+            filter_params = get_filterable_params_from_filter_params(
+                query_params, default_non_filter_params, filterset_class()
+            )
```
glennmatthews, line 186: It's possible to apply additional filters via query params "in addition to" a Saved View's filters, so this probably shouldn't be an `else` case but something more like: 

```suggestion
        filter_params = {}
        if "saved_view" in query_params:
            filter_params.update(SavedView.objects.get(pk=query_params["saved_view"]).config.get("filter_params", {}))
        filter_params.update(get_filterable_params_from_filter_params(
            query_params, default_non_filter_params, filterset_class()
        )
```

Let's please make sure to augment the testing for this job to cover this scenario.

dsoftait: Probably it is possible, but as I understand if we have `saved_view` then we shouldn't have more query_params so why it's needed ?

Additionally, I covered this case in testing.

mateusz-lachowski-codilime: I did a quick check on the demo. If we will add some filters on top of saved view filters, nautobot will redirect us to the url with both: query params from saved view and new filter added. 

So we have couple possible scenarios:
- Just query params: `/?parent=AMER&status=Active`
- Just saved view: `/?saved_view=314cd0db-9a3b-436a-a68f-7984c0b7c4bb`
- Both - for tests I created a saved view with `parent=AMER` and clicked `status=Active` and was redirected to: `/?saved_view=314cd0db-9a3b-436a-a68f-7984c0b7c4bb&parent=AMER&status=Active`
- Saved view, clicked status filter, removed built-in parent filter: `/?saved_view=314cd0db-9a3b-436a-a68f-7984c0b7c4bb&status=Active`

I think that Daria solution mimics the "frontend" way of working. In approach with merging filters from query params and saved view, we will fail in the fourth case from above list.

nautobot/core/tests/test_jobs.py
```diff
@@ -136,6 +139,56 @@ def test_export_devicetype_to_yaml(self):
         data = yaml.safe_load(yaml_data)
         self.assertEqual(data["manufacturer"], "Cisco")
 
+    def test_export_saved_view_to_csv_without_filters(self):
```
mateusz-lachowski-codilime, line 142: I think that you can merge those two tests into one and use `with subTest` construction to reuse the common code. 

Also it would be nice to add mixed test case for scenario with both saved view and custom filters: `Parsed query_params: {'saved_view': '314cd0db-9a3b-436a-a68f-7984c0b7c4bb', 'parent': 'AMER', 'status': 'Active'}`

nautobot/core/jobs/__init__.py
mateusz-lachowski-codilime: After double checking we should have something like `if "saved_view" in query_params and len(query_params) == 1` because if user will add some more filters, I believe we should focus on those "custom" one.

nautobot/core/jobs/__init__.py
```diff
@@ -150,6 +150,16 @@ class Meta:
         soft_time_limit = 1800
         time_limit = 2000
 
+    def _get_saved_view_filter_params(self, query_params):
```
mateusz-lachowski-codilime, line 153: Now it's a perfect candidate to perform a bunch of unittests with parametrized corner cases :)

nautobot/core/tests/test_jobs.py
```diff
@@ -140,54 +164,57 @@ def test_export_devicetype_to_yaml(self):
...
-        )  # just to be sure that we have more than 1 Location and filter works
-        self.assertEqual(len(rows), 1)  # because saved view has filter with name then should be only 1 record
+        filter_name2 = Status.objects.last().name
+        sv = self._create_saved_view(config={"filter_params": {"name": [filter_name]}})
+        rows = self._run_export_job(f"saved_view={sv.pk}&name={filter_name}&name={filter_name2}")
+        self.assertEqual(len(rows), 2)
         self.assertEqual(rows[0]["name"], filter_name)
+        self.assertEqual(rows[1]["name"], filter_name2)
+
+    def test_export_saved_view_to_csv_with_replaced_filters(self):
```
mateusz-lachowski-codilime, line 194: I would suggest either to add more comments, some line breaks to increase readability of this test or just completely switch to another simpler model or just create required statuses for this test purpose?

mateusz-lachowski-codilime (approved):
Regarding those two new comments to improve testing, I think that fix itself it's done right now :)

nautobot/core/tests/test_jobs.py
dsoftait: Changed it to use Manufacturer model

nautobot/core/jobs/__init__.py
dsoftait: done

HanlinMiao (approved):
Thank you for being thorough

nautobot/core/jobs/__init__.py
```diff
@@ -150,6 +150,16 @@ class Meta:
...
+        """Extract filter params from saved view if applicable."""
+        if "saved_view" in query_params and "all_filters_removed" not in query_params:
+            saved_view_filters = SavedView.objects.get(pk=query_params["saved_view"]).config.get("filter_params", {})
+            if len(query_params) > 1:
+                # Retain only filters also present in query_params
+                saved_view_filters = {key: value for key, value in saved_view_filters.items() if key in query_params}
```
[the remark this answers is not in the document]
dsoftait: The final results will be the same in both situations. In my case, I return all saved view filters that match the `query_params` keys, and I override them outside the method. In your case, you simply clear the dictionary and then add fields based on the `query_params` outside the method.

The only concern I have is that this method is called `_get_saved_view_filter_params`, so it might be a bit confusing if it returns an empty `{}` when there are actually filters matching `query_params` key in saved view instance.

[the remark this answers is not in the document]
glennmatthews: Good points - fine as-is then. :-)

glennmatthews (approved):

PR #7281, opened 2025-05-09
fix: add missing flatpickr styles

norbert-mieczkowski-codilime:
# What's Changed

Add missing flatpickr styles for both light and dark modes.
# Screenshots

| Light | Dark |
| - | - |
| light | dark |


nautobot/project-static/src/scss/nautobot.scss
```diff
@@ -477,7 +478,15 @@ $s2bs5-border-color: $border-color;
 
     /* Focus styles */
     --#{$prefix}focus-ring-color: #{$focus-ring-color-dark};
+
+    @include meta.load-css('flatpickr/dist/themes/dark.css');
+  }
+
+  @include color-mode(light, true) {
+    @include meta.load-css('flatpickr/dist/themes/light.css');
```
norbert-mieczkowski-codilime, line 486: I wanted to import `light.css` in root and only overwrite it with `dark.css` in dark mode but unfortunately they don't define the exact same styles and without constraining `light.css` to light mode, some of its styles still remain in dark mode.

HanlinMiao (approved):

nautobot/project-static/src/scss/nautobot.scss
```diff
@@ -477,7 +478,15 @@ $s2bs5-border-color: $border-color;
...
   }
+} @else {
+  @include meta.load-css('flatpickr/dist/themes/light.css');
```
glennmatthews, line 489: What's the case where this else is needed? Won't we always have `$enable-dark-mode`?

PR #7285, opened 2025-05-12
7250 fix custom fields

dsoftait:
# Closes #7250 
# What's Changed
Based on #4940, I added a JSON class to correct the output of custom fields in GraphQL when using cf_<label>. This update resolves an issue where JSON and MULTISELECT custom fields were being returned as repr() strings instead of proper data structures.
Changes include fixing GraphQL serialization for:
- JSON custom fields
- MULTISELECT custom fields

# Screenshots
## BEFORE JSON
before json field
## AFTER JSON
after json field
## BEFORE MULTISELECT
multiselect before
## AFTER MUTLISELECT
multiselect after
mutliselect after 2

# TODO
- [X] Explanation of Change(s)
- [X] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/core/#creating-changelog-fragments))
- [X] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example App Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

Issue #7250, opened 2025-05-02
GraphQL displays multi select custom fields as a stringified list when using `cf_<label>` syntax


jmcgill298:
### Environment
* Nautobot version (Docker tag too if applicable): 1.6.28
* Python version: N/A
* Database platform, version: mysql
* Middleware(s):

### Steps to Reproduce
1. Create multi-select custom field
2. Restart nautobot service
3. Make GraphQL query using `cf_<label>` syntax

### Expected Behavior
custom field data to be a list of strings

### Observed Behavior
custom field data is a stringified list of strings

NOTE: Using _custom_field_data does return that custom field value as a list of strings

lampwins:
Will be fixed in 1.6, noting that it will be a breaking change and also in 2.x.

lampwins:
See #4627

glennmatthews:
To be clear, this isn't a duplicate of #4627, as that was for `JSON` custom fields (and the linked PR #4940 *only* fixed such fields) and this is for `multiselect` custom fields, so a separate fix is still needed.

jmcgill298:
ah, thanks for clarifying @glennmatthews . I took that as both types of fields were having this issue from the same underlying design, and would both be resolved from the same fix.

jmcgill298:
Thanks @dsoftait

PR #7286, opened 2025-05-12
Fix MULTISELECT custom field in 2.x

dsoftait:
# Relates #7250 
# What's Changed
Based on changes in #7285 I added this change also to 2.x. Here change is only related with multiselect custom fields, because json field was already implemented.
# Screenshots
## BEFORE
multiselect before 2 x
## AFTER
mutliselect 2 x after

# TODO
- [ ] Unit, Integration Tests


changes/7250.fixed
```diff
@@ -0,0 +1 @@
+Fixed MULTISELECT custom field being returned as a `repr()` string when using GraphQL.
```
glennmatthews, line 1: Perhaps:

```suggestion
Fixed MULTISELECT custom field representation in GraphQL to be a JSON array instead of a string.
```

nautobot/core/graphql/schema.py
```diff
@@ -91,6 +91,7 @@
     CustomFieldTypeChoices.TYPE_URL: graphene.String(),
     CustomFieldTypeChoices.TYPE_SELECT: graphene.String(),
     CustomFieldTypeChoices.TYPE_JSON: JSON(),
+    CustomFieldTypeChoices.TYPE_MULTISELECT: JSON(),
```
glennmatthews, line 94: Would `graphene.List(graphene.String())` work here and be more explicit?

glennmatthews (approved):

nautobot/core/graphql/schema.py
dsoftait: Ok, I changed it to `graphene.List(graphene.String)`, because graphene.List(graphene.String()) returned an error
```
List could not have a mounted String() as inner type. Try with List(String)
```

